PR #170, opened 2015-02-20
WIP: list pointer files in the working directory

technoweenie:
Just a proof of concept of a plumbing tool that could eventually turn into a command to download arbitrary pointer files.


technoweenie:
Old!

PR #181, opened 2015-03-26
Configure Travis CI

bkeepers:
This configures Travis CI according to http://docs.travis-ci.com/user/languages/go/

We won't be able to test this until the repository is public, so we can wait to merge this at that time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 # http://docs.travis-ci.com/user/languages/go/
 language: go
 script: script/test
+sudo: required
 notifications:
   email: false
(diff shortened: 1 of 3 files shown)


bkeepers:
@rubyist @technoweenie: `user.name` and `user.email` have to be set for the tests to run. What do you think of setting it in `script/test`?

rubyist:
I'm thinking travis should probably run `script/cibuild`, and it could maybe be set there. We `script/test` as part of day to day development (and as part of the contributing instructions), that probably shouldn't modify your git config.

bkeepers:
:metal: good call.

PR #188, opened 2015-04-08
Specify that `size` property is in bytes

fj:
It's not clear from the documentation that `size` is in bytes, since it's only described as "an integer".


fj:
Currently failing because travisci is not configured correctly, it seems. See https://github.com/github/git-lfs/pull/181.

rubyist:
Nice catch, thanks for the contribution!

PR #191, opened 2015-04-08
Sudo is not necessary for Travis

bkeepers:
The build failures were actually caused by the git config issue that was fixed later in #181.


rubyist:
:+1:

PR #194, opened 2015-04-09
Go gettable

mattn:
Hope to install client with `go get`
```
go get github.com/github/git-lfs/cmd/git-lfs
```


rubyist:
This looks great @mattn, thanks for the contribution!

technoweenie:
Jeez, is that all it took?  Ha.  Thanks :)

PR #207, opened 2015-04-09
Use git submodule

PeterDaveHello:
Does Godeps works with git submodule?
If so, we can make the repo simpler!


technoweenie:
I want to see what happens with [the Go project's proposal for dependencies](https://groups.google.com/forum/?pli=1#!topic/golang-dev/nMWoEAG55v8%5B1-25%5D) first.  We'll probably move to whatever system they end up with.

> In Google’s internal source tree, we vendor (copy) all our dependencies into our source tree and have at most one copy of any given external library. We have the equivalent of only one GOPATH and rewrite our imports to refer to our vendored copy.

It doesn't sound like submodules will be part of it at all.

PR #212, opened 2015-04-10
Spec: clarify that filepath sharding is used in .git/lfs/objects/

rtyley:
I'd only read the LFS spec, and not the code - thanks to @mhagger for setting me straight last night at the Git Birthday Party! The support I'm trying to put into the [BFG](https://rtyley.github.io/bfg-repo-cleaner/) has to put files in the same place as the Git LFS client, obviously - I got it [wrong](https://github.com/rtyley/bfg-repo-cleaner/blob/git-lfs-alpha/bfg-library/src/main/scala/com/madgag/git/bfg/cleaner/LfsBlobConverter.scala#L78) because I didn't check the implementation:

https://github.com/github/git-lfs/blob/a0d389385/lfs/lfs.go#L44
https://github.com/github/git-lfs/blob/a0d389385/commands/smudge_test.go#L54-L55

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -81,6 +81,11 @@ global filters can be set up with `git lfs init`:
 $ git lfs init
 ```
 
+These filters ensure that large files aren't written into the repository proper,
+instead being stored locally at `.git/lfs/objects/{OID-PATH}` (where `{OID-PATH}`
+is a sharded filepath of the form `OID[0:2]/OID[2:4]/OID`), synchronized with
+the remote LFS server as necessary.
+
 The `clean` filter runs as files are added to repositories.  Git sends the
 content of the file being added as STDIN, and expects the content to write
 to Git as STDOUT.
@@ -93,8 +98,6 @@ signature.
   * Move the temp file to `.git/lfs/objects/{OID-PATH}`.
 * Delete the temp file.
 * Write the pointer file to STDOUT.
-
-`{OID-PATH}` is a sharded filepath of the form `OID[0:2]/OID[2:4]/OID`.
 
 Note that the `clean` filter does not push the file to the server.  Use the
 `git lfs sync` command to do that.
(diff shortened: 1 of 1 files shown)


technoweenie:
Good call.  Though it's a little awkward that the `{OID-PATH}` definition is in between uses of it.  Should it be above or below all the uses?

rtyley:
Above is probably best, though I think if it is there it needs to be a little longer, to introduce the concept of the `objects/lfs` file store. I'm happy to update if no objection?

technoweenie:
Sure, go for it.

rtyley:
Have tweaked the placement to the top as suggested.

> a service to other programs that want to interact with LFS at arm's length.

That might be useful to other programs, but I guess my hope is that LFS uses a simple&stable-enough format that it's never necessary to deal with it at arms length :-)

mhagger:
> > a service to other programs that want to interact with LFS at arm's length.
> 
> That might be useful to other programs, but I guess my hope is that LFS uses a simple&stable-enough format that it's never necessary to deal with it at arms length :-)

I agree that it is important for the format to be documented and stable so that other tools can interact with it. I'm just cautious because I've had to deal with many cases of low-grade corruption in vanilla Git repositories caused by buggy alternative Git clients *cough*grit*cough*. So if we can reduce the number of repo-touching apps that have to be written, debugged, and tested, there will be less scope for problems ending up in people's git histories where they are very awkward to repair.

@rtyley, just to be clear, I certainly don't mean to imply anything about your code!

rtyley:
> @rtyley, just to be clear, I certainly don't mean to imply anything about your code!

Ha! :smile: Well, I'll bear that in mind when I'm working out whether or not to make my own code to touch LFS repos!

technoweenie:
I like the spec change here.  Though I definitely agree with @mhagger that tools should use `git-lfs` where possible.

PR #213, opened 2015-04-10
Add gpm's license to its file

### LINKED ISSUE #211: Project probably in violation of gpm's license
This is probably a minor oversight. The project vendors [gpm](http://github.com/pote/gpm) without the appropriate copyright notice.

https://github.com/github/git-lfs/blob/master/script/gpm

GPMs license is MIT:

https://github.com/pote/gpm/blob/master/LICENSE

Which says:

```
The above copyright notice and this permission notice shall be included in
all copies or substantial portions of the Software.
```

While the projects MIT license file can be considered the permission notice, the copyright notice is definitely missing.

This can turn out to be a problem for anyone who wants to use git-lfs and is faced with a license audit.

This can easily be fixed by either having an additional file for third-party licenses or putting the notice in the gpm script itself.

rubyist:
Will fix #211 

/cc @bkeepers

### DIFF
--- a/script/gpm
+++ b/script/gpm
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# Copied from https://github.com/pote/gpm
 #
 # The MIT License
 # 
--- a/script/gpm
+++ b/script/gpm
@@ -1,5 +1,28 @@
 #!/usr/bin/env bash
-
+#
+# The MIT License
+# 
+# Copyright (c) 2013 Pablo Astigarraga
+# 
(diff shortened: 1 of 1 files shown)


bkeepers:
I think that's good, but let's get a :+1: from @talniv.

/cc @benbalter

technoweenie:
Should we add the repository link to the comments too?

technoweenie:
Merging this.  If it's not done correctly, we can move things around later.

PR #215, opened 2015-04-10
add default timeout to http client

Mistobaan:
See [Article](http://biasedbit.com/blog/golang-custom-transports/)

### DIFF
--- a/lfs/http.go
+++ b/lfs/http.go
@@ -6,9 +6,11 @@ import (
 	"fmt"
 	"github.com/rubyist/tracerx"
 	"io"
+	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


technoweenie:
Thanks for this.  I haven't had a chance to verify all this.  It looks good, but I really want to make sure this doesn't force requests to be under 30s.  I don't think it does though.

Might also be a good idea to make this an option that can be set via `git config` too.

technoweenie:
Whoop, almost forgot this.  I bumped some the dial and tls timeouts down since 30s is way long for a CLI tool.  Thanks!

Mistobaan:
Nice :)

PR #219, opened 2015-04-10
Docs for command_track

thekafkaf:
Very simple code, no need for much documentation, documented mostly the functions(godoc is awesome).

### DIFF
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -71,8 +71,8 @@ func trackCommand(cmd *cobra.Command, args []string) {
 
 // mediaPath represents a tracked path
 type mediaPath struct {
-	Path   string // The tracked path
-	Source string // The source attribute file
+	Path   string // the tracked path
+	Source string // the source attribute file
 }
 
 // findAttributeFiles returns the(existing) paths of the attributes files
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -133,6 +133,9 @@ func needsTrailingLinebreak(filename string) bool {
 	}
 
 	defer file.Close()
+
+	// Reading the file in chuncks of 16384 bytes,
+	// to avoid holding the entire file's content in memory
 	buf := make([]byte, 16384)
 	bytesRead := 0
 	for {
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -69,7 +69,7 @@ func trackCommand(cmd *cobra.Command, args []string) {
 	attributesFile.Close()
 }
 
-// mediaPath represents a tracked path
+// mediaPath represents a tracked path.
 type mediaPath struct {
 	Path   string // the tracked path
 	Source string // the source attribute file
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -124,6 +124,8 @@ func findPaths() []mediaPath {
 	return paths
 }
 
+// needsTrailingLinebreak returns 'true' if a file doesn't end in a newline,
+// 'false' otherwise (even if an error occures).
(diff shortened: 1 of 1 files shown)


commands/command_track.go
```diff
@@ -69,11 +71,15 @@ func trackCommand(cmd *cobra.Command, args []string) {
...
 type mediaPath struct {
-	Path   string
-	Source string
+	Path   string // the tracked path
+	Source string // the source attributes file
 }
 
+// findAttributeFiles returns the(existing) paths of the attributes files
+// (./.git/info/attributes, ./.gitattributes).
+// Which is where the tracked paths are configured.
```
technoweenie: How about something like this:

> findAttributeFiles returns the absolute paths to files in the current repository that Git uses to track attributes.

commands/command_track.go
```diff
@@ -96,6 +102,8 @@ func findAttributeFiles() []string {
 	return paths
 }
 
+// findPaths parsers the existing attributes files, and extracts the tracked paths
+// as an array of mediaPaths.
```
technoweenie: > findPaths extracts the paths tracked by Git LFS from the existing Git attributes files.

commands/command_track.go
```diff
@@ -19,6 +19,8 @@ var (
 	}
 )
 
+// trackCommand takes a list of paths as an argument, and adds each path to the default
+// attribtues file (.gitattributes), if it's not already exist.
```
synapsos: This should be:
``// attributes file (.gitattributes), if it not already exists.``

commands/command_track.go
thekafkaf: Used to python, so automatically documented the return type :P

bkeepers:
@thekafkaf thanks for the tweaks.

@technoweenie @rubyist look ok to you?

technoweenie:
Whoop, I missed the extra commits.  There's a few slight grammar changes I'd like to make first.

commands/command_track.go
```diff
@@ -19,6 +19,8 @@ var (
 	}
 )
 
+// trackCommand takes a list of paths as an argument, and adds each path to the default
+// attribtues file (.gitattributes), if it's not already exists.
```
michael-k: attribtues → attributes

Also, not completely true. It sounds like there is only one `.gitattributes` file.  Actually there is one per directory.

commands/command_track.go
```diff
@@ -124,13 +130,18 @@ func findPaths() []mediaPath {
 	return paths
 }
 
+// needsTrailingLinebreak returns 'true' if a file doesn't end in a newline,
+// 'false' otherwise (even if an error occures).
```
ttaylorr: I think this may be a misspelling of "occurs".

commands/command_track.go
```diff
@@ -124,13 +130,18 @@ func findPaths() []mediaPath {
...
 func needsTrailingLinebreak(filename string) bool {
 	file, err := os.Open(filename)
 	if err != nil {
 		return false
 	}
 
 	defer file.Close()
+
+	// Reading the file in chuncks of 16384 bytes,
```
ttaylorr: "chuncks" may be a misspelling of "chunks".

ttaylorr:
Thanks for this, @thekafkaf! Pending the few minor misspellings I noted, along with @michael-k's comment about the `trackCommand` documentation, this should be good to go 👍

ttaylorr:
Going to close this one out for now, if you want to come back to this in the future, please let me know! 😄

PR #220, opened 2015-04-10
Create git-lfs.bash_completion.sh for basic completion support

PeterDaveHello:
Refer to hub's completion (https://github.com/github/hub/blob/master/etc/hub.bash_completion.sh)

### DIFF
--- a/etc/git-lfs.bash_completion.sh
+++ b/etc/git-lfs.bash_completion.sh
@@ -27,6 +27,37 @@ EOF
     if [[ ${COMP_CWORD} -eq 2 ]] ; then
       __gitcomp "$cmds"
       return 0
+    elif [[ ${COMP_CWORD} -eq 3 ]] ; then
+      case "${prev}" in
+        clean|smudge)
+          _filedir -f
+          return
+          ;;
+        logs)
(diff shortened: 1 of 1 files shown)


PeterDaveHello:
This is not a perfect completion, but it'll help, I'll finish it when I have time :)

bkeepers:
Hey @PeterDaveHello, it looks like this is still in progress. Do you think you'll finish it soon, or do you want to close this for now?

technoweenie:
Hey @PeterDaveHello, haven't seen you in a while :)  I'm closing this because we're adding a new `git lfs integrate` command in https://github.com/github/git-lfs/pull/1133. The idea is that `integrate` can support zsh, bash, and other autocompletions. There's even the possibility we can grow beyond just autocompletion, depending on user needs. The really nice thing though is zsh autocomplete code is generated from the go package, reducing duplication, and making it easy for users to update with each new release.

Let me know if you (or anyone else that's reading this) wants to help add bash autocompletion to the upcoming `git lfs integrate` command.

PeterDaveHello:
Oops, sorry that I didn't notice this, that's fine, I may not have enough time to work on the new command, but I'll try it once it's finished :smile:

PR #228, opened 2015-04-13
A bunch of suggested tweaks to the man pages

mhagger:
See the individual commit messages for details.

Another thing that might be wrong but I'm not sure: The manpages for [`git-lfs-ls-files`](https://github.com/github/git-lfs/blob/master/docs/man/git-lfs-ls-files.1.ronn) and [`git-lfs-push`](https://github.com/github/git-lfs/blob/master/docs/man/git-lfs-push.1.ronn) both talk about "refspecs". In Git, a refspec is a specific thing: a mapping between remote references and local references, typically with a form like

    refs/heads/*:refs/remotes/origin/*

The `git-lfs-push` manpage says that the command can be invoked with arguments `<repo> [refspec]`. It is not clear, but I suppose that `<repo>` should be the name of a remote repository, so it should probably be called `<remote>`. Is `refspec` is a refspec as described above?

The `--stdin` documentation for the same command is confusing because it says that it "read[s] the repo and refspec on stdin". But according to `githooks(5)`, the input to a `pre-push` hook is

    <local ref> SP <local sha1> SP <remote ref> SP <remote sha1> LF

There is no "repo" in this input, nor is there a refspec, though I suppose `git-lfs-push` could infer those from the input plus values in the configuration. I'm puzzled.

The other manpage that refers to refspecs is the one for `git-lfs-ls-files`. From the text, it sounds like what is really meant here is not "refspec" but "reference" (or maybe "remote"?) or even "commit" or (to use Git's technical jargon) "commit-ish". But it's not really clear.

Both of these manpages would be improved by examples.


technoweenie:
These changes look great.  I suspect that we meant commitish instead of refspec.

As for `git lfs push`, I think the variable names are just wrong.  It's actually two commands:

1. `git lfs push [--dry-run] <remote> <ref>` (more like `git push`)
2. `git lfs push --stdin` (only called from the pre-push hook)

I think it'd make sense to split the 2nd option into a new `git lfs pre-push` command without the `--stdin` flag.  We can make it an alias for now so that pre-push hooks don't have to be updated _quite yet_.

@rubyist: Thoughts on this?

Also, can you sign up for the CLA? :)

mhagger:
CLA :white_check_mark:

`git lfs push --stdin` is invoked by the installed `pre-push` hook with the usual `pre-push` arguments, `<remote-name> <remote-URL>` (even if those args are currently unused). So I guess that is another reason to make it a separate command.

rubyist:
:+1: I don't mind having a separate command for the pre-push hook.

technoweenie:
I'm merging this as-is.  Splitting the push command can come from a separate PR.

PR #235, opened 2015-04-15
pre-push hook: properly quote args that are passed along

mhagger:
Properly quote the arguments that are passed to `git lfs push --stdin` just in case they contain any shell metacharacters.


technoweenie:
Thanks!  We're still not super sure how we're going to handle pre-push hook updates yet.  Before I work this change into an update though, I'd like to split the push commands like we're discussing in #228.

PR #237, opened 2015-04-15
use remote.{name}.lfsurl to specify an endpoint per remote

### LINKED ISSUE #206: Change `lfs_url` config key to just `lfsurl`
Apparently this is a convention used by Git.

```
```

We could support `lfs_url` until v0.6.x or something too.  Maybe even fix it in the code if it detects it.  Though possibly overkill for this :)

technoweenie:
Fixes #206

```
$ git config remote.master.lfs_url whatever
error: invalid key: remote.master.lfs_url
```

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -52,7 +52,7 @@ can have different Git LFS endpoints for different remotes.  Here is the list
 of rules that Git LFS uses to determine a repository's Git LFS server:
 
 1. The `lfs.url` string.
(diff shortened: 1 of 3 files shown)


technoweenie:
Normally we wouldn't just change config keys abruptly like this.  But I didn't realize that the `git config` command literally rejected them.  The current `lfs_url` key was a very recent rename from just `lfs`.

Issue #239, opened 2015-04-17
docs/spec.md sample out of date


jmarshall:
The docs/spec.md **## The Server** section discusses `lfs.url` and `remote.{name}.lfsurl` config settings, but the adjacent sample Git config file has different names for the settings:

```
[core]
  repositoryformatversion = 0
[lfs]
  endpoint = "https://github.com/github/git-lfs.git/info/lfs"      # url
[remote "origin"]
  url = https://github.com/github/git-lfs
  fetch = +refs/heads/*:refs/remotes/origin/*
  lfs = "https://github.com/github/git-lfs.git/info/lfs"           # lfsurl
```

rubyist:
I've merged these changes in, thanks for the report! :sparkles:

PR #201, opened 2015-04-09
Correct lfs url git config sample.

RyanMcG:
`lfs.endpoint` and `remote.NAME.lfs` are not used, `lfs.url` and `remote.NAME.lfs_url` are.

See [the spec](https://github.com/github/git-lfs/blob/master/docs/spec.md#the-server).

PR #240, opened 2015-04-17
Update spec to reflect lfs.url and lfsurl

rubyist:
Corrects some configuration documentation, reported in #239


rubyist:
Oh it looks like this was also reported in https://github.com/github/git-lfs/pull/201 but CLA issues prevent it from being merged. I'll merge this one and close out the other #201.

PR #246, opened 2015-04-19
stricter pointer spec

technoweenie:
Tightens up the pointer spec according from @swisspol in https://github.com/github/git-lfs/issues/233 and https://gist.github.com/swisspol/b9856fcca81029efb960.

* [x] implement git pointer commands

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -9,6 +9,27 @@ for other tools.
 The core Git LFS idea is that instead of writing large blobs to a Git repository,
 only a pointer file is written.
 
+* Pointer files are text files which MUST contain only UTF-8 characters.
+* Each line MUST be of the format `{key} {value}\n` (trailing unix newline).
+* Only a single space character between `{key}` and `{value}`.
+* Keys MUST only use the characters `[a-z] [0-9] . -`.  
+* The first key is _always_ `version`.
+* Lines of key/value pairs MUST be sorted alphabetically in ascending order
+(with the exception of `version`, which is always first).
+* Values MUST NOT contain return or newline characters.
+* Pointer files SHOULD not have the executable bit set when checked-in in Git.
+
+The required keys are:
+
+* `version` is a URL that identifies the pointer file spec.  Parsers MUST use
+simple string comparison on the version, without any URL parsing or
+normalization.  It is case sensitive, and %-encoding is discouraged.
+* `oid` tracks the unique object id for the file, prefixed by its hashing
+method.  Currently, only `sha256` is supported.
+* `size` is in bytes.
+
+Example of a v1 text pointer:
+
 ```
 version https://git-lfs.github.com/spec/v1
 oid sha256:4d7a214614ab2935c943f9e0ff69d22eadbb8f32b1258daaa5e2ca24d17e2393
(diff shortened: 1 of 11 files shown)


swisspol:
Awesome, thanks!

docs/spec.md
```diff
@@ -9,6 +9,27 @@ for other tools.
...
 
+* Pointer files are text files which MUST contain only UTF-8 characters.
+* Each line MUST be of the format `{key} {value}\n` (trailing unix newline).
+* Only a single space character between `{key}` and `{value}`.
+* Keys MUST only use the characters `[a-z] [0-9] . -`.  
+* The first key is _always_ `version`.
+* Lines of key/value pairs MUST be sorted alphabetically in ascending order
+(with the exception of `version`, which is always first).
+* Values MUST NOT contain return or newline characters.
+* Pointer files SHOULD not have the executable bit set when checked-in in Git.
```
rubyist: Capsify "NOT" here

rubyist:
I think this is a good approach :+1:

docs/spec.md
```diff
@@ -9,6 +9,27 @@ for other tools.
...
 
+* Pointer files are text files which MUST contain only UTF-8 characters.
+* Each line MUST be of the format `{key} {value}\n` (trailing unix newline).
+* Only a single space character between `{key}` and `{value}`.
+* Keys MUST only use the characters `[a-z] [0-9] . -`.  
+* The first key is _always_ `version`.
+* Lines of key/value pairs MUST be sorted alphabetically in ascending order
+(with the exception of `version`, which is always first).
+* Values MUST NOT contain return or newline characters.
+* Pointer files SHOULD NOT have the executable bit set when checked-in in Git.
```
davideg: s/checked-in in/checked into/

docs/spec.md
```diff
@@ -9,6 +9,27 @@ for other tools.
...
+
+The required keys are:
+
+* `version` is a URL that identifies the pointer file spec.  Parsers MUST use
+simple string comparison on the version, without any URL parsing or
+normalization.  It is case sensitive, and %-encoding is discouraged.
+* `oid` tracks the unique object id for the file, prefixed by its hashing
```
davideg: Do you want to be explicit about the prefix being followed by a ":" or is that clear from the example?

docs/spec.md
```diff
@@ -16,16 +37,15 @@ size 12345
...
-* Separate the property from its value with a single space.
-* Oid has a "sha256:" prefix.  No other hashing methods are currently supported
-for Git LFS oids.
-* Size is in bytes.
+* Run `git lfs pointer` to generate a pointer file for the given local file.
+* Run `git lfs pointer` to compare the blob OID of the generated pointer files.
+* Tools that parse and regenerate pointer files MUST preserve keys that they
+don't know or care about.
 
 Note: Earlier versions only contained the OID, with a `# comment` above it.
```
davideg: This wasn't touched by the current PR but including this part about earlier versions and some out-of-date code for parsing older pointer files has been bothering me :stuck_out_tongue_winking_eye: 

Any reason not to :fire: this section through current line 66?

docs/spec.md
swisspol: The spec should be 100% self contained: examples are not the spec.

I agree the column needs to be mentioned here.

docs/spec.md
technoweenie: Sure, then Git LFS tools don't know how to parse older pointer files.  I was thinking of removing this section though, because the chance of anyone besides me seeing these old blobs is pretty minimal.

swisspol: If this spec is supposed to indicate how to parse older versions then it
would need one section per version, and each section should be self
contained explaining exactly what version it is, how to identify it from
the header, and of course what the format is, and at least one example file
pointer.

It would be the best way to let third parties write correct tools.

technoweenie: I removed support for the really old pointer files.  Nothing but internal test repos ever used those.  But I did add an example of the pre-release pointer.  Again, the chances of any tool encountering this is pretty low...

docs/spec.md
technoweenie: I added a string format example here: `{hash-method}:{hash}`

technoweenie:
I impemented a `git lfs pointer` command.  You can check it out with `script/run pointer ...` if you have the repository cloned locally.  Or, check out [the sample `pointer` output in the spec](https://github.com/github/git-lfs/pull/246/files#diff-6e196c98aaadaa54d9ba9bbe92a13320R58).

Issue #242, opened 2015-04-18
Better support for SSH repositories


ddanier:
Currently the only way to get SSH setups running seems to be manually specifiing the URL to the git lfs server in the configuration. This is not necessary for HTTP setups as the URL gets automatically build (append /info/lfs). Of course such a mechanism is not possible for SSH, but:

Currently git lfs runs git-lfs-authenticate before any call to the HTTP storage backend. This is done to get the necessary authentication headers to access the storage backend. I think the response should be extended to include the URL of the storage backend, too. Example:

```
# remote: git@github.com:user/repo.git
$ ssh git@github.com git-lfs-authenticate user/repo.git download {oid}
{
  "header": {
    "Authorization": "Basic ..."
  },
  "url": "https://github.com/user/repo/info/lfs"
}
```

This could also be used to add some expiration date for the authentication (see the note about caching in docs/api.md).

Background: We are using gitolite over SSH for over 300 repositories and are thinking about switching big files over to git-lfs. This issue is currently is the biggest show-stopper, as it is nearly impossible (insane amout of work) to configure the correct URLs on every of the developer and designer computers.

technoweenie:
@ddanier Hey, thanks for the feedback.  Having gitolite support this would be awesome.

> I think the response should be extended to include the URL of the storage backend, too.

I think this is a good idea.  Though, it should be optional.  If unspecified, then the Git LFS client builds the url from the ssh remote. 

> This could also be used to add some expiration date for the authentication

If you look at the command output on GitHub, you'll notice we already do this.  However, I wasn't comfortable putting this into the spec without some solid credential caching code.  We would have to solve that before add this to the SSH remote output.  

I also thought about sending some kind of "scope" property.  Depending on your backend implementation, the given token could be scoped to the repository (`user/repo.git`), the operation (download or upload), and the OID.  Course, if you actually want the caching to reduce the number of SSH calls, the backend may want to reduce the scope to just the repo, or repo and operation only.

ddanier:
> I also thought about sending some kind of "scope" property. Depending on your backend implementation, the given token could be scoped to the repository (user/repo.git), the operation (download or upload), and the OID. Course, if you actually want the caching to reduce the number of SSH calls, the backend may want to reduce the scope to just the repo, or repo and operation only.

[...]

PR #247, opened 2015-04-19
add the optional HREF key to the SSH output

technoweenie:
/cc #242 @ddanier

I'm keeping the issue open until we can figure out how to securely cache the tokens.

* [x] Update client to look for the `href` property


rubyist:
Very reasonable :+1:

PR #248, opened 2015-04-19
Auth changes

### LINKED ISSUE #243: git-lfs asks for user/password although storage backend needs no authentication
I'm currently writing some proof of concept storage backend for git-lfs following the API (see docs/api.md). Currently I have implemented no authentication for accessing the files. This may not be a valid production case (probably besides having no auth for some internal network), but draw attention to git-lfs always asking for a user/password. As far as HTTP works in other cases/clients (like browsers) authentication should only be needed after a HTTP 401 response.

This should be true for subsequent API call, too. As the API will return the necessary authentication headers in these cases the client should not run into some kind of new authentication anyways (see "header" API response, should include basic auth). It even may not be necessary to cache the authentication because of this.

In any case: git-lfs should only ask for authentication if really necessary.

technoweenie:
This tweaks Git LFS's behavior regarding authentication.

## SSH auth

1. Run SSH command
2. If the SSH results return an Authorization header, or the domain is different than the Git remote's, make the request with the merged header from the SSH results.
3. If the domain matches the Git remote's and there is no Authorization header, then look for existing credentials for the domain _without prompting the user_.
4. Make the request (with the credentials if given).
5. If the request returns with a 401, prompt the user for a password and retry the request.

## HTTPS auth

1. Look for existing credentials for the domain _without prompting the user_.
2. Make the request (with the credentials if given).
3. If the request returns with a 401, prompt the user for a password and retry the request.

* [ ] implement above changes in the code

Closes #243 (/cc @ddanier)

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -15,7 +15,10 @@ too.
 
 The Git LFS API uses [git credentials](http://git-scm.com/docs/gitcredentials) to
 access the API username and password.  If the Git LFS domain matches the Git
-remote's, Git LFS will not prompt you again to enter the password.
+remote's, Git LFS will not prompt you again to enter the password.  If there are
+no git credentials for the domain, Git LFS will make the request without any
+authentication.  If the server returns a 401 request, then the Git LFS client
+can ask for credentials from the user before retrying the request.
 
 If the Git remote is using SSH, Git LFS will execute the `git-lfs-authenticate`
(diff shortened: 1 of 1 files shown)


rubyist:
:+1:

technoweenie:
Closing this. We decided to leave the current API alone, and introduce 401 support for the batch API.

PR #252, opened 2015-04-20
Link to roadmap in readme

bkeepers:
to make it a little easier to discover.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ display a progress bar while downloading or opt out of a large download.
 content locally on a CDN, resumable uploads and downloads, or performing
 requests in parallel for faster transfers.
 
+See the [ROADMAP](ROADMAP.md) for other planned and desired features.
+
 ## Known Implementations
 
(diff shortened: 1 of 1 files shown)


technoweenie:
Ah cool.  I thought I had a PR to link this in the [docs readme](https://github.com/github/git-lfs/tree/master/docs) too.  I think this change was left out when I split the PR into smaller ones yesterday.

Do you think the roadmap is appropriate in the root, or should it live in docs?

bkeepers:
> Do you think the roadmap is appropriate in the root, or should it live in docs?

Doesn't matter to me. The README is pretty simple right now, so maybe keep as much stuff in the root until it starts to feel overwhelming, then move content that doesn't apply to everyone into the docs directory?

wadewilliams:
> Doesn't matter to me. The README is pretty simple right now, so maybe keep as much stuff in the root until it starts to feel overwhelming

:+1:

PR #255, opened 2015-04-22
update godeps file to new format

Aorjoa:
Currently, this project use old format of godeps (dependency tool) and i would like to update to new format.


technoweenie:
This project uses the bundled `script/gpm` to handle dependencies.  I like it because it's only a simple shell script.  I'm also waiting to see what shakes out from the [official dependencies proposal](https://groups.google.com/forum/m/#!topic/golang-dev/nMWoEAG55v8) before making a big change like this.

Aorjoa:
Thanks for your inform.

technoweenie:
Bah, I got tired of waiting: https://github.com/github/git-lfs/pull/330

I did run into an interesting issue: The "github.com/olekukonko/ts" package is only used in a Go file with `+build windows`.  Since go isn't including that file in builds on my mac, Godeps originally left it out of `Godeps/Godeps.json`. I had to include a reference in the new `./git-lfs.go` file.

PR #258, opened 2015-04-23
Support concurrent uploads

rubyist:
This PR will add support to allow uploads to happen concurrently.

- [x] Create an upload queue
- [x] Synchronize initial `git config` call
- [x] Synchronize first credentials access
- [x] Display all upload errors
- [x] Add a configuration option with a sensible default
- [x] Nicer progress output
- [x] Docs
- [x] Fix all the TODOs I added


lfs/client.go
```diff
@@ -125,6 +126,58 @@ func (b *byteCloser) Close() error {
...
+}
+
+func (q *UploadQueue) Upload(oidPath, filename string, cb CopyCallback) {
+	q.uc <- upload{oidPath, filename, cb}
+}
+
+func (q *UploadQueue) Wait() {
+	close(q.uc)
+	q.wg.Wait()
+	close(q.ec)
```
technoweenie: Just to clarify, closing these channels will stop the for loops in the goroutines once they've processed the last messages.  Right?  Not that it matters much since the command will exit right after this happens.

Also, I think `Wait()` is a strange name.  It makes sense here, looking at the implementation of the queue.  But I don't want the upload queue in the push command to _wait_, I want it to _process_ or _run_ or something.  Prior art: http://golang.org/pkg/os/exec/#Cmd.Run

lfs/client.go
```diff
@@ -125,6 +126,58 @@ func (b *byteCloser) Close() error {
...
 
+type upload struct {
+	OIDPath  string
+	Filename string
+	CB       CopyCallback
+}
+
+type UploadQueue struct {
+	uc     chan upload
+	ec     chan *WrappedError
```
technoweenie: What do these variables mean?  UploadChannel and ErrorChannel?

technoweenie:
> Synchronize first credentials access

It might be cool to restructure `lfs.Upload` so it can take a number of files.  Then, it could post an array of objects and get back multiple link relations.  One Git LFS API request for every `n` files.

Probably out of scope for this PR.  Just thinking....

technoweenie:
How will upload progress work on the cli output?  We almost need something like the pb package that can track multiple, separate progress bars.

lfs/client.go
rubyist: Yeah, I'm working on renaming some things, my rough pass names tend to be ... rough :smile:

lfs/client.go
rubyist: Correct on the goroutines.

For `Wait()`, starting or running is implicit, we don't need to wait until the loop that figures out what to upload to finish before starting uploads. See http://golang.org/pkg/os/exec/#Cmd.Wait and http://golang.org/pkg/sync/#WaitGroup.Wait (which is what it's actually doing).

rubyist:
I think posting an array would be good, that'll save a lot of round trips when there's a lot of files. I would agree that's out of scope for this PR, that would require an API change because I don't think we currently say that is allowed.

For progress, it probably will have to be something with pb showing multiple bars. That should have been added to the TODO list.

lfs/client.go
technoweenie: <bike shedding comment>.  Cool :+1:

rubyist:
A progress bar that shows multiple entries is going to be problematic. It's not too complicated on unixy environments with ansi codes, but windows is a different story. I think you can access some terminal manipulation routines through the windows syscalls and kernel32, but that's a bit beyond my knowledge. There are a couple cross platform terminal ui libraries (e.g. go-termbox), but those want to take over the entire screen, and with git also dumping output to the screen that's not very practical.

We have a few options we can do:

* Don't show the progress bar(s) for concurrent uploads
* Show a single progress bar that covers the entire upload set
* Show a multi-bar on unix, and nothing or the single bar on windows (perhaps someone can contribute a windows multi-bar)

technoweenie:
> Show a single progress bar that covers the entire upload set

I like this idea.  Instead of a single "uploading foo.bar (100MB)" message per file, have a single message showing the number of files and total storage, as well as a single progress bar for the whole thing.

rubyist:
That's my leaning as well, because it's the simplest :).

rubyist:
pb doesn't really know about progress bars that span multiple items, but we can sneak it in with the newer `Prefix()` method by updating the prefix as transfers complete:

uploads

technoweenie:
Super cool :+1:

commands/command_ls_files.go
```diff
@@ -27,7 +27,7 @@ func lsFilesCommand(cmd *cobra.Command, args []string) {
 		}
 	}
 
-	pointers, err := scanner.Scan(ref, "")
+	pointers, err := lfs.Scan(ref, "")
```
technoweenie: I wonder if this should be ScanRefs or something?

rubyist:
This is now ready for :eyes:

The number of uploads that will happen concurrently is configured via `lfs.concurrentuploads` (better name?). It defaults to 3 (better default?).

When `push` or `pre-push` happens, the files to be uploaded are calculated as before and added to a queue. The queue is then processed. In order to prevent being prompted for authentication n times simultaneously, the queue is processed sequentially until an API call completes successfully. This should happen right after the first API call. Once an API call succeeds the upload workers will wake up and begin processing the queue concurrently until it's finished.

Progress output is now a single bar that covers the combined sizes of all uploads (see gif above).

technoweenie:
Found a small issue:

So I used the new `git lfs push` command twice in row, without a `git push` in between.  The progress counter stayed at "0 of 100 files" until the first upload.  It skipped the first 50 since the first `git lfs push` call already uploaded them.  It'd be cool if it could update the file counter and the byte counter on an upload that can be skipped because the remote server already has it.

```
$ git lfs push origin master 
(61 of 100 files) 3.91 MB / 19.53 MB  20.00 % 1m56s     
```

I'm using this to generate 50 Git LFS files to upload.

```ruby
require "time"
require "fileutils"

now = Time.now.strftime("%Y%m%d-%I%M%S")
FileUtils.mkdir_p(now)

50.times do |i|
  File.open("#{now}/#{i}.dat", "w") do |f|
    f.write(`openssl rand -hex 102400`)
  end
end

puts `git add #{now}`
puts `git commit -m "files for #{now}"`
```

technoweenie:
Here's how it looks now:

```
# create 50 files and commit

$ g lfs push origin master
(50 of 50 files) 9.77 MB / 9.77 MB  100.00 % 34s                                                                                                                                                                                                                                                                                                                            

# create 50 more files and another commit

$ git push origin master
(100 of 100 files) 19.53 MB / 19.53 MB  100.00 % 49s
```

The progress bar looks good.  There's a lot of overhead, but I think we'll need to make some major API changes to solve.

PR #261, opened 2015-04-23
Pre push cmd

technoweenie:
This renames `git lfs push` to `git lfs pre-push`, but leaves an alias around.  I'd like to remove the `git lfs push` alias by v0.6.x so we can eventually work towards a `git lfs push` command designed for users.


technoweenie:
Redoing this so we can split the commands.  Not just rename.  See https://github.com/github/git-lfs/pull/263

PR #263, opened 2015-04-24
Split push cmd

technoweenie:
As discussed in https://github.com/github/git-lfs/pull/228, the push command actually has two modes:

```
# run by a user
$ git lfs push origin master

# run by pre-push hook
$ echo "refs/heads/master {local-sha1} refs/heads/master {remote-sha1}" | git lfs push origin https://github.com/some/repo
```

This PR splits them apart into separate commands, and updates the docs.  The old functionality in `push` is kept, until we can safely roll pre-push updates out.

### DIFF
--- a/commands/command_push.go
+++ b/commands/command_push.go
@@ -1,56 +1,37 @@
 package commands
 
 import (
-	"fmt"
 	"github.com/github/git-lfs/git"
 	"github.com/github/git-lfs/lfs"
-	"github.com/github/git-lfs/pointer"
 	"github.com/github/git-lfs/scanner"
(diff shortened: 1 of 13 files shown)


technoweenie:
Okay, this is ready for some :eyes: 

@mhagger: This now has distinct `push` and `pre-push` commands.  The docs have been updated.  The confusing "repo" and "refspec" names have been replaced.

@rubyist: This now silently upgrades the `pre-push` hook if the contents match any that Git LFS ever sets.  You can run `git lfs update --force` to clobber any existing pre-push hook.

PR #264, opened 2015-04-24
require stdin content on certain commands

### LINKED ISSUE #210: Handle accidental `git lfs [smudge | clean]` runs
These are plumbing commands that aren't designed to be run by users.  However, running them should not hang indefinitely for users.

technoweenie:
This fails fast in a few commands (`clean`, `smudge`, `pointer`, and `push`) if they expect STDIN, but don't get it.  This commonly happens in `clean` and `smudge` commands if users run them interactively.

Fixes #210

### DIFF
--- a/commands/command_clean.go
+++ b/commands/command_clean.go
@@ -16,6 +16,7 @@ var (
 )
 
 func cleanCommand(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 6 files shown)


rubyist:
nice, :+1:

PR #265, opened 2015-04-24
add diff/merge properties to gitattributes

technoweenie:
@peff suggested setting diff/merge properties in `.gitattributes`.  This gives us and users a place to set custom Git LFS options for diffing and merging.  Since Git LFS has nothing for this now, it should fall back to the default diffing and merging behavior.

Note: This only affects the `git lfs track` command.  Users can still manually add the lines to `.gitattributes` with their own properties.

/cc @mhagger too.

### DIFF
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -58,7 +58,7 @@ func trackCommand(cmd *cobra.Command, args []string) {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


Argemiro76:
Help, I've been hacked!!

peff:
:+1: from me.

PR #270, opened 2015-04-28
Add link to S3 implementation

keithws:
I found an open source implementation of the LFS server API that stores large files on Amazon S3. It seems worthy of a link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ See the [ROADMAP](ROADMAP.md) for other planned and desired features.
 
 - [GitHub.com](https://github.com/early_access/large_file_storage) (support coming soon!)
 - [github/lfs-test-server](https://github.com/github/lfs-test-server) (reference server implementation)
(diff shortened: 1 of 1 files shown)


technoweenie:
I've had that starred, but I haven't had a chance to play around with it.  Is anyone using it?

bkeepers:
> Is anyone using it?

How much editorial do we want to do on this list? We could either include all "known implementations" (as the heading says) or wait until implementations have usage and are verified.

The code looks clean on this implementation, but it also looks like it doesn't stream the request/response data and reads the full objects into memory, which probably won't scale very well.

technoweenie:
I added an implementations page on the wiki: https://github.com/github/git-lfs/wiki/Implementations

This way the list can be maintained by the community without being blocked on PRs like this.

keithws:
The Implementations wiki page looks like a good solution.

PR #271, opened 2015-04-28
pass pointers straight through 'git lfs clean'

technoweenie:
This teaches `git lfs clean` to pass Git LFS pointers straight through.  There's no point in saving a pointer of a pointer.


technoweenie:
@rubyist: Sorry, this is probably going to cause some fun merge conflicts with #258.

rubyist:
-merge coooooooooonflicts

:+1:

Issue #268, opened 2015-04-28
Release: v0.5.1


technoweenie:
I've been keeping track in a draft release, but just realized that you need repo push access to see it.  So we're going to start tracking these in issues instead.  I'm hoping to push v0.5.1 out tomorrow.  Please comment if there's something missing that you think needs to go in.

* [x] Fix bug where `git lfs clean` will clean Git LFS pointers too #271 (@technoweenie)
* [x] Better timeouts #215 (@Mistobaan)
* [x] Concurrent uploads through `git lfs push` #258 (@rubyist)
* [x] Fix `git lfs smudge` behavior with zero-length file in `.git/lfs/objects` #267 (@technoweenie)
* [x] Separate out pre-push hook behavior from `git lfs push` #263 (@technoweenie)
* [x] Add diff/merge properties to .gitattributes #265 (@technoweenie)
* [x] Respect `GIT_TERMINAL_PROMPT ` #257 (@technoweenie)
* [x] Fix CLI progress bar output #185 (@technoweenie)
* [x] Fail fast in `clean` and `smudge` commands when run without STDIN #264 (@technoweenie)
* [x] Fix shell quoting in pre-push hook.  #235 (@mhagger)
* [x] Fix progress bar output during file uploads.  #185 (@technoweenie) 
* [x] Change `remote.{name}.lfs_url` to `remote.{name}.lfsurl` #237 (@technoweenie)
* [x] Swap `git config` order.  #245 (@technoweenie)
* [x] New `git lfs pointer` command for generating and comparing pointers #246 (@technoweenie)
* [x] Follow optional "href" property from git-lfs-authenticate SSH command #247 (@technoweenie)
* [x] Spec updates
  * `.git/lfs/objects` clarifications: #212 (@rtyley), #244 (@technoweenie)
  * man page updates: #228 (@mhagger)
  * pointer spec clarifications: #246 (@technoweenie)
* [x] Code comments
  * #225 (@thekafkaf)

PR #274, opened 2015-04-30
changelog file for next release

technoweenie:
This brings over the items from #268 to a new changelog file.


bkeepers:
:+1:  I think changelogs are really helpful. What do you think about including a link to a compare URL as well, like https://github.com/bkeepers/dotenv/blob/master/Changelog.md?

technoweenie:
Good call.  Should've opened this way sooner :)

PR #277, opened 2015-05-02
Fixed api.md not containing the correct information about href

ddanier:
While playing with https://github.com/ddanier/django-git-lfs I noticed the documentation for the href attribute is a bit off. If you point git-lfs to /objects it will produce /objects/objects. So the correct URL may not contain "objects". This commit fixes this documentation issue.


technoweenie:
Ah good catch.  I'd leave the trailing slash off too :)

Issue #262, opened 2015-04-24
git-lfs fsck (bitrot detection)


zeroshirts:
Would anyone else find it useful if git-lfs offered a fsck command that verified the integrity of tracked (large) files?  Re-calculating hashes for files under .git/lfs/objects and comparing against the object name (dir+filename) could serve as an integrity check.

My goal with this would be to detect bitrot in large files like photos and mp3s.

technoweenie:
I think this would be cool :+1:  Seems like it could go hand in hand with #260.

technoweenie:
FWIW this would've helped an issue like https://github.com/github/git-lfs/issues/266

zeroshirts:
I have a rough implementation up at https://github.com/zeroshirts/git-lfs.  @technoweenie , could you take a quick look and comment on my approach?  I need to add unit tests, at the very least.

technoweenie:
Cool, can you open a pull request?  You can create one [from the compare view](https://github.com/github/git-lfs/compare/master...zeroshirts:master).  It's also typically best to start off in your own branch.

PR #280, opened 2015-05-04
Add LFS fsck feature

zeroshirts:
This is a first-cut of the LFS fsck command to address issue #262.

I wanted to put this out there for review before adding stuff like on-the-fly status completion updates (which would be useful given the possibility of LFS fsck hashing gigs of media content).  I'm curious whether this is a good approach to the fsck problem.

After a fresh git clone and running script/bootstrap, I encountered a ton of errors in the unit tests for the commands package, so the changes you see in the non fsck unit tests were to fix those issues on my machine.

Example of the unit test errors I encountered:
```
--- FAIL: TestUpdateWithOldPrePushHook_2 (0.04s)
	update_test.go:174: Error writing pre-push in Before(): open /private/var/folders/0t/n69_kx414lg481jhdyf8n6xh0000gn/T/git-lfs-tests/empty/.git/hooks/pre-push: no such file or directory
```

### DIFF
--- a/commands/command_fsck.go
+++ b/commands/command_fsck.go
@@ -83,7 +83,7 @@ func fsckCommand(cmd *cobra.Command, args []string) {
 
 	err := doFsck(lfs.LocalGitDir)
 	if err != nil {
-		Panic(err, "Could not fsck Git LFS files")
+		Panic(err, "Git LFS fsck failed")
 	}
 	Print("Git LFS fsck OK")
 }
--- a/commands/command_fsck.go
+++ b/commands/command_fsck.go
@@ -61,7 +61,7 @@ func doFsck(localGitDir string) error {
 		if err != nil {
 			return err
 		}
-		_ = f.Close()
+		f.Close()
 
 		recalculatedOid := hex.EncodeToString(oidHash.Sum(nil))
 		if recalculatedOid != p.Pointer.Oid {
@@ -85,7 +85,7 @@ func fsckCommand(cmd *cobra.Command, args []string) {
 	if err != nil {
 		Panic(err, "Could not fsck Git LFS files")
 	}
-	Print("LFS fsck OK")
+	Print("Git LFS fsck OK")
 }
 
 func init() {
--- a/commands/command_fsck.go
+++ b/commands/command_fsck.go
@@ -1,0 +1,93 @@
+package commands
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"github.com/github/git-lfs/git"
+	"github.com/github/git-lfs/lfs"
+	"github.com/spf13/cobra"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+type fsckError struct {
+	name, oid string
+}
+
+func (e *fsckError) Error() string {
+	return "Object " + e.name + " (" + e.oid + ") is corrupt"
+}
+
+var (
+	fsckCmd = &cobra.Command{
+		Use:   "fsck",
+		Short: "Verifies validity of Git LFS files",
+		Run:   fsckCommand,
+	}
+)
+
+func doFsck(localGitDir string) error {
+	ref, err := git.CurrentRef()
+	if err != nil {
+		return err
+	}
+
+	pointers, err := lfs.ScanRefs(ref, "")
+	if err != nil {
+		return err
+	}
+
+	// TODO(zeroshirts): do we want to look for LFS stuff in past commits?
+	p2, err := lfs.ScanIndex()
+	if err != nil {
+		return err
+	}
+	// zeroshirts: assuming no duplicates...
+	pointers = append(pointers, p2...)
+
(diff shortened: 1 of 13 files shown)


zeroshirts:
Not sure why TestFsck() is breaking on the build box.  Will investigate.

commands/command_fsck.go
```diff
@@ -0,0 +1,93 @@
...
+	"path/filepath"
+)
+
+type fsckError struct {
+	name, oid string
+}
+
+func (e *fsckError) Error() string {
+	return "Object " + e.name + " (" + e.oid + ") is corrupt"
+}
```
technoweenie: If we're not exposing any of this, why not have something like:

```golang
func fsckError(name, oid string) error {
  return fmt.Errorf("Object %s (%s) is corrupt", name, oid)
}
```

commands/command_fsck.go
```diff
@@ -0,0 +1,93 @@
...
+// NOTE(zeroshirts): Ideally git would have hooks for fsck such that we could
+// chain a lfs-fsck, but I don't think it does.
+func fsckCommand(cmd *cobra.Command, args []string) {
+	lfs.InstallHooks(false)
+
+	err := doFsck(lfs.LocalGitDir)
+	if err != nil {
+		Panic(err, "Could not fsck Git LFS files")
+	}
+	Print("LFS fsck OK")
```
technoweenie: Let's go with "Git LFS" everywhere for consistency.

commands/command_fsck.go
```diff
@@ -0,0 +1,93 @@
...
+		if err != nil {
+			return err
+		}
+
+		oidHash := sha256.New()
+		_, err = io.Copy(oidHash, f)
+		if err != nil {
+			return err
+		}
+		_ = f.Close()
```
technoweenie: I'd just leave the `_ =` out.

technoweenie:
> do we want to look for LFS stuff in past commits?

I think there are at least 4 modes that we could support.  

1. Scan HEAD + Index (default)
2. Scan a commit by name.  `git lfs fsck --commit abcdef`
3. Scan `.git/lfs/objects` `git lfs fsck --all`
4. Scan EVERYTHING.

I think we can get away with 1-3.  Scanning the entire git history isn't interesting unless you have everything downloaded already.

zeroshirts:
Thanks for the comments, @technoweenie .  I'll make the changes.  Before I add support for different modes, I'm going to push some temporary debug code to diagnose the os.Stat() error.

technoweenie:
>  I'm going to push some temporary debug code to diagnose the os.Stat() error.

Use the [testing package's `Logf()`](http://golang.org/pkg/testing/#T.Logf).  They only display on test runs that have the error, so there's no harm leaving them in after you fix the bug.

zeroshirts:
Alright, I'm stuck.  I'm not sure why this works locally and not on the CI box, and the test is simple, so perhaps someone else could take a quick look.

TestFsck() is failing to find the LFS object corresponding to a LFS file after a commit.  I have test code that recursively prints out the repo dir, and on the CI machine, there are simply no LFS objects under .git/lfs/objects.  On my local machine, of course, everything works as expected.

If someone can figure this out, I'd still like to get clarification on how to run a bunch of commands on a TestRepository sequentially without the repo state getting cleared.  Currently -- and several other unit tests like TestCleanPointer depend on this behavior -- func (r *Repository) test(path string)() invokes repo.clone() before each testCommand Run(), so it isn't possible.

technoweenie:
Ugh, sorry about the sad state of integration testing here. I have some ideas on making it better.

The reason it's failing is because Git LFS isn't installed.  You can replicate it locally by removing the clean filter:

    $ git config --global --unset filter.lfs.clean

I'm going to update the test to manually write a bad file to `.git/lfs/objects`.

technoweenie:
I have [my own branch](https://github.com/github/git-lfs/compare/fsck-command?expand=1) that uses your commits up until you started all those debug commits.  I also changed the test so that it corrupts the file in the `before` callback.  Once the test passes, I'd add a check in the `after` callback that ensures the file is gone.

I want to change the fsck behavior so that it just prints the errors to STDOUT without logging an error.  We can probably just delete the corrupt object, and expect the client to download it on demand from the server.

```
$ git lfs fsck
Could not fsck Git LFS files:

Object a.dat (916f0027a575074ce72a331777c3478d6513f786a591bd892da1a577bf2335f9) is corrupt
```

technoweenie:
I did some more work and opened a PR: https://github.com/github/git-lfs/pull/290.  Still can't get the tests to pass though.

technoweenie:
Also, I'm going to make https://github.com/github/git-lfs/pull/291 a priority of mine.  Your testing experience in this PR is unacceptable.

zeroshirts:
Thanks for picking this up and taking on the testing issues.

PR #285, opened 2015-05-05
Endpoint for batch upload/download operations

rubyist:
This PR allows the LFS client to batch its upload and download operations.

Typically, the client has to ask the server about each object it wants to upload or download, resulting in many http round trips. This causes significant delay when dealing with a large amount of objects.

This PR proposes a change to the API that provides a batch endpoint that gives the client information about a number of objects at once.

The client will `POST` an array of oid/size objects to the batch endpoint, and the endpoint will return an array of link relations for each object asked about. The presence or absence of any of the `download`, `upload`, and `verify` relations can be used to determine the state of the object.

* `download` present - server has and has verified the object
* `upload` present - server does not have the object, client can upload and follow `verify` (if given)

TODO:
- [x] Spec the batch endpoints
- [x] Upload queue can work with batch endpoint
- [x] Add a `get` command to bulk download


docs/api.md
```diff
@@ -266,6 +266,66 @@ only appears on a 200 status.
...
+>     "size": 123
+>   }
+> ]
+>
+< HTTP/1.1 200 Accepted
+< Content-Type: application/vnd.git-lfs+json
+<
+< [
+<   {
+<     "oid": "1111111",
```
technoweenie: Missing the size.

technoweenie:
Interesting.  I imagined having endpoints like `/download`, `/upload`, and `/verify`.  Having one endpoint that can send the correct hypermedia relations is better :+1:

technoweenie:
Any thoughts on always sending and returning objects?  So instead of:

```
[
  { "oid": "abcdef" }, ...
]
```

Send a hash.  It gives us a place to add extra meta data for the request.  Maybe we want to tell the server what the local commit is that we're downloading the objects for?

```
{
  "commit": "sha1",
  "branch": "master",
  "objects": [
    { "oid": "abcdef" }, ...
  ]
}
```

rubyist:
I like that idea :+1:

meltingice:
Cool, I like this idea. Should help to speed things up a lot.

Is this going to be a mandatory endpoint, or is there some way the client can discover whether or not the server offers batching? Perhaps try the batch endpoint first, and then fall back to non-batched endpoints if it 404's?

technoweenie:
> Is this going to be a mandatory endpoint, or is there some way the client can discover whether or not the server offers batching? Perhaps try the batch endpoint first, and then fall back to non-batched endpoints if it 404's?

I'd like to transition the code to using the batch endpoint as the single required endpoint for servers to implement for Git LFS v1.0.

1. First, ship experimental support.  Enable some local `lfs.batch` config on a repository to use it.
2. Make the batch API default, and fallback to the original API if the server returns a 404 or `lfs.batch` is false.
3. Remove support for the old `/objects` API.

ddanier:
I think a batch job would be really nice.

However I don't like the idea of adding additional data to the requests (commit sha1, branch). The git-lfs server is currently pretty easy to implement as it's basically a simple key value store. This is - in my eyes - one of the key features that make it a feasible solution for many git providers (GitLab could for example add this without many problems, I did my own implementation in just under 10 hours). Now just adding more information might not complicate things, but allow us to do so. In the end this might make implementations much harder.

Another thing to keep in mind is that we are talking about big files. So the API request to check a files existance should have a pretty low impact on the overall performance, most of the time should be spended up/downloading the file itself. So although adding a batch job is a great thing from a technical perspective it might not necessarily be that much better from a usability perspective. My personal opinion is to just implement it anyways, as you cannot change this so easily when more servers exist.

technoweenie:
> However I don't like the idea of adding additional data to the requests (commit sha1, branch).

I want to add this meta data in a way that's not required for servers to process.  It should be easy to just focus on the proposed "objects" key if that's all your implementation cares about.  I'm hoping to keep the API simple for small servers, with optional features that bigger servers can choose to implement.

> So the API request to check a files existance should have a pretty low impact on the overall performance, most of the time should be spended up/downloading the file itself.

I hope to ship support for both APIs in the client first, defaulting to the current key/value methods.  If we can't prove that reducing the API overhead helps in a meaningful way, we can scrap it.  I'm anxious to compare this to the work done in https://github.com/github/git-lfs/pull/258.

Jericho25:
Good day!Let's just make it simply modern-convenient. . .

                          #jgaj   #jeckohero

Jericho25:
This thing is gonna be called the. #milkyway
  
              #jgaj #jeckohero

rubyist:
I think this is ready for some review. What this PR provides:

An extra endpoint in the API for batching operations. Instead of the initial `POST` of `oid` and `size` before a download or upload, the batch endpoint accepts a `POST` of an object containing an array of objects of `oid` and `size`. e.g. `{"objects":[{"oid": "...", "size": 42}, ...]}`. It returns an object containing an `objects` array where each entry will contain the normal upload or download link reference, depending on whether or not the server has that object. The client can then proceed to upload or download as normal. This saves an API round trip for each object.

An initial pass at a `git lfs get` command is also included. This command will download all objects for a given commit or branch. The object files will be added to the `.git/lfs/objects` directory as normal, and if the currently checked out branch/commit matches the downloaded branch/commit, the working directory files will be updated. We can further expand this command to take a range of commits or a single object by name or oid.

rubyist:
I should add that since github.com does not yet have a batch api, any testing will need to be done against the lfs test server. A matching PR containing a batch endpoint is here: https://github.com/github/lfs-test-server/pull/27

test/test-batch-transfer.sh
```diff
@@ -0,0 +1,74 @@
...
+  contents_oid=$(printf "$contents" | shasum -a 256 | cut -f 1 -d " ")
+
+  printf "$contents" > a.dat
+  git add a.dat
+  git add .gitattributes
+  out=$(git commit -m "add a.dat" 2>&1)
+  echo "$out" | grep "master (root-commit)"
+  echo "$out" | grep "2 files changed"
+  echo "$out" | grep "create mode 100644 a.dat"
+  echo "$out" | grep "create mode 100644 .gitattributes"
```
technoweenie: I've started using this pattern:

```shell
git commit -m "add a.dat" 2>&1 | tee commit.log
grep "master (root-commit)" commit.log
```

Nice thing is that the output is only printed to STDOUT once with `tee`.  If the last grep fails, you'll only see the commit output once, and the matches of the successful grep commands.

test/test-batch-transfer.sh
```diff
@@ -0,0 +1,74 @@
...
+
+  # Clone the repository again to $TRASHDIR/repo. This will be used to commit
+  # and push objects.
+  clone_repo "$reponame" repo
+
+  # This executes Git LFS from the local repo that was just cloned.
+  out=$($GITLFS track "*.dat" 2>&1)
+  echo "$out" | grep "Tracking \*.dat"
+
+  contents=$(printf "a")
```
technoweenie: Ha, you got this from my `test-happy-path.sh`. Unsure why `contents="a"` wouldn't work :)

Jericho25:
Congratulations to us!!!          #jgaj

commands/command_get.go
```diff
@@ -0,0 +1,75 @@
...
+
+	q.Process()
+
+	target, err := git.ResolveRef(ref)
+	if err != nil {
+	}
+
+	current, err := git.CurrentRef()
+	if err != nil {
+	}
```
technoweenie: Errors should be handled.

commands/command_get.go
```diff
@@ -0,0 +1,75 @@
...
+
+	if target == current {
```
technoweenie: Does it make sense to return early if they aren't equal, and remove a level of nesting?

lfs/transfer_queue.go
```diff
@@ -0,0 +1,212 @@
...
+			switch event {
+			case apiEventSuccess:
+				atomic.StoreInt32(&q.clientAuthorized, 1)
+				q.authCond.Broadcast() // Wake all remaining goroutines
+				return
+			case apiEventFail:
+				q.authCond.Signal() // Wake the next goroutine
+			}
+		}
+	}()
```
technoweenie: Does it make sense to add an `apiEventShutdown` so that a TransferQueue can clean up any lingering goroutines once it's done? I don't think we have any cases where multiple transfer queues are processed in the same process, so it may not matter.

lfs/transfer_queue.go
```diff
@@ -0,0 +1,212 @@
...
+// making only one POST call for all objects. The results are then handed
+// off to the transfer workers.
+func (q *TransferQueue) processBatch() {
+	q.files = 0
+	transfers := make([]*objectResource, 0, len(q.transferables))
+	for _, t := range q.transferables {
+		transfers = append(transfers, &objectResource{Oid: t.Oid(), Size: t.Size()})
+	}
+
+	objects, err := Batch(transfers)
```
technoweenie: We should define an upper bound for the number of transfers to batch at a time. This way the client isn't sending massive json payloads that servers have to process. We don't want to burden servers with the prospect of handling POST payloads with thousands of objects.

lfs/transfer_queue.go
technoweenie: On closer examination I see that `apiEventSuccess` returns early. :+1:

technoweenie:
Merging this, as it seems like a great improvement, and will serve as a base for future changes.

catphish:
As per #341 I'd really appreciate it if this capability could be made available via the SSH API in addition to HTTP.

PR #290, opened 2015-05-08
Fsck command

technoweenie:
This builds on #280 by @zeroshirts, adding an fsck command.  Right now, all it does is makes sure the `.git/lfs/objects` files all match the OID (a sha256 hash of the contents).  By default, it removes any bad files so that Git LFS can re-download them the next time they're referenced.

Questions:

Is `fsck` the right name?  It would be cool to eventually have something like Homebrew's `doctor` command for checking multiple things besides file system stuff.

Is deleting the corrupt objects the right default?  I'm not sure what we could do to fix files.

Observations (not necessarily blockers for this PR):

We should think hard about how `git fsck` decides what Git refs to scan.  It currently only looks in HEAD and the index.  Though I can come up with other useful scenarios:

* Scan a commit by name: `git lfs fsck --commit abcdef`
* Scan everything in `.git/lfs/objects`: `git lfs fsck --all`
* Scan all commits.

It'd also be nice if there was some consistency in how options work with `git lfs fsck` and potentially similar commands like `git lfs get`.

Also, we may want to build a higher level scanner in the `lfs` package, instead of calling `lfs.ScanRefs()` and `lfs.ScanIndex()` directly.  Something that we can re-use in other commands, without having to export internal things like `*lfs.wrappedPointer`.

```golang
cfg := &lfs.ScanConfig{
  StartCommit: "abcdef",
  EndCommit: "ghijkl",
}

err := lfs.Scan(cfg, func(path string, p *lfs.Pointer) error {
  ...
})
```

### DIFF
--- a/commands/clean_test.go
+++ b/commands/clean_test.go
@@ -227,15 +227,12 @@ func TestCleanWithCustomHook(t *testing.T) {
 
 	path := filepath.Join(repo.Path, ".git", "lfs", "objects")
 	prePushHookFile := filepath.Join(repo.Path, ".git", "hooks", "pre-push")
-	customHook := []byte("test")
+	customHook := "test"
 
 	cmd.Before(func() {
-		err := ioutil.WriteFile(prePushHookFile, customHook, 0755)
-		if err != nil {
-			t.Fatal(err)
-		}
+		repo.WriteFile(prePushHookFile, customHook)
 
-		_, err = os.Open(path)
+		_, err := os.Open(path)
 		if _, ok := err.(*os.PathError); !ok {
 			t.Fatalf("'%s' should not exist", path)
 		}
--- a/commands/command_fsck.go
+++ b/commands/command_fsck.go
@@ -27,9 +27,17 @@ func doFsck(localGitDir string) (bool, error) {
 		return false, err
 	}
 
+	// The LFS scanner methods return unexported *lfs.wrappedPointer objects.
+	// All we care about is the pointer OID and file name
+	pointerIndex := make(map[string]string)
+
 	pointers, err := lfs.ScanRefs(ref, "")
 	if err != nil {
 		return false, err
+	}
+
+	for _, p := range pointers {
+		pointerIndex[p.Oid] = p.Name
 	}
 
 	// TODO(zeroshirts): do we want to look for LFS stuff in past commits?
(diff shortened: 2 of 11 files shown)


zeroshirts:
Good catch finding the part where I leaked a file descriptor on that error return path ( https://github.com/github/git-lfs/commit/2f6c3beba36f5c67026339b1972814b043a1e5bf ).  The only problem with using the defer f.Close() trick is that you'll hold to all of the file descriptors in that for loop and then close them all at once on function exit.  Imagine if a user has thousands of tracked LFS files... we could potentially exhaust the pool of available file descriptors.  I suggest closing them as 'early' as possible to avoid this.

zeroshirts:
I have a bad feeling about deleting corrupted objects.  I would be cool with a 'recover' command that attempted to pull down uncorrupted object versions from Git LFS remotes (but of course, if the user hasn't pushed them, they won't exist).

Imagine that I have a photo that was deleted from my Git LFS managed photos dir, but is still in my commit history.  Imagine the photo's Git LFS object is corrupted, slightly.  I *would* like to know from Git LFS that it's corrupted, but I definitely don't want Git LFS making a value judgement for me that the entire photo is worthless..  the corruption might be imperceptible, or affect only a small region.  Same goes for an mp3 file.

There are other examples... imagine the user is running on consumer/prosumer RAID hardware that doesn't validate block/stripe checksums on-the-fly.  There could be a scenario where 'LFS fsck' iterates through the Git refs, and happens to get data from a RAID disk that has bitrot in the corresponding data.  If we adopted an "eager" delete-corruption policy, Git-LFS would nuke all of the RAID copies.  What's interesting (to me) is that the same 'LFS fsck' command run sometime later might pull the Git ref data from a different RAID drive that WASN'T suffering from silent bitrot.  Or, perhaps the user decides to kick-off a RAID scrub after LFS fsck detects the corruption.  The raid scrub might be able to heal the bitrot errors.

technoweenie:
I'm cool with swapping the default to the dry run, and requiring an addition flag to delete the corrupted objects.  I don't want it to log it as an exception like in the original implementation, though. I think it should list them out, along with a couple next steps:

1. Purge the corrupt files
2. Purge and re-download from a Git LFS server.  This flag should probably optionally take a remote name if you don't want to get it from `origin`.

technoweenie:
This is on hold until #306 is merged and we can properly test this!

michael-k:
link #262

Issue #293, opened 2015-05-10
git lfs not check git file system before start tracking files


Aorjoa:
File tracking not check necessary items before (git lfs require git file system and into git influence directory) , to prevent error I will hack some code for support this issue and I will show reproduce issue as below. ( I think it's like #200 but duplicate line of tracking file in my case)

System environments
git-lfs/0.5.1 (GitHub; darwin amd64; git 2.0.1; go 1.4.1) on Mac OS X : Yosemity 10.10.4

I woking in directory contain a testfile.png and use git lfs without git file system
```bash
$ git lfs track "testfile.png"
Tracking testfile.png
```
.gitattributes was created and I looking into file
```bash
$ cat .gitattributes
testfile.png filter=lfs diff=lfs merge=lfs -crlf
```
then use git lfs track again
```bash
$ git lfs track "testfile.png"
Tracking testfile.png
```
and read it
```bash
$ cat .gitattributes
testfile.png filter=lfs diff=lfs merge=lfs -crlf
testfile.png filter=lfs diff=lfs merge=lfs -crlf
```
It seem like git lfs not check git file system before. but if git already in the directory it’s warnning user about file exits in git lfs system (.gitattributes).

create git file system
```bash
$ git init
Initialized empty Git repository in /Users/AorJoa/repository/test_check_git_exists/.git/
```
git check tracking file before add any line
```bash
$ git lfs track "testfile.png"
testfile.png already supported
```
and file contains two line of tracking file
```bash
$ cat .gitattributes 
testfile.png filter=lfs diff=lfs merge=lfs -crlf
testfile.png filter=lfs diff=lfs merge=lfs -crlf
```

technoweenie:
Hey, I'm not able to replicate this:

```
$ git lfs track "testfile.png"
Tracking testfile.png

$ cat .gitattributes 
testfile.png filter=lfs diff=lfs merge=lfs -crlf

$ git lfs track "testfile.png"
testfile.png already supported

$ cat .gitattributes
testfile.png filter=lfs diff=lfs merge=lfs -crlf
```

I added a passing test for this in #323.  Are you still seeing this?  What version of Git LFS are you running?

Aorjoa:
I use `$ go install cmd/git-lfs/git-lfs.go` from commit 38657a8. It's still problem but I think test pass in #323 because `$ git init` was created .git directory. (I guess this issue coming from I run `$ git lfs track <some file>` out of git directory.

technoweenie:
[...]

PR #294, opened 2015-05-10
Prevent to track file outside git repository

Aorjoa:
according #293 my report issue, now i already hack code for fixed.
if anybody run `$ git lfs track` outside git repository it's show message "Git LFS cannot perform outside git repository."

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Git LFS Changelog
 
+## v0.5.1 (30 April, 2015)
+
 * Fix Windows install.bat script.  #223 (@PeterDaveHello)
 
 * Fix bug where `git lfs clean` will clean Git LFS pointers too #271 (@technoweenie)
--- a/lfs/lfs.go
+++ b/lfs/lfs.go
@@ -11,7 +11,7 @@ import (
 	"strings"
 )
 
-const Version = "0.5.0"
+const Version = "0.5.1"
 
 var (
 	LargeSizeThreshold = 5 * 1024 * 1024


rubyist:
Thanks for the PR! One piece of feedback I have is, we have some existing code that determines the local git directory and sets `lfs.LocalGitDir` on init. We should be able to just check whether or not that it set and avoid execing a git command here. :+1: for the test, that looks legit.

/cc https://github.com/github/git-lfs/issues/200

Aorjoa:
:+1:

Aorjoa:
I was change code for check lfs.LocalGitDir == ""

michael-k:
I think it's better to inline the function `workingInsideGit()` like this:
```go
if lfs.LocalGitDir == "" {
	Print("Git LFS cannot perform outside git repository.")
	return
}
```

Small helper functions are considered good style in go, but in this case it's just a bunch of negations (equal → false → negation) which makes it harder to understand what happens on a low level (maintainability).

The message in `Print()` makes it clear what happens on a high level (readability), so imho the “speaking name” `workingInsideGit` adds no further benefit.

And last, but not least, I'm voting in favor of squashing the commits together. :)

Aorjoa:
Thank you very much for your advice. :+1: I would like to change the message to "Not in a Git LFS repository". Do you think it's clear enough? @michael-k  :)

michael-k:
git itself says:
```
fatal: Not a git repository (or any parent up to mount point /home)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```
or with `GIT_DISCOVERY_ACROSS_FILESYSTEM=1`:
```
fatal: Not a git repository (or any of the parent directories): .git
```
`Not in a Git LFS repository` might not be correct, as `lfs.LocalGitDir` does not check if git-lfs was initialized.

technoweenie:
I redid this PR as #323.  Normally I wouldn't do something like that, but I wanted to test this on the new test suite.  The actual command change is pretty minimal.  My PR takes @michael-k's suggestion into account.  It also doesn't change the import formatting in `commands/command_track.go`.

PR #300, opened 2015-05-12
Codestyle

michael-k:
This is my first contribution to this project, so I'm not sure what's appreciated and what's not. Please let me know. I'm happy to change the PR in any way necessary.

As I worked myself through the codebase I made some changes to the codestyle and did some minor refactoring (switched if-else; returned nil as payload with error). I split it in small commits to make it easier to follow.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -136,13 +136,13 @@ func logPanic(loggedError error, recursive bool) string {
 	if len(os.Args) > 0 {
 		fmt.Fprintf(fmtWriter, " %s", strings.Join(os.Args[1:], " "))
 	}
-	fmt.Fprint(fmtWriter, "\n")
+	fmt.Fprintln(fmtWriter)
 
 	logEnv(fmtWriter)
-	fmt.Fprint(fmtWriter, "\n")
+	fmt.Fprintln(fmtWriter)
 
 	fmtWriter.Write(ErrorBuffer.Bytes())
-	fmt.Fprint(fmtWriter, "\n")
+	fmt.Fprintln(fmtWriter)
 
 	fmt.Fprintln(fmtWriter, loggedError.Error())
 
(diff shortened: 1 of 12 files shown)


technoweenie:
Looks good to me.  There's a lot of old Go code of mine in here written before I understood the popular Go conventions.

commands/commands.go
```diff
@@ -12,6 +10,9 @@ import (
 	"path/filepath"
 	"strings"
 	"time"
+
+	"github.com/github/git-lfs/lfs"
+	"github.com/spf13/cobra"
```
rubyist: If we're going to go with this style of import block (which I am in favor of, and is the only annoying difference between `goimports` and `go fmt`), then we should do it this way everywhere.

rubyist:
:+1: for me, other than hoping to see those import blocks consistent across the code base.

michael-k:
Thanks for the feedback. I'll change the imports.

technoweenie:
I don't think the imports are necessary to change in this PR.  Instead, I think we should setup a separate PR that somehow bundles `goimports`, and uses it in place of `script/fmt`.  I'd probably just add the `go get` in `script/bootstrap`.

rubyist:
I think this PR should leave them consistent, whichever way we go. I'm not entirely convinced bundling goimports is necessary, go fmt will honor the spaced blocks if it encounters them. I could be persuaded otherwise, though. I honestly don't use script/fmt and just run that stuff from my editor.

technoweenie:
Actually, you do use script/fmt, it's called during script/test :)

I'm totally cool with consistent imports in this PR.

On Tue, May 12, 2015 at 10:25 AM, Scott Barron <notifications@github.com>
wrote:

> I think this PR should leave them consistent, whichever way we go. I'm not
> entirely convinced bundling goimports is necessary, go fmt will honor the
> spaced blocks if it encounters them. I could be persuaded otherwise,
> though. I honestly don't use script/fmt and just run that stuff from my
> editor.
>
> —
> <https://github.com/github/git-lfs/pull/300#issuecomment-101339163>.
>



-- 
Rick Olson

michael-k:
I removed the import reformatting from this PR. (That's what I meant with “I'll change the imports.”)

technoweenie:
Thanks!  I think we'll just do that in an isolated PR.

PR #302, opened 2015-05-13
Add missing check for errors in smudge and clean commands

celgemark:
Make sure we exit with error code if the smudge or clean fails

### DIFF
--- a/commands/command_clean.go
+++ b/commands/command_clean.go
@@ -44,6 +44,10 @@ func cleanCommand(cmd *cobra.Command, args []string) {
 		file.Close()
 	}
 
+	if err != nil {
+		Panic(err, "Error cleaning asset.")
+	}
+
 	if cleaned != nil {
 		cleaned.Close()
 		defer cleaned.Teardown()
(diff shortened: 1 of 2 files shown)


commands/command_clean.go
```diff
@@ -44,6 +44,10 @@ func cleanCommand(cmd *cobra.Command, args []string) {
 		file.Close()
 	}
 
+	if err != nil {
+		Panic(err, "Error cleaning asset.")
```
celgemark: This causes a test to fail and I could need some advice on how to solve. Should we really ignore cleaning failures?
Do I understand the purpose of the test TestCleanPointer correctly that it verifies that we should do nothing when cleaning a file with content matching a cleaned pointer? Do we need this check? If so should we perhaps add that check explicitly and not bypass all clean errors?

technoweenie: The same check is [a few lines below this](https://github.com/github/git-lfs/blob/923be66fdd733be438a3fbdd1eb1201ec24bda0c/commands/command_clean.go#L57-L59).  But first, it attempts to close the cleaned file stream.  It also checks for a very specific type of error (`*lfs.CleanedPointerError`), before handling other errors.

celgemark: Thanks, I noticed that as well so I deleted my pull request. Should have looked more in detail after my rebase of this change. I will be more careful next time. Thank you for a great tool. I really like git-lfs!

PR #303, opened 2015-05-13
Use goimports instead of gofmt

technoweenie:
This updates `script/fmt` to use `goimports` instead of the built-in `gofmt`.  The main benefit here is that it organizes the import statements with stdlib packages above the 3rd party libs.  The main change is in [`script/fmt`](https://github.com/github/git-lfs/compare/goimports?expand=1#diff-f9a8d86d11fc6a80a862480ab36f3941) itself:

```diff
+hash goimports 2>/dev/null || {
+  echo "Installing goimports"
+  echo "go get golang.org/x/tools/cmd/goimports"
+  go get golang.org/x/tools/cmd/goimports
+}
+
 # don't run gofmt in these directories
 ignored=(/bin/ /docs/ /log/ /man/ /tmp/ .vendor)
 for i in */ ; do
   if [[ ! ${ignored[*]} =~ "/$i" ]]; then
-    gofmt -w -l "$@" "${i%?}"
+    goimports -w -l "$@" "${i%?}"
   fi
 done
```

Is this valid, or should this fall back to `gofmt`?  `gofmt` respects the import style that `goimports` forces, so we won't have to worry about competing code formatting tools.

### DIFF
--- a/commands/command_clean.go
+++ b/commands/command_clean.go
@@ -1,9 +1,10 @@
 package commands
 
 import (
+	"os"
+
 	"github.com/github/git-lfs/lfs"
 	"github.com/spf13/cobra"
-	"os"
 )
 
 var (
--- a/commands/command_logs.go
+++ b/commands/command_logs.go
@@ -2,11 +2,12 @@ package commands
 
 import (
 	"errors"
-	"github.com/github/git-lfs/lfs"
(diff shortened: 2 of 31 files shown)


rubyist:
:heart:

:+1: for falling back

PR #305, opened 2015-05-14
Refactored and improved Errorhandling

michael-k:
My changes are mostly refactoring.  Non-refactoring changes are:

- Don't return a file name we didn't write to eb12f25
- Write error to stderr, even if os.MkdirAll fails e8c6175

The cases where `os.MkdirAll` and `os.Create` fail, are (still) handled differently and I'm not sure if this was intended in the first place. (The previous code didn't make it that obvious.) Please review.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -106,33 +106,31 @@ func handlePanic(err error) string {
 		return ""
 	}
 
-	return logPanic(err, false)
+	return logPanic(err)
 }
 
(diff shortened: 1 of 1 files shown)


technoweenie:
Looks good to me.  :+1:

Thanks for looking though this.  When we started this repository over a year ago, we were very new to Go.  This command error logging code is probably the oldest code in here :)

PR #291, opened 2015-05-08
Integration Tests 2.0

technoweenie:
Our integration tests have a number of problems that make it difficult for new contributors to add commands.  The tests are a mix of explicit Go code, and a non-trivial test helper for setting up sample repositories.  They give me confidence to changes to [simple commands](https://github.com/github/git-lfs/blob/master/commands/pointer_test.go), but not for the important ones that communicate with external servers.

https://github.com/github/git-lfs/pull/280#issuecomment-99766136 really highlights how this friction impacts changes to this young project.

PR #306, opened 2015-05-14
Integration tests 3 (one more time with feeling)

technoweenie:
Code spike to see if https://github.com/github/git-lfs/pull/291 would work better with shell scripts.  The main rationale is that these aren't Go tests, so running them with the "testing" package isn't necessary.  Some of the tests will rely heavily on `os.Chdir()`, which affects the entire process.  Even with goroutines, there's no guarantee that parallel tests won't run in the same process.  Just running #291 with any kind of concurrency leads to weird errors.

Hopefully we can get around cross platform shell issues by using the common commands supported everywhere, and implement others in tiny Go commands.  For instance, a command like `lfstest-gitserver` could spin up an HTTP server that responds to Git and Git LFS calls that tests use.


test/cmd/lfstest-gitserver.go
```diff
@@ -0,0 +1,235 @@
...
+
+	w.WriteHeader(202)
+	w.Write(by)
+}
+
+func lfsGetHandler(w http.ResponseWriter, r *http.Request) {
+	parts := strings.Split(r.URL.Path, "/")
+	oid := parts[len(parts)-1]
+
+	if by, ok := largeObjects[oid]; ok {
```
michael-k: I'd prefer
```go
by, ok := largeObjects[oid]
if !ok {
    w.WriteHeader(404)
    return
}
```

test/cmd/lfstest-gitserver.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// handles any requests with "{name}.server.git/info/lfs" in the path
+func lfsHandler(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/vnd.git-lfs+json")
+	if r.Method == "POST" {
+		lfsPostHandler(w, r)
+	} else {
+		lfsGetHandler(w, r)
+	}
```
michael-k:
```go
switch r.Method {
case "PUT":
    lfsPostHandler(w, r)
case "GET":
    lfsGetHandler(w, r)
default:
    w.WriteHeader(405)
}
```

test/cmd/lfstest-gitserver.go
```diff
@@ -0,0 +1,235 @@
...
+		oid := parts[len(parts)-1]
+
+		if by, ok := largeObjects[oid]; ok {
+			w.Write(by)
+			return
+		}
+
+		w.WriteHeader(404)
+	default:
+		w.WriteHeader(500)
```
michael-k: Why not `405` (Method Not Allowed)?

technoweenie: Honestly, because it's a test server and I don't really care.  Though I suppose the first time I hit that bug in testing, I'll curse myself for not using something more descriptive like 405 :)

test/cmd/lfstest-gitserver.go
technoweenie: Good call.  I like early returns like that.

technoweenie:
This has progressed from a code spike to a PR that's ready for a review. I want to focus on the core test suite with the included "happy path" test as an example.  After this is merged, I'll work on porting existing tests from the `commands` package to this new format.

While I'm waiting for all of the glorious reviews to come in, I'm going to work on a `test/README.md` doc that explains how tests work at a high level.

@michael-k: I made all of your suggested improvements.

technoweenie:
I updated test/README.md and wrote some better docs in the shell scripts.  It should be a good starting point.

michael-k:
I'm being asked for username and password. o.O
```sh
$ test/test-happy-path.sh
compile git-lfs for test/test-happy-path.sh
LFSTEST_URL=/…/git-lfs/test/remote/url LFSTEST_DIR=/…/git-lfs/test/remote lfstest-gitserver
Username for 'http://127.0.0.1:60114': 
Password for 'http://127.0.0.1:60114': 
Username for 'http://127.0.0.1:60114': 
Password for 'http://127.0.0.1:60114': 
test: happy path ...
```
`LFSTEST_*` shortended by me.

When hitting `Crtl+C`, the next test fails:
```sh
test/test-happy-path.sh
Username for 'http://127.0.0.1:34260': 
Password for 'http://127.0.0.1:34260': 
Username for 'http://127.0.0.1:34260': 
Password for 'http://127.0.0.1:34260': 
test: happy path ...                                               FAILED
-- stdout --
    set up remote git repository: test-happy-path
    Reinitialized existing Git repository in /…/git-lfs/test/remote/test-happy-path.git/
    clone local git repository test-happy-path to repo
    Cloning into 'repo'...
    Downloading a.dat (1 B)
-- stderr --
    + set -e
    + basename test/test-happy-path.sh .sh
    + reponame=test-happy-path
    + setup_remote_repo test-happy-path
    + local reponame=test-happy-path
    + echo set up remote git repository: test-happy-path
    + repodir=/…/git-lfs/test/remote/test-happy-path.git
    + mkdir -p /…/git-lfs/test/remote/test-happy-path.git
    + cd /…/git-lfs/test/remote/test-happy-path.git
    + git init --bare
    + git config http.receivepack true
    + git config receive.denyCurrentBranch ignore
    + printf [filter "lfs"]
    	required = true
    	smudge = %s smudge %%f
    	clean = %s clean %%f
    [credential]
    	helper = %s
    [remote "origin"]
    	url = %s/%s
    	fetch = +refs/heads/*:refs/remotes/origin/*
     /…/git-lfs/bin/git-lfs /home/tux/repos/golang/src/github.com/github/git-lfs/bin/git-lfs lfstest http://127.0.0.1:34260 test-happy-path
    + clone_repo test-happy-path repo
    + cd /…/git-lfs/tmp/test-happy-path.sh-11398
    + local reponame=test-happy-path
    + local dir=repo
    + echo clone local git repository test-happy-path to repo
    + GIT_CONFIG=/…/git-lfs/test/remote/config-test-happy-path git clone http://127.0.0.1:34260/test-happy-path repo
    + out=Cloning into 'repo'...
    Downloading a.dat (1 B)
    + cd repo
    + git config credential.helper lfstest
    + echo Cloning into 'repo'...
    Downloading a.dat (1 B)
    + /…/git-lfs/bin/git-lfs track *.dat
    + out=*.dat already supported
    + + echo *.dat already supportedgrep
     Tracking \*.dat
    + test_status=1
```

After `rm test/remote/url` the test passes again.

technoweenie:
> I'm being asked for username and password. o.O

Ah, what's your git config look like?  The tests set `credential.helper` to the custom `git-credential-lfstest` command that the tests build.  The config is set on in the [test repo's local git config](https://github.com/github/git-lfs/blob/904145fcceef79f11b1da9bcf3829edb5055d8a0/test/testhelpers.sh#L107). Since clones have no local git config, the tests [generate a little git config](https://github.com/github/git-lfs/blob/904145fcceef79f11b1da9bcf3829edb5055d8a0/test/testhelpers.sh#L88-L89) file that it passes with `GIT_CONFIG` to all git clone commands.

Can you drop a `git config -l` towards the top of `test-happy-path.sh`?  Just hit enter to clear the prompts and see what's in your local git config (probably after [this](https://github.com/github/git-lfs/blob/904145fcceef79f11b1da9bcf3829edb5055d8a0/test/test-happy-path.sh#L19).  Maybe there's something else that we need?  

I think one solution would be to pass `GIT_CONFIG` to every Git command, but I like that they can [use regular git commands](https://github.com/github/git-lfs/blob/904145fcceef79f11b1da9bcf3829edb5055d8a0/test/test-happy-path.sh#L28-L36).  

> When hitting Crtl+C, the next test fails

This is an issue with the way tests are setup.  Test files check for `test/remote/url` to see if the tests have been bootstrapped by something else.  This is admittedly a lame solution, but it keeps us from having to recompile for every test file when running the suite with `script/integration`. I tried explaining this more in [test/README.md](https://github.com/github/git-lfs/blob/integration-tests-3/test/README.md#integration-tests) and I'm very open to suggestions here.

rubyist:
I can reproduce the username/password request by globally unsetting `credential.helper` first.

```
scott@ubuntu:~/git-lfs$ git config credential.helper
cache
scott@ubuntu:~/git-lfs$ ./test/test-happy-path.sh 
compile git-lfs for ./test/test-happy-path.sh
LFSTEST_URL=/home/scott/git-lfs/test/remote/url LFSTEST_DIR=/home/scott/git-lfs/test/remote lfstest-gitserver
test: happy path ...                                               OK
scott@ubuntu:~/git-lfs$ git config --global --unset credential.helper
scott@ubuntu:~/git-lfs$ ./test/test-happy-path.sh 
compile git-lfs for ./test/test-happy-path.sh
LFSTEST_URL=/home/scott/git-lfs/test/remote/url LFSTEST_DIR=/home/scott/git-lfs/test/remote lfstest-gitserver
Username for 'http://127.0.0.1:58629': 
```

Other than that, I think it looks great and I can't wait to start writing some tests for batch/concurrent using this.

rubyist:
On the [second clone](https://github.com/github/git-lfs/blob/904145fcceef79f11b1da9bcf3829edb5055d8a0/test/test-happy-path.sh#L56) into the `clone` dir, it needs to download lfs assets, which happens before setting the credential helper config. I think the only way we can handle that is by using the `GIT_CONFIG` env variable in those instances.

technoweenie:
I reworked the test to clone the initial empty repo twice.  The test then uses `cd` to change to the empty clone, and runs `git pull`.  @michael-k: does this work for you now?

This is a quick fix.  I think the correct thing is to somehow respect `GIT_CONFIG`.

michael-k:
Yes, `test/test-happy-path.sh`, `script/integration` and `script/cibuild` work without any username/password prompt.

I noticed, that some stuff is left behind in `test/remote`. I have some questions regarding that stuff:
- Is it reused in the next run?
- Are there any side effects between two test runs?
- Is there any cleanup script (planned)?

Sorry, atm I don't have the time to dig into this myself.

technoweenie:
Cool, thanks for confirming.  I'm going to merge this and start porting tests over.

`test/remote` is cleared during the setup process that compiles things.  So it's run either when `test/test-*.sh` files are run individually, or once at the beginning of `script/integration`.  

While there are no side effects between two runs, the `test/remote` directory (and the test git server that's booted up to expose it over HTTP) is "global" to all tests.  So different tests should use unique names.  The happy-path test uses `reponame="$(basename "$0" ".sh")"` to ensure it's unique from other test files.

The files are in .gitignore, so I don't know of any issues with leaving them there.  They could come in handy if you're troubleshooting some weird test failure.  Though that doesn't help us on cloud ci tools, and even then only very rarely. I may just clean them up now until it's proven that we should keep them around.

michael-k:
> (…) `test/remote` directory (and the test git server that's booted up to expose it over HTTP) is "global" to all tests.

Yeah, I read about that in the `README.md`. :+1:

> The files are in .gitignore, so I don't know of any issues with leaving them there.

No issues.  But it would be nice to have some script (e.g. `script/clean`) to clean them up.  There might be more of them in the future and in different places.  I'd say it's common practice for Makefiles to have a cleanup target, so it should be for tests too.

> They could come in handy if you're troubleshooting some weird test failure.

Good point.

> I may just clean them up now until it's proven that we should keep them around.

Maybe add an command line option to leave them behind?

technoweenie:
@michael-k: I've updated the suite to remove the files unless $KEEPTRASH is set.

technoweenie:
Unfortunately, I had to bring the explicit credential helper back due to unreliable CI builds:

* https://travis-ci.org/github/git-lfs/builds/63519523
* https://travis-ci.org/github/git-lfs/builds/63519527

I can retry them and get the failed ones to eventually pass, and vice versa.  On a whim, I pushed 3bdfa6ca672af4329b739603d0883b3ede79ed14 to separate branch, and got 5 passes in a row.  I'm merging this now.  Though there is an opportunity for someone to either write a smaller test credential helper in shell, or figure out why the inline helper was failing sometimes.

PR #309, opened 2015-05-15
Add debian package

ssgelm:
This adds support for building a debian package.  I believe there should be some documentation on building the package still.  Where would you like that to go?

Package can be built by running `dpkg-buildpackage -us -uc` from the root of the repo.  It is currently confirmed to work on Debian jessie and wheezy. On wheezy it requires `wheezy-backports` versions of `dh-golang`, `git`, and `golang`.


technoweenie:
What do you think about adding the docs to [docs/linux-build.md](https://github.com/github/git-lfs/blob/master/docs/linux-build.md)?  It's pretty short now, so a Debian section should be fine.  If not, I'd add it to docs/debian-build.md.

technoweenie:
@jsh: I'm thinking that we'd keep your [`script/debian-build`](https://github.com/github/git-lfs/blob/master/script/debian-build), and add a `script/debian-package` for this.  Unless you think debian users won't ever want to build manually once there's an official package available.

ssgelm:
Yeah, docs/linux-build.md seems reasonable to me.  I will add that now.

ssgelm:
Docs added.

jsh:
Rick,

I think keeping a script can't hurt. After all, there's a Makefile for git,
even though most folks install a package. :-)

On Fri, May 15, 2015 at 9:58 AM, risk danger olson <notifications@github.com
> wrote:

> @jsh <https://github.com/jsh>: I'm thinking that we'd keep your
> script/debian-build
> <https://github.com/github/git-lfs/blob/master/script/debian-build>, and
> add a script/debian-package for this. Unless you think debian users won't
> ever want to build manually once there's an official package available.
>
> —
> <https://github.com/github/git-lfs/pull/309#issuecomment-102445748>.
>



-- 
Jeffrey Haemer <jeffrey.haemer@gmail.com>
720-837-8908 [cell], http://seejeffrun.blogspot.com [blog],
http://www.youtube.com/user/goyishekop [vlog]
*פרייהייט? דאס איז יאַנג דינען וואָרט.*

technoweenie:
@ssgelm: Apologies that it took so long to do a proper review. Does this need anything else (besides conflict resolution)?

I have a few minor concerns:

* The `override_dh_auto_build` function in `debian/rules` symlinks the files manually.  Is there a reason you can't use `script/bootstrap`? This project's structure is a little odd for Go programs, so we may be making some changes.  Though we'll probably have to tweak these debian files regardless.
* Is the top level `./debian` directory required?  I'd like a more organized place for the various scripts as we support more package managers.

ssgelm:
@technoweenie: It's possible that `script/bootstrap` could be used.  I'll take a look.  Unfortunately though the top level debian directory is required. :disappointed:

technoweenie:
@ssgelm: In that case, I'll just merge this.  It's very possible I'm going to be changing the project structure so that the symlinks are not needed anymore.  I'll ping you on that PR for help updating the scripts if/when that happens.

ssgelm:
Awesome! Sounds good.

PR #311, opened 2015-05-15
Traced performance in case of error too

michael-k:
This also keeps the usages of variable 'start' close together.


rubyist:
:+1: looks good, thanks!

PR #313, opened 2015-05-18
Building on CentOS

andyneff:
The build script for CentOS could not build the man pages on CentOS 6 since ruby 2 is not available. However, there is a way to get ruby via the RVM. That has been added in so that CentOS 6 can build the man pages using this script

Getting golang on CentOS 5 isn't really a "supported" idea, but I looked into a number of work arounds and settled on the simplest one. I saw it mentioned in the comments and gave it a try. If you would rather say NO to CentOS 5, I can remove that, but I can also say that this will build and run git-lfs on a a CentOS 5 Machine (Tested on a 5.11 VM).

Also add the ability to use a currently cloned version of git-lfs, so it does not always clone into /tmp

### DIFF
--- a/script/centos-build
+++ b/script/centos-build
@@ -1,7 +1,9 @@
-#!/bin/bash -eu
+#!/usr/bin/env bash
 #
 # This script works with CentOS 6 or 7
 # The CentOS 5 kernel is too old for go's liking.
+
+set -eu
 
 trap 'echo FAIL' ERR
 
@@ -9,7 +11,7 @@ if grep -q ' 6' /etc/redhat-release; then
   rpm -q epel-release || rpm -Uvh http://download.fedoraproject.org/pub/epel/6/x86_64/epel-release-6-8.noarch.rpm
 fi
 
-yum install -y bison git golang make man
+yum install -y bison git golang make man which
 cd /tmp
 [ -d git-lfs ] || git clone https://github.com/github/git-lfs
 cd git-lfs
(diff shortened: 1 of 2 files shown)


technoweenie:
As someone that doesn't use CentOS, and is not even much of a devops person, PRs like this are difficult to review.  I'll probably ping you on any issues we get on these scripts, so I'd love any further help you can provide.  I think eventually I'd like to setup some VMs for running QA checks on stuff like this before we release.

/cc @jsh, author of the CentOS install script.

andyneff:
@technoweenie Not a problem. Just let me know when you got an issue, and I'll see if I can help. CentOS is my primary OS of choice ;) So I'm definitely testing that!

technoweenie:
@andyneff: can you confirm this works with master?  We've made some major changes (see #345). Looks like it should just work though.

andyneff:
@technoweenie You'll have to forgive me relative lack of experience with go, but... Now that you've removed the symlink, how is this supposed to build now?

on 47f735098b626d3d5536f907adbdc88a590e8170, ```scripts/bootstrap``` fails stating

```
script/build.go:13:2: cannot find package "github.com/github/git-lfs/lfs" in any of:
        /usr/lib/golang/src/github.com/github/git-lfs/lfs (from $GOROOT)
        ($GOPATH not set
```

which is the ```import ("github.com/github/git-lfs/commands")``` line. So... where is that import?


Centos 7.1.1503 (should be the easiest case)

technoweenie:
Gah, I should add this to the readme or contributing guide. Go projects expect you to clone the repository to `$GOPATH`. Its directory tree should look like this:

```
- src
  - github.com
    - github
      - git-lfs
- pkg
  (go build artifacts)
- bin
  (compiled go apps)
...
```

I just use `$HOME/go` as my GOPATH.

technoweenie:
Here's a great way to get started hacking on Git LFS, and really any Go project:

1. Install Go 1.3+
2. Make sure `$GOPATH` is set (I use `$HOME/go`).
3. Run `go get github.com/github/git-lfs`.
4. Optional: Symlink `$GOPATH/src/github.com/github/git-lfs` to your project directory if you have one.

andyneff:
Well... Thank you for the hand holding, I needed it apparently... let me just ~~rant~~ say that IS a paradigm I'm not used to, the repo != the build structure... And using golang like that enforces that... (by checking out the repo into this other build structure)... I do not like... but that is not a discussion for git-lfs, it looks like that is golang's fault. 

So this brings me to the inevitable question now, what is the purpose/scope of these scripts? Because BEFORE, they had two features (intended or not)

1) I could ```git clone``` the repo, and run that script, and everything builds
2) I could download JUST the script, it clones the repo for me, and everything builds

But as I see it now, the only course of action is to
a) Follow the 5 steps @technoweenie Gave me, and then run it and it will work

Which makes lines 37-39 pointless (in their current state)

So what do you want this to do now?

technoweenie:
You're right, the new setup is a bit of a regression. Getting this to actually _work_ on Windows (no more symlinking) is worth it IMO.

I would say option 1 is the primary goal of these scripts.

andyneff:
@technoweenie Ok, I have 1 working again. I went ahead add took your suggested optional step 4 and added a symlink myself. I guess whomever writes a similar windows script could do the same thing, but with a junction? Just guessing. At any rate, it all works now.

PR #315, opened 2015-05-19
Fix spec error - there is no 'git lfs sync' command

sinbad:


technoweenie:
Wow, you found an oldie :)

sinbad:
Super-trivial, but I thought I'd start with something easy :)

PR #323, opened 2015-05-21
Track tests

technoweenie:
This moves the track command's tests to the shell scripts.  It also fixes #294, taking the feedback into account.


test/test-track.sh
```diff
@@ -0,0 +1,90 @@
...
+
+  # no need to setup a remote repo, since this test doesn't need to push or pull
+
+  mkdir track
+  cd track
+  git init
+
+  # track *.jpg once
+  git lfs track "*.jpg" | grep "Tracking \*.jpg"
+  numjpg=$(cat .gitattributes | grep "\*.jpg" | wc -l)
```
michael-k:
```sh
grep "\*.jpg" .gitattributes
```

test/test-track.sh
```diff
@@ -0,0 +1,90 @@
...
+  if [ "$(printf "%d" "$numjpg")" != "1" ]; then
+    echo "wrong number of jpgs"
+    cat .gitattributes
+    exit 1
+  fi
+
+  # track *.jpg again
+  git lfs track "*.jpg" | grep "*.jpg already supported"
+  numjpg=$(cat .gitattributes | grep "\*.jpg" | wc -l)
```
michael-k: See line 17

PR #324, opened 2015-05-21
grep doesn't need cat

michael-k:


technoweenie:
Good call. I just got used to the `cmd | grep` pattern because the test output shows the actual value.

```
# THE TEST
out=$(git lfs track)
echo "$out" | grep "Listing tracked paths"

# TEST OUTPUT
    + echo 'Listing tracked paths
        *.mov (.git/info/attributes)
        *.jpg (.gitattributes)
        *.gif (a/.gitattributes)
        *.png (a/b/.gitattributes)'
    + grep 'Listing tracked pathss'
```

It only makes sense for `echo` though. Grepping files should probably look something like this so we can see what was actually in the file in the case of a failure on Travis:

```
grep "sub module" README || {
  cat README
  exit 0
}
```

We could turn that into a `file_grep` helper function or something too.

michael-k:
Why `exit 0` instead of `exit 1`? If grep fails, the test should fail.

technoweenie:
Ha, typo.  `exit 0` is redundant there, assuming README is a valid file.

technoweenie:
Some test refactoring made this conflict. But, the [end result](https://github.com/github/git-lfs/blob/4df64ebf8036277899617fc93f6492440316b000/test/test-submodule.sh#L28-L31) is the same. Thanks for the patch!

PR #329, opened 2015-05-25
Fix charset http Content-Type header

joerg:
I tried implementing an LFS Server in JAVA and came up with this issue. According to https://tools.ietf.org/html/rfc5987, it has to always be "key = value". I also tried this change and my implementation now works (or at least I am one small step further).

### DIFF
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -18,7 +18,7 @@ import (
 )
 
 const (
-	mediaType = "application/vnd.git-lfs+json; charset-utf-8"
+	mediaType = "application/vnd.git-lfs+json; charset=utf-8"
 )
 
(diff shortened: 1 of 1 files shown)


rubyist:
That dash is definitely a typo on our part, thanks for the fix! If you sign the CLA we can merge it in.

joerg:
Nice, thanks.
Btw: Read and followed the CONTRIBUTING.md there is no word about the CLA in it. Thats why I did not sign it up front. ;)

rubyist:
That's good feedback, we should definitely mention it in there. I think the only way one finds out about it is the link in a "failing" build after a PR is submitted. Thanks again!

PR #330, opened 2015-05-25
Godeps

technoweenie:
Updates the project to use Godeps instead of gpm.  Main goal is to get this project building with standard go tools and without [symlinking](https://github.com/github/git-lfs/blob/8ce6043f3d3f3a680b013ffe12b7a9c88400a89e/script/bootstrap#L11). This reduces the friction for new coders to get up and going, especially Windows go developers.


technoweenie:
Went with Nut instead.  #331.

PR #331, opened 2015-05-25
setup nut for dependency management

technoweenie:
Trying Nut out.  I like the simpler config file. Unfortunately, it doesn't seem to be rewriting the imports for me.  https://github.com/jingweno/nut/issues/24


technoweenie:
This is ready to go.  Changes:

* All 3rd party imports in the Git LFS code and the vendored libs should have a "github.com/github/git-lfs/vendor/nut/" prefix. 
* Somehow the "github.com/kr/pty" package was never vendored. It is now.
* The various scripts are now _much_ simpler.  Should work with vanilla go commands, even on Windows.
* Forked "github.com/bmizerany/assert" to fix an import statement in an example script.  This gets around [this issue](https://github.com/jingweno/nut/issues/24).
* There are references to "github.com/ogier/pflag". I think we need to clean up some references in some of spf13's packages upstream.

Why nut? Mostly because it worked out of the box. Godep didn't work quite right, and [required a hack](https://github.com/github/git-lfs/blob/godeps/git-lfs.go#L6-L8) to get around an issue. We've also had great interactions with Nut's author, @jingweno in past Go projects.  He [mentioned](https://github.com/jingweno/nut/issues/24#issuecomment-105285821) an intent to support the [suggested Go import conventions](https://groups.google.com/forum/#!topic/golang-dev/nMWoEAG55v8%5B1-25%5D) with a [community specified vendor file format](https://github.com/kardianos/vendor-spec).

sinbad:
Looks fine to me. I never had the issue you encountered with godep (didn't have any platform-specific imports) and found it worked fine with updates over time - I see nut update isn't implemented yet - but it's all much the same. Main thing is the rewriting of the imports so it'll build out of the box which is a big plus.

rubyist:
Looks good, a couple notes:

Right now nut doesn't have an option to just rewrite imports (https://github.com/jingweno/nut/issues/8). This means manually typing the full vendored import path, and tools like `goimports` will no longer be able to help with that. If nut had this, for example, that could be part of the editor's formatting command (or `script/fmt`), run after `goimports`. This isn't a deal breaker, just something we might trip over.

Updating support would be nice, but working around that is pretty much what we had to do with gpm, so not a big deal (actually seems a little easier, I always had problems with gpm and .git dirs).

Running `nut install` in a freshly checked out version of this branch I get this every time:

```diff
diff --git a/vendor/_nuts/github.com/olekukonko/ts/doc.go b/vendor/_nuts/github.com/olekukonko/ts/doc.go
index d73c57e..50c63ca 100644
--- a/vendor/_nuts/github.com/olekukonko/ts/doc.go
+++ b/vendor/_nuts/github.com/olekukonko/ts/doc.go
@@ -20,7 +20,7 @@ Example

        import (
                "fmt"
-               "github.com/github/git-lfs/vendor/_nuts/github.com/olekukonko/ts"
+               "github.com/olekukonko/ts"
        )

        func main() {
```

This is code in a commented block, I'm thinking it got rewritten on accident and nut is just putting it back to normal, though I would think that no tool should be rewriting imports in comments. I think it should be safe to put that back as the normal import so we don't have to fix it every time we run nut.

technoweenie:
@rubyist: Ah, that might've been rewritten by me originally, and nut is just fixing it.  I didn't realize it was in a giant comment block.

rubyist:
I went back and looked at the commits and realized that's probably what happened :+1:

jingweno:
Thanks for using `nut`! I'll prioritize https://github.com/jingweno/nut/issues/8 higher since quite a few people ask for it.

jadengore:
@technoweenie How do you fix the `assert` issue? I saw your fork but I'm not sure how to integrate!

Issue #328, opened 2015-05-24
RPM installs


andyneff:
I'm working on some rpm install for git-lfs, (Currently under andyneff/git-lfs@91b185b435e382a69598f61e6df8193ad4c1faf7) I eventually want to get it to work for the supported CentOS's (5, 6, 7), Fedora Versions as of July 2015 ( 21, 22, since 20 will be EOL by July), and maybe if I can EVER get to it, OpenSUSE... although I've never tried that or SLES. Any comments on that?

When I am done with this, would you like me to PR? If so, do you want a PR as soon as I am done with CentOS (this weekend I hope?) or just wait until all the stretch goals are done and do one PR?

(A little background of rpms, they are built using the rpmbuild command, so as you update your own build scripts, etc... only the few lines in the "spec" file need be updated, and rpmbuild will always be called to build the actual rpms.)

Side-note: Given the static nature of go executables, I can probably even make a generic set that can be compiled on any OS (since rpm is actually available on most OSes including Debian, etc...) so that as long as you build on a 64 bit ELF OS. Advantage is you can build the rpms from whatever Linux OS you are using for the releases, and Redhat OSes can install them... I think it will work?

technoweenie:
I'd love to see rpm builds for Git LFS.  Anything that makes Git LFS easier for people to install and use, AND can be eventually automated on our end through cloud build servers or something. Smaller PRs are nice. :metal:

technoweenie:
Thanks so much for taking this on!

PR #332, opened 2015-05-26
RPMs for Centos 5/6/7

andyneff:
A set of RPMs that will build git-lfs on CentOS 5, 6, and 7 for #328. A build_rpm.bsh script is included to basically automate installing (and build in the case of CentOS 5 and 6) the build dependencies, and build the git-lfs rpm and srpm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ debian/*.log
 debian/files
 debian/*.substvars
 obj-*
+
+rpm/BUILD*
+rpm/*RPMS
--- a/rpm/INSTALL.md
+++ b/rpm/INSTALL.md
@@ -14,8 +14,12 @@ and install a suitable golang/ruby so that git-lfs can be built.
 Simple run:
 
 ```
+./clean.bsh
 ./build_rpms.bsh
 ```
+
+The clean.bsh script removes previoeu rpms, etc... and removed the source tar.gz
+file. Otherwise you might end up creating an rpm with pieces from different versions
 
 Practice is to run rpmbuild as non-root user. This prevents inadvertently installing
 files in the operating system. The intent was to run build_rpms.bsh as a non-root user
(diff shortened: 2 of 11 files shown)


technoweenie:
Instead of requiring ruby for people downloading rpms, is there a way that the rpm build script could generate them locally and have the rpm just bundle them?

Also, we're looking at changing how Go dependencies are managed (#330 or #331), which could change the project structure. It looks like the script calls `script/bootstrap` which means nothing should change. Just FYI.

andyneff:
Ruby is NOT required for people downloading the rpm, it is only required to build the RPM from source (hence the BuildRequires:) It will never show up as a problem/dependency for someone downloading and using the final rpm. I only added ALLLLL the others for whomever chooses to build the RPMs for completeness.

It generates separate rpms for ruby/the gems/go ONLY for the purpose of building the final git-lfs. So in essence they are creating their own bundles (and installs them) just for the person creating the rpms (at release time... or for that enthusiastic user out there to DIY). This is best way to fit into how redhat does things.

If you do not like this, what I CAN look into is instead of installing the ruby/gem/go in the the system (prefix /) I can make the build script install them into the local prefix.

The person building the rpms also has the option of having thier own ruby/go setup, and just calling the rpmbuild command (really, the last line of the build_rpm script)  with --nodeps.

These are all things I want to cover in the README I have not written yet.

technoweenie:
@andyneff: can you confirm this works with master? We've made some major changes (see #345).

andyneff:
I will try and confirm tonight (and add an option to run off of currently checked out version instead of ONLY a tar ball)

andyneff:
@technoweenie, with these new changes, everything works for building RPMs against the new master from yesterday, and I added running scripts/test to the rpm (to be run under the %check section)

PR #336, opened 2015-05-26
Port tests

technoweenie:
This moves all the tests from the `commands` package to the new shell tests. 

* [x] clean
* [x] env
* [x] init - I removed these tests because they test that the global git config is updated. I don't want tests to modify existing global git configs.
* [x] ls-files
* [x] pointer
* [x] pre-push
* [x] push
* [x] smudge
* [x] status
* [x] update
* [x] version - also removed, since every test run calls this


michael-k:
I'm not happy with this.  Shell scripts are horrible in many ways.  Some arguments, why we shouldn't rely on shell scripts:
* It should be fun to write tests, with shell scripts (for integration tests), it's not.
* There is no way to see any coverage.
* Learning shell scripting might be to big an obstacle for some developers.

If there's no way to make the `*_test.go` files work, I'd vote in favor of separate go files (e.g. one per integration test), executed with `go run filename.go`.  We could still use `exec` where necessary and would be close to the functionality of shell scripts, but the code would be in go.

technoweenie:
Honestly, I'm not a huge fan of the shell tests either. I'd be down for support both sh and go tests so we can experiment with some single file go tests like you suggested. If we find a pattern that works, we can then move tests back to go one by one. But there's nothing about the [current Go test stuff](https://github.com/github/git-lfs/pull/336/files#diff-b65ac838606a4ab61fdb7724d458246d) that I want to keep.

technoweenie:
FWIW here's another attempt at Go tests containing simpler declarative statements like the shell tests. 

https://github.com/github/git-lfs/pull/291/files#diff-85b5fd21f7c36d3ba95b3a6fc0225941

I'm merging this so we can move forward on some other PRs. Still open to the idea of exploring single file go tests, though.

michael-k:
:+1:

Issue #333, opened 2015-05-26
I couldn't push code into github with git-lfs hook on git push


Sobrr:
git-lfs is currently not completely supported by github (only allow part of the users). Now I couldn't push any of my changes because of that.

"
You're signed up for early access and invitations will be going out soon. We're excited about Git LFS too!
Docs: https://github.com/contact
error: failed to push some refs to 'https://github.com/Sobrr/iOS.git'
"

After I tried to uninstall git-lfs so that I can push my files in, there is this error. 
"git: 'lfs' is not a git command. See 'git --help'."

Where do I change the push command to the old one and bypass the git lfs command?

rubyist:
Sorry for the trouble, you should be able to fix your pushes by removing the `.git/hooks/pre-push` file from your repository. This is where git is trying to run `git lfs` when you push.

We'll add some more documentation around uninstalling Git LFS as this is not very clear. Feel free to close this issue if this solves your problem, or if you run into any other problems just let me know how I can help. Thanks!

technoweenie:
Blah, this sucks. Git LFS will put that file back. @sobrr: If it keeps happening, you can probably get away with throwing a dummy pre-push hook in there.

However, why is Git LFS trying to push anything? It shouldn't even try to contact the Git LFS API unless you've committed some Git LFS pointers in that repo.

rubyist:
The problem, as I understand it, is that the git-lfs binary is uninstalled. The pre-push hook will always run, and if the git-lfs binary isn't there it will error out.

technoweenie:
Oh right right. I was thinking of the larger issue. You shouldn't have to uninstall Git LFS just because not all of your repo remotes have support for it.

rubyist:
This problem is fixed as of #339, a new push hook is installed that will allow git to continue pushing when the git-lfs binary is missing. Thanks for the report!

shadowwalkersb:
I was having problems pulling even after deleting ```.git/hooks/pre-push```. Then, I removed the global git configs that Git LFS sets up following instructions in #316 . I am posting this, in case it is not addressed with #339 .

Sobrr:
Hi @technoweenie, can you enable LFS for this account?

technoweenie:
@Sobrr Can you open a new issue with details on your problem? It looks like Git LFS is enabled, but you're hitting some error that I can't see in the screenshot. I'm locking this old thread issue from further updates though since it's a very different problem.

technoweenie:
[...]

PR #339, opened 2015-05-27
Make the pre-push hook check that git-lfs exists first

rubyist:
If git-lfs does not exist (e.g. it has been uninstalled) the pre-push hook should still exit cleanly so the push can continue. The script will output a warning stating that the repo has been set up for Git LFS but that it is not installed.

/cc #333


technoweenie:
:+1: Though unfortunately the command is now complex enough that we can't just copy and paste it around the tests.

PR #345, opened 2015-05-28
update build scripts

technoweenie:
We've changed how go dependencies are managed in Git LFS (#331). Hopefully the various build scripts still work ;) Two main differences:

* `script/bootstrap` no longer symlinks the root dir into `.vendor/src/github/git-lfs`.
* The main entry point is now `./git-lfs.go`, not `cmd/git-lfs/git-lfs.go`.

@ssgelm: I removed the symlinking from the debian package rules. 

@jsh: Since the `script/debian-build` and `script/centos-build` both use `script/bootstrap`, I think they're good to go.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -15,9 +15,6 @@ override_dh_clean:
 	dh_clean
 
 override_dh_auto_build:
-	for dir in .vendor/src/github.com/*; do \
-		ln -s ../../../$$dir $(BUILD_DIR)/src/github.com/; \
-	done
 	dh_auto_build
 	rm $(BUILD_DIR)/bin/script
 	./script/man


ssgelm:
@technoweenie this does seem to work although compiling against master currently fails with the following error:
```# github.com/github/git-lfs/test/cmd
src/github.com/github/git-lfs/test/cmd/lfstest-gitserver.go:27: main redeclared in this block
        previous declaration at src/github.com/github/git-lfs/test/cmd/git-credential-lfstest.go:23```

**Comment by technoweenie** [comment]
Ah, those are for tests only, and don't need to be compiled as part of git lfs. You should be able to build with something like:

    $ go build -o bin/git-lfs git-lfs.go

I didn't see where the build happened exactly in the debian files. The `override_dh_auto_build` only mentions a `dh_auto_build` task. Does it have something to do with that "golang" build system defined in the `%` task?

If this is a big deal, I'm pretty sure we can rename `./test` to `./_test` so that standard go commands ignore the integration tests.

**Comment by technoweenie** [comment]
So `go build` works great.  But I think that's because it's only building the `*.go` files in the root dir. I think the debian stuff is running against every `*.go` file in the project though:

```
$ go build ./...
# github.com/github/git-lfs/test/cmd
../../go/src/github.com/github/git-lfs/test/cmd/lfstest-gitserver.go:27: main redeclared in this block
	previous declaration at ../../go/src/github.com/github/git-lfs/test/cmd/git-credential-lfstest.go:23

# rename the test dir
$ git mv test _test
$ go build ./...
# it works!
```

ssgelm:
I'll see what I can do.  The dh build system is not the most straightforward...  Having said that building the package _does_ run the tests, so I'm not sure whether that affects this.

ssgelm:
This is the code for dh-golang, which handles the compilation: http://anonscm.debian.org/cgit/collab-maint/dh-golang.git/tree/lib/Debian/Debhelper/Buildsystem/golang.pm

technoweenie:
Ah, I wonder if we can add the `./test` dir to `DH_GOLANG_EXCLUDES`? `./vendor` would be good too. Running `script/bootstrap` for builds and `script/test` for tests would be fine too.

technoweenie:
Handled in #383

PR #346, opened 2015-05-28
Clean empty file

technoweenie:
Git LFS is not intended for zero length files, but it should handle them properly at least.


rubyist:
:+1:

PR #349, opened 2015-05-29
Rename `get` to `fetch`

rubyist:
Fetch is a more appropriate name for the command.


technoweenie:
:+1: Similar to `git fetch`

PR #354, opened 2015-05-30
Further improvements of LocalWorkingDir and LocalGitDir detection

### LINKED ISSUE #353: Track: Inconsistent behavior in .git directory compared to `git add`
`git add` doesn't allow you to run it when in a `.git` directory.

```
$ cd repo/.git
$ git add ../some.file
fatal: This operation must be run in a work tree
$ echo "$?"
128
```

`git lfs track` doesn't care and creates a `.gitattributes` file in `repo/.git`.

michael-k:
Exposes and fixes #353


rubyist:
:+1: thanks for the fix!

PR #358, opened 2015-06-01
If the batch operation fails, fall back to individual

rubyist:
The batch endpoint is still "experimental" and a server may not yet support it. It is off by default in the client, but if gets turned on it should fall back to individual processing if the batch processing throws an error.


technoweenie:
It'd be cool if a 404/410 response (indicating that the server doesn't even try to implement this) could turn off the `lfs.batch` config value.

lfs/client.go
```diff
@@ -595,3 +603,25 @@ func sendApiEvent(event int) {
...
+}
+
+func newNotImplError() error {
+	return notImplError{errors.New("Not Implemented")}
+}
+
+func isNotImplError(err *WrappedError) bool {
+	type notimplerror interface {
+		NotImplemented() bool
+	}
```
technoweenie: Why is this declared inside the function like this?

lfs/client.go
```diff
@@ -595,3 +603,25 @@ func sendApiEvent(event int) {
...
+	if e, ok := err.Err.(notimplerror); ok {
+		return e.NotImplemented()
+	}
```
technoweenie: Since the error itself has a static string, what about something like this:

```go
var notImplementedErr = errors.New("Not Implemented")

func newNotImplError() *WrappedError {
  return Error(notImplementedErr)
}

func isNotImplError(err *WrappedError) bool {
  if err == nil {
    return false
  }

  return err.Err == notImplementedErr
}

File: lfs/transfer_queue.go [hunk 3]
```diff
@@ -125,11 +127,17 @@ func (q *TransferQueue) processBatch() {
 
 	objects, err := Batch(transfers)
 	if err != nil {
-		q.errorc <- err
-		sendApiEvent(apiEventFail)
-		return
+		if isNotImplError(err) {
+			tracerx.Printf("queue: batch not implemented, disabling")
+			configFile := filepath.Join(LocalWorkingDir, ".gitconfig")
+			git.Config.UnsetLocalKey(configFile, "lfs.batch")
```
technoweenie: It may make sense to set `lfs.batch` to `false` to override any potential global config values.

rubyist: Makes sense, though I'm thinking that messing with `.gitconfig` is troublesome because now that file is modified and needs to be committed and pushed. I think we should set it in `.git/config`. That won't propagate, of course, but it'll get set in each repo on the first attempt.

lfs/client.go
rubyist: This is avoiding directly comparing error types, using this method http://dave.cheney.net/2014/12/24/inspecting-errors

lfs/client.go
rubyist: Because it's not needed outside the function.

lfs/transfer_queue.go
technoweenie: I think that's good. People can always set it in `~/.gitconfig` if they really want.

Though this makes me realize that the config applies to every remote. Eventually we may want also support keys like `remote.{name}.lfsbatch` so that a repository can happily use the batch API.

Considering that the batch API will likely be the ONLY API eventually, remote-specific config keys is probably too much hassle.

PR #360, opened 2015-06-01
[DO NOT MERGE] How to cache small binaries in git lfs?

larsxschneider:
Hi,

I am evaluating git lfs for large repositories (> 20GB) that contain a big number of small binary files. Our offices are distributed across the globe and the initial git clone operation is painfully slow.

__Idea__:
Setup a redis server in every office and use it to cache small files. The developers would need to configure an environment variable with the redis server IP to tell git lfs that there is a redis cache available.

__Obvious problem__:
The solution bypasses all security mechanisms. For now that's not a big problem as we run everything behind a firewall (plus the cached binaries contain no source code).

__Question__:
What do you think about this approach? Nuts? Do you see a better/easier solution?

PS: Please don't judge me by the go code attached to this PR... it's the first piece of go that I've ever written :smile:

### DIFF
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -14,6 +14,7 @@ import (
 	"regexp"
 	"strconv"
 
+	"github.com/garyburd/redigo/redis"
 	"github.com/github/git-lfs/vendor/_nuts/github.com/rubyist/tracerx"
 )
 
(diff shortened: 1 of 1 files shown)


technoweenie:
> The developers would need to configure an environment variable with the redis server IP to tell git lfs that there is a redis cache available.

This breaks one of the key principles of Git LFS: users shouldn't need to configure anything beyond their Git remote and some form of authentication (hopefully the same one their Git server takes). Also, if your set of objects is small enough to live in a single server's RAM, then you may not need Git LFS.

Did you actually benchmark this against the v0.5.1 client? I wonder how this redis cache compares to that, and how it will compare to the finished batch API #285. The batch API work is the start of an effort to reduce the Git LFS overhead, especially when there is a large number of files involved.

bkeepers:
Some kind of caching proxy server would be really nice. I'm not sure doing it purely in redis and making the client redis-aware is the right approach.

larsxschneider:
@technoweenie I benchmarked it against 0.5.1:
My repository contains 14237 files in LFS. Most of them are pretty small, less than 100kb. A git clone takes about 8.5h without Redis cache and ~3min with Redis cache. I would like to store these files in Git LFS instead of Git because they might get updated once in a while and that would blow up the Git repo. I haven't tested it with the batch API as I don't have a 0.5.2 compatible server, yet.

If I understand the code correctly, then it is slow for the following reason:
On checkout git applies the smudge filter on every file. This will start a new git-lfs process. This process makes a [HTTP request to read the meta data](https://github.com/github/git-lfs/blob/87264d01411046716bdf178da516e77cb958ed06/lfs/client.go#L111) and afterwards a [HTTP request to read the content](https://github.com/github/git-lfs/blob/87264d01411046716bdf178da516e77cb958ed06/lfs/client.go#L124). All this stuff is done 14237 times in my case... right?

@bkeepers 
Yeah, that's why I added "DO NOT MERGE" :smile: I just wanted to introduce the caching idea and hacking redis into the client was the fastest way to make a proof of concept.

technoweenie:
> This will start a new git-lfs process. This process makes a HTTP request to read the meta data and afterwards a HTTP request to read the content. All this stuff is done 14237 times in my case... right?

Git LFS caches objects in `.git/lfs/objects`. So it only does that the first time.

Also, one of the ideas with the `git lfs fetch` command (new in the upcoming v0.5.2) is handling batch downloads in the `post-checkout` hook. This won't be spawning a command for each file, and will run them in parallel. This could have a dramatic impact on the download speeds. The batch uploads in v0.5.2 improve the speed 20-50%. My current test repo has 5000+ objects at ~500k each.

Issue #318, opened 2015-05-20
GIT_CONFIG is ignored


technoweenie:
The shell integration tests pass its custom git config values to a `git clone` through the  `GIT_CONFIG` environment variable.  One of the custom values is a `credential.helper` setting to a test-only credential helper that generates a username and password for any "127.0.0.1:PORT" git servers.  This config is meant to be applied to test repos only.  

As diagnosed [here](https://github.com/github/git-lfs/pull/306#issuecomment-103958132), it seems this isn't working with a git clone.
1. `$ GIT_CONFIG=path/to/config git clone`
2. Git runs fetch internally.
3. As Git checks the files out, it calls `git lfs smudge` on each file.
4. before downloading, Git LFS calls `git credential`.

It seems that the custom `GIT_CONFIG` is being dropped somewhere after step 2.  I did try adding this to the `commands/commands_smudge.go` to see if Git is passing `GIT_CONFIG`:

```
    ++ out='Cloning into '\''clone'\''...
    panic: GIT_CONFIG ENV /Users/rick/github/git-lfs/test/remote/config-test-happy-path
```

So it looks like it's getting the value from Git properly.

A couple places to check:
1. The [git config loading code](https://github.com/github/git-lfs/blob/24c12a4a92e5d5474830618c89ec5c3e75638c32/lfs/config.go#L154-L202) always runs `git config -l`.  Instead, it should probably recognize `GIT_CONFIG` and use that.
2. The [git credential code](https://github.com/github/git-lfs/blob/24c12a4a92e5d5474830618c89ec5c3e75638c32/lfs/credentials.go#L35-L51) builds a `*os.Cmd` without specifying the Env.  The [docs](http://golang.org/pkg/os/exec/#Cmd) say that it should be passing the current process' env through, so in theory this should work.

michael-k:
> The docs say that it should be passing the current process' env through

The same should be true for `git config -l` and `git-config` respects `GIT_CONFIG`.

ttaylorr:
Roadmappin' this.

technoweenie:
Re-opening after adding this to a GH Project.

PR #361, opened 2015-06-01
Use .gitconfig in LocalWorkingDir instead of PWD

michael-k:
Also fixed a typo and added missing env tests.

All config/env related and therefore bundled in one PR.


technoweenie:
Nice. `ListFromFile(name string)` can be used to support the `GIT_CONFIG` env var (#318).

Issue #281, opened 2015-05-04
Release v0.5.2


technoweenie:
## DONE

* [x] Add experimental support for batch API calls #285 (@rubyist)
* [x] Add `git lfs fetch` command for downloading objects. #285 (@rubyist)
* [x] Fix `git lfs track` issues when run outside of a git repository #200, #312
* [x] Fix `git lfs track` for paths with spaces in them #327
* [x] Fix bug with zero length files #346 (@technoweenie)
* [x] The Git `pre-push` warns if `git-lfs` is not installed #339 (@rubyist)
* [x] Fix Content-Type header #329 (@joerg)
* [x] Trace performance in case of error while scanning refs #311 (@michael-k)
* [x] Improve build linux build instructions #299 (@jsh)
* [x] Various coding convention cleanups #300 (@michael-k)
* [x] Error handling refactoring #305 (@michael-k)
* [x] Performance tracing improvements #311 (@michael-k)
* [x] Fix detection of `LocalGitDir` and `LocalWorkingDir` #312 #354 #361 (@michael-k)
* [x] New integration test suite #306 (@technoweenie)
* [x] Prevent tracking a file outside of the git repository #294 (@Aorjoa)
* [x] Add script for building debian package #309 (@ssgelm)
* [x] Fix bug that prevented tracking files with spaces in the name. #327 (@technoweenie)
* [x] Fix charset parameter in Content-Type header #329 (@joerg)
* [x] Move to nut for dependency management. #331 (@technoweenie)
* [x] Fix `pre-push` hook so it doesn't exit with an error if `git lfs` does not exist. #339 (@rubyist)
* [x] Fix bug that prevented zero length files from being managed by Git LFS #346 (@technoweenie)
* [x] Fix inconsistent file mode bits for directories created by Git LFS #364 (@michael-k)
* [x] Simple optimization to the clean and smudge filters #377 (@bozaro)
* [x] Fix bug in `git lfs untrack` that incorrectly some removed entries from `.gitattributes` #398 (@michael-k)
* [x] Batch API improvements #358 (@rubyist)
* [x] Resolved race condition #396 (@michael-k)
* [x] Cleaned up arguments #392 (@michael-k)
* [x] Don't read configuration from nonexistent file #391 (@bozaro)
* [x] Do not defer close in loop #384 (@michael-k)
* [x] Collect HTTP transfer stats #366, #400 (@rubyist)
* [x] Fixed tracking issues #356 (@michael-k)
[...]

PR #362, opened 2015-06-01
Add support for configured URLs containing credentials per RFC1738

ewbankkit:
This PR adds support for URLs containing credentials per [RFC1738](https://www.ietf.org/rfc/rfc1738.txt) - ```user:pass@host``` - to be configured in the usual places.
Such URLs will trigger a warning.
Any credentials parsed from a configured URL will be saved via ```git credential approve```.

### DIFF
--- a/lfs/config.go
+++ b/lfs/config.go
@@ -58,17 +58,27 @@ func ObjectUrl(endpoint Endpoint, oid string) (*url.URL, error) {
 }
 
 func (c *Configuration) Endpoint() Endpoint {
+	e := c.endpoint()
+
+	if u, err := url.Parse(e.Url); err == nil && u.User != nil {
+		fmt.Fprintln(os.Stderr, "warning: configured LFS endpoint contains credentials")
+	}
+
+	return e
+}
+
+func (c *Configuration) endpoint() Endpoint {
 	if url, ok := c.GitConfig("lfs.url"); ok {
-		return Endpoint{Url: url}.newEndpoint()
+		return Endpoint{Url: url}
 	}
 
 	if len(c.CurrentRemote) > 0 && c.CurrentRemote != defaultRemote {
-		if endpoint := c.RemoteEndpoint(c.CurrentRemote); len(endpoint.Url) > 0 {
-			return endpoint.newEndpoint()
+		if e := c.RemoteEndpoint(c.CurrentRemote); len(e.Url) > 0 {
+			return e
 		}
 	}
 
-	return c.RemoteEndpoint(defaultRemote).newEndpoint()
+	return c.RemoteEndpoint(defaultRemote)
 }
 
 func (c *Configuration) ConcurrentTransfers() int {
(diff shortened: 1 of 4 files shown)


lfs/config.go
```diff
@@ -57,18 +57,28 @@ func ObjectUrl(endpoint Endpoint, oid string) (*url.URL, error) {
...
 
-func (c *Configuration) Endpoint() Endpoint {
+func (c *Configuration) Endpoint() (endpoint Endpoint) {
+	defer func() {
+		if u, err := url.Parse(endpoint.Url); err == nil {
+			if u.User != nil {
+				fmt.Fprintln(os.Stderr, "warning: configured LFS endpoint contains credentials")
+			}
+		}
+	}()
```
technoweenie: Using a `defer` here seems odd.  This isn't really cleaning up a resource.  I wonder if we should just have a `NewEndpoint()` function that encapsulates this?

technoweenie:
:+1: for the idea. I'm a little concerned that a function on `*Configuration` is printing to stderr though. I wonder if there's a better way to handle that.  Ideally the commands themselves have full control over what gets sent to stdout or stderr. The `lfs` package is more like a library.

ewbankkit:
Yeah, I don't like writing directly to **stderr** from within what is effectively library code but there is some precedence (e.g. ```pointer_smudge.go```). Another thought is to use the logging code in ```commands.go```.

technoweenie:
I think it makes sense for library code to return errors, and the `commands` package to log or output them correctly. I want to review and include this with a new v0.5.2 release next week.  #281

michael-k:
Code-style counterproposal:
```go
func (c *Configuration) Endpoint() Endpoint {
	e := c.endpoint()

	if u, err := url.Parse(e.Url); err == nil && u.User != nil {
		fmt.Fprintln(os.Stderr, "warning: configured LFS endpoint contains credentials")
	}

	return e
}

func (c *Configuration) endpoint() Endpoint {
 	if url, ok := c.GitConfig("lfs.url"); ok {
		return Endpoint{Url: url}
 	}
 
 	if len(c.CurrentRemote) > 0 && c.CurrentRemote != defaultRemote {
 		if e := c.RemoteEndpoint(c.CurrentRemote); len(e.Url) > 0 {
			return e
 		}
 	}
 
	return c.RemoteEndpoint(defaultRemote)
}
```

michael-k:
> I think it makes sense for library code to return errors, and the `commands` package to log or output them correctly.

I looked into it and didn't see a best way to do that.  The `endpoint` is never really exposed to the `commands` package and its usage seems to be deeply nested in package `lfs`.

technoweenie:
I came to the same conclusion. I think an stderr message here is fine. If someone can come up with a better idea, I have no problem slowly refactoring things to support it.

I do want to change `newEndpoint()` into a function without a receiver though. I'll make a PR based on this one soon to get the idea across.

ewbankkit:
Incorporated suggested code style proposal.

technoweenie:
Hey @ewbankkit, I noticed some issues with this when I actually _reviewed_ the code. The `Creds` struct is used for `git-credential` specifically. In the case that a remote URL already has auth info, we don't need to even call `git-credential`. Since your patch returns a `Creds` populated from the url auth info, it will still call `git credential approve` to store the url auth username/pass in the configured git credential helper.

I fixed it in #407, which builds on the new `Endpoint` constructors introduced in #403. It changes `getCreds()` to check `apiUrl` for url credentials. I was able to use your almost test verbatim.

Thanks for diving in and attempting this!

PR #365, opened 2015-06-03
Directory Permissions - Changed to use constants

thpang67:
I have updated the code to use constants in place of the 0755 along with a note about possibly using the umask of the system git-lfs is installed.

### DIFF
--- a/lfs/lfs.go
+++ b/lfs/lfs.go
@@ -17,15 +17,11 @@ const Version = "0.5.1"
 //
 // Setup permissions for the given directories used here.
 //
-// NOTE: Pull these values from the umask value of this
-//       process in the future. This would ensure the directories
-//       are created with the same value the Administrator of the
-//       system had intended without making any assumptions as the
-//       values do below.
-//
-const TEMP_DIR_PERMS        = 0755
-const LOCAL_MEDIA_DIR_PERMS = 0755
-const LOCAL_LOG_DIR_PERMS   = 0755
+const (
+    TempDirPerms       = 0755
+    LocalMediaDirPerms = 0755
+    LocalLogDirPerms   = 0755
+)
 
 var (
 	LargeSizeThreshold = 5 * 1024 * 1024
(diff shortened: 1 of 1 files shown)


rubyist:
This looks good, but I think the comment about future use of umask should be left out. The umask should already be honored when creating files and directories via Go code. That is, we ask for `0755` but if the umask is set to something other than `0022` we won't get it.

One formatting nit-pick, I think we'd rather see related consts blocked together. Also Go code will generally camel case the names rather than use all caps. You can find counterexamples such as `os.O_RDONLY` or `os.SEEK_SET` and friends, but those are usually coming from POSIX or underlying system names. I think we'd want to camel case here, e.g.

```go
const (
	TempDirPerms       = 0755
	LocalMediaDirPerms = 0755
	LocalLogDirPerms   = 0755
)
```

Thanks for the PR!

lfs/lfs.go
```diff
@@ -14,6 +14,15 @@ import (
...
 const Version = "0.5.1"
 
+//
+// Setup permissions for the given directories used here.
+//
+const (
+    TempDirPerms       = 0755
+    LocalMediaDirPerms = 0755
+    LocalLogDirPerms   = 0755
+)
```
michael-k: For now, there's no need to export these constants.

technoweenie:
@thpang67 Thanks, just merged this.  I have some Go feedback for you, under the assumption that you're pretty new to Go in general:

We don't need to expose those values because they only matter to the internals of the `lfs` package. I made them lowercase to keep them private in 4eb1660fa7e968e9e67931827197095989e6178d.

I noticed that commit [changed the spacing on the constants](https://github.com/github/git-lfs/commit/4eb1660fa7e968e9e67931827197095989e6178d#diff-0a0453ed1c5381f97debc0f1ffa878fcL20). I use the `go-plus` atom package to run `goimports` on every Go file after a save automatically. The project will also run `script/fmt` at the very start of every `script/test` or `script/cibuild` run too.

thpang67:
@technoweenie thanks for the information on the atom package. I have just started using atom and find it to be quite useful. I have added the go-plus package. Go is new to me. I am coming from many years of structured development and programming. Thanks again.

PR #372, opened 2015-06-07
Return a non-zero exit status if build fails

michael-k:
When running test scripts and the build fails, the tests are running anyway.  This obscures the build error in case an old executable is still present.

- Return a non-zero exit status if one or more builds fail.
- Exit test with a non-zero exit status if build fails, i.e. fail test.


technoweenie:
Nice :+1:

PR #373, opened 2015-06-07
Fsck command

### LINKED ISSUE #262: git-lfs fsck (bitrot detection)

michael-k:
This PR is built on #280 by @zeroshirts and #290 by @technoweenie.

I transferred the test to the new test environment. Should be enough to close #262.

@zeroshirts:
I squashed a05e331 (*Add LFS fsck command*) and 1c8c54d (*code review changes*) together and moved  `file.Close()` right after `_, err = io.Copy(oidHash, f)`.¹  I hope this is ok with you!?

¹ See comment in 2f6c3be.  If `io.Copy` fails, we still want to close the file.


michael-k:
Some comments on questions raised in #290:

**Is `fsck` the right name?**

Consistent with `git` and `git-annex`. → Yes

**Is deleting the corrupt objects the right default?**

Potential data loss. → No

>  I'm not sure what we could do to fix files.

Not git-lfs' job.

**Observations (not necessarily blockers for this PR)**

Nice things we should document somewhere in order to do them in other PRs.  I'd say the current user interface is a good default.

zeroshirts:
LGTM

technoweenie:
Awesome, thanks for picking this back up. I'd like to ship this with Git LFS v0.5.2 some time next week.  

We do need to add a [man page](https://github.com/github/git-lfs/tree/master/docs/man) for this.

michael-k:
:+1: I've put the man page on my todo list for the next days.

technoweenie:
> sha256sum: command not found

I use `shasum -a 256`, but it may be easier to write one in Go. I have no idea what we can use that will work cross platform.

technoweenie:
Here's one idea: https://gist3.github.com/technoweenie/85e1611dff2289bd7e18

michael-k:
The gist3 link returns a 404. This one works, though: https://gist.github.com/technoweenie/85e1611dff2289bd7e18

I changed `sha256sum` to `shasum -a 256`. Other tests are using `shasum -a 256` too, therefore I thing it's out of scope for this PR to move to a custom hash program.

michael-k:
I added a man page. From my POV this PR is ready.

technoweenie:
> Other tests are using shasum -a 256 too, therefore I thing it's out of scope for this PR to move to a custom hash program.

Ideally all shell scripts in git-lfs are written for lowest common denominator sh environments. I have no idea what the standard shasum command is, or if there even is one. I agree that it's not worth adding a custom command until we need to though.

Issue #376, opened 2015-06-10
git-lfs is unusable slow (especially on Windows)


bozaro:
I convert my repository to git-lfs and try to make simple checkout:

 * Linux: 1436 sec
 * Windows: 4970 sec

Git repository is accessible by ssh protocol. LFS-server is defined by simple ```git-lfs-authenticate``` script:
```
#!/bin/sh
echo '{
  "href": "http://git.test.local:9999/test/test",
  "header": {
    "Authorization": "Basic dGVzdDp0ZXN0"
  }
}'
```

The main performance issues:

 * git-lfs for every file establish new ssh connection and executes ```git-lfs-authenticate``` command (aditional ~0.3 sec per file on Windows, ~0.2 sec per file on Linux);
 * git-lfs for every file executed as separate process (additional ~0.2 sec per file on Windows);

It should be noted that in the case of:

 * if the git-lfs-authenticate knows how files are stored, then he could immediately return the "download" section and save one HTTP request;
 * if the git-lfs-authenticate don't knows how files are stored, it always returns the same result.

In addition, processing each file in a separate process does not allow reusing HTTP connections.

Working copy contains 4231 LFS files (total size: 3.6G).
All data hosted in local gigabit network. I am sure that in the case of slower network results will be much sadder.

Versions:

 * Windows test:
   * Windows 8.1
   * git-lfs: 0.5.1
   * git: 1.9.5.msysgit.1
   * ssh: OpenSSH_6.6p1, OpenSSL 1.0.1f 6 Jan 2014
 * Linux test
   * Ubuntu 14.04.2 LTS
   * git-lfs: 0.5.1
   * git: 1.9.1
   * ssh: OpenSSH_6.6.1p1 Ubuntu-2ubuntu2, OpenSSL 1.0.1f 6 Jan 2014

bozaro:
Also performance on Windows can be improvement by changing in ```.gitconfig```:

From:
```
[filter "lfs"]
	clean = git lfs clean %f
	smudge = git lfs smudge %f
```
To:
```
[filter "lfs"]
	clean = git-lfs clean %f
	smudge = git-lfs smudge %f
```

bozaro:
I try to use ssh multiplexing as workaround (http://en.wikibooks.org/wiki/OpenSSH/Cookbook/Multiplexing).
It's reduce clone time on Linux from 1436 sec to 541 sec (2.65X faster).
But I could not get this feature to work under Windows.

technoweenie:
[...]

PR #377, opened 2015-06-10
Remove unnecessary exec the process by filter wrapper

bozaro:
This change improve git-lfs performance on Windows, becuase Ms Windows function CreateProcess is too expensive (~0.2 sec per call).
This small change reduce my test repository clone time from 4970 sec to 4352 sec (1.14X faster).
See #376 for more details.


technoweenie:
Cool thanks! Nice low hanging fruit :)

PR #378, opened 2015-06-10
Pure SSH API implementation

sinbad:
Implements a pure SSH path allowing you to perform LFS operations entirely over SSH instead of only using SSH for auth and uploading/downloading over HTTP. Useful for those who want to simply self-host using existing SSH key access without using a web server. SSH key authentication is required.

Corresponding reference server implementation is at https://github.com/sinbad/git-lfs-ssh-serve (there's also a mock SSH client in test/cmd/lfstest-mockssh used for integration testing)

Other features of this PR:
* Support for ssh:// format URLs as well as bare SSH urls
* Support for custom SSH ports
* Support for GIT_SSH environment variable, and special port options for plink.exe and TortoisePlink.exe
* As many parallel SSH connections as defined in lfs.concurrenttransfers are opened & used at once (default 3)

Future enhancements:
* Hybrid mode in git-lfs-ssh-serve where the API is implemented entirely via SSH but upload/download can be via hypermedia links (e.g. store on S3)
* Remembering whether SSH is using the pure SSH route or the mixed git-lfs-authenticate route so that we don't waste time on failed calls & falling back each time
* Eventually merge features of git-lfs-ssh-authenticate (needs source to reference implementation)
* Manual authentication (only SSH key auth is supported right now)

Previously submitted as https://github.com/github/git-lfs/pull/350


sinbad:
Warning: I haven't been able to test the git-lfs-authenticate fallback because I don't have a server implementation available. This should be tested before merging.

sinbad:
Hrm, integration tests stalled on Travis, but the "ssh" integration test completed "OK" in the output - in all the cases the ssh path stalled for me while testing, that output didn't appear. Not sure what to do about that, it completes fine here.

sinbad:
Using technoweenie/glowing-octo-lana has exposed some issues in the git-lfs-ssh-authenticate fallback, so I'm on that.

sinbad:
Cloning and fetching with git@github.com:technoweenie/glowing-octo-lana now works as expected. Can't test pushing at this time but it should be the same.

technoweenie:
That looks like travis is timing the test out. I can replicate it locally:

```
$ script/test lfs -v
=== RUN TestSuccessStatus
--- PASS: TestSuccessStatus (0.00 seconds)
=== RUN TestErrorStatusWithCustomMessage
--- PASS: TestErrorStatusWithCustomMessage (0.00 seconds)
...
=== RUN TestSuccessfulDownloadWithRedirects
```

Also, the integration test asks for a pwd:

```
$ script/integration 
Git LFS: /Users/rick/github/git-lfs/bin/git-lfs
git-lfs/0.5.1 (GitHub; darwin amd64; git 2.3.0; go 1.3)
tmp dir: /Users/rick/p/git-lfs-temp
remote git dir: /Users/rick/github/git-lfs/test/remote
LFSTEST_URL=/Users/rick/github/git-lfs/test/remote/url LFSTEST_DIR=/Users/rick/github/git-lfs/test/remote lfstest-gitserver
Running this maxprocs=4
test: track ...                                                    OK
test: update ...                                                   OK
test: status ...                                                   OK
test: track directory ...                                          OK
test: track without trailing linebreak ...                         OK
test: track outside git repo ...                                   OK
test: status --porcelain ...                                       OK
test: track representation ...                                     OK
test: track absolute ...                                           OK
test: track in gitDir ...                                          OK
test: submodule local git dir ...                                  OK
Username for 'https://mock.com': test: submodule env ...                                            OK
```

Probably getting that from an ssh remote url somewhere.

technoweenie:
What do you think about breaking this up into smaller PRs? There's a lot going on here.

1. Smaller tweaks like GIT_SSH or ssh:// support.
2. The ApiContext interface. Maybe even take the opportunity to simplify it to just `DownloadObjects()` and `UploadObjects()`.
3. SSH implementation of ApiContext.

> Eventually merge features of git-lfs-ssh-authenticate (needs source to reference implementation)

I can't provide the source. I don't think this will be that difficult as long as the different JSON messages look consistent.

1. When the ApiContext is built from an ssh git remote, it makes an initial ssh call that all SSH servers supporting Git LFS use.
2. The response either reports the HTTPS url and header, or the ssh command.
3. If the full SSH protocol is supported, `NewEndpoint()` returns the SSH context with the open connection for further commands.
4. If it points to the HTTPS url, `NewEndpoint()` closes the SSH command and caches the url and header in the HTTP context.

```js
// $ ssh git@git-server.com git-lfs-authenticate my-repo download

// use https client
{
  "href": "https://lfs.git-server.com.com/lfs/my-repo",
  "header": {
    "Authorization": "Token foobar"
  },
  "expires_at": "2015-06-10T20:59:02Z"
}

// use ssh client
{
  "href": "ssh://git@git-server.com",
  "expires_at": "2015-06-10T20:59:02Z"
}
```

I don't care about the specific so much. I just need a way to maintain backwards compatibility for existing customers, and I want the entire git lfs API to feel consistent.

michael-k:
> (…) breaking this up into smaller PRs (…)

:+1:

sinbad:
> Also, the integration test asks for a pwd:

Hrm, they don't do that for me, I'll try another machine to see if I can reproduce.

> What do you think about breaking this up into smaller PRs?

Yeah. I think only 2 though: the SSH url form, custom port support & GIT_SSH, and the pulling out of ApiContext & implementation of the SSH API. I don't really want to change the API as well because that requires a corresponding PR on the HTTP reference server which makes things more complex. Trying to change things to just one call to DownloadObjects also has implications for how concurrent transfers works - does that move inside the APIContext, or does the caller to DownloadObjects partition the set and make several parallel calls (in which case that requires multiple ApiContext instances for stateful connections like SSH, as is happening now, which is part of the reason for GetApiContext/ReleaseApiContext).

> Eventually merge features of git-lfs-ssh-authenticate (needs source to reference implementation)
> The response either reports the HTTPS url and header, or the ssh command.

I don't really like making the server self-aware of its own SSH command, that's not really necessary and stops you doing clever relaying via scripts if that's what you want to do. Even if you just made it a 'marker' response that assumes you hit the same server but with a different command, it means people wanting to just run pure SSH have to have both commands installed. 

However I wonder if I could be smarter about this and run a compound command over ssh, like `git-lfs-authenticate foo bar || git-lfs-ssh-serve bar`, so that if you have git-lfs-authenticate the old behaviour happens but if not and you have git-lfs-ssh-serve then you get the full ssh path, but only calling SSH once. That way neither path suffers from the 'try-fail-fallback' inefficiency. I'll need to experiment.

I have some other work to do the next couple of days and will be back to this early next week.

technoweenie:
> Trying to change things to just one call to DownloadObjects also has implications for how concurrent transfers works - does that move inside the APIContext, or does the caller to DownloadObjects partition the set and make several parallel calls (in which case that requires multiple ApiContext instances for stateful connections like SSH, as is happening now, which is part of the reason for GetApiContext/ReleaseApiContext).

The API will be much easier to change before the interface is extracted. Let's figure out all the quirks in the HTTP client code before bringing in a new adapter. We can do it slow and carefully, making sure not to introduce new bugs. If we bring this in now, as is, we'll be stuck with the current API until we transition both adapters at the same time.

> does the caller to DownloadObjects partition the set and make several parallel calls

They could all be synchronous calls. The context object persists across multiple calls, maintaining any state or connections.

```go
context := NewContext()
defer context.Close()
context.DownloadObjects(array1)
context.DownloadObjects(array2)
```

The current ref scanner code may have to change though. Currently it scans the entire ref, and returns the whole array of object pointers. It could be changed to publish to a channel, or just fire a batch callback every `n` objects.

sinbad:
> They could all be synchronous calls. The context object persists across multiple calls, maintaining any state or connections.

So in your pseudocode the DownloadObjects calls for array1/2 are publishing their results asynchronously to a channel or something, so you're assuming some parallelisation internal to the context I assume. This means: 

1. The SSH state management moves internal to the context; probably becoming worker goroutines with a cycle of startup/connect, process 0:n commands in serial, then exit. Inputs and outputs become channels, although much more abstract, probably just a list of Oids & the operation. You probably still want the Batch interface as well as Download/Upload so that you can pre-determine the size for progress reporting.
2. A large part of the fetch code may need restructuring since it runs all its own goroutines right now and most of that will move inside the context. The trouble with that is that it might result in some structural duplication compared to right now, where all the concurrency is in the higher levels and the contexts are simpler.

I actually think it will be better to leave the concurrency in the outer code like it is now, and have the contexts stay much simpler, so concurrency is achieved just by code calling multiple context methods in parallel - just that in the case of stateful ones like SSH, each context is only allowed to be used in one goroutine at a time (GetApiContext() organises that for SSH while allowing the stateless HTTP context to be re-used in parallel)

rubyist:
There are a couple problems that are causing the tests to deadlock.

The first one is easy, some of the download tests are doing a `defer reader.Close()` in a loop when it should just be closing them immediately, causing the deadlock:

```diff
diff --git a/lfs/download_test.go b/lfs/download_test.go
index 02efc58..936d3bd 100644
--- a/lfs/download_test.go
+++ b/lfs/download_test.go
@@ -226,7 +226,7 @@ func TestSuccessfulDownloadWithRedirects(t *testing.T) {
 		if wErr != nil {
 			t.Fatalf("unexpected error for %d status: %s", redirect, wErr)
 		}
-		defer reader.Close()
+		reader.Close()
 
 		if size != 4 {
 			t.Errorf("unexpected size for %d status: %d", redirect, size)
@@ -568,7 +568,7 @@ func TestSuccessfulDownloadFromSeparateRedirectedHost(t *testing.T) {
 		if wErr != nil {
 			t.Fatalf("unexpected error for %d status: %s", redirect, wErr)
 		}
-		defer reader.Close()
+		reader.Close()
 
 		if size != 4 {
 			t.Errorf("unexpected size for %d status: %d", redirect, size)
```

The other is a little more problematic. The combination of calling `UploadCheck()` followed by `UploadObject()` will always deadlock (really, calling anything that would create a context before calling `UploadObject()` will have the same problem, but this is the real scenario). During `GetApiContext()` for `UploadObject()`, the context already exists so we return without unlocking the cache lock. When the call to `ReleaseApiContext()` is made it deadlocks waiting on the cache lock.

sinbad:
> There are a couple problems that are causing the tests to deadlock.

Thanks for spotting those, I'll take a look.

sinbad:
Argh, just spent many hours tracking down the stalling issue on Travis, turned out it was nothing to do with ssh - fixed it & was going to submit a PR but @michael-k already got to it: https://github.com/github/git-lfs/pull/396 Well played, wish I'd worked on this a day later ;)

So now it's all working again & re-integrated with latest changes, will split it up.

sinbad:
I've split out the SSH URL/port/GIT_SSH support as https://github.com/github/git-lfs/pull/404 & re-merged this one on top. Also split out a couple of other small ones in case you want to do those separately but https://github.com/github/git-lfs/pull/404 is the main one.

If you don't want to merge this until after the API changes then I won't do any more work on this PR for the moment, since merging it does take a bit of work each time (regular conflicts and I have to be careful to preserve modifications to moved code). I think there's a fairly significant discussion to be had around exactly DownloadObjects will work, I think you still need the Batch() up-front to estimate transfer times. FYI I'm splitting my time between Git LFS and something else now so I'll be back on this early next week barring small things.

ajohnson23:
I just wanted to throw a +1 here to show my support. I think this would be a great feature to have.

pedrocarmona:
👍

ttaylorr:
Hey @AndrewJDR, just came across this PR again in my inbox. We just shipped #1236 which should be a big step in making it possible to easily implement this with minimal effort. 😄

AndrewJDR:
I noticed this was closed. Does that mean there are no plans for a pure SSH implementation that ships with the official git-lfs package?

ttaylorr:
I think it is something that we're interested in long-term, but not right now.

AndrewJDR:
Well for what it's worth, I'll be very happy when this gets implemented some day. If I had more free time, I'd do it myself.

AndrewJDR:
Well for what it's worth, I'll be very happy when this gets implemented some day.  If I had more free time, I'd do it myself.
It seems like @sinbad  had pure SSH almost totally implemented at one point! I'm assuming that's no good due to all the api changes?

sinbad:
Yeah, it's way out of date now.

I'd like a pure SSH route one day too just on principle, I think if you've chosen to use SSH you should be able to complete everything that way if you want. But in practice most providers end up with the lfs storage accessible primarily over HTTPS anyway, and you get some nice automatic features that way, so in practice it turned out to be a bit niche in practice. I'm an old traditionalist though (shakes fist at cloud) so I'd like it to happen eventually.

AndrewJDR:
@sinbad If pure SSH does happen one day, please revive or at least throw a post into this thread so I get notified :) Thanks!

PS. I am with you!
oldman

johnwbyrd:
Yeah, I'm glad I saw this before I implemented LFS for our project.  The promise of "git lfs" is that "everything will just work."  If you have no interest in supporting ssh properly, then we need the special GitHub type server with the special GitHub type protocol which means that LFS isn't really a standard at all.  Will check back in a couple years once you guys figure this one out.

AndrewJDR:
@johnwbyrd This is a fair criticism of git-lfs imo. If you're looking for something that 'just works' over ssh, I can suggest looking at git-fat: https://github.com/abraithwaite/git-fat
It's very similar, but already works over plain old ssh.

From github's perspective, I think they made git-lfs to serve their own purposes as a company and opening it up was just gravy. I suppose I understand that there is maybe little incentive for them to implement this. That said... I really, really wish someone would implement this.

technoweenie:
Thanks for the feedback. Lack of an SSH protocol is definitely fair criticism about LFS. There are a lot of other core issues with LFS, and our small team is working on what we think is the most crucial. However, I'd love to work with someone on implementing a pure SSH protocol.

johnwbyrd:
The more I think about this problem, the more I think that the fundamental approach taken with LFS, annex, fat, lfs, etc.... are unsound.  They all take the basic approach of creating a parallel metadata repository, and filtering out certain files so they are not stored in git, but stored in the metadata repository.  This then creates a parallel problem of managing the metadata repository.  So now you've taken one problem and made two problems out of it.

Linus got it right the first time.  Git's secret weapon is its distributed nature -- any git install is a client and a server and both.  But everyone is trying to solve this problem in a client/server manner.  I believe there is a better way... but it's not going to be got by messing around with the porcelain of Git, or by building servers to manage big data in parallel.

Recall that if two git users have different local branches, their repositories can look quite different.  One user can have a branch called "master" and another branch called "bigfiles".  Another user can just have a branch called "master".  The size of their repositories will be quite different, even though they are both working on the same master branch.  

I posit that the central problem is not managing "big" files and rerouting them into a central server with a filter.  The central problem is deciding under which conditions certain refs and/or branches should be pushed or pulled between repositories, and then deciding how those refs can be combined and/or overlaid and presented to the user at checkout time.  It is possible that this might be accomplished merely with another level of indirection.  Like git lfs, this proposed method would have blobs containing SHA1 hashes which would be "fixed up" with a smudge filter... but unlike lfs, those smudge filters would look for blobs or trees in the current repository that match the SHA1 hash given.  If they exist, you have a "full" repository; if not, then you don't currently have access to the "big" files and still need to get them from someone else.

Look at what such a system would buy you.  You could decide whether you wanted to be up-to-date constantly with a detailed history of every "big" asset, you could say you only wanted to track the latest "big" asset in the repository (conveniently, git gc gets rid of dangling blobs for you), or you could say you don't want to see big assets at all.  And based on the branches and/or refs you receive at push/pull time, git itself could work out whether the "big" files need to go over the wire or not, and use its existing systems to transfer them if they need to go.  No special servers or protocols needed.

This is probably worthy of a blog post or longer discussion,  And I'd be very surprised if I were the first person who thought of this approach.  I'll get out of this pr now (I know this discussion does not belong here) and come back when I have something useful.  Thanks for listening.

Schroedingers-Cat:
Interesting point! I think being able to set up a file like .gitignore where you put in lines like ".png" that will be cloned/pulled to your repository only when you checkout a commit sounds great and would feel more git-like. You could have all these files garbage collected when you didn't check them out for a week or so. It would also be great if you could specifiy certain file sizes and combine then with file types for automatic purging from the repo.

As these actions are potentially dangerous (you are telling your vcs to drop content silently you put in it in the first place) it should show an info like "incomplete repository" when calling up "git status".

When you are interested in a native git-like solution, you can achieve kind-of-the-same by putting large files into submodules and shallow-cloning these submodules. This way, you keep the entire source code history but have only a subset of the big files shallow-cloned to you local submodul.

ioquatix:
@johnwbyrd did you ever follow up with an implementation or more discussion somewhere?

johnwbyrd:
Nope, Real Life has gotten in the way.  In my copious spare time.  Real Soon Now.

In the meantime, note that "git overlay" as a concept means several different things, so I'll need to go with "git layer" or some similar term.

Also note that the new "git worktree" concept is about two doors down from the thing I'm proposing.

ioquatix:
@johnwbyrd It's okay I understand that time is precious.

`git worktree` is pretty interesting. How does it relate to what you are proposing?

johnwbyrd:
git worktree is a new git feature that demonstrates, at least conceptually, that two branches of the same git repository can coexist at the same moment.  git worktree does nothing about reconciling these differences into a single directory, nor was it intended to do that.

PR #380, opened 2015-06-11
git-lfs ignores GIT_SSH enviroment variable

bozaro:


michael-k:
This is already covered by #378 (0ca3060).

lfs/ssh.go
```diff
@@ -22,7 +23,11 @@ func sshAuthenticate(endpoint Endpoint, operation, oid string) (sshAuthResponse,
 
 	tracerx.Printf("ssh: %s git-lfs-authenticate %s %s %s",
 		endpoint.SshUserAndHost, endpoint.SshPath, operation, oid)
-	cmd := exec.Command("ssh", endpoint.SshUserAndHost,
+	ssh := os.Getenv("GIT_SSH")
+	if len(ssh) == 0 {
+		ssh = "ssh"
+	}
```
technoweenie: This is great, but typically we've been putting this stuff in the `*lfs.Configuration` struct so that we only have to check the ENV once per git lfs process.

michael-k: Not really:
```
$ grep -Ri "os.Getenv" *
lfs/credentials.go:			if exitErr.ProcessState.Success() == false && os.Getenv("GIT_TERMINAL_PROMPT") == "0" {
lfs/http.go:	if sslVerify == "false" || len(os.Getenv("GIT_SSL_NO_VERIFY")) > 0 {
lfs/util.go:	logPath := os.Getenv("GIT_LFS_PROGRESS")
lfs/config.go:		isTracingHttp: len(os.Getenv("GIT_CURL_VERBOSE")) > 0,
```

technoweenie: Ah you're right, I'm thinking of one of our internal projects :) I still think it's a good thing to do, especially for anything that's called multiple times. Two of those 4 matches are cached.

bozaro:
@technoweenie 
Maybe, but I'm not ready to do it. I was first to meet any project in golang and not ready to write quality code :disappointed:

technoweenie:
@bozaro I want to see how #378 (which implements this) shakes out first though. 

I really appreciate the feedback and PRs you've submitted.  As @michael-k pointed out, I too put in several uncached `os.Getenv()` calls too :)

bozaro:
Update commit for using Config.Getenv provided by #387

bozaro:
Close as duplicate of #404

PR #381, opened 2015-06-11
Windows: hide git application window.

bozaro:
Original problem (https://code.google.com/p/tortoisegit/issues/detail?id=2518):

 1. Install Git LFS (https://git-lfs.github.com/)
   * Download and unpack archive;
   * Put git-lfs.exe on any avaialble in PATH directory;
   * Run "git lfs init"
 2. Clone repository from: https://github.com/bozaro/test
 3. Change image "tux.png"
 4. Show "Git Commit" window

On showing "Git Commit" window you will see black window blink for each changed LFS file (twice in this case).

You also will see black window blink on refreshing change list by F5 key.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -118,7 +118,7 @@ func (c *gitConfig) Version() (string, error) {
 // simpleExec is a small wrapper around os/exec.Command.
 func simpleExec(name string, args ...string) (string, error) {
 	tracerx.Printf("run_command: '%s' %s", name, strings.Join(args, " "))
-	cmd := exec.Command(name, args...)
+	cmd := execCommand(name, args...)
 
 	output, err := cmd.Output()
 	if _, ok := err.(*exec.ExitError); ok {
--- a/git/git_nix.go
+++ b/git/git_nix.go
@@ -1,0 +1,13 @@
+// +build !windows
+
+// Package git contains various commands that shell out to git
+package git
+
+import (
+	"os/exec"
+)
+
+// execCommand is a small platform specific wrapper around os/exec.Command
+func execCommand(name string, arg ...string) *exec.Cmd {
+	return exec.Command(name, arg...)
+}
--- a/git/git_win.go
+++ b/git/git_win.go
@@ -1,0 +1,16 @@
+// +build windows
+
+// Package git contains various commands that shell out to git
+package git
+
+import (
+	"os/exec"
+	"syscall"
+)
+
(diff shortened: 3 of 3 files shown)


technoweenie:
Awesome! This issue came up in internal testing, but we weren't sure what we could do about it.

I want to review and include this with a new v0.5.2 release next week.  #281

If other Windows users can vouch for this, that'd be great. Windows is not my primary OS.

@bozaro: Is there a way to replicate this on Windows without Tortoise?

git/git.go
```diff
@@ -7,7 +7,6 @@ import (
 	"io"
 	"os/exec"
 	"strings"
-
 	"github.com/github/git-lfs/vendor/_nuts/github.com/rubyist/tracerx"
```
michael-k: Can you please add the empty line again?  This is the import formatting done by [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports).  If you have it installed, you can use `script/fmt` to format the git-lfs code.

Here's how you can update this PR (replace editor with the editor of your choice; only needed if you don't have goimports):
```
$ editor git/git.go
$ script/fmt
$ git add --patch git/git.go
$ git commit --amend
$ git push <remote> <localbranch>:blinking
```

michael-k:
:+1: for using build tags!

bozaro:
@technoweenie 
To reproduce the problem you need to run git-lfs without windows.
I know two ways to do this:

* Run from another application (in this case TortoiseGit);
* Compile the program as a GUI-application and run:```go build -ldflags "-H windowsgui"```

bozaro:
Important note: when the problem was detected, the blinking was due to call ```git version```. This call has been removed in #382.

bozaro:
Do I need anything else from me to solve this problem?
The problem of flashing windows is showstopper for using git-lfs and TortoiseGit (perhaps it affect any more GUI for git).

michael-k:
> This pull request contains merge conflicts that must be resolved.

The signature of `execCommand` has changed upstream.  Please remove the parameter `stdin io.Reader` and
```go
if stdin != nil {
	cmd.Stdin = stdin
}
```

bozaro:
@michael-k 
Updated commit.

technoweenie:
I was hoping someone else on Windows would chime in. I'll do some Windows QA for the v0.5.2 launch though. Thanks so much for filling in our Windows knowledge gaps :+1:

PR #382, opened 2015-06-11
Remove useless fork per every file for "git version"

bozaro:
Because Ms Windows has very expensive CreateProcess this operation takes too long time on large file count.

### DIFF
--- a/lfs/lfs.go
+++ b/lfs/lfs.go
@@ -8,7 +8,6 @@ import (
 	"runtime"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


technoweenie:
It's nice to have the Git version from users for troubleshooting purposes. However, most of the time this info is unnecessary, and not even linked to the `git-lfs` binary. 

I'd like to merge this, and then include the version for Git LFS and Git at the top of `git lfs env`.

PR #383, opened 2015-06-12
Fix the deb package build and bump the version to 0.5.2

ssgelm:
Thanks to the new more standard project layout we no longer need to symlink things.  I also bumped the version in preparation for #281.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+git-lfs (0.5.2) stable; urgency=low
+
+  * New upstream version
+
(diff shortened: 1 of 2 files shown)


technoweenie:
Should I wait for v0.5.2 to merge this?

ssgelm:
It's your call.  Merging this will mean that the package will end up being version 0.5.2 so it might be worth waiting.  Having said that if one were to pull master and run the build they would end up with what is essentially version 0.5.2 anyway.

PR #384, opened 2015-06-12
Do not defer close in loop

michael-k:
See https://github.com/github/git-lfs/pull/378#issuecomment-111231779:

> There are a couple problems that are causing the tests to deadlock.
>
> The first one is easy, some of the download tests are doing a defer reader.Close() in a loop when it should just be closing them immediately, causing the deadlock:


rubyist:
Thanks for pulling this out :+1:

PR #386, opened 2015-06-12
Support chunked Transfer-Encoding

ryansimmen:
Fixes github/git-lfs#385

### DIFF
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -302,7 +302,11 @@ func UploadObject(o *objectResource, cb CopyCallback) *WrappedError {
 	if len(req.Header.Get("Content-Type")) == 0 {
 		req.Header.Set("Content-Type", "application/octet-stream")
 	}
-	req.Header.Set("Content-Length", strconv.FormatInt(o.Size, 10))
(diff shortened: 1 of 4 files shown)


technoweenie:
Easy peasy! This needs a test.

technoweenie:
I'd like to see a test that uses the [git and git lfs server that shell integration tests use](https://github.com/github/git-lfs/blob/master/test/cmd/lfstest-gitserver.go). I think we can do this by sending a `Transfer-Encoding` header in the [post](https://github.com/github/git-lfs/blame/968faba7e11786e3884dba1e797f816f2c216541/test/cmd/lfstest-gitserver.go#L131) and [batch](https://github.com/github/git-lfs/blame/968faba7e11786e3884dba1e797f816f2c216541/test/cmd/lfstest-gitserver.go#L207) endpoints if the integration test's repo name has "chunked-transfer-encoding" in the name.

ryansimmen:
I've updated the PR with a few small integration tests.

rubyist:
code and tests look :+1:

PR #391, opened 2015-06-15
Do not read the configuration from nonexistent file (remove redundant git call)

bozaro:


michael-k:
I think this is better placed in `ListFromFile` itself.  Just return an empty string and `nil` if the file doesn't exist.

bozaro:
@michael-k 
I move check to ```ListFromFile``` :)

michael-k:
:+1:

rubyist:
:+1: Thanks!

PR #392, opened 2015-06-15
Cleaned up arguments

michael-k:


rubyist:
I'm not sure why Travis is failing here, but this looks good to me. Thanks!

michael-k:
@rubyist Travis is failing, because of one of two reasons (or both):

- The defer stuff you mentioned: https://github.com/github/git-lfs/pull/378#issuecomment-111231779 (fix: #384)
- A deadlock in transfer queue. I investigated it during the last two days and the fix is nearly ready. I'm going to push it in the next couple of hours.

michael-k:
Btw. travis is failing for some pull requests due to this two issues.  Once the fixes are upstream, the PR tests can be rerun.

rubyist:
@michael-k are you experiencing tests hanging locally? So far I am unable to reproduce this outside of Travis. Looking forward to seeing what you found in the transfer queue code. While it's definitely bad form (and I will merge that PR, too), I don't think the defered closes are relevant outside of that particular SSH PR (#378), which toggles mutex states inside of a `Close()`.

michael-k:
I experienced the `test-ls-files.sh` hanging and maybe that's due to the defers. But I'm not sure and I didn't investigate it further.

> are you experiencing tests hanging locally?

Nope and that was a big problem.  Endless adjustments and my own travis tests (#393 would have been helpful) until I found the lock.  Upside: I know the test environment better. ;)

But it is possible to reproduce it locally. See #396 for details (and fix).

PR #393, opened 2015-06-15
Make it possible to enable Travis on forks

sinbad:
I needed this to debug Travis issues, might be useful for someone else. Usually trying to use Travis with a fork of a golang project fails because the clone path is $fork_user/git-lfs instead of github/git-lfs which breaks package references. Note that this code should work on any golang repo owned by github, in case that's useful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,6 @@ install: true
 script: script/cibuild
 notifications:
   email: false
+before_install:
+  - repo=`basename $PWD`; localDir=`dirname $PWD`; cfDir="`dirname $localDir`/github"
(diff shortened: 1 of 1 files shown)


michael-k:
:+1:
I had the same issue and made a temporary commit on top which changed the imports:
```sh
grep -RI "github.com/github/git-lfs" * | cut -d ":" -f 1 | sort | uniq | xargs sed -i -e 's/github.com\/github\/git-lfs/github.com\/michael-k\/git-lfs/g'
```

technoweenie:
Thanks!

PR #396, opened 2015-06-15
Resolved race condition

michael-k:
To reproduce the deadlock, put `runtime.Gosched()` right after `workersReady <- 1`.  The `q.authCond.Signal()` is the being sent without any worker waiting at `q.authCond.Wait()`.  Therefore all workers block at `q.authCond.Wait()` and the main goroutine blocks at `wg.Wait()` (deadlock).

Use case 1 `q.transferables` is empty:

> `<-workersReady` is not (endlessly) blocking, because the additional `go func` closes the channel.  The `q.authCond.Signal()` does not do anything.

Use case 2 `q.transferables` is not empty:

> `<-workersReady` in combination with `q.authCond.L.Lock()` blocks until at least one worker is waiting (`q.authCond.Wait()`). Therefore the `q.authCond.Signal()` signals one of the goroutines to
  run.

> The size of the channel `workersReady` makes sure that every worker that passed `q.authCond.L.Lock()` before the main goroutine does not block at `workersReady <- 1`.

### DIFF
--- a/lfs/transfer_queue.go
+++ b/lfs/transfer_queue.go
@@ -75,11 +75,11 @@ func (q *TransferQueue) processIndividual() {
 
 	for i := 0; i < q.workers; i++ {
 		go func() {
-			workersReady <- 1
 			for t := range apic {
 				// If an API authorization has not occured, we wait until we're woken up.
 				q.authCond.L.Lock()
 				if atomic.LoadInt32(&q.clientAuthorized) == 0 {
+					workersReady <- 1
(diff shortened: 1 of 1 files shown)


michael-k:
After the merge, all PR tests that timed out should be restarted.

rubyist:
Nice, that looks legit to me :+1:

michael-k:
Just for reference, here is the commit causing the deadlock: 5619c5c

PR #399, opened 2015-06-16
Michael k cleanup config

technoweenie:
Merges latest master into #387.

### DIFF
--- a/lfs/config.go
+++ b/lfs/config.go
@@ -17,13 +17,15 @@ import (
 
 type Configuration struct {
 	CurrentRemote         string
-	gitConfig             map[string]string
-	remotes               []string
 	httpClient            *HttpClient
 	redirectingHttpClient *http.Client
+	envVars               map[string]string
 	isTracingHttp         bool
 	isLoggingStats        bool
-	loading               sync.Mutex
+
+	loading   sync.Mutex // guards initialization of gitConfig and remotes
+	gitConfig map[string]string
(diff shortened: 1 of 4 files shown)


lfs/config.go
```diff
@@ -40,24 +42,22 @@ var (
...
 func NewConfig() *Configuration {
 	c := &Configuration{
-		CurrentRemote:  defaultRemote,
-		isTracingHttp:  len(os.Getenv("GIT_CURL_VERBOSE")) > 0,
-		isLoggingStats: len(os.Getenv("GIT_LOG_STATS")) > 0,
+		CurrentRemote: defaultRemote,
+		envVars:       make(map[string]string),
 	}
+	c.isTracingHttp = len(c.Getenv("GIT_CURL_VERBOSE")) > 0
+	c.isLoggingStats = len(c.Getenv("GIT_LOG_STATS")) > 0
```
technoweenie: Since this is already caching the bool check, we can use `os.Getenv()` here, and not worry about caching it in the `gitConfig` map. Perhaps a micro optimization that doesn't really matter.

Also, should bool checks actually look at the value, instead of seeing if the string is non empty? Common accepted values could be `1`, `true`, or `on`, while values like `0`, `false`, and `off` could still be false.

technoweenie:
I'm going to merge this. The git config bool checks can be looked at in another PR if it's important enough.

PR #403, opened 2015-06-17
Consistently construct Endpoint from urls the same way

sinbad:
There are 2 ways to construct LFS urls - from explicit URLs and from the
clone URL (which gets info/lfs appended). Previously there were multiple
paths where the Endpoint could be constructed in a short-circuit manner
which prevents future expansion in common functions. This standardises
the Endpoint construction so it always happens via NewEndpoint, and
NewEndpointFromCloneURL performs the extra step of appending info/lfs.

This has a current effect of supporting SSH urls in all config settings,
and not just the remote clone URL as before. But mostly this is to enable
future expansion (improved SSH support)

### DIFF
--- a/lfs/config.go
+++ b/lfs/config.go
@@ -62,7 +62,7 @@ func (c *Configuration) Getenv(key string) string {
 
 func (c *Configuration) Endpoint() Endpoint {
 	if url, ok := c.GitConfig("lfs.url"); ok {
-		return Endpoint{Url: url}
+		return NewEndpoint(url)
 	}
 
 	if len(c.CurrentRemote) > 0 && c.CurrentRemote != defaultRemote {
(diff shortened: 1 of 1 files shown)


michael-k:
Is it necessary to export the two functions?

sinbad:
> Is it necessary to export the two functions?

They're both safe for external use and conceptually robust, and Endpoint is already exported so really safe construction should be too.

michael-k:
I don't see it as a question of “safe for external use” but as a design decision.  Exported functions are the API of a package.  A narrow API provides guidance on how to use a package.  (Very helpful for people new to a project.)

Exposed functions are also restricting programmers in terms of signature changes.  Changing the signature of an exposed function, means changing the API.  For now, this is not really an issue as the packages are only used internally.  But in the future this might change.

sinbad:
If that's the case then Endpoint shouldn't be exposed either. IMO you either expose the type AND the construction functions or neither.

michael-k:
We need to export the type `Endpoint` to generate [documentation](https://godoc.org/github.com/github/git-lfs/lfs#Endpoint) for its members (which is not yet written).

I considered [`Endpoint()`](https://godoc.org/github.com/github/git-lfs/lfs#Configuration.Endpoint) and [`RemoteEndpoint()`](https://godoc.org/github.com/github/git-lfs/lfs#Configuration.RemoteEndpoint) being the construction functions.

Maybe my views are just to misguided by my OOP background.  So, I'm totally open to new insights and opinions here. (Which doesn't mean that I'll give up my position without a fight. ;) )

sinbad:
Fair point about Endpoint() & RemoteEndpoint() although that's specific to creating from other instances. NewFoo() is the idiom in Go for publicly constructible types since you can't have protected constructors like C++/Java to tell people they shouldn't just construct themselves. 

I don't feel super-strong about it though, if Rick wants this internalised before accepting I'll do it.

technoweenie:
:+1: on the change.

I think if we want to move towards this being a package for other Go libraries, we should start moving to `github.com/github/git-lfs`, as opposed to `github.com/github/git-lfs/lfs` as it is today. I think that should be done deliberately when someone's ready to think about the design. I'm cool with the current exported constructors for now.

technoweenie:
I'm going to merge this so it can be used in #362, unless anyone has any objections.

lfs/config.go
```diff
@@ -105,38 +105,47 @@ func (c *Configuration) RemoteEndpoint(remote string) Endpoint {
...
 
 	return Endpoint{}
 }
 
-func remoteEndpointFromUrl(url string) Endpoint {
+const ENDPOINT_URL_UNKNOWN = "<unknown>"
+
+// Create a new endpoint from a URL associated with a git clone URL
+// The difference to NewEndpoint is that it appends [.git]/info/lfs to the URL since it
+// is the clone URL
```
technoweenie: In the future I think method comments should follow [standard godoc conventions](http://blog.golang.org/godoc-documenting-go-code).

Obviously this project isn't strict about docs. But I think if someone takes the time to write something, it should follow the conventions. It's easy enough that I may just make the change live before merging this, or applying it in a quick PR.

technoweenie:
I reorganized the code a bit in #407 based on the assumption that this PR will be merged as is. Now that `Endpoint` is more than just a struct declaration, I moved it all to a separate go file.

2a6af23ce065e6fd25b7c54ec383362f7cd30993 is the actual commit.

PR #404, opened 2015-06-17
Improved SSH support: non-bare urls, custom ports, and GIT_SSH/plink

sinbad:
* Full SSH urls e.g. ssh://user@host/repo
* Custom ports in both bare and full urls
* GIT_SSH environment for alternate ssh clients
* Explicit support for plink.exe and TortoisePlink.exe

Depends on https://github.com/github/git-lfs/pull/403 (included in PR)

### DIFF
--- a/lfs/config.go
+++ b/lfs/config.go
@@ -51,6 +51,16 @@ func (c *Configuration) Getenv(key string) string {
 	return v
 }
 
+func (c *Configuration) Setenv(key, value string) error {
+	// Check see if we have this in our cache, if so update it
+	if _, ok := c.envVars[key]; ok {
+		c.envVars[key] = value
+	}
+
+	// Now set in process
+	return os.Setenv(key, value)
+}
+
 // GetenvBool parses a boolean environment variable and returns the result as a bool.
 // If the environment variable is unset, empty, or if the parsing fails,
 // the value of def (default) is returned instead.
--- a/lfs/endpoint.go
+++ b/lfs/endpoint.go
@@ -10,13 +10,12 @@ import (
 
 const EndpointUrlUnknown = "<unknown>"
 
-var httpPrefixRe = regexp.MustCompile("\\Ahttps?://")
-
 // An Endpoint describes how to access a Git LFS server.
 type Endpoint struct {
 	Url            string
 	SshUserAndHost string
 	SshPath        string
+	SshPort        string
 }
 
 // NewEndpointFromCloneURL creates an Endpoint from a git clone URL by appending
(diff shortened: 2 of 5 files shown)


technoweenie:
Can you merge the latest master now that #403 has been merged? Rebase if you like.

lfs/config.go
```diff
@@ -105,41 +116,99 @@ func (c *Configuration) RemoteEndpoint(remote string) Endpoint {
...
+	return e
+}
+
+// Create a new endpoint from a general URL
+func NewEndpoint(urlstr string) Endpoint {
+	endpoint := Endpoint{Url: urlstr}
+	// Check for SSH URLs
+	// Support ssh://user@host.com/path/to/repo and user@host.com:path/to/repo
+	u, err := url.Parse(urlstr)
+	if err == nil {
```
technoweenie: I'm not a fan of this giant method. I'd prefer to have something that calls internal functions:

```go
u, err := url.Parse(urlstr)
if err != nil {
  // right now just git@github.com, but maybe others one day?
  processSshRemote(urlstr)
}

switch u.Scheme {
case "http", "https": processHttpUrl(u)
case "ssh": processSshUrl(u)
default: NewEndpoint(ENDPOINT_URL_UNKNOWN)
}
``` 

Note: feel free to override my suggestions on method or variable names. 

Minor nitpick: I tend to use `rawurl`. I don't like having type related suffixes on variable names. Maybe it's from my experience using hungarian notation in VB/Access code :)

sinbad: OK - rebased and will refactor a little in next commit. 

You did VB/Access too eh? We should commiserate over a beer one day ;)

lfs/config.go
```diff
@@ -151,20 +162,94 @@ func NewEndpointFromCloneURL(url string) Endpoint {
...
+				newPath = strings.Join(parts, "/")
+			}
+			newrawurl := fmt.Sprintf("ssh://%v", newPath)
+			newu, err := url.Parse(newrawurl)
+			if err == nil {
+				return newu
+			}
+		}
+	}
+	return u
```
michael-k: I'd prefer early returns.
```go
if u.Scheme != "" || u.Path == "" {
	return u
}

// This might be a bare SSH URL
// Turn it into ssh:// for simplicity of extraction later
parts := strings.Split(u.Path, ":")
if len(parts) < 2 {
	return u
}

// Treat presence of ':' as a bare URL
var newPath string
if len(parts) > 2 { // port included; really should only ever be 3 parts
	newPath = fmt.Sprintf("%v:%v", parts[0], strings.Join(parts[1:], "/"))
} else {
	newPath = strings.Join(parts, "/")
}

newrawurl := fmt.Sprintf("ssh://%v", newPath)
newu, err := url.Parse(newrawurl)
if err != nil {
	return u
}

return newu
```

michael-k: Another possibility is a switch statement:
```go
var newPath string
switch {
case len(parts) < 2:
	return u
case len(parts) == 2:
	newPath = strings.Join(parts, "/")
default:
	// port included; really should only ever be 3 parts
	newPath = fmt.Sprintf("%v:%v", parts[0], strings.Join(parts[1:], "/"))
}
```

lfs/config.go
```diff
@@ -151,20 +162,94 @@ func NewEndpointFromCloneURL(url string) Endpoint {
...
+			endpoint.SshUserAndHost = fmt.Sprintf("%s@%s", u.User.Username(), host)
+		} else {
+			endpoint.SshUserAndHost = host
+		}
+		if len(match) > 2 {
+			endpoint.SshPort = match[2]
+		}
+	} else {
+		endpoint.Url = ENDPOINT_URL_UNKNOWN
+		return endpoint
```
michael-k: possible early return

lfs/ssh.go
```diff
@@ -38,3 +47,41 @@ func sshAuthenticate(endpoint Endpoint, operation, oid string) (sshAuthResponse,
...
+	basessh := filepath.Base(ssh)
+	// Strip extension for easier comparison
+	if ext := filepath.Ext(basessh); len(ext) > 0 {
+		basessh = basessh[:len(basessh)-len(ext)]
+	}
+	isPlink := strings.EqualFold(basessh, "plink")
+	isTortoise := strings.EqualFold(basessh, "tortoiseplink")
+	if ssh == "" {
+		ssh = "ssh"
+	}
```
michael-k: counterproposal:
```go
isPlink := false
isTortoise := false

ssh := Config.Getenv("GIT_SSH")
if ssh == "" {
	ssh = "ssh"
} else {
	basessh := filepath.Base(ssh)
	// Strip extension for easier comparison
	if ext := filepath.Ext(basessh); len(ext) > 0 {
		basessh = basessh[:len(basessh)-len(ext)]
	}
	isPlink := strings.EqualFold(basessh, "plink")
	isTortoise := strings.EqualFold(basessh, "tortoiseplink")
}
```

lfs/config.go
technoweenie: I definitely prefer early returns vs multiple levels of nesting. :+1:

bozaro:
Without this change, you can not use SSH-connections multiplexing on Windows.
I think this latest update, which can give a appreciable performance boost without using batch api.
It would be nice if it was included in v0.5.2 (#281).

technoweenie:
This PR doesn't look ready, so it likely won't make it in. We can release v0.5.3 whenever we want though.

sinbad:
AFAIK the only thing that's "not ready" about it is stylistic preferences?

technoweenie:
Yes, and a merge conflict with master.

sinbad:
OK, that must be new in the last day or two. Will come back to this next week.

rubyist:
Are we doing ssh multiplexing in this PR? If so, has it been benchmarked with large numbers of large files? It is my understanding that ssh multiplexing starts to fall over when connections with large amounts of data are involved, though I am admittedly unfamiliar with plink.

technoweenie:
There's also another patch I'm holding back because of insufficient testing. We can roll v0.5.3 when these two are ready. I'm anxious to get v0.5.2 out because it's been way too long since the last release.

@rubyist: We may be able to punt on that concern for now until we're reviewing the SSH API.

sinbad:
> Are we doing ssh multiplexing in this PR? If so, has it been benchmarked

All this PR does is support ssh:// url forms, custom ports, and the GIT_SSH env var. Its performance characteristics are identical to master since the ssh calls themselves happen exactly the same way, all it does is support cases that would previously fail.

My other PR on full SSH support opens only Config.ConcurrentTransfers() SSH connections to run the API over and pipes everything through those.

rubyist:
@sinbad :+1: that sounds OK then. My concerns were around using something like ssh's ControlMaster (or equivalents in plink, etc), using ConcurrentTransfers() connections with the transfer queue should be just fine.

sinbad:
> My concerns were around using something like ssh's ControlMaster (or equivalents in plink, etc), using ConcurrentTransfers() connections with the transfer queue should be just fine.

Yeah I did consider adding another config parameter that could override ConcurrentTransfers() for SSH connections, using a default lower limit but for simplicity I didn't, and the default 3 should be OK for all but the most totalitarian server admin.

bozaro:
> Are we doing ssh multiplexing in this PR? If so, has it been benchmarked
I make little benchmark on Windows/Linux with our project repository #376 (4231 LFS files, working copy size: 3.6G, original repository size: ~50G). Local gigabit network environment.


**On Linux:**
I create simple config ```~/.ssh/config```:
```
Host *
    ControlMaster auto
    ControlPath ~/.ssh/controlmasters/%r@%h:%p
    ControlPersist 10m
```
This changes enable using single SSH connection for all ```git-lfs-authenticate``` requests.
Clone time reduced from 1436 sec to 541 sec (2.65X faster, ~0.216 sec per file).

**On Windows:**
I try to use similar configurations with differ OpenSSH versions. It looks that OpenSSH this feature is not supported on Windows.

A similar function is in Putty (plink.exe), but its use needs support GIT_SSH environment  variable (this PR).

The effect of multiplex under Windows, I did not measure.

sinbad:
Rebased on master & added early-outs, should be ready for inclusion in 0.5.3 now.

lfs/endpoint.go
```diff
@@ -37,21 +36,95 @@ func NewEndpointFromCloneURL(url string) Endpoint {
...
+		switch u.Scheme {
+		case "ssh":
+			endpoint = endpointFromSshUrl(u)
+		case "http", "https":
+			endpoint = endpointFromHttpUrl(u)
+		default:
+			// Just passthrough to preserve
+			endpoint = Endpoint{Url: rawurl}
+		}
+	}
```
michael-k: Should we really return an empty `Endpoint` in case of `err != nil`?  Why not `Endpoint{Url: rawurl}`?

sinbad: Pretty much anything will go through the default: route, but for consistency with the other routes I've returned EndpointUrlUnknown in that case. You have to throw something pretty whacky at url.Parse for it to return an error (none of the tests with bare strings manage it).

lfs/ssh.go
technoweenie: I like it. I think the workflow is clearer. @sinbad: Thoughts? I can just make this change after merging.

sinbad: Yeah it's fine - I don't mind either way.

PR #405, opened 2015-06-17
Cleaned up old integration test leftovers

technoweenie:
Fixes github/git-lfs#357


technoweenie:
Jealous that you get the :red: blood commit :red: :)

PR #407, opened 2015-06-17
Tweaks to config/endpoints

technoweenie:
DO NOT MERGE

I started exploring a different approach to #362 based on the assumption that #403 will be merged.


technoweenie:
Code spike served it's purpose.

PR #411, opened 2015-06-18
bring windowless commands to the lfs package

technoweenie:
This brings the Windows-specific improvements from #381 to the lfs package. Specifically for the `git-credential` and `ssh` execs.

/cc @bozaro @sinbad

### DIFF
--- a/lfs/credentials.go
+++ b/lfs/credentials.go
@@ -33,7 +33,7 @@ type CredentialCmd struct {
 
 func NewCommand(input Creds, subCommand string) *CredentialCmd {
 	buf1 := new(bytes.Buffer)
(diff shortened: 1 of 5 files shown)


sinbad:
I think you only hit this problem if you use a git.exe other than the standard msys version - I haven't seen this behaviour myself on Windows, I tend to use msys git & plink. If you need me to do some specific testing I can but not for a couple of days (my time for LFS client work is Mon-Wed unless there's something urgent)

technoweenie:
Ah, right on. We'll worry about this when it's a problem.

bozaro:
I only see one potential problem: if the console application will try to ask the user any data.
But I think that everything should be fine :)

PR #413, opened 2015-06-18
Reorganize API docs

technoweenie:
This breaks up `docs/api.md` into separate files, as we start to transition the API. `docs/api/README.md` will have a high level overview of the API, and links to all the API versions.  Right now, we just have the "v1 original" and "v1 batch" APIs to describe.

I'm not incrementing the API version for the original and batch APIs since they don't affect each other. There aren't very many implementations of the original anyway. Breaking changes to existing endpoints will require new API versions and mime types. Additive changes are allowed without bumping the version.

* [x] Add API readme
* [x] Document v1 original api
* [x] Document v1 batch api


technoweenie:
I've got all my changes updated. I'm going to come back later today or tomorrow morning and re-read this before merging.

PR #415, opened 2015-06-18
Support 401 responses for Batch API

rubyist:
This PR adds full 401 support for the Batch API.

- [x] If access is not marked private, request Batch with no creds
- [x] If access is marked private, request Batch with creds
- [x] If a 401 is received, mark private, get creds, and reattempt the request

Related test server PR: https://github.com/github/lfs-test-server/pull/33

### DIFF
--- a/lfs/config.go
+++ b/lfs/config.go
@@ -131,7 +131,7 @@ func (c *Configuration) SetPrivateAccess() {
 	// Modify the config cache because it's checked again in this process
 	// without being reloaded.
 	c.loading.Lock()
-	c.gitConfig["lfs.access"] = "private"
+	c.gitConfig[key] = "private"
 	c.loading.Unlock()
 }
 
--- a/lfs/config.go
+++ b/lfs/config.go
@@ -108,6 +108,10 @@ func (c *Configuration) BatchTransfer() bool {
 	return false
 }
 
+// PrivateAccess will retrieve the access value and return true if
+// the value is set to private. When a repo is marked as having private
+// access, the http requests for the batch api will fetch the credentials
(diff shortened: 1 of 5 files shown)


technoweenie:
Don't worry about documentation, I already did that [here](https://github.com/github/git-lfs/pull/413/files#diff-f094a3f9eec03cad995d9ba33220d584R97). It's a PR, so still subject to changes of course.

lfs/config.go
```diff
@@ -117,6 +117,15 @@ func (c *Configuration) BatchTransfer() bool {
 	return false
 }
 
+func (c *Configuration) PrivateAccess() bool {
+	if v, ok := c.GitConfig("lfs.access"); ok {
+		if v == "private" || v == "PRIVATE" {
```
technoweenie: Does it make sense to use `strings.ToLower()` here?

rubyist: Yeah, I'll fix up nitpicks when things are closer to working. Just wanted to get this linked to your roadmap.

rubyist:
This is ready for some :eyes:. The gist of things:

A copy of the `newApiRequest` and `newClientRequest` code was made for batch operations (the old code will eventually go away, so I'm OK with the copy).

When a batch request is run, it will check for an `lfs.<url>.access` configuration key, where `<url>` is the endpoint url. If the value is set to "private", the request will run with the credential helpers as it normally has. If the value is missing or any other value, the request will be attempted without authorization.

If a request fails with a 401, the `lfs.<url>.access` key will be set to private in `.git/config`, the credentials will be retrieved, and the request will be retried. All future batch requests will then get the credentials when they're run.

Issue #414, opened 2015-06-18
Roadmap: Fast, efficient uploading and downloading


technoweenie:
This is a meta issue for the roadmap. Comments should be geared towards the addition and deletion of tasks. Leave detailed discussion to issues linked in tasks. If a task doesn't have an issue next to it, create one and leave a comment here so we can update the list.

* [x] Batch API (experimental in v0.5.2+, stable by v0.6.x)
* [x] Batch API: 401 responses #415
* [x] Batch API: upper limit of the number of objects per batch #419 #476
* [x] Batch API: define semantics for partial failures #417
* [x] Support chunked Transfer-Encoding #385
* [x] SSH auth: stop sending OID for batch requests #416
* [ ] Use `post-checkout` for downloads #616
* [x] Resumable downloads
* [x] Resumable uploads

## Ideas

* [ ] Chunked uploads
* [ ] Binary diffing
* [ ] Bulk tarball uploads or downloads through the storage API.

bozaro:
I think, the good idea for batch API is removing ```{oid}``` from ```git-lfs-authenticate``` command:
```
ssh git@github.com git-lfs-authenticate user/repo.git download {oid}
```
The presence of this argument needs to perform this command for each object is sent.

technoweenie:
@bozaro: Good point. We'll definitely make that change for the batch api.

PR #416, opened 2015-06-18
API documentation: removing {oid} from git-lfs-authenticate command

bozaro:
Suggestion for #414: I think, the good idea for batch API is removing {oid} from git-lfs-authenticate command:

```
ssh git@github.com git-lfs-authenticate user/repo.git download {oid}
```
The presence of this argument needs to perform this command for each object is sent.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -18,14 +18,13 @@ access the API username and password.  If the Git LFS domain matches the Git
 remote's, Git LFS will not prompt you again to enter the password.
 
 If the Git remote is using SSH, Git LFS will execute the `git-lfs-authenticate`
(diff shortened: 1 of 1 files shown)


technoweenie:
I have a PR to rearrange the API docs already: #413.

The difficulty here is making sure SSH implementations that require the OID continue to work when we stop sending it. I think we'll continue to make this call for every object with the current API, and then start omitting the OID for the upcoming batch API.

technoweenie:
I made the change in f0507599779735c57af928645d2f026d15f02fb3. How's that look?

bozaro:
It looks great.

technoweenie:
Cool, thanks for bringing this up.

PR #421, opened 2015-06-19
teach script/integration how to run against a specific git-lfs build

technoweenie:
This is useful if you want to run the existing tests against a specific git-lfs binary.


rubyist:
Cool :+1:

PR #422, opened 2015-06-19
ensure the oid/size are set

rubyist:
Older versions of the API may not include the oid and size in the json payload for the check POST. If they're blank, set them to what we currently have.


technoweenie:
:+1: We'll provide a json schema for the finalized batch API (and future APIs) to get around this in the future.

PR #423, opened 2015-06-19
Support concurrent transfers when outputting progress for desktop apps

rubyist:
This is a fix for a bug in the handling of the `GIT_LFS_PROGRESS` file which is used by desktop apps, or other things integrating with the git-lfs binary.

At a later point, we should move toward a more unified progress tracking which supports this file as well as the pb progress bars via a single API. This file, the pb output, and possibly the `watchers` channel could be unified into a single event stream, cleaning up some of this code.

### DIFF
--- a/lfs/transfer_queue.go
+++ b/lfs/transfer_queue.go
@@ -307,7 +307,6 @@ func (l *progressLogger) Close() error {
 // be used when writing causes an error.
 func (l *progressLogger) Shutdown() {
 	l.writeData = false
-	l.Close()
 }
 
 // newProgressLogger creates a progressLogger based on the presence of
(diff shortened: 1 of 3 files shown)


lfs/transfer_queue.go
```diff
@@ -247,3 +276,62 @@ func (q *TransferQueue) Process() {
...
+		return l.log.Sync()
+	}
+	return nil
+}
+
+// Close will call Close() on the underlying file
+func (l *progressLogger) Close() error {
+	if l.writeData {
+		return l.log.Close()
+	}
```
technoweenie: This should check if `l.log` is nil instead so `Close()` actually closes the handle after an error.

rubyist: :boom: good catch

PR #424, opened 2015-06-19
fix logs command

technoweenie:
You can't modify a 0 len array. I didn't want to append on each iteration, so I went with a slice instead.


rubyist:
Why not append? It's how we generally build arrays in other places in the code.

technoweenie:
Dumb notions about perf. Using `append` leads to a much simpler diff though :+1:  Also added an integration test for this command.

PR #429, opened 2015-06-20
Fix two bugs in fetch, add tests

rubyist:
This PR fixes two :bug: :beetle: in the `fetch` command.

1. When fetching the not current branch, the queue would not process, nothing would download.
1. When fetching the current branch, the main routine would not wait for the goroutine that copies files into the wd and updates the git index, possibly leaving the wd in an inconsistent state.

Integration tests for fetch are included.


technoweenie:
:+1: This should also go into the `release-0.5` branch too.

PR #437, opened 2015-06-24
initial json schemas for batch api

technoweenie:
[JSON Schema](http://json-schema.org) is a pretty good format for defining the schema of a data structure. This is the exact schema that we'll be using to validating requests and responses through GitHub's implementation of the server.

* [x] Change `_links` to `actions`. It accurately describes what they're for. Also, `_links` was only ever used for [HAL](http://stateless.co/hal_specification.html) compatibility. Git LFS has diverged enough that there's no point to maintaining that illusion.
* [x] Specify required attributes
* [x] Validate `href` and optional `header` objects under each action.
* [x] Validate action types? (`download`, `upload, `verify`).
* [x] Define `error` object.
* [x] Add simple json parsing test for the schema files #528

Current example request and response:

```js
// REQUEST
{
  "operation": "upload or download",
  "objects": [
    {
      "oid": "abc123",
      "size": 123
    }
  ]
}

// RESPONSE
{
  "objects": [
    {
      "oid": "abc123",
      "size": 123,

      "actions": {
        // objects will have either a download action,
        // or an upload + verify
        "download": {
          "href": "url",

          // optional
          "header": {
            "Some-Header": "some value"
          },

          // optional UTC timestamp in iso 8601
          "expires_at": "2015-07-20T21:10:55Z"
        }
      },

      // error property is optional
      "error": {
        "code": "fatal",
        "message": "some fatal error occurred"
      }
    }
  ]
}
```

/cc @sinbad: Curious on your thoughts. Ideally the SSH implementation shares the same JSON schema as much as possible.

EDIT: I don't intend to use the json schema files directly in the Git LFS client right now. It's just for API documentation.

### DIFF
--- a/docs/api/http-v1-batch-request-schema.json
+++ b/docs/api/http-v1-batch-request-schema.json
@@ -1,0 +1,26 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "title": "Git LFS HTTPS Batch API v1 Request",
+  "type": "object",
+  "properties": {
+    "operation": {
+      "type": "string"
+    },
+    "objects": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "oid": {
+            "type": "string"
+          },
+          "size": {
+            "type": "number"
+          }
+        },
+        "additionalProperties": false
+      }
+    }
+  },
+  "additionalProperties": false
+}
--- a/docs/api/http-v1-batch-request-schema.json
+++ b/docs/api/http-v1-batch-request-schema.json
@@ -18,9 +18,11 @@
             "type": "number"
           }
         },
+        "required": ["oid", "size"],
         "additionalProperties": false
       }
     }
   },
+  "required": ["objects", "operation"],
   "additionalProperties": false
 }
--- a/docs/api/http-v1-batch-response-schema.json
+++ b/docs/api/http-v1-batch-response-schema.json
@@ -58,7 +58,7 @@
             "additionalProperties": false
           }
         },
-        "required": ["oid", "size", "actions"],
+        "required": ["oid", "size"],
         "additionalProperties": false
       }
     }
--- a/docs/api/http-v1-batch-response-schema.json
+++ b/docs/api/http-v1-batch-response-schema.json
@@ -1,0 +1,23 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "title": "Git LFS HTTPS Batch API v1 Response",
+  "type": "object",
+  "properties": {
+    "objects": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "oid": {
+            "type": "string"
+          },
+          "actions": {
+            "type": "object"
+          }
+        },
+        "additionalProperties": false
+      }
+    }
+  },
+  "additionalProperties": false
+}
--- a/docs/api/http-v1-batch-response-schema.json
+++ b/docs/api/http-v1-batch-response-schema.json
@@ -12,7 +12,7 @@
             "type": "string"
           },
           "size": {
-            "type": "number",
+            "type": "number"
           },
           "actions": {
             "type": "object"
--- a/docs/api/http-v1-batch-response-schema.json
+++ b/docs/api/http-v1-batch-response-schema.json
@@ -11,6 +11,9 @@
           "oid": {
             "type": "string"
           },
+          "size": {
+            "type": "number",
+          },
           "actions": {
             "type": "object"
           }
--- a/docs/api/http-v1-batch-response-schema.json
+++ b/docs/api/http-v1-batch-response-schema.json
@@ -48,7 +48,7 @@
             "type": "object",
             "properties": {
               "code": {
-                "type": "string"
+                "type": "number"
               },
               "message": {
                 "type": "string"
--- a/docs/api/http-v1-batch-response-schema.json
+++ b/docs/api/http-v1-batch-response-schema.json
@@ -2,6 +2,27 @@
   "$schema": "http://json-schema.org/draft-04/schema",
   "title": "Git LFS HTTPS Batch API v1 Response",
   "type": "object",
+
+  "definitions": {
+    "action": {
+      "type": "object",
+      "properties": {
+        "href": {
+          "type": "string"
+        },
+        "header": {
+          "type": "object",
+          "additionalProperties": true
+        },
+        "expires_at": {
+          "type": "string"
+        }
+      },
+      "required": ["href"],
+      "additionalProperties": false
+    }
+  },
+
   "properties": {
     "objects": {
       "type": "array",
@@ -15,12 +36,33 @@
             "type": "number"
           },
           "actions": {
-            "type": "object"
+            "type": "object",
+            "properties": {
+              "download": { "$ref": "#/definitions/action" },
+              "upload": { "$ref": "#/definitions/action" },
+              "verify": { "$ref": "#/definitions/action" }
+            },
+            "additionalProperties": false
+          },
+          "error": {
+            "type": "object",
+            "properties": {
+              "code": {
+                "type": "string"
+              },
+              "message": {
+                "type": "string"
+              }
+            },
+            "required": ["code", "message"],
+            "additionalProperties": false
           }
         },
+        "required": ["oid", "size", "actions"],
         "additionalProperties": false
       }
     }
   },
+  "required": ["objects"],
   "additionalProperties": false
 }
--- a/docs/api/http-v1-batch.md
+++ b/docs/api/http-v1-batch.md
@@ -107,6 +107,7 @@ authentication info, regardless of how `lfs.<url>.access` is configured.
 > Authorization: Basic ... (if authentication is needed)
 >
 > {
+>   "operation": "upload",
 >   "objects": [
 >     {
 >       "oid": "1111111",
@@ -122,7 +123,8 @@ authentication info, regardless of how `lfs.<url>.access` is configured.
 <   "objects": [
 <     {
 <       "oid": "1111111",
-<       "_links": {
+<       "size": 123,
+<       "actions": {
 <         "upload": {
 <          "href": "https://some-upload.com",
 <           "header": {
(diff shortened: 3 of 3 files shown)


sinbad:
I think your TODOs cover most of the comments I'd have; I do think that the content of "actions" should be specified fairly tightly, i.e. an action type and a URL, and probably specify the 3 action types.

I like that you've included an 'operation' at the top of the request; although in practice this may not be required if this batch endpoint is the only request in the API, but it does offer options for extensibility later; assuming other operations can use the same request structure of course. In the SSH work before a 'method' was required because there was no URL or GET/POST/PUT differentiation to split the various requests, it's just that if everything moves to this batch endpoint that's no longer necessary as such. If you wanted to be minimalist and didn't need this for anything else, you could drop it and instead I could always wrap this JSON in an outer object in SSH which included any additional context only required by SSH.

And yep, definitely needs an error object; it would be good to have a code & message and have the code fulfil the same requirement as the HTTP response code - duplication in the HTTP API of course but required if we're unifying the JSON for non-HTTP too.

Otherwise, all looks good to me. I can re-work the SSH PR when the time comes to either use this JSON directly or wrap it if extra context is needed but still keep the data the same. That way HTTP & SSH can just become almost dumb transports with the real processing happening on the JSON instead.

sinbad:
Oh FYI I intend for the URL on the actions to have a 'special' URL form when the intention is to allow the client to submit all the uploads & downloads via the same connection (open to suggestions of what that form should be, but it shouldn't feature any hostnames as such) . But by using this form it will support the hybrid mode of having an SSH API (if your git host only wants to have that running, and is doing all the permissions checking there) but all the upload/download action happening via something else entirely e.g. S3. And keeps the responses standard of course.

technoweenie:
> I do think that the content of "actions" should be specified fairly tightly, i.e. an action type and a URL, and probably specify the 3 action types.

Agreed :+1:

> I like that you've included an 'operation' at the top of the request; although in practice this may not be required if this batch endpoint is the only request in the API

I added it so the server knows what permissions to check. It only needs to check read/pull access for downloads, but write/push access for uploads. 

Technically I could use GET (ElasticSearch receives search json docs as GET request bodies), but that feels weird. I'm open to other HTTP verbs, but a bit concerned about compatibility with various HTTP clients. Also, it may limit the types or names of future operations that we could add. 

Another possibly better option is specifying the operation in the path.

```
POST /objects/upload
POST /objects/download
POST /objects/delete # not specified, but it IS in the roadmap
```

> it would be good to have a code & message and have the code fulfil the same requirement as the HTTP response code - duplication in the HTTP API of course but required if we're unifying the JSON for non-HTTP too.

The errors are per-object. I didn't plan on adding a top level error object for things like authz problems. I think non HTTP transports can just wrap this structure with an envelope.

sinbad:
I had another thought about this based on some server work I'm doing. On the assumption that the storage location isn't the same as where the API is running (as supported by the hypermedia links), it's pretty common to use authorization tokens in the returned JSON to be included in the headers of the final upload/download request to implement security - GH does this with its use of S3, I'm using JWT http://jwt.io for a private server I've written. 

The problem is you really want those tokens to have an expiry time so they can't be used forever. If you only ever make the API request for all files you have to transfer up-front, that means you have to have a excessively long expiry time (probably 24h) to account for the chance that the client has a ton of transfers queued up and so the later requests might happen hours after the initial API call when the tokens were generated.

One way to address that would be to do an extra API call just before uploading each file (you still need the batch call up-front to do progress bars) but I don't like that. Instead I suggest adding an optional 'expires_at' property to each link, and have git-lfs detect when a link it's about to use has expired and can re-request new links. I think expiry on upload/download links is going to be common enough to justify this.

technoweenie:
@sinbad: I had a similar thought about expiration while designing the SSH fallback API:

```
$ ssh git@github.com git-lfs-authenticate github/git-lfs-test download
{
  "href": "https://api.github.com/lfs/github/git-lfs-test",
  "header": {
    "Authorization": "BOOYA"
  },
  "expires_at": "2015-07-06T22:34:50Z"
}
```

Even if we added `expires_at` as you suggest, we'll still have to deal with random HTTP issues due to intermittent bugs, bad deploys, or just flaky internet issues. We've been throwing around some ideas to support retries for these types of http responses (401, 403, 50x). I think I'd prefer handling these errors consistently, unless there's a really good reason to handle expiration errors differently.

sinbad:
Yeah using 'expires_at' makes sense universally for links, whether it's from git-lfs-authenticate or for download/upload links. I agree that retry behaviour should be shared for all cases where you'd need to retry, whether error or expiry.

technoweenie:
> So I guess in both cases 'expires_at' is maybe unnecessary since your recovery is always the same based on context?

This is actually why I didn't include `expires_at` in the spec for the SSH auth stuff. I suppose there's really no point in retrying 403 responses though, unless the client knows (through the existence of `expires_at`) that the token is temporary. No point retrying 403 responses from standard user/password requests.

Ok, I'll add it as an optional property.

sinbad:
> No point retrying 403 responses from standard user/password requests.

I think you should retry those, but after deleting the git creds so they get re-prompted - the most common case is that the user changed their password on the server but the local keychain / creds cache still has the old one.

technoweenie:
Good call with `git-credentials`. There's a low chance of with headers too. Maybe I have to update my S3 key, invalidating all of the expiring tokens. They're not expired, but the git lfs client should hit the batch api again with the failed OIDs to get a fresh token.

See, we don't need `expires_at` :)

sinbad:
> See, we don't need expires_at :)

I know, I know :) I just have this nagging feeling that it'll be useful in some disambiguation case we haven't thought of yet (I know from SourceTree that debugging client-side auth errors can be a PITA)

bozaro:
I want to write my own LFS-server implementation.

In my case, I want to set temporary token for upload/download files. It will allow the user to enter a login and password for the LFS-server (SSO).

At the time of the token generation it is enough for me that the user is allowed to execute a ```git-lfs-authenticate```. I don't knwon user password.
The moment of blocking a user or change the password, I can not be able to keep track of. As a result - token must be temporary. So I need ```expires_at``` attribute.

sinbad:
@bozaro you can implement expiry on the token in your server without the lfs client ever knowing it was there, it's usually part of the token, as in http://jwt.io. It's just that if it *is* in the JSON as well, the client has a better chance of knowing the reason a request has failed because of an expired token & that it's OK just to retry with the same auth mechanism.

technoweenie:
I specified some more things. This is what the JSON should look like now:

```js
// REQUEST
{
  "operation": "upload or download",
  "objects": [
    {
      "oid": "abc123",
      "size": 123
    }
  ]
}

// RESPONSE
{
  "objects": [
    {
      "oid": "abc123",
      "size": 123,

      "actions": {
        // objects will have either a download action,
        // or an upload + verify
        "download": {
          "href": "url",

          // optional
          "header": {
            "Some-Header": "some value"
          },

          // optional UTC timestamp in iso 8601
          "expires_at": "2015-07-20T21:10:55Z"
        }
      },

      // error property is optional
      "error": {
        "code": "fatal",
        "message": "some fatal error occurred"
      }
    }
  ]
}
```

Questions:

1. What should the error codes be? Right now it's a string. We could make it an integer so it has the same semantics as http error codes (403, 500, etc). Or we could define short string names for known errors.
2. Is ISO8601 fine for the `expires_at`? JSON doesn't really have a date type. We only need second resolution. I'd prefer ISO8601 or even just unix time.

rubyist:
:+1: for the json changes. ISO8601 is cool with me. Strings or ints for error codes - my first reaction would be ints, but I'm not super against short strings. Ints seem like they'd be a little bit easier to handle client side, particularly if a client wants to translate/localize, but not so much so that I'm hard line one way or the other.

sinbad:
Looks good to me, I'd just note that the size property is optional in the request (for "download"). 

1. Error codes: also think an int would be better and that we should predefine some common cases, similar to HTTP status codes. However we need to define what constitutes an error - for example do we define a missing object on "download" as an error case in this context, with (say) a 404 status code in this section, and have the client code able to detect a missing resource case from either a HTTP 404 or a 404 error code in the JSON? Previously in the SSH PR I was using a return size of -1 to differentiate between 'not found' and 'a real error' but something a bit more explicit like a 404 in the error response would be good for standardisation, it just means the HTTP route has some duplication between the response code and the error content.
2. ISO8601 is fine with me too.

technoweenie:
Reusing HTTP codes as the error code makes sense. Some of the status codes are pretty well defined (404 obviously), while others are really vague (400, 422). 

> for example do we define a missing object on "download" as an error case in this context, with (say) a 404 status code in this section, and have the client code able to detect a missing resource case from either a HTTP 404 or a 404 error code in the JSON?

Yes, I think the http response would be 200, even if every requested object is missing. The 404 would be recorded in the JSON. However, an authorization error should result in a 401/403 http response, as opposed to reporting it in the JSON.

sinbad:
> Yes, I think the http response would be 200, even if every requested object is missing. The 404 would be recorded in the JSON.

Ah yes, since the error structure is per object that would work. In the case of an auth error though that's going to mean potentially the same error on every object? Or do you think a top-level error structure is needed as well to handle that more cleanly?

technoweenie:
> you intend to do that in the HTTP status code - that does preclude handling those errors the same way in SSH though

I think the SSH protocol can use the same JSON format here, with an extra wrapper in place of the http response.

```json
{
  "status": 403,
  "body": {
    "objects": []
  }
}

sinbad:
Yep, agreed.

rubyist:
I think auth errors should be handled at the top level. Normal http codes and whatever we do for ssh. It makes the batching and concurrency a bit easier.

technoweenie:
I added some json schema tests in #528. The diff is massive because it's vendoring a [jsonschema package](https://github.com/xeipuuv/gojsonschema#working-with-errors) and its dependencies.

* [json_schema_test.go](https://github.com/github/git-lfs/blob/test-api-schema/docs/json_schema_test.go) verifies that each json file in `docs/**` is a valid json schema document.
* [http_v1_batch_test.go](https://github.com/github/git-lfs/blob/test-api-schema/docs/http_v1_batch_test.go) tests the request and response object validation using the schema files.

This may be too much. Though I like the idea of including json schema files for the API, and for having clear tests that don't depend on internal lfs objects. The API is fairly simple now, though has several optional properties (`expires_at`, `error`, etc). I think this will become even more important as we add forward compatible features to the API.

technoweenie:
Thanks for the feedback, everyone.

PR #444, opened 2015-06-27
Design proposal for making Git LFS extensible

sanoursa:
This is a design proposal for making Git LFS extensible.  The idea is that you can register an LFS extension, and the extension gets inserted into the LFS clean and smudge calls, and is also tracked in the pointer file to ensure that anything that is cleaned can be correctly smudged.

Our biggest use case is that we have many files that are both very large and very compressible, and we want to be able to compress those files before they are written into the LFS cache (and then pushed to the LFS server), and uncompress them when writing to the working directory.  Our proposal is to add extensibility to the core behavior of LFS, and then we will build our own compression extension for the teams that need it. 

We are also considering releasing our compression extension so that others can use it if they want to.  Ideally, this could even lead to a library of LFS extensions, contributed by various authors, so that users of LFS can mix and match the behaviors they want when using LFS.


docs/spec.md
```diff
@@ -200,3 +200,151 @@ $ cat .gitattributes
...
+
+```
+[lfs-extension "foo"]
+  clean = git-lfs-foo clean %f
+  smudge = git-lfs-foo smudge %f
+  priority = 1
+[lfs-extension "bar"]
+  clean = git-lfs-bar clean %f
+  smudge = git-lfs-bar smudge %f
+  priority = 2
```
technoweenie: Since this is using exec to spin up extensions, we should design the filters to accept multiple objects. Perhaps we could create a json document of filenames, oids, and local paths to pass to the command.

We're not tied to the clean/smudge interface from Git. Though I like keeping the names the same.

docs/spec.md
```diff
@@ -200,3 +200,151 @@ $ cat .gitattributes
...
+be transparent to extensions, since they are still invoked on a per-file basis.
+
+### Registration
+
+To register an LFS extension, it must be added to the Git config.  Each extension needs 
+to define:
+
+* Its unique name.  This will be used as part of the key in the pointer file.
+* The command to run on clean
+* The command to run on smudge/fetch
```
technoweenie: Let's call this just "smudge."

technoweenie:
First, I don't like the idea of adding stuff to the main docs before it's written. There are a few violations of this already, and they bug me.  What do you think about moving the bottom additions to `docs/proposals/extensions.md`?  

I like the proposal, but it's not a priority of mine yet. We can _almost_ get around compression by compressing the file in `.git/lfs/objects`, and supporting gzip encoding on http requests. But, this extension will ensure that these highly compressible files won't eat up space on cloud hosts that charge for storage usage. :+1: However, I don't want to bundle a compression extension until it can guarantee the same OID every time. Also ideally we could include binaries for all major platforms still.

One minor concern is that we have a few parsing or blob scanning functions that use blob size as a way to filter out obvious non pointers. Since the pointer files are all the same, this is easy. Too many extensions could lead to files not ever getting pushed.

As the spec mentions, it's important that extension work regardless of what Git LFS command is using `.git/lfs/objects`: `clean`, `fetch`, `fsck`, `pre-push`, `push`, `smudge`, etc. We may want to design some kind of interface for this, like a database. There are some proposed features that could take advantage of this as well:

* track partial downloads #414
* track partial uploads to various remotes #414
* automatic compression of files in `.git/lfs/objects`
* auto gc ([roadmap](https://github.com/github/git-lfs/blob/master/ROADMAP.md))

Not saying all of this has to be done at once. But I think whoever writes this code should have this in the back of their mind.

> Ideally, this could even lead to a library of LFS extensions, contributed by various authors, so that users of LFS can mix and match the behaviors they want when using LFS.

We need to be mindful of the end user experience here. Git LFS should detect when extensions aren't around. Some kind of extension manager would be cool, but probably a huge yak shave until there are more than a few plugins. At the least, maybe Git LFS maintains a small directory of extension names and install or doc URLs. The [wiki](https://github.com/github/git-lfs/wiki) could be a basic starting point.

Thanks for sending this!

docs/spec.md
sanoursa: I'm making a basic assumption here that since we're dealing with large files, the overhead of reading a file twice is higher than the overhead of spinning up a process to execute an extension.  Would you agree with that assumption?

If we do as you suggest, it would require the extension to have to read the file off disk, even though Git or LFS have already read that same file.  Streaming the file contents over STDIN avoids that double I/O, but it does add process overhead.

Longer term, if we see that process overhead is too high, we can make a more meaningful optimization, and allow extensions to run in-proc.  This will be especially helpful when using "git lfs fetch" since there is a long-running LFS process that can then make multiple calls to an extension API.  The downside is it places more constraints on extension developers.

Thoughts?

technoweenie: > the overhead of reading a file twice is higher than the overhead of spinning up a process to execute an extension. Would you agree with that assumption?

For linux/mac, yes :) My understanding is that os/exec perf in Windows is much worse.

I think what you have originally is sufficient for this proposal. But it's something to be aware of when building the first implementation.

nrser:
want to chime in support for clean / smudge extensions, specifically to support encryption. spent a little bit earlier trying to get git-crypt and lfs to play with each-other, and it seems like i got it to work in the "clean" direction at least, but i have a feeling it will be a somewhat fragile solution.

docs/spec.md
whoisj: > My understanding is that os/exec perf in Windows is much worse.

While true, process start up is around 10 ms with spinny disks, better with SSD. IO will likely still cost more - avoiding IO is always a good thing.

technoweenie:
Solid proposal :+1:

PR #446, opened 2015-06-28
add some basic init tests

technoweenie:
This simply tests the current behavior in preparation for future `git lfs init/uninit` work.


rubyist:
:+1: legit

PR #447, opened 2015-06-28
Clean fix

technoweenie:
`git push` errors if you have Git LFS pointers without an object in `.git/lfs/objects`. This prevents the client from making a successful Git push without a successful Git LFS push. But, the error message is very confusing.

> Cannot clean a Git LFS pointer.  Skipping.

The confusing error message was written to prevent `git add` from turning a Git LFS pointer to _another_ pointer in the clean filter. So here's an attempt at a better error:

> new.dat is an LFS pointer to 7aa7a5359173d05b63cfd682e3c38487f3cb4f7f1d60659fe59fab1505977d4c, which does not exist in .git/lfs/objects.
>    
> Run 'git lfs fsck' to verify Git LFS objects.

This hopefully makes a little more sense. It also points the user to the fsck command, which will show the user all of their missing objects. Thoughts?

The first commits are some light refactoring as I was going over some code. Some properties were no longer used, or could be simplified. For instance, no need to send an `*os.File` in the `*cleanedAsset` object if every caller is going to immediately close it. Just close it and send the filename instead.

Also, the error message change in `commands/command_pre_push.go` is a hack. `lfs.NewUploadable` should return a `*WrappedError` with the `Panic` bool set to false. This way both `push` and `pre-push` get the correct error message for free.

### DIFF
--- a/lfs/upload_queue.go
+++ b/lfs/upload_queue.go
@@ -18,6 +18,7 @@ type Uploadable struct {
 // NewUploadable builds the Uploadable from the given information.
 func NewUploadable(oid, filename string, index, totalFiles int) (*Uploadable, *WrappedError) {
 	path, err := LocalMediaPath(oid)
+
 	if err != nil {
 		return nil, Errorf(err, "Error uploading file %s (%s)", filename, oid)
 	}
--- a/lfs/upload_queue.go
+++ b/lfs/upload_queue.go
@@ -97,12 +97,12 @@ func ensureFile(smudgePath, cleanPath string) error {
 	}
 
 	cleaned, err := PointerClean(file, stat.Size(), nil)
-	if err != nil {
-		return err
+	if cleaned != nil {
+		cleaned.Teardown()
 	}
 
-	if cleaned != nil {
-		defer cleaned.Teardown()
+	if err != nil {
+		return err
 	}
 
 	if expectedOid != cleaned.Oid {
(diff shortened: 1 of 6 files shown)


lfs/upload_queue.go
```diff
@@ -100,7 +101,9 @@ func ensureFile(smudgePath, cleanPath string) error {
 		return err
 	}
 
-	cleaned.Close()
+	if cleaned != nil {
+		defer cleaned.Teardown()
```
rubyist: I don't think this section is quite right, for a couple reasons. Primarily, it's possible for `PointerClean` to return an error with and a valid `cleanedAsset` if the [CopyWithCallback](https://github.com/github/git-lfs/blob/8b3fa6aec1414a00f609f900b934b09d302ef5a1/lfs/pointer_clean.go#L46) returns an error. In this case, `ensureFile` will return without removing the cleaned file.

Less importantly, I don't think we need to defer the `Teardown()`, It's only removing the underlying file and we're not using it before the end of the function.

Maybe something like:
```diff
diff --git a/lfs/upload_queue.go b/lfs/upload_queue.go
index cbd075a..4a26f08 100644
--- a/lfs/upload_queue.go
+++ b/lfs/upload_queue.go
@@ -97,12 +97,12 @@ func ensureFile(smudgePath, cleanPath string) error {
 	}
 
 	cleaned, err := PointerClean(file, stat.Size(), nil)
-	if err != nil {
-		return err
+	if cleaned != nil {
+		cleaned.Teardown()
 	}
 
-	if cleaned != nil {
-		defer cleaned.Teardown()
+	if err != nil {
+		return err
 	}
 
 	if expectedOid != cleaned.Oid {
```

technoweenie:
Okay, I made the recommended change. :metal:

PR #458, opened 2015-07-03
Setup GOPATH if it isn't already set.

### LINKED ISSUE #388: Build scripts broken
I just tried compiling LFS in a clean Ubuntu 14.04 (trusty) Docker container and got an error. Something about package includes (sorry, I don't know much about GO). I got similar errors trying to build it on a Debian 8 machine. To reproduce use the following:

```
docker run -i -t --entrypoint /bin/bash ubuntu:trusty
apt-get install git
git clone https://github.com/github/git-lfs.git
cd git-lfs/
script/debian-build
```

which gives

```
<output omitted>
+ ./script/bootstrap
script/build.go:13:2: cannot find package "github.com/github/git-lfs/lfs" in any of:
	/root/.gvm/gos/go1.4.2/src/github.com/github/git-lfs/lfs (from $GOROOT)
	/root/.gvm/pkgsets/go1.4.2/global/src/github.com/github/git-lfs/lfs (from $GOPATH)
++ echo FAIL
FAIL
```

On CentOS 7 following the same steps I get: 

```
script/build.go:13:2: cannot find package "github.com/github/git-lfs/lfs" in any of:
	/usr/lib/golang/src/github.com/github/git-lfs/lfs (from $GOROOT)
	($GOPATH not set)
FAIL
```

mikemcquaid:
This variable is a bit of a mystery to non-Go developers and it's unclear how you set it up properly. As a result, if it's not set let's set everything up how Go likes it.

Also:
- This will also help keep the Homebrew formula simpler.
- Ignore the `src` directory that we're creating now.
- script/bootstrap: exit on any errors (to be more consistent with other scripts and safer).
- script/bootstrap: remove unneeded `pwd` (the rest of this script assumes it's already in the root of the repository anyway)

Closes #388.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ obj-*
 
 rpm/BUILD*
 rpm/*RPMS
+
+src
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -1,4 +1,5 @@
 #!/bin/sh
+set -e
 
 script/fmt
 rm -rf `pwd`/bin/*
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -1,6 +1,13 @@
 #!/bin/sh
 set -e
 
+if [ -z "$GOPATH" ]; then
+  export GOPATH="$(pwd)"
+  mkdir -p src/github.com/github
+  [ -h src/github.com/github/git-lfs ] ||
+    ln -s "$GOPATH" src/github.com/github/git-lfs
+fi
+
 script/fmt
 rm -rf bin/*
 go run script/*.go -cmd build "$@"
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -1,7 +1,9 @@
 #!/bin/sh
 set -e
 
-if [ -z "$GOPATH" ]; then
+# Only set GOPATH on non-Windows platforms as Windows can't do the symlinking
+# that we need.
+if [ -z "$GOPATH" ] && [ uname -s | grep -vq "_NT-" ]; then
   export GOPATH="$(pwd)"
   mkdir -p src/github.com/github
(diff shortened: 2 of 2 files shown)


gurugray:
:+1:

technoweenie:
We used to do something similar to this, but it didn't work at all for Windows devs. I wouldn't mind this being a homebrew-specific install script though. We have some for debian/centos now.

mikemcquaid:
@technoweenie I've scoped this to non-Windows. That check works on both MinGW and Cygwin.

technoweenie:
That `grep -vq "_NT-"` seems fragile. How long will Windows keep identifying itself as some NT variant?

Otherwise, this seems like a good compromise. If you need to do more than `script/bootstrap`, you should learn about `GOPATH`. We can certainly help in the git lfs docs too.

mikemcquaid:
> Otherwise, this seems like a good compromise. If you need to do more than script/bootstrap, you should learn about GOPATH. We can certainly help in the git lfs docs too.

:+1:

> That grep -vq "_NT-" seems fragile. How long will Windows keep identifying itself as some NT variant?

Given past experience: as long as it uses an NT kernel (or similar). If I was a betting man I'd put :dollar: on this not changing for at least 5-10 years. If @shiftkey has any ideas of a better test here I'm all ears, though!

shiftkey:
> How long will Windows keep identifying itself as some NT variant?

How's " for the foreseeable future" sound?

> We can certainly help in the git lfs docs too.

That'd be great, particularly for curious Windows users...

mikemcquaid:
@technoweenie Thanks!

gurugray:
thanks!

PR #461, opened 2015-07-05
Push deleted files

technoweenie:
This fixes the scanner to find all objects in a commit range, even if an object was added and deleted in that same range. We want all `git lfs push` commands to push every object so that every individual commit is has all the objects it needs.


technoweenie:
@rubyist: This is ready for some :eyeglasses:

rubyist:
:+1: looks good to me

andyneff:
:+1: Thank you for this!

PR #464, opened 2015-07-06
Refactor some of the code in #404

technoweenie:
Some small changes to #404:

1. Implemented @michael-k's suggestion in https://github.com/github/git-lfs/pull/404#discussion-diff-32717780. It's more clear where and how both `isPlink` and `isTortoise` are set.
2. Removed any ssh stuff from `NewEndpoint()`. It now returns the correct `Endpoint` based on the url scheme. I really hope we don't try to support any other strings without a uri scheme ;)

/cc @sinbad


sinbad:
I'm happy with this.

PR #467, opened 2015-07-07
bootstrap: fix syntax error.

mikemcquaid:
@technoweenie Sorry, pushed an old version in #458 with a syntax error. Works now.


technoweenie:
Doh!

PR #468, opened 2015-07-07
Fully re-submit batch requests on 401

rubyist:
When a batch request received a 401, the request was re-run without resetting the Body, causing the request to error and fall back to individual processing. This just reruns the `Batch()` function.

### DIFF
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -200,16 +200,21 @@ func Batch(objects []*objectResource, operation string) ([]*objectResource, *Wra
 	tracerx.Printf("api: batch %d files", len(objects))
 	res, objs, wErr := doApiBatchRequest(req, creds)
 	if wErr != nil {
-		if res != nil {
-			switch res.StatusCode {
-			case 404, 410:
-				tracerx.Printf("api: batch not implemented: %d", res.StatusCode)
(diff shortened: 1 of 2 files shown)


lfs/client.go
```diff
@@ -200,16 +200,21 @@ func Batch(objects []*objectResource, operation string) ([]*objectResource, *Wra
...
-			}
+		if res == nil {
+			sendApiEvent(apiEventFail)
+			return nil, wErr
+		}
+
+		switch res.StatusCode {
+		case 401:
+			Config.SetPrivateAccess()
+			tracerx.Printf("api: batch not authorized, submitting with auth")
```
technoweenie: Perhaps unrelated to this PR, but should this tracerx statement be inside `Config.SetPrivateAccess()`?

rubyist: I might be leaning toward putting a message in both. My thinking is: putting one at this level indicates that the auth failed and it's resubmitting, this is the appropriate level for that message. Having one in `Config.SetPrivateAccess()` shows the config setting being changed, which could come from anywhere (even though this is currently the only place it is called).

technoweenie: Works for me :+1:

PR #469, opened 2015-07-08
Some cleanup to make debian package pass lintian checks

ssgelm:
There were a few issues when running lintian (the debian deb linter) on the package.  This fixes them.

 - Add the machine-readable license header that debian requires
 - Update to the current standards version (this does not break compiling on older debian releases from what I can tell)
 - Add a lintian-overrides file for a few exceptions because golfing doesn't support a few things debian likes.


technoweenie:
Thanks!

PR #472, opened 2015-07-08
Check doHttpRequest() error in UploadObject()

### LINKED ISSUE #462: Intermittent error when pushing
https://gist.github.com/technoweenie/182d3197b22f15942a22

Fun (io?) error: `[signal 0xb code=0x1 addr=0x20 pc=0x947eb]`

Waiting on the Git LFS version...

rubyist:
This method was not checking the error from doHttpRequest() yet still manipulating the reponse object and body. In some scenarios this can cause a panic when trying to read from the nil reponse body.

Fixes #462


technoweenie:
:metal:

PR #475, opened 2015-07-09
Use -text instead of -crlf

technoweenie:
Fixing https://github.com/github/git-lfs/issues/473

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,8 @@ Git LFS uses `.gitattributes` files to configure which are managed by Git LFS.
 Here is a sample one that saves zips and mp3s:
 
     $ cat .gitattributes
-    *.mp3 filter=lfs -crlf
-    *.zip filter=lfs -crlf
+    *.mp3 filter=lfs -text
+    *.zip filter=lfs -text
 
 Git LFS can manage `.gitattributes` for you:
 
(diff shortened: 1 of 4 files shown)


technoweenie:
Apparently we can add macros. There's a builtin `binary` macro:

    [attr]binary -diff -merge -text

What if we did something like this?

```diff
-*.gif filter=lfs diff=lfs merge=lfs -crlf
-*.png filter=lfs diff=lfs merge=lfs -crlf
+[attr]lfs filter=lfs diff=lfs merge=lfs -text
+*.gif lfs
+*.png lfs
```

Though this will take some research to figure out what versions of Git support it. We're targeting Git 1.8.2, which has the `pre-push`.

technoweenie:
I compiled a version of git from the [v1.8.2 tag](https://github.com/git/git/releases/tag/v1.8.2), and it worked fine. I don't want to use macros in Git LFS v0.5.3, but probably for v0.6.0 next month.

shana:
Looks like crlf was renamed to text in 2010 - https://github.com/git/git/commit/5ec3e67052289217c84e53d2cda90d939ac5725b

That goes all the way back to git 1.7.2, so we're good just using text here methinks :smile:

technoweenie:
Cool. I'm merging this, but I hope to use a custom macro in the next major release (v0.6.0).

PR #476, opened 2015-07-09
Batches of n

rubyist:
This PR splits up batch requests into several batches of `n` (currently hardcoded to 100).

- [x] Progress counters are a little off
- [x] Handle errors to the batch call, resubmit batch or track failures
- [x] Test with many thousands of files

### DIFF
--- a/lfs/transfer_queue.go
+++ b/lfs/transfer_queue.go
@@ -1,11 +1,8 @@
 package lfs
 
 import (
-	"fmt"
-	"os"
 	"path/filepath"
 	"sync"
-	"sync/atomic"
 
 	"github.com/github/git-lfs/git"
 	"github.com/github/git-lfs/vendor/_nuts/github.com/rubyist/tracerx"
@@ -27,9 +24,7 @@ type Transferable interface {
 
 // TransferQueue provides a queue that will allow concurrent transfers.
 type TransferQueue struct {
-	filesAdded    int32 // Number of files added to the queue
-	transferIdx   int32 // Used to track transfer number for progress monitor
-	workers       int   // Number of transfer workers to spawn
+	workers       int // Number of transfer workers to spawn
 	transferKind  string
 	errors        []*WrappedError
 	transferables map[string]Transferable
@@ -37,9 +32,8 @@ type TransferQueue struct {
 	apic          chan Transferable  // Channel for processing individual API requests
 	transferc     chan Transferable  // Channel for processing transfers
 	errorc        chan *WrappedError // Channel for processing errors
-	progressc     chan string        // Channel for GIT_LFS_PROGRESS monitor
 	watchers      []chan string
-	monitors      []ProgressTracker
+	monitors      []*ProgressMeter
 	wait          sync.WaitGroup
 }
 
@@ -49,7 +43,6 @@ func newTransferQueue(workers int) *TransferQueue {
 		apic:          make(chan Transferable, batchSize),
 		transferc:     make(chan Transferable, batchSize),
 		errorc:        make(chan *WrappedError),
-		progressc:     make(chan string, batchSize),
 		workers:       workers,
 		transferables: make(map[string]Transferable),
 	}
@@ -61,8 +54,6 @@ func newTransferQueue(workers int) *TransferQueue {
 
 // Add adds a Transferable to the transfer queue.
 func (q *TransferQueue) Add(t Transferable) {
-	atomic.AddInt32(&q.filesAdded, 1)
-
 	q.wait.Add(1)
 	q.transferables[t.Oid()] = t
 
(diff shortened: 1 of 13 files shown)


rubyist:
I think this is ready for some review. There are two main areas of change, the `TransferQueue` and `Batcher`, and the `ProgressMeter`, here's an overview:

## TransferQueue

The `TransferQueue` API mostly stays the same, but the internals are a little different. The main change here is instead of waiting until `Process()` has been called to start up the transfers, the transfers can start as soon as items are `Add()`ed and the caller will use `Wait()` to wait for all of the transfers to finish. The `TransferQueue` is given a total file count and file size up front, but these are used as estimates for the `ProgressMeter`, not as an indication of files that get `Add()`ed.

There is also some refactoring around the code that prevents `workers` number of initial api requests, potentially requesting multiple auth prompts at once. Instead of the complicated `sync.Cond` and confusing mutex locking, we simply launch one api worker and when it has its first success it can launch `workers - 1` more api workers.

## Batcher

When operating in batch mode, the `TransferQueue` will use a `Batcher` to batch incoming transfers into groups of `n` (where `n = 100` currently). In this scenario, `TransferQueue` will block waiting for `Batcher` to queue up `n` requests. Once `n` files are added, `TransferQueue` starts transferring them. At this point, `Batcher` will queue up another 100 files and block until `TransferQueue` pulls its next round. This effectively blocks the scanner meaning we should only keep up to about `n*2` pointers in memory at once rather than reading them all in as before.

## ProgressMeter

Due to the nature of the `Batcher` and `TransferQueue` interaction, we need a better progress meter. The `pb` meter provides more than is needed for this feature and is a bit too inflexible to give us what we want. `ProgressMeter` provides a pretty simple output that makes the transfers and file skipping more obvious. Previously when a file was skipped (because it doesn't need to be transferred) we'd just increment the bytes transferred, making it look like the file was transferred. This confused just about everyone. The new output will appear like this:

```
$ git push origin b3
(5439 of 37164 files, 31725 skipped) 1.30 GB / 8.90 GB, 7.60 GB skipped
```

This is a repo that had some interrupted transfers and files amongst various branches, pushing to a new branch, `b3`. Here, the push checks a total of `37164` Git LFS files, but skips `31725` of them because the sever already has them. The byte counts follow similarly.

The progress code for the meter as well as the `GIT_LFS_PROGRESS` output was pulled out of the `TransferQueue` code to make things a little easier to follow.

lfs/transfer_queue.go
```diff
@@ -24,314 +24,248 @@ type Transferable interface {
...
-// making only one POST call for all objects. The results are then handed
-// off to the transfer workers.
-func (q *TransferQueue) processBatch() error {
-	transfers := make([]*objectResource, 0, len(q.transferables))
-	for _, t := range q.transferables {
-		transfers = append(transfers, &objectResource{Oid: t.Oid(), Size: t.Size()})
-	}
+// legacyFallback is used when a batch request is made to a server that does
+// not support the batch endpoint. When this happens, the Transferables are
+// feed from the batcher into apic to be processed individually.
```
technoweenie: "feed" => "fed"

technoweenie:
This is looking much better. I can't wait to kill all the legacy api client stuff for v0.7.0 :metal:

A couple things come to mind while reading this:

1. It looks like `TransferQueue` will work to drain the batcher's queue as fast as possible. It doesn't seem to wait for every item in the batch to transfer before grabbing the next batch of 100. 
2. Errors seem to collect forever. I wonder if `TransferQueue` should halt early if there are enough errors.
3. It doesn't look like there's a way for the `TransferQueue` to retry failed transfers. We may want to look for common transient errors that can safely be retried.
4. How does this affect the `CopyCallback` stuff? Will native apps still get a reasonably accurate progress log?

rubyist:
1. It will only drain 100 at a time. The channel that feeds transfers from `batchApiRoutine()` to `transferWorker()` has a `batchSize` buffer (ie 100). It will fill that up then block until a transfer is completed. `batchApiRoutine()` only calls `batcher.Next()` once all of the objects from that batch request have been sent to the `transferWorker`. As transfers from batch `A` finish, transfers from batch `B` will trickle into the queue, but batch `C` won't be requested until batch `B` is drained, effectively preventing it from draining any faster. I can add some commenting around that code to make the blocking nature a little more clear.

2. This is true, the batch code has always queued errors until the end. Now that we're realizing people are dealing with very large numbers of files we might want an early halt strategy and/or error counts in the progress output.

3. This is also true, I almost added the ability to push failures back onto the batcher but thought it might go into a separate PR. I think it would go pretty cleanly along with the schema change that adds the errors properties.

4. All of that stuff should still be intact, the progress meter writes out the log file for native apps, if the env var is set. I don't think we have a very strict test regarding the content of that file, though, I'll add one so we can make sure it continues to work.

technoweenie:
1. Okay, I missed that `q.transferc` has a capacity of 100. My main concern is that the time between the batch api completing and the 100th file transferring could extend beyond the expiration timeout. Perhaps this is just something servers need to be aware of. If a user is uploading 100 1GB files, the server should send back a pretty ridiculously high expiration.

I would really like to track timestamps for this stuff. Maybe `transferWorker()` can report to tracerx how long it had to wait since the batch (or individual) API returned the link relation for that object. Could be useful for debugging.

I agree that 2 and 3 should go in separate PRs. I really want to add better testing for the `CopyCallback` for 4, but also not a concern of this PR.

rubyist:
Servers doing a dynamic expiry time is an interesting idea. The concurrent transfer workers could also act as buckets where files are broken up by size ranges so a couple huge files don't hold things up for smaller files.

Tracing timing is def a good idea.

PR #480, opened 2015-07-12
Git for CentOS

andyneff:
-CentOS 6 only has git 1.7 available. So I had to take the existing git rpm, and update it to 2.4.5. Ever feature and capability should have been ported (Emacs, perl, etc...) So that it could probably replace the existing git with no ill effect...
-CentOS 5 was pulling an 1.7 version instead of an available 1.8.2 version. The scripts has been updated to pull the compatible version
-Building no longer fails when sudo in not installed requiring the developer to set the SUDO environment variable
-Building the SRPM no longer fails due to bad permissions
Docs updated
-Many (but far less now) integration tests still fail. I'm using this for active software development in CentOS 6.6, and it does work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,5 +16,7 @@ obj-*
 
 rpm/BUILD*
 rpm/*RPMS
+rpm/*.log
+rpm/SOURCES
 
 src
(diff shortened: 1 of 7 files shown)


technoweenie:
Awesome, thanks!

> Many (but far less now) integration tests still fail. I'm using this for active software development in CentOS 6.6, and it does work.

Once we get Git LFS v0.6.0 shipped, I want to look at getting CI builds to run on a wider variety of OS's to catch stuff like this.

PR #483, opened 2015-07-20
Proposals

technoweenie:
Documents the new proposals directory. It's basically a spot to talk about potential features at a high level.

Also fixes some really minor stuff in the extensions proposal.


rubyist:
:+1:

PR #485, opened 2015-07-21
Dependency linter

rubyist:
A quick script to warn about importing third party packages directly instead of via nut (i.e. vendored).

Example output if you mistakenly import directly (with an exit code of 1):

```
$ ./script/lint
Non vendored dependencies found:
	github.com/cheggaaa/pb
	github.com/rubyist/tracerx
```

Normal output (with an exit code of 0):

```
$  ./script/lint
Looks good!
```

### DIFF
--- a/script/fmt
+++ b/script/fmt
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 formatter=gofmt
 hash goimports 2>/dev/null && {
@@ -12,3 +12,10 @@ for i in */ ; do
     $formatter -w -l "$@" "${i%?}"
   fi
 done
+
+msg=`script/lint`
+if [ $? -ne 0 ];
+then
+  echo "$msg"
+  exit 1
+fi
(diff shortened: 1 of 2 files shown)


technoweenie:
Should we include a message like: "these dependencies should be tracked in nut, with a prefix of 'github.com/github/git-lfs/vendor/_nuts'"?

Also, I feel like this should be combined with `script/fmt` somehow.

```
$ ack script/fmt
script/bootstrap
13:script/fmt

script/build.go
31:	cmd, err := exec.Command("script/fmt").Output()

script/release
2:script/fmt

script/run
2:script/fmt

script/test
5:script/fmt
```

rubyist:
Agree with both points

technoweenie:
I'm thinking of just tacking this on to `script/fmt`. It should fail early, the same way a go syntax error would. It'll get run before builds and tests.

rubyist:
:+1:

technoweenie:
I added the message to script/lint, and tacked it to the end of script/fmt. It only outputs on error, just like script/fmt. I didn't change script/lint to be quiet on success when called directly.

PR #486, opened 2015-07-21
Git LFS Extensions

ryansimmen:
Code for proposal https://github.com/github/git-lfs/blob/master/docs/proposals/extensions.md

### DIFF
--- a/commands/command_ext.go
+++ b/commands/command_ext.go
@@ -29,12 +29,13 @@ func extListCommand(cmd *cobra.Command, args []string) {
 	n := len(args)
 	if n == 0 {
 		printAllExts()
-	} else {
-		config := lfs.Config
-		for _, key := range args {
-			ext := config.Extensions()[key]
-			printExt(ext)
-		}
+		return
+	}
+
+	config := lfs.Config
+	for _, key := range args {
+		ext := config.Extensions()[key]
+		printExt(ext)
 	}
 }
 
--- a/commands/command_ext.go
+++ b/commands/command_ext.go
@@ -1,0 +1,64 @@
+package commands
+
+import (
+	"fmt"
+
+	"github.com/github/git-lfs/lfs"
+	"github.com/github/git-lfs/vendor/_nuts/github.com/spf13/cobra"
+)
+
+var (
+	extCmd = &cobra.Command{
+		Use:   "ext",
+		Short: "View details for all extensions",
+		Run:   extCommand,
+	}
+
+	extListCmd = &cobra.Command{
+		Use:   "list",
+		Short: "View details for specified extensions",
+		Run:   extListCommand,
+	}
+)
+
+func extCommand(cmd *cobra.Command, args []string) {
+	printAllExts()
+}
+
+func extListCommand(cmd *cobra.Command, args []string) {
+	n := len(args)
+	if n == 0 {
+		printAllExts()
+	} else {
+		config := lfs.Config
+		for _, key := range args {
+			ext := config.Extensions()[key]
+			printExt(ext)
+		}
+	}
+}
+
+func printAllExts() {
+	config := lfs.Config
+
+	extensions, err := lfs.SortExtensions(config.Extensions())
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+	for _, ext := range extensions {
+		printExt(ext)
+	}
+}
+
+func printExt(ext lfs.Extension) {
+	Print("Extension: %s", ext.Name)
+	Print("    clean = %s", ext.Clean)
+	Print("    smudge = %s", ext.Smudge)
+	Print("    priority = %d", ext.Priority)
+}
+
+func init() {
+	extCmd.AddCommand(extListCmd)
+	RootCmd.AddCommand(extCmd)
+}
--- a/docs/man/git-lfs-ext.1.ronn
+++ b/docs/man/git-lfs-ext.1.ronn
@@ -14,6 +14,7 @@ during smudge and clean.
 
 * List details for all extensions
 
+    `git lfs ext`
     `git lfs ext list`
 
 * List details for the specified extensions
--- a/docs/man/git-lfs-ext.1.ronn
+++ b/docs/man/git-lfs-ext.1.ronn
@@ -1,0 +1,25 @@
+git-lfs-ext(1) - View extension details
+==============================================================
+
+## SYNOPSIS
+
+`git lfs ext list` [<name>...]
+
+## DESCRIPTION
+
+Git LFS extensions enable the manipulation of files streams
+during smudge and clean.
+
+## EXAMPLES
+
+* List details for all extensions
+
+    `git lfs ext list`
+
+* List details for the specified extensions
+
+    `git lfs ext list 'foo' 'bar'`
+
+## SEE ALSO
+
+Part of the git-lfs(1) suite.
--- a/lfs/extension.go
+++ b/lfs/extension.go
@@ -1,8 +1,16 @@
 package lfs
 
 import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
+	"hash"
+	"io"
+	"os"
+	"os/exec"
 	"sort"
+	"strings"
 )
 
 // An Extension describes how to manipulate files during smudge and clean.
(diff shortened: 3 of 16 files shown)


technoweenie:
I haven't done an exhaustive review of this yet. But I'm glad you were able to pull this off without touching too much of the code base.  Hopefully the extension piping can stay in `lfs.PointerClean` and `lfs. PointerSmudge`. The change in `commands/command_clean.go` is an example of something that I'd love to clean up in a future PR.

Also, what do you think about `git lfs fsck` changing so that it actually runs the smudge extension(s) to confirm that the object is still valid? Right now it just compares the hash of the content on disk to the object ID.

lfs/scanner.go
```diff
@@ -16,7 +16,7 @@ import (
 const (
 	// blobSizeCutoff is used to determine which files to scan for Git LFS
 	// pointers.  Any file with a size below this cutoff will be scanned.
-	blobSizeCutoff = 140
```
technoweenie: I opened an issue to turn this into a dynamic setting based on the repo config: https://github.com/github/git-lfs/issues/524

commands/command_ext.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+)
+
+func extCommand(cmd *cobra.Command, args []string) {
+	printAllExts()
+}
+
+func extListCommand(cmd *cobra.Command, args []string) {
+	n := len(args)
+	if n == 0 {
```
technoweenie: I tend to prefer short guard clauses as opposed to if/else:

```go
if len(args) == 0 {
  printAllExts()
  return
}

config := lfs.Config
for _, key := range args {
  ext := config.Extensions()[key]
  printExt(ext)
}

File: docs/man/git-lfs-ext.1.ronn [hunk 3]
```diff
@@ -0,0 +1,25 @@
...
+## EXAMPLES
+
+* List details for all extensions
+
+    `git lfs ext list`
+
+* List details for the specified extensions
+
+    `git lfs ext list 'foo' 'bar'`
+
```
technoweenie: Do we need the `list` command? What about:

```
# list all extensions
git lfs ext

# show the foo and bar extensions
git lfs ext "foo" "bar"
```

Now, maybe we will want sub commands like "add". In that case, how about:

```
# list all extensions
git lfs ext

# show extensions
git lfs ext show 'foo' 'bar'

# add extension (in future PR)
git lfs ext add "foo" ...
```

lfs/pointer.go
```diff
@@ -89,48 +111,106 @@ func verifyVersion(version string) error {
...
 	}
-	oid := oidParts[1]
 
-	var size int64
-	sizeStr, ok := parsed["size"]
+	value, ok = kvps["size"]
 	if !ok {
-		return nil, errors.New("Invalid Oid")
+		return nil, errors.New("Invalid size")
+	}
```
technoweenie: I think we can skip checking if `kvps` has a `size` key: http://play.golang.org/p/ij4GbJITJ4

```go
value = kvps["size"]
size, err := strconv.ParseInt(value, 10, 0)
if err != nil || size < 0 {
  return nil, fmt.Errorf("Invalid size: %q", value)
}
```

technoweenie:
On a first glance, it's difficult to tell the difference between `Extension` and `PointerExtension`. I think the documentation should be a little more clear. `Extension` objects are parsed from the Git config, while `PointerExtension` objects are parsed from the Git LFS Pointer.

lfs/pointer.go
```diff
@@ -17,27 +18,41 @@ var (
...
 
-func NewPointer(oid string, size int64) *Pointer {
-	return &Pointer{latest, oid, size, oidType}
+type ByPriority []PointerExtension
+
+func (p ByPriority) Len() int           { return len(p) }
+func (p ByPriority) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
+func (p ByPriority) Less(i, j int) bool { return p[i].Priority < p[j].Priority }
+
+func NewPointer(oid string, size int64, exts []PointerExtension) *Pointer {
```
technoweenie: I think this should use `*PointerExtension` where possible, so you don't have to worry about what functions take the direct objects, and which take pointers.

lfs/pointer_smudge.go
```diff
@@ -154,8 +154,76 @@ func readLocalFile(writer io.Writer, ptr *Pointer, mediafile string, cb CopyCall
...
+		var extsR []Extension
+		for i, _ := range exts {
+			ext := exts[len(exts)-1-i]
+			extsR = append(extsR, ext)
+		}
+
+		request := &pipeRequest{"smudge", reader, workingfile, extsR}
+
+		var response pipeResponse
+		response, err = pipeExtensions(request)
```
technoweenie: Is there a reason you're not using `response, err :=` here?

commands/command_ext.go
ryansimmen: That works for me.

docs/man/git-lfs-ext.1.ronn
ryansimmen: I originally wrote the ext command the way you suggested.  @sanoursa and I went back and forth on this.  I don't have a strong opinion either way.  Saeed, would you like to weigh in?

lfs/pointer.go
ryansimmen: Done

lfs/pointer_smudge.go
ryansimmen: err was already declared, but I suppose there's not much harm in declaring it again.

docs/man/git-lfs-ext.1.ronn
sanoursa: I thought we had agreed that list would be optional, so that "git lfs ext" shows all extensions.  My only other thought is that if we want to support additional verbs later, like add, we should make sure that it's not legal to say "git lfs ext foo" now.

lfs/pointer.go
ryansimmen: Agreed

docs/man/git-lfs-ext.1.ronn
ryansimmen: That's correct, both `git lfs ext` and `git lfs ext list` show details for all of the extensions.  I'll make that more clear in the documentation.

lfs/pointer_smudge.go
technoweenie: I think in general, reusing variables declared earlier in a function is fine. I'm asking specifically why `response` is declared right above where it's set when `pipeExtensions()` returns a `pipeResponse`. Totally not a big deal :)

docs/man/git-lfs-ext.1.ronn
technoweenie: I believe the current implementation looks like this now:

```
# list all
git lfs ext

# list all
git lfs ext list

# list some
git lfs ext list "foo"
```

Totally cool :+1: Part of my question was haggling on the difference between `list` and `show`. But since there are no extension details, I think `list` is fine. This is how `git remote` works too:

```
$ git remote
origin

$ git remote show
origin

$ git remote show origin
* remote origin
  Fetch URL: https://github.com/github/git-lfs
  Push  URL: https://github.com/github/git-lfs
  HEAD branch: master
  Remote branches:
...
```

sanoursa: Makes sense.  I'd be totally fine with show if you prefer that.

technoweenie: Yeah thinking about it more today -- I'd only want `show` if the command had more output. Right now it's just a list of extension names.

technoweenie:
So IMO the only things left before merging:

* Change `PointerExtension` to `*PointerExtension` (https://github.com/github/git-lfs/pull/486#discussion_r35574717)
* Fix the merge conflicts (Probably from #476 and #527)

After that, I really want to put this system through its paces. But we should have plenty of time before v0.6.0 for that.

lfs/pointer_smudge.go
ryansimmen: Fixed

docs/man/git-lfs-ext.1.ronn
ryansimmen: It lists the extension names and their configuration details.

ryansimmen:
Merge conflicts are resolved.  Once you give the thumbs up I'll squash the fixup! commit.

lfs/extension.go
```diff
@@ -0,0 +1,181 @@
...
+	for n, ext := range m {
+		p := ext.Priority
+		if _, exist := pMap[p]; exist {
+			err := fmt.Errorf("duplicate priority %d on %s", p, n)
+			return nil, err
+		}
+		pMap[p] = ext
+	}
+
+	var priorities []int
```
WillHipschman: You could combine this loop into the previous one

technoweenie: Good call.

Issue #490, opened 2015-07-22
Roadmap: local storage management


sinbad:
This is a meta issue for the roadmap. Comments should be geared towards the addition and deletion of tasks. Leave detailed discussion to issues linked in tasks. If a task doesn't have an issue next to it, create one and leave a comment here so we can update the list.

# Tasks

## Prune Command
Deletes objects which fall outside retention conditions in order to save disk space on the local clone. Can be invoked specifically (git lfs prune) or via the --prune option on fetch. The following must be retained:

- [x] Objects referenced in the current checkout (**required**)
- [x] Objects referenced by commits which have not been pushed, either to a named remote, or ANY remote, based on config (**required** - data only stored locally must never be deleted)
- [x] Other recent objects that you may want to use again without re-downloading (optional)
  - [x] Objects referenced at recent branches/tags (current date - N days, configurable, 0 to ignore)
  - [x] Objects referenced in other recent commits on HEAD (HEAD commit date - N days, 0 to ignore)
  - [x] Objects referenced in other recent commits of recent branches (latest commit - N days, 0 to ignore)
- [x] Take account of `git worktree` to avoid pruning HEAD objects from any referenced worktree

The reason for so many configurable options here is that deleting something too aggressively may result in large downloads which will frustrate the user. The prune command has to be useful enough to be regularly used to keep local storage at a reasonable level, but without annoying the user with repeat downloads if they happen to switch branches or want to look at recent changes. Omitting more options and only keeping the current HEAD + unpushed would probably make a lot of people not use it after they realised they had to re-download other recent branches.

All options should be kept in gitconfig so the user can specify their general preferences for retention and override them for specific repositories if they want. Having them as config options rather than command line options also means they only have to think about them once, making using the command much easier.

## Fetch command extensions

Update the fetch command so that it has the following features:

1. Fetch objects referenced by a list of refs provided as arguments (currently only 1 supported)
2. Fetch objects for the current checkout (no arguments) - as present
3. Fetch 'recent' objects in addition to either the list of refs or the current checkout, say via a '--recent' option, which could also be set on by default via gitconfig
4. Accept a --prune option to invoke `git lfs prune` afterwards
5. Accept --all to download all objects ever referenced in the history, for the purposes of migration between remotes. Incompatible with the --prune option.
6. Make 'fetch' **only** download objects, not alter the working copy. Introduce a 'checkout' command to ensure working copy files contain the correct content, and a 'pull' command which combines 'fetch' and 'checkout'.

[...]

Issue #460, opened 2015-07-04
git push failing with deleted files


jonathansick:
I tried to push some changes back up to GitHub and got this LFS error log:

```
» git lfs logs last     (preprint) jsick@Helvetica ~/manuscripts/phd_thesis 1 ↵
> git-lfs pre-push origin git@github.com:jonathansick/phd_thesis.git
LocalWorkingDir=/Users/jsick/manuscripts/phd_thesis
LocalGitDir=/Users/jsick/manuscripts/phd_thesis/.git
LocalMediaDir=/Users/jsick/manuscripts/phd_thesis/.git/lfs/objects
TempDir=/Users/jsick/manuscripts/phd_thesis/.git/lfs/tmp
GIT_PREFIX=

Error uploading file figs/resolvedsp/patches/patch_fit_chisq.pdf (a1abaf8316861a74933e5c72de59f4f8e80baa476f17d0505dceea0fec2a69aa)

stat figs/resolvedsp/patches/patch_fit_chisq.pdf: no such file or directory
goroutine 16 [running]:
github.com/github/git-lfs/lfs.Stack(0x0, 0x0, 0x0)
        /Users/rick/github/git-lfs/.vendor/src/github.com/github/git-lfs/lfs/errors.go:85 +0x89
github.com/github/git-lfs/commands.logPanic(0x65daa8, 0xc208025aa0, 0x69da00, 0x0, 0x0)
        /Users/rick/github/git-lfs/.vendor/src/github.com/github/git-lfs/commands/commands.go:156 +0xf17
github.com/github/git-lfs/commands.handlePanic(0x65daa8, 0xc208025aa0, 0x0, 0x0)
        /Users/rick/github/git-lfs/.vendor/src/github.com/github/git-lfs/commands/commands.go:108 +0x6c
github.com/github/git-lfs/commands.LoggedError(0x65daa8, 0xc208025aa0, 0xc208030090, 0x83, 0x0, 0x0, 0x0)
        /Users/rick/github/git-lfs/.vendor/src/github.com/github/git-lfs/commands/commands.go:65 +0x89
github.com/github/git-lfs/commands.Panic(0x65daa8, 0xc208025aa0, 0xc208030090, 0x83, 0x0, 0x0, 0x0)
        /Users/rick/github/git-lfs/.vendor/src/github.com/github/git-lfs/commands/commands.go:75 +0x68
[...]

PR #491, opened 2015-07-22
update roadmap to point to the new local storage management issue

technoweenie:
1. @sinbad is anxious to tackle local storage management: #490. I combined items since they're all covered by his task list.
2. This [Extensions proposal](https://github.com/github/git-lfs/blob/master/docs/proposals/extensions.md) is very important to @ryansimmen (#460). Hooks on hooks on hooks.
3. I demoted client side metrics reporting. While @rubyist added better HTTP stats to the client (#366), I don't know if we need clients reporting this data to servers.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -3,6 +3,13 @@
 This is a high level overview of some of the big changes we want to make for
 Git LFS. If you have an idea for a new feature, open an issue for discussion.
 
+## Releases
+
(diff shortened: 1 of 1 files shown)


rubyist:
> I demoted client side metrics reporting. While @rubyist added better HTTP stats to the client (#366), I don't know if we need clients reporting this data to servers.

:+1: That stuff was mostly for our use, and it's easy enough to get it from someone having trouble. It could serve as a starting point for more automated reporting but I don't think it's something we need to spend any time on now.

PR #510, opened 2015-07-23
Fixed pseudo pointer with extra data test-clean to work on CentOS

andyneff:
Handle special characters (\n) in sh script no matter what distro

(Unfortunately...) Fedora/RHEL/CentOS/etc... use the same executable for sh and bash. echo "\n" just happens to be one of the few cases when it will produces different (wrong?) results. In order to make this always work for everyone, when trying to print escape characters, using printf works

With this patch, ALL tests pass in Debian and CentOS!


technoweenie:
Nice catch. We should start compiling a list of sh gotchas for the tests :)

PR #511, opened 2015-07-23
Docker scripts

andyneff:
A set up docker scripts useful for building, packaging, signing, and testing on multiple Linuxes from a single Linux

@ssgelm I could use your help on adding some Debian repos deb. That's the only part missing here.


docker/README.md
```diff
@@ -0,0 +1,311 @@
...
+
+    ./docker/run_docker.bsh {image(s)} -- bash
+
+After listing the image(s) you want to run, add a double dash (--) and then any 
+command (and arguments) you want executed in the docker. Remember, the command
+you are executing has to be part of the docker image.
+
+### Building Dockers (Optional) ###
+
+```run_dockers.bsh``` calls ```build_dockers.bsh```, but you can still call the
```
technoweenie: This is a fenced code syntax error. You can use just a `.

technoweenie:
I think this is ready to merge once the formatting in the md file is fixed (I left a comment last week about it). I was able to build all the docker VMs and Git LFS packages on my mac. There's probably more to be done here, but since this code isn't compiled and run as part of git lfs, I think it's fine.

andyneff:
There, I think I got them all.

Thanks!

docker/README.md
```diff
@@ -0,0 +1,311 @@
...
+
+## Adding addition OSes ##
+
+To add another operating system, simply follow the already existing pattern, 
+and all the scripts will pick them up. A new Dockerfile should go in a directory
+named
+
+    ./docker/git-lfs_{OS NAME}_{OS VERSION #}
+    
+where **{OS NAME}** and **{OS VERSION #}** should not contain underscores (_).
```
technoweenie: Can you put backticks around the `_`?  The markdown rendering is italicizing the rest of the text below.

PR #517, opened 2015-07-23
Allow for special chars in password

memikequinn:
When using special-ish characters, `><-`, etc, in your password, the string would get garbled and base64 decoding breaks on the test server with the following err: `illegal base64 data at input byte...`

### DIFF
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -691,7 +691,7 @@ func getCreds(req *http.Request) (Creds, error) {
 
 func setRequestAuth(req *http.Request, user, pass string) {
 	token := fmt.Sprintf("%s:%s", user, pass)
-	auth := "Basic " + base64.URLEncoding.EncodeToString([]byte(token))
+	auth := "Basic " + base64.StdEncoding.EncodeToString([]byte(token))
 	req.Header.Set("Authorization", auth)
 }
 


technoweenie:
Do you have a test case? I can't make this fail: http://play.golang.org/p/3DEIuNdvnU

technoweenie:
I wrote [this test case](https://gist.github.com/technoweenie/7b27d4435bdabb6e9676) and can't get it to fail using [code from lfs-test-server](https://github.com/github/lfs-test-server/blob/b9b934bcb1da26cd0e1480e6dff5560db912699d/meta_store.go#L228). Based on my understanding of base64 encoding, changing `base64.URLEncoding` shouldn't have any effect on what characters it can take

memikequinn:
This was my error.  I was trying to decode a string encoded with `base64.URLEncoding` with `base64.StdEncoding`, which is why I was getting an error.  This seems to work with a standard string like `asdf` but not with `asdf-><`.  See this: http://play.golang.org/p/D2FmyaZrs8 
The bottom line is that I should use the same method to decode as was used to encode.  Sorry to waste your time.

technoweenie:
Interesting. My test case decodes with `base64.StdEncoding`, so I didn't think it'd be an issue. We may have to change the lfs test server.  Thanks for the test case! I'll investigate further.

PR #518, opened 2015-07-24
Allow a subset of tests to be run by script/integration with arguments

sinbad:
e.g. `script/integration fetch status` will run only status & fetch tests

### DIFF
--- a/script/integration
+++ b/script/integration
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 . "test/testenv.sh"
 set -e
--- a/script/integration
+++ b/script/integration
@@ -5,6 +5,7 @@ set -e
 
 SHUTDOWN_LFS=no
 SHOW_LOGS=yes
+TESTS=( "$@" )
 
 atexit() {
   res=${1:-$?}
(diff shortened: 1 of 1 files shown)


technoweenie:
You can just run the tests individually. Though I suppose this lets you run multiple tests without compiling the project multiple times or having to specify `SKIP_COMPILE=1`.

CI is showing a syntax error:

    script/integration: 8: script/integration: Syntax error: "(" unexpected

sinbad:
It's just a little more convenient & you get all the usual behaviour like compilation & output/log reporting. Not sure why Travis is complaining, it runs fine here - maybe a different version of 'sh', will check.

sinbad:
Argh, on OS X /bin/sh is actually silently bash (not even a symlink, but checkout 'sh --help'), there is no standard POSIX sh on OS X 10.10 at least. POSIX sh doesn't support arrays which is the issue. 

Since this can potentially cause discrepancies between local testing & Travis, should the scripts actually use /bin/bash instead?

technoweenie:
> Since this can potentially cause discrepancies between local testing & Travis, should the scripts actually use /bin/bash instead?

I don't have a strong opinion. I just want the scripts to run everywhere. I'm not sure changing to `/bin/bash` will hurt the chances of them running on Windows :)

sinbad:
Bash is shipped with the standard msysGit install so Windows should be no worse off. The problem will be the 1% minority Linux distributions which are so optimised they only come with an esoteric shell or precisely the shell you pick on install and no more. I don't think that's a huge deal though, since this only affects contributors and not users, and contributors already have to install other less common things like a Go dev environment.

technoweenie:
@sinbad: We all agreed with your reply in Gitter :+1:

I'd say just convert whatever scripts you need to bash in this PR. We can do a bigger sh to bash PR separately.

sinbad:
Done

PR #519, opened 2015-07-24
Fix install.sh for OS X, FreeBSD, etc.

nuta:
`install(1)`'s `-D` option used in install.sh is a coreutils specific feature. OS X's
implementation does not support it and some other implementation such
as FreeBSD behaves differently.

This PR fixes this problem so that we can use this install script in those OSes.

### DIFF
--- a/script/install.sh.example
+++ b/script/install.sh.example
@@ -8,9 +8,11 @@ elif [ "${BOXEN_HOME:-}" != "" ] ; then
   prefix=${BOXEN_HOME:-}
 fi
 
+mkdir -p $prefix/bin
+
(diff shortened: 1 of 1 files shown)


technoweenie:
Thanks for this. I thought I fixed when we released v0.5.2, but I guess it was a dumb hotfix that never made it in :(

pmidge:
Was this change supposed to make it into `0.5.4`'s `install.sh`? It still fails with the invalid option message. Manually updating `install.sh` allows install to complete.

technoweenie:
@pmidge Ah, looks like this wasn't backported properly.

pmidge:
Thank you!

PR #520, opened 2015-07-24
Fix annoying keychain errors when running integration tests on OS X

sinbad:
This was caused by the fact that you can't disable the osxkeychain helper
if it's set at the system gitconfig (which is the default in recent git),
and the git credential 'store' operation will call ALL helpers, not just
the special one created for the integration tests. Because the test user
environment doesn't have a keychain this caused lots of annoying popup
error messages saying "A keychain could not be found to store X".

### DIFF
--- a/test/testhelpers.sh
+++ b/test/testhelpers.sh
@@ -169,11 +169,11 @@ setup() {
     # user env has no keychain; so create one
     mkdir -p $TMPDIR
     mkdir -p $HOME/Library/Preferences # required to store keychain lists
-    security -v create-keychain -p pass $TMPDIR/temp.keychain
-    security -v list-keychains -s $TMPDIR/temp.keychain
-    security -v unlock-keychain -p pass $TMPDIR/temp.keychain
-    security -v set-keychain-settings -lut 7200 $TMPDIR/temp.keychain
-    security -v default-keychain -s $TMPDIR/temp.keychain
+    security create-keychain -p pass $TMPDIR/temp.keychain
+    security list-keychains -s $TMPDIR/temp.keychain
+    security unlock-keychain -p pass $TMPDIR/temp.keychain
+    security set-keychain-settings -lut 7200 $TMPDIR/temp.keychain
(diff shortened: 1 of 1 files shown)


sinbad:
Hmm, hold off this one for now, seems to be causing LocalItems keychain prompts further down the line, not sure why yet.

sinbad:
Tweaked this a bit so that it doesn't have knock-on effects on the keychain later while still running the integration tests without popups.

technoweenie:
Wow, nice. I have my system credential helper set to something custom, so I never noticed this.

test/testhelpers.sh
```diff
@@ -161,6 +161,21 @@ setup() {
   }
   cp "$HOME/.gitconfig" "$HOME/.gitconfig-backup"
 
+  if [[ `git config --system credential.helper | grep osxkeychain` == "osxkeychain" ]]
```
andyneff: Should this be

    if [[ "`git config --system credential.helper | grep osxkeychain`" == osxkeychain* ]]

to account for cases where someone may have arguments in their credential.helper? (It's at least possible with cache and store)

Also this makes it a bash, So the shebang should be updated to

    #!/usr/bin/env bash

OR

    if [ "`git config --system credential.helper | grep osxkeychain`" == "osxkeychain" ]

if arguments just aren't EVER going to happen with the osxkeychain

test/testhelpers.sh
```diff
@@ -175,6 +190,16 @@ shutdown() {
     if [ -s "$LFS_URL_FILE" ]; then
       curl "$(cat "$LFS_URL_FILE")/shutdown"
     fi
+
+    if [[ `git config --system credential.helper | grep osxkeychain` == "osxkeychain" ]]
```
andyneff: Same

test/testhelpers.sh
sinbad: The osxkeychain doesn't take any arguments. The extra quotes don't seem necessary?

PR #521, opened 2015-07-24
Update deb version to 0.5.3

ssgelm:


technoweenie:
Oh, that's it? I was looking for some config file or something :)

ssgelm:
Yeah, it's crazy simple.

PR #526, opened 2015-07-27
Debain cross compile

andyneff:
Workarounds and patches to get Debian to cross compile 32-bit on a 64-bit Debian (using the -ai386 flag)

@ssgelm Does this look okay?


ssgelm:
I think that looks okay.  I tested on arm also and it built so I don't think this breaks anything on other architectures.

PR #527, opened 2015-07-27
Add checkout & pull commands, changes and fixes to fetch

sinbad:
1. `fetch` no longer updates the working copy, it only downloads files
2. Added `checkout` command which updates the working copy with content if files are missing or are matching pointers. Modified files are left alone (user should reset if they want)
3. Added `pull` command which performs `fetch` and `checkout` for the current ref
4. ~~`fetch` now supports multiple refs as arguments~~
5. `checkout` will correctly populate multiple local files with the same content (old `fetch` would only populate the first one returned from `git rev-list`)
6. `fetch` only adds missing or incorrect objects to the download queue, not all of them. This improves the progress reporting; previously fetch would appear to be trying to download all objects referenced at a ref then only have to download the missing ones as the download queue figured out they were already local. Also means fetch reports nothing at all if you're completely up to date already, instead of reporting "(0 of 1000 files)".
7. A side effect is that 'pull' does all fetching first, then all checkout (not interleaved as before). This is because there may be more objects to checkout than to fetch (missing in working copy but already in store). Coupling these has unwanted edge cases & it's better to simplify & do what it says on the tin i.e. the same as running fetch then checkout.

Note that the ScanTree does something similar to ScanRefs but reports pointers for every file that uses the content and not just one where multiple files share the same content. I'm not sure if this is the intended behaviour elsewhere but for now I've left it as a separate method. It also doesn't have the race conditions on nameMap that I believe ScanRefs has (commented elsewhere).

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -95,15 +95,10 @@ func checkoutWithChan(in <-chan *lfs.WrappedPointer) {
 			continue
 		}
 		// OK now we can (over)write the file content
-		file, err := os.Create(pointer.Name)
+		err = lfs.PointerSmudgeToFile(pointer.Name, pointer.Pointer, nil)
 		if err != nil {
-			Panic(err, "Could not create working directory file")
+			Panic(err, "Could not checkout file")
 		}
-
-		if err := lfs.PointerSmudge(file, pointer.Pointer, pointer.Name, nil); err != nil {
-			Panic(err, "Could not write working directory file")
-		}
-		file.Close()
 
 		updateIdxStdin.Write([]byte(pointer.Name + "\n"))
 	}
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -18,6 +18,16 @@ var (
 )
 
 func checkoutCommand(cmd *cobra.Command, args []string) {
+
+	// No params
+	checkoutAll()
+}
+
+func init() {
+	RootCmd.AddCommand(checkoutCmd)
+}
+
+func checkoutAll() {
 	ref, err := git.CurrentRef()
 	if err != nil {
 		Panic(err, "Could not checkout")
@@ -39,10 +49,7 @@ func checkoutCommand(cmd *cobra.Command, args []string) {
 	}
 	close(c)
 	wait.Wait()
-}
 
-func init() {
-	RootCmd.AddCommand(checkoutCmd)
 }
 
 // Populate the working copy with the real content of objects where the file is
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -19,15 +19,15 @@ var (
 
 func checkoutCommand(cmd *cobra.Command, args []string) {
 
-	// No params
-	checkoutAll()
+	// Parameters are filters
+	checkoutWithIncludeExclude(args, nil)
 }
 
 func init() {
 	RootCmd.AddCommand(checkoutCmd)
 }
 
-func checkoutAll() {
+func checkoutWithIncludeExclude(include []string, exclude []string) {
 	ref, err := git.CurrentRef()
 	if err != nil {
 		Panic(err, "Could not checkout")
@@ -45,11 +45,18 @@ func checkoutAll() {
 
 	checkoutWithChan(c, &wait)
 	for _, pointer := range pointers {
-		c <- pointer
+		if lfs.FilenamePassesIncludeExcludeFilter(pointer.Name, include, exclude) {
+			c <- pointer
+		}
+
 	}
 	close(c)
 	wait.Wait()
 
+}
+
+func checkoutAll() {
+	checkoutWithIncludeExclude(nil, nil)
 }
 
 // Populate the working copy with the real content of objects where the file is
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -1,0 +1,104 @@
+package commands
+
+import (
+	"github.com/github/git-lfs/git"
+	"github.com/github/git-lfs/lfs"
+	"github.com/github/git-lfs/vendor/_nuts/github.com/spf13/cobra"
+	"os"
+	"os/exec"
+	"sync"
+)
+
+var (
+	checkoutCmd = &cobra.Command{
+		Use:   "checkout",
+		Short: "checkout",
+		Run:   checkoutCommand,
+	}
+)
+
+func checkoutCommand(cmd *cobra.Command, args []string) {
+	ref, err := git.CurrentRef()
+	if err != nil {
+		Panic(err, "Could not checkout")
+	}
+
+	pointers, err := lfs.ScanRefs(ref, "", nil)
+	if err != nil {
+		Panic(err, "Could not scan for Git LFS files")
+	}
+
+	var wait sync.WaitGroup
+	wait.Add(1)
+
+	c := make(chan *lfs.WrappedPointer)
+
+	checkoutWithChan(c, wait)
+	for _, pointer := range pointers {
+		c <- pointer
+	}
+	close(c)
+	wait.Wait()
+}
+
+func init() {
+	RootCmd.AddCommand(checkoutCmd)
+}
+
+// Populate the working copy with the real content of objects where the file is
+// either missing, or contains a matching pointer placeholder, from a list of pointers.
+// If the file exists but has other content it is left alone
+// returns immediately but a goroutine listens on the in channel for objects
+// calls wait.Done() when the final item after the channel is closed is done
+func checkoutWithChan(in <-chan *lfs.WrappedPointer, wait sync.WaitGroup) {
+	go func() {
+		// Fire up the update-index command
+		cmd := exec.Command("git", "update-index", "-q", "--refresh", "--stdin")
+		updateIdxStdin, err := cmd.StdinPipe()
+		if err != nil {
+			Panic(err, "Could not update the index")
+		}
+
+		if err := cmd.Start(); err != nil {
+			Panic(err, "Could not update the index")
+		}
+
+		// As files come in, write them to the wd and update the index
+		for pointer := range in {
+
+			// Check the content - either missing or still this pointer (not exist is ok)
+			filepointer, err := lfs.DecodePointerFromFile(pointer.Name)
+			if err != nil && !os.IsNotExist(err) {
+				if err == lfs.NotAPointerError {
+					// File has non-pointer content, leave it alone
+					continue
+				}
+				Panic(err, "Problem accessing %v", pointer.Name)
+			}
+			if filepointer != nil && filepointer.Oid != pointer.Oid {
+				// User has probably manually reset a file to another commit
+				// while leaving it a pointer; don't mess with this
+				continue
+			}
+			// OK now we can (over)write the file content
+			file, err := os.Create(pointer.Name)
+			if err != nil {
+				Panic(err, "Could not create working directory file")
+			}
+
+			if err := lfs.PointerSmudge(file, pointer.Pointer, pointer.Name, nil); err != nil {
+				Panic(err, "Could not write working directory file")
+			}
+			file.Close()
+
+			updateIdxStdin.Write([]byte(pointer.Name + "\n"))
+		}
+
+		updateIdxStdin.Close()
+		if err := cmd.Wait(); err != nil {
+			Panic(err, "Error updating the git index")
+		}
+		wait.Done()
+	}()
+
+}
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -1,12 +1,13 @@
 package commands
 
 import (
-	"github.com/github/git-lfs/git"
-	"github.com/github/git-lfs/lfs"
-	"github.com/github/git-lfs/vendor/_nuts/github.com/spf13/cobra"
 	"os"
 	"os/exec"
 	"sync"
+
+	"github.com/github/git-lfs/git"
+	"github.com/github/git-lfs/lfs"
+	"github.com/github/git-lfs/vendor/_nuts/github.com/spf13/cobra"
 )
 
 var (
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -8,6 +8,7 @@ import (
 	"github.com/github/git-lfs/git"
 	"github.com/github/git-lfs/lfs"
 	"github.com/github/git-lfs/vendor/_nuts/github.com/spf13/cobra"
+	"github.com/rubyist/tracerx"
 )
 
 var (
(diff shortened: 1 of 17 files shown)


sinbad:
Hrm, didn't spot the pre-push test fail until now, will investigate.

lfs/util.go
```diff
@@ -86,3 +88,64 @@ func wrapProgressError(err error, event, filename string) error {
...
+// Return whether a given filename passes the include / exclude path filters
+// Only paths that are in includePaths and outside excludePaths are passed
+// If includePaths is empty that filter always passes and the same with excludePaths
+// Both path lists support wildcard matches
+func FilenamePassesIncludeExcludeFilter(filename string, includePaths, excludePaths []string) bool {
+	if len(includePaths) == 0 && len(excludePaths) == 0 {
+		return true
+	}
+
+	// For Win32, becuase git reports files with / separators
```
technoweenie: Typo

commands/command_checkout.go
```diff
@@ -0,0 +1,118 @@
...
+	checkoutWithIncludeExclude(nil, nil)
+}
+
+// Populate the working copy with the real content of objects where the file is
+// either missing, or contains a matching pointer placeholder, from a list of pointers.
+// If the file exists but has other content it is left alone
+// returns immediately but a goroutine listens on the in channel for objects
+// calls wait.Done() when the final item after the channel is closed is done
+func checkoutWithChan(in <-chan *lfs.WrappedPointer, wait *sync.WaitGroup) {
+	go func() {
```
technoweenie: Any thoughts on making this a normal function, and calling `go checkoutWithChan(c, &wait)` above? I think it makes it more clear in `checkoutWithIncludeExclude()` that it's running as a goroutine.

commands/command_checkout.go
```diff
@@ -0,0 +1,118 @@
...
+				// while leaving it a pointer; don't mess with this
+				continue
+			}
+			// OK now we can (over)write the file content
+			file, err := os.Create(pointer.Name)
+			if err != nil {
+				Panic(err, "Could not create working directory file")
+			}
+
+			if err := lfs.PointerSmudge(file, pointer.Pointer, pointer.Name, nil); err != nil {
```
technoweenie: Is it worth defering `file.Close()` above this, so Go closes the file handle before it panics? I'm actually not sure if Go panics will leave file handles open or not.

commands/command_pull.go
```diff
@@ -0,0 +1,34 @@
...
+	ref, err := git.CurrentRef()
+	if err != nil {
+		Panic(err, "Could not pull")
+	}
+
+	// Previously we would only checkout files that were downloaded, as they
+	// were downloaded. However this would ignore files where the content was
+	// already present locally (since these are no longer included in transfer Q for
+	// better reporting purposes).
+	// So now we do exactly what we say on the tin, fetch then a separate checkout
```
technoweenie: I think this comment is unnecessary. Old commits and PRs provide the history lesson. :+1: on the approach though.

lfs/util.go
```diff
@@ -86,3 +88,64 @@ func wrapProgressError(err error, event, filename string) error {
 
 	return nil
 }
+
+// Return whether a given filename passes the include / exclude path filters
+// Only paths that are in includePaths and outside excludePaths are passed
+// If includePaths is empty that filter always passes and the same with excludePaths
+// Both path lists support wildcard matches
+func FilenamePassesIncludeExcludeFilter(filename string, includePaths, excludePaths []string) bool {
```
technoweenie: This func could use some golang tests.

lfs/util.go
```diff
@@ -86,3 +88,64 @@ func wrapProgressError(err error, event, filename string) error {
...
+
+		}
+	}
+
+	return true
+}
+
+// Are we running on Windows? Need to handle some extra path shenanigans
+func IsWindows() bool {
+	return runtime.GOOS == "windows"
```
technoweenie: I think we should cache this somehow.  Even just having a package level var. We don't need to check this on every iteration of the include/exclude path arrays.

technoweenie:
I think this looks good. Nothing major in my comments.

@rubyist: Would love for you to check this out too. What do you think about removing `ScanRefs()` in favor of `ScanTree()`? That should be a quick PR as long as the `ScanTree()` behavior is the same. If this happens though, I'd probably rename `ScanTree()` to `ScanRefs()`.

rubyist:
I'll review more thoroughly shortly, but just from reading the text here the big difference between `ScanRefs()` and `ScanTree()` is that the former only cares about the Git LFS files themselves where `ScanTree()` is also concerned with mapping Git LFS files to working directory files, am I correct in that?

I think `ScanTree()` is good and definitely works here, and would probably be a good fit for the `ls-files` and `fsck` commands. However, the transferring functions, `fetch`, `pre-push`, etc, aren't concerned with with working directory files. I suspect we can't just replace one with the other as is (since `fetch` was left using `ScanRefs()` in this PR). I'm not sure if combining the functions would be very clear, but they could perhaps use better names.

technoweenie:
Hmm, I think you're right about `ScanTree()` vs `ScanRefs()` :+1:

commands/command_checkout.go
sinbad: Yeah, I was just following the pattern that had been used elsewhere (in places like catFileBatch) but I realise now that's because there's some extra startup code before the grouting starts, where in this case it's just a goroutine. I'll change it.

commands/command_checkout.go
sinbad: This code was just moved verbatim from fetch. The problem with using defer here is that it's in a loop so none of the handles would be released until all the files are done which has its own issues. I'll refactor it & create lfs.PointerSmudgeFile which wraps the create & write, and the defer is localised per file.

commands/command_pull.go
sinbad: I dunno, I find source comments head off misunderstandings better, since they're right there when you wonder why something is done that way. People don't always want to start digging through old PRs / commits to find explanations when they think they've spotted something, I've seen many an accidental regression that way.

lfs/util.go
sinbad: Yeah, I was thinking this because they probably need to run on Windows too for complete coverage.

lfs/util.go
sinbad: OK will do, but FYI it's not expensive, will fail on the first byte.

sinbad:
Yes, on reflection it's better to keep ScanRefs as well and just be clear that it works uniquely on blobs whilst ScanTree works uniquely on tree paths (otherwise they do the same thing). It only really makes a difference when the same content is used in multiple files. ScanTree is slightly more efficient for a single ref than ScanRefs (since it can collect size at the same time without needing cat-file --batch-check), but it can't do a commit range, and for cases like fetch/push you're only interested in the unique oids anyway.

Although thinking about it, unique blob shas will not continue to equate 1:1 to unique oids longer term if the pointer metadata changes between versions so ScanRefs may not always return unique oids without further filtering.

lfs/util.go
technoweenie: We have no Windows CI yet :/  But yes, very important there. Now that `script/bootstrap` should work on Windows, I'll resume the search for a CI system ([AppVeyor](http://www.appveyor.com/)?).

commands/command_pull.go
rubyist: What the comment doesn't mention is the reason for doing the checkout concurrently, which was that it made a significant difference in total run time, especially when processing large files. There are tradeoffs between the two approaches and if we're going to put in comments like this such tradeoffs should be mentioned.

commands/command_checkout.go
```diff
@@ -0,0 +1,111 @@
+package commands
+
+import (
+	"github.com/github/git-lfs/git"
+	"github.com/github/git-lfs/lfs"
+	"github.com/github/git-lfs/vendor/_nuts/github.com/spf13/cobra"
+	"os"
+	"os/exec"
+	"sync"
+)
```
rubyist: This import block isn't synced up with what `goimports` will produce.

commands/command_fetch.go
```diff
@@ -1,21 +1,17 @@
...
-	"os"
-	"os/exec"
-	"sync"
-	"time"
-
 	"github.com/github/git-lfs/git"
 	"github.com/github/git-lfs/lfs"
 	"github.com/github/git-lfs/vendor/_nuts/github.com/rubyist/tracerx"
 	"github.com/github/git-lfs/vendor/_nuts/github.com/spf13/cobra"
+	"time"
```
rubyist: This import block isn't synced up with what `goimports` will produce.

commands/command_pull.go
sinbad: The problem is I don't know that it's even possible to use the old approach and be functionally complete as a checkout, it would ignore files that were already present in the store but not the working copy for example. Maybe you could parallelise the things that did need to be downloaded but you can't avoid having to check that there aren't things that didn't need to be downloaded (or were in multiple locations in the file system).

commands/command_checkout.go
sinbad: Can you explain? I can't see anything wrong with it, and the ordering is as per gofmt.

rubyist: `script/fmt` prefers `goimports` (but falls back to `go fmt`) (#303). `goimports` (kind of annoyingly) has a slightly different way of ordering import blocks. It outputs this block as such:

```go
import (
	"os"
	"os/exec"
	"sync"

	"github.com/github/git-lfs/git"
	"github.com/github/git-lfs/lfs"
	"github.com/github/git-lfs/vendor/_nuts/github.com/rubyist/tracerx"
	"github.com/github/git-lfs/vendor/_nuts/github.com/spf13/cobra"
)
```

So when any of the scripts are run that call `script/fmt` (pretty much all of them) on a machine that has `goimports` installed, it'll end up changing the file.

sinbad: OK never mind, added goimports to my Sublime save action & see it likes to separate stdlib & others. Do the contributor styles notes need updating?

rubyist: :+1: for adding that to the contributor notes

rubyist:
I'm :+1: overall, and agree with all the previous comments. I have two other pieces of feedback that could be discussed / addressed in separate PRs if needed.

**Checkout**

I'm wondering if `checkout` should have some kind of progress or feedback. In a repo with around 5000 reasonably sized files a complete checkout takes just under 3 minutes on my machine. Given the nature of `git update-index`, we don't really know when git has finished with it so doing a counter or percentage progress meter might not make a lot of sense, but we could output something that indicates we're working.

**Pull**

I'm not convinced the `pull` operation can't be done concurrently. The `fetch` can be started and return a channel it uses to pass pointers across to `checkout`. The `checkout` code uses `ScanTree()` to map oids to the file(s) that contain that data then uses that data to look up the pointers that come across from `fetch`. In the case of files already in the store the pointer can be passed over to `checkout` immediately when it skips the download queue. Does this cover all the bases?

I have some very quick and dirty PoC code in [this gist](https://gist.github.com/rubyist/4233626ac6d3acaf2a10) that might be a start. In a quick test, a repo with 1000 reasonably sized files and 200 "duplicates" (same data, different name) the serial `pull` took 10:28 and the concurrent version took 9:15 (I removed all of the files from the working directory and the entire `.git/lfs/objects` directory for each test). Everything was also properly restored when the files were missing from the wd but present in the store. What other edge cases should be checked?

sinbad:
Just discovered a bunch of issues related to paths when the user is in a subdir - I think this would have affected `fetch` before if run in a subdir (pointer.Name was rooted but files written as relative to cwd). I've fixed the most important ones (will write tests tomorrow).

However a Q: right now `checkout` operates on the entire repo even if you run it in a subdir, I think that's the best default especially as `pull` should always operate like that (and it includes `checkout`). However you can filter what `checkout` does with arguments so I think I should probably support `.` and `./` arguments to restrict `checkout` to the cwd. Sound ok?

sinbad:
Parallel commenting :) I'll take a look at the gist & think again about parallel fetch/checkout tomorrow, thanks.

sinbad:
Thanks for the gist, I've just tweaked things slightly (mainly function names & comments) and merged it in. 

I've also added progress reporting for `checkout` to my roadmap issue, I think I'd only do this for standalone checkout & not `pull` to avoid mixed progress updates from `fetch` & `checkout`.

I'm just going to add more tests around the cwd issues and checkout filtering args & then this is ready to be merged if you're happy.

sinbad:
Done, ready to be merged. Apologies for the force re-push on that last commit, Travis had a transient error about "fatal: empty ident name" which was nonsense, running things again works fine. Not sure what that's about.

rubyist:
Sorry, I conflicted your PR! The fixes are pretty easy, though. `command_fetch.go` is pretty straight forward, `transfer_queue.go` can just keep everything from `HEAD`, `test-pre-push.sh` can keep everything from this PR, and the following changes made to `progress_meter.go` will make it all pass:

```diff
diff --git a/lfs/progress_meter.go b/lfs/progress_meter.go
index c664054..26db777 100644
--- a/lfs/progress_meter.go
+++ b/lfs/progress_meter.go
@@ -84,7 +84,7 @@ func (p *ProgressMeter) Finish() {
 	close(p.finished)
 	p.update()
 	p.logger.Close()
-	if p.show {
+	if p.show && p.estimatedBytes > 0 {
 		fmt.Fprintf(os.Stdout, "\n")
 	}
 }
@@ -116,7 +116,7 @@ func (p *ProgressMeter) writer() {
 }
 
 func (p *ProgressMeter) update() {
-	if !p.show {
+	if !p.show || p.estimatedFiles == 0 {
 		return
 	}
 
```

sinbad:
@rubyist np, I spotted it and had most of it resolved, was just trying to figure out how to early-out on no transfers so thanks for the diff on that :) Should all be fine now.

rubyist:
Looks good to me :+1:

lfs/util_test.go
```diff
@@ -57,3 +57,80 @@ func TestCopyWithCallback(t *testing.T) {
...
+	assert.Equal(t, false, FilenamePassesIncludeExcludeFilter("test/filename.dat", []string{"test"}, []string{"test/filename.dat"}))
+	assert.Equal(t, true, FilenamePassesIncludeExcludeFilter("test/filename.dat", []string{"test/*"}, []string{"test/notfile*"}))
+	assert.Equal(t, false, FilenamePassesIncludeExcludeFilter("test/filename.dat", []string{"test/*"}, []string{"test/file*"}))
+	assert.Equal(t, false, FilenamePassesIncludeExcludeFilter("test/filename.dat", []string{"another/*", "test/*"}, []string{"test/notfilename.dat", "test/filename.dat"}))
+
+	if IsWindows() {
+		// Extra tests because Windows git reports filenames with / separators
+		// but we need to allow \ separators in include/exclude too
+		// Can only test this ON Windows because of filepath behaviour
+		// Inclusion
```
technoweenie: First, I love that this feature is being snuck in like this. I don't see a way to configure this via git config or a flag on `git lfs checkout`, but that can come later. If I'm understanding correctly, the only thing that uses this right now is the checkout command when run from a subdirectory. Also, big props on thinking about Windows right now, instead of waiting for the inevitable bug report. :metal:

However, I'm a bit concerned about the duplication. It'd be cool if the assertions could run with `/` separated paths and `\` separated paths, without having to specify everything twice. Though, that will require a nasty nested array structure of examples, making this test more difficult to read.

Also, when we start thinking about specifying default included/excluded paths in the git config, we will probably want to support both `/` and `\` so that users on different OS's can work together.

technoweenie:
I played around with the commands a bit. They feel great. :metal:

lfs/util_test.go
sinbad: The include/exclude gitconfigs are for later `fetch` & `prune` work rather than checkout, since the primary intention is to reduce download time & disk space (if you already have the data locally there's no harm in checking it out). However the args to checkout use exactly the same logic as inclusions so it made sense to include this as a general feature now. I'd like to preserve the consistency with `git checkout` and accept just pathspecs as arguments rather than use --include / --exclude params, it's nice when the commands feel the same as core git I think. 

I didn't think duplication was much of an issue in test code but take the point, I'll generalise it.

technoweenie: > I'd like to preserve the consistency with git checkout and accept just pathspecs as arguments rather than use --include / --exclude params, it's nice when the commands feel the same as core git I think.

Works for me :+1:

Just as a counter point, we changed `git lfs add` to `git lfs track` because people were using it like they use `git add`. But I think the similar names makes a lot of sense for the fetch/checkout/pull trio.

sinbad: Yeah I think when the command is conceptually different it totally makes sense to use different names / conventions.

PR #528, opened 2015-07-27
Test api schema

technoweenie:
This adds some basic json schema tests:

* The documented json schema values are all valid json schemas
* The schemas are tested with examples of expected valid or invalid json documents.

How do the tests look? Is it worth bringing in 5 tiny packages to validate json?


sinbad:
Hmm, this feels a little heavyweight for testing that the schema is valid & expresses what you think it does. It's certainly the most correct way to do it but it almost feels like maybe it should be separate? Apropos it would be nice to be able to do:

`curl https://myservers/rest/api/foo | lfs-api-validate`

...to easily prove whether a server is fulfilling the API properly. A different test I know and you still need to assert that the schema itself is correct but I wonder if that and a lfs-api-validate tool would sit better in their own repo without cluttering the client.

technoweenie:
Yeah, keeping it separate is probably a good call. I do want some kind of lint/test suite for servers too, eventually.

PR #530, opened 2015-07-27
teach 'uninit' how to remove git config sections

### LINKED ISSUE #529: git lfs uninit does not completely remove its traces from ~/.gitconfig
After running `git lfs init`, my `~/.gitconfig` ends like this:
```
[filter "lfs"]
	clean = git-lfs clean %f
	smudge = git-lfs smudge %f
	required = true
```
If I subsequently run `git lfs uninit`, my `~/.gitconfig` still ends with the line:
```
[filter "lfs"]
```
Running `git lfs init`, followed by `git lfs uninit`, repeatedly will result in another appended filter section line surviving each time:
```
[filter "lfs"]
[filter "lfs"]
[filter "lfs"]
```
[etc.]

technoweenie:
We need to run `git config` with the `--remove-section` arg. Fixes #529

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -87,6 +87,10 @@ func (c *gitConfig) UnsetGlobal(key string) {
 	simpleExec("git", "config", "--global", "--unset", key)
 }
 
+func (c *gitConfig) UnsetGlobalSection(key string) {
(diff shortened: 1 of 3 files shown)


rubyist:
:+1:

PR #534, opened 2015-07-28
newClientRequest() now accepts an initial header

### LINKED ISSUE #533: Client prompts for username/password auth after obtaining a token via SSH
`newApiRequest()` calls [`sshAuthenticate()`](https://github.com/github/git-lfs/blob/master/lfs/client.go#L557) to retrieve a token with git-lfs-authenticate, but the following [`newClientRequest()` call](https://github.com/github/git-lfs/blob/master/lfs/client.go#L573) always [invokes `getCreds()`](https://github.com/github/git-lfs/blob/master/lfs/client.go#L595). This results in a password prompt if a Git credential cacher isn't configured.

technoweenie:
This allows newClientRequest() to set an Authorization header from either an SSH response or link relation without calling git credentials. Fixes #533 

/cc @jatoben

### DIFF
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -581,10 +581,8 @@ func newClientRequest(method, rawurl string, header map[string]string) (*http.Re
 		return nil, nil, err
 	}
 
-	if header != nil {
-		for key, value := range header {
-			req.Header.Set(key, value)
-		}
+	for key, value := range header {
+		req.Header.Set(key, value)
 	}
 
 	req.Header.Set("User-Agent", UserAgent)
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -60,13 +60,9 @@ func (o *objectResource) NewRequest(relation, method string) (*http.Request, Cre
 		return nil, nil, objectRelationDoesNotExist
 	}
 
-	req, creds, err := newClientRequest(method, rel.Href)
+	req, creds, err := newClientRequest(method, rel.Href, rel.Header)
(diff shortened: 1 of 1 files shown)


technoweenie:
@sinbad/@rubyist: I'd appreciate a review, since you two have worked in this part of the codebase. It's a pretty trivial patch, but I still feel weird about merging it by myself.

lfs/client.go
```diff
@@ -570,27 +566,27 @@ func newApiRequest(method, oid string) (*http.Request, Creds, error) {
...
 }
 
-func newClientRequest(method, rawurl string) (*http.Request, Creds, error) {
+func newClientRequest(method, rawurl string, header map[string]string) (*http.Request, Creds, error) {
 	req, err := http.NewRequest(method, rawurl, nil)
 	if err != nil {
 		return nil, nil, err
 	}
 
+	if header != nil {
```
rubyist: This `nil` check isn't necessary. When passing `nil`, `header` ends up being the zero value for a `map[string]string` which can be ranged over just fine.

rubyist:
:+1:

sinbad:
Looks good to me, not sure why Travis is upset after that latest commit though, tests are green here.

technoweenie:
Intermittent test failures are my favorite :trollface: I've started putting some more details, but it hasn't been helping. I'll probably try a PR that changes them all to bash today.

PR #536, opened 2015-07-28
SHA in user agent

technoweenie:
I want the version string to include the git commit that built it:

```
$ git lfs version
git-lfs/0.5.3 (GitHub; darwin amd64; go 1.3; git 0cf7c47)
```

This PR gets it working with `script/run version`, but not `script/bootstrap`.

```
$ script/bootstrap
# github.com/github/git-lfs
/opt/boxen/chgo/versions/1.3/pkg/tool/darwin_amd64/6l: unknown flag -X github.com/github/git-lfs/lfs.GitCommit 22bb4ab

2015/07/28 16:04:56 exit status 2
exit status 1
```

I'm really not sure why it [fails here](https://github.com/github/git-lfs/blob/0cf7c479e403deee9f8e0cdafa9b6597d22fab71/script/build.go#L132-L139) when it works in `script/run` or running `go build` manually.

A workaround is to create a file like `lfs/aaaaa_version.go`:

```go
package lfs

func init() {
  GitCommit = "22bb4ab"
}
```

Such a file could be created and then trashed during every `script/bootstrap` run.

### DIFF
--- a/git-lfs.go
+++ b/git-lfs.go
@@ -7,6 +7,5 @@ import (
 
 func main() {
 	commands.Run()
-
 	lfs.LogHttpStats()
 }
--- a/lfs/lfs.go
+++ b/lfs/lfs.go
@@ -11,12 +11,8 @@ import (
 	"github.com/github/git-lfs/vendor/_nuts/github.com/rubyist/tracerx"
 )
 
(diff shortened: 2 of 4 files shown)


rubyist:
Excellent :+1:

PR #538, opened 2015-07-29
Correctly detect the absence of `ronn`

sinbad:
Just a little fix: without the quotes the -x test passes and you get a weird error along the lines of:
`script/man: line 10: man/git-lfs-checkout.1.ronn: Permission denied`
..if ronn isn't installed, instead of the correct error message: 
`Install the 'ronn' ruby gem to build the man pages.`

### DIFF
--- a/script/man
+++ b/script/man
@@ -1,7 +1,7 @@
 #!/bin/sh
 ronn=`which ronn`
 
-if [ -x $ronn ]; then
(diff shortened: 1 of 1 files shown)


technoweenie:
Ah thanks. My shell knowledge has leveled up a lot since I wrote this :)

PR #540, opened 2015-07-30
Fix test race condition

technoweenie:
This attempts to fix the race condition regarding `$HOME` in the integration tests.

Now, an initial `$TESTHOME` is setup at `$REMOTEDIR/home`. It has the starting git config and OSX keychain (if needed). `begin_test()` resets `$HOME` to `$TRASHDIR/home`, which is unique per test script. Tests run sequentially in a single script.

I also ran into an issue where `begin_test()` was silently failing, causing the scripts to bail after the first test. So, I added something in `shutdown()` that ensures every test runs by grepping the file for all `begin_test()` calls.

### DIFF
--- a/script/backport-pr
+++ b/script/backport-pr
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 #
 # Backports a PR into a release branch:
 #
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 set -e
 
 # Only set GOPATH on non-Windows platforms as Windows can't do the symlinking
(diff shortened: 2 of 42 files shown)


technoweenie:
Also updated all the scripts to use `#!/usr/bin/env bash`. There are a few hold outs:

* rpm/SPECS/git.spec
* The Git pre-push hook that Git LFS installs uses `#!/bin/sh`. I'll change that in another pr.

There were a few that pass args to the shebang. I changed those to use the same standard shebang with an explicit `set` for the args.

technoweenie:
Feels better, but not a complete fix :(  

* https://travis-ci.org/github/git-lfs/builds/73308232

rubyist:
Doh, this is my fault. #544

technoweenie:
With @sinbad's help, I was able to replicate the issue that he fixed in #520. We both confirmed that this PR doesn't undo those fixes.

PR #543, opened 2015-07-30
Support `git lfs checkout .` and similar correctly

sinbad:
Previously using `.` `./` and `.\` would only behave as expected in subdirs
(because calculating the relative dir would resolve it), using them in the
root would match nothing. Behave as expected instead.

### DIFF
--- a/lfs/util.go
+++ b/lfs/util.go
@@ -115,6 +115,11 @@ func FilenamePassesIncludeExcludeFilter(filename string, includePaths, excludePa
 	if len(includePaths) > 0 {
 		matched := false
 		for _, inc := range includePaths {
+			// Special case local dir, matches all (inc subpaths)
+			if inc == "." || inc == "./" || inc == ".\\" {
+				matched = true
+				break
+			}
 			matched, _ = filepath.Match(inc, filename)
 			if !matched && IsWindows() {
 				// Also Win32 match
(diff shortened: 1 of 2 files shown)


lfs/util.go
```diff
@@ -115,6 +115,11 @@ func FilenamePassesIncludeExcludeFilter(filename string, includePaths, excludePa
 	if len(includePaths) > 0 {
 		matched := false
 		for _, inc := range includePaths {
+			// Special case local dir, matches all (inc subpaths)
+			if inc == "." || inc == "./" || inc == ".\\" {
+				matched = true
+				break
+			}
```
technoweenie: Any thoughts on putting this in some top level var?

```go
var matchAllPaths = []string{".", "./", ".\\"}

func something(s string) {
  if _, ok := matchAllPaths[s]; ok {

  }
}
```

sinbad: Good call, done as a map[string]struct{} (set idiom)

technoweenie: Doh, I meant a `map[string]bool`, not a slice. But yours is fine. I'd just move it outside the func so that it doesn't get created every time `FilenamePassesIncludeExcludeFilter()` is called.

sinbad: I was trying to avoid polluting the namespace unnecessarily but I guess it'll get called a lot so worth it.

BTW using a map[]struct{} instead of a map[]bool saves 1 byte per entry since the empty struct takes 0 bytes. \o/ :D

technoweenie: A byte saved is a byte earned.

sinbad: I'd better not spend my 3 bytes all at once

PR #544, opened 2015-07-30
Suppress progress output for a push dry run

rubyist:
There should be a `SuppressProgress()` call on the upload queue when it's doing a dry run. I added it to the `pre-push` stuff but overlooked it for `push`. It only fails intermittently because the progress meter output is written in a goroutine, so during a dry run it's possible that the `push` command exits before the goroutine ever runs. On slower or more utilized machines, like this CI, it gets a chance to run and shows progress output.


technoweenie:
:+1: I like the second commit better.

PR #546, opened 2015-07-31
git worktree support (git 2.5+)

sinbad:
When using `git worktree`, store lfs objects in the same shared location as the git commit objects, rather than duplicating them. Temp files are still kept per worktree.

This requires resolving an extra level of indirection on the git dir; with worktree, the first level, identified by GIT_DIR when setting up and by a git-dir: redirect in the `.git` file afterwards, points at a git dir which just has an index and other worktree-specific data. A `commondir` file in that git dir points to where the git objects are stored, which is common with the original repo the worktree was created from. We store lfs files alongside the commit objects in this case, sharing with all worktrees.


lfs/lfs.go
```diff
@@ -103,9 +103,10 @@ func init() {
 
 	LocalWorkingDir, LocalGitDir, err = resolveGitDir()
 	if err == nil {
-		LocalMediaDir = filepath.Join(LocalGitDir, "lfs", "objects")
+		LocalGitStorageDir = resolveGitStorageDir(LocalGitDir)
+		LocalMediaDir = filepath.Join(LocalGitStorageDir, "lfs", "objects") // objects across all worktrees
 		LocalLogDir = filepath.Join(LocalMediaDir, "logs")
-		TempDir = filepath.Join(LocalGitDir, "lfs", "tmp")
+		TempDir = filepath.Join(LocalGitDir, "lfs", "tmp") // temp files per worktree
```
technoweenie: Should these var comments go next to the declaration like the one you added for `LocalGitStorageDir`?

lfs/util.go
```diff
@@ -275,3 +275,36 @@ func ConvertCwdFilesRelativeToRepo(cwdchan <-chan string) (<-chan string, error)
 func IsWindows() bool {
 	return GetPlatform() == PlatformWindows
 }
+
+// Determine if a file/dir exists, returns IsDir() results too
+func FileOrDirExists(path string) (exists bool, isDir bool) {
```
technoweenie: I'd like to see comments written to [Go convention](http://blog.golang.org/godoc-documenting-go-code). Specifically, start the comment with the name of the function, and end it with a period.

I know we're not that strict about comments in this project. Also, there are a lot of comments that could be added or reworded besides these. But if we're going to add more, I'd like them to follow conventions.

test/test-worktree.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
+
+. "test/testlib.sh"
+
+# git 2.5 or higher only
+compare_version $(git version | cut -d" " -f3) "2.5.0"
+if [[ $? != $VERSION_LOWER ]]; then
+  begin_test "git worktree"
```
technoweenie: Any thoughts on just exiting early before `begin_test()` to remove the extra level of nesting? Maybe even spit out an error message like "skipped worktree test because git {your_version} < 2.5.0".

technoweenie:
:metal: This is awesome.

lfs/lfs.go
sinbad: I've moved all except the one for TempDir since that changes during the life of the command

test/test-worktree.sh
sinbad: Doh, I didn't do that because I thought 'I don't want to exit the whole test suite' but of course exit  just affects the current script.

PR #547, opened 2015-07-31
Delay progress output until the transfer starts

rubyist:
If progress is not delayed it can interfere with credential helpers asking for credentials to make the API calls.


technoweenie:
Seems legit :+1:

PR #551, opened 2015-07-31
Add progress meter to checkout

sinbad:
Only displayed when running a separate checkout command, not on pull (lets the download queue do the progress to avoid mixing the 2)

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -101,7 +101,7 @@ func checkoutWithIncludeExclude(include []string, exclude []string) {
 	var wait sync.WaitGroup
 	wait.Add(1)
 
-	c := make(chan *lfs.WrappedPointer)
+	c := make(chan *lfs.WrappedPointer, 1)
 
 	go func() {
 		checkoutWithChan(c)
@@ -114,6 +114,7 @@ func checkoutWithIncludeExclude(include []string, exclude []string) {
 		totalBytes += pointer.Size
 	}
 	progress := lfs.NewProgressMeter(len(pointers), totalBytes, false)
+	progress.Start()
 	totalBytes = 0
 	for _, pointer := range pointers {
 		totalBytes += pointer.Size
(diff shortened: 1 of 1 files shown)


technoweenie:
It doesn't seem to output any progress until it's done:

```
$ GIT_TRACE=1 git lfs checkout
10:42:23.755943 git.c:555               trace: exec: 'git-lfs' 'checkout'
10:42:23.756430 run-command.c:351       trace: run_command: 'git-lfs' 'checkout'
trace git-lfs: run_command: 'git' rev-parse HEAD
trace git-lfs: run_command: git ls-tree -r -l --full-tree 089268aa349414ff230f5f3d72774477cb0844cb
(40 of 40 files) 483.14 MB / 483.14 MB
```

commands/command_checkout.go
```diff
@@ -107,14 +107,30 @@ func checkoutWithIncludeExclude(include []string, exclude []string) {
 		checkoutWithChan(c)
 		wait.Done()
 	}()
+
+	// Count bytes for progress
+	var totalBytes int64
+	for _, pointer := range pointers {
+		totalBytes += pointer.Size
+	}
+	progress := lfs.NewProgressMeter(len(pointers), totalBytes, false)
```
technoweenie: Ah, you need to call `Start()` on `progress`.  See #547

sinbad: Ah OK, Start() wasn't there until I rebased, done now. You have to use quite big files to see the progress otherwise most of the time (1-2s) is spent at the end waiting for the update-index process to stop.

rubyist:
This looks good, I added a note in #558 that mentions how we could use verbose output from `git update-index` to provide some progress around that extra end time, if needed. I don't think that needs to happen in this PR, necessarily.

technoweenie:
Merging this for now. Some `update-index` progress will be nice though.

sschuberth:
Is this supposed to display progress if a `checkout` actually needs to download a (big) file that's not in the local cache yet? Or is it only supposed to show progress of the copy from local cache to working tree? I'm asking because I don't see any progress in the first case with git-lfs 1.1.0 on Windows.

PR #552, opened 2015-07-31
build a helper for confirming the git version in tests

technoweenie:
I didn't notice until too late that the skip message from #546 never showed up. The problem is that `comparison_to_operator()` causes the script to halt if it returns something other than `$VERSION_EQUAL` (0) because of `set -e`.

/cc @sinbad


sinbad:
:+1:

PR #554, opened 2015-07-31
Support using credentials from multiple accounts against the LFS endpoint

clareliguori:
Support credentials from multiple accounts when useHttpPath=true is set in the git config

Provide the URL path to 'git credential fill' for credential helpers that need the path to return the right credentials
Use the objects API URL for credentials so there is a stable path passed to the credential helper for caching

### DIFF
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -653,7 +653,7 @@ func getCreds(req *http.Request) (Creds, error) {
 		}
 	}
 
-	creds, err := credentials(req.URL)
+	creds, err := credentials(apiUrl)
 	if err != nil {
 		return nil, err
 	}
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -653,7 +653,29 @@ func getCreds(req *http.Request) (Creds, error) {
 		}
 	}
 
-	creds, err := credentials(apiUrl)
+	credsUrl := apiUrl
+	if len(Config.CurrentRemote) > 0 {
+		if u, ok := Config.GitConfig("remote." + Config.CurrentRemote + ".url"); ok {
+			gitRemoteUrl, err := url.Parse(u)
+
+			if err == nil &&
+				gitRemoteUrl.Scheme == apiUrl.Scheme &&
+				gitRemoteUrl.Host == apiUrl.Host {
+
+				if gitRemoteUrl.User != nil {
+					if pass, ok := gitRemoteUrl.User.Password(); ok {
+						fmt.Fprintln(os.Stderr, "warning: current Git remote contains credentials")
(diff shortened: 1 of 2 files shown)


technoweenie:
Ah, good call. I have a few questions though:

Will passing the path to the credential `fill` command (https://github.com/github/git-lfs/pull/554/files#diff-1eed123d34cdc2b60cc1a63521a9ca88R20) cause any problems for those of us without `useHttpPath` enabled? I don't _think_ so, but I'll setup my laptop to use the osxkeychain helper instead of boxen's to see firsthand.

Do you by chance know what path git uses for pushes and pulls? We don't want end users to have to enter their password again for `/{user}/{repo}.git/info/lfs/objects` when they've already entered it for `/{user}/{repo}.git`.

technoweenie:
I think this needs to strip the leading `/` from the path:

```
Downloading droidtocat.gif (24.65 KB)
trace git-lfs: run_command: 'git' config -l -f .git/config
Username for 'https://github.com//technoweenie/glowing-octo-lana.git/info/lfs/objects':
```

The Username prompt looks correct here:

```
echo "protocol=https\nhost=github.com\npath=technoweenie/glowing-octo-lana.git/info/lfs/objects" | git credential fill 
Username for 'https://github.com/technoweenie/glowing-octo-lana.git/info/lfs/objects':
```

clareliguori:
Added trimming the leading slash, so you get this now:

```
$ git clone https://github.com/technoweenie/glowing-octo-lana
Downloading droidtocat.gif (24.65 KB)
Username for 'https://github.com/technoweenie/glowing-octo-lana.git/info/lfs/objects'
```

Good idea on re-using the base remote URL, I'll work on that

clareliguori:
Patch will now use the git remote's credentials if possible.

```
$ git clone https://github.com/technoweenie/glowing-octo-lana
Downloading droidtocat.gif (24.65 KB)
Password for 'https://clareliguori@github.com/technoweenie/glowing-octo-lana':
```

With this change, it seems to have started successfully picking up my credentials config for github.com from my gitconfig, which is great:
```
[credential "https://github.com"]
	username = clareliguori
```

technoweenie:
This looks great. There are a couple more things:

1. The code to set the auth from the git remote is directly copied from earlier in the function. It could be extracted into a small, private helper function. I'll probably just make a small refactoring PR after merging this.
2. I'd really like new tests for this. Unfortunately, we don't have adequate credential tests at ALL. So I'll probably tackle this on my own.

This will likely make it into Git LFS v0.6.0, and a v0.5.5 if another bug fix release is necessary. Thanks for sending this patch!

technoweenie:
Thanks again for the patch. I made a private helper and added some shell tests in #561.

PR #559, opened 2015-08-04
Adding parsing methods to extract LFS references from arbitrary history

sinbad:
This is part of my preparatory work to support the `prune` command; I thought I would submit this part early both to establish its use while I continue working on the specific features, and in case it is useful elsewhere. 

The focus of this PR is the `parseLogOutputToPointers` function which can scan a properly formatted `git log` output & extract relevant changes to LFS references. I've included an example of its use, ScanUnpushed() which shows what oids haven't been pushed anywhere & therefore only exist locally. However it will also be useful for other tasks (e.g. `fetch --recent`). The reason for it being able to scan for additions or removals is that depending on which direction in history you're scanning you might want either one (e.g. scanning backwards from a snapshot date between commits you want the '-' diffs to know what was being used beforehand).

The main feature of this method compared to alternatives is that it can extract WrappedPointers from arbitrary history with just a single `git log` command (and can be run in parallel with other tasks if needed)

### DIFF
--- a/lfs/scanner.go
+++ b/lfs/scanner.go
@@ -623,6 +623,8 @@ func parseLogOutputToPointers(log io.Reader, dir LogDiffDirection,
 				p, err := DecodePointer(&pointerData)
 				if err == nil {
 					results <- &WrappedPointer{Name: currentFilename, Size: p.Size, Pointer: p}
+				} else {
+					tracerx.Printf("Unable to parse pointer from log: %v", err)
 				}
 			}
 			pointerData.Reset()
--- a/lfs/scanner.go
+++ b/lfs/scanner.go
@@ -32,7 +32,7 @@ var (
 	logLfsSearchArgs = []string{
 		"-G", "oid sha256:", // only diffs which include an lfs file SHA change
 		"-p",   // include diff so we can read the SHA
-		"-U10", // Make sure diff context is always big enough to support extension lines to get whole pointer
+		"-U12", // Make sure diff context is always big enough to support 10 extension lines to get whole pointer
 		`--format=lfs-commit-sha: %H %P`, // just a predictable commit header we can detect
 	}
 )
--- a/lfs/scanner.go
+++ b/lfs/scanner.go
@@ -31,8 +31,8 @@ var (
 	// lfs changes and format the output suitable for parseLogOutput.. method(s)
 	logLfsSearchArgs = []string{
 		"-G", "oid sha256:", // only diffs which include an lfs file SHA change
-		"-p",  // include diff so we can read the SHA
-		"-U3", // Make sure diff context is always the same (used to ensure we always see all of pointer data)
+		"-p",   // include diff so we can read the SHA
+		"-U10", // Make sure diff context is always big enough to support extension lines to get whole pointer
 		`--format=lfs-commit-sha: %H %P`, // just a predictable commit header we can detect
 	}
 )
--- a/lfs/scanner.go
+++ b/lfs/scanner.go
@@ -611,7 +611,7 @@ func parseLogOutputToPointers(log io.Reader, dir LogDiffDirection,
 	commitHeaderRegex := regexp.MustCompile(`^lfs-commit-sha: ([A-Fa-f0-9]{40})(?: ([A-Fa-f0-9]{40}))*`)
 	fileHeaderRegex := regexp.MustCompile(`diff --git a\/(.+?)\s+b\/(.+)`)
 	fileMergeHeaderRegex := regexp.MustCompile(`diff --cc (.+)`)
-	pointerDataRegex := regexp.MustCompile(`^([\+\- ])(version https://git-lfs|oid sha256|size).*$`)
+	pointerDataRegex := regexp.MustCompile(`^([\+\- ])(version https://git-lfs|oid sha256|size|ext-).*$`)
 	var pointerData bytes.Buffer
 	var currentFilename string
 	currentFileIncluded := true
--- a/lfs/scanner.go
+++ b/lfs/scanner.go
@@ -26,6 +26,17 @@ const (
 	chanBufSize = 100
 )
 
+var (
+	// Arguments to append to a git log call which will limit the output to
+	// lfs changes and format the output suitable for parseLogOutput.. method(s)
+	logLfsSearchArgs = []string{
+		"-G", "oid sha256:", // only diffs which include an lfs file SHA change
+		"-p",  // include diff so we can read the SHA
+		"-U3", // Make sure diff context is always the same (used to ensure we always see all of pointer data)
+		`--format=lfs-commit-sha: %H %P`, // just a predictable commit header we can detect
+	}
+)
+
 // WrappedPointer wraps a pointer.Pointer and provides the git sha1
 // and the file name associated with the object, taken from the
 // rev-list output.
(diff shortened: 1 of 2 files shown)


lfs/scanner.go
```diff
@@ -517,3 +528,149 @@ func lsTreeBlobs(ref string) (chan TreeBlob, error) {
...
+
+// logUnpushedSHAs scans history for all LFS pointers which have been added but not pushed to any remote,
+// return progressively in a channel
+func logUnpushedSHAs() (chan *WrappedPointer, error) {
+	logArgs := []string{"log",
+		"--branches", "--tags", // include all locally referenced commits
+		"--not", "--remotes", // but exclude everything reachable from any remote
+	}
+	// Add standard search args to find lfs references
+	logArgs = append(logArgs, logLfsSearchArgs...)
```
technoweenie: Not a big deal, but why is this being joined with a global slice when none of the flags change? Couldn't this just use the global slice?

technoweenie:
Looks good. Will this work with pointers that use extensions?

```
ext-0-foo sha256:{original hash}
ext-1-bar sha256:{hash after foo}
oid sha256:{hash after bar}
size 123
```

The oid will definitely change as the extensions change, so I think it will.

lfs/scanner.go
rubyist: Does it need the global `logLfsSearchArgs`? I don't see it referenced anywhere else.

rubyist:
This looks cool :+1:

lfs/scanner.go
sinbad: In both cases the answer to this is it's there for my future use :) The idea is that the local `logArgs` represents the history filters for this specific function (unpushed), while `logLfsSearchArgs` are the common arguments that will be needed for *all* history scans. I'm going to need the distinction and the shared global args for the other work I'm doing on fetch and they're probably useful elsewhere too. So in other functions I'll use a different set of log filtering args but still the standard `logLfsSearchArgs` to get the format & oid change filter.

sinbad:
> Looks good. Will this work with pointers that use extensions? 
> The oid will definitely change as the extensions change, so I think it will.

Yes, so long as the oid changes the diff will come out in the log search. However if there's a chance that more than 2 extension lines will occur before the oid line then the `-U3` context won't be enough to pick up the whole pointer for decoding. Do we have an idea of the upper bar of extensions? I think I'll just use `-U10` to be sure since it won't hurt to be over-cautious.

lfs/scanner.go
sinbad: Also note how `parseLogOutputToPointers` requires that at least `logLfsSearchArgs` are specified in the `git log` command that produces the output it consumes, but other params are up to the caller.

technoweenie: Seems legit :+1:

technoweenie:
Hmm, I don't recall if we set an upper limit on the number of extensions. I can see setting it to something really low, like 3 or 5. 10 is fine for now until we can settle on a real limit.

sinbad:
Yeah `-U10` allows for 9 `ext-` lines before & after the oid line (the version & size lines use up the last context, assuming neither change), so this should be plenty; it doesn't make any difference for any pointers without extensions anyway. I'm not currently parsing out `ext-` lines into the pointer data because the current decode method can't handle them anyway. Even after it does I think they're probably not useful in this context (extensions are just for clean/smudge?).

sinbad:
Oh, I see extensions have been merged in now. I've updated this PR so `ext-` data is passed from the diff too.

technoweenie:
If you want to get the "actual" file from an lfs pointer with extensions, you'll have to parse all that out. This info is _only_ stored in these pointer files. A plain object in `.git/lfs/objects` does not have the necessary info to smudge itself.

Keep in mind: extensions have not shipped. Now's the time to make changes :)

sinbad:
Hrm, that's going to make pruning (& maybe other functions) a lot more complicated if you can no longer rely on the oid == the storage location. 

Doesn't this also mean that if an extension changed it could generate a different storage location for the same oid? So changing your gitconfig can completely change the storage location of all assets even if they had the same original oid?

It would be a lot simpler for many purposes if there was a 'storage oid' field in the pointer which easily identified where the data is stored, rather than having to locate the highest `ext-` entry. You could use it in all cases to identify what storage location this represents, even without any exts.  I think the extensions PR has already broken the 'fetch unique' logic for example:
```
		// Only add to download queue if local file is not the right size already
		// This avoids previous case of over-reporting a requirement for files we already have
		// which would only be skipped by PointerSmudgeObject later
		if !lfs.ObjectExistsOfSize(p.Oid, p.Size) {
			q.Add(lfs.NewDownloadable(p))
```
Also 'Size' is presumably not reliable any more because it represents the original size of the object, not what it is after the extension chain was applied?

sinbad:
And eek, even a small screw up in your extension code can utterly and completely screw up your storage integrity, and every developer has to get it right. I'd hate to support this ;)

sinbad:
Above comments might be wrong, extensions proposal said:
> The oid and size keys are calculated from the final bytes written into the LFS storage

If it's the case that `p.Oid` and `p.Size` are post-extensions then it's less of a problem (integrity of extension setup across all devs still worries me a bit). I read your comment to mean the "actual" storage location would have to be parsed out of the last ext entry. Code is a little confusing so not sure which it is.

technoweenie:
The OID is still the storage location, and the size matches the byte size of the object in local storage (`.git/lfs/objects`) or even at the server. Since it's the output of the extension chain, it could be a compressed, encrypted thing.

Now, if you want to turn the blob into the actual file during checkout, you'll either need all the pointer fields from the log, or you'll need to read the pointer from the git blob.

> Doesn't this also mean that if an extension changed it could generate a different storage location for the same oid? So changing your gitconfig can completely change the storage location of all assets even if they had the same original oid?

The Git config only covers the extensions for new files. Existing files will be smudged according to the info in the pointer.

> And eek, even a small screw up in your extension code can utterly and completely screw up your storage integrity, and every developer has to get it right. I'd hate to support this ;)

Yeaaaah. At least we only have to worry about the extension system working. But I think that we should advertise extensions as an "experimental" feature. I'm afraid people will rush to build compression filters based on a non deterministic algorithm.

sinbad:
> The OID is still the storage location, and the size matches the byte size of the object in local storage 

Phew. Thanks for the confirmation.

> Now, if you want to turn the blob into the actual file during checkout, you'll either need all the pointer fields from the log, or you'll need to read the pointer from the git blob.

OK, my latest commit on this PR includes all the `ext-` lines so that will all still work from these results too.

> Yeaaaah. At least we only have to worry about the extension system working. But I think that we should advertise extensions as an "experimental" feature. I'm afraid people will rush to build compression filters based on a non deterministic algorithm.

Agreed, all it takes is for the extension to change over time & 2 developers have different versions, or for people to choose a generic name like 'compress' and for that to have been used by different extensions :o Wild thought: maybe the extensions themselves should have an identifying SHA based on how they translate a known input string? Then if they're incompatible you know before you start trampling over real data.

ryansimmen:
Git LFS Extensions support 10 extensions numbered 0-9.

ryansimmen:
> Wild thought: maybe the extensions themselves should have an identifying SHA based on how they translate a known input string? Then if they're incompatible you know before you start trampling over real data. :+1:

sinbad:
> Git LFS Extensions support 10 extensions numbered 0-9.

OK `-U12` it is then to be absolutely sure. There's a crazy edge case with `-U10` where if 10 extensions were installed, there same content was re-added but the last extension output a different result, 10 lines of context wouldn't quite be enough (there could be 11 unchanged lines including version before the oid). It'll never happen except maybe it might ;)

technoweenie:
This looks good, but I'd like to see some more tests. For instance, none of the tests have extensions in the diffs.

One other thing: the `gitattributes` file sets these files up with a custom "lfs" diff handler:

    *.gif filter=lfs diff=lfs merge=lfs -crlf

It doesn't do anything yet, but I wonder how it'll affect this function if we add a custom "lfs" diff handler.

sinbad:
Good call on the diff handler, that should be explicit. I'll deal with that and add a few cases with extension lines.

sinbad:
Tests added for extension cases. I've tested the use of a custom diff tool in .gitattributes (basically just defining one that called `echo`) and actually it doesn't affect the output of `git log`, (or even `git show`), only that of `git diff`. So there is no impact on this code.

technoweenie:
Great, thanks for checking into that.

PR #560, opened 2015-08-04
refactor lfs.SortExtensions

technoweenie:
This was pointed out by @WillHipschman in https://github.com/github/git-lfs/pull/486#discussion_r36108775.

* fewer slice iterations
* slices are initialized with the correct capacity

/cc @ryansimmen since this came from your pr.


rubyist:
:+1:

PR #565, opened 2015-08-06
Fetch tests should test local storage

sinbad:
Spotted while working on other things that assert_pointer wasn't a good test for fetch, since it only validates that the git commit objects have the right pointer data, not that the content exists locally. 

Added assert_local_object and refute_local_object instead & used in fetch tests.


technoweenie:
Ah yeah, good call :+1:

PR #568, opened 2015-08-06
Adding Help Text To Commands

WillHipschman:
Currently using git lfs help checkout will not display anything other than the base help text of the Cobra command.

This change parses the .ronn man files in /docs/ and writes them to go variables as part of script/bootstrap using go generators. When making a man change (adding a new command/man file or modifying a current one) the help text can be regenerated by running 'go generate' from the git-lfs root or by running the bootstrap script.


commands/help-text_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	"io/ioutil"
+	"strings"
+	"testing"
+)
+
+//New commands must be added here
+var helpText = map[string]string {
+	"git-lfs-checkout.1.ronn":git_lfs_checkout_HelpText,
```
WillHipschman: Go doesn't really have a way to reflect over variables on a class, so I settled for a static map. If anyone has better ideas on how to map the man file names to variable names I'm all ears.

commands/help-text_test.go
```diff
@@ -0,0 +1,80 @@
...
+    "git-lfs-clean.1.ronn":git_lfs_clean_HelpText,
```
WillHipschman: tabbing

commands/help-text_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"git-lfs-status.1.ronn":git_lfs_status_HelpText,
+	"git-lfs-track.1.ronn":git_lfs_track_HelpText,
+	"git-lfs-uninit.1.ronn":git_lfs_uninit_HelpText,
+	"git-lfs-untrack.1.ronn":git_lfs_untrack_HelpText,
+	"git-lfs-update.1.ronn":git_lfs_update_HelpText,
+	"git-lfs.1.ronn":git_lfs_HelpText,
+}
+
+var(
+	suffix 		string 	= ".1.ronn"
```
WillHipschman: This file has some weird tabbing. I blame windows.

docs/man/git-lfs-clean.1.ronn
```diff
@@ -18,3 +18,4 @@ Git attributes.
 git-lfs-init(1), git-lfs-push(1), gitattributes(5).
 
 Part of the git-lfs(1) suite.
+
```
WillHipschman: I added some white space to the end of some man files to make them display better in the terminal

rubyist:
It looks like a couple `.exe` files were added that probably shouldn't have been.

docs/include-help-text.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+
+func convertManFileToString(fileName, root string) []byte {
+    helpText, _ := ioutil.ReadFile(root + "/" + fileName)
+	return []byte(strings.Replace(string(helpText), "`", "'", -1))
+
+}
+
+
```
rubyist: I think this should go along with the other stuff under `script/`. That's our general convention so far, and things under `docs/` do not get run through the formatter, which can lead to formatting and tab issues on different machines. There are several lines that don't have correct tabbing and there should be a space after `import`.

commands/help-text_test.go
rubyist: This could just be generated along with the rest of the data. It wouldn't really bother me if it lived at the bottom of `help-text.go`. I think it's too easy to miss adding something to this list. What it's really testing is that the generator has run before building, and if something is missed then the test doesn't have much value.

rubyist:
The `help-text_test.go` file has several formatting issues, can you run that through `goimports` (`script/fmt` should be doing this) and clean that up?

commands/command_checkout.go
```diff
@@ -15,6 +15,7 @@ var (
 	checkoutCmd = &cobra.Command{
 		Use:   "checkout",
 		Short: "Checks out LFS files into the working copy",
+		Long:  git_lfs_checkout_HelpText,
```
technoweenie: Go keywords tend to be in camelCase. Also, I think the `git_lfs` prefix here is redundant. I think you could name these after the subcommand: `checkoutHelp`, `cleanHelp`, etc.

docs/man/git-lfs-clean.1.ronn
technoweenie: Trailing linebreaks are always welcome :+1:

rubyist:
Higher level feedback:

How does everyone feel about it displaying raw markdown to the terminal? I don't :heart: it, but if it doesn't bother anyone else then I won't oppose it. I wouldn't want to pull in a markdown parser and display it formatted or anything, just maybe strip the markdown out.

I'm also thinking that this approach differs a little from `git`. I'm just spitballing here, but I'll throw this out for consideration/discussion.

When you do `git help checkout`, for example, it displays the formatted man page in your configured pager/browser/what have you. While `git help lfs` does work, it's unfortunate that `git` won't dig into sub commands and give us `git help lfs checkout`. However, it does work to do `git help git-lfs-checkout`. The `cobra.Command` setup allows us to provide our own `HelpFunc`, so it's possible that `git lfs help checkout` can just shell out to `git help git-lfs-checkout` rendering the way `git` typically would.

The caveat to that is it only works if the Git LFS man pages have been installed, but `git` does give a correct exit code if it's not found and we could fallback to this generated output, so we'd probably need to keep both (though ideally, man pages are installed).

commands/command_checkout.go
rubyist: :+1:, and along with this, I think i'd rather see `_` in the file names than `-`.

technoweenie:
> How does everyone feel about it displaying raw markdown to the terminal?

Markdown is designed to look good in plain text, unless you're doing some weird stuff like extra html tags or tables. So I'm ok with it. But there are some tags, like `<br>`, that need to be stripped out.

I think I'd prefer git ignoring `commands/help-text.go`, and relying on `script/bootstrap` to generate that each time. 

> The cobra.Command setup allows us to provide our own HelpFunc, so it's possible that git lfs help checkout can just shell out to git help git-lfs-checkout rendering the way git typically would.

I like this idea, actually. The markdown ronn files are converted to proper man pages, which git already knows how to display. It'd also keep us from storing the help text twice (in ronn files, and again in the go code).

Now, since we don't have a reliable way to install man pages cross platform yet, I think this PR is a good workaround. I think the `commands/help-text.go` file should be git ignored. The various scripts should all bootstrap before building/testing/running the code, so this should be fine.

Also, do we even need these `git_lfs_checkout_HelpText` variables? It could be in a big map that gets generated:

```go
var help = map[string]string {
  "checkout": `git-lfs-checkout(1) -- Update working copy with file content if available............`,
  "clean": "...........",
```

It's still very easy to set this to the `Long` property on `*cobra.Command` objects. It's also easy to see what commands are documented by checking or iterating through the map.

sinbad:
First: I love that this is on the radar to fix. :+1: 

> The caveat to that is it only works if the Git LFS man pages have been installed 

Just incase it hasn't come up yet, were you aware that the default behaviour in Windows for something like `git checkout --help` is that it launches the browser & displays the HTML version of the man page? The msys git install doesn't seem to include man so this is the alternative, I'm not sure what triggers it all & whether that will work for us too. Just a thought since we're generating HTML from ronn too.

technoweenie:
Closing this. This was a great idea, but @sinbad's approach with `go generate` in #665 keeps us from maintaining the help text in multiple places. Those ronn files are now a single source of truth :metal:

Issue #558, opened 2015-08-03
Failures during a checkout can leave the git index locked


rubyist:
I encountered a failure during an `lfs checkout` that left the git index locked after it bailed out. I'm not yet able to reproduce it reliably. I can manually run a `git update-index` and kill the process while it's working on a larger file resulting in the index being left locked, so I'm pretty sure that's what's happening. I'm not quite sure which part of the checkout process is exiting early.

The `update-index` command can provide output which we can watch to make sure we're not shutting down in the middle of processing. e.g.:

```
$ git update-index -q --refresh --verbose --stdin
1.dat
add '1.dat'
2.dat
3.dat
add '2.dat'
add '3.dat'
```

This could also be used to add some extra progress info along the lines of #551.

PR #570, opened 2015-08-07
Avoid panic during update index

rubyist:
This PR will avoid any use of `Panic()` while the `git update-index` loop is running. If the process exits while `update-index` is in the middle of processing a file, the index can be left in an inconsistent, locked state.

/cc #558

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -20,10 +20,10 @@ var (
 )
 
 func checkoutCommand(cmd *cobra.Command, args []string) {
-
 	// Parameters are filters
 	// firstly convert any pathspecs to the root of the repo, in case this is being executed in a sub-folder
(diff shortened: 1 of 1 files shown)


rubyist:
Marking this for review as is, it is only fixing the index lock problem. Using the update-index output to drive progress ended up being messier than I'd like so I'm going to punt on that for now. There are a couple of optimizations for speeding up the checkout, I'll put those in a new PR so this can ship.

sinbad:
:+1: Although this will conflict with https://github.com/github/git-lfs/pull/573 :sob:

sinbad:
I've pre-emptively merged this into #573 to head off any merge conflicts.

rubyist:
Thanks for the review, one of these times I'll submit a PR that doesn't conflict with yours!

PR #572, opened 2015-08-10
fix non-batch transfer

billygor:
non-batch transfer was stuck when objects already exist on lfs-server


rubyist:
Good catch, thanks!

PR #573, opened 2015-08-10
Add include/exclude config options & params to fetch

sinbad:
Note: don't merge until we've resolved https://github.com/github/git-lfs/issues/564.

Added `lfs.fetchinclude` and `lfs.fetchexclude` config options and `--include` and `--exclude` command line args to fetch. These cause certain paths to either be included or excluded when fetching binaries so that people who work in specific subfolders and don't want to waste time or disk space fetching irrelevant things can do so. This is most common in teams using a single shared repo for all assets but where not everyone needs all of them, e.g. only texture artists might need the original .psd files vs exported runtime equivalents, only modellers/animators might need the original .max files etc. Some teams have a more complex pipeline which avoids this but for those that don't wanting to download by subfolder is quite common. I expect the config option variants to be most used, `--include`/`--exclude` args to fetch are there for special occasions.

This affects the behaviour of `fetch`, `pull` and `smudge` since in all cases the `download` argument to PointerSmudge is set based on whether it matches include/exclude filters.

There is a later knock-on effect on `push` though because it means a user won't have all oids for a given commit. Right now that will cause `push` to fail later if the commits in question get included in the scan for pushing a new branch. This is another case of https://github.com/github/git-lfs/issues/564, which needs to be addressed before this can be merged.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -187,6 +187,30 @@ type ErrorWithStack interface {
 	Stack() []byte
 }
 
+// determineIncludeExcludePaths is a common function to take the string arguments
+// for include/exclude and derive slices either from these options or from the
+// common global config
+func determineIncludeExcludePaths(includeArg, excludeArg string) (include, exclude []string) {
+	var includePaths, excludePaths []string
+	if len(includeArg) > 0 {
+		for _, inc := range strings.Split(includeArg, ",") {
+			inc = strings.TrimSpace(inc)
+			includePaths = append(includePaths, inc)
+		}
+	} else {
+		includePaths = lfs.Config.FetchIncludePaths()
+	}
+	if len(excludeArg) > 0 {
+		for _, ex := range strings.Split(excludeArg, ",") {
+			ex = strings.TrimSpace(ex)
+			excludePaths = append(excludePaths, ex)
+		}
+	} else {
+		excludePaths = lfs.Config.FetchExcludePaths()
+	}
+	return includePaths, excludePaths
+}
+
 func init() {
 	log.SetOutput(ErrorWriter)
 }
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"io"
 	"os"
 	"os/exec"
 	"sync"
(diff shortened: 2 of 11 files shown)


sinbad:
This PR is safe to merge after https://github.com/github/git-lfs/pull/581

commands/commands.go
```diff
@@ -187,6 +187,30 @@ type ErrorWithStack interface {
 	Stack() []byte
 }
 
+// determineIncludeExcludePaths is a common function to take the string arguments
+// for include/exclude and derive slices either from these options or from the
+// common global config
+func determineIncludeExcludePaths(includeArg, excludeArg string) (include, exclude []string) {
+	var includePaths, excludePaths []string
```
technoweenie: Does it make sense to initialize the array with a capacity based on the number of commas in the strings? Not sure this matters much though. Probably best to keep the code simpler :)

technoweenie:
I think this PR looks good :+1: A couple things come to mind that won't necessarily block this PR:

As a user, I don't think this documentation is sufficient to show me how this works. What kinds of file paths can I use? Well, _I know_ based on [these tests](https://github.com/github/git-lfs/blob/57eec4f7d80e96cceaa0360959406684350fc4ce/lfs/util_test.go#L71-L103). How should the docs describe this? A man page? Or a link to something in ./docs from the man pages?

Related to that, is there a way to get some visibility into what this is doing? If I have some include/exclude properties set without knowing it, the fetch/pull commands may behave unexpectedly. Any thoughts on how we could show what files, if any, are skipped to lfs users?

rubyist:
:+1:, agree on some more documentation

commands/commands.go
sinbad: Yeah I figured the likelihood of having enough separate entries to warrant this was quite small.

sinbad:
The matching/wildcard rules are the same as gitignore (deliberately so); I'm updating the docs for my next PR anyway so I've included some explicit examples: https://github.com/sinbad/git-lfs/commit/6d35d8303b7346f5db7e35d14a96067bad917f95

technoweenie:
@sinbad: That's perfect :+1:

PR #576, opened 2015-08-11
major pre-push optimization

billygor:
this change improves drastically pre-push behaviour, by not sending
lfs objects which are already on a remote. Works perfectly with
pushing new branches and tags.

currently pre-push command analyse "local sha1" vs "remote sha1" of the
ref being pushed and if "remote sha1" is available locally tries to send
only lfs objects introduced with new commits.

why this is broken:
- remote branch might have moved forward (local repo is not up to date).
  In this case you have no chance to isolate new lfs objects ("remote sha1"
  does not exist locally) and git-lfs sends everything from the local
  branch history.
- remote branch does not exist (or new tag is pushed). Same consequences.

But what is important - local repository always have remote references,
from which user created his local branch and started making some local
changes. So, all we have to do is to identify new lfs objects which do
not exist on remote references. And all this can be easily achieved with
the same all mighty git rev-list command.

This change makes git-lfs usable with gerrit, where changes are uploaded
by using magic gerrit branches which does not really exist. i.e.
git push origin master:refs/for/master

in this case "refs/for/master" does not exist and git feeds all 0-s as
"remote sha1".

### DIFF
--- a/lfs/scanner.go
+++ b/lfs/scanner.go
@@ -3,6 +3,7 @@ package lfs
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"io"
 	"os/exec"
 	"regexp"
@@ -60,9 +61,15 @@ type indexFile struct {
 
 var z40 = regexp.MustCompile(`\^?0{40}`)
 
+const (
+	ScanTwoRefsDelta   = iota // 0 - or default scan mode
+	ScanAll            = iota
+	ScanDeltaToRemotes = iota
+)
+
 type ScanRefsOptions struct {
+	ScanMode         int
 	SkipDeletedBlobs bool
-	scanAll          bool
 	nameMap          map[string]string
 }
 
@@ -73,7 +80,9 @@ func ScanRefs(refLeft, refRight string, opt *ScanRefsOptions) ([]*WrappedPointer
 	if opt == nil {
 		opt = &ScanRefsOptions{}
 	}
-	opt.scanAll = refLeft == ""
+	if refLeft == "" {
+		opt.ScanMode = ScanAll
+	}
 	opt.nameMap = make(map[string]string, 0)
 
 	start := time.Now()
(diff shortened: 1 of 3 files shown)


lfs/scanner.go
```diff
@@ -187,6 +195,12 @@ func revListShas(refLeft, refRight string, opt ScanRefsOptions) (chan string, er
 		if refRight != "" && !z40.MatchString(refRight) {
 			refArgs = append(refArgs, refRight)
 		}
+	case ScanAll:
+		refArgs = append(refArgs, "--all")
+	case ScanDeltaToRemotes:
+		refArgs = append(refArgs, refLeft, "--not", "--remotes")
```
billygor: I'm using just "--remotes" here. Another option would be to use --remotes=<remote_name_from_pre-push_arg_1>. But since we only have one git-lfs server and single lfs.url configuration option, it is safe to assume that the same git lfs server is used for all remotes. So, we can do a better jobs isolating only lfs objects introduced locally.

technoweenie: > But since we only have one git-lfs server and single lfs.url configuration option, it is safe to assume that the same git lfs server is used for all remotes. So, we can do a better jobs isolating only lfs objects introduced locally.

Not necessarily. You can also configure an lfs server at `remote.{name}.lfsurl`. Ideally, each git host has its own lfs host too that's automatically configured. It should be easy to push and pull between different git hosts.

There is one big issue with this right now. Git doesn't tell us what remote we should be using in the smudge or post-checkout hooks. So pulling from a non-origin remote doesn't quite work.

billygor: ok, here is the change which makes use of current remote only: https://github.com/github/git-lfs/pull/578
I'll be happy with either of those 2.

sinbad:
> There is one big issue with this right now. Git doesn't tell us what remote we should be using in the smudge or post-checkout hooks. So pulling from a non-origin remote doesn't quite work.

I was thinking that perhaps the `fetch` & `pull` lfs commands should have a remote as their first arguments, but didn't add them yet. For smudge / post-checkout I suggest we use the default remote as chosen by `git pull`, which is:

1. Check if the current branch is tracking a remote branch (`git branch -vv`) and if so use the remote it's tracking (this applies to pull, not push so for example in forks you'd usually track the read-only upstream on a secondary remote to pull from, and push to `origin` as your fork)
2. Use `origin` if it exists
3. Fail

I agree the pre-push should be specific about what remote it's comparing with.

technoweenie:
@sinbad: I like your suggestion.

I also prefer #578 to this PR.

sinbad:
I've submitted PR https://github.com/github/git-lfs/pull/583 to implement the remote arguments & refinement to the default remote.

billygor:
superseded by https://github.com/github/git-lfs/pull/578

PR #577, opened 2015-08-11
Fix race condition meaning TransferQueue might not report errors in time

sinbad:
Previously error went to chan after `q.wait.Done()` so if a caller was blocked on `q.Wait()` it could end up checking `q.Errors()` before the error had been sent to the channel / extracted by errorCollector. Error needs to be sent before `q.wait.Done()`.


rubyist:
Doh! Good catch :+1:

PR #578, opened 2015-08-11
major pre-push optimization (alternative: look for deltas against current remote only)

billygor:
this change improves drastically pre-push behaviour, by not sending
lfs objects which are already on a remote. Works perfectly with
pushing new branches and tags.

currently pre-push command analyse "local sha1" vs "remote sha1" of the
ref being pushed and if "remote sha1" is available locally tries to send
only lfs objects introduced with new commits.

why this is broken:
- remote branch might have moved forward (local repo is not up to date).
  In this case you have no chance to isolate new lfs objects ("remote sha1"
  does not exist locally) and git-lfs sends everything from the local
  branch history.
- remote branch does not exist (or new tag is pushed). Same consequences.

But what is important - local repository always have remote references,
from which user created his local branch and started making some local
changes. So, all we have to do is to identify new lfs objects which do
not exist on remote references. And all this can be easily achieved with
the same all mighty git rev-list command.

This change makes git-lfs usable with gerrit, where changes are uploaded
by using magic gerrit branches which does not really exist. i.e.
git push origin master:refs/for/master

in this case "refs/for/master" does not exist and git feeds all 0-s as
"remote sha1".

### DIFF
--- a/lfs/scanner.go
+++ b/lfs/scanner.go
@@ -3,6 +3,7 @@ package lfs
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"io"
 	"os/exec"
 	"regexp"
@@ -60,9 +61,18 @@ type indexFile struct {
 
 var z40 = regexp.MustCompile(`\^?0{40}`)
 
+type ScanningMode int
+
+const (
+	ScanRefsMode         = ScanningMode(iota) // 0 - or default scan mode
+	ScanAllMode          = ScanningMode(iota)
+	ScanLeftToRemoteMode = ScanningMode(iota)
+)
+
 type ScanRefsOptions struct {
+	ScanMode         ScanningMode
+	RemoteName       string
 	SkipDeletedBlobs bool
-	scanAll          bool
 	nameMap          map[string]string
 }
 
@@ -73,7 +83,9 @@ func ScanRefs(refLeft, refRight string, opt *ScanRefsOptions) ([]*WrappedPointer
 	if opt == nil {
 		opt = &ScanRefsOptions{}
 	}
-	opt.scanAll = refLeft == ""
+	if refLeft == "" {
+		opt.ScanMode = ScanAllMode
+	}
 	opt.nameMap = make(map[string]string, 0)
 
 	start := time.Now()
@@ -174,9 +186,8 @@ func ScanIndex() ([]*WrappedPointer, error) {
 // channel from which sha1 strings can be read.
 func revListShas(refLeft, refRight string, opt ScanRefsOptions) (chan string, error) {
 	refArgs := []string{"rev-list", "--objects"}
-	if opt.scanAll {
-		refArgs = append(refArgs, "--all")
-	} else {
+	switch opt.ScanMode {
+	case ScanRefsMode:
 		if opt.SkipDeletedBlobs {
 			refArgs = append(refArgs, "--no-walk")
 		} else {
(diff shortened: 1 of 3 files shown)


technoweenie:
I like this :metal: Would like at least 1 more review though. Should be able to merge it early next week. Thanks!

lfs/scanner.go
```diff
@@ -174,9 +184,8 @@ func ScanIndex() ([]*WrappedPointer, error) {
 // channel from which sha1 strings can be read.
 func revListShas(refLeft, refRight string, opt ScanRefsOptions) (chan string, error) {
 	refArgs := []string{"rev-list", "--objects"}
-	if opt.scanAll {
-		refArgs = append(refArgs, "--all")
-	} else {
+	switch opt.ScanMode {
+	case ScanTwoRefsDelta:
```
sinbad: Nitpick but it's not always scanning two refs, it can scan one (refRight == "") and in fact that's the more common usage. I'd call it ScanRefs or something similar instead.

lfs/scanner.go
```diff
@@ -187,6 +196,12 @@ func revListShas(refLeft, refRight string, opt ScanRefsOptions) (chan string, er
 		if refRight != "" && !z40.MatchString(refRight) {
 			refArgs = append(refArgs, refRight)
 		}
+	case ScanAll:
+		refArgs = append(refArgs, "--all")
+	case ScanDeltaToRemote:
+		refArgs = append(refArgs, refLeft, "--not", "--remotes="+opt.RemoteName)
```
sinbad: How about making this support refRight != "" as well? It's a valid use case to want to scan an existing local range with both to & from AND exclude a remote as well, even if it's not used right now.

If you'd prefer not to add that I'd prefer if the enum indicated that refRight was ignored somehow.

lfs/scanner.go
```diff
@@ -60,9 +61,16 @@ type indexFile struct {
...
 var z40 = regexp.MustCompile(`\^?0{40}`)
 
+const (
+	ScanTwoRefsDelta  = iota // 0 - or default scan mode
+	ScanAll           = iota
+	ScanDeltaToRemote = iota
+)
+
 type ScanRefsOptions struct {
+	ScanMode         int
```
sinbad: You could strongly type ScanMode:
```
type ScanModeEnum int

sinbad:
:+1: except for my nitpick suggestions. Together with the recently merged https://github.com/github/git-lfs/pull/581 this should make pre-push much better.

technoweenie:
:+1: on the nitpicks. How about these names: `ScanRefs`, `ScanAll`, `ScanLeftToRemote`?

billygor:
rebased and uploaded new version.
@sinbad, thanks. I've strongly typed ScanMode.
As for adding "refRight" to "ScanDeltaToRemote" mode: In pre-push scenario I consider 3 options:
1. refRight is all zeros (means we are pushing new ref) - no need to include refRight.
2. local repo is up to date, which means we have refRight as one of remotes refs and there is no effect of adding refRight to `git rev-list --objects <left> --not --remotes="remote name"`
3. local repo is not up to date, ref on a remote is pointing to some commit we don't have yet, and so we cannot add refRight to `git rev-list`, it will return with failure.

May be I'm missing something here. If you'll explain me a scenario where this might be still needed, I can add a check that "refRight" exists locally (`git cat-file -e <sha1>` - will do), and add it to `git rev-list` if it does. And I would prefer to do it with a subsequent commit.

@technoweenie, I took your naming advice, but ScanRefs name is already used as the function name. Had to add "Mode" to ScanMode names.

sinbad:
You're right that in the current push use case you don't need to support left & right & remotes, but ScanRefs is a general utility function so is not limited to that use case in future. I can't say there will never be a use case for 'everything between 2 commits that also isn't on remote'. I don't mind that use case not being catered for but the signature should be clear that supplying the right ref will do nothing.

rubyist:
:+1: from me, looks rad

PR #579, opened 2015-08-11
don't use ldflags in credential tests

technoweenie:
gccgo silently ignores them, causing tests to fail.


technoweenie:
@billygor: Does this work for you?

billygor:
@technoweenie: Yes, script/integration tests works fine with gccgo now. Thanks!

PR #581, opened 2015-08-12
Check server for objects before failing push because local is missing

sinbad:
The local client doesn't have all the LFS objects so it's perfectly valid to push commits which reference objects that don't exist, so long as the server already has them. This could occur when deleting files and pushing new branches (https://github.com/github/git-lfs/issues/564) but is increasingly likely as fetch & prune functionality is introduced.

So now we perform a pre-check and any locally missing files are checked on the server. To gain the benefit of the existing parallelism and batching API this required adding a new type of transferable which only checks for download, and doesn't actually do it - instead the fact that the transfer *would* go ahead is used as an indicator that this oid can be skipped when it comes to push & therefore not cause the process to fail.

After merging this PR it's safe to merge https://github.com/github/git-lfs/pull/573 (which deliberately doesn't download all LFS objects).


rubyist:
Looks good to me :+1:

What do folks thing about embedding a `Downloadable` in `Checkable` and avoiding repeating all the methods? e.g.

```go
type Checkable struct {
	Pointer *WrappedPointer
	object  *objectResource
	*Downloadable
}

func NewCheckable(p *WrappedPointer) *Checkable {
	return &Checkable{Pointer: p, Downloadable: NewDownloadable(p)}
}

func (d *Checkable) Transfer(cb CopyCallback) *WrappedError {
	// just report completion of check but don't do anything
	cb(d.Size(), d.Size(), int(d.Size()))
	return nil
}
```

While we don't technically need `NewCheckQueue`, I think I'm OK with it because it helps make the intent clear.

sinbad:
> What do folks thing about embedding a Downloadable in Checkable and avoiding repeating all the methods?

That's cool but feels a bit funky having Checkable include a more specific type. I think it would make more sense conceptually to do the reverse and embed a Checkable in Downloadable?

technoweenie:
> I think it would make more sense conceptually to do the reverse and embed a Checkable in Downloadable?

It depends how you look at it. Is Downloadable embedding Checkable to add new functionality, or is Checkable embedding Downloadable to disable functionality?

I think it makes sense for both [Downloadable](https://github.com/github/git-lfs/blob/9a99504349a4640ba37dc9a1e0664fe87f2e3f86/lfs/download_queue.go#L24-L42) and [Uploadable](https://github.com/github/git-lfs/blob/9a99504349a4640ba37dc9a1e0664fe87f2e3f86/lfs/upload_queue.go#L53-L71) to embed a Checkable.

rubyist:
I think Checkable as the base makes the most sense. I'm also not opposed to keeping it as it is, just wondering aloud kind of thing.

test/cmd/lfstest-gitserver.go
```diff
@@ -228,19 +228,35 @@ func lfsBatchHandler(w http.ResponseWriter, r *http.Request, repo string) {
...
+				Actions: map[string]lfsLink{
+					"upload": lfsLink{
+						Href:   lfsUrl(repo, obj.Oid),
+						Header: map[string]string{},
+					},
+				},
+			}
+			if testingChunked {
+				o.Actions["upload"].Header["Transfer-Encoding"] = "chunked"
+			}
```
technoweenie: This duplication seems unnecessary. Can we get away with changing the action key from "upload" to "download" based on the bool return value from `largeObjects.Get()`?

sinbad:
> I think Checkable as the base makes the most sense. I'm also not opposed to keeping it as it is, just wondering aloud kind of thing.

Yeah this was a good point to make. On reflection I've renamed it DownloadCheckable because the check is specific to download, and the upload check is different (and Uploadable uses a different field structure so there's no re-use). Downloadable now just embeds DownloadCheckable and overrides Transfer().

test/cmd/lfstest-gitserver.go
sinbad: Done

sinbad:
Hrm, random failure in checkout test on Travis when it works fine here...

sinbad:
I discovered I can re-run the Travis test on my own fork (I can't here) and it passed at the same commit. https://travis-ci.org/sinbad/git-lfs/builds/75401728

sinbad:
Any chance you can re-run the Travis check on this PR please? I think it's bogus (see previous comment, same commit tested fine in my fork).

rubyist:
Passed!

PR #582, opened 2015-08-13
Pre-push hook must exit with a non-zero result when git-lfs isn't found

sinbad:
Otherwise the branch push will go ahead and data will be incomplete on the server.

Unfortunately even after this is fixed in the code, it won't fix problems in repos which were initialised beforehand. Need to edit .git/hooks/pre-push manually, even "git lfs init --force" won't fix it because the pre-push already exists.

### DIFF
--- a/lfs/setup.go
+++ b/lfs/setup.go
@@ -17,7 +17,7 @@ var (
 	valueRegexp           = regexp.MustCompile("\\Agit[\\-\\s]media")
 	NotInARepositoryError = errors.New("Not in a repository")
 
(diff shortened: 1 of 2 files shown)


technoweenie:
Good call. We changed it to exit 0 so that teams transitioning back off git lfs aren't blocked by this hook. But that's a silly idea in hindsight, especially in native tools that would hide the message on a successful response. Instead, I think the error should tell users how to remove the pre-push hook if they want.

sinbad:
Done, error message now explains how to remove the hook, and also is more specific about not finding git-lfs on the path, rather than saying it's not installed.

technoweenie:
Actually I finished this on the bus ride home last night (building off your first commit): #585

> also is more specific about not finding git-lfs on the path, rather than saying it's not installed. 

:metal:

sinbad:
Awesome, closing this one then.

PR #583, opened 2015-08-13
Add optional remote arg to fetch & pull, and default to tracking remote

sinbad:
The first arg to fetch & pull is now a remote. In addition, the default remote if you don't specify is now the tracking remote as in `git pull` if it exists, and origin if that's not set. This makes it more consistent with the underlying git workflow especially in triangular fork setups.

Note: I had to base this on top of https://github.com/github/git-lfs/pull/573 to avoid conflicts.

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"io"
 	"os"
 	"os/exec"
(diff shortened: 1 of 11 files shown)


technoweenie:
:+1:

rubyist:
:+1:

PR #585, opened 2015-08-14
Extend #582 with more tests

technoweenie:
This adds some tests to confirm that `git lfs update` can automatically update the former pre-push hooks. Also taught `git lfs init --force` how to update the pre-push hook too.


sinbad:
:+1: Nice, much better to have it update itself, I totally missed that you had that concept in there already. 

I think 'not on your path' rather than 'not installed' would be more accurate as a message though, especially to trigger thoughts about cause when using lfs in other environments.

technoweenie:
I brought your error message in. Thanks!

PR #586, opened 2015-08-14
Update install.sh.example

isserrano:
Install script was failing in ubuntu when PATH+=:$prefix/bin

### DIFF
--- a/script/install.sh.example
+++ b/script/install.sh.example
@@ -16,5 +16,5 @@ for g in git*; do
   install $g "$prefix/bin/$g"
 done
 
-PATH+=:$prefix/bin
+PATH=$PATH:$prefix/bin
 git lfs init


ssgelm:
@isserrano what version of ubuntu?  Also, how are you running the script?

isserrano:
14.04 Server

I ran this:

 sudo ./install.sh
./install.sh: 16: ./install.sh: *PATH+=:*/usr/local/bin: not found

ssgelm:
And you are running the install script unmodified?  It looks like the script is being executed with dash instead of bash, however the shebang specifies bash so I'm not sure why that would be happening.

isserrano:
I am using the script as it comes from the download.
And yes, with bash works, but not with dot.

ssgelm:
If you run `sudo env bash` do you end up at a bash prompt?

ssgelm:
Ah, I see the problem.  This is a bug in the currently released script that does not exist in master.  The currently shipping shebang is `#!/bin/sh -eu` which will indeed fail on ubuntu/debian.  This PR is not necessary for the current version in master.

isserrano:
ok, good to know , thanks!

Issue #593, opened 2015-08-17
git lfs fetch with batch on hangs indefinitely


WillHipschman:
Repro steps:

* Remove smudge filter and turn on batch and concurrent transfer:

[filter "lfs"]
       clean = git-lfs clean %f
[lfs]
       batch = true
       concurrenttransfers = 4

* Clone an existing repo
* run "git lfs fetch"

Expected: All pointers are fetched/populated
Actual: cmd shows status: "(0 of 199 files) 0 B / 577.44 MB" indefinitely

technoweenie:
What commit are you on? I think this was fixed a week or so ago.

Also, the smudge filter should have no effect on the fetch command.

WillHipschman:
It's from ~2 weeks ago. I'll sync and try it again :)

WillHipschman:
I'm still seeing it with the latest commit:

commit 4457d7c7c5906025f753579f67f975792235b717

technoweenie:
What does `GIT_TRACE=1 git lfs fetch` show?

WillHipschman:
trace git-lfs: run_command: 'git' rev-parse HEAD
trace git-lfs: run_command: git rev-list --objects --do-walk 1bb5d6f454a0c4c390a858c890c547588cfd42b6
trace git-lfs: run_command: git cat-file --batch-check
trace git-lfs: run_command: git cat-file --batch
trace git-lfs: run_command: 'git' config -l
trace git-lfs: run_command: 'git' config -l -f D:\dev\LfsTest8\.git\config
trace git-lfs: tq: starting 4 transfer workers
trace git-lfs: tq: running as batched queue, batch size of 100
trace git-lfs: tq: sending batch of size 100
trace git-lfs: api: batch 100 files
trace git-lfs: HTTP: POST https://mseng.visualstudio.com/DefaultCollection/Personal/_git/LFSTest.git/info/lfs/objects/batch
trace git-lfs: HTTP: 203
trace git-lfs: api: batch 99 files
trace git-lfs: HTTP: POST https://mseng.visualstudio.com/DefaultCollection/Personal/_git/LFSTest.git/info/lfs/objects/batch
trace git-lfs: HTTP: 203

WillHipschman:
I'm about to debug into the server code--it's possible it's an issue there. But I figured I'd report it until I know for sure.

technoweenie:
[...]

PR #594, opened 2015-08-17
Don't hang when batch calls error

rubyist:
A few problems with the batch code error handling can lead to a hang:

- [x] Actually return an error instead of nil
- [x] Properly manage the waitgroup when a batch call errors
- [x] Make sure we display errors instead of just a 0 count
- [x] Tests for this scenario

/cc #593

### DIFF
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -135,4 +135,12 @@ func fetchAndReportToChan(pointers []*lfs.WrappedPointer, include, exclude []str
 	processQueue := time.Now()
 	q.Wait()
 	tracerx.PerformanceSince("process queue", processQueue)
+
+	for _, err := range q.Errors() {
+		if Debugging || err.Panic {
+			LoggedError(err.Err, err.Error())
+		} else {
+			Error(err.Error())
+		}
+	}
 }
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -218,7 +218,7 @@ func Batch(objects []*objectResource, operation string) ([]*objectResource, *Wra
 	LogTransfer("lfs.api.batch", res)
 
 	if res.StatusCode != 200 {
-		return nil, Errorf(errors.New("batch error"), "Invalid status for %s %s: %d", req.Method, req.URL, res.StatusCode)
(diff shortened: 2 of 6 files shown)


rubyist:
First problem:

```go
return nil, Errorf(nil, "Invalid status for %s %s: %d", req.Method, req.URL, res.StatusCode)
```

Would actually return a `nil` error, so the error could never be reported.

Second problem: Even with properly reporting the error, the waitgroup was never decremented, leading to the actual hang.

Also added: error output for fetch, which had none.

technoweenie:
The fact that `Errorf(nil, ...)` returns nil is stupid. This tweak passes:

```
diff --git a/lfs/errors.go b/lfs/errors.go
index c66a8c4..e79bbf7 100644
--- a/lfs/errors.go
+++ b/lfs/errors.go
@@ -1,10 +1,13 @@
 package lfs
 
 import (
+       "errors"
        "fmt"
        "runtime"
 )
 
+var genericError = errors.New("Git LFS client error")
+
 type WrappedError struct {
        Err     error
        Message string
@@ -19,7 +22,7 @@ func Error(err error) *WrappedError {
 
 func Errorf(err error, format string, args ...interface{}) *WrappedError {
        if err == nil {
-               return nil
+               err = genericError
        }
 
        e := &WrappedError{

```

Now, since this specific case doesn't have an error to wrap, maybe it should just create one?

```golang
Error(fmt.Errorf("Invalid status for %s %s: %d", req.Method, req.URL, res.StatusCode))
```

The WrappedError struct will use that message for both the internal and user-visible error message.

Also, this WrappedError jank is not the best. It may make sense to look into [errgo](https://github.com/juju/errgo) instead. Definitely not for this PR :)

rubyist:
> The fact that Errorf(nil, ...) returns nil is stupid.

:+1:, I'll polish this up a bit.

> Also, this WrappedError jank is not the best. It may make sense to look into errgo instead. Definitely not for this PR :)

I feel the same, we could definitely smooth things out a bit.

rubyist:
I added the `genericError` and used that for now. I've got a couple more steps I can take as part of #595 to go toward cleaning up some error shenanigans.

PR #595, opened 2015-08-17
Retry various transfer failures

rubyist:
We want to retry some calls in the event of an http error. We would typically not retry if the http call was successful and returned a status code, with the exception of upload transfers that can expire and return a 403. There are three calls that can be retried in these cases:

- [x] Retry once when the transfer fails (timeout, http error, 403 with expiry)
- [x] Retry once when a batch api call fails (timeout, http error)
- [x] Retry once when old api call fails (timeout, http error)

And also:
- [ ] Tests!


rubyist:
This is ready for review. It's fairly straight forward, when the api, download, or upload operations run through the transfer queue and get an error without an http response (e.g. timeout or network error) they will be added to the queue again to retry. The queue tracks which objects it has retried and only retries once for a given object. Uploads that receive a 403 will also be retried, this generally happens when an upload token has expired.

I'm not sure how we can automate end to end testing of this because the typical http handlers don't seem to allow you to cause an error that won't return a valid http response. We could add another test server to the rig that could work at a lower level, but I'm not sure that's worth doing. Open to suggestions here.

lfs/batcher.go
```diff
@@ -51,16 +55,22 @@ func (b *Batcher) acceptInput() {
...
-				break Loop
+			select {
+			case t, ok := <-b.input:
+				if ok {
+					batch = append(batch, t)
+				} else {
+					exit = true // input channel was closed by Exit()
+					break Loop
+				}
+			case <-time.After(time.Millisecond * 300):
```
technoweenie: This timeout tracks how long it takes to iterate through a `[]*lfs.WrappedPointer` from the push/pre-push/fetch commands, correct? Is it hanging if some of the pointers are skipped or erroring for some reason?

rubyist: Not quite. For retrying, the transfer queue was changed slightly. Previously, you'd `Add()` all your items, then `Wait()` would start up everything, but it would call batcher's `Exit()` to ensure any remaining items were flushed from the batcher. The side effect of this is that once you called `Wait()` on the transfer queue, you can't `Add()` anything to it. The timeout ensures batches of less than 100 are flushed implicitly. e.g. if the caller is blocked on `Next()` and there are 50 items waiting to be batched, the caller will get them after 300ms. Doing this means we can just use `Add()` to retry an object.

sinbad: Is it worth introducing a Flush() operation that does this & allows further items to be added but without needing it to be based on a timeout? The timeout is cool in that it's automatic, but it also makes it less deterministic - unrelated changes can alter the behaviour just because of how the times work out. I've seen this problem a lot in debugging (stuff happens differently when you're debugging because different amounts of time pass when you stick on a breakpoint) particularly in real-time systems and it often needs a 'pausable' time source to eliminate. Although debugging is not really a practical problem right now due to the lack of decent Go debugging facilities, differences in server response times could cause difficult to repeat behaviour - some of that is unavoidable but I'd prefer a deterministic sequence unrelated to time, or at least without any parallel time component other than server responses, in our own code if that's possible.

lfs/transfer_queue.go
```diff
@@ -278,6 +306,21 @@ func (q *TransferQueue) run() {
 	}
 }
 
+func (q *TransferQueue) canRetry(err error, id string) bool {
+	if !IsRetriableError(err) {
+		return false
+	}
+
+	defer q.retrylock.Unlock()
+	q.retrylock.Lock()
```
sinbad: As an alternative to this lock & map, how about:
```diff
diff --git a/lfs/download_queue.go b/lfs/download_queue.go
index d8e00e2..5c1b0f4 100644
--- a/lfs/download_queue.go
+++ b/lfs/download_queue.go
@@ -2,6 +2,7 @@ package lfs
 
 // The ability to check that a file can be downloaded
 type DownloadCheckable struct {
+	*TransferableBase
 	Pointer *WrappedPointer
 	object  *objectResource
 }
@@ -15,6 +16,7 @@ func (d *DownloadCheckable) Check() (*objectResource, error) {
 }
 
 func (d *DownloadCheckable) Transfer(cb CopyCallback) error {
+	d.RecordAttempt()
 	// just report completion of check but don't do anything
 	cb(d.Size(), d.Size(), int(d.Size()))
 	return nil
@@ -58,6 +60,7 @@ func NewDownloadable(p *WrappedPointer) *Downloadable {
 }
 
 func (d *Downloadable) Transfer(cb CopyCallback) error {
+	d.RecordAttempt()
 	err := PointerSmudgeObject(d.Pointer.Pointer, d.object, cb)
 	if err != nil {
 		return Error(err)
diff --git a/lfs/transfer_queue.go b/lfs/transfer_queue.go
index 636fb2a..30d9992 100644
--- a/lfs/transfer_queue.go
+++ b/lfs/transfer_queue.go
@@ -20,6 +20,18 @@ type Transferable interface {
 	Size() int64
 	Name() string
 	SetObject(*objectResource)
+	AttemptCount() int
+}
+
+type TransferableBase struct {
+	attempts int
+}
+
+func (t *TransferableBase) AttemptCount() int {
+	return t.attempts
+}
+func (t *TransferableBase) RecordAttempt() {
+	t.attempts++
 }
 
 // TransferQueue provides a queue that will allow concurrent transfers.
diff --git a/lfs/upload_queue.go b/lfs/upload_queue.go
index ce06545..649952a 100644
--- a/lfs/upload_queue.go
+++ b/lfs/upload_queue.go
@@ -8,6 +8,7 @@ import (
 
 // Uploadable describes a file that can be uploaded.
 type Uploadable struct {
+	*TransferableBase
 	oid      string
 	OidPath  string
 	Filename string
@@ -42,6 +43,7 @@ func (u *Uploadable) Check() (*objectResource, error) {
 }
 
 func (u *Uploadable) Transfer(cb CopyCallback) error {
+	d.RecordAttempt()
 	wcb := func(total, read int64, current int) error {
 		cb(total, read, current)
 		return nil
```
... and then pass the actual Transferable to this function and retry if t.AttemptCount() < 2.

sinbad:
:+1: other than my general comments which you can either use or ignore :) Took me a little while to follow all the paths to try to look for race conditions but I think it's fine.

lfs/transfer_queue.go
rubyist: I think this would be fine. I'm lukewarm on pushing that responsibility into the Transferrable, but I do like avoiding the mutex.

lfs/batcher.go
rubyist: I think an explicit flush adds more complexity but doesn't make things any more deterministic. The transfer loop blocks on `Next()`, so you'd presumably call `Flush()` when you `Add()` something (otherwise you're just pushing the timeout up). Transfers start as soon as the queue has received a batch worth. Say a transfer fails while the thing feeding the queue has added 52 items, so we `Add()` and `Flush()`, sending off a batch of 53 items. Then another fails and we send off a batch of 10 items.

I think the way to make it operate in a more deterministic manner is to withhold all failures until the end and then retry them.

sinbad: I was imagining that original code that was adding the Transferables would call Flush() after it had finished adding, which would guarantee the batcher would send to batchReady at that point regardless of numbers. I agree that you'd have to separate the failures into a different queue (different batcher?) to submit at the end to be truly deterministic & ensure that every batch was the same regardless of timing. It does add a little code complexity but I think knowing that if you submit the same items in the same order that they will always result in the same requests at the end, regardless of timings, is a nice position to be in from a diagnostic point of view.

technoweenie: Gotcha, thanks for explaining.

sinbad: > My concern wasn't complexity added by withholding retries, rather adding an explicit Flush that's still nondeterministic.

:+1: Makes sense

rubyist:
This revised version withholds all the failed transfers until the end then does the retries by reseting the batcher and going through the flow again. This should keep things a little more deterministic and straight forward.

technoweenie:
This could use another full review with @rubyist's latest changes. I'll be looking at it first thing tomorrow.

lfs/batcher.go
```diff
@@ -39,11 +42,20 @@ func (b *Batcher) Next() []Transferable {
...
+	atomic.StoreUint32(&b.exited, 1)
 	close(b.input)
 }
 
+// Reset allows an Exit()ed batcher to be re-started.
+func (b *Batcher) Reset() {
+	if atomic.CompareAndSwapUint32(&b.exited, 1, 0) {
+		b.input = make(chan Transferable, b.batchSize)
+		go b.acceptInput()
+	}
```
ttaylorr: Should this return an error if the CAS failed?

lfs/batcher.go
```diff
@@ -9,6 +11,7 @@ package lfs
 // When a timeout, Flush(), or Exit() occurs, the group may be smaller than the
 // batch size.
 type Batcher struct {
+	exited     uint32
```
ttaylorr: Even though this field is private, maybe a quick note here about what 1 and 0 mean in the context of this member?

lfs/transfer_queue.go
```diff
@@ -231,7 +276,12 @@ func (q *TransferQueue) transferWorker() {
 		}
 
 		if err := transfer.Transfer(cb); err != nil {
-			q.errorc <- err
+			if q.canRetry(err) {
```
ttaylorr: Perhaps we could extract something to wrap this sort of behavior? At a glance, I see it in three places in this file. Perhaps such a thing may be overkill, but I'm curious of your thoughts.

lfs/batcher.go
rubyist: Yeah it probably should

lfs/batcher.go
rubyist: It's just boolean

lfs/transfer_queue.go
rubyist: One of them goes away when the old api support is removed, an extraction for the remaining two is probably overkill.

lfs/transfer_queue.go
```diff
@@ -76,6 +83,26 @@ func (q *TransferQueue) Wait() {
...
+		}
+		for _, t := range q.retries {
+			q.Add(t)
+		}
+		if q.batcher != nil {
+			q.batcher.Exit()
+		}
+		q.wait.Wait()
+	}
+
```
sinbad: Should set q.retrying back to 0 after this?

sinbad:
:+1: I like the predictability of this approach (under the same error conditions).

An aside - right now you can't re-use the TransferQueue for more operations after Wait() because the WaitGroups will never be re-initialised, I think that should be either catered for or explicitly ruled out in the docs.

rubyist:
> An aside - right now you can't re-use the TransferQueue for more operations after Wait() because the WaitGroups will never be re-initialised, I think that should be either catered for or explicitly ruled out in the docs.

TransferQueue has always worked that way (except for the brief period in this PR), but I agree we could document it better, I'll add that here. Changes to re-use a TransferQueue can be done separately.

lfs/batcher.go
rubyist: On a closer look, this is just acting as a noop if `Reset` is called and the batcher has not been `Exit`ed. Operationally this is fine so I'm just going to document it as such for now.

rubyist: Better yet, I'm going to remove the manual `Reset()` call and just have `Add()` check for the condition. This will also avoid the situation where a panic would occur if an `Add()` is called after an `Exit()` but before a `Reset()`.

lfs/transfer_queue.go
rubyist: It won't have any effect here, since the TQ isn't meant to be reused, but let's add it to pave the way for a reusable TQ in the future.

lfs/transfer_queue.go
technoweenie: I might push for a wrapper function too, especially considering this is the only place that uses `retry()` and `canRetry()`. However, the `if` statements are different enough that they would have to accept a function callback or something. 

```go
q.queueRetry(err, func() {
  tracerx.Printf("tq: retrying object %s", transfer.Oid())
  q.retry(transfer)
})
```

That may save you some lines of code, but I think the current form is easier to read.

technoweenie:
:+1: Looks good.

PR #599, opened 2015-08-21
describe how validation errors are returned from the batch api

technoweenie:
Updates the batch API spec to match https://github.com/github/git-lfs/issues/597.

### DIFF
--- a/docs/api/http-v1-batch.md
+++ b/docs/api/http-v1-batch.md
@@ -194,7 +194,7 @@ Object error codes should match HTTP status codes where possible:
 
 Validation errors can only occur on `upload` requests. Servers must verify
 that OIDs are valid SHA-256 strings, and that sizes are positive integers.
-Servers should also set an upper bound for the allowed object size too. Here's a
+Servers may also set an upper bound for the allowed object size too. Here's a
 response showing one uploadable object, and one with a validation error:
 
 ```
--- a/docs/api/http-v1-batch.md
+++ b/docs/api/http-v1-batch.md
@@ -29,58 +29,6 @@ slight change allows anonymous access to public Git LFS objects. The client
 stores the result of this in the `lfs.<url>.access` config setting, where <url>
 refers to the endpoint's URL.
 
-## API Responses
-
-This specification defines what status codes that API can return.  Look at each
-individual API method for more details.  Some of the specific status codes may
-trigger specific error messages from the client.
-
-* 200 - The request completed successfully.
-* 307 - A temporary redirect.  Keeps the original request method intact.
-* 400 - General error with the client's request.  Invalid JSON formatting, for
-example.
-* 401 - The authentication credentials are needed, but were not sent.
-* 403 - The requesting user has access to see the repository, but not to push
-changes to it.
-* 404 - Either the user does not have access to see the repository, or the
-repository or requested object does not exist.
-
-The following status codes can optionally be returned from the API, depending on
-the server implementation.
-
-* 406 - The Accept header needs to be `application/vnd.git-lfs+json`.
-* 429 - The user has hit a rate limit with the server.  Though the API does not
-specify any rate limits, implementors are encouraged to set some for
-availability reasons.
-* 501 - The server has not implemented the current method.  Reserved for future
-use.
-* 509 - Returned if the bandwidth limit for the user or repository has been
-exceeded.  The API does not specify any bandwidth limit, but implementors may
-track usage.
-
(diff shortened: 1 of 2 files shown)


docs/api/http-v1-batch.md
```diff
@@ -191,11 +191,47 @@ between the requesting user and the repository.
...
+< }
+
+The HTTP 422 response indicates that none of the objects were saved. The client
+may batch up the valid objects for another attempt at uploading them. However,
+clients should abort higher level operations like a Git push since not all of
+the objects were uploaded separately.
+
+It is possible for servers to respond with a 200, and just annotate specific
+objects that failed. Here's a sample request and response that includes one
+missing object:
```
sinbad: Is 'failed' the right word here? I think this example means that the upload wasn't required, that's why it's a download link instead of an upload link. Maybe "It is possible for servers to respond with a 200, and annotate objects that already exist, thus the valid operation is download and not upload."

sinbad: Sorry, previous comment based on diff but didn't apply once I'd expanded it (talking about download not upload). Should this paragraph specifically identify that this applies to the 'download' case?

sinbad: Also, a similar response can happen for upload but for different reasons, i.e. if the object already exists. 

Maybe this should be in a separate section headed '200 partial responses' and give examples for valid cases for download & upload.

technoweenie: I like the idea of splitting out a separate section for 200 partial responses. :+1:

IMO, a request to upload an existing file is not an error. The object should have no error and no actions set on it. The lack of an `upload` action means the client can skip it.

I think in previous API versions, the server returned a `download` action on existing objects instead. I don't think the client even bothers to check this though.

sinbad: > IMO, a request to upload an existing file is not an error. The object should have no error and no actions set on it. The lack of an upload action means the client can skip it.

100% agree, when I said 'a similar response' I meant a missing upload entry for the object in the response, sorry for the ambiguity.

> I think in previous API versions, the server returned a download action on existing objects instead. I don't think the client even bothers to check this though.

Yeah, and I've mirrored that in test servers I've built too :) But that's because the old API didn't have an intended operation; leaving the object out seems more consistent.

technoweenie:
Rewrote how we present the responses in the spec. It now shows the various possible responses: successful, partial error, validation error, 403, 401, etc. I also added a new header for 401 responses: `LFS-Authenticate`. This is an attempt to be forward compatible with NTLM auth (#617).

rubyist:
:+1:

docs/api/http-v1-batch.md
```diff
@@ -241,12 +184,102 @@ a sample request and response that includes one missing object:
...
+
+* 404 - The object does not exist on the server.
+* 410 - The object was removed by the owner.
+* 422 - Validation error.
 
-The error codes are integers that map to similar HTTP status codes where
-possible. Specifics can be included by the server in the `message` property.
+Validation errors can only occur on `upload` requests. Servers must verify
+that OIDs are valid SHA-256 strings, and that sizes are positive integers.
+Servers should also set an upper bound for the allowed object size too. Here's a
```
sinbad: Should this be "Servers *can* also set an upper bound"? I don't think an upper bound would be commonplace for self-hosting, only when you're providing hosting for other people.

sinbad:
Looks good, just one tiny comment on the docs.

docs/api/http-v1-batch.md
technoweenie: I was purposely trying to use one of the rfc conformance words. Though after looking at the [RFC for the RFC conformance words](http://www.ietf.org/rfc/rfc2119.txt), I think you're right:

> SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.

> MAY   This word, or the adjective "OPTIONAL", mean that an item is
   truly optional.  One vendor may choose to include the item because a
   particular marketplace requires it or because the vendor feels that
   it enhances the product while another vendor may omit the same item.
   An implementation which does not include a particular option MUST be
   prepared to interoperate with another implementation which does
   include the option, though perhaps with reduced functionality. In the
   same vein an implementation which does include a particular option
   MUST be prepared to interoperate with another implementation which
   does not include the option (except, of course, for the feature the
   option provides.)

PR #600, opened 2015-08-21
Refactor error wrapping

rubyist:
:warning: Experimental :warning:

This PR contains an experiment in refactoring our error wrapping system. I'm pushing it up for discussion, so don't feel like this is something you have to go along with if you don't like it.

In some [older code](https://github.com/github/git-lfs/blob/4457d7c7c5906025f753579f67f975792235b717/lfs/client.go#L751) I tried out some error inspection techniques outlined by Dave Cheney in [this post](http://dave.cheney.net/2014/12/24/inspecting-errors). I found this to be an interesting approach and agree with his points in the article.

In #595 I'm ending up doing a similar thing. Some errors are retriable conditions, and I need to indicate this down in the client code when they happen. I don't want the `TransferQueue` to be inspecting API response codes or errors to know whether it should retry an operation or not. This type of
behavior wrapping fits well here, and in a few other cases.

So I gave myself a little time this morning to see what this would look like if we went all in on this pattern. The `WrappedError` object we use has gotten in the way before and using this pattern should allow us to keep error wrapping internal to the `lfs` package, having only `error`s coming out of it. No error types are exported, only a handful of functions for checking the error
behavior. I feel this method allows for easier extension without breaking calling code.

Goals of this refactoring:
* Move away from a concrete error wrapper (`WrappedError`), use only `error`
* Avoid exporting any error types
* Avoid comparing any concrete error types
* Define behavior checks for things like retriability, fatal-ness, etc
* Allow behaviors to wrap each other
* Keep the current Stack and Context functionality

Once I got the error wrapping and checking functions working, it was very easy to move all of the existing code over. There aren't too many checks we actually do. There's a little more code involved in setting up a new kind of error to check, but I think that's OK because you should really be questioning whether you really need to inspect an error. Once the boilerplate is in, it's very easy and transparent to use. There's a good chuck of documentation in [errors.go](https://github.com/github/git-lfs/blob/554aef8692d3ffac1b244f00150c624671a7b3e3/lfs/errors.go).

There are a couple things that could use some cleaning up, but I don't want to get any more into it until there's consensus. Particularly, the `Errorf()` isn't fully working yet, and I really don't like the way the
`cleanPointerError` stuff works (I don't like the way it works in master, either).

What do other folks think of this pattern? I'm going to proceed with using the pattern specifically in #595, as the `notImplError` is used, but if there's interest in using this for more, this provides a good start. If you hate it, it's fine, it was fun anyway :satisfied:


commands/command_pointer.go
```diff
@@ -85,7 +85,7 @@ func pointerCommand(cmd *cobra.Command, args []string) {
 		fmt.Printf("Pointer from %s\n\n", pointerName)
 
 		if err != nil {
-			Error(err.Error())
+			Error("Not a valid Git LFS pointer file.")
```
technoweenie: Isn't this losing internal error details? That message is taking from the `lfs.NotAPointerError` var thats' removed in lfs/pointer.go.

technoweenie:
:+1: I really like this idea.

commands/command_pointer.go
rubyist: Yeah, it is. There are a few things left that aren't quite right and need to be polished up before this is fully ready.

rubyist:
I think this is in a good spot. From here we can do a little more work to polish it up as we go. At a later time we can un-export any error checking functions that are only used internally in the `lfs` package. It should also be possible to extract the base error stuff into its own package, but I don't think that's necessary right now.

sinbad:
:+1:  I've noticed that Go uses a mixture of exposed error types and error category functions (e.g. os.IsNotExist()) and I was never sure which was the canonical approach. I definitely prefer either to WrappedError and on reflection the case for functions over types is solid. Looks good.

ttaylorr:
Looks :cool: to me -- this is a really nice refactoring. It's nice that clients of the `lfs` package can accept `error` types instead of wrapped ones while keeping all of the benefits that we can stick on to wrapped errors.

technoweenie:
Thanks for checking this out so quickly :) Merging!

PR #563, opened 2015-08-04
Update LFS license file to reflect MIT license use

hoolio:
Adding language at the top of the file that makes clear that the project is licensed under the MIT license, with a link to the Wikipedia article explaining what that license is. 

Users should have an easy way to determine what standard OSS license applies to a project, if any, by quickly looking at it.

PR #603, opened 2015-08-23
Docker Windows Compatibility

andyneff:
Added Compatibility for Running docker in Windows via boot2docker

- Fixed CentOS createrepo incompatibility on samba shares (not patched until ~Fedora 22)
- Fixed Debian incompatibility with having too many permissions
- In order to get this working, .gitattributes was added so force all newlines to LF only. extentions.md was the only file that need CRLF converted to LF
- Other small fixes for MINGW compatibility

Fixed RPMs to handle #563 
Added DOCKER_OTHER_OPTIONS so that additional arguments can be sent to the containers


technoweenie:
This looks good. Since this isn't used directly by the tool, or even part of the build or test process, I'm going to just merge it. I'm actually setting up a windows vm to try this out (and mess with some windows installer stuff), so I'll get to see it in action soon.

Issue #606, opened 2015-08-25
Push on non-current branches produces incorrect results


sinbad:
I'm just dumping this one here so I don't lose it, I spotted this while trying test my 'fetch recent' feature.

If you create a secondary branch which includes a modification of an existing file, then try to push it from a different branch, it doesn't push the last update to the object but still updates the remote branch pointer.

I demonstrated this by adding the oid to the --dry-run output and running against a simple repo with this setup:

```
steve@Stevens-MacBook-Pro ~/p/G/s/g/g/g/t/t/fetch-recent> git branch
* master
  other_branch
steve@Stevens-MacBook-Pro ~/p/G/s/g/g/g/t/t/fetch-recent> git lfs push --dry-run origin other_branch
push file1.dat (ec3a53b52eb11808d59d29ffd01568edf2d62cdc9c02def84d0a11704d552de2)
push file2.dat (495e32a1c5615c5cf236d99c1ec53903086630eddcba51a1f8b543bf5f2275be)
push file1.dat (d9929b4e48b6ab5d9c5e0da4a5b9bf89f5c63dd0a08c7ebcd2afbbe19436bfdb)
push file1.dat (1a526e57eddca261e4da323bf7b8ca468d27aedf019884e30dfc3cd7342706e1)
steve@Stevens-MacBook-Pro ~/p/G/s/g/g/g/t/t/fetch-recent> git checkout other_branch
Switched to branch 'other_branch'
steve@Stevens-MacBook-Pro ~/p/G/s/g/g/g/t/t/fetch-recent> git lfs push --dry-run origin other_branch
push file1.dat (7bf3771066357ae85beb19c246d6177c5d8183f4fd8ee20eaae3fe54dffa0a72)
push file1.dat (d9929b4e48b6ab5d9c5e0da4a5b9bf89f5c63dd0a08c7ebcd2afbbe19436bfdb)
push file1.dat (1a526e57eddca261e4da323bf7b8ca468d27aedf019884e30dfc3cd7342706e1)
steve@Stevens-MacBook-Pro ~/p/G/s/g/g/g/t/t/fetch-recent>
```

`7bf3771066357ae85beb19c246d6177c5d8183f4fd8ee20eaae3fe54dffa0a72` is the final change on other_branch:

```
steve@Stevens-MacBook-Pro ~/p/G/s/g/g/g/t/t/fetch-recent> git checkout other_branch
Switched to branch 'other_branch'
[...]

PR #607, opened 2015-08-25
Add enhanced test setup tools for 'go test' and integration tests

sinbad:
* Use test.Repo and AddCommits to set up complex multi-branch tests
* Can commit in the past for history tests, branch, merge and tag
* Can generate random data for files or use what you give it
* Returns full detail of the commits created including SHAs and OIDs
* Can pipe JSON from a batch file into lfstest-testutils to get the same
  functionality in acceptance tests
* get_date bash util to create commit dates easily on BSD/Mac and GNU
* see test-push.sh for bash example and scanner_git_test.go for go test

### DIFF
--- a/lfs/pointer_clean.go
+++ b/lfs/pointer_clean.go
@@ -31,8 +31,7 @@ func PointerClean(reader io.Reader, fileName string, fileSize int64, cb CopyCall
 	var oid string
 	var size int64
 	var tmp *os.File
-	// Initialise to empty set rather than nil for consistency with pointer decode
-	var exts []*PointerExtension = make([]*PointerExtension, 0)
+	var exts []*PointerExtension
 	if len(extensions) > 0 {
 		request := &pipeRequest{"clean", reader, fileName, extensions}
 
--- a/lfs/pointer_clean.go
+++ b/lfs/pointer_clean.go
@@ -31,7 +31,8 @@ func PointerClean(reader io.Reader, fileName string, fileSize int64, cb CopyCall
 	var oid string
 	var size int64
 	var tmp *os.File
-	var exts []*PointerExtension
+	// Initialise to empty set rather than nil for consistency with pointer decode
+	var exts []*PointerExtension = make([]*PointerExtension, 0)
 	if len(extensions) > 0 {
 		request := &pipeRequest{"clean", reader, fileName, extensions}
 
--- a/lfs/scanner_git_test.go
+++ b/lfs/scanner_git_test.go
@@ -1,0 +1,76 @@
+package lfs_test // to avoid import cycles
+
+// This is for doing complete git-level tests using test utils
+// Needs to be a separate file from scanner_test so that we can use a diff package
+// which avoids import cycles with testutils
+
+import (
+	"testing"
+
+	. "github.com/github/git-lfs/lfs"
+	"github.com/github/git-lfs/test"
+	"github.com/github/git-lfs/vendor/_nuts/github.com/technoweenie/assert"
+)
+
+func TestScanUnpushed(t *testing.T) {
+	repo := test.NewRepo(t)
+	repo.Pushd()
+	defer func() {
+		repo.Popd()
+		repo.Cleanup()
+	}()
+
+	inputs := []*test.CommitInput{
+		{ // 0
+			Files: []*test.FileInput{
+				{Filename: "file1.txt", Size: 20},
+			},
+		},
+		{ // 1
+			NewBranch: "branch2",
+			Files: []*test.FileInput{
+				{Filename: "file1.txt", Size: 25},
+			},
+		},
+		{ // 2
+			ParentBranches: []string{"master"}, // back on master
+			Files: []*test.FileInput{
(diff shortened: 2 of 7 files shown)


sinbad:
I've pulled this out of my fetch-recent branch because it's useful to test #606

test/test-push.sh
```diff
@@ -84,3 +84,50 @@ begin_test "push object id(s)"
...
+      {\"Filename\":\"file1.dat\",\"Size\":${#content3}, \"Data\":\"$content3\"},
+      {\"Filename\":\"file2.dat\",\"Size\":${#content4}, \"Data\":\"$content4\"}]
+  }
+  ]" | lfstest-testutils addcommits
+
+  git push origin master
+  assert_server_object "$reponame" "$oid1"
+  assert_server_object "$reponame" "$oid2"
+  assert_server_object "$reponame" "$oid3"
+  assert_server_object "$reponame" "$oid4"
```
sinbad: FYI I've deliberately only pushed one branch here to avoid #606 - I'll use this branch as a base to prove & fix that bug by adding a secondary branch.

lfs/pointer_clean.go
```diff
@@ -31,7 +31,8 @@ func PointerClean(reader io.Reader, fileName string, fileSize int64, cb CopyCall
 	var oid string
 	var size int64
 	var tmp *os.File
-	var exts []*PointerExtension
+	// Initialise to empty set rather than nil for consistency with pointer decode
```
rubyist: I'm not sure I understand this

sinbad: If a pointer is initialised from DecodePointer() it gets an empty array rather than nil, see decodeKV(). PointerClean() initialised it to nil. Comparing the pointers for the same data which came through the 2 paths with assert() would always fail because of this. Besides being a pain for testing it should really be consistent whether a lack of extensions is nil or an empty array.

rubyist: Which assert would fail in this case? If I change the line back I don't get any failures here. It shouldn't matter if a slice is really `nil`, it can always be treated as a slice with empty checks being done with `len()`. I don't think anything should be comparing a slice with `nil` directly.

Aside from that, I don't think the comment adds much clarity. I also think we should prefer `:=` to `var =` when initializing with a value.

lfs/scanner_git_test.go
```diff
@@ -0,0 +1,76 @@
+package lfs_test // to avoid import cycles
+
+// This is for doing complete git-level tests using test utils
+// Needs to be a separate file from scanner_test so that we can use a diff package
+// which avoids import cycles with testutils
+
+import (
+	"testing"
+
+	. "github.com/github/git-lfs/lfs"
```
rubyist: I don't think it's necessary to use the `import .` here. It's fine to use `lfs.ScanUnpushed()` in the test.

sinbad: It's not necessary, it's just a pattern I've seen used in this case of avoiding import cycles with test utils. e.g. http://golang.org/src/strings/search_test.go

lfs/pointer_clean.go
sinbad: Turns out that test is still in my other branch & didn't come across with this. For example:
```
	expected := []*WrappedPointer{
		{Name: "folder/nested.txt", Size: outputs[3].Files[0].Size, Pointer: outputs[3].Files[0]},
		{Name: "folder/nested.txt", Size: outputs[0].Files[2].Size, Pointer: outputs[0].Files[2]},
		{Name: "folder/nested2.txt", Size: outputs[0].Files[3].Size, Pointer: outputs[0].Files[3]},
	}
	// Need to sort to compare equality
	sort.Sort(test.WrappedPointersByOid(expected))
	sort.Sort(test.WrappedPointersByOid(pointers))
	assert.Equal(t, expected, pointers)
```
.. since it's possible to be this specific with the new test setup tools. I don't think I ported the sort utils over to this branch but they're something else you need since the assert lib doesn't have a ConsistsOf() method like Ginkgo/Gomega.

The comment was there to document the reason for changing the line which isn't obvious unless you also look at decodeKV(). An alternative is to change decodeKV() to use nil for extensions by default (that would be my preference actually, avoid unnecessary allocations). I'll take it out of this PR since it's not needed & put it back in as a nil in my upcoming fetch-recent branch.

lfs/scanner_git_test.go
rubyist: They don't really need to use it there, either, they're only using two things from `strings` :wink: [Effective Go](https://golang.org/doc/effective_go.html#package-names) prefers to avoid it, but I really only noticed it because some tooling doesn't handle it well (e.g. `godef`).

rubyist:
:+1: from me, seems really useful.

lfs/scanner_git_test.go
sinbad: Yeah the general advice is to not use `import . foo`, but there are a couple of officially sanctioned exceptions in test code - it's mostly just so it feels like normal test code even though it has to be in a different package to break cycles.

PR #610, opened 2015-08-27
Support for `git lfs fetch --recent`

sinbad:
`git lfs fetch --recent` downloads files for things other than your current checkout, to pre-empt their use and thus avoid making the user wait when they switch a branch, or want to check out recent changes.

It can do 2 things:
* Fetch a snapshot of files at 'recent branches', committed within a configurable number of days of the current date. This means switching to those branches is far quicker than before.
* Fetch previous versions of files that existed before a branch tip, committed within a configurable number of days of the last commit on that branch. This applies to the current HEAD and all 'recent branches', if any. This means looking at recent changes on interesting branches is quick.

You can also make `--recent` the default via lfs.fetchrecentalways. 

The fetch operations are deliberately ordered so the most useful things are done first:

1. Fetch snapshot of the current HEAD
2. Fetch recent branches, most recent to least recent
3. Fetch recent previous versions, starting with HEAD and then recent branches, most recent to least

This means that this command can be left in the background if you want; once the current ref is finished you can carry on working and let the recent fetching happen in the background. Reporting on the command line has been upped a little bit so the user can see which part of the fetch is going on.

This PR also includes:
* A dedicated 'config' man page listing all of the `git config` options that Git LFS adds
* fetch at a ref now *only* downloads the state at that ref, and not previous versions (without `--recent`). Previously fetch would use the `--do-walk` option to `git rev-list` which means it would include any previous changes to the same files and also files which were deleted. This is more efficient and the `--recent` option gets time-bounded previous versions.
* Utility methods like CurrentRef now return a Ref struct instead of a single value; this saves round-trips when you need the name of the ref as well as the SHA (also categorises into branch/remote branch/tag)
* When files are skipped in fetch because of the include/exclude filter, this is reported if GIT_TRACE=1
* Tests added for CurrentRef and CurrentRemoteRef


commands/command_fetch.go
```diff
@@ -70,13 +89,74 @@ func fetchRefToChan(ref string, include, exclude []string) chan *lfs.WrappedPoin
...
+		Panic(err, "Could not scan for Git LFS previous versions")
+	}
+	fetchPointers(pointers, include, exclude)
+}
+
+// Fetch recent objects based on config
+func fetchRecent(alreadyFetchedRefs []*git.Ref, include, exclude []string) {
+	fetchconf := lfs.Config.FetchPruneConfig()
+
+	if fetchconf.FetchRecentRefsDays == 0 && fetchconf.FetchRecentCommitsDays == 0 {
```
technoweenie: It'd be more correct to use `< 1` here, though it looks like `FetchPruneConfig()` won't even set these properties unless the value is greater than 0.

sinbad: I thought about using <= 0 (< 1 works except if we allow fractional days in future) but it just seemed less intuitive. As you say the config prevents any value < 0 anyway.

docs/man/git-lfs-config.1.ronn
```diff
@@ -0,0 +1,92 @@
+git-lfs-config(1) -- Configuration options for git-lfs
+======================================================
+
+## CONFIGURATION FILES
+
+git-lfs uses the same configuration files as git-config(1) with the same
+precedence. Most options pertaining to git-lfs are contained in the `[lfs]`
+section, meaning they all named `lfs.foo` or similar, although Occasionally an
```
technoweenie: "Occasionally" should be lower case.

docs/man/git-lfs-config.1.ronn
```diff
@@ -0,0 +1,92 @@
+git-lfs-config(1) -- Configuration options for git-lfs
```
technoweenie: I'm certainly no `man` expert, but I think 5 is a more appropriate man section: "File formats and conventions" (https://en.wikipedia.org/wiki/Man_page). Git tracks the config values in [git-config(1)](https://www.kernel.org/pub/software/scm/git/docs/git-config.html), but that goes with the `git config` command. Hooks with no related command are documented in [githooks(5)](https://www.kernel.org/pub/software/scm/git/docs/githooks.html).

technoweenie:
This branch looks great :metal: I'm ready to merge this once the (nitpicky) man page stuff is sorted.

docs/man/git-lfs-config.1.ronn
sinbad: Makes sense, I just copied git-config.

sinbad:
I've made the man changes, and I've also changed the defaults for remote ref fetching after some reflection (latest commit).

PR #611, opened 2015-08-27
only check git-credentials for lfs api requests

technoweenie:
This changes Git LFS so it only checks `git-credential` for LFS API requests. The Git LFS API is intended to be mounted next to the Git server. So it makes sense to use git credentials, since it can take advantage of existing credentials for Git users that use HTTPS remotes.

    https://github.com/foo/bar.git # git remote
    https://github.com/foo/bar.git/info/lfs/... # lfs remote

However, the client shouldn't have to check git credentials for the "storage api" requests. These are the requests to download or upload raw objects from URLs linked from the LFS API response.  The Storage API may not require any authorization for public objects in public repositories. Basically, the Git LFS client should trust the LFS server to return enough meta data to authenticate the requests.

1. `doApiRequestWithRedirects()` is the low level function that handles API communications. It now fetches the credentials [if needed](https://github.com/github/git-lfs/blob/31ccc2293c11ea7e436b336aac0f42125c44feb7/lfs/client.go#L646-L662), and [saves the credentials on success](https://github.com/github/git-lfs/blob/31ccc2293c11ea7e436b336aac0f42125c44feb7/lfs/client.go#L397).
2. With the Batch API, there's now a distinction between public and private repos. Initially a repo is considered public, unless the batch API returns a 401. Then a git config value is set so future requests treat the repo as private. `doApiBatchRequest()` now [tells `doApiRequestWithRedirects()` if credentials are needed](https://github.com/github/git-lfs/blob/31ccc2293c11ea7e436b336aac0f42125c44feb7/lfs/client.go#L466)

/cc @rubyist @sinbad


lfs/client.go
```diff
@@ -156,12 +155,13 @@ func DownloadCheck(oid string) (*objectResource, *WrappedError) {
 }
 
 func DownloadObject(obj *objectResource) (io.ReadCloser, int64, *WrappedError) {
-	req, creds, err := obj.NewRequest("download", "GET")
+	tracerx.Printf("WTF: %v", obj)
```
WillHipschman: WTF is a fun trace message :)

rubyist: Surely it means "Waiting To Fetch"!

technoweenie: Gah, 2nd debug message that I left in here.

*force pushes*

These are not the commits you are looking for.

technoweenie:
I think we can fix #612 in this PR:

> In client.go in getCreds if we make a request against mymachinename but the remote is localhost then we will not authenticate correctly.
>
>```go
>if req.URL.Scheme != apiUrl.Scheme ||
>    req.URL.Host != apiUrl.Host {
>```

I believe that check was added so that the LFS client doesn't apply credentials from your LFS server at `https://lfs-server.com` to the storage requests hosted on a 3rd party system like S3. But since this PR only uses git credentials for LFS API requests, I think we can remove this check.

WillHipschman:
Sounds good to me :+1:

sinbad:
While personally I would always get the API server to generate sufficient authentication tokens to pass to the storage server, I think dropping support for storage credentials entirely might be a step too far. How about supporting 401 responses from the storage server and fetching credentials only then, just like the API calls?

Even though the storage server might not be hosted in the same place as the git server, git-credentials is still a pretty solid way to cache authentication data, and it would be nice to retain the ability to use basic auth on storage calls, for systems that don't support tokens.

lfs/client.go
```diff
@@ -380,10 +378,21 @@ func doHttpRequest(req *http.Request, creds Creds) (*http.Response, *WrappedErro
 	return res, wErr
 }
 
-func doApiRequestWithRedirects(req *http.Request, creds Creds, via []*http.Request) (*http.Response, *WrappedError) {
-	res, wErr := doHttpRequest(req, creds)
+func doApiRequestWithRedirects(req *http.Request, via []*http.Request, useCreds bool) (*http.Response, *WrappedError) {
+	var creds Creds
+	if useCreds {
+		c, err := getCreds(req)
```
ttaylorr: Could probably push this value onto `creds` itself, since we return anyway if there's an non-nil value in `err`.

lfs/client.go
```diff
@@ -422,15 +431,15 @@ func doApiRequestWithRedirects(req *http.Request, creds Creds, via []*http.Reque
...
 
-		return doApiRequestWithRedirects(redirectedReq, redirectedCreds, via)
+		return doApiRequestWithRedirects(redirectedReq, via, useCreds)
 	}
 
 	return res, nil
 }
 
-func doApiRequest(req *http.Request, creds Creds) (*http.Response, *objectResource, *WrappedError) {
+func doApiRequest(req *http.Request) (*http.Response, *objectResource, *WrappedError) {
```
ttaylorr: Both this and `doApiBatchRequest` have the same input and return parameters.  I can't think of one off the top of my head, but is there a type we could extract to represent the return parameters (or the entire method itself)?

lfs/client.go
```diff
@@ -559,22 +568,22 @@ func newApiRequest(method, oid string) (*http.Request, Creds, error) {
...
+		return nil, err
 	}
 
 	req.Header.Set("Accept", mediaType)
-	return req, creds, nil
+	return req, nil
 }
 
-func newClientRequest(method, rawurl string, header map[string]string) (*http.Request, Creds, error) {
+func newClientRequest(method, rawurl string, header map[string]string) (*http.Request, error) {
```
ttaylorr: It probably doesn't matter since this func is non-exported, but potentially we could grab a type `ClientRequestParam` and pass that as one parameter instead of these three?

lfs/client.go
rubyist: These and the batch api calls will be unified very soon, putting everything through the batch api. I wouldn't spend any time trying to build an abstraction that covers both cases.

lfs/client.go
technoweenie: We could, but I don't think it's necessary in this case. All but one uses of this pass all three arguments.

technoweenie:
@sinbad: While I think forcing users to maintain two sets of authentication is a poor user experience, I agree that having `git-credential` track stuff for the storage API could be useful.

lfs/client.go
technoweenie: I don't understand, push what into `creds`. The error?

ttaylorr: Whoops, I see what you were doing here, ignore my suggestion.

technoweenie:
I've updated this PR based on some feedback:

> How about supporting 401 responses from the storage server and fetching credentials only then, just like the API calls?

I've refactored the credential and client methods a bit. All credential related functions are in `lfs/credentials.go`. Some things were renamed. Some exported types were removed to simplify the code. I also moved a lot of the manual `Creds` handling to `doHttpRequest()`. `doHttpRequest()` now takes an optional `Creds` arg. If given, it will accept or reject it based on the response. There are now three entry points to run HTTP requests:

* `doAPIRequest()` talks to the LFS API using credentials pulled from `getCredsForAPI()`. This is only directly used for the "verify" API calls.
* `doApiBatchRequest()` and `doLegacyApiRequest()` both use `doAPIRequest()`, but also parse the response JSON.
* `doStorageRequest()` directly uploads or downloads files using credentials pulled from `getCreds()`.

One potential problem. Storage requests may add a lot of entries to the git credential helper if `useHttpPath` is set. I wonder if we can disable this for a specific `git credential` call, regardless of the git config?

> getCreds incorrectly compares a requests URL scheme when against localhost #612 

There are now two functions for filling an HTTP request with an Authorization header:

* `getCreds()` asks `git credential` for the request URL.
* `getCredsForAPI()` tries to get the credentials from the configured LFS or Git remote URLs before asking `git credential`. If the either the host or scheme of the request URL does not match the LFS or Git remote URLs, then it still asks `git credential` (instead of bailing).

> getCreds fails when CredentialManagers don't return optional fields #614 

It now ignores any exit code related errors from `git credential fill`. It still raises problems if the `reject` or `accept` subcommands exit unsuccessfully, however.

technoweenie:
This is looking good in my QA tests :metal:

PR #613, opened 2015-08-27
lfs/batcher: introduce test cases

ttaylorr:
This commit adds the notion of a test case into the `lfs.Batcher` tests. Each
test case has three parameters: a batch size, the number of items to add to that
batcher, and whether or not it should exit the batcher after adding all items.

This commit also adds a helper func called `assertAll` which processes a set of
test cases, throwing assertion errors into the `*testing.T` along the way.

On the other hand, this commit does not contain any utilities to assert that a
deadlock will occur when feeding items into a Batcher, i.e., when the batch size
is set to 0.

### DIFF
--- a/lfs/batcher_test.go
+++ b/lfs/batcher_test.go
@@ -7,28 +7,66 @@ import (
 )
 
 func TestBatcherSizeMet(t *testing.T) {
-	b := NewBatcher(2)
-
-	for i := 0; i < 4; i++ {
-		b.Add(&Downloadable{})
-	}
-
-	group := b.Next()
-	assert.Equal(t, 2, len(group))
-
-	group = b.Next()
-	assert.Equal(t, 2, len(group))
+	assertAll([]batcherTestCase{
+		{2, 4, false},
+		{3, 5, false},
+		{0, 0, false},
+	}, t)
 }
 
 func TestBatcherExit(t *testing.T) {
-	b := NewBatcher(4)
+	assertAll([]batcherTestCase{
+		{2, 4, true},
+		{3, 5, true},
+		{0, 0, true},
+	}, t)
+}
 
-	for i := 0; i < 2; i++ {
-		b.Add(&Downloadable{})
+// Type batcherTestCase specifies information about how to run a particular test
+// around the type lfs.Batcher.
+type batcherTestCase struct {
+	BatchSize  int
+	ItemCount  int
+	ShouldExit bool
+}
+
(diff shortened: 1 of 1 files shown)


lfs/batcher_test.go
```diff
@@ -7,28 +7,66 @@ import (
...
+	assertAll([]batcherTestCase{
+		{2, 4, true},
+		{3, 5, true},
+		{0, 0, true},
+	}, t)
+}
+
+// Type batcherTestCase specifies information about how to run a particular test
+// around the type lfs.Batcher.
+type batcherTestCase struct {
```
technoweenie: Go comments for types and funcs start with the name. The only godoc comments that have a prefix are package comments.  https://golang.org/doc/effective_go.html#commentary

lfs/batcher_test.go
```diff
@@ -7,28 +7,66 @@ import (
...
+		return 0
 	}
 
-	b.Exit()
+	return b.ItemCount / b.BatchSize
+}
+
+// Func assertAll processes all test cases, throwing assertion errors if they
+// fail.
+func assertAll(cases []batcherTestCase, t *testing.T) {
```
technoweenie: I'd like to see a more descriptive name here, like `runBatcherTests`. A name like `assertAll` seems too generic here. This seems to be more than just a simple assertion, as well.

ttaylorr:
Thanks for the suggestions, @technoweenie.  I've pushed commits to address both of your comments.

technoweenie:
Thanks for the tests!

lfs/batcher_test.go
```diff
@@ -7,28 +7,66 @@ import (
...
+// around the type lfs.Batcher.
+type batcherTestCase struct {
+	BatchSize  int
+	ItemCount  int
+	ShouldExit bool
+}
 
-	for i := 0; i < 4; i++ {
-		b.Add(&Downloadable{})
+// Batcher makes and retunrs a lfs.Batcher according to the specification given
```
WillHipschman: nit: typo, should be "returns"

ttaylorr:
Thanks for the review @technoweenie, @WillHipschman!

lfs/batcher_test.go
technoweenie: Doh! 3f1eec26a25b0c8e17f8b89c3d3dd419189d80c5

ttaylorr: Thanks for fixing that!  Totally missed it.

PR #615, opened 2015-08-28
lfs/batcher: refactor `run` goroutine

ttaylorr:
This commit introduces several refactorings to the `run` method (now called
`acceptInput`) on the Batcher type.

- A call to `acceptInput` directly does not spawn a goroutine. The recommended
  way to initialize is to spawn it as a goroutine, which is how the constructor
  method handles initialization of the Batcher type.

- The responsibility of creating new batches is pushed down into the `newBatch`
  func, which allocates a slice of `Transferable`s and sets its capacity at the
  maximum batch size.

### DIFF
--- a/lfs/batcher.go
+++ b/lfs/batcher.go
@@ -17,9 +17,9 @@ func NewLot(l, c int) Lot {
 }
 
 // IsFull returns whether or not the given instance of a Lot is full, as
-// specified by the capacity given as the first argument.
-func (l Lot) IsFull(capacity int) bool {
-	return len(l) == capacity
+// specified by the max length given as the first argument.
+func (l Lot) IsFull(maxLen int) bool {
+	return len(l) == maxLen
 }
 
 // Add returns a new Lot that contains all of the contents of the existing Lot,
--- a/lfs/batcher.go
+++ b/lfs/batcher.go
@@ -8,38 +8,18 @@ package lfs
 //   * Exit() is called
 // When a timeout, Flush(), or Exit() occurs, the group may be smaller than the
 // batch size.
-
-// A Lot represents a group of Transferables that was packaged up by a Batcher.
-type Lot []Transferable
-
-func NewLot(l, c int) Lot {
-	return Lot(make([]Transferable, l, c))
-}
-
-// IsFull returns whether or not the given instance of a Lot is full, as
-// specified by the max length given as the first argument.
-func (l Lot) IsFull(maxLen int) bool {
-	return len(l) == maxLen
-}
-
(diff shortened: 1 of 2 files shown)


lfs/batcher.go
```diff
@@ -8,21 +8,41 @@ package lfs
 //   * Exit() is called
 // When a timeout, Flush(), or Exit() occurs, the group may be smaller than the
 // batch size.
+
+// A Lot represents a group of Transferables that was packaged up by a Batcher.
+type Lot []Transferable
```
WillHipschman: I don't love the name 'Lot' I had to look up the definition (and it makes sense!) but it's not a commonly used word in this way, especially not in programming. Is there  a reason we don't use 'Batch" or "GroupOfTransferables" or something like that that's a bit more clear?

ttaylorr: Batch was my first choice, but is already defined in this package [here](https://github.com/github/git-lfs/blob/master/lfs/client.go#L177). I'm not to averse of `Lot`, but `TransferableGroup` would also work well imho.

rubyist:
I like the rename of `run` to `acceptInput`, and I'm good with pushing the `go` call up. I also agree with @WillHipschman about the name `Lot`, but looking at this from a higher level I'm wondering what advantage adding a `Lot` type brings?

`Batcher` is under the `lfs` package, but we can examine it as though it were its own package. In that light, adding an exported `Lot` type increases the surface area of the Batcher API. There are now more things a client has to know when using the API and there are more things to track when extending or maintaining the API to avoid breakage. We should always be mindful when doing this and ensure there is a clear advantage.

From my perspective, clients of the API now receive a `Lot` type from `Next()` instead of a slice, but none of the added functionality is particularly useful to (or currently used by) the client. The batching operation should be fairly opaque, so a client shouldn't care if it's a "full" batch or not, and while a client *can* modify the slice `Next()` gives back, providing an exported `Add()` indicates that it could be a reasonable thing to do.

The wrapping `Lot` provides is only used internally so it shouldn't be exported, but it's also only wrapping very simple builtins (`append()` and `len()`). I don't think the added code and complexity are worth doing this at this point.

ttaylorr:
I agree, making clients receive `Lot`s instead of slices was a choice I didn't consider enough when I wrote this patch.  From how I see it, we have two options:

- Remove the Lot type entirely. In the Batcher, we'll just operate directly on the `[]Transferable` instead.

- Make `Lot` exported and unwrap the underlying slice it to all exported methods.

Normally, I'd spring for the second one, but we're wrapping such basic operations, I think the first one is probably a better choice.

rubyist:
:+1: thanks for the refactor!

PR #617, opened 2015-08-28
Initial NTLM Proposal

WillHipschman:
Here's an initial proposal for NTLM auth in git lfs.

I need to update it with specifics on the Client interface but I want to wait until #611 is merged to do that.

Let me know if you have thoughts or questions.

Hacked together implementation of push here https://github.com/WillHipschman/git-lfs/tree/ntlm

### DIFF
--- a/docs/proposals/ntlm.md
+++ b/docs/proposals/ntlm.md
@@ -1,0 +1,27 @@
+++# NTLM Authentication With Git-Lfs
+++
+++Enterprise users in a windows ecosystem are frequently required to use integrated auth. Basic auth does not meet their security requirements and setting up SSH on Windows is painful. 
+++
+++There is an overview of NTLM at http://www.innovation.ch/personal/ronald/ntlm.html
+++
+++### Implementation
+++
+++If the LFS server returns a "Www-Authenticate: NTLM" header, we will set lfs.{endpoint}.access to be ntlm and resubmit the http request. Subsequent requests will
+++go through the ntlm auth flow.
+++
+++We will store NTLM credentials in the credential helper. When the user is prompted for their credentials they must use username:{DOMAIN}\{user} and password:{pass}
+++
+++The ntlm protocl will be handled by an ntlm.go class that hides the implementation of InitHandshake, Authenticate, and Challenge. This allows miminal changesto the existing
+++client.go class.
+++
+++### Tech
+++
+++There is a ntlm-go library available at https://github.com/ThomsonReutersEikon/go-ntlm that we can use. We will need to implementate the Negotiate method and publish docs on what NTLM switches we support. I think simple user/pass/domain is best here so we avoid supporting a million settings with conflicting docs.
+++
+++### Work
+++
+++Before supporting this as a mainstream scenario we should investigate making the CI work on windows so that we can successfully test changes.
+++
+++### More Info
+++
+++You can see a hacked-together implementation of git lfs push with NTLM at https://github.com/WillHipschman/git-lfs/tree/ntlm


technoweenie:
Does Git support NTLM authentication without having to go through an HTTP proxy? [This](http://gasparnagy.com/2013/11/connect-to-tfs-2013-git-repository-with-libgit2sharp/) is the only non-proxy info that I could find. I'm guessing you're trying to do the same thing here.

WillHipschman:
Yeah, I'm trying to do something similar. The link above points to a change made by some of my colleagues, to accomplish the same thing I'm trying to do for git-lfs.

technoweenie:
I'm not in favor of users setting the `lfs.ntlm` config value. This is an extra configuration step that I'd like to avoid if possible. The 401 response from the API should tell you what it needs to authenticate. For instance, the spec says to send a header like `WWW-Authenticate: Basic realm="myRealm"`. What if the LFS server could send back `WWW-Authenticate: NTLM` or something.

Also, I think instead of `lfs.ntlm`, we could have a configurable auth type:

    $ git config lfs.auth [ ntlm | basic ]

Note to self: the current API probably isn't returning `WWW-Authenticate`. I think it only _really_ matters for web browsers, but we should include it in the spec anyway.

technoweenie:
Noticed that the ntlm spec you linked uses that header too:

```
    2: C <--  S   401 Unauthorized
                  WWW-Authenticate: NTLM
```

WillHipschman:
Definitely agree with decided the auth type based on the server response. So we then have:

1. The client is configured by default to use basic auth using the lfs.auth switch. If we see an NTLM server response during this time we can fault into NTLM mode (or should we fail?)

2. If the user sets their client to use NTLM using the lfs.auth switch we immediately go into ntlm mode and skip the extra hop.

technoweenie:
The batch API should automatically skip this extra hop (if you're referring to the initial auth-less request that returns 401). Check out [the `PrivateAccess()` and `SetPrivateAccess()` methods](https://github.com/github/git-lfs/blob/3f1eec26a25b0c8e17f8b89c3d3dd419189d80c5/lfs/config.go#L125-L151). 

Maybe we can set `lfs.{endpoint}.access` to `basic` or `ntlm`, instead of `private`. The fact that it requires any auth at all means it's not public.

whoisj:
For performance reasons, it would be nice to have the optional config which says "hey, I know this will be NTLM so skip the failed auth challenge and proceed directly to Go and collection $200".

This can be hyper critical in fast build scenarios where build machines do not have branch or even repo affinity.

technoweenie:
@whoisj The batch API client already uses the `lfs.{endpoint}.access` git config key for that. Check out my previous reply.

WillHipschman:
I agree we can use the ````lfs.{endpoint}.access```` key. The hop I'm referring to is number 1 below.

**Without ntlm set**

    1: C  --> S   GET ... 
    2: C <--  S   401 Unauthorized
                  WWW-Authenticate: NTLM
    3: C  --> S   GET ...
                  Authorization: NTLM <base64-encoded type-1-message>
    4: C <--  S   401 Unauthorized
                  WWW-Authenticate: NTLM <base64-encoded type-2-message>
    5: C  --> S   GET ...
                  Authorization: NTLM <base64-encoded type-3-message>
    6: C <--  S   200 Ok

**With ntlm set**

    1: C  --> S   GET ...
                  Authorization: NTLM <base64-encoded type-1-message>
    2: C <--  S   401 Unauthorized
                  WWW-Authenticate: NTLM <base64-encoded type-2-message>
    3: C  --> S   GET ...
                  Authorization: NTLM <base64-encoded type-3-message>
    4: C <--  S   200 Ok

WillHipschman:
It's also worth noting that once we do the above exchange once, every request sent through that same httpClient won't have to re-authenticate, so this will be efficient from a batching perspective.

technoweenie:
Yes, that's handled [here](https://github.com/github/git-lfs/blob/0cca12bf80bc3a9612cfb788dacbf689431c9d50/lfs/client.go#L635-L645).

technoweenie:
I added a new `LFS-Authenticate` header to the API spec in #599. This should make the client forward compatible with NTLM and other potential authentication methods. We're purposely not using `WWW-Authenticate` for browser security reasons. With that, the flow should look like this:

```
# with no lfs.{endpoint}.access value
1: C  --> S   GET ...
2: C <--  S   401 Unauthorized
              LFS-Authenticate: NTLM
3: C  --> S   GET ...
              Authorization: NTLM <base64-encoded type-1-message>
4: C <--  S   401 Unauthorized
              WWW-Authenticate: NTLM <base64-encoded type-2-message>
5: C  --> S   GET ...
              Authorization: NTLM <base64-encoded type-3-message>
6: C <--  S   200 Ok

# with lfs.{endpoint}.access == "ntlm"
1: C  --> S   GET ...
              Authorization: NTLM <base64-encoded type-1-message>
2: C <--  S   401 Unauthorized
              WWW-Authenticate: NTLM <base64-encoded type-2-message>
3: C  --> S   GET ...
              Authorization: NTLM <base64-encoded type-3-message>
4: C <--  S   200 Ok
```

If the go package can work with `LFS-Authenticate`, I think that'd be preferable. But it's not worth forking it or adding some config value that only we'd ever use.

WillHipschman:
I've updated the proposal to include the lfs.{endpoint}.access scheme and included info on storing usernames.

Why can't we use Www-Authenticate? It's specific to the NTLM protocol, so I really don't think we should change it.

technoweenie:
> Why can't we use Www-Authenticate?

Browser security stuff. We don't want LFS API urls triggering login popups in browsers.

technoweenie:
Closing this since we merged #714 :metal:

PR #619, opened 2015-08-30
lfs: promote Hooks and Attributes to types

ttaylorr:
This commit introduces two new types into the API: Hook, and Filter.

Both `Hook` and `Filter` are abstractions built on Git hooks and filters
respectively. Each type knows how to install and uninstall itself. These
processes are utilized by the setup method in the `lfs` package, and the
appropriate calls have been updated in the init and uninit commands.

These abstractions were introduced to make adding/removing required filters and
hooks easier for future projects, including the migration away from the smudge
filter.

Eventually it seems appropriate to move both new types into the `git` package,
as opposed to the `lfs` package. At the time of writing this commit, there is
some coupling against state defined in the `lfs` package (i.e., whether or not
we're currently in a git repo, the local working directory, etc). At somepoint
it would be nice to remove that coupling and move these new types into the
`git` package.


ttaylorr:
Looking into why the build doesn't pass, `script/cibuild` passes for me locally, but not for Travis it seems.

lfs/filters.go
```diff
@@ -0,0 +1,76 @@
...
+		return true
+	}
+
+	return valueRegexp.MatchString(value)
+}
+
+// The Filters type represents a set of filters to install on the system. It
+// exists purely for the convenience of being able to stick the Teardown method
+// somewhere where it makes sense.
+type Filters []*Filter
```
technoweenie: > It exists purely for the convenience of being able to stick the Teardown method somewhere where it makes sense.

I think it makes sense to just have functions like `SetupFilters([]*Filter)` and `TeardownFilters([]*Filter)`, too. I'm not sure how much is gained by declaring a `Filters` type.

commands/command_uninit.go
```diff
@@ -20,7 +22,7 @@ var (
 )
 
 func uninitCommand(cmd *cobra.Command, args []string) {
-	if err := lfs.UninstallFilters(); err != nil {
+	if err := lfs.TeardownFilters(); err != nil {
 		Error(err.Error())
```
technoweenie: Why the name change from install/uninstall to setup/teardown? The `Hook` type still uses the install/uninstall/upgrade terms.

lfs/setup.go
```diff
@@ -1,160 +1,80 @@
...
+	// XXX(@ttaylorr) not sure if this makes sense as a filter? Perhaps a
+	// Settable or Attribute type may be more appropriate.
 
-	prePushHookPath := filepath.Join(LocalGitDir, "hooks", "pre-push")
-	file, err := os.Open(prePushHookPath)
-	if err != nil {
-		// hook doesn't exist, our work here is done
-		return nil
-	}
+	requireFilters = &Filter{Name: "required", Value: "true"}
```
technoweenie: I wonder if we're thinking about filters incorrectly? Instead of thinking of it as smudge/clean/required filters, think of it as the `lfs` filter, with values for smudge/clean/required. Check out the docs:

> If a filter must succeed in order to make the stored contents usable, you can declare that the filter is required, in the configuration:
>
>```
[filter "crypt"]
	clean = openssl enc ...
	smudge = openssl enc -d ...
	required
>```

http://git-scm.com/docs/gitattributes

technoweenie:
I like where this is going. Though I think filters are represented incorrectly, as I pointed out above. Not that the current code represents them more accurately though.

> Eventually it seems appropriate to move both new types into the git package

It may even be worth extracting this to a separate git extension. But I'd want to see it in action first (and ship git lfs 1.0) before even thinking about that.

technoweenie:
Ah, the test failure is coming from [here](https://github.com/github/git-lfs/blob/e8a21619a99edb2bdb407cc9d7cf649619869df7/test/testhelpers.sh#L181-L185).  Weird that it passes locally but not on travis.

I tweaked the test helpers to actually print the git config instead of just listing it:

```
$ gd
diff --git a/test/testhelpers.sh b/test/testhelpers.sh
index 20f6d8c..12a905d 100644
--- a/test/testhelpers.sh
+++ b/test/testhelpers.sh
@@ -180,7 +180,7 @@ setup() {
   git config --global user.email "git-lfs@example.com"
   grep "git-lfs clean" "$REMOTEDIR/home/.gitconfig" > /dev/null || {
     echo "global git config should be set in $REMOTEDIR/home"
-    ls -al "$REMOTEDIR/home"
+    cat "$REMOTEDIR/home/.gitconfig"
     exit 1
   }
 ```

NOW we can see the actual error. Here's what I get on my laptop:

```
$ script/cibuild
ok  	github.com/github/git-lfs/lfs	0.327s
Git LFS: /Users/rick/github/git-lfs/bin/git-lfs
git-lfs/0.5.3 (GitHub; darwin amd64; go 1.3; git 04cab98)
git version 2.3.0
Updated pre-push hook.
Git LFS initialized.
global git config should be set in /Users/rick/github/git-lfs/test/remote/home
[filter "lfs"]
	clean = git lfs clean %%f
	smudge = git lfs smudge %%f
	required = true
[credential]
	helper = lfstest
[user]
	name = Git LFS Tests
	email = git-lfs@example.com
```

Git LFS uses `git-lfs clean` specifically to remove one extra os exec in the hooks. This was changed in #377.

File: lfs/setup.go (hunk 3 above)
Comment (ttaylorr) [inline lfs/setup.go]: Thanks for pointing this out -- this makes a lot more sense conceptually.  I actually had a note on my local copy to come back to this, glad you reminded me!

**Comment by ttaylorr** [comment]
:metal: on for that quick change in the tests, I'll get right on changing that up!

**Comment by ttaylorr** [comment]
Looks like we're still red -- which is odd since I can't reproduce the failure locally.  I'll do some more digging tomorrow to get this PR green, but at a later date I'd like to figure out why the tests have environment dependent output for me.

**Comment by technoweenie** [comment]
I definitely see the failures. We should talk more about why _you_ can't in gitter :)

The bash test suite is a custom thing that GitHub wrote. Very different from go or ruby tests. The output is a lot to take in because it echoes the commands from the test until a bash statement fails. Here's one from my own `script/cibuild` run:

```
test: uninit outside repository ...                                FAILED
-- stdout --
-- stderr --
    + set -e
    ++ git config filter.lfs.smudge
    + '[' 'git-lfs smudge %f' = 'git-lfs smudge %%f' ']'
    + test_status=1
```

That's failing on [this first statement](https://github.com/github/git-lfs/blob/3f1eec26a25b0c8e17f8b89c3d3dd419189d80c5/test/test-uninit.sh#L9) from the test.

**Comment by ttaylorr** [comment]
Finally green! Sorry about going red while this PR was open, from my perspective it was green the whole time!  I'll do some digging to figure out why I can't run tests locally.

**Comment by ttaylorr** [comment]
Rebased off of master, so we're merge-able again :smile:

File: lfs/hooks.go [hunk 4]
```diff
@@ -0,0 +1,148 @@
...
+	PostCheckoutHook                  = "post-checkout"
+	PostMergeHook                     = "post-merge"
+	PrePushHook                       = "pre-push"
+	PreReceiveHook                    = "pre-receive"
+	UpdateHook                        = "update"
+	PostReceiveHook                   = "post-receive"
+	PostUpdateHook                    = "post-update"
+	PushToCheckoutHook                = "push-to-checkout"
+	PostRewriteHook                   = "post-rewrite"
+	RebaseHook                        = "rebase"
```
technoweenie: What value do we get from defining all of these hooks? You can turn any string into a `HookType` if you really want. I'd just make Hook's Type property a string and be done with it.

FWIW: this is similar to my own struggle with the `Config.Access()` [`AuthType` constants](https://github.com/github/git-lfs/blob/3226d2a652799ebf81026c1ed14d71b783239436/lfs/config.go#L17-L22). The only benefit the code gets is being able to compare against a constant instead of initializing the string. I'm not even sure that matters for a short lived command line tool.

So after being confronted with this again, I think I'm leaning the other way towards just using strings.

lfs/attribute.go
```diff
@@ -0,0 +1,88 @@
...
+// "attribute", as defined here: http://git-scm.com/docs/gitattributes.
+type Attribute struct {
+	// The Path of an Attribute refers to the path at which all properties
+	// are relative to. For example, for a Path with the value "core", Git
+	// will produce something like:
+	//
+	// [core]
+	//	autocrlf = true
+	//	...
+	Path string
```
technoweenie: `Path` makes me think of the file system. What about "section"?

PR #624, opened 2015-09-01
batch API changes for v0.6.0

technoweenie:
This makes two changes for a Git LFS v0.6.0 release:

1. It now sets `lfs.{endpoint}.access` to "basic", not "private". This is an attempt to make the config forward compatible with possible alternate authentication methods, such as NTLM #617.
2. It now tries the batch API by default, setting `lfs.batch` to "false" if it has to fall back to the legacy API.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -122,7 +122,7 @@ func (c *gitConfig) Find(val string) string {
 
 // SetGlobal sets the git config value for the key in the global config
 func (c *gitConfig) SetGlobal(key, val string) {
-	simpleExec("git", "config", "--global", "--add", key, val)
+	simpleExec("git", "config", "--global", key, val)
 }
 
 // UnsetGlobal removes the git config value for the key from the global config
@@ -136,7 +136,7 @@ func (c *gitConfig) UnsetGlobalSection(key string) {
 
 // SetLocal sets the git config value for the key in the specified config file
 func (c *gitConfig) SetLocal(file, key, val string) {
-	simpleExec("git", "config", "--file", file, "--add", key, val)
+	simpleExec("git", "config", "--file", file, key, val)
 }
 
 // UnsetLocalKey removes the git config value for the key from the specified config file
--- a/lfs/config.go
+++ b/lfs/config.go
@@ -133,11 +133,10 @@ func (c *Configuration) BatchTransfer() bool {
 		}
 
 		// Any numeric value except 0 is considered true
-		if n, err := strconv.Atoi(v); err == nil && n != 0 {
-			return true
-		}
+		n, _ := strconv.Atoi(v)
+		return n != 0
 	}
-	return false
+	return true
 }
 
 // PrivateAccess will retrieve the access value and return true if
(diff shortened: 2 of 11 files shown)


git/git.go
```diff
@@ -122,7 +122,7 @@ func (c *gitConfig) Find(val string) string {
 
 // SetGlobal sets the git config value for the key in the global config
 func (c *gitConfig) SetGlobal(key, val string) {
-	simpleExec("git", "config", "--global", "--add", key, val)
+	simpleExec("git", "config", "--global", key, val)
 }
```
technoweenie: The `--add` flag will not overwrite existing values. It just adds another entry with the same key.

ttaylorr:
Looks cool to me :+1: The config attributes will be a good candidate for #619

rubyist:
:+1:

lfs/config.go
```diff
@@ -134,11 +141,10 @@ func (c *Configuration) BatchTransfer() bool {
 		}
 
 		// Any numeric value except 0 is considered true
-		if n, err := strconv.Atoi(v); err == nil && n != 0 {
-			return true
-		}
+		n, _ := strconv.Atoi(v)
+		return n != 0
```
sinbad: You need to handle the "false" case here (to replace the "true" case above), only lfs.batch=0 will work to disable otherwise.

sinbad:
You just need to update the lfs.{remote}.access section of git-lfs-config.5.ronn to match the private->basic change. I don't know if you want to add the ls.batch setting too, I left it out originally thinking it would go away.

Otherwise :+1:

lfs/config.go
technoweenie: I updated the tests to use a matrix so it was easier to see what values were being tested. "false" wasn't specifically tested, but it works. `strconv.Atoi()` returns `0` even if there's an error parsing the string to a number, so it still passes.

technoweenie:
I updated the notes about the `lfs.{remote}.access` setting. I also updated the `lfs.batch` one too, since it was already there.

lfs/config.go
sinbad: Hrm, I think this isn't guaranteed - the convention is that in the case of functions returning a non-nil error, you shouldn't rely on the values of *any* of the other return values. Neither Atoi nor ParseInt make the guarantee that the return will be 0 if err != nil, it *probably* will be since that's the default value but it's not specified AFAICT.

technoweenie: That's true. Probably not something that'll change before Go 2.0 though. Could change it to:

```go
n, err := strconv.Atoi(v)
return n != 0 && err == nil
```

sinbad: I'd just invert the whole thing (including removing the "true" test part) & return true unless you see "false" or successfully parse a zero. You're right that it will probably continue to work, I just don't really like relying on undocumented / non-guaranteed results.

rubyist: It is documented, but it's documented under ParseInt, which Atoi calls. However, the value depends on the type of error. If it can't parse it as a number it returns 0, but if the number is too big to be represented it'll be the max number. Due to that, I agree with the inversion.

technoweenie: What about https://github.com/github/git-lfs/pull/628, then?

Issue #625, opened 2015-09-02
Integration tests randomly fail with Go 1.5 (batch related)


sinbad:
I've been getting random failures in any one of a handful of batch-related integration tests recently and I've realised that they only seem to happen when using Go 1.5. I can't reproduce them when switching back to 1.4.

Adopting Go 1.5 is important not only for the vast performance improvements in the GC but because 1.4 has a bug which prevents using LFS API URLs which include a username (https://github.com/golang/go/issues/11399). 

The errors can occur even when you're using GIT_LFS_TEST_MAXPROCS=1 so it's not a race condition between tests. It is however always in one of the batch tests (not the same one, and not always in the same command), the common response is:

```
1 of 1 files) 11 B / 11 B                                                     Error uploading file common1.dat (2bc27d3512c2b3ab35b75244a7c5a7508610096aed67d64d65b11b5574b897d9)

    Errors logged to /Users/steve/projects/Go/src/github.com/github/git-lfs/tmp/test-pre-push.sh-63176/missing-but-on-server-batch/.git/lfs/objects/logs/20150902T101809.825928472.log
```

Repeating just the single test doesn't reproduce it. Since GIT_LFS_TEST_MAXPROCS=1 still does, I suspect it might be some some leftover state issue on the test server in batch mode, although I don't know why Go 1.5 would trigger it.

sinbad:
I'll be looking into this further myself when I get the chance, just recording it here in case anyone else starts to see it.

technoweenie:
I noticed this yesterday too when I setup a travis build with go 1.5: https://travis-ci.org/github/git-lfs/builds/78316038

sinbad:
Current working theory is that it's because Go 1.5 is now genuinely using multiple cores all the time and not just using co-operative multitasking on one core when hitting syscalls etc. I've found one specific instance I think is dodgy, in TransferQueue.transferWorker:

```
func (q *TransferQueue) transferWorker() {
	for transfer := range q.transferc {
		cb := func(total, read int64, current int) error {
			q.meter.TransferBytes(q.transferKind, transfer.Name(), read, total, current)
			return nil
		}

		if err := transfer.Transfer(cb); err != nil {
			q.errorc <- err
		} else {
			oid := transfer.Oid()
			for _, c := range q.watchers {
				c <- oid // <------- A. RECEIVER DOESN'T NECESSARILY HAPPEN BEFORE B.
			}
		}

		q.meter.FinishTransfer(transfer.Name())

[...]

PR #627, opened 2015-09-02
Fix Go 1.5 integration test issues

sinbad:
The test for lack of batch support turned it off globally on the test server then turned it back on, but since the server is shared between parallel tests, this introduces a race condition which can break other batch tests randomly. Instead, use a specific repo name to flag it.

Found this while investigating #625 although this isn't the complete fix. [Edit]Now it is.

### DIFF
--- a/commands/command_pre_push.go
+++ b/commands/command_pre_push.go
@@ -151,12 +151,16 @@ func prePushCheckForMissingObjects(pointers []*lfs.WrappedPointer) (objectsOnSer
 	}
 	// this channel is filled with oids for which Check() succeeded & Transfer() was called
 	transferc := checkQueue.Watch()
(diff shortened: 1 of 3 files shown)


technoweenie:
:+1: I like the idea of encoding special behavior for tests. I was thinking about something like that for API responses with various objects too.

sinbad:
I've added a second fix to this PR which resolves the remaining issue with batch tests in Go 1.5, there was a specific issue with expecting q.Wait() to return after all the transfer channel results had been received. Together these 2 seem to solve all the integration test issues in Go 1.5 in my tests.

rubyist:
:+1:

ttaylorr:
:+1:

PR #628, opened 2015-09-02
Config bool checks

technoweenie:
Adds a more explicit bool parser based on `strconv.ParseBool()`. May want to write to stderr if a config value is invalid (such as `42` or `-1`).

### DIFF
--- a/lfs/config.go
+++ b/lfs/config.go
@@ -100,7 +100,7 @@ func (c *Configuration) GetenvBool(key string, def bool) bool {
 		return def
 	}
 
-	b, err := strconv.ParseBool(s)
+	b, err := parseConfigBool(s)
 	if err != nil {
 		return def
 	}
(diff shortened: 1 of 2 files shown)


rubyist:
:+1: cleaner

ttaylorr:
:+1: much

technoweenie:
@sinbad: this affects your fetch config stuff too. `FetchRecentAlways` and `FetchRecentRefsIncludeRemotes` now accept all the bool string values that all git config bools do. There is one behavior change though:

```diff
-			// Any numeric value except 0 is considered true		
-			if n, err := strconv.Atoi(v); err == nil && n == 0 {		
-				c.fetchPruneConfig.FetchRecentRefsIncludeRemotes = false
```

Any numeric values besides `1`/`0` are flat out ignored. I'm not positive how `git config` treats these values though, so I should probably check that out :)

sinbad:
Love it :+1:

PR #629, opened 2015-09-02
Access types

technoweenie:
This removes the `AuthType` type in favor of strings. Also adds some unit tests. The only behavior change is that `Config.Access()` returns "basic" if `lfs.{url}.access` is still using the legacy "private" value.

### DIFF
--- a/commands/command_update.go
+++ b/commands/command_update.go
@@ -1,6 +1,9 @@
 package commands
 
 import (
+	"regexp"
+
+	"github.com/github/git-lfs/git"
 	"github.com/github/git-lfs/lfs"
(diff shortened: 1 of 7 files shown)


technoweenie:
This also teaches `git lfs update` how to fix the old private access settings.

sinbad:
I quite liked the string-backed enums, why remove them? Are you expecting it to be extensible later?

technoweenie:
I just wasn't seeing any benefit to them in this case. And yes, Git LFS may get new auth types: #617

rubyist:
:+1:

ttaylorr:
Looks cool to me :+1:

PR #630, opened 2015-09-02
test with multiple go versions

technoweenie:


technoweenie:
Merging, since the go 1.5 issues were fixed in #627.

PR #631, opened 2015-09-03
Fix worktree test now that batch is enabled by default

sinbad:
Travis didn't spot this because it's using a git version < 2.5.0


sinbad:
Travis is failing for unrelated reasons, "batch error handling" test fails maybe 1 in 5 times locally too (and on master)

ttaylorr:
:+1:

PR #633, opened 2015-09-04
Fetch --all

sinbad:
Support for `git lfs fetch --all` for backup/migration purposes.

### DIFF
--- a/lfs/progress_meter.go
+++ b/lfs/progress_meter.go
@@ -2,8 +2,10 @@ package lfs
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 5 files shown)


sinbad:
Once again CI fail is the intermittent 'batch error handling' test again, unrelated.

sinbad:
Rebased and lights are green again.

ttaylorr:
Looks cool to me :+1:

lfs/progress_meter.go
```diff
@@ -212,3 +214,58 @@ func formatBytes(i int64) string {
...
+func (s *Spinner) Finish(out io.Writer, finishMsg string) {
+	s.msg = finishMsg
+	s.stage = 0
+	var sym string
+	if runtime.GOOS == "windows" {
+		// Windows console sucks, can't do nice check mark except in ConEmu (not cmd or git bash)
+		// So play it safe & boring
+		sym = "Done"
+	} else {
+		sym = fmt.Sprintf("%c", '\u2714')
```
technoweenie: Looks like this will output "Done 3 objects found". Here's what I see:

```
$ git lfs fetch --all    
✔ 3 objects found 
```

Something like `*` might be just fine.

sinbad: It’ll only use ’Done’ on Windows because Windows console sucks. I like the ’spin then checkmark when done’ style for tasks but if you prefer `*` I can change it.

technoweenie: I understand why you can't use the checkmark. I just want it to return a readable sentence.  I'd prefer `*`, or even `[ok]` or `[done]`.

sinbad: Gotcha, done. Or rather `*` :)

sinbad:
I rebased but now have another Travis fail with a smudge test, doesn't happen locally on Go 1.5.

sinbad:
Second time's a charm.

PR #635, opened 2015-09-04
Fix pre-push hook with multiple branches

### LINKED ISSUE #632: Pre-push hook doesn't cope with pushing multiple branches at once
If you call `git push` with multiple branches, the pre-push hook only pushes objects for the first branch.
Example:
```
git push origin branch1 branch2 branch3 branch4
```
...will only push objects referenced in branch1.

sinbad:
Fixes #632

### DIFF
--- a/commands/command_pre_push.go
+++ b/commands/command_pre_push.go
@@ -62,7 +62,7 @@ func prePushCommand(cmd *cobra.Command, args []string) {
 
 		left, right := decodeRefs(line)
 		if left == prePushDeleteBranch {
-			return
+			continue
 		}
 
 		// Just use scanner here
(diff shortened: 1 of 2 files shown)


technoweenie:
This looks great, but that's a lot of nesting. I'd have broken it up into smaller functions, but I don't think it's necessary here.

sinbad:
It was only 1 extra level so I didn't really think about it, but I guess the 'for' is quite long now. I'll separate it and also rebase so I get the transfer queue fix.

sinbad:
Dammit, bitten by the init random fail after rebase Xo It's fine, honest ;)

ttaylorr:
Nice :+1:

sinbad:
Lights are green :)

PR #637, opened 2015-09-04
Resolve race condition in transfer queue error collection

rubyist:
This was causing the intermittent batch error test failures under go 1.5. It was possible that the queue's `Wait()` function returned before all of the errors had been collected. This PR ensures error collection has finished.


technoweenie:
:metal:

sinbad:
You rock!

PR #638, opened 2015-09-04
test for bad clean or smudge attribute

technoweenie:
I think it's coming from [here](https://github.com/github/git-lfs/blob/10dbfb5da854b3a4a1aefbb073e4aef7d0f4e898/lfs/attribute.go#L32-L47). Sometimes it returns the smudge error before the clean error. This might be a slight regression from #619. I'll have to see if the older code returned errors for every attribute and not just the first one.

### DIFF
--- a/test/test-init.sh
+++ b/test/test-init.sh
@@ -30,7 +30,8 @@ begin_test "init with old settings"
 
   [ "$res" = 2 ]
 
-  grep "clean attribute should be" init.log
+  cat init.log
+  grep -E "(clean|smudge) attribute should be" init.log
   [ `grep -c "(MISSING)" init.log` = "0" ]
 
   [ "git lfs smudge %f" = "$(git config filter.lfs.smudge)" ]


ttaylorr:
Argh, this bit me at work once.  From the Golang spec:

> The iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next. If map entries that have not yet been reached are removed during iteration, the corresponding iteration values will not be produced. If map entries are created during iteration, that entry may be produced during the iteration or may be skipped. The choice may vary for each entry created and from one iteration to the next. If the map is nil, the number of iterations is 0.

Your fix looks good to me, but it looks like you dropped your Git credentials somewhere along the way.

test/test-init.sh
```diff
@@ -30,7 +30,8 @@ begin_test "init with old settings"
 
   [ "$res" = 2 ]
 
-  grep "clean attribute should be" init.log
+  cat init.log
```
ttaylorr: Do we need to `cat` this out?

technoweenie: It only shows up if the test fails. It's much easier to diagnose with this printed out.

technoweenie:
> it looks like you dropped your Git credentials somewhere along the way

Ah yeah, I've been messing with my git config while QAing with various git config settings. I've started using a docker image for this stuff instead. Thanks for the reminder to clean up my global git config :)

test/test-init.sh
ttaylorr: Works for me :+1:

PR #642, opened 2015-09-06
basic nsis script

technoweenie:
This attempts to detect Git in common dirs like:

* `C:\Program Files\Git\mingw64\bin` (Git For Windows)
* `C:\Program Files\Git\bin`
* `C:\Program Files (x86)\Git\bin`

We should try to support every common way to install Git on Windows. Not sure what to do about native tools though. Ideally, they'd just bundle Git LFS. Until then, we can add them to the list.

I also didn't try to run "git lfs init" for the user. It seems like Git For Windows and GitHub Desktop include their own "Git Bash" commands with the proper env. Users should run the init command there.

/cc @ssgelm: you expressed interest in a Windows installer.


script/nsis/installer.nsi
```diff
@@ -0,0 +1,53 @@
...
+VIAddVersionKey ProductName      "Git LFS"
+VIAddVersionKey Comments         "Git LFS tool"
+VIAddVersionKey CompanyName      "GitHub, Inc"
+VIAddVersionKey LegalCopyright   "GitHub, Inc"
+VIAddVersionKey FileDescription  "Git LFS Tool"
+VIAddVersionKey FileVersion      1
+VIAddVersionKey ProductVersion   1
+VIAddVersionKey InternalName     "Git LFS"
+VIAddVersionKey LegalTrademarks  "something, something, darkside"
+VIAddVersionKey OriginalFilename "GitLFSInstaller.exe"
```
technoweenie: These are probably ALL WRONG :) Need to figure out how to sign this properly, so Windows can stop saying it's coming from an "unknown publisher".

shiftkey: Probably need to look at [Authenticode signing](https://msdn.microsoft.com/en-us/library/bb756995.aspx) of some sort...

sinbad: Yes, I highly recommend EV code signing, Windows 8, 8.1 & 10 all have SmartScreen enabled by default and non-codesigned downloads are usually flagged. http://blogs.msdn.com/b/ie/archive/2012/08/14/microsoft-smartscreen-amp-extended-validation-ev-code-signing-certificates.aspx

In my experience even code signed apps can be flagged as untrusted to begin with until enough people have downloaded them if the cert is new, so if you have a cert you already use for say GitHub Desktop, definitely try to re-use it.

script/nsis/installer.nsi
```diff
@@ -0,0 +1,53 @@
...
+
+Function .onGUIInit
+  IfFileExists $PROGRAMFILES64\Git\mingw64\bin\git.exe 0 +3
```
sinbad: You might want to try running 'where git' first, if people have opted to update their path outside of Git Bash too this will detect the location.

script/nsis/installer.nsi
```diff
@@ -0,0 +1,53 @@
...
+  IfFileExists $PROGRAMFILES64\Git\bin\git.exe 0 +3
+    StrCpy $INSTDIR "$PROGRAMFILES64\Git\bin"
+    Goto exit
+
+  IfFileExists $PROGRAMFILES\Git\bin\git.exe 0 exit
+    StrCpy $INSTDIR "$PROGRAMFILES\Git\bin\git.exe"
+
+  exit:
+    SetOutPath $INSTDIR
+FunctionEnd
```
sinbad: Recommend a stage in the installer which confirms this path & explains that it's going to install LFS in it, option to browse if they want.

technoweenie:
Here are some screenshots. I messed with the "modern UI 2" plugin, but it seemed to require a components page. So I opted to go with the shorter script, even if the installer looks old school.

windows_10_x64_base 2

windows_10_x64_base

script/nsis/installer.nsi
technoweenie: Just posted a screenshot of that page.

script/nsis/installer.nsi
technoweenie: I ran that on my VM snapshot with just Git For Windows in the Git Cmd, and it returned `$PROGRAMFILES64\Git\cmd\git.exe`. May have to install this to multiple directories so it's available everywhere the user can open a "git" prompt.

sinbad: I believe that's a wrapper which just starts up Git Bash and runs the real git.exe from the bin dir, so it'll probably be fine to just pull out ..\bin from directories which end in cmd.

technoweenie: You're right. `git lfs` works just fine. However, due to #377, we need `git-lfs` to work too.

script/nsis/installer.nsi
sinbad: Ah thanks, I guess that's enough to explain things. Do you know what happens if someone uninstalls Git without uninstalling LFS first? Possible it might nuke the dir & remove the uninstaller, hopefully if it does LFS can still clean itself out of installed programs.

script/nsis/installer.nsi
sinbad: Hmm - we could just say that if you want that optimisation you have to be in Git Bash.

script/nsis/installer.nsi
technoweenie: If the directory is wiped out, I don't know if there's anything for the LFS uninstaller to do. I don't know if I can run `git lfs init` or `git lfs uninit` from nsis with the proper git environment, so the installer just deletes `git-lfs.exe`.

script/nsis/installer.nsi
technoweenie: Seems easier to just copy that everywhere. Changing the optimization requires changing the global git config. The current version of the nsi script does this.

technoweenie:
Okay, I'm happy enough with this to try it out with v0.6.0.

script/nsis/installer.nsi
```diff
@@ -0,0 +1,66 @@
...
+  IfFileExists $PROGRAMFILES64\Git\mingw64\bin\git.exe 0 +3
+    StrCpy $INSTDIR "$PROGRAMFILES64\Git\mingw64\bin"
+    Goto exit
+
+  IfFileExists $PROGRAMFILES64\Git\bin\git.exe 0 +3
+    StrCpy $INSTDIR "$PROGRAMFILES64\Git\bin"
+    Goto exit
+
+  IfFileExists $PROGRAMFILES\Git\bin\git.exe 0 +2
+    StrCpy $INSTDIR "$PROGRAMFILES\Git\bin\git.exe"
```
sinbad: You need to remove the '\git.exe', when you run this against a 32-bit install the "Install" button is disabled because it's pointing at an existing file instead of a folder.

sinbad:
I think this should have a build script, I had to manually copy git-lfs.exe to script/nsis in order to build the installer.

It's a shame NSIS doesn't support HiDPI displays, I haven't used it for a few years, and not since getting a retina Mac. Eww, pixels ;) Not a big issue though obviously.

technoweenie:
> I think this should have a build script, I had to manually copy git-lfs.exe to script/nsis in order to build the installer.

Doh!

> It's a shame NSIS doesn't support HiDPI displays

Well, there are a couple modern UI plugins. I tried "MUI2", but it forced me to use a components. If they don't support HiDPI displays, I can look into WiX. I went with nsis because I thought building the packages with the unix `makensis` command would be useful. Turns out I still need Windows to sign it :/

sinbad:
> Turns out I still need Windows to sign it :/

Ah crap, forgot about that. Yeah, WiX is quite usable for simple things, it just gets a bit messy the more complex you go but git-lfs doesn't need much.

technoweenie:
> Yeah, WiX is quite usable for simple things, it just gets a bit messy the more complex you go

Judging by these nsis tutorials, it's not much different. If I knew the "scripting language" used labels and gotos, I might've gone back to xml happily :)

sinbad:
> Judging by these nsis tutorials, it's not much different. If I knew the "scripting language" used labels and gotos, I might've gone back to xml happily :)

Haha, yeah it's super-basic, more like a batch file than a proper language. NSIS is a little more direct than WiX which is more declarative; it's more powerful but sometimes requires more 2nd-derivative thinking :)

PR #644, opened 2015-09-08
In repo check

technoweenie:
This prevents 7 commands from blowing up because they're not run from inside of a repository. It also makes a couple slight tweaks to the integration test suite:

1. The verbose logs at the end of a failure (server logs and your local env) are ONLY shown from `script/cibuild`. We care about this output on CI servers, but probably not on our local machines.
2. The trace file is truncated after every individual test. This way a test failure will show just its own git trace log, not the trace log for every test that has run in the file.

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -21,6 +21,11 @@ var (
 )
 
 func checkoutCommand(cmd *cobra.Command, args []string) {
+	if !lfs.InRepo() {
+		Print("Not in a git repository.")
+		os.Exit(128)
+	}
+
 	// Parameters are filters
 	// firstly convert any pathspecs to the root of the repo, in case this is being executed in a sub-folder
 	var rootedpaths []string
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -21,10 +21,7 @@ var (
 )
 
 func checkoutCommand(cmd *cobra.Command, args []string) {
-	if !lfs.InRepo() {
-		Print("Not in a git repository.")
-		os.Exit(128)
-	}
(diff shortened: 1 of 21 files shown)


commands/command_checkout.go
```diff
@@ -21,6 +21,11 @@ var (
 )
 
 func checkoutCommand(cmd *cobra.Command, args []string) {
+	if !lfs.InRepo() {
```
ttaylorr: Perhaps we could throw this into a `func` in the `commands` package?

```go
package commands

func AssertInRepo() {
        if !lfs.InRepo() {
                Print("Not in a git repository.")
                os.Exit(128)
        }
}
```

ttaylorr:
My only comment concerns moving the whole "check and exit if not in repo" business into a func and reusing that code in several of the commands.  Looks like some tests are failing as well, which I'd like to look at.

technoweenie:
I don't get the failures on my mac. I'll try in a docker image in a bit. Seems to be a difference in the exit code that's returned.

technoweenie:
Okay, I figured the problem out.

```
   trace: built-in: git 'config' '-l'
   trace git-lfs: run_command: 'git' config -l -f /home/travis/gopath/src/github.com/github/git-lfs/.git/config
   trace: built-in: git 'config' '-l' '-f' '/home/travis/gopath/src/github.com/github/git-lfs/.git/config'
```

It's running from inside the github/git-lfs directory, so it's not crashing. The tests pass for me because I set `GIT_LFS_TEST_DIR=/Users/rick/p/git-lfs-temp` for my tests.

PR #646, opened 2015-09-08
Push all to remote

technoweenie:
This uses the `lfs.ScanRefsMode` scan mode when pushing if the `--all` flag is given. This lets you confirm that everything referenced in the given ref has been pushed to the given remote.

### DIFF
--- a/commands/command_push.go
+++ b/commands/command_push.go
@@ -1,7 +1,6 @@
 package commands
 
 import (
-	"fmt"
 	"io/ioutil"
 	"os"
 
@@ -43,23 +42,7 @@ func uploadsBetweenRefAndRemote(remote string, refs []string) *lfs.TransferQueue
 
 	if pushAll {
 		if len(refs) == 0 {
-			// no ref given as an arg, so scan all refs
-			opts := &lfs.ScanRefsOptions{ScanMode: lfs.ScanAllMode, SkipDeletedBlobs: false}
-			// This could be a long process so use the chan version & report progress
-			Print("Scanning for all objects ever referenced...")
-			spinner := lfs.NewSpinner()
-			var numObjs int64
-			pointerchan, err := lfs.ScanRefsToChan("", "", opts)
-			if err != nil {
-				Panic(err, "Could not scan for Git LFS files")
-			}
-			pointers := make([]*lfs.WrappedPointer, 0)
-			for p := range pointerchan {
-				numObjs++
-				spinner.Print(OutputWriter, fmt.Sprintf("%d objects found", numObjs))
-				pointers = append(pointers, p)
-			}
-			spinner.Finish(OutputWriter, fmt.Sprintf("%d objects found", numObjs))
+			pointers := scanAll()
 			Print("Pushing objects...")
 			return uploadPointers(pointers)
 		} else {
--- a/commands/command_push.go
+++ b/commands/command_push.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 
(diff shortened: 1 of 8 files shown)


commands/command_push.go
```diff
@@ -19,6 +19,7 @@ var (
 	pushDryRun       = false
 	pushDeleteBranch = "(delete)"
 	pushObjectIDs    = false
+	pushAllToRemote  = false
```
ttaylorr: Do you think the `ToRemote` part of this identifier could be dropped?

ttaylorr:
Other than my one comment, looks good to me :+1:

sinbad:
You need to update the man page too. :)

I like the concept, but I think its interpretation of `--all` is potentially misleading. It's "all at one ref" not "all binaries ever", which is not symmetrical with `git lfs fetch --all` which downloads all referenced binaries ever (for backup). I'm concerned that people will think they can perform a complete migration of their LFS files using `fetch --all` then `push --all` and they'd be done - but they won't be because as currently implemented it would actually implicitly mean `push --all HEAD` only.

I think `git lfs push --all` with no ref parameter should mean `push all binaries ever` i.e. using ScanAllMode. I think the behaviour in this PR i.e. ScanRefsMode should be reserved for *either* a different option (`--all-ref`?) in which case ref args should be disallowed, or alternatively ScanRefsMode should only be used when a ref is given *explicitly* on the command line, not implied HEAD.

sinbad:
Also the non-migratory version of `--all` might need a step like `prePushCheckForMissingObjects` to cope with cases where the person pushing doesn't have the entire LFS state.

commands/command_push.go
technoweenie: I wanted to distinguish this from `fetch --all`, which downloads EVERYTHING. Though `pushAllForRef` would be a more accurate var name I think.

technoweenie:
I totally updated the man page, but forgot to push. Or I faked the commit times :trollface: 

`fetch` symmetry is a better idea. :+1:

```bash
# PUSH EVERYTHING
$ git lfs push --all origin

# just master branch
$ git lfs push --all origin master

# multiple refs
$ git lfs push --all origin master feature/1 v1.0
```

I'll add a test case with missing objects too. Good catch.

sinbad:
> I totally updated the man page, but forgot to push. Or I faked the commit times :troll face:

I believe you ;)

> `# PUSH EVERYTHING`...

:+1:

technoweenie:
The behavior matches https://github.com/github/git-lfs/pull/646#issuecomment-138901295 now. Still a couple issues:

1. Probably want to use `lfs.ScanRefsToChan()` like `fetch --all` does. It has its own spinner while it's scanning the entire history.
2. `prePushCheckForMissingObjects()` outputs a separate progress output. 

        $ git lfs push --all origin master 
        (1 of 1 files) 24.65 KB / 24.65 KB                                                                                                                                                                                                                                                                                                                                         
        (0 of 3 files, 2 skipped) 0 B / 1.45 MB, 1.43 MB skipped

technoweenie:
Basically copied a bunch of spinner code from `fetchAll()` to get this:

```
$ git lfs push --all origin
✔ 3 objects found                                                                                                                                                                                                                                                                                                                                                           
Pushing objects...
(1 of 1 files) 24.65 KB / 24.65 KB                                                                                                                                                                                                                                                                                                                                         
(0 of 3 files, 2 skipped) 0 B / 1.45 MB, 1.43 MB skipped     
```

technoweenie:
I removed the first progress meter by setting `dryRun` to true.

```
$ git lfs push --all origin
✔ 3 objects found                                                                                                                                                                                                                                                                                                                                                           
Pushing objects...
(0 of 3 files, 3 skipped) 0 B / 1.45 MB, 1.45 MB skipped   
```

It still writes the progress log if `GIT_LFS_PROGRESS` is passed, though.

docs/man/git-lfs-push.1.ronn
```diff
@@ -3,25 +3,28 @@ git-lfs-push(1) -- Push queued large files to the Git LFS endpoint
...
+of the remote.
 
 ## OPTIONS
 
+* `--dry-run`:
+    Print the files that would be pushed, without actually pushing them.
+
+* `--all`:
+    This pushes all objects for the given ref or current branch, regardless of
+    whether they are referenced in the local clone of the remote.
```
sinbad: Should really mention the difference between providing a ref on the command line & not when `--all` is provided.

sinbad:
> I removed the first progress meter by setting dryRun to true.

Wow, that's entirely *not* what I would expect `dryRun` to do. I think that option should be renamed to `suppressProgress` instead to avoid confusion in future, because that's what it actually does - looks to me like `dryRun=true` does everything `dryRun=false` does apart from the progress output. I can imagine a time when someone fixes it to *actually* be a dry run option (and thus not do anything except local scanning) and for that to break this code.

Other than that and my man comment, looks good to me! :+1:

docs/man/git-lfs-push.1.ronn
technoweenie: "This pushes all objects to the remote that are referenced by any reachable commit. If no ref is provided, then all refs are pushed."

technoweenie:
I agree about `dryRun`, but I want to update that in a separate PR.

PR #648, opened 2015-09-09
Added file format migration example to README

i59:
Provided example for moving all use cases of a specific file format to the README, very useful if migrating existing repositories to LFS usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,8 @@ Moving to LFS does not have to be hard! The Windows batch script below can be us
     git commit -m "Removed %file_extension% format from the repository"
     :: Tell LFS to track this format, has to be done before the files are added
     git lfs track "*.%file_extension%"
+    :: Make sure to add the changes made by LFS above
+    git add .gitattributes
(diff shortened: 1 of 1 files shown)


technoweenie:
Hey, I really appreciate you writing this up. Any thoughts on moving this to the [wiki](https://github.com/github/git-lfs/wiki), instead? It seems like a better place for things like this. The steps are only applicable for Windows, and probably not every use case. The wiki would be a good place for you and other contributors 

I want to keep all technical docs (specifications, man pages, etc) in the repository, but not user tutorials. While the wiki is a good place to _start_, I can't say that'll be the final place for user tutorials.

i59:
Sure thing, here you go: https://github.com/github/git-lfs/wiki/Moving-existing-file-formats-to-LFS-on-Windows

technoweenie:
Thanks, closing this out, for now. I'll experiment with organized wiki pages, so you could go to a page like `Migration`, with links to tips like this. Then we'd link to `Migration` from the README.

PR #650, opened 2015-09-10
always consult PrivateAccess() for batch API calls

technoweenie:
While running my QA checks, I noticed that the auth type wasn't being set at all:

```
$ git lfs env
git-lfs/0.6.0 (GitHub; darwin amd64; go 1.5; git 19105ed)
git version 2.3.0

Endpoint=https://github.com/github/git-lfs.git/info/lfs (auth=none)
```

#611 introduced various constructors for different types of requests that the client makes:

* `doApiBatchRequest()` - LFS Batch API requests
* `doLegacyApiRequest()` - Legacy LFS API requests.
* `doAPIRequest()` - Common function for ALL LFS API requests.
* `doStorageRequest()` - Storage requests (these actually upload and download the objects).

`doAPIRequest()` requires credentials for all non-GET requests. But all Batch API requests are POST. This is a pretty small change, but highlights that the project needs better tests around this stuff.

### DIFF
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -357,7 +357,7 @@ func UploadObject(o *objectResource, cb CopyCallback) error {
 	req.Header.Set("Content-Length", strconv.Itoa(len(by)))
 	req.ContentLength = int64(len(by))
 	req.Body = ioutil.NopCloser(bytes.NewReader(by))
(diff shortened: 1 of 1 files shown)


sinbad:
Nice catch

PR #656, opened 2015-09-11
Make sure random test data is seeded differently between commits

sinbad:
You could still end up with duplicated seeds between commits which could cause unintended consequences.


technoweenie:
:+1: Seems legit.

PR #657, opened 2015-09-11
Remove git requirement to make custom installs work

andyneff:


technoweenie:
Just for additional context, we're removing this requirement to support users that installed git manually.

PR #665, opened 2015-09-14
Use man content for `help <cmd>' and '<cmd> --help'

sinbad:
Call 'go generate ./commands' to generate code from .ronn files which the cobra command can then use. Added to the build scripts but needs to be done manually at least once to test it outside that. 

I've done the minimum re-formatting for now. The man format is pretty damn ugly on the command line and I'd like to do better but didn't want to spend time on that until the principle was approved. For now it just fixes the backtick issue, other things we could do:

* Strip off the 2-line headers
* Strip 'SYNOPSIS' and pull out the command summary as the first line
* Strip '##' section markers generally
* Use an abbreviated form for `command --help` (just contents of SYNOPSIS, DESCRIPTION and OPTIONS) and only print the full text for `help command`. Cobra splits this as Usage/Help, currently it's the same for both.

If you think this is a good idea I can hack on it some more.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -222,22 +222,27 @@ func determineIncludeExcludePaths(includeArg, excludeArg string) (include, exclu
 	return includePaths, excludePaths
 }
 
-// help is used for 'git-lfs help <command>'
-func help(cmd *cobra.Command, args []string) {
-	if txt, ok := ManPages[cmd.Use]; ok {
+func printHelp(commandName string) {
+	if txt, ok := ManPages[commandName]; ok {
 		fmt.Fprintf(os.Stderr, txt)
 	} else {
-		fmt.Fprintf(os.Stderr, "Sorry, no help text found\n")
+		fmt.Fprintf(os.Stderr, "Sorry, no usage text found for %q\n", commandName)
 	}
+}
+
+// help is used for 'git-lfs help <command>'
+func help(cmd *cobra.Command, args []string) {
+	if len(args) == 0 {
+		printHelp("git-lfs")
+	} else {
+		printHelp(args[0])
+	}
+
 }
 
 // usage is used for 'git-lfs <command> --help' or wen invoked manually
 func usage(cmd *cobra.Command) error {
-	if txt, ok := ManPages[cmd.Use]; ok {
-		fmt.Fprintf(os.Stderr, txt)
-	} else {
-		fmt.Fprintf(os.Stderr, "Sorry, no usage text found\n")
-	}
+	printHelp(cmd.Name())
 	return nil
 }
 
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -16,6 +16,9 @@ import (
 	"github.com/github/git-lfs/vendor/_nuts/github.com/spf13/cobra"
 )
 
+// Populate man pages
+//go:generate go run ../docs/man/mangen.go
+
 var (
 	Debugging    = false
 	ErrorBuffer  = &bytes.Buffer{}
@@ -29,6 +32,7 @@ var (
 			cmd.Usage()
 		},
 	}
+	ManPages = make(map[string]string, 20)
 )
 
 // Error prints a formatted message to Stderr.  It also gets printed to the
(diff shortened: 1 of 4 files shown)


sinbad:
Oops, 'go generate' is go 1.4+ so the 1.3 tests failed. We're dropping those though I hear :)

script/build.go
```diff
@@ -38,6 +38,11 @@ func mainBuild() {
 
 	fmt.Printf("Using %s\n", runtime.Version())
 
+	genOut, err := exec.Command("go", "generate", "./commands").CombinedOutput()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "go generate failed:\n%v", genOut)
+		os.Exit(1)
```
technoweenie: Don't print byte arrays :)

```
$ script/bootstrap
Using go1.3
go generate failed:
[103 111 58 32 117 110 107 110 111 119 110 32 115 117 98 99 111 109 109 97 110 100 32 34 103 101 110 101 114 97 116 101 34 10 82 117 110 32 39 103 111 32 104 101 108 112 39 32 102 111 114 32 117 115 97 103 101 46 10]exit status 1
```

commands/commands.go
```diff
@@ -218,6 +222,28 @@ func determineIncludeExcludePaths(includeArg, excludeArg string) (include, exclu
 	return includePaths, excludePaths
 }
 
+// help is used for 'git-lfs help <command>'
+func help(cmd *cobra.Command, args []string) {
+	if txt, ok := ManPages[cmd.Use]; ok {
+		fmt.Fprintf(os.Stderr, txt)
+	} else {
+		fmt.Fprintf(os.Stderr, "Sorry, no help text found\n")
+	}
```
technoweenie: I don't think `cmd.Use` is what you want. I updated the error message like this:

```go
fmt.Fprintf(os.Stderr, "Sorry, no help text found for %q.\n", cmd.Use)
```

And got: "help [command]". I think `args[0]` is what you want.

technoweenie:
I really dig this idea :+1: It's like #568, but takes advantage of go's code generation (which I've been meaning to check out since it was announced).

Also, to answer your final reply on that thread:

> were you aware that the default behaviour in Windows for something like `git checkout --help` is that it launches the browser & displays the HTML version of the man page? 

Nope, that sounds awesome. Then the man page references at the bottom would actually link to each other :metal:

commands/commands.go
sinbad: Weird, it worked fine for me & matched up with the man content; 'Use' is pretty consistently set like this:
```
	pushCmd = &cobra.Command{
		Use:   "push",
```
But I guess arguably someone could make a mistake on setting those up so args would be better.

script/build.go
sinbad: Doh.

sinbad:
> I really dig this idea :+1: It's like #568

Gah, I was looking for that yesterday but couldn't find it, I must have been looking at the issues. I prefer the map approach & general help/usage functions too since it means it automatically works for new commands if you provide a man page, without having any compile errors.

About the HTML on Windows; that's basically the other option; we could instead have 2 different behaviours for Linux/Mac and Windows, the former calling 'man git-lfs-<command>' and the latter opening HTML which we have to deploy somewhere. My main problem with that approach is that even though it's nice in a pre-packaged release, it's not very usable in source builds, because the man/HTML might not be present (or at the least the paths would be different)

I'm tempted just to go with this for now (with suitable reformatting) & consider extending things later so that when built with the release flags the man/HTML route is used instead.

commands/commands.go
sinbad: Ah yeah actually you can't use `args[0]` because the 'usage' function which is called for `--help` doesn't include them. I'm using `cmd.Name()` instead which is the external name for `Use` and also dealing with an edge case on the root topic.

sinbad:
Already looking better, just want to tackle the links next:
```bash
steve@Stevens-MacBook-Pro ~/p/G/s/g/g/git-lfs> git lfs help push

git lfs push [options] <remote> [<ref>...]<br>
git lfs push <remote> [<ref>...]<br>
git lfs push --object-id <remote> [<oid>...]

Upload Git LFS files to the configured endpoint for the current Git remote.  By
default, it filters out objects that are already referenced by the local clone
of the remote.

Options:

* --dry-run:
    Print the files that would be pushed, without actually pushing them.

* --all:
    This pushes all objects to the remote that are referenced by any commit
    reachable from the refs provided as arguments. If no refs are provided, then
    all refs are pushed.

* --object-id:
    This pushes only the object OIDs listed at the end of the command, separated
    by spaces.

* --stdin:
    Read the remote and branch on stdin. This is used in conjunction with the
    pre-push hook and must be in the format used by the pre-push hook:
    <local-ref> <local-sha1> <remote-ref> <remote-sha1>. If --stdin is used
    the command line arguments are ignored.  NOTE: This is deprecated in favor
    of the pre-push command.
```

technoweenie:
> I'm tempted just to go with this for now (with suitable reformatting) & consider extending things later so that when built with the release flags the man/HTML route is used instead.

Baby steps. :+1:

> Already looking better, just want to tackle the links next

That looks much better than what I saw. The man page headers were annoying. I like the more compact format.

sinbad:
OK I think this is ready to be merged now. I've made the links look nicer:
```
Options:

* -I <paths> --include=<paths>:
  Specify lfs.fetchinclude just for this invocation; see "Include and exclude"

* -X <paths> --exclude=<paths>:
  Specify lfs.fetchexclude just for this invocation; see "Include and exclude"

* --recent:
  Download objects referenced by recent branches & commits in addition to those
  that would otherwise be downloaded. See "Recent changes"

* --all:
  Download all objects referenced by any commit that is reachable; this is
  primarily for backup / migration purposes. Cannot be combined with --recent or
  --include/--exclude. Ignores any globally configured include and exclude paths
  to ensure that all objects are downloaded.

Include and exclude
-------------------
```
It's not 100% foolproof because it only processes the man pages line-wise and not with a full lexer (links split across lines wouldn't work for example) but I think it does the job much better for now.

technoweenie:
> It's not 100% foolproof because it only processes the man pages line-wise and not with a full lexer (links split across lines wouldn't work for example) but I think it does the job much better for now.

I think any edge cases for the parser should be tested. I'm not sure how that would work yet though.

Issue #664, opened 2015-09-14
Running `git lfs ls-files` before adding any files causes index out of range error


curzona:
Running `git lfs ls-files` before adding any files causes following error:

```
$ git init
Initialized empty Git repository in /blah/blah
$ git lfs version
git-lfs/0.6.0 (GitHub; linux amd64; go 1.5.1; git 183bdc8)
$ git lfs ls-files
panic: runtime error: index out of range

goroutine 1 [running]:
github.com/github/git-lfs/git.ResolveRef(0x882d90, 0x4, 0x1, 0x0, 0x0)
	/Users/rick/go/src/github.com/github/git-lfs/git/git.go:68 +0x2f6
github.com/github/git-lfs/git.CurrentRef(0xacf3a0, 0x0, 0x0)
	/Users/rick/go/src/github.com/github/git-lfs/git/git.go:73 +0x38
github.com/github/git-lfs/commands.lsFilesCommand(0xaa55c0, 0xacf3a0, 0x0, 0x0)
	/Users/rick/go/src/github.com/github/git-lfs/commands/command_ls_files.go:26 +0x205
github.com/github/git-lfs/vendor/_nuts/github.com/spf13/cobra.(*Command).execute(0xaa55c0, 0xacf3a0, 0x0, 0x0, 0x0, 0x0)
	/Users/rick/go/src/github.com/github/git-lfs/vendor/_nuts/github.com/spf13/cobra/command.go:477 +0x403
github.com/github/git-lfs/vendor/_nuts/github.com/spf13/cobra.(*Command).Execute(0xaa6ae0, 0x0, 0x0)
	/Users/rick/go/src/github.com/github/git-lfs/vendor/_nuts/github.com/spf13/cobra/command.go:551 +0x46a
github.com/github/git-lfs/commands.Run()
	/Users/rick/go/src/github.com/github/git-lfs/commands/commands.go:85 +0x23
main.main()
	/Users/rick/go/src/github.com/github/git-lfs/git-lfs.go:9 +0x14
```

technoweenie:
Oh good call! Looks like that's failing because your repo has no commits, and therefore no current ref.

curzona:
@technoweenie Thanks Rick

PR #668, opened 2015-09-17
Fixed git lfs ls-files then index out of range error.

Aorjoa:
This PR for command `git lfs ls-files` and then error `index out of range` issue according #664

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -64,6 +64,10 @@ func ResolveRef(ref string) (*Ref, error) {
 		return nil, err
 	}
 	lines := strings.Split(outp, "\n")
+	if len(lines) <= 1 {
+		err := errors.New("git can't resolve ref : " + ref)
+		return nil, err
+	}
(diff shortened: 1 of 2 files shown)


git/git.go
```diff
@@ -64,6 +64,10 @@ func ResolveRef(ref string) (*Ref, error) {
 		return nil, err
 	}
 	lines := strings.Split(outp, "\n")
+	if len(lines) <= 1 {
+		err := errors.New("git can't resolve ref : " + ref)
```
technoweenie: I prefer using `fmt.Errorf` simple error messages that build dynamic strings:

```go
err := fmt.Errorf("git can't resolve ref: %q", ref)
```

This is nitpicky, and not worth waiting on you to fix. Just some Go feedback :)

technoweenie:
Hey, welcome back :)  Thanks for the fix!

Aorjoa:
Thank you @technoweenie :+1:  and sorry for long time no talk.

PR #670, opened 2015-09-17
Docgen fixes

technoweenie:
A few minor tweaks. The output of `help subcmd` and `subcmd -h` are now identical. Also, leading and ending linebreaks are consistent.

/cc @sinbad

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -25,8 +25,7 @@ var (
 	ErrorWriter  = io.MultiWriter(os.Stderr, ErrorBuffer)
 	OutputWriter = io.MultiWriter(os.Stdout, ErrorBuffer)
 	RootCmd      = &cobra.Command{
-		Use:   "git-lfs",
-		Short: "Git LFS provides large file storage to Git.",
+		Use: "git-lfs",
 		Run: func(cmd *cobra.Command, args []string) {
 			versionCommand(cmd, args)
 			cmd.Usage()
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -224,7 +224,7 @@ func determineIncludeExcludePaths(includeArg, excludeArg string) (include, exclu
 
 func printHelp(commandName string) {
 	if txt, ok := ManPages[commandName]; ok {
-		fmt.Fprintf(os.Stderr, txt)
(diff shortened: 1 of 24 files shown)


commands/commands.go
```diff
@@ -224,7 +223,7 @@ func determineIncludeExcludePaths(includeArg, excludeArg string) (include, exclu
 
 func printHelp(commandName string) {
 	if txt, ok := ManPages[commandName]; ok {
-		fmt.Fprintf(os.Stderr, txt)
+		fmt.Fprintf(os.Stderr, "%s\n", strings.TrimSpace(txt))
```
sinbad: Hmm, this puts the command snug under the prompt which I find harder to read, e.g.:
```
steve@Stevens-MacBook-Pro ~/p/G/s/g/g/git-lfs> git lfs pull --help
git lfs pull [options] [<remote>]

Download Git LFS objects for the currently checked out ref, and update
```
I find this harder to read than
```
steve@Stevens-MacBook-Pro ~/p/G/s/g/g/git-lfs> git lfs pull --help

git lfs pull [options] [<remote>]

Download Git LFS objects for the currently checked out ref, and update
```

sinbad:
:+1: I originally didn't remove the "Short" entries because Cobra generated a nice command list from them, but that's defunct since I started handling the root help topic. Nice catch on the br's.

commands/commands.go
technoweenie: I agree that whitespace is usually good. However in this case, the empty space looks off to me:

```bash
$ git help
usage: git [--version] [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p|--paginate|--no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           [-c name=value] [--help]
           <command> [<args>]

$ ruby -h
Usage: ruby [switches] [--] [programfile] [arguments]
  -0[octal]       specify record separator (\0, if no argument)
  -a              autosplit mode with -n or -p (splits $_ into $F)
```

technoweenie:
Ok, merging this. We can always change the whitespace thing later :)

sinbad:
Yeah, having reviewed other command help I see not having the spare line is normal, just looked odd to me after getting used to it being there :)

PR #671, opened 2015-09-17
teach 'update' cmd how to install hooks in bare repos

technoweenie:
Instead of requiring `git lfs update` to be run inside a git working directory, this relies on `lfs.LocalGitDir` to install the hooks in the correct place.

/cc @ttaylorr since you touched hook code recently. Though I'm pretty sure I'm the one that added the invalid repo error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,4 @@ notifications:
 before_install:
   - repo=`basename $PWD`; localDir=`dirname $PWD`; cfDir="`dirname $localDir`/github"
   - if [[ "$localDir" != "$cfDir" ]]; then mv "$localDir" "$cfDir"; cd ../../github/$repo; export TRAVIS_BUILD_DIR=`dirname $TRAVIS_BUILD_DIR`/$repo; fi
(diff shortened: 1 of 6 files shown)


technoweenie:
Had to add `GIT_LFS_TEST_DIR` to travis. Some of the tests rely on commands being run outside of git repositories. By default, tests run inside the git-lfs repository, causing those to fail. Setting `GIT_LFS_TEST_DIR` outside of the git-lfs directory fixes it.

Ideally, the tests would warn and probably skip tests that should fail when this happens.

technoweenie:
Confirmed this improves the process for migrating a repository to use LFS with BFG:

```bash
$ git clone --mirror https://github.com/mventurelli/relichunterszero
Cloning into bare repository 'relichunterszero.git'...

$ cd relichunterszero.git 
$ java -jar /path/to/bfg.jar --no-blob-protection --convert-to-git-lfs "*.wav"
$ java -jar /path/to/bfg.jar --no-blob-protection --convert-to-git-lfs "*.mp3"
$ git lfs init # sets up pre-push hook for the migrated repo
```

After that, set up the remote and push away!

Issue #307, opened 2015-05-14
git lfs ls-files does not list duplicates


andyneff:
I was just using git lfs ls-files to verify which files were being tracked, and I noticed when I expected to see two files, I only saw one. It turns out this was because they were identical files. Everything seemed to be functioning properly though, only one POST to my test lfs server, and only one GET file request when cloning, but git lfs ls-files only shows one file instead of both. I found this very confusing, and think it would be nice if it would at least list both (and maybe even tell me it's a duplicate?)

```
git init
git lfs init
git config lfs.url http://localhost:8080/git/test/

git lfs track '*.tgz'
git add .gitattributes

ls /proc -la > one.tgz
cp one.tgz two.tgz

git add one.tgz
git add two.tgz

git commit -m "commit"

```
git lfs ls-files yields 

    one.tgz

Instead of something like

    one.tgz
    two.tgz (duplicate of one.tgz)

Using version 0.5.1 of git-lfs

technoweenie:
Nice idea :+1:

ttaylorr:
Fixed 😄

PR #674, opened 2015-09-19
support git lfs ls-files show duplicate files

Aorjoa:
according #307 report `$git lfs ls-files` not list duplicate, i think this feature importance to me and then i have hack some code please review my code and suggest me :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,22 @@ page][impl]. You can chat with the team at: https://gitter.im/github/git-lfs
 
 ## Getting Started
 
-Download the [latest client][rel] and run the included install script.  The
-installer should run `git lfs init` for you, which sets up Git's global
-configuration settings for Git LFS.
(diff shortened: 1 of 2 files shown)


technoweenie:
Thanks for the patch :)  What do you think about just listing the OIDs and filenames? I don't know if we need to see the "duplicate of other-file.ext" messages. But seeing multiple files with the same OID should tip off the fact that they're dupes.

technoweenie:
I've been thinking about this more. I think it's time to rewrite fsck. For one, it can take advantage of `lfs.ScanTree()`, instead of `lfs.ScanRefs()`. I think that it should always do the following:

1. Ensure every file in .git/lfs/objects matches its own OID
2. Ensure every file in the working directory that matches the configured include/exclude settings

Maybe even add include/exclude flags to fsck, like fetch takes.

Issue #673, opened 2015-09-18
Git push succeeds even if LFS server is down


davidbernard04:
The issue https://github.com/github/git-lfs/issues/609 addressed the 401 failure case, but what about other types of upload failures?

I've tested some *push* scenarios when the LFS server is not reachable (or down), and the *git push* always succeed, even though the files are never uploaded.

Here are the *git push* traces, deliberately using a bad FQDN for the LFS server:
```
[/Projects/Stash/test_dbernard_lfs]$ GIT_TRACE=1 git push origin master
14:02:37.004276 git.c:346               trace: built-in: git 'push' 'origin' 'master'
14:02:37.005836 run-command.c:351       trace: run_command: 'ssh' '-p' '7999' 'git@stash' 'git-receive-pack '\''/drag/test_dbernard_lfs.git'\'''
14:02:37.245966 run-command.c:351       trace: run_command: '.git/hooks/pre-push' 'origin' 'ssh://git@stash:7999/drag/test_dbernard_lfs.git'
14:02:37.255984 git.c:555               trace: exec: 'git-lfs' 'pre-push' 'origin' 'ssh://git@stash:7999/drag/test_dbernard_lfs.git'
14:02:37.257033 run-command.c:351       trace: run_command: 'git-lfs' 'pre-push' 'origin' 'ssh://git@stash:7999/drag/test_dbernard_lfs.git'
trace git-lfs: run_command: git rev-list --objects b6f0c92e6a4ad7d502dc2cdf5a4c5050146db619 --not --remotes=origin
trace git-lfs: run_command: git cat-file --batch-check
trace git-lfs: run_command: git cat-file --batch
trace git-lfs: run_command: 'git' config -l
trace git-lfs: run_command: 'git' config -l -f /Projects/Stash/test_dbernard_lfs/.gitconfig
trace git-lfs: run_command: 'git' config -l -f /Projects/Stash/test_dbernard_lfs/.git/config
trace git-lfs: tq: starting 3 transfer workers
trace git-lfs: tq: running as individual queue
trace git-lfs: api: uploading (ef89aa51c09cb36df05b35a9fd19e8dc59cea5f8b2cf26f14443f5759cd6870b)
trace git-lfs: HTTP: POST http://dbernard@artifactory_bad_fqdn/artifactory/api/lfs/lfs-local-dbernard/objects
(0 of 1 files) 0 B / 110.13 KB
[...]

PR #676, opened 2015-09-21
[wip] better tests for handling server failures

technoweenie:


technoweenie:
This is an attempt to test various failure scenarios so we can catch things like #675 before they happen. Ideally, this PR fails until #675 is merged in, but I don't think that's quite the case just yet. But before I dig in, is this on the right track? Do the tests look useful?

I went with magic content strings because then it should be clear that the integration test is trying to trigger specific failure behavior on the server. The case statements are a lot more clear, despite the lines of copied code. If this approach works, we can replace `testingChunkedTransferEncoding()` with a magic string too.

technoweenie:
While writing this, I expected all of the push tests with the batch api to pass, and the push tests with the legacy api to fail until #675 was merged. However, the batch api tests were failing too.

```
# STDOUT
    Tracking *.dat
    [master (root-commit) bdc7654] welp
     3 files changed, 7 insertions(+)
     create mode 100644 bad.dat
     create mode 100644 good.dat
(1 of 2 files, 1 skipped) 2 B / 18 B, 16 B skipped                             
    push successful?

# STDERR
    + git lfs track '*.dat'
    + printf hi
    + printf status-batch-403
    + git add .gitattributes good.dat bad.dat
    + git commit -m welp
    + set +e
    + git push origin master
    To http://127.0.0.1:58006/test-push-failures-status-batch-403
    + '[' 0 = 0 ']'
    + echo 'push successful?'
    + exit 1
```

Looks like the file was being skipped, but not treated as an error. 3671cc7f27130ecd468a04bc27c7554382d6f3fa changes the batch response handler too look for an object error before looking for a link relation.

technoweenie:
So I merged #675 into this branch and pushed it up to a [temporary branch](https://github.com/github/git-lfs/compare/github:a62e510...github:b22949b) so CI can run. It passed!

https://travis-ci.org/github/git-lfs/builds/81463755

My plan now is:

1. Trigger #673 through a test in this PR
2. Ensure it passes when merged with #675.
3. Merge #675 
4. Merge #676

rubyist:
:shipit:

technoweenie:
Added a DNS test to this PR. Still passes in the temp merge branch:

https://github.com/github/git-lfs/compare/github:a62e510...github:f9b7779
https://travis-ci.org/github/git-lfs/builds/81469622

Issue #616, opened 2015-08-28
Proposal: Drop smudge in favor of post-checkout hook


technoweenie:
The improvements of the batch API and the updated `fetch` command are wasted on the Git smudge filters. Each checked out file has to invoke `smudge`, which only operates on a single file at a time. There's no parallelism here (unless Git adds it). 

The `post-checkout` hook runs after a `git checkout` is run after updating the work tree. The hook can then call `git lfs pull`, which downloads the LFS objects for the current working directory and replaces the pointers. This means it can make full use of the batch API and concurrent transfers.

Some downsides:

1. Cloning is more difficult. I don't know of a way to clone with a `post-checkout` hook. We could solve this with a `git lfs clone` wrapper. 
2. What happens if users have their own `post-checkout` scripts? I know that the Unreal Engine uses one, for example. Git needs a way for multiple tools to add their hooks in harmony.

whoisj:
Are post-checkout hooks invoked with actions like reset?

technoweenie:
@whoisj I _think_ so, but that's definitely a use case we'd need to check.

ttaylorr:
I've started an implementation which I'm tracking over at ttaylorr/git-lfs#1 :smile:.

bozaro:
Unreal Engine uses post-checkot hook for very similar usecase: they run GitDependencies.exe tool (written by Epic Games) for downloading large binaries files.

I asked them about their git-lfs plans: https://answers.unrealengine.com/questions/293214/gitdependenciesexe-vs-git-lfs.html

ttaylorr:
Finally got a chance to sit down with this, and there are a few issues I'd like to address.  The idea was to use the `post-checkout` hook as a means to trigger `git lfs pull`, which would in effect run the smudge filter in parallel.

Once I implemented this, nearly a third of integration tests were failing. I discovered the culprit is usually cloning a bare repo, and then pulling down commits from the remote later. This edge case does not invoke the `post-checkout` hook.  To me, this seems like a correct implementation of `post-checkout` since going from no commits to some commits doesn't involve a checkout.  I am not immediately sure how to solve this problem.

@whoisj brings up another issue, `reset`s in Git also do not trigger a post-checkout.

To me, these two cases together are significant for users of Git LFS.  I'd like to discuss whether or not a `post-checkout` hook is the ideal way to implement this, and see if there are any other options.

technoweenie:
Thanks for checking this out! Let's start small: 

[...]

PR #679, opened 2015-09-23
smudge passthrough

technoweenie:
This is an attempt at a workaround for #616. It's not ideal, but it gives users a way to optimize their downloads at the cost of complicating the git workflow a bit.

```bash
# downloads all files through smudge
$ git clone foo/bar.git

# downloads all files in a single batch command instead
$ GIT_LFS_SKIP_SMUDGE=1 git clone foo/bar.git
$ git lfs pull
```

That workaround sucks. We can make this better for users with:

```bash
# update smudge filter for all repos
$ git lfs init --skip-smudge

# update smudge filter for only the local repo
$ git lfs init --skip-smudge --local
```

### DIFF
--- a/commands/command_env.go
+++ b/commands/command_env.go
@@ -44,6 +44,11 @@ func envCommand(cmd *cobra.Command, args []string) {
 	for _, env := range lfs.Environ() {
 		Print(env)
 	}
+
+	for _, key := range []string{"filter.lfs.smudge", "filter.lfs.clean"} {
+		value, _ := lfs.Config.GitConfig(key)
+		Print("git config %s = %q", key, value)
+	}
 }
 
 func init() {
--- a/commands/command_init.go
+++ b/commands/command_init.go
@@ -16,9 +16,9 @@ var (
 		Run: initHooksCommand,
 	}
 
-	forceInit = false
-	localInit = false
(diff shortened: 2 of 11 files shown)


technoweenie:
Ok, ready for a review!

rubyist:
Seems good to me. I kind of like the env + pull, it's easily scriptable without altering any configuration. That's not to say we shouldn't keep improving, but I think it's a reasonable place to start.

technoweenie:
Not only is it easily scriptable, but it's easily run by native tools too.

commands/command_env.go
```diff
@@ -44,6 +44,11 @@ func envCommand(cmd *cobra.Command, args []string) {
 	for _, env := range lfs.Environ() {
 		Print(env)
 	}
+
+	for _, key := range []string{"filter.lfs.smudge", "filter.lfs.clean"} {
```
ttaylorr: Could refer back to [this](https://github.com/github/git-lfs/blob/master/lfs/setup.go#L21) variable

ttaylorr:
Looks good to me, glad we got this in :+1:

Tqelung:
我手機真是爛

PR #680, opened 2015-09-25
change ls-files to list OID(s)

Aorjoa:
from #674 I just change to show OIDs and filenames.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ repos
 docker/*.key
 
 src
+commands/mancontent_gen.go
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,4 @@ notifications:
 before_install:
   - repo=`basename $PWD`; localDir=`dirname $PWD`; cfDir="`dirname $localDir`/github"
   - if [[ "$localDir" != "$cfDir" ]]; then mv "$localDir" "$cfDir"; cd ../../github/$repo; export TRAVIS_BUILD_DIR=`dirname $TRAVIS_BUILD_DIR`/$repo; fi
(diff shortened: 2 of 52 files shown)


Aorjoa:
Opps! sorry for duplicated PR ;-(

technoweenie:
First, no worries about duplicate PRs. That's all good :) It seems odd that those commits from Sinbad are showing up again. Is your master up to date with the root's master?

I want to simplify the output. The brackets make the filename difficult to see, especially when there are multiple entries:

```bash
$ git lfs ls-files -l
50eb175bbfde63504add621427a6a98910e89697 ==> [ gif/droidtocat.gif,gif/dupe.gif ]
ba2f35f8b0ec51d85ffcc973e5297e22682bb2cd ==> [ png/render.png ]
```

How about something like this? The output is easier to read. We can also include an indicator showing LFS objects (`*`) vs LFS pointers (`-`).

```bash
$ git lfs ls-files -l
50eb175bbfde63504add621427a6a98910e89697 * gif/droidtocat.gif
50eb175bbfde63504add621427a6a98910e89697 * gif/dupe.gif
ba2f35f8b0ec51d85ffcc973e5297e22682bb2cd - png/render.png
```

Also, I'm thinking we can drop the call to `ScanRefs` and just rely on `ScanTree`. That seems like what users would expect. What do you think?

technoweenie:
I made a new PR with your commits: https://github.com/github/git-lfs/pull/681

Aorjoa:
Thank you for all your advice. @technoweenie for you ask on previous PR, I think * and - look good and I think we can use scanTree instead (but I don't know impact in dept).

technoweenie:
@Aorjoa Thanks for your help here. I merged these commits in #681, plus a few other changes.

PR #681, opened 2015-09-25
Ls files fixes

technoweenie:
This builds on #680:

* `lfs.ScanTree()` is used _instead of_ `lfs.ScanRefs()`. Duplicate files are now shown.
* The output includes the LFS OID and whether the file has the lfs pointer or the actual contents.

### DIFF
--- a/commands/command_ls_files.go
+++ b/commands/command_ls_files.go
@@ -1,7 +1,7 @@
 package commands
 
 import (
-	"strings"
+	"os"
 
 	"github.com/github/git-lfs/git"
 	"github.com/github/git-lfs/lfs"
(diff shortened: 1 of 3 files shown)


technoweenie:
@aorjoa: One more thing. We're not using the CLI package's built-in usage text anymore, so there's no point adding it to the flags definition. The help output is now rendered from the man pages directly:

```bash
$ git lfs help ls-files
git lfs ls-files [<ref>]

Display paths of Git LFS files that are found in the tree at the given
reference.  If no reference is given, scan the currently checked-out branch.

Options:

* -l --long:
```

PR #684, opened 2015-09-25
Better handling for tiny files below the blob size threshold that contain 'git-lfs'

technoweenie:
Git LFS uses a few properties to determine whether something is a valid LFS pointer or not. Right now, the criteria is:

1. Below the blob size threshold (1024 bytes)
2. Contains `git-lfs` in the content.

This causes problems for small test files like "this is a sample git-lfs file." So, it assumes anything that doesn't even start with `version` isn't even trying to be a Git LFS file.

/cc @rubyist: Did this by elevating the bad pointer key message to its own error type.

### DIFF
--- a/lfs/errors.go
+++ b/lfs/errors.go
@@ -162,6 +162,19 @@ func IsNotAPointerError(err error) bool {
 	return false
 }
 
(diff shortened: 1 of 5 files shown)


rubyist:
:+1:

Issue #687, opened 2015-10-01
The First Push In an LFS Repo With Batch Enabled Fails


WillHipschman:
The first push ends in an infinite 401 auth loop because we set lfs.<url>.access to basic, but don't pick it up until the next push.

Tqelung:
like

technoweenie:
Doh, fixed by #690.

PR #690, opened 2015-10-01
Auth fix for #687

WillHipschman:


technoweenie:
Can we get a test in place for this?

WillHipschman:
I've updated the current batch transfer test to use a capital repo name.

I've verified the test fails with 1.0 and passes with the fix,

technoweenie:
Still didn't work. We need to downcase all gitconfig keys.

WillHipschman:
I'm pretty sure gitconfig keys are downcased when we load the config--they were only uppercase when we modified the config in-process after it had already been loaded.

Do we have a tracking issue with a repro of the failure?

technoweenie:
> Do we have a tracking issue with a repro of the failure?

Nah, but that would've been good :)

I was a little frustrated when I wrote that comment after dealing with what I thought were more casing issues. People were having auth issues, and I instructed them to set the `access` value to force auth. Doing so can lead to issues if multiple git config keys are written:

```bash
$ git config --list
lfs.https://github.com/github/git-lfs-test.git/info/lfs.access=ntlm
lfs.https://github.com/GitHub/Git-LFS-Test.git/info/lfs.access=whatevs

$ git lfs env
git-lfs/1.0.2 (GitHub; darwin amd64; go 1.5.1; git 090222f)
git version 2.4.9 (Apple Git-60)

Endpoint=https://github.com/GitHub/Git-LFS-Test.git/info/lfs (auth=whatevs)
```

Which one is right? 

I think having Git LFS write config values with lower cased keys is technically correct, or at least _better_, but it wouldn't stop people from writing their own access keys. Maybe the fix is to have `git lfs env` warn if duplicate keys are found.

```
$ git config -l
# added by git-lfs
lfs.https://github.com/GitHub/Git-LFS-Test.git/info/lfs.access=basic
# added by the user
lfs.https://github.com/GITHUB/GIT-LFS-TEST.git/info/lfs.access=none

$ git lfs env
Endpoint=https://github.com/GitHub/Git-LFS-Test.git/info/lfs (auth=none)

# HELLO AUTH LOOP
$ git lfs pull
```

technoweenie:
I had a branch to enforce lowercase config keys. After writing this reply out, I decided to delete that branch and added the warning to https://github.com/github/git-lfs/pull/837 instead.

```bash
$ git config --list
lfs.https://github.com/GitHub/Git-LFS-Test.git/info/lfs.access=basic
lfs.https://github.com/GITHUB/GIT-LFS-TEST.git/info/lfs.access=none

$ git lfs env
WARNING: Reading LFS config from ".gitconfig", not ".lfsconfig". Rename to ".lfsconfig" before Git LFS v2.0 to remove this warning.
WARNING: These git config values clash:
  git config "lfs.https://github.com/GitHub/Git-LFS-Test.git/info/lfs.access" = "basic"
  git config "lfs.https://github.com/GITHUB/GIT-LFS-TEST.git/info/lfs.access" = "none"
git-lfs/1.0.2 (GitHub; darwin amd64; go 1.5.1; git 265feee)
git version 2.4.9 (Apple Git-60)

Endpoint=https://github.com/GitHub/Git-LFS-Test.git/info/lfs (auth=none)
```

WillHipschman:
Ahh makes sense. Glad you found a fix :+1:

PR #692, opened 2015-10-01
Fix resolve of git dirs finding a parent non-bare repo if wd is a bare repo

sinbad:
This causes tests to fail if GIT_LFS_TEST_DIR is not set, since the bare repo tests actually cascade up to the git-lfs repo itself and start modifying config files in there. Might potentially affect other cases where a bare repo is inside a non-bare repo.

Resolution is to use `git rev-parse` to locate all dirs instead of manually recursing. Could potentially remove manual processing of GIT_DIR and GIT_WORK_TREE in future but have left as belt & braces for now.

Had to remove test for using relative GIT_WORK_TREE with no GIT_DIR, git rev-parse always fails in that case so it appears it's not supported by git even though man git-config claims it is.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


technoweenie:
:+1: Feels good to butcher `recursiveResolveGitDir()` :)

PR #702, opened 2015-10-05
Fixed fetch include/exclude examples, should not have '='

sinbad:


technoweenie:
Doh! I just thought you used some weird alternate syntax :)

PR #710, opened 2015-10-06
Display the specific error in the case of a 403

rubyist:
A status of 403 against the API would be masked with a generic message. This displays a more specific message - either one returned in the API json, or a default message about authentication.

### DIFF
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -216,6 +216,8 @@ func Batch(objects []*objectResource, operation string) ([]*objectResource, erro
 		case 404, 410:
 			tracerx.Printf("api: batch not implemented: %d", res.StatusCode)
 			return nil, newNotImplementedError(nil)
+		case 403:
+			return nil, Error(err)
 		}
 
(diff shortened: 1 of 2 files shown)


technoweenie:
Should we just do this with every status code that falls through? I think returning the `err` would be better than losing it [here](https://github.com/github/git-lfs/blob/6f5c95d1135bff1810f56507c2d8b79859bc965e/lfs/client.go#L227-L229):

```go
 	if res.StatusCode != 200 {
 		return nil, Error(fmt.Errorf("Invalid status for %s %s: %d", req.Method, req.URL, res.StatusCode))
 	}
```

Assuming that the response has valid json with a "message" property, the error should get this info [here](https://github.com/github/git-lfs/blob/6f5c95d1135bff1810f56507c2d8b79859bc965e/lfs/client.go#L537).

rubyist:
Yeah that makes sense. That latest commit handles all the fallthroughs. The changed test shows that if the call returns some weird status without json it'll display an error about being unable to parse the response.

PR #711, opened 2015-10-06
Use batch API for Download, fallback to legacy if unsupported

### LINKED ISSUE #708: Client still depends on legacy API
The [*Git LFS v1 Original API*](https://github.com/github/git-lfs/blob/v1.0.0/docs/api/http-v1-original.md) says:

> This describes the original API for Git LFS v0.5.x. It's already deprecated by the batch API.

and the [*Git LFS v1 Batch API*](https://github.com/github/git-lfs/blob/v1.0.0/docs/api/http-v1-batch.md) says:

> This is a newer API introduced in Git LFS v0.5.2, and made the default in Git LFS v0.6.0. The client automatically detects if the server does not implement the API yet, and falls back to the legacy API.

However v1.0.0 of the client still depends on the legacy API for clone operations. During **git clone**, GETing LFS objects will fail with the following error where the legacy API is not available:

```
Repository or object not found: http://localhost/scm/test/lfsrepo.git/info/lfs/objects/1b88b71728388427ca7c6df0948add966bb0799b6a2dbb669145e84526fa81b4
```

rubyist:
The smudge filters still used the direct legacy api via the `Download()` function. This PR provides a new `Download()` which uses the batch API and falls back to a `DownloadLegacy()` if the server does not support the batch API.

This fixes #708

### DIFF
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -15,6 +15,7 @@ import (
 	"regexp"
 	"strconv"
 
+	"github.com/github/git-lfs/git"
 	"github.com/github/git-lfs/vendor/_nuts/github.com/rubyist/tracerx"
 )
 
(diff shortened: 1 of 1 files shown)


rubyist:
I think there's still plenty to refactor around this code, but we might want to hold off on that until we completely remove support of the legacy API. Then we can just remove all the legacy cruft at once.

sinbad:
:+1:  
Yeah it's slightly confusing at the moment because DownloadCheck() also uses the old API but then it's only ever called by the individualApiRoutine, so the use of the batch version & fallback is higher up. Download() is only ever used by the smudge filter now though so this does address the core issue for now, so I think it's still worth doing ahead of a more general tidy up.

technoweenie:
I updated this so that `lfs.Download()` only attempts batch transfers if `lfs.batch` is not false.

> so I think it's still worth doing ahead of a more general tidy up.

We absolutely have to do this before starting on your SSH pet project :)

rubyist:
:+1: to your changes

PR #713, opened 2015-10-06
Improved default remote behaviour & validation for fetch/pull

### LINKED ISSUE #660: [v0.60] Git LFS Pull on newly cloned repo causes 'Error for POST objects/batch' error
[Running Git LFS v0.60]

I have a Git LFS active repo that I have cloned using non-git-lfs commands due to Jenkins CI plugin limitations. After this is done I had planned to run *git lfs pull* to download the LFS-managed files. However in doing so I get the following error:

> (0 of 534 files) 0 B / 446.51
> Error for POST objects/batch
> Error for POST objects/batch
> Error for POST objects/batch
> Error for POST objects/batch
> Error for POST objects/batch
> Error for POST objects/batch

After clearing the git lfs logs and trying git lfs pull again the logs were empty.

I'm using a GitHub private repo.

I'm not sure how to debug this further. Happy to dig deeper given a few pointers (How does one enable verbose logging for Git in Windows?).

sinbad:
This fixes #660 and also improves a few other areas:
* Default remote is now either the tracked remote, or "origin" or any other *single* remote defined. If all of these fail there's a better error message now.
* User-supplied remotes are now validated to avoid ambiguous fail cases
* General tidy up to make it clearer what each of these git package functions are intended to do

### DIFF
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -30,10 +30,12 @@ func fetchCommand(cmd *cobra.Command, args []string) {
 		// Remote is first arg
 		lfs.Config.CurrentRemote = args[0]
 	} else {
-		trackedRemote, err := git.RemoteForCurrentBranch()
-		if err == nil {
-			lfs.Config.CurrentRemote = trackedRemote
-		} // otherwise leave as default (origin)
+		// Actively find the default remote, don't just assume origin
+		defaultRemote, err := git.DefaultRemote()
+		if err != nil {
+			Panic(err, "No default remote")
+		}
+		lfs.Config.CurrentRemote = defaultRemote
 	}
 
 	if len(args) > 1 {
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -29,14 +29,14 @@ func fetchCommand(cmd *cobra.Command, args []string) {
 	if len(args) > 0 {
 		// Remote is first arg
 		if err := git.ValidateRemote(args[0]); err != nil {
-			Panic(err, fmt.Sprintf("Invalid remote name '%v'", args[0]))
+			Exit("Invalid remote name %q", args[0])
 		}
 		lfs.Config.CurrentRemote = args[0]
 	} else {
 		// Actively find the default remote, don't just assume origin
 		defaultRemote, err := git.DefaultRemote()
 		if err != nil {
-			Panic(err, "No default remote")
+			Exit("No default remote")
 		}
 		lfs.Config.CurrentRemote = defaultRemote
 	}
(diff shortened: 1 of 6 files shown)


commands/command_fetch.go
```diff
@@ -28,12 +28,17 @@ func fetchCommand(cmd *cobra.Command, args []string) {
 
 	if len(args) > 0 {
 		// Remote is first arg
+		if err := git.ValidateRemote(args[0]); err != nil {
+			Panic(err, fmt.Sprintf("Invalid remote name '%v'", args[0]))
+		}
```
technoweenie: Instead of `Panic()`, I think `Exit()` would be more appropriate. We don't need to show a stack trace or anything like that, do we?

Also, you can use `%q` in the format string to have the value quoted: `"Invalid remote name %q"`

git/git.go
```diff
@@ -81,15 +81,27 @@ func CurrentBranch() (string, error) {
...
-func CurrentRemote() (string, error) {
+// RemoteForCurrentBranch returns the name of the remote that the current branch is tracking
+func RemoteForCurrentBranch() (string, error) {
+	ref, err := RemoteRefNameForCurrentBranch()
+	if err != nil {
+		return "", err
+	}
+	// ref name is full name, split first part for remote
+	parts := strings.Split(ref, "/")
+	return parts[0], nil
```
technoweenie: Ref names can have a `/` in them :( Maybe have `RemoteForCurrentBranch()` return the remote and current branch. Then `RemoteRefNameForCurrentBranch()` could call that and append the remote branch to the ref name.

sinbad: Yeah I know, but that doesn't matter for this? I'm only pulling out the remote name which can't have a '/'.

You can't just append the current branch to the remote name to get the remote tracking ref, that's what the code was doing before and it doesn't deal with tracking differently named remote branches. I fixed that at the same time.

commands/command_fetch.go
sinbad: I always forget about %q! I'll switch to Exit() in a couple of places.

git/git.go
technoweenie: Blah, I meant remotes. Git seems fine with it:

```bash
$ git remote show
foo/bar
origin

$ git fetch foo/bar
From https://github.com/OWNER/REPO
 * [new branch]      master     -> foo/bar/master
 * [new branch]      technoweenie-patch-1 -> foo/bar/technoweenie-patch-1
```

sinbad: Wait, what? :scream: I'm sure that used to be disallowed. OK, I'll have to cope with that then...

technoweenie:
:metal: I want to experiment with throwing this "default remote" info in `git lfs env` too.

PR #714, opened 2015-10-06
Add NTLM Auth

WillHipschman:
This a first attempt at adding NTLM auth. It addresss the proposal in #617. Please let me know your thoughts.

It uses lfs.{endpoint}.access to toggle NTLM and most ntlm specific logic is in its own class. We store user/pass in the cred helper but could later go get them from AD.

Things that still need to be done:

1. Add Tests-I can add unit tests but I don't think migrating the whole test infrastructure to support windows environments should block this.
2. Add documentation.


WillHipschman:
Checks are failing because I haven't yet included the ntlm auth library. Assuming that folks are ok with this implementation and the libraries use I'll add it in the next iteration.

lfs/config.go
```diff
@@ -141,12 +143,16 @@ func (c *Configuration) BatchTransfer() bool {
...
+	return c.Access() == "ntlm"
+}
+
 // PrivateAccess will retrieve the access value and return true if
 // the value is set to private. When a repo is marked as having private
 // access, the http requests for the batch api will fetch the credentials
 // before running, otherwise the request will run without credentials.
 func (c *Configuration) PrivateAccess() bool {
-	return c.Access() != "none"
+	return c.Access() != "none" && c.Access() != "ntlm"
```
WillHipschman: Remove the ntlm check here--ntlm is private access.

lfs/credentials.go
```diff
@@ -30,8 +30,9 @@ func getCreds(req *http.Request) (Creds, error) {
 // This prefers the Git remote URL for checking credentials so that users only
 // have to enter their passwords once for Git and Git LFS. It uses the same
 // URL path that Git does, in case 'useHttpPath' is enabled in the Git config.
-func getCredsForAPI(req *http.Request) (Creds, error) {
-	if skipCredsCheck(req) {
+func getCredsForAPI(req *http.Request, skip bool) (Creds, error) {
+	
+	if skip && skipCredsCheck(req) {
```
WillHipschman: Find a cleaner way to do this skip.

vendor/_nuts/github.com/ThomsonReutersEikon
```diff
@@ -0,0 +1 @@
+Subproject commit 52b7efa603f1b809167b528b8bbaa467e36fdc02
```
WillHipschman: We probably need to fork this into the git-lfs repo if we're gonna use it.

lfs/client.go
```diff
@@ -692,13 +746,17 @@ func setRequestAuthFromUrl(req *http.Request, u *url.URL) bool {
 }
 
 func setRequestAuth(req *http.Request, user, pass string) {
-	if len(user) == 0 && len(pass) == 0 {
-		return
-	}
+	if(Config.NtlmAccess()){
+		//no-op. The NTLM manager will handle auth headers
```
technoweenie: I'd prefer just returning early here so this doesn't add a level of nesting.

lfs/client.go
```diff
@@ -259,11 +287,21 @@ func UploadCheck(oidPath string) (*objectResource, error) {
...
+	
+	io.Copy(ioutil.Discard, res.Body)
+	res.Body.Close()
+	
 	if err != nil {
 		if IsAuthError(err) {
-			Config.SetAccess("basic")
-			tracerx.Printf("api: upload check not authorized, submitting with auth")
-			return UploadCheck(oidPath)
+			if strings.ToLower(res.Header["Www-Authenticate"][0][0:4]) == "ntlm" {
```
technoweenie: This looks like it could lead to panics. I'd recommend doing the following:

* Use `res.Header.Get("Www-Authenticate")`, which always returns a string.
* Use `strings.HasPrefix(wwwAuth, "ntlm")`, which won't panic if the header value only has 3 characters.
* Wrap this in a small helper func that takes an `*http.Response`. Looks like this check is repeated in multiple places.

```golang
authType := getAuthType(res)
Config.SetAccess(authType)
tracerx.Printf("api: upload check not authorized, submitting with %s auth", authType)
return UploadCheck(oidPath)
```

lfs/client.go
```diff
@@ -125,8 +126,15 @@ func Download(oid string) (io.ReadCloser, int64, error) {
 		return nil, 0, err
 	}
 	LogTransfer("lfs.data.download", res)
+	
+	buf, _ := ioutil.ReadAll(res.Body)
+	body := myReader{bytes.NewBuffer(buf)}
```
technoweenie: This reads the body into memory, which could get pretty large. The body should be cleaned up by all callers of `Download()` instead.

lfs/ntlm.go
```diff
@@ -0,0 +1,172 @@
...
+)
+
+func (c *Configuration) NTLMSession(creds Creds) ntlm.ClientSession {
+	
+	if c.ntlmSession != nil {
+		return c.ntlmSession
+	}
+	
+	splits := strings.Split(creds["username"], "\\")
+	var session, _  = ntlm.CreateClientSession(ntlm.Version2, ntlm.ConnectionOrientedMode)
```
technoweenie: Looks like this is hiding a possible error. I think that this function should be private, and should return a possible error that other functions can bubble up.

lfs/ntlm.go
```diff
@@ -0,0 +1,172 @@
...
+	session.SetUserInfo(splits[1], creds["password"], strings.ToUpper(splits[0]))
+	
+	c.ntlmSession = session
+	
+	return session
+}
+
+func DoNTLMRequest(request *http.Request, retry bool) (*http.Response, error) {					
+	handReq := cloneRequest(request)	
+	res, nil := InitHandShake(handReq)
```
technoweenie: I think this is silently discarding a possible error too.

lfs/ntlm.go
```diff
@@ -0,0 +1,172 @@
...
+		
+		saveCredentials(creds, res)	
+		
+		return res, nil	
+	}
+	return res, nil	
+}
+
+func InitHandShake(request *http.Request) (*http.Response, error){
+	var response, err = Config.HttpClient().Do(request)
```
technoweenie: It's common in go to use `:=` here instead of `var`:

```go
response, err := Config.HttpClient().Do(request)
```

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "vendor/_nuts/github.com/ThomsonReutersEikon"]
+	path = vendor/_nuts/github.com/ThomsonReutersEikon
+	url = https://github.com/ThomsonReutersEikon/go-ntlm
```
sinbad: Is there any reason this is a submodule instead of just being in the Nut.toml and added directly like the other dependencies? Submodules tend to over-complicate things unless you really need them.

lfs/ntlm.go
```diff
@@ -0,0 +1,172 @@
...
+		clonedReq.Header.Add(k,v[0])
+	}
+	
+	clonedReq.ContentLength = request.ContentLength	
+	
+	return clonedReq
+}
+
+func getNegotiateMessage() string{
+	return "NTLM TlRMTVNTUAABAAAAB7IIogwADAAzAAAACwALACgAAAAKAAAoAAAAD1dJTExISS1NQUlOTk9SVEhBTUVSSUNB"
```
sinbad: Is this is a hardcoded test value, or is it generated from something?

WillHipschman:
Thanks for the feedback @technoweenie. Before I go address it, are you ok with overall approach and use of the ThomsonReutersEikon library?

.gitmodules
WillHipschman: I didn't know the Nut.toml existed. I'll move it there :+1:

lfs/ntlm.go
WillHipschman: So the NTLM negotiate message has about a million flags (https://msdn.microsoft.com/en-us/library/cc236650.aspx), but I don't think we need to support all of them. This value is the "barebones" message that will work with IIS 7.5 or later.

Technically, this won't support NTLM servers that use NTLMv1 (pre-Windows 7) or NTLM servers that have special requirements, but I don't think there are many of them.

If we decide to support the full range of NTLM flags (or add new flags we support) we could construct the negotiate message ourselves, but I don't think it's necessary for the first version of this.

.gitmodules
rubyist: As a side note, the contributing doc outlines how to add deps with nut, if you hadn't seen it. I always have to consult it because I never remember.

lfs/client.go
WillHipschman: The issue is that multiple threads call `Download()`. If the thread 1 makes a request, and then thread 2 makes a request before thread 1 closes the stream, then the `*Http.Client` will spin up a new connection because it can't reuse the old one. NTLM is connection-oriented, so we will have to redo the 3-5 hop auth cycle.

If we instead close/copy the stream before returning it to the caller, every caller will reuse the connection and we only have to do the auth negotiation once.

I agree that this doesn't make sense for non-NTLM scenarios, so I can wrap it in an `if` block to only dispose early in NTLM cases. Long term, we may want a pool of `*http.Client`s and then we can just pick one that has an available connection.

technoweenie: Each `*http.Client` is designed to be goroutine-safe in an application. Its [Transport](https://golang.org/pkg/net/http/#Transport) has some properties for configuring its keepalive connections. Other than `CancelRequest()` and `CloseIdleConnections()`, it doesn't seem to give much visibility into how it's managing the connections. I wonder if we can just set `MaxIdleConnsPerHost` to 1? That would hurt concurrency for NTLM users, but should force a single connection per host.

technoweenie: @rubyist just mentioned that my previous idea of setting the max idle hosts to 1 would not work at all. It'd more likely just force each request to create a new connection.

It seems weird that NTLM would require the exact same TCP connection. Is there some token or state value that we can store on the global `Config` object, perhaps? I'm starting to think I'll have to learn how NTLM works before being comfortable merging this...

WillHipschman: The connection isn't guaranteed to be kept alive unless the *entire* body is read and then closed. See https://github.com/golang/go/issues/5645

I think for batch to work we need to handle multiple connections. 

Maybe we could check the response size and not copy it (and force the re-auth) if it's under some limit

lfs/client.go
```diff
@@ -126,6 +127,17 @@ func Download(oid string) (io.ReadCloser, int64, error) {
 	}
 	LogTransfer("lfs.data.download", res)
 
+	if Config.NtlmAccess() {
+		buf, _ := ioutil.ReadAll(res.Body)
```
rubyist: If we read the entire response into memory, large files can definitely crash the client on low memory machines. We could _maybe_ do this _if_ content length is set _and_ it's under some value - but I would really prefer to not do it at all for downloads. This is also ignoring the error.

rubyist:
Reading the whole body into memory doesn't seem like it's going to help much, because it still takes time to do that. Reading into memory vs reading into a file will probably not differ significantly enough to matter while running concurrent transfers, particularly with groups of files of varying sizes.

I think the way to guarantee that the connection will be reused is to not run it concurrently. This could be done at a higher level by setting the number of workers for the transfer queue to 1. But then I'd have to wonder if that's really worth it vs re-authing connections. What's the cost there? Do we have benchmarks? I'm thinking setting the max idle conns to be >=  the number of concurrent workers should limit the number of connections that get tossed.

sinbad:
If you're reading the entire body in order to try to force the re-use of the authenticated connection, you're just serialising all the transfers anyway, so you might as well make that explicit by disabling concurrency for NTLM. The first request is always issued serially anyway (others are delayed until auth has been performed) so once you've detected NTLM in that first request you could disable parallel requests if you wanted to re-use the connection, without bulk reading into memory all the time. 

I'm actually not convinced that the ReadAll is guaranteed not to be interrupted by another goroutine anyway, Go makes no assurances on that, so using it to try to serialise the requests probably isn't 100% watertight - it might be in this Go version but not always. Only deliberately running only one request at a time would do that, and then the regular code paths would ensure the body is read & closed before the next request.

But, I also agree with @rubyist that the proper solution is to allow the parallel transfers and to find a way to do the NTLM auth per connection, although I suspect that needs a to be implemented in a Transport implementation. TBH there's probably a good argument that our existing auth sequence should be in a Transport and not in the calling code.

sinbad:
> there's probably a good argument that our existing auth sequence should be in a Transport and not in the calling code.

.. although the first request would still need to be serialised in order to ensure the user only gets one login prompt.

WillHipschman:
I think that we were just missing a close in an earlier version somewhere. I removed the body copying and upload works just fine with

```
ConcurrentTransfers=4
BatchTransfer=true
```

WillHipschman:
I lied. Upload works fine, but Download has race conditions where some number of files don't get downloaded depending on interleaving. Setting 'ConcurrentTransfers=1` works though.

lfs/client.go
```diff
@@ -178,6 +182,9 @@ func DownloadCheck(oid string) (*objectResource, error) {
 		return nil, Error(err)
 	}
 
+	io.Copy(ioutil.Discard, res.Body)
+	res.Body.Close()
```
technoweenie: We shouldn't need this. `res` comes from `doLegacyApiRequest()`, which [calls `decodeApiResponse()`](https://github.com/github/git-lfs/blob/b0318b25350b9abfe56af0afb01913a9b2ed56a4/lfs/client.go#L413). That function [parses json and immediately closes the response body](https://github.com/github/git-lfs/blob/b0318b25350b9abfe56af0afb01913a9b2ed56a4/lfs/client.go#L594-L595). While it doesn't hurt, these lines shouldn't be required.

lfs/client.go
```diff
@@ -225,7 +232,11 @@ func Batch(objects []*objectResource, operation string) ([]*objectResource, erro
 	tracerx.Printf("api: batch %d files", len(objects))
 
 	res, objs, err := doApiBatchRequest(req)
+
+	io.Copy(ioutil.Discard, res.Body)
+	res.Body.Close()
```
technoweenie: Ditto [above](https://github.com/github/git-lfs/pull/714/files#r44162099).

lfs/credentials.go
```diff
@@ -75,7 +76,7 @@ func getCredURLForAPI(req *http.Request) (*url.URL, error) {
 			if gitRemoteUrl.Scheme == apiUrl.Scheme &&
 				gitRemoteUrl.Host == apiUrl.Host {
 
-				if setRequestAuthFromUrl(req, gitRemoteUrl) {
+				if !Config.NtlmAccess() && setRequestAuthFromUrl(req, gitRemoteUrl) {
```
technoweenie: I think these `Config.NtlmAccess()` checks should go inside these functions. This way callers don't need to know to skip `skipCredsCheck()` or `setRequestAuthFromUrl()` if ntlm auth is active.

lfs/ntlm.go
```diff
@@ -0,0 +1,215 @@
...
+	authenticateMessage := concatS("NTLM ", base64.StdEncoding.EncodeToString(authenticate.Bytes()))
+	request.Header.Add("Authorization", authenticateMessage)
+	return Config.HttpClient().Do(request)
+}
+
+func parseChallengeResponse(response *http.Response) ([]byte, error) {
+	header := response.Header.Get("Www-Authenticate")
+
+	//parse out the "NTLM " at the beginning of the response
+	challenge := header[5:]
```
technoweenie: This can panic if the header value is too short. I'd definitely check the string length and return an error. Much better than a scary panic backtrace.

lfs/client.go
```diff
@@ -235,9 +246,13 @@ func Batch(objects []*objectResource, operation string) ([]*objectResource, erro
...
 
 		if IsAuthError(err) {
-			Config.SetAccess("basic")
-			tracerx.Printf("api: batch not authorized, submitting with auth")
-			return Batch(objects, operation)
+			if isNtlmRequest(res) {
+				toggleAuthType("ntlm", "api: response indicates ntlm, submitting with %s auth")
+				return Batch(objects, operation)
+			} else {
+				toggleAuthType("basic", "api: batch not authorized, submitting with %s auth")
```
technoweenie: I think it makes sense for `toggleAuthType()` to write a standard message, instead of requiring a custom one for each auth type. I think all auth types can just use a format string like "api: response indicates %q authentication. Resubmitting"

Also, I don't see a reason for `isNtlmRequest()`. I think I'd prefer something like `responseAuthType(res)` that read "www-authenticate" or "lfs-authenticate".

lfs/client.go
```diff
@@ -475,7 +508,12 @@ func doHttpRequest(req *http.Request, creds Creds) (*http.Response, error) {
 	}
 
 	if err != nil {
-		err = Error(err)
+		if IsAuthError(err) && isNtlmRequest(res) {
```
technoweenie: When would this case come up? It seems like a faulty ntlm server could cause an infinite loop here. Is there some way that `DoNTLMRequest()` could handle this?

technoweenie:
See #820

Issue #705, opened 2015-10-05
Git LFS v1.0.1


technoweenie:
We'll probably be shipping a v1.0.1:

* [x] Downcase git config keys (prevents Auth loop) https://github.com/github/git-lfs/pull/690
* [x] Show more info for unexpected http responses https://github.com/github/git-lfs/pull/710
* [x] Use separate stdout/stderr buffers for `git-lfs-authenticate` https://github.com/github/git-lfs/pull/718
* [x] Use LoggedError instead of Panic if update-index fails in checkout #735
* [x] `smudge` command exits with non-zero if the download fails #732
* [x] Use `git rev-parse` to find the git working dir #692 
* [x] Improved default remote behaviour & validation for fetch/pull #713
* [x] Make fetch return error code when 1+ downloads failed #734
* [x] Improve lfs.InRepo() detection in `init`/`update` #756
* [x] whitelist the valid keys read from .gitconfig #760
* [x] Teach smudge to use the batch api #711
* [x] Fix not setting global attribute when needed to b/c of local state #765
* [x] Fix clone fail when fetch is excluded globally #770 
* [x] Fix for partial downloads problem #763

## Investigate

* [ ] Detect when credential cacher is not setup #523 
* [ ] Fix issue with ssh shorthands #278 
* [ ] Dynamic size cutoff for blob scanning #524 
* [ ] git-lfs staged but unstaged changes with 0B tracked file #557 
* [ ] fetch/pull don't respect tracking graph #660
* [ ] pushing to a new remote fails if local storage isn't full #662

sinbad:
> fetch/pull don't respect tracking graph #660

They actually do, the actual issue is that the error message isn't nice if the current checkout isn't on a tracking branch **and** there isn't an origin remote. Seems this is the case with a Jenkins build (and maybe other CIs). Alternative behaviour might be to use *any single* remote if origin isn't defined, i.e. if there's no origin and only one remote, use that, fail with nicer error if there's >1 remote and no origin.

technoweenie:
Ah right, that was just cargo culted from the previous issue.

PR #715, opened 2015-10-07
Add netrc support for LFS API requests

rubyist:
This PR adds support for using netrc for LFS API requests. The priority becomes:

1. Check the LFS URL for authentication. Ex: http://user:pass@example.com
1. Check netrc for authentication.
1. Check the Git remote URL for authentication IF it's the same scheme and host of the LFS URL.
1. Ask 'git credential' to fill in the password from one of the above URLs.

TODO:

- [x] Add netrc support
- [x] Add automated tests
- [x] What happens on Windows? Does netrc work there?

//cc #705


lfs/credentials.go
```diff
@@ -87,6 +97,27 @@ func getCredURLForAPI(req *http.Request) (*url.URL, error) {
 	return credsUrl, nil
 }
 
+func setCredURLFromNetrc(req *http.Request) bool {
+	home := os.Getenv("HOME")
+	if len(home) == 0 {
+		return false
+	}
+
+	nrc := filepath.Join(home, ".netrc")
```
technoweenie: Do you think it's worth caching this in the global `lfs.Config` object? Looks like this is called for every API request, so it could be used multiple times in a single `git-lfs` process (especially with the legacy API).

Mr-Mak-Hacking:
img_20151021_011523

technoweenie:
I added central caching of the `*netrc.Netrc` object in `Config`. Also added a platform specific `netrcBasename` var that's set to `.netrc` everywhere except windows, which uses `_netrc`.

technoweenie:
This looks good, but `script/integration` is hanging locally after the last test. The "master" and even the latest "cred-fixes" branches both work flawlessly for me. 

I'm going to hold it back from v1.1 until that can be checked out, and the netrc feature itself can be tested a bit more.  Should be a shoe-in for v1.2 though.

rubyist:
It hangs because `netrcBasename` is blank because go on osx won't build the file `config_linux.go`, even if the build tag tells it to. You have to name it something that isn't a platform name:

```
Naming a file dns_windows.go will cause it to be included only when building the package for Windows; similarly, math_386.s will be included only when building the package for 32-bit x86.
```

I think I usually see them as `_nix` or `_unix`.

technoweenie:
Doh, good catch.

On Wed, Nov 18, 2015 at 11:47 AM, Scott Barron <notifications@github.com>
wrote:

> It hangs because netrcBasename is blank because go on osx won't build the
> file config_linux.go, even if the build tag tells it to. You have to name
> it something that isn't a platform name:
>
> Naming a file dns_windows.go will cause it to be included only when building the package for Windows; similarly, math_386.s will be included only when building the package for 32-bit x86.
>
> I think I usually see them as _nix or _unix.
>
> —
> <https://github.com/github/git-lfs/pull/715#issuecomment-157817845>.
>



-- 
Rick Olson

rubyist:
I pushed a change that renames the file, but also changes the way the variable is set. Instead of a var + init() in the build flagged files, the build flagged files should define the variable. With this pattern, the build would have failed immediately because nothing defined `netrcBasename`.

PR #718, opened 2015-10-08
Use separate buffers for stdout and stderr on executing git-lfs-authenticate

bozaro:
It's fix authentication issue on ssh error messages like:
```
X11 forwarding request failed
```

### DIFF
--- a/lfs/ssh.go
+++ b/lfs/ssh.go
@@ -5,6 +5,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"bytes"
 
 	"github.com/github/git-lfs/vendor/_nuts/github.com/rubyist/tracerx"
 )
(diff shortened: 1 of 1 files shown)


rubyist:
This looks reasonable to me :+1:

lfs/ssh.go
```diff
@@ -37,12 +38,22 @@ func sshAuthenticate(endpoint Endpoint, operation, oid string) (sshAuthResponse,
 
 	cmd := exec.Command(exe, args...)
 
-	out, err := cmd.CombinedOutput()
+	// Save stdout and stderr in separate buffers
+	var outbuf, errbuf bytes.Buffer
+	cmd.Stdout = &outbuf
+	cmd.Stderr = &errbuf
```
sinbad: Is that strictly allowed? I thought the only supported way was to read from `cmd.StdErrPipe()` and `cmd.StdOutPipe()`

bozaro: I dont know. This fragment was found on StackOverflow: http://stackoverflow.com/questions/31095699/return-cmd-stdout-and-stderr-as-string-instead-of-printing-to-console-in-golang/31095907
I have no golang experience.

sinbad: Never mind, I see it's OK before Start(). I've just never seen that approach before, useful :)

sinbad:
:+1:

technoweenie:
Thanks!

PR #727, opened 2015-10-10
Add full zsh completion support

mikkeloscar:
This adds full zsh completion support for git-lfs. I have used the same file structure as the other PR for bash support. Let me know if I should change it in any way.

This version depends on the [_git](https://github.com/zsh-users/zsh/blob/master/Completion/Unix/Command/_git) completion distributed with zsh, and it should be installed into something like: `/usr/share/zsh/site-functions/_git-lfs`.

### DIFF
--- a/etc/git-lfs.zsh_completion
+++ b/etc/git-lfs.zsh_completion
@@ -1,0 +1,124 @@
+#compdef git-lfs
+#description manage large files with git-lfs
+
+_git-lfs() {
+  local -a commands help
+  commands=(
+  'env:Display the Git LFS environment'
+  'checkout:Populate working copy with real content from Git LFS files'
+  'fetch:Download git LFS files from a remote'
+  'fsck:Check GIT LFS files for consistency'
+  'init:Ensure Git LFS is configured properly'
+  'logs:Show errors from the git-lfs command'
+  'ls-files:Show information about Git LFS files in the index and working tree'
+  'pull:Fetch LFS changes from the remote & checkout any required working tree files'
+  'push:Push queued large files to the Git LFS endpoint'
+  'status:Show the status of Git LFS files in the working tree'
+  'track:View or add Git LFS paths to Git attributes'
+  'untrack:Remove Git LFS paths from Git Attributes'
+  'update:Update Git hooks for the current Git repository'
+  'clean:Git clean filter that converts large files to pointers'
+  'pointer:Build and compare pointers'
+  'pre-push:Git pre-push hook implementation'
+  'smudge:Git smudge filter that converts pointer in blobs to the actual content'
+  )
+
+  help=(
+  'help:Command help'
+  )
+
+  _arguments \
+    "1: :{_describe 'command' commands -- help}" \
+    "*:: :->args"
+
+  case $state in
+    args)
+      case $words[1] in
+        checkout|clean|track|untrack)
+          _files
+          ;;
+
+        fetch)
+          _arguments \
+            {'(--include=)-I','(-I)--include='}'[Specify lfs.fetchinclude just for this invocation]:path:_files' \
+            {'(--exclude=)-X','(-X)--exclude='}'[Specify lfs.fetchexclude just for this invocation]:path:_files' \
+              '--recent[Download objects referenced by recent branches & commits]' \
+              '--all[Download all objects referenced by any commit that is reachable]' \
+              '::remote:__git_remote_branch_names' \
+              '::ref:__git_commits'
+          ;;
+
+        init)
+          _arguments \
+            '--force[Sets the "lfs" smudge and clean filters, overwriting existing values]'
+          ;;
+
+        logs)
+          _arguments \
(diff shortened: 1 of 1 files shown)


technoweenie:
Hey, this is awesome, especially since I use zsh :) This is a good start, but I'd really like Git LFS to generate this somehow. I'm going to hold off on this until Git LFS v1.1 though, so we can figure out how users can install this.

sinbad:
As a new zsh user, I really liked that way I could just add support for many things like go via http://ohmyz.sh/ plugins. I wonder if we should have a way to generate a zsh config which can be submitted as a plugin there, with maybe an entry point for custom installs as well (but which has to be manually configured)?

technoweenie:
There are a couple problems here. First, the commands are doomed to be out of perpetually out of date. Second, what's installation like? "clone the repo down and copy to some directory owned by root"?

I'm cool with shipping this now, with updated descriptions and command flags. But I think it'd be worth building it into the command:

1. Turn this into a template that Git LFS can generate from the `commands` package.
2. Add command to install to something in `$fpath`
3. Add command to install for oh-my-zsh as @sinbad suggested. Maybe even submit to https://github.com/zsh-users/zsh-completions.
4. Teach `git lfs install/update` how to update this. Maybe store the location in git config:

    ```bash
    $ git config --global lfs.etc.zshautocomplete
    /Users/rick/.zsh/site-functions/_git-lfs
    ```

Also, is there any way to get this to trigger for `git lfs`?

FWIW I happened to stumble on this project: https://github.com/bobthecow/git-flow-completion. Might be a good example that we can borrow from for bash/fish autocompletion :)

sinbad:
:+1: Thinking about this overnight I was also thinking this should just be an lfs command (`git lfs zsh`?) to generate the script, and that it should do it from the command data we already have in cobra so that it doesn't need to be manually updated. Suggested options:

- If no options: 
  - If `$ZSH_CUSTOM` is defined, write script to `$ZSH_CUSTOM/plugins/git-lfs/git-lfs.plugin.zsh`. This installs it just for this user with oh-my-ssh.
  - Otherwise write to `~/.git-lfs.zsh`?
- Option `--oh-my-ssh-core` (or something) to install to `~/.oh-my-zsh/plugins/git-lfs/git-lfs.plugin.zsh` (for submitting upstream to http://ohmyz.sh/
- Option `-f <foo>` to write to a custom location. I'm not sure if we should try to edit `.zshrc` to include this location or leave that to the user, I'd probably err on the side of the latter

> Bonus points if we somehow be added to git's autocompletions :)

Yeah, wonder if we need to send PRs to both core Git, http://ohmyz.sh/ and https://github.com/zsh-users/zsh-completions or whether there's a more efficient way?

sinbad:
I've been exploring generating a script using this as a template: https://github.com/github/git-lfs/blob/zsh-completions/commands/command_zsh.go

technoweenie:
I believe `$ZSH_CUSTOM` is [an oh-my-zsh thing](https://github.com/robbyrussell/oh-my-zsh/wiki/Customization). With plain `.zsh`, autocompletions are loaded as a site function in a directory defined in `$fpath`. The filename `$fpath/_git-lfs` seems to be significant, linking the autocompletions to `git-lfs`, but not `git lfs`. I think we can just use `$fpath`, and tell the user how to update it in their `.zshrc` if there's a problem adding the file.

Instead of `git lfs zsh`, how about something like `git lfs integrate zsh`? This would be a good spot to add other integrations, like bash, in the future. It should automatically detect oh-my-zsh if possible. If not, then we can add `git lfs integrate oh-my-zsh` or something.

> I've been exploring generating a script using this as a template

Would you mind throwing that on top of this PR? I want to include @mikkeloscar's work here in the commit list, since we're taking a lot of inspiration from it.

sinbad:
> I believe `$ZSH_CUSTOM` is an oh-my-zsh thing

Yeah that's what I was trying to use as a trigger to assume oh-my-zsh if given no other context.

> The filename `$fpath/_git-lfs` seems to be significant, linking the autocompletions to `git-lfs`

So if `$fpath` is specified, I just default to `$fpath/_git-lfs` with no `.zsh` ext? I'll probably need you to verify this works in practice once I do...

> Would you mind throwing that on top of this PR? I want to include @mikkeloscar's work here in the commit list, since we're taking a lot of inspiration from it.

Oh yeah that's an excellent idea. Rebased & done.

sinbad:
> how about something like `git lfs integrate zsh`?

That's a good shout too, I'll do this once I've figured out a couple of other niggles.

sinbad:
New PR at #1133

mikkeloscar:
Hello, I just wanna say that I have been following the discussion here and it's fine by me if you close this PR when you want to :)

Also, I'm glad that you could use some of the stuff from this PR.

sinbad:
Thanks! If you have any ideas of how to add completions for `git lfs` (i.e. with a space rather than `git-lfs`), I'd love to hear them at #1133, I totally failed when I tried to adapt the script :)

PR #732, opened 2015-10-12
Exit 1 if a smudge fails

rubyist:
This will cause the git checkout to abort instead of continue on as if things are OK.

/cc https://github.com/github/git-lfs/issues/731

### DIFF
--- a/commands/command_smudge.go
+++ b/commands/command_smudge.go
@@ -75,7 +75,7 @@ func smudgeCommand(cmd *cobra.Command, args []string) {
 
 	if err != nil {
 		ptr.Encode(os.Stdout)
-		Panic(err, "Error accessing media: %s (%s)", filename, ptr.Oid)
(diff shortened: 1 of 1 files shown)


technoweenie:
How about just `Error()`? Does it need to dump the stacktrace and all that?

rubyist:
It won't display the stacktrace in either case for clone, but we could use `Exit()` instead. `Error()` still exit 0's.

technoweenie:
Ah right, `Exit()` then. Most of these errors should be transient network issues or whatever. It should still write detailed enough errors to STDERR.

rubyist:
On third thought, maybe we should do `LoggedError()` like it was, and just add an `os.Exit(2)`. Skipping `LoggedError()` won't leave any logs behind.

bozaro:
:+1:

Issue #725, opened 2015-10-09
git lfs pull stuck


swolfish:
During the `git lfs pull` download gets stuck and after bunch of time I cancelled it (ctrl+c). And then it's not possible to continue/restart the process.
This is the output:
```bash
(10 of 33 files) 4.91 MB / 92.80 MB
Error opening media file buffer.
Error updating the git index

Errors logged to /home/***/.git/lfs/objects/logs/20151009T131458.580210367.log
Use `git lfs logs last` to view the log.
```
Running `git lfs logs last` got this:
```bash
git-lfs/1.0.0 (GitHub; linux amd64; go 1.5.1; git 7379fa9)
git version 2.1.0

$ git-lfs pull
Error opening media file buffer.
Error updating the git index

exit status 128
goroutine 51 [running]:
github.com/github/git-lfs/lfs.Stack(0x0, 0x0, 0x0)
/Users/rick/go/src/github.com/github/git-lfs/lfs/errors.go:557 +0x80
github.com/github/git-lfs/commands.logPanicToWriter(0x7f3a1a05d1e8, 0xc8200340b0, 0x7f3a1a05ee80, 0xc8200340a8)
/Users/rick/go/src/github.com/github/git-lfs/commands/commands.go:184 +0xf7f
github.com/github/git-lfs/commands.logPanic(0x7f3a1a05ee80, 0xc8200340a8, 0x0, 0x0)
/Users/rick/go/src/github.com/github/git-lfs/commands/commands.go:148 +0x421
github.com/github/git-lfs/commands.handlePanic(0x7f3a1a05ee80, 0xc8200340a8, 0x0, 0x0)
/Users/rick/go/src/github.com/github/git-lfs/commands/commands.go:123 +0x4e
github.com/github/git-lfs/commands.LoggedError(0x7f3a1a05ee80, 0xc8200340a8, 0x8da8e0, 0x1c, 0x0, 0x0, 0x0)
/Users/rick/go/src/github.com/github/git-lfs/commands/commands.go:73 +0x82
github.com/github/git-lfs/commands.Panic(0x7f3a1a05ee80, 0xc8200340a8, 0x8da8e0, 0x1c, 0x0, 0x0, 0x0)
/Users/rick/go/src/github.com/github/git-lfs/commands/commands.go:83 +0x5d
github.com/github/git-lfs/commands.checkoutWithChan(0xc8200fe060)
[...]

PR #735, opened 2015-10-13
Use LoggedError instead of Panic if update-index fails in checkout

sinbad:
Also report stderr/out since error object is often something useless like "exit status 128".
To help with #725.


technoweenie:
:metal:

PR #742, opened 2015-10-15
Prune feature (delete old / unreferenced local objects)

sinbad:
I think this is finally ready for review now. Please see `docs/man/git-lfs-prune.1.ronn` for details of what it does and how it determines what to retain and what to delete.

I've created quite a detailed set of tests and think I've covered everything, but please do scrutinise, since this does involve deleting user data!


sinbad:
Oh, the intention is to add a `--prune` option to `git lfs fetch` too, I haven't done that yet. All it will do is call the `prune()` function so doesn't stop review of the main guts.

commands/command_prune.go
```diff
@@ -0,0 +1,460 @@
...
+	pruneDryRunArg      bool
+	pruneVerboseArg     bool
+	pruneVerifyArg      bool
+	pruneDoNotVerifyArg bool
+)
+
+func pruneCommand(cmd *cobra.Command, args []string) {
+
+	// Guts of this must be re-usable from fetch --prune so just parse & dispatch
+	if pruneVerifyArg && pruneDoNotVerifyArg {
```
WillHipschman: It feels weird having both --verify-remote and --no-verify remote. I assume it's part of the prune spec? Maybe we could have setting one to true set the other to false so that we don't always have to check two things?

commands/command_prune.go
```diff
@@ -0,0 +1,460 @@
...
+
+const (
+	PruneProgressTypeLocal  = PruneProgressType(iota)
+	PruneProgressTypeRetain = PruneProgressType(iota)
+	PruneProgressTypeVerify = PruneProgressType(iota)
+)
+
+// Progress from a sub-task of prune
+type PruneProgress struct {
+	Type  PruneProgressType
```
WillHipschman: I would name this variable ProgressType.  'Type' makes me think of object types.

commands/command_prune.go
sinbad: This is the standard `git` pattern for options that can be enabled on the command line, or overridden to be false again e.g. when they're set to true in `gitconfig`. See `--ff` vs `--no-ff`, `--squash` vs `--no-squash` and many others. I prefer to feel like `git` whenever possible.

commands/command_prune.go
sinbad: It's a bit redundant given the scope but I guess coming from other languages especially this could be ambiguous. Changed.

technoweenie:
I merged master and had a few conflicts. Though based on the diffs, they were just cases where this PR and master introduced functions in the same place. I think the majority of it came from https://github.com/github/git-lfs/pull/801.

You can see my changes: https://github.com/github/git-lfs/compare/sinbad-prune

FWIW I'm not super thrilled about `lfs.NewScanRefsOptions()`. If you have thoughts, I'd be open to discussing those for a future PR.

sinbad:
Looks good, I've merged those changes into this PR.

kosikfll:
I upgraded everything can you all look at it and tell me what you think

commands/command_prune.go
GiniFa: There some HARDCORE RAP

PR #749, opened 2015-10-16
Update git-lfs.spec

glensc:
update Source0 url to use existing url

### DIFF
--- a/rpm/SPECS/git-lfs.spec
+++ b/rpm/SPECS/git-lfs.spec
@@ -6,7 +6,7 @@ Summary:        Git extension for versioning large files
 Group:          Applications/Archiving
 License:        MIT
 URL:            https://git-lfs.github.com/
-Source0:        https://github.com/github/git-lfs/archive/%{version}/%{name}-%{version}.tar.gz
+Source0:        https://github.com/github/git-lfs/archive/v%{version}/%{name}-%{version}.tar.gz
 BuildRoot:      %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 BuildRequires:  golang, tar, which, bison, rubygem-ronn, git
 BuildRequires:  perl-Digest-SHA
--- a/rpm/SPECS/git-lfs.spec
+++ b/rpm/SPECS/git-lfs.spec
@@ -6,7 +6,7 @@ Summary:        Git extension for versioning large files
 Group:          Applications/Archiving
 License:        MIT
 URL:            https://git-lfs.github.com/
-Source0:        https://github.com/github/git-lfs/archive/%{name}-%{version}.tar.gz
+Source0:        https://github.com/github/git-lfs/archive/%{version}/%{name}-%{version}.tar.gz
 BuildRoot:      %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 BuildRequires:  golang, tar, which, bison, rubygem-ronn, git
 BuildRequires:  perl-Digest-SHA


technoweenie:
@glensc Is there a reason this PR can't be combined with #750?

glensc:
err no. just created two pr's for two different problems

rpm/SPECS/git-lfs.spec
```diff
@@ -6,7 +6,7 @@ Summary:        Git extension for versioning large files
 Group:          Applications/Archiving
 License:        MIT
 URL:            https://git-lfs.github.com/
-Source0:        https://github.com/github/git-lfs/archive/%{name}-%{version}.tar.gz
+Source0:        https://github.com/github/git-lfs/archive/%{version}/%{name}-%{version}.tar.gz
 BuildRoot:      %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
```
jabr08: I'm Trying

andyneff:
Two problems

1. https://github.com/github/git-lfs/archive/1.0.0/git-lfs-1.0.0.tar.gz doesn't actually exist. This means either `https://github.com/github/git-lfs/archive/v%{version}.tar.gz` or `https://github.com/github/git-lfs/archive/v%{version}/%{name}-v%{version}.tar.gz` needs to be used. Ok, so now we've changed the filename to v1.0.0.tar.gz or git-lfs-v1.0.0.tar.gz. This is looking less spec like, because tradition is `git-lfs-1.0.0.tar.gz`
2. If we did change the filename part of the url here, it will break https://github.com/glensc/git-lfs/blob/master/rpm/build_rpms.bsh#L181. I would prefer if one be updated, so the other.

So which way should we go? 

1. Change Source0 to just `%{name}-%{version}.tar.gz` because lets face it, the rpmbuild just doesn't download that url for you, you have to use something like spectool which isn't really mainstream. So maybe "why even bother with the url"?
2. Change EVERYTHING to v%{version}.tar.gz
3. Change EVERYTHING to %{name}-v%{version}.tar.gz
4. Other ideas???

glensc:
sorry, will fix tag matching, this actually works:

https://github.com/github/git-lfs/archive/v%{version}/%{name}-%{version}.tar.gz

pushed another commit. let me know if need to squash

glensc:
> because lets face it, the rpmbuild just doesn't download that url for you.

yes it does. at least rpm 5.x.

andyneff:
So that's how the archive/{tag}/ thing works... anyway you want. Nice

:+1:

rpm/SPECS/git-lfs.spec
```diff
@@ -6,7 +6,7 @@ Summary:        Git extension for versioning large files
 Group:          Applications/Archiving
 License:        MIT
 URL:            https://git-lfs.github.com/
-Source0:        https://github.com/github/git-lfs/archive/%{name}-%{version}.tar.gz
+Source0:        https://github.com/github/git-lfs/archive/v%{version}/%{name}-%{version}.tar.gz
```
technoweenie: The URLs are actually `https://github.com/github/git-lfs/archive/{tag}.tar.gz`, like https://github.com/github/git-lfs/archive/v1.0.2.tar.gz. The "{name}-{version}" filename is set on a Content-Disposition header in the response.

glensc: but rpm needs last path component to be filename {name}-{version}.tar.gz, therefore the url hack. and it works fine.

you can inject any garbage after /archive/tag in url, github will ignore the junk and provide data under the complete url.

glensc: it "just works". see:

```
$ wget https://github.com/github/git-lfs/archive/v1.0.2/git-lfs-1.0.2.tar.gz
--2015-11-18 00:30:28--  https://github.com/github/git-lfs/archive/v1.0.2/git-lfs-1.0.2.tar.gz
Resolving github.com (github.com)... 192.30.252.129
Connecting to github.com (github.com)|192.30.252.129|:443... connected.
HTTP request sent, awaiting response... 302 Found
Location: https://codeload.github.com/github/git-lfs/tar.gz/v1.0.2 [following]
--2015-11-18 00:30:28--  https://codeload.github.com/github/git-lfs/tar.gz/v1.0.2
Resolving codeload.github.com (codeload.github.com)... 192.30.252.146
Connecting to codeload.github.com (codeload.github.com)|192.30.252.146|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 260484 (254K) [application/x-gzip]
Saving to: ‘git-lfs-1.0.2.tar.gz’

git-lfs-1.0.2.tar.gz            100%[========================================================>] 254.38K   417KB/s   in 0.6s   

2015-11-18 00:30:30 (417 KB/s) - ‘git-lfs-1.0.2.tar.gz’ saved [260484/260484]

$ 
```

technoweenie: Right on :metal: Thanks for the clarification.

PR #750, opened 2015-10-16
Update git-lfs.spec

glensc:
cleanup BR, the tools are not used during build.
they were added in initial commit, 4a716277, so likely bad copy-paste

### DIFF
--- a/rpm/SPECS/git-lfs.spec
+++ b/rpm/SPECS/git-lfs.spec
@@ -8,6 +8,7 @@ License:        MIT
 URL:            https://git-lfs.github.com/
 Source0:        https://github.com/github/git-lfs/archive/%{name}-%{version}.tar.gz
 BuildRoot:      %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
+BuildRequires:  perl-Digest-SHA
 BuildRequires:  golang, tar, rubygem-ronn, git
 
 %define debug_package %{nil}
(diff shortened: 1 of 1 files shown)


andyneff:
perl-Digest-SHA is used 
https://github.com/github/git-lfs/blob/7379fa9087b83ebd1f30d0af53320007fe3cc4be/test/test-fsck.sh#L25
https://github.com/github/git-lfs/blob/7379fa9087b83ebd1f30d0af53320007fe3cc4be/test/test-fsck.sh#L33
https://github.com/github/git-lfs/blob/4915979d5aadfc4daa4cdba6d84db694802f8086/test/testhelpers.sh#L345

which MAY have been left over from build_rpm.bsh, but I'm not 100% sure it'll build without until I test everything

And I can't tell you what used/used bison right now.

@glensc Were you able to build and pass the tests without these installed on Centos 5, 6 and 7? The dockers have these installed already, so that's not a good test.

glensc:
ok, `shasum` is in `perl-tools` in my distro, will be adding it back. thanks for review!

and sorry, i did not actually ran the tests :(

andyneff:
`which` is needed by `script/man`

```
./script/man: line 2: which: command not found
```

So the only `bison` can be removed. `which` isn't a requirement of yum or part of the "Development Tools", so it should be kept it.

Add `which` back in please, and then this is good to go!

glensc:
@andyneff BR which is back!

technoweenie:
Thanks @glensc for opening this, and @andyneff for reviewing :metal:

Issue #717, opened 2015-10-08
`git lfs init` should detect absence of `.git`


wkoszek:
Looks like `git lfs init` doesn't detect the lack of `git` correctly:
```
wk:~> cd /tmp/
wk:/tmp> mkdir lfs-exp1
wk:/tmp> cd lfs-exp1
wk:/tmp/lfs-exp1> git lfs init
Git LFS initialized.
wk:/tmp/lfs-exp1> ls -la
total 0
drwxr-xr-x   2 wk    wheel    68  7 paź 23:03 .
drwxrwxrwt  59 root  wheel  2006  7 paź 23:03 ..
wk:/tmp/lfs-exp1> git init
Initialized empty Git repository in /private/tmp/lfs-exp1/.git/
wk:/tmp/lfs-exp1> git lfs init
Updated pre-push hook.
Git LFS initialized.
wk:/tmp/lfs-exp1> la
total 0
drwxr-xr-x   3 wk    wheel   102  7 paź 23:04 .
drwxrwxrwt  59 root  wheel  2006  7 paź 23:03 ..
drwxr-xr-x  10 wk    wheel   340  7 paź 23:04 .git
```

I'd expect it to complain.

technoweenie:
Its primary purpose is to setup the global clean and smudge filters. IF it's inside a repository, it also installs the pre-push hook. Other commands will do that too, like the "track" command.

wkoszek:
@technoweenie can you confirm that the observed behavior is suboptimal and that we're agreeing that `git lfs init` reports success without actually doing anything?

technoweenie:
It sets the global clean and smudge filters:

```bash
$ git config --global --list | grep lfs
filter.lfs.clean=git-lfs clean %f
filter.lfs.smudge=git-lfs smudge %f
filter.lfs.required=true
```

AlicanC:
> Its primary purpose is to setup the global clean and smudge filters. IF it's inside a repository, it also installs the pre-push hook. Other commands will do that too, like the "track" command.

@technoweenie, it installs the pre-push hook even when it's not in a repository:
```
C:\Users\redacted\not-a-repository> ls -Name
C:\Users\redacted\not-a-repository> git lfs init
Updated pre-push hook.
Git LFS initialized.
C:\Users\redacted\not-a-repository> ls -Name
hooks
C:\Users\redacted\not-a-repository> ls hooks -Name
pre-push
```

```
[...]

PR #752, opened 2015-10-17
lfs complain not a git directory

Aorjoa:
For fixed issue #717 and I have an question.
Why we provide flag `local`and `force` in source code but in [help page](https://github.com/github/git-lfs/blob/master/docs/man/git-lfs-init.1.ronn) not found how to use it?

### DIFF
--- a/commands/command_init.go
+++ b/commands/command_init.go
@@ -22,7 +22,13 @@ var (
 )
 
 func initCommand(cmd *cobra.Command, args []string) {
-	requireInRepo()
+	if localInit {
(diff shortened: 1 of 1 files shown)


technoweenie:
I think you found the core problem, but this doesn't really solve the bug. It still creates that `hooks` dir. This is a better fix IMO: #756

technoweenie:
Closing in favor of #756

technoweenie:
I added the missing `local` and `skip-smudge` flags to the init man page: #761

PR #760, opened 2015-10-20
whitelist the valid keys from .gitconfig

technoweenie:
This whitelists the valid keys that are read from `.gitconfig`. Git LFS can't make Git read options from `.gitconfig` by default, but it should be explicit about what it allows.

* `lfs.url`
* `remote.{name}.lfsurl`
* `lfs.extension.{name}.priority`

The `.gitconfig` file is designed for common settings for all users cloning.  IMO the other settings should be set by a user's standard git config.

### DIFF
--- a/lfs/config.go
+++ b/lfs/config.go
@@ -301,12 +301,9 @@ func (c *Configuration) loadGitConfig() bool {
 		return false
 	}
 
-	uniqRemotes := make(map[string]bool)
-
 	c.gitConfig = make(map[string]string)
 	c.extensions = make(map[string]Extension)
 
-	var output string
 	listOutput, err := git.Config.List()
 	if err != nil {
 		panic(fmt.Errorf("Error listing git config: %s", err))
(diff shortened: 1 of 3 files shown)


rubyist:
:+1:

ptoomey3:
This seems to accomplish what we want. But, I do worry this code is a little fragile since it feels like there are so many locations that end up having to check for the `whitelist` and `allowed` variables. I wonder if having a common chokepoint at the top of the function would be useful to maintain this more easily going forward. I was thinking of something like:

``` go
key := strings.ToLower(pieces[0])
value := pieces[1]

keyParts := strings.Split(key, ".")
if whitelisted && !c.isSafeKeyAndValue(keyParts, value) {
  continue
}
...
```

Also, I found the term `whitelisted` a bit confusing. At first glance I was reading it as though `c.readGitConfig(fileOutput, uniqRemotes, true)` meant that the config was "whitelisted" because it was safe, where it actually means "please whitelist this config". I wonder if something like "safeLFSKeysOnly" (or whatever would be idiomatic Go naming) would be more descriptive.

lfs/config.go
```diff
@@ -315,48 +312,85 @@ func (c *Configuration) loadGitConfig() bool {
...
 				if err == nil && p >= 0 {
 					ext.Priority = p
 				}
 			}
+
 			ext.Name = name
 			c.extensions[name] = ext
-		} else if len(keyParts) == 2 && keyParts[0] == "lfs" && keyParts[1] == "fetchinclude" {
+		} else if len(keyParts) > 1 && keyParts[0] == "remote" {
+			if whitelisted && (len(keyParts) < 3 || keyParts[2] != "lfsurl") {
```
ptoomey3: Is there any reason we can't check `len(keyParts) != 3` here?

technoweenie: It would break for any subkeys under `remote.{name}.lfsurl`. I don't anticipate this being an issue in this case though.

ptoomey3: Are there any such subkeys under `remote.{name}.lfsurl` today? Not that I envision you adding any dangerous subkeys here, but it would be nice to be explicit and whitelist those subkeys as they are added rather than assume any new subkey is safe.

technoweenie:
I incorporated the easy feedback in #776. I agree about the fragility of this function though. It's providing a single map of git config values, setting up internal extension objects, and collecting a list unique remotes from the configuration. I've added a bigger refactoring task for v1.1: https://github.com/github/git-lfs/issues/755

lfs/config.go
technoweenie: Fair enough, see #776.

ptoomey3:
Much :heart:  @technoweenie. Thanks for the fix!

Issue #266, opened 2015-04-28
Git LFS partial download


technoweenie:
When a download in `git lfs smudge` is cancelled, it sometimes leave some trash in `.git/lfs/objects`:

```
$ ls -al .git/lfs/objects/ab/cd
total 43928
drwxr--r--  4 rick  staff    136 Apr 28 08:02 ./
drwxr--r--  3 rick  staff    102 Apr 28 08:02 ../
-rw-r--r--  1 rick  staff      0 Apr 28 08:02 abcd...
-rw-r--r--  1 rick  staff  13523 Apr 28 08:06 abcd...-temp
```

On a second `git checkout`, the 0-length file gets moved over to the working directory.  My hunch is that the file gets opened for file locking purposes, but never cleaned up because the command was aborted.

technoweenie:
Fixed in #267

andyneff:
I am still getting this error, with version 0.5.2 and using the lastest git-lfs (off of 2eef603f)

```
LocalWorkingDir=/dev/shm/blah2/vip_root
LocalGitDir=.git
LocalMediaDir=.git/lfs/objects
TempDir=.git/lfs/tmp
ConcurrentTransfers=3
BatchTransfer=false
GIT_DIR=.git
GIT_PREFIX=

Error accessing media: external/src/SOURCES/GDAL-1.11.2-cp27-none-win_amd64.whl (90bd3def1a082ef9a790a4a5820a431452524408a9d4229d6abc548c5e6efb24)

Error opening media file buffer.
open .git/lfs/objects/90/bd/90bd3def1a082ef9a790a4a5820a431452524408a9d4229d6abc548c5e6efb24-temp: file exists
goroutine 1 [running]:
github.com/github/git-lfs/lfs.Stack(0x0, 0x0, 0x0)
        /tmp/git-lfs/src/github.com/github/git-lfs/lfs/errors.go:85 +0x8d
github.com/github/git-lfs/lfs.Errorf(0x7fa8f3bfb598, 0xc2081d5f20, 0x8335d0, 0x20, 0x0, 0x0, 0x0, 0x4f)
        /tmp/git-lfs/src/github.com/github/git-lfs/lfs/errors.go:29 +0x5e
github.com/github/git-lfs/lfs.downloadFile(0x7fa8f3bf9c58, 0xc20802e008, 0xc20803c3c0, 0x7fffbec09550, 0x38, 0xc208050e40, 0x57, 0x0, 0x0)
[...]

PR #763, opened 2015-10-20
[wip] rewrite the atomic file writes

technoweenie:
This is an attempt to solve #266. If `git-lfs` is exited while downloading, it leaves some `.git/lfs/objects/../../OID-temp` files, causing future downloads (`smudge`, `fetch`, or `pull`) to error out until those temp files are deleted.

This used to work using my [contentaddressable package](https://github.com/technoweenie/go-contentaddressable) which used the `O_EXCL` file mode to ensure only one process writes to a file at a time. This breaks if the process exits without letting Go's `defer` clean the file up.

This PR changes a few things. First, it drops the contentaddressable package in favor of the [atomic package](https://github.com/natefinch/atomic). While `os.Rename()` works as an atomic operation on posix systems, it doesn't on Windows. The atomic package uses Windows APIs to provide the same atomic guarantee in `atomic.ReplaceFile()`.

It also throws all temp object files into `.git/lfs/tmp/objects`, so it's easy for `git-lfs` to clean the files up on exit. Multiple `git-lfs` processes will have their own sets of temp files. It's possible that multiple `git-lfs fetch` operations will download a file multiple times. But, it should only move the file to it's final spot (`.git/lfs/objects/../../OID`) IF the temp file is written successfully and the SHA-256 signature of the contents matches the expected OID.

This SHOULD work fine on linux/mac, but needs a bit more work for Windows support.

TODO:

* [x] drop `contentaddressable` package
* [x] Add `atomic` package, replace `os.Rename()` usage with `atomic.ReplaceFile()`.
* [x] have git-lfs cleanup old temp files automatically on exit.


sinbad:
Wow, I had no idea `os.Rename()` wasn't atomic on Windows, Go should really fix that & use the native calls the atomic package does. 

If you wanted to do cleanup even when someone hit Ctrl+C you could use `os.Signal` to catch SIGINT?

rubyist:
`os.Rename` isn't always atomic on POSIX, e.g. when moving across file systems or using NFS, if the system crashes during the operation, and probably some other oddball things. I think the go authors have explicitly suggested to not rely on its atomicity even though it _can_ be on some systems.

> If you wanted to do cleanup even when someone hit Ctrl+C you could use os.Signal to catch SIGINT?

:+1: I think we should definitely catch any signals we can and make sure we clean up whenever we can. I think that's the real heart of the issue over the atomic operations (though I'm :+1: on providing the atomic operations, too).

rubyist:
It looks like in Go 1.5, `os.Rename` uses `MoveFileEX`, which is what this atomic package does. https://github.com/golang/go/commit/92c57363e0b4d193c4324e2af6902fe56b7524a0

If we're sticking to 1.5+ we could drop the extra dep.

technoweenie:
> If you wanted to do cleanup even when someone hit Ctrl+C you could use os.Signal to catch SIGINT?

That's fine. Anything that isn't caught when Git LFS crashes can be picked up on its next run.

> when moving across file systems or using NFS, if the system crashes during the operation, and probably some other oddball things.

Hopefully no one sets up their local repositories like this :)

> It looks like in Go 1.5, os.Rename uses MoveFileEX, which is what this atomic package does.

Oh niiice.

lfs/lfs.go
```diff
@@ -28,6 +28,7 @@ var (
 	LocalGitDir        string // parent of index / config / hooks etc
 	LocalGitStorageDir string // parent of objects/lfs (may be same as LocalGitDir but may not)
 	LocalMediaDir      string // root of lfs objects
+	LocalObjectTempDir string // where temporariliy downloading objects are stored
```
WillHipschman: nit: typo, should be temporarily

technoweenie: You should've seen what it was before I committed.

technoweenie:
This now cleans the files up whenever `git-lfs` exits if the temp files match the following conditions:

1. The temp file doesn't match the `{oid}-{rand}` pattern.
2. The OID exists at `.git/lfs/objects`, meaning something has successfully downloaded it.
3. The temp file hasn't been modified in 1 hour. This should mean that no other `git-lfs` still has its greedy paws on it.

I didn't specify any signals so that it would notify on any of them. Go only supports `os.Interrupt` and `os.Kill` cross platform. Since this isn't a long running process, I don't expect users to be sending signals to Git LFS at all.

rubyist:
Ctrl-c on the command line is the primary signal we should catch. That's SIGINT, which should work on osx/Linux/Windows.

rubyist:
Oh, I see, you're just doing it for all of them, I misread. That's cool :+1:

technoweenie:
Turns out it's nice to actually exit when the Interrupt signal is received. Otherwise it just keeps on downloading :)

I pushed some changes, and added a stderr msg. 

```
Fetching objects...
(0 of 1 files) 99.02 MB / 340.56 MB                                                                                                                                                                                                                                                                                                                                        ^C
Exiting because of "interrupt" signal.
```

git-lfs.go
```diff
@@ -1,11 +1,37 @@
...
+		for {
+			sig := <-c
+			once.Do(lfs.ClearTempObjects)
+			switch sig {
+			case os.Interrupt, os.Kill:
+				fmt.Fprintf(os.Stderr, "\nExiting because of %q signal.\n", sig)
+				if commands.Erroring {
+					os.Exit(2)
+				}
+				os.Exit(0)
```
sinbad: POSIX states that the exit code when the process is terminated by a signal must be > 128 (128+n), so I think "Erroring" is redundant and we should just return 128 + signal number. Windows doesn't define this but since the environment is Git Bash adhering to POSIX makes sense. Either way a forcibly terminated process should never return 0.

git-lfs.go
```diff
@@ -1,11 +1,37 @@
...
 func main() {
+	c := make(chan os.Signal)
+	signal.Notify(c)
+
+	var once sync.Once
+
+	go func() {
+		for {
+			sig := <-c
+			once.Do(lfs.ClearTempObjects)
```
sinbad: This clears temporary objects on *any* signal, including non-terminating signals like SIGSTOP, SIGCONT and SIGTRAP (the latter would kill the temp dir when a debugger was attached ;) ). Cleanup should only happen on default terminating signals, list here: https://en.wikipedia.org/wiki/Unix_signal#POSIX_signals

git-lfs.go
```diff
@@ -1,11 +1,37 @@
...
+	"os/signal"
+	"sync"
+
 	"github.com/github/git-lfs/commands"
 	"github.com/github/git-lfs/lfs"
 )
 
 func main() {
+	c := make(chan os.Signal)
+	signal.Notify(c)
```
sinbad: I think you should actually specify the types i.e. `signal.Notify(c, os.Interrupt, os.Kill)` to avoid overrunning the channel buffer, the docs say: 
> Package signal will not block sending to c: the caller must ensure that c has sufficient buffer space to keep up with the expected signal rate. For a channel used for notification of just one signal value, a buffer of size 1 is sufficient.
Kill/Interrupt are OK with a single buffer value because they're only sent once & then we terminate, but other non-terminating signals can get sent more frequently. Sounds like they might either overwrite or overrun a single-entry channel in that case.

git-lfs.go
sinbad: Hmm, seems Go only exposes SIGINT and SIGKILL for filtering. I think if you want to handle more types you have to use the syscall package which will be less portable - you'd also have to make the signal channel large enough to handle the non-blocking calls for other signals. Maybe it's best just to handle those 2 in the Notify().

lfs/localstorage.go
```diff
@@ -0,0 +1,46 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/github/git-lfs/vendor/_nuts/github.com/rubyist/tracerx"
+)
+
+func ClearTempObjects() {
+	filepath.Walk(LocalObjectTempDir, func(path string, info os.FileInfo, err error) error {
```
sinbad: `filepath.Walk` isn't the fastest of ways to scan a directory because it does it in lexical order. Given that it's just one directory, `Readdirnames` is the fastest method, although you need `Readdir` if you also need the `FileInfo`.

lfs/localstorage.go
```diff
@@ -0,0 +1,46 @@
...
+		tracerx.Printf("Removing invalid tmp object file: %s", path)
+		return true
+	}
+
+	if FileExists(localMediaPathNoCreate(oid)) {
+		tracerx.Printf("Removing existing tmp object file: %s", path)
+		return true
+	}
+
+	if time.Since(info.ModTime()) > time.Hour {
```
sinbad: Why not just delete everything? Is this to try to avoid parallel `git-lfs` instances trampling over each other? Locks would normally prevent that anyway, except for maybe that sliver of time between finishing writing and moving a file to its final location.

lfs/localstorage.go
```diff
@@ -0,0 +1,46 @@
...
+	"strings"
+	"time"
+
+	"github.com/github/git-lfs/vendor/_nuts/github.com/rubyist/tracerx"
+)
+
+func ClearTempObjects() {
+	filepath.Walk(LocalObjectTempDir, func(path string, info os.FileInfo, err error) error {
+		if shouldDeleteTempObject(path, info) {
+			return os.RemoveAll(path)
```
sinbad: I think you should always return no error, otherwise a failure to delete one file will stop deletion of the rest. Stray locks on files can cause issues like this (especially on Windows). Maybe just trace any errors.

sinbad:
:+1: on the general approach, just had a few detail issues. Great to hear that go 1.5 is using MoveFileEx now.

rubyist:
I upvote all of @sinbad's feedback, particularly the signal handing stuff. Solid points.

lfs/localstorage.go
technoweenie: Yeah, I'd like to investigate locks in a future version.

technoweenie:
Thanks for the feedback. I think I incorporated all of it.

git-lfs.go
```diff
@@ -1,11 +1,37 @@
...
+
+	var once sync.Once
+
+	go func() {
+		for {
+			sig := <-c
+			once.Do(lfs.ClearTempObjects)
+			fmt.Fprintf(os.Stderr, "\nExiting because of %q signal.\n", sig)
+
+			var exitCode int
```
sinbad: Being super pedantic signal exits are supposed to be 129+ (signal value is never 0) so to avoid that if the cast below fails, better to initialise exitCode to a non-zero value.

lfs/localstorage.go
sinbad: I meant the write locks you already have will prevent the deletion unless you catch it in between the close & rename (unlikely). But this was the least important of my comments.

technoweenie: I tried that approach, but I don't think my locks were doing what I thought they were. The contentaddressable package opens them with these flags: `os.O_WRONLY|os.O_CREATE|os.O_EXCL`. Turns out `O_EXCL` + `O_CREATE` only confirm the file didn't exist previously. `os.RemoveAll` happily deleted it.

Cross platform locking seemed too complicated for v1.0.1, so I went with this workaround. Considering that temp files will be removed most of the time, it seems like a fair tradeoff.

technoweenie:
Seems to work fine on windows too. :+1:

Issue #759, opened 2015-10-20
global lfs.fetchexclude results in error accessing media


gogodave:
If you define a global lfs.fetchexclude with something like:
`git config --global lfs.fetchexclude dir1`

and dir1 contains a tracked file, then a clone of the repo results in "Error accessing media".

git lfs logs last reports this:

    git-lfs/1.0.0 (GitHub; linux amd64; go 1.5.1; git 7379fa9)
    git version 2.4.8

    $ git-lfs smudge test-sub2/f3.xz
    Error accessing media: test-sub2/f3.xz (edeaaff3f1774ad2888673770c6d64097e391bc362d7d6fb34982ddf0efd18cb)

    File missing and download is not allowed
    goroutine 1 [running]:
    github.com/github/git-lfs/lfs.Stack(0x0, 0x0, 0x0)
            /tmp/docker_run/git-lfs/rpm/BUILD/git-lfs-1.0.0/src/github.com/github/git-lfs/lfs/errors.go:557 +0x80
    github.com/github/git-lfs/commands.logPanicToWriter(0x7f413b8f91e8, 0xc820046160, 0x7f413b8f9a78, 0xc820011eb0)
            /tmp/docker_run/git-lfs/rpm/BUILD/git-lfs-1.0.0/src/github.com/github/git-lfs/commands/commands.go:184 +0xf7f
    github.com/github/git-lfs/commands.logPanic(0x7f413b8f9a78, 0xc820011eb0, 0x0, 0x0)
            /tmp/docker_run/git-lfs/rpm/BUILD/git-lfs-1.0.0/src/github.com/github/git-lfs/commands/commands.go:148 +0x421
    github.com/github/git-lfs/commands.handlePanic(0x7f413b8f9a78, 0xc820011eb0, 0x0, 0x0)
            /tmp/docker_run/git-lfs/rpm/BUILD/git-lfs-1.0.0/src/github.com/github/git-lfs/commands/commands.go:123 +0x4e
    github.com/github/git-lfs/commands.LoggedError(0x7f413b8f9a78, 0xc820011eb0, 0x8deee0, 0x1e, 0xc820203d00, 0x2, 0x2)
            /tmp/docker_run/git-lfs/rpm/BUILD/git-lfs-1.0.0/src/github.com/github/git-lfs/commands/commands.go:73 +0x82
    github.com/github/git-lfs/commands.smudgeCommand(0xab4080, 0xc820011980, 0x1, 0x1)
[...]

PR #770, opened 2015-10-22
Fix clone fail when fetch is excluded globally (#759)

sinbad:
Smudge command should return 0 if download was optional and didn't occur


technoweenie:
:+1:

PR #771, opened 2015-10-22
Make integration tests work on Windows

sinbad:


technoweenie:
Epic :clap:

technoweenie:
This is really close. The env tests fail because something is converting the `\t` in the paths (such as my home dir: `c:\Users\techn...`) to tabs.

technoweenie:
Sorry if you prefer the mystery approach to debugging, but here are some spoilers:

```
test: env with bare repo ...                                       FAILED
-- stdout --
    Initialized empty Git repository in C:/Users/techn/git-lfs-tests/test-env.sh-5768/env-with-bare-repo/
    ASSERT
    --- a.txt   2015-10-22 18:36:51.247455400 -0700
    +++ b.txt   2015-10-22 18:36:51.346460800 -0700
    @@ -3,7 +3,7 @@
     git version 2.5.1.windows.1
    -GIT_LFS_TEST_DIR=/c/Users/techn/git-lfs-tests
    +GIT_LFS_TEST_DIR=C:/Users/techn/git-lfs-tests
     git-lfs/1.0.0 (GitHub; windows amd64; go 1.5.1; git 6fad0c7)
     LocalGitDir=C:\Users\techn\git-lfs-tests\test-env.sh-5768\env-with-bare-repo
     LocalGitStorageDir=C:\Users\techn\git-lfs-tests\test-env.sh-5768\env-with-bare-repo
```

I changed `contains_same_elements` to use `diff`, making the results readable in stdout. Turns out the problem comes from my own GIT_LFS_TEST_DIR setting of `/c/Users/techn/git-lfs-tests`. However, it doesn't seem to work in Git Bash if I use `c:\Users\techn\git-lfs-tests` instead.

    test/testlib.sh: line 78: c:Userstechngit-lfs-tests/test-env.sh-3432/out: No such file or directory

technoweenie:
BTW I made my commits in a dummy branch. I like the idea of using `diff` for these long string comparisons. It also fixes the `\t` issue by using a `'%s'` arg for `printf`.

https://github.com/github/git-lfs/commits/sinbad-windows-tests

technoweenie:
One thought I had is that Git LFS could detect whether it's in a git bash env. If `$SHELL` is set (especially to something like `/Git/usr/bin/*sh\z`), then I think we can assume it's Git Bash. That env is empty in Git cmd. Then, the paths could all be be output as the unixified paths, like `/c/Users/...`.

That sounds like kind of a headache. The test suite would have to know about all this too.

Since a lot of this seems to be coming from `git lfs env`, we could even add some flags to disable ENV output or something.

sinbad:
> This is really close. The env tests fail because something is converting the \t in the paths 

Aaargh, I fixed loads of those cases but obviously missed a couple because they didn't blow up for me, nice catch with the `'%s'`.

>  I like the idea of using diff for these long string comparisons.

Why didn't I think of that instead of copy/pasting the terminal output and staring at it? Nice.

I'll have a poke around with `GIT_LFS_TEST_DIR` today.

sinbad:
I've ported the `diff` approach from your branch, I just redirected the output to `stderr` so it was easier to find in the output (next to the test failure). 

The workaround for `GIT_LFS_TEST_DIR ` is to use the form:
```
 export GIT_LFS_TEST_DIR=C:\\Users\\steve\\temp\\git-lfs-test-dir
```
That works fine in Git Bash for me and the tests pass now with `GIT_LFS_TEST_DIR ` (with one other small fix which is on this branch)

As for detecting bash & using different paths, I think it's simplest to just make the rule that git-lfs always deals in native paths? All we then need to do is be aware that some data from either env vars or git may come in with bash-style paths and we may need to clean it up before using it internally, or passing it back to git (we had this in SourceTree a lot). Luckily `filepath` does a lot of the heavy lifting internally and any external scripts can use `cygpath` if they know that output from git-lfs will be native and they want Bash (although my experience is that most of the time Windows tools would rather have native paths). 

I don't think it's too onerous to require that `GIT_LFS_TEST_DIR` is set to a native path (with escaping) just for the tests. I *could* write something which runs `cygpath` on every environment variable in the tests (it can be made safe for non-paths) but I think that's overkill.

technoweenie:
> some data from either env vars or git may come in with bash-style paths and we may need to clean it up before using it internally, or passing it back to git

So Git uses native (`c:\` ?) paths? I'm fine with using native paths on windows.

> I don't think it's too onerous to require that `GIT_LFS_TEST_DIR` is set to a native path

Agreed.

sinbad:
> So Git uses native (c:\ ?) paths? I'm fine with using native paths on windows.

No, you get paths in bash-style but Go's `filepath` can cope with those as well as native paths. Also when you pass paths to git it can handle native paths if they're relative (but not absolute paths because it thinks they're outside the repo - must do a string test). So the main thing is to be careful about using string tests on paths unless you clean them first - this is why `lfs.FilenamePassesIncludeExcludeFilter` uses `filepath.Clean` before doing any comparisons.

It's just easier to use native paths inside Go generally since they'll all behave the same way, you just have to be aware of the potential for non-native paths coming in from Git, which will still work anyway inside Go but will throw off string comparisons.

technoweenie:
Huzzah, works on my vm too :+1:

PR #784, opened 2015-10-24
Appveyor Proof of concept

andyneff:
First working version producing an unsigned git-lfs.exe

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,56 @@
+version: 1.0.{build}
+init:
+- cmd: >-
+    set REPO_DIR=%cd%
+
+
+    mkdir %BASHROOT%
+
+    cd %BASHROOT%
+
+
+    curl -LOk https://github.com/git-for-windows/git/releases/download/v2.6.2.windows.1/PortableGit-2.6.2-64-bit.7z.exe
+
+
+    7z x PortableGit-2.6.2-64-bit.7z.exe > nul
+
+
+    mkdir c:\go15
+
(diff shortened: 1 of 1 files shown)


technoweenie:
Cool, let's try it.

daryllabar:
Does this mean that LFS is compatible with Appveyor?  Is there anything special that I has to be done?

technoweenie:
I couldn't get it to work, unfortunately. Would love help with it though!

daryllabar:
Was this just an attempt to get Appveyor to be able to download the files itself?

andyneff:
This was an attempt to get appveyor to compile and the lfs executable for CI purposes.

I'm not aware of not being able to use git lfs in an appveyor itself. I would imagine all you REALLY need to do is

1. Download the [git-lfs.exe](https://github.com/github/git-lfs/releases/download/v1.3.1/git-lfs-windows-amd64-1.3.1.zip) file and unzip (I'm guessing something like [this](http://www.howtogeek.com/tips/how-to-extract-zip-files-using-powershell/) would help with that, I'm *not* a powershell genius)
2. Put git-lfs.exe in your path (dealers choice on how)
3. Run `git lfs install`

And then you can use git lfs to do everything you need (`git clone`, `git fetch`, etc...)

daryllabar:
Here is [how to download a file using AppVeyor](http://www.appveyor.com/docs/how-to/download-file).  I'm guessing using on of those apporaches to download the git-lfs.exe would work just fine.

PR #788, opened 2015-10-26
Allow lfs.fetchinclude and lfs.fetchexclude in .gitconfig

technoweenie:
The `lfs.fetchinclude` and `lfs.fetchexclude` keys seem save to read from the working directory's `.gitconfig` file. /cc @sinbad @ptoomey3 

Docs: https://github.com/github/git-lfs/blob/master/docs/man/git-lfs-fetch.1.ronn


ptoomey3:
Seems sane.

sinbad:
I doubt many people will actually do this because including / excluding is typically a personal workflow thing (different people in the team use different sets of files), but it's safe to do it.

PR #790, opened 2015-10-26
rewrite test runner in go to reduce output race conditions

technoweenie:
Replace 3 highly compact shell lines in `script/integration` with 100 lines of go code. It uses goroutines to manage the concurrency, and a channel to ensure output is dumped to stdout without race conditions.

/cc @andyneff @rubyist @sinbad


technoweenie:
This cleans up the output and now passes in the docker tests for me :metal:

PR #807, opened 2015-10-31
Docker and rpm updates

andyneff:
- `run_dockers.bsh` can now auto pull the latest docker images
- More verbose (and un cut) output of 64 bit rpm builds build. No more redirect to `build.log`
- RPMs use GIT_LFS_TEST_DIR in %check


technoweenie:
Thanks! Finally got my docker issues resolved, so I could confirm this all works.

Issue #789, opened 2015-10-26
Silent installation option for git-lfs Windows installer


FeodorFitsner:
Hello,

Windows installer is made with NSIS however it doesn't respect `/S` (silent) option for unattended installation ([this](https://github.com/github/git-lfs/blob/master/script/nsis/installer.nsi#L41) should be avoided in silent mode too). That would be especially useful for headless CI environment such as AppVeyor.

Thank you!

technoweenie:
Would it help to have a zip of just `git-lfs.exe`? I'm thinking about ditching nsis in a future release (maybe v1.1).

FeodorFitsner:
Sure, zip could be a great addition to MSI for "advanced" installations (when you know where to unpack the zip and what to do next).

technoweenie:
Just merged https://github.com/github/git-lfs/pull/812. Should work better for Git LFS 1.1.0.

PR #812, opened 2015-11-03
Update nsis to repsect silent mode #789

lbarbisan-ullink:


technoweenie:
Thanks! Now that I know that silent mode installs on windows are a thing, I'll make sure we keep the feature around if we move to something else like [InnoSetup](http://www.jrsoftware.org/isinfo.php).

pascalberger:
@lbarbisan-ullink I tried to run `git-lfs-windows-amd64-1.1.0.exe` in silent mode using `/s` parameter, but it doesn't seem to make a difference (license & installation folder dialog are still shown). How is silent mode supposed to work

lbarbisan-ullink:
`/S` in uppercase ?

pascalberger:
Thanks! Uppercase is the key :+1:

jed-frey:
If anyone finds this through Google (as I did) the command is now /SILENT & /VERYSILENT (I couldn't get /s or /S to work)

image

You can also do an install once and record the install settings and play them back:

image

tolache:
Can it be installed on Windows Server Core?

bk2204:
I don't know the answer to that, because it's not an environment we test.  You're welcome to try, though.

PR #816, opened 2015-11-05
Evaluate symlinks in directory path names

### LINKED ISSUE #811: git lfs track fails when git dirpath contains symlinks
git lfs track fails with `a.gif is outside repository` if current working dir(=gitdir) contains symlinks.

Steps to reproduce.

``` bash
$ mkdir /tmp/real
$ cd /tmp/
$ ln -sf real sym
$ cd sym
$ git init . ; git lfs init
# copy any binary file to this dir
$ git lfs track a.gif
a.gif is outside repositor
```

git lfs env output

```
git-lfs/1.0.2 (GitHub; linux amd64; go 1.5.1; git 9a9ab0386672)
git version 2.6.1

LocalWorkingDir=/tmp/real
LocalGitDir=/tmp/sym/.git
LocalGitStorageDir=/tmp/sym/.git
LocalMediaDir=/tmp/sym/.git/lfs/objects
TempDir=/tmp/sym/.git/lfs/tmp
ConcurrentTransfers=3
BatchTransfer=true
```

difro:
This will normalize all directories for proper comparison.

This fixes #811


lfs/lfs.go
```diff
@@ -101,6 +101,12 @@ func ResolveDirs() {
 	var err error
 	LocalWorkingDir, LocalGitDir, err = resolveGitDir()
 	if err == nil {
+		if LocalWorkingDir, err = filepath.EvalSymlinks(LocalWorkingDir); err != nil {
+			panic(fmt.Errorf("Error trying to evalsymlink '%s': %s", LocalWorkingDir, err))
```
technoweenie: A few small things: 

* There's no need to pass an error to panic. `fmt.Sprintf()` would be fine.
* You can use `%q` instead of `%s` to print a quoted string.
* "evalsymlink" in the error message should probably be split out.

difro: Was just mindlessly copy & paste'ing error-check+panic code from few lines below.
But I see your points.
Will push new commit soon.

PR #818, opened 2015-11-06
Fix "foo is outside repository" error when "git lfs track" in a symlinked dir

sinbad:
To reproduce, run this with an unmatched git lfs:
```bash
git init track-symlinkdst
ln -s track-symlinkdst track-symlinksrc
cd track-symlinksrc
git lfs track "*.png"
```

This was accidentally caused by the move to use `git rev-parse --top-level` to find the working dir. While more reliable than previous manual methods, git also silently resolves symlinks in the returned path, meaning that `lfs.LocalWorkingDir` is always the real path and not the symlink. This causes `HasPrefix` tests to fail unless you also check the resolved symlink version.


jiangxin:
Today I encounter the same issue, but has slight different solutions, see #822

PR #820, opened 2015-11-06
NTLM Auth attempt 2

technoweenie:
This makes some changes to #714:

* Moved some `!Config.NtlmAccess()` checks to lower level helper functions, so that every caller of these functions doesn't have to also check for ntml access.
* Use a standard tracer message for `toggleAuthType()` so every caller doesn't have to define one.
* Removed some ntlm specific helpers like `isNtlmRequest()`. Trying to remove as much ntlm special casing as possible.
* Remove some spots where http response bodies are closed multiple times.
* Automatically set `ConcurrentTransfers()` to 1 if ntlm authentication is used, so users aren't required to set this value themselves. 
* Removed some redundant internal helper functions in favor of the stdlib (`ioutil.NopCloser()` is pretty cool, for example).

@WillHipschman: Normally I'd just leave feedback for you to fix. But I didn't because the diff on #714 includes some changes in master that were made after the PR was opened. It made it really difficult to review. Seems like a GitHub PR issue, and nothing to do with your actual work. 

There is still an `ioutil.ReadAll()` when cloning an http request, though. If NTLM auth is used for any file requests (as opposed to just LFS API requests), it should definitely buffer to a file instead. Maybe buffer up to 1MB in memory, and put the overflow in a temp file.


technoweenie:
Trying to remove the `ioutil.ReadAll()` call was more difficult than I thought. It's in #821 by itself.

WillHipschman:
This looks good to me. Thanks for the changes :+1:

PR #821, opened 2015-11-06
NTLM: cloneable body

technoweenie:
This is a proposal to merge into #820. It's complicated enough that I think it could use its own review.

Instead of buffering the entire request body into an in-memory byte buffer, this uses an internal `cloneableBody` type that only buffers up to 1MB. Any overflow is then written to a local temp file. This type can also clone itself without reading the body again. It simply creates a new struct with the same 1MB byte buffer, and re-opens the temp file.

Also, assuming all file handles are properly closed, a `cloneableBody` buffer will only delete the tempfile when the last one is closed.

### DIFF
--- a/lfs/ntlm.go
+++ b/lfs/ntlm.go
@@ -173,12 +173,12 @@ func cloneRequestBody(req *http.Request) (io.ReadCloser, error) {
 
 	var cb *cloneableBody
 	var err error
-	hasExisting := false
+	isCloneableBody := true
 
 	// check to see if the request body is already a cloneableBody
 	body := req.Body
 	if existingCb, ok := body.(*cloneableBody); ok {
-		hasExisting = true
+		isCloneableBody = false
 		cb, err = existingCb.CloneBody()
 	} else {
 		cb, err = newCloneableBody(req.Body, 0)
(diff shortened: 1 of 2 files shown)


lfs/ntlm.go
```diff
@@ -169,13 +171,147 @@ func cloneRequestBody(req *http.Request) (io.ReadCloser, error) {
...
+		}
+
+		req.Body = cb2
+	}
+
+	return cb, nil
+}
+
+type cloneableBody struct {
+	b      []byte    // in-memory buffer of body
```
WillHipschman: I would change b to body, f to file, etc. It makes reading the code easier :)

lfs/ntlm.go
```diff
@@ -169,13 +171,147 @@ func cloneRequestBody(req *http.Request) (io.ReadCloser, error) {
...
+	}
+
 	if err != nil {
 		return nil, err
 	}
 
-	req.Body = ioutil.NopCloser(bytes.NewBuffer(buf))
-	return ioutil.NopCloser(bytes.NewBuffer(buf)), nil
+	// if the request body is NOT a cloneable body, replace it with one
+	if !hasExisting {
```
WillHipschman: I would rename hasExisting to isCloneableBody and remove the comment.

WillHipschman:
Other than some variable names this looks awesome.

lfs/ntlm.go
technoweenie: Some internal go code uses short vars like this. The go community seems to favor shorter names like that. But I think I learn more towards your opinion on this :)

PR #828, opened 2015-11-11
env: Show proper SshUserAndHost for remoteEndpoint

jiangxin:
If there are multiple remotes for the current repo, `git lfs env` shows wrong SshUserAndHost for remote endpoints.


technoweenie:
Yikes, nice catch :)

technoweenie:
Thanks for the patch!  I confirmed this works in this test: https://github.com/github/git-lfs/commit/b5fa32f7aa2cf11e6e8a8248421653511b47800a

Issue #827, opened 2015-11-11
With 'smudge --skip', if the file exists in cache use it?


ajohnson23:
TLDR: I like running with --skip by default to prevent the slow serial downloads, but it's inconvenient to have to always run 'git lfs pull' even when the objects are already cached locally. Please consider either changing the behavior of --skip or adding another flag that avoids *downloads* during smudge (outputting the placeholder in this case), but still smudges if the object is available in cache.

Long version:
Currently, with smudge --skip, you'll still get the lfs placeholder file even if you already had the file in your .git/lfs/objects cache.

It's mainly the *download* that hurts me in terms of wait time, so I'd prefer if there were a way to skip just downloading during smudge, but go ahead and do the smudge if the file already exists in the .git/lfs/objects cache. This way, if i do a 'git checkout tree-ish' and the lfs objects are already cached locally, they'll be smudged in. For files that aren't available in cache, I should still get a placeholder.

As far as I'm concerned the behavior of smudge --skip could just be changed to work this way, but if there's some reason for having a way to totally disable smudging (even when the file is cached), another flag such as --skip-downloads could be added.

ajohnson23:
I opened a pull request for this, #831

ajohnson23:
Merged a fix, closing this.

PR #831, opened 2015-11-12
smudge --skip will now smudge if object is in cache.

ajohnson23:
This change makes it so --skip does the smudge if the file
is in cache. It still skips automatic downloads as one would
expect.

This is more convenient, since it prevents one from having
to 'git lfs pull' after every 'git checkout' if the needed
objects are already available in .git/lfs/objects.

Note that the --skip option for smudge is documented as
follows:
"Skip automatic downloading of objects on clone or pull."
So even after this change we're still in line with the
documentation.

### DIFF
--- a/commands/command_smudge.go
+++ b/commands/command_smudge.go
@@ -23,14 +23,6 @@ func smudgeCommand(cmd *cobra.Command, args []string) {
 	requireStdin("This command should be run by the Git 'smudge' filter")
 	lfs.InstallHooks(false)
 
-	if smudgeSkip || lfs.Config.GetenvBool("GIT_LFS_SKIP_SMUDGE", false) {
-		_, err := io.Copy(os.Stdout, os.Stdin)
-		if err != nil {
-			Panic(err, "Error writing data to stdout:")
-		}
-		return
-	}
-
 	// keeps the initial buffer from lfs.DecodePointer
 	b := &bytes.Buffer{}
 	r := io.TeeReader(os.Stdin, b)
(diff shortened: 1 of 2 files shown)


ajohnson23:
Also see #827

technoweenie:
> This is more convenient, since it prevents one from having
to 'git lfs pull' after every 'git checkout' if the needed
objects are already available in .git/lfs/objects.

This feels like a weird feature. Though convenient, how are you supposed to know that every file was checked out properly?  Even with this feature, I think I'd still want to run `git lfs pull` anyway just to be sure something wasn't missed.

ajohnson23:
Not weird at all! :) There are many reasons, I think.

Maybe you had run git lfs pull --all in prep for a long stretch of offline work or a bisect with a huge range.

Maybe you're just aware you had just checked out that branch/revision so you know you were cached.

But the most common reason is that maybe you're the only person checking in largefiles because that is your role within the project. If you're the only one checking in the largefiles, it's guaranteed that you'll have them cached. This is the case for myself, and I don't think it's that unusual -- I'm basically the "build manager", so I'm the only one checking in dependency .so/.a/.dll files.

Plus, how is this feature harmful?

By the way, this is how git-fat worked and coming to lfs, it felt very weird that it didn't work this way. There are many cases where I 100% know I have large files cached (as mentioned above) and having to lfs pull is an annoyance.

technoweenie:
I'll merge this, but we'll keep recommending that people run `git lfs pull` with the skip-smudge setting. I would argue that your use cases are not common at all, unless you know how Git LFS works really well. Most people aren't, and should be running `git lfs pull` (or better yet, baking that into a script or something).

technoweenie:
Oh and thanks for the patch BTW :metal:

ajohnson23:
Thanks for the merge. Agreed that skip is not the best default.

That said, it'd be great to find a way to speed up the pulls when they're happening post checkout without the skip option. Here's hoping git lfs popularity encourages some smudge/clean api changes to git to allow for this!

PR #837, opened 2015-11-16
use .lfsconfig (with fallback to .gitconfig)

### LINKED ISSUE #780: Use .lfsconfig instead of .gitconfig
This change would make it clear that the file is used for lfs configuration only. Though we'd want to keep supporting `.gitconfig`, and just print a warning asking them to change it to `.lfsconfig`.

technoweenie:
This changes Git LFS to read committed lfs config settings from `.lfsconfig` instead of `.gitconfig`. This makes it more clear what settings should go in that file. It does continue to check for `.gitconfig` but only IF `.lfsconfig` does not exist. This is meant as a way for projects to gradually rename this file on their own. By Git LFS v2.0, we'll remove support completely.

This also makes a few other changes:

* `ListFromFile()` doesn't run `os.Stat` anymore. The new `readGitConfigFromFiles()` function minimizes `os.Stat` calls.
* The config order was tweaked slightly, reducing the number of `git config` calls. No need to call `git config --list` AND `git config --file .git/config --list`. Now it only calls `git config --file .lfsconfig --list` and `git config --list`.

Fixes #780

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -280,9 +280,6 @@ func (c *gitConfig) List() (string, error) {
 
 // ListFromFile lists all of the git config values in the given config file
 func (c *gitConfig) ListFromFile(f string) (string, error) {
-	if _, err := os.Stat(f); os.IsNotExist(err) {
-		return "", nil
-	}
 	return simpleExec("git", "config", "-l", "-f", f)
 }
 
--- a/lfs/config.go
+++ b/lfs/config.go
@@ -360,6 +360,12 @@ func (c *Configuration) readGitConfigFromFiles(filenames []string, filenameIndex
 	filename := filenames[filenameIndex]
 	_, err := os.Stat(filename)
 	if err == nil {
+		if filenameIndex > 0 && ShowConfigWarnings {
+			expected := ".lfsconfig"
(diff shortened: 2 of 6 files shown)


lfs/config.go
```diff
@@ -327,28 +328,22 @@ func (c *Configuration) loadGitConfig() bool {
...
-		panic(fmt.Errorf("Error listing git config: %s", err))
-	}
+	configFiles := []string{
+		filepath.Join(LocalWorkingDir, ".lfsconfig"),
 
-	configFile := filepath.Join(LocalWorkingDir, ".gitconfig")
-	fileOutput, err := git.Config.ListFromFile(configFile)
-	if err != nil {
-		panic(fmt.Errorf("Error listing git config from %s: %s", configFile, err))
+		// TODO: remove .gitconfig support for Git LFS v2.0
```
WillHipschman: Add the issue # tracking this work so that when we complete the work we know to remove the TODO.

git/git.go
```diff
@@ -280,9 +280,6 @@ func (c *gitConfig) List() (string, error) {
 
 // ListFromFile lists all of the git config values in the given config file
 func (c *gitConfig) ListFromFile(f string) (string, error) {
-	if _, err := os.Stat(f); os.IsNotExist(err) {
-		return "", nil
-	}
 	return simpleExec("git", "config", "-l", "-f", f)
```
khorshedalam: 837

PR #843, opened 2015-11-18
upload ubuntu packages and generate some markdown package download links

technoweenie:
Updated the script to upload packages for Ubuntu, AND generate a bunch markdown links for the releases page to manually download them.

NOTE: I already ran this, so the ship has sailed for v1.1 :)

/cc @ssgelm and @andyneff


technoweenie:
Cool, thanks.

Issue #845, opened 2015-11-19
Add ref information to upload request


dluksza:
Current upload request specification makes it hard to enforce branch based permissions in Gerrit. In order to enforce them Gerrit must first accept all incoming large files, then based on the pushed repository content either discard uploaded objects or keep them.

Such approach leads to waste of user's time, data transfer and server storage.

Adding destination ref information into upload request would solve this issue. Before sending back upload links Gerrit can check user access permissions to given ref, preventing or allowing large object upload.

Example upload request with ref information:

```
> POST https://git-lfs-server.com/objects/batch HTTP/1.1
> Accept: application/vnd.git-lfs+json
> Content-Type: application/vnd.git-lfs+json
> Authorization: Basic ... (if authentication is needed)
>
> {
>   "operation": "upload",
>   "ref": "refs/heads/master",
>   "objects": [
>     {
>       "oid": "1111111",
>       "size": 123
>     }
>   ]
> }
>
```

technoweenie:
:+1: Seems legit. This is something that existing implementations can ignore.

rubyist:
:+1: here too

brabil:
Would be interested in such feature.

technoweenie:
I just added this to my [Git LFS v1.2](https://github.com/github/git-lfs/issues/844) list. Now since it's a  simple addition that shouldn't affect things, it's likely it'll be in the next 1.1.x release. It really depends on how busy I am, and whether anyone else contributes the code instead :) Anyone that comments on this issue will get pinged on the pull request (unless you ask me not to).

ttaylorr:
This was implemented by @technoweenie in https://github.com/git-lfs/git-lfs/issues/2712.

Issue #779, opened 2015-10-23
Refactor API client


technoweenie:
There's a lot of cruft in the `lfs` package for dealing with both LFS APIs. This can make it difficult to follow the code path, and really difficult to add support for alternate protocols.  This is what I'd like to see:
- Simple go interface with methods like `DownloadObjects()` and `UploadObjects()`.
- Public, exported package at `github.com/github/git-lfs/client`.
- Consistent usage in `git lfs fetch`, `git lfs pull`, and `git lfs smudge`.

sinbad:
I've previously (#378) had a go at abstracting the APIs to add SSH as a usable for complete API round-trip. That needs completely revisiting now but were you happy with the general principles? If so I could take this on at the same time as revisiting SSH.

beni05:
On large layouts, such as the landing page of a website or a full-page print ad, the Apple Pay
mark and the Download on the App Store badge can appear on the same layout when each
is placed in a dedicated tile or other area with content that promotes either Apple Pay or an
app. Do not place the Apple Pay mark and the Download on the App Store badge together
in a tile. The Apple Pay mark and Download on the App Store badge can be placed only near
content related to their use.?

technoweenie:
I think your [interface](https://github.com/sinbad/git-lfs/blob/feature/ssh/lfs/api.go#L9-L33) is too big. Of course, this was before the batch API was finalized and all that. But I'd like to see something like:

```go
type ApiContext interface {
	// Get the endpoint this context was constructed from
	Endpoint() Endpoint
	// Close the context & any resources it's using
	Close() error

	Upload(objects []*ObjectResource) ([]*ObjectResource, *WrappedError)
	Download(objects []*ObjectResource) ([]*ObjectResource, *WrappedError)
}
```

I don't know if we'll need `*Check` too. Ideally all of the commands just use this high level interface. I think it'll be nice to get all the http client and batching code out of this catch-all `lfs` package.

sinbad:
Yeah, most of that is because of the previous need to support individual objects as well as batches. Whether we need `*Check` depends on whether this interface is supposed to complete the upload/download or just return resources through which the actual upload/download could subsequently be performed - which is what your interface suggests (it's akin to batch in my previous one). 

[...]

PR #850, opened 2015-11-21
Add metadata headers when uploading blobs

noamt:
Adding the following metadata headers to the upload request:
* Object ID
* Original file name
* Git branch
* Git ref

Adding these headers will benefit smart remote repositories and will help improve BLOB management and tracking.

### DIFF
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -121,7 +121,7 @@ func Download(oid string, size int64) (io.ReadCloser, int64, error) {
 		&ObjectResource{Oid: oid, Size: size},
 	}
 
-	objs, err := Batch(objects, "download")
+	objs, err := Batch(objects, "download", "")
 	if err != nil {
 		if IsNotImplementedError(err) {
 			git.Config.SetLocal("", "lfs.batch", "false")
@@ -207,12 +207,12 @@ func (b *byteCloser) Close() error {
 	return nil
 }
 
-func Batch(objects []*ObjectResource, operation string) ([]*ObjectResource, error) {
+func Batch(objects []*ObjectResource, operation string, gitRev string) ([]*ObjectResource, error) {
 	if len(objects) == 0 {
 		return nil, nil
 	}
 
-	o := map[string]interface{}{"objects": objects, "operation": operation}
+	o := map[string]interface{}{"objects": objects, "operation": operation, "rev": gitRev}
 
 	by, err := json.Marshal(o)
 	if err != nil {
@@ -245,7 +245,7 @@ func Batch(objects []*ObjectResource, operation string) ([]*ObjectResource, erro
 
 		if IsAuthError(err) {
 			setAuthType(res)
-			return Batch(objects, operation)
+			return Batch(objects, operation, gitRev)
 		}
 
 		switch res.StatusCode {
@@ -342,13 +342,6 @@ func UploadObject(o *ObjectResource, cb CopyCallback) error {
 	req, err := o.NewRequest("upload", "PUT")
 	if err != nil {
 		return Error(err)
-	}
-
-	currentRef, err := git.CurrentRef()
-	if err != nil {
-		PrintError(err, "Failed to resolve current git ref while uploading %s", o.Oid)
-	} else {
-		req.Header.Set("X-Lfs-Git-Ref", currentRef.Sha)
 	}
 
 	currentBranch, err := git.CurrentBranch()
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -121,7 +121,7 @@ func Download(oid string, size int64) (io.ReadCloser, int64, error) {
 		&ObjectResource{Oid: oid, Size: size},
 	}
 
-	metadata := newDownloadTransferMetadata()
+	metadata := newTransferMetadata("download", "", "")
 	objs, err := Batch(objects, metadata)
 	if err != nil {
 		if IsNotImplementedError(err) {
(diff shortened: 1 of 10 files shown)


technoweenie:
Nice idea, but I prefer adding it to the batch api json document as suggested in #845. Why would you want this on the storage upload requests, and not the batch API? LFS was designed to use a custom LFS API against a potentially generic storage backend (such as S3), which would ignore the extra headers. Would implementing #845 be enough, or do you _need_ these headers?

Calling `git.CurrentRef()` AND `git.CurrentBranch()` is unnecessary. `CurrentRef()` returns more information. Both functions also invoke `git rev-parse`. We don't want to add 1 or 2 extra redundant `git rev-parse` calls to every file that gets uploaded. Might be better to add extra properties to `*Uploadable` instead, that get passed in somewhere at a higher level.

Super nitpicky: the `X-` prefix on custom headers is unnecessary. I'd probably go with header keys like `Git-Lfs-*`.

lfs/client.go
```diff
@@ -795,3 +820,12 @@ func setErrorHeaderContext(err error, prefix string, head http.Header) {
...
 }
+
+func PrintError(err error, format string, args ...interface{}) {
+	line := format
+	if len(args) > 0 {
+		line = fmt.Sprintf(format, args...)
+	}
+	fmt.Fprintln(ErrorWriter, line)
+	fmt.Fprintln(ErrorWriter, err.Error())
+}
```
technoweenie: This is the type of error collecting thing that the `commands` package should take care of. The `lfs` package is more of a library with functions that should bubble errors up. The `commands` package is the CLI layer that prints and logs errors.

noamt: I understand. My reason for doing this was thinking that the headers are not critical to the operation, so the operation shouldn't fail if any problems come up.

noamt:
Thanks for the feedback, you're right. It's more fitting in the batch API.
This also means that the batch API responses can be smarter in handling blob/path duplication.

I'd still like to have this feature, so if I understand you correctly you'd like to:
* Move this info to the batch API JSON.
* Drop the invocation of ```git.CurrentBranch()```.
* Move the collection of git information up the stack.

Should I close this and open a new pull request?

lfs/client.go
technoweenie: We have a `tracerx` package that logs stuff if you have pass `GIT_TRACE=1` to git lfs. This actually piggy backs on a core git feature, so running with git tracing on will show you all kinds of stuff :)

technoweenie:
> *  Move this info to the batch API JSON.
> * Drop the invocation of git.CurrentBranch().
> * Move the collection of git information up the stack.

:+1: If you could update the batch api docs to go with it, that'd be cool too. Depends how involved you want to get with the project. Also, as you're messing around in the lfs client code, there IS an issue focused on cleaning it up: #779. So, don't hesitate to call out anything gross or weird.

> Should I close this and open a new pull request?

You're welcome to do whatever you're most comfortable with. If it was me, I'd just keep hacking on this PR. 

I appreciate you taking the time to dig in! Feel free to jump in [chat](https://gitter.im/github/git-lfs) if you want to talk. It's a very low volume channel, especially during a US holiday week like this week :)

noamt:
Cheers, I'll get right on it.

noamt:
OK, so I've shuffled things around and now each object of the batch request will contain the original file name and the batch request itself will contain the git revision.

So now I'd like to resolve and pass down the current branch. Assuming I'm at ```commands/command_pre_push.go:72```, what do you think would be the best way to do that?

lfs/client.go
```diff
@@ -204,12 +207,12 @@ func (b *byteCloser) Close() error {
 	return nil
 }
 
-func Batch(objects []*ObjectResource, operation string) ([]*ObjectResource, error) {
```
technoweenie: I wonder if we should introduce some kind struct here instead?

```go
type BatchRequest struct {
  Operation string
  Rev string
}

// example pseudo code
ref, _ := git.CurrentRef()
req := lfs.NewRequest("download", objects)
req.Ref = ref.Name
lfs.Batch(req)
```

technoweenie:
> So now I'd like to resolve and pass down the current branch. Assuming I'm at `commands/command_pre_push.go:72`, what do you think would be the best way to do that?

Probably call `git.CurrentRef()` in the command, and pass it to the `uploadQueue`. Maybe some kind of generic struct to represent the current state of the git repo, like the current ref. Or maybe all it needs is the ref?

Getting the filename may be tricky though. It'd have to come from `lfs.ScanRefs()`, and then passed to the uploadable.

```go
// example
u, _ := lfs.NewUploadable(pointer.Oid, pointer.Name)
// it is possible for a single pointer to match multiple filenames
u.Filenames = pointer.Filenames
```

noamt:
So yeah I've actually got the file name from the pointer and I enter it into each individual object so that's fine.

noamt:
OK I think I nailed it. Added a new struct object ```TransferMetadata``` to the queue instead of polluting all these methods as the amount of data will also probably grow.
If it looks good I'll continue on to add tests and documentation.

noamt:
Done and done

noamt:
Though it's possible I don't think I've come across many instances of that case (local branch name != remote branch name) out in the wild , but I agree that it's better

dluksza:
'ref' property can be used in Gerrit for branch based permissions, it would create a security hole if it is name of local branch not destination branch :)

noamt:
Fixed. Thanks for the feedback @dluksza

PR #860, opened 2015-11-25
Remove git current branch

technoweenie:
Having both `git.CurrentBranch()` and `git.CurrentRef()` seemed redundant.


sinbad:
Catching up with this stuff late, looks good to me :+1:

Issue #778, opened 2015-10-23
Refactor local storage


technoweenie:
We use "local storage" to describe how Git LFS interacts with the `.git/lfs/objects` directory. I'd like to make it a separate go package like `github.com/github/git-lfs/localstorage` with a well thought-out API for any Go developers to use, in addition to the internal `lfs` package.

Responsibilities:
- Reading LFS objects from local storage
- Writing objects to local storage
- Internal lock files
- Cleaning up temp files

@sinbad: curious if you have thoughts on this.

PR #862, opened 2015-11-25
Extract localstorage 1

technoweenie:
This is a quick experiment in extracting some stuff into a proper `localstorage` package. See #778 for details.


lfs/lfs.go
```diff
@@ -29,6 +30,7 @@ var (
 	LocalGitStorageDir string // parent of objects/lfs (may be same as LocalGitDir but may not)
 	LocalMediaDir      string // root of lfs objects
 	LocalObjectTempDir string // where temporarily downloading objects are stored
```
sinbad: I think LocalMediaDir and LocalObjectTempDir should be removed in favour of LocalStorage. Looking through the code I think this identifies a few other places that manual uses of these dirs needs to be abstracted through LocalStorage

lfs/lfs.go
```diff
@@ -29,6 +30,7 @@ var (
...
+	LocalStorage       *localstorage.LocalStorage
```
sinbad: I find it a little confusing that the var is called the same as the type (and package), maybe that's just me but I'd prefer a different name for the var. Not sure what, StorageInst?

localstorage/temp.go
```diff
@@ -10,27 +9,28 @@ import (
...
+		if shouldDeleteTempObject(s, path) {
 			os.RemoveAll(path)
 		}
 	}
+
+	return nil
 }
 
-func shouldDeleteTempObject(path string) bool {
+func shouldDeleteTempObject(s *LocalStorage, path string) bool {
```
sinbad: Maybe make this a LocalStorage method?

sinbad:
Does switching to passing around by-value Object vs *Pointer have any noticeable performance impact?

sinbad:
Looks :+1: to me, just small comments.

lfs/lfs.go
technoweenie: Agreed, this PR was intended to be just a first step. If this PR is on the right track, we can continue with the refactoring.

lfs/lfs.go
technoweenie: Hah, yeah. StorageInst doesn't excite me either though. Maybe `Objects` or `ObjectStorage`?

localstorage/temp.go
technoweenie: Yeah, I suppose. Not sure why I wrote it like this :)

technoweenie:
> Does switching to passing around by-value Object vs *Pointer have any noticeable performance impact?

I doubt it. I think I changed it because the objects are immutable. Do you want to keep using pointer?

lfs/lfs.go
sinbad: Yeah those are better :)

lfs/lfs.go
sinbad: Yeah I think this is the right way to go.

sinbad:
>> Does switching to passing around by-value Object vs *Pointer have any noticeable performance impact?

> I doubt it. I think I changed it because the objects are immutable. Do you want to keep using pointer?

No, I just wondered if you'd benchmarked it. General guidelines are that small / immutable structs are probably best passed by value so it seems sensible to do this, I was just thinking about the cost of the copy operations given that this is the primary data structure and sometimes it's used in bulk. However I had been thinking about the string as 64 bytes that would be copied, but actually in practice because strings are immutable in Go it's actually just a pointer internally anyway so that's only an 8-byte copy (assuming 64-bit arch), same as the pointer to the struct. In that context it's only an extra 8 bytes for the int64 which will not be significant (and pointer dereferencing has a cost anyway). So it's all good.

technoweenie:
I'm not in love with `lfs.Objects`, but it'll work. Should I remove the prefix on methods like `lfs.Objects.ObjectPath()`?

sinbad:
> Should I remove the prefix on methods like `lfs.Objects.ObjectPath()`?

You mean to `lfs.Objects.Path()`? From the perspective of the LocalStorage object, being called ObjectPath isn't duplication so it's probably OK as-is, but maybe for the purposes of other packages using `lfs`, we should proxy the methods we intend to be used from outside rather than allow direct access to the LocalStorage instance.

technoweenie:
> we should proxy the methods we intend to be used from outside rather than allow direct access to the LocalStorage instance

I like that idea. I'll rename it to `lfs.objects`.

PR #863, opened 2015-11-25
Export ObjectResource

sinbad:
objectResource is returned by several public functions but isn't exported itself, so fix that. 

Context: I'm using calls to Batch() in my server API test tool which is outside the LFS package & need to see the results. I remember I also needed it when refactoring the API for my old SSH PR so I think it's likely to be useful generally.


technoweenie:
Seems fine for now. I'd really like to get #779 out before too many other go packages start depending on our "lfs" one though.

PR #868, opened 2015-11-26
Add server API test tool

sinbad:
Created a tool called `git-lfs-test-server-api` which can be pointed at either an API URL or a clone URL and will run through a series of tests to ensure the responses are as expected. See the `Readme.md` for details of how to use it.

Currently only tests the API responses to upload/download and through inference correct authentication; HTTPS or SSH, works with both via `--clone=<url>`. Relies on the internal methods to test that content types etc are respected. In order to try to keep it simple to update in future, it just calls the `Batch()` method rather than trying to dig any deeper than that.

It can either generate its own test data (in which case it needs to be able to upload some initial state, although it doesn't push any commits, only LFS files), or it can be given lists of data to test with, in which case no server state is modified. In the latter case the content server is not hit at all, only the API.

I wrote this to provide a repeatable smoke test for API implementations, extensions for cases I haven't thought of are welcome, although as mentioned I didn't really want to go any lower than `Batch` to make it easier to keep in sync later.

### DIFF
--- a/lfs/client.go
+++ b/lfs/client.go
@@ -40,12 +40,12 @@ var (
 	}
 )
 
-type objectError struct {
+type ObjectError struct {
 	Code    int    `json:"code"`
 	Message string `json:"message"`
 }
 
-func (e *objectError) Error() string {
+func (e *ObjectError) Error() string {
 	return fmt.Sprintf("[%d] %s", e.Code, e.Message)
 }
 
@@ -54,7 +54,7 @@ type ObjectResource struct {
 	Size    int64                    `json:"size"`
 	Actions map[string]*linkRelation `json:"actions,omitempty"`
 	Links   map[string]*linkRelation `json:"_links,omitempty"`
-	Error   *objectError             `json:"error,omitempty"`
+	Error   *ObjectError             `json:"error,omitempty"`
 }
 
 func (o *ObjectResource) NewRequest(relation, method string) (*http.Request, error) {
--- a/lfs/config.go
+++ b/lfs/config.go
@@ -60,6 +60,7 @@ type Configuration struct {
 	fetchIncludePaths []string
 	fetchExcludePaths []string
 	fetchPruneConfig  *FetchPruneConfig
+	manualEndpoint    *Endpoint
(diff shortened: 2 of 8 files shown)


technoweenie:
Finally sat down to play around with this. I did have to fix a small merge conflict with your `lfs.pushurl` patch (#949) in https://github.com/github/git-lfs/commit/21bdc9ab1132cef9e9a35680fc279925ac8c91b2#diff-1e640384cfb813925cdf6f092abc5399.

Also, it kept saying `0 tests`. Turns out it wasn't building with the `testupload.go` and `testdownload.go` files. https://github.com/github/git-lfs/commit/295fd47f8748094faf09c8f4bb1ca110018d006e

I wonder if the compilation should happen in `script/bootstrap` instead? I don't think this command is used in any of the lfs shell tests.

sinbad:
> Also, it kept saying 0 tests. Turns out it wasn't building with the testupload.go and testdownload.go files.

Ah, sorry about that - we were using this tool but built it directly in the dir rather from the script, I forgot that because there are no dependencies from main.go to the other files the compiler wouldn't build them from outside the dir. I'd only put it in here to catch build errors & hadn't run it from there.

> I wonder if the compilation should happen in script/bootstrap instead? I don't think this command is used in any of the lfs shell tests.

I don't understand - it's in setup() which is called at the start of the integration tests, it's what builds the other tools like lfstest-gitserver?

technoweenie:
Those utilities are exclusively for the integration tests. I just thought running `script/bootstrap` to build `git-lfs-test-server-api` would be better, than having to run integration tests. How are you using it?

sinbad:
Oh I see, sure; I included it in the integration test as a way to not miss breaking changes during dev. When we were using it here we just built it separately. If you're ok with building it all the time it's fine with me to include it in the bootstrap. It might be easier to just `pushd` into the dir and build rather than have to list all the source files?

technoweenie:
Another option is something like `script/build-test-server-api` I guess. Not a big deal though, which is why I didn't hold this PR up :)

PR #869, opened 2015-11-26
Remove 'config clash' warning, this can be misleading

sinbad:
I'm not sure what problem this was originally solving but it causes false-positives in SourceTree and some scripts that I run. 

Common perfectly valid cases where this can happen include using the '-c' option to git on any setting that's already specified in the gitconfig. It's normal to want to override settings like this sometimes and the warnings look scary when everything is fine.

If there's a more specific way to solve the original problem please let me know.


technoweenie:
It was added to warn about clashes with `lfs.*.access` settings. I changed it to warn only for lfs keys and not git config keys in #861.

sinbad:
Cool, thanks!

Issue #835, opened 2015-11-14
Windows uninstaller freeze!


ve3:
1. Download Git lfs
2. double click to install. default path is my desktop.
3. don't run command that alert on install done.
4. there are 2 files git-lfs.exe and git-lfs-uninstaller.exe. now i seems to don't want it. i want to delete it.
5. right click on git-lfs-uninstaller.exe and run as admin because double click won't work.
6. everything on desktop freezed. can't right click. can't delete file. and can't find where is this file in task manager.

are these files safe?

technoweenie:
That uninstaller is generated by [nsis](http://nsis.sourceforge.net/Main_Page). I want to look at using another windows installer toolkit for future versions though.

There's nothing weird in the [uninstall process](https://github.com/github/git-lfs/blob/b0318b25350b9abfe56af0afb01913a9b2ed56a4/script/nsis/installer.nsi#L61-L67). It just deletes `git-lfs.exe` from your system.

ttaylorr:
Hi @ve3, can you confirm whether this is still happening? I am inclined to close this issue, since it is quite old and we haven't received any reports of similar behavior since.

I am going to close this issue for now, but please do not hesitate to leave a comment here if this issue persists for you, and I'd be happy to dive in further.

Issue #819, opened 2015-11-06
Windows installer is ignoring installation path


shana:
I just ran the windows installer and it had the Desktop as the default path for the installation (which isn't really the best place to put executables, tbh).

I changed the path, and it extracted the uninstaller to the custom path I gave it, but the `git-lfs.exe` executable went to the Destkop anyway and not the custom path.

technoweenie:
Doh. IIRC, nsis forced me to add an [install dir](https://github.com/github/git-lfs/blob/b0318b25350b9abfe56af0afb01913a9b2ed56a4/script/nsis/installer.nsi#L4), though it [looks for Git in common locations](https://github.com/github/git-lfs/blob/b0318b25350b9abfe56af0afb01913a9b2ed56a4/script/nsis/installer.nsi#L21-L34). I was hoping to get around having to edit the system path.

I wanted to try out InnoSetup for a windows installer in the next release. Hopefully it's better than this funky nsis scripting language.

shana:
The easiest way to make an installer is to use Wix, really. That's what I use for the visual studio extension, there's no funky/weird UIs or things, it's just command line tools and a simple template file to  create an installer, most straightforward thing ever.

shana:
I can throw a wix template together quickly if you want to try it out. What are the rules for the installer? (i.e., what does it have to put where, are there registry keys, are there arch-specific things it needs to do, etc)

technoweenie:
> I can throw a wix template together quickly if you want to try it out.

That would be amazing :metal: A cli interface is definitely preferable when cutting releases. Especially if I have multiple builds to deal with. I started with nsis because I found a `makensis` command in homebrew, and I had dreams of building windows installers from my Mac. Unfortunately, I needed Windows to sign the installer anyway, so I never actually used `makensis`.

> The default installation dir is so the uninstaller (or the original msi) can live somewhere for later repairs/uninstalls, but there's way better places than the desktop.

Yea, agreed.

> What's the lowest supported version of Windows?

Ideally, whatever the lowest supported version for Git For Windows.

jigawot:
So, I installed using chocolatey (as per instructions in readme.md) and it put git-lfs.exe in just plain old C:\.  Did I do something wrong or is this expected behavior?

sschuberth:
Closing this as obsolete as we're now using Inno Setup instead of NSIS.

Issue #799, opened 2015-10-28
Git LFS installers should run 'git lfs init' as part of installation


strich:
Given that `git lfs init` is a global command I figure it might make sense to do this as part of the installation script on any platforms that have it (Windows).

It would obviously still need to be in the installation documentation for any methods where the user is simply extracting the package on to the computer.

technoweenie:
Totally. We use the nsis installer for windows, but I've been thinking about switching to [another one](http://www.jrsoftware.org/isinfo.php) for v1.1. I haven't had much windows experience, so any help here is appreciated.

strich:
Yeah I saw that. I had planned to take a look at implementing that task myself soon if no one else picks it up.

technoweenie:
@strich We have an [Installation chat room](https://gitter.im/github/git-lfs/installation) (in addition to the main lfs chat) if you want to rap about it in more real time :)

dg9ngf:
I guess it shouldn't update the global config but the system config instead. The LFS setup is a systemwide thing and all users of the computer can use it. The global config only applies to a single user. So if one user installs this package and another one uses it, this part of the installation would be missing is only the global config was updated.

Instead of calling `git config --global ...`, it should be `git config --system ...`. If your installer writes to the files directly, act accordingly.

And including this step would remove a lot of the confusion people have around getting LFS to work. Having to install something twice that isn't obviously documented (and otherwise it would fail with very strange error messages or simply not do anything at all) is a bad user experience and leads to either support load or frustration and abandonment. (Actually LFS needs to be installed three times, with the third time being the repository-local config. This obviously isn't covered by the application installer.)

strich:
Hey @dg9ngf - Yep this is clearly an issue I had not thought of. It would be trivial to change this to a system-wide config change. @technoweenie do you foresee any issues with this? If not, I'll make the change and submit the PR.

technoweenie:
These config settings shouldn't affect any repositories that aren't using LFS, unless they happen to have their own separate LFS filter setting. Without `--force`, `git lfs install` won't overwrite an existing setting that it doesn't recognize.

I agree that the system config is a more appropriate default. I think both the `install` and `uninstall` commands should get `--system` and `--global` options, to go with `--local` as well for parity with `git config`. The windows installer is probably fine as-is.

@dg9ngf: thanks for the feedback!

dg9ngf:
Well, you should retest your setup. I have just installed the following components into a brand new Windows 10.0 system:

[...]

Issue #736, opened 2015-10-13
Windows installer location


andyneff:
I told the windows installer to install into `D:\Program Files (x86)\Git\bin` and it correctly put the uninstaller there, but not git-lfs.

1. I downloaded https://github.com/github/git-lfs/releases/download/v1.0.0/git-lfs-windows-amd64-1.0.0.exe to `C:\Users\Me\Desktop\`

I have git for windows on my computer, but do not have it in any standard location, furthermore, it probably wasn't installed, but just unzipped. So it correctly didn't detect and git installed location if it was supposed to. This is how I like my setup. That part is good. All I see is a prompt for install location. So I choose `D:\Program Files (x86)\Git\bin` and tell it to install there.

![image](https://cloud.githubusercontent.com/assets/7596961/10461963/f090621a-71a9-11e5-82b6-78ec886b543a.png)

`git-lfs.exe` is installed to `C:\Users\Me\Desktop\git-lfs.exe` while `git-lfs-uninstaller.exe` is installed in `D:\Program Files (x86)\Git\bin\git-lfs-uninstaller.exe`

I was expecting `git-lfs.exe` to go into `D:\Program Files (x86)\Git\bin\` too.

Also, since git-lfs wasn't in the same dir, the uninstaller only removed itself, not the git-lfs left on my desktop. If I moved `git-lfs.exe` into the same directory, the uninstaller successfully removes it.

awayken:
:+1: I had issues with this, too.

technoweenie:
I'd really love help from someone that _uses_ Windows on this :) We're using an nsis script, but I've been thinking about moving to wix. I'll take a look this weekend.

whoisj:
> I'd really love help from someone that uses Windows on this :) We're using an nsis script, but I've been thinking about moving to wix. I'll take a look this weekend.

Big fan of Inno Setup myself. Git for Windows uses it as well.

WIX is a special kind of painful.

technoweenie:
Cool, I'll look into that. I originally went with NSIS because I saw a `makensis` command in homebrew, meaning I could (in theory) do the build from my mac. Of course, I later discovered I should sign the exe, so I'm stuck using my Windows VM :/

brabil:
I was trying to build on windows and I was stuck at missing zip command on my windows running git bash using MINGW64. What type of zip you are using on your windows VM to build the windows installer ?

technoweenie:
[...]

PR #875, opened 2015-11-30
Win installer upgrade

strich:
A new INNO Setup based windows installer. Features:

* Issue #819 - The installer will search the PATH environment vars for the location of any actively installed Git instance and use it to find out where to place Git LFS.
* If Git isn't installed or isn't in the env vars then the installer will simply ask the user to tell it where to install Git LFS.
* The installer is dual arch and should run on 32/64bit systems. However no effort has been made to actually install the correct git-lfs.exe file yet.
* It hasn't been setup, but its possible to integrate code signing either in the compile commandline or in the script itself.
* Issue #799 - The installer and uninstaller run git lfs install/uninstall for you.

Should also fix:
* Issue #835.
* Issue #736.


script/windows-installer/inno-setup-git-lfs-installer.iss
```diff
@@ -0,0 +1,89 @@
+; Script generated by the Inno Script Studio Wizard.
+; SEE THE DOCUMENTATION FOR DETAILS ON CREATING INNO SETUP SCRIPT FILES!
+
+#define MyAppName "Git LFS"
+#define MyAppVersion "1.1.0.0"
+#define MyAppPublisher "GitHub, Inc"
+#define MyAppURL "https://git-lfs.github.com/"
+#define MyAppExeName "Git-LFS-Installer.exe"
```
technoweenie: Instead of defining this here, can we make it dynamic based on the version?  `git-lfs-windows-amd64-{#MyAppVersion}.exe`. Also, does the version need 4 digits? If we can get away with a `MyAppVersion` of `x.x.x`, I think I'd prefer that.

strich: Yep should be easy to do.

script/windows-installer/inno-setup-git-lfs-installer.iss
```diff
@@ -0,0 +1,89 @@
...
+  ResultCode: integer;
+
+begin      
+  TmpFileName := ExpandConstant('{tmp}') + '\git_location.txt';
+  
+  Exec(
+    ExpandConstant('{cmd}'),
+    '/C "for %i in (git.exe) do @echo. %~$PATH:i > "' + TmpFileName + '"', 
+    '', SW_HIDE, ewWaitUntilTerminated, ResultCode
+  );
```
technoweenie: This is problematic because git installers typically setup an alternate shell shortcut, like "Git CMD". But the default `%PATH%` will not have Git anywhere. Not sure what we can do about that though.

technoweenie:
What do you think about a step that asks for your Git For Windows installation dir? From what I can tell, we'd have to put `git-lfs.exe` in multiple places. Assuming the default of `c:\Program Files\Git`:

* `c:\Program Files\Git\cmd` - used by Git CMD.
* `c:\Program Files\Git\bin` - not sure, seems legit though.
* `c:\Program Files\Git\mingw64\bin` - used by Git Bash.

I don't really care about supporting GitHub Desktop or Sourcetree, since both bundle git lfs with their own bundled versions of Git.

Having an input so people can add whatever path they want is important, but I also would like to support the standard Git For Windows installer too.

strich:
This installer does already have a step where the user can optionally change the install directory. We simply try to find it for them and make it the default directory.

In my tests only 1 directory requires the lfs executable - `c:\Program Files\Git\mingw64\bin`. Putting it there works in Git Bash as well as Powershell/cmd.

strich:
As per the commit comments, the installer will now install LFS into its own program dir in program files. It will compile both x86 and x64 binaries into a single installer and install the correct one based on the OS arch.

Note that to allow the move to the new file location, we cannoy simply have the LFS dir at the start of the PATH env var as Git will choose to use the `git-lfs.exe` file in its own local directory prior to looking in PATH.

technoweenie:
Cool, worked great for me :metal:  Thanks so much for hacking on this!

dg9ngf:
#799 is still open, and according to me observation this is correct (the installer does not update the config files). Yet this is mentioned in the description and this is merged. Something went wrong here!

technoweenie:
Is this a system vs global issue, as you mentioned in https://github.com/github/git-lfs/issues/799#issuecomment-200340776 ? The installer should be running the command [here](https://github.com/github/git-lfs/blob/cc41ca541361372323ccafd9d4f14797151c421c/script/windows-installer/inno-setup-git-lfs-installer.iss#L108).

technoweenie:
@dg9ngf: The problem is just that git lfs is not installing to something in `$PATH`. I posted a new issue with the specific issues in: #1120. Please continue the discussion there, it's very easy to lose track of discussion on closed items. Thanks!

dg9ngf:
Sounds reasonable. I just can't see this as a regular user of the installer. I've subscribed to the other issue. BTW, I didn't change the default install path, so this should hit everybody.

PR #879, opened 2015-12-02
Add git version dependency to 

ssgelm:
Since git-lfs requires git >= 1.8.2 this ensures that people don't accidentally install it with older versions of git


technoweenie:
Seems legit :+1: Ready to merge?

ssgelm:
:shipit:

Issue #878, opened 2015-12-02
if git-credential does not exist, git push fails to upload LFS files


tpdownes:
On a computing cluster running Debian 7, which has git 1.7 by default. It turns out that git 1.7 as packaged for Debian 7 does not include a git-credential binary.

Without realizing this, I setup a credential helper (globally) to help https operations. Normal git operations were not affected by this vestigial configuration. I didn't notice it wasn't working because my operations were less frequent than the timeout.

```
[credential]
        helper = cache --timeout 900
```

But... with this configuration, git push in a repository with LFS tracking files fails to even attempt a file upload.

In some sense, this is user error since the config has a section that can't help, but git LFS should fail gracefully to interactive prompts when git-credential fails so the bug may have wider applicability. I was able to resolve the error by installing a backported version of git 1.9 which includes the git-credential binary.

```
git-lfs/1.1.0 (GitHub; linux amd64; go 1.5.1)
git version 1.7.10.4

$ git-lfs smudge large/git-lfs_1.1.0_amd64.deb
Error accessing media: large/git-lfs_1.1.0_amd64.deb (b4403d3b939557fa802c09b3b27d0e6cc80a34798a8c6f35cfe84167e0b49208)

Git credentials for https://my.host/my-name/lfs-test.git not found:
'git credential fill' error: exit status 1

goroutine 1 [running]:
github.com/github/git-lfs/lfs.Stack(0x0, 0x0, 0x0)
    /tmp/docker_run/git-lfs/obj-x86_64-linux-gnu/src/github.com/github/git-lfs/lfs/errors.go:557 +0x80
github.com/github/git-lfs/commands.logPanicToWriter(0x7f3eff4451e8, 0xc820144078, 0x7f3eff4471a8, 0xc8204603a0)
    /tmp/docker_run/git-lfs/obj-x86_64-linux-gnu/src/github.com/github/git-lfs/commands/commands.go:184 +0xf7f
github.com/github/git-lfs/commands.logPanic(0x7f3eff4471a8, 0xc8204603a0, 0x0, 0x0)
    /tmp/docker_run/git-lfs/obj-x86_64-linux-gnu/src/github.com/github/git-lfs/commands/commands.go:148 +0x421
github.com/github/git-lfs/commands.handlePanic(0x7f3eff4471a8, 0xc8204603a0, 0x0, 0x0)
[...]

Issue #871, opened 2015-11-27
Package cloud - GPG signatures


ghost:
"Generating and verfying GPG signatures is critical to establishing the authenticity of package objects and the repositories storing those objects." in [package cloud](http://blog.packagecloud.io/eng/2014/11/24/howto-gpg-sign-verify-rpm-packages-yum-repositories/)

I was building an ansible role to download these package and noticed that in [package cloud centos6](https://packagecloud.io/install/repositories/github/git-lfs/config_file.repo?os=el&dist=6&source=script)
and [package cloud centos7](https://packagecloud.io/install/repositories/github/git-lfs/config_file.repo?os=el&dist=6&source=script)

gpgcheck=0

Can you please take a look?

technoweenie:
Both of those have `repo_gpgcheck=1`. So that should verify Package Cloud through a gpg signature that they manage. Package Cloud doesn't let me upload my own gpg keys AFAIK, otherwise I would.

/cc @andyneff

andyneff:
Tricky....

> **repo_gpgcheck** Either '1' or '0'. This tells yum whether or not it should perform a GPG signature check on the repodata. When this is set in the [main] section it sets the default for all repositories. The default is '0'. [1](http://linux.die.net/man/5/yum.conf)

Now, a little from packagecloudio

> packagecloud will sign your repos with a GPG key. This means that anyone who uses your repo can be certain that the repo was generated by packagecloud. The install scripts we generate for your repos will automatically import our public GPG key allowing anyone who uses your packagecloud repo to verify that the repo's metadata has not been tampered with. [2](https://packagecloud.io/docs#gpg)

Looks like packagecloud is trying to walk a fine "techinical" line saying they support gpg signing... The rpms have no gpg signature, only md5/sha1

```
[root@86d1d2c4ab19 /]# rpm -K git-lfs-1.1.0-1.el7.x86_64.rpm
git-lfs-1.1.0-1.el7.x86_64.rpm: sha1 md5 OK
```


There is a (now old) piece of code to sign the our packages (https://github.com/github/git-lfs/tree/master/docker#gpg-signing). I'll look into it a little to see if it is still functional.

However, even IF we did that.

1. packagecloud does not appear to have the ability to enable gpgcheck,
2. packagecloud does not appear to have the ability to add our own gpg public key

IF signing the metadata is NOT enough, I would have to suggest:

1. Having the package generator sign the packages at release time before uploading
[...]

PR #891, opened 2015-12-06
Rpm require

andyneff:
Added require for git back into rpm

-Moved Centos 6 build/tests to use IUS for a more official-ish git rpm. Removed our own git.spec file, as discussed in #878
-Change docker images to auto build images, which resulted in a slight name change. The old images will be taken down at a later time
-Got tests working again with new auto builds. They were a low priority, so they were purposefully left behind. But with discussion like #871, I wanted to make sure this was working again. It still doesn't work against package cloud keys yet, but since packagecloud doesn't sign rpms, it would appear there's no point there
-Cleaned up all files duplicated in the git-lfs_dockers repo.
-Updated README
-Since the dockerfiles are no longer in the docker directory, changed the run_dockers script to go by name (i.e. `debian_7` instead of filename `git-lfs_debian_7.dockerfile`, since there was no purpose behind the longer version)


technoweenie:
Looks good :+1: Thanks for the patch!

PR #907, opened 2015-12-21
Update PackageCloud Link

wpsmith:
Change PackageCloud link to go to more intuitive install page since this is what the list bullet is about.


technoweenie:
Good call, thanks :)

PR #909, opened 2015-12-22
Fix fallthrough when git-lfs-authenticate returns an error

sinbad:
If using an SSH URL and git-lfs-authenticate returns an error, without this change the error is just reported to the trace but execution continues.

A common result is a completely made up LFS URL being returned which then 404's, which ends up setting lfs.batch=false. If the server doesn't implement the non-batch interface, this config then breaks LFS even after the SSH issue is resolved & has to be manually removed, which is quite confusing for the user.


technoweenie:
Yuck :(  Thanks for the fix :)

Issue #922, opened 2016-01-07
LFS in GIT submodule is skipping .gitconfig


panjandrum:
Hi guys!
I'm whole day struggle with one big problem. Maybe you will be able to help me?

Lets me explain my situation.
First of all I'm using *lfs-test-server* on localhost with basic authentication.
I have 2 repositories, where second one is used as a GIT submodule in first. In both I'm using the same LFS server.

For simplicity lets assume that it looks like that:
```
first
|-- .git
|-- .gitconfig
`-- second
    |-- .git
    `-- .gitconfig
```

Both `.gitconfig` file looks like:
```
[lfs]
    url = "http://127.0.0.1:9999/"
[lfs "http://127.0.0.1:9999/"]
    access = basic
```

When I'm cloning first one, everything is ok. Files are downloaded from LFS server. However afer i run `git submodule update --init --remote` I have following error:
```
Error accessing media: bin/makedepend (0759e66ca9ae5ab00efa3641c604bd8e6e827f3594b8e57d5055a99c2bd5e18a)

Post file:///home/emaclec/opt/git/remote/sandbox/mtas_ftfw_artifacts.git/info/lfs/objects/batch: unsupported protocol scheme "file"
```

It looks like `.gitconfig` from submodule is not loaded at all. When I add .gitconfig content to `first/.git/modules/second/config` it solves the problem, but in my case I would like to avoid this.

Do you know how to force load `.giconfig` when repo is used as a submodule? Or maybe this is a bug?

Thanks in advance
panjandrum

technoweenie:
Seems like lfs should detect that it's in a submodule and pick the right file. Thanks for the report.

technoweenie:
This should be fixed in v1.1.1: https://github.com/github/git-lfs/releases/tag/v1.1.1

Thanks again!

PR #927, opened 2016-01-11
Fix loading config file (.lfsconfig or .gitconfig) when git lfs repository is used as submodule

rikdev:
If git lfs repository is used as submodule then git lfs can't find configure file (.lfsconfig or .gitconfig), because git lfs looks for the configure file in the directory <root_repository>/.git/modules (GIT_WORK_TREE) and the configure file is located in directory <root_repository>/<submodule_path>.

### DIFF
--- a/lfs/config.go
+++ b/lfs/config.go
@@ -339,6 +339,14 @@ func (c *Configuration) loadGitConfig() bool {
 		// TODO: remove .gitconfig support for Git LFS v2.0 https://github.com/github/git-lfs/issues/839
 		filepath.Join(LocalWorkingDir, ".gitconfig"),
 	}
+
+	gitRootDir, _ := git.RootDir()
+	if gitRootDir != "" {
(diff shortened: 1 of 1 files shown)


technoweenie:
This looks good, but there are no tests. Ideally a test in [`test-env.sh`](https://github.com/github/git-lfs/blob/c57bcf0f96fc0dc1f9a756236f80ce3b9ea7619f/test/test-env.sh) sets up a repository using submodules and confirms that it loads the config correctly.

panjandrum:
Avesome!

I belive that will fix the problem which I described in #922
When can I expect official release cantains this fix?

Best regards

technoweenie:
Hi, thanks for the patch. There are no tests for this behavior though. What's the use case? Is this intended to fix #922?

technoweenie:
Jeez, I can't read. Yes, #922. I'll add tests for this real quick :)

technoweenie:
Thanks for the patch. While I went with a totally different fix (https://github.com/github/git-lfs/pull/963), this patch got me on the right track.

PR #933, opened 2016-01-14
enable OSX build on TravisCI

larsxschneider:


technoweenie:
What does this do? Does it run tests on linux and osx? Or does it save built artifacts somewhere?

larsxschneider:
this just runs all the tests on Linux and OSX:
https://travis-ci.org/github/git-lfs/builds/102405529

I don't know how likely it is that git-lfs brakes something on OSX but not on Linux... I thought the PR is easy enough to just bring it to the table :smile:

larsxschneider:
(to be honest the real reason was that the tests fail on my Mac... but apparently it must be my setup since TravisCI is green on OS X)

PR #949, opened 2016-01-25
Support remote.name.pushurl correctly to fix #945

### LINKED ISSUE #945: Endpoint needs to respect remote.name.pushurl during 'git lfs push'
Currently when falling back on the clone URL as a base for the LFS url, `remote.name.url` is always used regardless of the operation.

When pushing, however, to match Git behaviour we should be checking for the existence of `remote.name.pushurl` and using that instead. Otherwise pushes of LFS data will go to a completely different place to the commits when users have configured things this way (e.g. via a command like `git remote set-url --push origin https:/some/push/url.git`)

I'm happy to spend some time on resolving this but am just logging it here so it doesn't get lost for now.

sinbad:
Fixes #945 

Right now git-lfs ignores the use of 'remote.[name].pushurl' in asymmetrical pull/push situations (within a single remote). This PR changes the way Endpoints are built so that the operation ("download" or "upload") is taken into account and the pushurl is used as a basis instead of the url when pushing. This avoids git-lfs behaving inconsistently with Git itself & the resulting confusion.

Also add support for 'remote.[name].lfspushurl' and global 'lfs.pushurl' so that push versions of urls are supported everywhere you might configure them.

This also affects the Access flags which are stored & looked up alongside the correct url based on whether you're pulling or pushing, in case the 2 urls have different auth mechanisms.

### DIFF
--- a/commands/command_env.go
+++ b/commands/command_env.go
@@ -16,7 +16,7 @@ var (
 func envCommand(cmd *cobra.Command, args []string) {
 	lfs.ShowConfigWarnings = true
 	config := lfs.Config
-	endpoint := config.Endpoint()
+	endpoint := config.Endpoint("download")
 
 	gitV, err := git.Config.Version()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


lfs/credentials.go
```diff
@@ -51,7 +51,8 @@ func getCredsForAPI(req *http.Request) (Creds, error) {
 }
 
 func getCredURLForAPI(req *http.Request) (*url.URL, error) {
-	apiUrl, err := Config.ObjectUrl("")
```
sinbad: I dropped ObjectUrl because it seems defunct except for this check, where it doesn't even use an oid. Seems it's just looking for a base api url to look up the creds for and Endpoint does that.

commands/command_env.go
```diff
@@ -16,7 +16,7 @@ var (
 func envCommand(cmd *cobra.Command, args []string) {
 	lfs.ShowConfigWarnings = true
 	config := lfs.Config
-	endpoint := config.Endpoint()
+	endpoint := config.Endpoint("download")
```
technoweenie: Does it make sense to setup constants like `DownloadOperation` instead of passing in a string?

sinbad: Yeah, I was thinking of doing that but there's lots of other places (unrelated to this change) it should apply too; it was already leaking into more areas than I liked so I didn't do it. I can do though.

technoweenie: Well, we could make that part of the refactor I guess.

technoweenie:
Thanks for tackling this! I tend to forget `pushurl` exists.

PR #962, opened 2016-01-29
Rikdev fix submodule lfsconfig (with tests!)

technoweenie:
Adds tests to #927.


technoweenie:
Better fix in https://github.com/github/git-lfs/pull/963.

PR #963, opened 2016-01-29
Fix submodule lfsconfig

technoweenie:
This is another approach to fixing #922.

While #927 fixes the bug, I'm not completey happy with it. Calling `git.RootDir()` in `loadGitConfig()` adds another os exec that's run on _every_ lfs command. This is in addition to when it's run inside `resolveGitDir()` if the `GIT_DIR` is specified. I'd like to fix the behavior without running `git.RootDir()` again.

The problem: `git submodule update --init --remote` runs `git lfs smudge` commands with the following env:

```
GIT_DIR=/path/to/repo/.git/modules/sub
GIT_WORK_TREE=.
```

When run with the above ENV vars, `LocalWorkingDirectory` is `/path/to/repo/.git/modules`, which is wrong. It should be `/path/to/repo/sub`, where `sub` is the path for the submodule. Here's the [code](https://github.com/github/git-lfs/blob/84e9e66c9156564180c9621491865b985f3c838d/lfs/lfs.go#L187-L193):

```go
func processWorkTreeVar(gitDir, workTree string) (string, string, error) {
	// See `core.worktree` in `man git-config`:
	// “The value [of core.worktree, GIT_WORK_TREE, or --work-tree] can be an absolute path
	// or relative to the path to the .git directory, which is either specified
	// by --git-dir or GIT_DIR, or automatically discovered.”

	if filepath.IsAbs(workTree) {
		return workTree, gitDir, nil
	}

	base := filepath.Dir(filepath.Clean(gitDir))
	absWorkTree := filepath.Join(base, workTree)
	return absWorkTree, gitDir, nil
}
```

If we treat `GIT_WORK_TREE=.` relative to the process' current working directory, it works just fine.  It seems like the comments (which are taken directly from the git docs) are wrong in this case. I can pass both `--git-dir` and `--work-tree` to get the status of my local git-lfs repo, from a totally different directory:

```cli
~/test $ git --git-dir=/Users/rick/p/git-lfs/.git --work-tree="/Users/rick/p/git-lfs" status -sb
## rikdev-fix-submodule-lfsconfig...origin/rikdev-fix-submodule-lfsconfig
 M lfs/config.go
 M lfs/lfs.go
 M test/test-submodule-lfsconfig.sh
```

If I try relative to the `.git` directory, it doesn't work.

```
~/test $ git --git-dir=/Users/rick/p/git-lfs/.git --work-tree=".." status -sb
## rikdev-fix-submodule-lfsconfig...origin/rikdev-fix-submodule-lfsconfig
 D ../.gitattributes
```

If I try relative to the current working directory, while in the git-lfs directory, it works:

```cli
~p/git-lfs git:(rikdev-fix-submodule-lfsconfig) $ git --git-dir=/Users/rick/p/git-lfs/.git --work-tree="." status -sb
## rikdev-fix-submodule-lfsconfig...origin/rikdev-fix-submodule-lfsconfig
 M lfs/config.go
 M lfs/lfs.go
 M test/test-submodule-lfsconfig.sh
```

@sinbad @rikdev: Thoughts?

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -451,7 +451,38 @@ func GetCommitSummary(commit string) (*CommitSummary, error) {
 		msg := fmt.Sprintf("Unexpected output from git show: %v", string(out))
 		return nil, errors.New(msg)
 	}
+}
 
+func GitAndRootDirs() (string, string, error) {
+	cmd := execCommand("git", "rev-parse", "--git-dir", "--show-toplevel")
+	out, err := cmd.CombinedOutput()
+	output := string(out)
+	if err != nil {
+		return "", "", fmt.Errorf("Failed to call git rev-parse --git-dir --show-toplevel: %q", output)
+	}
+
+	paths := strings.Split(output, "\n")
+	pathLen := len(paths)
+
+	if pathLen == 0 {
+		return "", "", fmt.Errorf("Bad git rev-parse output: %q", output)
+	}
+
+	absGitDir, err := filepath.Abs(paths[0])
+	if err != nil {
+		return "", "", fmt.Errorf("Error converting %q to absolute: %s", paths[0], err)
+	}
+
+	if pathLen == 1 || len(paths[1]) == 0 {
+		return absGitDir, "", nil
+	}
+
+	absRootDir, err := filepath.Abs(paths[1])
+	if err != nil {
+		return "", "", fmt.Errorf("Error converting %q to absolute: %s", paths[1], err)
+	}
+
+	return absGitDir, absRootDir, nil
 }
 
 func RootDir() (string, error) {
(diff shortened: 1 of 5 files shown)


sinbad:
In #692 I suggested that a future extension was to remove the manual worktree processing entirely and just use `git rev-parse --show-toplevel` in all cases instead (exposed via `git.RootDir()`). Would this resolve it?

technoweenie:
I like that idea. I'm playing with a change that calls `git rev-parse --git-dir --show-toplevel` only. This cuts a lot of code, only runs _one_ exec, and should respect any git env or config values.

Not too many failures. Though the bare repo one is interesting...

```
test: env with bare repo ...                                       FAILED
test: install updates repo hooks ...                               FAILED
test: submodule env ...                                            FAILED
test: update ...                                                   FAILED
```

technoweenie:
Cool, got it working. How's this look?

sinbad:
Awesome, I didn't know you could use both options on the same command. Looks great, much simpler, more efficient and fixes that bug. :metal:

technoweenie:
I didn't either, just tried it and was pleasantly surprised :)

PR #966, opened 2016-02-01
Don't create LFS pointer for empty files

nathanhi:
This PR fixes the reported behaviour in https://github.com/github/git-lfs/issues/557, where Git reports
a delta, if an attempt was made to track empty files with LFS.
If there already have been empty files tracked with LFS, these
need to be untracked manually, as this fix only prevents that
newly created, empty files can be tracked by LFS.

Problem is, that Git never runs the clean filter on 0-byte files:

https://github.com/git/git/blob/80980a1d5c2678ab9031d7c60faf38b9631eb1ce/diff.c#L2777

This leads to a diff between the file in the repository (pointer file) and 
the actual file content. Apparently the git-fat project seems to be affected 
by this, too (https://github.com/jedbrown/git-fat/issues/18).
The fix/workaround for this in LFS is simple; just don't create pointers 
for 0-byte target files, it's useless to store them in LFS anyway.

I'm not too sure though if my current patch is sufficient - it just checks
if the given file from the parameters has a size == 0, but apparently the
pointer creation + hash calculation is done from the stdin content. So just
let me know if this needs a rework with length calculation of the stdin
io.Reader.

### DIFF
--- a/lfs/pointer.go
+++ b/lfs/pointer.go
@@ -67,14 +67,16 @@ func (p *Pointer) Encode(writer io.Writer) (int, error) {
 
 func (p *Pointer) Encoded() string {
 	var buffer bytes.Buffer
-	if p.Size != 0 {
-		buffer.WriteString(fmt.Sprintf("version %s\n", latest))
-		for _, ext := range p.Extensions {
-			buffer.WriteString(fmt.Sprintf("ext-%d-%s %s:%s\n", ext.Priority, ext.Name, ext.OidType, ext.Oid))
-		}
-		buffer.WriteString(fmt.Sprintf("oid %s:%s\n", p.OidType, p.Oid))
-		buffer.WriteString(fmt.Sprintf("size %d\n", p.Size))
+	if p.Size == 0 {
+		return buffer.String()
 	}
+
+	buffer.WriteString(fmt.Sprintf("version %s\n", latest))
+	for _, ext := range p.Extensions {
+		buffer.WriteString(fmt.Sprintf("ext-%d-%s %s:%s\n", ext.Priority, ext.Name, ext.OidType, ext.Oid))
+	}
+	buffer.WriteString(fmt.Sprintf("oid %s:%s\n", p.OidType, p.Oid))
+	buffer.WriteString(fmt.Sprintf("size %d\n", p.Size))
 	return buffer.String()
 }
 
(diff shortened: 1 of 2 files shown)


technoweenie:
> I'm not too sure though if my current patch is sufficient - it just checks
if the given file from the parameters has a size == 0, but apparently the
pointer creation + hash calculation is done from the stdin content.

I think this should be handled based on the stdin content only. The filename is given with an optional `%f` parameter. It is _possible_ (though unlikely) that someone can run the `git lfs clean` command without it.

Also, the shell tests are more for integration tests. I think it should actually init a new repo and add an empty file, with the appropriate checks. Maybe even test pushing and cloning afterwards. We care about what commands the user will actually be running. 

There should be a good spot in the Go tests to make sure that cleaning an empty file results in no pointer.

Thanks for hacking on this :)

nathanhi:
Sorry, took me a while, I'm fairly new to Go. I've moved the size verification from command_clean.go to pointer.go, so that Encode() directly returns an empty Buffer on an empty input string. Did, of course, also add a test for that.

lfs/pointer.go
```diff
@@ -67,12 +67,14 @@ func (p *Pointer) Encode(writer io.Writer) (int, error) {
 
 func (p *Pointer) Encoded() string {
 	var buffer bytes.Buffer
-	buffer.WriteString(fmt.Sprintf("version %s\n", latest))
-	for _, ext := range p.Extensions {
-		buffer.WriteString(fmt.Sprintf("ext-%d-%s %s:%s\n", ext.Priority, ext.Name, ext.OidType, ext.Oid))
+	if p.Size != 0 {
+		buffer.WriteString(fmt.Sprintf("version %s\n", latest))
```
technoweenie: I'd prefer an early return here:

```go
func (p *Pointer) Encoded() string {
  	var buffer bytes.Buffer		  	var buffer bytes.Buffer
  	if p.Size == 0 {
  	  	return buffer.String()
  	}
```

This fits common go conventions better.

nathanhi:
@technoweenie: Sure, makes sense. Done.

technoweenie:
Cool, thanks for digging in, especially in an unfamiliar language :)

technoweenie:
BTW, I added an integration test to check this behaves correctly in Git: [test-zero-len-file.sh](https://github.com/github/git-lfs/blob/2dd108fbb5212c16292e729d89e22c7cf54c41dd/test/test-zero-len-file.sh). The first test asserts that the git blob created for the empty file is actually blank, and not an LFS pointer with the SHA-256 signature of `""`. The second test asserts that the repo has a clean `git status` after cloning.

Issue #844, opened 2015-11-19
Git LFS v1.2


technoweenie:
Here are the fixes slated for v1.2:

* [x] netrc support https://github.com/github/git-lfs/pull/715 (@rubyist)
* [x] `clone` command #988 (@sinbad)
* [x] support self-signed certs #1067 (@sinbad)
* [x] support sslverify option for specific hosts #1081 (@sinbad)
* [x] fix clean/smudge filter string for files starting with `-` #1083 (@epriestley)
* [x] fix silent failure to push LFS objects when ref matches a filename in the working copy #1096 (@epriestley)
* [x] fix problems with using LFS in symlinked folders #818 (@sinbad)
* [x] touch existing git tracked files when tracked in LFS so they are flagged as modified #1104 (@sinbad)
* [x] support for git references #1007 (@jlehtnie)
* [x] fix git lfs push silently misbehaving on ambiguous refs; fail like git push instead #1118 (@sinbad)
* [x] whitelist `lfs.*.access` config in local `~/.lfsconfig` #1122 (@rjbell4)
* [x] Only write the encoded pointer information to Stdout #1105 (@sschuberth)
* [x] Use hardcoded auth from remote or lfs config when accessing the storage api #1136 (@technoweenie, @jonmagic)
* [x] SSH should be called more strictly with command as one argument #1134 (@sinbad)
* [x] Stop transferring duplicate objects on major push or fetch operations on multiple refs. #1128 (@technoweenie)
* [x] Support `url.*.insteadof` git config options #1117 (@artagnon)
* [ ] zsh auto completion https://github.com/github/git-lfs/pull/1133

sinbad:
I'm thinking maybe it would be worth trying to bring the significant speed improvements from 'git lfs clone' to the other main workflow tasks that trigger downloads, like pulling or checking out a different branch. Unfortunately we already used 'git lfs pull' and 'git lfs checkout', maybe we either create new wrapper commands like 'git lfs gitpull' or use an option to the existing commands to turn it into a wrapper command e.g. 'git lfs pull --git'. Any thoughts?

bazooka241:
You will excuse me busy place.
The other day I will try to was sorry I have to unpleasant feelings in my selfish behavior, be a twice like this does not occur.
Did not Thank really sorry.

larsxschneider:
[...]

PR #969, opened 2016-02-03
Add ref information to upload request

dluksza:
Adds "ref" parameter extracted from git pre-push hook into the upload
request. This way servers with branch access permissions can enforce
them before large objects are transfered.

Issue: #845


commands/command_push.go
```diff
@@ -45,7 +46,7 @@ func uploadsBetweenRefAndRemote(remote string, refs []string) *lfs.TransferQueue
 		if len(refs) == 0 {
 			pointers := scanAll()
 			Print("Pushing objects...")
-			return uploadPointers(pointers)
+			return uploadPointers(pointers, strings.Join(refs, ","))
```
technoweenie: Since `pre-push` is reading from the git hook, it's probably getting a full remote ref like `refs/heads/master`. However, the `push` command is receiving user input, and is likely a relative name (which could be a branch, tag, etc). Do we care? Should the server just do its best to validate it?

Also, instead of joining the refs array with commas, why not just pass a JSON array to the server.

technoweenie:
Thanks for resurrecting #850. That slipped through the cracks during the holiday break. I'm trying to put out a v1.1.1 bug fix release this week, and then I want to start looking at new things like this.

Compared to #850, I like that this gets the ref from existing data, instead of [another git call](https://github.com/github/git-lfs/pull/850/files#diff-b54b33b58b9d76939b70cd4d2f48f251R257).

I'm not thrilled about passing an extra string argument to functions like `NewUploadQueue()`. Some kind of meta struct like in #850 would be preferable. However, it does seem kind of silly if `Refs` is the only property. But, it would give us a place to potentially add more metadata properties in the future.

I think the refs given to the `fetch` command should be passed to the DownloadQueue as well. But I don't think there's any way to support this in the standard LFS workflow that uses the git smudge filter. Git runs the smudge filters during checkout, which is after the git pull (if there even was one). Therefore it doesn't know the remote ref.

commands/command_push.go
dluksza: True, on the server side we are not interested in the name of local branch, we just want to know where you want to push (put) your changes.

Will remove my modifications from the command_push.go

dluksza:
I've added UploadMetadata struct as you suggested. Also move "ref" property from the root of upload request object to the "meta" property. The "meta" (and "ref") will be only appended when there is something set.

technoweenie:
Why keep it specific to uploads? You should be able to pass the same info from a `git lfs fetch` call, right? If the lfs server is going to look at the ref on uploads to verify authorization, it seems like it should for downloads too. Though I suppose the server could maintain a mapping of oids to remote refs.

dluksza:
I can add this to the `fetch` as well, but IMO this should be also verified on the server. Relaying only on the client data is not secure, since it is easy to craft malicious payload, through guessing the OID is hard.

One can say that that keeping branch-oid mapping is the implementation detail of server and LFS client should not impose that.

technoweenie:
> One can say that that keeping branch-oid mapping is the implementation detail of server and LFS client should not impose that.

100% agreed :+1:. Also, LFS won't be collecting that info when called through the smudge filter, so it may not be there anyway.

Thanks for the patch. This should make it into v1.2.0: #844

dluksza:
Thanks, that is awesome! :dancer:

When v1.2.0 is planned to be released?

technoweenie:
I don't have a strict schedule. #844 has what I'd like to do for that release.  Though if it's taking too long, I may decide to cut a release with what I have at the time :)

sinbad:
I'm not crazy about how this makes the LFS push invoked by the pre-push hook semantically different to using `git lfs push`. Seems that a perfectly valid `git lfs push` would now fail if the server requires ref information, while a pre-push hook invoked version that does the same thing would succeed. Could lead to some odd support edge-cases?

Also the argument against passing refs for fetch was that client-side data can't be trusted, but that applies to pre-push too; I could easily craft my local branch state to include any branches I wanted and the pre-push hook will happily supply that to the server while uploading the LFS objects, even if the actual git push would be rejected, because by nature it happens before any other validation. If it's just to avoid wasting upload time on mistakes then fair enough but I hope no-one interprets it as a security feature.

technoweenie:
> I'm not crazy about how this makes the LFS push invoked by the pre-push hook semantically different to using git lfs push. 

Ah yeah, we should be able to send the same data for a `git lfs push`.

> I could easily craft my local branch state to include any branches I wanted and the pre-push hook will happily supply that to the server while uploading the LFS objects, even if the actual git push would be rejected, because by nature it happens before any other validation.

I think the idea is that their git server has special privileges for certain branches. It should authenticate based on the user AND the fetched ref. Though I imagine most LFS servers will ignore it.

dluksza:
> I think the idea is that their git server has special privileges for certain branches. It should authenticate based on the user AND the fetched ref. Though I imagine most LFS servers will ignore it.

We are mostly concerned about push operations when we wold like to prevent users from pushing large files "to branches" that they  don't have access to. This is a optimization that people won't waste time waiting for the huge blob to upload only just to know that they have access for particular branch.

sinbad:
> It should authenticate based on the user AND the fetched ref.

Yeah, my point is that the user can basically make up any ref they like, with a bit of local branch fudging & the correct `git push` call, and the the pre-push hook will just pass it in the request. So it's not really valid as an authorisation test. But @dluksza has confirmed it's really just to avoid them wasting their time with the LFS upload associated with a git push that subsequently gets rejected.

technoweenie:
@dluksza: Yo, we didn't forget about this :) I really wanted this to work with `git lfs push --all`, so I had to update the scanner code: #1040. #1071 is a PR on top of _that_ which includes this PR. 

It's still kind of a mess now, but should be ready next week :)

dluksza:
Good to know that this change was not forgotten :) Looking forward for this change being merged.

technoweenie:
@dluksza and @sinbad: would love your feedback on this:

So, this PR has no tests, and I caught a nasty bug in here. Now, testing this is pretty tricky (and I didn't even mention it in this PR), so I won't hold that against you. Here's the first pass of tests:

https://github.com/github/git-lfs/commit/620c94b9886452ef952244870238410fda6288f5

These tests are a little weird. Normally, the test lfs server doesn't even bother checking the ref. So, I added `failsMetaCheck()` that only fails if both are true:

* An object OID matching the content `meta-test` **exists**. This tells the test server to verify the given ref.
* An object OID matching the content `meta-test: {full-ref}` (example: `meta-test: refs/heads/master`) **does not exist**. 

Nice thing is that these tests caught the fact that the `strings.HasPrefix` args are swapped. I noticed the tests were failing with: 

    Fails meta check with: {refs/heads/refs/heads/master}

I fixed it by changing it to `strings.HasPrefix(ref, "refs/")`, which also allows full tag refs too.

The tests pass on top of this PR, after it's been applied to #1128. I tested all the scenarios I could think of that work. Two that didn't work that I want to follow up on:

* `git lfs push` with a tag
* `git lfs fetch` with a tag

I'm not positive how we'd fix this, but that can come in another commit. I think we'd just have to attempt to expand ref names (`master` => `refs/heads/master`, `v1.0` => `refs/tags/v1.0`) based on the local refs, erroring if there are any ambiguities. Maybe look into branch tracking stuff too.

sinbad:
> I'm not positive how we'd fix this, but that can come in another commit. I think we'd just have to attempt to expand ref names (master => refs/heads/master, v1.0 => refs/tags/v1.0) based on the local refs, erroring if there are any ambiguities. 

I'm still a bit uncomfortable with this feature being somewhat unreliable and locally exploitable. Leaving aside wilful exploitation, it also adds context in some cases but not all (e.g. pushing oids), and the context it sends is somewhat inconsistent currently - when invoked via `pre-push` it will send the remote branch name (because its passed the destRef), when invoked via `git lfs push` it will send the local branch name because it only uses the local ref. Sure most people use identical names for local/remote branches, but some don't.

> Maybe look into branch tracking stuff too.

This would address that last point, but remember the answer can be different depending on the context (pull vs push). We only have code for the pull case right now because that's all we've needed, we haven't needed to derive a remote branch for push cases before. That's a bit more complex, you need to check the `push.default` config because it's only the same as pull when that's `upstream` or `simple`. It also depends on whether you're pushing to the remote you're tracking too. 

I dunno, obviously this feature is useful in specific cases right now, I'm just a bit uncomfortable with things that are only reliable on the 'happy path'. In this case the bad outcome is only that the ref information given to the server might not be correct in odd cases, so I guess so long as we put a big red flag on this to say it's advisory only, can be manipulated by a smart user if they decide to, and may have some rare edge cases. Puts my teeth on edge a little but I can live with it ;)

dluksza:
+1 for expanded refs

For me it is all client facing REST-like API that one can try to exploit. The lfs-server URL and protocol are know, one can easily craft upload and download request with wrong `oid` and `size`. Same goes for the meta-data.

Server implementors must be aware of that. But wouldn't say that meta-data are less trusted in the worse case one would upload blob's and git data would be rejected (same as it is right now).

This is just a simple optimization for the end users so that they don't need to wait for the blob's being uploaded and then got access deny from the git server.

technoweenie:
This PR is sorely in need of an update. I'll push an updated PR that fixes the merge conflicts. I have a few items I'd like to see before shipping in v1.4 or v1.5:

* Ensure we're always sending full server side refs: `refs/tags/v1.0`, `refs/heads/master`. Never client ref names.
* Integration tests that test `git lfs push`, `git lfs fetch` on repos where the local and remote branches don't match.
* Allow additional properties in the json schema :)

    ```js
    "meta": {
      "type": "object",
      "properties": {
        "ref": {
          "type": "string"
        }
      },
      "required": [],
      "additionalProperties": true
    }
    ```

technoweenie:
Closing this, it's too far behind to merge. I've got an updated version though: #1456

PR #974, opened 2016-02-03
guard against invalid remotes passed to `git lfs push`

### LINKED ISSUE #973: git lfs push --all <remote> falls back unexpectedly
I'm not sure if this is like this because of #436 But its very strange an unexpected

When provided a remote URL `git lfs` wont use it. I've had a dig through the code and it seems it only supports configured git remotes.
That might be OK but the fallback in `git lfs push` is very confusing.

```
$ git lfs push --all http://sgoo@git-server.com/sgoo/test.git
```
results in the push happening against `origin`.

I would expect it to behave more like how `git lfs fetch` behaves:
```
$ git lfs fetch http://sgoo@git-server.com/sgoo/test.git --all
Invalid remote name "http://sgoo@git-server.com/sgoo/test.git"
```

technoweenie:
Fixes #973

### DIFF
--- a/commands/command_pre_push.go
+++ b/commands/command_pre_push.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/github/git-lfs/git"
 	"github.com/github/git-lfs/lfs"
 	"github.com/github/git-lfs/vendor/_nuts/github.com/spf13/cobra"
 )
@@ -48,6 +49,10 @@ func prePushCommand(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
+	// Remote is first arg
+	if err := git.ValidateRemote(args[0]); err != nil {
+		Exit("Invalid remote name %q", args[0])
+	}
 	lfs.Config.CurrentRemote = args[0]
 
 	// We can be passed multiple lines of refs
--- a/commands/command_push.go
+++ b/commands/command_push.go
@@ -4,6 +4,7 @@ import (
 	"io/ioutil"
 	"os"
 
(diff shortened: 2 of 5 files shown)


sgoo:
Should this also be applied to `pre-push`?
If I do:
```
$ GIT_TACE=1 git push http://sgoo@git-server.com/sgoo/test.git
...
12:47:18.836067 git.c:561               trace: exec: 'git-lfs' 'pre-push' 'http://sgoo@git-server.com/sgoo/test.git' 'http://sgoo@git-server.com/sgoo/test.git'
...
```
You can see `pre-push` is invoked with the URL I provided to `git push`. But the request that is made in `pre-push` is made to the remote configured as `origin`.

technoweenie:
@sgoo I think that tracing info is potentially useful.

The os/exec calls in the `git` package act as a basic api wrapper around git. It's used for [resolving refs as commit shas](https://github.com/github/git-lfs/blob/064043040e1003b51f86b65c14d3f9bedc137f30/git/git.go#L63-L64), [listing remotes](https://github.com/github/git-lfs/blob/064043040e1003b51f86b65c14d3f9bedc137f30/git/git.go#L52-L57), etc. I don't want to have to deal with parsing the trace statements out of the output.

The `git` package could probably be replaced with a libgit2 binding, but I don't think it's worth hassle of dealing with a cgo dependency.

technoweenie:
WOW, I thought you were replying to #971 for some reason. Ignore everything I said in my last reply. I'll take a look at `pre-push`.

technoweenie:
Thanks for the feedback!  I added it to `pre-push` as you suggested.

PR #975, opened 2016-02-03
pass gitattributes patterns (mostly) unchanged

### LINKED ISSUE #870: git lfs track "/somedir/*" error
I think a slash at the beginning of a pattern is valid and indicates the root of the repository. This is explicitly stated in [.gitignore](https://github.com/git/git/blob/f5b6079871904ba5b0a8548f91545f126caf898b/Documentation/gitignore.txt#L112-L114) but seems to work in the same way for .gitattributes.

git-lfs 1.1.0 returns " is outside repository" in that case.

technoweenie:
Attempt to fix #870. The `git lfs track` command currently checks given patterns against the current working directory, preventing users from tracking files outside the repository. As pointed out in #870, this is actually incorrect. Via [gitignore docs](https://github.com/git/git/blob/f5b6079871904ba5b0a8548f91545f126caf898b/Documentation/gitignore.txt#L112-L114):

> A leading slash matches the beginning of the pathname. For example, `/{asterisk}.c` matches `cat-file.c` but not `mozilla-sha1/sha1.c`.

This removes that check, and the related tests.

### DIFF
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -56,36 +56,27 @@ func trackCommand(cmd *cobra.Command, args []string) {
 	}
 
 	wd, _ := os.Getwd()
+	relpath, err := filepath.Rel(lfs.LocalWorkingDir, wd)
+	if err != nil {
+		Exit("Current directory %q outside of git working directory %q.", wd, lfs.LocalWorkingDir)
+	}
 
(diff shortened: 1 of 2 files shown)


larsxschneider:
:+1:

PR #976, opened 2016-02-04
Disable CGO for Centos 5 to assist #952

andyneff:


technoweenie:
Thanks, but now I realize it didn't affect the dockerfiles, since they're on docker hub :) Hopefully https://github.com/andyneff/git-lfs_dockers/pull/1 should work.

andyneff:
This fix won't actually work apparently. `rhel` isn't defined in RHEL 5 ;) I'll fix it in an upcoming PR

PR #977, opened 2016-02-04
Tunable timeouts

technoweenie:
Allow clients to tweak the lfs http timeouts through the git config. It also bumps the normal timeout from 5s to 30s. I think the newer values make more sense for a command line tool.


larsxschneider:
Thank you :+1:

Issue #996, opened 2016-02-11
git lfs client pre-push operation leaves zombies lying around


rlaakkol:
I have a very large repository where I have used BFG repo cleaner to convert files over the entire history to use git lfs. Otherwise this is working nicely, but when I push this bare repository, it will take a very long time (this is not the problem), as I suppose it does this for all (11k) commits:
```
trace git-lfs: run_command: git rev-list --objects ... --not --remotes=origin
trace git-lfs: run_command: git cat-file --batch-check
trace git-lfs: run_command: git cat-file --batch
```

The problem is, that after each trace print like this, 3 more zombies ([git] <defunct>) appear in my process table. And with 11k commits this means that I will run out of PIDs before the operation can finish. I do not know what the processes actually are, but limiting concurrent transfers to 1 did not change the amount of zombie processes generated.

This can probably be considered to be a bug in the client, but if someone comes up with a workaround, I'd be more than happy to try it out. Is there for example a way to limit the git lfs pre-push operation to a subset of commits that need to be handled by it? There is only a handful of commits that actually introduce changes to files that need to be handled by lfs.

rlaakkol:
Just a thought: The startCommand function that is used by revListShas, catFileBatchCheck and catFileBatch uses exec.Start to run the external git command, but exec.Wait may be never called to clean the resources. Does this sound reasonable?

rlaakkol:
I fiddled around a bit and put some cmd.Cmd.Wait()s in convenient places, and no more zombies!

technoweenie:
I haven't heard of or noticed zombie git procs. That's a great analysis of the issue. Do you have a diff or PR for it?

rlaakkol:
I created a PR!

You should probably notice the zombies when pushing multiple commits for a repo that takes sufficiently long to handle by either monitoring `top` or `ps aux | grep "[d]efunct"`

technoweenie:
Just merged #1012. Thanks so much for reporting this, and looking into a fix!

PR #999, opened 2016-02-12
Added zombie cleanups for external git command calls.

rlaakkol:
For issue #996 

When pushing a large amount of commits, an increase in the number of zombie processes can be detected from the process table, specifically 3 new zombies for each commit that is handled. This was fixed by calling exec.Wait() for the commands that were called using exec.Start().

I'm not entirely sure if the calls for exec.Wait() are in their appropriate places here, but at least this seems to do the trick for me.

### DIFF
--- a/lfs/scanner.go
+++ b/lfs/scanner.go
@@ -294,6 +294,7 @@ func revListShas(refLeft, refRight string, opt *ScanRefsOptions) (chan string, e
 			revs <- sha1
 		}
 		close(revs)
+		cmd.Cmd.Wait()
 	}()
 
 	return revs, nil
@@ -344,6 +345,7 @@ func revListIndex(cache bool, indexMap *indexFileMap) (chan string, error) {
 			}
 		}
 		close(revs)
+		cmd.Cmd.Wait()
 	}()
 
 	return revs, nil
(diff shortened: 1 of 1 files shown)


lfs/scanner.go
```diff
@@ -344,6 +345,7 @@ func revListIndex(cache bool, indexMap *indexFileMap) (chan string, error) {
 			}
 		}
 		close(revs)
+		cmd.Cmd.Wait()
```
rlaakkol: Hmm, this might cause some build failures. No idea why, though...

technoweenie:
I'm not sure how this is actually helping. Those `Wait()` calls are happening in goroutines. Go programs drop all goroutines on the floor when `main()` exits.

rlaakkol:
No the goroutines are not the problem, but the child processes launched by `exec.Start()` for those git commands are. This was my quick and dirty fix, as calling `Wait()` before the stdins from the commands can be closed will cause problems. Another solution could be to pass the `cmd` pointers to the calling function `ScanRefsToChan()` from each of the three functions and call the `Wait()`s there after the inputs and outputs of the git commands have been properly handled.

rlaakkol:
Hmm, actually I think the problem in issue #1005 was because of these changes after all. My Wait()s may be closing something too soon and the stdout scanners in the goroutines don't receive everything. Moving the waits into these reading goroutines seems to work better!

Issue #766, opened 2015-10-21
No way to share large content between repo directories.


jwittner:
My objective in our CI is to be able to sync multiple branches one to one with folders, but right now that means a new download of our large content - roughly 15 GB - for every new branch. Seems like I should be able to move the storage for the local media and it's metadata out to a shared folder which git-lfs could leverage.  As far as I can tell there is no way to adjust the local media directory that reports from "git lfs env". If so then this might just be a documentation clarity issue. I'm not certain if there are implementation details preventing this either.

technoweenie:
I'd be into allowing the `lfs.LocalMediaDir` constant to change based on an environment variable. It wouldn't be difficult:

https://github.com/github/git-lfs/blob/f56cd54ad7f8e13df6415a568f5681394bd44b31/lfs/lfs.go#L99-L119

This would lead to a shared LFS directory on each CI server. I also wanted to experiment with some kind of LFS proxy, which could let you have your own shared storage for ALL your CI servers. This could greatly reduce the bandwidth with your upstream LFS server if it isn't already hosted inside your network.

jwittner:
Shared folders and the environment variable could get you close to a single storage location solution.

jwittner:
What about driving lfs.LocalMediaDir via the git-config - it would match the ways we configure the endpoint in git-lfs.

git config --add lfs.localMediaDir <path>?

technoweenie:
I like the `git config` option. Easy to configure per repo or globally :+1:

tristanz:
I have workflows that would also be helped by this option, so :+1: .

sinbad:
Have you tried using `git worktree`? Git LFS already supports that so if you create a separate worktree on the other branch, the `lfs/objects` directory will be shared between them just like the git repo itself. Another advantage is that this will automatically be respected by other functionality like the upcoming `git lfs prune`.

tristanz:
That's a good suggestion, but my applications is slightly different.  I want a local cache of all my objects across all my projects.  We have large binaries that are shared across multiple projects and get added and removed fairly frequently.  Ideally there would be a global LRU cache of objects.  The folder for this cache could be configured per project.

technoweenie:
Sounds like there are two use cases here:

1. Multiple clones for a CI server. Git worktrees may work here if the CI server supports them.
[...]

Issue #515, opened 2015-07-23
I'd like a local mirror/proxy/cache to improve performance.


jsh:
Instead of having to pull down every object from the remote server, every time, I'd like to have local copies without having to maintain and rely on my own server.

I can imagine several flavors of solutions that provide this, but an example would be a local cache: when I request an object during a clone, git-lfs could look in the cache and, if the object is there, return it immediately. If not, it would pull from the server and puts a copy in the cache, too.

We have projects where lots of developers are using a single server, and all of them pull hundreds of megs of files from the server on every clone. It'd be nice to make those clone operations lighter-weight and faster.

technoweenie:
So, the easy option is to spin up a LFS server, and manually configure it in `lfs.url` for your repo. This will require some additional config:

* Extra user/name pass for local LFS server.
* Some mapping so the LFS server knows what auth token to use for the remote LFS server (whether that's GitHub or whatever).

Oh yeah, someone needs to write this intelligent LFS cache server :)

Another idea would be a way to configure some proxy/cache in Git LFS somehow. I'm not totally sure how this would work, but it's something I'm interested in.

PR #1007, opened 2016-02-15
Add support for reference clone

### LINKED ISSUE #459: git-lfs should make use of a previous repo when I use git clone --reference
Using --reference [previous-repo] makes many checkouts 10x faster. But with lfs files, it still downloads them from the server. I wish the --reference argument would also hint to the lfs system to look there for files it can reuse.

jlehtnie:
Fixes #459 and possibly helps for #515 and #766.


technoweenie:
Thanks for the PR. I'd love to get this into v1.2 (#844). I don't have a lot of experience with git references though, so it may take a little bit to review and test this properly though.

lfs/pointer_smudge.go
```diff
@@ -41,6 +41,17 @@ func PointerSmudge(writer io.Writer, ptr *Pointer, workingfile string, download
...
+	if statErr != nil || stat == nil {
+		altMediafile := LocalReferencePath(ptr.Oid)
+		if altMediafile != "" && FileExistsOfSize(altMediafile, ptr.Size) {
+			copyErr := LinkOrCopy(altMediafile, mediafile)
+			if copyErr == nil {
+				stat, statErr = os.Stat(mediafile)
+			}
+		}
+	}
+
```
sinbad: I think this should be wrapped in a utility function (same sequence used in fetch too, DRY)

sinbad:
This is looking good; I'd like to do a bit more testing with it too but I'd support this for 1.2. It's similar to the 'shared object store' idea I've tinkered with before, but I like that this uses an existing git principle. I'm also glad to see that hard links are now supported on Windows via os.Link in go 1.5; that wasn't the case in earlier Go versions and I've had write my own Windows implementation before. Benefitting from that does depend on NTFS partitions but your fallbacks cope with that. :+1:

lfs/util.go
```diff
@@ -310,3 +310,35 @@ func FileExistsOfSize(path string, sz int64) bool {
...
+	_, err = io.Copy(out, in)
+	if err != nil {
+		return err
+	}
+	err = out.Sync()
+	return err
+}
+
+func LinkOrCopy(src string, dst string) error {
+	err := os.Link(src, dst)
```
sinbad: I would make this safer by doing 2 things:

1. Validate using `os.SameFile` to make sure `src` and `dst` are not the same
2. Instead of calling `CopyFileContents` with `dst` as the direct output, instead copy to a temp file and then `os.Rename` on success. This will be more robust under unusual cases like forced termination.

sinbad:
Other than the safety suggestions this works great for me. It actually behaves better than core git, which can potentially break if the referenced repo is garbage collected on shared objects (can't happen here because the local repo LFS object is either a hard link or a copy). Nice work!

lfs/util.go
jlehtniemi-broadsoft: @sinbad Thanks for your feedback. About the temporary file location: do you think it would be better to use `os.TempDir()` or store the tempfile in the destination directory? Latter is probably better from performance point of view because `os.TempDir()` might be in a different file system.

lfs/lfs.go
```diff
@@ -207,3 +232,17 @@ func traceHttpReq(req *http.Request) string {
...
+		altMediafile := LocalReferencePath(oid)
+		mediafile, err := LocalMediaPath(oid)
+		if err != nil {
+			return err
+		}
+		if altMediafile != "" && FileExistsOfSize(altMediafile, size) {
+			return LinkOrCopy(altMediafile, mediafile)
+		}
+	}
+	return nil
```
technoweenie: I tend to prefer smaller guard clauses. What about:

```go
func IfNoLocalObjExistsLinkOrCopyFromReferenceMedia(oid string, size int64) error {
    if ObjectExistsOfSize(oid, size) {
        return nil
    }

    // rest of the function
}
```

Also, I'm not thrilled with the name `IfNoLocalObjExistsLinkOrCopyFromReferenceMedia`. What about just `LinkOrCopyFromReference` ?

lfs/util.go
technoweenie: Use `lfs.TempDir`, which is defined [here](https://github.com/github/git-lfs/blob/9c7155dc780b47b704d21309c12eceef797fc893/lfs/lfs.go#L26), and set to `.git/lfs/tmp` [here](https://github.com/github/git-lfs/blob/9c7155dc780b47b704d21309c12eceef797fc893/lfs/lfs.go#L101).

commands/command_fetch.go
```diff
@@ -278,6 +278,9 @@ func fetchAndReportToChan(pointers []*lfs.WrappedPointer, include, exclude []str
 		// This avoids previous case of over-reporting a requirement for files we already have
 		// which would only be skipped by PointerSmudgeObject later
 		passFilter := lfs.FilenamePassesIncludeExcludeFilter(p.Name, include, exclude)
+
+		lfs.IfNoLocalObjExistsLinkOrCopyFromReferenceMedia(p.Oid, p.Size)
```
technoweenie: This should be added to the appropriate place in `command_smudge.go` too. Possibly [here](https://github.com/github/git-lfs/blob/70e5df64acdcb4f86077e1399b0df0925435ed9b/commands/command_smudge.go#L39).

technoweenie:
First, thanks again for posting this, and for your patience while I take my sweet time reviewing it. I have two bigger pieces of feedback:

According to [gitrepository-layout](https://git-scm.com/docs/gitrepository-layout), `objects/info/alternates` can include multiple alternates. They can also come from `$GIT_ALTERNATE_OBJECT_DIRECTORIES`.  Instead of setting `LocalReferenceDir`, I wonder if this could be added to `lfs.Configuration` instead. For example, the netrc support [reads from the env](https://github.com/github/git-lfs/blob/9c7155dc780b47b704d21309c12eceef797fc893/lfs/config.go#L215-L227) and caches the [parsed file](https://github.com/github/git-lfs/blob/8f90f594162d0a7e90e2f5785287f4327f097235/lfs/config_netrc.go#L20-L32). Then `IfNoLocalObjExistsLinkOrCopyFromReferenceMedia()` would loop through all the possible alternates, returning the first matching object.

The second, is that I'm not sure we need to link/copy those files from alternates to the current repository. It could result in a lot of duplicate files. It seems like `git lfs smudge` should just output the file from whatever path is necessary.

I'd say once the safety issue is resolved, and `git lfs smudge` supported, we could ship this as-is for v1.2, and improve/refactor things for v1.3.

sinbad:
> The second, is that I'm not sure we need to link/copy those files from alternates to the current repository. It could result in a lot of duplicate files. It seems like git lfs smudge should just output the file from whatever path is necessary.

When it's a hardlink it won't actually be a duplicate, just a reference counted file object shared in both places. The only time this wouldn't work & fall back on a copy is if the reference is on a separate filesystem, or on Windows if you're using FAT instead of NTFS (should be pretty unlikely now). 

I actually like the way it's done here better than the way git does it with `.git/objects`, which explicitly says that if you orphan objects in the reference repo they may disappear and therefore break referencing git repos. The link/copy stops this happening although in theory you would have pushed the lfs objects before orphaning them so it would just result in a repeat download in the referencing repo instead. Still, I quite like the inbuilt extra safety.

technoweenie:
I think the safety aspect is less important here than in Git. Git _needs_ every object locally, otherwise basic operations don't work. The only thing Git LFS absolutely must have in the local `.git/lfs/objects`, are objects that are waiting to be pushed for the first time. Everything else can be read from the git reference/alternate, or downloaded from the server.

commands/command_fetch.go
jlehtnie: Is this early call solely for the `smudgeInfo`? I mean it will be called anyway later on by `PointerSmudge`.

jlehtnie:
Personally I think that both of you have valid points here and maybe the behaviour should be configurable. E.g. one typical use case would be that the reference repository would be located on a network share. Now, if one is cloning a personal working repository to a laptop, copying the content to the local disk would be naturally preferable. On the other hand on a CI server it could safe lots of space if one could avoid the data duplication and simply refer to the network share.

technoweenie:
> Now, if one is cloning a personal working repository to a laptop, copying the content to the local disk would be naturally preferable.

That's a great use case for the current behavior. I'm cool with shipping this now, and adding my suggestion later possibly :+1:

lfs/util.go
```diff
@@ -326,3 +327,39 @@ func FileExistsOfSize(path string, sz int64) bool {
...
 }
+
+func CopyFileContents(src string, dst string) error {
+	tmp, err := ioutil.TempFile(TempDir, filepath.Base(dst))
+	if err != nil {
+		return err
+	}
+	defer func() {
+		tmp.Close()
+		_ = os.Remove(tmp.Name())
```
technoweenie: Not worth changing unless you have to make another change, but I tend to omit the `_ =` if I'm not going to do anything with it. Just some microfeedback :)

lfs/util.go
```diff
@@ -326,3 +327,39 @@ func FileExistsOfSize(path string, sz int64) bool {
...
+	in, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer in.Close()
+	_, err = io.Copy(tmp, in)
+	if err != nil {
+		return err
+	}
+	err = tmp.Sync()
```
technoweenie: Why `Sync()` instead of `Close()`? Seems like you'd want `Sync()` in cases where you want to flush to disk while keeping the handle open. Seems like we just want to close it here.

technoweenie:
Thanks for all the tweaks! I still think there are more improvements that can be made, and will add them to the roadmap if you or someone else decides to take it on.

jlehtnie:
Thank you once more both of you for your valuable reviews! I'm happy to get it merged and I hope I will have the opportunity to work on the improvements later on too.

PR #1008, opened 2016-02-16
2nd attempt at zombie eradication

rlaakkol:
Adding the Wait()s in the goroutines which handled the stdin:s of the processes caused some random failures, but this way seems to work better!

### DIFF
--- a/lfs/scanner.go
+++ b/lfs/scanner.go
@@ -294,6 +294,7 @@ func revListShas(refLeft, refRight string, opt *ScanRefsOptions) (chan string, e
 			revs <- sha1
 		}
 		close(revs)
+		cmd.Cmd.Wait()
 	}()
 
 	return revs, nil
@@ -344,6 +345,7 @@ func revListIndex(cache bool, indexMap *indexFileMap) (chan string, error) {
 			}
 		}
 		close(revs)
+		cmd.Cmd.Wait()
 	}()
 
 	return revs, nil
@@ -388,6 +390,7 @@ func catFileBatchCheck(revs chan string) (chan string, error) {
 			cmd.Stdin.Write([]byte(r + "\n"))
 		}
 		cmd.Stdin.Close()
+		cmd.Cmd.Wait()
 	}()
 
 	return smallRevs, nil
@@ -445,6 +448,7 @@ func catFileBatch(revs chan string) (chan *WrappedPointer, error) {
 			cmd.Stdin.Write([]byte(r + "\n"))
 		}
 		cmd.Stdin.Close()
+		cmd.Cmd.Wait()
 	}()
 
 	return pointers, nil
--- a/lfs/scanner.go
+++ b/lfs/scanner.go
@@ -383,6 +383,7 @@ func catFileBatchCheck(revs chan string) (chan string, error) {
 			}
 		}
 		close(smallRevs)
+		cmd.Cmd.Wait()
 	}()
 
 	go func() {
@@ -390,7 +391,6 @@ func catFileBatchCheck(revs chan string) (chan string, error) {
 			cmd.Stdin.Write([]byte(r + "\n"))
 		}
 		cmd.Stdin.Close()
-		cmd.Cmd.Wait()
 	}()
 
 	return smallRevs, nil
@@ -441,6 +441,7 @@ func catFileBatch(revs chan string) (chan *WrappedPointer, error) {
 			}
 		}
 		close(pointers)
+		cmd.Cmd.Wait()
 	}()
 
 	go func() {
@@ -448,7 +449,6 @@ func catFileBatch(revs chan string) (chan *WrappedPointer, error) {
 			cmd.Stdin.Write([]byte(r + "\n"))
 		}
 		cmd.Stdin.Close()
-		cmd.Cmd.Wait()
 	}()
 
 	return pointers, nil


technoweenie:
> No the goroutines are not the problem, but the child processes launched by `exec.Start()` for those git commands are.

I'm just saying that since Go doesn't wait for goroutines to finish, there's no guarantee those `Wait()` calls will finish. So, I'm not quite sure how this helps.

I'm also not seeing the problem at all. I wonder if it has to do with the fact that I'm running on the mac, or using Go 1.5.3, vs Go 1.4.2 (https://github.com/github/git-lfs/issues/1005#issuecomment-184304425).

rlaakkol:
It happened to me on go 1.5.3 too, so I suppose it's the mac then... It
doesn't matter if the goroutines stop when the parent exits as zombies will
get cleaned at that point anyway (a process can't be a zombie and an orphan
simultaneously, I think)!

technoweenie:
Maybe I'm not understanding the problem, then.  Are you seeing these commands for _every_ commit? 

```
trace git-lfs: run_command: git rev-list --objects ... --not --remotes=origin
trace git-lfs: run_command: git cat-file --batch-check
```

`rev-list` does have to scan the commits that you're pushing. Meaning it scans the entire history on the initial push. However, it should make these 3 calls just once per push. When I run locally, I see 2 `git` procs, and 1 `git-remote-http` proc.  PIDs aren't changing either. If it's making those calls multiple times for a single push, that's a problem.

rlaakkol:
Yeah, that's what it looks like, at least according to the amount of trace prints. I'll provide some trace output when I get back to work tomorrow! Maybe the problem is in my repo... I used the bfg repo cleaner (https://rtyley.github.io/bfg-repo-cleaner/) with the --convert-to-git-lfs switch and some custom index-filters to edit the history, so something might go wrong at that point, maybe?

rlaakkol:
I'll try to get more useful output in a while, but maybe this gives you a hint:
When I do the initial git push, it looks like this:
...
Git LFS: (0 of 991 files, 991 skipped) 0 B / 5.96 GB, 5.96 GB skipped
Git LFS: (0 of 991 files, 991 skipped) 0 B / 5.96 GB, 5.96 GB skipped
Git LFS: (0 of 991 files, 991 skipped) 0 B / 5.96 GB, 5.96 GB skipped
Git LFS: (0 of 914 files, 914 skipped) 0 B / 5.42 GB, 5.42 GB skipped                                                 
Git LFS: (0 of 991 files, 991 skipped) 0 B / 5.96 GB, 5.96 GB skipped                                                 
Git LFS: (0 of 914 files, 914 skipped) 0 B / 5.42 GB, 5.42 GB skipped                                                 
Git LFS: (0 of 991 files, 991 skipped) 0 B / 5.96 GB, 5.96 GB skipped
...
And at least when I tested with a smaller repo, the amount of those lines was about the same as the number of commits in the repo...

rlaakkol:
Here's a short bit of a push trace output:
stdout:
$ GIT_TRACE=1 git -c http.sslVerify=false push -f 2> ../trace.txt
Git LFS: (0 of 1606 files, 1606 skipped) 0 B / 18.84 GB, 18.84 GB skipped
Git LFS: (0 of 1798 files, 1798 skipped) 0 B / 20.67 GB, 20.67 GB skipped
Git LFS: (0 of 1744 files, 1744 skipped) 0 B / 19.68 GB, 19.68 GB skipped
Git LFS: (0 of 1518 files, 1518 skipped) 0 B / 16.92 GB, 16.92 GB skipped
Git LFS: (0 of 1502 files, 1502 skipped) 0 B / 16.56 GB, 16.56 GB skipped
^C
[stderr.txt](https://github.com/github/git-lfs/files/133840/stderr.txt)

technoweenie:
Ah, I bet you're pushing multiple branches because it's an initial push. We have to run those 3 commands. Even if your branches share a lot of history, git lfs still has to sift through the entire history for LFS objects. Those skipped objects were likely uploaded in another ref.

I tried adding the `Wait()` calls _before_ closing the channels ([diff](https://github.com/github/git-lfs/files/134816/lfs.diff.txt)), but the behavior is the same on my mac. The same 3 extra git processes stay running for the whole multi-ref push operation.

The interesting thing here is that the `Wait()` calls returned immediately, but the processes stick around in activity monitor and `ps`. The "DONE" messages print after the `*exec.Command` `Wait()` function returns.

```
$ git push ghe --all
PRE PUSH START 16f09066000e3328fb2d5c54beb55fc25c1af15c => ^0000000000000000000000000000000000000000
DONE: /Library/Developer/CommandLineTools/usr/libexec/git-core/git [git rev-list --objects 16f09066000e3328fb2d5c54beb55fc25c1af15c --not --remotes=ghe]
DONE: /Library/Developer/CommandLineTools/usr/libexec/git-core/git [git cat-file --batch-check]
DONE: /Library/Developer/CommandLineTools/usr/libexec/git-core/git [git cat-file --batch]
Git LFS: (501 of 501 files) 1.85 KB / 1.85 KB
PRE PUSH STOP 16f09066000e3328fb2d5c54beb55fc25c1af15c => ^0000000000000000000000000000000000000000
PRE PUSH START 16f09066000e3328fb2d5c54beb55fc25c1af15c => ^0000000000000000000000000000000000000000
DONE: /Library/Developer/CommandLineTools/usr/libexec/git-core/git [git rev-list --objects 16f09066000e3328fb2d5c54beb55fc25c1af15c --not --remotes=ghe]
DONE: /Library/Developer/CommandLineTools/usr/libexec/git-core/git [git cat-file --batch-check]
DONE: /Library/Developer/CommandLineTools/usr/libexec/git-core/git [git cat-file --batch]
Git LFS: (0 of 501 files, 501 skipped) 0 B / 1.85 KB, 1.85 KB skipped
```

BTW, you don't need the extra `Cmd` in `cmd.Cmd.Wait()`. The `Cmd` property is embedded in the wrapped struct:

```go
type wrappedCmd struct {
	Stdin  io.WriteCloser
	Stdout *bufio.Reader
	*exec.Cmd
}
```

You can tell because the `*exec.Cmd` declaration has no property name. That means all public exported properties and functions of `*exec.Cmd` are available in `*wrappedCmd`. You can then wrap those public functions, like I did with `Wait()` in my diff above.

technoweenie:
Managed to replicate this in an ubuntu vm. I did a normal push with the master version LFS, while watching the processes with `top -bc | grep git | grep vagrant`.

Saw this:

```
27419 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27420 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27421 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27472 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27473 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27474 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27525 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27526 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27527 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27576 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27577 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27578 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27629 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27630 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27631 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27695 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27696 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27697 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27748 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27749 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
27750 vagrant   20   0       0      0      0 Z  0.0  0.0   0:00.00 [git] <defunct>
```

Once I tried with your patch, I only saw one of those defunct git processes.

I also swapped the order of the `Wait()` calls. The `lfs.ScanRefs()` function is synchronous, and returns once all the downloads are done. It blocks while [looping through this channel](https://github.com/github/git-lfs/blob/7f63e585eeda53a4157b0a8feecd4160e3e1b303/lfs/scanner.go#L112-L114), until it's closed. This should ensure that the command executes before the channel is closed, unblocking the for loop in `lfs.ScanRefs()`.

Also it likely means there's at least one more process still sticking around. I'll add some more tracing for it.

rlaakkol:
Great! I appreciate the thorough explanations and go tips!

PR #1012, opened 2016-02-17
the walking dead (pt 3)

technoweenie:
Building on #1008 (/cc @rlaakkol). Just moved the `Wait()` call before closing the channels, as described in https://github.com/github/git-lfs/pull/1008#issuecomment-185353312.

### DIFF
--- a/lfs/scanner.go
+++ b/lfs/scanner.go
@@ -293,8 +293,9 @@ func revListShas(refLeft, refRight string, opt *ScanRefsOptions) (chan string, e
 			}
 			revs <- sha1
 		}
+
+		cmd.Wait()
 		close(revs)
-		cmd.Cmd.Wait()
 	}()
 
 	return revs, nil
@@ -344,8 +345,9 @@ func revListIndex(cache bool, indexMap *indexFileMap) (chan string, error) {
 				revs <- sha1
 			}
 		}
+
+		cmd.Wait()
 		close(revs)
-		cmd.Cmd.Wait()
 	}()
 
 	return revs, nil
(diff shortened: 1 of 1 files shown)


technoweenie:
FWIW, I audited the app's other uses of the `os/exec` package. Looks like we're calling `Wait()` everywhere else. I'm not sure what that extra defunct git process is from.  I added trace messages around `startCommand()` in `lfs/scanner.go` ([diff](https://github.com/github/git-lfs/files/135009/lfs.diff.txt)) and saw every command it started was finishing properly:

```bash
$ git push ghe --all
CMD START /Library/Developer/CommandLineTools/usr/libexec/git-core/git [git rev-list --objects 16f09066000e3328fb2d5c54beb55fc25c1af15c --not --remotes=ghe]
CMD START /Library/Developer/CommandLineTools/usr/libexec/git-core/git [git cat-file --batch-check]
CMD START /Library/Developer/CommandLineTools/usr/libexec/git-core/git [git cat-file --batch]
CMD STOP /Library/Developer/CommandLineTools/usr/libexec/git-core/git [git rev-list --objects 16f09066000e3328fb2d5c54beb55fc25c1af15c --not --remotes=ghe]
CMD STOP /Library/Developer/CommandLineTools/usr/libexec/git-core/git [git cat-file --batch-check]
CMD STOP /Library/Developer/CommandLineTools/usr/libexec/git-core/git [git cat-file --batch]
```

rubyist:
The waits look right to me :+1:

sinbad:
Yeah, looks good - I bet Windows was especially susceptible to this, I've had issues before of git leaving files locked sometimes because of things like this. Nice one.

technoweenie:
Thanks for the reviews!

PR #1016, opened 2016-02-19
update "ls-tree" command to use -z

technoweenie:
This updates the `ls-tree` command parser to use `-z`, which tells git to return lines split by a null terminator. This output is designed for scripts, and represents filenames with spaces and unicode better. See https://github.com/github/git-lfs/issues/987#issuecomment-186337634 for a more detailed writeup.

/cc @sinbad: I think we should try and use `-z` where possible. I copied `bufio.ScanLines()` and made `scanNullLines()`, so it should be easy to apply to other commands.

I also made sure `lsTreeBlobs()` calls `Wait()` on the `*exec.Command` object. Don't want another zombie herd (#1012).

### DIFF
--- a/lfs/scanner.go
+++ b/lfs/scanner.go
@@ -606,7 +606,7 @@ func lsTreeBlobs(ref string) (chan TreeBlob, error) {
 
 func parseLsTree(reader io.Reader, output chan TreeBlob) {
 	scanner := bufio.NewScanner(reader)
-	scanner.Split(ScanNullLines)
+	scanner.Split(scanNullLines)
 	for scanner.Scan() {
 		line := scanner.Text()
 		parts := strings.SplitN(line, "\t", 2)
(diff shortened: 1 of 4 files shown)


sinbad:
OK, FYI a more explicit way of dealing with encoding issues like this is to use `-c core.quotepath=false` as part of the git command. I believe `-z` just makes this implicit.

technoweenie:
Ah, I'm just used to using `-z` for this stuff. Pretty sure I fixed a similar bug in grit when parsing tree entries with filenames that include linebreaks. People do weird stuff in their repos :)

sinbad:
Yeah, whatever you prefer, I don't mind. I'm used to using core.quotepath=false because nulls have special meaning in C (string termination) so are kind of a PITA, plus most existing parsing utilities like newlines.

PR #1017, opened 2016-02-19
kill more process zombies

technoweenie:
Found a few more places that weren't explicitly calling `Wait()`. /cc #1012 @sinbad


sinbad:
:+1:

PR #1023, opened 2016-02-23
Better errors on push

technoweenie:
This fixes an issue when pushing or fetching where the "inner error" is dropped, making it difficult to diagnose problems.

This also tweaks an error message from `ensureFile()` /cc @rjbell4

```bash
$ git push origin master
Trying to push "a.wav" with OID 87428fc522803d31065e7bce3cf03fe475096631e5e07bbd7a0fde60c4cf25c7.
Not found in /Users/rick/p/studious-goggles/.git/lfs/objects/87/42.
error: failed to push some refs to 'https://git-server.com/technoweenie/studious-goggles'
```

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -57,6 +57,17 @@ func Exit(format string, args ...interface{}) {
 	os.Exit(2)
 }
 
+func ExitWithError(err error) {
+	if Debugging || lfs.IsFatalError(err) {
+		Panic(err, err.Error())
+	} else {
+		if inner := lfs.GetInnerError(err); inner != nil {
+			Error(inner.Error())
+		}
+		Exit(err.Error())
+	}
+}
+
 // Debug prints a formatted message if debugging is enabled.  The formatted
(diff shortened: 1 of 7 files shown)


commands/commands.go
```diff
@@ -57,6 +57,17 @@ func Exit(format string, args ...interface{}) {
 	os.Exit(2)
 }
 
+func ExitWithError(err error) {
+	if Debugging || lfs.IsFatalError(err) {
+		Panic(err, err.Error())
+	} else {
+		if inner := lfs.GetInnerError(err); inner != nil {
```
sinbad: Technically err.InnerError can be another `lfs.wrappedError` so maybe this should be a loop until that's nil?

technoweenie: No need because of [this](https://github.com/github/git-lfs/blob/3daf3d4a135a6e42f0e9f0b45cb60a2d922d27c5/lfs/errors.go#L296-L301). Good call though, I added a test confirming it.

Issue #1021, opened 2016-02-22
Build fails with go 1.6


jacobmischka:
Running `./scripts/bootstrap` fails to build for me on x64 Arch, no `bin` directory is created afterward. 

Output: https://gist.github.com/jacobmischka/0bd07fc66bfded7cd1e2

`go version go1.6 linux/amd64`

technoweenie:
Oh, they enabled the vendor stuff in 1.6. You'll have to use go 1.5.3, or use one of the precompiled ones in the release.

rubyist:
FWIW, you can build with 1.6 if you set `GO15VENDOREXPERIMENT=0` to turn off the vendoring stuff.

jacobmischka:
Okay, that works thank you. Will you be planning on updating the package to work with 1.6 by default soon, or should I expect to need to use that env variable for the foreseeable future?

rubyist:
I think nut hardcodes `vendor` so we'd probably have to move away from that (or it'd have to change). We could also just set the var in the bootstrap script, though it might resurface again with 1.7 if they drop the env var entirely (I thought I saw that somewhere, but I could be mistaken).

jacobmischka:
Okay, well this works for now, thanks again!

technoweenie:
Builds should work on 1.6 without specifying any env vars now: #1025

PR #1024, opened 2016-02-23
Interim update for basic compatibility with go 1.6

sinbad:
Addresses #1021 in the most minimal way possible for now.

I suggest moving to standard Go vendoring and flipping this option to `GO15VENDOREXPERIMENT=1` but that will be a bigger change. We can either use `git submodule` or `git subtree` but personally I'd prefer `subtree` because it has fewer edge cases (and we've embedded the vendor packages in this repo until now anyway). I can do that as a follow up PR if you agree.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -1,4 +1,8 @@
 #!/usr/bin/env bash
+
+# Compatibility with go 1.6
+export GO15VENDOREXPERIMENT=0
+
 set -e
 
 # Only set GOPATH on non-Windows platforms as Windows can't do the symlinking
--- a/script/release
+++ b/script/release
@@ -1,3 +1,7 @@
 #!/usr/bin/env bash
+
+# Compatibility with go 1.6
+export GO15VENDOREXPERIMENT=0
+
 script/fmt
 go run script/*.go -cmd release "$@"
--- a/script/run
+++ b/script/run
@@ -1,4 +1,8 @@
 #!/usr/bin/env bash
+
+# Compatibility with go 1.6
+export GO15VENDOREXPERIMENT=0
+
 script/fmt
 commit=`git rev-parse --short HEAD`
 go run -ldflags="-X github.com/github/git-lfs/lfs.GitCommit $commit" ./git-lfs.go "$@"
--- a/script/test
+++ b/script/test
@@ -2,6 +2,9 @@
 #/ Usage: script/test          # run all tests
 #/        script/test <subdir> # run just a package's tests
 
+# Compatibility with go 1.6
+export GO15VENDOREXPERIMENT=0
+
 script/fmt
 suite="./${1:-"lfs"} ./${1:-"git"}"
 if [ $# -gt 0 ]; then
--- a/test/testenv.sh
+++ b/test/testenv.sh
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 # Including in script/integration and every test/test-*.sh file.
 
+# Compatibility with go 1.6
+export GO15VENDOREXPERIMENT=0
+
 set -e
 
 # The root directory for the git-lfs repository by default.


technoweenie:
Doh! #1025

sinbad:
Lol

technoweenie:
> We can either use git submodule or git subtree but personally I'd prefer subtree because it has fewer edge cases (and we've embedded the vendor packages in this repo until now anyway). I can do that as a follow up PR if you agree.

We can keep the existing layout with nut by changing the imports matching `github.com/github/git-lfs/vendor/_nuts/*` to just `_nuts/*`.

sinbad:
I guess, but it would be good to not need nut any more, especially has it doesn't appear to be being maintained any more.

technoweenie:
Yeah, that's true. My only requirements are:

* Some kind of file showing packages and git versions
* Files in `vendor/host/user/repo/*`

I'd probably prefer subtree too, but I've never used it. I typically just do `git add vendor/host/user/repo/`. We don't necessarily need the commits, but it doesn't hurt, I guess.

rubyist:
If we just modify the import path, we'll have to do a bunch of fix ups when nut is run, because it'll always dump to vendor/. Actually, it doesn't look like nut even properly installs on a 1.6 setup, unless you turn off vendoring, because it uses nut itself.

technoweenie:
Ah, I hadn't even tried running nut itself on go 1.6. I'm cool with dropping nut. 

Before making a big change, I want to make sure we can still build all the linux packages too. I'm afraid that some centos doesn't have go 1.5 yet.

sinbad:
Subtree doesn't store much reference data by default. You can extract it from the history by looking for git-subtree-dir and git-subtree-split commit entries, but we could just choose to use a custom file similar to Nut.toml with just package/commit pairs, and have the vendor script run the appropriate `git subtree` commands for us based on that?

technoweenie:
Sounds good :+1:

@andyneff Are we stuck on Go 1.4 for older centos or debian versions? I think debian just bundles the static binary, but it looks like the [centos builds depend on golang](https://github.com/github/git-lfs/blob/e49c3aead2e258ae7b95ac22b50fa63664ffd340/rpm/SPECS/git-lfs.spec#L12).

andyneff:
@technoweenie The golang line is there for technical completeness, but since we wanted the ability to use any version we wanted, I use the `--nodeps` option [here](https://github.com/github/git-lfs/blob/master/rpm/build_rpms.bsh#L179) The dockers all download the desired version of go for [centos5](https://github.com/andyneff/git-lfs_dockers/blob/master/centos_5.bootstrap#L16-L20), [centos6](https://github.com/andyneff/git-lfs_dockers/blob/master/centos_6.bootstrap#L14-L18), [centos7](https://github.com/andyneff/git-lfs_dockers/blob/master/centos_7.bootstrap#L14-L18), [debian7](https://github.com/andyneff/git-lfs_dockers/blob/master/debian_7.dockerfile#L16-L20), and [debian8](https://github.com/andyneff/git-lfs_dockers/blob/master/debian_8.dockerfile#L16-L20). 

So there should be no problem moving forward trying to use other versions of go to build the packages using the dockers. Looks like I may need to add GO15VENDOREXPERIMENT=1???

technoweenie:
For now, the build scripts hard code `GO15VENDOREXPERIMENT=0`, until we update the import paths. But once we do that, we're stuck with go 1.5+. Older versions of Go won't understand those vendored import paths.

I'm in no rush to do this. I think we just have to do it before `GO15VENDOREXPERIMENT=0` stops working, which could be Go 1.7.

sinbad:
:+1: I might tinker with it a bit because I have a personal interest in subtree automation for deps post go 1.6 but it's not high on my list.

PR #1025, opened 2016-02-23
Go 1.6

technoweenie:
This is a quick hack that ensures builds on go 1.6 work. Eventually, I want to start using `GO15VENDOREXPERIMENT=1`.


sinbad:
:+1:

PR #1038, opened 2016-02-24
Show refs

technoweenie:
This is a quick code spike to getting the `git lfs push` command to run on individual local refs when given `--all`. Could be useful for sending ref info in #1036.

This may run into issues if the local branch doesn't match the remote tracking branch...

### DIFF
--- a/commands/command_push.go
+++ b/commands/command_push.go
@@ -24,7 +24,7 @@ var (
 	// shares some global vars and functions with command_pre_push.go
 )
 
-func uploadsBetweenRefs(left string, right string) *lfs.TransferQueue {
+func uploadsBetweenRefs(left string, right string) {
 	tracerx.Printf("Upload between %v and %v", left, right)
 
 	// Just use scanner here
@@ -32,10 +32,11 @@ func uploadsBetweenRefs(left string, right string) *lfs.TransferQueue {
 	if err != nil {
 		Panic(err, "Error scanning for Git LFS files")
 	}
-	return uploadPointers(pointers)
+
+	uploadPointers(pointers)
 }
 
-func uploadsBetweenRefAndRemote(remote string, refs []string) *lfs.TransferQueue {
+func uploadsBetweenRefAndRemote(remote string, refs []string) {
 	tracerx.Printf("Upload refs %v to remote %v", remote, refs)
 
 	scanOpt := lfs.NewScanRefsOptions()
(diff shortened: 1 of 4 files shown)


technoweenie:
Leaving these tests alone. If this and #1040 are satisfactory, I'll merge that first, and then fix these tests.

commands/command_push.go
```diff
@@ -44,16 +45,20 @@ func uploadsBetweenRefAndRemote(remote string, refs []string) *lfs.TransferQueue
...
+			refs = make([]string, len(gitrefs))
+			for idx, gitref := range gitrefs {
+				refs[idx] = gitref.Name
+			}
 		}
-	}
 
-	// keep a unique set of pointers
-	oidPointerMap := make(map[string]*lfs.WrappedPointer)
+		scanOpt.ScanMode = lfs.ScanRefsMode
```
sinbad: I think you meant scanOpt.ScanMode

commands/command_push.go
```diff
@@ -61,22 +66,11 @@ func uploadsBetweenRefAndRemote(remote string, refs []string) *lfs.TransferQueue
...
-			oidPointerMap[p.Oid] = p
-		}
-	}
-
-	i := 0
-	pointers := make([]*lfs.WrappedPointer, len(oidPointerMap))
-	for _, pointer := range oidPointerMap {
-		pointers[i] = pointer
-		i += 1
+		uploadPointers(pointers)
```
sinbad: This going to create a bit more traffic on the API when pushing multiple refs which share unpushed ancestry, and especially so in the --all case. If you take the case where there's a long history and 4 branches in the last few commits, all the oids in the history they shared would be submitted again for each branch. The API would only return links for those that hadn't been completed in the last branch, but still, it's increasing the load quite a bit.

git/git.go
```diff
@@ -153,15 +153,51 @@ func RemoteList() ([]string, error) {
...
 		ret = append(ret, strings.TrimSpace(scanner.Text()))
 	}
+
 	return ret, nil
 }
 
+// Refs returns all of the local and remote branches and tags for the current
+// repository. Other refs (HEAD, refs/stash, git notes) are ignored.
+func LocalRefs() ([]*Ref, error) {
+	cmd := execCommand("git", "show-ref")
```
sinbad: Recommend adding "--heads" and "--tags" to give you just those 2

technoweenie:
Merged this into #1040, so I only have to update the tests once.

PR #1040, opened 2016-02-24
improve initial pre-push imports

technoweenie:
When pushing a repository to a new host, Git LFS pushes each branch or tag individually. This leads to a lot of skipped objects. Assuming all branches are based on master, then there should be a lot of duplicate objects. Here's a push with a repository with 502 objects: 501 in master, and an extra in another branch:

```bash
$ time git push 5 --all
Git LFS: (501 of 501 files) 1.85 KB / 1.85 KB
Git LFS: (0 of 501 files, 501 skipped) 0 B / 1.85 KB, 1.85 KB skipped
Git LFS: (1 of 502 files, 501 skipped) 3 B / 1.86 KB, 1.85 KB skipped
Git LFS: (0 of 501 files, 501 skipped) 0 B / 1.85 KB, 1.85 KB skipped
Counting objects: 516, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (511/511), done.
Writing objects: 100% (516/516), 70.52 KiB | 0 bytes/s, done.
Total 516 (delta 1), reused 512 (delta 0)
To https://git-server/technoweenie/symmetrical-octo-pancake
 * [new branch]      1 -> 1
 * [new branch]      2 -> 2
 * [new branch]      3 -> 3
git push 5 --all  2.06s user 1.51s system 7% cpu 51.051 total
```

With this PR:

```bash
$ time git push 4 --all
Git LFS: (501 of 501 files) 1.85 KB / 1.85 KB
Git LFS: (1 of 1 files) 3 B / 3 B
Counting objects: 516, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (511/511), done.
Writing objects: 100% (516/516), 70.52 KiB | 0 bytes/s, done.
Total 516 (delta 1), reused 512 (delta 0)
To https://git-server/technoweenie/symmetrical-octo-funicular
 * [new branch]      1 -> 1
 * [new branch]      2 -> 2
 * [new branch]      3 -> 3
git push 4 --all  1.62s user 0.89s system 5% cpu 45.280 total
```

### DIFF
--- a/commands/command_pre_push.go
+++ b/commands/command_pre_push.go
@@ -55,6 +55,8 @@ func prePushCommand(cmd *cobra.Command, args []string) {
 	}
 	lfs.Config.CurrentRemote = args[0]
 
+	uploadedOids := lfs.NewStringSet()
+
 	// We can be passed multiple lines of refs
 	scanner := bufio.NewScanner(os.Stdin)
 	for scanner.Scan() {
@@ -69,12 +71,11 @@ func prePushCommand(cmd *cobra.Command, args []string) {
 			continue
 		}
 
-		prePushRef(left, right)
-
+		prePushRef(left, right, uploadedOids)
 	}
 }
 
-func prePushRef(left, right string) {
+func prePushRef(left, right string, uploadedOids lfs.StringSet) {
 	// Just use scanner here
 	scanOpt := lfs.NewScanRefsOptions()
 	scanOpt.ScanMode = lfs.ScanLeftToRemoteMode
@@ -85,17 +86,16 @@ func prePushRef(left, right string) {
 		Panic(err, "Error scanning for Git LFS files")
 	}
 
+	pointers = filteredPointers(pointers, uploadedOids)
+	pointers = filteredPointers(pointers, prePushCheckForMissingObjects(pointers))
+
 	totalSize := int64(0)
 	for _, p := range pointers {
 		totalSize += p.Size
 	}
 
-	// Objects to skip because they're missing locally but on server
-	var skipObjects lfs.StringSet
-
-	if !prePushDryRun {
-		// Do this as a pre-flight check since upload queue starts immediately
-		skipObjects = prePushCheckForMissingObjects(pointers)
+	if totalSize < 1 {
+		return
 	}
 
 	uploadQueue := lfs.NewUploadQueue(len(pointers), totalSize, prePushDryRun)
@@ -103,11 +103,6 @@ func prePushRef(left, right string) {
 	for _, pointer := range pointers {
 		if prePushDryRun {
 			Print("push %s => %s", pointer.Oid, pointer.Name)
-			continue
-		}
-
-		if skipObjects.Contains(pointer.Oid) {
-			// object missing locally but on server, don't bother
 			continue
 		}
 
@@ -120,6 +115,7 @@ func prePushRef(left, right string) {
 			}
 		}
 
+		uploadedOids.Add(pointer.Oid)
 		uploadQueue.Add(u)
 	}
 
@@ -137,7 +133,19 @@ func prePushRef(left, right string) {
 			os.Exit(2)
 		}
 	}
+}
 
+func filteredPointers(pointers []*lfs.WrappedPointer, filter lfs.StringSet) []*lfs.WrappedPointer {
+	filtered := make([]*lfs.WrappedPointer, 0, len(pointers))
+	for _, pointer := range pointers {
+		if filter.Contains(pointer.Oid) {
+			continue
+		}
+
+		filtered = append(filtered, pointer)
+	}
+
+	return filtered
 }
 
 func prePushCheckForMissingObjects(pointers []*lfs.WrappedPointer) (objectsOnServer lfs.StringSet) {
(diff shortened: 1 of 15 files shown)


sinbad:
This only addresses the issue when using the pre-push hook, it'll still be there when using `git lfs push ref1 ref2 ref3` or `git lfs push --all`. So actually my comment in #1038 that I edited to ref this one still stands right now.

commands/command_pre_push.go
```diff
@@ -137,7 +133,19 @@ func prePushRef(left, right string) {
 			os.Exit(2)
 		}
 	}
+}
+
+func filteredPointers(pointers []*lfs.WrappedPointer, filter lfs.StringSet) []*lfs.WrappedPointer {
```
sinbad: Instead of building a brand new list, how about either doing this as a channel or accepting a callback function for unfiltered items?

technoweenie:
> This only addresses the issue when using the pre-push hook

Yup, that's on purpose. Trying to get feedback on the approach. Then we can take it to the fetch and push commands :metal:

commands/command_pre_push.go
technoweenie: This seemed a lot simpler. Though I am curious enough to try writing some benchmark code for it.

sinbad:
Ah fair enough, approach is fine to me, very similar to what was done before just in a different place. I wondered if it could be generalised & centralised instead but I recognise that would probably have to be in the transfer queue which is just going to cause a whole lot of new race condition possibilities. This way is probably better just for sanity.

commands/command_pre_push.go
technoweenie: How about this? https://gist.github.com/technoweenie/34671a1296ba375b477a

```bash
$ go test -bench . -run BenchmarkFilterByChan
testing: warning: no tests to run
PASS
BenchmarkFilterByList-4    	    1000	   1871779 ns/op
BenchmarkFilterByCallback-4	    1000	   1653444 ns/op
BenchmarkFilterByChan-4    	     200	   9165917 ns/op
```

Still leaning towards the current approach. It's not much slower than using a callback, and has the simplest implementation. Unless there's some radical way to make the channel approach simpler and faster, I'm not even really considering it.

sinbad: I meant more like your callback example, channels as output and still lists as input; but I see the issue is that you need 2 iterations with shared reference data which is causing the channel way to be slow in parallel because of the mutex so it's no benefit over the callback; they'd have to be serialised to avoid the mutex, and it's more complex. 

I do prefer the callback style to returning new lists but that's probably the grumpy old man in me seeing 30% more allocations and freaking out.

technoweenie: I tried using callbacks:

```diff
-       pointers = filteredPointers(pointers, uploadedOids)
-       pointers = filteredPointers(pointers, prePushCheckForMissingObjects(pointers))
+       filtered := make([]*lfs.WrappedPointer, 0, len(pointers))
+       cb := func(p *lfs.WrappedPointer) {
+               filtered = append(filtered, p)
+       }
+
+       filteredPointers(pointers, uploadedOids, cb)
+
+       serverOids := prePushCheckForMissingObjects(filtered)
+       filteredPointers(filtered, serverOids, cb)
```

This is more complex, and introduces a concurrency issue because the second filter call is looping through the same slice while the callback is appending to it. The original approach, though 30% more wasteful with allocations, is the best option IMO. Though, that's assuming filtering twice is a good idea. My intention is to filter out everything that has been uploaded from this same process before checking the server to see what it has. It'll result in fewer batch calls in large import pushes.

technoweenie:
> I wondered if it could be generalised & centralised 

Me too! It may or may not happen in this PR. I just wanted to get feedback on this before spending more time on it. I'll reply again when I'm ready for round 2 :)

commands/command_pre_push.go
sinbad: > introduces a concurrency issue because the second filter call is looping through the same slice while the callback is appending to it

I don't think it does - the filteredPointers and prePushCheckForMissingObjects calls are completely synchronous/serial. But in any case, if you're happier with this style it's fine, I just can't help myself looking for allocation optimisations given my background ;)

technoweenie:
Okay, this works with `push --all` now. Basically, instead of scanning with `lfs.ScanAllMode`, it uses `git.LocalRefs()`, and scans each ref separately. I think the behavior is now correct for both `lfs push` and `lfs pre-push`. Though I'd like to take a stab at consolidating the similar push logic between the two commands

commands/command_pre_push.go
sinbad: Now that filteredPointers is used in multiple places, maybe promote it to the lfs package for re-use rather than keep it in the pre-push source file?

sinbad:
Looks good :+1: I've seen that checkout failure that's red-flagging the Travis build before, I think it's unrelated to this PR.

commands/command_pre_push.go
technoweenie: BTW just wanted to acknowledge you're absolutely right about the concurrency thing. Temporary brain fart.

I'm working on an update that attempts to centralize the uploading logic across pre-push and push. I'll probably post that as a PR against this PR so we can review that, and discuss package options there.

technoweenie:
This conflicts with the scanner changes in master, so I'm going to redo it.

Issue #995, opened 2016-02-11
empty/malformed manpages when installed via deb


whitty:
Installed from the packagecloud location on ubuntu trusty the man-pages are malformed or empty - note only 20 bytes long:

```
$ ls -l /usr/share/man/man1/git-lfs*
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs.1.gz
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs-checkout.1.gz
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs-clean.1.gz
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs-env.1.gz
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs-ext.1.gz
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs-fetch.1.gz
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs-fsck.1.gz
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs-init.1.gz
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs-install.1.gz
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs-logs.1.gz
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs-ls-files.1.gz
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs-pointer.1.gz
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs-pre-push.1.gz
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs-prune.1.gz
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs-pull.1.gz
-rw-r--r-- 1 root root 20 Feb  6 05:34 /usr/share/man/man1/git-lfs-push.1.gz
[...]

Issue #719, opened 2015-10-08
Debian package 1.0.0 for jessie 64bit is broken


megies:
Man pages in `usr/share/man/man1/` inside the package are all just empty gzip files.

technoweenie:
Thanks for the report. We'll look into it.

megies:
I bet you're aware of this, just to make double-sure (and since your deb packages don't have a debversion part). Please use an incremented debversion suffix (e.g. `..._1.0.0-1_...`) to make sure updated packages of same software version get installed (semi-)automatically by package management.

ssgelm:
@megies I'll take a look at this.  Thanks for the heads up!

ssgelm:
Sorry for the delay in looking at this.  Something definitely changed between 0.6.0 and 1.0.  Still looking into it.

ssgelm:
According to the git bisect I did this was broken in e970264f007f77384c03a4d5913a1948e83001fd.  For some reason dh_installman seems to be picking up that go file.  I will track it down.

brbsix:
I just got the update to 1.1.0 on Jessie 64bit and noticed man pages are still empty gzip files.

ssgelm:
Sorry I have not fixed it yet. I will try to do so asap!

carsten-AEI:
When I tried to rebuild the 1.1.0 Debian packages from the release tarball I got stuck with 
```
[...]
make[1]: Entering directory `/tmp/git-lfs-1.1.0'
mkdir -p debian/git-lfs/usr/bin
cp obj-x86_64-linux-gnu/bin/git-lfs debian/git-lfs/usr/bin/
make[1]: Leaving directory `/tmp/git-lfs-1.1.0'
   dh_install -O--buildsystem=golang
   dh_installdocs -O--buildsystem=golang
   dh_installchangelogs -O--buildsystem=golang
   dh_installexamples -O--buildsystem=golang
   dh_installman -O--buildsystem=golang
Converting man pages into code...
Failed to create go file: open ../commands/mancontent_gen.go: no such
file or directory
Converting man pages into code...
Failed to open man dir: <nil>
dh_installman: man --recode UTF-8 ./git\-lfs\-checkout\.1 >
git\-lfs\-checkout\.1\.new returned exit code 2
make: *** [binary] Error 2
dpkg-buildpackage: error: debian/rules binary gave error exit status 2
```
where `man` seems to be contained in the tarball.

Maybe this may be related to the man pages never showed up in the first place.

[...]

PR #1045, opened 2016-02-25
Fix for #719 and #995

andyneff:
Removes the stray `man` binary causing havoc for man page creation.


technoweenie:
THAT'S IT????

PR #1048, opened 2016-02-25
Add core team

technoweenie:
I've been wanting a Git LFS core team for a while. I wanted something more "official", but this at least gets us started.

/cc @andyneff @rubyist @sinbad


technoweenie:
As my first act as part of the LFS core team, I will merge a PR with no outside review :trollface:

sinbad:
\o/

Issue #1014, opened 2016-02-18
Pre-push hook problem when branches are deleted server-side


sinbad:
We just came across an edge case which can cause data loss with LFS in the following scenario:
1. A branch is deleted on the server but the client doesn't know yet
2. The server deletes unreferenced LFS objects to optimise space

Reproduction:
1. Create branch 'branch1': `git checkout -b branch1`
2. Add a new LFS file: `git add test.jpg`
3. Push branch1: `git push origin branch1`. Note that the new LFS file is uploaded
4. Delete branch1 on **server**: `git push origin --delete branch1` (or via web interface)
5. Note that this leaves .git/refs/remotes/origin/branch1 locally but the server now thinks that the commit containing test.jpg is now orphaned and can be GC'd. If GC is implemented on LFS objects too, the LFS objects would be deleted if it's not referenced elsewhere. Let's assume that's happened.
6. Create another branch at the same commit: `git checkout -b branch2`
7. Push branch2: `git push origin branch2`
8. No LFS objects are pushed. If the server GC'd the unreferenced LFS object already, this is data loss

In that last push step, the pre-push hook doesn't push any LFS objects, because .git/refs/remotes/origin/branch1 still exists and therefore `git rev-list --objects <commit> --not --remotes=origin` returns nothing. If `git fetch --prune` had been run, or the user had also deleted the locally cached version of the remote ref (e.g. `git branch -D -r branch1`) before pushing branch2, then it would have been OK. However we cannot guarantee the sequence that these things happen in, and therefore cannot guarantee that the remote ref we have locally is correct.

Git itself resolves this discrepancy during the push but it doesn't call the pre-push hook again with the updated commit range it will actually push to the server. So LFS never knows. Once this has happened, there's potentially a 'gap' in the commit sequence where the LFS client thinks the server must have everything, but if it GC's itself, it won't. 

I think this is a bug in our use of the pre-push hook and that we should run `git ls-remote` to confirm the existence of the remote refs we have cached locally before relying on them to determine what to push. We don't have to check the SHAs, because they may have moved forward anyway. There's another potential edge case where a force push has moved the remote ref to a non-descendant of our cached version, but in that case `git push` will fail and require that our local versions are updated anyway, so the subsequent push will resolve that.

I'm happy to work on this if you agree with the approach @technoweenie

technoweenie:
Great catch! I think calling `git ls-remote` is a good solution. Though, LFS servers should probably respect Git's built-in prune settings.

[...]

PR #1051, opened 2016-02-29
Fix issues with server-side branch deletion out of sync with cached remote refs

sinbad:
Fixes problem described in more detail in #1014 

Verifies that server still has all the remote refs we're using to diff which LFS objects to upload, in case one has been deleted without local knowledge and LFS objects have been GC'd in the mean time, which would lead to data loss. Proved that the test I added fails without the code change.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -714,16 +714,15 @@ func CloneWithoutFilters(args []string) error {
 	return nil
 }
 
-// RemoteBranchList returns a list of branches for a remote, without any
-// remote/ or refs/heads prefixes
+// RemoteRefsList returns a list of branches & tags for a remote
 // If retrieveFromRemote is true, this information is built by actually calling
 // the remote using 'git ls-remote'. Otherwise the cached remote branch
 // information held locally is used instead.
-func RemoteBranchList(remoteName string, retrieveFromRemote bool) ([]string, error) {
+func RemoteRefs(remoteName string, retrieveFromRemote bool) ([]*Ref, error) {
 
-	var ret []string
+	var ret []*Ref
 	if retrieveFromRemote {
-		cmd := execCommand("git", "ls-remote", "--heads", "-q", remoteName)
+		cmd := execCommand("git", "ls-remote", "--heads", "--tags", "-q", remoteName)
 
 		outp, err := cmd.StdoutPipe()
 		if err != nil {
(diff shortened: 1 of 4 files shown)


git/git.go
```diff
@@ -714,6 +714,63 @@ func CloneWithoutFilters(args []string) error {
 	return nil
 }
 
+// RemoteRefs returns a list of branches & tags for a remote
+// If retrieveFromRemote is true, this information is built by actually calling
+// the remote using 'git ls-remote'. Otherwise the cached remote branch
+// information held locally is used instead.
+func RemoteRefs(remoteName string, retrieveFromRemote bool) ([]*Ref, error) {
+
```
technoweenie: I think I'd rather have two separate functions, `RemoteRefs()` and `CachedRemoteRefs()`. There's not much shared between the two commands, and switching between the two values based on a boolean arg feels weird.

I'd also prefer the scanner loop use `continue`, instead of continuing to nest `if` statements. It results in a few more lines, but less indentation, which is preferable to me personally.

```golang
func RemoteRefs() {
	// init command, build scanner, etc
	for scanner.Scan() {
		match := r.FindStringSubmatch(scanner.Text())
		if match == nil {
			continue
		}
		
		name := strings.TrimSpace(match[3])
		// Don't match head
		if name == "HEAD" {
			continue
		}
		
		sha := match[1]
		if match[2] == "heads" {
			ret = append(ret, &Ref{name, RefTypeRemoteBranch, sha})
		} else {
			ret = append(ret, &Ref{name, RefTypeRemoteTag, sha})
		}
	}
}
```

sinbad: > I think I'd rather have two separate functions, RemoteRefs() and CachedRemoteRefs()

Yeah, fair enough.

> I'd also prefer the scanner loop use continue, instead of continuing to nest if statements

Hrm, double-edged sword IMO, continue is a pseudo-GOTO in many ways and can have it's own problems if over-used. But it's a short loop body so ok.

sinbad: Compromise, I'm keeping the `if match := r.FindStringSubmatch(scanner.Text()); match != nil {` because that's such a standard golang idiom. The HEAD test is now a continue.

sinbad:
Although this is technically a bugfix, because it changes behaviour a little and because the bug isn't likely to be being encountered just yet (will only start happening when servers implement GC of some kind, and will still be quite rare), I'm not planning on backporting this to 1.1.x and will just merge to master if you're ok with it.

technoweenie:
> I'm not planning on backporting this to 1.1.x and will just merge to master if you're ok with it.

:+1:

Issue #308, opened 2015-05-15
git-lfs will not use sslCAInfo or urlmatch


andyneff:
Slightly related to #276, but git-lfs does not support sslCAInfo, so if I have a ca to point to, no dice.

Further more it does not support "urlmatch" for additional http in the config either, so if in my global ~/.gitconfig I...

```
[http]
    sslVerify = False
```
This works of course, 

I can not
```
[http "https://only.one.site.com/"]
    sslVerify = False
```
I think this it mainly due to b94361f1f001011e5312d0de2db9ae3376f6a9a3 git-lfs/lfs/http.go Line: 41, but That's my 5 second guess

Needless to say, 
```
[http "https://only.one.site.com/"]
    sslCAInfo = /home/me/.ssl/trusted.crt
```
Won't work either.

technoweenie:
Thanks for the report.

rseymour:
I want to fire up another issue w/r/t git-lfs not looking at windows' certificate store either, but it's windows and... I'm at my limit of caring.

technoweenie:
Just added this to the v1.2 checklist in #844.

nkoterba:
Jumping on the bandwagon here, but is it possible for git-lfs to use the Mac OSX keychain?  We have all our self-signed certs for our various internal machines stored in Keychain and it would be nice for git-lfs (and its internal use of SSL) to "apply" any that match the URL/address of the GIT LFS backend server.

I know there's something similar for using Mac OSX keychain to store GIT HTTP basic Auth credentials:  https://help.github.com/articles/caching-your-github-password-in-git/

It would be great if it would also use certs from there.

bmabey:
I just ran into this problem with our GHE install as well. It turns out `hub` is facing the same issue and some people have done some investigation around using the OSx keychain: https://github.com/github/hub/issues/852#issuecomment-185303915

sinbad:
I'm working on this now, I plan to try to support SSLCAInfo, Mac OS X keychain and Windows certificate store.

sinbad:
Quick update as I've been doing some experiments & testing, turns out that git-lfs will already work with self-signed certificates that are trusted in the Windows Certificate Store, because the core golang implementation for root certs does that. Worked for me when I imported them on Windows 10 into the Trusted Root Certificates area in certmgr.msc anyway. Git, however, doesn't look there (as of 2.7.1) so you still need to set http.sslcainfo to make that part work. 

[...]

PR #1067, opened 2016-03-10
Support self signed certs

sinbad:
Adds support for trusting self-signed CA certs from the following locations:

- `http.sslcainfo` in gitconfig (non-specific CA cert file)
- `http.sslcapath` in gitconfig (non-specific dir of CA certs)
- `http.<url>.sslcainfo`, `http.<url>.sslcapath` in gitconfig; as above but specific to server:port. <url> must be of the form `https://host[:port]/` as currently subpaths are not supported (but it's impractical to have different certs per subpath anyway so support would not really beuseful)
- `GIT_SSL_CAINFO` and `GIT_SSL_CAPATH` environment variables; as non-specific gitconfig options
- On Mac OS X, certs marked 'Always Trust' in the System keychain in Keychain Access (just like Chrome and others)

On Windows, the Windows Certificate Store already works for git-lfs access in fact because Golang itself includes it in its Root CA search; however since git (2.7.1) doesn't support the WCS it's not a complete solution, unless you're using SSH clone URLs for git. You're better to stick to the gitconfig / environment options for consistency.

Related: #308


git/git.go
```diff
@@ -59,14 +59,14 @@ func LsRemote(remote, remoteRef string) (string, error) {
 		return "", errors.New("remote required")
 	}
 	if remoteRef == "" {
-		return simpleExec("git", "ls-remote", remote)
+		return subprocess.SimpleExec("git", "ls-remote", remote)
```
sinbad: I refactored SimpleExec and ExecCommand into their own subprocess package since I needed them elsewhere

technoweenie: I may have preferred a `github.com/github/git-lfs/lfsutil` sub package or something, but this works too :+1:

lfs/certs.go
```diff
@@ -0,0 +1,135 @@
...
+
+func appendRootCAsForHostFromGitconfig(pool *x509.CertPool, host string) *x509.CertPool {
+	// Accumulate certs from all these locations:
+
+	loaded := false
+	// GIT_SSL_CAINFO first
+	if cafile, ok := Config.envVars["GIT_SSL_CAINFO"]; ok {
+		pool = appendCertsFromFile(pool, cafile)
+		loaded = true
+	}
```
technoweenie: Is there reason this isn't just returning that `pool`? Seems like we could remove all these `loaded` checks.

git/git.go
sinbad: My first pass was like that but then I read https://blog.golang.org/package-names and realised it was considered a 'bad package name' so I changed it :smile:

lfs/certs.go
sinbad: I swear there was a reason, but I seem to have refactored it out at some point...

git/git.go
technoweenie: :+1:

lfs/http.go
```diff
@@ -118,8 +119,11 @@ func (c *Configuration) HttpClient() *HttpClient {
 	}
 
 	sslVerify, _ := c.GitConfig("http.sslverify")
+	tr.TLSClientConfig = &tls.Config{}
 	if sslVerify == "false" || Config.GetenvBool("GIT_SSL_NO_VERIFY", false) {
-		tr.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+		tr.TLSClientConfig.InsecureSkipVerify = true
+	} else {
+		tr.TLSClientConfig.RootCAs = getRootCAsForHost(host)
```
technoweenie: Should we cache the root CAs per host here? Or maybe the entire `*tls.Config` object.

sinbad: It might be best to cache the whole HttpClient?

sinbad:
I'm not sure what's going on with the Travis tests, they run fine here on Mac. Complaining of a cert issue it seems but I'm using the internal Go TLS server & its certs (1.6 here) and it works for me :/

lfs/http.go
technoweenie: Doh, much better idea. It already is cached, meaning this PR breaks it since we'll want different http client objects per host. That `c.httpClient` needs to become a `map[string]*HttpClient`.

sinbad: I thought of that as I was eating dinner :) Done now.

sinbad:
By chucking lots of debug data out I've proved that Travis is definitely using the same cert content that I am seeing locally, byte for byte, and yet the git clone against the SSL endpoint fails (before it even gets to test these changes). The error even reports that it's using the correct CAFile. I can see no reason why this doesn't work unless there's some additional TLS rules applied on Travis that I don't know how to look for.
@technoweenie IIRC you said you had a contact at Travis? If ever I needed an insider tip it's now :/

sinbad:
Discovered that certain builds of git seem to dislike the directly exported ASN.1/DER cert so exported as PEM instead which has got me past that error. Now getting a new one about a subject name mismatch that again only Travis' git seems to care about, maybe there's a setting I can turn that off in.

sinbad:
Welp, this is total nonsense after all. All I can surmise is that Travis' git is broken when it comes to SSL. I tried doing a full debug dump of connecting to the test server with openssl, this is the result (TL;DR - works totally fine and accepts the cert):
[travis_openssl_output.txt](https://github.com/github/git-lfs/files/176434/travis_openssl_output.txt)

And yet when Travis' git tries to clone from there, it rejects the cert:
[travis_test_output.txt](https://github.com/github/git-lfs/files/176435/travis_test_output.txt)

I'm done with this. If I can't get an explanation from Travis I'm just going to remove this integration test or get it to run only outside Travis. We've proved it works on multiple Macs and I've tested on Win10 too, it's fine outside Travis' environment.

technoweenie:
I think skipping the test in travis is fine. Great work diving into this :+1:

PR #1071, opened 2016-03-11
extract common pre-push and push uploading code

technoweenie:
I have some issues with this code. But at least now pre-push and push both call shared functions to upload objects.

From https://github.com/github/git-lfs/pull/1040#discussion_r55821731:

> Now that filteredPointers is used in multiple places, maybe promote it to the lfs package for re-use rather than keep it in the pre-push source file?

This PR is a start towards shared functions for pre-push and push commands. While it seems to work, there are a number of things I'd like to change:

* I don't like using `lfs.Configuration` like this. I wanted something more than a `StringSet` to share between multiple ref uploads in the same command.
* Filtering could probably be cleaned up.

EDIT: Fixed this in a later commit ^^^

Instead of promoting to the lfs package, what if we started on a high level client like we discussed in https://github.com/github/git-lfs/issues/779? I am a little concerned that my yak shave trying to get #969 merged is going on too long though :)

I don't think the commands should be assembling upload queue objects or anything. I think they should just be responsible for turning the CLI or STDIN arguments into code like this:

```go
c := client.New("origin")
c.DryRun = false

for _, ref := range args[1:] {
  c.Upload(scanObjectsLeftToRight(ref, "", lfs.ScanRefsMode))
}

errs := c.Errors()
// error handling, os.Exit(2), etc
```

### DIFF
--- a/commands/command_pre_push.go
+++ b/commands/command_pre_push.go
@@ -71,35 +71,41 @@ func prePushCommand(cmd *cobra.Command, args []string) {
 			continue
 		}
 
-		left, right := decodeRefs(line)
+		left, right, destRef := decodeRefs(line)
 		if left == prePushDeleteBranch {
 			continue
 		}
 
 		pointers, err := lfs.ScanRefs(left, right, scanOpt)
+
 		if err != nil {
 			Panic(err, "Error scanning for Git LFS files")
 		}
 
-		cli.Upload(pointers)
+		metadata := lfs.NewTransferMetadata(destRef)
+		cli.Upload(metadata, pointers)
 	}
 }
 
-// decodeRefs pulls the sha1s out of the line read from the pre-push
+// decodeRefs pulls the sha1s and destination ref out of the line read from the pre-push
 // hook's stdin.
-func decodeRefs(input string) (string, string) {
+func decodeRefs(input string) (string, string, string) {
 	refs := strings.Split(strings.TrimSpace(input), " ")
-	var left, right string
+	var left, right, destRef string
 
 	if len(refs) > 1 {
 		left = refs[1]
+	}
+
+	if len(refs) > 2 {
+		destRef = refs[2]
 	}
 
 	if len(refs) > 3 {
 		right = "^" + refs[3]
 	}
 
-	return left, right
+	return left, right, destRef
 }
 
 func init() {
--- a/commands/command_pre_push.go
+++ b/commands/command_pre_push.go
@@ -55,8 +55,6 @@ func prePushCommand(cmd *cobra.Command, args []string) {
 	}
 	lfs.Config.CurrentRemote = args[0]
 
-	uploadedOids := lfs.NewStringSet()
-
 	// We can be passed multiple lines of refs
 	scanner := bufio.NewScanner(os.Stdin)
 	for scanner.Scan() {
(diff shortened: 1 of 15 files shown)


technoweenie:
Feeling a lot better about this `*clientContext` struct. It's no longer coupled to the git scanner. Here's a simple example of usage:

```go
cli := newClient()
cli.RemoteName = lfs.Config.CurrentRemote
cli.DryRun = prePushDryRun

scanOpt := lfs.NewScanRefsOptions()
scanOpt.ScanMode = lfs.ScanLeftToRemoteMode
scanOpt.RemoteName = cli.RemoteName

pointers, err := lfs.ScanRefs(left, right, scanOpt)
if err != nil {
  Panic(err, "Error scanning for Git LFS files")
}

cli.Upload(pointers)
```

technoweenie:
Hopefully this wasn't a bad idea (if it is, I'll just merge b5356900f5354f681957d8b5e76fa64978da76b3 into #1040), but I merged #969 in here, fixing merge conflicts. It's obvious where we're not filling out the transfer meta data for `lfs push`. I guess I'll have to figure out how to get remote tracking branches given a list of local branches.

technoweenie:
Here's a compare view if you just want to review the push/pre-push changes to #1040.

https://github.com/github/git-lfs/compare/skip-pre-push-dupes...b5356900f5354f681957d8b5e76fa64978da76b3

git/git.go
```diff
@@ -40,6 +40,22 @@ type Ref struct {
...
+		return r.Name
+	}
+
+	if RemoteForBranch(r.Name) != remotename {
+		return r.Name
+	}
+
+	return RemoteBranchForLocalBranch(r.Name)
+}
+
```
technoweenie: This returns the remote tracking branch name for a local ref, returning the ref name as a fallback. Now, we may want to support the `local:remote` syntax, as seen in commands like `git push heroku heroku:master`. I think that'd be parsed in the command itself and passed to the initialization of the `*Ref`, which would be preferred over `r.Name`.

Thoughts?

git/git.go
```diff
@@ -80,6 +96,19 @@ func ResolveRef(ref string) (*Ref, error) {
...
+	for i, name := range refnames {
+		ref, err := ResolveRef(name)
+		if err != nil {
+			return refs, err
+		}
+
+		refs[i] = ref
+	}
+	return refs, nil
+}
```
technoweenie: This sucks, because `ResolveRef()` runs `simpleExec()`. You can pass multiple refs though:

```bash
$ git rev-parse master push-extraction push-optimizations --symbolic-full-name master push-extraction push-optimizations
2cf082fa14a738f73db6d172eba800ca118e6d97
ce738288364c00185a51c6fd2c9ce42d4ddec155
8aae3553fa708503571fa8fea90ac54a82b7b31b
refs/heads/master
refs/heads/push-extraction
refs/heads/push-optimizations
```

git/git.go
sinbad: I think this is fine for auto selection which is what the function is designed to do, but yeah if we want to support full syntax this may be overridden; the problem is I'm not sure we get the remote-side name in the pre-push hook?

I'm somewhat skeptical of the actual usefulness of this ref in the api anyway except as a convenience hint for early rejection. It's so easy to game it's of no real use security wise.

git/git.go
sinbad: Yeah this should be a single exec otherwise there's little point of it existing, calling code could loop on ResolveRef. Also (moot if changed) I don't think it should early-return on failure with the separate execs but should tolerate intermediate errors in case caller just wants to deal with the ones that work & skip failures. That would imply a map return value.

commands/command_pre_push.go
```diff
@@ -55,7 +54,13 @@ func prePushCommand(cmd *cobra.Command, args []string) {
 	}
 	lfs.Config.CurrentRemote = args[0]
 
-	uploadedOids := lfs.NewStringSet()
+	cli := newClient()
+	cli.RemoteName = lfs.Config.CurrentRemote
```
sinbad: Hmm, feels weird to have to copy the remote name when everything else uses Config directly, feels like this might be fragile - either forgetting to assign it, or other code which changes lfs.Config.CurrentRemote thinking it applies everywhere (some tests do this already I think). I think we should keep a single source of truth.

test/test-push.sh
```diff
@@ -333,29 +333,30 @@ begin_test "push object id(s)"
...
+  git show
 
   git lfs push --object-id origin \
-    4c48d2a6991c9895bcddcf027e1e4907280bcf21975492b1afbade396d6a3340 \
-    82be50ad35070a4ef3467a0a650c52d5b637035e7ad02c36652e59d01ba282b7 \
+    c1a338929270651d8fb132411d725518353650bfc9ac991991dd4a8c2139d97a \
+    7098284b1b16d67681313247191c217c5f185acaa40b5f00731574d70d75f2e1 \
     2>&1 | tee push.log
-  grep "(0 of 2 files, 2 skipped)" push.log
+  grep "(1 of 1 files)" push.log
```
sinbad: This now feels like non-intuitive output to get when asked to push 2 oids. It looks like it's ignored the second parameter rather than validly skipping one because it already exists on the remote. What changed this? It's not the skipping dupes code because this is a self-contained command so c1a33 hasn't been previously uploaded in this run.

sinbad:
Are you intending to wrap downloading in clientContext too? Although there's somewhat less for it to do than in the upload case. If not, or even if 'not right away' then maybe rename clientContext to uploadContext for now.

git/git.go
technoweenie: It already was a loop: https://github.com/github/git-lfs/blob/2cf082fa14a738f73db6d172eba800ca118e6d97/commands/command_fetch.go#L46-L51

I extracted this function, and left this note to work on after the weekend. Guess I could've said as much on my msg from Friday :)

sinbad: Haha, tale as old as time: 1. criticise code 2. realise you wrote it 3. Facepalm xD

Still, if it's going to be its own function it might as well improve things too. Clutching at the shreds of my dignity... ;D

technoweenie:
#1040 now conflicts after your scanner fixes. Totally fine. I'm going to chalk this up as experimentation, merge this PR into #1040, and then redo the changes on master.

PR #1078, opened 2016-03-15
Fix an alignment typo in Git LFS batch API v1 documentation

epriestley:
This JSON key is only indented one space, but should be indented two spaces to align with the rest of the structure.


technoweenie:
Thanks!

Issue #1056, opened 2016-03-02
Pointer file displayed on GitHub instead of actual file


cmaumet:
We (with @AlexBowring) are using git-lfs in our public repository: https://github.com/incf-nidash/nidmresults-examples and for one of our commits (https://github.com/incf-nidash/nidmresults-examples/commit/f3dc7eaafd7f7d69ea673479e8311515e077c103), the files stored by git-lfs are displayed as pointers (instead of the actual file). 

An example of file is https://github.com/incf-nidash/nidmresults-examples/blob/f3dc7eaafd7f7d69ea673479e8311515e077c103/spm_t_test/RPV.nii showing up as 
```
version https://git-lfs.github.com/spec/v1
oid sha256:e88c0b9c2cb607d5684467c9143e2ba4c96ea7979a2d14175d4dd1122da32bfc
size 4743512
```
while in the full commit listing (https://github.com/incf-nidash/nidmresults-examples/commit/f3dc7eaafd7f7d69ea673479e8311515e077c103) this file is actually listed as "Git LFS file not shown". 

We are not experiencing this issue with other commits (e.g. https://github.com/incf-nidash/nidmresults-examples/commit/e92760f172d62cd80782d848afd2b55b76eb1ba3) and could not track down what the error was. 

Would you have any advice for us? Thank you!

technoweenie:
It's because that object hasn't been pushed to the server. Try running `git lfs push --all` from your local repository to fix.

Also, this is a GitHub.com issue, and has nothing to do with this open source project. In the future, you can just hit me up at https://github.com/contact.

cmaumet:
@technoweenie: I have described our issue at https://github.com/contact but we got no solution so far and it's getting worse... There are now non-pushed git-lfs files in the master branch which makes it impossible to clone the repository. To avoid spreading the issue further, we are on hold awaiting a fix. Could you let me know what is the best thing for me to do now? Thank you!

technoweenie:
Sorry, clearly I was too quick to take this private. I found the bug in a command that Git LFS uses to scan the repository history:

```bash
$ git rev-list --objects --do-walk spm_t_test
fatal: ambiguous argument 'spm_t_test': both revision and filename
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

I gave you a workaround in our support thread, which is basically just rename the branch and try again:

```bash
$ git checkout -b spm_t_test-pr
[...]

PR #1079, opened 2016-03-17
push with ambiguous branch fails

technoweenie:
The scanner code doesn't handle git errors at all. I would expect this `git lfs push` command to fail, but it just returns successfully with no output. The problem is that the rev-list call is erroring, and git lfs is just rolling with it.

```
    + git lfs push --dry-run origin ambiguous-branch-name
    + tee push.log
    + echo 'actual git command:'
    + git rev-list --objects ambiguous-branch-name --not --remotes=origin
    fatal: ambiguous argument 'ambiguous-branch-name': both revision and filename
    Use '--' to separate paths from revisions, like this:
    'git <command> [<revision>...] -- [<file>...]'
```

### DIFF
--- a/test/test-push.sh
+++ b/test/test-push.sh
@@ -2,6 +2,29 @@
 
 . "test/testlib.sh"
 
+begin_test "push with filename / branch mismatch"
+(
+  set -e
+  reponame="$(basename "$0" ".sh")-ambiguous-branch-name"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" repo
+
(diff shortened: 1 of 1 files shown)


epriestley:
Oh -- I missed this, but #1096 is likely a fix for this, and presumably #1056 is the root issue.

epriestley:
(Although those just correct the command, rather than detecting/handling errors.)

sinbad:
Closing as addressed in #1118 now.

PR #1080, opened 2016-03-17
Move TTY wrapper to subprocess package

sinbad:
TTY abstraction and subprocess came out of different PRs, makes sense to move TTY into subprocess now they're both in master.


technoweenie:
:+1:

PR #1082, opened 2016-03-17
Explain a subtlety of client handling for "download" batch action

epriestley:
This may not be worth pulling or the best method or place to express the idea, but the behavior wasn't self-evident to me from the documentation while implementing an LFS server, even though it makes sense now that I understand the interaction.

### DIFF
--- a/docs/api/http-v1-batch.md
+++ b/docs/api/http-v1-batch.md
@@ -105,7 +105,10 @@ when the object has not been previously uploaded.
 successfully uploading an object.  This is an optional relation for the case that
 the server has not verified the object.
 * `download` - This relation describes how to download the object content.  This
-only appears if an object has been previously uploaded.
+only appears if an object has been previously uploaded.  If object content is
+already available, the server should provide this action even if the request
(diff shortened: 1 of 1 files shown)


technoweenie:
Looks good to me :+1:

sinbad:
Hmm, I'm not sure this is correct, it implies that the server *must* respond with a download link when the operation is `upload` in order for the upload to be skipped, when in fact simply *omitting* the upload link in the response does that. You don't have to provide a download link instead and doing so makes no difference vs omitting - git-lfs just doesn't find a link of type `upload` in both cases. I don't think we should be telling server implementors that they have to provide a download link in upload requests.

epriestley:
Ah, I must have misunderstood the `prePushCheckForMissingObjects()` / `NewDownloadCheckQueue()` interaction -- and, in retrospect, my test case would have worked for either behavior. I'll fix this.

epriestley:
#1087 reverts this and properly attributes the behavior to the `upload` action instead. Sorry about that!

sinbad:
Thanks for the fast turnaround!

PR #1083, opened 2016-03-17
Fix smudge/clean on files that begin with a dash

### LINKED ISSUE #1072: Incorrect behavior when a filename starts with dash
When a filename starts with dash I get a error message:
```
git lfs clean <path>

Read the contents of a large file from standard input, and write a Git
LFS pointer file for that file to standard output.

Clean is typically run by Git's clean filter, configured by the repository's
Git attributes.

Error: unknown shorthand flag: 'T' in -Total_Eclipse_Of_The_Heart-84093257.mp3
error: cannot feed the input to external filter git-lfs clean %f
error: external filter git-lfs clean %f failed
fatal: -Total_Eclipse_Of_The_Heart-84093257.mp3: clean filter 'lfs' failed

```

Git itself doesn't fail if I disable lfs extension for repository. It seems that lfs calls utilities without -- argument that splits filenames section from the options section of the utility command line (for example, you can call them like this `cat -- -file-starting-with-dash`).

epriestley:
Fixes #1072. Currently, the filters run a command like this when a tracked file begins with a dash:

`$ git-lfs clean -filename`

The filename is then misinterpreted as short arguments.

This installs `git-lfs clean -- %f` instead of `git-lfs clean %f`, and a similar smudge filter.

A limitation of this patch is that existing installs don't automatically upgrade to the new filters until they run `git lfs install --force`. It looks like this is consistent with the `git lfs clean` vs `git-lfs clean` precedent, but maybe encountering this behavior is realistic enough to justify trying to build some sort of auto-upgrade feature which promotes known older variants of the filter (it looks like the pre-push hook has such a list already).

In any case, this should fix new installs, and allow existing installs to use `git lfs install --force` to manually apply the fix.

### DIFF
--- a/lfs/setup.go
+++ b/lfs/setup.go
@@ -21,8 +21,8 @@ var (
 	filters = &Attribute{
 		Section: "filter.lfs",
 		Properties: map[string]string{
-			"clean":    "git-lfs clean %f",
-			"smudge":   "git-lfs smudge %f",
+			"clean":    "git-lfs clean -- %f",
+			"smudge":   "git-lfs smudge -- %f",
 			"required": "true",
 		},
 	}
@@ -30,8 +30,8 @@ var (
 	passFilters = &Attribute{
 		Section: "filter.lfs",
 		Properties: map[string]string{
-			"clean":    "git-lfs clean %f",
-			"smudge":   "git-lfs smudge --skip %f",
+			"clean":    "git-lfs clean -- %f",
+			"smudge":   "git-lfs smudge --skip -- %f",
 			"required": "true",
 		},
 	}
--- a/test/test-env.sh
+++ b/test/test-env.sh
@@ -2,8 +2,8 @@
 
 . "test/testlib.sh"
 
-envInitConfig='git config filter.lfs.smudge = "git-lfs smudge %f"
-git config filter.lfs.clean = "git-lfs clean %f"'
+envInitConfig='git config filter.lfs.smudge = "git-lfs smudge -- %f"
+git config filter.lfs.clean = "git-lfs clean -- %f"'
 
 begin_test "env with no remote"
 (
--- a/test/test-install.sh
+++ b/test/test-install.sh
@@ -41,8 +41,8 @@ begin_test "install with old settings"
   [ "git lfs clean %f" = "$(git config --global filter.lfs.clean)" ]
 
   git lfs install --force
-  [ "git-lfs smudge %f" = "$(git config --global filter.lfs.smudge)" ]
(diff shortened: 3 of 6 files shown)


epriestley:
I'm not sure how to reproduce that test failure locally, it passes in my environment:

```
$ ./test/test-smudge.sh 
compile git-lfs for ./test/test-smudge.sh
commands/mancontent_gen.go
Using go1.6
Git LFS: /Users/epriestley/dev/git-lfs/bin/git-lfs
git-lfs/1.1.1 (GitHub; darwin amd64; go 1.6; git fa2ec2f)
git version 2.5.4 (Apple Git-61)
Updated pre-push hook.
Git LFS initialized.

HOME: /Users/epriestley/dev/git-lfs/test/remote/home
TMP: /Users/epriestley/dev/git-lfs/tmp
CREDS: /Users/epriestley/dev/git-lfs/test/remote/creds
lfstest-gitserver:
  LFSTEST_URL=/Users/epriestley/dev/git-lfs/test/remote/url
  LFSTEST_SSL_URL=/Users/epriestley/dev/git-lfs/test/remote/sslurl
  LFSTEST_CERT=/Users/epriestley/dev/git-lfs/test/remote/cert
  LFSTEST_DIR=/Users/epriestley/dev/git-lfs/test/remote
GIT:
filter.lfs.clean git-lfs clean -- %f
filter.lfs.smudge git-lfs smudge -- %f
filter.lfs.required true
credential.helper lfstest
user.name Git LFS Tests
user.email git-lfs@example.com

test: smudge ...                                                   OK
test: smudge --info ...                                            OK
test: smudge with temp file ...                                    OK
test: smudge with invalid pointer ...                              OK
test: smudge include/exclude ...                                   OK
test: smudge with skip ...                                         OK
test: smudge clone with include/exclude ...                        OK
```

technoweenie:
Hey, thanks so much for stopping by and contributing all these patches :metal: 

This is the first time that the clean/smudge filters have changed. I think it'd be good to have `git lfs update` or `git lfs install` (which you should be running after every update) automatically fix the filters, so users don't have to run it with `--force`. As [you can tell](https://github.com/github/git-lfs/blob/c3c7e90609f2a334fa0bbf80b95ecd8c66d89466/lfs/setup.go#L8-L13), we've already run into this with the content of the `pre-push` hooks.

Also, that test failure is a wonderful intermittent issue, apparently. Don't worry about it in the context of this PR: https://github.com/github/git-lfs/issues/1086

epriestley:
Ah, thanks for the pointer on the test failure. Spooky! :ghost:

PR #1084, opened 2016-03-17
Fix "Uploading refs <remote> to remote <refs>" log string

epriestley:
The argument order and string order are flipped. Before:

```
$ GIT_TRACE=1 git-lfs push origin master
trace git-lfs: Upload refs origin to remote [master]
```

After:

```
$ GIT_TRACE=1 git-lfs push origin master
trace git-lfs: Upload refs [master] to remote origin
```


technoweenie:
Doh, busted. Thanks!

Issue #1053, opened 2016-02-29
Document remote name requirements, show better errors


rubyist:
Currently the smudge filters can only work with a remote named `origin`. Generally, by the time the smudge filters are called, the name of the remote pulled from cannot be determined, or there may never have been one (e.g. from checkout). We should add documentation to the smudge command indicating this requirement.

Right now, for example, when doing a `git pull` using a remote with a name other than `origin`, Git LFS will attempt to use an empty string as its endpoint, which results in a confusing error about unsupported protocols. We can detect this and fail earlier, with a better message.

The easiest workaround for this problem is to use the name `origin` for the remote. Alternatively, the smudge filter can be skipped and lfs operations run manually.

sinbad:
We should probably just be calling & checking `git.DefaultRemote()` as in `git lfs fetch` which will pick up the tracking remote for the current branch. It's not free though, it will add a couple of extra git calls per smudge which is not ideal & reinforce the need for `git lfs clone` and potentially for more efficient wrappers for `git pull/checkout` in future.

apeyser:
This also happens if the default branch isn't origin, even if you're pulling from origin.

apeyser:
#2710 explains why this would be the case. There's a fundamental lack of information for LFS to properly handle the data pull --- the server would need to be stored with the file pointer to pull the object from the right repo, but then there could only be one storage back end for the blobs.

I'd guess git would have to be patched to pass to smudge the current repo that is being handled?

quinngrier:
I just ran into this issue while trying to merge a branch from a remote whose name is not `origin`. In fact, none of my remotes are named `origin`. I resolved it by creating a remote named `origin` that points to the same URL and merging from that.

Here is a log so you can see what I went through as an inexperienced user of Git LFS:

```
$ git fetch --all
...

$ git merge some_remote/master
git-lfs filter-process: 1: git-lfs filter-process: git-lfs: not found
fatal: The remote end hung up unexpectedly

$ sudo apt install git-lfs
...

$ git merge some_remote/master
error: The following untracked working tree files would be overwritten by merge:
        .gitattributes
Please move or remove them before you merge.
Aborting

$ git status -s
 D some_file_1
?? .gitattributes

$ git checkout .
Encountered 1 file(s) that should have been pointers, but weren't:
        some_file_1

$ git status -s
 M some_file_1
?? .gitattributes

$ git reset --hard
HEAD is now at 7654321 some_commit_title
[...]

Issue #930, opened 2016-01-12
Error accessing media: unsupported protocol scheme ""


cmaumet:
When trying to pull from https://github.com/incf-nidash/nidmresults-examples, we get the following error:
```
$ git pull
Updating 17d5080..ff51e6d
Downloading spm_explicit_mask/explicit_mask_2.nii (579.34 KB)
Error accessing media: spm_explicit_mask/explicit_mask_2.nii (a543db67399ab02aefbf3dd2aad2f4e2980c9a026b657498c79f3573ac821952)

Errors logged to .git/lfs/objects/logs/20160112T102310.547408339.log
Use `git lfs logs last` to view the log.
error: external filter git-lfs smudge %f failed 2
error: external filter git-lfs smudge %f failed
fatal: spm_explicit_mask/explicit_mask_2.nii: smudge filter lfs failed
```

Full log:
```
$ git lfs logs last
git-lfs/1.1.0 (GitHub; darwin amd64; go 1.5.1)
git version 2.6.3

$ git-lfs smudge spm_explicit_mask/explicit_mask_2.nii
Error accessing media: spm_explicit_mask/explicit_mask_2.nii (a543db67399ab02aefbf3dd2aad2f4e2980c9a026b657498c79f3573ac821952)

Post objects/batch: unsupported protocol scheme ""
goroutine 1 [running]:
github.com/github/git-lfs/lfs.Stack(0x0, 0x0, 0x0)
	/private/tmp/git-lfs20151202-15548-y7a9pj/git-lfs-1.1.0/src/github.com/github/git-lfs/lfs/errors.go:557 +0x80
github.com/github/git-lfs/commands.logPanicToWriter(0xb641c0, 0xc820082228, 0xb64de0, 0xc82012a480)
	/private/tmp/git-lfs20151202-15548-y7a9pj/git-lfs-1.1.0/src/github.com/github/git-lfs/commands/commands.go:184 +0xf7f
github.com/github/git-lfs/commands.logPanic(0xb64de0, 0xc82012a480, 0x0, 0x0)
	/private/tmp/git-lfs20151202-15548-y7a9pj/git-lfs-1.1.0/src/github.com/github/git-lfs/commands/commands.go:148 +0x421
github.com/github/git-lfs/commands.handlePanic(0xb64de0, 0xc82012a480, 0x0, 0x0)
[...]

Issue #864, opened 2015-11-25
Git LFS is unaware of url configurations


dentuzhik:
So basically if you will do in your `.gitconfig`:
```
[url "git@github.com:"]
	insteadOf = "gh:"
	pushInsteadOf = "github:"
	pushInsteadOf = "git://github.com/"
	pushInsteadOf = "https://github.com/"
```

And then clone the repo, which uses git lfs, and then do `git lfs pull`, it wil fail with:
```
Post gh:username/reponame.git/info/lfs: unsupported protocol scheme "gh"
```

technoweenie:
Thanks for the report.

andre487:
I have the same issue

QuLogic:
Is this not fixed by #1443?

technoweenie:
Ah, it doesn't implement `pushinsteadof`. But `insteadof` should work as of v1.4.0. If not, please let us know what specific case didn't work.

eddsteel:
I had this problem with two `insteadOf` lines in the same stanza. Changing one to `pushInsteadOf` fixed it.

But `man git-config` says `insteadOf` "allows people to specify any of the equivalent URLs", plural. Doesn't break anything for me but seems incorrect.

Issue #798, opened 2015-10-28
ssh remote URL must include "git@" prefix


boblaublaw:
I'm noticing that when my remote is set to either:

"github.com:/Team-Name/Repo-Name"  OR
"ssh://github.com:/Team-Name/Repo-Name"

i get unsupported scheme errors despite being able to confirm ssh authentication using command line ssh to verify.  This is because my .ssh/config file specifies my username as "git".

I believe this is because git is correctly detecting the ssh repo specification, but not using the username "git" when talking to the LFS HTTP API.  

This may well be intended behavior, but if so, I think it is important to call out the requirement to use "git@..."

technoweenie:
It's not _intended_, but the shortcuts and alternatives haven't been implemented yet. There's a really old issue tracking this, but haven't found anyone interested in working on them yet: #278 #295 

I'll add them to the list for v1.1 if we have time for it: #755

ttaylorr:
Duplicate of: https://github.com/git-lfs/git-lfs/issues/278.

PR #1085, opened 2016-03-17
Accept raw remote URLs as valid

epriestley:
Fixes issue #1069, sort of, in theory. This is likely more of a starting point. Major issues:

  - No meaningful test coverage.
  - Pushing to a raw URL scans and tries to push every object in the repository.

Overview:

This treats any "ssh", "http", "https", "git" or protocol-free-with-a-colon-in-it remote name as a valid raw remote URL.

I'm not sure how to effectively test this, because it seems difficult to stub/simulate any of these protocols locally, and `git-lfs` won't work with `file://`, etc. Existing tests pass locally, and the existing test to verify that `git-lfs push not-a-remote` encounters an error passes (this is correctly detected as not being a valid remote).

I manually tested positive cases with, e.g.:

```
$ GIT_TRACE=1 git-lfs push git@github.com:epriestley/poems.git master
...
Git LFS: (1 of 7 files, 6 skipped) 11 B / 87.16 KB, 87.15 KB skipped    
```

One possible issue is that this causes this `rev-list` command to execute:

```
trace git-lfs: run_command: git rev-list --objects master --not --remotes=git@github.com:epriestley/poems.git
```

`git` runs this, but it fails to exclude anything and lists every object in the repository.

A better strategy might be:

  - If the URL is raw, run `git ls-remote <url>` instead.
  - Pass the list of remote refs to `--not` instead of `--remote <remote>`.

I can pursue this refinement, but didn't want to get too far into things if I'm on the wrong track.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"path/filepath"
 	"regexp"
@@ -172,7 +173,38 @@ func ValidateRemote(remote string) error {
 			return nil
 		}
 	}
+
+	err = ValidateRemoteURL(remote)
+	if err == nil {
+		return nil
+	}
+
 	return errors.New("Invalid remote name")
+}
+
+// ValidateRemoteURL checks that a string is a valid Git remote URL
+func ValidateRemoteURL(remote string) error {
+	u, err := url.Parse(remote)
+	if err != nil {
+		return err
+	}
+
+	switch u.Scheme {
+	case "ssh", "http", "https", "git":
+		return nil
+	case "":
+		// This is either an invalid remote name (maybe the user made a typo
+		// when selecting a named remote) or a bare SSH URL like
+		// "x@y.com:path/to/resource.git". Guess that this is a URL in the latter
+		// form if the string contains a colon ":", and an invalid remote if it
+		// does not.
+		if strings.Contains(remote, ":") {
+			return nil
+		}
+		return errors.New("Invalid remote name")
+	default:
+		return errors.New("Invalid Remote URL protocol")
+	}
 }
 
 // DefaultRemote returns the default remote based on:
--- a/lfs/config.go
+++ b/lfs/config.go
@@ -126,6 +126,11 @@ func (c *Configuration) GitRemoteUrl(remote string, forpush bool) string {
 		return u
 	}
 
+	err := git.ValidateRemoteURL(remote)
+	if err == nil {
+		return remote
+	}
+
 	return ""
 
 }


epriestley:
From #306, #295, etc., maybe I didn't dig deep enough into the test suite and `http://` is already stubbed, which would be suitable here.

This issue is also probably adjacent to #295, at least, and maybe a few other things -- I can fish through issues and see what else I can dig up, this just seemed like the kind of thing that might be stepping into "major infrastructure change" territory so I didn't want to dig too deep if this going to snowball into more serious spelunking than I'm currently equipped to handle.

technoweenie:
Awesome. Definitely want some test coverage. The shell tests do setup a git server with an https endpoint that can be used. This should test both a remote that is named in `git remote`, and a remote that that isn't.

> One possible issue is that this causes this rev-list command to execute:
>     $ git rev-list --objects master --not --remotes=git@github.com:epriestley/poems.git

I don't want to use `git ls-remote` for this, since that will hit the server. The `--remotes` option looks in the local ref from the remote in `refs/remotes/*`. LFS should only be scanning the range of commits from that ref to the current HEAD that it's going to be pushing. Using `git ls-remote` will hit the server, which could show a remote HEAD that is further ahead of the local repository.

What do you think about attempting to look up the remote name for a URL? If the URL has a remote name, continue passing `--remotes`. Otherwise, assume the repo has no local cache of that remote, and scan the entire history.

It may be worth creating a function that returns a `git.RemoteList` struct from:

```
$ git remote -v
dev	http://github.dev/technoweenie/whatever.git (fetch)
dev	http://github.dev/technoweenie/whatever.git (push)
origin	https://github.com/github/git-lfs (fetch)
origin	https://github.com/github/git-lfs (push)
```

Could even cache it in `lfs.Configuration`, to prevent potentially calling `git remote` multiple times in the same Git LFS process.

epriestley:
Cool, I'll see about getting proper coverage.

> Using git ls-remote will hit the server, which could show a remote HEAD that is further ahead of the local repository.

It can't ever find any descendants of the current HEAD, can it (or if it somehow does, we don't need to push the file data)? The remote could be ahead chronologically, but it can't be _ahead_ ahead unless something terrible happened and all bets are already off anyway (like someone pushed with `lfs` but skipped transmitting data), I think? But maybe I'm just misunderstanding or there's some scenario I'm not considering.

We (a code review tool called Phabricator) have an unusual situation where we sometimes push unreviewed code into a "staging area" remote which is mostly hidden from the user, identified only by URL, and never fetched, so the local repository will never know about most of the refs in it (and will know about none of the refs on the first push). This is unusual and you may not want to support it, but in our case there is never a named remote and the local repository will rarely have much information about refs present in the remote.

I'm not sure how common this is, although I'd imagine that if a push is using a raw URL, it's probably because there isn't a named local remote at least some of the time? But maybe there are situations where other tools pre-resolve the remote and drop the symbolic name for it, but it's still fundamentally `origin` or whatever.

epriestley:
For small repositories it's also believable that `rev-list`-ing everything and POSTing a little extra data is faster than eating an extra round trip with `ls-remote`, but that presumably breaks down by the time a repository reaches some realistic number of LFS objects.

epriestley:
> Using git ls-remote will hit the server, which could show a remote HEAD that is further ahead of the local repository.

Oh, sorry, I misunderstood this at first until I read the message in f3bf4ea9. The issue is not that a ref ahead of us is wrong, just that it's useless, since `rev-list ... --not <ref we don't know the history of>` is meaningless and `git` can't do anything with it.

I'm going to do a little more digging but don't have a better approach to suggest than the one you outline.

epriestley:
After some digging:

#144 has some discussion on the "what do we list to push" issue. I think that issue is actually resolved in `master` (since f3bf4ea)? @mhagger suggests the `ls-remote` approach there, with `--ignore-missing`.

#798, #278, and #295 are possibly adjacent. It looks like there may be some peril with host aliases if we try to guess which named remote corresponds to a URL, but that's fine. #693 is a vaguely related peril with `insteadOf` config. #864 is a copy of this, more or less. #754 looks like another one of the SSH host issues.

#431 and #436 have some cases with multiple remotes which LFS can't reasonably have perfect behavior on, which likely represent limits to how much we can possibly get right here. I think #580 is another copy of these. #930 is another one, although some users also reported the SSH aliases bug there. #1053 is approximately related.

---

So I think this is a maybe-reasonable way to break this down:

  - I'll cover this with tests and make `rev-list` fall back to a full list if it's interacting with a URL so we don't pass a nonsense `--remote <url>` argument. This should be strictly better than the behavior today ("don't work"), but not ideal.
  - Next, I'll try to guess which named remote a URL corresponds to so we can exclude refs with `--not --remote <remote>`. I think this will run into the `insteadOf` and "SSH Host Alias" issues, and I may be able to fix or mitigate them.
  - I'll keep an eye out on the `origin` stuff. I think it's better in modern code in `master` than some of those reports suggest and some of it fundamentally isn't tractable, but there may be opportunities to improve it.
  - When we still miss on finding local refs for the remote, I think the `ls-remote` + `--ignore-missing --not <every ref>` behavior from #144 is still better than nothing since it has a relatively small, relatively constant cost but potentially a big upside if even one ref isn't ahead of the local (and, selfishly, it will improve __my__ bizarre case), but I can make a case for it later if I make it this far.

git/git.go
```diff
@@ -172,9 +173,40 @@ func ValidateRemote(remote string) error {
...
 
+// ValidateRemoteURL checks that a string is a valid Git remote URL
+func ValidateRemoteURL(remote string) error {
+	u, err := url.Parse(remote)
+	if err != nil {
+		return err
+	}
+
+	switch u.Scheme {
+	case "ssh", "http", "https", "git":
```
grimreaper: Will this honor pushInsteadOf ?

technoweenie: Nope, but I think that'd be worthy of a separate pull request.

technoweenie:
Mockingbird is cool :+1:

technoweenie:
Dang, what did I do??? The GitHub Platform team uses Marvel Comic characters in their media types. Since I'm a big comic nerd, they ping me on all new features to "approve" them. I approved one in this PR for some reason.

technoweenie:
I just commented [here](https://github.com/github/git-lfs/issues/844#issuecomment-205541466) that I've moved this to the "nice to have" list for v1.2. However, I wouldn't mind shipping this mostly as-is, with more testing: 

> I'll cover this with tests and make rev-list fall back to a full list if it's interacting with a URL so we don't pass a nonsense --remote <url> argument. This should be strictly better than the behavior today ("don't work"), but not ideal.

I personally agree with this notion:

> I'm not sure how common this (hidden remotes) is, although I'd imagine that if a push is using a raw URL, it's probably because there isn't a named local remote at least some of the time? 

Having a command that works, but attempts to upload too many objects seems preferable to it not working at all.

lfs/config.go
```diff
@@ -126,6 +126,11 @@ func (c *Configuration) GitRemoteUrl(remote string, forpush bool) string {
 		return u
 	}
 
+	err := git.ValidateRemoteURL(remote)
```
naufraghi: Could be usefull to convert `err := git.ValidateRemoteURL` in `remote, err := git.NormalizeRemoteURL`. If the user clones an url with a trailing `/`, es `git clone https://github/repo.git/` the `NewEndpointFromCloneURLWithConfig` [1] adds an unwanted `repo.git/.git/info/lfs` layer to the Endpoint that should be `repo.git/info/lfs`

[1] [lfs/endpoint.go#L41](https://github.com/github/git-lfs/blob/8f90f594162d0a7e90e2f5785287f4327f097235/lfs/endpoint.go#L41)

omonnier:
I'm currently working on integrating git-lfs into Gerrit repo tool, and I'm facing the same issue, because repo has an automatic fallback mechanism for ssh to https, in case the git push over ssh is failing. For that reason, repo is not using the remote name in the git push command, but the full url instead. I'm not sure to fully understand the plan for fixing that issue, but just wanted to let you know that it's preventing the integration of git-lfs into repo. I'm available for helping in writing some unit tests if needed.

technoweenie:
I re-applied @epriestley's commit against master in #1451. I also added some basic tests for `git lfs [fetch | pull | push]` and applied @naufraghi's suggestion about handling remotes with trailing slashes.

swisspol:
@technoweenie Why did you close this PR? The issue it fixes is still not fixed.

technoweenie:
I closed it because the fix was applied in #1451. I confirmed that the issue is fixed just now too. Please open a new issue with details if you're still having problems.

PR #1087, opened 2016-03-18
Properly correct batch upload/download skip detection documentation

epriestley:
I got this wrong in #1082, see:

<https://github.com/github/git-lfs/pull/1082#issuecomment-198308275>

I misread the `prePushCheckForMissingObjects()` logic and, in my local test case, I had the server send back `download` _instead of_ `upload` and saw it skip, but that isn't an accurate test because either rule (presence of `download`, absence of `upload`) passes it.

To verify this behavior properly, I had the server send back a made up action (`quack`) instead of `upload`, and observed that the client skipped the transfer:

```
$ GIT_TRACE=1 git lfs push origin master
...
trace git-lfs: HTTP: POST http://local.phacility.com/diffusion/POEMS/poems.git/info/lfs/objects/batch
...
trace git-lfs: HTTP: {... "actions":{"quack":{ ... } }}
...
Git LFS: (0 of 1 files, 1 skipped) 0 B / 87.12 KB, 87.12 KB skipped                                                                                                                                                            
```

### DIFF
--- a/docs/api/http-v1-batch.md
+++ b/docs/api/http-v1-batch.md
@@ -99,16 +99,17 @@ schemas:
 
 Here are the valid actions:
 
-* `upload` - This relation describes how to upload the object.  Expect this with
-when the object has not been previously uploaded.
+* `upload` - This relation describes how to upload the object.  If the object
+has not previously been uploaded the server should provide this action.  If
+the object has been previously uploaded and the object content is known to the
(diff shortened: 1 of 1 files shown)


sinbad:
Thanks!

technoweenie:
Doh, thanks for catching that, @sinbad :)

PR #1091, opened 2016-03-18
Fully resolve symlinks in all cases for consistency

sinbad:
I discovered this problem while setting up the integration tests to always run in an isolated temp dir if GIT_LFS_TEST_DIR is not set. On Mac that's `/var/temp`, which is a symlink to `/private/var/temp` and which immediately failed in all kinds of ways. The 2 main issues were: 

1. `git rev-parse` resolves symlinks but os.Getwd(), filepath.Abs() does not, causing internal inconsistencies and some weird relative paths from filepath.Rel() that didn't always work
1. Many integration tests failed because bash wasn't resolving symlinks vs Go doing so sometimes

I've chosen to resolve symlinks always to fix this. In bash this was more tricky because `readlink -f` doesn't exist on Mac so there's a helper function to do it.

### DIFF
--- a/test/testenv.sh
+++ b/test/testenv.sh
@@ -3,9 +3,48 @@
 
 set -e
 
+UNAME=$(uname -s)
+IS_WINDOWS=0
+IS_MAC=0
+if [[ $UNAME == MINGW* || $UNAME == CYGWIN* ]]
+then
+  IS_WINDOWS=1
+elif [[ $UNAME == *Darwin* ]]
+then
+  IS_MAC=1
+fi
+
+resolve_symlink() {
+  local arg=$1
+  if [ $IS_WINDOWS == "1" ]; then
+    printf '%s' "$arg"
+  elif [ $IS_MAC == "1" ]; then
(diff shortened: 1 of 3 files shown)


test/testenv.sh
```diff
@@ -3,9 +3,48 @@
...
+IS_MAC=0
+if [[ $UNAME == MINGW* || $UNAME == CYGWIN* ]]
+then
+  IS_WINDOWS=1
+elif [[ $UNAME == *Darwin* ]]
+then
+  IS_MAC=1
+fi
+
+resolve_symlink() {
```
sinbad: It's not ideal to have this function in `testenv.sh` but it's simpler to have it available as early as this so that all derived directory vars can start with a resolved version, rather than trying to do it afterwards.

sinbad:
Hrm, small problem specific to go 1.5 reporting LocalWorkingDir as "." rather than blank.

sinbad:
Merging this for now as it appears people need it (or rather the follow-on temp dir work I've done)

Issue #1092, opened 2016-03-18
Broken link in CONTRIBUTING.md


anatolyborodin:
The link https://github.com/github/git-lfs.github.com/blob/gh-pages/_config.yml#L4 "Update Git LFS website" is 404.

See also commit 0fb9c7b10f0cd204899214282a9d2913a94ed1a4 by @technoweenie

technoweenie:
It's a valid link to the private repository for https://git-lfs.github.com, which is accessible to people that have access to make official Git LFS releases.

anatolyborodin:
Wouldn't it be good to add a corresponding remark to that link?

PR #1093, opened 2016-03-19
Mention releng access rights to the git-lfs.github.com repository

anatolyborodin:
One needs to have release engineer access rights to https://github.com/github/git-lfs.github.com to open the [link](https://github.com/github/git-lfs.github.com/blob/gh-pages/_config.yml#L4), it will show `404` otherwise.

See #1092


sinbad:
:+1:

anatolyborodin:
Thank you @sinbad!

technoweenie:
Thanks for this. I think I'll just make that repo public though. Pretty sure there's no good reason to keep it private.

anatolyborodin:
@technoweenie the text should be changed again then?

PR #1096, opened 2016-03-19
Fix `git rev-list` when a file exists with the same name as the branch

epriestley:
Fixes issue #1056.

When you try to push `master` but also have a file named  `master`, we run
`git rev-list ... master`. This is ambiguous, because `rev-list` accepts
paths or refs and `git` is unsure whether you mean the ref `master` or the
path `master`.

Instead, run `git rev-list ... master --`, which is unambiguous.

This adds a failing test (trying to push a file named `master` to LFS), then
fixes the test by appending `--` to the command.


sinbad:
:metal: thanks!

PR #1102, opened 2016-03-22
Report errors generated by packagecloud, exit on failure

javabrett:
I ran `mitmproxy` so I could notice that packagecloud package create was failing with an interesting error.    Prior to this change it fails silently in the script.

This change reports the failure and exits.


technoweenie:
Awesome :metal:

Issue #1074, opened 2016-03-15
push packagecloud versions for linuxmint/{rebecca,rafaela,rosa}


javabrett:
This is a request for releng to push packages to packagecloud for at least `linuxmint/rebecca`, `linuxmint/rafaela` and `linuxmint/rosa`, the current linuxmint LTS releases.  Those are also the push-string specs required.  I imagine that these packages would be identical to their `ubuntu/trusty` upstream equivalents.

The current `git-lfs/script.deb.sh` will create repo specs such as:

```
deb https://packagecloud.io/github/git-lfs/linuxmint/ rebecca main
deb-src https://packagecloud.io/github/git-lfs/linuxmint/ rebecca main
```

Currently, these have to be changed to:

```
deb https://packagecloud.io/github/git-lfs/ubuntu/ trusty main
deb-src https://packagecloud.io/github/git-lfs/ubuntu/ trusty main
```

... in order for installation to be able to proceed.

linuxmint remains the most popular distribution, so it makes sense to push debs for it too.

andyneff:
AFAIK the same Ubuntu packages should work on Mint.

javabrett:
@andyneff Correct.

javabrett:
I'm going to open a conversation with packagecloud about how the same package might be able to be shared across distro and perhaps distro-versions without duplicating the uploads, and still being able to use their nice install repo-setup script.

technoweenie:
@javabrett That sounds great. :metal:

javabrett:
Just to be clear, the motivation here is: packagecloud host the installer package repos, and provide neat scripts to help those on Linux who are less familiar with their (apt, rpm/Yum) package-manager and repo URLs.  These scripts are `script.deb.sh`, `script.rpm.sh` etc.

The problem is that these resolve specifically the installer's exact distribution and version (scripts call these `os` and `dist` respectively, somewhat confusingly to me anyway), so on say LinuxMint, these will automatically resolve to e.g. `os=linuxmint`, `dist=rebecca`.  If we don't upload such a package, the script succeeds, creates an invalid `deb` URL as above, then the install fails.  Presumably the same applies for ol ~= el,, centos ~= el etc.

[...]

PR #1076, opened 2016-03-15
Added Linux Mint distro to packagecloud push, issue #1074.

javabrett:
I haven't got my deb build running yet to be able to test this via push to my packagecloud.  Can build OK but the dpkg build complains about imports.

So I'm not sure this is correct, but someone with a working build would be able to test it, perhaps for the next release.

PR #1103, opened 2016-03-22
Create a new packagecloud Package for each upload

javabrett:
Current API cannot re-read package File contents.  Found during work on #1074. See also https://github.com/computology/packagecloud-ruby/issues/14.

The symptoms here are that a second and subsequent upload based on the same `Package` object fail.  Proxy debug shows that the POSTed form contains empty file data.  This is because `Package` opens the `File` (when passed as a string path), but `Client` reads the contents into MIME during the put package.  There's no close/reopen/rewind on the file, so on subsequent calls there is no data left.

Ideally `Package` bug above could be fixed such that that class reads the MIME bytes and closes the file it opens, so that it can be reused.  But for now, just create a fresh `Package` for each upload.

This may or may not account for sparse package coverage for 1.1.2 on packagecloud.

### DIFF
--- a/script/packagecloud.rb
+++ b/script/packagecloud.rb
@@ -58,7 +58,6 @@ end
 package_files = Dir.glob("repos/**/*.rpm") + Dir.glob("repos/**/*.deb")
 package_files.each do |full_path|
   next if full_path =~ /repo-release/
-  pkg = Packagecloud::Package.new(:file => full_path)
   distro_names = distro_names_for(full_path)
   distro_names.map do |distro_name|
     distro_id = $distro_id_map[distro_name] ||= $client.find_distribution_id(distro_name)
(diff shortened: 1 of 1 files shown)


technoweenie:
Nice deep dive :) Since this is fixed in the gem, should we be ok to close this without merging? 

I'm going to file an issue to add post-release packagecloud tests and a packagecloud gem requirement before doing another release. That shouldn't block #1076 though.

javabrett:
Yes this is now defunct and can be closed.  There's a [further improvement](https://github.com/computology/packagecloud-ruby/issues/16) possible in the packagecloud gem, to avoid re-reading/rewinding the file over and over, but I updated and tested the new 1.0.3 gem sans this change and it works fine.  And #1076 now looks good.  Closing.

PR #1104, opened 2016-03-22
Touch tracked files that are already added to git

### LINKED ISSUE #1057: Lfs detection inconsistent
I'm using Ubuntu 14.04, git version 1.9.1, git-lfs/1.1.1 (GitHub; linux amd64; go 1.5.3).
I seem to be able to create a situation where a committed binary file doesn't get picked up by lfs when lfs tracking is added later.  I've used the following series of commands to demonstrate the issue:
```
git init git-lfs-bug
cd git-lfs-bug
echo 0 > lfs.bin
git add lfs.bin
git commit -m "add binary file"
git lfs track *.bin
git add .gitattributes
git commit -m "track with lfs"
````
After this series of commands I expect that the lfs.bin file should be identified as modified, because it is matched by the git lfs tracking pattern, but the pointer file hasn't been committed yet.  However, when I run git status it says there are no modifications.  And if I run gitk it shows two commits, and no modified files.  Interestingly, gitk shows the first commit as adding a pointer file, which is also incorrect.  

One last detail is that I tried adding the above commands to a shell script and running them that way, but when run from the script the problem did not happen, ie. git status reports lfs.bin as modified and wants me to commit the pointer file.

### PR DESCRIPTION
**Pull request by sinbad**
Resolves #1057 

If a file is already tracked in git when it's added to git-lfs tracking, that file is now 'touched' (`mtime`/`atime` file attributes updated) to ensure that it appears as modified in `git status` along with the `.gitattributes` and therefore prompts the user to re-commit it. Without doing this git doesn't detect that the index is dirty for this file and the user could commit `.gitattributes` but with the file content remaining as non-LFS, which can cause problems when checking out this commit in future. Checking out this commit will forever after show the file as modified in the working copy because `smudge` will not have LFS data to checkout from even though `.gitattributes` says it should have, and `clean` will show a diff.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -777,6 +777,6 @@ func GetTrackedFiles(pattern string) ([]string, error) {
 		line := scanner.Text()
 		ret = append(ret, strings.TrimSpace(line))
 	}
-	return ret, nil
+	return ret, cmd.Wait()
 
 }
--- a/git/git.go
+++ b/git/git.go
@@ -753,3 +753,30 @@ func RemoteRefs(remoteName string) ([]*Ref, error) {
 	}
 	return ret, nil
 }
+
+// GetTrackedFiles returns a list of files which are tracked in Git which match
+// the pattern specified (standard wildcard form)
+// Both pattern and the results are relative to the current working directory, not
+// the root of the repository
+func GetTrackedFiles(pattern string) ([]string, error) {
+	var ret []string
+	cmd := subprocess.ExecCommand("git",
+		"-c", "core.quotepath=false", // handle special chars in filenames
+		"ls-files",
+		"--cached", // include things which are staged but not committed right now
+		"--",       // no ambiguous patterns
+		pattern)
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: git/git.go [hunk 1]
```diff
@@ -753,3 +753,30 @@ func RemoteRefs(remoteName string) ([]*Ref, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed to call git ls-files: %v", err)
+	}
+	cmd.Start()
+	scanner := bufio.NewScanner(outp)
+	for scanner.Scan() {
+		line := scanner.Text()
+		ret = append(ret, strings.TrimSpace(line))
+	}
+	return ret, nil
```
technoweenie: I think this should wait for the command to finish before returning.

technoweenie:
Looks great :+1:. I updated `git.GetTrackedFiles()` so it waits for the command to finish.

sinbad:
> I updated git.GetTrackedFiles() so it waits for the command to finish.

Doh, thanks for catching that.

PR #1105, opened 2016-03-22
pointer: Only write the encoded pointer information to Stdout

sschuberth:
This allows to create a pointer file by redirecting Stdout like

    $ git lfs pointer --file=path/to/file > pointer-to-file

Before this change, "pointer-to-file" would also contain the line saying

    "Git LFS pointer for path/to/file"

which makes the pointer file invalid.

### DIFF
--- a/commands/command_pointer.go
+++ b/commands/command_pointer.go
@@ -52,13 +52,13 @@ func pointerCommand(cmd *cobra.Command, args []string) {
 		}
 
 		ptr := lfs.NewPointer(hex.EncodeToString(oidHash.Sum(nil)), size, nil)
-		fmt.Printf("Git LFS pointer for %s\n\n", pointerFile)
+		fmt.Fprintf(os.Stderr, "Git LFS pointer for %s\n\n", pointerFile)
 		buf := &bytes.Buffer{}
 		lfs.EncodePointer(io.MultiWriter(os.Stdout, buf), ptr)
 
 		if comparing {
 			buildOid = gitHashObject(buf.Bytes())
-			fmt.Printf("\nGit blob OID: %s\n\n", buildOid)
+			fmt.Fprintf(os.Stderr, "\nGit blob OID: %s\n\n", buildOid)
 		}
 	} else {
 		comparing = false
@@ -81,22 +81,22 @@ func pointerCommand(cmd *cobra.Command, args []string) {
 		if !pointerStdin {
 			pointerName = pointerCompare
 		}
-		fmt.Printf("Pointer from %s\n\n", pointerName)
+		fmt.Fprintf(os.Stderr, "Pointer from %s\n\n", pointerName)
 
 		if err != nil {
 			Error(err.Error())
 			os.Exit(1)
 		}
 
-		fmt.Printf(buf.String())
+		fmt.Fprintf(os.Stderr, buf.String())
 		if comparing {
 			compareOid = gitHashObject(buf.Bytes())
-			fmt.Printf("\nGit blob OID: %s\n", compareOid)
+			fmt.Fprintf(os.Stderr, "\nGit blob OID: %s\n", compareOid)
 		}
 	}
 
 	if comparing && buildOid != compareOid {
-		fmt.Printf("\nPointers do not match\n")
+		fmt.Fprintf(os.Stderr, "\nPointers do not match\n")
 		os.Exit(1)
 	}
 


sinbad:
Please can you agree to the Contributor License Agreement since we can't merge this until you do (I think there's a link to do it if you click Details by the error above).

Apart from that this looks good to me but I'd like @technoweenie to OK it as well before merging.

sschuberth:
@sinbad Unfortunately, until I saw the bot check I wasn't aware a CLA is required. I need to get the CLA reviewed by our legal department which might take some some.

sschuberth:
On a related note, I just had a discussion in upstream Git about what kind of information should go to Stdout vs. Stderr, and here are some good answers:

* "Anything that is not the primary output of the program should go to stdout.": http://article.gmane.org/gmane.comp.version-control.git/289512
* "I think the stance of Git is to write only machine readable stuff to stdout [...]": http://article.gmane.org/gmane.comp.version-control.git/289514

sinbad:
Yep, I think that's a pretty good guide.

technoweenie:
Hey @sschuberth, thanks for submitting this. I'm certainly not a linux expert, and will usually defer to whatever Git does in cases like this. I think those guidelines make sense for git plumbing commands, but not git porcelain commands.

```bash
$ git status 2>/dev/null
On branch lfs-repo-usage
Your branch is based on 'origin/lfs-repo-usage', but the upstream is gone.
  (use "git branch --unset-upstream" to fixup)
nothing to commit, working directory clean
```

The default `git status` output isn't designed for machine readable content. So, my interpretation is that those guidelines are only for porcelain commands.

It looks like this PR declares `git lfs pointer` as a plumbing command, while sending some user output to STDERR so it's still readable to humans. Is that common with linux commands? Are there weird edge cases with commands returning STDERR output and a 0 exit code? Based on your profile, I'm assuming you know what you're talking about. I want to make sure that I understand this correctly ;)

sschuberth:
I wouldn't say I'm a Linux expert either (I was mostly working on Git for Windows), but the discussion I was referring to is about `git submodule`, which is a porcelain command. Also, the referred commit message by @peff says

> Putting messages like "Cloning into.." and "done" on stdout is un-Unix and uselessly clutters the stdout channel.

Note how he says "un-Unix", not "un-git-ish" or the like.

Your counter-example of `git status` is a good one, but on the other hand @gitster says (also in my linked thread):

> We may not have been very strict in code reviews to enfore it, and
> especially on the fringes of the system it may be easy to find
> violators, though.

Finally, even if this PR would declare `git lfs pointer` as a plumbing command I'd argue that's a correct thing to do as `git-lfs` itself lists `git lfs pointer` under "Low level commands" in its command line help.

technoweenie:
Nice counter point with `git submodule`. SOLD :+1:

Any idea what it'll take to get the CLA approved? Seems silly to hold up a trivial patch like this. I can totally ninja-merge this, or recreate the changes personally, but I'd rather not so you can get the sweet credit, and potentially contribute in the future ;)

If you can't, or don't feel like going through the hassle, that's totally OK too. I'll just recreate the patch. In the future, you can just open issues for us.

peff:
Yeah, I think you can get into pretty vague, opinionated arguments on stdout vs stderr. Basically, we should strive for usefulness. So if something is the "main" output of a command (as it is with `git status`), it should go to stdout, regardless of whether it is machine readable or not.  If it is extra chattiness that would otherwise make the stdout output harder to parse or reuse, then it should go to stderr (so I think @sschuberth's example shows that).

Something like `clone` is in a weird no-man's land, because it doesn't really produce a "main" output in the first place. Its function is to _do_ something, and all output is basically human-readable chat.

I don't think we have to care about plumbing versus porcelain when making this distinction.

> Are there weird edge cases with commands returning STDERR output and a 0 exit code?

tcl is notorious for treating anything on stderr like an error. IMHO that is just insane.

sschuberth:
> Any idea what it'll take to get the CLA approved?

@technoweenie I just need to push our legal department a bit. Give me a few more days. My patch is not super-urgent, and I'd like to get this sorted out anyway.

> tcl is notorious for treating anything on stderr like an error. IMHO that is just insane.

@peff Just FYI, [PowerShell does the same](https://stackoverflow.com/questions/2095088/error-when-calling-3rd-party-executable-from-powershell-when-using-an-ide). But I don't think it's insane per se. It's called standard *error* after all.

technoweenie:
Honestly, it doesn't sound insane to me either. That's why I wanted to clarify. Should we file a PR to change STDERR to STDUSER or something? :trollface: 

@sschuberth no worries, I'll wait patiently :)

sinbad:
This is such a common sticking point...in SourceTree for quite a while we coloured stderr output red because it was usually best way to highlight errors in large output based on observed Git behaviour, but it could be confusing when people had verbose mode on but the result code was successful. Our eventual solution was to display everything in standard text while the command runs but remembering which was stdout & stderr & applying the highlighting only when the result code was non-zero. Occasionally we highlight more than required in that case but *most* of the time in fatal cases it does pick out just the bad stuff. If only there were 3 defined output streams... :)

sschuberth:
@technoweenie I was able to accept the CLA now :-)

technoweenie:
Thanks! I should've asked about tests. But since I didn't, you get a free ride this time. Here's what I came up with:

https://github.com/github/git-lfs/commit/73f1af9c5fd0a4d59b91a20dce8158ab0cde6a14

It's kind of a good thing though, you would've spent a bunch of time banging your head against the wall because of this:

https://github.com/github/git-lfs/commit/dcf92498ae8ef2b314a1f68e21954fb0b9c8b73b

PR #1106, opened 2016-03-22
Automatic git lfs test dir

chittunoo:


sinbad:
These are my commits ;)

I deliberately only submitted the symlink part as #1091 for review first since the test dir part relied on that & I didn't want the duplication, I'd planned to submit that one after the merge of symlinks. Since it looks like Rick's out at the moment I'll optimistically merge them separately instead of this one.

Issue #1075, opened 2016-03-15
Test "track outside git repo" fails with 51e28193614730498ce78a3c5bf128e8383aa208


javabrett:
**Disclaimer** - this is a fresh build environment by someone who has not built git-lfs previously.

In my `cibuild`, the track-outside test, which asserts that you cannot `git lfs track '../*.foo'` from the top-level of the test-repo, where `..` is also not a repo, fails (in other words the command succeeds whereas the test asserts failure):

```
test: track outside git repo ...                                   FAILED
-- stdout --
    Tracking *.foo
    Initialized empty Git repository in <redacted>/git-lfs/tmp/test-track.sh-4149/track-outside/.git/
    Tracking *.file
    Tracking ../*.foo
-- stderr --
    + set -e
    + git lfs track '*.foo'
    + '[' -n '' ']'
    + git init track-outside
    + cd track-outside
    + git lfs track '*.file'
    + git lfs track '../*.foo'
    + exit 1
    + test_status=1
-- git trace --
   17:20:48.996693 git.c:561               trace: exec: 'git-lfs' 'track' '*.foo'
   17:20:48.996749 run-command.c:334       trace: run_command: 'git-lfs' 'track' '*.foo'
   17:20:49.041419 git.c:344               trace: built-in: git 'init' 'track-outside'
   17:20:49.043785 git.c:561               trace: exec: 'git-lfs' 'track' '*.file'
   17:20:49.043821 run-command.c:334       trace: run_command: 'git-lfs' 'track' '*.file'
   17:20:49.052885 git.c:561               trace: exec: 'git-lfs' 'track' '../*.foo'
   17:20:49.052948 run-command.c:334       trace: run_command: 'git-lfs' 'track' '../*.foo'
```

If I revert in current `master` commit 51e28193614730498ce78a3c5bf128e8383aa208, which changes relevant code in `command_track.go`, the test passes again (git-lfs refuses to add track in non-git parent).  The only complicating factor I can think of is that `track-outside/../../..` is my `git-lfs` clone, although the interceding directories are obviously not part of that repo.

The changed-code seems satisfied as long as my command is run from within the repo.  That might be the intention based on the commit comment "pass gitattributes patterns (mostly) unchanged".

Is allowing that track path now allowed, making the test incorrect?  Does the test need modifying or removal?

Is there a central CI build hosted anywhere, and is the test failing there?

technoweenie:
[...]

PR #1107, opened 2016-03-23
Always run integration tests in a separate GIT_LFS_TEST_DIR

sinbad:
If user doesn't specify a specific one, create one in the temp area & clean up when done if KEEPTRASH=0


sinbad:
Related: #1075 but also safer under fail conditions, doesn't 'infect' git-lfs repo config by accident.

tpdownes:
@singbad I confirm that I can build from master with `debuild -i -us -uc -b`. Thanks!

PR #1115, opened 2016-03-29
Added GO15VENDOREXPERIMENT=0 to dpkg dh build, follow-on from #1025.

javabrett:
Following from #1025.  If you want to run the Debian dpkg build using Go 1.6, you need this change, otherwise the import problems noted in https://github.com/github/git-lfs/pull/1076#issuecomment-199062143 will occur.

It is super-likely that the same need to be added to RPM build `rpm/build_rpms.bsh`, and this could be added on speculation, but I don't have a current build environment on a yum-packaged dist.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -1,6 +1,7 @@
 #!/usr/bin/make -f
 
 export DH_OPTIONS
+export GO15VENDOREXPERIMENT=0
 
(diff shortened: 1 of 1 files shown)


technoweenie:
Thanks :metal:

PR #1117, opened 2016-03-31
Account for url.<url>.insteadOf in all operations

artagnon:
We tackle the problem in two places: first, when a raw clone url is given, we
must load possible replacement for substrings from the config; second, when we
are fetching a certain remote's url from the config, we must try to replace the
relevant part of the url.

The implementation iterates over the map insteadOfs, and attempts to
strings.Replace() the url with all the key-value pairs.

### DIFF
--- a/lfs/config.go
+++ b/lfs/config.go
@@ -59,6 +59,7 @@ type Configuration struct {
 	gitConfig         map[string]string
 	origConfig        map[string]string
 	remotes           []string
+	insteadOfs        map[string]string
 	extensions        map[string]Extension
 	fetchIncludePaths []string
 	fetchExcludePaths []string
@@ -304,6 +305,11 @@ func (c *Configuration) Remotes() []string {
 	return c.remotes
 }
 
+func (c *Configuration) InsteadOfs() map[string]string {
+	c.loadGitConfig()
+	return c.insteadOfs
+}
+
 // GitProtocol returns the protocol for the LFS API when converting from a
 // git:// remote url.
 func (c *Configuration) GitProtocol() string {
@@ -416,6 +422,8 @@ func (c *Configuration) loadGitConfig() bool {
 	c.gitConfig = make(map[string]string)
 	c.extensions = make(map[string]Extension)
 	uniqRemotes := make(map[string]bool)
+	uniqInsteadOfs := make(map[string]string)
+	c.insteadOfs = make(map[string]string)
 
 	configFiles := []string{
 		filepath.Join(LocalWorkingDir, ".lfsconfig"),
@@ -423,12 +431,17 @@ func (c *Configuration) loadGitConfig() bool {
 		// TODO: remove .gitconfig support for Git LFS v2.0 https://github.com/github/git-lfs/issues/839
 		filepath.Join(LocalWorkingDir, ".gitconfig"),
 	}
-	c.readGitConfigFromFiles(configFiles, 0, uniqRemotes)
 
 	listOutput, err := git.Config.List()
+
 	if err != nil {
 		panic(fmt.Errorf("Error listing git config: %s", err))
 	}
+
+	c.readGitInsteadOfConfig(listOutput, uniqInsteadOfs)
+	c.insteadOfs = uniqInsteadOfs
+
+	c.readGitConfigFromFiles(configFiles, 0, uniqRemotes)
 
 	c.readGitConfig(listOutput, uniqRemotes, false)
 
@@ -537,6 +550,11 @@ func (c *Configuration) readGitConfig(output string, uniqRemotes map[string]bool
 			continue
 		}
 
+		if len(keyParts) == 3 && (keyParts[2] == "url" || keyParts[2] == "pushurl") {
+			for aliasurl, replacement := range c.insteadOfs {
+				value = strings.Replace(value, aliasurl, replacement, 1)
+			}
+		}
 		c.gitConfig[key] = value
 
 		if len(keyParts) == 2 && keyParts[0] == "lfs" && keyParts[1] == "fetchinclude" {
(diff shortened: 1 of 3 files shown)


technoweenie:
Hi @artagnon, thanks for the patch :metal: This is just an initial triage of the PR, so I haven't reviewed the code yet, but it now has a `review` label so one of us can come back to it. You do have a test that you should be able to replicate by running `test/test-config.sh`. The stderr output is echoing all the commands, until a command returns a non-zero exit code, failing the test. It's failing on this `grep 'Endpoint=http://actual-url (auth=none)' env.log` line near the bottom:

```
-- stderr --
    +set -e
    +reponame=default-config
    +mkdir default-config
    +cd default-config
    +git init
    +git remote add origin http://127.0.0.1:40053/default-config
    +tee env.log
    +git lfs env
    +grep 'Endpoint=http://127.0.0.1:40053/default-config.git/info/lfs (auth=none)' env.log
    +git config --file=.gitconfig lfs.url http://gitconfig-file-ignored
    +git config --file=.lfsconfig lfs.url http://lfsconfig-file
    +tee env.log
    +git lfs env
    +grep 'Endpoint=http://lfsconfig-file (auth=none)' env.log
    +git config lfs.url http://local-lfsconfig
    +tee env.log
    +git lfs env
    +grep 'Endpoint=http://local-lfsconfig (auth=none)' env.log
    +git config url.http://actual-url.insteadOf alias:
    +git lfs env
    +tee env.log
    +grep 'Endpoint=http://actual-url (auth=none)' env.log
    +end_test
    +test_status=1
    +set +x -e
```

https://travis-ci.org/github/git-lfs/jobs/119704573

artagnon:
Oops, fixed the minor detail. Thanks.

```diff
diff --git a/test/test-config.sh b/test/test-config.sh
index a535651..bf4a63e 100755
--- a/test/test-config.sh
+++ b/test/test-config.sh
@@ -22,9 +22,10 @@ begin_test "default config"
   git lfs env | tee env.log
   grep "Endpoint=http://local-lfsconfig (auth=none)" env.log

-  git config url."http://actual-url".insteadOf alias:
+  git config url."http://actual-url/".insteadOf alias:
+  git config lfs.url alias:rest
   git lfs env | tee env.log
-  grep "Endpoint=http://actual-url (auth=none)" env.log
+  grep "Endpoint=http://actual-url/rest (auth=none)" env.log
 )
 end_test
```

lfs/config.go
```diff
@@ -544,6 +565,38 @@ func (c *Configuration) readGitConfig(output string, uniqRemotes map[string]bool
 	}
 }
 
+// Special handling for url.<url>.insteadOf, since <url> will contain '.'
+func (c *Configuration) readGitInsteadOfConfig(output string, uniqInsteadOfs map[string]string) {
+	lines := strings.Split(output, "\n")
+	uniqKeys := make(map[string]string)
+	for _, line := range lines {
+		// Why is this not "="?
+		pieces := strings.SplitN(line, " ", 2)
```
technoweenie: > // Why is this not "="?

This is the first time the `pieces` var is used in this scope, so it must be declared. Same goes for values like `key` or `value` below. I suppose we could declare them above the `for` block, but I don't think it's worth the hassle.

lfs/config.go
```diff
@@ -409,16 +415,26 @@ func (c *Configuration) loadGitConfig() bool {
...
 		// TODO: remove .gitconfig support for Git LFS v2.0 https://github.com/github/git-lfs/issues/839
 		filepath.Join(LocalWorkingDir, ".gitconfig"),
 	}
+
+	listOutput, err := git.Config.ListRegexp(`^url.*`)
+	if err != nil {
+		panic(fmt.Errorf("Error listing git config with regexp `^url.*`: %s", err))
+	}
+	c.readGitInsteadOfConfig(listOutput, uniqInsteadOfs)
+	c.insteadOfs = uniqInsteadOfs
```
technoweenie: Does this need to call `git config` again? I wonder if it'd be better to just filter the values from the `git.Config.List()` call below? 

We try to limit the number of process execs where reasonable, especially in a spot like this that's called on every `git lfs` command. With default settings, doing an initial clone on a repository with lots of LFS objects will call `git lfs smudge` over and over again. This is especially bad on Windows, which is slower at execing a new process than linux. 

It's probably worth writing a quick benchmark so we can compare different approaches on the various platforms.

lfs/config.go
artagnon: No, I meant why the string we're splitting on is not "=": inconsistency in the way `git config` reports; see line 489.

lfs/config.go
artagnon: Excuse my thinko there. Fixed. We don't really need the `--get-regexp` function anymore; do you want to keep it around anyway?

lfs/config.go
technoweenie: Doh, sorry. That's silly of me. No clue why `git config --list` differs from `--get-regexp`, either. Looks like the only way we could parse the output of both commands the same is by using `-z`.

lfs/config.go
technoweenie: Nope, I think you were right to remove it. :+1:

technoweenie:
Hey @artagnon, thanks for reworking this based on my feedback so quickly. I'm beginning the QA phase for Git LFS 1.2 now, but this will probably just barely make the cut :)

sinbad:
If we're going to support `insteadOf`, should we also support `pushInsteadOf`? I added upload/download context on most operations to support `pushurl` (#949) so it should be doable.

lfs/config.go
```diff
@@ -416,20 +422,27 @@ func (c *Configuration) loadGitConfig() bool {
...
 	listOutput, err := git.Config.List()
+
 	if err != nil {
 		panic(fmt.Errorf("Error listing git config: %s", err))
 	}
 
+	c.readGitInsteadOfConfig(listOutput, uniqInsteadOfs)
+	c.insteadOfs = uniqInsteadOfs
+
+	c.readGitConfigFromFiles(configFiles, 0, uniqRemotes)
```
sinbad: There's a silent dependency here on `readGitInsteadOfConfig` being called before `readGitConfigFromFiles`, and having assigned `c.insteadOfs` before it, even though that member var isn't used anywhere else. 

I think the call to `readGitInsteadOfConfig` should be nested *inside* the  `readGitConfigFromFiles` to make this explicit, and the output `uniqInsteadOfs` not stored in the `Config` at all, just used to affect the behaviour of the rest of `readGitConfigFromFiles`, as it currently does. This would eliminate the hidden 'magic' ordering and unused state.

technoweenie:
You think it's worth shipping this, or waiting until `pushInsteadOf` is supported? I think it's useful as-is, and `pushInsteadOf` could be added in 1.2.1 or 1.3 (whatever we do next).

I'm inclined to let https://github.com/github/git-lfs/pull/1117/files#r59379179 slide until the next version too.

@sinbad: if you feel strongly about it, we can hold off though.

sinbad:
TBH this is the first time I've ever heard of these features, I'd never heard from anyone who used them before :) When I looked them up they just seemed to go together. If it's important to ship now I don't have major objections.

technoweenie:
Hey @artagnon, I found a bug during my QA where it looks like this is doing the `insteadof` replacements twice for default remote urls in `remote.<name>.url`. It works great for `lfs.url` and `remote.<name>.lfsurl` though.

Unfortunately, I'm going to hold this back from Git LFS v1.2. I'd like the following to be fixed before I merge this:

* [ ] Fix the double-replacement issue I mentioned above ^
* [ ] Support `url.<url>.pushinsteadof` as suggested [here](https://github.com/github/git-lfs/pull/1117#issuecomment-208913572).
* [ ] Fix this `readGitInsteadofConfig()` dependency issue described [here](https://github.com/github/git-lfs/pull/1117#discussion_r59379179).

Thanks for your work on this. Let me know if I can help you out in anyway on here or [gitter](https://gitter.im/github/git-lfs).

artagnon:
Thanks for catching those bugs @technoweenie. I'll be on vacation for 10 days, and I'll continue development when I'm back.

technoweenie:
@artagnon Hey, friendly nudge. Do you have time to pick this back up? If not, that's totally cool. I or someone else should be able to push this over the finish line.

artagnon:
@technoweenie Hey, I don't think I have the time to finish this up in the
near future. In any case, I'll send a PR if I find this incomplete when I
find some time. Don't wait up for me though.

Thanks.

technoweenie:
@artagnon Thanks again for submitting this PR. Unfortunately, I dropped the ball here. Lately, we've been on a refactoring kick in the `config` package (see #1425), meaning this PR doesn't merge cleanly anymore. I've rewritten the PR: #1443.

I also changed the implementation, since `strings.Replace()` doesn't work with some of the edge cases from git's documentation. The specific items are in bold:

> Any URL that **starts with** this value will be rewritten to start, instead, with <base>. In cases where some site serves a large number of repositories, and serves them with multiple access methods, and some users need to use different access methods, this feature allows people to specify any of the equivalent URLs and have Git automatically rewrite the URL to the best alternative for the particular user, even for a never-before-seen repository on the site. **When more than one insteadOf strings match a given URL, the longest match is used.**

Just FYI :)

Thanks again for your original patch. Hope to see you around again!

PR #1118, opened 2016-03-31
Catch previously undetected errors from git exit codes in async scans

### LINKED ISSUE #1100: `git lfs push <remote> <ref>` has poor behavior when tag and branch share a name
I don't think users can actually hit this directly, so I'm mostly filing this for completeness.

#1096,  #1056, and #1079 are related, and deal with a situation where a branch (like `master`) has the same name as a _file_ in the working copy. Users _can_ hit those cases, but #1096 should always resolve them because they are straightforward to disambiguate.

A more subtle case is when a where a branch has the same name as a _tag_ instead of the same name as a _file_. Setup is roughly:

  - Check out a branch.
  - Track, add, and commit some LFS stuff.
  - Tag anything with a tag that has the same name as the branch.
  - Run `git lfs push origin <branch>`.

When you run `git push` in a similar scenario, it stops you immediately because the ref name is ambiguous:

```
$ GIT_TRACE=1 git push origin branch-and-tag
07:32:46.028980 git.c:348               trace: built-in: git 'push' 'origin' 'branch-and-tag'
error: src refspec branch-and-tag matches more than one.
error: failed to push some refs to 'git@github.com:epriestley/poems.git'
```

However, when you run `git lfs push origin master`, it runs "successfully" but with the wrong behavior:

```
$ GIT_TRACE=1 git lfs push origin branch-and-tag
07:32:49.058547 git.c:558               trace: exec: 'git-lfs' 'push' 'origin' 'branch-and-tag'
07:32:49.058924 run-command.c:335       trace: run_command: 'git-lfs' 'push' 'origin' 'branch-and-tag'
trace git-lfs: Upload refs [branch-and-tag] to remote origin
trace git-lfs: run_command: git rev-list --objects branch-and-tag --not --remotes=origin --
trace git-lfs: run_command: git cat-file --batch-check
trace git-lfs: tq: starting 3 transfer workers
trace git-lfs: tq: running as individual queue
```

This command exits `0`, but should not. Internally, `git rev-list` has failed:

```
$ git rev-list --objects branch-and-tag --not --remotes=origin --
warning: refname 'branch-and-tag' is ambiguous.
```

`git-lfs` fails to detect this, interprets "no output" as "no matching objects", and concludes it doesn't need to push anything.

Instead, it should probably detect that `git rev-list` has failed and exit with an error similar to the one `git` exits with.

sinbad:
Fixes #1100 as well as surfacing a number of other fail states previously missed.

Wraps all scanner channels so that consumer only has to worry about primary results as channels, then check error result from `wrapper.Wait()` at the end to pick up any bad exit conditions (rather than having to handle 2 channels themselves).

Fixed a bunch of other issues while I was at it.

### DIFF
--- a/lfs/scanner.go
+++ b/lfs/scanner.go
@@ -1055,13 +1055,14 @@ type PointerChannelWrapper struct {
 }
 
 // Construct a new channel wrapper for WrappedPointer
-// Caller can use s.Pointers directly for normal processing then call Wait() to finish & check for errors
+// Caller can use s.Results directly for normal processing then call Wait() to finish & check for errors
 // Scan function is required to create error channel large enough not to block (usually 1 is ok)
 func NewPointerChannelWrapper(pointerChan <-chan *WrappedPointer, errorChan <-chan error) *PointerChannelWrapper {
 	return &PointerChannelWrapper{&BaseChannelWrapper{errorChan}, pointerChan}
 }
 
 // ChannelWrapper for string channel functions to more easily return async error data via Wait()
+// Caller can use s.Results directly for normal processing then call Wait() to finish & check for errors
 // See NewStringChannelWrapper for construction / use
 type StringChannelWrapper struct {
 	*BaseChannelWrapper
@@ -1069,6 +1070,7 @@ type StringChannelWrapper struct {
 }
 
 // Construct a new channel wrapper for string
+// Caller can use s.Results directly for normal processing then call Wait() to finish & check for errors
 func NewStringChannelWrapper(stringChan <-chan string, errorChan <-chan error) *StringChannelWrapper {
 	return &StringChannelWrapper{&BaseChannelWrapper{errorChan}, stringChan}
 }
@@ -1081,6 +1083,7 @@ type TreeBlobChannelWrapper struct {
 }
 
 // Construct a new channel wrapper for TreeBlob
+// Caller can use s.Results directly for normal processing then call Wait() to finish & check for errors
 func NewTreeBlobChannelWrapper(treeBlobChan <-chan TreeBlob, errorChan <-chan error) *TreeBlobChannelWrapper {
 	return &TreeBlobChannelWrapper{&BaseChannelWrapper{errorChan}, treeBlobChan}
 }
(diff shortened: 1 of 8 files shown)


lfs/scanner.go
```diff
@@ -397,11 +433,20 @@ func revListIndex(cache bool, indexMap *indexFileMap) (chan string, error) {
 			}
 		}
 
+		// Note: deliberately not checking result code here, because doing that
+		// can fail fsck process too early since clean filter will detect errors
+		// and set this to non-zero. How to cope with this better?
+		// stderr, _ := ioutil.ReadAll(cmd.Stderr)
+		// err := cmd.Wait()
+		// if err != nil {
+		// 	errchan <- fmt.Errorf("Error in git diff-index: %v %v", err, string(stderr))
```
sinbad: I'm not totally sure what to do about this yet, any ideas?

lfs/scanner.go
```diff
@@ -930,7 +1045,75 @@ func parseLogOutputToPointers(log io.Reader, dir LogDiffDirection,
...
+	errorChan <-chan error
+}
+
+func (w *BaseChannelWrapper) Wait() error {
+
+	var err error
+	for e := range w.errorChan {
+		if err != nil {
+			// Combine in case multiple errors
+			err = fmt.Errorf("%v\n%v", err, e)
```
technoweenie: This will lose the backtrace from any scans that return multiple errors. Any panics will show this as the culprit of the bug, and not a scanner that returns a channel wrapper. If it becomes a problem, I could see us using some `MultiError` type that preserves them, and teaching `commands.Panic()` how to handle it.

technoweenie:
This looks good to me. :+1:

> Fixed a bunch of other issues while I was at it.

What issues? Anything worth adding a test for?

lfs/scanner.go
sinbad: True, I did consider pulling in https://github.com/hashicorp/go-multierror but thought it was overkill, and concentrated on making the error text itself return more useful info instead.

sinbad:
> What issues? Anything worth adding a test for?

The main specific one was in `pre-push` and `push` where it would not correctly detect delete branch push modes. You're right that should have tests on it, I'll add a couple.

Other issues were just clarifications really, most of all the prevalence of `exit code 128` errors (now includes `stderr` so you know what the problem was) and just generally the failure to detect `git` errors in all of these methods, some of which are used outside of the scan context.

PR #1127, opened 2016-04-06
update roadmap for v1.3+

technoweenie:
I have one more larger proposal for a feature (file expiration), and a bunch of smaller bugs to add.


technoweenie:
I'm ready to merge this. Definitely doesn't shut the door for v1.3 stuff. But, it has everything from #844 that I won't make it into 1.2.

sinbad:
:+1:

PR #1128, opened 2016-04-06
push optimizations (take 2)

technoweenie:
This PR brings back the push optimizations from #1040. Hit up #1040 and #1071 to get the original context. This makes a few changes based on feedback:

* #969 is no longer included. That will come in the next PR.
* https://github.com/github/git-lfs/pull/1071#issuecomment-196280487

    > Are you intending to wrap downloading in clientContext too? Although there's somewhat less for it to do than in the upload case. If not, or even if 'not right away' then maybe rename clientContext to uploadContext for now.

    It's now called `uploadContext`. It's still unexported, so I expect the type to change and move from the `commands` package some day.

* https://github.com/github/git-lfs/pull/1071/files#diff-d540c48abbcd5b13b0acd4cbef433700R57

    > Hmm, feels weird to have to copy the remote name when everything else uses Config directly

    Handled in `newUploadContext()`

* https://github.com/github/git-lfs/pull/1071/files#r55989899

    > This now feels like non-intuitive output to get when asked to push 2 oids.

    Agreed. There are now fewer test changes. The reason the "skipped" output changed, is because it was being handled in [the old filter()](https://github.com/github/git-lfs/pull/1071/files#diff-3cff4ed9c5b60b5b81240c88cd74e404R69) _before_ the `*lfs.TransferQueue` object for the upload is created. Now, the only objects the `*uploadContext` skips those that have already been uploaded in the current process (usually from another ref).

### DIFF
--- a/commands/uploader.go
+++ b/commands/uploader.go
@@ -22,7 +22,7 @@ func newUploadContext(dryRun bool) *uploadContext {
 
 // AddUpload adds the given oid to the set of oids that have been uploaded in
 // the current process.
-func (c *uploadContext) AddUpload(oid string) {
+func (c *uploadContext) SetUploaded(oid string) {
 	c.uploadedOids.Add(oid)
 }
 
@@ -97,7 +97,7 @@ func (c *uploadContext) checkMissing(missing []*lfs.WrappedPointer, missingSize 
 	done := make(chan int)
 	go func() {
 		for oid := range transferc {
-			c.AddUpload(oid)
+			c.SetUploaded(oid)
 		}
 		done <- 1
 	}()
@@ -115,7 +115,7 @@ func upload(c *uploadContext, unfiltered []*lfs.WrappedPointer) {
 			}
 
 			Print("push %s => %s", p.Oid, p.Name)
-			c.AddUpload(p.Oid)
+			c.SetUploaded(p.Oid)
 		}
 
 		return
@@ -133,7 +133,7 @@ func upload(c *uploadContext, unfiltered []*lfs.WrappedPointer) {
 		}
 
 		q.Add(u)
-		c.AddUpload(p.Oid)
+		c.SetUploaded(p.Oid)
 	}
 
 	q.Wait()
--- a/commands/uploader.go
+++ b/commands/uploader.go
@@ -1,0 +1,130 @@
+package commands
+
+import (
+	"os"
+
+	"github.com/github/git-lfs/lfs"
+)
+
+var uploadMissingErr = "%s does not exist in .git/lfs/objects. Tried %s, which matches %s."
+
+type uploadContext struct {
+	RemoteName   string
+	DryRun       bool
+	uploadedOids lfs.StringSet
+}
+
+func newUploadContext() *uploadContext {
+	return &uploadContext{
+		uploadedOids: lfs.NewStringSet(),
+		RemoteName:   lfs.Config.CurrentRemote,
+	}
+}
+
+func (c *uploadContext) Upload(unfilteredPointers []*lfs.WrappedPointer) {
+	filtered := c.filterUploadedObjects(noSkip, unfilteredPointers)
+
+	totalSize := int64(0)
+	for _, p := range filtered {
+		totalSize += p.Size
+	}
+
+	uploadQueue := lfs.NewUploadQueue(len(filtered), totalSize, c.DryRun)
+
+	if c.DryRun {
+		for _, pointer := range filtered {
+			Print("push %s => %s", pointer.Oid, pointer.Name)
+			c.uploadedOids.Add(pointer.Oid)
+		}
+		return
+	}
+
+	c.filterServerObjects(filtered)
+	pointers := c.filterUploadedObjects(uploadQueue, filtered)
+
+	for _, pointer := range pointers {
+		u, err := lfs.NewUploadable(pointer.Oid, pointer.Name)
+		if err != nil {
+			if lfs.IsCleanPointerError(err) {
+				Exit(uploadMissingErr, pointer.Oid, pointer.Name, lfs.ErrorGetContext(err, "pointer").(*lfs.Pointer).Oid)
+			} else {
+				ExitWithError(err)
+			}
+		}
+
+		uploadQueue.Add(u)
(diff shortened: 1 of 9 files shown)


technoweenie:
@dluksza: I applied #969 on top of this in https://github.com/github/git-lfs/compare/include-refs-in-transfers?expand=1. Still working on it :)

I wouldn't check that branch out yet, I plan to rebase it against this PR once it's merged.

git/git.go
```diff
@@ -163,6 +176,39 @@ func RemoteList() ([]string, error) {
...
+// repository. Other refs (HEAD, refs/stash, git notes) are ignored.
+func LocalRefs() ([]*Ref, error) {
+	cmd := subprocess.ExecCommand("git", "show-ref", "--heads", "--tags")
+
+	outp, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to call git show-ref: %v", err)
+	}
+	cmd.Start()
+	defer cmd.Wait()
```
sinbad: I think any error from `cmd.Wait()` should be returned at the end.

commands/uploader.go
```diff
@@ -0,0 +1,130 @@
...
+
+func newUploadContext() *uploadContext {
+	return &uploadContext{
+		uploadedOids: lfs.NewStringSet(),
+		RemoteName:   lfs.Config.CurrentRemote,
+	}
+}
+
+func (c *uploadContext) Upload(unfilteredPointers []*lfs.WrappedPointer) {
+	filtered := c.filterUploadedObjects(noSkip, unfilteredPointers)
```
sinbad: How about making filterUploadedObjects return the unfiltered size as well to avoid the repeat iteration below?

commands/uploader.go
```diff
@@ -0,0 +1,130 @@
...
+			q.Skip(pointer.Size)
+		} else {
+			filtered = append(filtered, pointer)
+		}
+	}
+
+	return filtered
+}
+
+func (c *uploadContext) filterServerObjects(pointers []*lfs.WrappedPointer) {
```
sinbad: I think "filterServerObjects" is a misleading name as this doesn't filter a list, it adds any missing objects which are on the server to the list of already uploaded objects. `checkSkipMissingObjects` maybe?

commands/uploader.go
```diff
@@ -0,0 +1,130 @@
...
+	if c.DryRun {
+		for _, pointer := range filtered {
+			Print("push %s => %s", pointer.Oid, pointer.Name)
+			c.uploadedOids.Add(pointer.Oid)
+		}
+		return
+	}
+
+	c.filterServerObjects(filtered)
+	pointers := c.filterUploadedObjects(uploadQueue, filtered)
```
sinbad: If you made the `c.filterServerObjects(filtered)` call at the start of the function you could avoid calling `c.filterUploadedObjects` twice I think? You could pre-check each against `c.uploadedOids` inside `filterServerObjects` to avoid any unnecessary work across round trips.

commands/uploader.go
```diff
@@ -0,0 +1,130 @@
...
+				Error(inner.Error())
+			}
+			Error(err.Error())
+		}
+	}
+
+	if len(uploadQueue.Errors()) > 0 {
+		os.Exit(2)
+	}
+}
```
sinbad: I kinda feel like the `Upload` method should return errors rather than log & exit itself, and leave it to the commands to control the messaging & exit. I know this way eliminates more duplication and that this is still in the `command` package so it's fair for it to be there, but it feels like it'll probably end up in `lfs` or somewhere later? No big deal if you want to leave it here for now but I initially wondered where the error handling had gone.

sinbad:
:+1: just stylistic & potential optimisation comments from me. This is a really big improvement for overlapping refs.

commands/uploader.go
technoweenie: It filters in memory first, so it doesn't have to re-check the server for things that have been transferred in the same process for another ref. It should make a big difference when running `git lfs push --all` on a big repo with lots of branches.

commands/uploader.go
technoweenie: I could just move that to an unexported `commands.handleUpload()` func or something. Maybe that'll communicate my intentions with this class better?

commands/uploader.go
sinbad: Yeah I know, I meant that if you did this inside  `filterServerObjects`:
```
for _, pointer := range pointers {
 		if !c.uploadedOids.Contains(pointer.Oid) && !lfs.ObjectExistsOfSize(pointer.Oid, pointer.Size) {
 			// We think we need to push this but we don't have it
```
.. then you could do this ahead of a single call to `filterUploadedObjects` instead?

git/git.go
technoweenie: Fixed in 0935817c976bfd4bb8bc6f2564ee1f642326cb8d.

technoweenie:
Big refactor in 0935817c976bfd4bb8bc6f2564ee1f642326cb8d taking most of your feedback into account.

* A `commands.uploadContext` has one main responsibility now: build an `lfs.TransferQueue` to upload a set of objects, ignoring objects that have been uploaded already in the same process.
* A `commands.uploadContext` only creates fewer slices when filtering pointers to upload. 
* `commands.upload()` handles dry run behavior vs real behavior, and all the cli error handling for it.

sinbad:
:+1: Like it, lots of iteration has collapsed too & I like the new methods. I have maybe one more suggestion but it's optional, this would be fine.

commands/uploader.go
```diff
@@ -0,0 +1,155 @@
...
+func newUploadContext(dryRun bool) *uploadContext {
+	return &uploadContext{
+		DryRun:       dryRun,
+		uploadedOids: lfs.NewStringSet(),
+	}
+}
+
+// AddUpload adds the given oid to the set of oids that have been uploaded in
+// the current process.
+func (c *uploadContext) AddUpload(oid string) {
```
sinbad: Nit picky comment: `AddUpload` sounds like a request to add an upload to be done, how about either `AddUploaded` or `SetUploaded` to match `HasUploaded`?

technoweenie: :+1:

PR #1133, opened 2016-04-08
Zsh completions

sinbad:
Replaces #727 

Generates the zsh completion script dynamically from a combination of Cobra data and the ManPages dictionary. Tested with ohmyzsh so far.

So far only works with `git-lfs` command, not `git lfs`. I've tried to make this work and I can get as far as overriding the `git` completion in this script, but can't figure out how to only override it for the `lfs` subcommand and delegate back to the original completions for everything else. My script-fu is not strong enough yet.


commands/command_integrate.go
```diff
@@ -0,0 +1,206 @@
...
+import (
+	"fmt"
+	"os"
+	"os/user"
+	"path/filepath"
+	"strings"
+
+	"github.com/github/git-lfs/vendor/_nuts/github.com/spf13/pflag"
+
+	"github.com/github/git-lfs/vendor/_nuts/github.com/spf13/cobra"
```
technoweenie: Do you run `goimports`? Seems like it should've removed that blank line on line 11.

commands/command_integrate.go
```diff
@@ -0,0 +1,206 @@
...
+		ohmyzsh := getOhMyZshRootDir()
+		if len(ohmyzsh) == 0 {
+			Exit("Could not locate oh-my-zsh root dir via $ZSH")
+		}
+		outputFile = filepath.Join(ohmyzsh, "plugins", "git-lfs", "git-lfs.plugin.zsh")
+	} else if ohmyzsh := getOhMyZshRootDir(); len(ohmyzsh) > 0 {
+		// default user-specific oh-my-zsh
+		outputFile = filepath.Join(getOhMyZshRootDir(), "custom", "plugins", "git-lfs", "git-lfs.plugin.zsh")
+	} else {
+		outputFile = filepath.Join(getHomeDir(), "_git-lfs.zsh")
```
technoweenie: Is this the plain zsh install method? Pretty sure it needs to be in one of the `$fpath` directories, named `_git-lfs` exactly (no extension). Now, how do we get `$fpath` outside of the zsh shell?

sinbad: It doesn't have to be, you can just `source` it in your `~/.zshrc. It can auto-load if you put it in `$fpath` and it's called `_git-lfs` but I have no way to tell which folder in `$fpath` to put it in; I'd have to put it somewhere else and maybe add it to `$fpath` which I didn't really like either. Thought the best option was to let the user decide & just tell them where it is.

commands/command_integrate.go
sinbad: Yeah, my Sublime setup runs `goimports` every time I save, it obviously doesn't strips blanks.

commands/command_integrate.go
technoweenie: Cool, works for me. At the least, it's easy to script this myself if I'm particular about the file location :+1:

commands/command_init.go
```diff
@@ -1,33 +1,28 @@
...
-		Use: "hooks",
-		Run: initHooksCommand,
+		Use:        "hooks",
+		Deprecated: "Use 'git lfs install' now",
+		Run:        initHooksCommand,
 	}
 )
 
 func initCommand(cmd *cobra.Command, args []string) {
-	fmt.Fprintf(os.Stderr, "WARNING: 'git lfs init' is deprecated. Use 'git lfs install' now.\n")
```
technoweenie: Why'd you take these warnings out?

technoweenie:
This works pretty well, but I don't think it's quite ready for v1.2.

First, subcommand arguments don't work:

```
$ git-lfs fetch _arguments:450: command not found: __git_remotes
_arguments:450: command not found: __git_branch_names
_arguments:450: command not found: __git_remotes
_arguments:450: command not found: __git_branch_names
_arguments:450: command not found: __git_remotes
_arguments:450: command not found: __git_branch_names

$ git-lfs pre-push _arguments:450: command not found: __git_remote_branch_names
_arguments:450: command not found: __git_remote_branch_names
_arguments:450: command not found: __git_remote_branch_names
```

Second, I'd still really like to figure out how to autocomplete with `git lfs`, so it'd be useful to a wide variety of users.

A "nice to have" would be some integration with `git lfs install` or `git lfs update`. I'd love to set it up (maybe through a `--global` git config value) to write the zsh (and other potential integrations) to some location I setup. Then they automatically update every time I update Git LFS.

commands/command_init.go
sinbad: I'm using the `Deprecated` string on `cobra.Command` instead which prints a warning when you use it (also lets me detect deprecated commands and ignore them)

sinbad:
Do you have the `git` auto completions enabled? As in the original PR this script depends on them, that's where the `__git_*` functions are defined.

Yeah I'd really like to autocomplete `git lfs` too and I couldn't figure out how to do it. Would be great if someone with better knowledge of zsh could pitch in now it's automated.

technoweenie:
I have some form of git autocompletions, though I may want to investigate further. 

1__technoweenie_ops-bastion4-cp1-prd_____zsh__and_zsh_completions_by_sinbad_ _pull_request__1133_ _github_git-lfs

technoweenie:
I hate to kill this, you did great work @sinbad. I just don't think it's worth doing if it doesn't work with `git lfs ...`.

ttaylorr:
@robobenklein I wasn't able to make this work with the snippet you provided (switching `flow` for `lfs`). Can you advise?

robobenklein:
I have yet to define any of the subcommands yet, but I get first-level completion for `git lfs *` using what I have currently: https://github.com/robobenklein/configs/blob/e6b0c03566f025641a7d76fd26b662fa96911923/zsh/fpath/_git-lfs.zsh_completion

Note that the subcommand completions need to be sourced to run the `zstyle` setup for it. For `git-lfs` it works without the need to source it as long as it's `fpath` registered.

It gives me completion like this:

```
robo@robo-unseptium ~> git lfs <tab>
checkout   -- Populate working copy with real content from Git LFS files.
clone      -- Efficiently clone a Git LFS-enabled repository.
env        -- Display the Git LFS environment.
fetch      -- Download Git LFS files from a remote.
fsck       -- Check Git LFS files for consistency.
install    -- Install Git LFS configuration.
lock       -- Set a file as "locked" on the Git LFS server.
locks      -- List currently "locked" files from the Git LFS server.
logs       -- Show errors from the git-lfs command.
ls-files   -- Show information about Git LFS files in the index and working tree.
migrate    -- Migrate history to or from git-lfs
prune      -- Delete old Git LFS files from local storage
pull       -- Fetch LFS changes from the remote & checkout any required working tree files.
push       -- Push queued large files to the Git LFS endpoint.
status     -- Show the status of Git LFS files in the working tree.
track      -- View or add Git LFS paths to Git attributes.
uninstall  -- Uninstall Git LFS by removing hooks and smudge/clean filter configuration.
unlock     -- Remove "locked" setting for a file on the Git LFS server.
untrack    -- Remove Git LFS paths from Git Attributes.
update     -- Update Git hooks for the current Git repository.
```

ttaylorr:
Hi @robobenklein, I can confirm that your script completes correctly for command invocations of the form `git-lfs _`, but I agree with technoweenie above and would prefer that we complete `git lfs _` (note the missing `-`) as well.

It looks like https://github.com/git/git/commit/085e2ee0e62a97c2dc56d4a6baf3e03ce124bba9 allows us to do just that.

If you're interested in working on this topic, can you please prepare something that will work with the above?

robobenklein:
Got it working with both `git lfs` and `git-lfs` as of this version: https://github.com/robobenklein/configs/blob/433eca4021efd31f7bc989a2f7e25f0ba88e898c/zsh/fpath/_git-lfs

You should just be able to drop it in the fpath now. Also implemented fetch/ls-files/push subcommands.

https://asciinema.org/a/MXbve4MbOHo7dpnc4r2lhdpPC (alias g=git)

I don't know what that `_git` completion error is though, can somebody check if that's just me?

robobenklein:
I have working ZSH completion for `git-lfs` and `git lfs` now.

https://github.com/robobenklein/configs/blob/433eca4021efd31f7bc989a2f7e25f0ba88e898c/zsh/fpath/_git-lfs

With the last updates to git, the only step required to install is to add that `_git-lfs` file to ZSH's `fpath`.

I can continue to expand/finish this handwritten version, or I could start on the automatic generation of the script like this PR had - but I'm not familiar with Go so it might take be a bit longer.

@sinbad are you still interested in working on this issue? If so, I'd be willing to help out with whatever ZSH-specific issues you have.

commands/command_integrate.go
robobenklein: If you're shipping the completions in a debian package, I think it should be placed into `/usr/share/zsh/vendor-completions/`. (tested on Ubuntu)

ttaylorr:
> I have working ZSH completion for `git-lfs` and `git lfs` now.
> 
> https://github.com/robobenklein/configs/blob/007dcd1f4eb9a4cb42ba7558d161d9b2d6bd89c9/zsh/fpath/_git-lfs

Hi, when I attempt to install this, I get:

```ShellSession
$ echo $0
-zsh
$ pbpaste > _git-lfs.zsh
$ source _git-lfs.zsh
_arguments:comparguments:325: can only be called from completion function
```

But I am still hesitant to ship a ZSH-only solution, when built-in Git support is coming in a future release (as noted above.)

robobenklein:
Ah, there's not really the `install` step like there was with the previous solution.

The _git-lfs completion file needs to placed in the `fpath` which can be done when packaging a deb for example.

On Ubuntu, it would look like this: `/usr/share/zsh/functions/Completion/Unix/_git-lfs`, next to the git completion `_git` in the same folder.

PR #1134, opened 2016-04-08
SSH should be called more strictly with command as one argument

### LINKED ISSUE #1132: Custom GIT_SSH client not invoked correctly
When Git-LFS is invoking a custom `GIT_SSH` client, it passes the command as *separate* parameters, like `'git-lfs-authenticate', 'account/priv.git', 'upload'` however according to the [specification](https://git-scm.com/docs/git), the command must be passed in as exactly one parameter, like `'git-lfs-authenticate account/priv.git upload'`.

sinbad:
Fixes #1132

### DIFF
--- a/lfs/ssh.go
+++ b/lfs/ssh.go
@@ -3,6 +3,7 @@ package lfs
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


technoweenie:
Thanks for fixing this :)

PR #1135, opened 2016-04-08
You should add the .gitattributes file before committing

jonmagic:
While following the instructions in the README made my first commit and then realized I hadn't added the `.gitattributes` file. I think the instructions should be explicit about this step.


technoweenie:
I'd have put it on the same line :)

jonmagic:
> I'd have put it on the same line :)

I did that and then changed it to a two liner, but I can change it back :+1:

PR #1141, opened 2016-04-13
Update README.md

nirgeier:
Improve README file


technoweenie:
Thanks for the PR, but I'm not sure I want markdown link references for links that are only used once. Are you reading this from the command line? If link refs help you lot, I think I'd rather see them below the paragraphs that use them.

Issue #1120, opened 2016-04-03
Windows installer issues


technoweenie:
- When `c:\Program Files\Git LFS` (or anything not in the `$PATH`), either warn or add it to the `$PATH`.
- Mention when `git lfs install` fails to run, especially due to a `$PATH` issue.
- Rename uninst filename to something like `git-uninstall.exe`.

strich:
The installer does already add `C:\Program Files\Git LFS` to `$PATH`. There is however no error reporting or recovery, so if there is something going wrong we won't know about it. I'll look into ensuring that all steps are confirmed and errors caught and either dealt with or reported to the user.

technoweenie:
I wonder if there's something that needs to be refreshed between when it adds to the `$PATH` and when `git lfs install` runs? I restarted Git Bash, and now `/c/Program Files/Git LFS:/bin` is in there. Installing and uninstalling works fine with that. 

I'm pretty sure the installer added it to `$PATH` because I have no idea how to change it on Windows :)

dg9ngf:
Why does this extension to Git even need its own installation directory? Can't the git-lfs.exe binary just be placed somewhere git.exe will find it like every other Git command, too? %PATH% is growing and growing...

strich:
@technoweenie a console window does need to be refreshed after `$PATH` changes. However we use a new console session to perform the `git lfs install` so that cannot be the problem.

However this discussion has led me to think there might be an issue with the setup not performing the steps in the order I expected them to. This could be the root cause. I'll check it out in a day or so and report back.

@dg9ngf the reason we chose to give Git LFS its own installation directory is because there can be many forms of Git in windows with their own ways of installing themselves, and it is really difficult to figure out where to put git-lfs.exe. Furthermore, it created a dependency on another project - If, for example, Git for Windows changed the way it installs itself it could instantly break the Git LFS installer. It just seemed cleaner to give Git LFS its own installation directory, and there are no drawbacks to doing so.

sinbad:
> and there are no drawbacks to doing so

Unfortunately there are: primarily the limited `PATH` length on Windows. Anything over 2047 characters gets a bit dodgy, Win 7 can extend that to 4095 but requires a restart before it works (after you go over the old limit). The results of exceeding this length are randomly failing to find things on the PATH all of a sudden; in SourceTree for Windows we deliberately keep PATH additions to a minimum because of this, some developers have a lot of stuff on their PATH especially if they work in a lot of environments or forget to clean up.

[...]

PR #1143, opened 2016-04-13
Fixed an issue that caused the Win installer to fail to init Git LFS

strich:
* Fixed an issue (#1120) where the newly added Git LFS `PATH` location wasn't refreshed in time for use later in the installer script and silently failed to initialize Git LFS.
* If for any reason there is an issue running `git lfs install`, the installer will display a message box asking the user to manually do so.
* Added a `GIT_LFS_PATH` global env var for later use.


technoweenie:
Worked for me :metal:

PR #1144, opened 2016-04-14
fix netrc matching when host has no port

technoweenie:
Whew, glad I QA'd against a real server, and not just a local test one :)

The `net.SplitHostPort` function expects the host string to have a port (like `localhost:8008`). Currently, it blows up if you use a host on a standard http port.


sinbad:
👍

PR #1146, opened 2016-04-14
Change to the git-lfs meta package for Chocolatey

pascalberger:


technoweenie:
Thanks for the patch. I'm going to take this opportunity to finally install chocolatey and check this out.

PR #1147, opened 2016-04-14
Removed wrong html tag

pabloguerrero:


technoweenie:
Hi @pabloguerrero, this `<br>` tag is necessary for the `ronn` to `man` conversion. I assume you submitted this because it shows up when you run `git lfs fetch -h` or `git lfs help fetch` though. Is that correct?

If so, I think the code generation code should strip that out, instead of removing it here.

technoweenie:
Turns out the [help generator replaces `<br>`, not `<br/>`. I fixed it in 7472995bcf0900c79ec6ae2d1fe86610b599744b.

Thanks for the report!

pabloguerrero:
Hi @technoweenie - sorry for the belated reply. Yes, indeed I suggested this PR because it showed up in the command-line help. Glad that it helped anyways!

PR #1149, opened 2016-04-15
Fixed #719 missing /usr/share/man/man5/git-lfs-config.5.gz .

### LINKED ISSUE #719: Debian package 1.0.0 for jessie 64bit is broken

javabrett:
I was able to test this for a deb on Ubuntu/LinuxMint, and it worked, `man git-lfs-config` now works after local build and deb install.

Warning: RPM untested.


technoweenie:
😍 So, what would be my best course of action regarding v1.2? Should I just re-upload the debian flavor packages? I don't think this warrants a v1.2.1 release.

javabrett:
This does not appear severe enough to me to justify either re-upload or a new version IMO.  Fixed in 1.3.  Reuploading sets a bad precedent for mutable release artifacts with the same version number.

Easy enough read from [git-lfs-config.5.ronn](https://github.com/github/git-lfs/blob/master/docs/man/git-lfs-config.5.ronn) for now.  Maybe nice if one day HTML-rendered man-pages can be published somewhere.

technoweenie:
> Maybe nice if one day HTML-rendered man-pages can be published somewhere.

I want to get this on git-lfs.github.com for v1.3.

Thanks for the patch!

PR #1154, opened 2016-04-18
add homebrew update to release process

larsxschneider:
As homebrew is pretty popular amongst Mac users I think it would be nice to bump the version with every update. However, I also understand that you guys can't update for every package manager out there... therefore this is just a suggestion and I am not sad if you reject it 😄


technoweenie:
Thanks, adding this to the list is a great idea. I normally do this, but totally spaced on it.

PR #1158, opened 2016-04-18
More roadmap updates!

technoweenie:
This is focused on build/compilation stuff. Adding these items now so I don't forget later.


sinbad:
:metal:

PR #1160, opened 2016-04-19
Pass `git lfs clone` flags through to `git clone` correctly, respect some options

### LINKED ISSUE #1155: `git lfs clone` does not pass through git clone options?
Hi,

it looks like as if `git lfs clone` does not pass through git clone options properly. Or maybe I am doing something wrong? Consider this:

```
Ξ ~tmp → git lfs version
git-lfs/1.2.0 (GitHub; darwin amd64; go 1.6.1)

Ξ ~tmp → git lfs clone --verbose https://repo.url
git lfs clone [git clone options] <repository> [<directory>]

Clone an LFS enabled Git repository more efficiently by disabling LFS during the
git clone, then performing a 'git lfs pull' directly afterwards.

This is faster than a regular 'git clone' because that will download LFS content
using the smudge filter, which is executed individually per file in the working
copy. This is relatively inefficient compared to the batch mode and parallel
downloads performed by 'git lfs pull'.

Options:

All options supported by 'git clone'

Error: unknown flag: --verbose

Ξ ~tmp → git lfs clone -v https://repo.url
git lfs clone [git clone options] <repository> [<directory>]

Clone an LFS enabled Git repository more efficiently by disabling LFS during the
git clone, then performing a 'git lfs pull' directly afterwards.

This is faster than a regular 'git clone' because that will download LFS content
using the smudge filter, which is executed individually per file in the working
copy. This is relatively inefficient compared to the batch mode and parallel
downloads performed by 'git lfs pull'.

Options:

All options supported by 'git clone'

Error: unknown shorthand flag: 'v' in -v
``` 

Thanks,
Lars

/cc @sinbad

sinbad:
Fixes #1155 

- `git lfs clone` now correctly passes through flags to `git clone`
- `--no-checkout` and `--bare` now cause `git lfs clone` to `git lfs fetch` rather than pull
- `--origin` option is respected and changes how lfs fetches/pulls after clone

### DIFF
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -64,8 +64,8 @@ func cloneCommand(cmd *cobra.Command, args []string) {
 	// Now just call pull with default args
 	lfs.Config.CurrentRemote = "origin" // always origin after clone
 
-	if cloneFlags.NoCheckout {
-		// If --no-checkout then we shouldn't check out, just fetch instead
+	if cloneFlags.NoCheckout || cloneFlags.Bare {
+		// If --no-checkout or --bare then we shouldn't check out, just fetch instead
 		fetchRef("HEAD", nil, nil)
 	} else {
 		pull(nil, nil)
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -63,7 +63,13 @@ func cloneCommand(cmd *cobra.Command, args []string) {
 
 	// Now just call pull with default args
 	lfs.Config.CurrentRemote = "origin" // always origin after clone
-	pull(nil, nil)
+
+	if cloneFlags.NoCheckout {
+		// If --no-checkout then we shouldn't check out, just fetch instead
+		fetchRef("HEAD", nil, nil)
+	} else {
+		pull(nil, nil)
+	}
 
 }
 
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -16,12 +16,14 @@ var (
 		Use: "clone",
 		Run: cloneCommand,
 	}
+
+	cloneFlags git.CloneFlags
 )
 
 func cloneCommand(cmd *cobra.Command, args []string) {
 
 	// We pass all args to git clone
-	err := git.CloneWithoutFilters(args)
+	err := git.CloneWithoutFilters(cloneFlags, args)
 	if err != nil {
 		Exit("Error(s) during clone:\n%v", err)
 	}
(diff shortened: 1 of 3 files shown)


sinbad:
I foolishly thought just passing through arguments was enough, forgetting that cobra would reject any flags it didn't understand. Couldn't find a way to respect them without re-stating them all.

commands/command_clone.go
```diff
@@ -60,11 +62,47 @@ func cloneCommand(cmd *cobra.Command, args []string) {
...
+	cloneCmd.Flags().StringVarP(&cloneFlags.SeparateGit, "separate-git-dir", "", "", "See 'git clone --help'")
+	cloneCmd.Flags().StringVarP(&cloneFlags.Depth, "depth", "", "", "See 'git clone --help'")
+	cloneCmd.Flags().BoolVarP(&cloneFlags.Recursive, "recursive", "", false, "See 'git clone --help'")
+	cloneCmd.Flags().BoolVarP(&cloneFlags.RecurseSubmodules, "recurse-submodules", "", false, "See 'git clone --help'")
+	cloneCmd.Flags().StringVarP(&cloneFlags.Config, "config", "c", "", "See 'git clone --help'")
+	cloneCmd.Flags().BoolVarP(&cloneFlags.SingleBranch, "single-branch", "", false, "See 'git clone --help'")
+	cloneCmd.Flags().BoolVarP(&cloneFlags.NoSingleBranch, "no-single-branch", "", false, "See 'git clone --help'")
+	cloneCmd.Flags().BoolVarP(&cloneFlags.Verbose, "verbose", "", false, "See 'git clone --help'")
+	cloneCmd.Flags().BoolVarP(&cloneFlags.Ipv4, "ipv4", "", false, "See 'git clone --help'")
+	cloneCmd.Flags().BoolVarP(&cloneFlags.Ipv6, "ipv6", "", false, "See 'git clone --help'")
```
larsxschneider: Uh, I guess there is no other way with cobra ... that screams maintenance hell, though 😢  ... would it make sense to read the flags at run time (parsing `git clone`)? But then we don't know the type...

@spf13 do you happen to see a way to pass through flags with cobra?

sinbad: I looked through the cobra code and there’s no way to not error on undefined flags right now. I thought about going deeper & adding the option to cobra myself but it also uses pflag to do the parsing so it would probably be a doubly nested upstream change.

The alternative is to not use cobra for the clone command but that has maintenance issues of its own, this seemed the lesser evil given that git clone args don’t change much.

technoweenie:
> The alternative is to not use cobra for the clone command but that has maintenance issues of its own, this seemed the lesser evil given that git clone args don’t change much.

I agree with this. It's unfortunate, and possibly worth fixing in cobra. But I don't think we should hold up a v1.2.1 release to do that.

spf13:
@technoweenie @sinbad @larsxschneider appologize for not being more responsive. I was travelling with limited access. 

Cobra does use pflag, but it's our own fork of pflag so while technically it's doubly upstream, it's really just part of Cobra. 

I'm happy to help further. I'm not sure I yet totally understand what you are trying to do, but in reviewing the code above it seems like a simpler way would be possible. 

Do either of these help?:
https://godoc.org/github.com/spf13/pflag#ErrorHandling
If you change the value to `ContinueOnError` it may be all you need. I'm not sure if it skips unknown flags or not. 

The usage variable is a function that is called on error and may be customizable to what you need.
https://godoc.org/github.com/spf13/pflag#pkg-variables

sinbad:
Thanks @spf13 - I did look at `ContinueOnError` but it seems that Cobra already uses that in all uses of `pflag` in order to fully parse everything even in the case of intermediate errors. The issue is then this code in `execute()`:
```
	err = c.ParseFlags(a)
	if err != nil {
		return err
	}
```
So although at that stage it has used `pflag.ContinueOnError` in order not to stop at the first error, the presence of any unknown flags will cause an early exit from execute with an error. I don't think the usage method can prevent that, unless I'm missing something?

spf13:
So pflag.ContinueOnError do ignore parsing errors, but not unknown flags
errors.

It seems like another error value should be defined with this additional
behavior (ignore unknown flags) and then respected in pflag.

If any of you want to take a stab at it I'd gladly merge and provide help.

While I believe this will fix your immediate issue, I still feel like
there's may be a better overall approach that could be taken that would
have less maintenance. Admittedly I'm not quite clear on what the overall
goal is.

sinbad:
Thanks. It's an unusual case because we want to call a child process within our own and just pass through all the flags we've received without necessarily parsing all of them, letting the child process parse & validate everything else, because we don't control the arguments of the child process.

I will come back to this another time since I'm relatively comfortable with how static the arguments to `git clone` are; it's in `git`'s own interests not to make breaking changes to them (and since starting to use `git` in 2009 I can't think of much that's changed in that time).

spf13:
There isn't yet, but there should be an option in Cobra to skip flag
parsing and treat it all as args which you can then pass along. That should
make it very easy to wrap things as you are doing.

It shouldn't be the default, and should be configurable on a command. I
think it should be easy to add unless I'm not considering all cases.

Feel free to contribute it or discuss further if you think it will help.

PR #1173, opened 2016-04-22
Use proxy from git config if available

jonmagic:
This may be a possible solution to https://github.com/github/git-lfs/issues/1125 but we are Go and git-lfs newbs so we could use some guidance.

* Is this approach even close to how y'all would do this?
* Should we create a new test file for lfs/http.go so that we can test this method specifically or should we add an integration test, or something else entirely?

/cc @LizzHale

### DIFF
--- a/lfs/http.go
+++ b/lfs/http.go
@@ -141,12 +141,18 @@ func (c *Configuration) HttpClient(host string) *HttpClient {
 }
 
 func proxyFromGitConfigOrEnvironment(req *http.Request) (*url.URL, error) {
+	proxyURL, err := http.ProxyFromEnvironment(req)
+
+	if proxyURL != nil {
+		return proxyURL, err
+	}
+
 	if proxy, ok := Config.GitConfig("http.proxy"); ok {
 		proxyURL, err := url.Parse(proxy)
 		return proxyURL, err
 	}
 
-	return http.ProxyFromEnvironment(req)
+	return proxyURL, err
 }
 
 func checkRedirect(req *http.Request, via []*http.Request) error {
--- a/lfs/http.go
+++ b/lfs/http.go
@@ -10,6 +10,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/httputil"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -115,7 +116,7 @@ func (c *Configuration) HttpClient(host string) *HttpClient {
 	tlstime := c.GitConfigInt("lfs.tlstimeout", 30)
 
 	tr := &http.Transport{
-		Proxy: http.ProxyFromEnvironment,
+		Proxy: proxyFromGitConfigOrEnvironment,
 		Dial: (&net.Dialer{
 			Timeout:   time.Duration(dialtime) * time.Second,
 			KeepAlive: time.Duration(keepalivetime) * time.Second,
@@ -137,6 +138,15 @@ func (c *Configuration) HttpClient(host string) *HttpClient {
 	c.httpClients[host] = client
 
 	return client
+}
+
+func proxyFromGitConfigOrEnvironment(req *http.Request) (*url.URL, error) {
+	if proxy, ok := Config.GitConfig("http.proxy"); ok {
+		proxyURL, err := url.Parse(proxy)
+		return proxyURL, err
+	}
+
+	return http.ProxyFromEnvironment(req)
 }
 
 func checkRedirect(req *http.Request, via []*http.Request) error {
--- a/lfs/proxy_test.go
+++ b/lfs/proxy_test.go
@@ -1,0 +1,71 @@
+package lfs
+
+import (
+	"net/http"
+	"net/url"
+	// "os"
+	"testing"
+
+	"github.com/github/git-lfs/vendor/_nuts/github.com/technoweenie/assert"
+)
+
+// --- FAIL: TestProxyFromEnv (0.00s)
+// panic: runtime error: invalid memory address or nil pointer dereference [recovered]
+// panic: runtime error: invalid memory address or nil pointer dereference
+//
+// func TestProxyFromEnv(t *testing.T) {
+// 	os.Setenv("HTTPS_PROXY", "https://proxy-from-env:8080")
+//
+// 	u, err := url.Parse("http://some-host.com:123/foo/bar")
+// 	if err != nil {
+// 		t.Fatal(err)
+// 	}
+// 	req := &http.Request{
+// 		URL:    u,
+// 		Header: http.Header{},
+// 	}
+// 	proxyURL, err := proxyFromGitConfigOrEnvironment(req)
+//
+// 	assert.Equal(t, "proxy-from-env:8080", proxyURL.Host)
+// 	assert.Equal(t, nil, err)
+// }
+
+func TestProxyFromGitConfig(t *testing.T) {
+	oldGitConfig := Config.gitConfig
+	defer func() {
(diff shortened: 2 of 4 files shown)


technoweenie:
This looks great :) My main question is whether this should prefer the git config or the environment first? I'd lean towards using the git config as a last resort. A couple use cases:

1. Someone has a global `http_proxy` set in their `bashrc` because of their local network setup.
2. Someone runs `http_proxy=.... git clone ...`

I'd like to see how git handles this.

A go test for `proxyFromGitConfigOrEnvironment()` could go in `lfs/proxy_test.go`. Seems like that'd be enough to test the various use cases. 

An integration test would be lovely. It'd require setting up a fake proxy server for testing purposes. I can look into this separately, since it should've existed for the current env proxy support.

sinbad:
+1 to @technoweenie's comments. Environment vars should always override gitconfig settings, it's how you typically change things temporarily rather than persistently.

jonmagic:
@LizzHale and I paired on this again today and we think we've got the logic right but haven't figured out how to correctly test all of them. One of our tests works and two don't, we left them there but commented out with the error. Any *pointers are appreciated :grin:

lfs/http.go
```diff
@@ -139,6 +140,21 @@ func (c *Configuration) HttpClient(host string) *HttpClient {
 	return client
 }
 
+func proxyFromGitConfigOrEnvironment(req *http.Request) (*url.URL, error) {
+	proxyURL, err := http.ProxyFromEnvironment(req)
+
+	if proxyURL != nil {
+		return proxyURL, err
```
ttaylorr: Looks like we return the environment's proxy before checking if there's one in the user's `.gitconfig`. Since this method is called `proxyFromGitConfigOrEnvironment`, the name makes me think that it'll return the `.gitconfig`'s proxy _over_ the environment's one.

Either is fine, but I think it may be beneficial to update the method name if the implementation prioritizes environment configuration over `.gitconfig.`

ttaylorr:
Hi @jonmagic and @LizzHale! This looks awesome 👍 I definitely agree with @technoweenie and @sinbad's comment about the environment vs. `.gitconfig` debate, and your patch looks sweet.

Tests look like they're off to a good start, too. If you post more of the stacktrace, I'd be more than happy to take a look at it with you and try and figure out the cause of the failure 😄

lfs/proxy_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"testing"
+
+	"github.com/github/git-lfs/vendor/_nuts/github.com/technoweenie/assert"
+)
+
+// --- FAIL: TestProxyFromEnv (0.00s)
+// panic: runtime error: invalid memory address or nil pointer dereference [recovered]
+// panic: runtime error: invalid memory address or nil pointer dereference
+//
+// func TestProxyFromEnv(t *testing.T) {
```
jonmagic: Stacktrace:

```
warning: current Git remote contains credentials
warning: current Git remote contains credentials
--- FAIL: TestProxyFromEnv (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal 0xb code=0x1 addr=0x0 pc=0xe1a2a]

goroutine 287 [running]:
panic(0x4cef80, 0xc82000a0f0)
	/usr/local/Cellar/go/1.6.2/libexec/src/runtime/panic.go:481 +0x3e6
testing.tRunner.func1(0xc820366000)
	/usr/local/Cellar/go/1.6.2/libexec/src/testing/testing.go:467 +0x192
panic(0x4cef80, 0xc82000a0f0)
	/usr/local/Cellar/go/1.6.2/libexec/src/runtime/panic.go:443 +0x4e9
github.com/github/git-lfs/lfs.TestProxyFromEnv(0xc820366000)
	/Users/jonmagic/gopath/src/github.com/github/git-lfs/lfs/proxy_test.go:29 +0x34a
testing.tRunner(0xc820366000, 0x7d4368)
	/usr/local/Cellar/go/1.6.2/libexec/src/testing/testing.go:473 +0x98
created by testing.RunTests
	/usr/local/Cellar/go/1.6.2/libexec/src/testing/testing.go:582 +0x892
FAIL	github.com/github/git-lfs/lfs	1.285s
ok  	github.com/github/git-lfs/git	1.636s
```

lfs/proxy_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	assert.Equal(t, "proxy-from-git-config:8080", proxyURL.Host)
+	assert.Equal(t, nil, err)
+}
+
+// --- FAIL: TestProxyIsNil (0.00s)
+// panic: runtime error: invalid memory address or nil pointer dereference [recovered]
+// panic: runtime error: invalid memory address or nil pointer dereference
+//
+// func TestProxyIsNil(t *testing.T) {
```
jonmagic: Stacktrace:

```
warning: current Git remote contains credentials
warning: current Git remote contains credentials
--- FAIL: TestProxyIsNil (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal 0xb code=0x1 addr=0x0 pc=0xe1e32]

goroutine 307 [running]:
panic(0x4ceee0, 0xc82000a0f0)
	/usr/local/Cellar/go/1.6.2/libexec/src/runtime/panic.go:481 +0x3e6
testing.tRunner.func1(0xc820144090)
	/usr/local/Cellar/go/1.6.2/libexec/src/testing/testing.go:467 +0x192
panic(0x4ceee0, 0xc82000a0f0)
	/usr/local/Cellar/go/1.6.2/libexec/src/runtime/panic.go:443 +0x4e9
github.com/github/git-lfs/lfs.TestProxyIsNil(0xc820144090)
	/Users/jonmagic/gopath/src/github.com/github/git-lfs/lfs/proxy_test.go:69 +0x2b2
testing.tRunner(0xc820144090, 0x7d4380)
	/usr/local/Cellar/go/1.6.2/libexec/src/testing/testing.go:473 +0x98
created by testing.RunTests
	/usr/local/Cellar/go/1.6.2/libexec/src/testing/testing.go:582 +0x892
FAIL	github.com/github/git-lfs/lfs	1.251s
ok  	github.com/github/git-lfs/git	1.685s
```

lfs/proxy_test.go
technoweenie: So the relevant bit here is this:

```
github.com/github/git-lfs/lfs.TestProxyFromEnv(0xc820366000)
    /Users/jonmagic/gopath/src/github.com/github/git-lfs/lfs/proxy_test.go:29 +0x34a
```

This is telling me that the `TestProxyFromEnv()` function on `proxy_test.go:29` is failing due to a nil pointer reference. Going by the commented test line numbers, that's coming from the middle line here, because `proxyURL` is nil.

```go
proxyURL, err := proxyFromGitConfigOrEnvironment(req)
assert.Equal(t, "proxy-from-env:8080", proxyURL.Host)
assert.Equal(t, nil, err)
```

You should check `err` before the value to avoid issues like this.

lfs/proxy_test.go
technoweenie: This is similar to the other stack trace. However, I think you want to assert that `proxyURL` is nil, not `proxyURL.Host`. In Go, only pointers and interfaces can be nil. Built-in types like integers or strings are initialized to a default value, instead of nil.

```go
var foo string // default value is ""
var age int // default is 0
```

The `proxyFromGitConfigOrEnvironment()` returns a `*url.URL`, which can be nil since it's a pointer to a `url.URL` struct.

jonmagic:
Closing in favor of #1358

PR #1176, opened 2016-04-25
Added some building-on-RHEL prerequisites.

javabrett:


technoweenie:
This looks good. Should it mention that the dependencies are needed for building Go and the LFS man pages? It should be possible to build LFS with only the Go binary for your platform. Not that many linux user would want to though...

javabrett:
Unfortunately, this is not an attempt at complete coverage of the prerequisite packages - it's only the result of my most-recent attempt to build on an RPM-based distro, starting from a fairly skinny VM image that lacked things like `gcc`.

I did not strike issues with man-generation.  `shasum` (so not `sha1sum`, depends on a specific Perl package) and the need for `ruby-devel` (required activation of a latent repo config, failed with an obtuse Ruby can't-build-native error) were somewhat obtuse so I thought worth writing them down.  I will need to check my APT notes, but from memory that was more of a getting go-lang installed and working experience.

javabrett:
Looking at my notes and history for an APT-based system build/deb-build, I see these:

```
$ sudo apt-get install ruby-ronn
$ sudo apt-get install dh-golang
$ sudo apt-get install libxml2-dev libxslt1-dev python-dev
$ sudo apt-get install ruby1.9.1-dev
$ sudo apt-get install libxcb-xtest0
```

If any of these look correct or familiar, would it be worth adding them to the build doc on-suspicion?

andyneff:
LGTM :+1:

PR #1178, opened 2016-04-25
Fix failure to return non-zero exit code when lfs install/update fails to install hooks

### LINKED ISSUE #1174: `git lfs install` returns zero exit code even when it doesn't install hooks
Reproduction:

```
> rm .git/lfs/pre-push
> touch .git/hooks/pre-push
> git lfs install
Hook already exists: pre-push



Run `git lfs update --force` to overwrite this hook.
Git LFS initialized.
> echo $?
0
```
This should return a non-zero exit code since currently this repository is not correctly initialised.

This also applies to `git lfs update`.

sinbad:
Fixes #1174 

Also tidies up the output a little and automatically replaces blank pre-push hooks now.

### DIFF
--- a/commands/command_update.go
+++ b/commands/command_update.go
@@ -24,7 +24,7 @@ func updateCommand(cmd *cobra.Command, args []string) {
 
 	if err := lfs.InstallHooks(updateForce); err != nil {
 		Error(err.Error())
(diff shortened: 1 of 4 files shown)


technoweenie:
:metal:

PR #1182, opened 2016-04-26
Add `git lfs update --manual` option & promote it on hook install fail

sinbad:
Related to #1174 

As discussed in chat, implemented as a command option rather than in docs so that we don't have to maintain the hook script in more than one location.


technoweenie:
🚢

PR #1193, opened 2016-05-02
Applied same -ldflags -X name value -> name=value fix as in a62e510f.

javabrett:
Squashes script/run warning:
link: warning: option -X github.com/github/git-lfs/lfs.GitCommit 54c5efa may not work in future releases; use -X github.com/github/git-lfs/lfs.GitCommit=54c5efa

### DIFF
--- a/script/run
+++ b/script/run
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 script/fmt
 commit=`git rev-parse --short HEAD`
-GO15VENDOREXPERIMENT=0 go run -ldflags="-X github.com/github/git-lfs/lfs.GitCommit $commit" ./git-lfs.go "$@"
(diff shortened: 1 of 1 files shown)


technoweenie:
IIRC, this breaks on older versions of Go. However, I see that [Go 1.7 is nearing feature freeze](https://groups.google.com/forum/#!topic/golang-dev/lcvpM-vAoE8), so we may want to lock LFS to Go 1.6+.

javabrett:
Right, I think this change was made in the main build already for the Go 1.5 change, a62e510f .  So if it doesn't work for < 1.5 it should show-up already?  I confess I am using 1.6.

I think moving to at least Go >= 1.5, or 1.6 if the vendor imports are reworked, has been discussed in another issue.

technoweenie:
Turns out I already had this conversation, and decided supporting whatever old Go version wasn't worth it. https://github.com/github/git-lfs/commit/a62e510fff9f75be823d153e0228702f91c0a13e

I just checked in Go 1.3-1.6. It _works_ on all of them, but only prints the commit in Go 1.5+.

Issue #1156, opened 2016-04-18
v1.2.0 does not build on Debian 7 or 8


tpdownes:
After downloading source and running `dpkg-buildpackage -us -uc` on both Deb7 and Deb8. Deb8 error trace is below.

```
=== RUN TestExistingUpload
--- FAIL: TestExistingUpload (0.00 seconds)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
    panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x617447]

goroutine 183 [running]:
runtime.panic(0x7fc0a0, 0xc767d3)
    /usr/lib/go/src/pkg/runtime/panic.c:279 +0xf5
testing.func·006()
    /usr/lib/go/src/pkg/testing/testing.go:416 +0x176
runtime.panic(0x7fc0a0, 0xc767d3)
    /usr/lib/go/src/pkg/runtime/panic.c:248 +0x18d
github.com/github/git-lfs/localstorage.localObjectDir(0x0, 0x8c4190, 0x28, 0x0, 0x0)
    /home/thomas.downes/build/git-lfs-1.2.0/obj-x86_64-linux-gnu/src/github.com/github/git-lfs/localstorage/localstorage.go:57 +0x127
github.com/github/git-lfs/localstorage.(*LocalStorage).BuildObjectPath(0x0, 0x8c4190, 0x28, 0x0, 0x0, 0x0, 0x0)
    /home/thomas.downes/build/git-lfs-1.2.0/obj-x86_64-linux-gnu/src/github.com/github/git-lfs/localstorage/localstorage.go:48 +0x81
github.com/github/git-lfs/lfs.LocalMediaPath(0x8c4190, 0x28, 0x0, 0x0, 0x0, 0x0)
    /home/thomas.downes/build/git-lfs-1.2.0/obj-x86_64-linux-gnu/src/github.com/github/git-lfs/lfs/lfs.go:57 +0x62
github.com/github/git-lfs/lfs.TestExistingUpload(0xc20806ea20)
    /home/thomas.downes/build/git-lfs-1.2.0/obj-x86_64-linux-gnu/src/github.com/github/git-lfs/lfs/upload_test.go:109 +0x3f7
testing.tRunner(0xc20806ea20, 0xc7b200)
    /usr/lib/go/src/pkg/testing/testing.go:422 +0x8b
created by testing.RunTests
[...]

PR #1194, opened 2016-05-03
Sets -ldflags -X ... GitCommit in deb build for short-hash in version…

### LINKED ISSUE #1192: git --short commit-hash lost from version/UserAgent in 1.2.0 deb packages
- 1.1.2/packagecloud/deb `git version` : `git-lfs/1.1.2 (GitHub; linux amd64; go 1.5.3; git bf6a6c9)`
- 1.2.0/packagecloud/deb `git version` : `git-lfs/1.2.0 (GitHub; linux amd64; go 1.5.3)`
- master-54c5efa/localbuild/bin `git version` : `git-lfs/1.2.0 (GitHub; linux amd64; go 1.6; git 54c5efa)`
- master-54c5efa/localbuild/extracted-from-deb `git version` : `git-lfs/1.2.0 (GitHub; linux amd64; go 1.6)`

So this problem is in the package build and is reproducible in sandbox.

It looks like when we build from source in deb, we aren't correctly setting `-ldflags` to `-X github.com/github/git-lfs/lfs.GitCommit=<shortsha>`.  This is likely because the deb is built from source from a non-git-working-directory, which is correct, but of course does not have access to the output of `exec.Command("git", "rev-parse", "--short", "HEAD").Output()` (ref [script/build.go](https://github.com/github/git-lfs/blob/master/script/build.go#L47)).

Not sure how this regressed in 1.2.0 - possibly the deb build was improved to build from disconnected sources.

The fix is probably to inject this into the build via a new env var in [build.go](https://github.com/github/git-lfs/blob/master/script/build.go#L156) and hope that this can be passed-through to `dh_lang` helper when it launches its build.

javabrett:
… string. Fixed #1192.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -18,6 +18,8 @@ export PATH := $(CURDIR)/$(BUILD_DIR)/bin:$(PATH)
 # by-default, dh_golang only copies *.go and other source - this upsets a bunch of vendor test routines
 export DH_GOLANG_INSTALL_ALL := 1
 
+COMMIT=$(shell git rev-parse --short HEAD)
+
 %:
 	dh $@ --buildsystem=golang --with=golang
 
@@ -27,7 +29,7 @@ override_dh_clean:
 	dh_clean
 
 override_dh_auto_build:
-	dh_auto_build
+	dh_auto_build -- -ldflags="-X github.com/github/git-lfs/lfs.GitCommit=$(COMMIT)"
 	#dh_golang doesn't do anything here in deb 8, and it's needed in both
 	if [ "$(DEB_HOST_GNU_TYPE)" != "$(DEB_BUILD_GNU_TYPE)" ]; then\
 		cp -rf $(BUILD_DIR)/bin/*/* $(BUILD_DIR)/bin/; \


technoweenie:
This looks good to me. Can I get a sanity check from @ssgelm or @andyneff ?

ssgelm:
I don't love this change for similar reasons as were discussed in #1156.  The deb should be possible to build from the source tarball without it needing to be a git repo.  Is there a better way we can capture the right ref without needing to be in a git repo when building the package?

technoweenie:
The ref isn't required. So if someone is building from just a directory of files, it should just work.

> Is there a better way we can capture the right ref without needing to be in a git repo when building the package?

Would it be worth preparing a source tarball just for this purpose? 

* Teach this script to check for a local file like `./VERSION` or something
* Include the actual man pages so ruby isn't required to build.

ssgelm:
I think a source tarball sounds like a good idea.  But as long as this still builds successfully without being in a git repo I think it's fine.

javabrett:
I'm not currently able to get the source-tarball from `master` to `dpkg-buildpackage -us -uc`, and it fails identically with this change.  I could raise an issue for that, unless there's an existing issue or list of prerequisites for it to work.

javabrett:
In order to build-from-tarball, would we need to replace the `ldflags` approach with some sort of `go:generate`, so that the source-code file containing the version-string with short-hash is an actual source-file in the source tarball, rather than being injected by the linker?  Guessing, this is a bit beyond my Go ...

ttaylorr:
`github.com/github/git-lfs/lfs.GitCommit` is no longer the correct identifier (the `github` organization is now `git-lfs`, instead). @javabrett if you get a chance to look at this and update it, I would be happy to merge it 😄

PR #1198, opened 2016-05-04
Fix installer error on win32.

teo-tsirpanis:
On 32-bit versions of Windows (in my case, Windows XP), the windows installer fails because it cannot expand the "pf64"constant.

By changing this very constant to "pf", it gets resolved properly on all versions of Windows.

See http://www.jrsoftware.org/ishelp/index.php?topic=consts for further explanation of the constants.


sinbad:
Thanks for catching that!

PR #1200, opened 2016-05-04
Fix bug in Windows installer under Win32.

teo-tsirpanis:
In a previous pull request I submitted, there was an error.
The installer would write to a "mingw64" directory even on 32-bit versions of Windows.
With this patch, the installer will write to the correct "mingw32" directory under Win32.

### DIFF
--- a/script/windows-installer/inno-setup-git-lfs-installer.iss
+++ b/script/windows-installer/inno-setup-git-lfs-installer.iss
@@ -68,7 +68,11 @@ begin
   if LoadStringFromFile(TmpFileName, ExecStdOut) then begin
       if not (Pos('Git\cmd', ExtractFilePath(ExecStdOut)) = 0) then begin
         // Proxy Git path detected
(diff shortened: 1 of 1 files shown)


technoweenie:
/cc @strich for a sanity check.

@teo-tsirpanis thanks for your patches! If you don't mind, I'll probably ping you to try the installer once we get closer to the next release.

strich:
Yep looks good. Oversight on my part with that manually defined Git for Windows dir. Nice one @teo-tsirpanis

technoweenie:
Aaaand now I wish we had CI for the windows installer :) How do you even test that stuff???

I'm going to wait until I can confirm this works before merging.

strich:
Unit tests would be the ideal way to do it. But you'll need all the appropriate platforms, CI and a unit testing framework. Lots of work. :)

technoweenie:
Looks good, and I confirmed it compiles and builds the installer successfully.  @teo-tsirpanis Thanks for the patch!

PR #1212, opened 2016-05-10
Enforce minimum git version check (currently >= 1.8.2) when running a…

### LINKED ISSUE #1210: Fail-fast if minimum version not met (>= 1.8.2), esp important for non-package installs
I just wasted a significant amount of time assisting a colleague; it was not noticed that having installed LFS from a tarball, the minimum Git version (currently git >= 1.8.2) was not being met.  As observed by many others, this is the default situation for old RHEL and variants, where Git 1.7.1 is commonly pre-installed.

`git-lfs` commands should fail-fast if the minimum Git version is not met, lest they fail in interesting and obscure ways, especially when communicating and authenticating with remote servers.

There is already version-specific code, supported by `IsGitVersionAtLeast`, so it should just be a matter of adding a similar call to all commands, or at the bare-minimum, when `install` runs.

javabrett:
…ny command.  Fixed #1210 .

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -31,7 +31,8 @@ var (
 			cmd.Usage()
 		},
 	}
-	ManPages = make(map[string]string, 20)
+	ManPages          = make(map[string]string, 20)
+	MinimumGitVersion = "1.8.2"
 )
 
 // Error prints a formatted message to Stderr.  It also gets printed to the
@@ -96,6 +97,10 @@ func Panic(err error, format string, args ...interface{}) {
 }
 
 func Run() {
+	gitver, _ := git.Config.Version()
+	if !git.Config.IsGitVersionAtLeast(MinimumGitVersion) {
+		Exit(fmt.Sprintf("git version >= %s is required for Git LFS, your version: %s", MinimumGitVersion, gitver))
+	}
 	RootCmd.Execute()
 }
 


technoweenie:
Thanks for the PR! @sinbad: Any thoughts on checking the git version on every command, or just specific ones? It seems fast, but may become an issue when running the `git lfs smudge` command on thousands of files on a Windows machine. At the least, I'd add this check to these commands: `install`, `update`, `env`, `track`, `pre-push`, `pull`, `checkout`.

Unfortunately, this is adding a couple extra `git version` calls to every `git lfs` call. The reason is that [`IsGitVersionAtLeast` also runs `git version`](https://github.com/github/git-lfs/blob/5cc78bcc4368ae20c86213b810919ed70e27dc2d/git/git.go#L336-L345). Instead, the check could use `IsVersionAtLeast()` directly.

sinbad:
I'm amazed anyone is still running git that old :) I'd prefer to keep the checks to a minimum - although `git --version` is a relatively fast operation, on Windows we pay a high cost for every fork so it's better if we don't have to add any we don't have to. At a minimum it would be a good idea to cache the current git version in `git.Version()` rather than always forking (we can assume it's not going to change in the life of the process) if we're starting to make this more common - some specific functions do check the version in other places and it would be best not to call it more than once.

As an even tighter optimisation I could suggest adding the check only to `git lfs install` and `git lfs update` since those are gatekeepers and it's unlikely that people will downgrade. However since it's only once I can live with a universal call if it's cached for other uses later.

technoweenie:
> At a minimum it would be a good idea to cache the current git version in git.Version()

:+1:

> I could suggest adding the check only to `git lfs install` and `git lfs update`

I picked the other commands to guard against cases where people just dump the binary into their PATH and try to push ahead without checking things. I don't think you can get very far without running those. They're also not ever called multiple times in a tight loop like `git lfs smudge`.

sinbad:
Yeah, that's fair - I'm happy so long as we avoid smudge & clean then.

technoweenie:
I'd probably add `clone` and `push` to that list too.

samrocketman:
RedHat 6 machines are pretty wide spread and in use.  The version of git in the RH6/CentOS6 repos is 1.7.9.  I typically [compile git](https://gist.github.com/samrocketman/8671036) in those cases.

technoweenie:
@samrocketman This PR is about communicating to end users what the minimum git version is, not setting a new minimum git version. It's a good idea to let them know ahead of time, and not later when they're wondering why LFS isn't working correctly.

Closing this in favor of #1461, which only runs in the commands specified above, and adds git version caching. Here's sample output, proving that `git version` is only run once:

```bash
$ GIT_TRACE=1 script/run track
16:20:45.908860 git.c:350               trace: built-in: git 'rev-parse' '--short' 'HEAD'
trace git-lfs: run_command: 'git' version
Listing tracked paths
```

PR #1213, opened 2016-05-10
Allow smudge filter to return 0 on download failure

### LINKED ISSUE #1195: git checkout doesn't fail cleanly
TL;DR: When git lfs failes to fetch a file, the entire git command fails, and it doesn't leave the repo in the same state as it started in.

    > git clone https://github.com/jeffeb3/lfs_test.git
    Cloning into 'lfs_test'...
    remote: Counting objects: 29, done.
    remote: Compressing objects: 100% (23/23), done.
    remote: Total 29 (delta 6), reused 24 (delta 4), pack-reused 0
    Unpacking objects: 100% (29/29), done.
    Downloading Image.jpg (4.93 KB)
    Downloading blue.png (1.38 KB)
    Downloading green.png (24.46 KB)

    > git status
    On branch master
    
    nothing to commit, working directory clean

Here, I disconnect my connection to github.com (by pulling out the Ethernet cable)

    > git checkout HEAD^
    Downloading green.png (1.38 KB)
    Error downloading object: green.png           (d6991edfa7ab08200f68151eda1283ffa2839c721c401304acb744c9502b4c63)
    
    Errors logged to /home/jeffe/tmp/test/lfs_test/.git/lfs/objects/logs/20160503T104313.4193124.log
    error: external filter git-lfs smudge -- %f failed 2
    error: external filter git-lfs smudge -- %f failed
    fatal: green.png: smudge filter lfs failed

    > git status
    On branch master
    
    Changes not staged for commit:
      (use "git add/rm <file>..." to update what will be committed)
      (use "git checkout -- <file>..." to discard changes in working directory)
    
            deleted:    green.png

   no changes added to commit (use "git add" and/or "git commit -a")
 
So, first off. Obviously lfs can't get the file. You're not magic.

Secondly though, if the operation fails, then lfs needs to leave it in the beginning state. This is really the bug here.

Thirdly, and this maybe belongs in it's own feature, I'd rather at least have the option of letting checkouts like this succeed, leaving just the pointer around instead of the file that failed to fetch. I definitely do work when I can't get to the server, and I'd like to be able to still do all the funky git stuff (branch, checkout by hash, cherry-pick, rebase, etc) when not connected to the server. I can make lfs.fetchexclude="*" and then do all that work, as a workaround.

    git lfs version
    git-lfs/1.2.0 (GitHub; linux amd64; go 1.5.3)

P.S. I'm assuming the log isn't useful for you, because we know why it failed. Also the steps to reproduce it are above. I'll save it anyway though.

sinbad:
Fixes #1195 

Add option `lfs.skipdownloaderrors` / `GIT_LFS_SKIP_DOWNLOAD_ERRORS` to allow the smudge filter to return success on download errors, so that checkout etc can continue even without access to content.

It's an option because using this means exit codes are not propagated and scripts would be unable to detect problems with clone/checkout when LFS downloads failed. Also users if they weren't paying attention. However it's useful as an opt-in if wanting to switch branches etc when disconnected.

### DIFF
--- a/commands/command_smudge.go
+++ b/commands/command_smudge.go
@@ -63,7 +63,7 @@ func smudgeCommand(cmd *cobra.Command, args []string) {
 	cfg := lfs.Config
 	download := lfs.FilenamePassesIncludeExcludeFilter(filename, cfg.FetchIncludePaths(), cfg.FetchExcludePaths())
 
-	if smudgeSkip || lfs.Config.GetenvBool("GIT_LFS_SKIP_SMUDGE", false) {
+	if smudgeSkip || cfg.GetenvBool("GIT_LFS_SKIP_SMUDGE", false) {
 		download = false
 	}
 
@@ -77,7 +77,9 @@ func smudgeCommand(cmd *cobra.Command, args []string) {
 		// Download declined error is ok to skip if we weren't requesting download
 		if !(lfs.IsDownloadDeclinedError(err) && !download) {
 			LoggedError(err, "Error downloading object: %s (%s)", filename, ptr.Oid)
-			os.Exit(2)
+			if !cfg.SkipDownloadErrors() {
+				os.Exit(2)
+			}
 		}
 	}
 }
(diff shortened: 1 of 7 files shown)


technoweenie:
:metal:

PR #1217, opened 2016-05-12
Add missing config details to `env` command

sinbad:
Adds all Fetch/Prune config details to the `env` command, as well as extensions which weren't reported.

### DIFF
--- a/lfs/lfs.go
+++ b/lfs/lfs.go
@@ -82,7 +82,25 @@ func Environ() []string {
 		fmt.Sprintf("ConcurrentTransfers=%d", Config.ConcurrentTransfers()),
 		fmt.Sprintf("BatchTransfer=%v", Config.BatchTransfer()),
 		fmt.Sprintf("SkipDownloadErrors=%v", Config.SkipDownloadErrors()),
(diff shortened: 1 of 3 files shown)


sinbad:
Oh also exposes the Access settings since I thought that was useful.

technoweenie:
So the [roadmap](https://github.com/github/git-lfs/blob/e72efe65d238af3eeaea98a51df1b7b63b016887/ROADMAP.md) vaguely mentions "add all lfs.* git config keys to git lfs env output".  I'd love some automated way to do this, so we don't have to update it every time we add a new property.

However, I think this PR shows why the manual approach is good:

1. We output the value that Git LFS is using. It may come from an env var, a git config value, or implicit default in the tool.
2. The tests have to be updated with all the new properties. I'd be open to another solution to this if you have any thoughts though. Definitely not a blocker for this PR.

:ship:

sinbad:
Although we could dump the contents of the git config fairly easily, it doesn't totally reflect how we use the values internally (defaults, derived values) so I kinda like the hand crafted approach. Could maybe use reflection to automate if it became a problem? Otherwise I think this is ok for now.

PR #1222, opened 2016-05-16
Create Makefiles for building with gccgo

zeldin:
This pull request adds the ability to build git-lfs with gccgo simply by typing "make".  No "go" command is required.

### DIFF
--- a/script/genmakefile/genmakefile.go
+++ b/script/genmakefile/genmakefile.go
@@ -1,0 +1,49 @@
+package main
+
+import (
+	"fmt"
+	"go/build"
+	"os"
+	"strings"
+)
+
+var packages map[string]string = make(map[string]string)
+
+func generate_target(srcdir string, pkgdir string, prefix string, ctx build.Context) string {
+	pkg, _ := ctx.ImportDir(srcdir+pkgdir, 0)
+	name := pkg.Name
+	var deps []string
+	for _, imp := range pkg.Imports {
+		if strings.HasPrefix(imp, prefix) {
+			imp = strings.TrimPrefix(imp, prefix)
+			if packages[imp] == "" {
+				packages[imp] = generate_target(srcdir, imp, prefix, ctx)
+			}
+			deps = append(deps, "$(LIBS_"+packages[imp]+")")
+		}
+	}
+	if pkgdir != "" {
+		fmt.Printf("SRCDIR_%s := $(SRCDIR)%s/\n", name, pkgdir)
+	} else {
+		fmt.Printf("SRCDIR_%s := $(SRCDIR)\n", name)
+	}
+	fmt.Printf("SRC_%s := $(addprefix $(SRCDIR_%s), %s)\n", name, name, strings.Join(pkg.GoFiles, " "))
+	fmt.Printf("DEPS_%s := %s\n", name, strings.Join(deps, " "))
+	if pkgdir != "" {
+		fmt.Printf("OBJ_%s := $(LIBDIR)/%s.o\n", name, pkgdir)
+		fmt.Printf("LIB_%s := $(LIBDIR)/%s.a\n", name, pkgdir)
+		fmt.Printf("LIBS_%s := $(LIB_%s) $(DEPS_%s)\n", name, name, name)
+		fmt.Printf("$(OBJ_%s) : $(SRC_%s) $(DEPS_%s)\n", name, name, name)
+		fmt.Printf("\t@mkdir -p $(dir $@)\n")
+		fmt.Printf("\t$(GOC) $(GOFLAGS) -c -o $@ $(SRC_%s)\n", name)
+	}
+	return name
+}
+
+func main() {
+	srcdir := os.Args[1]
+	prefix := os.Args[2]
+	ctx := build.Default
+	ctx.CgoEnabled = false
+	generate_target(srcdir, "", prefix, ctx)
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 bin/
 benchmark/
+out/
 
 # only allow man/*.\d.ronn files
 man/*
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,52 @@
+GOC ?= gccgo
+AR ?= ar
+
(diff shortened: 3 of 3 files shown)


technoweenie:
This looks interesting. Are there any setup instructions beyond [installing gccgo](https://golang.org/doc/install/gccgo)? 

There is a test error:

```
script/script.go:15: main redeclared in this block
	previous declaration at script/govars.go:8
```

What if the script lived in `script/gccgo/govars.go` instead? That seems like a small utility command that's only needed for the makefiles. Otherwise, a `vars` command would have to be added to [the existing `main()`](https://github.com/github/git-lfs/blob/1e15699219657f9ee957a42b92cad5855cd45680/script/script.go#L17-L23). That works too.

I wonder if it'd be possible to generate these makefiles, so they don't have to be hand-touched as the packages are updated. Some of those makefiles are in `vendor/_nuts`, which are all 3rd party packages. I know that some of them will get outdated once @sinbad merges his next refactoring PR.

zeldin:
> Are there any setup instructions beyond installing gccgo?

As I'm running Gentoo, I simply emerged sys-devel/gcc with the "go" flag enabled.  :smile: 
Adding "go" to --enable-languages when building gcc as described on the webpage you refer to should be enough I think.

Yes, moving the govars script into a separate directory is fine, I'll add that to the PR when I get around to it.

As for generating the makefiles:  In order for it to be fully automatic one would need a tool that checks the (explicit and implicit) +build flags for each source file.  While creating such a tool would be interresting (maybe someone already has?  I haven't actually checked...) I considered it a bit out of scope.  I might change my mind on that point though.  :smiley_cat:

zeldin:
I have updated the PR with a reworked commit.  The make targets for all dependencies are now generated, and parallel make now works as well (it was broken before).

The make target generator uses library functions which are not available in gccgo 4.8, but I think git-lfs doesn't compile with that version anyway.

technoweenie:
Hi @zeldin This is looking good. Would you feel comfortable being listed as the maintainer of this code? No one on the LFS team uses gccgo. Also, do you know what version of go the latest version of gccgo is comparable to? We've been talking about making Go 1.5 the minimum required version.

zeldin:
That would be fine.  A list of mappings of gccgo versions to go language versions can be found on https://golang.org/doc/install/gccgo#Releases.  Allegedly the recently released gcc 6.1 supports version 1.6 of the language, but I haven't verified this for myself.  (I'm still on gcc 5.3, which implements go 1.4.2.)

script/genmakefile/genmakefile.go
```diff
@@ -0,0 +1,49 @@
...
+		fmt.Printf("LIB_%s := $(LIBDIR)/%s.a\n", name, pkgdir)
+		fmt.Printf("LIBS_%s := $(LIB_%s) $(DEPS_%s)\n", name, name, name)
+		fmt.Printf("$(OBJ_%s) : $(SRC_%s) $(DEPS_%s)\n", name, name, name)
+		fmt.Printf("\t@mkdir -p $(dir $@)\n")
+		fmt.Printf("\t$(GOC) $(GOFLAGS) -c -o $@ $(SRC_%s)\n", name)
+	}
+	return name
+}
+
+func main() {
```
ttaylorr: This may be a good thing to integrate with the [existing script.go](https://github.com/github/git-lfs/blob/1e15699219657f9ee957a42b92cad5855cd45680/script/script.go#L17-L23), as @technoweenie suggested [here](https://github.com/github/git-lfs/pull/1222#issuecomment-219529538).

script/genmakefile/genmakefile.go
```diff
@@ -0,0 +1,49 @@
...
+	fmt.Printf("DEPS_%s := %s\n", name, strings.Join(deps, " "))
+	if pkgdir != "" {
+		fmt.Printf("OBJ_%s := $(LIBDIR)/%s.o\n", name, pkgdir)
+		fmt.Printf("LIB_%s := $(LIBDIR)/%s.a\n", name, pkgdir)
+		fmt.Printf("LIBS_%s := $(LIB_%s) $(DEPS_%s)\n", name, name, name)
+		fmt.Printf("$(OBJ_%s) : $(SRC_%s) $(DEPS_%s)\n", name, name, name)
+		fmt.Printf("\t@mkdir -p $(dir $@)\n")
+		fmt.Printf("\t$(GOC) $(GOFLAGS) -c -o $@ $(SRC_%s)\n", name)
+	}
+	return name
```
ttaylorr: Looks 🆒.. Have you thought about using Go's [templating system](https://golang.org/pkg/text/template/) to accomplish the same? I think I'd be more comfortable if we had a template that included ``{{if pipeline}}`s instead of `if`'s and `fmt.Printf`'s.

ttaylorr: As an aside, you'd have to precompute all of the values that you want to print out in your template. That might not be a bad thing, since it would separate the "templating" from the data calculation, which would be 👍

script/genmakefile/genmakefile.go
zeldin: I don't think he ever suggested that.  Keep in mind that this is built as a separate binary, not run as a script.  (Since the whole point is removing the build dependency on the "go" interpreter.)

script/genmakefile/genmakefile.go
zeldin: I agree that the template variant would be slightly more elegant (although the data calulcation issue is rather minor; the only calculation done is converting an array to a space separated string, and the calculated values are used once each).  However, this feature does not seem mature.  In fact, I couldn't get it to work in any version of gccgo that I had installed.  So using it would be detrimental to my purpose (which is to help people who just want to use git-lfs, not fight against bleeding-edge google code).

script/genmakefile/genmakefile.go
ttaylorr: Ah, of course! Sounds good.

script/genmakefile/genmakefile.go
ttaylorr: OK, works for me 😄

ttaylorr:
Thanks for this! Looks good to me.

Everyone else ready to merge?

joykama:
Transfer my account pay pal account

PR #1223, opened 2016-05-16
Locking commands

ttaylorr:
This PR adds to @sinbad's original proposal and extends it to document some particular commands that we will have to teach the LFS client.

All of the documentation surrounding these new commands and API methods are in the contents of this PR, so I will keep this description brief 😄

Thanks in advance for reading! 👋

### DIFF
--- a/docs/proposals/locking.md
+++ b/docs/proposals/locking.md
@@ -248,7 +248,7 @@ type UnlockResult struct {
 ```
 
 Clients can determine whether or not their lock was removed by calling the
-`Active()` method on the returned Lock, if `UnlockResult.Err` is non-nil.
+`Active()` method on the returned Lock, if `UnlockResult.Err` is nil.
 
 #### `git lfs locks (-r <remote>|-b <branch|-p <path>)|(-i id)`
 
--- a/docs/proposals/locking.md
+++ b/docs/proposals/locking.md
@@ -169,6 +169,8 @@ type LockRequest struct {
         // Remote is the remote on which the client would like to obtain the
         // lock.
         Remote    string `json:"remote"`
+        // Head is a ref pointing to the latest commit that the client has.
+        Head string `json:"head"`
         // Committer is the individual that wishes to obtain the lock.
         Committer struct {
               // Name is the name of the individual who would like to obtain the
@@ -195,6 +197,9 @@ type LockResponse struct {
         // If an error was experienced in creating this lock, then the
         // zero-value of Lock should be sent here instead.
         Lock Lock `json:"lock"`
+        // CommitNeeded holds the minimum commit SHA that client must have to
+        // obtain the lock.
+        CommitNeeded string `json:"commit_needed"`
         // Err is the optional error that was encountered while trying to create
         // the above lock.
         Err error `json:"error,omitempty"`
--- a/docs/proposals/locking.md
+++ b/docs/proposals/locking.md
@@ -136,6 +136,10 @@ type Lock struct {
                Name  string `json:"name"`
                Email string `json:"email"`
         } `json:"creator"`
+        // CommitSHA is the commit that this Lock was created against. It is
+        // strictly equal to the SHA of the minimum commit negotiated in order
+        // to create this lock.
+        CommitSHA string `json:"commit_sha"
         // LockedAt is a required parameter that represents the instant in time
         // that this lock was created. For most server implementations, this
         // should be set to the instant at which the lock was initially
--- a/docs/proposals/locking.md
+++ b/docs/proposals/locking.md
@@ -125,6 +125,9 @@ sub-package.
 // Locks returned from the API may or may not be currently active, according to
 // the Expired flag.
 type Lock struct {
+        // Id is the unique identifier corresponding to this particular Lock. It
+        // must be consistent with the local copy, and the server's copy.
+        Id string `json:"id"`
         // Path is an absolute path to the file that is locked as a part of this
         // lock.
         Path string `json:"path"`
@@ -216,17 +219,17 @@ type LockResponse struct {
 
 The `unlock` command is responsible for releasing the lock against a particular
 file. The command takes a `<path>` argument which the LFS client will have to
-internally resolve into a UUID to unlock.
+internally resolve into a Id to unlock.
 
 The API associated with this command can also be used on the server to remove
 existing locks after a push.
 
 ```go
 // An UnlockRequest is sent by the client over the API when they wish to remove
-// a lock associated with the given UUID.
+// a lock associated with the given Id.
 type UnlockRequest struct {
-        // UUID is the identifier of the lock that the client wishes to remove.
-        UUID string `json:"uuid"`
+        // Id is the identifier of the lock that the client wishes to remove.
+        Id string `json:"id"`
 }
 ```
 
@@ -247,7 +250,7 @@ type UnlockResult struct {
 Clients can determine whether or not their lock was removed by calling the
 `Active()` method on the returned Lock, if `UnlockResult.Err` is non-nil.
 
-#### `git lfs locks (-r <remote>|-b <branch|-p <path>)|(-u uuid)`
+#### `git lfs locks (-r <remote>|-b <branch|-p <path>)|(-i id)`
 
 For many operations, the LFS client will need to have knowledge of existing
 locks on the server. Additionally, the client should not have to self-sort/index
@@ -264,8 +267,8 @@ field in the `LockListRequest`.
 type Property string
 
 const (
-        UUID   Property = "uuid"
         Branch Property = "branch"
+        Id     Property = "id"
         // (etc) ...
 )
 
--- a/docs/proposals/locking.md
+++ b/docs/proposals/locking.md
@@ -133,22 +133,22 @@ type Lock struct {
                Name  string `json:"name"`
                Email string `json:"email"`
         } `json:"creator"`
-        // CreatedAt is a required parameter that represents the instant in time
+        // LockedAt is a required parameter that represents the instant in time
         // that this lock was created. For most server implementations, this
         // should be set to the instant at which the lock was initially
         // received.
-        CreatedAt time.Time `json:"created_at"`
+        LockedAt time.Time `json:"locked_at"`
         // ExpiresAt is an optional parameter that represents the instant in
         // time that the lock stopped being active. If the lock is still active,
         // the server can either a) not send this field, or b) send the
         // zero-value of time.Time.
-        ExpiresAt time.Time `json:"expired_at,omitempty"`
+        UnlockedAt time.Time `json:"unlocked_at,omitempty"`
 }
 
 // Active returns whether or not the given lock is still active against the file
 // that it is protecting.
 func (l *Lock) Active() bool {
-        return time.IsZero(l.ExpiresAt)
+        return time.IsZero(l.UnlockedAt)
 }
 ```
 
--- a/docs/proposals/locking.md
+++ b/docs/proposals/locking.md
@@ -128,8 +128,8 @@ type Lock struct {
         // Path is an absolute path to the file that is locked as a part of this
         // lock.
         Path string `json:"path"`
-        // Creator is the author who initiated this lock.
-        Creator struct {
+        // Committer is the author who initiated this lock.
+        Committer struct {
                Name  string `json:"name"`
                Email string `json:"email"`
         } `json:"creator"`
--- a/docs/proposals/locking.md
+++ b/docs/proposals/locking.md
@@ -112,7 +112,7 @@ support:
 ## Implementation details
 ### Types
 To make the implementing locking on the lfs-test-server as well as other servers
-in the future easier, it makes sense to create a `lfs/lock` package that can be
+in the future easier, it makes sense to create a `lock` package that can be
 depended upon from any server. This will go along with Steve's refactor which
 touches the `lfs` package quite a bit.
 
--- a/docs/proposals/locking.md
+++ b/docs/proposals/locking.md
@@ -108,3 +108,188 @@ support:
 |9  |Cater for read-only binary files when merging locally<ul><li>Because files are read-only this might prevent merge from working when actually it's valid.</li><li>Always fine to merge the latest version of a binary file to anywhere else</li><li>Fine to merge the non-latest version if user is aware that this may cause merge problems (see Breaking the rules)</li><li>Therefore this feature is about dealing with the read-only flag and issuing a warning if not the latest</li></ul>|
 |10 |List current locks<ul><li>That the current user has</li><li>That anyone has</li><li>Potentially scoped to folder</li></ul>|`git lfs lock --list [paths...]`
 |11 |Reject a push containing a binary file currently locked by someone else|pre-receive hook on server, allow --force to override (i.e. existing parameter to git push)
+
+## Implementation details
+### Types
+To make the implementing locking on the lfs-test-server as well as other servers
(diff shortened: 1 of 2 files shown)


ttaylorr:
Hmm, after thinking more about this, I realized that will need to expand the API to cover the commit negotiation that @sinbad described in his original pull-request. I added partial support for this in 6864ac5cc439199fea06ddd3c9ee2f103a26fdb9, but I will have to think about a more flexible flow to support more implementations.

docs/proposals/locking.md
```diff
@@ -108,3 +108,188 @@ support:
 |9  |Cater for read-only binary files when merging locally<ul><li>Because files are read-only this might prevent merge from working when actually it's valid.</li><li>Always fine to merge the latest version of a binary file to anywhere else</li><li>Fine to merge the non-latest version if user is aware that this may cause merge problems (see Breaking the rules)</li><li>Therefore this feature is about dealing with the read-only flag and issuing a warning if not the latest</li></ul>|
 |10 |List current locks<ul><li>That the current user has</li><li>That anyone has</li><li>Potentially scoped to folder</li></ul>|`git lfs lock --list [paths...]`
 |11 |Reject a push containing a binary file currently locked by someone else|pre-receive hook on server, allow --force to override (i.e. existing parameter to git push)
+
+## Implementation details
+### Types
+To make the implementing locking on the lfs-test-server as well as other servers
+in the future easier, it makes sense to create a `lfs/lock` package that can be
+depended upon from any server. This will go along with Steve's refactor which
+touches the `lfs` package quite a bit.
```
technoweenie: I think the idea is that `github.com/github/git-lfs/lfs` is _not_ a package that we want 3rd parties to use. This should be extracted to a top level package like `github.com/github/git-lfs/lock`. May make sense to call it `github.com/github/git-lfs/lfslock` so that it's obvious when callers use it:

```go
lfslock.Lock{Path: "img/foo.psd", ...}
```

I dont'e care about the name so much, but I do want it as a top level package under the repo.

ttaylorr: I agree, but I think just `lock` should be fine. The generic nature of that package's name is more of a concern of any code that depends on it, not the LFS client itself. If someone that depends on this package and needs a more specific name, they can just rename it:

```go
import (
        lfslock github.com/github/git-lfs/lock
)
```

docs/proposals/locking.md
```diff
@@ -108,3 +108,198 @@ support:
...
+
+The API associated with this command can also be used on the server to remove
+existing locks after a push.
+
+```go
+// An UnlockRequest is sent by the client over the API when they wish to remove
+// a lock associated with the given UUID.
+type UnlockRequest struct {
+        // UUID is the identifier of the lock that the client wishes to remove.
+        UUID string `json:"path"`
```
technoweenie: json tag typo.

ttaylorr:
Demoting this PR to a `[WIP]` as I still have to consider some final thoughts for the lock-listing API. In particular:

Pagination is a bit of a tricky subject, since there is no guarantee that a client will get all of the locks they requested if a multipage response is returned and a new lock was created while paginating through the n-th page of that response. @technoweenie and I discussed that perhaps the best way around this was to eliminate the notion of "pages" and instead make it a requirement that the server return results in a reverse-chronological ordering. Clients could instead send the last UUID that they received and the server would send them `n` results after that, and we'd still keep the `HasMore bool`.

The other thing I have yet to decide upon is the filtering mechanism. The current filter definition that I proposed:

```go
type Filter struct {                                               
        // Prop is the property to search against.         
        Prop Property `json:"prop"`                        
        // Value is the value that the property must take. 
        Value string `json:"value"`                        
}
```

works well for equality operations, but fails for `<`, `>`, and `!=` operations. This is sort of a rabbit-hole, since once you consider adding operations for `<` and `>`, it's easy to add nested filters and `AND`s, `OR`s, and `XOR`s. I'm not sure how far we should take this idea, but if we pursue the idea of nested filters, I could work on an API to accommodate that.

technoweenie:
I like the type definitions, but I think it'd be helpful thinking about the API too. For example, I'm not sure unlocking will require a struct like this:

```
// An UnlockRequest is sent by the client over the API when they wish to remove
// a lock associated with the given UUID.
type UnlockRequest struct {
        // UUID is the identifier of the lock that the client wishes to remove.
        UUID string `json:"path"`
}
```

Given an LFS base url of `https://git-server.com/repo/info/lfs`, the unlock API call could be something like `POST https://git-server.com/repo/info/lfs/locks/{uuid} -d {"operation":"unlock", ...}`.

Also, I think the spec should recommend that server implementations keep locks for auditing purposes. This is why the unlock action is a POST to the resource, and not a DELETE.

docs/proposals/locking.md
```diff
@@ -108,3 +108,198 @@ support:
...
+        // CreatedAt is a required parameter that represents the instant in time
+        // that this lock was created. For most server implementations, this
+        // should be set to the instant at which the lock was initially
+        // received.
+        CreatedAt time.Time `json:"created_at"`
+        // ExpiresAt is an optional parameter that represents the instant in
+        // time that the lock stopped being active. If the lock is still active,
+        // the server can either a) not send this field, or b) send the
+        // zero-value of time.Time.
+        ExpiresAt time.Time `json:"expired_at,omitempty"`
```
sinbad: I'm not clear on how this field is intended to work. The past tense on "time that the lock stopped being active" suggests it's only for historical reporting, so it would only be filled in after the user had deliberately unlocked the file? Or, are you intending on making locks time-limited so they unlock on their own? I think the latter would cause some issues, files unlocking on a time limit could increase the likelihood of accidental parallel changes - the 'break lock' function is intended to recover from cases where a lock is left active by someone else by accident.

docs/proposals/locking.md
```diff
@@ -108,3 +108,198 @@ support:
...
+type LockRequest struct {
+        // Path is the path that the client would like to obtain a lock against.
+        Path      string `json:"path"`
+        // Remote is the remote on which the client would like to obtain the
+        // lock.
+        Remote    string `json:"remote"`
+        // LatestRemoteCommit is the SHA of the last known commit from the
+        // remote that we are trying to create the lock against, as found in
+        // `.git/refs/origin/<name>`.
+        LatestRemoteCommit string `json:"latest_remote_commit"`
```
sinbad: I think just `LastCommit` since there is no distinction between what is a commit on a remote and a commit locally. The important thing is that the remote must be able to confirm that there have been no other edits on this file since this point (or the common ancestor point) on any branch it knows about.

docs/proposals/locking.md
```diff
@@ -108,3 +108,198 @@ support:
...
+        // If the lock was unable to be created, this field will hold the
+        // zero-value of Lock and the Err field will provide a more detailed set
+        // of information.
+        //
+        // If an error was experienced in creating this lock, then the
+        // zero-value of Lock should be sent here instead.
+        Lock Lock `json:"lock"`
+        // CommitNeeded holds the minimum commit SHA that client must have to
+        // obtain the lock.
+        CommitNeeded string `json:"commit_needed"`
```
sinbad: Presumably `CommitNeeded` will only be populated if there is an error (and therefore no lock)? Implication being the client needs to checkout that commit or cherry pick the state, and also may have to `git fetch` if it doesn't have that commit locally.

docs/proposals/locking.md
```diff
@@ -108,3 +108,198 @@ support:
...
+// Lock represents a single lock that against a particular path.
+//
+// Locks returned from the API may or may not be currently active, according to
+// the Expired flag.
+type Lock struct {
+        // Path is an absolute path to the file that is locked as a part of this
+        // lock.
+        Path string `json:"path"`
+        // Creator is the author who initiated this lock.
+        Creator struct {
```
sinbad: This is called `Committer` in `LockRequest`, suggest they should be the same for consistency.

docs/proposals/locking.md
```diff
@@ -108,3 +108,198 @@ support:
...
+
+The API associated with this command can also be used on the server to remove
+existing locks after a push.
+
+```go
+// An UnlockRequest is sent by the client over the API when they wish to remove
+// a lock associated with the given UUID.
+type UnlockRequest struct {
+        // UUID is the identifier of the lock that the client wishes to remove.
+        UUID string `json:"uuid"`
```
sinbad: Where did the client find out what the `UUID` was? It probably should be in the `Lock` struct?

For clarity, I assume the reason for `UUID` (rather than just `Path`) is to allow for parallel locks when `--force` is used as described in the 'breaking the rules' section, which works for me.

docs/proposals/locking.md
```diff
@@ -108,3 +108,198 @@ support:
...
+// lock.
+type UnlockResult struct {
+        // Lock is the lock corresponding to the asked-about lock in the
+        // `UnlockPayload` (see above). If no matching lock was found, this
+        // field will take the zero-value of Lock, and Err will be non-nil.
+        Lock Lock `json:"lock"`
+        // Err is an optional field which holds any error that was experienced
+        // while removing the lock.
+        Err error `json:"error,omitempty"`
+}
```
sinbad: Maybe worth saying that the presence of a valid `Lock` and lack of an `Err` implies success. I also assume HTTP status code are going to signal general success/fail.

docs/proposals/locking.md
```diff
@@ -108,3 +108,198 @@ support:
...
+```
+
+Clients can determine whether or not their lock was removed by calling the
+`Active()` method on the returned Lock, if `UnlockResult.Err` is non-nil.
```
sinbad: "is non-nil" -> "is nil" I think?

docs/proposals/locking.md
```diff
@@ -108,3 +108,198 @@ support:
...
+
+#### `git lfs locks (-r <remote>|-b <branch|-p <path>)|(-u uuid)`
+
+For many operations, the LFS client will need to have knowledge of existing
+locks on the server. Additionally, the client should not have to self-sort/index
+this (potentially) large set. To remove this need, both the `locks` command and
+corresponding API method take several filters.
```
sinbad: Good call.

docs/proposals/locking.md
```diff
@@ -108,3 +108,198 @@ support:
...
+                // Prop is the property to search against.
+                Prop Property `json:"prop"`
+                // Value is the value that the property must take.
+                Value string `json:"value"`
+        } `json:"filters"`
+        // Page is the numerical page of the result set to return.
+        Page int `json:"page"`
+        // Limit is the maximum number of locks to return in a single page.
+        Limit int `json:"limit"`
+}
```
sinbad: Nice, pagination in v1 :) 👍

docs/proposals/locking.md
technoweenie: I wonder if the key name should be `UnlockedAt` or something then? It'd likely only be shown when viewing an old lock by its UUID.

docs/proposals/locking.md
technoweenie: While talking, we came up with three scenarios where a lock would be rejected:

1. Lock already exists. Return status 409, with the `lock` json object.
2. Commit is old. Return status 409, with `commit_needed` to the current HEAD on the remote.
3. Lock exists and commit is old. Return status 409 with both `lock` and `commit_needed`.

This `commit_needed` field is why I think `latest_remote_commit` in the lock request should be a commit the server knows. If the server HEAD is at commit `abc`, what good does it do to send `def`? If the server has never seen that commit, it's just an empty string of bytes to it.

My thinking is that if a user has already made a local edit `def`, the client still sends `abc`, since that's a commit that it knows. If the server has been updated to `abc2`, AND the commits `abc...abc2` touched the file of the requested lock, `commit_needed` should force the client to update to the latest version.

docs/proposals/locking.md
technoweenie: Should we just call this `Id` too? It may not matter if the uuid for a single repository's locks is just an auto increment ID of the table, or scoped to the repository somehow.

docs/proposals/locking.md
technoweenie: We talked about this. Another thought we had was relying on servers to sort all locks in reverse chronological, and use cursors for paging.

REQ1: `LockListRequest{}` (first page of all locks for the repository)
RES1: `LockList{Locks: []Lock{...}, HasMore: true}`
REQ2: `LockListRequest{Cursor: "lock-uuid-1"}` (first page of all locks for the repository)
RES2: `LockList{Locks: []Lock{...}, HasMore: true}` (locks after `lock-uuid-1`)

Another thought is to replace `HasMore` with `NextCursor`, so the client doesn't have to think about what the cursor should be on requests after the initial one.

docs/proposals/locking.md
sinbad: > My thinking is that if a user has already made a local edit `def`, the client still sends `abc`

Hmm, but if the client sent `def` then the server would immediately know that the client had made an edit without locking the file, which probably should be something that is caught. For clarity I was imagining that the client would send the SHA of the commit where the file was last modified, not necessarily `HEAD`.

docs/proposals/locking.md
sinbad: Yeah that's a good call, having explicit `Next` data rather than have the client figure it out is more friendly

docs/proposals/locking.md
ttaylorr: @sinbad good point. I had mentioned to @technoweenie that I thought having a lock with a given expiry time would be tricky, since that adds an extra layer of complexity to the locking mechanism. My original intention with this field was make it so clients (and the server) would be able to determine which locks were active or not, while also maintaining a historical list of locks that had been placed for posterity.

The tense is a little weird here, so I think I'll adopt @technoweenie's suggestion and rename it to `UnlockedAt`, and still retain the `omitempty` property of the JSON tag. That way, we're not using `nil` as a value (by changing this field's type to *time.Time), and we have the benefit of Go's zero-type initialization that we can take advantage of with the `time.IsZero` method.

docs/proposals/locking.md
ttaylorr: wfm!

ttaylorr: I think @technoweenie brings up a good point in [this](https://github.com/github/git-lfs/pull/1223#discussion_r63512395).

docs/proposals/locking.md
ttaylorr: 👍

docs/proposals/locking.md
ttaylorr: I'm fine with calling it `Id`. I also added this field to the `Lock` type definition, so that clients can know about the ID, too.

docs/proposals/locking.md
ttaylorr: 👍

docs/proposals/locking.md
ttaylorr: Thanks! 😄

docs/proposals/locking.md
ttaylorr: I second @sinbad! Good call with having the server return a NextCursor. Added.

ttaylorr:
On second thought, the currently proposed filter API might be alright for V1. I'm not sure the `>`/`<` operators are that useful for anything other than `created_at` time. It would be nice to greater support for nested filters, but I move that we adopt the current model for now, and expand on it in v2 if we find a greater need for more filtering operations.

technoweenie:
Merging this. Great job finishing the proposal, time to get crackin on a prototype :metal:

ttaylorr:
Thanks, @technoweenie! Let's do it.

PR #1226, opened 2016-05-17
Package refactor no.1

sinbad:
In preparation for extending git-lfs to allow more transfer (and later, api) options, I've done the first (most significant) stage of a refactor of the `lfs` package into more modular parts.

The key idea is to align parts of the functionality into more cohesive partitions so that each of them can be adapted more easily in future. This has involved the creation of a lot more packages in order to prevent circular dependencies. The general principles of the dependency tree are:

- `lfs` increasingly pulls functionality together rather than providing most of it
- `api` contains logic associated with calling the remote LFS server
- `httputil`, `errutil` collect together different utilities that are used in multiple packages
- `tools` brings together some file system and I/O helpers, too small to be their own package each right now
- `localstorage` existed before but now has more responsibility for the local state, `lfs` delegates more to it
- `config` collects all the environment, git config and lfs config together for use by all other packages
- `auth` covers credentials and the ssh authentication
- `progress` holds the spinners, progress meters etc

It's not done yet but I think I've come a long way to untangling the ball of string that was the `lfs` package. I think we should merge it at this point to head off conflicts since it's already useful. 

The main missing part is the `transfer` package. This will focus on the transferring of content (not the metadata) including parallelisation, so that this can be extended more easily (the original trigger for this work). However, to complete this part of the refactor requires me to move a couple of bits from `api` (`UploadObject` and `DownloadObject`) and also to pull apart the transfer queue, which currently groups together both the logic for batching up calls to the API, and calling the actual upload/download hooks - these need to be separated so batching & API is separate to transferring. This is non-trivial so needs a few more days, I'll do that next but it only affects a handful of files so I think it's worth merging this first.

Aside: I think we might want to create a branch for 1.2.x which excludes this and keep master for this and other big feature work like locking.


technoweenie:
This is a great first step in extracting stuff out of lfs. Do you worry that even by merging to master, we're communicating to the Go community that a) the `github.com/github/git-lfs/lfs` package is broken and 2) `github.com/github/git-lfs/api` is ready for your 3rd party go libs to use?

I think my ideal scenario would go like this:

1. Add some go comment to the top of the packages, and probably README.md too that these go packages are **experimental**, and subject to break. We currently only care that the client works, while the house cleaning is happening.
2. Create `release-1.2` branch.
3. Merge this PR
4. Start smaller PRs to clean the public API of the packages. These could be as simple as renaming `progress.NewProgressMeter` to `progress.NewMeter`, fixing the use of `config.Config`, or just refactoring the api client as we've discussed in issues like #779.

How's that sound?

(Also: not saying _you_ have to worry about all the changes. I'd be totally cool with you focusing on the changes you need for the transfer features stuff. Hope that goes without saying. But just in case...)

sinbad:
Yeah I hadn't really been clear on external Go use and was focussed on internal structuring. My feeling is that either:

1. External Go source should only depend on the `lfs` package, or 
1. We expose a top-level facade in the root `git-lfs` package as an explicit external api, which is more tightly controlled

I think the second option is the best one because it's clearer about what's a fixed external API and what's internal structuring. I'm happy with your steps and saying in the docs that we'll expose a more dependable api in the root later if you agree.

sinbad:
Docs done, I suggest we branch `release-1.2` at 1e15699219657f9ee957a42b92cad5855cd45680?

PR #1228, opened 2016-05-18
vendor the testify package

technoweenie:
@ttaylorr requested the testify package. Since it doesn't make sense to have multiple test helpers, I think we should also remove `github.com/technoweenie/assert`. The `github.com/technoweenie/assert` package is a fork of a dead project. It also halts test execution after each assertion.

The testify package supports multiple assertion failures in a single test. It also includes a mocking package that @ttaylorr wants to use in his api client tests (https://github.com/ttaylorr/git-lfs/pull/2).

@sinbad: Any thoughts?


sinbad:
Seems good to me. I wouldn't go too overboard with the mocking though, even though mockify can generate the code it can quickly become a lot to manage, I'm pro-minimalism where possible.

ttaylorr:
Sweet! 👍

PR #1236, opened 2016-05-21
API Client and Lock Service

ttaylorr:
This PR introduces the Git LFS API client, as well as the `LockService`, for interacting with the [new LFS file locking API](https://github.com/github/git-lfs/pull/1223).

# API Client
The API client is broken down into several key components, all of which are described here:

## Request Schemas
A request schema provides a definition or template type from which to build an actual request. Right now, the definition of this type is as follows:

```go
// RequestSchema provides a schema from which to generate sendable requests.
type RequestSchema struct {
	// Method is the method that should be used when making a particular API
	// call.
	Method string
	// Path is the relative path that this API call should be made against.
	Path string
	// Query is the query parameters used in the request URI.
	Query map[string]string
	// Body is the body of the request.
	Body interface{}
	// Into is an optional field used to represent the data structure into
	// which a response should be serialized.
	Into interface{}
}
```

## Request/Response Lifecycle
The lifecycle type is responsible for a few things:

1. Building an executable request based on the given schema.
2. Executing that request against an API server (and serialize the result, if applicable)
3. Cleaning up after that request has finished.

Right now, things are a little `net/http` specific, and the only Lifecycle implementation is the `*api.HttpLifecycle` type. That type knows how to do all of the things described above in an HTTP-specific context. Internally, it uses the `*http.Client` type to execute the requests.

In the future, I would like to abstract away the HTTP-specific stuff, and open the API up to an SSH, or even WebSocket implementation.

## Services
Services are responsible for implementing API methods by providing code to generate `*api.RequestSchema`s, which can be passed to the `*api.Client` in use to be sent through the lifecycle subsystem, and eventually, a response is returned.

### Lock Service

This PR includes the `LockService` type, which implements all of the APIs type definitions, as well as the three methods specified in the File Locking API proposal (Lock, Unlock, and Search).

An example use of the `LockService` would be as follows.
 
```go
client := api.NewClient(...)
locks, schema := client.Locks.Search(&api.LockSearchRequest{
        // ...
})

resp, err := client.Do(schema)
if err != nil {
        panic(err.Error())
}

for _, l := range locks.Locks {
        // ...
}

fmt.Fprint("API call succeeded with status %d\n", resp.StatusCode())
```

## `schema` package

We also introduced a way to test that our implementation of the API as defined in the `docs/` directory is correct, by using JSON schema.

This is a nice boundary to pick, since we have tests for the Service methods, and tests for the API client code, so the actual API implementation is tested separately. 

To test a new API type, simply add a `*_schema.json` file to the `api/schema/` directory, and you're good to go:

```go
func TestLockResponseSchema(t *testing.T) {
        schema.Validate(t, schema.LockResponse, &api.LockResponse{
                // ...
        })
}
```

or, alternatively:

```go
func TestInvalidLockResponseSchema(t *testing.T) {
        schema.Refute(t, schema.LockResponse, &api.LockResponse{
                // ...
        })
}
```

-------------------------------------

Thanks in advance to anyone who reads through this PR, I appreciate you taking the time to look through this rather lengthy diff. 😄 

Special thanks to @technoweenie and @sinbad for listening to my design proposals, as well as providing invaluable suggestions and advice along the way. 👋 

Enjoy!


ttaylorr:
Hmm... looks like there was some weirdness going on with one of `gojsonschema`'s dependencies not being properly vendored. @technoweenie: any thoughts?

technoweenie:
Argh, we have to rewrite all the imports.

technoweenie:
Try this: 

```diff
diff --git a/vendor/_nuts/github.com/xeipuuv/gojsonreference/reference.go b/vendor/_nuts/github.com/xeipuuv/gojsonreference/reference.go
index d4d2eca..3713a81 100644
--- a/vendor/_nuts/github.com/xeipuuv/gojsonreference/reference.go
+++ b/vendor/_nuts/github.com/xeipuuv/gojsonreference/reference.go
@@ -27,7 +27,7 @@ package gojsonreference

 import (
        "errors"
-       "github.com/xeipuuv/gojsonpointer"
+       "github.com/github/git-lfs/vendor/_nuts/github.com/xeipuuv/gojsonpointer"
        "net/url"
        "path/filepath"
        "runtime"
diff --git a/vendor/_nuts/github.com/xeipuuv/gojsonschema/jsonLoader.go b/vendor/_nuts/github.com/xeipuuv/gojsonschema/jsonLoader.go
index 0b40574..53b3733 100644
--- a/vendor/_nuts/github.com/xeipuuv/gojsonschema/jsonLoader.go
+++ b/vendor/_nuts/github.com/xeipuuv/gojsonschema/jsonLoader.go
@@ -38,7 +38,7 @@ import (
        "runtime"
        "strings"

-       "github.com/xeipuuv/gojsonreference"
+       "github.com/github/git-lfs/vendor/_nuts/github.com/xeipuuv/gojsonreference"
 )

 var osFS = osFileSystem(os.Open)
diff --git a/vendor/_nuts/github.com/xeipuuv/gojsonschema/schema.go b/vendor/_nuts/github.com/xeipuuv/gojsonschema/schema.go
index 577c786..c4cb9f4 100644
--- a/vendor/_nuts/github.com/xeipuuv/gojsonschema/schema.go
+++ b/vendor/_nuts/github.com/xeipuuv/gojsonschema/schema.go
@@ -32,7 +32,7 @@ import (
        "reflect"
        "regexp"

-       "github.com/xeipuuv/gojsonreference"
+       "github.com/github/git-lfs/vendor/_nuts/github.com/xeipuuv/gojsonreference"
 )

 var (
diff --git a/vendor/_nuts/github.com/xeipuuv/gojsonschema/schemaPool.go b/vendor/_nuts/github.com/xeipuuv/gojsonschema/schemaPool.go
index 70ae731..ab14242 100644
--- a/vendor/_nuts/github.com/xeipuuv/gojsonschema/schemaPool.go
+++ b/vendor/_nuts/github.com/xeipuuv/gojsonschema/schemaPool.go
@@ -30,7 +30,7 @@ import (
        "errors"
        "net/http"

-       "github.com/xeipuuv/gojsonreference"
+       "github.com/github/git-lfs/vendor/_nuts/github.com/xeipuuv/gojsonreference"
 )

 type schemaPoolDocument struct {
diff --git a/vendor/_nuts/github.com/xeipuuv/gojsonschema/subSchema.go b/vendor/_nuts/github.com/xeipuuv/gojsonschema/subSchema.go
index b249b7e..a2e72a5 100644
--- a/vendor/_nuts/github.com/xeipuuv/gojsonschema/subSchema.go
+++ b/vendor/_nuts/github.com/xeipuuv/gojsonschema/subSchema.go
@@ -31,7 +31,7 @@ import (
        "regexp"
        "strings"

-       "github.com/xeipuuv/gojsonreference"
+       "github.com/github/git-lfs/vendor/_nuts/github.com/xeipuuv/gojsonreference"
 )

 const (
```

ttaylorr:
@technoweenie thanks! Applied your commit, but it looks like we may still be missing some of `github.com/stretchr/testify`'s dependencies. We may want to look into a vendoring solution that doesn't force us to vendor all subdeps of a given package. Did you end up messing around with your own scripts this weekend?

sinbad:
On deps, I'd say this is a good time to move to go's 1.6's standard `vendor/` dir and switch to either [GoDep](https://github.com/tools/godep) or [Glide](https://glide.sh/), both of which understand this approach. Then we can drop the `GO15VENDOREXPERIMENT=0` requirement when people are running the latest go. It means requiring go 1.5+ though.

As for vendoring all sub-deps, this is a given if you expect the project to build straight after clone. The alternative is to not commit *any* of the vendored packages and require a post-clone command instead (e.g. `glide install`). The downside of that is you have to remember to run it whenever the dependencies change in a breaking way or scratch your head for a couple of minutes before you remember. I don't see any benefit of a halfway house where direct deps are committed but subdeps aren't.

Personally I prefer committing all deps because then the only time you have to worry about your package management is when you personally change the dependencies. But either works.

api/lifecycle.go
```diff
@@ -0,0 +1,32 @@
...
+// lifecycle of a request in a platform-agnostic fashion.
+type Lifecycle interface {
+	// Build creates a sendable request by using the given RequestSchema as
+	// a model.
+	Build(req *RequestSchema) (*http.Request, error)
+
+	// Execute transforms generated request into a wrapped repsonse, (and
+	// optionally an error, if the request failed), and serializes the
+	// response into the `into interface{}`, if one was provided.
+	Execute(req *http.Request, into interface{}) (Response, error)
```
sinbad: I thought part of the idea of abstracting this was to make the interface non-specific to HTTP, but you're including `http.Request` in the interface here. `Response` conversely is non-protocol specific, which is more what I expected.

api/lock_api.go
```diff
@@ -0,0 +1,233 @@
...
+// on that file, then the CommitNeeded field will be populated in the
+// LockResponse, signaling that more commits are needed.
+//
+// In the successful case, a new Lock will be returned and granted to the
+// caller.
+func (s *LockService) Lock(req *LockRequest) (*RequestSchema, *LockResponse) {
+	var resp LockResponse
+
+	return &RequestSchema{
+		Method: http.MethodPost,
```
sinbad: I don't think `Method` should expect HTTP-specific values. The HTTP protocol implementation could just be an implementation detail even though I know that's the expected default implementation. It's just that dependency on HTTP-specific elements subtly scattered around is what made the `lfs` package more rigid than it needed to be and since you're refactoring...

How about making this a custom enum and just translating it to `http.MethodPost` in the http lifecycle?

api/http_lifecycle.go
```diff
@@ -0,0 +1,160 @@
...
+// is returned.
+//
+// Otherwise, the api.Response is returned, along with no error, signaling that
+// the request completed successfully.
+func (l *HttpLifecycle) Execute(req *http.Request, into interface{}) (Response, error) {
+	resp, err := l.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
```
sinbad: There are some extra things you'll have to deal with here if this is going to replace the old api code eventually:

- Redirects
- NTLM
- Self-signed certificates
- Tracing
- Stats 

See the `httputil` package for details.

api/http_lifecycle.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+var _ Lifecycle = new(HttpLifecycle)
+
+// NewHttpLifecycle initializes a new instance of the *HttpLifecycle type with a
+// new *http.Client, and the given root (see above).
+func NewHttpLifecycle(root *url.URL) *HttpLifecycle {
+	return &HttpLifecycle{
+		root:   root,
+		client: new(http.Client),
```
sinbad: See `httputil.NewHttpClient` for some cases you won't be handling by just creating a standard `http.Client`.

technoweenie:
> We may want to look into a vendoring solution that doesn't force us to vendor all subdeps of a given package. 

Yup, until then, I'd just update the imports here. 

> Did you end up messing around with your own scripts this weekend?

Nah, I think I'll try [glide](https://github.com/Masterminds/glide) instead.

ttaylorr:
Hey @sinbad, thanks for the feedback 👍  I went ahead and moved all of our dependencies into `vendor/` using Glide over in https://github.com/github/git-lfs/pull/1243. Grabbing lunch in a few, but will address your comments shortly.

api/lifecycle.go
ttaylorr: You're right, but I had mentioned to @technoweenie that I thought this was a good intermediate step between the old HTTP-specific code and the new API client stuff. Eventually, I'd like to find the correct abstraction between the `*http.Request` and `api.Request`, but until we know what the SSH API is going to look like, this is difficult.

If it's okay with you, I think we should leave this and the `Method` field on `RequestSchema` as is until we can see what an SSH api would need specifically.

api/http_lifecycle.go
ttaylorr: Replacing the *http.Client with an `*httputil.HttpClient` seems to cover redirects, self-signed certs, tracing and stats, but not NTLM. Is there something special I need to do in order to make NTLM work>

ttaylorr:
OK, pushed a commit which uses the `*httputil.HttpClient`, and migrates those tests to run using a `testify.Suite` so we can `Setup()` and `TearDown()` the creds check bypass.

api/lifecycle.go
sinbad: Fair enough. Maybe a TODO comment though?

api/http_lifecycle.go
sinbad: I don't entirely understand it myself but it appears that there's a pre-negotiation stage which calls the NTLM service and provides an alternative pre-authenticated request which you call instead. You can see it in action in `httputil.doHttpRequest` - I don't think this code is as tidy as it could be, it currently takes different paths to creating a HttpClient which I think is unnecessary, but I just moved it as-is when refactoring because I don't understand it enough or have any real world test cases. You can probably eliminate the duplicate HttpClient creation and refactor it to match your Build/Execute sequencing but would be worth finding someone who uses NTLM to prove it still works beyond our tests.

api/lifecycle.go
ttaylorr: [Here](https://github.com/github/git-lfs/pull/1236/files#diff-25507c8a6cd1614550f235443ae8e53bR6) you go 😄

ttaylorr:
Huh, not sure why Travis ran the last test against Go 1.3.1 (not in the `.travis.yml`), but that's what caused the failure. Must have been an intermediate Travis bug.

technoweenie:
Looks good :metal:

ttaylorr:
Thanks @technoweenie, @sinbad! 👋

PR #1243, opened 2016-05-23
vendor: vendor dependencies in vendor/ using Glide

ttaylorr:
- script/vendor received an update in order to work with Glide
- import paths have been rewritten to work with GO15VENDOREXPERIMENT


technoweenie:
It's highly suspicious that this diff removes so many lines of code, since it should just be moving the files. Turns out we had both `github.com/ThomsonReutersEikon/log4go` and `code.google.com/p/log4go` vendored, and `code.google.com/p/log4go` was removed since that import path is not used anymore.

I think this is a good change :+1: A couple more things:

* [ ] Fix any `GO15VENDOREXPERIMENT=0` in the repo to `GO15VENDOREXPERIMENT=1`.
* [ ] Update the contributing guide to mention Glide and the new vendor path [here](https://github.com/github/git-lfs/blob/3efa04cbb4564d8e1a5afbb863956459eac75353/CONTRIBUTING.md#L97).

I'm not 100% sold on glide, but hopefully this makes it easier to switch those tools later. Once everything lives in `vendor/`, all we'll have to change is `script/vendor` and some docs.

ttaylorr:
Yep, just need to re-add `code.google.com/p/log4go` and we should be back to normal 👍

technoweenie:
Do we? If everything is using `github.com/ThomsonReutersEikon/log4go` instead, we should probably just keep that. AFAIK only the ntlm package uses any log4go.

grahamking:
Looks like [this commit](https://github.com/ThomsonReutersEikon/go-ntlm/commit/b00ec39bbdd04f845950f4dbb4fd0a2c3155e830) removes `go-ntlm`'s dependency on log4go, so maybe you can drop that repo.

ttaylorr:
@grahamking great find! @technoweenie and I spent a while wrestling with glide before I 👀 this.

grahamking:
We tried a few Go package management tools, but ended up using none of them. We just copy the code into our `vendor/` directory, delete it's `.git` dir and `*_test.go` files, and commit that. Couldn't get much simpler.

To update you checkout the latest version of your dependency somewhere else and copy it into your vendor directory again. Easier than learning a new tool.

technoweenie:
Glide seems to be working mostly okay. The problem is that `go test ./...` started failing because it was testing all the vendored packages. That's why I filed a PR against your log4go fork to fix the imports from its `examples` package. Turned out that didn't matter because it couldn't compile the `examples/*.go` files into a single package.

```bash
../go/src/github.com/github/git-lfs/vendor/github.com/ThomsonReutersEikon/log4go/examples/FileLogWriter_Manual.go:17: main redeclared in this block
```

I'm really glad you removed the log4go dependency from `go-ntlm` though :metal:

sinbad:
Small point but you should update `script/lint` to not mention Nut and instead say to run `glide up`.

technoweenie:
Looks good. I didn't want to make any updates to dependencies in this PR, but the `go-ntlm` package made a single commit that removed `log4go` 🎉 

https://github.com/ThomsonReutersEikon/go-ntlm/compare/52b7efa603f1b809167b528b8bbaa467e36fdc02...b00ec39bbdd04f845950f4dbb4fd0a2c3155e830

PR #1247, opened 2016-05-24
git: fully qualify remote refs, use testify/assert

ttaylorr:


technoweenie:
Went with #1248, see https://github.com/github/git-lfs/pull/1248#issuecomment-221428764

PR #1248, opened 2016-05-24
Return a fully remote ref to reduce chances of ref clashes

technoweenie:
Returning simply `{remote}/{branch}` doesn't help if someone creates a local branch `refs/heads/master/{remote}/{branch}`. This `RemoteRefNameForCurrentBranch()` function is only used for `ResolveRef()`. `git rev-parse` can definitely handle full refs.

### DIFF
--- a/git/git_test.go
+++ b/git/git_test.go
@@ -66,7 +66,7 @@ func TestCurrentRefAndCurrentRemoteRef(t *testing.T) {
 
 	refname, err := RemoteRefNameForCurrentBranch()
 	assert.Equal(t, nil, err)
-	assert.Equal(t, "origin/someremotebranch", refname)
+	assert.Equal(t, "refs/remotes/origin/someremotebranch", refname)
 
 	remote, err := RemoteForCurrentBranch()
 	assert.Equal(t, nil, err)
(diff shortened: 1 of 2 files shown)


git/git_test.go
```diff
@@ -66,7 +66,7 @@ func TestCurrentRefAndCurrentRemoteRef(t *testing.T) {
 
 	refname, err := RemoteRefNameForCurrentBranch()
 	assert.Equal(t, nil, err)
-	assert.Equal(t, "origin/someremotebranch", refname)
+	assert.Equal(t, "refs/remotes/origin/someremotebranch", refname)
```
ttaylorr: Could use this as an excuse to swap out `github.com/technoweenie/assert` for `github.com/stretchr/testify/assert`, if you're cool with that 😄

ttaylorr:
Looks good! Great find.

technoweenie:
@ttaylorr Even though you beat me to it in #1247 (just _barely_ 🐌 ), I'm going with this one with the doc update and the use of sprintf. I think sprintf is easier to read (and it's shorter by 13c). The testify change is fine, but I think that should come in a single PR that 🔥  my `assert` package fork to the ground :)

ttaylorr:
Sounds good to me! I'll nuke your `assert` package either tonight or tomorrow.

PR #1250, opened 2016-05-25
Remove old asserts

ttaylorr:
This pull-request removes @technoweenie's `assert` library so that we only use one assertions library throughout Git LFS.


technoweenie:
To be clear, it's not my library. I just forked it to fix some bug :)

Looks good though :+1:

PR #1252, opened 2016-05-25
api{client,lifecycle}: use config.Endpoint to resolve root

ttaylorr:


technoweenie:
:+1: Now the new api client can use the proper LFS endpoint for a given remote by asking the `*config.Configuration`.

PR #1255, opened 2016-05-26
WIP transferqueue race fix

technoweenie:


technoweenie:
I wish I remembered why I fixed this :(

PR #1256, opened 2016-05-26
Lock commands

ttaylorr:
This pull-request is an incremental step towards the new lock related commands within the Git LFS client.

As an added benefit, these three new commands make extensive use of the new API client, so they serve as an example of its use.

##### `git lfs locks [-i id] [-p path] [-l limit] [--remote remote]`
The `git lfs locks` allows the client to list locks active on the server corresponding to the given filters, along with using the optional cursor and limit properties to paginate through a longer list of results.

##### `git lfs lock [--remote remote] <path>`
The `lock` command allows the client to create a lock against a particular file by making the appropriate API call to the server.

##### `git lfs unlock [--force] [--remote remote] <--id | path>`
The `unlock` command allows a client to break a lock created by any user against any particular file [in accordance with the original specification](#https://github.com/github/git-lfs/pull/1223).

## TODO

Still todo, are the following:

- [x] Implement the shell tests for these three new commands
  - [x] Implement the LFS locking API within the test server
- [x] Make the command that take a `<path>` argument resolve to a path relative to the root of a repository, such that they are consistent on the server.
- [x] Add optional --id param to unlock a lock according to its ID, and not the path.
- [x] Add optional --force param to force an unlock.
- [x] Add optional --remote param to all the lock commands to specify a different remote.


technoweenie:
This looks great. A couple suggestions for the cli api:

* Add optional `--id` param to unlock a lock according to its ID, and not the path.
* Add optional `--force` param to force an unlock.
* Add optional `--remote` param to all the lock commands to specify a different remote.

sinbad:
I'd recommend using the naming convention from `git log` to limit & skip in the results list:
```
-<number>, -n <number>, --max-count=<number>
       Limit the number of commits to output.
--skip=<number>
       Skip number commits before starting to show the commit output.
```
I just think it's better to use git conventions wherever possible, plus `cursor` is kind of an ambiguous term.

commands/command_lock.go
```diff
@@ -0,0 +1,89 @@
...
+
+	path, err := lockPath(args[0])
+	if err != nil {
+		Error(err.Error())
+	}
+
+	s, resp := API.Locks.Lock(&api.LockRequest{
+		Path:               path,
+		Committer:          api.CurrentCommitter(),
+		LatestRemoteCommit: latest.Sha,
```
sinbad: I'm still dubious about why you use the latest *remote* commit here. This is not necessarily the state of the local file. If the server agrees that this is the latest commit known and grants the lock, but this remote ref is *ahead* of the current HEAD and the file is changed, locking it in its current state and allowing the edit is guaranteed to cause a merge conflict, which is what we're trying to avoid. 

Really you need to do one of the following:

1. Pass the current HEAD commit SHA. Server should reject the lock unless it's the latest one it knows about which includes a change to this file. Or,
1. Pass the latest known remote ref as here, but the *client* has to check whether there are changes to this file between HEAD and that remote ref. At this point the user either needs to update their HEAD as a whole, or cherry pick the file from that remote ref into HEAD, or accept that they will create a conflict.

ttaylorr:
@sinbad:

> to limit & skip in the results list:

Good call. @technoweenie and I were discussing this, and it would definitely be silly to make every user write their own script to traverse through multiple pages of locks. As a first pass, we thought it may be good to have the client perform this traversal, and then temporarily recommend that you call `git lfs locks --limit=1000 | less`. `--skip` would be challenging to implement in this way, since we would either have to teach the server how to skip results, or have the client load all locks and cut off the first `n` skipped. We may not need this feature, but I will think more about this nonetheless.

commands/command_lock.go
ttaylorr: Hey, @sinbad. @technoweenie and I spent a little while today talking about what you had discussed here. We came up with a few different scenarios and were unable to find a common unifying "lock granting" implementation that a server could use.

The cases that we found are:

1. Client attempts to obtain a lock on a branch that both the client and server know about, and the client has no local changes. The lock should, in this case, be granted.

2. Client attempts to obtain a lock on a branch that both the client and server know about, but there is a non-empty diff between the `HEAD` and the latest remote ref. It was not clear whether or not the lock should be granted.

3. Client attempts to obtain a lock on a branch that the server does not know about, in which case an error should be returned from the API and the client should instruct the user to push that branch, then making it identical to the first case (see above).

As I mentioned above, we had a hard time determining what locking rules should be put in place on the server, and what the API would need to look like to support it. @technoweenie suggested that we should, as a first pass, simply have a list of locks, which are granted to clients as long as there are no other locks currently in place.

I agree with his suggestion, since no one else is using Git in this way, it would be best to see how this feature gets used in the wild during the experimentation phase, and then design a ruleset (and a corresponding API) to match.

Thoughts?

sinbad: Yeah, this is why I said there's branching cases both before and after the client requests the lock. In case 2 for example, the client will have to take action **before even requesting** the lock from the server, because it knows even before asking the server that the non-empty diff means there's a problem with obtaining a lock with the current working copy state. 

I covered some of the sequence I thought might be required, both pre-request (client), during request (server) and post-grant (client) in https://github.com/github/git-lfs/blob/master/docs/proposals/locking.md. But I agree that it's fine to concentrate on getting the basic locking API working on both ends first. I just wanted to flag up the conflict cases we're trying to address to make sure they can be dealt with. On reflection I think they can with this API, so long as the pre/post functionality is there. 

FWIW some other cases not covered above:

- Client tries to obtain a lock on a remote ref and thinks there's an empty diff, but server knows that client's record of the remote ref is out of date (behind) and there is a diff now.
- Client branch is tracking a remote ref and they both agree, but server has another remote ref which is a descendant which has a change to this file in it (lock and unlock happened separately). Granting the lock just based on this remote ref will cause a conflict later on.
- Client's tracked remote ref has a sibling branch on the server where changes have occurred to this file in parallel. Again even though the file is not locked, granting the lock will cause a conflict later. 

I talked about these cases in my proposal by suggesting the user would have to choose between switching branches, cherry-picking file state or accepting that they would have to resolve a conflict (discard one side) later. Some of these conditions have to be detected before requesting the lock (the server isn't aware of the working copy state so the client must act on any known issues there before requesting the lock), and some after (only the server knows the state of all remote branches).

We already know these cases exist so I don't think we need to wait until the community finds them, but at the same time they don't have to all be tackled right now. Just so long as we're aware of them this can be the next phase after the base API is working.

shana:
> As a first pass, we thought it may be good to have the client perform this traversal, and then temporarily recommend that you call git lfs locks --limit=1000 | less. --skip would be challenging to implement in this way, since we would either have to teach the server how to skip results, or have the client load all locks and cut off the first n skipped.

Just a reminder that on windows a user might not have a shell that allows for the type of piping and filtering that you can do on a unix shell (sad, but true).

Is there a template already for the output that these commands will produce?

commands/command_lock.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	wd, err := os.Getwd()
+	if err != nil {
+		return "", err
+	}
+
+	abs := filepath.Join(wd, file)
+
+	return strings.TrimPrefix(abs, repo), nil
```
technoweenie: I'd prefer truncating the leading `/` here. A path of `foo/bar/baz.zip` is relative to the root of the repository, and not the local hard drive. Also, should this check that the path is a file? Not sure what to do if someone tries to lock a directory, or even `.` (the entire repo?).

ttaylorr: Good call. I took care of stripping the leading slash in 8fb44158b51fdd9644a8f2bc4bc8b1afe2a6e6bf, and also added a client-side check to prevent locking directories 👍

technoweenie:
Hey @shana, sorry for ignoring your comment. Definitely should've replied before merging this :/

> Is there a template already for the output that these commands will produce?

There's no "official" template other than [printf calls like this](https://github.com/github/git-lfs/blob/d96c55328eb599cc6126e4ee078ae619e6bab0d8/commands/command_lock.go#L67). Are you asking because you're interested in integrating git lfs with some native git tool? We haven't thought about that much, and I'm personally open to whatever integrators think is best.

Would you prefer:

1. We add a `--format` flag with custom `%arg` values that integrators can use to customize output. This would protect you from changes to the command output.
2. How about we teach the command to optionally spit JSON out? Then a native tool could parse that, instead of relying on any text output.

Issue #1251, opened 2016-05-25
pull -I: / at end of path changes behavior


dipnlik:
`git lfs pull -I dir/` is different from `git lfs pull -I dir`: the first "doesn't work" (meaning it does not download any files in `dir`), while the second works as expected and downloads the requested files.

This is bad because other command line tools usually don't care about the final `/`, and because tab-completion usually includes the final `/`.

ttaylorr:
Hi @dipnlik! 👋  I applied a small refactoring to this part of the LFS code, and took note to fix your issue. It should be resolved once we merge #1257 😄

PR #1257, opened 2016-05-27
Config include exclude

ttaylorr:
@dipnlik reported #1251 the other day that the include/exclude filters made a distinction between `./foo/bar/` (with the trailing slash) and `./foo/bar` (without).

This pull-request introduces a patch that essentially runs filepath.Clean() against the paths before processing them. In addition, it removes the `command.determineIncludeExcludeFilters` function, and instead introduces a set of more granular replacement functions in the `config.Configuration` type. Those functions are all documented within the changeset.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -15,6 +15,7 @@ import (
 	"github.com/github/git-lfs/errutil"
 	"github.com/github/git-lfs/git"
 	"github.com/github/git-lfs/lfs"
+	"github.com/github/git-lfs/tools"
 	"github.com/spf13/cobra"
 )
 
@@ -208,6 +209,11 @@ type ErrorWithStack interface {
 	Context() map[string]string
 	InnerError() string
 	Stack() []byte
+}
+
+func determineIncludeExcludePaths(config *config.Configuration, includeArg, excludeArg string) (include, exclude []string) {
+	return tools.CleanPathsDefault(includeArg, ",", config.FetchIncludePaths()),
+		tools.CleanPathsDefault(excludeArg, ",", config.FetchExcludePaths())
 }
 
 func printHelp(commandName string) {
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -210,30 +210,6 @@ type ErrorWithStack interface {
 	Stack() []byte
 }
 
-// determineIncludeExcludePaths is a common function to take the string arguments
-// for include/exclude and derive slices either from these options or from the
-// common global config
-func determineIncludeExcludePaths(includeArg, excludeArg string) (include, exclude []string) {
-	var includePaths, excludePaths []string
-	if len(includeArg) > 0 {
-		for _, inc := range strings.Split(includeArg, ",") {
-			inc = strings.TrimSpace(inc)
-			includePaths = append(includePaths, inc)
-		}
-	} else {
-		includePaths = config.Config.FetchIncludePaths()
-	}
-	if len(excludeArg) > 0 {
-		for _, ex := range strings.Split(excludeArg, ",") {
-			ex = strings.TrimSpace(ex)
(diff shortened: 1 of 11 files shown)


ttaylorr:
As a side note, the `config.Configuration` type is quickly becoming a dumping ground for miscellaneous methods. I will try and spend this weekend thinking about alternative designs or patterns to refactor against, but am definitely more than open to suggestions on this as well.

For now, I think this is fine, as I assume it will be a pretty drawn-out process to preform this refactor, since so many parts of LFS depend on this type.

/cc @sinbad

commands/commands.go
```diff
@@ -210,30 +210,6 @@ type ErrorWithStack interface {
 	Stack() []byte
 }
 
-// determineIncludeExcludePaths is a common function to take the string arguments
-// for include/exclude and derive slices either from these options or from the
-// common global config
-func determineIncludeExcludePaths(includeArg, excludeArg string) (include, exclude []string) {
```
sinbad: I actually think this function was better where it was; it's specifically about converting string arguments to slices or defaulting and so belongs firmly unexported in `command` in my view. I don't see any benefit in promoting it to an exported function in `Configuration`, doing so widens that scope outside its remit IMO (which I think you were feeling too based on your follow up comment).

config/config.go
```diff
@@ -587,20 +588,68 @@ func (c *Configuration) readGitConfig(output string, uniqRemotes map[string]bool
...
+
+	for _, part := range strings.Split(paths, delim) {
+		part = strings.TrimSpace(part)
+
+		cleaned = append(cleaned, filepath.Clean(part))
+	}
+
+	return cleaned
+}
+
```
sinbad: I think these would be more at home in `command` or at most in `filetools.go` in the `tools` package (but really they're quite specifically about command line input).

sinbad:
👍 except for packaging comments

sinbad:
Oh, should have a test too.

ttaylorr:
@sinbad great points. I was definitely feeling the weight of `*config.Configuration`. I refactored according to your comments, and added tests as appropriate. Let me know your thoughts!

ttaylorr:
Looks like the `prune` tests are still flaky. Will investigate tomorrow. :/

sinbad:
Sorry, I should have been more specific: there's a test for the `--include` options in `test-fetch.sh`, you could just add a dir there and prove that using `--include="dir/"` works after your changes.

config/config.go
```diff
@@ -76,6 +78,29 @@ func NewConfig() *Configuration {
 	return c
 }
 
+// NewFromValues returns a new *config.Configuration instance as if it had
+// been read from the .gitconfig specified by "gitconfig" parameter.
+//
+// NOTE: this method should only be called during testing.
+func NewFromValues(gitconfig map[string]string) *Configuration {
+	config := &Configuration{
```
sinbad: You could actually put this in config_test.go to enforce the 'only testing' requirement.

sinbad:
Looks good to me, the prune failure is puzzling (auth error), I re-ran the test to be sure it wasn't just an intermittent thing but it happened again.

ttaylorr:
@sinbad yeah, not sure what's up with the prune failure, either. @technoweenie and I re-ran the tests in #1267 and the failures were intermittent, so you may have just gotten unlucky.

Thoughts on moving OS X `git-latest-go-1.5` test to the "allowed failures" section?

config/config.go
ttaylorr: Originally this `func` was in the `config_test.go` file, but I needed to use it in the [`commands_test.go`](https://github.com/github/git-lfs/pull/1257/files#diff-b65ac838606a4ab61fdb7724d458246dR11). This is sort of smelly to me, but I'm not quite sure what to do about it.

PR #1259, opened 2016-05-27
api: add Force option to unlock requests

ttaylorr:
This pull-request adds the relevant schema, tests, and type to support the ability for a client to indicate to an LFS server that it wishes to forcibly remove a particular lock.


ttaylorr:
Thanks! 👋

PR #1260, opened 2016-05-29
Implement support for GIT_SSH_COMMAND

### LINKED ISSUE #1142: Unclear how GIT_SSH env var works
With Git, when providing custom ssh commands with GIT_SSH, you always had to wrap the command in a shell script because Git would only allow a single argument to be passed. This lead to the introduction of [GIT_SSH_COMMAND](https://github.com/git/git/blob/master/Documentation/RelNotes/2.3.0.txt#L25) in version 2.3, which allowed arguments.

Which behaviour is correct when using GIT_SSH with git-lfs?

pdf:
Per the spec, `GIT_SSH_COMMAND` takes precedence over `GIT_SSH`.

Fixes #1142

### DIFF
--- a/auth/ssh.go
+++ b/auth/ssh.go
@@ -70,6 +70,12 @@ func sshGetExeAndArgs(endpoint config.Endpoint) (exe string, baseargs []string) 
 	isTortoise := false
 
 	ssh := config.Config.Getenv("GIT_SSH")
+	cmdArgs := strings.Fields(config.Config.Getenv("GIT_SSH_COMMAND"))
+	if len(cmdArgs) > 0 {
+		ssh = cmdArgs[0]
+		cmdArgs = cmdArgs[1:]
+	}
+
 	if ssh == "" {
 		ssh = "ssh"
 	} else {
@@ -82,7 +88,11 @@ func sshGetExeAndArgs(endpoint config.Endpoint) (exe string, baseargs []string) 
 		isTortoise = strings.EqualFold(basessh, "tortoiseplink")
 	}
 
-	args := make([]string, 0, 4)
+	args := make([]string, 0, 4+len(cmdArgs))
+	if len(cmdArgs) > 0 {
+		args = append(args, cmdArgs...)
+	}
+
 	if isTortoise {
 		// TortoisePlink requires the -batch argument to behave like ssh/plink
 		args = append(args, "-batch")
--- a/auth/ssh_test.go
+++ b/auth/ssh_test.go
@@ -11,6 +11,8 @@ import (
 func TestSSHGetExeAndArgsSsh(t *testing.T) {
 	endpoint := config.Config.Endpoint("download")
 	endpoint.SshUserAndHost = "user@foo.com"
+	oldGITSSHCommand := config.Config.Getenv("GIT_SSH_COMMAND")
+	config.Config.Setenv("GIT_SSH_COMMAND", "")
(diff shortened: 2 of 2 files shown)


pdf:
Not sure I can see why Travis borked the first run?

pdf:
One other thought I had is that perhaps we should avoid mangling the args at all if `GIT_SSH_COMMAND` is set, not sure which is the best route there.

hbons:
@pdf awesome! I'll try to compile this and feed it to SparkleShare. :)

sinbad:
Thanks for this! I agree to a degree with your last comment that when `GIT_SSH_COMMAND` is set the principle should be that args should only really come from that. However I would make the exception for the port argument, since that may only be known based on the URL which could vary and `GIT_SSH_COMMAND` is probably global. So maybe drop the `-batch` option with GIT_SSH_COMMAND and expect that the user will have provided it if they wanted it but leave the port argument?

pdf:
@sinbad the scenario I was considering was that someone may use `GIT_SSH_COMMAND` to invoke some binary that will barf if it gets passed the port arguments.  I don't believe `-batch` is an issue, because it's only added if the tortoise is matched, and the code in this PR avoids duplicating args if they already exist in `GIT_SSH_COMMAND`.

My main concern is that we blindly append the port arg, without testing that the command is `ssh`, but I'm hesitant to add that check because I'm not certain whether the existing code omitted the check because there are clients other than `ssh` that may expect the `-p` arg (eg, via `GIT_SSH`), or whether it was omitted just to save a couple of lines - I wouldn't like to break anyone's existing workflow.

sinbad:
Understood. I think the best thing is to do what git does, which is to always add the port. This is from `connect.c` in the git source:

```c
			ssh = getenv("GIT_SSH_COMMAND");
			if (!ssh) {
				const char *base;
				char *ssh_dup;

				/*
				 * GIT_SSH is the no-shell version of
				 * GIT_SSH_COMMAND (and must remain so for
				 * historical compatibility).
				 */
				conn->use_shell = 0;

				ssh = getenv("GIT_SSH");
				if (!ssh)
					ssh = "ssh";

				ssh_dup = xstrdup(ssh);
				base = basename(ssh_dup);

				tortoiseplink = !strcasecmp(base, "tortoiseplink") ||
					!strcasecmp(base, "tortoiseplink.exe");
				putty = tortoiseplink ||
					!strcasecmp(base, "plink") ||
					!strcasecmp(base, "plink.exe");

				free(ssh_dup);
			}

			argv_array_push(&conn->args, ssh);
			if (flags & CONNECT_IPV4)
				argv_array_push(&conn->args, "-4");
			else if (flags & CONNECT_IPV6)
				argv_array_push(&conn->args, "-6");
			if (tortoiseplink)
				argv_array_push(&conn->args, "-batch");
			if (port) {
				/* P is for PuTTY, p is for OpenSSH */
				argv_array_push(&conn->args, putty ? "-P" : "-p");
				argv_array_push(&conn->args, port);
			}
			argv_array_push(&conn->args, ssh_host);
```

pdf:
That assuages any doubts I had, thanks.  Think this should be good to go as is then.

pdf:
In fact, based on that logic, I can probably drop the duplicate arg handling, which will simplify the patch and reduce the test cases - if it wouldn't work in git, there's no problem producing the same behaviour here.

pdf:
Simplified as above (now rather trivial) and rebased against master.

sinbad:
Thanks!

PR #1262, opened 2016-05-30
Add checkout --unstaged flag

orivej:
This change introduces `--unstaged` ('-u') flag to the `checkout` subcommand. Originally it is intended to skip the costly `git update-index` call in the continuous integration setup where `git lfs checkout` is the last git operation performed in the clone, and the state of the index does not matter anymore.

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -20,6 +20,7 @@ var (
 		Use: "checkout",
 		Run: checkoutCommand,
 	}
+	checkoutUnstagedArg bool
 )
 
 func checkoutCommand(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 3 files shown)


orivej:
This change could potentially be more useful if it left oids intact, so that `git reset --hard` after `git lfs checkout -u` would revert it, rather than update the index.

levonet:
+1

technoweenie:
Nice :+1: What do you think about adding this to `git lfs pull` too, which is essentially `git lfs fetch` + `git lfs checkout` right after it?

ttaylorr:
:wave: This looks great, thanks @orivej. I'm happy to merge this as-is. If you'd like to add the `--unstaged` flag to `git fls pull`, that'd be great too, but not necessary. We can just merge this PR for now, and then come back to the `pull` command in a future revision.

Let me know what works for you! Thanks again.

orivej:
I'm going to rebase this pull request onto the master, and also add the flag to the pull command.

ttaylorr:
Hey, @orivej -- sorry I missed this. Can you pull in the latest changes from master again?

sinbad:
I've proposed reverting this for a couple of reasons discussed in #1352 - I think it should be resubmitted in another form.

arikon:
@orivej Did you have time to do it?

PR #1265, opened 2016-05-31
Enhanced transfers: part 1

### LINKED ISSUE #1121: Serious usability issues prompting for credentials with an separate LFS server
I have observed serious usability issues prompting for credentials, when using a separate LFS server.  In my case, I am using GitHub Enterprise, and a separate Artifactory server for LFS, though I don't believe those specifics would be necessary.  The usability issues are severe enough for me to consider the case "unusable" and a bug.

Once credentials are cached, everything is smooth.  However, if credentials have not been cached, prompting for them is basically usable.  The workaround for the situation is to ensure credentials are cached (or stored) via an "out-of-band" mechanism.  This isn't a good sell to potential Git LFS users, though. :disappointed: 

First, I appear to be prompted for the password once for each of the "ConcurrentTransfers" threads.  These prompts are all strung together on a single line, and I have to enter credentials over and over again.

Additionally, after entering a username, the prompt for a password appears, but is immediately overwritten with an update on the status.  Therefore, it's not clear that's there's even a prompt, much less what is required.

I've attached an image to attempt to demonstrate the usability issue in this situation.  My configuration for the example is simply:
```
credential.helper=cache
http.sslverify=false
lfs.concurrenttransfers=8
```

It is the output of ```git lfs fetch``` from a repository cloned with ```GIT_LFS_SKIP_SMUDGE=1```.  The progress being made is as I attempt to enter my username and password multiple times, but I'm not even clear what I'm being prompted for, so I eventually give up.  I don't want other users to go through the same thing!

image

sinbad:
Part 1 of supporting more flexible transfer mechanisms for upload & download. This PR changes no visible behaviour or APIs but prepares the ground for features like resumable transfers and binary chunking/diffing.

The major changes are:

1. the code for actually performing the uploads and downloads is now isolated in the `transfer` package
1. paths for uploading / downloading are abstracted via a `TransferAdapter`, and the existing path of non-resumable HTTP GET/PUT is implemented as `basicTransferAdapter`
1. TransferAdapter implementations are unidirectional, to allow for paths that can only do one or the other (e.g. tus.io is only upload, HTTP range headers are only download)
1. a registration system for adding new named transfer adapters is in place but currently unused, will be in the next stage
1. rationalisation of much of the API code so there is a clearer line between the API which requests URLs to upload/download, and the code which actually performs the upload/download. 
1. `TransferQueue` is still used since it orchestrates the end-to-end process, which includes calling the API. However it no longer has any transfer workers, it delegates that to `TransferAdapter`. 
1. Legacy code which will be removed when we drop the non-batch API route is clearly marked `// TODO LEGACY API`. Use of this code has been minimised so it's no longer generally mixed in even when you have batch support.
1. Fixes #1121 - this was caused by only the API part of the download/upload exchange waiting for a password prompt in serial, the actual transfers all kicked off in parallel so if the storage server had different credentials and had to prompt, all goroutines would prompt. The transfer system now handles this itself.
1. The same transfer code path is used for all cases, whether you're using e.g. `git lfs fetch` or `git lfs smudge`. Previously there were multiple paths.
1. Removed the `DownloadCheckable` and simplified things so it just uses the `dryRun` flag instead when probing the API for possible downloads without actually downloading. No other code used this anyway.
1. `Transferable` still exists but is now just a statement of intent in the `lfs` package for the entire process of queueing through the API and eventually being transferred if `dryRun=false`. It still needs to be there as an abstraction since the data comes from different places, but beyond data wrapping has no real functionality anymore barring a legacy hook which will get removed when we move to batch only. The simpler `Transfer` struct in the `transfer` package is used to schedule the actual transfers (doesn't share `Transferable` both to avoid import cycles and because the intent is different)

As mentioned this is only an internal change right now (albeit a fairly large one) and behaviour is unaffected. The next step is to use the adapter registration system to provide additional transfer routes, negotiated between client & server through optional new fields in the API.


api/api.go
```diff
@@ -21,6 +17,36 @@ import (
 	"github.com/rubyist/tracerx"
 )
 
+// BatchOrLegacy calls the Batch API and falls back on the Legacy API
+// This is for simplicity, legacy route is not most optimal (serial)
+// TODO LEGACY API: remove when legacy API removed
+func BatchOrLegacy(objects []*ObjectResource, operation string) ([]*ObjectResource, error) {
```
ttaylorr: How about a note in the documentation that says announces we set "lfs.batch" to false locally if the batch operation is unsupported?

lfs/batcher.go
```diff
@@ -11,16 +11,16 @@ import "sync/atomic"
 type Batcher struct {
 	exited     uint32
 	batchSize  int
-	input      chan Transferable
-	batchReady chan []Transferable
+	input      chan interface{}
```
ttaylorr: Why can't we use the `Transferable` type?

lfs/transfer_queue.go
```diff
@@ -17,45 +18,52 @@ const (
...
 	SetObject(*api.ObjectResource)
+	// Legacy API check - TODO remove this and only support batch
+	LegacyCheck() (*api.ObjectResource, error)
 }
 
-// TransferQueue provides a queue that will allow concurrent transfers.
+// TransferQueue organises the wider process of uploading and downloading,
+// including calling the API, passing the actual transfer request to transfer
+// adapters, and dealing with progress, errors and retries
 type TransferQueue struct {
```
ttaylorr: This type is getting to hold a lot of state. Are there perhaps smaller sub-types that we could pull out of TransferQueue, and then it could simply orchestrate those types?

api/api.go
sinbad: I actually want us to stop doing this in 1.3 & remove the legacy API. The persistent `lfs.batch=false` setting is a regular source of support calls when something has gone wrong either on the server, or some step in between and then persistently the LFS client refuses to use batch any more. If we keep `lfs.batch` in 1.3 then I'll push for adding a big warning banner in every fetch/push/pull that tells people `lfs.batch=false` and suggests removing it. What we do with batch/legacy is outside the scope of this PR though.

lfs/batcher.go
sinbad: The question is why it uses a specific type when this code has no need of it. I was going to re-use the `Batcher` for something else and may still do (and will probably move it to another package), but didn't need to in the end. But what was clear is that the Batcher simple takes a stream of individual objects and turns them into a slice of N size - it has nothing to do with `Transferable`.

lfs/transfer_queue.go
sinbad: Interestingly it was _because_ I pulled out the actual upload/download functionality to an external type that it ended up holding extra instance vars. While we could further split the type by changing the way the API part works, that wasn't what I needed to do for this PR - the API calling remains basically the same, it's the upload/download that is externalised. I suggest we do what you suggest when moving to your new API interface later.

ttaylorr: 👍

PR #1267, opened 2016-05-31
test: use noop credential helper for auth tests

ttaylorr:
Some tests are failing on branches of mine because the LFS test credential helper is being overridden by the OS X system-level cred-helper. Instead of `--unset`-ing during the tests that needed it, this PR introduces a no-op credential helper, which fixes those tests.

### DIFF
--- a/test/cmd/git-credential-lfsnoop.go
+++ b/test/cmd/git-credential-lfsnoop.go
@@ -1,0 +1,4 @@
+package main
+
+func main() {
+}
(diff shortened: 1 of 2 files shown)


technoweenie:
FWIW: the no-op test is implemented in Go just to fit in with the integration tests' build system. It compiles `*.go` files in that directory to the `./bin` directory, which gets added to the `PATH` for the duration of the tests.

PR #1273, opened 2016-06-01
ignore all osx ci failures

technoweenie:
Seems that all 3 of our osx builds have different failures associated with them. 

* https://travis-ci.org/github/git-lfs/jobs/134590303
* https://travis-ci.org/github/git-lfs/jobs/134590297
* https://travis-ci.org/github/git-lfs/jobs/134590299

These 3 seem like test related issues on travis, and not lfs bugs, so they're now ignored.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,6 @@ matrix:
   fast_finish: true
   allow_failures:
     - os: osx
-      go: 1.6
   include:
     - env: git-latest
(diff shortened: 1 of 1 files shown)


technoweenie:
Also, there is an issue to investigate that test failure: https://github.com/github/git-lfs/issues/1272. Would be nice for all the builds to be required, of course.

PR #1276, opened 2016-06-02
Allow additional fields on request & response schema

sinbad:
This is to allow us to extend the API with optional fields in future versions without breaking the schema.

We should notify server implementations to update their schema if they're using it to validate requests since we need to add optional fields for the transfer features in v1.3 even if they don't implement those features.

### DIFF
--- a/docs/api/http-v1-batch-request-schema.json
+++ b/docs/api/http-v1-batch-request-schema.json
@@ -23,6 +23,5 @@
       }
     }
   },
-  "required": ["objects", "operation"],
-  "additionalProperties": false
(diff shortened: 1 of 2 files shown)


ttaylorr:
Seems cool to me 👍 . As an aside, we'll want to migrate these files over to [the schema directory](https://github.com/github/git-lfs/tree/master/api/schema) and test them against the request payloads we generate.

sinbad:
I've re-run the Travis job multiple times but am still getting the OS X failures (no failures locally on OS X), they're turned off in master now of course. Merging this anyway.

Issue #1290, opened 2016-06-05
Debian build tests fail with missing vendor/github.com/xeipuuv/gojsonschema/json_schema_test_suite/type/schema_0.json


javabrett:
Recent changes to vendor deps might interact with the Debian build - at least I have not seen this one before.  Could also be my environment - someone can deb build master and confirm.

```
$ dpkg-buildpackage -us -uc
...
=== RUN   TestJsonSchemaTestSuite
Test (0) | integer type matches integers :: an integer is an integer
--- FAIL: TestJsonSchemaTestSuite (0.00s)
	schema_test.go:379: Error (open <redacted>/go/src/github.com/github/git-lfs/obj-x86_64-linux-gnu/src/github.com/github/git-lfs/vendor/github.com/xeipuuv/gojsonschema/json_schema_test_suite/type/schema_0.json: no such file or directory)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x8 pc=0x4b69bb]

goroutine 6 [running]:
panic(0x8073e0, 0xc820010170)
	/usr/local/go/src/runtime/panic.go:464 +0x3e6
testing.tRunner.func1(0xc8200201b0)
	/usr/local/go/src/testing/testing.go:467 +0x192
panic(0x8073e0, 0xc820010170)
	/usr/local/go/src/runtime/panic.go:426 +0x4e9
github.com/github/git-lfs/vendor/github.com/xeipuuv/gojsonschema.TestJsonSchemaTestSuite(0xc8200201b0)
	<redacted>/go/src/github.com/github/git-lfs/obj-x86_64-linux-gnu/src/github.com/github/git-lfs/vendor/github.com/xeipuuv/gojsonschema/schema_test.go:381 +0xe57b
testing.tRunner(0xc8200201b0, 0xabf958)
	/usr/local/go/src/testing/testing.go:473 +0x98
created by testing.RunTests
	/usr/local/go/src/testing/testing.go:582 +0x892
FAIL	github.com/github/git-lfs/vendor/github.com/xeipuuv/gojsonschema	0.010s
[...]

PR #1291, opened 2016-06-05
[WIP] Debian build vendor test excludes

javabrett:
Makes Debian build work on my machine, needs checks for correctness. #1290.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -12,7 +12,8 @@ endif
 
 BUILD_DIR := obj-$(DEB_HOST_GNU_TYPE)
 export DH_GOPKG := github.com/github/git-lfs
-export DH_GOLANG_EXCLUDES := test
+# DH_GOLANG_EXCLUDES typically incorporates vendor exclusions from script/test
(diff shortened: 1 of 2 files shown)


ttaylorr:
Looks good! I think it may be useful to have some vendor tests run on these builds, since we use them in `script/test` on Travis. To remove the failing ones, you could try something similar to [what we have currently](https://github.com/github/git-lfs/blob/bf9bc1f29907cdf26cf29eb01df8baa9f17b45c2/script/test#L11-L13), which is basically just a combination of `go list` and a bunch of `grep -v`s.

javabrett:
@ttaylorr You are right, and I had that initially in combination with `DH_GOLANG_INSTALL_ALL=1`.  Need to check the impact of that too, since a lot more files will be copied around and I don't know where they will end up, i.e. whether they pass through the build.

But I stopped when I hit the next package not on your list to fail, which was `github.com/spf13/cobra` - I should have added it, looks like it is the only remaining one to fail.

ttaylorr:
Looks good to me. Thanks! 👍

technoweenie:
LGTM :+1: 

I think a note should be added to `script/test` so we know to keep both excluded packages in sync.

PR #1292, opened 2016-06-06
Enforced a minimum gem version of 1.0.4 for packagecloud-ruby

javabrett:
Tested by:

* `sudo gem uninstall packagecloud-ruby`
* `sudo gem install packagecloud-ruby -v 1.0.2`
* `sudo gem install packagecloud-ruby`

... running `ruby script/packagecloud.rb` each time.  You get the version warning until you hit `1.0.4`, which is current.  Could make this `1.0.3`, the bare minimum, if preferred.

### DIFF
--- a/script/packagecloud.rb
+++ b/script/packagecloud.rb
@@ -8,9 +8,13 @@ end
 
 require "json"
 
+packagecloud_ruby_minimum_version = "1.0.4"
 begin
+  gem "packagecloud-ruby", ">=#{packagecloud_ruby_minimum_version}"
   require "packagecloud"
+  puts "Using packagecloud-ruby:#{Gem.loaded_specs["packagecloud-ruby"].version}"
 rescue LoadError
+  puts "Requires packagecloud-ruby >=#{packagecloud_ruby_minimum_version}"
   puts %(gem install packagecloud-ruby)
   exit 1
 end


script/packagecloud.rb
```diff
@@ -8,9 +8,13 @@
 
 require "json"
 
+packagecloud_ruby_minimum_version = "1.0.4"
 begin
+  gem "packagecloud-ruby", ">=#{packagecloud_ruby_minimum_version}"
   require "packagecloud"
```
ttaylorr: I'm not a huge fan of using the exit code on L19 as control flow. How about moving stuff out of the block to look like this:

```rb
begin
  gem "packagecloud-ruby", ">=#{pkg_cloud_ruby_min_version}"
  require "packagecloud"
rescue LoadError
  puts "Requires ..."
  exit 1
end

require "packagecloud"
```

javabrett: Thanks @ttaylorr for the feedback.  I am struggling a bit to apply it though.

Did you mean to move, rather than copy the `require "packagecloud"` statement, to be outside of the `rescue` block?  I myself found the ruby mechanism for forcing the specific version or version minimum confusing here, but I have intentionally left both `gem` and `require` statements inside the `begin/rescue/end` block.  The purpose of this block is to inform the developer that they don't have the required package/version, and fail-fast, preventing a deployment attempt.  My understanding is that you cannot enforce the version with `require`, and `gem` will test that, but the `require` is still required.

So, I don't know whether `gem` could pass but `require` fail - this might be what you are angling at - that the `require` no longer needs to be in the block.  I don't mind either way really - it looks OK to me now, but I'm no Ruby expert.

ttaylorr: Wow, huge oops on my end. Sorry about that, @javabrett.

It's fine as is, my suggestion should have omitted the last "require" statement outside of the begin/rescue block. I'm still not a _huge_ fan of using the `LoadError` for control flow, so if you'd like, you could use [`Gem#latest_spec_for`](http://apidock.com/ruby/Gem/latest_spec_for/class) instead, but either works for me 👍

ttaylorr:
Looks good to me 👍  If you can merge the latest changes from upstream's `master`, I'd be happy to merge this in.

PR #1297, opened 2016-06-08
Enhanced transfers: part 3 (resumable downloads)

sinbad:
This PR implements resumable downloads for all storage hosts which support HTTP Range/Content-Range exchanges. Since it seamlessly falls back to re-downloading the entire file if the server doesn't support a range request, it doesn't hurt to always ask to resume and to proceed depending on what the server responds with (often these servers will just return 200 instead of 206 and send the entire file, which this caters for).

### DIFF
--- a/transfer/basic_download.go
+++ b/transfer/basic_download.go
@@ -101,7 +101,7 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb TransferProgressCallback
 			return fmt.Errorf("Cannot restart %v from %d without a file & hash", t.Object.Oid, fromByte)
 		}
 		// We could just use a start byte, but since we know the length be specific
-		req.Header.Set("Range", fmt.Sprintf("bytes=%d-%d", fromByte, t.Object.Size))
+		req.Header.Set("Range", fmt.Sprintf("bytes=%d-%d", fromByte, t.Object.Size-1))
 	}
 
 	res, err := httputil.DoHttpRequest(req, true)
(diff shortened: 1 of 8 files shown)


transfer/basic_download.go
```diff
@@ -0,0 +1,240 @@
...
+		rangeRequestOk := false
+		var failReason string
+		// check 206 and Content-Range, fall back if either not as expected
+		if res.StatusCode == 206 {
+			// Probably a successful range request, check Content-Range
+			if rangeHdr := res.Header.Get("Content-Range"); rangeHdr != "" {
+				regex := regexp.MustCompile(`bytes (\d+)\-.*`)
+				match := regex.FindStringSubmatch(rangeHdr)
+				if match != nil && len(match) > 1 {
+					contentStart, _ := strconv.ParseInt(match[1], 10, 32)
```
benhumphreys: I suspect this should be `ParseInt(match[1], 10, 64)` otherwise you can't resume when more than 2^31 bytes have already been downloaded

sinbad: Oops, good catch!

transfer/basic_download.go
```diff
@@ -0,0 +1,242 @@
...
+	if err != nil {
+		return err
+	}
+
+	if fromByte > 0 {
+		if dlFile == nil || hash == nil {
+			return fmt.Errorf("Cannot restart %v from %d without a file & hash", t.Object.Oid, fromByte)
+		}
+		// We could just use a start byte, but since we know the length be specific
+		req.Header.Set("Range", fmt.Sprintf("bytes=%d-%d", fromByte, t.Object.Size))
```
benhumphreys: Should this be `t.Object.Size -1` ?
RFC 7233 seems to suggest the range is zero based inclusive. See RFC 7233 section 2.1.

sinbad: You're right, will fix. Although by fluke the spec also says it will be corrected even though I got it wrong ;D

PR #1298, opened 2016-06-08
Removed CentOS 5 from dockers.  Fixed #1295.

### LINKED ISSUE #1295: Remove centos_5 from docker build
It doesn't work:

```
# github.com/github/git-lfs/tools
/usr/bin/ld: unrecognized option '--build-id=none'
/usr/bin/ld: use the --help option for usage information
collect2: ld returned 1 exit status
```

Seems this is expected on account of go not supporting Centos 5: https://github.com/golang/go/issues/9520 .

Looks like Centos 5 packages are not built anymore, or at least they aren't pushed to packagecloud, so this option could be removed from the default list for `run_dockers.bsh` and `test_dockers.bsh`.

/cc @andyneff

javabrett:


ttaylorr:
Looks good to me, if you pull in the latest changes from master, I'd be happy to merge 👍

PR #1299, opened 2016-06-09
Add `lfs.basictransfersonly` option to disable non-basic transfer adapters

sinbad:
Will be useful as a workaround where there are server bugs & avoid having to retry across adapters without knowing whether the cause is transient or not.

Basically error handling with non-basic adapters gets a lot more complex if we try to allow fallbacks to other adapters when an adapter fails. We can't know if errors are due to a transient problem or whether there's a bug in the server implementation and that we should stop negotiating this adapter. Even if we could determine that we'd have to perform the re-negotiation again. 

Instead with this PR we assume that we continue to use the same definition of retriable errors and fatal errors, and abort if a non-basic adapter fails. If the user then knows it's because the server is buggy then they can enable this option to go back to the basic transfer adapter in all cases until it's fixed.


config/config.go
```diff
@@ -227,6 +227,21 @@ func (c *Configuration) ConcurrentTransfers() int {
 	return uploads
 }
 
+// BasicTransfersOnly returns whether to only allow "basic" HTTP transfers
```
ttaylorr: Perhaps add a note about the default value, or what happens if the parse fails?

config/config_test.go
```diff
@@ -336,6 +336,35 @@ func TestConcurrentTransfersNegativeValue(t *testing.T) {
...
 
+func TestBasicTransfersOnlySetValue(t *testing.T) {
+	config := &Configuration{
+		gitConfig: map[string]string{
+			"lfs.basictransfersonly": "true",
+		},
+	}
+
+	b := config.BasicTransfersOnly()
+	assert.Equal(t, true, b)
```
ttaylorr: I updated this suite to use the `github.com/stretchr/testify` package so you can use `assert.True` and `assert.False` here 😄

ttaylorr:
Minor comments, otherwise good to go 👍

config/config_test.go
sinbad: I'll leave it as-is since there's no real benefit & the rest of the tests in this file use this style.

PR #1303, opened 2016-06-14
Enhanced transfers part 4: resuming uploads via tus.io protocol

sinbad:
Adds a new "tus" transfer adapter which uses the [tus.io resumable upload protocol](http://tus.io/protocols/resumable-upload.html) (HEAD/PATCH methods). This allows a server to share upload action links to an endpoint which supports this protocol and therefore allow upload resuming. 

The included tests implement the server protocol too to prove it works, and I'm also submitting a PR to [lfs-test-server](https://github.com/github/lfs-test-server) which uses the official `tusd` server to prove it works using the official implementation and to demonstrate how you might go about using it.

### DIFF
--- a/test/cmd/lfstest-gitserver.go
+++ b/test/cmd/lfstest-gitserver.go
@@ -499,23 +499,21 @@ func storageHandler(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(404)
 	case "HEAD":
 		// tus.io
-		if len(r.Header.Get("Tus-Resumable")) == 0 {
-			log.Fatal("Missing Tus-Resumable header in request")
+		if !validateTusHeaders(r) {
 			w.WriteHeader(400)
 			return
 		}
 		parts := strings.Split(r.URL.Path, "/")
 		oid := parts[len(parts)-1]
+		var offset int64
 		if by, ok := largeObjects.GetIncomplete(repo, oid); ok {
-			w.Header().Set("Upload-Offset", strconv.FormatInt(int64(len(by)), 10))
-		} else {
-			w.Header().Set("Upload-Offset", "0")
+			offset = int64(len(by))
 		}
+		w.Header().Set("Upload-Offset", strconv.FormatInt(offset, 10))
 		w.WriteHeader(200)
 	case "PATCH":
 		// tus.io
-		if len(r.Header.Get("Tus-Resumable")) == 0 {
-			log.Fatal("Missing Tus-Resumable header in request")
+		if !validateTusHeaders(r) {
 			w.WriteHeader(400)
 			return
 		}
@@ -592,6 +590,15 @@ func storageHandler(w http.ResponseWriter, r *http.Request) {
 	default:
 		w.WriteHeader(405)
 	}
+}
+
+func validateTusHeaders(r *http.Request) bool {
+	if len(r.Header.Get("Tus-Resumable")) == 0 {
+		log.Fatal("Missing Tus-Resumable header in request")
+		return false
+	}
+	return true
+
 }
 
 func gitHandler(w http.ResponseWriter, r *http.Request) {
--- a/test/cmd/lfstest-gitserver.go
+++ b/test/cmd/lfstest-gitserver.go
@@ -311,7 +311,18 @@ func lfsBatchHandler(w http.ResponseWriter, r *http.Request, repo string) {
 
 	res := []lfsObject{}
 	testingChunked := testingChunkedTransferEncoding(r)
+	testingTus := testingTusUploadInBatchReq(r)
+	testingTusInterrupt := testingTusUploadInterruptedInBatchReq(r)
 	var transferChoice string
+	if testingTus {
+		for _, t := range objs.Transfers {
+			if t == "tus" {
+				transferChoice = "tus"
+				break
+			}
+
+		}
+	}
 	for _, obj := range objs.Objects {
 		action := objs.Operation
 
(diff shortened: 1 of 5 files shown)


test/cmd/lfstest-gitserver.go
```diff
@@ -592,8 +707,9 @@ func repoFromLfsUrl(urlpath string) (string, error) {
 }
 
 type lfsStorage struct {
-	objects map[string]map[string][]byte
-	mutex   *sync.Mutex
+	objects    map[string]map[string][]byte
+	incomplete map[string]map[string][]byte
+	mutex      *sync.Mutex
```
ttaylorr: Why do you need the pointer?

test/cmd/lfstest-gitserver.go
```diff
@@ -481,6 +497,98 @@ func storageHandler(w http.ResponseWriter, r *http.Request) {
...
+	case "HEAD":
+		// tus.io
+		if len(r.Header.Get("Tus-Resumable")) == 0 {
+			log.Fatal("Missing Tus-Resumable header in request")
+			w.WriteHeader(400)
+			return
+		}
+		parts := strings.Split(r.URL.Path, "/")
+		oid := parts[len(parts)-1]
+		if by, ok := largeObjects.GetIncomplete(repo, oid); ok {
```
ttaylorr: I would throw a `var offset int` in there, and then do the `w.Header().Set(...)` thing once. Since the zero-value of `int` is zero, this allows to you to remove the `else` case 😄

ttaylorr: As another note, [`strconv.Atoi`](https://golang.org/pkg/strconv/#Itoa) is shorthand for the `FormatInt` call, so you can save some characters.

test/cmd/lfstest-gitserver.go
```diff
@@ -481,6 +497,98 @@ func storageHandler(w http.ResponseWriter, r *http.Request) {
...
+			w.Header().Set("Upload-Offset", strconv.FormatInt(int64(len(by)), 10))
+		} else {
+			w.Header().Set("Upload-Offset", "0")
+		}
+		w.WriteHeader(200)
+	case "PATCH":
+		// tus.io
+		if len(r.Header.Get("Tus-Resumable")) == 0 {
```
ttaylorr: Not a big deal, but perhaps you could extract this to:

```go
func AssertTusResumableHeader(r *http.Request, w http.ResponseWriter) (present bool) {
        if len(r.Header.Get("Tus-Resumable")) == 0 {
                log.Fatal("Missing Tus-Resumable header in request")
                w.WriteHeader(http.StatusBadRequest)

                return false
        }

        return true
}
```

and use that? That method is sort of side-effect-y, but I'm generally not as concerned about that in the testserver code.

test/cmd/lfstest-gitserver.go
```diff
@@ -481,6 +497,98 @@ func storageHandler(w http.ResponseWriter, r *http.Request) {
...
+			log.Fatal("Missing Tus-Resumable header in request")
+			w.WriteHeader(400)
+			return
+		}
+		parts := strings.Split(r.URL.Path, "/")
+		oid := parts[len(parts)-1]
+
+		offsetHdr := r.Header.Get("Upload-Offset")
```
ttaylorr: nit: temporary variable.

test/cmd/lfstest-gitserver.go
```diff
@@ -481,6 +497,98 @@ func storageHandler(w http.ResponseWriter, r *http.Request) {
...
+		offset, err := strconv.ParseInt(offsetHdr, 10, 64)
+		if err != nil {
+			log.Fatal("Unable to parse Upload-Offset header in request: ", err)
+			w.WriteHeader(400)
+			return
+		}
+		hash := sha256.New()
+		buf := &bytes.Buffer{}
+		out := io.MultiWriter(hash, buf)
```
ttaylorr: Nice 👍

ttaylorr:
👍 , just some minor comments from me.

test/cmd/lfstest-gitserver.go
technoweenie: `strconv.Atoi` takes an int, not an int64, so I don't think that'll work for larger files.

test/cmd/lfstest-gitserver.go
sinbad: Which pointer? The Mutex is unchanged if that's the one you mean.

test/cmd/lfstest-gitserver.go
sinbad: TBH the test server can't handle files longer than 2gb anyway because it stores the data in slices; that's why `len(slice)` only returns `int` and not `int64`.

test/cmd/lfstest-gitserver.go
sinbad: Compromise: I've removed the side effect but generalised the validation.

test/cmd/lfstest-gitserver.go
sinbad: I'm not following - what's the issue?

sinbad:
Tests are currently failing because updating to master has pulled in a flaky test: #1334

test/cmd/lfstest-gitserver.go
ttaylorr: Sorry, this wasn't clear. I was suggesting that you inline the `r.Header.Get` into L530's `strconv.ParseInt` like:

```go
offset, err := strconv.ParseInt(r.Header.Get("Upload-Offset"), 10, 64)
```

But it was super minor anyway 😄

ttaylorr:
Welcome back, @sinbad, and thanks for the speedy fixes. This looks great, and I am really excited to get this into master. 👍

PR #1310, opened 2016-06-16
Ignore HEAD ref when fetching with --all

### LINKED ISSUE #1307: git lfs fetch --all fails in bare git repo
From my understanding, it should work. However, it fails with:
> Git can't resolve ref: "HEAD"

Problem is probably https://github.com/github/git-lfs/blob/master/commands/command_fetch.go#L54
git.CurrentRef() should be skipped when --all is specified

ttaylorr:
Since there is no HEAD ref to resolve when a repository is cloned with the `--bare` flag, the `git lfs fetch` command learned to ignore it. Kudos to @xen2 for bringing this up in #1307.

I piggybacked the pseudo-test for bare clones onto the end of the `fetch-all` test in `test/test-fetch.sh`. I am using all of the fixture data from the test, but if we want to extract that out, I'd be more than happy to do that 😄 

Either way, this pull-request should resolve #1307!

--------------

/cc @rubyist

### DIFF
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -50,7 +50,7 @@ func fetchCommand(cmd *cobra.Command, args []string) {
 			Panic(err, "Invalid ref argument: %v", args[1:])
 		}
 		refs = resolvedrefs
-	} else {
+	} else if !fetchAllArg {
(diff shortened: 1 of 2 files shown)


rubyist:
:+1:

PR #1323, opened 2016-06-23
Fix 'git lfs fetch' with a sha1 ref

omonnier:
The commit 39676192da53261efd441ccf0c67c85a341ae1a1
introduced a regression in the capability to fetch a
sha1 ref.
The command 'git lfs fetch origin <sha1>' was returning
following error: 'Invalid ref argument: [<sha1>]'.

This is due to the fact that 'git rev-parse --symbolic-full-name <ref>'
returns an empty content if the ref is a sha1.

Signed-off-by: Olivier Monnier <olivier.monnier@intel.com>

### DIFF
--- a/test/test-fetch.sh
+++ b/test/test-fetch.sh
@@ -80,7 +80,14 @@ begin_test "fetch"
   assert_local_object "$contents_oid" 1
   assert_local_object "$b_oid" 1
 
-  # test with commit sha1 specified
+  # test with master commit sha1 specified
+  rm -rf .git/lfs/objects
+  master_sha1=$(git rev-parse master)
+  git lfs fetch origin "$master_sha1"
+  assert_local_object "$contents_oid" 1
+  refute_local_object "$b_oid" 1
+
+  # test with newbranch commit sha1 specified
   rm -rf .git/lfs/objects
   newbranch_sha1=$(git rev-parse newbranch)
   git lfs fetch origin "$newbranch_sha1"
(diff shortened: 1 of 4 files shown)


test/test-fetch.sh
```diff
@@ -80,6 +80,13 @@ begin_test "fetch"
   assert_local_object "$contents_oid" 1
   assert_local_object "$b_oid" 1
 
+  # test with commit sha1 specified
+  rm -rf .git/lfs/objects
+  newbranch_sha1=$(git rev-parse newbranch)
+  git lfs fetch origin "$newbranch_sha1"
+  assert_local_object "$contents_oid" 1
+  assert_local_object "$b_oid" 1
```
ttaylorr: Do you think it may be helpful to add a commit to the remote that is a descendant of `$newbranch_sha1` and refute that the local object exists, just to make sure that this fetch doesn't just include everything?

ttaylorr:
Other than my minor comment, 👍

test/test-fetch.sh
omonnier: done in https://github.com/github/git-lfs/pull/1323/commits/7c246de3d6b268eb8ff6c87f47a07d126ddf05ab (if I correctly understood your comment)

ttaylorr:
Nice! Thanks for the quick update. If one more person can sign off I think this is good to merge 👍

ttaylorr:
Hey @omonnier, would you mind merging in the latest changes from upstream master into your branch? Thanks!

omonnier:
Hi @ttaylorr, my patches were rebased on upstream master

technoweenie:
Nice catch, looks good to me :metal:

ttaylorr:
Thanks for the update, @omonnier! Looks good to me. 👍

PR #1324, opened 2016-06-23
script/test: propagate extra args to go test

omonnier:
As described in the test/README.md, it is possible to
pass extra args to the script/test, like for example:
  $ script/test lfs -run TestSuccessStatus -v
Those args must be propagated to the 'go test' script.

Signed-off-by: Olivier Monnier <olivier.monnier@intel.com>

### DIFF
--- a/script/test
+++ b/script/test
@@ -4,7 +4,7 @@
 
 script/fmt
 if [ $# -gt 0 ]; then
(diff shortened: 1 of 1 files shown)


ttaylorr:
Thanks! 👍

Issue #1334, opened 2016-06-28
"checkout --unstaged" test is flaky


ttaylorr:
The `test: checkout --unstaged` test is flaky. I have not been able to reproduce the failure (yet) on my laptop, but it is currently breaking Travis CI builds on Go 1.6.

Here are some failures:

- https://travis-ci.org/github/git-lfs/jobs/140899419
- https://travis-ci.org/github/git-lfs/jobs/140900519
- https://travis-ci.org/github/git-lfs/jobs/140899428

/cc @orivej could you please take a look at these test and see if you can figure out what's causing them to fail in that particular environment? Thanks!

orivej:
Still investigating…

Meanwhile, you should disable `git diff-files` checks in the `checkout --unstaged` test as they are only meant to differentiate `checkout` and `checkout --unstaged` functionality, and the actual output of `git diff-files` is not interesting to the user.

javabrett:
Is `git diff-files` supposed to be empty, and therefore also not contain the string **file1.dat**?  If the name is not supposed to be there, `grep` in that case is going to exit 1 and by `set -e` the test fails.  I don't know why it passes sometimes - perhaps `git diff-files` completes so-fast with no output that `bash` doesn't even bother to create the pipe and start `grep`?

I may be totally misunderstanding the sense of the test - please confirm whether `diff-files` should be empty (exits 0), and therefore the grep should also fail.

orivej:
The test checks the behaviour of `checkout --unstaged` command that performs `checkout` skipping `git update-index`. Therefore large files should be present only in the working copy and not in the index, and `git diff-files` will show the difference between the working copy and the index. Normally during this test `git diff-files` outputs:

```
:100644 100644 5731942191cb145b334b73f1a2a5ae4346041780 0000000000000000000000000000000000000000 M  file1.dat
```

However, due to interaction with the smudge filter that I don't yet understand, after performing some git commands in the checkout, e.g. `git status`, large files appear to be indexed (and future `git diff-files` output nothing). Ideally after `checkout --unstaged` `git status` and other commands will not affect the apparent state of the index. (Or, even better but unlikely, it is possible to achieve the same result as with `checkout` without doing `git update-index`.)

I have not yet reproduced test failure locally, even with the version of git used in Travis.

ttaylorr:
@orivej I haven't been able to reproduce it locally either :-(. Are you able to debug from the output of one of the failing tests I listed above?

orivej:
[...]

PR #1335, opened 2016-06-29
docs/man: move "logs" subcommands from OPTIONS to COMMANDS

ttaylorr:
The `git-lfs-logs.1` man-page incorrectly listed the `clear` and `boomtown` subcommands of `git lfs logs` as `--options` rather than subcommands.

Closes github/git-lfs#1311


ttaylorr:
The 1.6 test is currently failing because of a flaky `checkout --unstaged` test. 

Tracking #1334.

PR #1344, opened 2016-07-01
cmd,doc,test: teach `git lfs track --{no-touch,verbose,dry-run}

ttaylorr:
This pull request introduces three new flags into the `git lfs track` command: `--no-touch`, `--verbose`, and `--dry-run`. It also updates the man pages and adds test for this new behavior accordingly.

- Specifying `--no-touch` will cause LFS to skip modifying the file's `atime` and `mtime`, leaving the file unmodified from the filesystem's perspective.

- Specifying `--verbose` will cause LFS to print the name of each file that it touches, so that it's more apparent what `git lfs touch` is doing when acting upon a large glob of files (see https://github.com/github/git-lfs/issues/1304).

- Specifying `--dry-run` will cause `git lfs track` to log out all of the actions it would normally preform (as given with `git lfs track --verbose`), without actually modifying files or `.gitattributes` on disk.

### DIFF
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -90,13 +90,17 @@ ArgsLoop:
 		// Since all `git-lfs track` calls are relative to the root of
 		// the repository, the leading slash is simply removed for its
 		// implicit counterpart.
-		Print("Searching for files matching pattern: %s", pattern)
+		if trackVerboseLoggingFlag {
+			Print("Searching for files matching pattern: %s", pattern)
+		}
 		gittracked, err := git.GetTrackedFiles(pattern)
 		if err != nil {
 			LoggedError(err, "Error getting git tracked files")
 			continue
 		}
-		Print("Found %d files matching pattern: %s", len(gittracked), pattern)
+		if trackVerboseLoggingFlag {
+			Print("Found %d files previously added to Git matching pattern: %s", len(gittracked), pattern)
+		}
 		now := time.Now()
 
 		var matchedBlocklist bool
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -25,6 +25,9 @@ var (
 		Use: "track",
 		Run: trackCommand,
 	}
+
+	trackNoTouchFlag        bool
+	trackVerboseLoggingFlag bool
 )
 
 func trackCommand(cmd *cobra.Command, args []string) {
@@ -114,11 +117,18 @@ ArgsLoop:
 		}
 		Print("Tracking %s", pattern)
 
-		for _, f := range gittracked {
-			err := os.Chtimes(f, now, now)
-			if err != nil {
-				LoggedError(err, "Error marking %q modified", f)
-				continue
+		if !trackNoTouchFlag {
+			now := time.Now()
+			for _, f := range gittracked {
+				if trackVerboseLoggingFlag {
+					Print("Git LFS Track: touching %s", f)
+				}
+
+				err := os.Chtimes(f, now, now)
+				if err != nil {
+					LoggedError(err, "Error marking %q modified", f)
+					continue
+				}
 			}
 		}
 	}
@@ -217,5 +227,8 @@ func blocklistItem(name string) string {
 }
 
 func init() {
+	trackCmd.Flags().BoolVarP(&trackNoTouchFlag, "no-touch", "n", false, "skip modifying files matched by the glob")
+	trackCmd.Flags().BoolVarP(&trackVerboseLoggingFlag, "verbose", "v", false, "log which files are being tracked and modified")
+
 	RootCmd.AddCommand(trackCmd)
 }
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -26,7 +26,6 @@ var (
 		Run: trackCommand,
 	}
 
-	trackNoTouchFlag        bool
 	trackVerboseLoggingFlag bool
 	trackDryRunFlag         bool
 )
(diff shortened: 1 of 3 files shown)


ttaylorr:
Tests are failing because of calling `stat` on Travis infra incorrectly. Will take a look after 🍔

ttaylorr:
🍔  was delicious. Fixed tests and added the `--dry-run` flag.

docs/man/git-lfs-track.1.ronn
```diff
@@ -3,14 +3,35 @@ git-lfs-track(1) - View or add Git LFS paths to Git attributes
...
+
+* `--no-touch` `-n`:
+  If enabled, prevent `git lfs track` from touching files matched by the
+  provided pattern glob. Disabled by default.
+
+* `--verbose` `-v`:
+  If enabled, have `git lfs track` log files which it will touch. Disabled by
+  default.
+
+* `--dry-run` `d`:
```
technoweenie: `-d`

docs/man/git-lfs-track.1.ronn
```diff
@@ -3,14 +3,35 @@ git-lfs-track(1) - View or add Git LFS paths to Git attributes
...
+  If enabled, have `git lfs track` log all actions it would normally take
+  (adding entries to .gitattributes, touching files on disk, etc) without
+  preforming any mutative operations to the disk.
```
technoweenie: `performing`

technoweenie:
This looks good. However, it's missing a message for a basic `git lfs track {pattern}` call, like `Modifying %d files matching pattern %q`.

ttaylorr:
Good call on adding that extra messages, @technoweenie 👍  I got those in in fdb8f7ff390aaa62742f443d490d5f4ad366662c 😄

ttaylorr:
New `track` tests are passing. Looks like the `checkout --unstaged` test is broken on go1.6.

/cc #1334

commands/command_track.go
```diff
@@ -74,32 +78,43 @@ ArgsLoop:
...
-		// deliberately not done in parallel as a chan because we'll be marking modified
-		gittracked, err := git.GetTrackedFiles(pattern)
-		if err != nil {
-			LoggedError(err, "Error getting git tracked files")
-			continue
-		}
-		now := time.Now()
-		for _, f := range gittracked {
-			err := os.Chtimes(f, now, now)
+		if !trackNoTouchFlag || trackDryRunFlag {
```
sinbad: If `--no-touch` is used then we should still be checking whether there are matching files which are already added to git, because failing to do this will cause nasty historical problems if the `.gitattributes` is committed without also re-committing the affected tracked files. 

At an absolute minimum it should warn the user that this is the case, but really I think it should abort. Not `touch`ing the files just causes too many unfixable issues (without history re-writing) otherwise.

sinbad: I actually question whether the `--no-touch` option is a good idea at all in fact. `track` only touches files when it absolutely has to, and *not* touching them in this case results in really nasty problems down the line, because `.gitattributes` wants to run the filters but the content of the files in the git commit doesn't reflect that. The only case where you'd be avoiding touching files with this option will probably cause more problems than it solves.

sinbad:
👍 to `--verbose` and `--dry-run`, but 👎 to `--no-touch` IMO

ttaylorr:
Hey, @sinbad - thanks for your thoughts. I agree with what you're saying and went ahead and removed the `--no-touch` option from the track command.

😄

commands/command_track.go
```diff
@@ -87,11 +90,13 @@ ArgsLoop:
 		// Since all `git-lfs track` calls are relative to the root of
 		// the repository, the leading slash is simply removed for its
 		// implicit counterpart.
+		Print("Searching for files matching pattern: %s", pattern)
```
sinbad: I think I'd only print this in verbose mode.

commands/command_track.go
```diff
@@ -87,11 +90,13 @@ ArgsLoop:
...
 		gittracked, err := git.GetTrackedFiles(pattern)
 		if err != nil {
 			LoggedError(err, "Error getting git tracked files")
 			continue
 		}
+		Print("Found %d files matching pattern: %s", len(gittracked), pattern)
```
sinbad: I'd only print this in verbose mode, and I'd make it more specific that it's talking about files *already added* to git. Otherwise it would sound weird if you hadn't added any files to git yet but had files on disk with that pattern and it said "Found 0 files".

ttaylorr:
@sinbad works for me 👍 

As a thought, it would be interesting to add the `--verbose` flag to the root `git lfs` command and then introduce a `PrintVerbose` (or similarly named) function in the `commands` package to call `Print` only when given the  `--verbose` flag.

Not sure where we should draw the boundary between `--verbose` logs and using the `tracerx` package, so I'm fine with just implementing this in the `track` command for now.

sinbad:
Yeah, I think having a `PrintVerbose` and a global option would be good. `tracerx` is more for our debugging purposes although the line is somewhat blurry.

technoweenie:
Isn't this still vulnerable to the original issue? Running `git lfs track "*.psd"` with no flags can still take a _long_ time, with no hints to the user why. My original idea was to add a simple `Running all "*.psd" files through Git LFS`. Something to indicate to the user what's taking so long.

ttaylorr:
@technoweenie I was thinking giving callers the option of passing `--verbose` would be good if they were curious what was taking so long. I think being more explicit about our logging could be good, too, so I'd be fine reverting https://github.com/github/git-lfs/pull/1344/commits/26d3cc1571699d172c1272e9b47d9b9ee1f450a4. 😄

technoweenie:
I just thought `--verbose` was only going to be used when listing _all_ the files as they're touched. Guess the lines got crossed somewhere.

Maybe instead of this:

> Found %d files previously added to Git matching pattern: %s

How about:

> Adding %d files matching pattern %q to Git LFS

sinbad:
Ah, if the issue is about the time the `git ls-files` is taking, maybe always print the "Searching for files matching pattern: %s" even if non-verbose. Maybe even explain it more "Searching for files already added to git matching pattern: %s"

> Adding %d files matching pattern %q to Git LFS

Hmm, the number is just the files which are being added to Git LFS that are already added to git as non-LFS files - it's not all the files which are being tracked in LFS. It doesn't count any files which are on disk and which match the pattern but never added to git before. The point of this scan is to identify files which are already in git so need to be re-added after being changed to LFS files.

technoweenie:
Also, you should remove the `no-touch` references from the docs :) I think the flag was my idea, but I also fully support @sinbad's reply above about removing it.

ttaylorr:
@technoweenie Thanks, I forgot to remove it from this PR before I merged it, but I added a commit on master to remove: https://github.com/github/git-lfs/commit/2a420e13355c29c057fb932e9f8806e38d0a6de4

PR #1346, opened 2016-07-01
Safe track patterns

ttaylorr:


technoweenie:
Found a case where `filepath.Match()` doesn't behave exactly like Git:

```go
package main

import (
	"fmt"
	"path/filepath"
)

func main() {
    	files := []string{"foo/.gitattributes", "foo/bar/.gitattributes"}
	patterns := []string{"foo/*", "foo/**/*"}
	
	for _, pattern := range patterns {
		for _, file := range files {
			m, err := filepath.Match(pattern, file)
			fmt.Printf("%s on %s: %v (%s)\n", pattern, file, m, err)
		}
	}
}
```

https://play.golang.org/p/9rshqsIqXYf

Here it is with Git:

```bash
$ tree -a foo
foo
├── .gitattributes
└── bar
    └── .gitattributes

$ git -c core.quotepath=false ls-files --cached -- "foo/*"
foo/.gitattributes
foo/bar/.gitattributes

$ git -c core.quotepath=false ls-files --cached -- "foo/**/*"
foo/bar/.gitattributes
```

I don't want to try replicating git's match behavior in Go, as attractive as a shaved yak as that would be. I think a better idea would be to [get the tracked files](https://github.com/github/git-lfs/blob/7ee533f646ea7d76afcd073556c445bab420aa2b/commands/command_track.go#L85-L89) earlier in the command, and look for any file entries with a base name starting with `.git` or `.lfs`.

We may want come up with a test case containing a million files to see what the experience is like on such a large repository.

ttaylorr:
I did not know of the differences between `filepath.Match` and Git's internal filename filters. Good catch, @technoweenie.

I think the prefix-checking solution is a good one. I went ahead and implemented it in 6c5f738ee5b6df3989151e51d8033822c63b4f67, and ran benchmarks on it as compared to my prior commit, with a repository containing 10,000 files.

10,000 is rather small as compared to your original suggestion of 1,000,000, but `touch`-ing a million files in a row takes quite a bit of time on my computer 😜.

It looks like using `filepath.Match` is somewhat faster than checking all of the results from Git, but about 60 milliseconds. I assume this will trend fairly linearly with number of files, but this difference is rather small (for now).

In my opinion, I would gladly trade 60msec for parity with Git's filepath filters, but that's just me 😄

Here are the benchmarks in their entirety:

## Using Git with prefix matching:
| User | System | CPU | Total | 
|------|--------|-----|-------| 
| 0.05 | 0.11   | 95  | 0.165 | 
| 0.04 | 0.11   | 94  | 0.163 | 
| 0.04 | 0.12   | 88  | 0.184 | 
| 0.04 | 0.11   | 97  | 0.155 | 
| 0.04 | 0.11   | 93  | 0.167 | 

### CPU Summary
| Max | Mean | Median | 
|-----|------|--------| 
| 97  | 93.4 | 94     | 

### Total Summary
| Max   | Mean  | Median | 
|-------|-------|--------| 
| 0.184 | 0.167 | 0.165  | 


## Using `filepath.Match`
| User | System | CPU | Total | 
|------|--------|-----|-------| 
| 0.04 | 0.06   | 97  | 0.100 | 
| 0.03 | 0.07   | 98  | 0.105 | 
| 0.04 | 0.08   | 98  | 0.121 | 
| 0.04 | 0.07   | 98  | 0.117 | 
| 0.03 | 0.05   | 100 | 0.079 | 

### CPU Summary
| Max | Mean | Median | 
|-----|------|--------| 
| 100 | 98.2 | 98     | 

### Total Summary
| Max   | Mean  | Median | 
|-------|-------|--------| 
| 0.121 | 0.104 | 0.105  |

commands/command_track.go
```diff
@@ -17,6 +17,10 @@ import (
 )
 
 var (
+	blacklistedPrefixes = []string{
+		".git", ".lfs", // TODO
```
technoweenie: Do we need this todo anymore?

ttaylorr: I don't think so. I left it in to see if there were any other prefixes that we wanted to block out, but I think this list is fine for now 👍

technoweenie:
I'd prefer correctness (calling git directly) over a small speed increase. However, I think the check should happen before updating `.gitattributes`.  The order of track operations now looks like:

1. [Write encoded value to .gitattributes](https://github.com/github/git-lfs/blob/e34f7b33e14451db947c3be628112c70618b8567/commands/command_track.go#L81-L86)
2. [Get tracked files](https://github.com/github/git-lfs/blob/e34f7b33e14451db947c3be628112c70618b8567/commands/command_track.go#L93-L97)
3. [Skip touching files matching an entry in the block list](https://github.com/github/git-lfs/blob/e34f7b33e14451db947c3be628112c70618b8567/commands/command_track.go#L100-L103)
4. [Touch files](https://github.com/github/git-lfs/blob/e34f7b33e14451db947c3be628112c70618b8567/commands/command_track.go#L105-L109)

I think the order should be more like this:

1. Get tracked files
2. Cancel command if any tracked files match something in the block list.
3. Update .gitattributes
4. Touch tracked files

ttaylorr:
@technoweenie good call! I addressed this and re-ordered everything to fit in 3ad1222cf1552988402f9c87201ee06070adb730. Much smoother now ✨ 

As an aside, `git ls-files` seems to have some trouble listing absolute paths unless they are already present in that repo's `.gitattributes` file. Since fetching a list of tracked files now happens before updating .gitattributes (steps 1, and 3, respectively) I went ahead and stripped the leading slash in the call to `ls-files`. As discussed, this has the same effect since all `git lfs track` calls are relative to the absolute root of the repository. 😄

.gitattributes
```diff
@@ -1,3 +1,4 @@
 * text=auto
 * eol=lf
 *.bat eol=crlf
+/foo filter=lfs diff=lfs merge=lfs -text
```
technoweenie: Is this just for testing? I don't see it referenced anywhere.

technoweenie:
:ship:

.gitattributes
ttaylorr: My bad, must have slipped from my own testing.

PR #1349, opened 2016-07-04
Safety check to `comands.requireStdin`

ttaylorr:
This pull-request introduces an additional safety check around the `requireStdin` function of the `commands` package as seen [here](https://github.com/github/git-lfs/blob/79bedebb2536485317ed5470d4e6d72cdc92d0c4/commands/commands.go#L116).

As reported in https://github.com/github/git-lfs/issues/1347, the function was panicking when trying to call Mode on a nil `os.FileInfo`. Since `os.File.Stat` is guaranteed to return a non-nil value so long as no error is present, a sanity check was added to catch and log these errors. If no error was present, the bitmask is applied and checked as normal.

An aside: I added the error logging in my second commit, whereas it was previously simply discarding the error as before. If we want to continue to discard the blanked error from `os.Stdin.Stat()`, then we can simply not apply the second commit contained in this PR.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -120,9 +120,17 @@ func PipeCommand(name string, args ...string) error {
 }
 
 func requireStdin(msg string) {
-	stat, _ := os.Stdin.Stat()
-	if stat == nil || (stat.Mode()&os.ModeCharDevice) != 0 {
-		Error("Cannot read from STDIN. %s", msg)
+	var out string
+
+	stat, err := os.Stdin.Stat()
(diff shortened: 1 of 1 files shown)


ttaylorr:
CI is failing again due to the `checkout --unstaged` test.

Tracking: https://github.com/github/git-lfs/issues/1334

technoweenie:
:+1:  Quite silly of me to ignore the error like that.

ttaylorr:
Thanks for the review! 👋

PR #1350, opened 2016-07-05
⏳ Retry transfers with expired actions

ttaylorr:
This pull request eagerly retries transfers that come back with expired actions, per https://github.com/github/git-lfs/issues/1345.

This is still a "work-in-progress", as I'm not 100% certain that I've implemented this in the right place. I followed the upload from `commands/command_push.go`, into the LFS code and the transfer code, into the `handleTransferResult`, which I think is where we want to be.

Just in case, I've left some `// NOTE`s throughout my code, which I would very much appreciate the validation of from any potential reviewers (/cc @sinbad).

Pending that validation, I'll go ahead and implement a change in the test server to validate that this behavior is used correctly and that requests are indeed retried. Since that component of the test server touches a good portion of our tests, I wanted to make sure that my fix is correct before going in and modifying a critical path in the test server.

### DIFF
--- a/transfer/adapterbase.go
+++ b/transfer/adapterbase.go
@@ -1,8 +1,11 @@
 package transfer
 
 import (
+	"fmt"
 	"sync"
+	"time"
 
+	"github.com/github/git-lfs/errutil"
 	"github.com/rubyist/tracerx"
 )
 
@@ -103,8 +106,15 @@ func (a *adapterBase) worker(workerNum int) {
 			}
 		}
 		tracerx.Printf("xfer: adapter %q worker %d processing job for %q", a.Name(), workerNum, t.Object.Oid)
+
 		// Actual transfer happens here
-		err := a.transferImpl.DoTransfer(t, a.cb, authCallback)
+		var err error
+		if t.Object.IsExpired(time.Now()) {
+			tracerx.Printf("xfer: adapter %q worker %d found job for %q expired, retrying...", a.Name(), workerNum, t.Object.Oid)
+			err = errutil.NewRetriableError(fmt.Errorf("lfs/transfer: object %q has expired", t.Object.Oid))
+		} else {
+			err = a.transferImpl.DoTransfer(t, a.cb, authCallback)
+		}
 
 		if a.outChan != nil {
 			res := TransferResult{t, err}
--- a/transfer/adapterbase.go
+++ b/transfer/adapterbase.go
@@ -9,6 +9,12 @@ import (
 	"github.com/rubyist/tracerx"
 )
 
+const (
+	// objectExpirationGracePeriod is the grace period applied to objects
+	// when checking whether or not they have expired.
+	objectExpirationGracePeriod = 5 * time.Second
+)
+
 // adapterBase implements the common functionality for core adapters which
 // process transfers with N workers handling an oid each, and which wait for
 // authentication to succeed on one worker before proceeding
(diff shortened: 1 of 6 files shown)


sinbad:
Glad you're taking a look at this! Sadly not the right place :)

The primary case of `expires_at` being an issue is that time has lapsed between making the API call and getting around to instigating the transfer that was returned from it, most likely because there was a lot of data before, the connection wasn't fast enough, and an auth token will expire. To correctly handle this case requires that you request a new upload/download link from the API before making a request to the transfer system. 

As at https://github.com/github/git-lfs/pull/1350/commits/81ce51728adb963a97a44d708b3e909ac2af14c5 you're checking in `handleTransferResult`, which is the result of the transfer system, so this is after the request has already been made (using potentially expired link data). Also you're only checking on success, when in fact if the auth token was going to expire for example, the transfer system would have errored so this code will never be hit.

However the way you're retrying is correct, since `retry()` retries the entire process, starting at the batcher which calls the API again. The question is how to do that just before starting each transfer item, rather than just after as with `handleTransferResult`. That code is actually in the transfer adapters, not in the transfer queue, since the time must be compared at the exact moment a queued transfer is about to be invoked. Specifically, you'll find that code in `adapterbase.go` (excerpted for clarity):

```
func (a *adapterBase) worker(workerNum int) {
....
	for t := range a.jobChan {
...
		// Actual transfer happens here
		err := a.transferImpl.DoTransfer(t, a.cb, authCallback)
```
Just before calling `DoTransfer` you should check `t.Object` to see if the link is expired (and I would personally check if it's going to expire in say the next ~10s to avoid edge cases), and if it is you can immediately add a retriable error result to `a.outChan` instead of proceeding.

transfer/adapterbase.go
```diff
@@ -103,8 +106,15 @@ func (a *adapterBase) worker(workerNum int) {
 			}
 		}
 		tracerx.Printf("xfer: adapter %q worker %d processing job for %q", a.Name(), workerNum, t.Object.Oid)
+
 		// Actual transfer happens here
-		err := a.transferImpl.DoTransfer(t, a.cb, authCallback)
+		var err error
+		if t.Object.IsExpired(time.Now()) {
```
sinbad: As suggested earlier I would add a few seconds grace to `time.Now()` before passing it to `IsExpired` to avoid the case where the link is not expired exactly when you test, but is by the time the request goes through.

ttaylorr:
Currently trying to figure out why retried transfers increment the progress meter's count by calling `Add()`. Right now the output of pushing a single file with a retry is:

```
Git LFS: (2 of 1 files) ...
```

which is incorrect. I left a `TODO` in my latest push, but am going to return to this later.

ttaylorr:
Woo-hoo! @rubyist provided some invaluable pointers in finding the source of the `(2 of 1 files)` bug. Looks like the progress meter's `Finish` method was being called even if an error was encountered, which was incrementing the internal finished count.

By moving the Finish call into the path where no error occurred (as in https://github.com/github/git-lfs/pull/1350/commits/ea60ea048d3845dcc7bd6b49342ee4b1787bf0c7) fixed it!

Ready to merge 😄

sinbad:
I'd still like a grace few seconds on the expired time test but otherwise 👍

ttaylorr:
Thanks for the help and review, @sinbad!

PR #1352, opened 2016-07-05
Revert "Add checkout --unstaged flag"

### LINKED ISSUE #1334: "checkout --unstaged" test is flaky

sinbad:
Fixes #1334
Reverts github/git-lfs#1262

Reverting for possible re-submission later. The reasons are:

1. Failures in the test are causing other subsequent PRs to fail. The reason is that skipping the `update-index` is unreliable at creating a clean git state, because the index cache may be out of date if the timestamp of files changes during checkout. Depending on the exact timing, the version of git, and the file system, precision issues can result in a non-clean status without `update-index`.

1. I think the flag `--unstaged` is misleading and the documentation "Do not add files to the index, keeping them unstaged." is incorrect. `checkout` never added anything to the index, it only made sure the index cache was not out of date.

If the intention is to save the expense of calling `update-index` then perhaps this feature can come back later in the form of a `--no-update-index` flag or something, with a more accurate description of what it does; including mentioning that although it's faster, it might prevent the state returning to clean correctly in some cases.

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -20,7 +20,6 @@ var (
 		Use: "checkout",
 		Run: checkoutCommand,
 	}
-	checkoutUnstagedArg bool
 )
 
 func checkoutCommand(cmd *cobra.Command, args []string) {
@@ -44,7 +43,6 @@ func checkoutCommand(cmd *cobra.Command, args []string) {
 }
 
 func init() {
-	checkoutCmd.Flags().BoolVarP(&checkoutUnstagedArg, "unstaged", "u", false, "Do not add files to the index")
 	RootCmd.AddCommand(checkoutCmd)
 }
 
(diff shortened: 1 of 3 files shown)


technoweenie:
:+1: Thanks for clarifying what `update-index` is doing.

ttaylorr:
Works for me, thanks for the explanation 👍

ttaylorr:
Tracking: #1334

arikon:
@orivej Could you resubmit github/git-lfs#1262 in another form, please?

PR #1358, opened 2016-07-08
use proxy from git config 

technoweenie:
This updates #1173 with the latest master. Since that PR, a lot of subpackages have been split out, and we've switched assertion libraries. 

@jonmagic @LizzHale: I also changed the main `ProxyFromGitConfigOrEnvironment()` function from #1173 for these reasons:

1. `proxyFromGitConfigOrEnvironment()` is coupled to the global `Config` value. I updated the function to accept only a `*config.Configuration` argument, and return a `func(*http.Request) (*url.URL, error)`. That cleaned up the test setup for `TestProxyFromGitConfig` a bit.
2. It returns a valid url OR error from `http.ProxyFromEnvironment` now. I didn't want to drop a potential error if users have a bad HTTPS_PROXY value, but a good git `http.proxy` value.
3. The end of the function returns `nil` now, explicitly.

Unfortunately, I couldn't test the environment, because of [how `http.ProxyFromEnvironment` is implemented](https://golang.org/src/net/http/transport.go?s=7918:7975#L202). It reads from the env the _first_ time that function is called. Calling `os.Setenv` doesn't affect it.

However, I'm thinking about pulling all that code into Git LFS. There are a few internal functions that would be useful for proxy values specified in the git config, such as `NO_PROXY` handling.

### DIFF
--- a/lfs/proxy_test.go
+++ b/lfs/proxy_test.go
@@ -2,7 +2,6 @@ package lfs
 
 import (
 	"net/http"
-	"net/url"
 	"testing"
 
 	"github.com/github/git-lfs/config"
@@ -18,14 +17,9 @@ func TestProxyFromEnvironment(t *testing.T) {
 		"HTTPS_PROXY": "https://proxy-from-env:8080",
 	})
 
-	u, err := url.Parse("https://some-host.com:123/foo/bar")
+	req, err := http.NewRequest("GET", "https://some-host.com:123/foo/bar", nil)
 	if err != nil {
 		t.Fatal(err)
-	}
-
-	req := &http.Request{
-		URL:    u,
-		Header: http.Header{},
 	}
 
 	proxyURL, err := httputil.ProxyFromGitConfigOrEnvironment(cfg)(req)
@@ -39,14 +33,9 @@ func TestProxyFromGitConfig(t *testing.T) {
 		"http.proxy": "https://proxy-from-git-config:8080",
 	})
 
-	u, err := url.Parse("http://some-host.com:123/foo/bar")
+	req, err := http.NewRequest("GET", "http://some-host.com:123/foo/bar", nil)
 	if err != nil {
 		t.Fatal(err)
-	}
-
-	req := &http.Request{
-		URL:    u,
-		Header: http.Header{},
 	}
 
 	proxyURL, err := httputil.ProxyFromGitConfigOrEnvironment(cfg)(req)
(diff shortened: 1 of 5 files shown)


jonmagic:
@technoweenie awesome! Just getting back from vacation, anything I can do to help on this?

lfs/proxy_test.go
```diff
@@ -0,0 +1,59 @@
...
+	req, err := http.NewRequest("GET", "http://some-host.com:123/foo/bar", nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	proxyURL, err := httputil.ProxyFromGitConfigOrEnvironment(cfg)(req)
+
+	assert.Nil(t, proxyURL)
+	assert.Nil(t, err)
+}
```
sinbad: I think there should be some tests with `NO_PROXY` for completeness.

sinbad:
👍 looks good

lfs/proxy_test.go
ttaylorr: Added in https://github.com/github/git-lfs/pull/1358/commits/f0acdb285435d78fc425e3169e4e7d168fa393fa

technoweenie:
Hey, funny story time. @ttaylorr asked me why the function is named `ProxyFromGitConfigOrEnvironment`, when it's reading from the env first, _before_ git config. Turns out it's named that way because the original implementation read from the git config first. Then, I chimed in:

> My main question is whether this should prefer the git config or the environment first? I'd lean towards using the git config as a last resort. A couple use cases:
>
> * Someone has a global http_proxy set in their bashrc because of their local network setup.
> * Someone runs http_proxy=.... git clone ...
>
> I'd like to see how git handles this.

I never looked into it until just now though, while looking for any other proxy related git config values. The [docs](https://git-scm.com/docs/git-config) say:

> Override the HTTP proxy, normally configured using the http_proxy, https_proxy, and all_proxy environment variables 

Looks like @jonmagic and @LizzHale are right :) I wonder if a git `http.proxy` value should override a `HTTPS_PROXY` env var for https requests though?

sinbad:
This is a great point - I even checked the precedence and agreed with it, but now you raise it again I think this does need to be the other way around in this case. 

Usually environment variables should override gitconfig, but in this case the environment variables are general ones, not git specific ones. Thinking about it, it makes sense that the overall order of precedence should be:

1. Git specific environment variables
2. git config
3. General environment variables

.. because otherwise you have no way of persistently overriding generic environment settings specifically for git. The confusion I've had is that we've only really had 1 & 2 most of the time.

technoweenie:
I updated the code to check the git config first :+1:

ttaylorr:
Looks good! ✨

LizzHale:
:sparkles:

PR #1360, opened 2016-07-11
Unbreak building httputil on OpenBSD

jasperla:
This effectively allows for building git-lfs on OpenBSD.

### DIFF
--- a/httputil/certs_openbsd.go
+++ b/httputil/certs_openbsd.go
@@ -1,0 +1,8 @@
+package httputil
+
+import "crypto/x509"
+
+func appendRootCAsForHostFromPlatform(pool *x509.CertPool, host string) *x509.CertPool {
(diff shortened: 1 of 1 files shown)


ttaylorr:
I'm glad that this fixes building LFS on OpenBSD. I just have one comment:

So that we don't break building on other platforms, you might consider adding a `+build` flag to the top of the `httputil/certs_openbsd.go` file that you added, signaling Go to only build this file on the `openbsd` platform. Your file would then look like:

```go
// +build openbsd

package httputil

// ...
```

Other than that, 👍

ttaylorr:
Ah, I totally forgot about [build suffixes](https://golang.org/pkg/go/build/). LGTM!

jasperla:
Thanks for the quick merge!

PR #1361, opened 2016-07-11
travis-ci: add a build job to test against upcoming versions of Git

larsxschneider:
The Git source version is defined by the variables GIT_SOURCE_REPO and
GIT_SOURCE_BRANCH (which can also be a Git hash). The Git source is
checked out, compiled, and installed on the Travis CI Linux machine.
Afterwards all Git-LFS tests are executed against the this Git version.

This setup helps to test features that depend on changes in the
development branches of Git and Git-LFS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,12 +10,25 @@ os:
 env:
   global:
     - GIT_LFS_TEST_DIR="$HOME/git-lfs-tests"
+    - GIT_SOURCE_REPO="https://github.com/git/git.git"
+    - GIT_SOURCE_BRANCH="next"
 
 matrix:
   fast_finish: true
   allow_failures:
     - os: osx
+    - env: git-from-source
   include:
+    - env: git-from-source
+      os: linux
+      before_script:
+        - >
+          git clone $GIT_SOURCE_REPO git-source;
+          cd git-source;
+          git checkout $GIT_SOURCE_BRANCH;
+          make --jobs=2;
+          make install;
+          cd ..;
     - env: git-latest
       os: linux
       addons:


ttaylorr:
Cool!

larsxschneider:
Oops.. that was fast :-) - I was offline for a bit and couldn't check the Travis CI result, yet.
Apparently the test run fails with exit code 1 although all integration tests return OK:
https://travis-ci.org/github/git-lfs/jobs/143974381

Do you see what is going on?

ttaylorr:
I do not see any reason that the `script/integration` would be `exit 1`-ing in this case. It shouldn't be too big of a deal for now, since the `git@next` stuff isn't in a required build, but it would still be nice to have an accurate build status.

Maybe modify `script/integration` to run with `set -x` to get a closer look at what's going on?

larsxschneider:
There is no test "pull zero len file" in your Travis CI build:
https://travis-ci.org/github/git-lfs/jobs/143974381

... apparently this test fails in my own fork (both use the Git "next" branch):
https://travis-ci.org/larsxschneider/git-lfs/jobs/143974380

This might even be a problem with Git "next"... I tried the exact hash from Git 2.9 and all tests pass (on my fork):
https://travis-ci.org/larsxschneider/git-lfs/jobs/143974721

Questions:
(1) Why is a test missing in Travis CI?
(2) What Git branch should we test against by default? "next" or "master"?

larsxschneider:
@technoweenie @ttaylorr :
Oh oh:
I changed the Git from source to build Git 2.9 and it works: https://github.com/github/git-lfs/pull/1362
https://travis-ci.org/github/git-lfs/jobs/144080704

However, the regular Travis build fails because Git was updated to 2.9.1 last night:
https://travis-ci.org/github/git-lfs/jobs/144080705

... and the test case "test: pull zero len file ..." brakes

PR #1367, opened 2016-07-13
Enhanced transfers part 5: external custom transfer adapters

sinbad:
This PR allow external processes to act as relays for uploads and downloads, allowing anyone to implement custom transfer techniques of their choosing without cluttering the core with N supported methods.

LFS talks to the process over stdin/stdout with a simple linewise JSON protocol, with content being exchanged by file path (process and lfs must be on the same machine). The process can then talk to the server however it likes. 

The name of the custom adapter must be agreed between the client config and server, since the adapter to use will be negotiated at the API stage, and may change the information the API returns in the `actions` section. The content of the`upload` and `download` actions from the API are passed to the custom adapter so it can use that information as it sees fit.

This PR also contains a reference implementation of a custom adapter process, `test/cmd/lfstest-customadapter.go` which is used in the integration tests. It just uses the same HTTP protocol under the hood to keep things simple and to re-use the `lfstest-gitserver`, but demonstrates how the process comms work and how progress is reported etc. Other implementations can transfer data however they like.


sinbad:
Test failure is down to the zero-len file issue with latest Git: https://github.com/github/git-lfs/issues/1363 :(

test/cmd/lfstest-customadapter.go
```diff
@@ -0,0 +1,253 @@
...
+	"github.com/github/git-lfs/tools"
+)
+
+// This test custom adapter just acts as a bridge for uploads/downloads
+// in order to demonstrate & test the custom transfer adapter protocols
+// All we actually do is relay the requests back to the normal storage URLs
+// of our test server for simplicity, but this proves the principle
+func main() {
+
+	scanner := bufio.NewScanner(os.Stdin)
```
ttaylorr: Why do you need buffered versions of these?

transfer/custom.go
```diff
@@ -0,0 +1,376 @@
...
+	// send initiate message
+	initReq := NewCustomAdapterInitRequest(a.getOperationName(), a.concurrent, a.originalConcurrency)
+	resp, err := a.exchangeMessage(ctx, initReq)
+	if err != nil {
+		a.abortWorkerProcess(ctx)
+		return nil, err
+	}
+	if resp.Error != nil {
+		a.abortWorkerProcess(ctx)
+		return nil, fmt.Errorf("Error initializing custom adapter %q worker %d: %v", a.name, workerNum, resp.Error.Error())
```
technoweenie: Don't think we need to call `Error()` here specifically, `%v` should figure it out. It's also consistent with the error formatting above.

transfer/custom.go
```diff
@@ -0,0 +1,376 @@
...
+func (a *customAdapter) sendMessage(ctx *customAdapterWorkerContext, req interface{}) error {
+	b, err := json.Marshal(req)
+	if err != nil {
+		return err
+	}
+	tracerx.Printf("xfer: Custom adapter worker %d sending message: %v", ctx.workerNum, string(b))
+	// Line oriented JSON
+	b = append(b, '\n')
+	_, err = ctx.stdin.Write(b)
+	if err != nil {
```
technoweenie: We could cut down on some lines here by returning nil, right?

transfer/custom.go
```diff
@@ -0,0 +1,376 @@
...
+
+	tracerx.Printf("xfer: Shutting down adapter worker %d", ctx.workerNum)
+	termReq := NewCustomAdapterTerminateRequest()
+	err := a.sendMessage(ctx, termReq)
+	if err != nil {
+		return err
+	}
+	ctx.stdin.Close()
+	ctx.stdout.Close()
+	return ctx.cmd.Wait()
```
technoweenie: We think this should have some kind of timeout before calling `cmd.Process.Kill()` to kill it by force. Otherwise, a bad custom adapter could cause the git operation to wait forever.

technoweenie:
@rubyist, @ttaylorr, and I started going through this together. We're not quite done yet, but we've sent a few review comments. Things look cool so far. We definitely want to build our own custom adapter real quick to play around with it for real, too.

test/cmd/lfstest-customadapter.go
sinbad: Robustness, since we're dealing with cross-process comms and need to not assume how the other end flushes the streams. For example if the sender wrote 2 messages in quick succession to the stream and called flush only after both, a plain Read() could read both at once and you need somewhere to stash the second - to cope with this would basically be doing exactly what `bufio` code does.

transfer/custom.go
sinbad: Doh, yeah I cut some code out the middle there.

transfer/custom.go
sinbad: OK, but that's an persistent problem unless we add timeouts to _all_ requests to the adapter, in which case we need to define what is a 'reasonable' time between receiving some kind of response - including a progress update or a transfer completion. Right now the equivalent for HTTP transfers is 30 minutes (the keepalive), since the dial / TLS timeouts only deal with unresponsiveness on the initial connection.

I'll add it to this shutdown sequence specifically but there are definitely other stall paths which can be caused by either servers or now adapters.

test/cmd/lfstest-customadapter.go
sinbad: Oh if you meant the writers rather than the reader, I just like the explicitness of `Flush()` and the flexibility to build up more complex messages in stages if I wanted to (I didn't need to but hey).

sinbad:
I spotted a few cosmetic things while I was looking again this morning so tidied those up too. Tests passing now that I've merged master too \o/

transfer/custom.go
```diff
@@ -0,0 +1,382 @@
...
+func ConfigureCustomAdapters() {
+	pathRegex := regexp.MustCompile(`lfs.customtransfer.([^.]+).path`)
+	for k, v := range config.Config.AllGitConfig() {
+		if match := pathRegex.FindStringSubmatch(k); match != nil {
+			name := match[1]
+			path := v
+			var args string
+			var concurrent bool
+			var direction string
+			// retrieve other values
```
technoweenie: I would change the operation on the substring match, so it continues if there's no match. This reduces the level of indentation. 

Also, it wasn't immediately clear what the 3 vars are for, until we got to the closure below. I think a quick comment would be helpful, since vars are unusual.

transfer/custom.go
```diff
@@ -0,0 +1,382 @@
...
+					return fmt.Errorf("Failed to copy downloaded file: %v", err)
+				}
+			} else if a.direction == Upload {
+				if err = api.VerifyUpload(t.Object); err != nil {
+					return err
+				}
+			}
+			wasAuthOk = true
+			complete = true
+		}
```
technoweenie: This should handle the `default` case. A bad transfer adapter could cause Git LFS to hang, and then we'd get the bug report for it :)

technoweenie:
This looks good, once the last comments have been addressed :metal:

docs/custom-transfers.md
```diff
@@ -0,0 +1,276 @@
...
+the LFS store.
+
+##### Progress
+
+In order to support progress reporting while data is uploading / downloading, 
+the transfer process should post messages to stdout as follows before sending
+the final completion message:
+
+```json
+{ "id":"progress", "oid": "22ab5f63670800cc7be06dbed816012b0dc411e774754c7579467d2536a9cf3e", "bytesSoFar": 1234, "bytesSinceLast": 64 }
```
rubyist: Until now, I don't think any of our API JSON has had multi-word keys. Here we're camelcasing them, e.g. `bytesSoFar`, but in the [locking proposal](https://github.com/github/git-lfs/blob/master/docs/proposals/locking_api.md) we've got snakecase going on. Should we standardize on one or the other? I have no opinion of one over the other.

technoweenie: Dang, good catch. I prefer snake casing for JSON. But, arguments can be made for using camel casing. Go fields typically use camel case. Javascript fields do too, and it's where JSON came from (even if the spec doesn't really specify a preference).

docs/custom-transfers.md
```diff
@@ -0,0 +1,276 @@
...
+Immediately after invoking a custom transfer process, git-lfs sends initiation
+data to the process over stdin. This tells the process useful information about
+the configuration.
+
+The message will look like this:
+```json
+{ "id":"init", "operation":"download", "concurrent": true, "concurrenttransfers": 3 }
+```
+
+* `id`: Always "init" to identify this message
```
technoweenie: I think `id` is a confusing name for this field. Some suggestions we had were `state`, `operation`, `event`, `type`, `command`. This came up a couple times in our group review.

ttaylorr: I think `operation` is the clearest, but `Op` is a nice way to write it. My vote is for that :-)

rubyist: :+1: for op

technoweenie: It's also already used in the batch api requests. I'm not sure if that's a good or bad thing :)

transfer/custom.go
sinbad: I keep forgetting Ruby style ;)

Actually the explicit var declarations aren't needed any more; they were at one point but the code iterated out of needing them without me realising, I'll take them out.

docs/custom-transfers.md
sinbad: I'm with Richard Hendrix in preferring not to waste all those bytes on underscores ;)

I've mostly seen style guides that state camelCase (like [Googles](https://google.github.io/styleguide/jsoncstyleguide.xml#Property_Name_Format)) but I've seen real-world examples that use underscores or dashes too. Guess it's still a personal thing - as the first person to hit multi-word properties I'm temped to stick my flag in for cameCase but won't bust a gasket if you want me to change it.

docs/custom-transfers.md
sinbad: I thought about this far too long as well - naming eh? ;) I deliberately avoided `operation` because of the ambiguity with the batch api requests, and it's already used in the init structure to reflect that very purpose. I originally used `type` but that clashes with reserved words in most languages so I didn't want to cause people with autogen tools more hassle than I needed to.

I'd like `op` if it wasn't a contraction of `operation` which is already used. 😕 

I like `event` best out of the remaining options.

docs/custom-transfers.md
```diff
@@ -0,0 +1,276 @@
...
+
+Or if there was an error in the transfer:
+
+```json
+{ "event":"complete", "oid": "bf3e3e2af9366a3b704ae0c31de5afa64193ebabffde2091936ad2e7510bc03a", "error": { "code": 2, "message": "Explain what happened to this transfer" }}
+```
+
+* `event`: Always "complete" to identify this message
+* `oid`: the identifier of the LFS object
+* `error`: Should contain a `code` and `message` explaining the error
```
rubyist: This documentation for `error` is slightly different than the documentation for `error` on the download complete message. This one doesn't say what error should be if there is no error, the other doesn't say what error should be if there is an error. And if there is no error _must_ it be null or can it also be absent?

docs/custom-transfers.md
```diff
@@ -0,0 +1,276 @@
...
+The transfer process should post one or more [progress messages](#progress) and 
+then a final completion message as follows:
+
+```json
+{ "event":"complete", "oid": "22ab5f63670800cc7be06dbed816012b0dc411e774754c7579467d2536a9cf3e", "path": "/path/to/file.png", "error": null}
+```
+
+* `event`: Always "complete" to identify this message
+* `oid`: the identifier of the LFS object
+* `path`: the path to a file containing the downloaded data, which the transfer
```
rubyist: Should this be specified as an absolute path?

rubyist:
Are we saying that all LFS servers MUST support the basic protocol? Let's say I'm using a custom adapter to store my objects somewhere else entirely and there is no manner in which I can access my objects using the basic http protocols. What does it look like if my client does not have the custom adapter installed and thus cannot negotiate it? Along with its preferred negotiated transfer type should the server return information about all the transfer types it supports so we could give better error messages?

rubyist:
Is there an intended way for a custom adapter to get any configuration options or is that left up to each adapter? I'm thinking things like service auth keys, preferred local tmp/download directories, stuff like that.

docs/custom-transfers.md
sinbad: `error` can always be absent if there is no error - I tried to make it clearer by showing 2 versions of the complete message (it's the same in both the upload and download sections) - it first shows the struct if the upload/download is complete (with no error) and then what it should look like with an error. In practice it's just one structure with `error` (and `path` in the case of download) both being optional depending on the circumstances, but I thought it was clearer showing an explicit example in each case. I might be wrong 😄

docs/custom-transfers.md
sinbad: Technically it's not required, since the adapter process will have the same initial working directory as the git-lfs binary, so any relative path would be resolved to the same place. So I didn't specify in case someone wanted to exploit that to store temp files in a subdir for example.

rubyist: Ah yeah, fair points :+1:

sinbad:
> Is there an intended way for a custom adapter to get any configuration options or is that left up to each adapter?

I considered passing the git-lfs config or even the entire gitconfig to the process in the init struct at first, but I decided against it and pared it down to a minimum, because it would put unnecessary overhead on parsing the init struct (gitconfigs can get pretty large), and it might not even include the settings an adapter might need anyway. So beyond `args`, I'm leaving any further config to the adapter itself.

sinbad:
> Are we saying that all LFS servers MUST support the basic protocol? Let's say I'm using a custom adapter to store my objects somewhere else entirely and there is no manner in which I can access my objects using the basic http protocols. What does it look like if my client does not have the custom adapter installed and thus cannot negotiate it?

The `basic` protocol is always implicit (required for backwards compatibility) so yeah if you have an LFS server which doesn't implement HTTP transfers and the client doesn't support the custom protocol the server solely supports, it will fail during the transfer process in the same way as if there was no HTTP endpoint to connect to. 

So we're not saying LFS servers _have_ to support `basic`, but if they don't then clients without the sole custom adapter they support will not work.

docs/custom-transfers.md
sinbad: Helps if I read what I actually wrote vs what I think I wrote - actually it's the other `error` doc which is wrong and shouldn't be talking about `null`.

rubyist:
Looking good. It was raining last night, limiting my Pokemon hunting, so I wrote my own toy adapter to see how it'd go.  Pretty straight forward to implement and test using the protocol as documented. I dig it.

My adapter used minio as a backend, which is all http so a custom adapter isn't really necessary, but it's what I happened to be looking at so I went with it. Some code is [here](https://gist.github.com/rubyist/0fcdda7186b82b3a30435bb8fa8d6ebf).

sinbad:
@rubyist nice, glad you found it easy enough to implement, that was my hope 👍 Although HTTP get/put is pretty easy & ubiquitous I like that this gets you out of a bind whether you want to do something just a little more specific but still over HTTP (outside just changing url/headers), or if you want to do something more wacky with totally different protocols. I can think of cases of both but this way we don't have to worry about all the variations.

ttaylorr:
🎉

PR #1371, opened 2016-07-13
Handle artifactory responses

ttaylorr:
This pull-request rejects pushes with actions that have size less than zero bytes, as according to the spec changes.


ttaylorr:
:+1: from @rubyist as discussed in meeting.

sinbad:
:+1: barring test failure on one of the configurations which needs investigating.

ttaylorr:
Investigating the test failure now! 😄

PR #1372, opened 2016-07-14
travis-ci: require successful tests against upcoming Git core release

larsxschneider:
The Git-core 'master' branch tracks the commits that should go into the next release (see https://github.com/git/git/blob/master/Documentation/gitworkflows.txt). 

Run Git-LFS tests against a build of the Git-core 'master' branch and require a successful pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,13 +11,12 @@ env:
   global:
     - GIT_LFS_TEST_DIR="$HOME/git-lfs-tests"
     - GIT_SOURCE_REPO="https://github.com/git/git.git"
-    - GIT_SOURCE_BRANCH="next"
+    - GIT_SOURCE_BRANCH="master"
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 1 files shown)


sinbad:
👍 Seems like a good idea to me.

ttaylorr:
Works for me, too 👍

ttaylorr:
Looks sweet, @larsxschneider 😄

technoweenie:
> * 'master' tracks the commits that should go into the next release;
>
> * 'next' is intended as a testing branch for topics being tested for
  stability for master.

It makes sense for the builds to allow failures in git `next`. But depending on tests against git `master` means we get an early warning on failures for upcoming releases. At least this way, we can get fixes out before end users can encounter them.

ttaylorr:
I agree with what @technoweenie said. I would love to have both, but I hate adding stuff to the build matrix.

PR #1373, opened 2016-07-14
In git 2.9+, run "git lfs pull" in submodules after "git lfs clone"

### LINKED ISSUE #1172: git lfs clone for submodules
I realized that the `git lfs clone` [speed up](https://developer.atlassian.com/blog/2016/04/git-lfs-12-clone-faster/) does not work with Git Submodules and I wonder if anyone has a clever idea to approach this problem. Here is my first wrapper script attempt:

```bash
git-lfs clone $@
if [[ -z $2 ]]; then
    CLONE_PATH=$(basename ${1%.git});
else
    CLONE_PATH=$2;
fi
pushd "$CLONE_PATH"
    git submodule init
    wt_prefix=$(git rev-parse --show-prefix)
    git submodule--helper list --prefix "$wt_prefix" | {
        while read mode sha1 stage sm_path
        do
            name=$(git submodule--helper name "$sm_path") || exit
            url=$(git config submodule."$name".url)
            if ! test -d "$sm_path"/.git && ! test -f "$sm_path"/.git
            then
                git submodule--helper clone --prefix "$wt_prefix" --path "$sm_path" --name "$name" --url "$url"
                pushd "$sm_path"
                    git -c filter.lfs.smudge=cat checkout -q $sha1 || exit
                    git-lfs pull || exit
                popd
            fi
        done
    }
popd
```

This works by taking the relevant pieces from the [`git submodule`](https://github.com/git/git/blob/6a6636270fbaf74609cd3e1bd207dd2c420d640a/git-submodule.sh#L686-L731) command. However, this is probably a pretty fragile solution. I would prefer to use higher level Submodule commands. @sinbad @stefanbeller any better idea or do you think this is the way to go?

sinbad:
Fixes #1172

git 2.9+ preserves the config settings which disable the LFS filters in submodules so any LFS files will be pointers. This makes it more optimal (previous versions will suppress these settings so the smudge filter will be used in submodules) but we must run "git lfs pull" to update the submodule working copies.

/cc @larsxschneider

### DIFF
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -6,6 +6,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/github/git-lfs/subprocess"
+
 	"github.com/github/git-lfs/config"
 	"github.com/github/git-lfs/git"
 	"github.com/github/git-lfs/localstorage"
@@ -79,8 +81,36 @@ func cloneCommand(cmd *cobra.Command, args []string) {
 		fetchRef("HEAD", include, exclude)
 	} else {
 		pull(include, exclude)
+
+		err := postCloneSubmodules(args)
+		if err != nil {
+			Exit("Error performing 'git lfs pull' for submodules: %v", err)
+		}
 	}
 
+}
+
+func postCloneSubmodules(args []string) error {
+	// In git 2.9+ the filter option will have been passed through to submodules
+	// So we need to lfs pull inside each, if the original clone command
+	if !git.Config.IsGitVersionAtLeast("2.9.0") {
+		// In earlier versions submodules would have used smudge filter
+		return nil
+	}
+	// Also we only do this if --recursive or --recurse-submodules was provided
+	if !cloneFlags.Recursive && !cloneFlags.RecurseSubmodules {
+		return nil
+	}
+
+	// Use `git submodule foreach --recursive` to cascade into nested submodules
+	// Also good to call a new instance of git-lfs rather than do things
+	// inside this instance, since that way we get a clean env in that subrepo
+	cmd := subprocess.ExecCommand("git", "submodule", "foreach", "--recursive",
+		"git lfs pull")
+	cmd.Stderr = os.Stderr
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	return cmd.Run()
 }
 
 func init() {
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -92,7 +92,7 @@ func cloneCommand(cmd *cobra.Command, args []string) {
 
 func postCloneSubmodules(args []string) error {
 	// In git 2.9+ the filter option will have been passed through to submodules
-	// So we need to lfs pull inside each, if the original clone command
+	// So we need to lfs pull inside each
 	if !git.Config.IsGitVersionAtLeast("2.9.0") {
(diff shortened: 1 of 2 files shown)


commands/command_clone.go
```diff
@@ -79,8 +81,36 @@ func cloneCommand(cmd *cobra.Command, args []string) {
...
+		if err != nil {
+			Exit("Error performing 'git lfs pull' for submodules: %v", err)
+		}
+	}
+
+}
+
+func postCloneSubmodules(args []string) error {
+	// In git 2.9+ the filter option will have been passed through to submodules
+	// So we need to lfs pull inside each, if the original clone command
```
technoweenie: Is this an incomplete thought, or should it read "in the original clone command"?

commands/command_clone.go
```diff
@@ -79,8 +81,36 @@ func cloneCommand(cmd *cobra.Command, args []string) {
...
+	}
+	// Also we only do this if --recursive or --recurse-submodules was provided
+	if !cloneFlags.Recursive && !cloneFlags.RecurseSubmodules {
+		return nil
 	}
 
+	// Use `git submodule foreach --recursive` to cascade into nested submodules
+	// Also good to call a new instance of git-lfs rather than do things
+	// inside this instance, since that way we get a clean env in that subrepo
+	cmd := subprocess.ExecCommand("git", "submodule", "foreach", "--recursive",
```
ttaylorr: As a thought, we could support the `--jobs` flag from the [`git-clone`](https://git-scm.com/docs/git-clone) documentation.

Instead of running `git submodule foreach`, we could just get a list of _all_ submodule paths, feed all of those paths into a channel, and then slurp them from N goroutines. As a first pass, `--jobs` could be set to 1 so that it effectively behaves the same. I'm thinking something like:

```go
var wg sync.WaitGroup
var paths chan string

modules, err := subprocess.ExecCommand("git", "get-all-submodule-paths")
if err != nil {
        // ...
}

wg.Add(len(modules))
go func() {
        for _, path := range modules {
                paths <- path
        }
}()

var jobs int
for i range jobs {
        go func(paths <-chan string) {
        select {
        case path := <-paths:
                // run git-lfs pull
                wg.Done()
        }
        }
}

wg.Wait()
```

commands/command_clone.go
sinbad: Bah, stuff moved around and left this hanging, it's covered below now so I'll trim it.

commands/command_clone.go
sinbad: As discussed in #1172 the reason for using `git submodule foreach` is that you get recursion (submodules within submodules) for free - re-implementing the recursion ourselves is a complication I wanted to avoid (your code template above won't handle it and would actually fail the integration test since it specifically tests for this). 

Also since transfers in lfs pull are already parallelised I don't think we necessarily gain very much by also parallelising across submodules. The main bottleneck is the transfers so I don't think the added benefit of parallel submodule parsing is worth the extra complexity.

larsxschneider:
Thank you! 👍 😄

commands/command_clone.go
ttaylorr: I agree, we probably don't need two independent layers of concurrency. If we did end up pursuing this, though, it would be easy to convince `git submodule foreach` to work with us. We could just have it run `pwd` in each submodule directory, and then apply the same code above to use that.

Either way, I think this is fine for now 😄

PR #1374, opened 2016-07-15
Windows fixes

sinbad:
A bunch of fixes on Windows, details in the individual commits. 

Integration tests all appear OK but require some user interaction to cancel GUI auth prompts which the Windows version of OpenSSH seems to be throwing up.

### DIFF
--- a/api/schema/schema_validator.go
+++ b/api/schema/schema_validator.go
@@ -4,7 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/xeipuuv/gojsonschema"
@@ -24,11 +24,12 @@ func NewSchemaValidator(t *testing.T, schemaName string, got interface{}) *Schem
 	if err != nil {
 		t.Fatal(err)
 	}
+	// Platform compatibility: use "/" separators always for file://
+	dir = strings.Replace(dir, "\\", "/", -1)
 
 	schema := gojsonschema.NewReferenceLoader(fmt.Sprintf(
-		"file:///%s",
-		filepath.Join(dir, "schema/", schemaName),
-	))
+		"file:///%s/schema/%s", dir, schemaName),
+	)
 
 	marshalled, err := json.Marshal(got)
 	if err != nil {
--- a/api/schema/schema_validator.go
+++ b/api/schema/schema_validator.go
@@ -4,8 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"runtime"
-	"strings"
+	"path/filepath"
 	"testing"
 
 	"github.com/xeipuuv/gojsonschema"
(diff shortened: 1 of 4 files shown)


sinbad:
Figured out how to suppress the OpenSSH prompts, all definitely good on integration tests now.

api/schema/schema_validator.go
```diff
@@ -24,11 +24,12 @@ func NewSchemaValidator(t *testing.T, schemaName string, got interface{}) *Schem
 	if err != nil {
 		t.Fatal(err)
 	}
+	// Platform compatibility: use "/" separators always for file://
+	dir = strings.Replace(dir, "\\", "/", -1)
```
ttaylorr: Should we use the `os.PathSeparator` here?

```go
dir = strings.Replace(dir, os.PathSeparator, "/", -1)
```

sinbad: No, because then it would be pointlessly replacing '/' with '/' on non-Windows platforms ;) This is the only case where it's useful.

ttaylorr:
I'm a little bit anxious about the `s/filepath/path` change in `tools/filetools.go`, but I think we should be good to merge after you figure out how to suppress those OpenSSH prompts 😄

api/schema/schema_validator.go
ttaylorr: No, `strings.Replace` bails out if `old == new` (see: https://golang.org/src/strings/strings.go?s=17447:17493#L634).

I think it would be good to use `os.PathSeparator` so that this string is definitely correct and consistent with everything else on Windows.

sinbad: I could surround this code with a `GOOS` check to not do it at all on non-Windows actually, will do that.

ttaylorr: I don't think we should do that. Let's not introduce additional complexity. I am all for the comment that I [originally suggested](https://github.com/github/git-lfs/pull/1374/files/ee5df4971f3a819607d31ce91841417141ae1b84#r70994673).

sinbad:
> I'm a little bit anxious about the s/filepath/path change in tools/filetools.go

`path.Clean` is the same as `filepath.Clean` except that it uses non-platform-specific separators, which is exactly what we need in all the cases we need it right now. This has always been an annoying problem on Git for Windows, that git still uses forward slashes even though the platform in general doesn't.

> after you figure out how to suppress those OpenSSH prompts

Already did 😄

api/schema/schema_validator.go
sinbad: Hmm, I disagree. The specific case it's dealing with is the separators being wrong on Windows. The early-out inside the Go code (should you dig into it) isn't the point, it's that generalising it via `os.PathSeparator` obfuscates the purpose of it in our code. I prefer clarity over no-op generalisation.

ttaylorr: I agree that the no-op generalization is a little bit confusing, but I feel that the os.PathSeparator + a comment explaining what is going on is far preferable than depending on the `runtime` package.

rubyist: Isn't this what `filepath.ToSlash()` is for?

technoweenie: I'd prefer using `os.PathSeparator` too. I actually didn't know `strings.Replace()` behaved that way, but it makes sense.

sinbad: > Isn't this what filepath.ToSlash() is for?

Ha, this is by far the best option, I didn't spot that one 😄 Thanks @rubyist!

ttaylorr:
Glad we went with `filepath.ToSlash`. I'm ready to ship this 👍

PR #1378, opened 2016-07-17
Updated request.GetAuthType to handle multi-value auth headers

VladimirKhvostov:
GetAuthType method in https://github.com/github/git-lfs/blob/master/httputil/request.go#L174 only takes first WWW-Authenticate header into account. If server sends the following headers:
```
WWW-Authenticate: Bearer
WWW-Authenticate: Negotiate
WWW-Authenticate: NTLM
```
the code returns "basic", which is definitely not correct in this case.

This change fixes this issue. 

Bug: https://github.com/github/git-lfs/issues/1377

### DIFF
--- a/httputil/request.go
+++ b/httputil/request.go
@@ -21,6 +21,13 @@ type ClientError struct {
 	RequestId        string `json:"request_id,omitempty"`
 }
 
+const (
+	WwwAuthenticateHeader = "Www-Authenticate"
+	LfsAuthenticateHeader = "Lfs-Authenticate"
+	BasicAuthType         = "basic"
+	NtlmAuthType          = "ntlm"
+)
+
 func (e *ClientError) Error() string {
 	msg := e.Message
 	if len(e.DocumentationUrl) > 0 {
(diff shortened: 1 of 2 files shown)


httputil/request.go
```diff
@@ -172,14 +179,25 @@ func SetAuthType(req *http.Request, res *http.Response) {
...
 	}
 
-	if strings.HasPrefix(strings.ToLower(auth), "ntlm") {
-		return "ntlm"
+	return authType
+}
+
+func GetAuthTypeFromHeader(res *http.Response, headerName string) string {
+	authHeaders := res.Header[headerName]
+	for i := range authHeaders {
```
ttaylorr: You can grab the values contained in this slice by using the two-value `range` like so:

```go
for _, auth := range authHeaders {
        if strings.HasPrefix(...) {
        }
}
```

httputil/request.go
```diff
@@ -172,14 +179,25 @@ func SetAuthType(req *http.Request, res *http.Response) {
...
+		auth := authHeaders[i]
+
+		if strings.HasPrefix(strings.ToLower(auth), NtlmAuthType) {
+			return NtlmAuthType
+		}
 	}
 
-	return "basic"
+	return BasicAuthType
```
ttaylorr: It looks like this code only returns either the `Ntlm` or `Basic` auth types. I'm not super familiar with how this spec works: are there any other authentication types that we should support?

httputil/request_test.go
```diff
@@ -0,0 +1,60 @@
+package httputil
+
+import (
+	"github.com/stretchr/testify/assert"
+	"net/http"
+	"testing"
+)
+
+func TestGetAutTypeNoAuthHeaders(t *testing.T) {
+	res := &http.Response{Header: make(http.Header)}
```
ttaylorr: This initialization could be good to extract to a helper function that, given a `map[string]string` representing the header values, returns a `*http.Response` initialized with those header values. I'm thinking something like:

```go
func httpResponseFromHeaders(headers map[string]string) *http.Response {
        header := make(http.Header, len(headers))
        for k, v := range headers {
                header.Add(k, v)
        }

        return &http.Response{Header: header}
}
```

httputil/request.go
VladimirKhvostov: Makes sense. I'll make a change.

httputil/request.go
VladimirKhvostov: Great question. It looks to me that git-lfs currently only supports NTML and basic authentication.  Kerberos for example, is more secure than NTLM, which is more secure than basic auth, Since Kerberos is not supported in git-lfs, it does not make sense to recognize it. I must say that according to https://blogs.msdn.microsoft.com/chiranth/2014/04/17/setting-up-kerberos-authentication-for-a-website-in-iis/ when server sends Negotiate header, it indiacate that but both Kerberos and NTLM are supported. Since a lot of clients do not handle Negotiate, people normally configure servers to send both Negotiate and NTLM headers. Do you want to handle Negotiate explicitly?

VladimirKhvostov:
Made changes based Taylor's feedback. Added code to handle Www-Authenticate: Negotiate.

VladimirKhvostov:
Unfortunately, GitHub does not handle case when I amend my commit and force push it. VSTS handles this beautifully. I'll close this PR and create a new one. Sorry about this.

PR #1379, opened 2016-07-18
Updated request.GetAuthType to handle multi-value auth headers

VladimirKhvostov:
GetAuthType method in https://github.com/github/git-lfs/blob/master/httputil/request.go#L174 only takes first WWW-Authenticate header into account. If server sends the following headers:
```
WWW-Authenticate: Bearer
WWW-Authenticate: Negotiate
WWW-Authenticate: NTLM
```
the code returns "basic", which is definitely not correct in this case.

This change fixes this issue.  With this change we will also handle Negotiate authentication.

Bug: https://github.com/github/git-lfs/issues/1377

### DIFF
--- a/httputil/request.go
+++ b/httputil/request.go
@@ -21,6 +21,14 @@ type ClientError struct {
 	RequestId        string `json:"request_id,omitempty"`
 }
 
+const (
+	WwwAuthenticateHeader = "Www-Authenticate"
+	LfsAuthenticateHeader = "Lfs-Authenticate"
+	BasicAuthType         = "basic"
+	NtlmAuthType          = "ntlm"
+	NegotiateAuthType     = "negotiate"
+)
+
 func (e *ClientError) Error() string {
 	msg := e.Message
 	if len(e.DocumentationUrl) > 0 {
@@ -172,14 +180,19 @@ func SetAuthType(req *http.Request, res *http.Response) {
 }
 
 func GetAuthType(res *http.Response) string {
-	auth := res.Header.Get("Www-Authenticate")
-	if len(auth) < 1 {
-		auth = res.Header.Get("Lfs-Authenticate")
+
+	for _, headerName := range []string{WwwAuthenticateHeader, LfsAuthenticateHeader} {
+		authHeaders := res.Header[headerName]
+		for _, auth := range authHeaders {
+
+			authLower := strings.ToLower(auth)
+			// When server sends Www-Authentication: Negotiate, it supports both Kerberos and NTML.
+			// Since git-lfs current does not support Kerberos, we will return NTLM in this case.
+			if strings.HasPrefix(authLower, NtlmAuthType) || strings.HasPrefix(authLower, NegotiateAuthType) {
+				return NtlmAuthType
+			}
+		}
 	}
 
-	if strings.HasPrefix(strings.ToLower(auth), "ntlm") {
-		return "ntlm"
-	}
-
-	return "basic"
+	return BasicAuthType
 }
--- a/httputil/request.go
+++ b/httputil/request.go
@@ -22,11 +22,13 @@ type ClientError struct {
 }
 
 const (
-	WwwAuthenticateHeader = "Www-Authenticate"
-	LfsAuthenticateHeader = "Lfs-Authenticate"
-	BasicAuthType         = "basic"
-	NtlmAuthType          = "ntlm"
-	NegotiateAuthType     = "negotiate"
+	BasicAuthType     = "basic"
+	NtlmAuthType      = "ntlm"
+	NegotiateAuthType = "negotiate"
+)
+
+var (
+	AuthenticateHeaders = []string{"Www-Authenticate", "Lfs-Authenticate"}
 )
 
 func (e *ClientError) Error() string {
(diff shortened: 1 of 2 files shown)


httputil/request.go
```diff
@@ -172,14 +180,19 @@ func SetAuthType(req *http.Request, res *http.Response) {
...
 
 func GetAuthType(res *http.Response) string {
-	auth := res.Header.Get("Www-Authenticate")
-	if len(auth) < 1 {
-		auth = res.Header.Get("Lfs-Authenticate")
-	}
 
-	if strings.HasPrefix(strings.ToLower(auth), "ntlm") {
-		return "ntlm"
+	for _, headerName := range []string{WwwAuthenticateHeader, LfsAuthenticateHeader} {
```
ttaylorr: The inner contents of this loop look good. I wonder if we could make things a little bit clearer by making removing the `WwwAuthenticateHeader` and `LfsAuthenticateHeader` `consts` in favor of a single:

```go
var (
        AuthenticateHeaders = []string{"Www-Authenticate", "Lfs-Authenticate"}
)
```

since we never use them individually, only in this form on line 184.

httputil/request.go
```diff
@@ -172,14 +180,19 @@ func SetAuthType(req *http.Request, res *http.Response) {
...
+		authHeaders := res.Header[headerName]
```
ttaylorr: I don't think we need this temporary variable. We should shorten this to:

```go
for _, auth := range res.Header[http.CannonicalHeaderKey(headerName)] {
        // ...
}
```

httputil/request_test.go
```diff
@@ -0,0 +1,76 @@
+package httputil
+
+import (
+	"github.com/stretchr/testify/assert"
```
ttaylorr: This is a little bit pedantic, but I think `gofmt` would format these imports like:

```go
import (
        "net/http"
        "testing"

        "github.com/stretchr/testify/assert"
)
```

httputil/request_test.go
```diff
@@ -0,0 +1,76 @@
...
+		"WWW-Authenticate": {"Basic"},
+		"LFS-Authenticate": {"Ntlm"},
+	}
+
+	authType := GetAuthType(httpResponseFromHeaders(headers))
+
+	assert.Equal(t, NtlmAuthType, authType)
+}
+
+func httpResponseFromHeaders(headers map[string][]string) *http.Response {
```
ttaylorr: OK, I'm going going to make one more annoying suggestion about this function. I hope you'll forgive me 😜 

All of the tests in this file follow the same structure:

1. Initialize some headers.
2. Get the resulting auth type from those headers.
3. Assert that it equals some other auth type.

I'm proposing the following test-case struct, and the following test structure:

```go
type AuthenticateHeaderTestCase struct {
        AuthType string
        Headers map[string][]string
}
```

func (c *AuthenticateHeaderTestCase) Assert(t *testing.T) {
        t.Log("lfs/httputil: asserting auth type: %q for: %v", c.AuthType, c.Headers)

        assert.Equal(t, c.AuthType, GetAuthType(c.HttpResponse())
}

func (c *AuthenticateHeaderTestCase) HttpResponse() *http.Response {
        res := &http.Response{Header: make(http.Header)}
        for k, vv := range c.Headers {
                for _, v := range vv {
                        res.Header.Add(k, v)
                }
        }

        return res
}
```

and the following test structure:

```go
func TestAuthenticationCases(t *testing.T) {
        for _, c := range []AuthenticateHeaderTestCase{
                {NtlmAuthType, map[string]string{"Www-Authenticate" : {"Basic", "Negotiate"}}},
                {NtlmAuthType, map[string]string{"LFS-Authenticate" : {"Basic", "Ntlm"}}},
                // ...
        } {
                c.Assert(t)
        }
}
```

This is akin to the test-case pattern found in the [`batcher_test.go`](https://github.com/github/git-lfs/blob/d25c0eb03e45d788f971ebd50f4a47772db39ac7/lfs/batcher_test.go) file.

Feel free to not go with this implementation, I don't mind either way :smile:. What you have right now works great, and is cool with me, too :+1:

ttaylorr:
Thanks for re-opening this PR 😄 . Few more comments from me, but this looks great!

Like I said, feel free to not go with my test-case suggestion. I'd be fine merging this in either way.

httputil/request.go
VladimirKhvostov: I agree. I was thinking about this, but was not sure what is the right way to do this in GO lang. Thanks for teaching me best practices.

httputil/request.go
VladimirKhvostov: I C# code, I like to create temp variables in cases like this - it makes debugging easier in some cases. For example, I can set breakpoint when variable is set to specific value. I remove temp variable.

httputil/request_test.go
VladimirKhvostov: Makes sense. I updated the code. I was not aware of this pattern.

httputil/request_test.go
VladimirKhvostov: I like this idea. I updated test code. Thanks!

ttaylorr:
Great, thanks! 👍

technoweenie:
Thanks for the patch! I made a few changes in 29e8876427183ec01e3b88e14720c68ab5d2007c so that the `httputil` package doesn't export values like `BasicAuthType` or `AuthenticateHeaders`. Those are only used for the internals of the `httputil` package, so they don't need to be exported.

PR #1382, opened 2016-07-18
WIP: long running clean/smudge filter protocol

larsxschneider:
This is my first WIP version of a Git/Git-LFS stream filter. Please keep in
mind that I have little go-lang knowledge and experience. Therefore
I would be happy to receive a very strict review to improve my go-lang
skills 😄 👍 
## What is the problem with Git LFS?

Git LFS is an application that is executed via Git clean/smudge filter.
The process invocation of these filters requires noticeable time (especially
on Windows). An individual filter process is required for every single file 
that Git touches during its operations (e.g. `checkout` etc).
## Proposed solution

Instead of a single Git LFS process per file, I propose a single Git LFS
process per Git invocation. That means Git invokes the filter process 
(e.g. Git LFS) only once and then continuously talks to the same filter 
process via a pipes.

You can find the corressponding WIP Git core implementation here: 
https://github.com/larsxschneider/git/tree/filter-stream
## Performance tests

I executed both test runs on a 2,5 GHz Intel Core i7 with SSD and OS X.
A test run is the consecutive execution of four Git commands:
1. clone the repo
2. checkout to the "removed-files" branch
3. timed: checkout the "master" branch
4. timed: checkout "removed-files" branch

Test command:

```
set -x; git lfs clone https://github.com/larsxschneider/lfstest-manyfiles.git repo; cd repo; git checkout removed-files; time git checkout master; time git checkout removed-files
```

I compiled Git with the following flags:

```
NO_OPENSSL=YesPlease APPLE_COMMON_CRYPTO=YesPlease NO_GETTEXT=YesPlease make -j 8
```
### TEST RUN A -- Default Git 2.9 (ab7797d) and Git LFS 1.2.1

```
+ git lfs clone https://github.com/larsxschneider/lfstest-manyfiles.git repo
Cloning into 'repo'...
warning: templates not found /Users/lars/share/git-core/templates
Receiving objects: 100% (15012/15012), 2.02 MiB | 1.77 MiB/s, done.
Checking out files: 100% (15001/15001), done.
Git LFS: (15000 of 15000 files) 0 B / 77.04 KB
+ cd repo
+ git checkout removed-files
Branch removed-files set up to track remote branch removed-files from origin.
Switched to a new branch 'removed-files'
+ git checkout master
Checking out files: 100% (12000/12000), done.
Switched to branch 'master'

real    6m2.979s
user    2m39.066s
sys 2m41.610s
+ git checkout removed-files
Switched to branch 'removed-files'
Your branch is up-to-date with 'origin/removed-files'.

real    0m1.310s
user    0m0.385s
sys 0m0.881s
```
### TEST RUN B -- Git and Git LFS with stream filter support

```
+ git lfs clone https://github.com/larsxschneider/lfstest-manyfiles.git repo
Cloning into 'repo'...
warning: templates not found /Users/lars/share/git-core/templates
Receiving objects: 100% (15012/15012), 2.02 MiB | 1.30 MiB/s, done.
Git LFS: (15000 of 15000 files) 0 B / 77.04 KB
+ cd repo
+ git checkout removed-files
Branch removed-files set up to track remote branch removed-files from origin.
Switched to a new branch 'removed-files'
+ git checkout master
Checking out files: 100% (12000/12000), done.
Switched to branch 'master'

real    0m2.528s
user    0m0.209s
sys 0m1.602s
+ git checkout removed-files
Switched to branch 'removed-files'
Your branch is up-to-date with 'origin/removed-files'.

real    0m2.280s
user    0m0.066s
sys 0m0.637s
```
### Results

```
Default Git:                      6m2.979s + 0m1.310s = 364s
Git and Git LFS with stream filter support: 0m2.528s + 0m2.280s = 5s
```

The Git stream filter solution is almost :sparkles: **70x faster** :sparkles: when switching branches
on my local machine with a test repository containing 12,000 Git LFS files.
Based on my previous experience with Git LFS clone I expect even more
dramatic results on Windows.
## Next Steps
1. Make Travis-CI tests pass (anyone an idea what is wrong with the  "clone with submodules" test 1a24a7c1c9b25fdb3a6d5a55c9429e045e60fd89 ?)
2. Make the pipe protocol more robust against errors (e.g. by adding ACK messages). Do you have other suggestions for the protocol?
3. Cleanup Git-core patch, propose patch to mailing list (/cc heads up @peff).
4. Cleanup code duplication in `command_smudge.go`/`command_clean.go` and `command_filter.go`.
## Questions
1. Would you be OK with this approach in general? /cc @technoweenie @sinbad @ttaylorr 
2. How should I handle integration tests? Git LFS would need to support and test both protocols ("per file" and "filter stream"). I was thinking about running all integration tests twice with different Git filter configs.

Thanks,
Lars


.travis.yml
```diff
@@ -10,58 +10,27 @@ os:
...
+      ;;
+    osx)
+      brew install gettext
+      brew link --force gettext
+      NO_OPENSSL=YesPlease APPLE_COMMON_CRYPTO=YesPlease make --jobs=2
+      ;;
+    esac;
+    make install;
+    export PATH=$PWD:$PATH;
+    cd ..;
```
larsxschneider: These Travis-CI changes won't be in the final patch, of course.

test/cmd/lfstest-gitserver.go
```diff
@@ -660,7 +660,7 @@ func gitHandler(w http.ResponseWriter, r *http.Request) {
 		r.Body.Close()
 	}()
 
-	cmd := exec.Command("git", "http-backend")
+	cmd := exec.Command("git-http-backend")
```
larsxschneider: for some reason that was necessary on my local machine to make the integration tests pass

test/test-clone.sh
```diff
@@ -365,81 +365,81 @@ begin_test "clone (with .lfsconfig)"
...
-  popd
-
-
-)
-end_test
+# begin_test "clone with submodules"
+# (
+#   set -e
+
+#   # set up a doubly nested submodule, each with LFS content
```
sinbad: This worked in master/the recent PR it was added for, we'll need to figure out why it breaks with this change.

commands/command_filter.go
```diff
@@ -0,0 +1,215 @@
...
+		if err != nil {
+			Panic(err, "Error reading inputData pointer length for smudging.")
+		}
+		inputDataPtrLen := binary.LittleEndian.Uint32(buf)
+
+		// Read inputData
+		var outputData []byte
+		if inputDataPtrLen > 0 {
+			buf = make([]byte, inputDataPtrLen)
+			_, err = reader.Read(buf)
```
sinbad: Rather than reading all the data into an in-memory buffer then creating a reader on that, you'd be better to just pass `reader` directly to the clean/smudge function.

commands/command_filter.go
```diff
@@ -0,0 +1,215 @@
...
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		requestShutdown = true
+	}()
+
+	reader := bufio.NewReader(os.Stdin)
+	for {
+		buf := make([]byte, 4)
+		readBytes, err := reader.Read(buf)
+		if readBytes == 0 && requestShutdown {
```
sinbad: Should this be `||` rather than `&&`?

commands/command_filter.go
```diff
@@ -0,0 +1,215 @@
...
+
+func filterCommand(cmd *cobra.Command, args []string) {
+	requireStdin("This command should be run by the Git 'clean' filter")
+	lfs.InstallHooks(false)
+
+	requestShutdown := false
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		requestShutdown = true
```
sinbad: I think you meant to have a line above this that said `sig := <-c` so that `requestShutdown` is only set on an interrupt; right now it becomes true as soon as the goroutine runs. 

If this was the primary way of handling interrupts you'd probably also want to change the exit code to non-zero, and also catch `os.Kill`. However see `git-lfs.go` - this already handles both and has a constant goroutine which  will `os.Exit` and which will probably handle this case. Maybe you don't need it?

sinbad:
This is going to be great, it will remove the need for more lfs-specific commands and eventually could lead to us deprecating `git lfs clone` for recent git versions. 

I've started a few comments but haven't finished reviewing yet, will pick up again tomorrow.

commands/command_filter.go
larsxschneider: Yeah, I thought about that. But how do I ensure that the implementation does read exactly `inputDataPtrLen` bytes?

commands/command_filter.go
larsxschneider: I think `&&` is correct. This loop runs in cycles and reads zero bytes if Git is not sending anything (maybe I should add a `sleep`?).

commands/command_filter.go
larsxschneider: You are right. This is not necessary anymore (at some point in my testing it was, but I probably deactivated something important in my test setup 😄 ). Fix: c3bfdd733a08e1f436b3a70e0d37c1901847813c

test/test-clone.sh
larsxschneider: Yes! The test works on my local machine, though. No idea why TravisCI is different. See here for the log:
https://travis-ci.org/larsxschneider/git-lfs/jobs/145543925

Relevant part:
```sh
    +git lfs clone --recursive http://127.0.0.1:37063/submod-root submod-clone
    Cloning into 'submod-clone'...
    Submodule 'sub1' (http://127.0.0.1:37063/submod-level1) registered for path 'sub1'
    Cloning into '/home/travis/git-lfs-tests/test-clone.sh-10051/submod-clone/sub1'...
    Submodule 'sub2' (http://127.0.0.1:37063/submod-level2) registered for path 'sub1/sub2'
    Cloning into '/home/travis/git-lfs-tests/test-clone.sh-10051/submod-clone/sub1/sub2'...
    +cd submod-clone
    +assert_local_object eafdeb40fa710b531ca3aaf6ba737268ed021ec58315697b4cc3a4868f76b6a1 35
    +local oid=eafdeb40fa710b531ca3aaf6ba737268ed021ec58315697b4cc3a4868f76b6a1
    +local size=35
    ++grep LocalMediaDir
    ++git lfs env
    +local cfg=LocalMediaDir=/home/travis/git-lfs-tests/test-clone.sh-10051/submod-clone/.git/lfs/objects
    +local f=/home/travis/git-lfs-tests/test-clone.sh-10051/submod-clone/.git/lfs/objects/ea/fd/eafdeb40fa710b531ca3aaf6ba737268ed021ec58315697b4cc3a4868f76b6a1
    ++tr -d '[[:space:]]'
    ++wc -c
    +actualsize=35
    +'[' 35 '!=' 35 ']'
    ++wc -c
    +'[' 35 -eq 35 ']'
    +cd sub1
    +assert_local_object 757c94395f3a19f920e507ae4b1e85348f9b5c897d18ed24a994cd9414176b65 15
    +local oid=757c94395f3a19f920e507ae4b1e85348f9b5c897d18ed24a994cd9414176b65
    +local size=15
    ++grep LocalMediaDir
    ++git lfs env
    +local cfg=LocalMediaDir=/home/travis/git-lfs-tests/test-clone.sh-10051/submod-clone/.git/modules/sub1/lfs/objects
    +local f=/home/travis/git-lfs-tests/test-clone.sh-10051/submod-clone/.git/modules/sub1/lfs/objects/75/7c/757c94395f3a19f920e507ae4b1e85348f9b5c897d18ed24a994cd9414176b65
    ++tr -d '[[:space:]]'
    ++wc -c
    test/testhelpers.sh: line 35: /home/travis/git-lfs-tests/test-clone.sh-10051/submod-clone/.git/modules/sub1/lfs/objects/75/7c/757c94395f3a19f920e507ae4b1e85348f9b5c897d18ed24a994cd9414176b65: No such file or directory
    +actualsize=
    +'[' 15 '!=' '' ']'
    +exit 1
    +end_test
    +test_status=1
```

commands/command_filter.go
```diff
@@ -0,0 +1,207 @@
...
+
+func filterCommand(cmd *cobra.Command, args []string) {
+	requireStdin("This command should be run by the Git filter")
+	lfs.InstallHooks(false)
+
+	reader := bufio.NewReader(os.Stdin)
+	for {
+		buf := make([]byte, 4)
+		readBytes, err := reader.Read(buf)
+		if readBytes == 0 {
```
ttaylorr: Are you checking whether or not you're at the end of `stdin`? Checking this may be more appropriate:

```go
if _, err := reader.Read(buf); err == io.EOF {
        return
}

// <snip>
```

commands/command_filter.go
```diff
@@ -0,0 +1,207 @@
...
+			return
+		}
+		command := binary.LittleEndian.Uint32(buf)
+
+		if err != nil {
+			continue
+		}
+
+		// Read fileName length
+		buf = make([]byte, 4)
```
ttaylorr: Ah, disregard my comment above. It looks like you're doing a lot of the:

- make buffer
- read into buffer
- check error
- decode little endian uint32

pattern. The [`binary.Read`](https://golang.org/pkg/encoding/binary/#Read) function might help you out in this case. For instance, the first read may look something more like this:

```go
var command uint32
if err := binary.Read(reader, binary.LittleEndian, &command); err == io.EOF {
        return
} else if err != nil {
        continue
}

// <snip>
```

commands/command_filter.go
ttaylorr: You could pass a [`LimitedReader`](https://golang.org/pkg/io/#LimitedReader).

```go
inputData := io.LimitReader(reader, inputDataPtrLen)
```

commands/command_filter.go
```diff
@@ -0,0 +1,207 @@
...
+		if inputDataPtrLen > 0 {
+			buf = make([]byte, inputDataPtrLen)
+			_, err = reader.Read(buf)
+			if err != nil {
+				Panic(err, "Error reading inputData pointer for smudging.")
+			}
+			inputData := bytes.NewReader(buf)
+
+			switch command {
+			case 1:
```
ttaylorr: Perhaps we could promote these to consts?

```go
type FilterOperation uint32

const (
        CleanOperation FilterOperation = iota + 1
        SmudgeOperation
)
```

commands/command_filter.go
```diff
@@ -0,0 +1,207 @@
...
+				outputData, _ = clean(inputData, fileName)
+			case 2:
+				outputData, _ = smudge(inputData, fileName)
+			default:
+				panic("Unrecognized filter command")
+			}
+		}
+
+		writer := bufio.NewWriter(os.Stdout)
```
ttaylorr: Do we need to re-instantiate this buffered version of STDOUT during each loop invocation?

commands/command_filter.go
```diff
@@ -0,0 +1,207 @@
...
+		binary.Write(writer, binary.LittleEndian, resLength)
+		if resLength > 0 {
+			writer.Write(outputData)
+		}
+		writer.Flush()
+	}
+}
+
+func init() {
+	filterCmd.Flags().BoolVarP(&filterSmudgeSkip, "skip", "s", false, "")
```
ttaylorr: Does this apply both to the clean and smudge filter, or just the smudge filter?

commands/command_filter.go
```diff
@@ -0,0 +1,207 @@
...
+			fileName = string(buf)
+		}
+
+		// Read inputData length
+		buf = make([]byte, 4)
+		_, err = reader.Read(buf)
+		if err != nil {
+			Panic(err, "Error reading inputData pointer length for smudging.")
+		}
+		inputDataPtrLen := binary.LittleEndian.Uint32(buf)
```
ttaylorr: This may be a good case to introduce a `type InputDataHeader struct`. I'm thinking something like:

```go
type InputDataHdr struct {
        Name string
        Len  uint32
}

func (h *InputDataHdr) Read(r io.Reader) error {
        // Read name length, check error, buffer the `[]byte`, read the name itself
        // Read the next uin32, which is the input data length
}
```

commands/command_filter.go
```diff
@@ -0,0 +1,207 @@
...
+
+	return buf.Bytes(), nil
+}
+
+func filterCommand(cmd *cobra.Command, args []string) {
+	requireStdin("This command should be run by the Git filter")
+	lfs.InstallHooks(false)
+
+	reader := bufio.NewReader(os.Stdin)
+	for {
```
ttaylorr: Was wondering about having a type that could accept a `chan` of these `InputData` header types, and then hand them off to the appropriate processor for processing. Maybe something like:

```go
type Processor interface {
        Process(h *InputDataHdr, r io.Reader) (*OutputData, error)
}

// TODO: not a great name :-/
type Dispatcher struct {
        reader io.Reader
        writer io.Writer

        In <-chan *InputDataHdr
        Out chan *OutputData


        closer chan struct{}
        wg sync.WaitGroup
}

func (d *Dispatcher) Dispatch() error {
        for {
                select {
                case hdr := <-d.in:
                        d.wg.Add(1)
                        // Pick the appropriate processor, have it process
                        go func() { defer d.wg.Done(); d.Out <- processor.Process(hdr) }()
                case out := <-d.Out:
                        // Write out
                case <-d.closer:
                        d.wg.Wait()
                        return
                }
        }
}
```

I'm not attached at all to the implementation that I proposed, but I do want to get as much processing logic out of the `commands` package as possible 😄

ttaylorr:
I like this approach. This will definitely be a huge speed gain for what are currently pretty slow operations under certain scenarios.

I made a few comments on some the code above, mainly boiling down to a few suggestions:

1. A `InputDataHdr` struct
2. Using `binary.Read` instead of buffering and calling `binary.LittleEndian.Uint32(...)`
3. A `Processor` type

I think that the header/data idea is a good one for this protocol. I am wondering, however, about the benefits of implementing a multiplexed chunked protocol. Would it make sense to interleave the data that was being sent across these file descriptors? I am not entirely sure.

On one hand, this would allow the clean and smudge filters to start processing one file before they had finished others, which would enable increased parallelism.

On the other hand, perhaps this sort of optimization is not necessary. One concern with this approach is the additional complexity that would be incurred by this sort of muxing. I am thinking in particular of the approach that is implemented in the RTMP protocol, which is certainly complex. The relevant parts of the documentation can be found [here](wwwimages.adobe.com/content/dam/Adobe/en/devnet/rtmp/pdf/rtmp_specification_1.0.pdf), and a reference implementation that I wrote in Go can be found [here](https://github.com/WatchBeam/rtmp/tree/master/chunk) ([docs](http://godoc.org/github.com/WatchBeam/rtmp/chunk)). There is _far_ more going on in that `chunk` package than we would implement in LFS, but it'd still be more than we have now.

Your thoughts?

commands/command_filter.go
sinbad: No `Read` will i/o block if there's nothing in the stream, it won't return 0 unless the stream is closed (in which case err == EOF). You don't need a sleep because Read won't return until there's some data or an EOF/stream error.

In that sense I don't know what `&& requestShutdown` is supposed to be doing. Either one of an interrupt or a stream close should exit the loop, surely? But as I say in another comment you probably don't need to handle the interrupts anyway since git-lfs.go does and there's no special cleanup code here.

commands/command_filter.go
larsxschneider: Agreed. Thanks! https://github.com/larsxschneider/git-lfs/commit/59115d5e952a018ebe7862f350d5b59f8ea83ed3

commands/command_filter.go
larsxschneider: Genius! https://github.com/github/git-lfs/pull/1382/commits/6e111003cd7318b7d8afc2f912b792a171312904

commands/command_filter.go
larsxschneider: Agreed! https://github.com/github/git-lfs/pull/1382/commits/0deb0aa84d30a1b51aca85403a553747ad92e236

commands/command_filter.go
larsxschneider: No. https://github.com/github/git-lfs/pull/1382/commits/479ede0e6fa2fb90939e8ba310a7cde491282c74

commands/command_filter.go
larsxschneider: It just applies to the smudge filter to mimic the existing behavior.

larsxschneider:
@sinbad and @ttaylorr: Thanks a lot for your feedback!

**Re: Multiplexing**
Multiplexing would considerably complicate the protocol and is therefore
more error prone. Plus, with the Git clean/smudge interface that we have
today, the parallelism wouldn't buy us anything as Git processes the
files sequentially anyway. However, I have some vague ideas how we could
approach that. Therefore, I plan to define a "protocol version" field
that allows Git to support different filter protocols in the future.

commands/command_filter.go
larsxschneider: You're right. Thanks!

commands/command_filter.go
larsxschneider: OK!

Questions:
1. Are one letter variables in functions expected "go-style"?
2. Is this the correct way to handle errors? See: https://github.com/github/git-lfs/pull/1382/commits/268f2369567b9ad69630dd703efbf2cada8c5da6

commands/command_filter.go
larsxschneider: That's a lot of "go" foo for me to grasp but I think I have an idea what is going on/what you want. However, if we can't process the files in parallel then the `chan` would buy us anything, right?

commands/command_filter.go
larsxschneider: Ah, I just got what you mean. I will rename the parameter to `--skip-smudge`, ok?

commands/command_filter.go
```diff
@@ -20,6 +20,37 @@ const (
...
+	if err := binary.Read(r, binary.LittleEndian, &fileNameLen); err != nil {
+		return errutil.Errorf(err, "Error reading filename length.")
+	}
+
+	// Read file name
+	h.FileName = ""
+	if fileNameLen > 0 {
+		buf := make([]byte, fileNameLen)
+		readLen, err := r.Read(buf)
+		if err != nil || readLen != int(fileNameLen) {
```
ttaylorr: Hmm, I would split out these two cases individually. Returning the error if it's `!= nil` would be the first, and the unexpected EOF would be the second. This makes things a little clearer since we can get an "error" when the read succeeded, just read less than we wanted. I'm thinking:

```go
if readLen, err := r.Read(buf); err != nil {
        return errutil.Errorf(err, "Unexpected error")
} else if readLEn != int(fileNameLen) {
        return fmt.Errorf("unexpected EOF when reading file (got %d, wanted %d)", readLen, fileNameLen)
}
```

commands/command_filter.go
```diff
@@ -20,6 +20,37 @@ const (
...
+
+func (h *InputFileHdr) Read(r io.Reader) error {
+	// Read file name length
+	var fileNameLen uint32
+	if err := binary.Read(r, binary.LittleEndian, &fileNameLen); err != nil {
+		return errutil.Errorf(err, "Error reading filename length.")
+	}
+
+	// Read file name
+	h.FileName = ""
```
ttaylorr: Is there a reason you have to explicitly empty this out? Go's zero-value initialization will make this string be `""` by default.

commands/command_filter.go
ttaylorr: 1. Yes!
2. Left a few comments, but mostly yes 😄

commands/command_filter.go
ttaylorr: I would love to chat about this in IRC with you. Right now the `chan` buys us nothing since we can't write the files out out of order, but it does give us the benefit of not having to change a lot of code if/when we do introduce that behavior.

commands/command_filter.go
larsxschneider: OK. I didn't know that.

git/git.go
```diff
@@ -733,15 +733,18 @@ func CloneWithoutFilters(flags CloneFlags, args []string) error {
 	// not working. You can get around that with https://github.com/kr/pty but that
 	// causes difficult issues with passing through Stdin for login prompts
 	// This way is simpler & more practical.
-	filterOverride := ""
+	filterDriverOverride := ""
```
ttaylorr: Empty variable initializations are typically done in go using the `var` block. I would write this like so:

```go
var (
        filterDriverOverride bool
        smudgeFilterOverride string
)
```

and then use the appropriate [`fmt` verbs](https://golang.org/pkg/fmt/) to turn the above `string` and `bool` into `cmdargs` down below 😄

lfs/setup.go
```diff
@@ -26,8 +26,7 @@ var (
 	filters = &Attribute{
 		Section: "filter.lfs",
 		Properties: map[string]string{
```
ttaylorr: I wonder if we'll have to support a way to "upgrade" these (similar to how we treat the hooks in the `lfs` package, according to https://github.com/github/git-lfs/pull/1083)

git/git.go
larsxschneider: Thanks! 👍  https://github.com/github/git-lfs/pull/1382/commits/9aebb9caed58492bc3b2944bc58b4585ded1e5cb

Although "false" is actually a string because I am referring to the bash built in:
http://tldp.org/LDP/abs/html/internal.html 

> **false**: A command that returns an unsuccessful exit status, but does nothing else.

lfs/setup.go
larsxschneider: This is just my test setup. Eventually I would keep "clean" and "smudge" because the Git implementation will fall back to these commands if the "driver" fails... this would be a good safety net especially when the new implementation is rolled out for the first time... (although then we won't learn about the errors 😄 ).

BTW: do you think "driver" is an acceptable name? if not, do you have better ideas?

test/cmd/lfstest-gitserver.go
technoweenie: Weird that it only affects that command...

commands/command_env.go
```diff
@@ -47,7 +47,7 @@ func envCommand(cmd *cobra.Command, args []string) {
 		Print(env)
 	}
 
-	for _, key := range []string{"filter.lfs.smudge", "filter.lfs.clean"} {
+	for _, key := range []string{"filter.lfs.driver"} {
```
technoweenie: Why was this changed?

commands/command_filter.go
```diff
@@ -0,0 +1,221 @@
...
+		Panic(err, "Error reading input data length.")
+	}
+	fileLenStr = strings.TrimSuffix(fileLenStr, "\n")
+	fileLen, err := strconv.ParseUint(fileLenStr, 10, 64)
+	if err != nil {
+		Panic(err, fmt.Sprintf("Error parsing input data length '%s'.", fileLenStr))
+	}
+	h.FileLen = fileLen
+
+	// Panic(err, fmt.Sprintf("Errorddddddddddddgth '%d'.", fileLen))
```
technoweenie: You can use `tracerx.Printf()` to dump trace info when `GIT_TRACE=1` is set :)

commands/command_filter.go
```diff
@@ -0,0 +1,221 @@
...
+		file.Close()
+	}
+
+	if cleaned != nil {
+		defer cleaned.Teardown()
+	}
+
+	if errutil.IsCleanPointerError(err) {
+		// TODO: What is happening here? (triggered by ./test/test-clone.sh)
+		return errutil.ErrorGetContext(err, "bytes").([]byte), nil
```
technoweenie: This error comes up if it's trying to run an LFS pointer through `git lfs clean`, instead of the actual contents of the file.

commands/command_filter.go
```diff
@@ -0,0 +1,221 @@
...
+			default:
+				panic(fmt.Sprintf("Unrecognized filter command '%s'.", command))
+			}
+		}
+
+		writer.WriteString(fmt.Sprintf("%d\n", len(outputData)))
+		writer.Flush()
+		if len(outputData) > 0 {
+			writer.Write(outputData)
+		}
```
technoweenie: It's probably not a good idea to let `smudge()` return a `[]byte` of the binary file contents. It may be better to just pass `outputData` to `clean()` and `smudge()` instead.

larsxschneider:
**relevant to Git LFS in general:**
I am writing tests for the Git core side of the protocol and I discovered that Git calls `clean` way more than necessary: http://thread.gmane.org/gmane.comp.version-control.git/300028

commands/command_filter.go
rubyist: :+1: This definitely needs to be re-thought. `ptr.Smudge` is intended to write to a stream, not an in-memory buffer. This is potentially gigabytes of data.

commands/command_filter.go
```diff
@@ -0,0 +1,221 @@
...
+
+	return buf.Bytes(), nil
+}
+
+func filterCommand(cmd *cobra.Command, args []string) {
+	requireStdin("This command should be run by the Git filter")
+	lfs.InstallHooks(false)
+
+	reader := bufio.NewReader(os.Stdin)
+	writer := bufio.NewWriter(os.Stdout)
```
rubyist: stdio buffering should already be handled by the kernel, is there a reason we're buffering stdout here?

commands/command_filter.go
```diff
@@ -0,0 +1,221 @@
...
+
+	writer.WriteString(fmt.Sprintf("git-filter-protocol\nversion %d", FilterDriverVersion))
```
rubyist: This could be a `fmt.Fprintf()`, and it should probably handle the error. I notice there is no newline after the version in this protocol, how does git know how much data to read here?

rubyist:
It'll be pretty awesome if this makes it into git! I have two primary critiques so far. I think there's too much use of `Panic()` in deeper parts of the code, and the protocol parsing is happening in multiple places. I don't think the `InputFileHdr` is quite the right abstraction and the `Read` method on it is a little awkward. As it is, this parsing code would be very difficult to unit test.

I think something similar to `bufio.Scanner` would look pretty clean here, leaving us with something like this:

```go
func filterCommand() {
	// ...

	scanner := NewObjectScanner(os.Stdin)
	for scanner.Scan() {
		obj := scanner.Object()
		r := obj.Reader()

		// ...

		switch obj.Command {
		case cmdClean:
			clean(r, obj.Name)
		case cmdSmudge:
			smudge(r, obj.Name)
		}
	}

	if err := scanner.Err(); err != nil {
		// ...
	}

	// write output
}

```

[Here's a gist](https://gist.github.com/rubyist/7a8c43d18913e3ca1bdc3fb352660822) with a quick pass at an implementation.

ttaylorr:
Oh snap, that's way better. I like the idea of an object scanner, that puts the parsing implementation in the right place I think. Still not sold on the `switch` block, there may be some other abstraction that we can reach for, but I think this is awesome.

rubyist:
There probably is a better abstraction there, could be a part of the work to fix [this part](https://github.com/github/git-lfs/pull/1382#discussion_r71786735). If I'm thinking right, git is only performing one operation at a time per process invocation, so I'm not sure clean/smudge are really appropriate at the per-object level. I could be (probably am) overlooking something, though, it's been a while since I've dug around down there.

larsxschneider:
Thanks for all the feedback. I'll address it shortly. I just posted the Git core patch: http://thread.gmane.org/gmane.comp.version-control.git/300057 😅

larsxschneider:
Update: This PR is not stall, but I focus on getting the necessary changes into Git first:
http://public-inbox.org/git/20160722154900.19477-1-larsxschneider%40gmail.com/T/#t

larsxschneider:
@sinbad @ttaylorr @technoweenie :
If you have time, please take a quick look at the protocol:
https://github.com/larsxschneider/git/blob/e128326070847ac596e8bb21adebc8abab2003fc/Documentation/gitattributes.txt#L382-L457

The currently proposed protocol is a little different from what I've implemented here. Most notably  the data is split into "pkt-line" packets. This shouldn't be a problem in golang as I think I can implement a Reader/Writer to handle this transparently.

If you don't like the mailing list then feel free to add comments in my PR 😄 :
https://github.com/larsxschneider/git/pull/7

larsxschneider:
**Update:** I posted v5 of the necessary Git core changes today
http://public-inbox.org/git/20160810130411.12419-1-larsxschneider%40gmail.com/

larsxschneider:
**Update:** I posted v6 of the necessary Git core changes today
http://public-inbox.org/git/20160825110752.31581-1-larsxschneider@gmail.com/

On GitHub:
https://github.com/larsxschneider/git/pull/10

larsxschneider:
**Update:** I posted v7 of the necessary Git core changes today
http://public-inbox.org/git/20160908182132.50788-1-larsxschneider@gmail.com/

On GitHub:
https://github.com/larsxschneider/git/pull/11

larsxschneider:
Believe it or not but I am still working on the related Git core patch. I just posted v9:
http://public-inbox.org/git/20161004125947.67104-1-larsxschneider@gmail.com/

On GitHub: https://github.com/larsxschneider/git/pull/13

larsxschneider:
Junio merged the Git core patch series required for this PR to `next` ( git/git@ffd0de0 ) and it is scheduled for master ( http://public-inbox.org/git/xmqqk2d2ein7.fsf@gitster.mtv.corp.google.com/ ) 🎉 🎉 🎉 .
I'll update this PR with a working version soon, so that we can discuss how to use the new Git filter protocol feature in the most efficient way.

sinbad:
Woooo, amazing work Lars. 🤘

ttaylorr:
@larsxschneider great to hear. I'm looking forward to seeing your up-to-date version of this PR once it's ready. I'll be around to provide any help or answer any questions that you might have.

technoweenie:
@larsxschneider Congrats, great work!

larsxschneider:
I didn't had much time today and I won't have too much time in the next 1.5 weeks. That's why I push a *very* rough versions here. A few thoughts:

1. The implementation as it is right now would smudge files from the Git LFS cache or download files if they are not in the cache. That works but we cannot take advantage of parallel downloads. Ideally I would extend the filter protocol in a way that a filter can tell Git "I can't process this file just yet, ask me later again". However, this wouldn't be straight forward in Git core and the filter protocol change was already pretty large. Until the filter protocol is improved that way we could use a trick, though. If Git asks Git LFS for a file that is not in cache, then Git LFS could return the pointer file as-is and download the file in background (maybe even batched with other files). When Git shuts down, then it sends an EOF to Git LFS via pipe and waits until Git LFS terminates. In this step Git LFS could finish the downloads and replace the pointer files with the actual content. This should give us `git lfs clone` like speed and we could get rid of the wrapper commands. Credit: I think @sinbad posted that idea somewhere but I can't find the link.

1. `ReadRequest` reads the entire content into a `[]byte` which is then passed to the `clean/smudge` functions. I would rather pass a `Reader` to `clean/smudge`.

1. `clean/smudge` returns a `[]byte` which is then written to the pipe in PktLine format using `WriteResponse`. I would rather pass a `Writer` to `clean/smudge`.

1. `clean/smudge` functions in `command_filter.go` are pretty much the same as the ones in `command_clean.go` and `command_smudge.go`. I want to reuse code here.

1. The `process filter` takes precedence over `clean/smudge filter`. That's nice as it makes the filter update for the user very easy. However, I would like to ensure in the tests that Git versions > 2.10 actually use the `process filter`. I am thinking in the direction of an env variable (see `GIT_LFS_USE_LEGACY_FILTER`) to control that. Any good idea how to do that in a nice way?

1. Unit test are missing for `git/git_filter_protocol.go`

git/git_filter_protocol.go
```diff
@@ -0,0 +1,248 @@
...
+	requestList, err := o.readPacketList()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	requestMap := make(map[string]string)
+	for _, pair := range requestList {
+		v := strings.Split(pair, "=")
+		requestMap[v[0]] = v[1]
+	}
```
larsxschneider: @peff I finally grokked your "parse as dictionary" comment completely. As you have already recognized it parses this nicely:
```
packet:          git> command=smudge
packet:          git> pathname=path/testfile.dat
```
But not this:
```
packet:          git> capability=clean
packet:          git> capability=smudge
```
... I wonder what you think about this?
```
packet:          git> clean=capability
packet:          git> smudge=capability
```
However, it's probably too late to change it. I guess in the end it is not that important and I don't want to annoy Junio with this kind of late change now that the code is in `next`.

peff: Well, it's always nice to be vindicated eventually. ;)

If you feel there's room for improvement in the protocol, I don't think being in `next` is too late. You are welcome to submit patches on top, and nothing is cemented until the feature is in a released version of git.

It's up to you whether you think the change is worth making on top.

peff: > It's up to you whether you think the change is worth making on top.

To clarify: this is whether you think it's worth your time in dealing with the patch and any additional review.

From the maintainer's perspective, I think "I implemented a protocol, and now that it is close to cemented, I was fleshing out the other end of the protocol and realized there are some deficiencies" seems like a perfectly good reason to add more patches to the original topic.

technoweenie:
@larsxschneider :+1: on your 6 points. Regarding the first one:

> When Git shuts down, then it sends an EOF to Git LFS via pipe and waits until Git LFS terminates. 

How is this handled in the filter? Would it be after the `for` loop exits, just before `filterCommand()` returns?

ttaylorr:
> Unit test are missing for git/git_filter_protocol.go

I'm working on this right now and will post a PR to merge into this one as soon as I have the tests fleshed out! 🤘

sinbad:
Short of time this week but I totally agree with your "trick" in point 1 (I think we discussed it before as you said) - just smudging to the pointer in serial and doing the actual download in batch/parallel in the background, gated on the termination at the end seems like the best approach. The one Q I had outstanding about that is whether the stat info in the index would then be out of date and would need a `git update-index` to avoid files being displayed as modified.

larsxschneider:
@sinbad I haven't worked with `git update-index`, yet. Here is what would need to happen:
1. Files that GitLFS does not have in cache would end up as LFS pointer files in the working tree
2. After Git is done, GitLFS exchanges the pointer files with the actual content. I guess then we need to tell Git that the new content is OK and the working tree clean. That's what `git update-index` is for ?!

sinbad:
@larsxschneider yeah what I'm not sure about is exactly when Git updates the index for a file; it must be after the smudge filter is run so the size & date are correct, but I don't know if it does it _immediately_ after calling the filter, or at the end of the entire checkout. If it can do it at any time before LFS replaces the pointer file with the real content then the `stat` in the index will be out of date and probably a `git update-index` will be needed to avoid it appearing to be modified.

larsxschneider:
I close this PR as the work is continued in https://github.com/github/git-lfs/pull/1617

PR #1386, opened 2016-07-21
commands/test: make lock commands opt-in using GITLFSLOCKSENABLED=1

ttaylorr:
To prepare for the upcoming release of Git LFS v1.3, the `git-lfs lock` and
`git-lfs locks` commands are now hidden behind a enviornment variable
`GITLFSLOCKSENABLED`. Since most (all?) implementations of LFS do not yet
support this new locking API, it makes sense to allow users to experiment with
the command while at the same time, not making it a "full" part of the release.

If users wish to run any `git-lfs lock{,s}` command, they may do so according
to the following:

```
$ GITLFSLOCKSEANBLED=1 git lfs lock <flags> [args]
$ GITLFSLOCKSEANBLED=1 git lfs locks <flags> [args]
```

This commit, when applied, adds two new things:

1. The `isCommandEnabled(cmd string) bool` func in the commands package. This
   new func checks the OS's enviornment variables and determines whether or not
   a command is "enabled".

2. Updates the `lock` and `locks` tests to use the
   GITLFSLOCKSENABLED=1 flag such that they are able to run.

Since the `isCommandEnabled` func defaults to "false", it should only guard
commands which are deemed "experimental", as noted in the godoc.

-------

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
@@ -251,6 +252,19 @@ func help(cmd *cobra.Command, args []string) {
 func usage(cmd *cobra.Command) error {
 	printHelp(cmd.Name())
 	return nil
+}
+
+// isCommandEnabled returns whether the environment variable GITLFS<cmd>ENABLED
+// is equal to 1, returning false by default if the enviornment variable is not
+// specified.
+//
+// This function call should only guard commands that do not yet have stable
+// APIs or solid server implementations.
+func isCommandEnabled(cmd string) bool {
+	env := os.Getenv(fmt.Sprintf("GITLFS%sENABLED", strings.ToUpper(cmd)))
+	v, _ := strconv.Atoi(env)
+
+	return v == 1
 }
 
 func init() {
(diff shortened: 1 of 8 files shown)


commands/commands.go
```diff
@@ -253,6 +254,19 @@ func usage(cmd *cobra.Command) error {
...
 
+// isCommandEnabled returns whether the environment variable GITLFS<cmd>ENABLED
+// is equal to 1, returning false by default if the enviornment variable is not
+// specified.
+//
+// This function call should only guard commands that do not yet have stable
+// APIs or solid server implementations.
+func isCommandEnabled(cmd string) bool {
+	env := os.Getenv(fmt.Sprintf("GITLFS%sENABLED", strings.ToUpper(cmd)))
+	v, _ := strconv.Atoi(env)
```
technoweenie: [`strconv.ParseBool`](https://golang.org/pkg/strconv/#ParseBool) accepts 1, 0, t, f, etc.

ttaylorr: As discussed, went for [`config.Configuration.GetenvBool`](https://godoc.org/github.com/github/git-lfs/config#Configuration.GetenvBool) in https://github.com/github/git-lfs/pull/1386/commits/ab9842c3f97113e8c304b32f4040fbb2284dd9a5. 😄

PR #1389, opened 2016-07-21
Skip sending the `transfers` key if custom transfer adapters are disabled

technoweenie:
We have old versions of GH Enterprise that still support only the v1 Batch API.  This gives us a way to let customers use Git LFS v1.3 until their servers are updated.

/cc @sinbad


sinbad:
Good call, I don't have any test servers that are that strict and I didn't know GHE was. However I don't think this is going to fix it unless the client also defines `lfs.basictransfersonly=true` in their gitconfig. Without this the `basic` transfer will still be in the list since it's added on `init`. If we took that out it would make the defaulting more complicated, so I suggest:

```diff
diff --git a/api/api.go b/api/api.go
index c5873a8..f84c03e 100644
--- a/api/api.go
+++ b/api/api.go
@@ -54,6 +54,11 @@ func Batch(objects []*ObjectResource, operation string, transferAdapters []strin
                return nil, "", nil
        }

+       // Compatibility; omit transfers list when only basic
+       // older schemas included `additionalproperties=false`
+       if len(transferAdapters) == 1 && transferAdapters[0] == "basic" {
+               transferAdapters = nil
+       }
        o := &batchRequest{Operation: operation, Objects: objects, TransferAdapterNames: transferAdapters}
        by, err := json.Marshal(o)
        if err != nil {
```

Downside is having to use `"basic"` instead of `transfer.BasicAdapterName` because of import cycles; could avoid this by doing it in the callers to `api.Batch()` but it feels more natural here.

sinbad:
(if you add that change I'd revert the changes to `transfer.go` because they're no longer needed)

sinbad:
👍

technoweenie:
Thanks for the suggestion, but this still needs `lfs.basictransfersonly=true` for uploads because of the `tus` adapter.

> Good call, I don't have any test servers that are that strict and I didn't know GHE was.

I fixed it on GH when you asked, but it takes a little while to filter down to older GHE versions (assuming customers upgrade them in a timely manner).

sinbad:
> Thanks for the suggestion, but this still needs lfs.basictransfersonly=true for uploads because of the tus adapter.

Argh, of course. Sorry, was really hoping this wouldn't be breaking for anyone 😞 `additionalproperties=false` is evil...

technoweenie:
Any thoughts on dark shipping the `tus` adapter (by flipping the logic so it's only attempted if something like `lfs.tustransfers=true` is on)? I don't feel super comfortable shipping the new adapters without real use. Even the batch API originally shipped disabled, and with a fallback to the legacy api, so we could have people opt in specifically to try it out first.

I realize that the API change was done correctly, and it's only due to our

sinbad:
Yeah, agreed. Would be nice to make it automatic at some point in the future if we can though, after there are more tus server implementations out there.

config/config.go
```diff
@@ -244,6 +244,22 @@ func (c *Configuration) BasicTransfersOnly() bool {
 	return basicOnly
 }
 
+// TusTransfersAllowed returns whether to only use "tus.io" HTTP transfers.
+// Default is false, including if the lfs.tustransfers is invalid
+func (c *Configuration) TusTransfersAllowed() bool {
+	value, ok := c.GitConfig("lfs.tustransfers")
```
ttaylorr: Looks like we have the [`GitConfigBool`](https://github.com/github/git-lfs/blob/c2ae16a30553ec38a21af5e174613f14c6d6fc0c/config/config.go#L424-L437) func which may simplify some of this 😄

transfer/transfer.go
```diff
@@ -209,5 +209,8 @@ func initCoreAdaptersIfRequired() {
 	// That's why this isn't in an init() block
 	initCoreOnce.Do(func() {
 		ConfigureCustomAdapters()
+		if !config.Config.TusTransfersAllowed() {
```
ttaylorr: Maybe we could add a small note that this is strictly for the darkship in the 1.3 release?

technoweenie:
> Would be nice to make it automatic at some point in the future if we can though, after there are more tus server implementations out there.

Yeah, def. I think once the [lfs-test-server](https://github.com/git-lfs/lfs-test-server) supports it, it'll be good to go. I think the lfs-test-server should also include a custom one too.

ttaylorr:
Thanks for making those changes! ✨ 

Looks good to me. 👍

sinbad:
You'll need to document `lfs.tustransfers` in `docs/man/git-lfs-config.5.ronn`, and fix the env tests now that it's spitting out a new line (gotta think of a way to automate that one day)

sinbad:
> Yeah, def. I think once the lfs-test-server supports it, it'll be good to go.

Actually I added tus support to lfs-test-server over a month ago ;) Although mainly so that I could test using the official tus server implementation rather than my own, to make sure I'd implemented the protocol correctly. I'm fine with waiting until real-world servers support it though (we don't on Bitbucket yet)

technoweenie:
So you did :) Apparently I wasn't watching that repo for some reason, and totally missed it.

PR #1394, opened 2016-07-25
Bump version number

dakotahawkins:
I'm not sure what the convention is here. Version 1.3.0 is everywhere in this branch except in version.go.


ttaylorr:
Huh, looks like I missed that in the `v1.2.1` release and this one. Good catch!

PR #1397, opened 2016-07-27
Report actual list of transfers that will be tried from `git lfs env`

sinbad:
I found this useful when debugging transfers, above and beyond just showing the booleans or having to dig into the actual JSON that was sent. 

@technoweenie would be great if this could make it into 1.3 but no big deal if not.

### DIFF
--- a/lfs/lfs.go
+++ b/lfs/lfs.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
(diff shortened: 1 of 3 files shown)


technoweenie:
This looks good. Well, `lfs.Environ()` needs to be rethought, but you know what I mean :)

sinbad:
Doh, I missed the same tests you did when adding the tus setting!

PR #1398, opened 2016-07-27
Linux build fixes

technoweenie:
This updates the debian and rpm build scripts to go with https://github.com/andyneff/git-lfs_dockers/pull/3.

/cc @andyneff

### DIFF
--- a/rpm/build_rpms.bsh
+++ b/rpm/build_rpms.bsh
@@ -141,7 +141,7 @@ fi
 pushd ${CURDIR}/..
   #Yes, compile lfs before compiling lfs...
   ./script/bootstrap
-  #Use the version output to grab the version number and short sha 
+  #Use the version output to grab the version number and short sha
   #(that yes, I could have gotten from git myself)
   LFS_VERSION=$(./bin/git-lfs version | sed -r 's|.*/([0-9.]*).*|\1|')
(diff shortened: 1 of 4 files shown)


rpm/build_rpms.bsh
```diff
@@ -152,27 +152,26 @@ popd
...
+echo "Cleaning ${CURDIR}/tmptar"
+rm -rf ${CURDIR}/tmptar
+
 mkdir -p ${CURDIR}/tmptar/git-lfs-${LFS_VERSION}
 pushd ${CURDIR}/..
   #I started running out of space in the docker, so I needed to copy a little less waste
   tar -c . --exclude tmptar --exclude repos | tar -x -C ${CURDIR}/tmptar/git-lfs-${LFS_VERSION}/
 popd
-pushd ${CURDIR}/tmptar/git-lfs-${LFS_VERSION}
-  git clean -xdf
```
andyneff: If I remember correctly the only reason the git clean was needed was to remove any untracked files that were in your working copy when running the scripts. So this is definitely NOT needed, just FYI

ttaylorr:
In accordance with https://github.com/git-lfs/git-lfs_dockers/pull/3#issuecomment-235625615, 👍 .

technoweenie:
@andyneff Thanks, I've made a note of that.

I'm going to merge this now so we can get LFS v1.3 out the door, but I think there are more opportunities to clean the scripts up even more. I plan on opening an issue with some TODOs.

PR #1399, opened 2016-07-27
Trim custom adapter output before tracing to avoid extra newline

sinbad:
Tiny tweak to tracing, makes custom adapters not add newlines everywhere when GIT_TRACE=1


ttaylorr:
Pending a successful build, 👍

sinbad:
Bah, those annoying ephemeral test failures...

PR #1400, opened 2016-07-27
properly report `git update-index` errors

technoweenie:
I've been baffled by the various issues about `git update-index` failing with no reporting. Then it dawned on me that `CombinedOutput()` is the culprit. The [implementation](https://golang.org/src/os/exec/exec.go?s=11924:11970#L433) clearly shows that the stdout and stderr on the command aren't captured until `CombinedOutput()` is called. It's designed for situations where you want to run a simple command and capture it's output in a single line.


dakotahawkins:
Nice find!

technoweenie:
Welp, I had another setback trying to release LFS today. Good news is I can sneak this into v1.3.0 too :metal:

Issue #1403, opened 2016-07-28
sslCAInfo config lookup fails when host in config file doesn't have a trailing slash


dakotahawkins:
For example:

This works for git but not git-lfs:
```
[http "https://github.com"]
    sslCAInfo = ...
```

This works for both:
```
[http "https://github.com/"]
    sslCAInfo = ...
```

We had somebody here have this problem and noticing that config difference was pretty frustrating :)

PR #1404, opened 2016-07-28
Fix sslCAInfo config lookup when host in config doesn't have a trailing slash

dakotahawkins:
Corresponding issue: #1403

appendRootCAsForHostFromGitconfig() modified to check hosts with and without a trailing slash.

Cert unit tests made "table driven" and extended to cover this case.

### DIFF
--- a/httputil/certs.go
+++ b/httputil/certs.go
@@ -51,10 +51,14 @@ func appendRootCAsForHostFromGitconfig(pool *x509.CertPool, host string) *x509.C
 	if cafile := config.Config.Getenv("GIT_SSL_CAINFO"); len(cafile) > 0 {
 		return appendCertsFromFile(pool, cafile)
 	}
-	// http.<url>.sslcainfo
+	// http.<url>/.sslcainfo or http.<url>.sslcainfo
 	// we know we have simply "host" or "host:port"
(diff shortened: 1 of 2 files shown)


dakotahawkins:
From the new test:

Comment out the fix:

    ```
    $ script/test httputil
    --- FAIL: TestCertFromSSLCAInfoConfig (0.35s)
            Error Trace:    certs_test.go:81
            Error:          Not equal: true (expected)
                                    != false (actual)
            Messages:       Cert lookup for "git-lfs.local" should have succeeded with "http.https://git-lfs.local.sslcainfo"

    FAIL
    FAIL    github.com/github/git-lfs/httputil      1.645s
    ```

Re-add the fix:

    ```
    $ script/test httputil
    ok      github.com/github/git-lfs/httputil      1.637s
    ```

dakotahawkins:
Hopefully this can go in release-1.3 but I'm a little unclear on your branching strategy.

technoweenie:
@dakotahawkins This looks good, thanks :metal:. It won't make the cut for v1.3.0 since I already have most of the release packages built. I'll merge this after v1.3.0 is shipped, so it'll be in v1.4.0. I'll also backport it to `release-1.3` so it'll be in v1.3.1.

We have no concrete plans on when v1.3.1 is shipped, but we can do it in a week or so.

PR #1405, opened 2016-07-28
Added branch name (master) to CONTRIBUTING.md

dakotahawkins:
New section (Branching strategy). This could be filled out a bit but I
don't know enough about your workflow to write it up with any more detail.

Explicitly mention master in the PR section.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,16 +49,20 @@ The Git LFS teams mark issues and pull requests with the following labels:
 * `roadmap` - A checklist issue with tasks to fulfill something from the
 [roadmap](./ROADMAP.md)
 
+## Branching strategy
+
+In general, contributors should develop on branches based off of `master` and pull requests should be to `master`.
+
 ## Submitting a pull request
 
 0. [Fork][] and clone the repository
 0. Configure and install the dependencies: `script/bootstrap`
 0. Make sure the tests pass on your machine: `script/test`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -49,16 +49,20 @@ The Git LFS teams mark issues and pull requests with the following labels:
 * `roadmap` - A checklist issue with tasks to fulfill something from the
 [roadmap](./ROADMAP.md)
 
+## Branching strategy
+
+In general, contributors should develop on branches based off of `master` and pull requests should be to `master`.
+
```
ttaylorr: This is :sparkles:! What do you think about adding another small note in this `##` section about our strategy for release branches, i.e., `release-1.2`, `release-1.3` and so on?

technoweenie:
:metal: I had made note to update this, but you beat me to it.

CONTRIBUTING.md
technoweenie: I think we should just add that separately, rather than dictating how it all works to @dakotahawkins so he can just type what we said :)

ttaylorr: Sure, sorry!

dakotahawkins: I would be happy to do it but I don't entirely understand your strategy for those.

I'm used to git-flow, myself, but if somebody would explain how you handle major and point releases (e.g. @technoweenie mentioned a script you use to squash fixes for point releases) I could take a stab at it.

PR #1409, opened 2016-08-01
lfs/hook: teach `lfs.Hook` about `core.hooksPath`

ttaylorr:
This commit gives the `lfs.Hook` knowledge of the new `core.hooksPath`
configuration value that was introduce in Git 2.9.0.

When `core.HooksPath` is found in the Git configuration AND is supported (i.e.,
installed Git binary has a version greater than or equal to "2.9.0"), `git-lfs
install` will place new hooks in that directory. If the `core.hooksPath` is
specified, but the installed verison of Git does NOT support it, then it will
be ignored and installed in `.git/hooks` as per usual.

To test this behavior, two new shell tests were added:
  - One that runs on Git >= 2.9.0, and tests that `core.hooksPath` is respected
  - One that runs on Git < 2.9.0, and tests that `core.hooksPath` is ignored

Unfortunately, our current testing framework does not support skipping
individual tests, only skipping entire files, so two new shell test files were
added.

Resolves github/git-lfs#1407.

--------------

/cc @technoweenie

### DIFF
--- a/lfs/hook.go
+++ b/lfs/hook.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/github/git-lfs/config"
 	"github.com/github/git-lfs/errutil"
+	"github.com/github/git-lfs/git"
 )
 
 // A Hook represents a githook as described in http://git-scm.com/docs/githooks.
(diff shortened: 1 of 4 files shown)


test/test-install-custom-hooks-path.sh
```diff
@@ -0,0 +1,28 @@
...
+
+  hooks_dir="custom_hooks_dir"
+  mkdir -p "$hooks_dir"
+
+  git config --local core.hooksPath "$hooks_dir"
+
+  git lfs install 2>&1 | tee install.log
+  grep "Updated pre-push hook" install.log
+
+  [ -e "$hooks_dir/pre-push" ]
```
technoweenie: Should this have `[ ! -e ".git/hooks/pre-push" ]` too? Also, both tests misspell "perform" :)

lfs/hook.go
```diff
@@ -27,17 +28,30 @@ func (h *Hook) Exists() bool {
...
+
+// Dir returns the directory used by LFS for storing Git hooks. By default, it
+// will return the hooks/ sub-directory of the local repository's .git
+// directory. If `core.hooksPath` is configured and supported (Git verison is
+// greater than "2.9.0"), it will return that instead.
+func (h *Hook) Dir() string {
+	customHooksSupported := git.Config.IsGitVersionAtLeast("2.9.0")
+	if hp, ok := config.Config.GitConfig("core.hooksPath"); ok && customHooksSupported {
+		return hp
+	}
```
technoweenie: This is pretty nitpicky, but there's no point in even checking the git config if `customHooksSupported == false`. Probably fine to keep as-is from a code readability standpoint.

ttaylorr: I think I may keep this as-is because then we'd have to do the `filepath.Join` up front, regardless of whether or not `core.hooksPath` is supported. I have a feeling that a call to `filepath.Join` is not as efficient as a map lookup, but this is such an uncommon operation I think either way would be fine 😄

PR #1411, opened 2016-08-02
distinguish between empty include/exclude paths

### LINKED ISSUE #1249: Command-line switches don't override .lfsconfig file
Currently I have a LFS-enabled repo with the following `.lfsconfig` file:

```
[lfs]
	fetchexclude = spec/fixtures/vcr_cassettes/**/*.yml
```

The expected behavior is that when anyone clones the repo, by default none of the big LFS-enabled .yml files are downloaded.  I cloned the repo and this worked as expected.

Then I ran `git lfs pull -I spec/fixtures/vcr_cassettes/gbcoffice`, expecting the command line options to override the config file and actually download the files I requested, but it did not.

Also tried `git lfs pull -I spec/fixtures/vcr_cassettes/gbcoffice -X ""`, same behavior.

If I move/remove the config file, the commands work as expected, but moving/removing the file should not be needed when I'm being explicit about my intentions via command-line switches.

If you need any other information, feel free to ask.

technoweenie:
This uses a `pflag.Flag`'s `Changed` property to determine if the `--include` or `--exclude` flags are empty or omitted. I'm not convinced using `*string` is any better than setting the default flag value to some magic string like `OPTIONAL`. It does remove any possibility of clashing with a directory in a user's repository, I suppose.

I also ran into some failures in large `test-fetch.sh` tests that run `git lfs fetch` multiple times. Those instances were really difficult to debug, so I split a couple of them up into separate files.

Fixes #1249

### DIFF
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -16,21 +16,19 @@ var (
 		Use: "fetch",
 		Run: fetchCommand,
 	}
-	fetchIncludeArg string
-	fetchExcludeArg string
-	fetchRecentArg  bool
-	fetchAllArg     bool
-	fetchPruneArg   bool
+	fetchRecentArg bool
+	fetchAllArg    bool
+	fetchPruneArg  bool
 )
 
-func getIncludeExcludeArgs(cmd *cobra.Command, includeFlagArg, excludeFlagArg *string) (includeArg, excludeArg *string) {
+func getIncludeExcludeArgs(cmd *cobra.Command) (include, exclude *string) {
 	includeFlag := cmd.Flag("include")
 	excludeFlag := cmd.Flag("exclude")
 	if includeFlag.Changed {
-		includeArg = includeFlagArg
+		include = &includeArg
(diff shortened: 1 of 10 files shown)


commands/command_fetch.go
```diff
@@ -58,11 +69,13 @@ func fetchCommand(cmd *cobra.Command, args []string) {
...
 
 	success := true
+	include, exclude := getIncludeExcludeArgs(cmd)
+
 	if fetchAllArg {
 		if fetchRecentArg || len(args) > 1 {
 			Exit("Cannot combine --all with ref arguments or --recent")
 		}
-		if fetchIncludeArg != "" || fetchExcludeArg != "" {
+		if include != nil || exclude != nil {
```
technoweenie: @ttaylorr This is why I need to call `getIncludeExcludeArgs()` separately.

ttaylorr: Ah, I see. 😄

PR #1415, opened 2016-08-03
config-next: introduce Fetcher, EnvFetcher, config integration

ttaylorr:
This pull request introduces types `Fetcher`, `*EnvFetcher` and integrates them into the `config.Config` type.

In doing so, we are phasing out behavior like:

```go
var cfg *config.Config

cfg.GetenvBool("foo", true)
```

and replacing it with

```go
var cfg *config.Config

cfg.Env.Bool("foo", true)
```

This is the first in a series of many pull requests which will be focused on making the `*config.Config` type oriented towards data-fetching, and less domain-specific. Eventually, all of the domain-specific knowledge will disappear from the `config` package. In order to get there, however, we must first introduce a set of well factored types that will make this sort of data-fetching easy. I have described those types in https://github.com/github/git-lfs/commit/e32fabe37fc150b21a5d2d7594a5300c1388c5cb.

Timeline-wise, I am thinking:

1. Introduce `Fetcher`, implement `EnvFetcher`, and integrate with `*config.Config`.
2. Update tests to not use mutable behavior like `Set` and `SetAll`.
3. Implement the `GitFetcher`.
4. Refactor `Fetcher` type as described in https://github.com/github/git-lfs/commit/e32fabe37fc150b21a5d2d7594a5300c1388c5cb.
5. Piece by piece, remove legacy calls to `GetenvBool`, replacing them with `.Env.Bool` (and etc).
6. Piece by piece, remove domain knowledge from the `*config.Config` type, in favor of the methodology described in https://github.com/github/git-lfs/pull/1390#issuecomment-236068550.

For a more in-depth description, please read the commit bodies of e32fabe37fc150b21a5d2d7594a5300c1388c5cb, b2272a8c165877dc31bf2eaf8f7b9c06cd139315, and f6e8c528abab7889c893274e97202f3fdf325fe1.

-----------------

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -47,6 +47,11 @@ type FetchPruneConfig struct {
 }
 
 type Configuration struct {
+	// Env provides a Fetcher implementation used to access to the system's
+	// environment through os.Getenv. It is the point of entry for all
+	// system environment configuration.
+	Env *EnvFetcher
+
 	CurrentRemote   string
 	NtlmSession     ntlm.ClientSession
 	envVars         map[string]string
@@ -69,6 +74,8 @@ type Configuration struct {
 
 func New() *Configuration {
 	c := &Configuration{
+		Env: NewEnvFetcher(),
+
 		CurrentRemote: defaultRemote,
 		envVars:       make(map[string]string),
 	}
@@ -84,6 +91,8 @@ func New() *Configuration {
 // NOTE: this method should only be called during testing.
 func NewFromValues(gitconfig map[string]string) *Configuration {
 	config := &Configuration{
+		Env: NewEnvFetcher(),
+
 		gitConfig: make(map[string]string, 0),
 		envVars:   make(map[string]string, 0),
 	}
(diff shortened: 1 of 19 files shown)


technoweenie:
This looks great :metal:  One comment on your note about refactoring the `Fetcher` types in https://github.com/github/git-lfs/commit/e32fabe37fc150b21a5d2d7594a5300c1388c5cb:

You mention the new type being called `config.Environment`. If I understand that correctly, the `config.Configuration` struct would look something like this:

```go
type Configuration struct {
  Env Environment
  Git Environment
  // ...
}
```

I like the idea of `Environment` being the generic name for it. What do you think about the `config.Configuration` property being called `Os`?

```go
type Configuration struct {
  Os Environment
  Git Environment
  // ...
}

cfg.Os.Get("LFS_WHATEVER")
cfg.Git.Get("lfs.whatever")
```

ttaylorr:
I like that idea! Thanks for the feedback and review 😄

config/fetcher.go
```diff
@@ -0,0 +1,14 @@
...
+// Fetcher provides an interface to get typed information out of a configuration
+// "source". These sources could be the OS enviornment, a .gitconfig, or even
+// just a `map`.
+type Fetcher interface {
+	// Get returns the string value assosicated with a given key, or an
+	// empty string if none exists.
+	Get(key string) (val string)
+
+	// Bool returns the boolean value assosicated with a given key, or an
+	// empty string if none exists.
```
sinbad: "or an empty string if none exists" -> "the value passed as 'def' if none exists"

Probably worth mentioning which string values equate to true/false too.

config/map_fetcher.go
```diff
@@ -0,0 +1,28 @@
+package config
+
+import "strconv"
+
+// mapFetcher provides an implementation of the Fetcher interface by wrapping
+// the `map[string]string` type.
+type mapFetcher map[string]string
```
sinbad: It's important to point out that mapFetcher is not safe from concurrent map access and must therefore be protected by other mutexes.

config/config.go
```diff
@@ -78,18 +85,30 @@ func New() *Configuration {
...
+// NewFrom returns a new `*config.Configuration` that reads both its Git
+// and Enviornment-level values from the ones provided instead of the actual
+// `.gitconfig` file or `os.Getenv`, respectively.
 //
-// NOTE: this method should only be called during testing.
-func NewFromValues(gitconfig map[string]string) *Configuration {
+// This method should only be used during testing.
+func NewFrom(v Values) *Configuration {
 	config := &Configuration{
+		Env: mapFetcher(v.Env),
```
sinbad: This has subtle concurrent map issues embedded in it since mapFetcher is not safe. I know it's only for testing but it's likely to result in weird random errors. I think mapFetcher needs to have a mutex like the real EnvFetcher etc.

config/map_fetcher.go
sinbad: Seeing how it's actually used as a drop-in test replacement for other types which are safe in concurrent use I think it needs its own mutex.

sinbad:
Couple of comments on docs/map safety for your next iteration. I like how you've split this up into phases to make it easier to review 👍

config/config.go
ttaylorr: We shouldn't need to synchronize access to this map, since all the data is provided up front in #1416 and all subsequent map operations are read-only. No writers and `n` readers will not result in a concurrent map access error.

config/fetcher.go
ttaylorr: Ah, whoops. This code is actually removed in #1419, but I will apply the typo fix there 😄

config/config.go
sinbad: :+1: yep as discussed in chat, my bad - I didn't spot that it had become immutable.

ttaylorr:

PR #1416, opened 2016-08-04
config-next: Remove config.{Setenv,SetAllEnv}

ttaylorr:
This pull-request removes all instances of `config.Config.Setenv` and `config.config.SetAllEnv`. References to `SetConfig` still exist, but those are `.gitconfig` related, meaning they'll be addressed in a future pull-request.

There is also now considerably more repetition in the `auth/ssh_test.go` file. It's a good candidate for table-style tests, but I also appreciate its clarity right now.

This completes step 2 of config-next:

> Update tests to not use mutable behavior like Set and SetAll.

-----------------

### DIFF
--- a/auth/ssh_test.go
+++ b/auth/ssh_test.go
@@ -9,213 +9,225 @@ import (
 )
 
 func TestSSHGetExeAndArgsSsh(t *testing.T) {
-	cfg := config.New()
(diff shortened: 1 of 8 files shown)


technoweenie:
Whoo, I like having an immutable `config.Configuration`.

ttaylorr:

PR #1419, opened 2016-08-04
config: demote `Fetcher`, introduce `Environment`

ttaylorr:
Previously, to fetch data out of the `*config.Configuration` type, a reference
to a `Fetcher` was used, a-la:

```go
cfg.Env.Get(...)
```

This is quite convenient, however, it forces the LFS client to implement
several methods more than once. Consider the interface:

```go
type Fetcher interface {
        Get(key string) (val string)

        Bool(key string, def bool) (val bool)
        // et. al.
}
```

In order to return typed information from a configuration instance, _each_
`Fetcher` must implement its own `N` methods for `Int`, `Bool`, etc.

To remedy this, the `Environment` type was introduced. It instead _has_ a
`Fetcher`, and defines its own type conversions, like so:

```go
type Environment struct {
        f Fetcher
}

func (e *Environment) Bool(key string, def bool) (val bool) { }
func (e *Environment) Int(key string, def int)   (val int) { }

// et. al.
```

Now, the `config.Configuration` type holds a reference to an `Environment`, and
all type conversion methods are defined only once, saving time, and enforcing
consistency across multiple sources.

------------

/cc @technoweenie @sinbad @rubyist @larsxschneider


config/fetcher_test.go
```diff
@@ -0,0 +1,11 @@
+package config_test
+
+import "github.com/stretchr/testify/mock"
+
+type MockFetcher struct {
+	mock.Mock
+}
+
+func (m *MockFetcher) Get(key string) (val string) {
+	return m.Called(key).String(0)
```
technoweenie: This is a case where a fetcher that reads from a `map[string]string` would make more sense IMO.

```golang
env := EnvironmentOf(NewFetcher(map[string]string{"FOO": "FOO"}))
got := c.GotFn(env, c.Val)

if c.Expected != got {
  t.Errorf("....")
}
```

I just don't see what mocking gets us here.

ttaylorr: Good call. I originally wanted the mock to assert that the `Get` method was definitely being called with the correct arguments, but having a `map[string]string` buys us the same result, and lifts a dependency on the `mock` package of `testify`.

Removed the mocks in: https://github.com/github/git-lfs/pull/1419/commits/38c452b28359b64e8c4cd2600f09ba456bcf6b8f :metal:

sinbad:
👍

technoweenie:
:ship:

ttaylorr:
Thanks for the review! 👋

ttaylorr:

PR #1420, opened 2016-08-05
config-next: .gitconfig parsing

technoweenie:
This pull-request introduces a `Fetcher` implementation for `.gitconfig`-related interactions that occur in the `config.Configuration` type.

Pulling out the `.gitconfig` stuff is a bit trickier than the OS environment, because most of the state that `config.Configuration` keeps track of is created while parsing is going on, and is deeply coupled to this logic (see: [here](https://github.com/github/git-lfs/blob/49bf4d5178fe7fab99cec7f26eaa434eeb152bd9/config/config.go#L518-L594)).

To make this PR small, and to make it possible to do small refactorings in the future, the `ReadConfig` function, which produces the `.gitconfig` Fetcher, also returns all of the other state that the `config.Configuration` type needs to keep track of (see: [here](https://github.com/github/git-lfs/blob/7bcc69bf583aa1dbe010b0bc09b6f7f56b4cb2c5/config/git_fetcher.go#L31)).

Over time, this function should become simpler as we introduce stronger types that keep track of this state, thereby decreasing the length of that function's signature. Once the signature is simple, (i.e., it only returns a `*GitFetcher`, then it will be safe to add tests.)

The other notable hack that this PR introduces is some legacy support around the `ClearConfig` and `ResetConfig` functions (see: [here](https://github.com/github/git-lfs/blob/7bcc69bf583aa1dbe010b0bc09b6f7f56b4cb2c5/config/config.go#L475-L478), and [here](https://github.com/github/git-lfs/blob/7bcc69bf583aa1dbe010b0bc09b6f7f56b4cb2c5/config/config.go#L485-L487)). This was done to keep mutability out of the `GitFetcher` implementation, while still supporting those legacy methods. The next step is to remove those methods, thereby eradicating all mutability from the `config` package.

-


ttaylorr:

ttaylorr:
👍 from @technoweenie, @rubyist and myself

PR #1426, opened 2016-08-09
config-next: Unmarshalling

ttaylorr:
❗ The relevant diff is contained in [`config-next-prune..config-next-unmarshalling`](https://github.com/github/git-lfs/compare/config-next-prune...config-next-unmarshalling).

-

Calling `Unmarshal` on a struct-pointer will inspect each of that struct's
field's tags, and unmarshal configuration data into said field according to
three parameters:

- The environment specified, either `git` or `os`.
- The name of the key given (in the tag).
- The type of the field (either `string`, `int`, or `bool`).

If there is already a non-zero value stored in the field, then that field will
be skipped, allowing constructors to set up default behavior.

-

### DIFF
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -100,9 +100,10 @@ func fetchCommand(cmd *cobra.Command, args []string) {
 	}
 
 	if fetchPruneArg {
-		verify := cfg.FetchPruneConfig().PruneVerifyRemoteAlways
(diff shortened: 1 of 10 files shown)


ttaylorr:
Tracking: https://github.com/github/git-lfs/issues/1425.

ttaylorr:
👍 from @technoweenie, @rubyist and myself.

Issue #1425, opened 2016-08-09
config-next: Tracking Issue


ttaylorr:
This is a tracking issue to keep tabs on the `config-next` refactoring.

The broad goals of this refactoring are to move away from the domain-specific API that package `config` currently exposes, to a data-fetching type, and a series of well-factored types and packages throughout the rest of LFS.

Specifically, this refactoring comes in a number of parts:

- [x] Introduce `Fetcher`, implement `EnvFetcher`, and integrate with `*config.Config`.
- [x] Update tests to not use mutable behavior like `Set` and `SetAll`.
- [x] Refactor `Fetcher` type as described in https://github.com/github/git-lfs/commit/e32fabe37fc150b21a5d2d7594a5300c1388c5cb.
- [x] Implement the `GitFetcher`.
- [ ] Remove mutable `.gitconfig`-related operations
- [ ]  Piece by piece, remove legacy calls to `GetenvBool`, replacing them with `.Env.Bool` (and etc).
- [ ] Piece by piece, remove domain knowledge from the `*config.Config` type, in favor of the methodology described in https://github.com/github/git-lfs/pull/1390#issuecomment-236068550.

ttaylorr:
Acking for internal script

ttaylorr:
Acking for internal script.

ttaylorr:
> Piece by piece, remove domain knowledge from the `*config.Config` type, in favor of the methodology described in https://github.com/github/git-lfs/pull/1390#issuecomment-236068550.

Taking this one off of the tracking issue, since this is now outside the scope of the `config-next` refactoring. As discussed with @technoweenie, a good first step would be to outline all of the places the `config` package is used irresponsibly from `api`.

ttaylorr:
Woo-hoo! 🎉

PR #1427, opened 2016-08-09
config-next: remove `SetConfig` from package `api`, `lfs`, `httputil`

ttaylorr:
This pull request removes mutable `.gitconfig` behavior from the `api`, `lfs`, and `httputil` packages. Most of the work involved here is removing calls to `SetConfig` and `ResetConfig`, and also threading an instance of `*config.Configuration` through the `api` package instead of falling back to the `config.Config` global.

To retain the same behavior outside of the tests, the `api` package did gain a few references to `config.Config`, but this is a temporary measure until we remove that package entirely for v2.0. Alternatively, package `api` could own its own private instance of `*config.Configuration`, but I think that's outside the scope of this PR.

Next (and last!) up is the `transfer` package. This requires a little bit more of a fundamental change than just removing calls. The `transfer` package maintains package-global state that is involved with the `config.Config` singleton, so splitting that up is going to be tricky, but doable.

-

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -66,7 +66,7 @@ func Batch(cfg *config.Configuration, objects []*ObjectResource, operation strin
 		return nil, "", errutil.Error(err)
 	}
 
-	req, err := NewBatchRequest(operation)
+	req, err := NewBatchRequest(cfg, operation)
 	if err != nil {
 		return nil, "", errutil.Error(err)
 	}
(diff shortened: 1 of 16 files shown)


ttaylorr:
Tracking: #1425.

technoweenie:
Yay, more baby steps :ship:

PR #1429, opened 2016-08-09
config-next: use `Unmarshal` to fill out `FetchPruneConfig`

ttaylorr:
❗ The relevant diff is contained in: [`config-next-load-unmarshal..config-next-fetch-unmarshal`](https://github.com/github/git-lfs/compare/config-next-load-unmarshal...config-next-fetch-unmarshal).

-

This pull request uses the new `Unmarshal(v interface{}) error` function to fill out the result of calling `FetchPruneConfig`.

It adds the relevant `git:"lfs.foo"` tags to the `FetchPruneConfig` type definition, and sets the defaults during the struct initialization. Currently, it `panic`s on any errors being returned from `Unmarshal`, but those errors are all internal, and the presence of one would indicate a state wherein LFS is not able to function properly (thus they are panic-able).

-

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -28,21 +28,21 @@ var (
 type FetchPruneConfig struct {
 	// The number of days prior to current date for which (local) refs other than HEAD
 	// will be fetched with --recent (default 7, 0 = only fetch HEAD)
-	FetchRecentRefsDays int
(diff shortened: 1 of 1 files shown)


ttaylorr:
Tracking: #1425.

technoweenie:
:ship:

PR #1430, opened 2016-08-09
Transfer manifest

technoweenie:
This extracts the configured transfer adapters to an explicit `Manifest` type, instead of storing them on the go package itself. We get a few benefits here:

* Tests operate on immutable `*config.Configuration` and `*transfer.Manifest` objects. No resetting global state.
* The transfer configs are only loaded as needed.

I made sure that a manifest is loaded at most once per `git-lfs` command.

### DIFF
--- a/lfs/lfs.go
+++ b/lfs/lfs.go
@@ -68,11 +68,12 @@ func ObjectExistsOfSize(oid string, size int64) bool {
 }
 
 func Environ() []string {
+	manifest := transfer.ConfigureManifest(transfer.NewManifest(), config.Config)
 	osEnviron := os.Environ()
 	env := make([]string, 0, len(osEnviron)+7)
-	dltransfers := transfer.GetDownloadAdapterNames()
+	dltransfers := manifest.GetDownloadAdapterNames()
 	sort.Strings(dltransfers)
-	ultransfers := transfer.GetUploadAdapterNames()
+	ultransfers := manifest.GetUploadAdapterNames()
 	sort.Strings(ultransfers)
 
 	env = append(env,
--- a/lfs/transfer_queue.go
+++ b/lfs/transfer_queue.go
@@ -53,6 +53,7 @@ type TransferQueue struct {
 	retrywait         sync.WaitGroup
 	wait              sync.WaitGroup // Incremented on Add(), decremented on transfer complete or skip
 	oldApiWorkers     int            // Number of non-batch API workers to spawn (deprecated)
+	manifest          *transfer.Manifest
 }
 
 // newTransferQueue builds a TransferQueue, direction and underlying mechanism determined by adapter
@@ -67,6 +68,7 @@ func newTransferQueue(files int, size int64, dryRun bool, dir transfer.Direction
 		oldApiWorkers: config.Config.ConcurrentTransfers(),
 		transferables: make(map[string]Transferable),
 		trMutex:       &sync.Mutex{},
+		manifest:      transfer.ConfigureManifest(transfer.NewManifest(), config.Config),
 	}
 
 	q.errorwait.Add(1)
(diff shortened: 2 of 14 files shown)


transfer/transfer.go
```diff
@@ -24,12 +17,6 @@ const (
 // name and dir are to provide context if one func implements many instances
 type NewTransferAdapterFunc func(name string, dir Direction) TransferAdapter
```
sinbad: Just need to update the comments in this area to point people at the new `Manifest` type to find registration functions etc

transfer/manifest.go
```diff
@@ -0,0 +1,128 @@
...
+func (m *Manifest) GetDownloadAdapterNames() []string {
+	return m.getAdapterNames(m.downloadAdapterFuncs)
+}
+
+// GetUploadAdapterNames returns a list of the names of upload adapters available to be created
+func (m *Manifest) GetUploadAdapterNames() []string {
+	return m.getAdapterNames(m.uploadAdapterFuncs)
+}
+
+// GetUploadAdapterNames returns a list of the names of upload adapters available to be created
```
sinbad: `getAdapterNames`

transfer/custom_test.go
```diff
@@ -71,28 +32,26 @@ func TestCustomTransferBasicConfig(t *testing.T) {
...
 }
 
-func TestCustomTransferDownloadConfig(t *testing.T) {
-	saveTransferSetupState()
-	defer func() {
-		config.Config.ResetConfig()
-		restoreTransferSetupState()
-	}()
-
+func xTestCustomTransferDownloadConfig(t *testing.T) {
```
sinbad: xTest - typo? Probably means this test isn't running

transfer/custom_test.go
```diff
@@ -101,28 +60,26 @@ func TestCustomTransferDownloadConfig(t *testing.T) {
...
 }
 
-func TestCustomTransferUploadConfig(t *testing.T) {
-	saveTransferSetupState()
-	defer func() {
-		config.Config.ResetConfig()
-		restoreTransferSetupState()
-	}()
-
+func xTestCustomTransferUploadConfig(t *testing.T) {
```
sinbad: Ditto - maybe not accidental then but I don't understand

lfs/lfs.go
```diff
@@ -68,11 +68,12 @@ func ObjectExistsOfSize(oid string, size int64) bool {
 }
 
 func Environ() []string {
+	manifest := transfer.ConfigureManifest(transfer.NewManifest(), config.Config)
```
sinbad: I think this method should take at least a `Config` and maybe a `Manifest` as well, otherwise it could well be printing something entirely different to what is being used in practice.

lfs/transfer_queue.go
```diff
@@ -67,6 +68,7 @@ func newTransferQueue(files int, size int64, dryRun bool, dir transfer.Direction
 		oldApiWorkers: config.Config.ConcurrentTransfers(),
 		transferables: make(map[string]Transferable),
 		trMutex:       &sync.Mutex{},
+		manifest:      transfer.ConfigureManifest(transfer.NewManifest(), config.Config),
```
sinbad: Shouldn't this be passed in rather than re-created?

sinbad:
👍 Great work. A couple of places where the manifest is re-created when I think it should be passed in if that's the principle we're going with; otherwise the manifest that is created in the command may be different to the one being used in for example the transfer queue. Doesn't affect anything now because manifests are only created as default from config so they match but could cause issues in future.

technoweenie:
Thanks for pointing out my sloppy test and doc typos. I was having a hell of a time getting those custom tests to pass, and just forgot to re-enable them.

I'm going to leave the `lfs.Environ()` and `TransferQueue` changes to future refactoring PRs that will remove a reliance on `config.Config`.

ttaylorr:
👍  indeed!

PR #1434, opened 2016-08-10
Command var scope

technoweenie:
I thought changing `lfs.Environ()` to accept a `*config.Configuration` and `*transfer.Manifest` (instead of using global values) would be easy.  However, I wanted a spot for various "global" config objects to be created for all commands, so I picked `commands.Run()`. I may look for a way to pass some kind of context to each command and remove those too.

I built this PR off #1430, so view this diff instead: [transfer-manifest...command-var-scope](https://github.com/github/git-lfs/compare/transfer-manifest...command-var-scope)

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -165,6 +165,8 @@ func checkoutWithChan(in <-chan *lfs.WrappedPointer) {
 
 	// As files come in, write them to the wd and update the index
 
+	manifest := TransferManifest()
+
 	for pointer := range in {
 
 		// Check the content - either missing or still this pointer (not exist is ok)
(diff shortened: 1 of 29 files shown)


ttaylorr:
Looks good, I like this isolation. Curious why we need to give a `func() *cobra.Command` as an argument to `RegisterSubcommand`. It would be cool to only have to create commands as needed, but it looks like they all get created anyway.

technoweenie:
Returning a function lets us initialize the cobra commands whenever we want (as in, after all the config objects are initialized). Also, the lock commands currently return nil if they're not enabled. Obviously that won't always be the case, but I think we can keep `isCommandEnabled()` around for other future commands.

ttaylorr:
Ah, delaying until we have the config loaded is pretty sweet. Works for me 👍

ttaylorr:
Pending a succesful build: 🚢

PR #1438, opened 2016-08-11
Update ROADMAP.md

technoweenie:
Triaging old issues, and adding to the roadmap...

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -28,6 +28,7 @@ Git LFS. If you have an idea for a new feature, open an issue for discussion.
 | | Investigate hanging pushes/pulls when git credential helper is not set | [#197](https://github.com/github/git-lfs/issues/197) |
 | | Support git ssh shorthands | [#278](https://github.com/github/git-lfs/issues/278) |
 | | Support `GIT_CONFIG` | [#318](https://github.com/github/git-lfs/issues/318) |
+| | LFS Migration tool | [#326](https://github.com/github/git-lfs/issues/326) |
 
 ## Upcoming Features
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -29,7 +29,7 @@ Git LFS. If you have an idea for a new feature, open an issue for discussion.
 | | Support git ssh shorthands | [#278](https://github.com/github/git-lfs/issues/278) |
 | | Support `GIT_CONFIG` | [#318](https://github.com/github/git-lfs/issues/318) |
 | | LFS Migration tool | [#326](https://github.com/github/git-lfs/issues/326) |
-| | Warn when Git version is unsupported | [#410] |
+| | Warn when Git version is unsupported | [#410](https://github.com/github/git-lfs/issues/410) |
 
 ## Upcoming Features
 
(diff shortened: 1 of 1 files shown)


ttaylorr:
:metal: Want to keep this open until we go through all of the open issues? Should only be a few days.

technoweenie:
I'm not sure I'm ready to commit to the LFS migration tool on the roadmap like that. Perhaps under "Possible Features" instead?

ttaylorr:
I agree, not sure why I put it there in the first place. Moved it down to the "Possible Features" section!

ROADMAP.md
```diff
@@ -25,23 +25,43 @@ Git LFS. If you have an idea for a new feature, open an issue for discussion.
...
+| | Support tracking files by size | [#282](https://github.com/github/git-lfs/issues/282)
 | | Binary diffing - reduce the amount of content sent over the wire. | |
 | | Client side metrics reporting, so the Git LFS server can optionally track how clients are performing. | |
 | | Pure SSH: full API & transfer support for SSH without redirect to HTTP | |
+| | Compression of files in `.git/lfs/objects` | [#260](https://github.com/github/git-lfs/issues/260) |
+| | LFS Migration tool | [#326](https://github.com/github/git-lfs/issues/326) |
+| | Automatic upgrades | [#531](https://github.com/gihtub/git-lfs/issues/531) |
+| | Investigate `git add` hash caching | [#574](https://github.com/github/git-lfs/issues/574) |
+| | `git lfs archive` command | [#1322](https://github.com/github/git-lfs/issues/1322) |
+| | Support 507 http responses | [#1327](https://github.com/github/git-lfs/issues/1327) |
```
dpursehouse: I have a couple of questions about HTTP 507 support.

- Does adding support involve anything more than updating the list of optional status codes to mention it?  If not, I can send a PR to make that update.

- Is it going to mean a new API version?

Reason I ask about the API version is that I'd like to [include support for this in JGit](https://git.eclipse.org/r/#/c/79315/) (which currently implements Git LFS v1 Batch API) but if it's in a new API version (likely also including other stuff) it might not be as easy.

technoweenie:
> Does adding support involve anything more than updating the list of optional status codes to mention it? If not, I can send a PR to make that update.

Nope, just looking for a more specific error message. If you think 507 or any other status should do something special, we can talk about that too.

> Is it going to mean a new API version?

Since it's an addition, I'm not considering it a new API version. An old client that only understands v1 API stuff should always be able to talk to any v1.x API.

Also, I'm not happy with the API docs specifying v1, v1.2, and v1.3 API versions. I want to rewrite it as a single document, with specific properties calling out what Git LFS client supports them. If you're adding this to JGit, then I'll probably ping you for some feedback :)

dpursehouse:
> just looking for a more specific error message

You mean like in this block?

```
defaultErrors = map[int]string{
        400: "Client error: %s",
        401: "Authorization error: %s\nCheck that you have proper access to the repository",
        403: "Authorization error: %s\nCheck that you have proper access to the repository",
        404: "Repository or object not found: %s\nCheck that it exists and that you have proper access to it",
         500: "Server error: %s",
}
```

There aren't any special messages there for example for 422 or the other optional status codes (406, 429, etc).  Also as far as I have tested, the client will use the message that was given by the server anyway, so for example if the server returns 404 with a body:

```
{"message" : "repository foo not found"}
```

then it will display that message, rather than the message in the `defaultErrors` block.

> If you think 507 or any other status should do something special, we can talk about that too.

Right now I don't need it to do anything special.  Displaying the error message is enough, like it does for the others.

technoweenie:
Yes, that block. I realize it's not exactly a major change. It's also pretty far down the backlog :)

Issue #1439, opened 2016-08-12
discrepency between custom transfer docs and implementation


ericfrederich:
Not sure where it should be fixed.

I'm seeing my client send a message like the following with an `"Event"` key instead of an `"event"` key like the docs suggest [here](https://github.com/github/git-lfs/blob/master/docs/custom-transfers.md#stage-1-intiation)

```json
{"Event":"init","operation":"upload","concurrent":false,"concurrenttransfers":3}
```

If I had my choice I'd probably want the code fixed to send a lower case key rather than fixing the docs.

ttaylorr:
Hi @ericfrederich -- thanks for opening this.

I definitely agree that the code should send a lowercase `"event"` for consistency with other keys in the custom transfer API.

I'll go ahead and make that change now 😄

technoweenie:
Thanks for the report @ericfrederich! We just shipped a small tweak to the code.

PR #1441, opened 2016-08-12
transfer/custom: encode "event" as lowercase

ttaylorr:
As pointed out by @ericfrederich in #1439, the `Event` field of the custom transfer "init request" is encoded as `"Event":`, not `"event":` as it should be.

Running the custom transfer integration tests confirm that this is OK to do 👍 

I'm also thinking that per @technoweenie's suggestion, we backport this to our `release-1.3` branch and get this out as a part of the `v1.3.2` release.

-

/cc @technoweenie @sinbad @ericfrederich


ttaylorr:
Decided to target this in for the `1.4.0` release instead. 😄

sinbad:
Sorry I missed this earlier, yeah it's fine - typo on my part 😄

PR #1443, opened 2016-08-12
add support for url.*.insteadof in git config

technoweenie:
Rewrite of #1117 after the config refactoring in #1425. There's no `readGitInsteadOfConfig()` anymore to hook into anymore.

We can't easily support `pushinsteadof` without bigger changes. `config.NewEndpointWithConfig()` doesn't know if the request is a push or pull.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -73,7 +73,7 @@ type Configuration struct {
 	extensions     map[string]Extension
 	manualEndpoint *Endpoint
 	parsedNetrc    netrcfinder
-	urlAliases     map[string]string
+	urlAliasesMap  map[string]string
 	urlAliasMu     sync.Mutex
 }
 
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -449,6 +452,49 @@ func (c *Configuration) AllGitConfig() map[string]string {
...
+	aliases := c.urlAliases()
+	for alias, _ := range aliases {
+		if !strings.HasPrefix(rawurl, alias) {
+			continue
+		}
+
+		if longestalias < alias {
+			longestalias = alias
+		} else if longestalias == alias {
+			fmt.Fprintf(os.Stderr, "WARNING: Multiple 'url.*.insteadof' keys with the same alias: %q", alias)
```
ttaylorr: Should we have an integration test for this?

technoweenie: Why yes, we should, because it caught a bug :)

ttaylorr:
Looks good, let's 🚢

PR #1436, opened 2016-08-10
config-next: remove legacy methods

ttaylorr:
🚧  This pull-request removes legacy methods from the `*config.Configuration` object such as:

|       | Was | Becomes |
| --- | :---- | :---------- |
| ✅   | `*config.Configuration.GetEnv` | `cfg.Os.Get` |
| ✅  | `*config.Configuration.GetEnvBool` | `cfg.Os.Bool` |
| ⬜  | `*config.Configuration.GitConfig` | `cfg.Git.Get` |
| ⬜  | `*config.Configuration.GitConfigInt` | `cfg.Git.Int` |
| ⬜  | `*config.Configuration.GitConfigBool` | `cfg.Git.Bool` |

-

PR #1450, opened 2016-08-15
config-next: Wrapped `gitEnvironment`

ttaylorr:
This pull-request introduces a wrapped `gitEnvironment` to support the legacy behavior found in `loadGitConfig` in order to make #1436 possible.

-

In order to support lazily loading the values stored in a user's `.gitconfig`,
we must wait until calling `*config.Configuration.loadGitConfig()` until it is
_absolutely necessary_.

To accomplish this, it was proposed that we introduce a wrapped variant of the
`*Environment` type, only for interacting with the `GitFetcher` that was
capable of supporting such beahvior.

As such, a new implementation of the `Environment` type must be defined. Since
previously there only existed the concrete type `*Environment`, this commit
demotes that down to `*enviornment`, and introduces the interface
`Environment`, which it implements.

-

To wrap the old behavior found in `*config.Configuration.loadGitConfig`, a
`*gitEnvironment` implementation was introduced to wrap the behavior of another
Environment wholesale while at the same time prepending a call to
`loadGitConfig()`.

It should be noted that in order to preserve legacy behavior with using certain
member variables in `*config.Configuration`, there is a circular dependency
between the two types.

-

/cc @technoweenie

### DIFF
--- a/api/lock_api.go
+++ b/api/lock_api.go
@@ -149,8 +149,8 @@ type Committer struct {
 // "user.name" and "user.email" configuration values are used from the
 // config.Config singleton.
 func CurrentCommitter() Committer {
-	name, _ := config.Config.GitConfig("user.name")
-	email, _ := config.Config.GitConfig("user.email")
+	name, _ := config.Config.Git.Get("user.name")
+	email, _ := config.Config.Git.Get("user.email")
 
 	return Committer{name, email}
 }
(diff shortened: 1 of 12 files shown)


ttaylorr:
Tracking #1425.

technoweenie:
Looks good. I did a few find/replace ops and removed the remaining `.GitConfig*` functions :metal:

technoweenie:
Fixed a merge conflict from merging #1436: 691bbbe9c070839793422e98ea8de505d0707774

PR #1451, opened 2016-08-15
Fetch remote urls

technoweenie:
This adds support for pushing and pulling LFS objects with a raw remote url. This was originally submitting in #1085, and updated for the current master branch. This also adds tests.

### DIFF
--- a/config/endpoint.go
+++ b/config/endpoint.go
@@ -37,12 +37,17 @@ func NewEndpointFromCloneURLWithConfig(url string, c *Configuration) Endpoint {
 		return e
 	}
 
+	if strings.HasSuffix(url, "/") {
+		e.Url = url[0 : len(url)-1]
+	}
+
 	// When using main remote URL for HTTP, append info/lfs
-	if path.Ext(url) == ".git" {
+	if path.Ext(e.Url) == ".git" {
 		e.Url += "/info/lfs"
 	} else {
 		e.Url += ".git/info/lfs"
 	}
+
 	return e
 }
 
--- a/test/test-fetch.sh
+++ b/test/test-fetch.sh
@@ -503,6 +503,29 @@ begin_test "fetch --prune"
 )
 end_test
 
+begin_test "fetch raw remote url"
+(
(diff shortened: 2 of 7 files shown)


config/endpoint.go
```diff
@@ -37,12 +37,17 @@ func NewEndpointFromCloneURLWithConfig(url string, c *Configuration) Endpoint {
 		return e
 	}
 
+	if strings.HasSuffix(url, "/") {
```
ttaylorr: Since we already have `path` imported, would it be possible to use `path.Clean` here?

technoweenie: We could, but it does a lot more stuff besides just chomping the trailing `/` that we don't need. These urls either come from the git remote config (`remote.origin.url`), or from custom lfs config (`lfs.url`, `remote.origin.lfsurl`).

test/test-fetch.sh
```diff
@@ -503,6 +503,29 @@ begin_test "fetch --prune"
...
+  cd raw
+  git init
+  git lfs install --local --skip-smudge
+
+  git remote add origin $GITSERVER/test-fetch
+  git pull origin master
+
+  # LFS object not downloaded, pointer in working directory
+  refute_local_object "$contents_oid"
+  ack "$content_oid" a.dat
```
ttaylorr: I don't think the containers that Travis gives us have `ack` installed.

test/test-pull.sh
```diff
@@ -111,6 +111,34 @@ begin_test "pull"
...
+
+  git remote add origin $GITSERVER/test-pull
+  git pull origin master
+
+  contents="a"
+  contents_oid=$(calc_oid "$contents")
+
+  # LFS object not downloaded, pointer in working directory
+  refute_local_object "$contents_oid"
+  ack "$contents_oid" a.dat
```
ttaylorr: Same here.,

ttaylorr:
After you switch out `ack` for `grep`, I think this is ready to merge. 😄

PR #1452, opened 2016-08-15
add object Authenticated property

technoweenie:
This lets servers set an `Authenticated` property on objects, that the client uses to determine if it needs to call `git-credential` or not. This is a forward compatible API change, and won't affect existing implementations.

When a client hits the batch api, it sends a payload like this:

```js
{
  "operation": "download",
  "objects": [{"oid": "SOMEOID", "size": 123}]
}
```

A typical server response returns a URL to transfer the object with any values for the HTTP request header.

```js
{"objects": [{
  "oid": "SOMEOID",
  "size": 123,
  "actions": {
    "download": {
      "href": "https://download-this-object",
      "header": {
        "Authorization": "Basic ..."
      }
    }
  }
}
```

Git LFS will use `git-credential` to find the user and pass for that download unless the `Authorization` header is set. What if the service uses a non-standard header, or what if the file requires no authentication? This is where the `authenticated` property comes in:

```js
{"objects": [{
  "oid": "SOMEOID",
  "size": 123,
  "authenticated": false,
  "actions": {
    "download": {
      "href": "https://download-this-object",
      "header": {
        "Custom-Token": "Basic ..."
      }
    }
  }
}
```

### DIFF
--- a/api/object.go
+++ b/api/object.go
@@ -19,11 +19,12 @@ func (e *ObjectError) Error() string {
 }
 
 type ObjectResource struct {
-	Oid     string                   `json:"oid,omitempty"`
-	Size    int64                    `json:"size"`
-	Actions map[string]*LinkRelation `json:"actions,omitempty"`
-	Links   map[string]*LinkRelation `json:"_links,omitempty"`
-	Error   *ObjectError             `json:"error,omitempty"`
+	Oid           string                   `json:"oid,omitempty"`
+	Size          int64                    `json:"size"`
+	Authenticated bool                     `json:"authenticated,omitempty"`
+	Actions       map[string]*LinkRelation `json:"actions,omitempty"`
+	Links         map[string]*LinkRelation `json:"_links,omitempty"`
+	Error         *ObjectError             `json:"error,omitempty"`
 }
 
 // TODO LEGACY API: remove when legacy API removed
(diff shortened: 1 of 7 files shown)


api/object.go
```diff
@@ -19,11 +19,12 @@ func (e *ObjectError) Error() string {
...
 
 type ObjectResource struct {
-	Oid     string                   `json:"oid,omitempty"`
-	Size    int64                    `json:"size"`
-	Actions map[string]*LinkRelation `json:"actions,omitempty"`
-	Links   map[string]*LinkRelation `json:"_links,omitempty"`
-	Error   *ObjectError             `json:"error,omitempty"`
+	Oid           string                   `json:"oid,omitempty"`
+	Size          int64                    `json:"size"`
+	Authenticated bool                     `json:"authenticated,omitempty"`
```
ttaylorr: Is `omitempty` behavior that we want? In the case of `bool`s, we would only require having the `authenticated` property if the value were true.

ttaylorr:
Pending my one comment, looks great 👍

api/object.go
technoweenie: That's fine, since the default is false. Apparently `omitempty` is good enough for `oid` too :trollface:

ttaylorr:
🚢

PR #1445, opened 2016-08-13
Fixes #1356: Don't show error message twice

dpursehouse:
Change-Id: Iff591518853b1c5c229b676d08d8f914f1eb6f62
Signed-off-by: David Pursehouse <dpursehouse@collab.net>

PR #1453, opened 2016-08-16
fix duplicate error reporting

technoweenie:
This is an extension to #1445 which also cleans up some of the duplicate logic from `commands.ExitWithError()`.  /cc @dpursehouse


ttaylorr:
Looks great 👍

PR #1454, opened 2016-08-16
retry if file download failed

larsxschneider:
A file download can fail for multiple reasons, e.g. due to various
network errors. Use the `TransferQueue` retry mechansim to try it again.

Possible errors due to spotty network connections:
https://github.com/golang/go/blob/64214792e214bbacb8c00ffea92a7131e30fa59e/src/io/io.go#L26-L47

### DIFF
--- a/tools/iotools.go
+++ b/tools/iotools.go
@@ -6,6 +6,7 @@ import (
 	"hash"
 	"io"
 
+	"github.com/github/git-lfs/errutil"
 	"github.com/github/git-lfs/progress"
 )
 
@@ -84,4 +85,23 @@ func (r *HashingReader) Read(b []byte) (int, error) {
 	}
 
 	return w, err
+}
+
+// RetriableReader wraps a error response of reader as RetriableError()
+type RetriableReader struct {
+	reader io.Reader
+}
+
+func NewRetriableReader(r io.Reader) io.Reader {
+	return &RetriableReader{r}
+}
+
+func (r *RetriableReader) Read(b []byte) (int, error) {
+	n, err := r.reader.Read(b)
+	// EOF is a successful response as it is used to signal a graceful end of input
+	// c.f. https://git.io/v6riQ
+	if err == nil || err == io.EOF {
+		return n, err
+	}
+	return n, errutil.NewRetriableError(err)
 }
--- a/tools/iotools.go
+++ b/tools/iotools.go
@@ -6,7 +6,7 @@ import (
 	"hash"
 	"io"
 
-	"github.com/github/git-lfs/errutil"
+	"github.com/github/git-lfs/errors"
 	"github.com/github/git-lfs/progress"
 )
 
@@ -103,5 +103,5 @@ func (r *RetriableReader) Read(b []byte) (int, error) {
 	if err == nil || err == io.EOF {
 		return n, err
 	}
-	return n, errutil.NewRetriableError(err)
+	return n, errors.NewRetriableError(err)
 }
--- a/transfer/basic_download.go
+++ b/transfer/basic_download.go
@@ -204,11 +204,7 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb TransferProgressCallback
 	}
 	written, err := tools.CopyWithCallback(dlFile, hasher, res.ContentLength, ccb)
 	if err != nil {
-		wrappedErr := fmt.Errorf("cannot write data to tempfile %q: %v", dlfilename, err)
-		if errutil.IsRetriableError(err) {
-			return errutil.NewRetriableError(wrappedErr)
-		}
-		return wrappedErr
+		return errors.Wrapf(err, "cannot write data to tempfile %q", dlfilename)
 	}
 	if err := dlFile.Close(); err != nil {
 		return fmt.Errorf("can't close tempfile %q: %v", dlfilename, err)
--- a/transfer/basic_download.go
+++ b/transfer/basic_download.go
@@ -177,11 +177,13 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb TransferProgressCallback
 	}
 
 	var hasher *tools.HashingReader
+	httpReader := tools.NewRetriableReader(res.Body)
+
 	if fromByte > 0 && hash != nil {
 		// pre-load hashing reader with previous content
-		hasher = tools.NewHashingReaderPreloadHash(res.Body, hash)
+		hasher = tools.NewHashingReaderPreloadHash(httpReader, hash)
 	} else {
-		hasher = tools.NewHashingReader(res.Body)
+		hasher = tools.NewHashingReader(httpReader)
 	}
 
 	if dlFile == nil {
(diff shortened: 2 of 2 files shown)


larsxschneider:
This should address https://github.com/github/git-lfs/issues/1433
But I need to perform a large number of clones to trigger the bad behavior. Therefore I don't know yet if it really solves the issue. But I think it should help 😄 
Would that be an acceptable solution for you or is it too broad?

technoweenie:
Niiiice. I'd rather distinguish between network and local disk errors though. Maybe they're so infrequent, that it makes sense to retry everything?

ttaylorr:
I think distinguishing between the two would definitely be a nice to have 👍

larsxschneider:
Alright. How would you do it? Would it be OK for you if we wrap the repeatable error here:
https://github.com/github/git-lfs/blob/32b91777a2f57534738b7962d770b4d198cab647/tools/iotools.go#L51

This would mean all errors happening here actually wrapped:
https://github.com/golang/go/blob/64214792e214bbacb8c00ffea92a7131e30fa59e/src/io/io.go#L376-L414

I haven't dug deeper but I think this will emit more then network errors?!

technoweenie:
The body is read from the request through [this hashing reader](https://github.com/github/git-lfs/blob/62473bdd544a391613d6ff1636f41ff32f8a2cc0/transfer/basic_download.go#L181-L186). Maybe we can wrap that `res.Body` argument with a `Reader` implementation that wraps all errors as retriable errors?

```diff
diff --git a/tools/iotools.go b/tools/iotools.go
index 1b99f07..8fdbc74 100644
--- a/tools/iotools.go
+++ b/tools/iotools.go
@@ -7,6 +7,7 @@ import (
        "io"

        "github.com/github/git-lfs/progress"
+       "github.com/github/git-lfs/errutil"
 )

 type readSeekCloserWrapper struct {
@@ -85,3 +86,19 @@ func (r *HashingReader) Read(b []byte) (int, error) {

        return w, err
 }
+
+type retriableReader struct {
+       reader io.Reader
+}
+
+func NewRetriableReader(r io.Reader) io.Reader {
+       return &retriableReader{r}
+}
+
+func (r *retriableReader) Read(b []byte) (int, error) {
+       n, err := r.reader.Read(b)
+       if err != nil {
+               return n, errutil.NewRetriableError(err)
+       }
+       return n, nil
+}
diff --git a/transfer/basic_download.go b/transfer/basic_download.go
index 8c83b1b..e523ccc 100644
--- a/transfer/basic_download.go
+++ b/transfer/basic_download.go
@@ -178,11 +178,12 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb TransferProgressCallback
        }

        var hasher *tools.HashingReader
+       httpReader := tools.NewRetriableReader(res.Body)
        if fromByte > 0 && hash != nil {
                // pre-load hashing reader with previous content
-               hasher = tools.NewHashingReaderPreloadHash(res.Body, hash)
+               hasher = tools.NewHashingReaderPreloadHash(httpReader, hash)
        } else {
-               hasher = tools.NewHashingReader(res.Body)
+               hasher = tools.NewHashingReader(httpReader)
        }

        if dlFile == nil {
```

larsxschneider:
@technoweenie Elegant solution! I fixed the PR based on your idea. I am still testing the approach locally here...

technoweenie:
@larsxschneider Cool, let me know if it helps. I want to do a release this week, and would love to get this in :)

larsxschneider:
It is very hard for me to reproduce the error because:
* only happens on Windows
* only with our company VPN
* only on repos with a large number of LFS files
* only sporadically

I really would like to get this fix in, but I also want to be sure we really fix the issue 😄 
Apparently one of our engineers saw the exact problem *with* this fix... still investigating. I just build a special version with my name in the debug output to ensure the correct git-lfs.exe binary is in use 😉

larsxschneider:
@technoweenie Please review. 

The fix seems to substantially improve the situation for us. However, if we stress test it (multiple clones over a bad VPN connection) then we still see failures. I assume the reason is that we only run a single retry right now.

transfer/basic_download.go
```diff
@@ -203,7 +205,11 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb TransferProgressCallback
 	}
 	written, err := tools.CopyWithCallback(dlFile, hasher, res.ContentLength, ccb)
 	if err != nil {
-		return fmt.Errorf("cannot write data to tempfile %q: %v", dlfilename, err)
+		wrappedErr := fmt.Errorf("cannot write data to tempfile %q: %v", dlfilename, err)
+		if errutil.IsRetriableError(err) {
+			return errutil.NewRetriableError(wrappedErr)
+		}
+		return wrappedErr
```
larsxschneider: This looks cluttered. @technoweenie @ttaylorr @sinbad do you see a nice golang-ish solution?

ttaylorr: 👋  thanks for the ping. Handling of errors has changed significantly since v1.3.1, most of the work change are contained in https://github.com/github/git-lfs/pull/1463 and https://github.com/github/git-lfs/pull/1466. The gist of it is that we're now using Dave Cheney's `errors` package, which is described [here](https://godoc.org/github.com/pkg/errors).

How does this one feel?

```go
written, err := tools.CopyWithCallback(...)
if err != nil {
        err = errors.Wrapf(err, "cannot write data to tempfile %q", dlfilename)

        return errors.NewRetriableError(err)
}

// ...
```

technoweenie: I like that... as long as `NewRetriableError` won't rewrap an existing retriable error :)

larsxschneider: @ttaylorr Is your suggestion equivalent to my original code? (re: @technoweenie 's comment) ?

ttaylorr: Right now you'll have to check to make sure that the error isn't already retriable, like so:

```go
written, err := tools.CopyWithCallback(...)
if err != nil {
        err = errors.Wrapf(err, "cannot write data to tempfile %q", dlfilename)

        if errutil.IsRetriableError(err) {
                return err
        }

        return errors.NewRetriableError(err)
}

// ...
```

larsxschneider: After looking carefully at it I think your initial suggestion was correct. Please double check my error fix here: https://github.com/github/git-lfs/pull/1454/commits/a9418052bca0951916177f433ef1f5270380aff2

Other than that the PR is ready from my point of view 😄

tools/iotools.go
```diff
@@ -85,3 +86,22 @@ func (r *HashingReader) Read(b []byte) (int, error) {
...
+}
+
+func (r *RetriableReader) Read(b []byte) (int, error) {
+	n, err := r.reader.Read(b)
+	// EOF is a successful response as it is used to signal a graceful end of input
+	// c.f. https://git.io/v6riQ
+	if err == nil || err == io.EOF {
+		return n, err
+	}
+	return n, errors.NewRetriableError(err)
```
ttaylorr: Technically this could re-wrap an already retriable error in the case that `r`'s `reader` has another `RetriableReader` underneath, but I don't think it's a huge deal 😄

ttaylorr:
@larsxschneider This looks good. Eventually I'd like to look at being more flexible at the `transfer` layer by supporting things like multiple retries per OID, backoffs and other resiliency measures like that.

PR #1455, opened 2016-08-16
Update fetch include/exclude docs for pattern matching

ralfthewise:
Previously the docs indicated that pattern matching was per gitignore, but actually it is done per golang's `filepath.Match()`.  Ideally it will eventually mirror gitignore, but until then the docs should be corrected.

This will help with user frustration regarding https://github.com/github/git-lfs/issues/912


technoweenie:
Ah, good catch!

PR #1456, opened 2016-08-16
include ref in upload request

technoweenie:
Updates #969 against the current master. This was not a straight forward update because the code base has gone through a pretty major refactoring since then.
## To-do:
- [ ] Ensure we're always sending full server side refs: refs/tags/v1.0, refs/heads/master. Never client ref names.
- [ ] Integration tests that test git lfs push, git lfs fetch on repos where the local and remote branches don't match.


ttaylorr:
Are the checkboxes to-do items?

technoweenie:
Yup. Add to it if you think of anything else.

ttaylorr:
Tentatively milestone-ing this into v2.2.0 per https://github.com/git-lfs/git-lfs/issues/2190#issuecomment-298380482.

technoweenie:
Superseded by #2706

PR #1459, opened 2016-08-17
Vendor errs

ttaylorr:
This PR vendors the `github.com/pkg/errors` package for use in the upcoming `error` package refactoring :metal: 

-

/cc @technoweenie


glide.lock
```diff
@@ -25,16 +27,27 @@ imports:
   version: 580b9be06c33d8ba9dcc8757ea56b7642472c2f5
 - name: github.com/stretchr/testify
   version: 6cb3b85ef5a0efef77caef88363ec4d4b5c0976d
+  subpackages:
```
ttaylorr: Some of these misc changes like here and L38 are related to the latest update of `glide`.

PR #1461, opened 2016-08-17
check the git version is ok in some key commands

technoweenie:
/cc #1212

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -336,9 +336,13 @@ func isCommandEnabled(cfg *config.Configuration, cmd string) bool {
 
 func requireGitVersion() {
 	minimumGit := "1.8.2"
-	gitver, _ := git.Config.Version()
+	gitver, err := git.Config.Version()
+	if err != nil {
+		Exit("Error getting git version: %s", err)
+	}
+
 	if !git.Config.IsGitVersionAtLeast(minimumGit) {
-		Exit(fmt.Sprintf("git version >= %s is required for Git LFS, your version: %s", minimumGit, gitver))
+		Exit("git version >= %s is required for Git LFS, your version: %s", minimumGit, gitver)
 	}
 }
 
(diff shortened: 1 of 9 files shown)


commands/commands.go
```diff
@@ -334,6 +334,18 @@ func isCommandEnabled(cfg *config.Configuration, cmd string) bool {
 	return cfg.Os.Bool(fmt.Sprintf("GITLFS%sENABLED", strings.ToUpper(cmd)), false)
 }
 
+func requireGitVersion() {
+	minimumGit := "1.8.2"
```
ttaylorr: I wonder if it would be worth elevating this to a method argument, and having a const in the `commands` package.

This would allow us to easily require different versions of Git for different commands, but I'm not sure that's behavior we need.

ttaylorr:
Aside from my one comment, 👍

commands/commands.go
technoweenie: This is the absolute minimum requirement for Git LFS to run safely. We can add a const if something else needs this specific value, but nothing does right now. Other bits in the codebase that check the git version already use `IsGitVersionAtLeast()`:

```bash
$ ack IsGitVersionAtLeast
commands/command_clone.go
89:    	if !git.Config.IsGitVersionAtLeast("2.9.0") {

git/git.go
406:func (c *gitConfig) IsGitVersionAtLeast(ver string) bool {
790:   	if !Config.IsGitVersionAtLeast("2.8.0") {

git/git_test.go
205:   	if !Config.IsGitVersionAtLeast("2.5.0") {

lfs/hook.go
41:    	customHooksSupported := git.Config.IsGitVersionAtLeast("2.9.0")
```

ttaylorr: Ah, this makes sense as an absolute min.

ttaylorr:
🌐

PR #1462, opened 2016-08-18
Technoweenie/errors next

technoweenie:
@ttaylorr: Once I fixed `errorWrapper` to embed an `errorWithCause`, it was just a matter of fixing up some error messages:

* the http errors are being wrapped with no `message`. Since `errorWrapper` needs an `errorWithCause`, I added a default `"LFS"` context message.
* Fixed a few spots during lfs pointer parsing where a `nil` error was being passed. I think the new pointer error messages are better.

Those two issues are because `errors.Wrap()` is fundamentally different from the current error wrapper:

## OLD

```golang
err := errors.New("fatal error writing to disk blah blah")
wrapped := newWrappedError(err, "error writing to .git/lfs/objects")
```

It currently replaces the error message with a message more applicable to LFS. Seeing LFS die with some internal io error or whatever is not helpful. Seeing something like this was my original intention:

```
error writing to .git/lfs/objects
fatal error writing to disk blah blah
```

For whatever dumb reason, the old pointer code passed a nil error, resulting in a message like:

```
Not a valid Git LFS pointer file.
<nil> # just a blank space, but this is where an inner error should be
```

## NEW

The new `errors.Wrap()` adds the context as a prefix. So, the pointer errors in my failing test looked like this:

```
Not a valid Git LFS pointer file.: Error
```

At first I thought something was weirdly reversing the string. Instead, I sent real error messages, and shortened the new error prefix in a09c4642b4b730fe179e58eb57da65aff39c9e89.

## TODO

I think this is good to merge into `errors-next`. I think we should also review all the errors and clean up the other spots that are passing `nil` errors.

Also, probably always import `github.com/pkg/errors` instead of just `errors` for consistency.

### DIFF
--- a/errutil/errors.go
+++ b/errutil/errors.go
@@ -465,7 +465,7 @@ func (e notAPointerError) NotAPointerError() bool {
 }
 
 func NewNotAPointerError(err error) error {
-	return notAPointerError{newWrappedError(err, "Not a valid Git LFS pointer file.")}
+	return notAPointerError{newWrappedError(err, "Pointer file error")}
 }
 
 type badPointerKeyError struct {
--- a/errutil/errors.go
+++ b/errutil/errors.go
@@ -51,7 +51,6 @@ package errutil
 
 import (
 	"fmt"
-	"reflect"
 	"runtime"
 
 	"github.com/pkg/errors"
@@ -73,7 +72,6 @@ func parentOf(err error) error {
 // IsFatalError indicates that the error is fatal and the process should exit
 // immediately after handling the error.
 func IsFatalError(err error) bool {
-	fmt.Println(err, reflect.TypeOf(err))
 	if e, ok := err.(interface {
 		Fatal() bool
 	}); ok {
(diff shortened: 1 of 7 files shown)


errutil/errors.go
```diff
@@ -51,18 +51,18 @@ package errutil
...
 
 	"github.com/pkg/errors"
 )
 
-func parentOf(err error) error {
-	type causer interface {
-		Cause() error
-	}
+type errorWithCause interface {
+	Error() string
```
ttaylorr: I think to be more explicit we could embed the `error` interface instead.

test/test-pointer.sh
```diff
@@ -120,9 +120,9 @@ begin_test "pointer --stdin with bad pointer"
 
   expected="Pointer from STDIN
 
-Not a valid Git LFS pointer file."
+Pointer file error: invalid header"
 
-  [ "$expected" = "$output" ]
+  diff -u <(printf "$expected") <(printf "$output")
```
ttaylorr: Nice 😉

ttaylorr:
Thanks so much for doing this, that interface-shadowing was tricky to figure out. LGTM!

PR #1463, opened 2016-08-18
Errors Next

ttaylorr:
This pull-request uses the `github.com/pkg/errors` package as a replacement for the old [`wrappedError`](https://github.com/github/git-lfs/blob/ac5a532665dc35c5c108726e5632e1d288224b64/errutil/errors.go#L300-L330) type that previously existed in the `errutil` package.

`github.com/pkg/errors` gives us the same sort of wrapping that we previously enjoyed, but it puts the onus on a third-party dependency. We still have to keep the `wrappedError` type around for now, since `pkg/errors` doesn't support any sort of `context map[string]interface{}`, but my plan is to remove that type in a future PR by either:

a. Dropping it (+ support for contexts) entirely, or:
b. Only using it where we _need_ contextual data

-

/cc @technoweenie

### DIFF
--- a/api/download_test.go
+++ b/api/download_test.go
@@ -321,8 +321,9 @@ func TestDownloadAPIError(t *testing.T) {
 		return
 	}
 
-	if err.Error() != fmt.Sprintf(httputil.GetDefaultError(404), server.URL+"/media/objects/oid") {
-		t.Fatalf("Unexpected error: %s", err.Error())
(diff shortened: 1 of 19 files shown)


technoweenie:
This looks good to me. I think we should merge this and then audit the error wrapping call sites (similar to c5036f63d0b7cacfa292363972b589aa26e289bd) to ensure it's not writing awkward error messages.

ttaylorr:
> I think we should merge this and then audit the error wrapping call sites 

Let's do it! 😄

PR #1467, opened 2016-08-19
Reword the description of HTTP 509 status

dpursehouse:
to make it consistent with the others

The others don't begin with "Returned if ".


technoweenie:
Thanks!

PR #1470, opened 2016-08-19
don't report "seen" objects as ready to checkout

technoweenie:
`fetchAndReportToChan()` can optionally receive a channel for sending `*lfs.WrappedPointer` objects. This is [used in `git lfs pull` to send files to checkout](https://github.com/github/git-lfs/blob/177984f2212a9232e1614c5ffc4e0bd55204783f/commands/command_pull.go#L41-L42).  The [checkout process](https://github.com/github/git-lfs/blob/177984f2212a9232e1614c5ffc4e0bd55204783f/commands/command_checkout.go#L74-L80) processes files as it reads them off the channel.

This PR renames `skippedAndUniqPointers()` to `readyAndMissingPointers()`, and changes it so the only pointers that are considered "ready" are those that already exist in `.git/lfs/objects`.


ttaylorr:
👍

Issue #1327, opened 2016-06-27
Add HTTP error 507 to the git-lfs protocol


green-coder:
I think it makes sense to have the HTTP error 507 (Insufficient Storage) in a Large File Server. The LFS protocol currently doesn't list this error in the optional codes to be returned by the server in case of problem.

I request to add it.

technoweenie:
Thanks for the suggestion. Sounds good to me :)

technoweenie:
Added this to the roadmap in #1438

PR #1473, opened 2016-08-20
Add HTTP 507 (Insufficient Storage) to the list of optional statuses

dpursehouse:
Mention that the server implementation may optionally return status
code 507.

This does not add any specific handling of 507 in either the client or
the test server implementation. If a server returns this status code,
it will be handled by the client as a generic "Server Error".

See #1327 and #1438.

### DIFF
--- a/docs/api/v1/http-v1-batch.md
+++ b/docs/api/v1/http-v1-batch.md
@@ -285,6 +285,7 @@ specify any rate limits, implementors are encouraged to set some for
 availability reasons.
 * 501 - The server has not implemented the current method.  Reserved for future
 use.
+* 507 - The server has insufficient storage capacity to complete the request.
 * 509 - The bandwidth limit for the user or repository has been exceeded.  The
 API does not specify any bandwidth limit, but implementors may track usage.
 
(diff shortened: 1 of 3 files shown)


ttaylorr:
Looks good. Would you mind also adding an entry for a `507` response [here](https://github.com/github/git-lfs/blob/ddfaae3c3a5ada9a0afe4c24c596b662bbc28ffc/httputil/response.go#L23-L29), as per https://github.com/github/git-lfs/pull/1438#issuecomment-241061811?

dpursehouse:
> Would you mind also adding an entry for a `507` response

That was omitted intentionally, such that the error code is mentioned in the document allowing servers to optionally return it, but the client behaviour is unchanged and a generic error will be shown.  This is consistent with the behaviour for other optional codes such as `429` and `529`.

If it's OK to add a specific message for `507`, I would prefer to do that in a separate PR along with the other ones that are also omitted.

dpursehouse:
@ttaylorr I added the messages in c7022e7c6f53c4a66134b53c2ffd3c91ce5d1d8d.

technoweenie:
@dpursehouse I would've accepted a separate PR, but this works too :+1:

ttaylorr:
Merged this in as a single commit by accident, I think GitHub remembered that from when I merged the 1.4.0 release branch. I'll unset that default so that this doesn't happen again.

Thanks anyways!

dpursehouse:
\o/

thanks.

Issue #1431, opened 2016-08-10
LFS cannot clone into current (empty) directory


tarka:
In LFS 1.2 this worked:

    mkdir MYPROJECT
    cd MYPROJECT
    git lfs clone git@bitbucket.org:ssmith/MYPROJECT.git .

With 1.3.1 it fails:

    $ git lfs clone git@bitbucket.org:ssmith/MYPROJECT.git .
    fatal: destination path '.' already exists and is not an empty directory.
    Error(s) during clone:
    git clone failed: exit status 128

This appears to be related to the premature initialisation of the LFS storage at the top level (see also #1231):

    $ find
    .
    ./lfs
    ./lfs/objects
    ./lfs/objects/logs
    ./lfs/tmp
    ./lfs/tmp/objects

Note that this top-level `lfs` dir is created even when the repo is cloned into it's default directory:

    $ git lfs clone git@bitbucket.org:ssmith/MYPROJECT.git
    Cloning into 'dacng'...
    remote: Counting objects: 2363, done.
    remote: Compressing objects: 100% (1724/1724), done.
    remote: Total 2363 (delta 779), reused 1846 (delta 440)
    Receiving objects: 100% (2363/2363), 34.38 MiB | 2.71 MiB/s, done.
    Resolving deltas: 100% (779/779), done.
    Checking connectivity... done.
    Git LFS: (498 of 498 files) 129.13 MB / 130.33 MB
    [] $ ll
    total 0
    drwxr-xr-x  4 ssmith wheel 136 Aug 10 10:54 .
    drwxrwxrwt 11 root   wheel 374 Aug 10 10:37 ..
    drwxr-xr-x 15 ssmith wheel 510 Aug 10 10:54 MYPROJECT
    drwxr-xr-x  4 ssmith wheel 136 Aug 10 10:54 lfs

(A workaround for this is to do a `git clone` with GIT_LFS_SKIP_SMUDGE and then `git lfs pull`.)

tarka:
Still present in LFS 1.4.

ttaylorr:
Hey @tarka, sorry I missed this when triaging. I knocked this one out in #1478 with a little write-up, and it should no longer be an issue. I'm planning to ship this in v1.4.1, but you're welcome to build and install LFS yourself if you need it sooner 😄

ttaylorr:
Merged!

PR #1478, opened 2016-08-22
Support wrapped clone in current directory

ttaylorr:
This PR supports cloning into the current directory, `"."`, a-la:

```bash
$ git lfs clone git@github.com:foo/bar.git .
```

Previously, this was broken because an invocation to the `git-lfs clone` command would create a `localstorage` directory due to a call to the [`localstorage.ResolveDirs`](https://github.com/github/git-lfs/blob/c21de411141a7536689a63e1f82bcc02f46fdf83/localstorage/currentstore.go#L28-L47) function causing both of [these](https://github.com/github/git-lfs/blob/c21de411141a7536689a63e1f82bcc02f46fdf83/localstorage/currentstore.go#L34-L35) files to be created. The presence of those files violated a `git-clone` invariant that the clone target must be empty. That source of that call is from the `init()` function in the `lfs` package, [here](https://github.com/github/git-lfs/blob/68c0e18d05de0e60c8ce774b5b3163062b47e5a0/lfs/lfs.go#L144).

In https://github.com/github/git-lfs/commit/e447be2172b4b767089516432065003f31ccf9d2, I removed that call, and instead replaced it with a `PreRun` hook on all of our commands, except `clone`. `PreRun` hooks, as dictated by the `cobra` package will run before the actual `Run` function of the command.

In a further PR, we should definitely take a look at which commands _actually_ need to have the `localstorage` stuff initialized.

--

/cc @technoweenie @rubyist @sinbad for review
/cc @tarka #1431 for a heads-up

### DIFF
--- a/commands/command_clean.go
+++ b/commands/command_clean.go
@@ -77,8 +77,9 @@ func cleanCommand(cmd *cobra.Command, args []string) {
 func init() {
 	RegisterSubcommand(func() *cobra.Command {
 		return &cobra.Command{
-			Use: "clean",
-			Run: cleanCommand,
+			Use:    "clean",
+			Run:    cleanCommand,
+			PreRun: resolveLocalStorage,
 		}
 	})
 }
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -233,8 +233,9 @@ func checkoutWithChan(in <-chan *lfs.WrappedPointer) {
 func init() {
 	RegisterSubcommand(func() *cobra.Command {
 		return &cobra.Command{
-			Use: "checkout",
-			Run: checkoutCommand,
+			Use:    "checkout",
+			Run:    checkoutCommand,
+			PreRun: resolveLocalStorage,
 		}
 	})
 }
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -6,10 +6,10 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/github/git-lfs/localstorage"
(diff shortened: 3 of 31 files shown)


commands/command_clean.go
```diff
@@ -77,8 +77,9 @@ func cleanCommand(cmd *cobra.Command, args []string) {
 func init() {
 	RegisterSubcommand(func() *cobra.Command {
 		return &cobra.Command{
-			Use: "clean",
-			Run: cleanCommand,
+			Use:    "clean",
+			Run:    cleanCommand,
+			PreRun: resolveLocalStorage,
```
sinbad: I wonder if we could remove all the duplication of setting `PreRun` (and potentially forgetting to  include it) by standardising on a wrapper function like `NewLfsCommand("clean", cleanCommand)` or similar?

sinbad:
That's a fascinating edge case 😄 

👍 from me

commands/command_clean.go
ttaylorr: I think I am going to leave this as-is, and only give the PreRun option to commands that need it.

sinbad: Isn't that all of them except clone? I expect all new commands will need it too, would be easier to remember with a utility function.

ttaylorr: I would rather be explicit about which commands are using it up front. If it ends up being the majority, then perhaps it makes sense to add a helper function, but not now.

technoweenie: I think there are a few commands that don't need it, like `env`, but I'd bet that the majority of commands will need it.

PR #1482, opened 2016-08-24
Test `RetriableReader`

ttaylorr:
This pull-request is a small addendum to @larsxschneider's #1454 to add tests around the `RetriableReader`.

The only behavior change in this PR is that `RetriableReader` will no longer re-wrap already retriable errors, per https://github.com/github/git-lfs/pull/1454#discussion_r76088729.

-

/cc @technoweenie @larsxschneider for 💭  and 👀

### DIFF
--- a/tools/iotools.go
+++ b/tools/iotools.go
@@ -98,10 +98,16 @@ func NewRetriableReader(r io.Reader) io.Reader {
 
 func (r *RetriableReader) Read(b []byte) (int, error) {
 	n, err := r.reader.Read(b)
-	// EOF is a successful response as it is used to signal a graceful end of input
-	// c.f. https://git.io/v6riQ
-	if err == nil || err == io.EOF {
+
+	// EOF is a successful response as it is used to signal a graceful end
+	// of input c.f. https://git.io/v6riQ
+	//
+	// Otherwise, if the error is non-nill and already retriable (in the
+	// case that the underlying reader `r.reader` is itself a
+	// `*RetriableReader`, return the error wholesale:
+	if err == nil || err == io.EOF || errors.IsRetriableError(err) {
 		return n, err
 	}
+
 	return n, errors.NewRetriableError(err)
 }
--- a/tools/iotools.go
+++ b/tools/iotools.go
@@ -102,7 +102,7 @@ func (r *RetriableReader) Read(b []byte) (int, error) {
 	// EOF is a successful response as it is used to signal a graceful end
 	// of input c.f. https://git.io/v6riQ
 	//
-	// Otherwise, if the error is non-nill and already retriable (in the
+	// Otherwise, if the error is non-nil and already retriable (in the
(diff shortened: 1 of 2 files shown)


tools/iotools.go
```diff
@@ -98,10 +98,16 @@ func NewRetriableReader(r io.Reader) io.Reader {
...
-	// c.f. https://git.io/v6riQ
-	if err == nil || err == io.EOF {
+
+	// EOF is a successful response as it is used to signal a graceful end
+	// of input c.f. https://git.io/v6riQ
+	//
+	// Otherwise, if the error is non-nill and already retriable (in the
+	// case that the underlying reader `r.reader` is itself a
+	// `*RetriableReader`, return the error wholesale:
+	if err == nil || err == io.EOF || errors.IsRetriableError(err) {
```
larsxschneider: Nice catch!

tools/iotools_test.go
```diff
@@ -0,0 +1,76 @@
+package tools_test
+
+import (
+	"bytes"
+	"io"
+	"testing"
+
+	"github.com/github/git-lfs/errors"
+	. "github.com/github/git-lfs/tools"
```
larsxschneider: What does the `.` mean here?

ttaylorr: A `.` import (called "dot-import") imports all of that packages publicly exported identifiers into the namespace of the current package. This way, I can refer to things that package `tools` exports in the `tools_test` package without having to say `tools.X` (I can instead just refer to `X`).

They are generally discouraged in production code, but I quite like them for tests since it allows me to isolate tests from private fields/methods while still keeping the code clean.

tools/iotools_test.go
```diff
@@ -0,0 +1,76 @@
...
+// ErrReader implements io.Reader and only returns errors.
+type ErrReader struct {
+	// err is the error that this reader will return.
+	err error
+}
+
+// Read implements io.Reader#Read, and returns (0, e.err).
+func (e *ErrReader) Read(p []byte) (n int, err error) {
+	return 0, e.err
+}
```
larsxschneider: Very nice 👍 ! I felt really bad that I did not provide a proper test case... but my golang foo was not sufficient to write a test quickly. Thanks for this great tutorial here! Next time I will try to write the test myself 😄

ttaylorr: Thanks for your great patch 😄

tools/iotools_test.go
ttaylorr: As @rubyist points out, on second thought, this is not such a great idea. Having dot-imports makes it hard for humans and tools to figure out where identifiers are coming from. We already isolate ourselves from private members, which is all that matters 😄

ttaylorr:
Travis seems to be misreporting builds as having failed, even though all of the tests are passing as expected. I kicked off a restart of the builds and will begin the 1.4.1 release process as soon as this is merged in.

PR #1495, opened 2016-08-31
commands: try out RegisterCommand()

technoweenie:
This is an incomplete PR that I wanted to try after reading [this conversation](https://github.com/github/git-lfs/pull/1478#discussion-diff-75845058).

Unlike the existing `RegisterSubcommand()`, `RegisterCommand()` wraps the initialization of the `*cobra.Command` with the following properties:

* `Use` - The lfs sub command name, needed by every command.
* `Run` - The function for the command.
* `PreRun` - This is the controversial part. _Most_ commands want this set.

I agree that it's better that the few commands that don't need `.git/lfs` initialized can disable the `PreRun` hook. By default though, all `RegisterCommand` callers should work.

The `bool` return value also lets commands choose if they're activated. Currently only used for experimental commands like `locks`.

If this approach is favorable, I'll update all `RegisterSubcommand()` callers, kill it, and change how `subcommandFuncs` works.

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -231,11 +231,5 @@ func checkoutWithChan(in <-chan *lfs.WrappedPointer) {
 }
 
 func init() {
-	RegisterSubcommand(func() *cobra.Command {
-		return &cobra.Command{
-			Use:    "checkout",
-			Run:    checkoutCommand,
-			PreRun: resolveLocalStorage,
-		}
-	})
+	RegisterCommand("checkout", checkoutCommand, nil)
 }
--- a/commands/command_clean.go
+++ b/commands/command_clean.go
@@ -75,11 +75,5 @@ func cleanCommand(cmd *cobra.Command, args []string) {
 }
 
 func init() {
-	RegisterSubcommand(func() *cobra.Command {
-		return &cobra.Command{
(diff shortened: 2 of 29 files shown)


sinbad:
👍  Exactly what I had in mind. My only thought is it would be nice to have some docs on `Register[Sub]Command` to make the process a bit clearer (what the bool means, why choice of enable is deferred until that function call returns).

sinbad:
When you apply this to the rest of the commands, please can you disable the pre-run for `git lfs version` as well? This will fix #1496

technoweenie:
I updated everything to use `RegisterCommand()` and removed `RegisterSubcommand()`. Updated the docs as well. Going to see if I can get rid of the bool return value...

technoweenie:
The cfg is available in `init()`, so I was able to pull the `isCommandEnabled()` calls out of the `RegisterCommand()` fn callback, and in `init()` directly.

ttaylorr:
Looks awesome 👍  I wasn't totally sure whether or not the `RegisterCommand` idea would be clear enough, but since it's documented it works fantastically. I'm really glad that we're saving space and still being clear about what's going on 👍

sinbad:
:metal:

PR #1497, opened 2016-09-01
Automatically upgrade old filters instead of requiring —force

sinbad:
Came across a case recently where someone hadn't used git-lfs for a while and suddenly got errors on `git lfs install` of the form:

```
filter.lfs.clean attribute should be "git-lfs clean --%f" but is "git-lfs clean %f"
```

This surprised me because I thought we had upgrade paths for our config, but turns out that was only for hooks. So rather than making them run `git lfs install --force` for something we already know about, I introduced the same concept as for hooks, a list of upgradeable values we used to use. 

While I was at this I removed the attribute upgrade path for `git-media`, since it was vague and over-general really and I don't think it's necessary any more? I can put it back if there was a specific reason it's still there but looked like a relic to me.

### DIFF
--- a/lfs/attribute.go
+++ b/lfs/attribute.go
@@ -2,14 +2,9 @@ package lfs
 
 import (
 	"fmt"
-	"regexp"
 	"strings"
 
 	"github.com/github/git-lfs/git"
-)
-
-var (
-	valueRegexp = regexp.MustCompile("\\Agit[\\-\\s]media")
 )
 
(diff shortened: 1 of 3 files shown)


technoweenie:
Awesome, thanks for adding this! 

The `git-media` stuff is _way_ outdated. I don't think there's any need to find every filter we used during development :)

ttaylorr:
This looks awesome. Thanks!

Issue #1501, opened 2016-09-04
Lfs plumbing commands to follow git conventions


kingces95:
A small nit: Allow a file to be passed to `git lfs clean` so that (on windows):

    type foo.dll | git lfs clean 

could be executed as

    git lfs clean foo.dll

`git hash-object` accepts a file path by default and reads from stdin when `--stdin` is passed. It may be a breaking change to make `git lfs clean` follow that convention but maybe the installer can update the filters to allow for this change.

sinbad:
Both `git lfs clean` and `git lfs smudge` are _solely_ provided to interact with the git filter system, they're not "plumbing" commands per se. I don't think it's worth breaking compatibility for this.

I think the command you're looking for is `git lfs pointer`.

ttaylorr:
Yes, both the `clean` and `smudge` commands are internal commands that are not intended for use by a user. I've gone ahead and updated our documentation for the `git-lfs clean` command to point to `git-lfs-pointer(1)`'s documentation for this sort of use in #1503.

PR #1503, opened 2016-09-05
docs/man: reference git-lfs-pointer(1) in clean documentation

ttaylorr:
This pull-request adds some new documentation to `git-lfs-clean`'s man page to reflect that it is not a part of the user-facing plumbing commands. Instead, it points to the `git-lfs-pointer(1)` documentation in order to preview what the generated pointer data would look like for a particular object.

--

/cc @sinbad @kingces95 @technoweenie #1501


sinbad:
👍 thanks!

Issue #1486, opened 2016-08-26
Client does not display error message when server returns 503


dpursehouse:
Using the current head of the master branch:

```
git-lfs $ git describe
v1.4.0-9-g89f37f4
```

When the server returns 503 with a message `LFS is not available`, the client does not display the message:

```
test-project $ git push origin HEAD:refs/for/master
Git LFS: (0 of 3 files) 0 B / 1.31 MB

Errors logged to /Users/dpursehouse/temp/test-project/.git/lfs/objects/logs/20160826T144116.556014811.log
Use `git lfs logs last` to view the log.
error: failed to push some refs to 'ssh://admin@localhost:29418/test-project'
```

The user has to use `git lfs logs last` to see the message:

```
[snip]
batch response: Fatal error: http: LFS is not available
[snip]
```

This seems to be a recent change in behaviour.  With version 1.3.1 the error was displayed:

```
test-project $ git push origin HEAD:refs/for/master
Git LFS: (0 of 3 files) 0 B / 1.31 MB
LFS is not available

Errors logged to /Users/dpursehouse/temp/test-project/.git/lfs/objects/logs/20160826T145330.6359595.log
Use `git lfs logs last` to view the log.
error: failed to push some refs to 'ssh://admin@localhost:29418/test-project'
```

technoweenie:
Thanks for the report, looks like a regression with the new errors package.

ttaylorr:
👋  @dpursehouse, thanks for opening this.

Just to clarify, are you talking about the v1.3 batch API wherein an error is reported through the [batch response](https://github.com/github/git-lfs/blob/master/docs/api/v1.3/http-v1.3-batch-response-schema.json#L54-L63)?

I set up a branch to try and replicate this, but was unable to reproduce. In [`git-lfs/503-no-report`](https://github.com/github/git-lfs/compare/503-no-report?expand=1), the error was reported as follows:

```
$ git lfs push origin master
Git LFS: (1 of 1 files, 1 skipped) 2 B / 2 B, 16 B skipped                     
   [432cf07eb9a1b9d66f3009f001064c2466c237015f64ac05177b835d7effd138] welp: [503] welp
   error: failed to push some refs to 'http://127.0.0.1:50160/test-push-failures-status-batch-503'
```

The error traces itself through the code path as follow:

[...]

PR #1505, opened 2016-09-05
Log fatal error messages in command output

ttaylorr:
This pull-request ensures Git LFS will log an error message when it encounters a fatal error using `commands.FullError`.

In versions of Git LFS prior to v1.4.0, an error message would be logged out to the console and the full error (along with command and other useful debugging information) would be logged to a file on disk, viewable by `git lfs logs`.

This pull-request ensures that the first line of an error message is always logged to the console and the stack trace and expanded debugging information are logged out to disk. This diff (contained in 1dabeb5) makes that happen:

```diff
diff --git a/commands/commands.go b/commands/commands.go
index 6964396..ffcf1cf 100644
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -85,7 +85,7 @@ func FullError(err error) {
 
 func errorWith(err error, fatalErrFn func(error, string, ...interface{}), errFn func(string, ...interface{})) {
        if Debugging || errors.IsFatalError(err) {
-               fatalErrFn(err, "")
+               fatalErrFn(err, "%s", err)
                return
        }
 
```

The `%s, err` format was intentionally used instead of `%s, err.Error()` since `err` could be `nil`.

To test this behavior, I first added a new [content handler](https://github.com/github/git-lfs/blob/7ccc6b5cec1f8eec1d05f8e582c8e4b825d3bcc2/test/cmd/lfstest-gitserver.go#L42-L54) to the test gitserver to [return some JSON](https://github.com/github/git-lfs/blob/7ccc6b5cec1f8eec1d05f8e582c8e4b825d3bcc2/test/cmd/lfstest-gitserver.go#L493-L501) with 503 status code and a `message` that we assert is contained in the client's output.

I used the `/storage` handler instead of the `lfsBatchHandler`, since it seemed inappropriate to change the outer `message` server error in response to the contents of a single file scoped at the batch-level.

To test that the error made its way to the output of the original `git lfs push`, I added an optional second `$msg` parameter to the `push_fail_test` function in https://github.com/github/git-lfs/commit/50ba7eb794249b6c0fee9b1baae55c449783cb51. When `$msg` is non-empty according to `[ ! -z "$msg" ]`, `grep` is run against the output of `git lfs push` with `$msg` as the capture expression. Since this adds some additional complexity to the `push_fail_test` function, I added some documentation to explain, but if there's a more appropriate place to make this assertion I would appreciate your thoughts 😄 

👋  Thanks in advance for the review!

--

/cc @technoweenie @rubyist @sinbad for review
/cc @dpursehouse #1486 for 👀


dpursehouse:
👍 works as expected. Thanks.

commands/commands.go
```diff
@@ -85,7 +85,7 @@ func FullError(err error) {
 
 func errorWith(err error, fatalErrFn func(error, string, ...interface{}), errFn func(string, ...interface{})) {
 	if Debugging || errors.IsFatalError(err) {
-		fatalErrFn(err, "")
+		fatalErrFn(err, "%s", err)
```
sinbad: Instead of having to do this, would it make more sense for `LoggedError` to just print the error contents if `format` is blank? Seems more intuitive.

ttaylorr: I was wondering about this when I wrote this patch. My thinking is to leave it as-is (`"%s", err.Error()`), since that's what we're doing with [the errFn call](https://github.com/github/git-lfs/blob/7ccc6b5cec1f8eec1d05f8e582c8e4b825d3bcc2/commands/commands.go#L92).

I like that giving a blank format string means no output is generated (other than the standard panic message about logging to a file) and that that behavior is explicit.

I'm definitely open to changing this, though I think we'd want to do it at both call sites.

sinbad: That's a fair point too. I don't feel that strongly either way, perhaps just make it clearer in the doc for `LoggedError`?

sinbad:
:+1:

commands/commands.go
ttaylorr: > perhaps just make it clearer in the doc for LoggedError

Good call 😄 . I went ahead and added this in 0a40ef2.

technoweenie:
:+1: Nice catch. I think I was thinking more along @sinbad's line of reasoning when I added this code. I think being explicit is better though.

PR #1506, opened 2016-09-06
Make AppVeyor CI for Windows work again

sschuberth:
This build the binaries and installer for now. Not running tests yet.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,56 +1,17 @@
-version: 1.0.{build}
-init:
-- cmd: >-
-    set REPO_DIR=%cd%
-
-
-    mkdir %BASHROOT%
-
-    cd %BASHROOT%
-
-
-    curl -LOk https://github.com/git-for-windows/git/releases/download/v2.6.2.windows.1/PortableGit-2.6.2-64-bit.7z.exe
-
-
-    7z x PortableGit-2.6.2-64-bit.7z.exe > nul
-
-
-    mkdir c:\go15
-
-    cd \go15
-
-
-    curl -LO https://storage.googleapis.com/golang/go1.5.1.windows-amd64.zip
-
-    unzip -o go1.5.1.windows-amd64.zip > nul
-
-
-    set PATH=%BASHROOT%\bin;%GOROOT%\bin;%PATH%
-
-
-    cd %REPO_DIR%
-environment:
-  GOROOT: c:\go15\go
-  BASHROOT: c:\bash2
 install:
-- cmd: 
+  - cinst InnoSetup -y
+  - set PATH="C:\Program Files (x86)\Inno Setup 5";%PATH%
+
 build_script:
-- cmd: >-
-    mkdir src\github.com\github
+  - bash -c 'GOARCH=386 script/bootstrap'
+  - mv bin\git-lfs git-lfs-x86.exe
+  - bash -c 'GOARCH=amd64 script/bootstrap'
+  - mv bin\git-lfs git-lfs-x64.exe
 
+after_build:
+  - iscc script\windows-installer\inno-setup-git-lfs-installer.iss
 
-    mklink /J src\github.com\github\git-lfs .
-
-
-    set GOPATH=%CD%
-
-
-    go version
-
-
-    bash script\bootstrap
-
-
-    mv bin\git-lfs bin\git-lfs.exe
 artifacts:
-- path: bin\git-lfs.exe
+  - path: git-lfs-x86.exe
+  - path: git-lfs-x64.exe
+  - path: git-lfs-windows-*.exe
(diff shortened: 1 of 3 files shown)


script/bootstrap
```diff
@@ -1,9 +1,10 @@
 #!/usr/bin/env bash
 set -e
 
-# Only set GOPATH on non-Windows platforms as Windows can't do the symlinking
-# that we need.
-if [ -z "$GOPATH" ] && uname -s | grep -vq "_NT-"; then
+# Tell MSYS to really create symbolic links.
+export MSYS=winsymlinks:nativestrict
```
ttaylorr: Are you able to do this from within the `appveyor.yml`?

sschuberth: I guess so, but then calling script/bootstrap on a local Windows developer machine (from Git Bash) would not properly work unless that variable is already set.

sschuberth: If you feel better about it, I can limit setting that variable to Windows...

appveyor.yml
```diff
@@ -1,56 +1,17 @@
-version: 1.0.{build}
```
ttaylorr: Do we still need this `version`?

script/fmt
```diff
@@ -6,7 +6,7 @@ hash goimports 2>/dev/null && {
 }
 
 # don't run gofmt in these directories
-ignored=(/bin/ /docs/ /log/ /man/ /tmp/ /vendor/ /rpm/ /docker/ /debian/)
+ignored=(/bin/ /docs/ /log/ /man/ /tmp/ /vendor/ /rpm/ /docker/ /debian/ /src/)
```
ttaylorr: Is this `src/` directory used by AppVeyor as a workspace to build the Windows installer? If so, perhaps it would be good to document this above:

```bash
# don't run gomt in these directories...
#  - `src/`: used by AppVeyor to ...
```

ttaylorr:
This is looking good. I left a few minor comments, mostly me asking questions about how AppVeyor works 😅 .

One thing to note is that this won't generate signed releases of Git LFS that we can upload as our official Windows installer. I don't think this is a problem for now, since getting some sort of test harness in place for verifying that the Windows installer builds correctly is definitely a step forward.

I'd love to look into how we could get AppVeyor to sign our releases for us in the future, since that's definitely something that we'd like to do.

appveyor.yml
sschuberth: IMO it's not needed (I believe it's only used in AppVeyor's web interface to refer to the builds), and it's wrong anyway.

script/bootstrap
ttaylorr: Ah, good idea. If we're in a Windows environment while running `script/bootstrap`, _then_ we can set `$MSYS`.

script/fmt
sschuberth: No, this is the src directory that is created by script/bootstrap, i.e. it's Golang related.

script/bootstrap
sschuberth: Right. I'll simply re-use parts of the existing check for that.

script/fmt
ttaylorr: Ack! I forgot that we do that as a part of our bootstrap process. OK, this is 👍

sschuberth:
> One thing to note is that this won't generate signed releases of Git LFS that we can upload as our official Windows installer.

The Inno Setup compiler `iscc` supports the `/S` option to specify a signing tool, so I guess that should be easy to add.

sschuberth:
I've pushed another version that only exports the MSYS (and now also CYGWIN) variable on Windows.

ttaylorr:
Thanks for being so quick to respond to my comments! I'm going to wait for the tests to pass and then merge this to master and test it out.

sschuberth:
Please do not merge yet,  there's a small bug I'll fix from home in an hour or so!

sschuberth:
Fixed, see the AppVeyor artifacts at https://ci.appveyor.com/project/sschuberth/git-lfs/build/1.0.19/artifacts

ttaylorr:
Hooray! https://ci.appveyor.com/project/git-lfs/git-lfs/build/artifacts

sschuberth:
Interesting, why's this at *git-lfs/git-lfs* and not at https://ci.appveyor.com/project/github/git-lfs/build/artifacts?

ttaylorr:
We're planning to move to https://github.com/git-lfs soon anyway.

sschuberth:
Also, looking at https://github.com/github/git-lfs/pull/1505 it seems two AppVeyor builds get kicked off, *continuous-integration/appveyor/branch* and *continuous-integration/appveyor/pr*. Is this intended?

ttaylorr:
Good catch. I enabled the "Do not build feature branches with open Pull Requests" option.

PR #1507, opened 2016-09-06
Allow fetch to run without a remote configured

sschuberth:
For newly initialized local repositories that are configured to track
existing LFS objects by manually creating pointer files via
git-lfs-pointer, there might be no remote configured yet. Still it should
be possible to fetch the LFS objects. Allow this by postponing the check
for a remote until it is really required.

### DIFF
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -42,12 +42,7 @@ func fetchCommand(cmd *cobra.Command, args []string) {
 		}
 		cfg.CurrentRemote = args[0]
 	} else {
-		// Actively find the default remote, don't just assume origin
-		defaultRemote, err := git.DefaultRemote()
-		if err != nil {
-			Exit("No default remote")
-		}
-		cfg.CurrentRemote = defaultRemote
+		cfg.CurrentRemote = ""
 	}
 
 	if len(args) > 1 {
@@ -246,6 +241,16 @@ func fetchPointers(pointers []*lfs.WrappedPointer, include, exclude []string) bo
 // Fetch and report completion of each OID to a channel (optional, pass nil to skip)
 // Returns true if all completed with no errors, false if errors were written to stderr/log
 func fetchAndReportToChan(allpointers []*lfs.WrappedPointer, include, exclude []string, out chan<- *lfs.WrappedPointer) bool {
+	// Lazily initialize the current remote.
+	if len(cfg.CurrentRemote) == 0 {
+		// Actively find the default remote, don't just assume origin
+		defaultRemote, err := git.DefaultRemote()
+		if err != nil {
+			Exit("No default remote")
+		}
+		cfg.CurrentRemote = defaultRemote
+	}
+
 	ready, pointers, totalSize := readyAndMissingPointers(allpointers, include, exclude)
 	q := lfs.NewDownloadQueue(len(pointers), totalSize, false)
 


ttaylorr:
@sschuberth -- thanks for another contribution. I think there are a few test failures that I'm seeing happening:

```
commands/command_fetch.go:45: cannot use nil as type string in assignment
commands/command_fetch.go:160: invalid operation: cfg.CurrentRemote == nil (mismatched types string and nil)
```

This is because the `CurrentRemote` field on the `*config.Configuration` type is a `string`, meaning you should assign empty string (`""`) to it when emptying it out, or check if the `len(str) > 0)` when checking if it has value.

Once you fix those up, however, I think this looks 👍.

sschuberth:
@ttaylorr Thanks, fixed.

ttaylorr:
@sschuberth thanks for getting back to me so quickly. The good news is that LFS compiles now. The bad news is that now the tests are failing for a legitimate reason.

The output of the test failures that I'm looking at can be found: [here](https://travis-ci.org/github/git-lfs/builds/157896061).

I'm seeing the following:

```
+git remote rename origin something
+git lfs fetch
http: Post objects/batch: unsupported protocol scheme ""
```

which means that [after the remote is renamed to something other than `origin`](https://github.com/github/git-lfs/blob/a06accadc8a4bcd6a9f225f8d5eb3806c393b7d2/test/test-fetch.sh#L436-L437), LFS tries to preform a `fetch`, but fails due to trying to parse an unsupported protocol: empty-string. This makes me think that the an empty string was returned from `config.CurrentRemote`, and the default remote was missing. The order of execution that would cause this to happen is as follows:

0. The fetch command calls the `api.Batch` function, and a new batch request is generated [here](https://github.com/github/git-lfs/blob/a06accadc8a4bcd6a9f225f8d5eb3806c393b7d2/api/api.go#L69).
0. Upon trying to generate that new batch request, an SSH authentication is preformed [here](https://github.com/github/git-lfs/blob/ddfaae3c3a5ada9a0afe4c24c596b662bbc28ffc/api/v1.go#L119).
0. That SSH authentication attempts to grab the correct server-side endpoint for the given operation [here](https://github.com/github/git-lfs/blob/ddfaae3c3a5ada9a0afe4c24c596b662bbc28ffc/auth/ssh.go#L26).
0. And when resolving that endpoint, the `config` package picks either the `cfg.CurrentRemote` (if not empty), or the default remote, [here](https://github.com/github/git-lfs/blob/master/config/config.go#L246-L252).

We can't rely on the default remote in this test-case, because there is no `origin`, anymore. Instead, we have to rely on the `cfg.CurrentRemote`, which I have a hunch is being set in the wrong place.

In the `commands/command_fetch.go` file, you can see there are many paths that the fetch command can take. Not all of them, however, lead to `fetchRecent`, which is where you have the current remote check. All of the paths do, however, lead to `fetchAndReportToChan`, which is where I would make the current remote check.

Would you mind trying that one out and seeing if it resolves these test failures?

sschuberth:
Thanks for the detailed analysis! Let's see how my update goes...

ttaylorr:
@sschuberth Nice! I think we fixed it. Travis reported all green except for one build which failed with a message about unresponsive output, making me think that the worker had died.

I went ahead and restarted that build [here](https://travis-ci.org/github/git-lfs/jobs/157930650). Once it passes, I think we're ready to merge.

Thanks again for a great contribution!

PR #1509, opened 2016-09-06
[WIP] appveyor: run `cibuild` on Windows

ttaylorr:


appveyor.yml
```diff
@@ -1,6 +1,16 @@
+clone_folder: c:\go\src\github.com\github\git-lfs
+
+environment:
+  GOPATH: c:\go
+
 install:
+  - set PATH=%GOPATH%\bin;%PATH%;
```
sschuberth: This is not necessary, system Go is already in `PATH`.

appveyor.yml
```diff
@@ -1,6 +1,16 @@
...
+  - go version
```
sschuberth: Even if you left this in for debugging, I don't think it adds much value as we have the standard `Using go1.6.2` lines anyway.

appveyor.yml
```diff
@@ -1,6 +1,16 @@
...
   - cinst InnoSetup -y
-  - set PATH="C:\Program Files (x86)\Inno Setup 5";%PATH%
+  - set PATH="C:\Program Files (x86)\Inno Setup 5";%PATH%;
```
sschuberth: Why the addition of the trailing `;`?

appveyor.yml
```diff
@@ -1,6 +1,16 @@
...
+
+test_script:
+  - bash -c 'script/cibuild'
```
sschuberth: This can be simplified a bit to just `bash script/cibuild`.

appveyor.yml
```diff
@@ -1,6 +1,16 @@
...
+  GOPATH: c:\go
+
 install:
+  - set PATH=%GOPATH%\bin;%PATH%;
+  - go version
   - cinst InnoSetup -y
-  - set PATH="C:\Program Files (x86)\Inno Setup 5";%PATH%
+  - set PATH="C:\Program Files (x86)\Inno Setup 5";%PATH%;
+
+test_script:
```
sschuberth: How about moving this block after `after_build`? I feel that would be more natural as it matches execution order.

ttaylorr:
Hi @sschuberth, thanks for an early review. This PR was expiremental. I'm going to close this out for now, since Windows tests are not at the top of my immediate list of things to do. I will re-open this in the future and get Windows tests running in the wild.

PR #1512, opened 2016-09-09
testhelpers: do not set credential.helper multiple times

sschuberth:
We already set this globally in setup().

### DIFF
--- a/test/test-credentials-no-prompt.sh
+++ b/test/test-credentials-no-prompt.sh
@@ -27,7 +27,6 @@ begin_test "attempt private access without credential helper"
   git add .gitattributes
   git commit -m "initial commit"
 
-  git config --global credential.helper lfsnoop
   git config credential.helper lfsnoop
   git config -l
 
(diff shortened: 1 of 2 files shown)


ttaylorr:
Hi, @sschuberth - thanks for this 😄 . I think this is kept in because we do override the `--global` `credential.helper` a few places throughout our tests (one example is [here](https://github.com/github/git-lfs/blob/ebf7fbd3ec3771497e2beb8911d856d5d273e920/test/test-credentials-no-prompt.sh#L30-L31)).

I'm going to defer to @technoweenie, since we worked on some credential helper stuff a few months ago.

sschuberth:
Your example in `test-credentials-no-prompt.sh` is also the only one I've found. I don't see a reason why `lfsnoop` is set both globally and locally there. Locally should be enough, I guess, so I've added a commit to remove setting it globally. Let's see how the tests go :-)

sschuberth:
Getting the simplification right involves more work, closing for now.

PR #1518, opened 2016-09-11
Simplify getting the absolute Git root directory

sschuberth:
"git rev-parse --show-toplevel" is defined to already return an absolute
path, so there is no need to convert it.


ttaylorr:
Thanks!

Issue #1519, opened 2016-09-12
Packaging install/uninstall issues


petermarko:
## Issue 1

I'm using debian:jessie package and tried to upgrade to 1.4.1.
It failed when configuring package with

```
Setting up git-lfs (1.4.1) ...
The filter.lfs.clean attribute should be "git-lfs clean -- %f" but is "git-lfs clean %f"
Run `git lfs install --force` to reset git config.
dpkg: error processing package git-lfs (--configure):
 subprocess installed post-installation script returned error exit status 2
Errors were encountered while processing:
 git-lfs
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

My configuration was valid (and default) in old version.
The old format is no longer valid and I need to correct it manually.
The package should include migration script to fix this automatically.
With current git-lfs code, `debian/postinst` should include `--force` flag:

```
#!/bin/sh
git lfs install --system --force
```
## Issue 2

When investigating I have noticed that the debian package contains postinst and prerm
scripts installing and uninstalling git-lfs configuration _unconditionally_.
This seems to be wrong, because `git lfs uninstall` command removes all configuration entries.
I have (in addition to 3 mandatory default entries) also setting of url of my server.
This is valid place to configure it per man page of `git-lfs-config`

```
git-lfs  uses the same configuration files as git-config(1) with the same precedence.
```

So now any upgrade of git-lfs will remove my configuration, which is obviously an error.
To fix this, I propose following `debian/prerm`:

```
#!/bin/sh
if [ "$1" = "remove" ] ; then
  git lfs uninstall
fi
```

This will skip the removal when the package is being upgraded,
while handling uninstall of package properly.
## Note 1

Previous suggestions were made to have a quick solution without any code change.
In my opinion, the proper solution would be to add additional functions in addition to
`git lfs install` and `git lfs uninstall`, which would do migration and
disabling/enabling of git-lfs while keeping othe configuration untouched.

This would be to properly support all relevant debian package management options
- install - install package and configure it
- upgrade - upgrade package and update configuration if needed
- remove - remove the package (keeping config files, just disabling the feature)
- purge - remove configuration files

Note that upgrade should not enable git-lfs is user manually disabled it,
it should just migrate configuration.
## Note 2

This was written for debian packaging.
But the same issue is for sure present also for rpm (or any other package type).
I have no experience there, so I do not have any code/configuration proposals.

This is also reason I'm not providing pull request, since it would be incpomlete.
[...]

PR #1521, opened 2016-09-12
git,lfs: improve behavior of `git-lfs uninstall`

ttaylorr:
@petermarko points out in https://github.com/github/git-lfs/issues/1519 that `git-lfs uninstall` will remove all `.gitconfig` values in the `filters.lfs` section forcefully.

In a scenario where a user runs `git lfs install`, their .gitconfig looks something like:

```bash
$ git config -l | grep "filter\.lfs"
filter.lfs.required=true
```

If they either added a new value to the `filter.lfs` section, or _changed_ one of the existing values (i.e., `git config filter.lfs.smudge "my_custom_thing"`), then upon running `git-lfs uninstall`, both of those changes would be overwritten. This PR addresses that.

The new behavior of the `(a *Attribute) Uninstall()` method now only removes:

- Paths that are explicitly set by `git lfs install`, and
  - matches the value that `git lfs install` set it to, or
  - matches one of the values that is "upgradable" for that path

To do this, LFS takes advantage of the `value_regexp` parameter in `git-config`:

> If you want to update or unset an option which can occur on multiple lines, a POSIX regexp value_regex needs to be given. Only the existing values that match the regexp are updated or unset.

This PR is a first step to some upgrade improvements that I'd like to ship to our package installations. For more information on this, see: https://github.com/github/git-lfs/issues/1519#issuecomment-246502641

--

/cc @technoweenie @petermarko #1519

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -352,9 +352,21 @@ func (c *gitConfig) UnsetGlobal(key string) (string, error) {
 	return subprocess.SimpleExec("git", "config", "--global", "--unset", key)
 }
 
+// UnsetSystemMatchingValues removes the git config value for the key from the
+// global config whose values match the given regular expression.
+func (c *gitConfig) UnsetGlobalMatchingValue(key string, re *regexp.Regexp) (string, error) {
+	return subprocess.SimpleExec("git", "config", "--global", "--unset", key, re.String())
+}
+
 // UnsetSystem removes the git config value for the key from the system config
 func (c *gitConfig) UnsetSystem(key string) (string, error) {
 	return subprocess.SimpleExec("git", "config", "--system", "--unset", key)
+}
+
+// UnsetSystemMatchingValues removes the git config value for the key from the
+// system config whose values match the given regular expression.
+func (c *gitConfig) UnsetSystemMatchingValue(key string, re *regexp.Regexp) (string, error) {
+	return subprocess.SimpleExec("git", "config", "--system", "--unset", key, re.String())
 }
 
 // UnsetGlobalSection removes the entire named section from the global config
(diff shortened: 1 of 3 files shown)


technoweenie:
Nice solution here. I didn't even realize you could pass a regex to `git config --unset` like that. However, I don't think this change is solving a real problem. If a user doesn't like `git lfs uninstall` removing all `filter.lfs.*` keys, they're doing something custom, and should just not run that command. All it does is makes some `git config` changes, and removes some LFS hooks in `.git/hooks`.

I think @petermarko's point in "Issue 2" of #1519 is that `git lfs uninstall` should _only_ be run when removing the package. Upgrades should probably only run `git lfs update`, instead of `git lfs uninstall && git lfs install`. I agree with that.

If I misunderstood #1519, or someone else posts a compelling argument, I think this would be a worthy change.

ttaylorr:
@technoweenie I agree. Thinking about this patch a little bit more, I think that you're 👍 in saying that if a user wants to remove LFS in a unique way, they should just run those commands themselves.

> Upgrades should probably only run `git lfs update`

What do you think about adding two new flags to `update`: `--system`, and `--local`? Right now `update` only works inside the repository because it always attempts to upgrade the contents of `.git/lfs`. Even though it works inside the repository, my hunch is that this constraint only exists so it can do the local upgrades, since it _also_ touches the system configuration in doing a hook installation.

By adding these flags, we could tell the RPM and Debian packages to run `git lfs update --system` after installing the package to either upgrade, or install the system hooks.

In terms of supporting more Debian package management options (from #1519):

> remove - remove the package (keeping config files, just disabling the feature)
> purge - remove configuration files

I'm trying to figure out where we should draw the line between `purge` and `remove`. My instinct is to make them functionally the same, per your comments above.

ttaylorr:
As a side note, teaching `update` some new tricks would also lend itself well to the different installation options we get over in RPM world. Specifically: install, uninstall, and upgrade (running `git lfs install`, `git lfs uninstall`, and `git lfs update --system`, respectively).

PR #1523, opened 2016-09-13
Make path comparison robust against Windows short / long path issues

sschuberth:
On Windows 10, this test was failing because the current user's home
directory was in DOS-style 8.3 short path notation in the expected Git
path, whereas the actual path was using the long path notation. Fix that
by comparing files based on Stat() instead of comparing strings.


ttaylorr:
👍

PR #1528, opened 2016-09-19
lfs/batcher: support immediate Flush()-ing

ttaylorr:
This pull-request adds the ability to immediately flush the currently accumulating batch.

The motivation behind this pull-request is the ability to support multiple retries per object in the current transfer queue implementation. Currently, we [keep track of the items that failed to transfer](https://github.com/github/git-lfs/blob/b7b1f2b7985c8f7a18c563c70898f669ffb8c13c/lfs/transfer_queue.go#L420) and then [retry them once more](https://github.com/github/git-lfs/blob/b7b1f2b7985c8f7a18c563c70898f669ffb8c13c/lfs/transfer_queue.go#L233-L242) before we exit in [`Wait()`](https://github.com/github/git-lfs/blob/b7b1f2b7985c8f7a18c563c70898f669ffb8c13c/lfs/transfer_queue.go#L218-L221).

In the future, we'd like to support retrying objects multiple times, by re-submitting them into the Batcher, and then getting batches of retries out. In order to do this, we need to know how many retries there are left. In my initial pass, I simply re-added failed items to the batcher, and waited for a batch to come out. However, a batch isn't delivered until we either [fill the batch size](https://github.com/github/git-lfs/blob/b7b1f2b7985c8f7a18c563c70898f669ffb8c13c/lfs/batcher.go#L63) (unlikely for retries), or [Exit()](https://github.com/github/git-lfs/blob/b7b1f2b7985c8f7a18c563c70898f669ffb8c13c/lfs/batcher.go#L65-L68), [and close all of the channels](https://github.com/github/git-lfs/blob/b7b1f2b7985c8f7a18c563c70898f669ffb8c13c/lfs/batcher.go#L51).

To allow us to force a batch to be delivered immediately, we guarentee that an item has been delivered on the channel (more information in https://github.com/github/git-lfs/commit/98873d87c4b06beecbba57221e70dd975ce7d7d4), and then immediately Flush() the batch, giving us a batch of 1 item to be retried.

In future PRs, I plan to make this retry behavior smarter, by overflowing retry buckets, or batching the retries together. As a first pass, I just want it to work 😄 

As a side 📝 , the test cases that I introduced in https://github.com/github/git-lfs/pull/613 (coincidentally, my first LFS patch) did not fully cover the behavior of the batcher, namely what happens when an uneven amount of items is added. This is what was causing the deadlock in https://github.com/github/git-lfs/pull/1528/commits/98873d87c4b06beecbba57221e70dd975ce7d7d4. With some debugging, the old tests produced:

```
Creating batcher with 5 items and size 3
Added item
Added item
Added item
Added item # <- gone
Added item # <- gone
Grabbing batch #1
Got batch size of 3
<end>
```

The remaining two items are gone! No one ever checks for them. I fixed this behavior in https://github.com/github/git-lfs/pull/1528/commits/a0d0635c82cb8fec724ff24ea7664b28d3dc9d03.

--

/cc @rubyist since you were the original author
/cc @technoweenie @sinbad for 👀  and 💭 s

### DIFF
--- a/lfs/batcher.go
+++ b/lfs/batcher.go
@@ -66,10 +66,10 @@ func (b *Batcher) Exit() {
 }
 
 // acceptInput runs in its own goroutine and accepts input from external
-// clients. Without truncation, it fills and dispenses batches in a sequential
-// order: for a batch size N, N items will be processed before a new batch is
-// ready. If a batch is truncated while still filling itself, it will be
-// returned immediately, opening up a new batch for all subsequent items.
+// clients. Without flushing, the batch is filled completely in a sequential
+// order, and then dispensed. If, while filling a batch, it is flushed part-way
+// through, the batch will be dispensed with its current contents, and all
+// subsequent Add()s will be placed in the next batch.
 func (b *Batcher) acceptInput() {
 	var exit bool
 
--- a/lfs/batcher.go
+++ b/lfs/batcher.go
@@ -13,6 +13,7 @@ type Batcher struct {
 	batchSize  int
 	input      chan interface{}
 	batchReady chan []interface{}
+	truncate   chan interface{}
 }
 
 // NewBatcher creates a Batcher with the batchSize.
@@ -21,6 +22,7 @@ func NewBatcher(batchSize int) *Batcher {
 		batchSize:  batchSize,
 		input:      make(chan interface{}),
 		batchReady: make(chan []interface{}),
+		truncate:   make(chan interface{}),
 	}
 
 	go b.acceptInput()
@@ -32,6 +34,7 @@ func NewBatcher(batchSize int) *Batcher {
 func (b *Batcher) Add(t interface{}) {
 	if atomic.CompareAndSwapUint32(&b.exited, 1, 0) {
 		b.input = make(chan interface{})
+		b.truncate = make(chan interface{})
 		go b.acceptInput()
 	}
 
@@ -44,16 +47,25 @@ func (b *Batcher) Next() []interface{} {
 	return <-b.batchReady
 }
 
+// Truncate causes the current batch to halt accumulation and return
+// immediately, even if it is smaller than the given batch size.
+func (b *Batcher) Truncate() {
+	b.truncate <- struct{}{}
+}
+
 // Exit stops all batching and allows Next() to return. Calling Add() after
 // calling Exit() will reset the batcher.
 func (b *Batcher) Exit() {
 	atomic.StoreUint32(&b.exited, 1)
 	close(b.input)
+	close(b.truncate)
 }
 
 // acceptInput runs in its own goroutine and accepts input from external
-// clients. It fills and dispenses batches in a sequential order: for a batch
-// size N, N items will be processed before a new batch is ready.
+// clients. Without truncation, it fills and dispenses batches in a sequential
+// order: for a batch size N, N items will be processed before a new batch is
+// ready. If a batch is truncated while still filling itself, it will be
+// returned immediately, opening up a new batch for all subsequent items.
 func (b *Batcher) acceptInput() {
 	exit := false
 
@@ -61,12 +73,16 @@ func (b *Batcher) acceptInput() {
 		batch := make([]interface{}, 0, b.batchSize)
 	Loop:
 		for len(batch) < b.batchSize {
-			t, ok := <-b.input
-			if !ok {
-				exit = true // input channel was closed by Exit()
+			select {
+			case t, ok := <-b.input:
+				if !ok {
+					exit = true // input channel was closed by Exit()
+					break Loop
+				}
+				batch = append(batch, t)
+			case <-b.truncate:
 				break Loop
 			}
-			batch = append(batch, t)
 		}
 
 		b.batchReady <- batch
--- a/lfs/batcher.go
+++ b/lfs/batcher.go
@@ -31,8 +31,8 @@ func NewBatcher(batchSize int) *Batcher {
 	return b
 }
 
-// Add adds an item (or many items) to the batcher. Add is safe to call from
-// multiple goroutines.
+// Add adds one or more items to the batcher. Add is safe to call from multiple
+// goroutines.
 func (b *Batcher) Add(ts ...interface{}) {
 	if atomic.CompareAndSwapUint32(&b.exited, 1, 0) {
 		b.input = make(chan interface{})
--- a/lfs/batcher.go
+++ b/lfs/batcher.go
@@ -1,6 +1,9 @@
 package lfs
 
-import "sync/atomic"
+import (
+	"sync"
+	"sync/atomic"
+)
 
 // Batcher provides a way to process a set of items in groups of n. Items can
 // be added to the batcher from multiple goroutines and pulled off in groups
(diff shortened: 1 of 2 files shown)


lfs/batcher.go
```diff
@@ -6,36 +6,43 @@ import "sync/atomic"
...
 	}
 
 	go b.acceptInput()
 	return b
 }
 
-// Add adds an item to the batcher. Add is safe to call from multiple
-// goroutines.
-func (b *Batcher) Add(t interface{}) {
+// Add adds an item (or many items) to the batcher. Add is safe to call from
```
technoweenie: How about "Add sends one or more items to the batcher"?

ttaylorr: Way clearer! Changed in https://github.com/github/git-lfs/pull/1528/commits/49489c0f43011c5f0638eb2fd17b993e7488a0a6.

lfs/batcher.go
```diff
@@ -6,36 +6,43 @@ import "sync/atomic"
...
 	batchReady chan []interface{}
+	flush      chan interface{}
 }
 
 // NewBatcher creates a Batcher with the batchSize.
 func NewBatcher(batchSize int) *Batcher {
 	b := &Batcher{
 		batchSize:  batchSize,
-		input:      make(chan interface{}, batchSize),
+		input:      make(chan interface{}),
```
sinbad: Making input a single length channel is going to force adding of more than one item to block until `acceptInput()` pulls an item, meaning these goroutines are now forced to work in lock step. This seems a bit restrictive?

ttaylorr: I think this is an interesting thought. The reason I un-buffered the channel in https://github.com/github/git-lfs/pull/1528/commits/98873d87c4b06beecbba57221e70dd975ce7d7d4 and https://github.com/github/git-lfs/pull/1528/commits/a27d92d1a48f5d6d57218501b60f00ebbbf75826 was so that I could ensure that after calling `Add()` a number of times in test, that they would _for sure_ be in the batch so that after `Flush()`-ing, I'd get them out.

However, I'm not actually sure we ever needed the buffered behavior in the first place. I could be wrong, since I haven't been in this part of the code in a while, but I ran some benchmarks to get a second opinion.

With buffering (on master):
```
~/g/g/lfs (master) $ g rev-parse HEAD
b7b1f2b7985c8f7a18c563c70898f669ffb8c13c

~/g/g/lfs (master!) $ go test -bench=.
BenchmarkLsTreeParser-4           500000              4087 ns/op
PASS
ok      github.com/github/git-lfs/lfs   2.889s
```

Without buffering (on 49489c0f43011c5f0638eb2fd17b993e7488a0a6):
```
~/g/g/lfs (batcher-truncate!) $ g rev-parse HEAD
49489c0f43011c5f0638eb2fd17b993e7488a0a6

~/g/g/lfs (batcher-truncate!) $ go test -bench=.
BenchmarkLsTreeParser-4           500000              4096 ns/op
PASS
ok      github.com/github/git-lfs/lfs   2.877s
```

I ran these benchmarks a number of times locally to minimize any jitter, but the numbers stayed remarkably consistent.

If we were to buffer the channel again, I'm thinking we could either:
- Upon receiving a signal from `<-b.flush`, enter a spin-loop until we have drained the channel (which has the 👎  side of thrashing the user's CPU)
- Increment a counter of "number of times flushed", and check after either a `close(b.input)` or some other signal, and preform the same drain as above.

But, given the added complexity of the ideas I'm thinking about above, and the minimal difference in performance, I think we'd be fine to keep this as-is.

For reference, the benchmark I ran is in [this gist]().

ttaylorr: I think this is an interesting thought. The reason I un-buffered the channel in https://github.com/github/git-lfs/pull/1528/commits/98873d87c4b06beecbba57221e70dd975ce7d7d4 and https://github.com/github/git-lfs/pull/1528/commits/a27d92d1a48f5d6d57218501b60f00ebbbf75826 was so that I could ensure that after calling `Add()` a number of times in test, that they would _for sure_ be in the batch so that after `Flush()`-ing, I'd get them out.

However, I'm not actually sure we ever needed the buffered behavior in the first place. I could be wrong, since I haven't been in this part of the code in a while, but I ran some benchmarks to get a second opinion.

With buffering (on master):
```
~/g/g/lfs (master!) $ git rev-parse HEAD
b7b1f2b7985c8f7a18c563c70898f669ffb8c13c

~/g/g/lfs (master!) $ go test -run=BenchmarkAddingToBatcher -bench=./lfs
BenchmarkAddingToBatcher-4      10000000               151 ns/op
BenchmarkLsTreeParser-4           500000              4288 ns/op   # <- ignore
PASS
ok      github.com/github/git-lfs/lfs   3.880s
```

Without buffering (on 49489c0f43011c5f0638eb2fd17b993e7488a0a6):
```
~/g/g/lfs (batcher-truncate!) $ git rev-parse HEAD                                
49489c0f43011c5f0638eb2fd17b993e7488a0a6

~/g/g/lfs (batcher-truncate!) $ go test -run=BenchmarkAddingToBatcher -bench=./lfs
BenchmarkAddingToBatcher-4       3000000               551 ns/op
BenchmarkLsTreeParser-4           500000              4245 ns/op   # <- ignore
PASS
ok      github.com/github/git-lfs/lfs   4.422s
```

My initial thought after seeing these numbers was to buffer the channel. For future reference, one of the ways we could go about doing this is:

Upon receiving a `<-b.flush` signal, spin until we have completely drained the input channel:
```go
case <-b.input:
for {
        select {
                case t, ok := <-b.input:
                // Do the thing that we're doing above :-(
                default:
                // Channel is empty!
                break Acc
        }
}
```

However, @rubyist made a great point in that given that we're not adding to this batcher from multiple goroutines currently in LFS, having this sort of contention isn't a big deal. He also pointed out that since we're using this code to do network transfers, an extra 400ns/op isn't a huge deal either.

I think given our current use-case, and the additional complexity necessary to buffer the input channel, we should leave this as-is until we need to change it.

--

For reference, the benchmark I ran is in [this gist](https://gist.github.com/ttaylorr/dabc204dce7f0f808a1f69d9857229cd).

sinbad: Yeah right now, we drive the batcher sequentially and don't try to parallelise the API requests so this probably isn't an issue, just wanted to point out that it is more restrictive than before (it is now more explicitly a single in, single out style). If @rubyist is cool with that then so am I.

lfs/batcher.go
```diff
@@ -44,29 +51,43 @@ func (b *Batcher) Next() []interface{} {
...
 func (b *Batcher) Exit() {
 	atomic.StoreUint32(&b.exited, 1)
 	close(b.input)
+	close(b.flush)
 }
 
 // acceptInput runs in its own goroutine and accepts input from external
-// clients. It fills and dispenses batches in a sequential order: for a batch
-// size N, N items will be processed before a new batch is ready.
+// clients. Without truncation, it fills and dispenses batches in a sequential
```
technoweenie: This message still mentions truncation. Can you rewrite it in a way so it calls it "flushing" somehow?

ttaylorr: https://github.com/github/git-lfs/pull/1528/commits/9f9b53266140154872795d0da202702a263728e6

lfs/batcher.go
technoweenie: Here's what he said in internal chat:

> yeah it looks like it's all done sequentially. if it's cleaner to not buffer, i say do it, 400ns is noise when we're talking network transfers

I agree with the assessment here. Removing buffering lets us keep this to a simple select loop without a noticeable performance impact.

PR #1535, opened 2016-09-21
lfs/transfer_queue: support multiple retries per object

ttaylorr:
This pull-request teaches the `lfs/transfer_queue` how to retry an object more than once per transfer.

### Previously
The transfer queue used to follow these steps when executing a transfer:

0. Receive a bunch of objects from calls to [`Add()`](https://github.com/github/git-lfs/blob/878165d746549404cf1ddbac0db58573904abbe0/lfs/transfer_queue.go#L84-L97)
0. Start either the [`batchApiRoutine`](https://github.com/github/git-lfs/blob/878165d746549404cf1ddbac0db58573904abbe0/lfs/transfer_queue.go#L330-L408) or [`individualApiRoutine`](https://github.com/github/git-lfs/blob/878165d746549404cf1ddbac0db58573904abbe0/lfs/transfer_queue.go#L266-L303) based on [its capabilities](https://github.com/github/git-lfs/blob/878165d746549404cf1ddbac0db58573904abbe0/lfs/transfer_queue.go#L442-L457).
0. Attempt to transfer the object(s), and commit to a [`retry`](https://github.com/github/git-lfs/blob/878165d746549404cf1ddbac0db58573904abbe0/lfs/transfer_queue.go#L459-L461) if it failed with a [retriable error](https://github.com/github/git-lfs/blob/878165d746549404cf1ddbac0db58573904abbe0/lfs/transfer_queue.go#L463-L469).
  0. Collect the transfers to be retried in the [`retryCollector`](https://github.com/github/git-lfs/blob/878165d746549404cf1ddbac0db58573904abbe0/lfs/transfer_queue.go#L418-L423)
  0. [Close the transfer channel to kick off all of the objects to be retried.](https://github.com/github/git-lfs/blob/878165d746549404cf1ddbac0db58573904abbe0/lfs/transfer_queue.go#L228-L231)
  0. [Re-add all of the objects to be retried](https://github.com/github/git-lfs/blob/878165d746549404cf1ddbac0db58573904abbe0/lfs/transfer_queue.go#L233-L242) and re-process them _once_.
0. Exit.

This works fine, but limits the transfer queue to:
* Responding to one set of retries per transfer.
* Responding to one retry per unique OID in the queue.

### Now
Now, the transfer queue knows how to immediately retry an object. It works like this:

0. Steps 1-4 from above.
0. If an object failed, and [can be retried](https://github.com/github/git-lfs/blob/8d6102854fe14be05602ebe7fe213518685b30ed/lfs/transfer_queue.go#L492-L512), [enqueue a retry](https://github.com/github/git-lfs/blob/8d6102854fe14be05602ebe7fe213518685b30ed/lfs/transfer_queue.go#L488-L490).
0. Collect the retry, and...
  0. Make sure that we [have the budget](https://github.com/github/git-lfs/blob/8d6102854fe14be05602ebe7fe213518685b30ed/lfs/transfer_queue.go#L438-L441) to retry the object.
  0. [`Add()`](https://github.com/github/git-lfs/blob/8d6102854fe14be05602ebe7fe213518685b30ed/lfs/transfer_queue.go#L445) it to the next batch, or API channel (in legacy mode)
  0. If in batch mode, [immediately flush the batch](https://github.com/github/git-lfs/blob/8d6102854fe14be05602ebe7fe213518685b30ed/lfs/transfer_queue.go#L446-L449), forcing the `batchApiRoutine` function to [receive a new batch](https://github.com/github/git-lfs/blob/8d6102854fe14be05602ebe7fe213518685b30ed/lfs/transfer_queue.go#L348).
0. [Wait](https://github.com/github/git-lfs/blob/8d6102854fe14be05602ebe7fe213518685b30ed/lfs/transfer_queue.go#L249) until all objects have been either transferred, or abandoned.

(The ability to flush a batch was introduced in #1528, and enables the ability to immediately retry an object in a batch.) 

In order to make sure that all items are processed before exiting, the way we treat the internal waitgroup `q.wait` changed slightly. Previously, `q.wait` was incremented every time we try to preform a transfer on an item. To prevent a situation where the `WaitGroup` would reach zero while waiting in between failing an object and retrying it, the `WaitGroup` is now [only incremented the _first_ time](https://github.com/github/git-lfs/blob/8d6102854fe14be05602ebe7fe213518685b30ed/lfs/transfer_queue.go#L95-L100) an object begins a transfer.

The transferqueue also now keeps track of the number of retries made per OID in order to prevent infinitely retrying that transfer. Currently, the maximum number of retries per object is set to 1, in order to keep the same behavior as from before this PR which is why there are no new tests.

I've also left some line comments throughout the code to clarify some further things.

--

/cc @technoweenie @rubyist @sinbad for 👀 and 💭

### DIFF
--- a/lfs/transfer_queue.go
+++ b/lfs/transfer_queue.go
@@ -361,7 +361,6 @@ func (q *TransferQueue) batchApiRoutine() {
 		}
 
 		objs, adapterName, err := api.Batch(config.Config, transfers, q.transferKind(), transferAdapterNames)
-		tracerx.Printf("_> resp %v %v %v", objs, adapterName, err)
 		if err != nil {
 			if errors.IsNotImplementedError(err) {
 				git.Config.SetLocal("", "lfs.batch", "false")
--- a/lfs/transfer_queue.go
+++ b/lfs/transfer_queue.go
@@ -2,7 +2,6 @@ package lfs
 
 import (
 	"sync"
-	"sync/atomic"
 
 	"github.com/github/git-lfs/api"
 	"github.com/github/git-lfs/config"
@@ -14,7 +13,8 @@ import (
 )
 
 const (
-	batchSize = 100
+	batchSize         = 100
+	defaultMaxRetries = 1
 )
 
 type Transferable interface {
@@ -30,7 +30,7 @@ type Transferable interface {
 
 // TransferQueue organises the wider process of uploading and downloading,
 // including calling the API, passing the actual transfer request to transfer
-// adapters, and dealing with progress, errors and retries
+// adapters, and dealing with progress, errors and retries.
 type TransferQueue struct {
 	direction         transfer.Direction
 	adapter           transfer.TransferAdapter
@@ -38,11 +38,9 @@ type TransferQueue struct {
 	adapterResultChan chan transfer.TransferResult
 	adapterInitMutex  sync.Mutex
 	dryRun            bool
-	retrying          uint32
 	meter             *progress.ProgressMeter
 	errors            []error
 	transferables     map[string]Transferable
-	retries           []Transferable
 	batcher           *Batcher
 	apic              chan Transferable // Channel for processing individual API requests
 	retriesc          chan Transferable // Channel for processing retries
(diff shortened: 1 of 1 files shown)


lfs/transfer_queue.go
```diff
@@ -354,23 +361,26 @@ func (q *TransferQueue) batchApiRoutine() {
...
-
 				go q.legacyFallback(batch)
 				return
 			}
 
-			if q.canRetry(err) {
-				for _, t := range batch {
-					q.retry(t.(Transferable))
+			var errOnce sync.Once
+			for _, o := range batch {
```
ttaylorr: If eventually we end up grouping retried objects and "fresh" objects in the same batch, we should only fail objects that have exceeded their retry budget, and not all of the items in the batch at once. To keep the old behavior of reporting to the errorc channel the same, I wrapped it in a sync.Once, so it can only happen once.

PR #1540, opened 2016-09-26
Status initial commit

### LINKED ISSUE #1536: `git lfs status` crashes on new repos
### Steps to reproduce

```bash
$ git init .
$ git lfs status
Could not get the current ref

Errors logged to /path/to/repo/.git/lfs/objects/logs/20160926T120454.38945701.log
```

### Error log
```
git-lfs/1.4.1 (GitHub; darwin amd64; go 1.7)
git version 2.10.0

$ git-lfs status
Could not get the current ref

Git can't resolve ref: "HEAD"
```

sinbad:
Fixes #1536 

Also makes added files appear in the staged area before the first commit.

### DIFF
--- a/lfs/scanner.go
+++ b/lfs/scanner.go
@@ -30,6 +30,9 @@ const (
 	// chanBufSize is the size of the channels used to pass data from one
 	// sub-process to another.
 	chanBufSize = 100
+
+	// Equivalent to git mktree < /dev/null, useful for diffing before first commit
+	nullTreeish = "4b825dc642cb6eb9a060e54bf8d69288fbee4904"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


lfs/scanner.go
```diff
@@ -192,7 +195,8 @@ func (m *indexFileMap) Set(sha string, index *indexFile) {
 // ScanIndex returns a slice of WrappedPointer objects for all
 // Git LFS pointers it finds in the index.
 // Reports unique oids once only, not multiple times if >1 file uses the same content
-func ScanIndex() ([]*WrappedPointer, error) {
+// isFirstCommit must be true if this repo doesn't have any commits yet
+func ScanIndex(isFirstCommit bool) ([]*WrappedPointer, error) {
```
ttaylorr: Nice. What do you think about changing this to parameter to instead be the first ref to begin with? I think that might make the behavior of the function more obvious to the caller (instead of having to look and see what `true`/`false` means).

The `nullTreeish` constant makes this a little tricky to deal with, so maybe it could be called `InitialCommit` or something similar?

`ScanIndex("HEAD")`, `ScanIndex(InitialCommit)`.

technoweenie: I like this idea. Could even add a global like `HeadRef` or something too.

sinbad: I wanted callers to ScanIndex not to have to know exactly what they needed to do to support pre-initial commit but you're right that when reading the code afterwards just seeing `false` is less useful than seeing `InitialCommit` so I'll do that. 

`HeadRef` would be good in theory but because that would introduce an extra `git` call just to derive `HEAD` in 99% of cases, and when most of the time it actually makes sense to abort before the first commit - `status` is one of the few where it makes sense to do something useful - I won't do this for now. If you can think of some other cases where it's useful to get the null tree automatically we can add it?

PR #1541, opened 2016-09-26
add test to replicate `track` case insensitivity bug

technoweenie:
Wrote a test attempting to replicate https://github.com/github/git-lfs/issues/1537.


ttaylorr:
Looks cool. (Acking out of our first responder rotation).

PR #1545, opened 2016-09-28
lfs/tq: extract `RetryCounter`

ttaylorr:
This pull-request introduces the extracted `RetryCounter` type, which serves two purposes:
- Allows configuration via the `lfs.transfer.maxretries` `.gitconfig` option
- Serves as a base on which to implement retry budgets

📝  I left a few notes throughout the code to point out some choices I made, but this should be a pretty minor change regardless.

I want to investigate a higher-level integration test for this behavior, but I'm not sure what it will look like after retry budgets are implemented. As a first pass, I'm thinking that we could encode "number of retries before success" on the server side and assert that we retried as many times as we could on the client side.

:wave: thanks in advance for the review.

---

/cc @technoweenie @rubyist @sinbad for 👀

### DIFF
--- a/lfs/transfer_queue.go
+++ b/lfs/transfer_queue.go
@@ -28,6 +28,68 @@ type Transferable interface {
 	LegacyCheck() (*api.ObjectResource, error)
 }
 
+type retryCounter struct {
+	// MaxRetries is the maximum number of retries a single object can
+	// attempt to make before it will be dropped.
+	MaxRetries int `git:"lfs.transfer.maxretries"`
+
+	// cmu guards count
+	cmu sync.Mutex
+	// count maps OIDs to number of retry attempts
+	count map[string]int
+}
+
+// newRetryCounter instantiates a new *retryCounter. It parses the gitconfig
+// value: `lfs.transfer.maxretries`, and falls back to defaultMaxRetries if none
+// was provided.
+//
+// If it encountered an error in Unmarshaling the *config.Configuration, it will
+// be returned, otherwise nil.
+func newRetryCounter(cfg *config.Configuration) *retryCounter {
+	rc := &retryCounter{
+		MaxRetries: defaultMaxRetries,
+
+		count: make(map[string]int),
+	}
+
+	if err := cfg.Unmarshal(rc); err != nil {
+		tracerx.Printf("rc: error parsing config, falling back to default values...: %v", err)
+		rc.MaxRetries = 1
+	}
+
+	if rc.MaxRetries < 1 {
+		tracerx.Printf("rc: invalid retry count: %d, defaulting to %d", rc.MaxRetries, 1)
+		rc.MaxRetries = 1
+	}
+
+	return rc
+}
+
+// Increment increments the number of retries for a given OID. It is safe to
+// call across multiple goroutines.
+func (r *retryCounter) Increment(oid string) {
+	r.cmu.Lock()
+	defer r.cmu.Unlock()
+
+	r.count[oid]++
+}
+
+// CountFor returns the current number of retries for a given OID. It is safe to
+// call across multiple goroutines.
+func (r *retryCounter) CountFor(oid string) int {
+	r.cmu.Lock()
+	defer r.cmu.Unlock()
+
+	return r.count[oid]
+}
+
+// CanRetry returns the current number of retries, and whether or not it exceeds
+// the maximum number of retries (see: retryCounter.MaxRetries).
+func (r *retryCounter) CanRetry(oid string) (int, bool) {
+	count := r.CountFor(oid)
+	return count, count < r.MaxRetries
+}
+
 // TransferQueue organises the wider process of uploading and downloading,
 // including calling the API, passing the actual transfer request to transfer
 // adapters, and dealing with progress, errors and retries.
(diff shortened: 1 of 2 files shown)


lfs/transfer_queue.go
```diff
@@ -55,16 +117,14 @@ type TransferQueue struct {
...
-	// maxRetries is the maximum number of retries a single object can
-	// attempt to make before it will be dropped.
-	maxRetries uint32
+	rc            *retryCounter
 }
 
 // newTransferQueue builds a TransferQueue, direction and underlying mechanism determined by adapter
 func newTransferQueue(files int, size int64, dryRun bool, dir transfer.Direction) *TransferQueue {
-	logPath, _ := config.Config.Os.Get("GIT_LFS_PROGRESS")
+	cfg := config.Config
```
ttaylorr: This is a pattern that we have in a few places. Instead of wiring up sending a `*config.Configuration` through all of `newTransferQueue`
's call-sites, leave a tiny reminder that the value is interchangeable.

lfs/transfer_queue.go
```diff
@@ -432,10 +491,8 @@ func (q *TransferQueue) errorCollector() {
 // retryCollector runs in its own goroutine.
 func (q *TransferQueue) retryCollector() {
 	for t := range q.retriesc {
-		q.rmu.Lock()
-		q.retryCount[t.Oid()]++
-		count := q.retryCount[t.Oid()]
-		q.rmu.Unlock()
+		q.rc.Increment(t.Oid())
+		count := q.rc.CountFor(t.Oid())
```
ttaylorr: The old behavior kept the same lock for both operations: add and retrieve. This isn't necessary, since only `retryCollector` is running at a time.

lfs/transfer_queue.go
```diff
@@ -28,6 +28,68 @@ type Transferable interface {
...
+// be returned, otherwise nil.
+func newRetryCounter(cfg *config.Configuration) *retryCounter {
+	rc := &retryCounter{
+		MaxRetries: defaultMaxRetries,
+
+		count: make(map[string]int),
+	}
+
+	if err := cfg.Unmarshal(rc); err != nil {
+		tracerx.Printf("rc: error parsing config, falling back to default values...: %v", err)
```
ttaylorr: This should never fail. If the value stored in the `.gitconfig` isn't parsable as an `int`, we'll just [fall back to the default](https://github.com/github/git-lfs/blob/06b4c1463df60e4c480ff6d9c14fd81fcada3c17/config/config.go#L165), the current value (which happens to be `defaultMaxRetries`).

lfs/transfer_queue.go
```diff
@@ -28,6 +28,68 @@ type Transferable interface {
...
+		rc.MaxRetries = 1
+	}
+
+	if rc.MaxRetries < 1 {
+		tracerx.Printf("rc: invalid retry count: %d, defaulting to %d", rc.MaxRetries, 1)
```
ttaylorr: Here's another note in the same vein. Quicker than wiring some logging up, we can just dump this straight into the trace, and re-run the command with `GIT_LFS_TRACE=1` if there's a problem.

lfs/transfer_queue.go
```diff
@@ -28,6 +28,68 @@ type Transferable interface {
 	LegacyCheck() (*api.ObjectResource, error)
 }
 
+type retryCounter struct {
+	// MaxRetries is the maximum number of retries a single object can
+	// attempt to make before it will be dropped.
+	MaxRetries int `git:"lfs.transfer.maxretries"`
```
sinbad: I still love this annotation config feature BTW 😄

lfs/transfer_queue.go
sinbad: 👍 in an evaluation of "what's the worst that could happen" it makes total sense just to log & keep going in this case, panicking would not help anyone.

ttaylorr:
> I think you already mentioned previously that you'd look at batching this instead later, which would make more sense I think

This makes sense to me, and was exactly what I was considering for next-steps on this project.

Thanks for the review! ❤️

ttaylorr:
Here are some post-merge numbers for your 👀 :

Using my LFS benchmark tool, and this configuration:

```go
func BenchmarkMultipleRetries(b *bench.B) {                                    
        b.FailureRate(.25)                                             
        b.MaxRetryCount(1)                                            
                                                                       
        b.InstrumentUntilSuccessful("git lfs push --all origin master")
}                                                                      
```

A number of trials had to run more than once in order to successfully push all objects:

```sh
~/g/bench (next!) $ go build *.go && sudo ./main 2>&1 | xargs
1 1 1 3 1 5 2 1
```

Using the same `Benchmark` function, with the following diff applied:

```diff
diff --git a/main.go b/main.go
index f89edfb..276bbc8 100644
--- a/main.go
+++ b/main.go
@@ -8,7 +8,7 @@ import (
 
 func BenchmarkMultipleRetries(b *bench.B) {
        b.FailureRate(.25)
-       b.MaxRetryCount(1)
+       b.MaxRetryCount(10)
 
        b.InstrumentUntilSuccessful("git lfs push --all origin master")
 }
```

All trials of the command only took one invocation to transfer all objects successfully 😍 👍 

```sh
~/g/bench (next!) $ go build *.go && sudo ./main 2>&1 | xargs
1 1 1 1 1 1 1 1
```

PR #1557, opened 2016-10-05
Travis: Only install git if not installed yet

sschuberth:
It seems some osx10.11 Travis workers already have git installed via
homebrew, resulting in

    Error: git-2.10.0 already installed

    To install this version, first `brew unlink git`

So only install git if not installed yet, and upgrade it otherwise.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ matrix:
       before_script:
         - >
           brew update;
-          brew install git;
+          brew ls --versions git && brew upgrade git || brew install git;
     - env: git-latest-go-1.5
       os: linux
       go: 1.5
@@ -58,7 +58,7 @@ matrix:
       before_script:
         - >
           brew update;
-          brew install git;
+          brew ls --versions git && brew upgrade git || brew install git;
 
 before_install:
   - >


sschuberth:
One example of a job failing with the above error is https://travis-ci.org/sschuberth/git-lfs/jobs/165230172.

sschuberth:
Also, I'm not super-happy about the way I had to implement the check, it seems very laborious, but I was unable to find a better solution so far. Suggestions welcome :-)

ttaylorr:
Thanks for opening this. I'm wondering if `brew install` is actually the thing that we want to do. A `brew install git` will always give the latest version, right?

sschuberth:
Right. Provided that you did `brew update` before, obviously. Why would we not want to do `brew install git`? For workers where git is not installed (via `brew`) we need this to ensure to run the latest version of git, right?

ttaylorr:
> For workers where git is not installed (via brew) we need this to ensure to run the latest version of git, right?

Not always, no. Homebrew isn't always the installation path for Git, since some of our builds use different branches of Git, or install it directly from source.

screenshot 2016-10-06 13 20 11

ttaylorr:
Ah, sorry -- this is correct. In the compressed diff, I didn't see that this was only applied to workers that ran the latest git. I'm 👍  with this, thanks!

Issue #1546, opened 2016-09-29
`git lfs track` does not track existing files in sub directories


larsxschneider:
### Setup

```
git-lfs/1.4.1 (GitHub; darwin amd64; go 1.7)
git version 2.10.0
macOS 10.11.6
```

### Steps to reproduce

```bash
# Create the repo and add some files to plain Git
$ git init .
$ mkdir subdir
$ echo "content1" >root.dat
$ echo "content2" >subdir/sub.dat
$ git add .
$ sleep 1   # <-- sleep is only required if you run this test as script!
$ git commit -m "add content to Git"

# Move files to Git LFS
$ git lfs track "/root.dat"
$ git lfs track "/subdir/sub.dat"
$ git add .
$ git commit -m "move content to LFS"

# Check results
$ git cat-file -p :root.dat
version https://git-lfs.github.com/spec/v1
oid sha256:47d741b6059c6d7e99be25ce46fb9ba099cfd6515de1ef7681f93479d25996a4
size 9

$ git cat-file -p :subdir/sub.dat
content2
```

`git lfs track` did not track the existing file in the sub directory. This seems to happen to all files that are located in sub directories and tracked individually with a leading slash (e.g. `/subdir/sub.dat`).

ttaylorr:
Thanks for opening this! I'll put this on my to-do list.

ttaylorr:
Fixed in #1563.

Issue #1531, opened 2016-09-21
tracking of files anchored at the root of the repository fails on Windows


larsxschneider:
### Setup

```
git-lfs/1.4.1 (GitHub; windows amd64; go 1.7; git 2070e4c)
git version 2.9.2.windows.1 
Win 7 & Win 8.1
```

### Steps to reproduce

```bash
MINGW64 ~/my-repo (master)
$ echo "test" > test1.dat

MINGW64 ~/my-repo (master)
$ git lfs track /test1.dat
Error getting git tracked files

Errors logged to ....
Use `git lfs logs last` to view the log.
```

Tracking of files anchored at the root of the repository fails on Windows in Git Bash.

### Log

```
git-lfs/1.4.1 (GitHub; windows amd64; go 1.7; git 2070e4c)
git version 2.9.2.windows.1

$ git-lfs track C:/path/to/my/Git/test1.dat
Error getting git tracked files

exit status 128
```

technoweenie:
Thanks for the report. Added to a v1.5.0 milestone. If @ttaylorr or someone else can submit a patch, I can see getting this into a v1.4.x release, but I personally don't have the time now.

ttaylorr:
Thanks for the ping! I can take a look at this after I finish some work I'm doing on the transferqueue.

ttaylorr:
Sorry for the late response! I just finished digging into this, here's what I found.

When you type `/a.dat` at the Windows shell, it turns that into `C:/Program Files/Git/a.dat`. This is usually outside of the Git repository (unless for some reason you're working in `C:/Program Files/Git`) so running `git ls-files` fails.

I don't think it's worth trying to change or guess the path that Windows is prepending, since this will break the expectations of Windows scripts. I do think, however, that we should improve the error message we're getting from `git lfs track` to include the error that we failed with when running the `ls-files` command.

Does this sound like a good approach?

larsxschneider:
Hm... on Linux/OSX `/something` refers to something directly at the root of the file system. However, Git interprets that as "something at the root of the Git repository". I believe that is similar to what you mean with "break the expectations of Windows scripts", right?

When is `/a.dat` turned into `C:/Program Files/Git/a.dat`? I assume you see the latter as parameter in Go? What if we put parenthesis around the parameter (`git lfs track "/a.dat"`)?

ttaylorr:
[...]

PR #1563, opened 2016-10-07
git: ignore non-root wildcards

ttaylorr:
This pull-request fixes a mismatch between LFS's file-matching algorithm and the one used by `.gitattributes`.

Previously, all patterns that began with a leading `/` and matching files in subdirectories were ignored (i.e., `/a/dir/b.dat`). This is incorrect, since Git actually _will_ match these files, unless they contain a wildcard pattern (i.e., `/*.dat`).

This pull-request fixes that by only allow patterns with a leading slash that match files by name, thereby not using wildcard characters.

--

/cc @technoweenie @larsxschneider #1546.


technoweenie:
Same bug as #1531. Is there a reason you want to reference paths with a leading slash? I'm not sure this is something we can fix. I don't see anything in the env that would let git-lfs identify that the `C:/Program Files/Git/` (yes, _forward_ slashes. weird) prefix can be stripped without hardcoding that value.

ttaylorr:
Ha, looks good to me!

commands/path_windows.go
```diff
@@ -0,0 +1,59 @@
...
+	defer winBashMu.Unlock()
+
+	// check if path starts with windows drive letter
+	if !winPathHasDrive(pattern) {
+		return pattern
+	}
+
+	if len(winBashPrefix) < 1 {
+		// cmd.Path is something like C:\Program Files\Git\usr\bin\pwd.exe
+		cmd := subprocess.ExecCommand("pwd")
```
sschuberth: Instead of doing this, one thought that I had was to read the `Inno Setup: App Path` key's value from the Windows registry inside `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Git_is1`. That would return e.g. `C:\Program Files\Git`. Not sure if that's much more elegant, though.

PR #1565, opened 2016-10-07
use filepath.Clean() when comparing filenames to include/exclude patterns

technoweenie:
This removes the windows-specific hacks, and uses `filepath.Clean()` to sanitize the paths. This broke previously because it was `filepath.Match()` expects back slashes on windows.

### DIFF
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -76,6 +76,7 @@ func fetchCommand(cmd *cobra.Command, args []string) {
 
 	} else { // !all
 		includePaths, excludePaths := determineIncludeExcludePaths(cfg, include, exclude)
+		fmt.Println("FETCH INCLUDE", includePaths)
 
 		// Fetch refs sequentially per arg order; duplicates in later refs will be ignored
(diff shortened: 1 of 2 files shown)


sschuberth:
@technoweenie Thanks for this. Out of curiosity, are you running on Windows 10? Because on Windows 10 I see more test failures than on Windows 8.1. These additional failures seem to be related to Windows short vs. long path issues where Windows 10 seems to behave differently, and I'm trying to address them as part of my [windows-tests](https://github.com/sschuberth/git-lfs/commits/windows-tests) branch.

sinbad:
👍 Looks like the main issue here was people using '\' separators in their include/exclude config, which I don't usually advocate (all other git settings tend to work better using '/' even on Windows), but this is good belt & braces. The tests didn't used to fail on Windows for that, maybe something else has been changed in the meantime.

technoweenie:
> Out of curiosity, are you running on Windows 10? Because on Windows 10 I see more test failures than on Windows 8.1.

I ran them on my Windows 10 gaming pc :)

PR #1571, opened 2016-10-10
lfs/tq: send error before terminating object

ttaylorr:
Previously, decrementing the surrounding `sync.WaitGroup` allowed the
`q.errorc` channel to be closed before this send could occur. This racy
behavior introduced a send-on-closed-channel panic, which is fixed in this
commit.

--

/cc @technoweenie @sinbad @rubyist


sinbad:
Nice.

PR #1572, opened 2016-10-11
script/windows: remove WizardSmallImageFile

ttaylorr:
The old `WizardSmallImageFile` compiles successfully under InnoSetup v5.5.9(a), but fails at runtime with an error saying, "Bitmap image invalid".

I re-exported the image on macOS using 24bit colorspace with RGB (no alpha) and 8-bits per channel. This causes the installer to build and run successfully, but the image is invisible in the installer.

--

/cc @technoweenie @sschuberth


sschuberth:
I'd like to investigate this a bit more before this gets merged. @ttaylorr I remember you having some problems with the wizard image already before, but as you can see from [this installer](https://ci.appveyor.com/api/buildjobs/h24avlfj8nld7fhu/artifacts/git-lfs-windows-1.4.2.exe) created on [AppVeyor CI](https://ci.appveyor.com/project/git-lfs/git-lfs/build/1.0.136/artifacts) the wizard image is fine. What is the machine you're creating the official Windows release on, is that your VM? Why not simply use the AppVeyor CI created installer artifact instead?

ttaylorr:
@sschuberth the installer was created on a Windows 10 VM, and was reproduced by @technoweenie on his Windows 7 (?) VM. I don't mind waiting an hour or two to fix this, but my priority is getting this released today.

I'm going to defer to @technoweenie's judgement on whether or not we should wait on merging this or not.

sschuberth:
> compiles successfully under InnoSetup v5.5.9(a)

I guess the "(a)" means you're using the non-Unicode (Ascii) version of Inno Setup? I'm using the Unicode version on AppVeyor, and also locally on my machine. Would you mind also trying to use the Unicode version of Inno Setup to see if it solves the problem for some reason?

ttaylorr:
BOOM! Great call. I uninstalled whatever version of InnoSetup I had installed and used the Unicode one, and it worked. Proceeding with the release!

sschuberth:
FYI, I'll follow up on the Inno Setup newsgroup about this error.

technoweenie:
Thanks for your help on this @sschuberth :metal:

PR #1575, opened 2016-10-12
Also add a large wizard image to the Windows installer

sschuberth:
And add compatibility with the ANSI version of Inno Setup with has issues with decoding 16-bits-per-pixel BMP files.


ttaylorr:
This looks awesome! I went ahead and built the installer on my Windows VM, and it ran as expected. For reference:

screenshot 2016-10-12 13 32 29

Thanks for giving the installer some ❤️ , @sschuberth!

Issue #1580, opened 2016-10-14
Versio 1.4.2 introduces "invalid memory address or nil pointer dereference" on fetch


sschuberth:
When doing `git lfs fetch` with version 1.4.2 on Windows x64 I get:

    Fetching master
    &{0xc042044120} WARNING: Git LFS is using a deprecated API, which will be removed in v1.5.0.
    &{0xc042044120}          Consider enabling the latest API by running: `git config lfs.batch true`.
    panic: runtime error: invalid memory address or nil pointer dereference
    [signal 0xc0000005 code=0x0 addr=0x0 pc=0x525a18]

    goroutine 41 [running]:
    panic(0x79e920, 0xc04200c0a0)
            c:/go/src/runtime/panic.go:500 +0x1af
    github.com/github/git-lfs/lfs.(*TransferQueue).individualApiRoutine(0xc042200000, 0xc0423738c0)
            c:/go/src/github.com/github/git-lfs/lfs/transfer_queue.go:341 +0xe8
    created by github.com/github/git-lfs/lfs.(*TransferQueue).launchIndividualApiRoutines.func1
            c:/go/src/github.com/github/git-lfs/lfs/transfer_queue.go:515 +0x72

with version 1.4.1 everything is fine. So my guess one of these introduced this:

    $ git log v1.4.2 --not v1.4.1  -- lfs/transfer_queue.go
    commit 6c0b9373ee7490ab0453e7f1fb65ea1743ec0a90
    Author: Taylor Blau <me@ttaylorr.com>
    Date:   Mon Oct 10 15:57:12 2016 -0600

        lfs/tq: send error before terminating object

        Previously, decrementing the surrounding `sync.WaitGroup` allowed the
        `q.errorc` channel to be closed before this send could occur. This racy
        behavior introduced a send-on-closed-channel panic, which is fixed in this
        commit.

    commit b5cba63f36656a75624b1cee4d1fdaf7525f4060
    Author: Taylor Blau <me@ttaylorr.com>
    Date:   Wed Sep 28 12:25:37 2016 -0600

        lfs/tq: extract `RetryCounter`

    commit 4b194496c15d584820ceaf40524f2492b53782da
    Author: Taylor Blau <me@ttaylorr.com>
    Date:   Wed Sep 21 17:38:32 2016 -0600

        lfs/transfer_queue: remove debug stmt

    commit 9b1bd0471c061c306c07f18ed48dd80e895f8d32
[...]

PR #1584, opened 2016-10-14
Test legacy check retries behavior

ttaylorr:
This pull-request fixes an issue described in https://github.com/github/git-lfs/issues/1580, where, when using the transferqueue with `lfs.batch=false`, a failed legacy check would result in a nil-pointer dereference `panic`.

To ensure further safety, I added two new `oidHandler`s to force a finite number of retries into both the upload/download check for legacy transfers. I will follow this up in my next pull-request with a similar solution for the storage endpoint so we can also test this behavior when using the batch API.

--

/cc @technoweenie @rubyist @sschuberth #1580

### DIFF
--- a/lfs/transfer_queue.go
+++ b/lfs/transfer_queue.go
@@ -338,7 +338,7 @@ func (q *TransferQueue) individualApiRoutine(apiWaiter chan interface{}) {
 	for t := range q.apic {
 		obj, err := t.LegacyCheck()
 		if err != nil {
-			if q.canRetryObject(obj.Oid, err) {
+			if q.canRetryObject(t.Oid(), err) {
 				q.retry(t)
 			} else {
 				q.errorc <- err
--- a/test/testhelpers.sh
+++ b/test/testhelpers.sh
@@ -70,7 +70,8 @@ refute_server_object() {
   curl -v "$GITSERVER/$reponame.git/info/lfs/objects/$oid" \
     -u "user:pass" \
     -o http.json \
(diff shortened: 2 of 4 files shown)


lfs/transfer_queue.go
```diff
@@ -338,7 +338,7 @@ func (q *TransferQueue) individualApiRoutine(apiWaiter chan interface{}) {
 	for t := range q.apic {
 		obj, err := t.LegacyCheck()
 		if err != nil {
-			if q.canRetryObject(obj.Oid, err) {
+			if q.canRetryObject(t.Oid(), err) {
```
ttaylorr: This is the actual bug.

test/testhelpers.sh
```diff
@@ -70,7 +70,8 @@ refute_server_object() {
   curl -v "$GITSERVER/$reponame.git/info/lfs/objects/$oid" \
     -u "user:pass" \
     -o http.json \
-    -H "Accept: application/vnd.git-lfs+json" 2>&1 |
+    -H "Accept: application/vnd.git-lfs+json" \
+    -H "X-Ignore-Retries: true" 2>&1 |
```
ttaylorr: `X-Ignore-Retries` is added as a header value here so as to not fail these `assert_server_object` when the objects that are being looked for have a failure count assosciated with them.

lfs/transfer_queue.go
sschuberth: Confirmed that this fixes the issue, and that the new `test-legacy-retries` would catch it if it still was present. Thanks!

PR #1590, opened 2016-10-17
WIP (!) Locking read-only behaviour

sinbad:
OK this branch has been running for too long and it's still not ready yet 😞 **Not ready for merging**, this is just for visibility/discussion.

The general thrust of this work is to try to expand on the locking feature to assist with workflow, to avoid accidents that require work to be lost later. This part of the work is about setting files to be read-only when they're marked as needing locking, and are not locked by the current user, as a kind of 'rumble strip' to hopefully stop you from accidentally doing work on a file before you realise you should have checked the locking situation. This has required the introduction of a bunch of new concepts:

* A custom attribute `lockable`, which can be enabled / disabled on a file pattern using `git lfs track --lockable`. Files flagged like this are assumed to need locking because they're unmergeable (some LFS files may actually be mergeable e.g. text data)
* `git lfs lock|unlock` now changes the write flag of these files
* `post-checkout` and `post-commit` hooks to make files read-only when they're marked as lockable and not currently locked by this user. `post-merge` is still TODO
* A local cache of locked file state to make it feasible to perform the above without hitting the server (uses [boltdb](https://github.com/boltdb/bolt))

Open questions / risk factors right now:

* Practical performance of keeping tabs on write flags in practice on large repos
* Gaps in available hook behaviour that may make write flag state not 100% watertight; known cases discussed in proposals/locking.md
* Edge cases / undefined behaviour with read-only working copy files

I think this work is going to be required to make locking workable in practice but it's not there yet. Feel free to comment but I'm not proposing we merge this yet.

### DIFF
--- a/commands/command_post_checkout.go
+++ b/commands/command_post_checkout.go
@@ -68,6 +68,5 @@ func postCheckoutFileChange() {
 }
 
 func init() {
-	RegisterCommand("post-checkout", postCheckoutCommand, func(cmd *cobra.Command) {
-	})
+	RegisterCommand("post-checkout", postCheckoutCommand, nil)
 }
--- a/commands/command_post_checkout.go
+++ b/commands/command_post_checkout.go
@@ -1,0 +1,67 @@
+package commands
+
+import (
+	"os"
+
+	"github.com/github/git-lfs/git"
+
+	"github.com/github/git-lfs/locking"
+
+	"github.com/spf13/cobra"
+)
+
+// postCheckoutCommand is run through Git's post-checkout hook. The hook passes
+// up to 3 arguments on the command line:
+//
+//   1. SHA of previous commit before the checkout
+//   2. SHA of commit just checked out
+//   3. Flag ("0" or "1") - 1 if a branch/tag/SHA was checked out, 0 if a file was
+//      In the case of a file being checked out, the pre/post SHA are the same
+//
+// This hook checks that files which are lockable and not locked are made read-only,
+// optimising that as best it can based on the available information.
+func postCheckoutCommand(cmd *cobra.Command, args []string) {
+	if len(args) != 3 {
+		Print("This should be run through Git's post-commit hook.  Run `git lfs update` to install it.")
+		os.Exit(1)
+	}
+
+	requireGitVersion()
+
+	if args[2] == "1" {
+		postCheckoutRevChange(args[0], args[1])
+	} else {
+		postCheckoutFileChange()
+	}
+
+}
+
+func postCheckoutRevChange(pre, post string) {
+	// We can speed things up by looking at the difference between previous HEAD
+	// and current HEAD, and only checking lockable files that are different
+	files, err := git.GetFilesChanged(pre, post)
+
+	if err != nil {
+		LoggedError(err, "Warning: post-checkout rev diff %v:%v failed: %v\nFalling back on full scan.", pre, post, err)
+		postCheckoutFileChange()
+	}
+	err = locking.FixLockableFileWriteFlags(files)
+	if err != nil {
+		LoggedError(err, "Warning: post-checkout locked file check failed: %v", err)
+	}
+
+}
+
+func postCheckoutFileChange() {
+	// Sadly we don't get any information about what files were checked out,
+	// so we have to check the entire repo
+	err := locking.FixAllLockableFileWriteFlags()
+	if err != nil {
+		LoggedError(err, "Warning: post-checkout locked file check failed: %v", err)
+	}
+}
+
+func init() {
+	RegisterCommand("post-checkout", postCheckoutCommand, func(cmd *cobra.Command) {
+	})
+}
--- a/commands/command_post_checkout.go
+++ b/commands/command_post_checkout.go
@@ -28,6 +28,11 @@ func postCheckoutCommand(cmd *cobra.Command, args []string) {
 
 	requireGitVersion()
 
+	// Skip this hook if no lockable patterns have been configured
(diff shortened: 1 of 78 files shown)


technoweenie:
Thanks for submitting this. I have a review in progress. Though my first question is: why bolt? That seems like a pretty heavy dependency for caching. It seems like we could get away with flat files:

```bash
# key is hashed path abcdef...
$ cat .git/lfs/lockcache/path/ab/cd/abcdef...
123

$ cat .git/lfs/lockcache/id/123
abcdef...
```

That seems simple, but probably very inefficient for large numbers of locks...

sinbad:
> Though my first question is: why bolt?

Yeah, I know what you mean. Mostly because if I was going to do it, it would be a single file loaded into memory once & written out once at the end too (separate files for each lock is overkill since all you need is a boolean). Given that bolt already does this and has handling in place for parallel processes trying to access those files already, seemed like not reinventing was preferable, especially since to do a good job handling all the edge cases would have distracted me from what I was actually trying to do. 

If we decide this whole PR is viable as an approach and that bolt is a bit heavyweight we can replace it for the final version.

commands/command_post_checkout.go
```diff
@@ -0,0 +1,73 @@
...
+	// Sadly we don't get any information about what files were checked out,
+	// so we have to check the entire repo
+	err := locking.FixAllLockableFileWriteFlags()
+	if err != nil {
+		LoggedError(err, "Warning: post-checkout locked file check failed: %v", err)
+	}
+}
+
+func init() {
+	RegisterCommand("post-checkout", postCheckoutCommand, func(cmd *cobra.Command) {
```
technoweenie: Ew, `RegisterCommand()` should probably just take a nil callback :)

sinbad:
> This PR is way too big to review (beyond just chrome having issues rendering this page 😉). I'd really like to see it split out into individual components that we can discuss separately

Yep, I totally agree 😞  Unfortunately I needed to hit a certain number of features before I could even demo where I was trying to go with it, which was necessary both for my own evaluation of the approach and to be able to justify it to anyone else. This PR wasn't for merging or proper review, more for "does going down this path look reasonable or am I insane". I'm glad you think that on principle it seems worth carrying on with.

I'm going to go a few steps further in this branch to try to dig to the bottom of a few of the concerns I still have (like performance and potential unknown edge cases with read-only files). Then when I'm happy it's practical I'll do the whole thing over as smaller PRs.

sinbad:
> I really want to get away from static caches (like locking.cachedLockablePatterns) in all packages except commands.

Can you explain why? It would be vastly less performant not to cache these patterns, and passing caches backwards and forwards to `command` breaks encapsulation. Unless you mean having some sort of encapsulated extensible 'state' object which `command` manages the lifecycle of but which is opaque to it so that other packages can keep data in it for performance reasons?

locking/cache.go
```diff
@@ -0,0 +1,213 @@
...
+func cacheLock(filePath, id string) error {
+	return runLockDbFunc(func(tx *bolt.Tx) error {
+		path2id, err := tx.CreateBucketIfNotExists(pathToIdBucketName)
+		if err != nil {
+			return err
+		}
+		id2path, err := tx.CreateBucketIfNotExists(idToPathBucketName)
+		if err != nil {
+			return err
+		}
```
sinbad: I think when I do this over I'm going to replace boltdb, but for reference, the reason for using `CreateBucketIfNotExists` is that you have to `Get` the bucket inside the transaction anyway because you're not allowed to store bucket references outside of transactions in Bolt. So there's no benefit to having a separate creation process done once at init because you can't hold on to the reference anyway. You might as well not have separate code to create *and* when one call does either/or.

I didn't use `api.Lock` because I prefer to avoid introducing transitive dependencies in function signatures; if I did that anything calling it then depends on the `api` package and from experience that sort of thing leads to eventual circular deps. If you'd prefer a structure I'd introduce a new one in this package.

technoweenie:
I think of packages like `locking`, `scanner`, etc, as libraries that other go devs could use for their own lfs apps. I'd prefer to have `GetLockablePatterns()` return some `*locking.Lockables` struct, with funcs like:

```
func (*Lockables) IsLockable(path string) bool

// replace locking.FixLockableFileWriteFlags()
func (*Lockables) FixWriteFlags(files []string) error
```

Would something like that work? I wonder if this and `config.GetAttributePaths()` could go into their own package? `attributes.GetPaths(gitDir, workingDir string) *attributes.Paths`

sinbad:
Closing; now I'm relatively happy this is a sane direction it will return as several smaller PRs later.

sinbad:
> I'd prefer to have GetLockablePatterns() return some *locking.Lockables struct, with funcs 

Hmm, I guess so, more of an OO approach then. It does imply either most of the package interface would become part of that struct (since a lot of it depends on the lockables list), or that it becomes are required function parameter.

locking/cache.go
technoweenie: >  didn't use api.Lock because I prefer to avoid introducing transitive dependencies in function signatures

:+1: I'm thinking we could even eventually get rid of `api`. Callers should just be using packages like `locking`, `transfer`, `batch`, etc.

technoweenie:
Yes, I'd prefer a more OO approach. It will give people more flexibility if they want to build on the internal LFS packages, as well as making testing more straightforward.

Can you submit the doc changes as a small PR so we can retain that valuable info? If not, I'll just plagiarize it :)

sinbad:
Sure, I'll do a separate PR in a sec.

PR #1593, opened 2016-10-18
Update .mailmap

dpursehouse:
Change-Id: I139809347d7e59d85758e3696a01f2e2e8329c91
Signed-off-by: David Pursehouse <dpursehouse@collab.net>

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -7,8 +7,10 @@ Evan Priestley <epriestley@javelinjs.com> <git@epriestley.com>
 Josh Vera <josh@joshvera.com> <vera@github.com>
 Lars Schneider <larsxschneider@gmail.com> <lars.schneider@autodesk.com>
 Lee Reilly <lee@leereilly.net> <lee@github.com>
+Noam Y. Tenne <noam@10ne.org> noamt <noam@10ne.org>
 Rick Olson <technoweenie@gmail.com> rick <technoweenie@gmail.com>
 Rick Olson <technoweenie@gmail.com> risk danger olson <technoweenie@gmail.com>
+Rick Olson <technoweenie@gmail.com> Your Name <you@example.com>
 Riku Lääkkölä <riku.laakkola@iki.fi> <riku.laakkola@coriant.com>
 Ryan Simmen <ryan.simmen@gmail.com> <ryansim@microsoft.com>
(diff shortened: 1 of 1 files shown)


sschuberth:
Note sure what the policy is here, but I find it confusing to have Gerrit's `Change-Id` in the commit message body (BTW, it's a pity that GitHub still doesn't let me add review comments to lines of commit messages). Similarly for the `Sign-off-by`, this project does not require it, but is it ok to leave it in?

dpursehouse:
@sschuberth I have the Gerrit hook installed in my templates so it gets added on all new clones.  I could manually remove it, but meh.  If project owners don't like it they remove it (see commits on facebook/buck for example) and I'm fine with that.

sschuberth:
@dpursehouse I just checked, given that your previous commits with both `Change-Id` and `Sign-off-by` were accepted, I guess this is fine, too :-)

.mailmap
```diff
@@ -7,8 +7,10 @@ Evan Priestley <epriestley@javelinjs.com> <git@epriestley.com>
 Josh Vera <josh@joshvera.com> <vera@github.com>
 Lars Schneider <larsxschneider@gmail.com> <lars.schneider@autodesk.com>
 Lee Reilly <lee@leereilly.net> <lee@github.com>
+Noam Y. Tenne <noam@10ne.org> noamt <noam@10ne.org>
 Rick Olson <technoweenie@gmail.com> rick <technoweenie@gmail.com>
 Rick Olson <technoweenie@gmail.com> risk danger olson <technoweenie@gmail.com>
+Rick Olson <technoweenie@gmail.com> Your Name <you@example.com>
```
ttaylorr: Is this supposed to map to `Your Name <you@example.com>`?

sschuberth: No, *from* `Your Name <you@example.com>` to `Rick Olson <technoweenie@gmail.com>`. I remember also stumbling upon a commit from this user when I created my `.mailmap`, but I did not find sufficient proof that this is in fact @technoweenie. I guess @dpursehouse still remembered that this is the case.

ttaylorr: 👍 . Cool, thanks for clarifying!

dpursehouse:
No 100% absolute concrete evidence, but the commit with this name seems to
be from Rick.

PR #1595, opened 2016-10-19
scanner: refactor RevListsShas into type

ttaylorr:
This pull-request refactors the original implementation of the `RevListShas` method into a type that retains the same behavior.

This is done such that we can easily manipulate a channel of revs coming out of this function. That is not made use of (yet), but will become the case in a subsequent PR.

--

/cc @technoweenie @rubyist @sinbad @sschuberth @larsxschneider

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -43,18 +43,6 @@ type Ref struct {
 	Sha  string
 }
 
-// MatchesNameAndType returns whether or not the receiving *Ref "r" has the same
-// name and type as the "other" *Ref. Nil *Refs are invalid values and thus do
-// not possess equality.
-func (r *Ref) MatchesNameAndType(other *Ref) bool {
-	if r == nil || other == nil {
-		return false
-	}
-
-	return r.Name == other.Name &&
-		r.Type == other.Type
-}
-
 // Some top level information about a commit (only first line of message)
 type CommitSummary struct {
 	Sha            string
--- a/git/git.go
+++ b/git/git.go
@@ -43,6 +43,18 @@ type Ref struct {
 	Sha  string
 }
 
+// MatchesNameAndType returns whether or not the receiving *Ref "r" has the same
+// name and type as the "other" *Ref. Nil *Refs are invalid values and thus do
+// not possess equality.
+func (r *Ref) MatchesNameAndType(other *Ref) bool {
+	if r == nil || other == nil {
+		return false
+	}
+
+	return r.Name == other.Name &&
+		r.Type == other.Type
+}
+
 // Some top level information about a commit (only first line of message)
 type CommitSummary struct {
 	Sha            string
--- a/scanner/rev_cache.go
+++ b/scanner/rev_cache.go
@@ -1,0 +1,30 @@
+package scanner
+
+import "sync"
+
+type RevCache struct {
+	m  map[string]string
+	mu *sync.Mutex
+}
+
+func NewRevCache(m map[string]string) *RevCache {
+	return &RevCache{
+		m:  m,
+		mu: &sync.Mutex{},
+	}
+}
+
+func (c *RevCache) GetName(sha string) (string, bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	rev, ok := c.m[sha]
+	return rev, ok
+}
+
+func (c *RevCache) Cache(sha, name string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.m[sha] = name
+}
--- a/scanner/revs.go
+++ b/scanner/revs.go
@@ -2,12 +2,10 @@ package scanner
 
 import (
 	"bufio"
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"regexp"
-	"strconv"
 	"strings"
 
 	"github.com/github/git-lfs/git"
@@ -27,6 +25,10 @@ var (
 	z40            = regexp.MustCompile(`\^?0{40}`)
 	ambiguousRegex = regexp.MustCompile(`warning: refname (.*) is ambiguous`)
 )
+
+func RevListShas(refLeft, refRight string, opt *ScanRefsOptions) (<-chan string, <-chan error, error) {
+	return NewRevListScanner(opt).Scan(refLeft, refRight)
+}
 
 type RevListScanner struct {
 	ScanMode         ScanningMode
@@ -93,18 +95,13 @@ func (r *RevListScanner) scanAndReport(cmd *wrappedCmd, revs chan<- string, errs
 		revs <- sha1
 	}
 
-	var err error
 	stderr, _ := ioutil.ReadAll(cmd.Stderr)
 	if err := cmd.Wait(); err != nil {
-		err = fmt.Errorf("Error in `git rev-list --objects`: %v %s", err, stderr)
+		errs <- fmt.Errorf("Error in `git rev-list --objects`: %v %s", err, stderr)
 	} else {
 		if match := ambiguousRegex.FindStringSubmatch(string(stderr)); len(match) > 0 {
-			err = fmt.Errorf("Error: ref %s is ambiguous", match[1])
+			errs <- fmt.Errorf("Error: ref %s is ambiguous", match[1])
 		}
-	}
-
-	if err != nil {
-		errs <- err
 	}
 
 	close(revs)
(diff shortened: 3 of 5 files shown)


git/git.go
```diff
@@ -43,6 +43,18 @@ type Ref struct {
 	Sha  string
 }
 
+// MatchesNameAndType returns whether or not the receiving *Ref "r" has the same
+// name and type as the "other" *Ref. Nil *Refs are invalid values and thus do
+// not possess equality.
+func (r *Ref) MatchesNameAndType(other *Ref) bool {
```
sinbad: Is there any (valid) situation where Name & Type are equal but Sha is not? Since we never modify refs I don't think so. In which case the struct equality operator is enough?

scanner/rev_cache.go
```diff
@@ -0,0 +1,30 @@
...
+
+func (c *RevCache) GetName(sha string) (string, bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	rev, ok := c.m[sha]
+	return rev, ok
+}
+
+func (c *RevCache) Cache(sha, name string) {
```
sinbad: I find this a little confusing since sha->name would logically be a multi-map (shas can have multiple refs pointing at them).

scanner/revs.go
```diff
@@ -22,139 +21,194 @@ const (
...
-	// Use "--" at the end of the command to disambiguate arguments as refs,
-	// so Git doesn't complain about ambiguity if you happen to also have a
-	// file named "master".
-	refArgs = append(refArgs, "--")
+// Scan reports a channel of revisions, and a channel of errors encountered
+// while scanning those revisions, over the given range from "left" to "right".
+//
+// If there was an error encountered before beginning to scan, it will be
+// returned by itself, with two nil channels.
+func (r *RevListScanner) Scan(left, right string) (<-chan string, <-chan error, error) {
```
sinbad: StringChannelWrapper has a more convenient interface for handling strings and errors rather than separate channels.

sinbad: (Realise the scanner methods predated this but might be worth taking advantage of)

git/git.go
ttaylorr: Thanks for bringing this up. I can't think of any legal situations where this would/could happen off-hand, so I simplified the behavior in https://github.com/github/git-lfs/pull/1595/commits/cf8fac17aeb2c0c64b3c413d72af3f06019108a6.

scanner/rev_cache.go
ttaylorr: Ha, I totally misunderstood the code when I was reading it for the first time. The `Cache(sha, name string)` caches the SHA1 hash of a file to its name on disk (as used [here](https://github.com/github/git-lfs/blob/e0db9cc16262528c476aa04420461315e8bfca23/lfs/scanner.go#L157-L162)).

I rename this type in https://github.com/github/git-lfs/pull/1595/commits/53f2abb59df469bcf8371580056ef32336274358, and added some docs in https://github.com/github/git-lfs/pull/1595/commits/e16e2de4a11885b5c996f722f24830e0210bb515.

scanner/revs.go
ttaylorr: I was debating whether or not to use the *XChannelWrapper` types when I was refactoring this. IMHO, I think passing the channels around directly and using the `tools.CollectErrsFromChan` method is more lightweight than a new type, so I am going to leave this as-is for now.

This will become less of a sore spot a few more PRs down the line.

ttaylorr:
@sinbad thanks for the review! I pushed some more changes out to address your comments 👍

technoweenie:
Moving things to a new package gives us another opportunity to examine the interface. The `lfs` package has a ton of things exported that probably don't need to be. The `scanner` package's main task is to stream LFS pointers given a range of git commits. With that in mind, do `scanner. NameCache`, `scanner.RevListScanner` or even `scanner.RevListShas()` need to be exported? These seem like implementation details for an LFS object scanner.

Also, I don't think a `scanner.RevListScanner` needs to take a `scanner.ScanRefsOptions` in its constructor. What about something like this?

```go
revlistScanner := newRevListScanner("origin", ScanRefsMode)
revlistScanner.SkipDeletedBlobs = true
revlistScanner.Scan(left, right)
```

If `scanner.ScanRefsOptions` is going to get removed in a future PR, that's totally cool :+1:

technoweenie:
Actually, ignore the last part. I think we still need `ScanRefsOptions` so callers outside the `scanner` pkg can set this info. Then let each inner type like a `revListScanner` use that.

ttaylorr:
> With that in mind, do scanner. NameCache, scanner.RevListScanner or even scanner.RevListShas() need to be exported?

I don't think that these types need to be exported. To keep these PRs small, I think it makes sense to export the same functions that exist in `lfs/scanner.go` until we've moved all of the code over. Once that's done, I figure we can reduce this package down to a single public function, and leave all of the implementation private.

I did, however, make the `nameCache` type private in https://github.com/github/git-lfs/pull/1595/commits/d900abdda13933434b4b3a2387cb8abe65f7d3d4, since that doesn't need to be exported today.

technoweenie:
> This is done such that we can easily manipulate a channel of revs coming out of this function.

Can you provide some example of how this will work, and why a type is needed? The type shares all of the same properties of `ScanRefsOptions`. It looks like all that was changed is `RevListShas()` calls `(*RevListScanner) Scan()`, which is broken up to a few smaller funcs like `refArgs()` and `scanAndReport()`. Seems like `RevListScanner` is unnecessary, and definitely does not need to be exported.

ttaylorr:
> Can you provide some example of how this will work, and why a type is needed?

Sure. When I started this refactor, the presence of the `ScanRefOptions` argument to the function was a smell that indicated there might be a type to extract that would be more appropriate than a large function.

My first inclination was to extract a 2-3 smaller, static functions, and call those, but I hit another smell of passing that `ScanRefOptions` argument all the way down. I then extracted a type that has the same fields as `ScanRefOptions`. This is still a smell, but temporary one.

Since there is code from the `lfs` package that uses the `ScanRefOptions` type, and I wanted to keep this change isolated, I decided to de-structure the data on that type into fields when we construct the `RevListScanner` type. Eventually, when all of the stuff that was in `lfs/scanner.go` is in the `scanner` package, then I'll remove the type, and each of the three types that I'm proposing will only have the information they need.

I think having three types (one that scans refs, one that does the batch check, and one that converts blobs into LFS pointers) is a good idea in that each is isolated in their concern, and doesn't have access to data that they don't need. This makes the types easily testable, and compose-able, which I think is a solid step forwrad.

technoweenie:
I see `ScanRefOptions` as something similar to `http.Request` (with worse type names). If a caller wants to scan for lfs objects, it creates a new `ScanRefOptions`, calls `Scan()`, and gets back a result. It only has 3 immutable properties (I'm purposely ignoring `nameCache`, which I think we both dislike). Destructuring it just feels overkill to me.

* The three classes are private, and closely related right now. 
* There are three because Git doesn't give us a better way to scan a repo. What if they did, an we could drop it to 1 command? Definitely isn't changing any time soon though.
* Adding a property to `ScanRefOptions` would require also adding it to the 3 commands that need it. Also unlikely.

One upside, is if the three types end up with similar method names, they won't need prefixes to keep them apart:

```go
revListScannerScan()
catFileBatchScan()

(*revListScanner) Scan()
(*catFileBatch) Scan()
```

If you really want to add the types, go for it. I was just asking a question. Some times I wish I would've gotten more questions about types I've written. If you do keep this, I still think `RevListScanner` should be private. I don't think anything uses it outside of the scanner package.

> Eventually, when all of the stuff that was in `lfs/scanner.go` is in the `scanner` package, then I'll remove the type (`ScanRefOptions`), and each of the three types that I'm proposing will only have the information they need.

If you do that, then you'll need a public Scan function with 3 positional args:

```go
scanner.Scan(scanner.ScanAllMode, "origin", false)
```

Positional args, especially for optional ones like `SkipDeletedBlobs`, are not my favorite. So I think it'd make more sense to just rename `lfs.ScanRefOptions` to something like `scanner.Options`, instead of removing it in one PR, and bringing it back in another.

ttaylorr:
Thanks again for this feedback 👍 . I didn't do a great job of articulating what I'm going for in this series of PRs, which I think can make it difficult to look at each by itself. I did my best at responding to some of your questions below.

My inclination is to merge this PR and introduce the type after changing the name of `ScanRefOptions` to `scanner.Options`, and making the `RevListScanner` a private member of the package. I think this is a good idea, but if this does become overkill after all of the PRs in this series have been merged, then I definitely agree that we can change it fairly easily. 

> If a caller wants to scan for lfs objects, it creates a new ScanRefOptions, calls Scan(), and gets back a result.

I may be misunderstanding you, but my thinking is that the type could be made private so the entrypoint to this API is simple. More importantly, the caller doesn't have to concern him or herself with the implementation, or existence, of the `*RevListScanner`. I started working towards that [here](https://github.com/github/git-lfs/blob/d900abdda13933434b4b3a2387cb8abe65f7d3d4/scanner/revs.go#L34-L38), but still need to make the type private to finish it off.

> There are three because Git doesn't give us a better way to scan a repo. What if they did, an we could drop it to 1 command?

I think in either case we'd have to change a lot of the internals of the code, but I agree that this isn't a likely thing to happen in the near future.

> Positional args, especially for optional ones like SkipDeletedBlobs, are not my favorite.

I definitely agree. In the final iteration of this series of PRs, I'm imagining that something like `ScanRefOptions` (maybe called `*scanner.Options`) will still exist, but we'd de-structure the fields on that struct into the fields that each component needs, that way they only have access to what they need to do their jobs.

technoweenie:
>>If a caller wants to scan for lfs objects, it creates a new ScanRefOptions, calls Scan(), and gets back a result.
>
>I may be misunderstanding you, but my thinking is that the type could be made private so the entrypoint to this API is simple. More importantly, the caller doesn't have to concern him or herself with the implementation, or existence, of the `*RevListScanner`

I'm referring to `scanner.ScanRefs()` (`scanner.ScanTree()`, `scanner.ScanIndex()`, `scanner.ScanUnpushed()`, etc,  are already simple). You're going to revert back to 2 positional and an optional arg? I'm not talking about `RevListScanner`, which has never been exported.

ttaylorr:
Hi, all. I wanted to post a quick update after I've given some thought to this PR over the weekend. 

First of all, I'd like to thank @technoweenie and @sinbad for their feedback, and for asking hard questions.

My goal with this series of PRs was for them to be small, and easily digestible in pieces. I failed, however, to clearly articulate my vision and reasoning behind removing/adding some code in both the comments and PR description. I will try and further articulate my vision in the future.

For now, I am going to close this PR. My thinking here is that we have code in `lfs/scanner.go` that works, and is simple. Let's start by moving that code over to the `scanner` package, making it work in a parallel fashion (where there are no individual bottlenecks that can't be solved by increasing the parallelism), and then refactor later, if necessary.

I think this allows us to keep the simple code that we already have while making small, incremental changes moving us towards our goals. My new plan, going forward, is:

0. In three separate PRs, move over the code from `lfs/scanner.go` into package `scanner`, leaving the `lfs/scanner.go` code to delegate into the new package.
0. In the next PR, apply small, calculated refactorings as described above.
0. Transition all calling code to use the `scanner` package, instead of the `lfs/scanner.go` functions.
0. 🎉 

Once again, I really do appreciate the feedback, and can't wait to get cracking on the rest of the PRs in this series. 🤘

PR #1596, opened 2016-10-19
scanner: move RevListShas to new package

ttaylorr:
This pull-request moves the `RevListShas` implementation into the new `scanner` package, and delegates calls into it.

This is the first pull-request in a series that I'm working on which will introduce a refactored implementation of what's found in `lfs/scanner.go` into a new `scanner` package, which will operate as an async pipeline. Currently, the scanner must complete each operation in its entirety before we even kick off the transfer queue. This works fine when dealing with small working trees, but is problematic when we have to buffer a large amount of Git blobs in memory before transferring them.

Since the rev scanner is the first operation in that chain, I started there. A diagram of where I'd like to end up is below:

I'm planning on leaving all of the existing calling code as-is, and keep relying on the `lfs` package (delegating into the `scanner` package from `lfs/scanner.go`) until I am completely done with this refactoring.

--

/cc @technoweenie @rubyist @sinbad @sschuberth @larsxschneider


ttaylorr:
This PR pretty much just moves the code from one package to another, but I wanted to note in more detail the reason I did not adopt the `TChannelWrapper` pattern.

- The `TChannelWrapper` types are defined in the `lfs` package, which makes sense, since most packages depend on `lfs`, not the other way around. Since the `lfs` package is delegating into `scanner` for the time being, we'd have to additionally move the `TChannelWrapper` types to `tools` or similar, and use it from there. This seems undesirable to me because my intent is to no longer use these types on the `scanner` package, because...
- Even though most LFS packages are designated as being unsafe for public use due to the amount of flux in our publicly exported APIs, a green-field `scanner` package seems like a good place to start stabilizing our APIs. Thus, we shouldn't make the assumption that callers of the `scanner.ScanRevsToChan` function necessarily _want_ the behavior that the chan wrappers provide. We should provide APIs at the lowest common denominator so that callers can decide how they want to treat that data, and not be constrained by decisions made at the API-layer. My plan going forward with this is to make the signatures return two `(<-chan string, <-chan error)`, and introduce a helper function at the call-sites of these functions to provide the behavior that the `TChannelWrapper` types provided.

PR #1597, opened 2016-10-19
scanner: move `lfs.catFileBatchCheck` to `scanner` package

ttaylorr:
This pull-request migrates the `lfs.catFileBatchCheck` method over from `lfs/scanner.go` and into the new `scanner` package.

Since there are no public uses of this method outside of `lfs/scanner.go`, this migration was pretty straightforward.

The next PR will refactor this function into a new type in the `scanner` package capable of accepting a stream of revs, moving us further towards the queue/worker pattern as described in #1596.

--

/cc @technoweenie @sinbad @sschuberth @rubyist @larsxschneider

### DIFF
--- a/lfs/scanner.go
+++ b/lfs/scanner.go
@@ -15,7 +15,6 @@ import (
 
 	"github.com/github/git-lfs/git"
 	"github.com/github/git-lfs/scanner"
-	"github.com/github/git-lfs/tools"
 	"github.com/rubyist/tracerx"
 )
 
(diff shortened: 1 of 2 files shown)


sinbad:
I know this applies more to the target branch than this one, but I think the `startCommand` function should be relocated to the `subprocess` package rather than being duplicated now in `lfs` and `scanner`.

Otherwise 👍

ttaylorr:
Rebased locally and changed the base of this PR to point at `rev-scanner`, instead of `rev-scanner-refactor`.

PR #1600, opened 2016-10-20
commands,lfs/tq: Only send unique OIDs to the Transfer Queue

ttaylorr:
This pull-request fixes a regression where sending a non-unique OID to a transfer queue more than once would result in a `panic()` after having decremented the `sync.WaitGroup` past zero.

The fix comes in two parts:

0. https://github.com/github/git-lfs/commit/047fe19830fd497dd9882ed38a3030ec81938491: Only allow calling `lfs.TransferQueue.Add()` once per OID.
0. https://github.com/github/git-lfs/commit/d770dfa09eb289ff4d4167d91ce4a9030f9886ed: De-dup incoming pointers such that the progress meter displays the right number of bytes and files.

Regarding https://github.com/github/git-lfs/commit/d770dfa09eb289ff4d4167d91ce4a9030f9886ed, I am not hugely excited about adding another conditional against a potentially large set of data, but given the expected lifetime of this code, and the relative speed of the implementation (`tools.StringSet`), I am OK with releasing this.

As an additional note, this is a similar piece of functionality as compared what we do when calling `git lfs fetch` in `lfs.readyAndMissingPointers`. This function accepts:

- a list of all pointers
- an include filter
- an exclude filter

and returns (in order):

- the pointers we already have
- the pointers we need
- the total number of bytes we're expecting to transfer.

To make this calculation, it uses a similar mechanism [here](https://github.com/github/git-lfs/blob/b7b1f2b7985c8f7a18c563c70898f669ffb8c13c/commands/command_fetch.go#L317-L322) to de-dup pointers that we have seen already.

--

/cc @technoweenie @rubyist @sinbad @sschuberth

### DIFF
--- a/commands/uploader.go
+++ b/commands/uploader.go
@@ -42,14 +42,23 @@ func (c *uploadContext) prepareUpload(unfiltered []*lfs.WrappedPointer) (*lfs.Tr
 	totalSize := int64(0)
 	missingSize := int64(0)
 
+	// XXX(taylor): temporary measure to fix duplicate (broken) results from
+	// scanner
+	uniqOids := tools.NewStringSet()
+
 	// separate out objects that _should_ be uploaded, but don't exist in
 	// .git/lfs/objects. Those will skipped if the server already has them.
 	for _, p := range unfiltered {
-		// object already uploaded in this process, skip!
-		if c.HasUploaded(p.Oid) {
+		// object already uploaded in this process, or we've already
+		// seen this OID (see above), skip!
+		if uniqOids.Contains(p.Oid) || c.HasUploaded(p.Oid) {
 			continue
 		}
+		uniqOids.Add(p.Oid)
 
+		// increment numObjects and totalSize early (even if it's not
+		// going into uploadables), since we will call Skip() based on
+		// the results of the download check queue
 		numObjects += 1
 		totalSize += p.Size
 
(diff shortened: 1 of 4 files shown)


ttaylorr:
This bug can be triggered in situations where the LFS pointer file changes, but the OID remains the same. Since a change in size would imply a change in OID, the only way that we can trigger this is by changing the `version` field, which was something that happened often in the early LFS days.

Here's an example:

```diff
diff --git a/git-lfs/psds/100.psd b/git-lfs/psds/100.psd
index 06ce4d1..4e72f9c 100644
--- a/git-lfs/psds/100.psd
+++ b/git-lfs/psds/100.psd
@@ -1,3 +1,3 @@
-version http://git-media.io/v/2
+version https://git-lfs.github.com/spec/v1
oid sha256:2d9668b1766d9fc1f2cc9ba188ae7f1676299f1d1ac649091705ce953d0932ef
size 174884
```

@technoweenie do you think this is worth testing for?

technoweenie:
> do you think this is worth testing for?

Absolutely. Even though the vast majority of objects use the version uri in the first public release, we have to continue to support old pointer formats since git blobs are immutable.

ttaylorr:
@technoweenie I agree! Here's a commit with that test: https://github.com/github/git-lfs/pull/1600/commits/0195476bce19fbdf137aa038b003da64834070a3.

PR #1603, opened 2016-10-21
transfer: more descriptive "expired at" errors

ttaylorr:
This pull request enhances of the descriptiveness of the error message returned when an object has expired.

To accomplish this, we needed to teach the `api.Object` how to return which action it was expired for, so we have a basis to compare the duration against the time that the transfer was attempted.

One open ❓ I'm thinking about is whether or not we should change the behavior for the `*api.Object.IsExpired` method to also take a direction so that it can compare against a specific action, and not all of them (as it currently does). This may not actually matter so much, since spec-compliant implementations of the LFS API will only send back the action that we're interested in.

The new error message is down below:

```
Git LFS: (0 of 1 files) 0 B / 29 B
    LFS: lfs/transfer: object "54899ec3690514f121a8f55729e9f802cde71091c3245966175a0bab6134573d" has expired at 2016-10-21 14:24:33.728774785 -0600 MDT, -4m55.024738112s ago
    error: failed to push some refs to 'http://127.0.0.1:57180/push_no_retry_expired_action'
```

--

cc @technoweenie @rubyist @sinbad @sschuberth

### DIFF
--- a/transfer/adapterbase.go
+++ b/transfer/adapterbase.go
@@ -124,11 +124,18 @@ func (a *adapterBase) worker(workerNum int, ctx interface{}) {
 		}
 		tracerx.Printf("xfer: adapter %q worker %d processing job for %q", a.Name(), workerNum, t.Object.Oid)
 
+		// tt is the time that we are to compare the transfer's
+		// `expired_at` property against.
+		tt := time.Now().Add(-objectExpirationGracePeriod)
+
 		// Actual transfer happens here
 		var err error
-		if t.Object.IsExpired(time.Now().Add(objectExpirationGracePeriod)) {
+		if expAt, expired := t.Object.IsExpired(tt); expired {
 			tracerx.Printf("xfer: adapter %q worker %d found job for %q expired, retrying...", a.Name(), workerNum, t.Object.Oid)
-			err = errors.NewRetriableError(errors.Errorf("lfs/transfer: object %q has expired", t.Object.Oid))
+			err = errors.NewRetriableError(errors.Errorf(
+				"lfs/transfer: object %q has expired at %s, %s ago",
+				t.Object.Oid, expAt, tt.Sub(expAt),
+			))
 		} else if t.Object.Size < 0 {
 			tracerx.Printf("xfer: adapter %q worker %d found invalid size for %q (got: %d), retrying...", a.Name(), workerNum, t.Object.Oid, t.Object.Size)
 			err = fmt.Errorf("Git LFS: object %q has invalid size (got: %d)", t.Object.Oid, t.Object.Size)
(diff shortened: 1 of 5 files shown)


transfer/adapterbase.go
```diff
@@ -124,11 +124,18 @@ func (a *adapterBase) worker(workerNum int, ctx interface{}) {
 		}
 		tracerx.Printf("xfer: adapter %q worker %d processing job for %q", a.Name(), workerNum, t.Object.Oid)
 
+		// tt is the time that we are to compare the transfer's
+		// `expired_at` property against.
+		tt := time.Now().Add(-objectExpirationGracePeriod)
```
sinbad: You've used `-objectExpirationGracePeriod`, but it should be a positive offset as it was before. I think the naming might have misled you; it's not really a 'grace period', it's trying to account for the fact that actually using the link will take a little time and so if it expires in the very near future it's not worth trying to use it.

transfer/adapterbase.go
```diff
@@ -124,11 +124,18 @@ func (a *adapterBase) worker(workerNum int, ctx interface{}) {
...
+
 		// Actual transfer happens here
 		var err error
-		if t.Object.IsExpired(time.Now().Add(objectExpirationGracePeriod)) {
+		if expAt, expired := t.Object.IsExpired(tt); expired {
 			tracerx.Printf("xfer: adapter %q worker %d found job for %q expired, retrying...", a.Name(), workerNum, t.Object.Oid)
-			err = errors.NewRetriableError(errors.Errorf("lfs/transfer: object %q has expired", t.Object.Oid))
+			err = errors.NewRetriableError(errors.Errorf(
+				"lfs/transfer: object %q has expired at %s, %s ago",
```
sinbad: Maybe just 'expires at' rather than trying to give a time offset since in anything expiring in the very near future is not worth trying and the 'ago' could look backwards - it's reasonable to reject a link that expires in 1s so we fetch new ones from the API but `-1s ago` would look weird ;)

Worth noting these messages should never be seen by the user anyway. The expected behaviour is that they trigger another API request for fresh links internally.

technoweenie: :+1: The time is also too precise, it doesn't need to show sub second values. Also, it should be in local time if possible, so the tz doesn't have to be set. Hopefully doing that conversion isn't too terrible :)

```diff
-has expired at 2016-10-21 14:24:33.728774785 -0600 MDT
+expires at 2016-10-21 14:24:33
```

ttaylorr:
```
Git LFS: (0 of 1 files) 0 B / 29 B
    LFS: lfs/transfer: object "54899ec3690514f121a8f55729e9f802cde71091c3245966175a0bab6134573d" expires at 24 Oct 16 13:38 MDT
    error: failed to push some refs to 'http://127.0.0.1:62541/push_no_retry_expired_action'
```

PR #1605, opened 2016-10-24
Allow usage of proxies even when contacting localhost

chalstrick:
Sometimes it makes sense to use a proxy even when the remote host is localhost
or 127.0.0.1. git-lfs in contrast to e.g. native git does not allow to use a
proxy when talking to localhost. Remove this special treatment of localhost
and react only on http_proxy or no_proxy env variables.

My use case: I wanted to debug/analyze network traffic from a git client to the
git server and the lfs server running on my developer machine. I set up
mitmproxy [1] as a proxy server running on localhost and I set http_proxy to
point to mitmproxy. I was able to see all the traffic between the client and
the local git server in mitmproxy but the the communication between the client
and the lfs server was hidden because git-lfs ignored the proxy settings
because the hostname was localhost.

### DIFF
--- a/httputil/proxy.go
+++ b/httputil/proxy.go
@@ -1,7 +1,6 @@
 package httputil
 
 import (
-	"net"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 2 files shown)


technoweenie:
This is a great use case :+1: You can fix the ci error by removing [this line](https://github.com/chalstrick/git-lfs/blob/e4e53254d91fcdbbf9069260fb48068170300391/httputil/proxy.go#L4), which is no longer needed since this file is no longer using `net.SplitHostPort()` or `net.ParseIP()`. 

FWIW: `script/fmt` or `script/test` would have removed that line automatically if you have `goimports` installed.

ttaylorr:
After you fix the CI error this looks good to me 👍

sschuberth:
Since the code has been modified now, I wonder whether [this comment](https://github.com/github/git-lfs/pull/1605/files#diff-a57c4db7daf415113329bcf4ee5437e6R89) should also go away.

chalstrick:
I updated the comment and cleaned the imports. Looks good for me now

sschuberth:
> Looks good for me now

Except maybe for the missing dot at the end of the code comment :-D (Nitpick!)

chalstrick:
Agree, I added the missing dot

PR #1607, opened 2016-10-25
remove scratch file 'q'

ttaylorr:
cc @technoweenie


ttaylorr:
Looks like this file was introduced by mistake in https://github.com/github/git-lfs/pull/1605.

PR #1610, opened 2016-10-26
Remove HTML artifacts

russplaysguitar:
Just saw them and thought I'd fix them really quick


ttaylorr:
I think these are here intentionally, but @technoweenie would know for sure.

technoweenie:
They are intentional, since these files are processed with [ronn](http://rtomayko.github.io/ronn/).

russplaysguitar:
Ah good to know, thanks

PR #1616, opened 2016-11-01
Fast directory walk

### LINKED ISSUE #1611: running "git lfs track" in a large repository is very slow
Executing `git lfs track` in a large repository can be very slow. It takes >30s to run `git lfs track` in a repository with 31k files under version control and > 600k files in the same folder (mostly ignored in .gitignore) on modern hardware (most recent MacBook Pro at time of writing).

Being in itself an issue, it makes the usage of the popular tool Sourcetree unusable as it will run a `git lfs track` after every staging action.

sinbad:
Fixes #1611

Provides a number of more optimal replacements for `filepath.Walk`. The easiest of which is `tools.FastWalkGitRepo` which automatically avoids walking `.git` and anything that matches any `.gitignore` it discovers as it walks.

Lower-level versions are available to include/exclude custom patterns.

I've used this in `git lfs track` to significantly speed up the enumeration of `.gitattributes` files, particularly when a repo contains a lot of files, most of which are in the `.gitignore`, as in #1611. 

I've also refactored `FilenamePassesIncludeExcludeFilter` into the `tools` package, and extracted a utility function `FileMatch` which replaces `filepath.Match` by mirroring the behaviour of `gitignore` pattern matching, which wasn't completely supported before. This has a side effect of enhancing the capabilities of the include/exclude filters used elsewhere to support a more complete range of wildcard options including the "**" folder match.


ttaylorr:
Looks like the `clone (with .lfsconfig)` test is at least flaky on Windows, per [this](https://ci.appveyor.com/project/git-lfs/git-lfs/build/1.0.295) run. Going to give this a full review now 👍

commands/command_track.go
```diff
@@ -181,16 +184,24 @@ func findAttributeFiles() []string {
 		paths = append(paths, repoAttributes)
 	}
 
-	filepath.Walk(config.LocalWorkingDir, func(path string, info os.FileInfo, err error) error {
-		if err != nil {
-			return err
+	fchan, errchan := tools.FastWalkGitRepo(config.LocalWorkingDir)
```
ttaylorr: Nice. What do you think about pulling this out to a private helper function in the `commands` package? I know the logic is duplicated in a few places across this PR, but I'm not so much thinking about that as I want the `findAttributeFiles` func to read cleanly.

tools/filetools.go
```diff
@@ -121,3 +126,250 @@ func VerifyFileHash(oid, path string) error {
...
+// Only paths that are in includePaths and outside excludePaths are passed
+// If includePaths is empty that filter always passes and the same with excludePaths
+// Both path lists support wildcard matches
+func FilenamePassesIncludeExcludeFilter(filename string, includePaths, excludePaths []string) bool {
+	if len(includePaths) == 0 && len(excludePaths) == 0 {
+		return true
+	}
+
+	if len(includePaths) > 0 {
+		matched := false
```
ttaylorr: We could get rid of the above `if` by changing this line to:

```go
matched := len(includePaths) == 0
```

tools/filetools.go
```diff
@@ -121,3 +126,250 @@ func VerifyFileHash(oid, path string) error {
...
+		for _, inc := range includePaths {
+			matched = FileMatch(inc, filename)
+			if matched {
+				break
+			}
+		}
+		if !matched {
```
ttaylorr: Or if we keep the `if` on L139, I don't think we need this one, since we `break` out whenever `matched` is true on L144.

sinbad:
> Looks like the clone (with .lfsconfig) test is at least flaky on Windows, per this run. Going to give this a full review now 👍

Yeah and yet the very same commit worked fine in the [other AppVeyor run](https://ci.appveyor.com/project/git-lfs/git-lfs/build/1.0.294). Seems dubious to me, I can't reproduce it.

commands/command_track.go
sinbad: It's only actually used once in production code, the other place is in the tests. Since what this `findAttributeFiles` function is doing is only hooking out items from the channel that are `.gitattributes` files it seems to me to make sense like it is, processing the channel and skipping anything else. I certainly don't want another function that builds a full list and then needs to be iterated, the other option is a function which takes search criteria but that's pretty specific and nothing else is using it yet. Maybe if something else needs it later it could be refactored but wrapping it speculatively right now might not actually be re-usable anyway.

tools/filetools.go
sinbad: But that conditional guards a lot more than just setting `matched := false`, in fact `matched` is deliberate scoped inside that block. It's that way because early-out by lack of inclusion requires checking every item and only concluding at the end, as opposed to early-out by exclusion which can return as soon as it's matched. That's why the exclusion block is simpler.

tools/filetools.go
sinbad: I don't understand your logic, sorry. You'll need to write me the complete block as you see it because to me that doesn't seem to equate to the same result.

canra:
(USTA)

sinbad:
Re-tested everything on Windows this morning, totally fine. Looks like it's passing on AppVeyor now too, did someone re-run it? If so thanks, I don't have access to do that and was going to push a pointless change just for force it 😄 

Appreciate if someone could approve this, I don't think any changes are needed. @ttaylorr @technoweenie

sinbad:
Since they’re all related variations of the same functionality it seemed sensible to expose them all in case it was useful to re-use them, but can be hidden if you prefer.

tools/filetools.go
ttaylorr: Ah, you're right; nice catch.

tools/filetools.go
```diff
@@ -121,3 +126,250 @@ func VerifyFileHash(oid, path string) error {
...
+
+	waitg.Wait()
+	close(fiChan)
+	close(errChan)
+
+}
+
+// Main recursive implementation of fast walk
+// Increment waitg.Add(1) for each new goroutine launched internally
+func fastWalkItem(parentDir string, itemFi os.FileInfo, excludeFilename string,
```
ttaylorr: This method's name is confusing, mainly because it does not specify what an `item` is in context. Is an item a directory? A file? Something else?

If you could specify what that refers to specifically, and improve the documentation to point out what some of the parameters mean (in particular, I'm thinking of `itemFi` and `fiChan`) that would be 👍

tools/filetools.go
```diff
@@ -121,3 +126,250 @@ func VerifyFileHash(oid, path string) error {
...
+	// still need the Stat() to know whether something is a dir, so use
+	// File.Readdir instead. Means we can provide os.FileInfo to callers like
+	// filepath.Walk as a bonus.
+	df, err := os.Open(fullPath)
+	if err != nil {
+		errChan <- err
+		return
+	}
+	defer df.Close()
+	jobSize := 100
```
ttaylorr: This variable is effectively a constant. Would you mind extracting it and giving it a name outside of this method?

tools/filetools.go
```diff
@@ -121,3 +126,250 @@ func VerifyFileHash(oid, path string) error {
...
+	for children, err := df.Readdir(jobSize); err == nil; children, err = df.Readdir(jobSize) {
+		// Parallelise all dirs, and chop large dirs into batches
+		waitg.Add(1)
+		go func(subitems []os.FileInfo) {
```
ttaylorr: Nice and clean. Love it!

tools/filetools.go
```diff
@@ -121,3 +126,250 @@ func VerifyFileHash(oid, path string) error {
...
+	}
+
+}
+
+// loadExcludeFilename reads the given file in gitignore format and returns a
+// revised array of exclude paths if there are any changes.
+// If any changes are made a copy of the array is taken so the original is not
+// modified
+func loadExcludeFilename(filename, parentDir string, excludePaths []string) ([]string, error) {
+
```
ttaylorr: Nit: please remove this newline.

tools/filetools.go
```diff
@@ -121,3 +126,250 @@ func VerifyFileHash(oid, path string) error {
...
+	f, err := os.OpenFile(filename, os.O_RDONLY, 0644)
+	if err != nil {
+		return excludePaths, err
+	}
+	defer f.Close()
+
+	retPaths := excludePaths
+	modified := false
```
ttaylorr: For these variables that hold the zero-value of their type, I prefer the:

```go
var modified bool
```

declaration, but that's just a stylistic preference, which you should feel free to ignore.

technoweenie:
I think `FastWalk` can be removed, and `FastWalkWithExcludeFiles` should not be exported.

tools/filetools.go
technoweenie: Why bother, unless something else needs to use the same size?

tools/filetools.go
technoweenie: I prefer the way it is, lol. If this is later changed to require a default value of `true`, then the diff of that update would just be:

```diff
-modified := false
+modified := true
```

ttaylorr: Good call!

tools/filetools.go
sinbad: > Is an item a directory? A file? Something else?

It's both, that's why I used `Item` and not `File` or `Dir`. Guess it could be `fastWalkFileOrDir`

tools/filetools.go
sinbad: Yeah I don't want to pollute the namespace, it's only a named var at all to self-document what it's for.

tools/filetools.go
sinbad: Yeah I prefer explicitness of defaults, but maybe just because I come from languages that don't guarantee initialised variables

sinbad:
All done - CLA check seems to be stalled though?

PR #1617, opened 2016-11-02
Filter Protocol Support

ttaylorr:
This pull-request is a continuation of #1382, originally authored by @larsxschneider.

It should serve as a place to merge in all of PRs in the filter-stream series (see description below), and then be merged into master as one huge thing. I'm suggesting this since the PR as of https://github.com/github/git-lfs/commit/46a8ee599d0da1fdee59bcdcdbfcbdc966e5db8e is incomplete, and not in a state that I would feel comfortable landing into master.

In terms of the PR series, here's what I'm thinking:

1. https://github.com/github/git-lfs/pull/1617: Initial implementation of the filter protocol from @larsxschneider, that's this PR. 
2. https://github.com/github/git-lfs/pull/1618: Integration tests to make sure that all subsequent PRs keep working
3. https://github.com/github/git-lfs/pull/1619: Unit tests around the protocol reading/writing to verify that its behavior is correct.
4. https://github.com/github/git-lfs/pull/1620: Make the ObjectScanner behave like a Scanner. 
5. https://github.com/github/git-lfs/pull/1621: An `io.Reader` implementation for reading data off of the filter stream, and wiring that `io.Reader` up throughout the code
6. https://github.com/github/git-lfs/pull/1622: An `io.Writer` implementation for writing streams of data to the filter stream, and wiring that up
7. ~~[pending]:~~ A cleanup PR to rename types, document anything that I missed, and generally apply some 💅  EDIT: this was done in place

-----

/cc @larsxschneider @technoweenie @sinbad @rubyist @sschuberth


ttaylorr:
Ready for review 🎉

lfs/setup.go
```diff
@@ -28,11 +28,13 @@ var (
 		Properties: map[string]string{
 			"clean":    "git-lfs clean -- %f",
 			"smudge":   "git-lfs smudge -- %f",
+			"process":  "git-lfs filter",
```
larsxschneider: I wonder if we should call the `git-lfs` parameter `process` instead of `filter` to align the naming with Git?

lfs/setup.go
```diff
@@ -28,11 +28,13 @@ var (
...
 			"required": "true",
 		},
 		Upgradeables: map[string][]string{
 			"clean":  []string{"git-lfs clean %f"},
 			"smudge": []string{"git-lfs smudge %f"},
+			// TODO: process here, too?
```
larsxschneider: I think we should add `process` here, too.

larsxschneider:
@ttaylorr 

I wonder if we should rebase this PR before merging to make reading the history easier in the future.
Maybe we could rebase it into 3 commits:

#### Commit: Math Min/Max Addition
tools/math.go
tools/math_test.go

#### Commit: Pktline reader/writer
git/packet_reader.go
git/packet_reader_test.go
git/packet_writer.go
git/packet_writer_test.go

#### Commit: Filter protocol
all the remaining changes

What do you think?

commands/command_filter.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+Scan:
+	for s.Scan() {
+		var err error
+		var w io.Writer
+
+		req := s.Request()
+		if req == nil {
+			break
```
larsxschneider: This would only happen in case of an error, right?  If this is true, then maybe we should add a comment here? This error would kill the `git-lfs` process. This is good because `git` would detect that and restart `git-lfs`.

commands/command_filter.go
```diff
@@ -0,0 +1,215 @@
...
+		case "smudge":
+			w = git.NewPacketWriter(os.Stdout, smudgeFilterBufferCapacity)
+			err = smudge(w, req.Payload, req.Header["pathname"])
+		default:
+			fmt.Errorf("Unknown command %s", cmd)
+			break Scan
+		}
+
+		var status string
+		if _, ferr := w.Write(nil); ferr != nil {
```
larsxschneider: Maybe add a comment here that `w.Write(nil)` actually writes a pkt-line flush packet (`0004`) as this is not obvious?

I wonder if `w` should be a `PacketWriter` instead of a `io.Writer`? Then we could use the actual `flush()` function. I feel that could be helpful for our future-selves reading this code 😄

ttaylorr:
> I wonder if we should rebase this PR before merging to make reading the history easier in the future.
> Maybe we could rebase it into 3 commits:

Interesting thought. I'm a little bit cautious about rewriting the history in this branch since a lot of these commits are order dependent and I would _really_ hate to break the tests at this point.

Do you think, for reviewing, the browsing file-by-file could work?

lfs/setup.go
ttaylorr: Good call. I think `process` is a little generic, but I'm on board with `git lfs filter-process`.

commands/command_filter.go
```diff
@@ -0,0 +1,215 @@
...
+				os.Exit(2)
+			}
+		}
+
+		_, err = ptr.Encode(to)
+		return err
+	}
+
+	return nil
+}
```
ttaylorr: We should definitely tackle this before merging. I'm crossing my fingers that this is simple enough to do... I designed the interfaces so that it _should_ be interoperable with the old clean/smudge code, but I'll have to try it out to be sure.

Adding it to my to-do list!

commands/command_filter.go
ttaylorr: Not quite, this happens in the case of a final flush packet, where both the error and request body are `nil`. I think the scanner already has some logic in place for this, so I think this should be removable.

After some quick inspection... turns out this totally _is_ removable. I applied some 🔥  in https://github.com/github/git-lfs/pull/1617/commits/31c80c4e47ae704e010896bc821ebf7678252a2f.

commands/command_filter.go
ttaylorr: Good call. My original intent was to make this interface solely an `io.Writer`, but I think this `w.Write(nil)` business is sort of unintuitive. I think the main goal should have been that the interface is usable as an `io.Writer` so that we can use it throughout LFS and not special-case any sort of flushing mid-way through.

Since the only flush that we need to do happens here, I've gone ahead and made this a little clearer in https://github.com/github/git-lfs/pull/1617/commits/f73fd2344f7eb911e3c711a0698b309a606e6b2d.

git/packet_writer.go
```diff
@@ -57,15 +58,7 @@ func (w *PacketWriter) Write(p []byte) (int, error) {
...
-		// sequence.
-
-		if _, err := w.flush(); err != nil {
-			return 0, err
-		}
-
-		if err := w.proto.writeFlush(); err != nil {
+		if err := w.Flush(); err != nil {
 			return 0, err
 		}
```
larsxschneider: Since we don't need `write(nil)` anymore, shouldn't we remove that feature completely?

ttaylorr:
Ah, failing builds due to some changes on master. Merging in to fix that up.

ttaylorr:
💅  has been applied, so this should be a lot cleaner to review! Some notable changes:

- Renamed `git.PacketWriter` and `git.packetReader` to be `git.PktlineWriter` and `git.pktlineReader`, respectively.
- Renamed `git.ObjectScanner` to `FilterProcessScanner`.
- Renamed `git.protocol` to `git.pktline`.
- Refactored the `clean` command to share code between legacy and latest version
- Refactored the `smudge` command to share code between legacy and latest version
- Added documentation to anything that was lacking it

larsxschneider:
@ttaylorr : 
I rebased your e6eb985 in branch `filter-stream-rebased` with 9 commits: 
https://github.com/github/git-lfs/compare/filter-stream-rebased

```
1. fix typo
2. travis-ci: add build to test latest Git master on macOS
3. travis-ci: test GitLFS with the earlierst supported Git version
4. tools/math: add function to return smaller/larger number of two ints
5. lfs: suppress io.EOF errors during pointer decodes
6. lfs: only combine readers when necessary
7. commands: make clean/smudge functions reusable for new filter protocol
8. git: add Git pkt-line format reader/writer
9. implement new Git filter-process protocol
```
`1-7` are preparation only as they do not contain any filter-process related code.
`8` prepares the filter-process implementation and `9` is the actual implementation.

`e6eb985` contains the exact same content as `filter-stream-rebased` (check the diff - it's empty!)

I think if you use Git blame in the future then it would easier to see what belonged to the initial filter protocol implementation. Plus reviewers can focus on `1-7` to see all changes to the existing code more easily.

If you like this structure then just force push the `filter-stream-rebased` branch to the branch of this PR.

larsxschneider:
In case you wonder how I did that: I just made a soft git reset to the first commit (that keeps all changes) and then I created 9 new commits picking the right line/files. Finally I changed the author for your commits --> 3min work 😄

commands/command_filter_process.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+Scan:
+	for s.Scan() {
+		var err error
+		var w *git.PktlineWriter
+
+		req := s.Request()
+
+		s.WriteStatus("success")
```
larsxschneider: I wonder if we should use `s.WriteStatus(statusFromErr(nil))` here? Because we duplicate knowledge of the protocol (the status flag "success") here. That might be nit-picking...

commands/command_filter_process.go
```diff
@@ -0,0 +1,128 @@
...
+			fmt.Errorf("Unknown command %s", cmd)
+			break Scan
+		}
+
+		var status string
+		if ferr := w.Flush(); ferr != nil {
+			status = "error"
+		} else {
+			status = statusFromErr(err)
+		}
```
larsxschneider: Same here, knowledge about the status "error" is here and in `statusFromErr`. Do you see a nice way to have the status strings only in one location?

test/test-filter-process.sh
```diff
@@ -0,0 +1,97 @@
...
+  printf "$contents" > a.dat
+
+  git add a.dat
+
+  expected="$(pointer "$contents_oid" "${#contents}")"
+  got="$(git cat-file -p :a.dat)"
+
+  diff -u <(echo "$expected") <(echo "$got")
+)
+end_test
```
larsxschneider: I think we should add a third test here to check that Git chooses the filter protocol over the legacy protocol. Any good idea how we could do that? Can could run Git with `GIT_TRACE=1` and then count the invocations of `trace git-lfs: Read filter protocol request.` or something?

ttaylorr: Unsure if this is something that LFS should be testing, but the way I'd do it is to use some no-op legacy filters, and then use the `filter-process` command as the new filter process, as such:

```bash
git \                             
  -c "filter.lfs.process=git-lfs filter-process" \      
  -c "filter.lfs.smudge=cat" \    
  -c "filter.lfs.required=false" \
```

That way Git would only correctly clean/smudge files if it chose the operational `git-lfs filter-process` over the no-op `cat`.

commands/command_filter_process.go
ttaylorr: I do -- https://github.com/github/git-lfs/pull/1617/commits/d236d06a578215dd4500300aaa85ff01df041ffa.

ttaylorr:
> If you like this structure then just force push the filter-stream-rebased branch to the branch of this PR.

I think I prefer the state of this branch as-is. I appreciate you squashing things down in a neat way, but I think the development history is important too.

git/pkt_line.go
```diff
@@ -0,0 +1,137 @@
...
+		return err
+	}
+
+	if _, err := p.w.Write(data); err != nil {
+		return err
+	}
+
+	if err := p.w.Flush(); err != nil {
+		return err
+	}
```
larsxschneider: Maybe add a TODO here to check how this behaves with large (1+GB) files. These files need a large number of pkt-line packets and I wonder if there is a measurable performance gain if we only flush after the last packet. I don't think we need to address this in this PR. The code as is is correct from my point of view.

git/pkt_line_writer.go
```diff
@@ -0,0 +1,134 @@
...
+// As many bytes are removed from "p" as possible and stored in an internal
+// buffer until the amount of data in the internal buffer is enough to write a
+// single packet. Once the internal buffer is full, a packet is written to the
+// underlying stream of data, and the process repeats.
+//
+// When the caller has no more data to write in the given chunk of packets, a
+// subsequent call to `Write(p []byte)` MUST be made with a nil slice, to flush
+// the remaining data in the buffer, and write the terminating bytes to the
+// underlying packet stream. Alternatively, the caller can use the public
+// `Flush() error` func to preform the same operation.
```
larsxschneider: `s/preform/perform/`

git/pkt_line_writer.go
```diff
@@ -0,0 +1,134 @@
...
+// error is returned, along with the number of bytes written to the underlying
+// protocol stream, as described above.
+func (w *PktlineWriter) Write(p []byte) (int, error) {
+	var n int
+
+	if p == nil {
+		if err := w.Flush(); err != nil {
+			return 0, err
+		}
+	}
```
larsxschneider: Since we don't use `Write(nil)` anywhere in the code anymore I would remove that feature as it could be considered a little confusing (you would need to adjust the comment in L47, too).

git/pkt_line_writer.go
```diff
@@ -0,0 +1,134 @@
...
+		m := tools.MinInt(len(p), MaxPacketLength-len(w.buf))
+
+		// Append on all of the data that we could into the internal
+		// buffer.
+		w.buf = append(w.buf, p[:m]...)
+		// Truncate "p" such that it no longer includes the data that we
+		// have in the internal buffer.
+		p = p[m:]
+
+		n = n + m
```
larsxschneider: Ultra pedantic nit: In the reader you used `n += m` (same is true for the flush method below)🤓

git/pkt_line_writer.go
```diff
@@ -0,0 +1,134 @@
...
+	return n, nil
+}
+
+// Flush empties the internal buffer used to store data temporarily and then
+// writes the pkt-line's FLUSH packet, to signal that it is done writing this
+// chunk of data.
+func (w *PktlineWriter) Flush() error {
+	if _, err := w.flush(); err != nil {
+		return err
+	}
```
larsxschneider: As with the current implementation this shouldn't be necessary AFAIK. However, we don't flush after each packet (see performance comment above) then this would be a nice place to run the final flush.

lfs/pointer.go
```diff
@@ -113,7 +113,7 @@ func DecodeFrom(reader io.Reader) ([]byte, *Pointer, error) {
 	written, err := reader.Read(buf)
 	output := buf[0:written]
 
-	if err != nil {
+	if err != nil && err != io.EOF {
```
larsxschneider: Is an `EOF` during pointer decoding really OK? That would be an error in the protocol, no?

lfs/setup.go
```diff
@@ -28,11 +28,13 @@ var (
...
 			"smudge":   "git-lfs smudge -- %f",
+			"process":  "git-lfs filter-process",
 			"required": "true",
 		},
 		Upgradeables: map[string][]string{
-			"clean":  []string{"git-lfs clean %f"},
-			"smudge": []string{"git-lfs smudge %f"},
+			"clean":   []string{"git-lfs clean %f"},
+			"smudge":  []string{"git-lfs smudge %f"},
+			"process": []string{"git-lfs filter"},
```
larsxschneider: **Important:** Shouldn't that be `filter-process` here?

lfs/setup.go
```diff
@@ -41,11 +43,13 @@ var (
...
 			"smudge":   "git-lfs smudge --skip -- %f",
+			"process":  "git-lfs filter-process --skip",
 			"required": "true",
 		},
 		Upgradeables: map[string][]string{
-			"clean":  []string{"git-lfs clean %f"},
-			"smudge": []string{"git-lfs smudge --skip %f"},
+			"clean":   []string{"git-lfs clean %f"},
+			"smudge":  []string{"git-lfs smudge --skip %f"},
+			"process": []string{"git-lfs filter --skip"},
```
larsxschneider: **Important:** Shouldn't that be `filter-process` here?

test/test-filter-process.sh
```diff
@@ -0,0 +1,97 @@
...
+. "test/testlib.sh"
+
+# HACK(taylor): git uses ".g<hash>" in the version name to signal that it is
+# from the "next" branch, which is the only (current) version of Git that has
+# support for the filter protocol.
+#
+# Once 2.11 is released, replace this with:
+#
+# ```
+# ensure_git_version_isnt $VERSION_LOWER "2.11.0"
```
larsxschneider: Maybe we should teach `ensure_git_version` the `g` format?

test/test-filter-process.sh
larsxschneider: I think I would prefer it if we set clean and smudge to `false` here to ensure they are not used.

test/test-install.sh
```diff
@@ -8,14 +8,17 @@ begin_test "install again"
 
   smudge="$(git config filter.lfs.smudge)"
   clean="$(git config filter.lfs.clean)"
+  filter="$(git config filter.lfs.process)"
 
   printf "$smudge" | grep "git-lfs smudge"
   printf "$clean" | grep "git-lfs clean"
+  printf "$filter" | grep "git-lfs filter"
```
larsxschneider: `s/lfs filter/lfs filter-process/`

larsxschneider:
The PR looks really good from my point of view. During my review I found [2 important problems](https://github.com/github/git-lfs/pull/1617#discussion_r87366072) that I think need to be addressed before merge and a couple of nits that could be/should be addressed.

(Yeah, I should have used the "Request changes" button instead of "Approve" ... but well 😄 )

git/pkt_line_writer.go
technoweenie: This is becoming a problem. How long until you can stop blaming your new keyboard :trollface:

commands/command_filter_process.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+Scan:
+	for s.Scan() {
+		var err error
+		var w *git.PktlineWriter
+
+		req := s.Request()
+
+		s.WriteStatus(statusFromErr(nil))
```
ttaylorr: Good thought. This is not possible given the current implementation, because the `*git.FilterProcessScanner` [returns false](https://github.com/github/git-lfs/blob/d236d06a578215dd4500300aaa85ff01df041ffa/git/filter_process_scanner.go#L141-L144) for scan operations that yielded in a non-nill error, causing this for loop to bail out immediately.

I think that it might be cleaner to do:

```go
s.WriteStatus(statusFromErr(s.Err()))
```

_or_, even cleaner, change the signature of `s.WriteStatus` to take an error.

commands/command_filter_process.go
```diff
@@ -0,0 +1,128 @@
...
+	ptr, err := lfs.DecodePointer(from)
+	if err != nil {
+		// If we tried to decode a pointer out of the data given to us,
+		// and the file was _empty_, write out an empty file in
+		// response. This occurs because when the clean filter
+		// encounters an empty file, and writes out an empty file,
+		// instead of a pointer.
+		//
+		// TODO(taylor): figure out if there is more data on the reader,
+		// and buffer that as well.
```
ttaylorr: After discussing in meat-space, the consensus was there will be no more data on the io.Reader coming from Git, since we will have decoded the entire pointer (or at least, tried to) by this point, so I went ahead and removed the TODO.

@larsxschneider can clarify anything that I left out.

commands/command_filter_process.go
```diff
@@ -0,0 +1,128 @@
...
+		return err
+	}
+
+	lfs.LinkOrCopyFromReference(ptr.Oid, ptr.Size)
+
+	return smudge(to, ptr, filename, filterSmudgeSkip)
+}
+
+func filterCommand(cmd *cobra.Command, args []string) {
+	requireStdin("This command should be run by the Git filter process")
```
ttaylorr: Awesome catch. I agree that this is out of scope of the current PR, so I opened #1639 to address.

git/filter_process_scanner.go
```diff
@@ -0,0 +1,199 @@
...
+//     }
+type FilterProcessScanner struct {
+	// pl is the *pktline instance used to read and write packets back and
+	// forth between Git.
+	pl *pktline
+
+	// req is a temporary variable used to hold the value accessible by the
+	// `Request()` function. It is cleared at the beginning of each `Scan()`
+	// invocation, and written to at the end of each `Scan()` invocation.
+	req *Request
```
ttaylorr: Ha, good nit. I did not separate these by a newline because `req` and `err` are both used in _very_ similar ways when calling the `Scan()` (and internal `readRequest()`) function.

test/test-filter-process.sh
ttaylorr: Nice. https://github.com/github/git-lfs/pull/1617/commits/792f5aaa5c07ff97387f31f525f9c31ebb76e9d9!

git/pkt_line_writer.go
ttaylorr: 5c4fb03!

git/pkt_line_writer.go
ttaylorr: 5c4fb03!

lfs/setup.go
ttaylorr: Good question. This is in the `Upgradeables` section of the setup code, meaning it's legit to overwrite this value so long as it matches one of the values in that slice. Since there was code publicly available during the course of this PR that referred to `git-lfs filter` instead of `git-lfs filter-process`, I'm going to keep this.

lfs/setup.go
ttaylorr: See https://github.com/github/git-lfs/pull/1617#discussion_r87466563.

test/test-filter-process.sh
ttaylorr: Good thought. Since adding support for the `git-describe` format would complicate this function (and I'm not planning on gating any more LFS features behind Git versions other than those released), I'm going to forgo this change for now, unless we need something like this in the future.

test/test-install.sh
ttaylorr: Nice catch, ha! Fixed it right up in: 4ae8afc.

git/pkt_line_writer.go
ttaylorr: I think this is a good candidate to look at improving in a future PR.

git/pkt_line.go
```diff
@@ -0,0 +1,137 @@
...
+
+	pktLen, err := strconv.ParseInt(string(pktLenHex[:]), 16, 0)
+	if err != nil {
+		return nil, err
+	}
+
+	if pktLen == 0 {
+		return nil, nil
+	}
```
ttaylorr: 63f5418.

git/filter_process_scanner.go
```diff
@@ -0,0 +1,199 @@
...
+	// of the file in the index.
+	Payload io.Reader
+}
+
+// Scan scans for the next request, or error and returns whether or not the scan
+// was successful, indicating the presence of a valid request. If the Scan
+// failed, there was either an error reading the next request (and the results
+// of calling `Err()` should be inspected), or the pipe was closed and no more
+// requests are present.
+func (o *FilterProcessScanner) Scan() bool {
```
ttaylorr: 129030a.

commands/command_filter_process.go
larsxschneider: Hm. Maybe we can improve this part. This is how a typical request looks like:
```
packet:          git> command=smudge
packet:          git> pathname=path/testfile.dat
packet:          git> 0000
packet:          git> CONTENT
packet:          git> 0000
packet:          git< status=success         # (1) This success is L86 above.
packet:          git< 0000
packet:          git< SMUDGED_CONTENT
packet:          git< 0000
packet:          git< status=success         # (2) This status is set later.
packet:          git< 0000
```

Right now (1) will never set the status to `error`. Instead, GitLFS will just shutdown. Right now (2) sets the status to `error` in case of an error which makes Git shutdown GitLFS (because the filter has the `required=true` flag).

I think we should be consistent here and I think I prefer the (1) approach. If GitLFS detects an error then it should shutdown. This way we ensure that Git gets our "something is wrong" message very clearly.

git/pkt_line_writer.go
larsxschneider: Please see comment in https://github.com/github/git-lfs/commit/5c4fb035af2f564252ab4d6f674ba14a650852bf

ttaylorr:
Thanks everyone! 🎉

eminence:
Tremendous work!

commands/command_filter_process.go
ttaylorr: > I think we should be consistent here

I agree, but I'm not sure what you're suggesting. Should we `Panic()` if an error occurs (a situation where we currently write `status=error\n`)? This code would go [here][1].

[1]: https://github.com/git-lfs/git-lfs/blob/5e48c70305d010b0c9c0092e0f59371f5911cb40/commands/command_filter_process.go#L98-L102

larsxschneider: Yeah, I would panic whenever we run into an error and only report back `status=success`. I think that would be the least amount of code, the easiest to maintain, and the most error resistant.

The status flag is really intended for filters that *can* fail. GitLFS is all or nothing: If it fails then we are done.

lfs/setup.go
lockyluchiano: 1

lfs/setup.go
lockyluchiano: 1

commands/command_filter.go
lockyluchiano: 1

commands/command_smudge.go
```diff
@@ -14,10 +14,58 @@ import (
 )
 
 var (
+	// smudgeInfo is a command-line flag belonging to the "git-lfs smudge"
+	// command specifying whether to skip the smudge process and simply
+	// print out the info of the files being smudged.
+	//
+	// As of v1.5.0, it is deprecated.
 	smudgeInfo = false
+	// smudgeInfo is a command-line flag belonging to the "git-lfs smudge"
```
lockyluchiano: 1

commands/command_filter_process.go
lockyluchiano: 1

git/packet_writer.go
lockyluchiano: 1

test/test-filter-process.sh
lockyluchiano: 1

git/filter_process_scanner.go
lockyluchiano: 1

lockyluchiano:
1

git/pkt_line.go
```diff
@@ -0,0 +1,137 @@
...
+		r: bufio.NewReader(r),
+		w: bufio.NewWriter(w),
+	}
+}
+
+// readPacket reads a single packet entirely and returns the data encoded within
+// it. Errors can occur in several cases, as described below.
+//
+// 1) If no data was present in the reader, and no more data could be read (the
+//    pipe was closed, etc) than an io.EOF will be returned.
```
lockyluchiano: 1

git/pkt_line_writer.go
lockyluchiano: 1

lfs/pointer.go
lockyluchiano: 1

test/test-install.sh
lockyluchiano: 1

lfs/setup.go
marlazinger200: 1

lfs/setup.go
marlazinger200: 1

commands/command_filter.go
marlazinger200: 1

git/packet_writer.go
marlazinger200: 1

commands/command_smudge.go
marlazinger200: 1

commands/command_filter_process.go
marlazinger200: 1

test/test-filter-process.sh
marlazinger200: 1

git/filter_process_scanner.go
marlazinger200: 1

git/pkt_line.go
marlazinger200: 1

lfs/pointer.go
marlazinger200: 1

marlazinger200:
1

git/pkt_line_writer.go
marlazinger200: 1

test/test-install.sh
marlazinger200: 1

PR #1618, opened 2016-11-02
filter_stream: Integration tests

ttaylorr:
This pull request adds an integration test making sure the filter protocol works as expected, when it is enabled.

The test I added creates two files tracked with LFS on two separate branches, and asserts that on each branch, each one of the files:

- exists
- has the correct "smudged" contents
- has a valid LFS pointer underneath.

Currently, running the test is gated on the `$GIT_LFS_USE_LEGACY_FILTER` being enabled (`eq "1"`), and the installed version of Git including support for the filter process. Usually, this is done with:

```bash
ensure_git_version_isnt "$VERSION_LOWER" "2.11.0"
```

but since `2.11.0` isn't released yet, the best we can do is check that the version is a pre-release and contains a "g" in the patch number, like: `git version 2.10.1.710.g6539e97`. This is done by using the hack below, which should be removed once version 2.11.0 is released.

```bash
if [ "1" -ne "$(git version | cut -d ' ' -f 3 | grep -c "g")" ]; then
  echo "skip: $0 git version does not include support for filter protocol"
  exit
fi
```

One open question is how we should treat other existing integration tests once using the filter protocol is the blessed way to clean/smudge. I'm thinking we should ensure that our build matrix covers versions before/after 2.11, and then run the best clean/smudge mechanism we can on each. That way, all the tests will use filter protocol on the versions of Git that support it, and we'll still test the old behavior on machines that don't.

---

/cc @larsxschneider @technoweenie @rubyist @sinbad @sschuberth

### DIFF
--- a/test/test-filter-process.sh
+++ b/test/test-filter-process.sh
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+
+. "test/testlib.sh"
+
+if [ "$GIT_LFS_USE_LEGACY_FILTER" == "1" ]; then
+  echo "skip: $0 (filter process disabled)"
+  exit
+fi
+
+# HACK(taylor): git uses ".g<hash>" in the version name to signal that it is
+# from the "next" branch, which is the only (current) version of Git that has
+# support for the filter protocol.
+#
+# Once 2.11 is released, replace this with:
+#
+# ```
+# ensure_git_version_isnt $VERSION_LOWER "2.11.0"
+# ```
+if [ "1" -ne "$(git version | cut -d ' ' -f 3 | grep -c "g")" ]; then
+  echo "skip: $0 git version does not include support for filter protocol"
+  exit
+fi
+
+begin_test "filter process: checking out a branch"
+(
+  set -e
+
+  reponame="filter_process_checkout"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" repo
+
+  git lfs track "*.dat"
+  git add .gitattributes
+  git commit -m "initial commit"
+
+  contents_a="contents_a"
+  contents_a_oid="$(calc_oid $contents_a)"
+  printf "$contents_a" > a.dat
+
+  git add a.dat
+  git commit -m "add a.dat"
+
+  git checkout -b b
+
+  contents_b="contents_b"
+  contents_b_oid="$(calc_oid $contents_b)"
+  printf "$contents_b" > b.dat
+
+  git add b.dat
+  git commit -m "add b.dat"
+
+  git push origin --all
+
+  pushd ..
+    git \
+      -c "filter.lfs.process=git-lfs filter" \
+      -c "filter.lfs.clean="\
+      -c "filter.lfs.smudge=" \
+      -c "filter.lfs.required=true" \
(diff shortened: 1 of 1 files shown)


test/test-filter-process.sh
```diff
@@ -0,0 +1,77 @@
...
+. "test/testlib.sh"
+
+if [ "$GIT_LFS_USE_LEGACY_FILTER" == "1" ]; then
+  echo "skip: $0 (filter process disabled)"
+  exit
+fi
+
+# HACK(taylor): git uses ".g<hash>" in the version name to signal that it is
+# from the "next" branch, which is the only (current) version of Git that has
+# support for the filter protocol.
```
technoweenie: FYI that `g` prefix is from [`git describe`](https://git-scm.com/docs/git-describe). 

> The "g" prefix stands for "git" and is used to allow describing the version of a software depending on the SCM the software is managed with.

Sorry if you already know that, the comment made it seem like you didn't.  It's not something I've ever encountered in normal git use, just github feature dev :)

ttaylorr:
> Should there be another test for adding files through the filter stream too? 

Definitely! I added one such test in https://github.com/github/git-lfs/pull/1618/commits/bc23d69257f1d5e317e78087b7a2dd1cbe958b27.

> If the version is >= 2.11, it should clear filter.lfs.clean and filter.lfs.smudge manually.

Good call. As I recall, @larsxschneider mentioned to me that Git >=2.11.0 will pick the filter process over the smudge/clean combo, even if all three are given. I think we should still remove the smudge/clean options on older versions of Git, that way we're not relying on Git to pick the right one for us.

> Sorry if you already know that, the comment made it seem like you didn't.

I didn't, thanks for explaining 😄

larsxschneider:
I would vote against "remove the smudge/clean options on older versions of Git" because some people have different versions of Git installed at the same time (e.g. Git installed with OS and Git installed with SourceTree). If we don't keep the smudge/clean filter then Git LFS could appear broken for the older Git version (and some users are not even aware of the fact that they have multiple Git versions...).

The protocol filter taking precedence is hardcoded in Git core. Therefore I think we would be OK if we keep all 3 config values.

larsxschneider:
+1 for "ensure that our build matrix covers versions before/after 2.11". I think we can assume that protocol filter is available on master (BTW: it actually is already - we can leave "next" 🎉).

We could add a build that builds Git 1.9.x or whatever is the minimum requirement for Git LFS from Git source. Would that be OK?

technoweenie:
I think `git lfs install` should continue writing all 3 clean/smudge/process config settings forever. Extra, ignored config settings won't hurt.

As for testing, it may make sense to run ci with all 3 settings too, since that's what `git lfs install` will be doing. A specific 2.11-only test could confirm that `process` is preferred over `clean` or `smudge`. I'm not sure how that would even work though... If somehow Git starts choosing the wrong filter type, things should keep working for end users.

ttaylorr:
> A specific 2.11-only test could confirm that process is preferred over clean or smudge.

Rad, I definitely dig that idea. I think this can all happen in a later PR, so I'd love to merge this one right now if possible 👍

PR #1619, opened 2016-11-02
filter_stream: Protocol unit tests

ttaylorr:
This pull-request adds unit tests ensuring that the Go implementation of the filter protocol is correct.

This PR contains a fair number of changes, so here's the breakdown:

1. https://github.com/github/git-lfs/commit/8bc726dc87d4763d083c4c958842f48c34db6362: Split the high-level protocol (reading requests, writing responses) from the low-level protocol (reading/writing packets, length headers, chunking large packets, etc). This was done so that the unit tests at each "strata" of the protocol could be easily separated, and it hides the underlying implementation of the protocol from the higher-level use cases.
2. https://github.com/github/git-lfs/commit/17f9ead05c0cf3fa81f3d5bbf55a7121b11359bb: add some unit tests for the low-level protocol.
3. https://github.com/github/git-lfs/commit/31d04161d90fc522bcb76e00566eda006979c595: add some unit tests for the high-level protocol.

Though I should have done this sooner, my plan going forward is to rename these types in a future PR to more accurately reflect what they're used for. The `ObjectScanner` type will actually implement the scanner pattern in a future PR in this series, so that name may end up staying the same.

---

### DIFF
--- a/git/filter_protocol.go
+++ b/git/filter_protocol.go
@@ -1,0 +1,114 @@
+package git
+
+import (
+	"bufio"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"strconv"
+	"strings"
+)
+
+type protocol struct {
+	r *bufio.Reader
+	w *bufio.Writer
+}
+
+func newProtocolRW(r io.Reader, w io.Writer) *protocol {
+	return &protocol{
+		r: bufio.NewReader(r),
+		w: bufio.NewWriter(w),
+	}
+}
+
+func (p *protocol) readPacket() ([]byte, error) {
+	pktLenHex, err := ioutil.ReadAll(io.LimitReader(p.r, 4))
+	if err != nil || len(pktLenHex) != 4 { // TODO check pktLenHex length
+		return nil, err
+	}
+
+	pktLen, err := strconv.ParseInt(string(pktLenHex), 16, 0)
+	if err != nil {
+		return nil, err
+	}
+
+	if pktLen == 0 {
+		return nil, nil
+	}
+	if pktLen <= 4 {
+		return nil, errors.New("Invalid packet length.")
+	}
+
+	return ioutil.ReadAll(io.LimitReader(p.r, pktLen-4))
+}
+
+func (p *protocol) readPacketText() (string, error) {
+	data, err := p.readPacket()
+	return strings.TrimSuffix(string(data), "\n"), err
+}
+
+func (p *protocol) readPacketList() ([]string, error) {
+	var list []string
+	for {
+		data, err := p.readPacketText()
+		if err != nil {
+			return nil, err
+		}
+
+		if len(data) == 0 {
+			break
+		}
+
+		list = append(list, data)
+	}
+
+	return list, nil
+}
+
+func (p *protocol) writePacket(data []byte) error {
+	if len(data) > MaxPacketLength {
+		return errors.New("Packet length exceeds maximal length")
+	}
+
+	if _, err := p.w.WriteString(fmt.Sprintf("%04x", len(data)+4)); err != nil {
+		return err
+	}
+
+	if _, err := p.w.Write(data); err != nil {
+		return err
+	}
+
+	if err := p.w.Flush(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *protocol) writeFlush() error {
+	if _, err := p.w.WriteString(fmt.Sprintf("%04x", 0)); err != nil {
+		return err
+	}
+
+	if err := p.w.Flush(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *protocol) writePacketText(data string) error {
+	//TODO: there is probably a more efficient way to do this. worth it?
+	return p.writePacket([]byte(data + "\n"))
+}
+
+func (p *protocol) writePacketList(list []string) error {
+	for _, i := range list {
+		if err := p.writePacketText(i); err != nil {
+			return err
+		}
+	}
+
+	return p.writeFlush()
+}
--- a/git/filter_protocol.go
+++ b/git/filter_protocol.go
@@ -22,13 +22,30 @@ func newProtocolRW(r io.Reader, w io.Writer) *protocol {
 	}
 }
 
+// readPacket reads a single packet entirely and returns the data encoded within
+// it. Errors can occur in several cases, as described below.
+//
+// 1) If no data was present in the reader, and no more data could be read (the
+//    pipe was closed, etc) than an io.EOF will be returned.
+// 2) If there was some data to be read, but the pipe or reader was closed
+//    before an entire packet (or header) could be ingested, an
+//    io.ErrShortBuffer error will be returned.
+// 3) If there was a valid header, but no body associated with the packet, an
+//    "Invalid packet length." error will be returned.
+// 4) If the data in the header could not be parsed as a hexadecimal length in
+//    the Git pktline format, the parse error will be returned.
+//
+// If none of the above cases fit the state of the data on the wire, the packet
+// is returned along with a nil error.
 func (p *protocol) readPacket() ([]byte, error) {
-	pktLenHex, err := ioutil.ReadAll(io.LimitReader(p.r, 4))
-	if err != nil || len(pktLenHex) != 4 { // TODO check pktLenHex length
+	var pktLenHex [4]byte
+	if n, err := io.ReadFull(p.r, pktLenHex[:]); err != nil {
 		return nil, err
+	} else if n != 4 {
+		return nil, io.ErrShortBuffer
 	}
 
-	pktLen, err := strconv.ParseInt(string(pktLenHex), 16, 0)
+	pktLen, err := strconv.ParseInt(string(pktLenHex[:]), 16, 0)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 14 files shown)


ttaylorr:
As a further note, this PR should eventually be merged into #1617, but the diff is more readable when it's based off of #1618. Once the later is merged into the former, the base of this PR will change, and its tests should be ran again before merging.

larsxschneider:
Look what I just found by accident:
https://github.com/bargez/pktline/blob/master/pktline.go

Useful?

git/filter_protocol.go
```diff
@@ -0,0 +1,114 @@
...
+	"bufio"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"strconv"
+	"strings"
+)
+
+type protocol struct {
```
technoweenie: This can probably be renamed in a later PR (or maybe it already has, I haven't gotten that far yet!): `FilterProcessProtocol`?

git/filter_protocol_test.go
```diff
@@ -0,0 +1,211 @@
...
+		0x30, 0x30, 0x30, 0x38, // 0008 (hex. length)
+		0x1, 0x2, 0x3, 0x4, // payload
+	}, buf.Bytes())
+}
+
+func TestFilterProtocolDoesNotWritePacketsExceedingMaxLength(t *testing.T) {
+	var buf bytes.Buffer
+
+	rw := newProtocolRW(nil, &buf)
+	err := rw.writePacket(make([]byte, MaxPacketLength+1))
```
larsxschneider: Would it make sense to check for MaxPacketLength, too?

ttaylorr:
> Look what I just found by accident:
> https://github.com/bargez/pktline/blob/master/pktline.go

Hey! That's pretty neat, I didn't know it existed. Looking at its implementation, I actually like ours a little better. Even though we grow the buffers internally (`read() ([]byte, error)` instead of `read(p []byte) (int, error)`), I think it makes sense to do that when those functions are private (which they are in our case).

Returning appropriately sized buffers makes the calling code a lot simpler, since we don't have to guess how much data we're going to read up front.

I'm fine maintaining our own version of this for now. I figure the smaller we can keep our `glide.yaml` the better, even for small dependencies. I'd hate to have to fork the dependency if we had to change it.

git/filter_protocol.go
ttaylorr: Ah, this should _definitely_ be renamed. I gave this a temporary name when I was working on it last week and never got around to changing it. I'll touch this up in my final PR in the series. 💅

ttaylorr:
My bad on the `<=` vs `<` 4 debacle! I grabbed that code from what was originally there and should have looked into it further. Fixed that right up in https://github.com/github/git-lfs/pull/1619/commits/c23a3838c76063b03432c320ed6ba6c0424c19df.

git/filter_protocol_test.go
ttaylorr: Good call! Added one such test in https://github.com/github/git-lfs/pull/1619/commits/91e54e36f6cca92edf9f13b860fdbac07d987d66.

larsxschneider:
I just realized that actually I messed that up in Git core 😦. I need to fix this:
https://github.com/git/git/blob/master/contrib/long-running-filter/example.pl#L30

ttaylorr:
@larsxschneider no worries. I'm going to remove 30476aa6e3a8d6a936f609ca216f83651877938d and 64e1d75f9f76c6fbb5d8edc16467717dd1814f9f, and we'll be back to where we were before 😄

git/filter_protocol.go
```diff
@@ -0,0 +1,114 @@
...
+		return nil, err
+	}
+
+	pktLen, err := strconv.ParseInt(string(pktLenHex), 16, 0)
+	if err != nil {
+		return nil, err
+	}
+
+	if pktLen == 0 {
+		return nil, nil
```
larsxschneider: How do we distinguish between a "flush" packet here and "no" packet at all?
Do we need that distinction? I am not sure but I think so 🤔

git/filter_protocol.go
```diff
@@ -0,0 +1,114 @@
...
+func (p *protocol) readPacketList() ([]string, error) {
+	var list []string
+	for {
+		data, err := p.readPacketText()
+		if err != nil {
+			return nil, err
+		}
+
+		if len(data) == 0 {
+			break
```
larsxschneider: E.g. here we want to wait for a flush. If there is "no" package then we want to read from the pope, again.

git/filter_protocol.go
```diff
@@ -0,0 +1,114 @@
...
+	if _, err := p.w.WriteString(fmt.Sprintf("%04x", len(data)+4)); err != nil {
+		return err
+	}
+
+	if _, err := p.w.Write(data); err != nil {
+		return err
+	}
+
+	if err := p.w.Flush(); err != nil {
+		return err
```
larsxschneider: Do we want to flush after each packet? Would this have any performance implications if we send a large number of packets (big big file)?

git/filter_protocol_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+	pkt, err := rw.readPacket()
+
+	if len(c.Payload) > 0 {
+		assert.Equal(t, c.Payload, pkt)
+	} else {
+		assert.Empty(t, pkt)
+	}
+
+	if len(c.Err) > 0 {
```
larsxschneider: I am just curious - why do you check 'len' here? Why don't you check against nil?

ttaylorr: `c.Err` is a `string`, and `nil` is not a valid value for that type./

git/filter_protocol.go
ttaylorr: Interesting question. This is how I found the code originally, but I'm wondering if we don't flush after every packet, when should we flush?

git/filter_protocol.go
ttaylorr: As of https://github.com/github/git-lfs/pull/1619/commits/9bc1d68b77ca8ad8dc94161c3a74f88e4d462125, this will _always_ be a flush, since invalid packets will have errors that are handled above, and packets that are only partially present on the wire (or have yet to be sent) will block on the `io.ReadFull` method. 👍

git/filter_protocol.go
ttaylorr: _Fantastic_ question. This kept me busy most of the morning trying to figure out the best way to handle this sort of thing. The result of my search is summarized in https://github.com/github/git-lfs/pull/1619/commits/9bc1d68b77ca8ad8dc94161c3a74f88e4d462125.

The 🔑  to this being clean is that `0004 <EOF>` is an invalid packet, so the only two cases we have to disambiguate are: 

1. Got a packet with a body (or err)
2. Got a flush (or err)

Here's what I thought:
- Use [`io.ReadFull`](https://github.com/github/git-lfs/blob/9bc1d68b77ca8ad8dc94161c3a74f88e4d462125/git/filter_protocol.go#L42) to wait until we have a _full_ header present on the wire before continuing. This means that if we call `readPacket()` and Git hasn't sent us any data over the pipe yet, the call will block (simplifying the call-sites and removing the need to spin waiting for a packet) instead of return early. This is, IMHO, great behavior, since we don't need to add a nth distinction to the return values of this method where we read a partial packet or didn't have any packet at all.
- Return early if we got less bytes than we wanted, meaning the pipe closed with invalid/partial data left on the wire. This happens when `io.ReadFull` encounters an error, halting the read, we can inspect the data that it slurped up, and exit early.
- Utilize the fact that "empty" packets (those beginning with `0004` and no body) are invalid to disambiguate between invalid packets `(nil, err)`, and flush packets `(nil, nil)`.

To summarize what's in the documentation, this method now has four return scenarios:

1. `([]byte, nil)` - an entire packet was read successfully
2. `(nil, io.EOF)` - no more data, the pipe was closed
3.  `(nil, io.ErrShortBuffer)` - partial data was read, but the pipe was closed before the packet's entire data was pushed onto the pipe
4. `(nil, nil)` - empty body (a flush packet) and no error.

(and the last case, which is nil , and some random error if the header couldn't be parsed, or etc.)

git/filter_protocol.go
```diff
@@ -0,0 +1,132 @@
...
+// 3) If there was a valid header, but no body associated with the packet, an
+//    "Invalid packet length." error will be returned.
+// 4) If the data in the header could not be parsed as a hexadecimal length in
+//    the Git pktline format, the parse error will be returned.
+//
+// If none of the above cases fit the state of the data on the wire, the packet
+// is returned along with a nil error.
+func (p *protocol) readPacket() ([]byte, error) {
+	var pktLenHex [4]byte
+	if n, err := io.ReadFull(p.r, pktLenHex[:]); err != nil {
```
sinbad: I think this should have an explicit timeout on it since this is a separate process and could be orphaned if the git process terminated unexpectedly on an OS that doesn't automatically terminate child processes (e.g. Windows). 

You could still use `ReadFull` and just use a signal channel e.g. 
```go
	var n int
	var err error
	sig := make(chan int, 1)
	go func() {
		n, err = io.ReadFull(p.r, pktLenHex[:])
		sig <- 1
	}()
	select {
	case <-sig:
		break
	case <-time.After(time.Second * 30):
		return nil, fmt.Errorf("Timeout reading from git")
	}
	// n & err OK to process
```

PR #1620, opened 2016-11-02
filter_stream: Make ObjectScanner behave like a Scanner

ttaylorr:
This pull-request cleans up the `ObjectScanner` API to better fit the scanner-pattern, an example of which can be found in [`bufio.Scanner`](https://golang.org/pkg/bufio/#Scanner).

Commit by commit, here's the breakdown:

1. https://github.com/github/git-lfs/commit/78e583e48e83f5c3889e58079f0242e9e1736aa1: instead of logging errors directly to `os.Stderr` and returning a `bool` from `Init()`, return those messages in `error`s.
2. https://github.com/github/git-lfs/commit/6728507d5e33e832083e97be1ec53b893795864a: cleanup the `"Error: "` prefixes from the messages, since that formatting will be handled by the `commands` package now.
3. https://github.com/github/git-lfs/commit/18ac2724fd84a76a95ed494a0b3c337974d78ba9, https://github.com/github/git-lfs/commit/b2e568f56d1efb533de01accdd58348e03aa0f65: Do the same thing to `NegotiateCapabilities()`, in the same order.
4. https://github.com/github/git-lfs/commit/c326fcb8c55ae654143d308bfb6a1e145fc1c8f1: Introduce the `*git.Request` type, to clean up the signature of the upcoming `Object()` function to return a type with a `Header` and `Payload` instead of adding more return types, i.e.,:

```go 
type Request struct {
        Header map[string]string
        Payload []byte
}

func (s *ObjectScanner) Object() *Request { ... }
```

instead of:

```go
func (s *ObjectScanner) Object() (map[string]string, []byte)
```

by giving them names, and a meaningful type, I think of that function below is a lot cleaner.

5. https://github.com/github/git-lfs/commit/a778118b384ed352ddf9158e945e688c6e7fbe8a: Implement the scanner itself.

The scanner ends of being a lot cleaner at the call site. Instead of looping forever, and then breaking out at different points in the body, the loop condition encapsulates all of those tiny other exit conditions in one.

The error gets handled once, outside of the loop, which is essentially what was happening in before when it was handled inside of the loop before breaking, now it's just clearer what is actually going on.

This idea was originally a suggestion from @rubyist way back in https://github.com/github/git-lfs/pull/1382#issuecomment-234579799, so credit where credit is due for a great idea!

---

### DIFF
--- a/git/git_filter_protocol.go
+++ b/git/git_filter_protocol.go
@@ -42,23 +42,23 @@ func (o *ObjectScanner) Init() error {
 
 	initMsg, err := o.p.readPacketText()
 	if err != nil {
-		return fmt.Errorf("Error: reading filter initialization failed with %s", err)
+		return fmt.Errorf("reading filter initialization failed with %s", err)
 	}
 	if initMsg != "git-filter-client" {
-		return fmt.Errorf("Error: invalid filter protocol welcome message: %s", initMsg)
+		return fmt.Errorf("invalid filter protocol welcome message: %s", initMsg)
 	}
 
 	supVers, err := o.p.readPacketList()
 	if err != nil {
-		return fmt.Errorf("Error: reading filter versions failed with %s", err)
+		return fmt.Errorf("reading filter versions failed with %s", err)
 	}
 	if !isStringInSlice(supVers, reqVer) {
-		return fmt.Errorf("Error: filter '%s' not supported (your Git supports: %s)", reqVer, supVers)
+		return fmt.Errorf("filter '%s' not supported (your Git supports: %s)", reqVer, supVers)
 	}
 
 	err = o.p.writePacketList([]string{"git-filter-server", reqVer})
 	if err != nil {
-		return fmt.Errorf("Error: writing filter initialization failed with %s", err)
+		return fmt.Errorf("writing filter initialization failed with %s", err)
 	}
 	return nil
 }
--- a/git/git_filter_protocol.go
+++ b/git/git_filter_protocol.go
@@ -67,17 +67,17 @@ func (o *ObjectScanner) NegotiateCapabilities() error {
 
 	supCaps, err := o.p.readPacketList()
 	if err != nil {
-		return fmt.Errorf("Error: reading filter capabilities failed with %s", err)
+		return fmt.Errorf("reading filter capabilities failed with %s", err)
 	}
 	for _, reqCap := range reqCaps {
 		if !isStringInSlice(supCaps, reqCap) {
-			return fmt.Errorf("Error: filter '%s' not supported (your Git supports: %s)", reqCap, supCaps)
+			return fmt.Errorf("filter '%s' not supported (your Git supports: %s)", reqCap, supCaps)
 		}
 	}
 
 	err = o.p.writePacketList(reqCaps)
 	if err != nil {
-		return fmt.Errorf("Error: writing filter capabilities failed with %s", err)
+		return fmt.Errorf("writing filter capabilities failed with %s", err)
 	}
 
 	return nil
--- a/git/git_filter_protocol.go
+++ b/git/git_filter_protocol.go
@@ -5,7 +5,6 @@ package git
 import (
 	"fmt"
 	"io"
-	"os"
 	"strings"
 
 	"github.com/rubyist/tracerx"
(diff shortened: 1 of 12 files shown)


git/git_filter_protocol.go
```diff
@@ -36,103 +38,75 @@ func NewObjectScanner(r io.Reader, w io.Writer) *ObjectScanner {
...
+func (o *ObjectScanner) Init() error {
 	tracerx.Printf("Initialize filter")
 	reqVer := "version=2"
 
 	initMsg, err := o.p.readPacketText()
 	if err != nil {
-		fmt.Fprintf(os.Stderr,
-			"Error: reading filter initialization failed with %s\n", err)
-		return false
+		return fmt.Errorf("reading filter initialization failed with %s", err)
```
technoweenie: Big :+1: for using errors and not a return :ghost: bool here. Any thoughts on using error wrapping instead? It doesn't seem like this would be clobbering any interesting error info though...

git/git_filter_protocol.go
```diff
@@ -154,6 +128,40 @@ func (o *ObjectScanner) WriteResponse(outputData []byte) error {
...
+			// TODO: should we check the err of this call, to?!
+			o.writeStatus("error")
+			return nil, err
+		}
+		if len(chunk) == 0 {
+			break
+		}
+		req.Payload = append(req.Payload, chunk...) // probably more efficient way?!
+	}
+	o.writeStatus("success")
```
technoweenie: Another ignored error...

git/git_filter_protocol.go
```diff
@@ -154,6 +128,40 @@ func (o *ObjectScanner) WriteResponse(outputData []byte) error {
...
+	for _, pair := range requestList {
+		v := strings.Split(pair, "=")
+		req.Header[v[0]] = v[1]
+	}
+
+	for {
+		chunk, err := o.p.readPacket()
+		if err != nil {
+			// TODO: should we check the err of this call, to?!
+			o.writeStatus("error")
```
technoweenie: > should we check the err of this call, to?!

Yes, I think the error should be checked. Or at least logged.

git/git_filter_protocol.go
ttaylorr: Ooh, definitely agree. This was moving some old code around, and I didn't even think to add that. Got it in d484908fb8418dcaedf9a69fb3d1c6087345c6af.

ttaylorr:
Made sure that the errors of calling `writeStatus` were returned properly in https://github.com/github/git-lfs/pull/1620/commits/24cb3890c5f14d8451b22c797421242dfaf74cab! Not a huge fan that an error gets squashed [here](https://github.com/github/git-lfs/blob/24cb3890c5f14d8451b22c797421242dfaf74cab/git/git_filter_protocol.go#L122-L124), but I think if the inner-error happened, we're already in a pretty bad place.

git/object_scanner_test.go
```diff
@@ -228,3 +233,16 @@ func TestObjectScannerWritesResponsesInMultipleChunks(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Equal(t, []string{"status=success"}, status)
 }
+
+// readRequest preforms a single scan operation on the given `*ObjectScanner`,
+// "s", and returns: an error if there was one, or a request if there was one.
+// If neither, it returns (nil, nil).
+func readRequest(s *ObjectScanner) (*Request, error) {
```
larsxschneider: How do we detect EOF? That happens when Git exits and closes the pipe. This event should signal LFS to cleanup and exit the process. I think I remember that io.Reader communicates this with an ErrClosedPipe or something?

canra:
[USTA]

git/git_filter_protocol.go
```diff
@@ -36,122 +39,135 @@ func NewObjectScanner(r io.Reader, w io.Writer) *ObjectScanner {
...
+func (o *ObjectScanner) Err() error        { return o.err }
+
+func (o *ObjectScanner) WriteResponse(outputData []byte) error {
+	for {
+		chunkSize := len(outputData)
+		if chunkSize == 0 {
+			o.p.writeFlush()
+			break
+		} else if chunkSize > MaxPacketLength {
+			chunkSize = MaxPacketLength // TODO check packets with the exact size
```
sinbad: I don't think this TODO is needed any more? Looks OK to me at exact size.

PR #1621, opened 2016-11-02
filter_stream: provide io.Reader implementation

ttaylorr:
This pull-request provides an implementation of the `io.Reader` interface that can read data encoded with the protocol used for filter process communication.

This provides a huge savings on memory usage when cleaning files, as the shasum of the contents can be calculated as data is streamed in, and then discarded.

This pull request is implemented in three parts:

- https://github.com/github/git-lfs/commit/4524d1221efc54d82f4a409208e0d83a56a1c78e: Implement the underlying `PacketReader`.
- https://github.com/github/git-lfs/commit/d23c0445f4580d6b82345ae3613bf47ecaff482a: Make the `WriteStatus` method public since reading packets no longer reports success/error statuses, and this responsibility will be delegated to the caller in the following commit...
- https://github.com/github/git-lfs/commit/2db63047c028811deeab1b8462e44251304164ca: use an `io.Reader` as the `Payload` field of a request, instead of buffering the data upfront.

Some notes on the implementation:

- There is no "offset" encoded into where the reader starts and stops. This means that a given packet run must be _completely_ consumed before the next one is read. The code as it currently stands within this branch, does obey that rule as the synchronization is done with the loop condition, `scanner.Scan()`.
- The implementation is safe to call with buffers shorter or longer than the packet that is being read. Any data that wont fit into the current given buffer will be sent to an internal buffer, and subsequent calls to `Read(p []byte)` will read first from the internal buffer until the internal buffer is drained.
- An `io.EOF` is returned on the first call to `Read()` _after_ all of the data has been consumed. This is done to avoid a complicated implementation dealing with peeking bytes to see if more data is incoming. Luckily, the Go documentation allows for this (meaning things like `io.Copy`, etc will respect it):

> An instance of this general case is that a Reader returning a non-zero number of bytes at the end of the input stream may return either err == EOF or err == nil. The next Read should return 0, EOF.

(See: https://golang.org/pkg/io#Reader).

---

/cc @larsxschneider @technoweenie @sinbad @rubyist @sschuberth

### DIFF
--- a/commands/command_filter.go
+++ b/commands/command_filter.go
@@ -140,24 +140,32 @@ func filterCommand(cmd *cobra.Command, args []string) {
 		ExitWithError(err)
 	}
 
+Scan:
 	for s.Scan() {
 		req := s.Request()
 
 		// TODO:
-		// ReadRequest should return data as Reader instead of []byte ?!
 		// clean/smudge should also take a Writer instead of returning []byte
 		var outputData []byte
+		var err error
+
 		switch req.Header["command"] {
 		case "clean":
-			outputData, _ = clean(bytes.NewReader(req.Payload), req.Header["pathname"])
+			outputData, err = clean(req.Payload, req.Header["pathname"])
 		case "smudge":
-			outputData, _ = smudge(bytes.NewReader(req.Payload), req.Header["pathname"])
+			outputData, err = smudge(req.Payload, req.Header["pathname"])
 		default:
 			fmt.Errorf("Unknown command %s", cmd)
-			break
+			break Scan
 		}
 
-		s.WriteResponse(outputData)
+		if err != nil {
+			s.WriteStatus("error")
+
+		} else {
+			s.WriteStatus("success")
+			s.WriteResponse(outputData)
+		}
 	}
 
 	if err := s.Err(); err != nil {
--- a/commands/command_filter.go
+++ b/commands/command_filter.go
@@ -161,7 +161,6 @@ Scan:
 
 		if err != nil {
 			s.WriteStatus("error")
-
 		} else {
 			s.WriteStatus("success")
 			s.WriteResponse(outputData)
--- a/commands/command_filter.go
+++ b/commands/command_filter.go
@@ -79,16 +79,12 @@ func clean(reader io.Reader, fileName string) ([]byte, error) {
 }
 
 func smudge(reader io.Reader, filename string) ([]byte, error) {
+	var pbuf bytes.Buffer
+	reader = io.TeeReader(reader, &pbuf)
+
 	ptr, err := lfs.DecodePointer(reader)
 	if err != nil {
-		// mr := io.MultiReader(b, reader)
-		// _, err := io.Copy(os.Stdout, mr)
-		// if err != nil {
-		// 	Panic(err, "Error writing data to stdout:")
-		// }
-		var content []byte
-		reader.Read(content)
-		return content, nil
+		return pbuf.Bytes(), err
 	}
 
 	lfs.LinkOrCopyFromReference(ptr.Oid, ptr.Size)
--- a/commands/command_filter.go
+++ b/commands/command_filter.go
@@ -84,6 +84,18 @@ func smudge(reader io.Reader, filename string) ([]byte, error) {
 
 	ptr, err := lfs.DecodePointer(reader)
 	if err != nil {
+		// If we tried to decode a pointer out of the data given to us,
+		// and the file was _empty_, write out an empty file in
+		// response. This occurs because when the clean filter
+		// encounters an empty file, and writes out an empty file,
+		// instead of a pointer.
+		//
+		// TODO(taylor): figure out if there is more data on the reader,
+		// and buffer that as well.
+		if len(pbuf.Bytes()) == 0 {
+			return pbuf.Bytes(), nil
+		}
+
 		return pbuf.Bytes(), err
 	}
 
@@ -155,7 +167,7 @@ Scan:
 			break Scan
 		}
 
-		if err != nil {
+		if err != nil && err != io.EOF {
 			s.WriteStatus("error")
 		} else {
 			s.WriteStatus("success")
--- a/commands/command_filter.go
+++ b/commands/command_filter.go
@@ -14,11 +14,24 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	// cleanFilterBufferCapacity is the desired capacity of the
+	// `*git.PacketWriter`'s internal buffer when the filter protocol
+	// dictates the "clean" command. 512 bytes is (in most cases) enough to
+	// hold an entire LFS pointer in memory.
+	cleanFilterBufferCapacity = 512
+
+	// smudgeFilterBufferCapacity is the desired capacity of the
+	// `*git.PacketWriter`'s internal buffer when the filter protocol
+	// dictates the "smudge" command.
+	smudgeFilterBufferCapacity = git.MaxPacketLength
+)
+
 var (
 	filterSmudgeSkip = false
 )
 
-func clean(reader io.Reader, fileName string) ([]byte, error) {
+func clean(to io.Writer, reader io.Reader, fileName string) error {
 	var cb progress.CopyCallback
 	var file *os.File
 	var fileSize int64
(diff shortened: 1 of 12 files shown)


commands/command_filter.go
```diff
@@ -140,24 +140,32 @@ func filterCommand(cmd *cobra.Command, args []string) {
...
 		case "clean":
-			outputData, _ = clean(bytes.NewReader(req.Payload), req.Header["pathname"])
+			outputData, err = clean(req.Payload, req.Header["pathname"])
 		case "smudge":
-			outputData, _ = smudge(bytes.NewReader(req.Payload), req.Header["pathname"])
+			outputData, err = smudge(req.Payload, req.Header["pathname"])
 		default:
 			fmt.Errorf("Unknown command %s", cmd)
-			break
+			break Scan
```
larsxschneider: Is this golang's "goto"?

git/packet_reader.go
```diff
@@ -0,0 +1,62 @@
...
+
+var _ io.Reader = new(packetReader)
+
+func (r *packetReader) Read(p []byte) (int, error) {
+	var n int
+
+	if len(r.buf) > 0 {
+		// If there is data in the buffer, shift as much out of it and
+		// into the given "p" as we can.
+		n = int(math.Min(float64(len(p)), float64(len(r.buf))))
```
larsxschneider: Why do we cast to float64 here?

commands/command_filter.go
ttaylorr: Sort of. This is a [label](https://golang.org/ref/spec#Labeled_statements) so I can `break`/`continue` into a different loop than the enclosing one at the callsite. Useful for things like this 👍

git/packet_reader.go
ttaylorr: `math.Min` accepts two float64s as its arguments, and the `len` builtin returns an `int`, so the cast must be preformed 😭

commands/command_filter.go
```diff
@@ -140,24 +140,32 @@ func filterCommand(cmd *cobra.Command, args []string) {
...
 		}
 
-		s.WriteResponse(outputData)
+		if err != nil {
+			s.WriteStatus("error")
+
```
sschuberth: Nit: Superfluous blank line.

git/packet_reader.go
sschuberth: Doing these casts is not a good idea as outlined [here](http://mrekucci.blogspot.de/2015/07/dont-abuse-mathmax-mathmin.html). I guess we're better off having something like

    n = len(r.buf)
    if n > 0 {
        if n > len(p) {
            n = len(p)
        }
        // ...
    }

git/packet_reader.go
```diff
@@ -0,0 +1,62 @@
...
+		if len(chunk) == 0 {
+			// If we got an empty chunk, then we know that we have
+			// reached the end of processing for this particular
+			// packet, so let's terminate.
+
+			return n, io.EOF
+		}
+
+		// Figure out how much of the packet we can read into "p".
+		nn := int(math.Min(float64(len(chunk)), float64(len(p))))
```
sschuberth: Same thing here about the casts.

commands/command_filter.go
sinbad: You could have just used `continue` since that doesn't interact with `switch`.

git/packet_reader.go
ttaylorr: Nice catch. Though there aren't any mismatches under MaxPacketLength (see: https://play.golang.org/p/hUhRZhv3n1), I think this is still a good thing to guard against.

I added two functions to the `tools` packet, `MinInt(a, b int) int` and `MaxInt(a, b int) int` in https://github.com/github/git-lfs/pull/1621/commits/7dea0af323d96c60cbd5e508c629a4bb209d01de and used them in both places here.

ttaylorr:
I fixed the hang in the integration test from yesterday, but there's a new open question that I'd love some feedback on. The test that is currently broken (and what is causing CI to go 🔴 ) is the `pull zero len file` test.

This test is broken because LFS tries to write the zero-len file out to the filter protocol, but the protocol code rejects it because packets with no body are invalid (according to: https://github.com/github/git-lfs/pull/1619#issuecomment-258039097 and https://github.com/github/git-lfs/pull/1619#issuecomment-258039532).

So long as that assumption is correct, I think we have two options:

1. Fall back to the legacy clean/smudge process instead of the filter protocol, which would allow users to track empty files with LFS (not sure if this is useful), but binds us to use legacy APIs which may not be supported in all versions of Git going forward.
2. We could also just specify that tracking empty files is not allowed and just enforce that requirement earlier on in LFS, so that this situation doesn't happen. I think this course of action makes more sense.

In the case that I misread @larsxschneider's comment and my previous assumptions about lengths were incorrect, then I'll have to re-do some work as described in https://github.com/github/git-lfs/pull/1619#discussion_r86431429, which will propagate its way back to this PR and make this a non-issue.

Curious as to your thoughts.

sschuberth:
> We could also just specify that tracking empty files is not allowed and just enforce that requirement earlier on in LFS

I agree here. **L**FS is for **large** files after all, and zero is not large ;-)

larsxschneider:
Your assumption was indeed correct and pktline shouldn't send `0004`. I was wrong and I am sorry that I confused you. That also explains why [my demo code](https://github.com/git/git/blob/master/contrib/long-running-filter/example.pl#L30) considers `0004` invalid. I just wrote that too long ago to remember correctly.

---

*Re: 1.*
I think the new protocol should be able to handle everything. If we would need the legacy protocol in any way then there would be something wrong with the new protocol.

I actually test the case of an empty file in the protocol and therefore it should work:
https://github.com/git/git/blob/be5a750939c212bc0781ffa04fabcfd2b2bd744e/t/t0021-conversion.sh#L375

One trick to see what is going on is to run the tests with `GIT_TRACE=1 GIT_TRACE_PACKET=1`.
This is how an empty file is processed:

```
packet:          git> command=clean            # Git wants to clean
packet:          git> pathname=test4-empty.r   # ... a file with this name
packet:          git> 0000                     # end of request header
packet:          git> 0000                     # end of content (= no content!)
packet:          git< status=success           # filter sets status success
packet:          git< 0000                     # end of status
packet:          git< 0000                     # end of cleaned content (= no content!) 
packet:          git< 0000                     # end of status (no status change)
```

Would that work for you?

*Re: 2.*
An empty file is certainly not large. However, people use the `*.bin` pattern and this can easily catch empty files (I saw this many times already @Autodesk).

princess75:
On Nov 6, 2016 5:27 PM, "Lars Schneider" <notifications@github.com> wrote:
>
> Your assumption was indeed correct and pktline shouldn't send 0004. I was
wrong and I am sorry that I confused you. That also explains why my demo
code considers 0004 invalid. I just wrote that too long ago to remember
correctly.
>
> ________________________________
>
> Re: 1.
> I think the new protocol should be able to handle everything. If we would
need the legacy protocol in any way then there would be something wrong
with the new protocol.
>
> I actually test the case of an empty file in the protocol and therefore
it should work:
>
https://github.com/git/git/blob/be5a750939c212bc0781ffa04fabcfd2b2bd744e/t/t0021-conversion.sh#L375
>
> One trick to see what is going on is to run the tests with GIT_TRACE=1
GIT_TRACE_PACKET=1.
> This is how an empty file is processed:
>
> packet: git> command=clean # Git wants to clean packet: git>
pathname=test4-empty.r # ... a file with this name packet: git> 0000 # end
of request header packet: git> 0000 # end of content (= no content!)
packet: git< status=success # filter sets status success packet: git< 0000
# end of status packet: git< 0000 # end of cleaned content (= no content!)
packet: git< 0000 # end of status (no status change)
>
> Would that work for you?
>
> Re: 2.
> An empty file is certainly not large. However, people use the *.bin
pattern and this can easily catch empty files (I saw this many times
already @Autodesk).

ttaylorr:
> One trick to see what is going on is to run the tests with GIT_TRACE=1 GIT_TRACE_PACKET=1.

This is awesome, thanks for letting me know it exists.

> packet:          git< 0000                     # end of cleaned content (= no content!)

Makes sense to me, I can make this work! Thanks for the info 🤘

PR #1622, opened 2016-11-03
filter_stream: provide io.Writer implementation

ttaylorr:
This pull-request provides an implementation of the `io.Writer` interface capable of writing packets using the Git "pktline" format to an underlying datastream.

Here's a 🔎  at some the different parts of this PR:

1. https://github.com/github/git-lfs/commit/5723bb8c016d422fa098fcdaaf1dd15e146c3e42: implement the `PacketWriter` type and add an exhaustive set of unit tests.
2. https://github.com/github/git-lfs/commit/8ab75794997d64c92dc7649abb216ce33e32fbe4: add public constructor `NewPacketWriter` which takes any `io.Writer` and returns a `*PacketWriter` wrapping that underneath. If the given writer is itself already a `*PacketWriter`, it returns that wholesale to avoid re-wrapping.
3. https://github.com/github/git-lfs/commit/a90c16285bc98ae8f7e71fea832494eeabb0caf2: Use the `io.Writer` in the `clean`/`smudge` implementation in `commands/commands_filter.go`.
4. https://github.com/github/git-lfs/commit/5e72db6a3541b21d9bbb1981e09a981a8827f26f: A quick bit of 💅  to use the new `tools.MinInt` func instead of casting `int`s to `float64`s and relying on (potentially) inaccurate results from `math.Min`.

In terms of the implementation of the `io.Writer` itself, here's a quick tour:

- Since the pktline protocol expects us to write as much data as we can in a single packet (as many bytes as we can up to and including `MaxPacketLength`), we have an internal buffer that is written to until we have ingested enough data to fill a single packet. 
- To avoid leaving data in the buffer at the end of writing a chunk of data, calling `w.Write(nil)` is a special-case signaling to write the `0000` flush packet.
- `Write(p []byte)` reports only the amount of _data_ (not including the 4-byte headers) _actually_ written to the stream. This means that `Write()` with less data than `MaxPacketLength` will return `(0, nil)`, unless there was enough data in the buffer to cause a packet to be written.
- `Write` is smart enough to write multiple packets if the given buffer "p" holds more information than a single packet can hold. It will buffer extra data that does not fit past the first `n` packets.

---

### DIFF
--- a/commands/command_filter.go
+++ b/commands/command_filter.go
@@ -131,8 +131,7 @@ func smudge(to io.Writer, reader io.Reader, filename string) error {
 		download = false
 	}
 
-	sbuf := new(bytes.Buffer)
-	err = ptr.Smudge(sbuf, filename, download, TransferManifest(), cb)
+	err = ptr.Smudge(to, filename, download, TransferManifest(), cb)
 	if file != nil {
 		file.Close()
 	}
@@ -151,8 +150,7 @@ func smudge(to io.Writer, reader io.Reader, filename string) error {
 		return err
 	}
 
-	_, err = ptr.Encode(to)
-	return err
+	return nil
 }
 
 func filterCommand(cmd *cobra.Command, args []string) {
--- a/commands/command_filter.go
+++ b/commands/command_filter.go
@@ -171,7 +171,13 @@ Scan:
 	for s.Scan() {
 		var err error
 
-		switch req := s.Request(); req.Header["command"] {
+		req := s.Request()
+		if req == nil {
+			break
+		}
+		s.WriteStatus("success")
+
+		switch req.Header["command"] {
 		case "clean":
 			err = clean(w, req.Payload, req.Header["pathname"])
 		case "smudge":
--- a/commands/command_filter.go
+++ b/commands/command_filter.go
@@ -17,12 +17,13 @@ import (
 const (
 	// cleanFilterBufferCapacity is the desired capacity of the
 	// `*git.PacketWriter`'s internal buffer when the filter protocol
-	// dictates the "clean" command
+	// dictates the "clean" command. 512 bytes is (in most cases) enough to
+	// hold an entire LFS pointer in memory.
 	cleanFilterBufferCapacity = 512
 
 	// smudgeFilterBufferCapacity is the desired capacity of the
 	// `*git.PacketWriter`'s internal buffer when the filter protocol
-	// dictates the "smudge" command
+	// dictates the "smudge" command.
 	smudgeFilterBufferCapacity = git.MaxPacketLength
 )
 
@@ -158,7 +159,6 @@ func filterCommand(cmd *cobra.Command, args []string) {
 	lfs.InstallHooks(false)
 
 	s := git.NewObjectScanner(os.Stdin, os.Stdout)
-	w := git.NewPacketWriter(os.Stdout, 0)
 
 	if err := s.Init(); err != nil {
 		ExitWithError(err)
@@ -170,6 +170,7 @@ func filterCommand(cmd *cobra.Command, args []string) {
 Scan:
 	for s.Scan() {
 		var err error
+		var w io.Writer
 
 		req := s.Request()
 		if req == nil {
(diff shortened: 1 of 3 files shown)


canra:
[USTA]

commands/command_filter.go
```diff
@@ -138,29 +160,32 @@ func filterCommand(cmd *cobra.Command, args []string) {
...
+			w = git.NewPacketWriter(os.Stdout, smudgeFilterBufferCapacity)
+			err = smudge(w, req.Payload, req.Header["pathname"])
 		default:
 			fmt.Errorf("Unknown command %s", cmd)
 			break Scan
 		}
 
+		if err == nil {
+			_, err = w.Write(nil)
+		}
```
larsxschneider: I assume we flush the content here?

commands/command_filter.go
```diff
@@ -138,29 +160,32 @@ func filterCommand(cmd *cobra.Command, args []string) {
...
+		var status string
 		if err != nil {
-			s.WriteStatus("error")
+			status = "error"
 		} else {
-			s.WriteStatus("success")
-			s.WriteResponse(outputData)
+			status = "success"
 		}
+		s.WriteStatus(status)
```
larsxschneider: We set the final (2nd!) status ("after write") here, correct? I am not sure right now where the status "after read" is set. Shouldn't that be in a similar scope?

git/packet_writer.go
```diff
@@ -0,0 +1,124 @@
...
+
+import (
+	"io"
+
+	"github.com/github/git-lfs/tools"
+)
+
+type PacketWriter struct {
+	// buf is the internal buffer of bytes used to store data given in calls
+	// to Write that don't warrant a protocol write.
```
larsxschneider: "...don't warrant a protocol write." --> might be the late hour or my non-native english ... but I don't get this sentence 🤔

git/packet_writer.go
```diff
@@ -0,0 +1,124 @@
...
+
+// NewPacketWriter returns a new *PacketWriter, which will write to the
+// underlying data stream "w". The internal buffer is initialized with the given
+// capacity, "c".
+//
+// If "w" is already a `*PacketWriter`, it will be returned as-is.
+func NewPacketWriter(w io.Writer, c int) *PacketWriter {
+	if pw, ok := w.(*PacketWriter); ok {
+		return pw
+	}
```
larsxschneider: When would this happen?

ttaylorr:
Huge thank you to @larsxschneider for pairing with me on: df9514a1384336f6a7089063f94dabd3f2379819...4c1615432a126826717d6a632486198aad8faaec.

commands/command_filter.go
```diff
@@ -49,7 +61,10 @@ func clean(reader io.Reader, fileName string) ([]byte, error) {
 	if errors.IsCleanPointerError(err) {
 		// TODO: report errors differently!
 		// os.Stdout.Write(errors.GetContext(err, "bytes").([]byte))
-		return errors.GetContext(err, "bytes").([]byte), nil
+		// return errors.GetContext(err, "bytes").([]byte), nil
+
+		// TODO(taylor): what does this mean?
```
ttaylorr: 👍  this came up during our pairing and was addressed in https://github.com/github/git-lfs/pull/1622/commits/1d2b5ce985cb508c9d25ab5029561227e66d7b9f.

commands/command_filter.go
ttaylorr: I think this was originally incorrect but has since been fixed in https://github.com/github/git-lfs/pull/1622/commits/4c1615432a126826717d6a632486198aad8faaec and https://github.com/github/git-lfs/pull/1622/commits/10660fc0b8f3eee4fa77a7979470d9de4496ee50.

git/packet_writer.go
ttaylorr: Definitely confusing! Cleaned it up in https://github.com/github/git-lfs/pull/1622/commits/7be86e55842f9ca4805484a0163399845e1bbae8.

git/packet_writer.go
ttaylorr: If a caller (not us, since we never make this mistake) passes tries to wrap an `io.Writer` that _is_ underneath itself a `*git.PacketWriter` to this constructor, we return that writer `w` instead of re-wrapping it and producing invalid sequences of bytes.

PR #1625, opened 2016-11-07
Locking part 1: locking package

sinbad:
Part 1 resubmission of #1590

This refactors logic into a dedicated package called `locking`, so that secondary functionality can be added in a re-usable place. Makes calls to a lock cache which will be implemented in a later PR (a replacement for BoltDB which was used in #1590).

Also involved moving `ChannelWrapper` into tools since it is a useful abstraction of working with parallel result and error channels while avoiding mistakes; used for the refactored `SearchLocks` so streamed results are avaialble if needed.

### DIFF
--- a/locking/locks.go
+++ b/locking/locks.go
@@ -1,0 +1,198 @@
+package locking
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/github/git-lfs/api"
+	"github.com/github/git-lfs/config"
+	"github.com/github/git-lfs/git"
+	"github.com/github/git-lfs/tools"
+)
+
+var (
+	// API is a package-local instance of the API client for use within
+	// various command implementations.
+	API = api.NewClient(nil)
+	// errNoMatchingLocks is an error returned when no matching locks were
+	// able to be resolved
+	errNoMatchingLocks = errors.New("lfs: no matching locks found")
+	// errLockAmbiguous is an error returned when multiple matching locks
+	// were found
+	errLockAmbiguous = errors.New("lfs: multiple locks found; ambiguous")
+)
+
+// Lock attempts to lock a file on the given remote name
+// path must be relative to the root of the repository
+// Returns the lock id if successful, or an error
+func Lock(path, remote string) (id string, e error) {
+	// TODO: API currently relies on config.Config but should really pass to client in future
+	savedRemote := config.Config.CurrentRemote
+	config.Config.CurrentRemote = remote
+	defer func() { config.Config.CurrentRemote = savedRemote }()
+
+	// TODO: this is not really the constraint we need to avoid merges, improve as per proposal
+	latest, err := git.CurrentRemoteRef()
+	if err != nil {
+		return "", err
+	}
+
+	s, resp := API.Locks.Lock(&api.LockRequest{
+		Path:               path,
+		Committer:          api.CurrentCommitter(),
+		LatestRemoteCommit: latest.Sha,
+	})
+
+	if _, err := API.Do(s); err != nil {
+		return "", fmt.Errorf("Error communicating with LFS API: %v", err)
+	}
+
+	if len(resp.Err) > 0 {
+		return "", fmt.Errorf("Server unable to create lock: %v", resp.Err)
+	}
+
+	if err := cacheLock(resp.Lock.Path, resp.Lock.Id); err != nil {
+		return "", fmt.Errorf("Error caching lock information: %v", err)
+	}
+
+	return resp.Lock.Id, nil
+}
+
+// Unlock attempts to unlock a file on the given remote name
+// path must be relative to the root of the repository
+// Force causes the file to be unlocked from other users as well
+func Unlock(path, remote string, force bool) error {
+	// TODO: API currently relies on config.Config but should really pass to client in future
+	savedRemote := config.Config.CurrentRemote
+	config.Config.CurrentRemote = remote
+	defer func() { config.Config.CurrentRemote = savedRemote }()
+
+	id, err := lockIdFromPath(path)
+	if err != nil {
+		return fmt.Errorf("Unable to get lock id: %v", err)
+	}
+
+	return UnlockById(id, remote, force)
+
+}
+
+// Unlock attempts to unlock a lock with a given id on the remote
+// Force causes the file to be unlocked from other users as well
+func UnlockById(id, remote string, force bool) error {
+	s, resp := API.Locks.Unlock(id, force)
+
+	if _, err := API.Do(s); err != nil {
+		return fmt.Errorf("Error communicating with LFS API: %v", err)
+	}
+
+	if len(resp.Err) > 0 {
+		return fmt.Errorf("Server unable to unlock lock: %v", resp.Err)
+	}
+
+	if err := cacheUnlockById(id); err != nil {
+		return fmt.Errorf("Error caching unlock information: %v", err)
+	}
+
+	return nil
+}
+
+// ChannelWrapper for lock search to more easily return async error data via Wait()
+// See NewPointerChannelWrapper for construction / use
+type LockChannelWrapper struct {
+	*tools.BaseChannelWrapper
+	Results <-chan api.Lock
+}
+
+// Construct a new channel wrapper for api.Lock
+// Caller can use s.Results directly for normal processing then call Wait() to finish & check for errors
+func NewLockChannelWrapper(lockChan <-chan api.Lock, errChan <-chan error) *LockChannelWrapper {
+	return &LockChannelWrapper{tools.NewBaseChannelWrapper(errChan), lockChan}
+}
+
+// SearchLocks returns a channel of locks which match the given name/value filter
+// If limit > 0 then search stops at that number of locks
+func SearchLocks(remote string, filter map[string]string, limit int) *LockChannelWrapper {
+	// TODO: API currently relies on config.Config but should really pass to client in future
+	savedRemote := config.Config.CurrentRemote
+	config.Config.CurrentRemote = remote
+	errChan := make(chan error, 5) // can be multiple errors below
+	lockChan := make(chan api.Lock, 10)
+	c := NewLockChannelWrapper(lockChan, errChan)
+	go func() {
+		defer func() {
+			close(lockChan)
+			close(errChan)
+			// Only reinstate the remote after we're done
+			config.Config.CurrentRemote = savedRemote
+		}()
+
+		apifilters := make([]api.Filter, 0, len(filter))
+		for k, v := range filter {
+			apifilters = append(apifilters, api.Filter{k, v})
+		}
+		lockCount := 0
+		query := &api.LockSearchRequest{Filters: apifilters}
+	QueryLoop:
+		for {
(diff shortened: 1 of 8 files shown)


technoweenie:
Since this is adding a new `locking` package, I want it to have a well designed API that external users can depend on. Can we design something with a `locking.Client` object that has per-command properties like `CurrentRemote`, and some lock caching implementation? 

I realize at this point, it's a pipe dream, since both the `api` and `auth` packages use the `*config.Configuration`.

sinbad:
I’d planned that for a later PR when I introduce the caching (started the bolt replacement already), trying to chop it up into smaller incremental changes.

locking/locks.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/github/git-lfs/api"
+	"github.com/github/git-lfs/config"
+	"github.com/github/git-lfs/git"
+	"github.com/github/git-lfs/tools"
+)
+
+var (
+	// API is a package-local instance of the API client for use within
+	// various command implementations.
+	API = api.NewClient(nil)
```
ttaylorr: Is there a reason that this is public?

locking/locks.go
```diff
@@ -0,0 +1,198 @@
...
+	// errNoMatchingLocks is an error returned when no matching locks were
+	// able to be resolved
+	errNoMatchingLocks = errors.New("lfs: no matching locks found")
```
ttaylorr: I'd be 👍  increasing the surface-area of this package's API and exporting these as sentinel error values.

locking/locks.go
```diff
@@ -0,0 +1,198 @@
...
+	// were found
+	errLockAmbiguous = errors.New("lfs: multiple locks found; ambiguous")
+)
+
+// Lock attempts to lock a file on the given remote name
+// path must be relative to the root of the repository
+// Returns the lock id if successful, or an error
+func Lock(path, remote string) (id string, e error) {
+	// TODO: API currently relies on config.Config but should really pass to client in future
+	savedRemote := config.Config.CurrentRemote
```
ttaylorr: Could you pull out a temporary variable:

```go
cfg := config.Config
```

to signal that references to `cfg` should be replaced with a parameter, or pass the `*config.Configuration` instance explicitly? This is a pattern that we use throughout other parts of LFS to signal that the usage of that config should be replaced with a more specific type that holds only the data that it needs, _or_ that the usage of that configuration instance is not necessarily bound to the global one.

I believe that both of those situations apply here.

locking/locks.go
```diff
@@ -0,0 +1,198 @@
...
+
+	return resp.Lock.Id, nil
+}
+
+// Unlock attempts to unlock a file on the given remote name
+// path must be relative to the root of the repository
+// Force causes the file to be unlocked from other users as well
+func Unlock(path, remote string, force bool) error {
+	// TODO: API currently relies on config.Config but should really pass to client in future
+	savedRemote := config.Config.CurrentRemote
```
ttaylorr: Same note about the `config.Configuration` use here.

locking/locks.go
```diff
@@ -0,0 +1,198 @@
...
+	config.Config.CurrentRemote = remote
```
ttaylorr: This is not safe to use across multiple goroutines. Ideally we should fix this and either:

- duplicate the configuration instance
- extract a type that holds this information and has only one local instance
- have some sort of locking/coordination mechanism guarding (a `sync.RWMutex` or similar) read/write access to this instance.

If neither of those seem immediately feasible or you'd rather focus on other stuff for the time being, let's at least document this behavior in the method's godoc.

locking/locks.go
```diff
@@ -0,0 +1,198 @@
...
+	if err := cacheUnlockById(id); err != nil {
+		return fmt.Errorf("Error caching unlock information: %v", err)
+	}
+
+	return nil
+}
+
+// ChannelWrapper for lock search to more easily return async error data via Wait()
+// See NewPointerChannelWrapper for construction / use
+type LockChannelWrapper struct {
```
ttaylorr: I am still not totally sold on this `<T>ChannelWrapper` pattern. I think introducing this new `locking` package is a great call, but I'd love if right off the bat we could make better promises about the stability of it.

My biggest concern is that public consumers of this package expect results from functions that return one of these channel wrappers in a different way that is provided. In other words, I'm not sure if they're flexible enough.

From what I can gather, this type provides two APIs:

- `Results <-chan api.Lock`
- `Wait() error`

What do you think about returning the `<-chan api.Lock` wholesale (allowing callers to interact with it asynchonously as before, and exporting a function in `tools` a-la:

```go
package tools

func CollectErrsFromChan(errs <-chan error) error
```

that provides the same behavior as `Wait() error`?

I think this solution would allow callers/consumers of this method to handle errors in a synchronous _or_ asynchronous manner, which would provide a sufficient amount of flexibility.

locking/locks.go
```diff
@@ -0,0 +1,198 @@
...
+// Caller can use s.Results directly for normal processing then call Wait() to finish & check for errors
+func NewLockChannelWrapper(lockChan <-chan api.Lock, errChan <-chan error) *LockChannelWrapper {
+	return &LockChannelWrapper{tools.NewBaseChannelWrapper(errChan), lockChan}
+}
+
+// SearchLocks returns a channel of locks which match the given name/value filter
+// If limit > 0 then search stops at that number of locks
+func SearchLocks(remote string, filter map[string]string, limit int) *LockChannelWrapper {
+	// TODO: API currently relies on config.Config but should really pass to client in future
+	savedRemote := config.Config.CurrentRemote
```
ttaylorr: Same note here about the `config.Config` usage.

locking/locks.go
```diff
@@ -0,0 +1,198 @@
...
+	config.Config.CurrentRemote = remote
+	errChan := make(chan error, 5) // can be multiple errors below
+	lockChan := make(chan api.Lock, 10)
+	c := NewLockChannelWrapper(lockChan, errChan)
+	go func() {
```
ttaylorr: What do you think about extracting this goroutine out to a separate (private) function to add some clarity and cut down on the indentation? I'm thinking:

```go
func scanLocks(offsetId string, locks chan<- api.Lock, errs chan<- error) { ... }
```

This function could also recurse and handle the offset in a really nice way.

locking/locks.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// SearchLocks returns a channel of locks which match the given name/value filter
+// If limit > 0 then search stops at that number of locks
+func SearchLocks(remote string, filter map[string]string, limit int) *LockChannelWrapper {
+	// TODO: API currently relies on config.Config but should really pass to client in future
+	savedRemote := config.Config.CurrentRemote
+	config.Config.CurrentRemote = remote
+	errChan := make(chan error, 5) // can be multiple errors below
+	lockChan := make(chan api.Lock, 10)
```
ttaylorr: Why 5 and 10?

locking/locks.go
sinbad: OK - my experience is that a lot of people aren't immediately comfortable with the 'dual asynchronous channel' pattern and that mistakes get made quite easily; the channel wrapper is intended to make consuming from it idiomatically simple and less prone to error. 

Maybe I can think of a way to make default use simple & error free but allowing manual access to the underlying channels possible if desired. Although realistically I don't see the benefit; you can already process primary results in parallel and processing errors in parallel is rarely very useful.

locking/locks.go
sinbad: Again, taken from the previous implementation. I really need to change the API to properly decouple this as mentioned in the TODO. Fair point though since this is a supposedly re-usable package now, maybe this first step needs to include more changes after all.

locking/locks.go
sinbad: Arbitrary; > 1 because parallel, errors less likely than results. What would you have used?

sinbad:
My take away summary: step 1 needs to be more than a simple move refactor. I'll have to come back to this later.

locking/locks.go
ttaylorr: Sounds good, I'll keep my 👀  peeled. My motivation--one I should have mentioned earlier--was that I'd like to provide the flexibility for people to handle these errors on a case-by-case basis, halting the goroutine/worker/etc if an error is deemed "fatal", and ignoring it to move on otherwise.

locking/locks.go
ttaylorr: I'd remove the buffer, since it is not obvious to the caller when this function will block and when it wont.

ttaylorr:
> My take away summary: step 1 needs to be more than a simple move refactor. I'll have to come back to this later.

Thanks, and sorry for some of these heavier :bomb:s. The only reason I want to address these now rather than later is that this PR is targeted against master.

sinbad:
> The only reason I want to address these now rather than later is that this PR is targeted against master.

OK, so if I built a major feature branch then submitted individual stage PRs against that, with the final PR to master just being one big merge-of-merges, you'd be OK with that? I think that will be more manageable as I'm stuck between trying to make PRs that are simultaneously small enough / incremental enough to review, and acceptable at a higher level design wise.

technoweenie:
IMO that approach did not work well at all with the big filter process PR (#1382). At one point I was making a review comment in PR 3 that was already fixed in PR 5. 

I'd prefer a branch that merges master multiple times until it's ready to merge. This leads to a messier git history, but it preserves our true public collaboration history.

That said, I imagine this is pretty frustrating for you. I think part of the problem is that this PR is doing 2 or 3 things at the same time.

1. Introduce `locking` package
2. Use locking package in commands
3. Extract channel wrappers to `tools`.

Items 1 and 2 could probably stay in 1 PR, but item 3 should definitely have been separate to minimize conflicts. We could've fast tracked a PR to extract the channel wrappers, while debating the finer points of the locking package API.

I know your LFS time is limited so feel free to toss us some tasks like item 3 (which I tackled separately). Also, I do plan for @ttaylorr or me to help out on the locking front some time after we finish the scanner and transfer queue changes. Those are core problems for all LFS users that I've wanted to fix for a long time.

sinbad:
> IMO that approach did not work well at all with the big filter process PR (#1382). At one point I was making a review comment in PR 3 that was already fixed in PR 5.

If it makes a difference, I don't intend to run PRs in parallel like that (I won't have the bandwidth). I expect to only have one in review at any one time.

I've ditched the channel wrappers for my next iteration anyway since we're moving away from channels as an API-exposed construct (on reflection, a good shout), so they're not really needed any more. It actually addresses most of the feedback on this PR in one go in fact, barring the configuration stuff. Turns out the methods in this case didn't really even need callbacks anyway given the way they were working. So I think I'll nail 1 & 2 in a mo and 3 goes away. I'll iterate again on config later but it's a 'no change' for this PR. If you think what's there can go into master, feel free to merge it.

technoweenie:
Thanks for the new PR. The `locking-master` PR target is interesting. Let's us approve and merge outside of master, so technically any Go APIs aren't shipped yet. Unline my git scanner stuff which IS technically shipped :/

I've also enabled branch protections on `locking-master`, so it verifies code reviews, ci, cla, etc :)

PR #1626, opened 2016-11-08
travis-ci: test GitLFS with ancient Git version

larsxschneider:
According to @technoweenie GitLFS should work with Git 1.8.2 or later:
https://github.com/github/git-lfs/issues/410#issuecomment-112976174

Git 1.8.2 is sufficient for Linux but macOS seems to require at least
Git 1.8.5 to run all integration tests. On Git 1.8.2 "test: pull" and
"test: pull with raw remote url" fail on macOS.

Details: https://travis-ci.org/github/git-lfs/jobs/174177362


technoweenie:
:+1:

PR #1627, opened 2016-11-08
filter stream cleanup

larsxschneider:
a few patches that I thought could be useful after my first review


test/test-install.sh
```diff
@@ -192,7 +192,9 @@ begin_test "install --local"
   [ "git lfs clean %f" = "$(git config --global filter.lfs.clean)" ]
   [ "git-lfs filter" = "$(git config filter.lfs.process)" ]
   [ "git-lfs filter" = "$(git config --local filter.lfs.process)" ]
-  [ "" = "$(git config --global filter.lfs.process)" ]
+
+  # Why is this not empty??
+  [ "git-lfs filter" = "$(git config --global filter.lfs.process)" ]
```
ttaylorr: Thinking that this has to do with some other test installing LFS globally. I think we can remove this assertion, since we can't guarantee in what order these tests run.

larsxschneider: fixed. please feel free to merge if you think that everything is OK

ttaylorr:
Going to wait for Travis to pass, then will merge away.

PR #1631, opened 2016-11-08
'smudge --info' is deprecated in favor of 'ls-files'

technoweenie:
This deprecates `git lfs smudge --info`, since `git lfs ls-files` provides us better info. 

```
$ git lfs pointer --file CHANGELOG.md | git lfs smudge -i
Git LFS pointer for CHANGELOG.md

WARNING: 'smudge --info' is deprecated and will be removed in v2.0
USE INSTEAD:
  $ git lfs pointer --file=path/to/file
  $ git lfs ls-files

20704 --
```

It also deprecates `git lfs push --stdin`, since `git lfs pre-push` exists.

```
$ echo hi | git lfs push origin master --stdin
WARNING: 'git lfs push --stdin' is deprecated, and will be removed in v2.0.
Run 'git lfs update' or ensure .git/hooks/pre-push uses 'git lfs pre-push'.
```

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -217,7 +217,7 @@ var warningOnce sync.Once
 
 func legacyWarning() {
 	warningOnce.Do(func() {
-		fmt.Fprintln(os.Stderr, "WARNING: Git LFS is using a deprecated API, which will be removed in v1.5.0.")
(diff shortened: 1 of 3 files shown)


ttaylorr:
Awesome. I'll merge this into master, update the `filter-stream` branch with these changes, and then continue my refactoring from there.

PR #1633, opened 2016-11-09
clarify auth in batch response headers

technoweenie:
@ksurent How’s this look?

/cc https://github.com/github/git-lfs/issues/1628#issuecomment-259301295


ksurent:
Looks good :)

But would look even better if there was at least a one sentence explanation or a link to where expiration tokens are also explained. Aside from your commit they're not even mentioned in the document.
With a little bit of googling I found this: https://github.com/github/git-lfs/pull/1350. Maybe copy-paste a few sentences from there?

technoweenie:
Thanks for the feedback. Closing this in favor of a much larger reorganization of the API docs. I'll post a PR once I have something legible.

PR #1634, opened 2016-11-10
Remove `AllGitConfig()`

technoweenie:
I thought it'd be cool to decouple some things from a `*config.Configuration`, specifically `cfg.Endpoint()`. When converting a raw url to an `Endpoint`, [`config.NewEndpointWithConfig()`](https://github.com/github/git-lfs/blob/a45eaee6fa35993763f87aa7611586faea08dfd5/config/endpoint.go#L54-L56) calls [`ReplaceUrlAlias()`](https://github.com/github/git-lfs/blob/a45eaee6fa35993763f87aa7611586faea08dfd5/config/config.go#L451-L454), which [calls `AllGitConfig()` inside `urlAliases()`](https://github.com/github/git-lfs/blob/a45eaee6fa35993763f87aa7611586faea08dfd5/config/config.go#L430-L438).

This PR adds `All()` to the `config.Environment` interface. But, it also had to add `Del(key string)` and `Set(key, value string)` because of a single spot that modifies that internal `gitConfig` map. I'm not crazy about it, but I think it's better than dealing with that `gitConfig`.

One thought I had was specififying a `MutableEnvironment` interface:

```golang
type Environment interface {
	Bool(key string, def bool) (val bool)
	Int(key string, def int) (val int)
	All() map[string]string
}

type MutableEnvironment interface {
	Set(key, value string)
	Del(key string)
}

type Configuration struct {
	Os Environment
	Git Environment
	mutableGit MutableEnvironment
// ...
```

They would need to point to the same data so that changes in `SetEndpointAccess()` affect `Git.Get()` calls elsewhere, but it could be some extra hassle to ensure callers don't start adding `Set()` and `Delete()` calls on `Environment` objects.

* [x] add `MutableEnvironment`?
* [x] Rename Fetcher?
* [x] Docs for all exported funcs

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -57,8 +57,6 @@ type Configuration struct {
 	// configuration.
 	Git Environment
 
-	gitConfig map[string]string
-
 	CurrentRemote   string
 	NtlmSession     ntlm.ClientSession
 	envVars         map[string]string
@@ -106,9 +104,8 @@ type Values struct {
 // This method should only be used during testing.
 func NewFrom(v Values) *Configuration {
 	return &Configuration{
-		Os:        EnvironmentOf(mapFetcher(v.Os)),
-		Git:       EnvironmentOf(mapFetcher(v.Git)),
-		gitConfig: v.Git,
+		Os:  EnvironmentOf(mapFetcher(v.Os)),
+		Git: EnvironmentOf(mapFetcher(v.Git)),
 
 		envVars: make(map[string]string, 0),
 	}
@@ -349,16 +346,10 @@ func (c *Configuration) SetEndpointAccess(e Endpoint, authType string) {
 	switch authType {
 	case "", "none":
 		git.Config.UnsetLocalKey("", key)
-
-		c.loading.Lock()
 		c.Git.Del(key)
-		c.loading.Unlock()
 	default:
 		git.Config.SetLocal("", key, authType)
-
-		c.loading.Lock()
-		c.gitConfig[strings.ToLower(key)] = authType
-		c.loading.Unlock()
+		c.Git.Set(key, authType)
 	}
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -351,7 +351,7 @@ func (c *Configuration) SetEndpointAccess(e Endpoint, authType string) {
 		git.Config.UnsetLocalKey("", key)
 
 		c.loading.Lock()
-		delete(c.gitConfig, strings.ToLower(key))
+		c.Git.Del(key)
 		c.loading.Unlock()
 	default:
 		git.Config.SetLocal("", key, authType)
--- a/config/config.go
+++ b/config/config.go
@@ -421,12 +421,6 @@ func (c *Configuration) SortedExtensions() ([]Extension, error) {
 	return SortExtensions(c.Extensions())
 }
 
-func (c *Configuration) AllGitConfig() map[string]string {
-	c.loadGitConfig()
-
-	return c.gitConfig
-}
-
 func (c *Configuration) urlAliases() map[string]string {
 	c.urlAliasMu.Lock()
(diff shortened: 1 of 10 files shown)


config/config.go
```diff
@@ -106,9 +104,8 @@ type Values struct {
 // This method should only be used during testing.
 func NewFrom(v Values) *Configuration {
 	return &Configuration{
-		Os:        EnvironmentOf(mapFetcher(v.Os)),
-		Git:       EnvironmentOf(mapFetcher(v.Git)),
-		gitConfig: v.Git,
```
ttaylorr: 🎉

config/config.go
```diff
@@ -349,16 +346,10 @@ func (c *Configuration) SetEndpointAccess(e Endpoint, authType string) {
 	switch authType {
 	case "", "none":
 		git.Config.UnsetLocalKey("", key)
-
-		c.loading.Lock()
-		delete(c.gitConfig, strings.ToLower(key))
-		c.loading.Unlock()
+		c.Git.Del(key)
```
ttaylorr: I definitely agree that this isn't my first choice of things to add to the public API for the `config` package, since it ~~encourages~~ allows calling code to preform these same mutable operations.

I think the `MutableEnvironment` is an interesting option. Here's another thought:

```go
func del(e *Environment, key string) *Environment { ... }
func set(e *Environment, key, val string) *Environment { ... }
```

and

```go
c.Git = del(c.Git, key)
```

This would allow us to make the mutable operations (`del` and `set`) package-private, while still supporting this single instance of mutation. I'm not immediately sure how you'd implement such a function "`del`", but on first glance it may be more complicated than simply making a copy of all keys save the one(s) that you want to remove.

Curious as to your thoughts!

config/git_fetcher.go
```diff
@@ -126,6 +126,31 @@ func (g *GitFetcher) Get(key string) (val string, ok bool) {
...
+}
+
+func (g *GitFetcher) Set(key, value string) {
+	g.vmu.RLock()
+	defer g.vmu.RUnlock()
+	g.vals[strings.ToLower(key)] = value
+}
+
+func (g *GitFetcher) Del(key string) {
+	g.vmu.RLock()
```
ttaylorr: Since this is a write operation, shouldn't we use the `Lock()` and `Unlock()` methods, since `RLock` and `RUnlock` are designated to readers?

config/git_fetcher.go
```diff
@@ -126,6 +126,31 @@ func (g *GitFetcher) Get(key string) (val string, ok bool) {
...
+	for key, value := range g.vals {
+		newmap[key] = value
+	}
+
+	return newmap
+}
+
+func (g *GitFetcher) Set(key, value string) {
+	g.vmu.RLock()
+	defer g.vmu.RUnlock()
```
ttaylorr: Same note about the `RLock` vs `Lock` method call here.

config/os_fetcher.go
```diff
@@ -53,3 +53,11 @@ func (o *OsFetcher) Get(key string) (val string, ok bool) {
 
 	return v, ok
 }
+
+func (o *OsFetcher) All() map[string]string {
+	return nil
+}
+
+func (o *OsFetcher) Set(key, value string) {}
```
ttaylorr: If the `Set` and `Del` methods remain on this type, I think a `panic` would be appropriate here, since these methods explicitly do nothing which does not agree with the caller's expectations.

config/config.go
technoweenie: A private func could work, as long as `SetEndpointAccess()` stays in the `config` package. I don't think it will, but I also think a better design will remove the need for it.

config/git_fetcher.go
technoweenie: Ah yeah, copy pasta.

PR #1646, opened 2016-11-12
WIP: promised downloads (proof of concept)

larsxschneider:
Goal (implemented as proof of concept demo):
If a file is not in the GitLFS cache, then GitLFS returns an empty file upon Git's smudge request and starts the download right away. At the end of the "process-filter" protocol GitLFS waits for the downloads to finish and writes the files to the Git working tree.

**This should allows us to get rid of the `git lfs clone|pull|fetch` commands.**

---

@ttaylorr If you like it then please change the code to use the proper parallel download machinery. 😉

### DIFF
--- a/lfs/pointer_smudge.go
+++ b/lfs/pointer_smudge.go
@@ -1,6 +1,7 @@
 package lfs
 
 import (
+	"bufio"
 	"fmt"
 	"io"
 	"os"
@@ -16,6 +17,33 @@ import (
 	"github.com/github/git-lfs/progress"
 	"github.com/rubyist/tracerx"
 )
+
+type DeferredDownload struct {
+	Ptr         *Pointer
+	WorkingFile string
+}
+
+var deferredDownloads []DeferredDownload
+
+func PointerWaitForDownloads(manifest *transfer.Manifest) {
+	// Here we should wait for downloads. This proof of concept demo only
+	// starts to download the files here.
+	for _, dd := range deferredDownloads {
+		fmt.Fprintf(os.Stderr, "Deferred download: %s\n", dd.WorkingFile)
+		mediafile, _ := LocalMediaPath(dd.Ptr.Oid)
+		stat, statErr := os.Stat(mediafile)
+		f, _ := os.Create(dd.WorkingFile)
+		w := bufio.NewWriter(f)
+		if statErr != nil || stat == nil {
+			downloadFile(w, dd.Ptr, dd.WorkingFile, mediafile, manifest, nil)
+		} else {
+			// This case would happen if we smudge a file multiple times
+			// or if a LFS file is multiple times in the worktree.
+			readLocalFile(w, dd.Ptr, mediafile, dd.WorkingFile, nil)
+		}
+		w.Flush()
+	}
+}
 
 func PointerSmudgeToFile(filename string, ptr *Pointer, download bool, manifest *transfer.Manifest, cb progress.CopyCallback) error {
 	os.MkdirAll(filepath.Dir(filename), 0755)
@@ -57,7 +85,22 @@ func PointerSmudge(writer io.Writer, ptr *Pointer, workingfile string, download 
 
 	if statErr != nil || stat == nil {
 		if download {
-			err = downloadFile(writer, ptr, workingfile, mediafile, manifest, cb)
+			// TODO: Check if the `ptr.Oid` is already being downloaded.
+			// If no download is running then start the download right away!
+			// If a download of the `ptr.Oid` is running then check if the
+			// `workfile` is the same. If this is the case then no further
+			// action is required. If the `workfile` is not the same then
+			// ensure that the file is written on `WaitForDownloads`.
+			//
+			// Important: GitLFS should not write the files directly to the
+			// worktree when they finish. It should write the files only to the
+			// local media path. Only at the end, when Git signals GitLFS with
+			// an EOF its exit then GitLFS should write the files (in the
+			// `WaitForDownloads` step).
+			var dd DeferredDownload
+			dd.Ptr = ptr
+			dd.WorkingFile = workingfile
+			deferredDownloads = append(deferredDownloads, dd)
 		} else {
 			return errors.NewDownloadDeclinedError(statErr, "smudge")
 		}
--- a/commands/command_filter_process.go
+++ b/commands/command_filter_process.go
@@ -104,6 +104,12 @@ Scan:
 		s.WriteStatus(status)
 	}
 
+	// TODO: Detect an EOF after a successful filter-process request (EOF at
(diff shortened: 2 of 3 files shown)


ttaylorr:
Looking like a great start. I'm really psyched to get this in, it's going to be a great addition combined with the filter process stuff we implemented in #1617.

I can take a stab at this and have it handle transfers in parallel, shouldn't be too tricky.

What's required on the Git side?

lfs/pointer_smudge.go
```diff
@@ -57,7 +85,22 @@ func PointerSmudge(writer io.Writer, ptr *Pointer, workingfile string, download
...
+			//
+			// Important: GitLFS should not write the files directly to the
+			// worktree when they finish. It should write the files only to the
+			// local media path. Only at the end, when Git signals GitLFS with
+			// an EOF its exit then GitLFS should write the files (in the
+			// `WaitForDownloads` step).
+			var dd DeferredDownload
+			dd.Ptr = ptr
+			dd.WorkingFile = workingfile
+			deferredDownloads = append(deferredDownloads, dd)
```
larsxschneider: Here we write nothing to `writer` which means the `smudge` result is an empty file. We might want to write the GitLFS pointer there to `writer` or a message like `file is being downloaded`. It doesn't matter what we write ... the file is overwritten at the end anyways. This intermedia step is only relevant in case GitLFS dies...

larsxschneider:
Nothing required on the Git side. [Regular 2.11 release candidate](https://github.com/git/git/releases/tag/v2.11.0-rc1) is sufficient.

ttaylorr:
> Nothing required on the Git side. Regular 2.11 release candidate is sufficient.

My concern is that by writing empty contents to the file and then doing the copy on the LFS side, the story of failing a transfer part of the way through is not great. If our filter dies before we have a chance to copy the files from their temporary locations, the working tree is left in a dirty state which would show all of the pointer files being removed.

I see a few options on how to remediate this:

- Implement a small patch in Git adding a protocol capability to "roll back" the conversions performed by the filter in the event of a failure. We could send this packet at the very end if a transfer failed.
- Implement the patch as originally discussed where Git would know about the temporary locations of the files on disk and would perform the copy itself, respected the `filter.<driver>.required` config option.
- Maintain a set of those temporary files ourselves and write _another_ filter which would execute the copy _ONLY_ if the transfer succeeded. With this approach, I think our safest bet would be to echo back the contents of the file as given to us by the filter, but this doesn't really solve the problem since we would have to write a potentially huge amount of data over the pipe, leaving us where we started.

My preference would be to teach Git the promise-based approach, that way we can respect the `filter.<driver>.required` option and defer the responsibility of bringing the changes into the working-tree/index to Git itself. This way, if a transfer fails, Git can choose to not copy the files over, not wasting disk I/O and (more importantly) not leaving the working tree or index in a dirty state.

larsxschneider:
I looked a bit deeper into a possible Git core implementation and realized that the "temporary file" idea might be too complicated with the current Git core filter code. I proposed an alternative idea as RFC to the Git list: http://public-inbox.org/git/D10F7C47-14E8-465B-8B7A-A09A1B28A39F@gmail.com/

Note the last sentence of Eric's reply:
> Anyways, I'll plan on doing something similar (in Perl) with the
> synchronous parts of public-inbox which relies on "cat-file --batch"
> at some point... (my rotational disks are sloooooooow :<)

That might be interesting for #1649 /cc @technoweenie !

technoweenie:
Interesting. Both this idea, and the comment about `cat-file --batch`.

larsxschneider:
I posted a patch to support this "promised downloads" natively in Git:
http://public-inbox.org/git/20170108191736.47359-1-larsxschneider@gmail.com/

rjbell4:
Just found this.  Love the idea!  It would be great for general LFS support.  Though hopefully we can still support "skip smudge" for some use cases.

ttaylorr:
> Just found this. Love the idea! It would be great for general LFS support. Though hopefully we can still support "skip smudge" for some use cases.

Don't see any reason why we couldn't 👍 .

Spunkie:
If this was implemented as is would there be any way of supporting workflows that make use of the `-X`/`--exclude=` and `-I`/`--include=` params when using straight `git clone|pull|fetch` or would I still need to fallback on `git lfs clone|pull|fetch` cmds?

ttaylorr:
> If this was implemented as is would there be any way of supporting workflows that make use of the `-X`/`--exclude=` and `-I`/`--include=` params when using straight `git clone|pull|fetch` or would I still need to fallback on `git lfs clone|pull|fetch` cmds?

In Git today, you'd have to use the `git lfs` variants to make use of LFS-specific flags. That being said, there are a few alternatives:

1. Git forwards all un-recognized flags to filter drivers, hooks, etc.
2. You `git lfs install --skip-smudge`, and then do a `git lfs checkout -I ... -X ...` once inside your repository.

larsxschneider:
I posted my v3 of the relevant Git parts for this change:
http://public-inbox.org/git/20170409191107.20547-1-larsxschneider@gmail.com/

Feedback would be highly appreciated for Git filter protocol change described here:
https://github.com/larsxschneider/git/commit/08a461f103#diff-6fb066a3d66afe00c0995f97f5c680fdR521

larsxschneider:
The necessary Git filter protocol changes have been merged to the Git core `next` branch and are scheduled for the `master` branch: https://github.com/git/git/commit/2841e8f81cb2820024804b9341577be1d0ce1240

I hope they make it into the Git 2.14 release 😊

ttaylorr:
> The necessary Git filter protocol changes have been merged to the Git core `next` branch and are scheduled for the `master` branch: [git/git@2841e8f](https://github.com/git/git/commit/2841e8f81cb2820024804b9341577be1d0ce1240)

🎉 Hooray! Excellent job on this patch series, as always.

With regards to implementing support for this in LFS, I am more than happy to help. My current roadmap is focused on performance improvements to the migrator which I anticipate taking about a month.

This change should be self-contained enough that I am able to work on it at the same time as the other items on my roadmap.

@larsxschneider how do you feel about adopting a similar workflow to when we worked on the initial `process` filter?

larsxschneider:
Thanks @ttaylorr ! Yeah, I think a similar workflow as before would be great. Please let me know if I can help in any way.

ttaylorr:
> Thanks @ttaylorr ! Yeah, I think a similar workflow as before would be great. Please let me know if I can help in any way.

Will do, thanks! 😄

technoweenie:
@larsxschneider Thanks and congrats on the git patch! Excited to see how this improves clone/fetch times :metal:

larsxschneider:
This PR is superseded by #2511

PR #1650, opened 2016-11-14
rewrite new catfilebatch implementation for upcoming gitscanner pkg

technoweenie:
This is the first step towards https://github.com/github/git-lfs/issues/1649. This is a pattern that worked for a small scanner prototype I hacked up this weekend. The scanner currently relies on `git rev-list`, `git cat-file --batch-check`, and `git cat-file --batch`. Feeding data through requires 3 channels:

* An output channel that is sent results from the current command.
* An input channel that receives values (usually git sha1s) from something.
* A single error channel that all commands share.

The commands themselves continue running in goroutines until the input channel is closed. Once closed, the stdout reader will eventually return `io.EOF`, which will eventually close the output channel. The output channel of `git cat-file --batch-check` would be an input channel for `git cat-file --batch`, so its closure would eventually close `git cat-file --batch` too.

This code is written in the lfs package with prefixed filenames and prefixed types where necessary. The goal is to get the new design in there, and then start phasing out coupling on the channel wrappers, `*lfs.WrappedPointer`, etc. Doing this in the `lfs` package is vital, so we can watch it run our test suite and find issues earlier rather than later.


technoweenie:
That failure is due to `cat-file --batch` [returning](https://ci.appveyor.com/project/git-lfs/git-lfs/build/1.0.436) this:

    126fd41019b623ce1621a638d2535b26e0edb4df missing

It passes locally because it's referencing a prototype branch that I have locally, but isn't pushed to GitHub. It should pass with those objects, but I think we have a better way of testing git stuff in our go test. I don't want this to be purely tested by our bash tests.

lfs/gitscanner_catfilebatch.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	go catFileBatchOutput(pointerCh, cmd, errCh)
+	go catFileBatchInput(cmd, sha1Ch, errCh)
+	return nil
+}
+
+func catFileBatchOutput(pointerCh chan *WrappedPointer, cmd *wrappedCmd, errCh chan error) {
+	for {
+		l, err := cmd.Stdout.ReadBytes('\n')
```
ttaylorr: I think using a `*bufio.Reader` might be better here, that way you can just call `Text()` or `Bytes()`.

lfs/gitscanner_catfilebatch.go
```diff
@@ -0,0 +1,84 @@
+package lfs
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"strconv"
+
+	"github.com/pkg/errors"
```
ttaylorr: Did you mean to import `github.com/github/git-lfs/errors`?

lfs/gitscanner_catfilebatch.go
ttaylorr: On second thought, maybe the scanner pattern could work to return the three fields:

```go 
scanner := newCatFileBatchScanner(cmd.Stdout)

for scanner.Scan() {
        scanner.Sha1()
        scanner.Type()
        scanner.Size()
}

if err := scanner.Err(); err != nil && err != io.EOF {
        // ...
}
```

This should make it a little easier to test this code and I dig the logical separation.

lfs/gitscanner_catfilebatch.go
```diff
@@ -0,0 +1,84 @@
...
+		fields := bytes.Fields(l)
+		s, _ := strconv.Atoi(string(fields[2]))
+
+		nbuf := make([]byte, s)
+		_, err = io.ReadFull(cmd.Stdout, nbuf)
+		if err != nil {
+			break // Legit errors
+		}
+
+		p, err := DecodePointer(bytes.NewBuffer(nbuf))
```
ttaylorr: You could use a limit reader here to avoid buffering this data.

```go
p, err := DecodePointer(io.LimitReader(cmd.Stdout, s)
```

There might be some internal buffering going on in `DecodePointer`, but I at least think this looks a little cleaner.

lfs/gitscanner_catfilebatch.go
```diff
@@ -0,0 +1,84 @@
...
+			pointerCh <- &WrappedPointer{
+				Sha1:    string(fields[0]),
+				Size:    p.Size,
+				Pointer: p,
+			}
+		}
+
+		_, err = cmd.Stdout.ReadBytes('\n') // Extra \n inserted by cat-file
+		if err != nil {
+			if err != io.EOF {
```
ttaylorr: Could you reverse this conditional, making it a little easier to read?

```go
if err != nil {
        if err == io.EOF {
               break
        }

        errCh <- errors.Wrap(err, "...")
}

File: lfs/gitscanner_catfilebatch.go [hunk 5]
```diff
@@ -0,0 +1,84 @@
...
+				errCh <- errors.Wrap(err, "git cat-file --batch:")
+			}
+			break
+		}
+	}
+
+	stderr, _ := ioutil.ReadAll(cmd.Stderr)
```
ttaylorr: I should really push my branch that wraps this behavior in the command itself, but I think that's outside the scope of this PR.

lfs/gitscanner_catfilebatch.go
technoweenie: I think this would make a good follow up PR. Trying not to change things _too much_.

lfs/gitscanner_catfilebatch.go
technoweenie: Well duh, `break != continue`. I wrote a scanner impl as you suggested, which was pretty easy. Simplifies the flow a tad too.

technoweenie:
Welp, I give up. I can't use the `test` package, since it imports `lfs`. So, trying to use it to test the git scanner in the `lfs` package results in an import loop.

So, I went back to my original idea, of just adding fake lfs pointers to the repository and running the scanner against itself. That worked well, until I tried to push.

```
7e9e5ac30f2216fd0fd6f5faed316f2d5983361a4203c3330cfa46ef65bb4767 does not exist in .git/lfs/objects. Tried git/fixtures/add.lfs.txt, which matches 7e9e5ac30f2216fd0fd6f5faed316f2d5983361a4203c3330cfa46ef65bb4767.
error: failed to push some refs to 'https://github.com/github/git-lfs'
```

So unfortunately, I can't add go tests for this right now :( I'll bring the test back once this code has been extracted to a separate package.

technoweenie:
Converted `catFileBatchCheck` as well. I think this is ready to go. I want to start on the new public scanner interface in the next PR.

lfs/gitscanner_catfilebatch.go
```diff
@@ -28,6 +28,33 @@ func runCatFileBatch(pointerCh chan *WrappedPointer, sha1Ch <-chan string, errCh
 	return nil
 }
 
+func catFileBatchOutput(pointerCh chan *WrappedPointer, cmd *wrappedCmd, errCh chan error) {
+	scanner := &catFileBatchScanner{r: cmd.Stdout}
+	for scanner.Scan() {
+		pointerCh <- scanner.Pointer()
```
ttaylorr: So clean 😍 ! Awesome work.

Issue #1640, opened 2016-11-10
Filter protocol follow-up issue


ttaylorr:
As we're getting close to merging support for the new Filter Protocol introduced in Git 2.11 into LFS, there are a few issues that I'd like to follow up on that are _not necessary to address before merging_.

| | Description | Refs | 
| - | ---------- | ---------- |
|~~◽️~~| ~~Reconcile `io.EOF`s that are legitimate, vs. `io.EOF`s that just mean we're done reading/writing a packet.~~ | https://github.com/github/git-lfs/pull/1617#discussion_r87365141 https://github.com/github/git-lfs/pull/1617#discussion_r87361200 |
|:white_check_mark:| Flush the `io.Writer` only when necessary, not after every packet write. | https://github.com/github/git-lfs/pull/1617#discussion_r87362237 |
|:construction:| ~~Use empty list to repeat `status=` packets.~~ |  |
| 🚫  | ~~Fix `requireStdin` on Windows.~~ | https://github.com/github/git-lfs/issues/1639 |
|🚫| Make error behavior consistent. | https://github.com/github/git-lfs/pull/1617#discussion_r87562930 https://github.com/github/git-lfs/pull/1617#discussion_r87356653 --> Flush part |

---

/cc @larsxschneider @technoweenie

ttaylorr:
Going to take a look at this this week. (+acking out of our internal first-responder queue.)

ttaylorr:
> Fix requireStdin on Windows.

Not sure if this is possible. Tried a number of different approaches locally, but I think there is always a `stdin` since the input can come from your terminal in interactive mode.

ttaylorr:
Tracking: https://github.com/git-lfs/git-lfs/pull/1617#discussion_r88133403.

ttaylorr:
Also tracking: [`git-lfs/git-lfs/fewer-status-bytes`][1]. The implementation appears correct to me based on conversations that I've had with @larsxschneider and inspection of the [relevant Git code][2], but causes a segfault in practice.

@larsxschneider would you be open to 🍐-ing on this tomorrow?

[1]: https://github.com/git-lfs/git-lfs/tree/fewer-status-bytes
[2]: https://github.com/git/git/blob/3ab228137f980ff72dbdf5064a877d07bec76df9/convert.c#L739-L749

ttaylorr:
> Reconcile io.EOFs that are legitimate, vs. io.EOFs that just mean we're done reading/writing a packet.

[...]

PR #1658, opened 2016-11-15
filter-process: less flushing

ttaylorr:
Previously, we flushed intermediate data out of the underlying *bufio.Writer
used by the implementation of `pkt-line` in `*git.pktline` after _every_ write.
This means that we would flush after partial packets, whole packets, and after
pkt-line flush sequences. This is unnecessary.

Let us instead flush only after an entire packet has been written, after
writing the `0000` flush sequence.

/ref git-lfs/git-lfs#1640

---

/cc @technoweenie @larsxschneider

### DIFF
--- a/git/filter_process_scanner_test.go
+++ b/git/filter_process_scanner_test.go
@@ -35,6 +35,7 @@ func TestFilterProcessScannerRejectsUnrecognizedInitializationMessages(t *testin
 
 	pl := newPktline(nil, &from)
 	require.Nil(t, pl.writePacketText("git-filter-client-unknown"))
+	require.Nil(t, pl.writeFlush())
 
 	fps := NewFilterProcessScanner(&from, &to)
 	err := fps.Init()
@@ -106,8 +107,7 @@ func TestFilterProcessScannerReadsRequestHeadersAndPayload(t *testing.T) {
 	// Multi-line packet
 	require.Nil(t, pl.writePacketText("first"))
 	require.Nil(t, pl.writePacketText("second"))
-	_, err := from.Write([]byte{0x30, 0x30, 0x30, 0x30}) // flush packet
-	assert.Nil(t, err)
+	require.Nil(t, pl.writeFlush())
 
 	req, err := readRequest(NewFilterProcessScanner(&from, &to))
 
@@ -121,13 +121,11 @@ func TestFilterProcessScannerReadsRequestHeadersAndPayload(t *testing.T) {
 }
 
 func TestFilterProcessScannerRejectsInvalidHeaderPackets(t *testing.T) {
-	var from bytes.Buffer
+	from := bytes.NewBuffer([]byte{
+		0x30, 0x30, 0x30, 0x34, // 0004 (invalid packet length)
+	})
 
-	pl := newPktline(nil, &from)
-	// (Invalid) headers
-	require.Nil(t, pl.writePacket([]byte{}))
-
-	req, err := readRequest(NewFilterProcessScanner(&from, nil))
+	req, err := readRequest(NewFilterProcessScanner(from, nil))
 
 	require.NotNil(t, err)
 	assert.Equal(t, "Invalid packet length.", err.Error())
--- a/git/pkt_line.go
+++ b/git/pkt_line.go
@@ -68,11 +68,17 @@ func (p *pktline) readPacket() ([]byte, error) {
 	return payload, err
 }
 
+// readPacketText follows identical semantics to the `readPacket()` function,
+// but additionally removes the trailing `\n` LF from the end of the packet, if
+// present.
 func (p *pktline) readPacketText() (string, error) {
 	data, err := p.readPacket()
 	return strings.TrimSuffix(string(data), "\n"), err
 }
 
+// readPacketList reads as many packets as possible using the `readPacketText`
+// function before encountering a flush packet. It returns a slice of all the
+// packets it read, or an error if one was encountered.
 func (p *pktline) readPacketList() ([]string, error) {
 	var list []string
 	for {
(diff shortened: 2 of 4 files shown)


larsxschneider:
I measured if we can observe and performance improvement/degradation with this change.

## Test Call
```
git --version; git-lfs version; for i in `seq 1 5`; do printf "RUN $i: "; rm -rf perftest; mkdir perftest && cd perftest && git init . >/dev/null 2>&1 && mkfile -n 1G test.dat && git-lfs track *.dat >/dev/null 2>&1 && time git add .; cd ..; done
```

## Results
```
# More flushing
git version 2.11.0.rc1
git-lfs/1.4.4 (GitHub; darwin amd64; go 1.7.3; git 5e48c70)
RUN 1: git add .  4.92s user 1.58s system 126% cpu 5.137 total
RUN 2: git add .  4.84s user 1.52s system 138% cpu 4.600 total
RUN 3: git add .  4.85s user 1.47s system 137% cpu 4.577 total
RUN 4: git add .  4.83s user 1.50s system 137% cpu 4.605 total
RUN 5: git add .  4.95s user 1.55s system 134% cpu 4.838 total

# Less flushing
git version 2.11.0.rc1
git-lfs/1.4.4 (GitHub; darwin amd64; go 1.7.3; git b3a4e84)
RUN 1: git add .  4.95s user 1.51s system 126% cpu 5.119 total
RUN 2: git add .  4.98s user 1.61s system 117% cpu 5.596 total
RUN 3: git add .  4.91s user 1.51s system 132% cpu 4.861 total
RUN 4: git add .  4.86s user 1.54s system 136% cpu 4.689 total
RUN 5: git add .  4.90s user 1.50s system 129% cpu 4.940 total
```
--> turns out it does not make a difference 😄 ! But I agree it is the right thing to do.

Out of curiosity I also tested the legacy filters:
```
git version 2.10.2
git-lfs/1.4.4 (GitHub; darwin amd64; go 1.7.3)
RUN 1: git add .  3.00s user 1.24s system 113% cpu 3.716 total
RUN 2: git add .  3.01s user 1.23s system 107% cpu 3.930 total
RUN 3: git add .  3.00s user 1.25s system 113% cpu 3.757 total
RUN 4: git add .  3.01s user 1.21s system 111% cpu 3.798 total
RUN 5: git add .  3.00s user 1.25s system 113% cpu 3.734 total
```

As you can see adding a file is faster propably because Git can write the entire file without modification to the pipe. Using the filter protocol the file needs to be chunked in pkt-line packages.

In the future I would like to pick up the [`smudgeToFile/cleanFrom` patch series from Joey Hess](http://public-inbox.org/git/20160616203259.5886-1-joeyh@joeyh.name/) (developer behind git-annex) to tackle this problem. This should give us a performance that is way better than the legacy filters 🤘

larsxschneider:
Scratch that. I just realized that I was benchmarking "clean" and Git-LFS does not write many packets during clean. Here is the smudge benchmark:

## Test Call
```
git --version; git-lfs version; mkfile -n 1G test.dat; for i in `seq 1 5`; do printf "RUN $i: "; rm -rf perftest; mkdir perftest && cp test.dat perftest && cd perftest && git init . >/dev/null 2>&1 && git-lfs track *.dat >/dev/null 2>&1 && git add . && git commit -m "test" >/dev/null 2>&1; rm test.dat; time git checkout .; cd ..; done
```

## Results
```
# More flushing
git version 2.11.0.rc1
git-lfs/1.4.4 (GitHub; darwin amd64; go 1.7.3; git 5e48c70)
RUN 1: git checkout .  0.79s user 9.39s system 103% cpu 9.834 total
RUN 2: git checkout .  0.77s user 9.17s system 102% cpu 9.700 total
RUN 3: git checkout .  0.78s user 9.11s system 103% cpu 9.537 total
RUN 4: git checkout .  0.78s user 9.27s system 103% cpu 9.665 total
RUN 5: git checkout .  0.76s user 8.91s system 104% cpu 9.262 total

# Less flushing
git version 2.11.0.rc1
git-lfs/1.4.4 (GitHub; darwin amd64; go 1.7.3; git b3a4e84)
RUN 1: git checkout .  0.76s user 9.13s system 107% cpu 9.223 total
RUN 2: git checkout .  0.76s user 9.30s system 106% cpu 9.442 total
RUN 3: git checkout .  0.76s user 9.24s system 105% cpu 9.436 total
RUN 4: git checkout .  0.76s user 9.22s system 106% cpu 9.356 total
RUN 5: git checkout .  0.74s user 9.21s system 106% cpu 9.350 total
```
--> again, it does not make a difference 😄 ! 

Legacy filters:
```
git version 2.10.2
git-lfs/1.4.4 (GitHub; darwin amd64; go 1.7.3)
RUN 1: git checkout .  0.03s user 8.34s system 96% cpu 8.633 total
RUN 2: git checkout .  0.03s user 8.40s system 96% cpu 8.697 total
RUN 3: git checkout .  0.04s user 8.37s system 99% cpu 8.445 total
RUN 4: git checkout .  0.03s user 8.51s system 100% cpu 8.507 total
RUN 5: git checkout .  0.04s user 8.28s system 99% cpu 8.372 total
```

Again, legacy filters are a bit faster for the reason described above.
What I don't get is why the relative difference is sooo different between clean and smudge. Shouldn't the relative difference be similar? Maybe Git is not writing the pkt-line packets in the optimal way or GitLFS is not receiving them in the optimal way?

Any ideas @ttaylorr @technoweenie ?

ttaylorr:
> As you can see adding a file is faster propably because Git can write the entire file without modification to the pipe. Using the filter protocol the file needs to be chunked in pkt-line packages.

Yeah, it's too bad that this new hotness doesn't help with the case of few, large files as much as it does with many small ones. Still, I think this is a good change 😉 

Thanks again for benchmarking, @larsxschneider. Even though the perf wins weren't huge, I still think this is a good thing to do.

PR #1659, opened 2016-11-15
run osx builds on circle

technoweenie:


ttaylorr:
Rad. Bummer that CircleCI doesn't allow us to run matrix builds. What do you think about running one long build against three different versions of Git to accomplish the same effect?

technoweenie:
Ready to go :metal:  CircleCI tests finished before AppVeyor and Travis could even _start_. Just need dat review!

sschuberth:
Neither the commits nor this PR seem to give a rationale *why* macOS builds were moved from Travis CI to Circle CI. Mind explaining why? Is it to have better chances to run the builds for all platforms in parallel and thus reduce the feedback time for PRs?

ttaylorr:
> Neither the commits nor this PR seem to give a rationale why macOS builds were moved from Travis CI to Circle CI.

We definitely should have mentioned this earlier, you're correct.

> Mind explaining why?

Not at all! 😄  The parallelism wasn't so much our concern as was the boot times for macOS containers on Travis. TravisCI is a great service, and continues to run our Linux build matrix with speed, but the startup time of macOS containers on Travis was prohibitive. In terms of numbers, we could be sitting macOS queue for upwards of 90 minutes before our builds would start. This isn't such a problem normally, but LFS has at least a few PRs pushed every day, and this wait time is not sustainable.

CircleCI was recommended to me by a member of the Atom team, for they use the same suite of CI tools that we use now. It works well for us.

PR #1661, opened 2016-11-15
Allow non lfs test files

technoweenie:
#1650 needs a way to test against generated git repos. I wanted to use `*test.Repo` for this, but it only ever writes LFS objects. So, this PR refactors that code so that it uses a `test.BlobInput` interface, giving me an opportunity to write a `test.FileInput` implementation.

This changes the interface to construct the `test.LFSInput` objects:

```diff
-{Filename: "file1.txt", Size: 25},
+test.NewLFSInput("file1.txt", test.RandInput(t, 25)),
```

It's longer, but explicit. Also, it's easy to add a static string:

```diff
+test.NewLFSInput("file1.txt", "booya"),
```

This also removes the custom random content generator. I'm not sure we need anything fancy, when tests only generate up to 42 bytes, and the api command only generates up to 250.

/cc @sinbad

### DIFF
--- a/git/git_test.go
+++ b/git/git_test.go
@@ -25,26 +25,26 @@ func TestCurrentRefAndCurrentRemoteRef(t *testing.T) {
 	// only interested in branches
 	inputs := []*test.CommitInput{
 		{ // 0
-			Files: []*test.LFSInput{
-				{Filename: "file1.txt", Size: 20},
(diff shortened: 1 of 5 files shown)


technoweenie:
Blah, forgot about `lfstest-testutils`. Going to start over.

PR #1666, opened 2016-11-16
refactor testutils.FileInput so it's a little more clear

technoweenie:
I want to use `testutils` for testing #1650. However, it seems that `testutils.FileInput` can _only_ write LFS objects. This refactors the code into `(*FileInput) AddToCommit()`, which can eventually support writing plain git objects.

### DIFF
--- a/test/testutils.go
+++ b/test/testutils.go
@@ -20,10 +20,10 @@ import (
 	"sync"
 	"time"
 
+	"github.com/git-lfs/git-lfs/errors"
 	"github.com/git-lfs/git-lfs/git"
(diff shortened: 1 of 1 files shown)


test/testutils.go
```diff
@@ -23,6 +23,7 @@ import (
 	"github.com/git-lfs/git-lfs/git"
 	"github.com/git-lfs/git-lfs/lfs"
 	"github.com/git-lfs/git-lfs/localstorage"
+	"github.com/pkg/errors"
```
ttaylorr: `github.com/git-lfs/git-lfs/errors`?

PR #1668, opened 2016-11-16
commands: prefer `ExitWithError` over writing `status=error`

ttaylorr:
Here's a PR that prefers exiting with a non-zero code from the LFS process if we encountered an error, rather than writing the `status=error\n` sentinel.

@larsxschneider suggested this to me as a cleaner and more consistent alternative to writing the above sentinel thereby placing the responsibility of exiting in LFS's hands rather than Git's.

I'm not sure that I agree that this is necessarily cleaner, but I'd love to get some outside thoughts as well. Either way should work.

---

/cc @technoweenie @larsxschneider

### DIFF
--- a/commands/command_filter_process.go
+++ b/commands/command_filter_process.go
@@ -61,6 +61,8 @@ func filterSmudge(to io.Writer, from io.Reader, filename string) error {
 }
 
 func filterCommand(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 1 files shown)


ttaylorr:
Ha, looks like this is broken anyway. Thinking we'll be okay sending a few extra bytes of data 👍

Issue #1649, opened 2016-11-14
Proposal: git/gitscanner package for v2.0


technoweenie:
I'd like to propose a new design for the scanner code in the `lfs` package. The scanner functions are concerned with extracting LFS pointers from a range of git objects:

* `lfs.ScanRefs()` scans a range of commits.
* `lfs.ScanTree()` scans just the tree of a single commit.
* `lfs.ScanIndex()` scans uncommited objects in the git index.
* `lfs.ScanUnpushed()` looks for local objects that have not been pushed to a given remote.
* `lfs.ScanPreviousVersions()` scans commits in a ref since a given time.

There are a number of problems with it, as it has organically grown:

1. Common usage of the scanner functions buffer the LFS pointers to a slice. This means operations that depend on the scanner can't start until it's done. This is most apparent when you notice that LFS uploads of large git pushes don't start until `git rev-list` has run on the entire history.
2. No shared context, each `lfs.ScanRefs()` in a `git push` with lots of branches/tags results in duplicated work, such as [comparing local and remote refs](https://github.com/github/git-lfs/blob/49354864921c6d4acaedf70ccd33a6ff5abff431/lfs/scanner.go#L296-L310), or spinning up extra `git cat-file` processes for each ref.
3. Lots of coupling with the `config` and `lfs` packages.

@ttaylorr and I came up with a design for a new scanner, focusing on async communication so it can begin feeding LFS Pointers to the transfer queue almost immediately.

![](https://cloud.githubusercontent.com/assets/443245/19458195/d66b94b6-9487-11e6-8a1b-828cfd15d86a.png)

Essentially, I want to create a `gitscanner.Scanner` object that can handle multiple scan options. Here's an illustration in pseudo go code for a pre-push implementation.

```go
// The hook receives commit information on stdin in the form:
// <local ref> <local sha1> <remote ref> <remote sha1>
// Pushes with lots of branches/tags will have to scan multiple commit ranges

// this spins up `git cat-file --batch-check` and `git cat-file --batch` commands
gscanner := gitscanner.New()

// the scanner can calculate the remote refs needed for a push, just once,
// instead of for each ref line in stdin.
// https://github.com/github/git-lfs/blob/49354864921c6d4acaedf70ccd33a6ff5abff431/lfs/scanner.go#L296-L310
gscanner.Remote("origin")

go func() {
  for pointer := range gscanner.Output {
    // pass to the transfer queue
  }
}()

[...]

PR #1670, opened 2016-11-16
experimental gitscanner api

technoweenie:
This PR builds on #1650 and introduces a `GitScanner` type. It's pretty empty, but it illustrates where I want to go with #1649. As of this PR, nothing outside of the `lfs` package uses `lfs.ScanRefs()` or `lfs.ScanRefsToChan()`.

TODO

* [x] merge #1650 
* [x] change base to `master`
* [x] :ship:


ttaylorr:
This looks good to me.

technoweenie:
I just pushed a change that stores the skipped refs on the `*GitScanner`. This means that initial git pushes with lots of refs will only calculate this data once, instead of on every `lfs.ScanRefs()` or `(*GitScanner) ScanLeftToRemote()` call.

ttaylorr:
> I just pushed a change that stores the skipped refs on the `*GitScanner`.

Looks sweet. I'm a little bit worried about that list being potentially large, but not sure what other alternatives I'd go for.

technoweenie:
Okay, I think this is ready to go now. `lfs/scanner.go` has no more public functions. Everything is using the new gitscanner api. I'd like to merge this so I can continue on more focused refactoring.

technoweenie:
As a bonus, I exported the base channel wrapper stuff to `tools` (from #1625).

lfs/gitscanner.go
```diff
@@ -0,0 +1,170 @@
...
+	return &GitScanner{started: time.Now()}
+}
+
+// Close stops exits once all processing has stopped, and all resources are
+// tracked and cleaned up.
+func (s *GitScanner) Close() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if s.closed {
```
ttaylorr: I think you can remove the mutex lock and unlock here by using a `uint32` and either storing `0` or `1` by using the `sync/atomic` package.

lfs/gitscanner.go
```diff
@@ -0,0 +1,170 @@
...
+func (s *GitScanner) ScanPreviousVersions(ref string, since time.Time) (*PointerChannelWrapper, error) {
+	return logPreviousSHAs(ref, since)
+}
+
+// ScanIndex scans the git index for modified LFS objects.
+func (s *GitScanner) ScanIndex(ref string) (*PointerChannelWrapper, error) {
+	return scanIndex(ref)
+}
+
+func (s *GitScanner) opts(mode ScanningMode) *ScanRefsOptions {
```
ttaylorr: Leaving a note here that this is a temporary measure at least until we can get rid of `*ScanRefsOptions`.

PR #1671, opened 2016-11-17
start reading off the Watch() channel before sending any input

technoweenie:
This ensures that all transfers by the `checkQueue` are set as uploaded. I thought this might also be the cause of a deadlock, but now I don't think so. Nothing is waiting for `len(missing)` items to flow through `transferc`.

### DIFF
--- a/commands/command_prune.go
+++ b/commands/command_prune.go
@@ -114,6 +114,7 @@ func prune(fetchPruneConfig config.FetchPruneConfig, verifyRemote, dryRun, verbo
 	var totalSize int64
 	var verboseOutput bytes.Buffer
 	var verifyc chan string
+	var verifywait sync.WaitGroup
 
 	if verifyRemote {
 		cfg.CurrentRemote = fetchPruneConfig.PruneRemoteName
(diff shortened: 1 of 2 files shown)


ttaylorr:
Auditing the rest of the code to make sure that we haven't made any similar mistakes:

```bash
~/g/git-lfs (fix-dl-check) $ ack "Watch()"                                        
commands/command_fetch.go                                                         
265:            dlwatch := q.Watch()                                              
                                                                                  
commands/command_prune.go                                                         
125:            verifyc = verifyQueue.Watch()                                     
                                                                                  
commands/uploader.go                                                              
105:    transferCh := checkQueue.Watch()                                          
                                                                                  
lfs/transfer_queue.go                                                             
328:// Watch returns a channel where the queue will write the OID of each transfer
330:func (q *TransferQueue) Watch() chan string {
```

- :ok: __`commands/command_fetch.go`__:
The monitoring channel is opened with `Watch()`, and a goroutine is opened to listen _before_ calling `Add()` on the transfer_queue.

- :ok: __`commands/command_prune.go`__:
This command only watches the transfer queue if the `verifyRemote` bool is set to true, and leaves the channel `nil` by default. `Watch()` is called first, then the goroutine is spawned, and then `Add()`s begin to occur. This function correctly handles the case where `verifyRemote` is false, since it does _not_ spawn a goroutine that would block forever, since `range`-ing over a `nil` channel blocks infinitely.

- 🆗  __`commands/uploader.go`__: fixed in this PR.

ttaylorr:
> 🆗 commands/command_prune.go:

I misspoke. `commands/command_prune` _does_ duplicate the same behavior that @technoweenie fixed in this PR, so we need to break [this][1] code out and ensure that it happens before [this][2] code.

[1]: https://github.com/git-lfs/git-lfs/blob/ced527048cab09e4353567df95be44339a8dad66/commands/command_prune.go#L138-L140
[2]: https://github.com/git-lfs/git-lfs/blob/ced527048cab09e4353567df95be44339a8dad66/commands/command_prune.go#L148-L155

PR #1673, opened 2016-11-17
re-enable push and pre-push dry-run flags

technoweenie:
My push/pre-push refactoring (https://github.com/git-lfs/git-lfs/pull/1128) inadvertently broke dry runs of these commands.


technoweenie:
Doh, setting `DryRun` in https://github.com/git-lfs/git-lfs/blob/ced527048cab09e4353567df95be44339a8dad66/commands/uploader.go#L18-L20. False alarm, people.

PR #1685, opened 2016-11-18
1.5/fedora 24

technoweenie:
After shipped v1.5.1, I looked into why Fedora 24 isn't working. Turns out I just needed to add it to the list.

@JonasT: Does this work for you?

### DIFF
--- a/script/packagecloud.rb
+++ b/script/packagecloud.rb
@@ -35,6 +35,7 @@ $distro_name_map = {
     el/7
     fedora/22
     fedora/23
+    fedora/24
   ),
   "debian/7" => %w(
     debian/wheezy


JonasT:
Please note Fedora 25 will be released next Tuesday, so you might want to add that one too

technoweenie:
Doesn't look like [packagecloud supports fedora 25 yet](https://packagecloud.io/docs). If you can confirm that you can install Git LFS v1.5.1 on Fedora 24, I'll go ahead and merge this now.

JonasT:
I just tried it and install with the packagecloud.io script works. I also did some pull, merge & push work on an LFS repo just now, no issues. Thanks!

technoweenie:
Thanks for bringing this up!

PR #1686, opened 2016-11-18
lfs/tq: completely remove legacy support

ttaylorr:
This pull-request removes the remaining relics for supporting the legacy API from the transfer queue.

In particular, we no longer need:

- the `apic` channel (previously used for manually issuing legacy API requests out-of-band with the transfer adapter)
- the `oldApiWorkers` field which used control the number of out-of-band legacy API workers
- some references in the documentation referring to the legacy API

---


ttaylorr:
> Is it possible for q.batcher to be nil?

Great question. I was thinking that I should have removed the `q.batcher` in this PR as I was getting into the one that I'm about to submit. To answer, `q.batcher` can _no longer_ be nil, but it used to be, which is why this check exists.

When we booted up the `*lfs.TransferQueue` in versions of LFS prior to https://github.com/git-lfs/git-lfs/pull/1629, we had to check whether or not we were using the batcher, or operating in legacy mode:

```go
if config.Config.BatchTransfer() {
	tracerx.Printf("tq: running as batched queue, batch size of %d", batchSize)
	q.batcher = NewBatcher(batchSize)
	go q.batchApiRoutine()
} else {
	tracerx.Printf("tq: running as individual queue")
	q.launchIndividualApiRoutines()
}
```

(see: https://github.com/git-lfs/git-lfs/blob/cbf91a9337313449cb45fc09fb80d467f88d1142/lfs/transfer_queue.go#L542-L549)

The batcher is now initialized [here](https://github.com/git-lfs/git-lfs/blob/3a11d6cda0e6d7da54c3eb134d10bb7d186b3ac7/lfs/transfer_queue.go#L453). It _does_ hold the zero-value `(*lfs.Batcher)(nil)` during initialization (for the underlying type is a pointer to an `lfs.Batcher`), but this is before any `Add()`, `Retry()` or similar functions are called.

ttaylorr:
Huh, looks like CircleCI cancelled the build for https://github.com/git-lfs/git-lfs/pull/1686/commits/9eb01b40f4929c553b15304e8c2b6418e401048f as soon as I pushed https://github.com/git-lfs/git-lfs/pull/1686/commits/452dbb00254f60c5b93e65d0f5644c79e20335dc. It appears that AppVeyor did the same, but didn't report any meaningful message.

Going to need to retry these builds to get them off red, even though there are no test failures.

PR #1687, opened 2016-11-18
Release fixes

technoweenie:
Some release fixes that were made for the v1.5.1 release.


script/release.go
```diff
@@ -39,6 +39,11 @@ func mainRelease() {
 		release(rel)
 		fmt.Println()
 	}
+
+	fmt.Println("SHA-256 hashes:")
+	for _, rel := range buildMatrix {
+		fmt.Printf("**%s**\n%s\n\n", rel.Filename, rel.SHA256)
```
ttaylorr: Could move the `\n\n` into a `strings.Join` call to remove the trailing ones:

```go
shas := make([]string, len(buildMatrix)
for i, rel := range buildMatrix {
        shas[i] = fmt.Sprintf("**%s**\n%s", rel.Filename, rel.SHA256)
}

fmt.Println(strings.Join(shas, "\n\n"))
```

PR #1688, opened 2016-11-18
Catfilebatch unittests

technoweenie:
Added unit tests that would have caught the bug that required https://github.com/git-lfs/git-lfs/pull/1680.

Here's the test run on a slightly old version of master, with these 2 commits cherry picked:

```bash
$ gl
commit 35509bfc8836bdd2f5493404caea28c257db90d4
Author: risk danger olson <technoweenie@gmail.com>
Date:   Fri Nov 18 13:48:41 2016 -0700

    add unit tests for 'git cat-file --batch'

commit 65fdf9b6df5fabd19099528f18d95ced3301d333
Author: risk danger olson <technoweenie@gmail.com>
Date:   Fri Nov 18 12:54:56 2016 -0700

    add unit tests for 'cat-file --batch-check'

commit f92bd3b2145b8504753d7b08c27b37befb7556eb
Author: risk danger olson <technoweenie@gmail.com>
Date:   Fri Nov 18 07:44:20 2016 -0700

    remove deprecated features

$ script/test lfs -run TestCatFileBatch
--- FAIL: TestCatFileBatchScanner (0.00s)
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 1000
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 1000
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 1000
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 1000
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 1000
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 128
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 1000
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 1000
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 1000
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 1000
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 1000
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 128
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 1000
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 1000
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 1000
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 1000
	gitscanner_catfilebatchscanner_test.go:88: 0000000000000000000000000000000000000000 blob 1000
        Error Trace:    gitscanner_catfilebatchscanner_test.go:50
	Error:		Expected nil, but got: git cat-file --batch:: Invalid: "\x01\x84\xe2\xa6%\x0e\xaaN\x04\xc1.{K<[`\xb8\xc1\x8fkf\x1ds\xcd15*\xa5\xf3\x9fHWqG\xe8\xb7\xe3\x82F\x1e\x82\xa1\xf4\xfcA\x06m\x8b=s\x17K\x9a2\xbd7\xc1\xa7]\xec\x00;!\xc2Y\x9a?\xc8\xd0EE\x1a\xef|z\xe4\x04\xbf$\x87\xa6\x17Ӵ\x97\xd2\r\xbb\xb7^\xe7\xa5\xcc9F,\xeb߅&\n"

FAIL
FAIL	github.com/git-lfs/git-lfs/lfs	0.010s
```

It's reading the invalid data because `io.LimitReader` wasn't ensuring that the `catFileBatchScanner` is reading all of the non-lfs pointer bytes.


lfs/gitscanner_tree.go
```diff
@@ -41,38 +41,20 @@ func catFileBatchTree(treeblobs *TreeBlobChannelWrapper) (*PointerChannelWrapper
...
-			}
-
-			_, err = cmd.Stdout.ReadBytes('\n') // Extra \n inserted by cat-file
+			p, err := scanner.Next()
 			if err != nil {
+				errchan <- err
 				break
+			} else if p != nil {
+				p.Name = t.Filename
+				pointers <- p
```
technoweenie: This is an interesting pattern. This command synchronously writes to stdin, and then reads the response from stdout. This is so it can associate the returned pointer with the given filename from `ls-tree`. This is incompatible with `Scan()`, since it will keep reading until there's a _valid_ lfs pointer.

I think this synchronous approach is a solid pattern though. Pretty much every scan call needs to associate the filename at least. Other funcs do this by keeping some map of shas and filenames, which is :-1:.

lfs/gitscanner_catfilebatchscanner_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+func fakeReaderWithRandoData(t *testing.T, blobs []*Pointer) io.Reader {
+	buf := &bytes.Buffer{}
+	rngbuf := make([]byte, 1000) // just under blob size cutoff
+	rng := rand.New(rand.NewSource(0))
+
+	for i := 0; i < 5; i++ {
+		n, err := io.ReadFull(rng, rngbuf)
+		if err != nil {
+			t.Errorf("error reading from rng: %+v", err)
```
ttaylorr: `t.Fatalf` will allow you to get rid of the `return` below 👍

lfs/gitscanner_tree.go
ttaylorr: Thanks for the explanation. I'm on board 👍

lfs/gitscanner_catfilebatchcheckscanner_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"bufio"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCatFileBatchCheckScanner(t *testing.T) {
```
ttaylorr: Hm, I think this test name could be more specific: `TestCatFileBatchCheckScansCorrectlyFormattedBlobs`.

ttaylorr: Alternatively could be broken down into smaller tests, i.e., one for "short line", one for invalid sizes, one for long lines, etc.

At that point I'd probably extract a test helper function or type, but I think that'd be a cool improvement in that it'd allow you to name the tests more specifically.

lfs/gitscanner_catfilebatchscanner_test.go
technoweenie: I tend to like `Errorf` so a single test can return multiple errors. Since this is for generating test data, and also should _never happen_ unless something is really wrong, I've changed it to `Fatalf` :+1:

technoweenie:
I think `Next()` works better for these synchronous stdin/stdout commands. It's not very fun to use though:

```go
	scanner := &catFileBatchScanner{r: cmd.Stdout}
	for {
		p, err := scanner.Next()
		if err != nil {
			if err != io.EOF {
				errCh <- err
			}
			break
		} else if p != nil {
			pointerCh <- p
		}
	}
```

One thought is leaving the `io.EOF` logic to the scanner, and returning another bool:

```go
	scanner := &catFileBatchScanner{r: cmd.Stdout}
	for {
		p, shouldBreak, err := scanner.Next()
		if err != nil {
			errCh <- err
		}

		if p != nil {
			pointerCh <- p
		}

		if shouldBreak {
			break
		}
	}
```

Ended up being longer, but it's not nested.

A third option would be passing the channels to the scanner:

```go
scanner := &catFileBatchScanner{r: cmd.Stdout}
for scanner.Scan(pointerCh, errCh) {}

// ok empty for loops are strange...
scanner.ScanInto(pointerCh, errCh)
```

ttaylorr:
I like the second option. Maybe `more` or `next` is more appropriate? I like the way that this reads:

```go
for {
        p, hasNext, err := scanner.Next()
        // ...

        if !hasNext {
                break
        }
}
```

technoweenie:
Cool, going with `hasNext`. Check out the two uses of it in this PR. They're nearly identical, except `catFileBatchTree()` can set `p.Name`. That wasn't possible with my `Into` ideas above.

technoweenie:
All 3 of these have a similar pattern for running commands. They spin up a `*exec.Cmd`, start up a goroutine that synchronously passes things to stdout, waiting for it to return something via stdout.

* `runCatFileBatchCheck`
* `runCatFileBatch`
* `catFileBatchTree`

I want to see if I can clean up `lsTreeBlobs` a bit and get rid of the duplicate `catFileBatchTree` altogether.

lfs/gitscanner_catfilebatch.go
```diff
@@ -23,42 +23,41 @@ func runCatFileBatch(pointerCh chan *WrappedPointer, revs *StringChannelWrapper,
...
 
-	go catFileBatchOutput(pointerCh, cmd, errCh)
-	go catFileBatchInput(cmd, revs, errCh)
-	return nil
-}
+	go func() {
+		scanner := &catFileBatchScanner{r: cmd.Stdout}
+		for r := range revs.Results {
+			cmd.Stdin.Write([]byte(r + "\n"))
+			hasNext := scanner.Scan()
```
ttaylorr: I'm a little confused by how this method works 😱 . My inclination would be that `Scan()` scans for a pointer, and returns whether or not a _subsequent_ scan will succeed.

However, from looking at the code below, it looks like it only returns whether the current scan was OK, and returns false even if we're at the end of the stream.

Should the scanner buffer two (or n) pointers instead of just one? If you end up implementing this, you'll have to be careful that `io.Reader`s can return EOF at the last read (n>0, io.EOF) or directly _after_ the last read (0, io.EOF).

lfs/gitscanner_catfilebatch.go
```diff
@@ -77,59 +76,52 @@ func (s *catFileBatchScanner) Err() error {
...
-			}
-		}
+	if size != read {
+		return nil, fmt.Errorf("expected %d bytes, read %d bytes", size, read)
+	}
 
-		_, err = r.ReadBytes('\n') // Extra \n inserted by cat-file
-		if err != nil {
-			return nil, err
+	p, err := DecodePointer(bytes.NewBuffer(buf[0:read]))
```
ttaylorr: Nit: this could be `DecodePointer(bytes.NewBuffer(buf[:read]))`.

lfs/gitscanner_catfilebatchcheck.go
```diff
@@ -19,99 +18,87 @@ func runCatFileBatchCheck(smallRevCh chan string, revs *StringChannelWrapper, er
...
-	err     error
 }
 
 func (s *catFileBatchCheckScanner) BlobOID() string {
 	return s.blobOID
 }
 
 func (s *catFileBatchCheckScanner) Err() error {
-	return s.err
+	return s.s.Err()
```
ttaylorr: Interesting. I think I might prefer keeping the old way (maintaining an `err error` member of the struct) and assigning that member when `next()`-ing. That way, if we change the underlying implementation to no longer use a `*bufio.Scanner` (or anything that has an `Err()` method) we're only updating the `next()` method, not this one.

lfs/gitscanner_catfilebatchcheckscanner_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"bufio"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCatFileBatchCheckScannerWithValidOutput(t *testing.T) {
+	lines := []string{
+		"short line",
```
ttaylorr: Digging the `assertNextOID` helper func. I think that combined with the contents of each line ("not-a-size", "short-line", etc) is descriptive enough 👍

lfs/gitscanner_catfilebatch.go
ttaylorr: OK, after some meat-space conversation, I get this now 👍 . `hasNext` returns whether or not we can scan again, and we may or _may not_ get a pointer, or an error. I think renaming this to `canScanAgain` and documenting the `Scan()` method itself would be ✨

lfs/gitscanner_catfilebatchcheck.go
technoweenie: Yeah, I could go either way here. I just went with the one with fewer LOC. I have a hunch that if this code is refactored to get rid of `*bufio.Scanner`, we'll be moving to libgit2 or something completely different.

PR #1689, opened 2016-11-18
Embed the version information into the Windows executable

sschuberth:
I'm currently looking into reducing the number of places where the version is defined, so that less places need to be touched when bumping the version for a new release, like e.g. in 4e5ba415c41c42ceb9b91b89f0a3655740f3c276.

But I have a few questions in this regard. To start with, where are the values for the variables in [this line](https://github.com/git-lfs/git-lfs/blob/master/rpm/SPECS/git-lfs.spec#L9) of the RPM spec defined? Do these come from [a few lines above](https://github.com/git-lfs/git-lfs/blob/master/rpm/SPECS/git-lfs.spec#L1-L2)? I'm asking because these definitions differ in the case from the variable use.


ttaylorr:
> I'm currently looking into reducing the number of places where the version is defined, so that less places need to be touched when bumping the version for a new release, like e.g. in 4e5ba41.

I'm on board with this 👍 . I think the reason that they are defined in more than one place has to do with preventing circular dependencies, though I'm neither happy with the existence of this problem or its (current) solution. 

> To start with, where are the values for the variables in this line of the RPM spec defined?

I'm not sure, but @andyneff might have a better idea 👍 

[1]: https://github.com/git-lfs/build-dockers

sschuberth:
While looking more into this, I realized we should probably also add proper version information to the Windows executable (and an icon, while at it). My final goal would be to create the official Windows distribution on CI to be fully reproducible, and not on some "blessed" developer's machine.

sschuberth:
Ok, so I changed the topic of this PR a bit ... in fact it's now not less places where some sort of version number is defined. But at least there now is a canonical place where to bump the version information for the Windows build: In the new `resources.rc` file.

lfs/lfs.go
```diff
@@ -16,10 +16,6 @@ import (
 	"github.com/rubyist/tracerx"
 )
 
-const (
-	Version = "1.5.0"
```
ttaylorr: I think I'd rather have the version stay here than in `config`, since it makes more sense to say `lfs.Version` than `config.Version` and have to look at your import paths to realize that the `config` package comes from `github.com/git-lfs/git-lfs`.

script/windows-installer/resources.rc
```diff
@@ -0,0 +1,20 @@
+main ICON "git-lfs-logo.ico"
+1 VERSIONINFO
+FILEVERSION 1,5,0,0
```
ttaylorr: Is this a part of the LFS version, or a version for the `.rc` format?

lfs/lfs.go
sschuberth: I can do that. It also solves the issue of `Version` being a `var` in `config/version.go`. IMO it should be a `const` like it is here.

sschuberth: Turns out that when I import `github.com/git-lfs/git-lfs/lfs` in `config/version.go` in order to use `lfs.Version` I get `import cycle not allowed` :-( Any suggestion how to solve that?

ttaylorr: Ah, that's because the `github.com/git-lfs/git-lfs/lfs` package already has a dependency pointing towards `github.com/git-lfs/git-lfs/config`, introducing another dependency the other way around would be a circular dependency issue. Let's just leave it in the `config` package for now, even though it's kind of unfortunate.

Untangling that stuff would be more time than I think it's worth at this stage.

technoweenie: FWIW I think it should _eventually_ in `./git-lfs.go` or something:

```go
import gitlfs "github.com/git-lfs/git-lfs"
import "fmt"

func main() {
  fmt.Println(gitlfs.Version)
}
```

Oh god, I just realized `git-lfs` is not a valid go package. What have we donnnnnne.

As far as this PR is concerned, `config.Version` is just fiiiine :+1:

sschuberth: BTW, I was wondering why the version still is 1.5.0 in master. Don't you merge back the release branches to master?

script/windows-installer/inno-setup-git-lfs-installer.iss
```diff
@@ -1,9 +1,9 @@
-; Script generated by the Inno Script Studio Wizard.
-; SEE THE DOCUMENTATION FOR DETAILS ON CREATING INNO SETUP SCRIPT FILES!
-
 #define MyAppName "Git LFS"
-#define MyAppVersion "1.5.0"
-#define MyAppPublisher "GitHub, Inc"
+
+; Misuse RemoveFileExt to strip the 4th patch-level version number.
+#define MyAppVersion RemoveFileExt(GetFileVersion("..\..\git-lfs-x86.exe"))
```
sschuberth: Nope, I arbitrarily chose the x86 version. As both executables have the version information embedded it does not matter which one Inno Setup is looking at here. But I should make that clearer by adding a comment.

lfs/lfs.go
technoweenie: Nope, `release-1.5` is basically its own thing. Bug fixes will have to be backported to it. The `master` branch is diverging a lot already, with the scanner refactoring PRs.

I suppose we should change it to `2.0.0-dev` or something though.

sschuberth:
I made a few minor updates, but I still wonder about how to conveniently update the version numbers across all places we now have. Should I add a shell script that runs `sed` or something?

technoweenie:
We basically just look at previous releases to see what needs to change. I don't think a script is necessary for this PR, but would be welcome.

sschuberth:
> I don't think a script is necessary for this PR, but would be welcome.

I've rebased the PR and added such a script.

sschuberth:
Looks like this is again waiting forever for CircleCI to return its result. Is this a known issue?

ttaylorr:
> Looks like this is again waiting forever for CircleCI to return its result. Is this a known issue?

We had some issues last week where we'd exceeded our allotted build time for the month, but that issue has since been resolved. I'm not seeing any other weird changes to the `circle.yml` file in your tree, so perhaps you could try updating your branch (again) and hope that that kicks of a build?

sschuberth:
> so perhaps you could try updating your branch (again) and hope that that kicks of a build?

Well, I *did* that just a few minutes ago, but I certainly can try again ...

sschuberth:
I don't see my PR being built at https://circleci.com/gh/git-lfs/git-lfs, so maybe we're indeed out of build time again?

ttaylorr:
> maybe we're indeed out of build time again?

Hm, doesn't seem like it: the last PR I just pushed is [building now][1]. Going to loop in @technoweenie since he owns the Circle account.

[1]: https://circleci.com/gh/git-lfs/git-lfs/343?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

sschuberth:
Circle CI still is "Waiting for status to be reported", @technoweenie could you please advise?

technoweenie:
@sschuberth Our Circle trial expired, but we updated the billing recently. Can you update the branch with the latest from master, to kick the jobs off again? I'd do it, but you posted this on your fork.

sschuberth:
@technoweenie How recently did you update the billing? My last change to this PR is from 6 hours ago, but still Circle CI did not seem to kick off.

technoweenie:
Hmm... We're at 18% of our allotted build minutes. No errors from Circle CI. SHRUG. Just going to merge it, it's not like this PR adds anything remotely risky from mac builds.

sschuberth:
Thanks!

PR #1694, opened 2016-11-21
Key value store

sinbad:
This PR adds an in-memory, optimistically locked key-value store. It's basically a replacement for Boltdb for caching active locks locally as requested in the lock PR.

Submitted separately because it's generic and hopefully easy to review. Rewrite of lock PR is still ongoing but will use this.

### DIFF
--- a/tools/keyvaluestore.go
+++ b/tools/keyvaluestore.go
@@ -105,6 +105,7 @@ func (k *KeyValueStore) Save() error {
 		k.loadAndMergeReaderIfNeeded(f)
 		// Now we overwrite the file
 		f.Seek(0, os.SEEK_SET)
+		f.Truncate(0)
 	}
 
 	k.version++
--- a/tools/keyvaluestore.go
+++ b/tools/keyvaluestore.go
@@ -50,15 +50,17 @@ type keyValueStoreData struct {
 func NewKeyValueStore(filepath string) (*KeyValueStore, error) {
 
 	var kvdata *keyValueStoreData
+	stat, _ := os.Stat(filepath)
 	f, err := os.OpenFile(filepath, os.O_RDONLY, 0664)
-	if err == nil {
+	// Only try to load if file has some bytes in it
+	if err == nil && stat.Size() > 0 {
 		defer f.Close()
 		dec := gob.NewDecoder(f)
 		err := dec.Decode(&kvdata)
 		if err != nil {
 			return nil, fmt.Errorf("Problem loading key/value data from %v: %v", filepath, err)
 		}
-	} else if !os.IsNotExist(err) {
+	} else if err != nil && !os.IsNotExist(err) {
 		return nil, err
 	}
 	if kvdata == nil {
@@ -114,27 +116,33 @@ func (k *KeyValueStore) Save() error {
 	if err != nil {
 		return err
 	}
+	stat, _ := os.Stat(k.filename)
 
 	defer f.Close()
-	// Decode *only* the version field to check whether anyone else has
-	// modified the db; gob serializes structs in order so it will always be 1st
-	dec := gob.NewDecoder(f)
-	var versionOnDisk int64
-	err = dec.Decode(&versionOnDisk)
-	if err != nil {
-		return fmt.Errorf("Problem checking version of key/value data from %v: %v", k.filename, err)
-	}
-	if versionOnDisk != k.data.version {
-		// Reload data & merge
-		var dbOnDisk map[string]interface{}
-		err = dec.Decode(&dbOnDisk)
+
+	// Only try to peek if > 0 bytes, ignore empty files (decoder will fail)
+	if stat.Size() > 0 {
+		var versionOnDisk int64
+		// Decode *only* the version field to check whether anyone else has
+		// modified the db; gob serializes structs in order so it will always be 1st
+		dec := gob.NewDecoder(f)
+		err = dec.Decode(&versionOnDisk)
 		if err != nil {
-			return fmt.Errorf("Problem reading updated key/value data from %v: %v", k.filename, err)
+			return fmt.Errorf("Problem checking version of key/value data from %v: %v", k.filename, err)
 		}
-		k.reapplyChanges(dbOnDisk)
+		if versionOnDisk != k.data.version {
+			// Reload data & merge
+			var dbOnDisk map[string]interface{}
+			err = dec.Decode(&dbOnDisk)
+			if err != nil {
+				return fmt.Errorf("Problem reading updated key/value data from %v: %v", k.filename, err)
+			}
+			k.reapplyChanges(dbOnDisk)
+		}
+		// Now we overwrite
+		f.Seek(0, os.SEEK_SET)
 	}
-	// Now we overwrite
-	f.Seek(0, os.SEEK_SET)
+
 	k.data.version++
 
 	enc := gob.NewEncoder(f)
--- a/tools/keyvaluestore.go
+++ b/tools/keyvaluestore.go
@@ -170,6 +170,7 @@ func (k *KeyValueStore) loadAndMergeReaderIfNeeded(f io.Reader) error {
 			return fmt.Errorf("Problem reading updated key/value data from %v: %v", k.filename, err)
 		}
 		k.reapplyChanges(dbOnDisk)
+		k.version = versionOnDisk
 	}
 	return nil
 }
--- a/tools/keyvaluestore.go
+++ b/tools/keyvaluestore.go
@@ -1,0 +1,164 @@
+package tools
+
+import (
+	"encoding/gob"
+	"fmt"
+	"os"
+	"sync"
+)
+
+// KeyValueStore provides an in-memory key/value store which is persisted to
+// a file. The file handle itself is not kept locked for the duration; it is
+// only locked during load and save, to make it concurrency friendly. When
+// saving, the store uses optimistic locking to determine whether the db on disk
+// has been modified by another process; in which case it loads the latest
+// version and re-applies modifications made during this session. This means
+// the Lost Update db concurrency issue is possible; so don't use this if you
+// need more DB integrity than Read Committed isolation levels.
+type KeyValueStore struct {
+	mu       sync.RWMutex
+	filename string
+	data     *keyValueStoreData
+	log      []keyValueChange
+}
+
+type keyValueOperation int
+
+const (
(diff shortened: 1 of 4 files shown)


sinbad:
Hold off a minute, I think there may be an issue with writing a DB that's truncated smaller, need to add a test for that.

sinbad:
OK, handled that case now; actually still worked (`gob` protocol is pretty robust) but shouldn't leave old data lying around.

tools/keyvaluestore.go
```diff
@@ -0,0 +1,201 @@
...
+	k.mu.Lock()
+	defer k.mu.Unlock()
+
+	// Short-circuit if we have no changes
+	if len(k.log) == 0 {
+		return nil
+	}
+
+	// firstly peek at version; open read/write to keep lock between check & write
+	f, err := os.OpenFile(k.filename, os.O_RDWR|os.O_CREATE, 0664)
```
ttaylorr: Thinking that we could do this eagerly and maintain a handle on the `*os.File` from the instance, instead of this method.

Since I assume that we `Save()` more than once, only having to do this somewhat expensive operation once would be a win for me.

tools/keyvaluestore.go
```diff
@@ -0,0 +1,201 @@
...
+
+// Save persists the changes made to disk
+// If any changes have been written by other code they will be merged
+func (k *KeyValueStore) Save() error {
+	k.mu.Lock()
+	defer k.mu.Unlock()
+
+	// Short-circuit if we have no changes
+	if len(k.log) == 0 {
+		return nil
```
ttaylorr: Does `k.mu` lock `k.log`?

tools/keyvaluestore.go
```diff
@@ -0,0 +1,201 @@
...
+	if stat.Size() > 0 {
+		k.loadAndMergeReaderIfNeeded(f)
+		// Now we overwrite the file
+		f.Seek(0, os.SEEK_SET)
+		f.Truncate(0)
+	}
+
+	k.version++
+
+	enc := gob.NewEncoder(f)
```
ttaylorr: Actually, we may want to just hold onto this instance instead of the file. Perhaps both, since we need to `f.Seek` and `f.Truncate`.

tools/keyvaluestore.go
```diff
@@ -0,0 +1,201 @@
...
+	err = enc.Encode(k.version)
+	if err != nil {
```
ttaylorr: Nit: inline this:

```go
if err = enc.Encode(k.Version); err != nil {
        // ...
}
```

tools/keyvaluestore.go
```diff
@@ -0,0 +1,201 @@
...
+		return fmt.Errorf("Error while writing version data to %v: %v", k.filename, err)
+	}
+	err = enc.Encode(k.db)
+	if err != nil {
```
ttaylorr: See above.

tools/keyvaluestore.go
```diff
@@ -0,0 +1,201 @@
...
+// KeyValueStore provides an in-memory key/value store which is persisted to
+// a file. The file handle itself is not kept locked for the duration; it is
+// only locked during load and save, to make it concurrency friendly. When
+// saving, the store uses optimistic locking to determine whether the db on disk
+// has been modified by another process; in which case it loads the latest
+// version and re-applies modifications made during this session. This means
+// the Lost Update db concurrency issue is possible; so don't use this if you
+// need more DB integrity than Read Committed isolation levels.
+type KeyValueStore struct {
+	mu       sync.RWMutex
```
ttaylorr: I think it would be helpful if you documented that this mutex locks the _entire_ `*KeyValueStore` instance.

tools/keyvaluestore.go
```diff
@@ -0,0 +1,201 @@
...
+	filename string
+	log      []keyValueChange
+
+	// This is the persistent data
+	// version for optimistic locking, this field is incremented with every Save()
+	version int64
+	db      map[string]interface{}
+}
+
+type keyValueOperation int
```
ttaylorr: Possible to use a `byte`, since it's smaller than an `int`?

tools/keyvaluestore.go
```diff
@@ -0,0 +1,201 @@
+package tools
```
ttaylorr: I think this would be appropriate to put in a sub-package of `tools`: `github.com/git-lfs/git-lfs/tools/kv`, which would allow us to get rid of a lot of `KeyValue*` prefixes.

tools/keyvaluestore.go
```diff
@@ -0,0 +1,201 @@
...
+	key       string
+	value     interface{}
+}
+
+// NewKeyValueStore creates a new store and initialises it with contents from
+// the named file, if it exists
+func NewKeyValueStore(filepath string) (*KeyValueStore, error) {
+	kv := &KeyValueStore{filename: filepath, db: make(map[string]interface{})}
+	err := kv.loadAndMergeIfNeeded()
+	return kv, err
```
ttaylorr: Nit: since you can return non-nil values for _both_, this could be simplified to:

```go
return kv, kv.loadAndMergeIfNeeded()
```

tools/keyvaluestore.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// Remove removes the key and its value from the store in memory
+// Changes are not persisted until you call Save()
+func (k *KeyValueStore) Remove(key string) {
+	k.mu.Lock()
+	defer k.mu.Unlock()
+
+	delete(k.db, key)
+	k.log = append(k.log, keyValueChange{keyValueRemoveOperation, key, nil})
```
ttaylorr: Looks like these transactional methods are pretty similar. What do you think about teaching the `keyValueChange` type how to perform these operations?

tools/keyvaluestore.go
```diff
@@ -0,0 +1,201 @@
...
+	// Do nothing if empty file
+	if stat.Size() == 0 {
+		return nil
+	}
+
+	f, err := os.OpenFile(k.filename, os.O_RDONLY, 0664)
+	if err == nil {
+		defer f.Close()
+		return k.loadAndMergeReaderIfNeeded(f)
+	} else if !os.IsNotExist(err) {
```
ttaylorr: Is this possible given the above?

tools/keyvaluestore.go
sinbad: Yeah, although because of data alignment, in practice you don't actually save anything unless you have an array of them :)

tools/keyvaluestore.go
sinbad: No, this is completely deliberate, and key to the optimistic locking technique. If you hold on to the handle, it will mean that parallel instances of git-lfs will block each other, which is limiting. The whole point of the optimistic locking system is to allow multiple instances of git-lfs to operate on shared key value stores and not end up locking each other out, while also resolving any parallel update issues.

Opening a file isn't that expensive and not unexpected on a `Save()` call. I doubt we'll be calling it very often; optimally you only call it once just before exit because it's an in-memory store. Conflicts are resolved at save time to avoid having to maintain stronger locks (this is a common DB trick).

tools/keyvaluestore.go
sinbad: OK, can I call it 'keyvalue' for clarity? I even considered the option of making it a separate top-level package since it's not even specific to `git-lfs`...

ttaylorr: Let's stick with `tools/kv` since `kv` is a common abbreviation.

tools/keyvaluestore.go
ttaylorr: Ah, damn!

tools/keyvaluestore.go
sinbad: Implicitly yes; it's a requirement that `mu` is always locked in the body of functions, either read-only or a full lock on mutating functions.

tools/keyvaluestore.go
sinbad: No, see comments above. Being able to let go of the file handle is a deliberate feature :)

tools/keyvaluestore.go
sinbad: Not any more 👍

tools/kv/keyvaluestore.go
```diff
@@ -0,0 +1,204 @@
...
+	if len(k.log) == 0 {
+		return nil
+	}
+
+	// firstly peek at version; open read/write to keep lock between check & write
+	f, err := os.OpenFile(k.filename, os.O_RDWR|os.O_CREATE, 0664)
+	if err != nil {
+		return err
+	}
+	stat, _ := os.Stat(k.filename)
```
technoweenie: Small nit, but you can call `f.Stat()` here.

tools/kv/keyvaluestore.go
```diff
@@ -0,0 +1,204 @@
...
+
+// reapplyChanges replays the changes made since the last load onto baseDb
+// and stores the result as our own DB
+func (k *Store) reapplyChanges(baseDb map[string]interface{}) {
+	for _, change := range k.log {
+		switch change.op {
+		case setOperation:
+			baseDb[change.key] = change.value
+		case removeOperation:
+			delete(baseDb, change.key)
```
technoweenie: There's a bit of duplication here with `Set()` and `Remove()`. Though since this is just a wrapper around a map, it's probably fine for now. One way to solve it is to turn `change` into an interface, and calling `Apply()` on `*setKey` or `*removeKey` implementations in `reapplyChanges()` and `logChange()`.

It looks like a mutex is needed here, but I think it's fine this code path is only called on `Save()` (which has a mutex) and `NewStore()` (which shouldn't need one).

ttaylorr:
Hmm.. I think we have a new intermittent failure:

```
test: clone in current directory ...                               FAILED
```

I've seen this one pop up a few times here and there, and only on Windows. Haven't done any investigation, so I'm not sure what's going on 😭 

I'm going to re-run the tests for now, since your PR didn't introduce the transient failure.

sinbad:
Gah, yeah I saw this a couple of weeks ago in an unrelated PR. Only seems to affect AppVeyor and I can't reproduce it on Windows :/

sinbad:
Re-targetted this at `locking-master` which will be my integration point for locking stuff from now on, so I can submit incrementally without the overhead of needing APIs to be final at every point. Appveyor problems went away again after I rebased.

PR #1695, opened 2016-11-21
rewrite logScanner using the scanner pattern

technoweenie:


technoweenie:
This rewrites the log parsing stuff to use the scanner pattern. Ready for review and merge.

There are a lot of ways to rewrite `scan()`, but I'd rather move on with the larger refactoring package. I'll put any style complaints there into an issue for someone else to tackle later.

PR #1696, opened 2016-11-21
extract a filepathfilter package

technoweenie:
This is an exploration of what a `filepathfilter` package could look like. Some benefits:

1. `*logScanner` only has to pass a single value around. It no longer needs to be bound to the internal `tools` package.
2. The `*logScanner` constructor only takes 2 args now. The filter is an optional thing to set. The default filter `nil` is just fine.
3. An immutable `filepathfilter.Filter` gives us a place to optimize repeated calls. More below.
4. A bit pie-in-the-sky, but we could depend on a small interface that's just `Allows(string) bool`, in case there are other filter implementations we may want. Doubt it though :)

`tools.FileMatch()` works well, and has a large test suite, some patterns (wildcards with no path separators, like `*.zip`, or double `**` wildcards) have to create regexes on each call. By simply introducing some new types, and doing this work ahead of time gives us the following savings:

```
BenchmarkToolsIncludeWildcardOnly-8      	     100	  16927573 ns/op
BenchmarkFilterIncludeWildcardOnly-8     	    2000	    982018 ns/op
BenchmarkToolsIncludeDoubleAsterisk-8    	     100	  21409532 ns/op
BenchmarkFilterIncludeDoubleAsterisk-8   	     300	   4350653 ns/op
```


ttaylorr:
Looks good. Not sure I'm totally on board with making a new package just for this type, do you think there are any other appropriate places it could go?

technoweenie:
I put it in a new package since it's the kind of thing I could see another tool wanting to use from Git LFS. Otherwise I'd just lump it in `tools`.

ttaylorr:
Hm, I kind of want to put it in `tools`, but that would be 👎 since it'd make _some_ of that package's API stable and others not.

technoweenie:
Both `lfs` and `tools` are "internal" packages... as in we _hope_ no one else is importing and using them :) I'd like to move towards having well designed packages for Git LFS devs and any 3rd party devs importing these packages. 

I like what https://blog.golang.org/package-names says about package names:

>Avoid meaningless package names. Packages named `util`, `common`, or `misc` provide clients with no sense of what the package contains. This makes it harder for clients to use the package and makes it harder for maintainers to keep the package focused. Over time, they accumulate dependencies that can make compilation significantly and unnecessarily slower, especially in large programs. And since such package names are generic, they are more likely to collide with other packages imported by client code, forcing clients to invent names to distinguish them.

technoweenie:
Just removed `tools.FilenamePassesIncludeExcludeFilter()` and `tools.FileMatch`. Everything in the code bass uses `filepathfilter`. :metal:

commands/command_filter_process.go
```diff
@@ -73,6 +72,12 @@ func filterCommand(cmd *cobra.Command, args []string) {
 		ExitWithError(err)
 	}
 
+	skip := filterSmudgeSkip
```
ttaylorr: I think you could simplify this to:

```go
skip := filterSmudgeSkip || cfg.Os.Bool("GIT_LFS_SKIP_SMUDGE", false)
```

which I find to be more readable in context.

Issue #1697, opened 2016-11-21
external filter 'git-lfs filter-process' failed


eminence:
Currently using git 2.11.0.rc2 (I want to test the new process filter stuff), and git-lfs 1.5.1

I'm seeing this error which I can't explain:

```
lxd@sulfursim:/scratch/lxd/08> git init test-lfs
Initialized empty Git repository in /a/tmp/lxd/08/test-lfs/.git/

lxd@sulfursim:/scratch/lxd/08> cd test-lfs/

lxd@sulfursim:/scratch/lxd/08/test-lfs> unzip $HOME/repro.zip
Archive:  /u/users/lxd/repro.zip
   creating: prereqs/
   creating: prereqs/A/
   creating: prereqs/A/foo.m/
 extracting: prereqs/A/foo.m/A.txt
   creating: prereqs/A/bar.m/
  inflating: prereqs/A/bar.m/B.txt

lxd@sulfursim:/scratch/lxd/08/test-lfs> git lfs track "prereqs/**"
Tracking prereqs/**

lxd@sulfursim:/scratch/lxd/08/test-lfs> cat .gitattributes
prereqs/** filter=lfs diff=lfs merge=lfs -text

lxd@sulfursim:/scratch/lxd/08/test-lfs> git add prereqs/
error: external filter 'git-lfs filter-process' failed
fatal: prereqs/A/foo.m/A.txt: clean filter 'lfs' failed

lxd@sulfursim:/scratch/lxd/08/test-lfs> find prereqs/
prereqs/
prereqs/A
prereqs/A/foo.m
prereqs/A/foo.m/A.txt
prereqs/A/bar.m
prereqs/A/bar.m/B.txt

lxd@sulfursim:/scratch/lxd/08/test-lfs> git lfs logs

lxd@sulfursim:/scratch/lxd/08/test-lfs> git lfs env
git-lfs/1.5.1 (GitHub; linux amd64; go 1.7.3; git 4e5ba41)
git version 2.11.0.rc2

LocalWorkingDir=/a/tmp/lxd/08/test-lfs
LocalGitDir=/a/tmp/lxd/08/test-lfs/.git
LocalGitStorageDir=/a/tmp/lxd/08/test-lfs/.git
LocalMediaDir=/a/tmp/lxd/08/test-lfs/.git/lfs/objects
LocalReferenceDir=
TempDir=/a/tmp/lxd/08/test-lfs/.git/lfs/tmp
ConcurrentTransfers=3
[...]

PR #1698, opened 2016-11-22
filter-process: exit with error if we detect an unknown command from Git

larsxschneider:
This fixes the error reporting for https://github.com/git-lfs/git-lfs/issues/1697


ttaylorr:
Thanks, this looks good 👍. I switched from the `%s` formatter, which yields: `string` to the `'%q` formatter, which yields: `'string'`, and then removed the surrounding `'`'s.

PR #1699, opened 2016-11-22
filter-process:  fix reading 1024 byte files

larsxschneider:
c.f. https://github.com/git-lfs/git-lfs/issues/1697

### DIFF
--- a/test/test-filter-process.sh
+++ b/test/test-filter-process.sh
@@ -6,15 +6,7 @@
 # from the "next" branch, which is the only (current) version of Git that has
 # support for the filter protocol.
 #
-# Once 2.11 is released, replace this with:
-#
-# ```
-# ensure_git_version_isnt $VERSION_LOWER "2.11.0"
-# ```
-if [ "1" -ne "$(git version | cut -d ' ' -f 3 | grep -c "g")" ]; then
-  echo "skip: $0 git version does not include support for filter protocol"
-  exit
-fi
+ensure_git_version_isnt $VERSION_LOWER "2.11.0"
 
 begin_test "filter process: checking out a branch"
 (
--- a/test/test-filter-process.sh
+++ b/test/test-filter-process.sh
@@ -95,8 +95,8 @@ begin_test "filter process: add a file with 1024 bytes"
 (
   set -e
 
-  mkdir repo
-  cd repo
+  mkdir repo-issue-1697
+  cd repo-issue-1697
   git init
   git lfs track "*.dat"
   dd if=/dev/zero of=first.dat bs=1024 count=1
--- a/test/test-filter-process.sh
+++ b/test/test-filter-process.sh
@@ -97,3 +97,21 @@ begin_test "filter process: adding a file"
   diff -u <(echo "$expected") <(echo "$got")
 )
 end_test
+
+# https://github.com/git-lfs/git-lfs/issues/1697
+begin_test "filter process: add a file with 1024 bytes"
+(
+  set -e
+
+  mkdir repo
+  cd repo
+  git init
+  git lfs track "*.dat"
+  dd if=/dev/zero of=first.dat bs=1024 count=1
+  printf "any contents" > second.dat
+  git add .
+)
+end_test
+
+
+
--- a/git/pkt_line_reader.go
+++ b/git/pkt_line_reader.go
@@ -54,7 +54,7 @@ func (r *pktlineReader) Read(p []byte) (int, error) {
 		// Mark that we have read "nn" bytes into "p"
 		n += nn
 
-		if n >= len(p) {
+		if n > len(p) {
 			break
 		}
 	}
--- a/git/pkt_line_reader.go
+++ b/git/pkt_line_reader.go
@@ -53,10 +53,6 @@ func (r *pktlineReader) Read(p []byte) (int, error) {
 
 		// Mark that we have read "nn" bytes into "p"
 		n += nn
-
-		if n > len(p) {
-			break
-		}
 	}
 
 	return n, nil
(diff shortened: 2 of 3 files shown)


test/test-filter-process.sh
```diff
@@ -6,15 +6,7 @@
...
-# Once 2.11 is released, replace this with:
-#
-# ```
-# ensure_git_version_isnt $VERSION_LOWER "2.11.0"
-# ```
-if [ "1" -ne "$(git version | cut -d ' ' -f 3 | grep -c "g")" ]; then
-  echo "skip: $0 git version does not include support for filter protocol"
-  exit
-fi
+ensure_git_version_isnt $VERSION_LOWER "2.11.0"
```
larsxschneider: This seems to work but it still generates a few error messages:
```
test/testhelpers.sh: line 416: 10#rc2: value too great for base (error token is "10#rc2")
test/testhelpers.sh: line 420: 10#rc2: value too great for base (error token is "10#rc2")
```
See: https://travis-ci.org/git-lfs/git-lfs/jobs/177907753

... Oh, a few lines down I read this:
```
skip: test/test-install-custom-hooks-path-unsupported.sh (git version > 2.9.0)
```
The git version is clear greater than 2.9, right?

However, this is a separate issue and unrelated to the 1024 error.

ttaylorr: These "custom hook paths" tests have two parts:

- The `test/test-install-custom-hooks-path-unsupported.sh` runs in versions of Git that _don't_ support custom hooks paths (i.e., Git <2.9.0), and ensures that the custom hooks directory is ignored, and the hooks are placed in `.git/hooks/`.
- the `test/test-install-custom-hooks-path.sh` runs on versions of Git that _do_ support custom hooks paths (i.e., Git >=2.9.0), and ensures that the custom hooks directory is respected. 

There's no (current) clean way to do this in one file, so I split it into two, and only one of them runs at a time.

larsxschneider:
Heads up: this seems to work for me on macOS and Linux but I have a hanging process on Windows. Unfortunately I cannot share the repository that creates the hanging process and I haven't been able to create a small test case, yet.

ttaylorr:
> I have a hanging process on Windows.

Are you running Go 1.7.3 on the Windows box?

larsxschneider:
Ah!! There was something about 1.7...  ... let me try!

ttaylorr:
OK, I am convinced that this is the correct solution. Here's why:

When we try to clean a file, the first thing that we do is see if that file is already an LFS pointer, and if it is, we pass that file through directly. To check if the file is an LFS pointer, we only read part of it, since the whole file could be rather large. We read the first 1024 bytes of the file (as dictated by the [`blobSizeCutoff`][1].

We buffer the data in that read by passing it through a `io.TeeReader` into a `bytes.Buffer` (so we can use the data later on if the file wasn't already a pointer). This happens [here][2]. Here's where things get interesting. We combine the buffered data with the actual reader (an `*os.File`, usually) _if_ we buffered less data than the total length of the file. That happens [here][3]. We only concatenate if there is more data, since if we had read the entire contents of the file, there'd be no more data left to read, and the call would either a) block forever, or b) return an `io.EOF` immediately.

However! Our `pkt-line` implementation of `io.Reader` splits the last read and the EOF read into two. So if you're at the end of a file, you'll get:

1. `Read([1024]byte)` => `(1024, nil)`
2. `Read[n]byte)` => `(0, io.EOF)`

This is totally valid according to the Go documentation for the `io.Reader` interface, but introduces a __very subtle__ problem in our code. Consider a 1024 byte file, and the `pkt-line` representation of it. We'd have:

1. A header with the pathname and command (consumed by the filter-process scanner)
2. A flush packet (consumed by the filter-process scanner)
3. 1024 bytes of the file's contents (consumed by the reader above)
4. A flush packet (NOT CONSUMED!)

Since we filled the buffer, and didn't reach into the next packet (which would have been a `0000` flush), the stream is left 4 bytes behind where it should be, thus the next command isn't recognized, and the command exits in a dirty state.

This fix is correct because it combines the last read into:

1. `Read([1024]byte)` => `(1024, io.EOF)`

because the reader reads packets until it has to buffer, _not_ until the given `p` buffer is full. This causes it to read the flush packet against a perfectly sized buffer, which advances the buffer to the right state, even though we don't concatenate the readers. 

This behavior is covered under our unit tests [here][4] and [here][5].

[1]: https://github.com/git-lfs/git-lfs/blob/13018d77986b37df09aea937352365f8891ce022/lfs/scanner.go#L6-L8
[2]: https://github.com/git-lfs/git-lfs/blob/13018d77986b37df09aea937352365f8891ce022/lfs/pointer_clean.go#L79-L83
[3]: https://github.com/git-lfs/git-lfs/blob/13018d77986b37df09aea937352365f8891ce022/lfs/pointer_clean.go#L85-L90
[4]: https://github.com/git-lfs/git-lfs/blob/161d2c853365342c6153d38a4b424ff871d93994/git/pkt_line_reader_test.go#L138
[5]: https://github.com/git-lfs/git-lfs/blob/161d2c853365342c6153d38a4b424ff871d93994/git/pkt_line_reader_test.go#L142

eminence:
I'd be happy to test this new version on my set of test data (the data that found this issue).   It totals about 10GB and 138000 files.  Would you happen to have a 64-bit linux binary handy?

ttaylorr:
@eminence: sure. Here are builds for both AMD64 and 386 architectures built against Linux containing all of the commits in this PR:

```bash
~/g/git-lfs (lars/test-1024-error) $ git show -q HEAD
commit 0899760263ad35ef11bd7e8af2799b8ea33ae624
Author: Taylor Blau <me@ttaylorr.com>
Date:   Tue Nov 22 10:57:18 2016 -0700

    git: remove extraneous length check

    `n`, representing the total number of bytes we have read into `p`, can never
    become `>len(p)`. As such, let's remove a check for that impossible condition.
```

[git-lfs-linux-386-1.5.0.tar.gz](https://github.com/git-lfs/git-lfs/files/607193/git-lfs-linux-386-1.5.0.tar.gz)
[git-lfs-linux-amd64-1.5.0.tar.gz](https://github.com/git-lfs/git-lfs/files/607194/git-lfs-linux-amd64-1.5.0.tar.gz)

eminence:
Thanks!  It's been running for about 30 minuets now with no errors.  I'll report back here when it completes.

Issue #1690, opened 2016-11-18
Support for Windows Long Paths


DoDoENT:
Hi!

We have a very deep git repository (6 submodules one inside another), and the most deep one has some prebuilt binaries on LFS. When cloning the entire hierarchy on Windows 10, I get an error when the most deep submodule tries to fetch LFS data. I even tried manually with `git lfs fetch`, but same error happens:

    open C:\Users\DoDo\Desktop\Microblink\core-recognizer-api\.git\modules\core-recognizer\modules\core-ocr-engine\modules\core-neural-network\modules\core-utils\modules\opencv-bin\modules\Windows\lfs\objects\incomplete\f98b3d50d1f150b17e3c77b1f6cfb0f77edf9922f93f653a885b6997f380cf9a.tmp: The system cannot find the path specified.

I believe the error is caused by path being too long (280 characters) - I even tried creating a symbolic link on c:\ which points to root git repository and tried fetching from it, but I got the same error. What finally worked was to put my root git directly on c:\ and rename it to a short single-letter name.

Is there any LFS option that will turn on support for long paths? The git has option [`core.longpaths`](http://stackoverflow.com/a/26111092) which I do have enabled, but it seems LFS is not honoring it.

ttaylorr:
Hi, @DoDoENT -- thanks for brining this up! Git LFS should be respecting this option if it's in your `.gitconfig`, but there may be some issues with opening files that have names that long on Windows, which I'm not sure how to handle.

Going to cc @larsxschneider who has some more expertise on debugging these sorts of Windows issues to 👂  his 💭 .

ttaylorr:
Actually, this looks like it might be coming in the next release of Go: ([refs](https://github.com/golang/go/blob/2925427a47f41622f28f84889ad7aade27581144/src/os/path_windows.go#L131-L139)). If the next version of Go hasn't released by the time that we're ready to push the next release of LFS, I'd be okay with temporarily back-porting this function into LFS and using it in conjunction with the `core.longpaths` setting that you mentioned above 👍 .

DoDoENT:
Thank you for quick answers! I am looking forward to a fix.

technoweenie:
I think we should wait for Go 1.8 (or whatever the first go release to include this is), since it seems like it's coming soon. We can ship a new LFS pretty much the day a new Go version is out, even if there are no LFS fixes. We'll just update the binaries. 

[...]

PR #1705, opened 2016-11-22
Fix windows long paths

technoweenie:
Attempting a fix to #1690 for v1.5.x only. I assume that Go 1.8 (which includes this fix in the stdlib) will be out by the time LFS 2.0 is out.

I'm having some problems getting the [failing test](https://github.com/git-lfs/git-lfs/compare/release-1.5...fix-windows-long-paths-test?expand=1) to work (details at https://github.com/git-lfs/git-lfs/issues/1690#issuecomment-262324701). I'd really like to confirm a failing test before these fixes are applied, so that a passing test in this PR proves it actually worked.


ttaylorr:
Yikes. Did not mean to click "Request Changes". Merge away! 😄

technoweenie:
Here's a failing test against [19326a77397647088adc863144fb5428f3fb64f1](https://github.com/git-lfs/git-lfs/commits/19326a77397647088adc863144fb5428f3fb64f1):

https://ci.appveyor.com/project/git-lfs/git-lfs/build/1.0.643

```
    + out='Cloning into '\''000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000'\''...
    Downloading foo/bar/objects/0000000000000000000000000000000000000000000000000000000000000000.bin (14 B)
    Error downloading object: foo/bar/objects/0000000000000000000000000000000000000000000000000000000000000000.bin (7da4fb02e8b23ed26d05e4a955565a7b1ff77846d0bf0d9ce5651bf20ed2e5d3)
    
    Errors logged to C:\Users\appveyor\AppData\Local\Temp\1\git-lfs_TEMP.tMOi0d\test-long-paths.sh-3652\000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000\.git\lfs\objects\logs\20161122T205741.5768557.log
    error: external filter git-lfs smudge -- %f failed 2
    error: external filter git-lfs smudge -- %f failed
    fatal: foo/bar/objects/0000000000000000000000000000000000000000000000000000000000000000.bin: smudge filter lfs failed
    You can inspect what was checked out with '\''git status'\''
    and retry the checkout with '\''git checkout -f HEAD'\'''
```

Here's the passing test on this PR: https://ci.appveyor.com/project/technoweenie/git-lfs/build/1.0.193

To top it off, @DoDoENT confirmed my custom build worked in https://github.com/git-lfs/git-lfs/issues/1690#issuecomment-262347384.

knime:
I believe that the fix didn't cover all situations, at least we are having still problems:

```
$ git-lfs.exe pull knime-jenkins
Could not checkout file

Could not create working directory file: open org.knime.build.config/director/plugins/org.eclipse.equinox.launcher.cocoa.macosx.x86_64_1.1.300.v20150602-1417/eclipse_1611.so: The system cannot find the path specified.
```
After a quick code inspection I believe the problem is that the _fixLongPath_ function assumes it gets an absolute path but according to the error message above only a relative path seems passed to _PointerSmudgeToFile_ and therefore _longpathos.Create_. Could this be the problem?

PR #1710, opened 2016-11-22
Fast walk no channels

technoweenie:
This is an exploration of a new `FastWalkGitRepo` api that accepts a function callback, instead of returning channels. I've been thinking about the `gitscanner` public interface for a few weeks, and I keep coming back to the fact that we should not be exposing channels to end users (even if _we_ are the end users too). I picked `FastWalkGitRepo()` as a guinea pig, since it's much simpler. 

The `commands/command_track.go` changes look much better to me. The new version is focused on what the command needs, and not reading channels concurrently.

I wrote this as a parallel implementation so I could compare the benchmark side by side, and don't see a big difference:

```
$ script/test tools -v -run Bench -bench . -count 5
BenchmarkFastWalkGitRepoChannels-8   	       1	3059166517 ns/op
--- BENCH: BenchmarkFastWalkGitRepoChannels-8
	filetools_test.go:79: files: 20008, errors: 0
BenchmarkFastWalkGitRepoChannels-8   	       1	3052510253 ns/op
--- BENCH: BenchmarkFastWalkGitRepoChannels-8
	filetools_test.go:79: files: 20008, errors: 0
BenchmarkFastWalkGitRepoChannels-8   	       1	2763236760 ns/op
--- BENCH: BenchmarkFastWalkGitRepoChannels-8
	filetools_test.go:79: files: 20008, errors: 0
BenchmarkFastWalkGitRepoChannels-8   	       1	2656352784 ns/op
--- BENCH: BenchmarkFastWalkGitRepoChannels-8
	filetools_test.go:79: files: 20008, errors: 0
BenchmarkFastWalkGitRepoChannels-8   	       1	2676479922 ns/op
--- BENCH: BenchmarkFastWalkGitRepoChannels-8
	filetools_test.go:79: files: 20008, errors: 0
BenchmarkFastWalkGitRepoCallback-8   	       1	2679861166 ns/op
--- BENCH: BenchmarkFastWalkGitRepoCallback-8
	filetools_test.go:102: files: 20007, errors: 0
BenchmarkFastWalkGitRepoCallback-8   	       1	2744682179 ns/op
--- BENCH: BenchmarkFastWalkGitRepoCallback-8
	filetools_test.go:102: files: 20008, errors: 0
BenchmarkFastWalkGitRepoCallback-8   	       1	2691585917 ns/op
--- BENCH: BenchmarkFastWalkGitRepoCallback-8
	filetools_test.go:102: files: 20008, errors: 0
BenchmarkFastWalkGitRepoCallback-8   	       1	2903895731 ns/op
--- BENCH: BenchmarkFastWalkGitRepoCallback-8
	filetools_test.go:102: files: 20007, errors: 0
BenchmarkFastWalkGitRepoCallback-8   	       1	2701274878 ns/op
--- BENCH: BenchmarkFastWalkGitRepoCallback-8
	filetools_test.go:102: files: 20008, errors: 0
PASS
ok  	github.com/git-lfs/git-lfs/tools	27.942s
```

If you all agree this is a favorable change, I'll remove all the old code, and start on a change to the `gitscanner` like this:

```go
// pseudo code, I haven't exactly looked all these funcs up
tq := newBatchTransferQueue()

// non blocking!
scanner := gitscanner.New(func(p *lfs.WrappedPointer, err error) {
  // queue up to download this object
  tq.Add(p)
})
scanner.RemoteForPush("origin")

// not implemented yet, but would add file path inc/exc filtering to all applicable scans
scanner.Filter = filepathfilter.New(inc, exc)

for _, ref := range pushRefs {
    // blocking call until `rev-list` is done
    // then calls callback given to New()
    scanner.ScanLeftToRemote(ref)
}

// any internal channel clean up, make sure all callbacks have been called
scanner.Close()
```

### DIFF
--- a/tools/filetools.go
+++ b/tools/filetools.go
@@ -129,19 +129,19 @@ func VerifyFileHash(oid, path string) error {
 // Returned from FastWalk with parent directory context
 // This is needed because FastWalk can provide paths out of order so the
 // parent dir cannot be implied
-type FastWalkInfo struct {
+type fastWalkInfo struct {
 	ParentDir string
 	Info      os.FileInfo
 }
 
-type FastWalkCallback func(string, os.FileInfo, error)
+type FastWalkCallback func(parentDir string, info os.FileInfo, err error)
 
 // FastWalkGitRepo is a more optimal implementation of filepath.Walk for a Git repo
 // It differs in the following ways:
 //  * Provides a channel of information instead of using a callback func
 //  * Uses goroutines to parallelise large dirs and descent into subdirs
 //  * Does not provide sorted output; parents will always be before children but
-//    there are no other guarantees. Use parentDir in the FastWalkInfo struct to
+//    there are no other guarantees. Use parentDir in the fastWalkInfo struct to
 //    determine absolute path rather than tracking it yourself like filepath.Walk
 //  * Automatically ignores any .git directories
 //  * Respects .gitignore contents and skips ignored files/dirs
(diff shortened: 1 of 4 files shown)


tools/filetools.go
```diff
@@ -134,18 +134,7 @@ type FastWalkInfo struct {
...
-func fastWalkWithExcludeFiles(dir, excludeFilename string,
-	excludePaths []filepathfilter.Pattern) (<-chan FastWalkInfo, <-chan error) {
-	fiChan := make(chan FastWalkInfo, 256)
-	errChan := make(chan error, 10)
-
-	go fastWalkFromRoot(dir, excludeFilename, excludePaths, fiChan, errChan)
-
-	return fiChan, errChan
-}
+type FastWalkCallback func(string, os.FileInfo, error)
```
ttaylorr: I think assigning names to the arguments of this function would be helpful 👍

technoweenie:
> What do you want to do with the channel-based code?

I'm trying to get rid of it, but I'm getting intermittent failures. The nice thing about the channel code is that they serve as a mechanism for the reader to ensure they've gotten all the data. I just have to make sure that the function blocks until all the callbacks have been called.

```
--- FAIL: TestFastWalkBasic (0.16s)
        Error Trace:    filetools_test.go:61

...

			Diff:
			--- Expected
			+++ Actual
			@@ -1,2 +1,2 @@
			-([]string) (len=538 cap=1024) {
			+([]string) (len=532 cap=1000) {
			  (string) (len=8) "testroot",
			@@ -102,4 +102,2 @@
			  (string) (len=39) "testroot/folder2/subfolder1/file158.txt",
			- (string) (len=39) "testroot/folder2/subfolder1/file159.txt",
			- (string) (len=38) "testroot/folder2/subfolder1/file16.txt",
			  (string) (len=38) "testroot/folder2/subfolder1/file17.txt",
			@@ -107,3 +105,2 @@
			  (string) (len=38) "testroot/folder2/subfolder1/file19.txt",
			- (string) (len=37) "testroot/folder2/subfolder1/file2.txt",
			  (string) (len=38) "testroot/folder2/subfolder1/file20.txt",
			@@ -129,3 +126,2 @@
			  (string) (len=38) "testroot/folder2/subfolder1/file39.txt",
			- (string) (len=37) "testroot/folder2/subfolder1/file4.txt",
			  (string) (len=38) "testroot/folder2/subfolder1/file40.txt",
			@@ -186,3 +182,2 @@
			  (string) (len=38) "testroot/folder2/subfolder1/file90.txt",
			- (string) (len=38) "testroot/folder2/subfolder1/file91.txt",
			  (string) (len=38) "testroot/folder2/subfolder1/file92.txt",
			@@ -190,3 +185,2 @@
			  (string) (len=38) "testroot/folder2/subfolder1/file94.txt",
			- (string) (len=38) "testroot/folder2/subfolder1/file95.txt",
			  (string) (len=38) "testroot/folder2/subfolder1/file96.txt",
```

technoweenie:
I think we have to keep using channels internally, so that we can ensure the callback isn't called in parallel. I think there's a slight chance of this happening on larger directories, because `Readdir(100)` spins up goroutines for each chunk of 100 directory entries.

technoweenie:
I think I want the callback to have an optional `error` return value, which would halt the file walk early. That'll take a bit of work though :)

sinbad:
Yeah the main issue with the code right now is that the callback isn't serialised at all, so multiple goroutines can call it at once. You'd have to put the same kind of channel handling code inside a non-exposed function and have that call the callback instead. 

As for halting the process in the middle, that's tricky but doable. By nature there are N walkers at any one time to balance I/O blocking so you'd have to halt all of them. That typically means having a global sentinel value which will trigger all goroutines to stop iterating. If you can guarantee that its value will only ever be set once, in one direction (from false to true say) by a single goroutine (the one aggregating the channels), then you don't need to lock it, which means you avoid another bottleneck.

sinbad:
I suggest changing to a single channel of 'data or error' returns instead of 2 channels, this is what I'm looking at now in response to feedback on getting rid of channels on the locking APIs.

technoweenie:
> Yeah the main issue with the code right now is that the callback isn't serialised at all, so multiple goroutines can call it at once. You'd have to put the same kind of channel handling code inside a non-exposed function and have that call the callback instead.

I've updated `FastWalkGitRepo()` to do this.

> As for halting the process in the middle, that's tricky but doable.

I played around with a few approaches, but gave up because I don't think it's a big deal. Seems like errors can only come from reading directories or the exclude file, which would point to a bigger issue than what LFS can handle (and would probably halt the directory reading anyway).

> I suggest changing to a single channel of 'data or error' returns instead of 2 channels

I just did that in https://github.com/git-lfs/git-lfs/pull/1710/commits/3b9e629d7440c5297ec33666e875b3fd50dd72fe. I like that I only have to worry about reading from a single channel, which lets me get rid of a `sync.WaitGroup`. Is there some other reason this is preferable that I'm missing?

sinbad:
> I just did that in 3b9e629. I like that I only have to worry about reading from a single channel, which lets me get rid of a sync.WaitGroup. Is there some other reason this is preferable that I'm missing?

👍 This is fine, you get a simpler processing model, it just implies that you generally consume from a single goroutine. The advantage of separate results/error channels is that you can consume results from >1 channel and have a dedicated error collection routine to centralise error handling. It's more tedious to collect errors in a combined channel if you have multiple consumers of results. I think the trade off is fine in this case.

PR #1711, opened 2016-11-23
GitLFS requires Go 1.7+

larsxschneider:
Apparently the process-filter does not work with Go below version 1.7
cf. https://github.com/git-lfs/git-lfs/pull/1699#issuecomment-262303880


ttaylorr:
I think we should stay on the safe side and make the minimum required version Go 1.7.3, since we know it works in that version.

I also went ahead and updated the README to reflect this as well.

PR #1712, opened 2016-11-23
Remove the license description from the license itself

benbalter:
This PR removes the "This project is licensed under the MIT license" line from the license, and replaces it with just the license title (which itself, is optional).

This will allow GitHub to properly detect the license, because right now it doesn't know what that extra line of custom text means, or if it has any legal implications (and thus doesn't call the license MIT).

/cc @technoweenie

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,4 @@
-This project is licensed under the [MIT license](https://en.wikipedia.org/wiki/MIT_License). 
+MIT License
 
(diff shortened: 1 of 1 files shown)


benbalter:
@ttaylorr I did when I submitted the PR.

screen shot 2016-11-23 at 11 59 43 am

ttaylorr:
Huh, not sure why the CLA app isn't reporting that you did. I'm going to merge anyway 👍

PR #1723, opened 2016-11-28
Locking: move lock funcs to own package

sinbad:
Re-implementation of #1625. Based on PR feedback (except config changes which are still pending). Channel wrappers are gone & so is much of complexity, simpler for clients (and added local `Lock` struct so remove dependency on `api.Lock`). 

Merge target is `locking-master` to denote that this isn't a finished API yet but to enable reviewing in bite-size chunks.

### DIFF
--- a/commands/command_lock.go
+++ b/commands/command_lock.go
@@ -28,7 +28,7 @@ func lockCommand(cmd *cobra.Command, args []string) {
 		Exit(err.Error())
 	}
 
-	id, err := locking.Lock(path, lockRemote)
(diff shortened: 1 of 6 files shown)


sinbad:
> I'm approving this on the condition that the locking exports are changed as suggested (#1625 (comment)) in a future PR

Yeah, I think that's going to need a bigger refactor of other code's use of `Configuration` to do properly so wanted to come back to that. Using `locking-master` as an intermediate step will keep this sane for everyone I hope.

PR #1725, opened 2016-11-29
Gitscanner callbacks

technoweenie:
This changes the `gitscanner` API to use callbacks, instead of returning a `PointerChannelWrapper`. See [the fsck command](https://github.com/git-lfs/git-lfs/compare/fast-walk-no-channels...gitscanner-callbacks?expand=1#diff-3b7fa953c43592e4af73f324813e6c2fR35) for an _ideal_ usage of it:

```go
gitscanner := lfs.NewGitScanner(func(p *lfs.WrappedPointer, err error) {
  // handle pointers or scan errors
})

if err := gitscanner.ScanRefWithDeleted(ref.Sha, nil); err != nil {
  // handle cmd error
}

gitscanner.Close()
```

You can now pass a callback to `NewGitScanner()` for all the `.Scan*()` funcs, or pass a callback to individual functions if you want those pointers to be handled differently.

Unfortunately, this introduces some ugly code in the core push/pull/checkout related commands. A lot of code still expects to receive a `[]*lfs.WrappedPointer` after the scan has completed. This cleanup will have to happen in future PRs. The main blockers are the progress meter and transfer queues need to accept a stream of pointers.

My hunch is that the transfer queue is all ready setup for this, but our _use_ of it prefers receiving a `[]*lfs.WrappedPointer`. However, the progress meter definitely expects to receive the total number and size of the pointers up front. Since this process is becoming async, updating the progress meter stuff is my next target.

Also, I was a doofus and did most of this work on top of #1710, so I have to wait for that to be merged before merging this :)

* [ ] merge #1710
* [ ] reset base to `master`

### DIFF
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -107,11 +107,27 @@ func fetchCommand(cmd *cobra.Command, args []string) {
 }
 
 func pointersToFetchForRef(gitscanner *lfs.GitScanner, ref string) ([]*lfs.WrappedPointer, error) {
-	pointerCh, err := gitscanner.ScanTree(ref)
-	if err != nil {
+	var pointers []*lfs.WrappedPointer
+	var multiErr error
+	tempgitscanner := lfs.NewGitScanner(func(p *lfs.WrappedPointer, err error) {
+		if err != nil {
+			if multiErr != nil {
+				multiErr = fmt.Errorf("%v\n%v", multiErr, err)
+			} else {
+				multiErr = err
+			}
+			return
+		}
+
+		pointers = append(pointers, p)
+	})
+
+	if err := tempgitscanner.ScanTree(ref, nil); err != nil {
 		return nil, err
 	}
-	return collectPointers(pointerCh)
+
+	tempgitscanner.Close()
+	return pointers, multiErr
 }
 
 func fetchRefToChan(gitscanner *lfs.GitScanner, ref string, filter *filepathfilter.Filter) chan *lfs.WrappedPointer {
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -241,21 +241,32 @@ func scanAll(gitscanner *lfs.GitScanner) []*lfs.WrappedPointer {
 	spinner := progress.NewSpinner()
 	var numObjs int64
 
-	pointerCh, err := gitscanner.ScanAll()
-	if err != nil {
-		Panic(err, "Could not scan for Git LFS files")
-	}
+	// use temp gitscanner to collect pointers
+	var pointers []*lfs.WrappedPointer
+	var multiErr error
+	tempgitscanner := lfs.NewGitScanner(func(p *lfs.WrappedPointer, err error) {
+		if err != nil {
+			if multiErr != nil {
+				multiErr = fmt.Errorf("%v\n%v", multiErr, err)
+			} else {
+				multiErr = err
+			}
+			return
+		}
 
-	pointers := make([]*lfs.WrappedPointer, 0)
-
-	for p := range pointerCh.Results {
 		numObjs++
 		spinner.Print(OutputWriter, fmt.Sprintf("%d objects found", numObjs))
 		pointers = append(pointers, p)
-	}
-	err = pointerCh.Wait()
-	if err != nil {
+	})
+
+	if err := tempgitscanner.ScanAll(nil); err != nil {
 		Panic(err, "Could not scan for Git LFS files")
+	}
+
+	tempgitscanner.Close()
+
+	if multiErr != nil {
+		Panic(multiErr, "Could not scan for Git LFS files")
 	}
 
 	spinner.Finish(OutputWriter, fmt.Sprintf("%d objects found", numObjs))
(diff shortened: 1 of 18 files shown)


commands/command_fsck.go
```diff
@@ -17,110 +17,99 @@ var (
...
+	_, err = io.Copy(oidHash, f)
+	f.Close()
 	if err != nil {
-		Panic(err, "Error checking Git LFS files")
+		return false, err
 	}
 
-	if ok {
-		Print("Git LFS fsck OK")
+	recalculatedOid := hex.EncodeToString(oidHash.Sum(nil))
```
ttaylorr: Oh man, this is so much better than

```go
fmt.Sprintf("%x", oidHash.Sum(nil))
```

commands/command_pre_push.go
```diff
@@ -77,14 +78,37 @@ func prePushCommand(cmd *cobra.Command, args []string) {
...
+			} else {
+				multiErr = err
+			}
+			return
+		}
+
+		pointers = append(pointers, p)
+	}
+
+	if err := g.ScanLeftToRemote(ref, cb); err != nil {
```
ttaylorr: Are we OK to do this if `multiErr != nil`?

lfs/gitscanner.go
```diff
@@ -134,6 +174,17 @@ func (s *GitScanner) opts(mode ScanningMode) *ScanRefsOptions {
...
 
+func firstGitScannerCallback(callbacks ...GitScannerCallback) (GitScannerCallback, error) {
+	for _, cb := range callbacks {
+		if cb == nil {
+			continue
+		}
+		return cb, nil
+	}
+
+	return nil, errors.New("No callback given")
```
ttaylorr: We talked about this in 🍖 -space, but this would be a good candidate to extract to a sentinel value if you want to keep the `firstGitScannerCallback` business 👍

commands/command_fetch.go
```diff
@@ -107,11 +107,27 @@ func fetchCommand(cmd *cobra.Command, args []string) {
...
 
 func pointersToFetchForRef(gitscanner *lfs.GitScanner, ref string) ([]*lfs.WrappedPointer, error) {
-	pointerCh, err := gitscanner.ScanTree(ref)
-	if err != nil {
+	var pointers []*lfs.WrappedPointer
+	var multiErr error
+	tempgitscanner := lfs.NewGitScanner(func(p *lfs.WrappedPointer, err error) {
+		if err != nil {
+			if multiErr != nil {
+				multiErr = fmt.Errorf("%v\n%v", multiErr, err)
```
ttaylorr: Coming back to this, if `CombineErrors` were to go in the `errors` package, I think it'd be better to just call it `errors.Combine`/

commands/command_fsck.go
```diff
@@ -17,110 +17,99 @@ var (
...
+	}
 
-		oidHash := sha256.New()
-		_, err = io.Copy(oidHash, f)
-		f.Close()
-		if err != nil {
-			return false, err
+	for _, oid := range corruptOids {
+		badFile := filepath.Join(badDir, oid)
+		if err := os.Rename(lfs.LocalMediaPathReadOnly(oid), badFile); err != nil {
```
technoweenie: How else do you rename?

commands/command_fetch.go
technoweenie: Yeah, we could probably expose the inner errors somehow too. I'll experiment with this in another PR.

commands/command_pre_push.go
technoweenie: Yes. If this returns an error, the command that `ScanLeftToRemote()` attempts to spin up errored, so the callback potentially sets `multiErr` is never called.

ttaylorr: Awesome, thanks for the explanation 👍

lfs/gitscanner.go
```diff
@@ -10,6 +10,14 @@ import (
 	"github.com/rubyist/tracerx"
 )
 
+var missingCallbackErr = errors.New("No callback given")
+
+// CallbackMissing returns a boolean indicating whether the error is reporting
+// that a GitScanner is missing a required GitScannerCallback.
+func CallbackMissing(err error) bool {
```
technoweenie: It was a 50/50 toss-up for me, but your recommendation made it a 51/49 split :)

commands/command_fetch.go
sinbad: I added `errors.Combine` in my original locking PR so 👍 (although it only concatenated with a line break as you're doing here, nothing smart like storing inner errors)

sinbad:
It feels a little odd to me to supply the callback in `NewGitScanner` instead of per `Scan...` function. Are you really likely to want to use the same callback for more than one invocation of `Scan...`? Feels to me like you'd mostly want different callbacks each time, and the callback implementation flowing after the `Scan..` call feels more natural and readable to having to look back at the construction.

technoweenie:
Part of the point in the [git scanner redesign](https://github.com/git-lfs/git-lfs/issues/1649) is spawning the `git cat-file` processes a single time for large pushes, reducing how many `git` commands are spawned overall. Most uses of the git scanner can be changed to pass the pointers asynchronously to the transfer queue, but a lot of [cruft](https://github.com/git-lfs/git-lfs/pull/1725/files#diff-d540c48abbcd5b13b0acd4cbef433700L80) prevents us from doing it in this PR. I plan to tackle the progress meter next, while @ttaylorr works on the transfer queue.

That said, I do think passing the callback on the constructor is pretty strange from an API perspective. There's still time before the API is finalized a `git/gitscanner` package, so I expect the callback situation will change.

Issue #1730, opened 2016-12-01
Trouble with equal (=) signs in filename


smorin-primero:
I tried to clone a repository with a lfs file where the filename contains an equal sign, and the clean filter failed:

```
git-lfs/1.5.2 (GitHub; windows amd64; go 1.7.3; git 547a06aa)
git version 2.11.0.windows.1

$ git-lfs clone ...
Error updating the git index:
Unknown command ""
error: packet write failed
error: external filter 'git-lfs filter-process' failed
fatal: Test=.png: clean filter 'lfs' failed
```

Also when we try to add a file with an equal sign, we get a similar error:

```
$ GIT_TRACE=1 git add Test\=.png
09:22:43.235434 git.c:371               trace: built-in: git 'add' 'Test=.png'
09:22:43.240449 run-command.c:350       trace: run_command: 'git-lfs filter-process'
trace git-lfs: run_command: 'git' version
trace git-lfs: run_command: 'git' config -l
trace git-lfs: Initialize filter-process
trace git-lfs: Read filter-process request.
trace git-lfs: Read filter-process request.
09:22:44.165040 run-command.c:42        trace: run_command: running exit handler for pid 8720
Unknown command ""
```

larsxschneider:
Thanks for the report! Unfortunately I cannot reproduce the error on my windows machine.
Can you post the output of the following command?

```
GIT_TRACE=1 git clone --branch equal-bug https://github.com/larsxschneider/lfstest-individual-files.git smudge-error
``

larsxschneider:
Plus, do you use the Git bash or Cygwin?

smorin-primero:
I use Git bash, and I can't reproduce using your example. I will try to create a little example where it works.

I know that we have the problem on a couple of our repositories, so it's not isolated. I doubt it's the servers fault, but we use the self-hosted Bitbucket server.

ttaylorr:
@smorin-primero thanks for opening this. My initial thought is that the `=` character in the filename is unrelated to the issue at hand, and it is rather related to something in the file's contents. The ASCII hexcode for `=` is `0x3d` which has no special bearing in the LFS or the `pkt-line` protocol.

Are you able to share the contents of a repo that is able to reproduce the problem? That would be very helpful in debugging whatever issue is going on here 👍.

smorin-primero:
Ok, so it looks like it's more complicated. I was able to reproduce the problem using the image in attachment, only when there is an equal sign in the filename.

[...]

PR #1731, opened 2016-12-02
git: parse filter process header values containing '=' properly

larsxschneider:
A filter process key-value pair must be split at the first '=' character
into into at most two substrings. The value *can* contain a '='
character. The key *must never* contain a '=' character.

Should fix https://github.com/git-lfs/git-lfs/issues/1730


technoweenie:
Yes, with #1727.

PR #1733, opened 2016-12-02
Backport #1731 for v1.5.x: git: parse filter process header values containing '=' properly

ttaylorr:
This backports #1731.


eteka83:
Leave a comment it is important to always note on how to cope with two related systems. 
especially should these kind of concepts consist of subtle relationships among them . 
for instance it is good to note that sound and light are related to each other  on the other hand , light is electromagnetic while sound is pneumatic. 
these two quantities are represented in two abstract positions

PR #1734, opened 2016-12-02
build: use Go 1.7.4

ttaylorr:
This pull-request uses Go 1.7.4 on all of our build services. Go 1.7.4 matches what we ship with (after bumping the build dockers) and contains security fixes over 1.7.3.

- **TravisCI**: Bumped version.
- **CircleCI**: Uses latest version from Homebrew, already upgraded.
- **AppVeyor**: Doesn't ship their workers with 1.7.4 yet, working on installing it ourselves.

---

/cc @technoweenie

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 # http://docs.travis-ci.com/user/languages/go/
 language: go
 
-go: 1.7.1
+go: 1.7.4
 
 os:
   - linux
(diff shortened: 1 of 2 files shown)


ttaylorr:
Going to go ahead and backport this into the `release-1.5` tree before we release, since the version bump contains security fixes.

PR #1742, opened 2016-12-06
tq: implement `*tq.WorkerQueue` type

ttaylorr:
This pull-request implements the `*tq.WorkerQueue` type, responsible for distributing work among an available set of workers.

## What

Functionally, this type is very similar to the existing [`*transfer.adapterBase`][1] type, but I chose to re-implement in this new package for a few reasons:

1. **Model of execution**. This implementation has slightly different conditions for deciding when workers are available, and when new work can be assigned to the queue. Since this is an "all or nothing" scenario, where these execution changes have to be implemented atomically, I figured it'd be easier to start with something similar and port over functionality as necessary.
2. **New package**. This is going in a new package, `tq`, so at the very least I'd be duplicating the `transfer.adapterBase` type as `tq.adapterBase` and making changes there. I figure there won't be a huge difference in time spent whether I port over functionality up front or as needed. My immediate next step is going to be to make the existing interface and functionality from the `transfer` package work in the new `tq` package, so we should know pretty soon whether or not there are major flaws.

## How

This `*tq.WorkerQueue` has design invariants that make it easy to satisfy our goals for new work done in the `*tq.TransferQueue`. One of those goals is prioritizing retries to happen immediately to avoid a large amount of buffering, and in order for that to happen, we need to know about retries as soon as possible. 

This queue reports "failed" items that need to be retried through the retry channel that it returns in response to an `Add()` call:

```go
func (q *WorkerQueue) Add(batch []string) (retries <-chan string) { ... }
```

Callers of this method (namely, the `*tq.TransferQueue`) can read items from this channel until it closes and prioritize those items in the next batch of items. In fact, this is exactly what the new transfer queue will do, in a pull-request that I'll push shortly.

One downside of this method is that we have to wait for work to _completely_ finish in a batch before accumulating the next one. Right now, I'm not convinced that this is yet a showstopper, but just something that we need to keep our 👀  on. Implementing this new transfer queue in a new package will allow us to have clear boundaries of functional types that we can change in isolation should this behavior become prohibitive to operation. 

## Misc

Following @sinbad's lead, I've based this PR off of `tq-master`, the de-facto "master" branch for the transfer queue refactorings. This means I can avoid the nasty PR chain that was the filter-process project, while still making these fairly short/easy to review.

Once the project is done, we'll fast-forward master to include all the changes from `tq-master`.

---

/cc @git-lfs/core  

[1]: https://github.com/git-lfs/git-lfs/blob/7ffac54d7662d0ca3f919ec68d481d604970cea7/transfer/adapterbase.go

### DIFF
--- a/tq/worker_queue.go
+++ b/tq/worker_queue.go
@@ -1,0 +1,116 @@
+package tq
+
+import "sync"
+
+// WorkerFn is a function type, held by a `*WorkerQueue`, which is called
+// anytime there is work to be done. The parameter of the function "oid" is
+// given as the OID of the object to transfer. The return value dictates whether
+// or not the object was transferred successfully, true if so, false if
+// otherwise.
+type WorkerFn func(oid string) bool
+
+type WorkerQueue struct {
+	// tasks holds `*task`s to be worked upon by `WorkerFn`s. It is written
+	// to from a single source (see: `Add(batch []string)` below), and
+	// consumed from many workers.
+	tasks chan *task
+	// fn is the WorkerFn that executes when an OID is available to be
+	// processed. If the function returns true, then the OID is marked as
+	// successfully processed, and no retries are enqueued. Otherwise, the
+	// items is marked as having failed, and a retry is enqueued by writing
+	// the OID to the returned `retries` channel.
+	fn WorkerFn
+
+	// wg waits for all items given in an `Add()` call to be processed
+	// _completely_ by a worker. After latching, the given retries channel
+	// will be closed.
+	wg *sync.WaitGroup
+	// wwg waits for all workers to terminate after the `tasks` channel
+	// closes, and is a terminating condition of the Wait() function below.
+	wwg *sync.WaitGroup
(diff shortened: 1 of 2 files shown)


tq/worker_queue.go
```diff
@@ -0,0 +1,116 @@
...
+
+	go func(wg *sync.WaitGroup, retries chan<- string) {
+		defer close(retries)
+
+		for _, oid := range batch {
+			q.tasks <- &task{oid, retries, wg}
+		}
+
+		wg.Wait()
+	}(q.wg, retries)
```
technoweenie: I'm not sure this needs the arguments, since this . If so, wouldn't you have to pass `q.tasks` and `batch` too?

ttaylorr: Good call. Got rid of that in https://github.com/git-lfs/git-lfs/pull/1742/commits/8a063172539a9aa93398f9c65fd13f0dfe222b81.

PR #1744, opened 2016-12-06
tq: implement `*tq.Queue` type

ttaylorr:
~~⚠️  this PR is based against the [`tq-worker-queue`][1] tree to make the diff easier to view. Once [that PR][2] is merged into `tq-master`, I'll update the base of this to point at `tq-master` as well.~~

---

This pull-request implements a type similar to `*lfs.TransferQueue` as `*tq.Queue`. This type serves as a conduit between a stream of OIDs (i.e., the `*lfs.gitscanner`) and a pool of available workers (represented by the `*tq.WorkerQueue`). 

Though the public API between `*tq.Queue` and `*tq.WorkerQueue` is similar (both have `Add(oid string)` and `Wait()` funcs), their underlying purpose is different. While the `*tq.WorkerQueue` is responsible for distributing work and reporting failures, the `*tq.Queue` is responsible for batching units of work and including retries back into that batch _before_ new items.

Couple of neat things going on:

1. **Buffered channels**: One thing we can do to make sure batches are more readily available is wait until we have one _extra_ batch's worth of data buffered before applying back-pressure to the caller. This is done through channels alone by [buffering the incoming channel][3] to the desired size of a batch so that 
2. **Priority retries**: By waiting for a batch to complete, we can process retried items into the batch first before accepting new writes. That means that even if _all_ of the items fail in a batch, we can just skip reading new items, and immediately kick off that batch in the next iteration of the loop, since it'll already be full by that time.

Couple of things to consider:

1. As noted in #1742, we _have_ to wait for one batch to completely finish processing before sending off items for the next batch into the worker pool. I don't think this will be a huge deal, unless we have batches of _drastically_ different sizes (i.e., 99 objects <5MB, and a single 2GB object). I think there are a few things we can do to handle this better:
   - One thing we can do is kick off multiple batches and check retries before we send the batch off, pushing back new items in favor of retried ones.
   - Another thing we could do is sort the batches by size of item, prioritizing larger items first, so we have less of a chance of having to wait on them at the end. This could be interesting, and would also be fairly easy to implement.
2. Should we address [this][4] comment and do something meaningful if a caller tries to call `Add()` after `Wait()`-ing? Currently, the function will panic after trying to send on the now-closed `q.incoming` channel. I figure we could store the state in an `atomic`-managed `uint32`, but I'm not sure about what we can do when we _are_ closing and will reject the write other than log it and drop the OID on the floor. That may be good enough, but I'd love to hear your thoughts on this.

---

/cc @git-lfs/core 

[1]: https://github.com/git-lfs/git-lfs/tree/tq-worker-queue
[2]: https://github.com/git-lfs/git-lfs/pull/1742
[3]: https://github.com/git-lfs/git-lfs/pull/1744/commits/24fc67f8554573d74cbadbe594b71965234c1765#diff-aa8152c7da8b9e591de386d91ba55952R32
[4]: https://github.com/git-lfs/git-lfs/pull/1744/commits/24fc67f8554573d74cbadbe594b71965234c1765#diff-aa8152c7da8b9e591de386d91ba55952R97

### DIFF
--- a/tq/queue.go
+++ b/tq/queue.go
@@ -1,0 +1,115 @@
+package tq
+
+import "sync"
+
+// Queue organizes and distributes work on OIDs against a set of available
+// workers (represented by a *workerQueue). Any failed items returned back by
+// that `*workerQueue` will be prioritized into the next batch.
+type Queue struct {
+	// incoming serves as the barrier between new calls to `Add()` and
+	// accumulating the next batch of items to be sent into the worker
+	// queue.
+	//
+	// incoming has a finite non-zero buffer equal to the size of a batch
+	// such that one extra batch of data can be buffered before applying
+	// back-pressure to callers.
+	incoming chan string
+	// workers maintains a handle on the *workerQueue instance used to
+	// distribute batched sets of OIDs
+	workers *workerQueue
+
+	// wg holds onto the run() goroutine and latches only after the
+	// goroutine has terminated, enabling safe shutdowns.
+	wg *sync.WaitGroup
+}
+
+// New instantiates a new `*tq.Queue` type with a given batch size "size",
+// number of workers "workers", and a function to do that work, "fn".
+//
+// Once returned, the `*Queue` will be active and able to accept new writes.
+func New(size, workers int, fn WorkerFn) *Queue {
+	q := &Queue{
+		incoming: make(chan string, size),
+		workers:  newWorkerQueue(workers, fn),
+
+		wg: new(sync.WaitGroup),
+	}
+
+	q.wg.Add(1)
+	go q.run()
+
+	return q
+}
+
+// run accumulates new items into a batch for sending to the pool of available
+// workers, prioritizing retried objects over new objects.
+//
+// The function executes as follows:
+//
+//   1. Allocate a new batch of items for sending to the worker pool.
+//   2. Accept as many writes as we can before overfilling the batch from the
+//      `q.incoming` channel. If the channel closes, mark that we are closing
+//      and continue.
+//   3. Send the batch to the worker pool
+//   4. Empty the batch, begin re-filling it with items from the retry channel
+//      until the channel is closed.
+//   5. If closing, and the batch is empty, return.
+//   6. Otherwise, go to step 2.
+//
+// run runs in its own goroutine.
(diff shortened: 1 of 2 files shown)


tq/queue.go
```diff
@@ -0,0 +1,115 @@
...
+//   5. If closing, and the batch is empty, return.
+//   6. Otherwise, go to step 2.
+//
+// run runs in its own goroutine.
+func (q *Queue) run() {
+	defer q.wg.Done()
+
+	var closing bool
+
+	batch := make([]string, 0, cap(q.incoming))
```
technoweenie: I think this should just hardcode 100. Pretty sure that's a hardcoded limit for batches. The batch size should not be determined by the size of the `q.incoming` buffer, since we may want to increase the buffer to estimate more objects in the mutable progress meter from #1747.

tq/queue.go
```diff
@@ -0,0 +1,115 @@
...
+// Add()-ed already, this function will block.
+//
+// Add cannot be called after the `*Queue` has been marked as `Wait()`-ing.
+func (q *Queue) Add(oid string) {
+	// TODO(taylor): potentially store whether or not we're closing and
+	// check that before writing to a closed channel? Not sure what to do
+	// with that "oid" though... drop it? Return an error? Panic? All seem
+	// uniquely bad...
+
+	q.incoming <- oid
```
technoweenie: I'm cool with letting the channel write panic here. Calling `Add()` after `Wait()` is something easily preventable. Should be noted in the `Queue` godocs.

ttaylorr: Cool, I'm on board with that idea. Added some docs in https://github.com/git-lfs/git-lfs/pull/1744/commits/5983939c429b662df622c57c90986e4d36328ff0.

tq/queue.go
ttaylorr: Made the batch size and queue depth flexible with option types in https://github.com/git-lfs/git-lfs/pull/1744/commits/e1e746f063660e7d6a7bac6c34472fe95ee76a80. If a caller neglects to give the `tq.WithBatchSize` option, it'll just default to the `tq.DefaultBatchSize` const.

ttaylorr:
Superseded by #1758.

PR #1745, opened 2016-12-06
Teach gitscanner how to filter in ScanTree()

technoweenie:
This removes the inline filepathfilter checking from the `commands`, relying on the gitscanner to do it.

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -38,7 +38,7 @@ func checkoutCommand(cmd *cobra.Command, args []string) {
 	checkoutWithIncludeExclude(filter)
 }
 
-func checkoutFromFetchChan(in chan *lfs.WrappedPointer) {
+func checkoutFromFetchChan(in chan *lfs.WrappedPointer, filter *filepathfilter.Filter) {
 	ref, err := git.CurrentRef()
 	if err != nil {
 		Panic(err, "Could not checkout")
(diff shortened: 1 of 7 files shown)


commands/command_checkout.go
```diff
@@ -35,12 +35,10 @@ func checkoutCommand(cmd *cobra.Command, args []string) {
...
 
 	filter := filepathfilter.New(rootedpaths, nil)
-	gitscanner := lfs.NewGitScanner(nil)
-	defer gitscanner.Close()
-	checkoutWithIncludeExclude(gitscanner, filter)
+	checkoutWithIncludeExclude(filter)
 }
 
-func checkoutFromFetchChan(gitscanner *lfs.GitScanner, filter *filepathfilter.Filter, in chan *lfs.WrappedPointer) {
+func checkoutFromFetchChan(in chan *lfs.WrappedPointer) {
```
ttaylorr: I think this function still needs a `*filepathfilter.Filter`. Additionally, we should probably put a safety net around this and guard it with an integration test.

technoweenie:
Added a test that guards the original behavior: https://github.com/git-lfs/git-lfs/pull/1745/commits/bce58fc6d9ca57e847bf17fb01b3681a86cf4a56

I really want to refactor `pull()`, but there are still more changes coming to the progress meter and transfer queue so it doesn't require a pointer slice upfront.

PR #1746, opened 2016-12-07
Introduce transfer queue options

technoweenie:
This replaces the argument list in the transfer queue constructors with a variadic option func arg. Both `meter` and `dryRun` are technically optional. Also, the `*TransferQueue` has that `run()` that's required.

This was mostly an experiment. I'm now thinking that `newTransferQueue(dir transfer.Direction)` is all we need if `tq.Run()` is called after the first `tq.Add(...)`. Then you wouldn't need optional args in any fashion:

```go
tq := newTransferQueue(transfer.Download)
// defaults to false
tq.DryRun = true
// defaults to progress.Noop()
tq.Meter = progress.NewMeter("...")
```

### DIFF
--- a/lfs/transfer_queue.go
+++ b/lfs/transfer_queue.go
@@ -115,22 +115,38 @@ type TransferQueue struct {
 	rc       *retryCounter
 }
 
-// newTransferQueue builds a TransferQueue, direction and underlying mechanism determined by adapter
-func newTransferQueue(dir transfer.Direction, meter progress.Meter, dryRun bool) *TransferQueue {
-	if meter == nil {
-		meter = progress.Noop()
-	}
+type TransferQueueOption func(*TransferQueue)
 
+func DryRun(dryRun bool) TransferQueueOption {
+	return func(tq *TransferQueue) {
+		tq.dryRun = dryRun
+	}
+}
+
+func WithProgress(m progress.Meter) TransferQueueOption {
+	return func(tq *TransferQueue) {
+		tq.meter = m
+	}
+}
+
+// newTransferQueue builds a TransferQueue, direction and underlying mechanism determined by adapter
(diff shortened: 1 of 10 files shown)


lfs/transfer_queue.go
```diff
@@ -133,6 +141,14 @@ func newTransferQueue(dir transfer.Direction, meter progress.Meter, dryRun bool)
 		rc:            newRetryCounter(config.Config),
 	}
 
+	for _, opt := range options {
+		opt(q)
+	}
+
+	if q.meter == nil {
+		q.meter = progress.Noop()
```
ttaylorr: Nit: assign this in the struct initialization. That'll remove this conditional.

lfs/transfer_queue.go
```diff
@@ -115,16 +115,24 @@ type TransferQueue struct {
...
+}
+
+func WithProgress(m progress.Meter) TransferQueueOption {
+	return func(tq *TransferQueue) {
+		tq.meter = m
 	}
+}
 
+// newTransferQueue builds a TransferQueue, direction and underlying mechanism determined by adapter
+func newTransferQueue(dir transfer.Direction, options ...TransferQueueOption) *TransferQueue {
```
ttaylorr: 😻 I love the `options ...TransferQueueOption` pattern.

lfs/transfer_queue.go
technoweenie: It'll remove the conditional, but it'll also create a dummy struct that we never need when the `lfs.WithProgress()` option is applied.

technoweenie:
I applied the options pattern to `progress.NewMeter()`. It's looking good, so I'll go ahead and merge this into #1747. I'm thinking about unexporting the option types `TransferQueueOptions` and `MeterOptions`, since I don't think they need to be exported. Also, there are likely other small changes (such as docs on public properties) before #1747 will be ready to merge.

PR #1747, opened 2016-12-07
Mutable progress meter

technoweenie:
This adds `(*progress.ProgressMeter) AddEstimate(size int64)`, which lets us build up the estimated files and total bytes to transfer a file at a time. This is the first step towards relieving us of the requirement that we need to gather all the pointers in a `[]*lfs.WrappedPointer` before any transfers can _start_.

This also changes the transfer queue constructors: `lfs.NewUploadQueue()`, `lfs.NewDownloadQueue()`, and `lfs.NewDownloadCheckQueue()` to accept the minimum args: a `*progress.ProgressMeter` and `dryrun bool`. I didn't like these args, so I tried #1746 as an experiment.


progress/meter.go
```diff
@@ -48,7 +52,7 @@ func NewProgressMeter(estFiles int, estBytes int64, dryRun bool, logPath string)
 		finished:       make(chan interface{}),
 		estimatedFiles: int32(estFiles),
 		estimatedBytes: estBytes,
-		dryRun:         dryRun,
+		DryRun:         dryRun,
```
technoweenie: Going to un-export this, move `Start()` to this constructor, and then add a `MeterOption` func to set `dryRun`.

commands/command_checkout.go
```diff
@@ -127,29 +127,22 @@ func checkoutWithIncludeExclude(filter *filepathfilter.Filter) {
...
-	progress := progress.NewProgressMeter(len(pointers), totalBytes, false, logPath)
-	progress.Start()
-	totalBytes = 0
+	meter := progress.NewMeter(logPath)
+	meter.Start()
+	var totalBytes int64
 	for _, pointer := range pointers {
 		totalBytes += pointer.Size
-		progress.Add(pointer.Name)
+		meter.StartTransfer(pointer.Name)
```
technoweenie: Need to add a `meter.AddEstimate()` here.

technoweenie:
Welp, we can't move `Start()` to the constructor, because it starts writing to the log immediately.  Everything else I said I'd do has been done. Review away!

progress/meter.go
```diff
@@ -32,24 +33,70 @@ type ProgressMeter struct {
...
 
-	return &ProgressMeter{
-		logger:         logger,
+	return func(m *ProgressMeter) {
+		if len(name) == 0 {
+			return
+		}
+
+		if !filepath.IsAbs(name) {
+			printErr("GIT_LFS_PROGRESS must be an absolute path")
```
ttaylorr: Not sure I dig using the `printErr` function like this. What do you think about:

```go
return func(m *ProgressMeter) {
        var err error
        defer func() {
                if err == nil {
                        return
                }

                err = errors.Wrap("Error creating progress logger", err)
                fmt.Fprintf(os.Stderr, "%s", err)
        }()

        if len(name) == 0 {
                return
        }

        if !filepath.IsAbs(name) {
                err = errors.New("GIT_LFS_PROGRESS must be an absolute path")
                return
        }
}
```

test/git-lfs-test-server-api/main.go
```diff
@@ -147,17 +149,16 @@ func buildTestData() (oidsExist, oidsMissing []TestObject, err error) {
 	defer repo.Cleanup()
 	// just one commit
 	commit := test.CommitInput{CommitterName: "A N Other", CommitterEmail: "noone@somewhere.com"}
-	var totalSize int64
 	for i := 0; i < oidCount; i++ {
 		filename := fmt.Sprintf("file%d.dat", i)
 		sz := int64(rand.Intn(200)) + 50
 		commit.Files = append(commit.Files, &test.FileInput{Filename: filename, Size: sz})
-		totalSize += sz
+		meter.AddEstimate(sz)
```
ttaylorr: What do you think about leaving the `totalSize` variable and adding to that inside the loop, then calling the `AddEstimate` function once? If `oidCount` is large, I figure that the overhead of calling `atomic.Add<T>` functions will add up.

progress/meter.go
```diff
@@ -58,15 +105,6 @@ func (p *ProgressMeter) Start() {
...
-func (p *ProgressMeter) Add(name string) {
-	idx := atomic.AddInt64(&p.transferringFiles, 1)
-	p.fileIndexMutex.Lock()
-	p.fileIndex[name] = idx
-	p.fileIndexMutex.Unlock()
-}
-
 // Skip tells the progress meter that a file of size `size` is being skipped
 // because the transfer is unnecessary.
 func (p *ProgressMeter) Skip(size int64) {
```
ttaylorr: Per our 🍖 -space discussion, I definitely agree that it'd be nice for this function to have a name that agrees with the `AddEstimate` function.

What do you think about `DecrementEstimate` or `DecreaseEstimate`?

progress/meter.go
technoweenie: I think my version is more straight forward than relying on a `defer` func to check a local var.

progress/meter.go
```diff
@@ -32,24 +33,70 @@ type ProgressMeter struct {
 	dryRun            bool
 }
 
-// NewProgressMeter creates a new ProgressMeter for the number and size of
-// files given.
-func NewProgressMeter(estFiles int, estBytes int64, dryRun bool, logPath string) *ProgressMeter {
-	logger, err := newProgressLogger(logPath)
-	if err != nil {
+type env interface {
```
technoweenie: It's not needed, but it saves this package from having to import `config` :)

progress/meter.go
```diff
@@ -32,24 +33,70 @@ type ProgressMeter struct {
...
+			printErr(err.Error())
+			return
+		}
+
+		m.logger.writeData = true
+		m.logger.log = file
+	}
+}
+
+func WithOSEnv(os env) meterOption {
```
technoweenie: They have the same interface, but git and os envs are very different. Just made sense to make it more explicit in the method name. I was envisioning a type that accepted OS and git env values.

progress/meter.go
technoweenie: Well, we were actually talking about the `Add() -> StartTransfer()` rename so it's consistent with `FinishTransfer()`. However, this is a valid point. I don't like any of the method names with `Estimate` TBH, and was hoping to come up with a better name. Maybe just `Add()` and `Skip()`, even if this new `Add()` is very different from the old one (that is now `StartTransfer()`)?

test/git-lfs-test-server-api/main.go
technoweenie: I wrote https://gist.github.com/technoweenie/f5ba7b696fb7eded1473d336e99a80d3 to benchmark atomic.  Clearly a channel is not a good way to optimize `atomic.AddInt64()` calls:

```
~p/atomic-bench $ go test . -bench . -v
testing: warning: no tests to run
BenchmarkChannels-8   	2000000000	         0.18 ns/op
BenchmarkAtomic-8     	2000000000	         0.03 ns/op
PASS
ok  	_/Users/rick/p/atomic-bench	6.014s

~p/atomic-bench $ go test . -bench BenchmarkChannels -v
testing: warning: no tests to run
BenchmarkChannels-8   	2000000000	         0.18 ns/op
PASS
ok  	_/Users/rick/p/atomic-bench	5.436s

~p/atomic-bench $ go test . -bench BenchmarkAtomic -v
testing: warning: no tests to run
BenchmarkAtomic-8   	2000000000	         0.04 ns/op
PASS
ok  	_/Users/rick/p/atomic-bench	0.617s
```

Happy to try other approaches if anyone think it'd be worth it, but `sync/atomic` is fast enough for me :)

Issue #1726, opened 2016-11-30
lfs tracked files are sometimes incorrectly shown as modified in "git status" output


AndrewJDR:
I relatively frequently run into a situation where ```git status``` will show one or more lfs-tracked files as being modified when they are not in fact modified. I have yet to nail down exact reproduction steps, but I do have some bits of information about it:

- ```git checkout -- <the lfs tracked files>``` Will **not** resolve the issue; The files will still be shown as modified in an immediately subsequent ```git status``` output.
- It can be "fixed" by doing the following:
  * Emptying the .gitatttributes file (i.e. run ```cp /dev/null .gitattributes``` or similar)
  * Running ```git status```
  * Running ```git checkout -- .gitattributes``` (Restores the .gitattributes file)
  * Running ```git status``` again. The lfs-tracked files are now not shown as being modified anymore, as expected.
  * (Note that these steps did **not** include a step like ```git checkout -- <lfs tracked files>```, which tells me the files were indeed completely unmodified the entire time.)
- It seems like problem might occur immediately after a ```git reset --hard <tree-ish>``` or ```git checkout <tree-ish>```, but I am not 100% sure on this.

Has anyone else encountered this?

AndrewJDR:
I have noticed #1544, but that issue seems to only happen after a merge conflict on the .gitattributes file, so I'm not sure that is the same issue.

I can confirm that this has happened to me many times even when there was no merge conflict on .gitattributes.

AndrewJDR:
Also, just to be clear --

The lfs tracked files in question in my example have already been committed as lfs-tracked files long ago and have been that way for quite some time.

I just want to clarify that don't have any confusion about how to use git-lfs, and how it works, etc. :)
fwiw, I've observed a similar problem with other tools that use git smudge/clean filters. This might a problem in git-core, but it would be very nice to get it tracked down and resolved/worked around one way or the other, because less experienced users really have no idea how to fix this when it happens, and it prevents them from being able to merge or do any other action that requires a clean working directory/checkout.

AndrewJDR:
I'll post an example of one of my .gitattributes files here just for completeness, but I'll also note that I've experienced this problem on many different repositories with varying .gitattributes files:
```
*.dll filter=lfs diff=lfs merge=lfs -text
*.lib filter=lfs diff=lfs merge=lfs -text
*.ilk filter=lfs diff=lfs merge=lfs -text
*.pdb filter=lfs diff=lfs merge=lfs -text
[...]

PR #1749, opened 2016-12-08
[WIP] test: reproduce git-lfs/git-lfs#1726

ttaylorr:
This pull request introduces a failing test designed to reproduce the bug found in #1726.

---

/cc @AndrewJDR @git-lfs/core

### DIFF
--- a/test/test-dirty-index.sh
+++ b/test/test-dirty-index.sh
@@ -1,0 +1,32 @@
+#!/usr/bin/env bash
+
+. "test/testlib.sh"
+
+begin_test "dirty index (git-lfs/git-lfs#1726)"
+(
+  set -e
+  reponame="dirty-index-prep"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" repo_before_all
+
+  git lfs track "*.dat"
+  git add .gitattributes
+  git commit -m "initial commit"
+
+  printf "contents" > a.dat
+  ln -s a.dat a.symlink.dat
+
+  git add *.dat
+  git commit -m "contents"
+  git push origin master
+
+  # ~
+
+  git config --global core.symlinks false
+
+  clone_repo "$reponame" repo
+  git status --porcelain 1>&1 | tee status.log
+
+  [ 0 -eq "$(grep -c "M a.symlink.dat" status.log)" ]
+)
+end_test


ttaylorr:
Interesting, it appears that this test-case alone isn't enough to reproduce the failure on Windows (see [these][1] [two][2] builds).

[1]: https://ci.appveyor.com/project/git-lfs/git-lfs/build/1.0.772
[2]: https://ci.appveyor.com/project/git-lfs/git-lfs/build/1.0.773

AndrewJDR:
@ttaylorr I'm not sure what the behavior of the ln -s command is on windows, but it might boil down to that.
In my repro, I had to ln -s from Linux, commit/push, and clone on Windows.

sschuberth:
> I'm not sure what the behavior of the ln -s command is on windows

That actually depends on the environment you use, whether it's Cygwin, MSYS2 (i.e. Git for Windows), Bash on Ubuntu on Windows 10 (i.e. Windows Subsystem for Linux) etc. In the first two cases is also depends on whether the `CYGWIN` environment variable contains the `winsymlinks` option, and whether it's set to `lnk`, `native` or `nativestrict`.

> I had to ln -s from Linux, commit/push, and clone on Windows

In that case, Git for Windows [usually creates plain text files](http://stackoverflow.com/a/11664406/1127485) containing the path to the target file.

AndrewJDR:
@ttaylor are you saying you can actually repro this issue on Linux? I have not been able to do this. For me, on Linux, even if core.symlinks is false, git clone still produces actual symlinks, so the issue doesn't happen.

@sschuberth in order to repro this issue on cygwin/msys, I must not be using the winsymlinks setting -- it must be in the mode that creates the plain text files containing the path to the target file.

AndrewJDR:
I should also mention that with (git for Windows), winsymlinks is off by default unless you opt in during installation - and even that option is very new.

Also symlinks seem to need admin privileges to actually work.

So we should probably count on many Windows users not having symlinks enabled, thus they'll run into this issue if the other repro steps come into play for them...

sschuberth:
> I should also mention that with (git for Windows), winsymlinks is off by default unless you opt in during installation

Sorry, I believe you're mixing up some things here. Since you were asking about `ln -s`, my answer also referred to programs that link against the POSIX emulation layer, and `ln.exe` is one of them. For those, the setting of the `winsymlinks` option as part of the `CYGWIN` environment variable is important.

For Git for Windows' `git.exe`, however, which is a native executable that does *not* link against the POSIX emulation layer, the setting of the `winsymlinks` option is not taken into account. All that Git for Windows does care about is the Git-specific `core.symlinks` setting, which the installer only sets to `true` if the option is chosen during the installation process.

> Also symlinks seem to need admin privileges to actually work.

Not necessarily. As explained on the [Git for Windows Wiki](https://github.com/git-for-windows/git/wiki/Symbolic-Links), having the `SeCreateSymbolicLinkPrivilege` privilege basically is enough. 

> So we should probably count on many Windows users not having symlinks enabled

Agreed.

test/test-dirty-index.sh
```diff
@@ -0,0 +1,32 @@
...
+  reponame="dirty-index-prep"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" repo_before_all
+
+  git lfs track "*.dat"
+  git add .gitattributes
+  git commit -m "initial commit"
+
+  printf "contents" > a.dat
+  ln -s a.dat a.symlink.dat
```
sschuberth: The reason why the test does not fails as expect on Windows is that on MSYS2 this line will not actually create a symbolic link unless you do `export MSYS=winsymlinks:nativestrict` beforehand.

sschuberth: The reason why the test does not fail (as would be expect) on Windows is that on MSYS2 this line will not actually create a symbolic link unless you do `export MSYS=winsymlinks:nativestrict` beforehand. Right now it just copies the file.

AndrewJDR:
> Sorry, I believe you're mixing up some things here. 

Yeah, you're absolutely right.

After some testing, I see that Git for Windows "only" needs the core.symlinks=true and SeCreateSymbolicLinkPrivilege on the account being used. The core.symlinks is not a huge deal, but setting the SeCreateSymbolicLinkPrivilege is a bit of a pita for the average user. It seems like it'd be nice if the git for windows installer offered to do the necessary group policy edit for you if you check the "Use symlinks" box during installation, but I digress...

In any case, it'd be great to have this fixed. I poked around for a bit myself, but I think this gets deeper into the intricacies of the smudge/clean stuff than I am able to devote time to learning about at the moment.

AndrewJDR:
@ttaylor Was this fixed? If so, by what commit?

ttaylorr:
> @ttaylor Was this fixed? If so, by what commit?

Not to my knowledge. Can you try and reproduce this issue with the latest tip of master? If it's still an issue, I will be happy to reopen this issue.

PR #1752, opened 2016-12-09
Add more meta-data to the Windows installer executable

sschuberth:
On Windows, this makes the detailed file properties dialog show the file
version and copyright information also for the Windows installer
executable. Previously only the git-lfs executables had this information.


script/windows-installer/inno-setup-git-lfs-installer.iss
```diff
@@ -15,7 +17,7 @@
 AppId={{286391DE-F778-44EA-9375-1B21AAA04FF0}
 AppName={#MyAppName}
 AppVersion={#MyAppVersion}
-;AppVerName={#MyAppName} {#MyAppVersion}
+AppCopyright=GitHub, Inc. and Git LFS contributors
```
ttaylorr: I think this is 🆗 since it matches the text in `LICENSE.md`, but I would love to get some extra :eyes: on this just to be sure.

cc @bkeepers

technoweenie: I think the fact that it's in our license file means it's a-ok. I consider it the source of truth over this `iss` file. Can you verify that it compiles, and that the file properties show it so we can approve and merge this?

sschuberth: I did already verify this. You can also easily do it yourself, it you want, by simply downloading the [installer artifact from the AppVeyor build](https://ci.appveyor.com/api/buildjobs/ycarh44lni2qr7so/artifacts/git-lfs-windows-1.5.0.exe) (this does not have the most recent version number as in `master` we still are at `1.5.0`). See this screenshot:

ttaylorr:
@sschuberth Thanks for verifying. I booted up my Windows VM this morning to check for myself, but I can't build the Windows binaries as I used to. Previously, I did a `script/bootstrap` and copied the artifacts up as `git-lfs-windows-<arch>.exe` and was able to build the installer. This no longer works, as I get an error saying that the build artifacts are missing metadata.

I'm assuming there's some other step that I have to do in order to apply metadata onto the build artifacts, but that step isn't clear to me. Could you please document how to do this somewhere?

sschuberth:
> [...] copied the artifacts up as `git-lfs-windows-<arch>.exe` and was able to build the installer.

@ttaylorr I believe your naming is just wrong: The Windows executables should be named `git-lfs-x64.exe` and `git-lfs-x86.exe`, without the `-windows` infix. But AFAIR that has always been like that.

ttaylorr:
@sschuberth, sorry I didn't mean to type the `-windows` portion of the filename, my filenames are correct. The error seems unrelated.

screen shot 2016-12-13 at 12 27 22 pm

sschuberth:
Ah, right, now I recall: In `appveyor.yml` I added `C:\Ruby23\DevKit\mingw\bin\windres.exe script\windows-installer\resources.rc -o resources.syso`, which I deliberately did *not* (yet) include to `script/bootstrap` as the path to `windres.exe` is unclear in the general case. Anyway, it's inconvenient that this step is missing from the script for local builds. I'll think about a possible fix. Do you have `windres.exe` in your Windows VM?

ttaylorr:
> Do you have `windres.exe` in your Windows VM?

I do not. Seems like to get this you need to install MinGW, which I did, but it is still not present anywhere in my `$PATH`.

sschuberth:
Ok. I think I'll rework embedding of resources to use https://github.com/josephspurrier/goversioninfo instead of `windres`, and add the changes to this PR.

sschuberth:
@ttaylorr Ok, it took me several tries to get this right, but now running `script/bootstrap` locally should work just fine again as I do not rely on `windres` to be present anymore, but install and use `goversioninfo` as part of Windows builds.

ttaylorr:
@sschuberth awesome work! Everything works as expected:

screen shot 2016-12-15 at 10 55 43 am

script/bootstrap
```diff
@@ -14,6 +14,14 @@ if [ -z "$GOPATH" ]; then
     ln -s "$GOPATH" src/github.com/git-lfs/git-lfs
 fi
 
+if uname -s | grep -q "_NT-"; then
+  echo "Installing goversioninfo to embed resources into Windows executables..."
```
ttaylorr: Do we have to do this every time `script/bootstrap` is run? I think we should wrap this in a:

```sh
if [ ! -d "$GOPATH/src/github.com/josephspurrier/goversioninfo" ]; then
  echo "Installing goversioninfo ..."
  go get github.com/josephspurrier/goversioninfo/cmd/goversioninfo
fi

export PATH="$GOPATH/bin/windows_386:$PATH"
echo "Creating the resource.syso version information file..."

go generate
```

update-version.sh
```diff
@@ -4,7 +4,8 @@ VERSION_STRING=$1
 VERSION_ARRAY=( ${VERSION_STRING//./ } )
```
ttaylorr: Sorry I didn't catch this before: is it possible we could put this in the `script/` directory?

versioninfo.json
```diff
@@ -0,0 +1,18 @@
+{
```
ttaylorr: Let's call this file just `version.json`.

script/bootstrap
sschuberth: I believe `go get` is smart enough to do something like that itself. Here's some timing:

    $ time go get github.com/josephspurrier/goversioninfo/cmd/goversioninfo

    real    0m4.039s
    user    0m0.015s
    sys     0m0.000s

    $ time go get github.com/josephspurrier/goversioninfo/cmd/goversioninfo

    real    0m0.245s
    user    0m0.000s
    sys     0m0.000s

As you can see, the second time `go get` takes only 0.2s.

Anyway, I was considering to use Glide for `goversioninfo` and commit it under `vendor`, but when I tried quickly I wasn't able to get it working correctly. Would be using Glide be the better approach, even though `goversioninfo` is only used for the Windows build?

versioninfo.json
sschuberth: The `versioninfo.json` name [seems to be hard-coded](https://github.com/josephspurrier/goversioninfo#usage) into `goversioninfo`, unfortunately.

Issue #1651, opened 2016-11-14
Proposal: `transfer` package for v2.0.0


ttaylorr:
This issue proposes a new design for the `transfer` package, which will supersede the `lfs/transfer_queue.go` code and serve as the sole mechanism to transfer objects to and from LFS servers.

## Motivation

The current transfer-queue lacks some features that we’d like to implement for the [v2.0.0][1] release of Git LFS:

1. __Less buffering.__ LFS should be better at dealing with really large working trees during initial pushes, or `push —all`’s. Currently, when retrying an object, it is re-inserted at the last spot in the queue, meaning that we could be holding a potentially large number of OIDs in memory before retrying them.
2. __Async.__ Currently, we have to wait for the entire working tree scan to complete before beginning the transfer. This is fine for small trees, but yields undesirable behavior for larger ones. We should be able to start uploading as soon as we have enough data to do so. 
3. __Backpressure.__ We should synchronize the work being done by the [scanner][2] package with the work being done by the `transfer` package in uploading objects that the scanner finds. In other words: if we’re network-bound while uploading, we should delay scanning for more objects until we can resume uploading.
4. __Resumable transfers, by default.__ We should be able to resume uploads and downloads as the default if a transfer fails part-way through. This, combined with better retry behavior will yield for much more resilient and efficient transfers.
5. __Better design.__ Designing a new package for the `transfer` code, thus moving it out of the `lfs` package affords us the opportunity to re-imagine it from first principles, and define a solid, public API that can be used from other packages, and other programs.

[1]: https://github.com/github/git-lfs/milestone/3
[2]: https://github.com/github/git-lfs/issues/1649

## Proposal

What follows is a high-level overview of the mechanism I’d like to implement in the `scanner` package, and a description of each of its major components (+ how it relates to the motivations above):

![](https://cloud.githubusercontent.com/assets/443245/19457770/1bf3c4b6-9485-11e6-84f5-5a5049305cb0.png)

### Async & backpressure

We should halt the scanning process when there are no available workers free to transfer objects. We can do this by using the following flow:

1. Accept items using a `<-chan *lfs.Pointer` to build up a single batch of items
2. If the currently accumulating batch has a cardinality smaller than twice the desired size of a batch (2*100), _OR_ if the `<-chan` remains open, go to step 1.
3. Make a batch API call with the accumulated items
4. Distribute the transfers across `n` workers.
5. While all workers are busy, go to step 4.
6. As soon as a worker is free, go to step 1.
[...]

PR #1758, opened 2016-12-12
tq: prioritize transferring retries before new items

ttaylorr:
This pull-request implements priority retries, and moves `TransferQueue`-related code into a new package, `tq`, as discussed in #1651.

## Why?

In an effort to better support transferring large numbers of objects through the `*tq.TransferQueue`, we need to buffer less data than we currently do. One way to do this is to process retries up front, instead of keeping data in memory about failed objects around until the end of the transfer. Normally, this isn't a huge problem, but if the transfer queue fails half a million objects, we can't afford to keep that in memory.

## What?

This pull request brings a number of benefits to the `*TransferQueue` type:

1. **Back-pressure**. If the `*TransferQueue` can't accept more items, there's no sense in wasting a large amount of disk and CPU usage scanning the Git data. Previously, instances of `git rev-list`, `git cat-file` and `git cat-file --batch` would run ad nauseam even if the `TransferQueue` couldn't accept new items. Now, `Add()` will block after a maximum buffer depth has been reached. In other words, by default, `Add()` will accept 200 (twice the default batch size of 100) items before apply back-pressure up to the `gitscanner`, causing it to wait. 
2. **Support for many objects**: Right now, we keep a large amount of data in memory per each object retry. Before https://github.com/git-lfs/git-lfs/pull/1535, that data was kept around until the very end of the transfer and then retried all at once. That got better after that PR was merged, but at the cost of many more API calls. This PR encompasses the best of both worlds: by retrying objects closer to when they were initially `Add()`-ed, data about that object can be kept in memory for less time. By treating retries as part of the next batch and prioritizing them, we make less API calls.
3. **Less API calls**. As of https://github.com/git-lfs/git-lfs/pull/1535, we make an API call (about) _once per object retry_. Normally this isn't such a huge problem, but if many objects fail, we'll be making a proportional number of expensive API calls. Not cool. This pull-request, by prioritizing object retries before accepting new items cuts the number of batch API calls by the size of a batch (100, by default). Much better.

## Inner-workings

This pull-request fundamentally changes the way that we process batches and retries in the transfer queue. Here's a breakdown of what goes on:

1. Start with an empty batch.
2. Until the batch is full OR the `<-q.incoming` channel is closed, append from `Add()` via `<-q.incoming` items into the batch.
3. Make a batch API request, and send the results to the transfer adapters.
4. Clear the batch.
5. Collect the failed items into the next batch.
6. If the `<-q.incoming` channel is closed AND the next batch is empty, exit.
7. Otherwise, go to step 2. 

This requires us to change how the `*tq.adapterBase` (previously `*transfer.adapterBase`) type is implemented. The changes can be summarized by the signature change. What used to be:

```go
func (a *adapterBase) Add(t *Transfer)
```

is now:

```go
func (a *adapterBase) Add(ts ...*Transfer) (results <-chan TransferResult)
```

by synchronizing the results with a batch of `*Transfer` items, we can deterministically pre-fill the next batch with failed items from the last one.

This change comes at a cost, which is that in order to know when to `close()` the `results` channel, we must wait for all of the transfer adapters to finish their in-progress tasks. This effect will be un-noticeable if the items in a transfer set are of uniform size, but is more pronounced when items with wildly different sizes are processed in a random order with a poor network connection.

To combat this, items in a batch are sorted in descending object size to minimize the chance that a worker will get tied up on a large object as the last item in a batch. This change, implemented in 6decfe4 should make this a non-issue.

---

Since this PR is a on the bigger side, here's a convenient break-down of the changes:

1. 453d02a: Introduce `batchSize` instance variable (and `OptionFn`) to allow configuring the batch size.
2. bdcacb0: Implement priority retries.
3. 6decfe4: Sort batches by descending object size to minimize chance of idle workers.
4. d0c4ccd...55275ca: Move `lfs/transfer_queue.go` and `transfer` package into new package, `tq`.
5. 70f1479...8c94383: Rename various exported methods/types in `tq` package to avoid [stuttering][1].

[1]: https://blog.golang.org/package-names#TOC_3.

---

/cc @git-lfs/core

### DIFF
--- a/tq/adapterbase.go
+++ b/tq/adapterbase.go
@@ -25,7 +25,6 @@ type adapterBase struct {
 	transferImpl transferImplementation
 	jobChan      chan *job
 	cb           ProgressCallback
-	outChan      chan TransferResult
 	// WaitGroup to sync the completion of all workers
 	workerWait sync.WaitGroup
 	// WaitGroup to sync the completion of all in-flight jobs
@@ -70,9 +69,8 @@ func (a *adapterBase) Direction() Direction {
 	return a.direction
 }
 
-func (a *adapterBase) Begin(maxConcurrency int, cb ProgressCallback, completion chan TransferResult) error {
+func (a *adapterBase) Begin(maxConcurrency int, cb ProgressCallback) error {
 	a.cb = cb
-	a.outChan = completion
 	a.jobChan = make(chan *job, 100)
 
 	tracerx.Printf("xfer: adapter %q Begin() with %d workers", a.Name(), maxConcurrency)
@@ -93,25 +91,17 @@ func (a *adapterBase) Begin(maxConcurrency int, cb ProgressCallback, completion 
 type job struct {
 	T *Transfer
 
-	listeners []chan<- TransferResult
-	wg        *sync.WaitGroup
+	results chan<- TransferResult
+	wg      *sync.WaitGroup
 }
 
 func (j *job) Done(err error) {
-	for _, l := range j.listeners {
-		l <- TransferResult{j.T, err}
-	}
-
+	j.results <- TransferResult{j.T, err}
 	j.wg.Done()
 }
 
 func (a *adapterBase) Add(transfers ...*Transfer) <-chan TransferResult {
 	results := make(chan TransferResult, len(transfers))
-
-	listeners := []chan<- TransferResult{results}
-	if a.outChan != nil {
-		listeners = append(listeners, a.outChan)
-	}
 
 	a.jobWait.Add(len(transfers))
 
@@ -121,7 +111,7 @@ func (a *adapterBase) Add(transfers ...*Transfer) <-chan TransferResult {
 		for _, t := range transfers {
 			// BUG(taylor): End() is race-y here, and can close
 			// jobChan before we want it to
-			a.jobChan <- &job{t, listeners, a.jobWait}
+			a.jobChan <- &job{t, results, a.jobWait}
 		}
 
 		a.jobWait.Wait()
@@ -138,9 +128,6 @@ func (a *adapterBase) End() {
 
 	// wait for all transfers to complete
 	a.workerWait.Wait()
-	if a.outChan != nil {
-		close(a.outChan)
-	}
 
 	tracerx.Printf("xfer: adapter %q stopped", a.Name())
 }
--- a/tq/adapterbase.go
+++ b/tq/adapterbase.go
@@ -24,7 +24,7 @@ type adapterBase struct {
 	direction    Direction
 	transferImpl transferImplementation
 	jobChan      chan *job
-	cb           TransferProgressCallback
+	cb           ProgressCallback
 	outChan      chan TransferResult
 	// WaitGroup to sync the completion of all workers
 	workerWait sync.WaitGroup
@@ -49,7 +49,7 @@ type transferImplementation interface {
 	// Implementations can clean up per-worker resources here, context is as returned from WorkerStarting
 	WorkerEnding(workerNum int, ctx interface{})
 	// DoTransfer performs a single transfer within a worker. ctx is any context returned from WorkerStarting
-	DoTransfer(ctx interface{}, t *Transfer, cb TransferProgressCallback, authOkFunc func()) error
+	DoTransfer(ctx interface{}, t *Transfer, cb ProgressCallback, authOkFunc func()) error
 }
 
 func newAdapterBase(name string, dir Direction, ti transferImplementation) *adapterBase {
@@ -70,7 +70,7 @@ func (a *adapterBase) Direction() Direction {
 	return a.direction
 }
 
-func (a *adapterBase) Begin(maxConcurrency int, cb TransferProgressCallback, completion chan TransferResult) error {
+func (a *adapterBase) Begin(maxConcurrency int, cb ProgressCallback, completion chan TransferResult) error {
 	a.cb = cb
 	a.outChan = completion
 	a.jobChan = make(chan *job, 100)
@@ -211,7 +211,7 @@ func (a *adapterBase) worker(workerNum int, ctx interface{}) {
 	a.workerWait.Done()
 }
 
-func advanceCallbackProgress(cb TransferProgressCallback, t *Transfer, numBytes int64) {
+func advanceCallbackProgress(cb ProgressCallback, t *Transfer, numBytes int64) {
 	if cb != nil {
 		// Must split into max int sizes since read count is int
 		const maxInt = int(^uint(0) >> 1)
--- a/tq/adapterbase.go
+++ b/tq/adapterbase.go
@@ -109,8 +109,6 @@ func (a *adapterBase) Add(transfers ...*Transfer) <-chan TransferResult {
 		defer close(results)
 
 		for _, t := range transfers {
-			// BUG(taylor): End() is race-y here, and can close
-			// jobChan before we want it to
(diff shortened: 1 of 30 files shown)


tq/adapterbase.go
```diff
@@ -82,19 +90,58 @@ func (a *adapterBase) Begin(maxConcurrency int, cb TransferProgressCallback, com
...
+
+	a.jobWait.Add(len(transfers))
+
+	go func() {
+		defer close(results)
+
+		for _, t := range transfers {
+			// BUG(taylor): End() is race-y here, and can close
+			// jobChan before we want it to
+			a.jobChan <- &job{t, listeners, a.jobWait}
```
sinbad: Maybe I'm missing something, but if you're adding `len(transfers)` to `jobWait`, and `Done` is only ever called as a response to completion of items in `jobChan` on a 1:1 basis, `jobWait.Wait()` should never return before this loop is done, therefore the `close` in `End()` is safe?

tq/adapterbase.go
```diff
@@ -148,10 +197,8 @@ func (a *adapterBase) worker(workerNum int, ctx interface{}) {
 			err = a.transferImpl.DoTransfer(ctx, t, a.cb, authCallback)
 		}
 
-		if a.outChan != nil {
-			res := TransferResult{t, err}
-			a.outChan <- res
-		}
+		// Mark the job as completed, and alter all listeners
+		job.Done(err)
```
sinbad: I'm unreasonably happy with how `job.Done()` scans 😄

tq/transfer_queue.go
```diff
@@ -87,70 +87,101 @@ func (r *retryCounter) CanRetry(oid string) (int, bool) {
 	return count, count < r.MaxRetries
 }
 
+type Batch []Transferable
+
+func (b Batch) IsFull(max int) bool { return len(b) >= max }
```
sinbad: Since you have to supply the max as an argument I'm not sure what value this function adds?

tq/transfer_queue.go
```diff
@@ -165,12 +196,266 @@ func (q *TransferQueue) Add(t Transferable) {
...
+				break
+			}
+
+			batch = append(batch, t)
+		}
+
+		// Before enqueuing the next batch, sort it's items by size
+		// (largest first) so that workers can process the largest
+		// transfers first.
+		sort.Sort(sort.Reverse(batch))
```
sinbad: Sorry, I don't understand what benefit this sort is giving you. Since you're only sorting within the batch, and the batch is the unit that is blocked on, what does it matter what the ordering is inside the batch?

I could imagine that if you sorted all the transfers as a whole so that bigger files made it into the first batch that this would alter the blocking behaviour, but sorting inside the batch seems incidental. Please explain?

tq/transfer_queue.go
```diff
@@ -165,12 +196,266 @@ func (q *TransferQueue) Add(t Transferable) {
...
+		retries, err := q.enqueueAndCollectRetriesFor(batch)
+		if err != nil {
+			q.errorc <- err
+		}
+
+		if closing && len(retries) == 0 {
+			break
+		}
+
+		batch = retries
```
sinbad: I like how the original & retries are merged into one batch, makes everything simpler.

technoweenie:
I'd really like to remove the coupling on the `config` package. I would've preferred to review "prioritize transferring retries" first, and then work out the `tq` move/extraction stuff separately. What do you think about changing the base of this PR to `tq-master`?

tq/adapterbase.go
ttaylorr: This is correct, if the calling code occurs in a single goroutine. Consider this:

```go
func main() {
        var a *adapterBase

        go a.Add(myBatch)

        a.Wait()
}
```

If we're say, 50 items into `myBatch` before the execution scheduler takes us to the `a.Wait()` call, it'll close `a.jobChan`, which is what we expect. Once the scheduler takes us back to the goroutine, it'll try and process the 51st item and write into a closed channel, which is a panic. Even if it weren't a panic, we don't want to lose those items.

ttaylorr: Excuse me, I hit enter a little too early :-). What I wanted to add is that even though this is a fairly fringe scenario, I still think it's important to consider.

tq/transfer_queue.go
ttaylorr: Yeah, I was a little hesitant about adding this function. I think it provides minimal value for the reason that you stated, and it expands the `Batch` interface. I've removed it in ced37c6c.

tq/transfer_queue.go
ttaylorr: No worries, I wasn't quite clear with my reasoning here. The idea is to reduce the likelihood of a single worker getting tied up processing one very large item at the end of a batch while others are sitting idle. Since enqueuing the next batch is contingent on processing the current batch completely, idle workers should be treated as something we need to minimize as much as possible.

Specifically, the scenario I'm thinking about is a batch of 100 items where the first 99 are 5MB and the last item is 2GB. Processed in sequence, the workers will quickly finish transferring the small items. Towards the very end of the sequence, one worker will be sitting transferring a very large object, while the remaining workers are sitting doing nothing.

If the 2GB item is processed first, it ties a worker up for the duration of the transfer as it always would, but it allows the other workers to tear through the small items, minimizing the chance that a small number of large transfers will prolong the duration of the batch.

The sort makes this property much less noticeable, as does a larger number of workers, or a more evenly sized workload.

I went ahead and clarified this in b9aa0986.

ttaylorr:
> I would've preferred to review "prioritize transferring retries" first, and then work out the `tq` move/extraction stuff separately.

This makes much more sense than the approach I pursued. I was finished with the in-place refactoring on Saturday night, and wanted to keep moving forward. However, I didn't want to create two PRs that would be in-flight at once, since I figured that'd be tough to review. Instead, I opted to create one large PR, which was a mistake.

> What do you think about changing the base of this PR to `tq-master`?

I think that's a good idea. I reset `tq-master` to the current HEAD of `master`, and made the base of this PR points at the `tq-master` branch.

tq/transfer_queue.go
sinbad: OK I see, so it's saving you from the case where one massive file is the absolute last thing off the queue when everything else is already finished. I'm not sure if in practice it's likely to make much of difference but I understand where you're coming from now :)

tq/adapterbase.go
ttaylorr: Doh! Sorry, I thought you were asking about my use of `a.jobWait.Wait()` in `Done()`, which I added after I added that `// BUG(taylor)` comment while debugging. Looks like I just forgot to remove that comment, which I did in ce08f661. I think we just found out in a round-about way that we're on the same page!

ttaylorr:
@technoweenie I implemented the changes that we discussed this morning in cf409ac... d7bf608. 😄

ttaylorr:
@technoweenie sounds good, and thanks for the 👍. I'm thinking that the follow-up PRs can happen in the order that you suggested, since the only strict dependency is that the `tq.Object` stuff happen last.

Additionally before we merge `tq-master` into `master`, I'd like to look at the `*adapterBase` and narrow the interface down to `WorkerStarting()`, `WorkerEnding()`, and `DoTransfer()`. That will have to happen after the `tq.Object`, since I'm thinking that one is a dependency of the other.

I think after all of that, this should be 🆗 to merge into `master`.

ttaylorr:
Opened up a tracking issue and am tracking progress in https://github.com/git-lfs/git-lfs/issues/1764.

PR #1760, opened 2016-12-12
Implement local lock cache, support querying it

sinbad:
The next incremental rewrite of the locking work; this time we're using our key value store to locally cache locks for the current user for more efficient access. This will be used in the next PR to provide efficient support for marking files read-only on checkout if not locked.

Also adds some minor API improvements such as returning a full `Lock` structure on acquiring a lock rather than just an Id.

### DIFF
--- a/locking/cache.go
+++ b/locking/cache.go
@@ -44,7 +44,7 @@ func (c *LockCache) isIdKey(key string) bool {
 }
 
 // Cache a successful lock for faster local lookup later
-func (c *LockCache) CacheLock(l Lock) error {
+func (c *LockCache) Add(l Lock) error {
 	// Store reference in both directions
 	// Path -> Lock
 	c.kv.Set(l.Path, &l)
@@ -54,7 +54,7 @@ func (c *LockCache) CacheLock(l Lock) error {
 }
 
 // Remove a cached lock by path becuase it's been relinquished
-func (c *LockCache) CacheUnlockByPath(filePath string) error {
+func (c *LockCache) RemoveByPath(filePath string) error {
 	ilock := c.kv.Get(filePath)
 	if lock, ok := ilock.(*Lock); ok && lock != nil {
 		c.kv.Remove(lock.Path)
@@ -65,7 +65,7 @@ func (c *LockCache) CacheUnlockByPath(filePath string) error {
 }
 
 // Remove a cached lock by id because it's been relinquished
-func (c *LockCache) CacheUnlockById(id string) error {
+func (c *LockCache) RemoveById(id string) error {
 	// Id as key is encoded
 	idkey := c.encodeIdKey(id)
 	ilock := c.kv.Get(idkey)
@@ -77,7 +77,7 @@ func (c *LockCache) CacheUnlockById(id string) error {
 }
 
 // Get the list of cached locked files
-func (c *LockCache) CachedLocks() []Lock {
+func (c *LockCache) Locks() []Lock {
 	var locks []Lock
 	c.kv.Visit(func(key string, val interface{}) bool {
 		// Only report file->id entries not reverse
--- a/locking/cache.go
+++ b/locking/cache.go
@@ -56,8 +56,7 @@ func (c *LockCache) CacheLock(l Lock) error {
 // Remove a cached lock by path becuase it's been relinquished
 func (c *LockCache) CacheUnlockByPath(filePath string) error {
 	ilock := c.kv.Get(filePath)
-	if ilock != nil {
-		lock := ilock.(*Lock)
+	if lock, ok := ilock.(*Lock); ok && lock != nil {
 		c.kv.Remove(lock.Path)
 		// Id as key is encoded
 		c.kv.Remove(c.encodeIdKey(lock.Id))
@@ -70,8 +69,7 @@ func (c *LockCache) CacheUnlockById(id string) error {
 	// Id as key is encoded
 	idkey := c.encodeIdKey(id)
 	ilock := c.kv.Get(idkey)
-	if ilock != nil {
-		lock := ilock.(*Lock)
+	if lock, ok := ilock.(*Lock); ok && lock != nil {
 		c.kv.Remove(idkey)
 		c.kv.Remove(lock.Path)
 	}
--- a/locking/cache.go
+++ b/locking/cache.go
@@ -1,0 +1,103 @@
+package locking
+
+import (
+	"strings"
+
+	"github.com/git-lfs/git-lfs/tools/kv"
+)
+
+const (
+	// We want to use a single cache file for integrity, but to make it easy to
+	// list all locks, prefix the id->path map in a way we can identify (something
+	// that won't be in a path)
+	idKeyPrefix string = "*id*://"
+)
+
+type LockCache struct {
+	kv *kv.Store
+}
+
+func NewLockCache(filepath string) (*LockCache, error) {
+	kv, err := kv.NewStore(filepath)
+	if err != nil {
+		return nil, err
+	}
+	return &LockCache{kv}, nil
+}
+
+func (c *LockCache) encodeIdKey(id string) string {
+	// Safety against accidents
+	if !c.isIdKey(id) {
+		return idKeyPrefix + id
+	}
+	return id
+}
+func (c *LockCache) decodeIdKey(key string) string {
+	// Safety against accidents
+	if c.isIdKey(key) {
+		return key[len(idKeyPrefix):]
+	}
+	return key
+}
+func (c *LockCache) isIdKey(key string) bool {
+	return strings.HasPrefix(key, idKeyPrefix)
+}
+
+// Cache a successful lock for faster local lookup later
+func (c *LockCache) CacheLock(l Lock) error {
+	// Store reference in both directions
+	// Path -> Lock
+	c.kv.Set(l.Path, &l)
+	// EncodedId -> Lock (encoded so we can easily identify)
+	c.kv.Set(c.encodeIdKey(l.Id), &l)
+	return nil
+}
+
+// Remove a cached lock by path becuase it's been relinquished
+func (c *LockCache) CacheUnlockByPath(filePath string) error {
+	ilock := c.kv.Get(filePath)
+	if ilock != nil {
+		lock := ilock.(*Lock)
+		c.kv.Remove(lock.Path)
+		// Id as key is encoded
+		c.kv.Remove(c.encodeIdKey(lock.Id))
+	}
+	return nil
+}
+
+// Remove a cached lock by id because it's been relinquished
+func (c *LockCache) CacheUnlockById(id string) error {
+	// Id as key is encoded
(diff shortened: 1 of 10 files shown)


sinbad:
I think with the merging of this PR, `locking-master` can be merged into `master` because it's a stable mid-stage. There will then be a few further staging posts between here and when the locking is completely finished.

locking/locks.go
```diff
@@ -47,15 +50,20 @@ func NewClient(cfg *config.Configuration) (*Client, error) {
...
+// Close this client instance; must be called to dispose of resources
+func (c *Client) Close() error {
+	return c.cache.Save()
+}
+
 // LockFile attempts to lock a file on the current remote
 // path must be relative to the root of the repository
 // Returns the lock id if successful, or an error
-func (c *Client) LockFile(path string) (id string, e error) {
+func (c *Client) LockFile(path string) (Lock, error) {
```
ttaylorr: I think returning a `*Lock` would be appropriate here, since this method doesn't guarantee that locking a file will succeed all of the time, and the zero-value `Lock{}` doesn't make sense.

locking/locks.go
```diff
@@ -127,39 +135,55 @@ type Lock struct {
...
 	Email string
-	// CommitSHA is the commit that this Lock was created against. It is
-	// strictly equal to the SHA of the minimum commit negotiated in order
-	// to create this lock.
-	CommitSHA string
-	// LockedAt is a required parameter that represents the instant in time
-	// that this lock was created. For most server implementations, this
-	// should be set to the instant at which the lock was initially
-	// received.
+	// LockedAt tells you when this lock was acquired.
```
ttaylorr: `s/tells you/is the/`

locking/locks.go
```diff
@@ -127,39 +135,55 @@ type Lock struct {
...
 	}
 }
 
 // SearchLocks returns a channel of locks which match the given name/value filter
 // If limit > 0 then search stops at that number of locks
-func (c *Client) SearchLocks(filter map[string]string, limit int) (locks []Lock, err error) {
+// If localOnly = true, don't query the server & report only own local locks
+func (c *Client) SearchLocks(filter map[string]string, limit int, localOnly bool) (locks []Lock, err error) {
+
+	if localOnly {
```
ttaylorr: 😻

locking/locks.go
```diff
@@ -127,39 +135,55 @@ type Lock struct {
...
+
+func (c *Client) searchCachedLocks(filter map[string]string, limit int) ([]Lock, error) {
+	cachedlocks := c.cachedLocks()
+	path, filterByPath := filter["path"]
+	id, filterById := filter["id"]
+	lockCount := 0
+	locks := make([]Lock, 0, len(cachedlocks))
+	for _, l := range cachedlocks {
+		// Manually filter by Path/Id
+		if (filterByPath && path != l.Path) ||
```
ttaylorr: I follow this logic, but I think we can make it clearer:

```go
for _, l := range cachedLocks {
        matchesPath := path == l.Path
        matchesId := id == l.Id

        if (filterByPath && matchesPath) || (filterById && matchesId) {
                // ...
        }
}
```

This does add a level of indentation to the body of the loop, but I think the extra clarity is worth it.

locking/locks.go
```diff
@@ -225,33 +249,101 @@ func (c *Client) lockIdFromPath(path string) (string, error) {
 	}
 }
 
+// We want to use a single cache file for integrity, but to make it easy to
+// list all locks, prefix the id->path map in a way we can identify (something
+// that won't be in a path)
+const idKeyPrefix = "*id*://"
+
+func (c *Client) encodeIdKey(id string) string {
```
ttaylorr: Not sure if `(c *Client)` is the right receiver for this method. What do you think about wrapping `tools/kv` in this package with functions that understand `*Lock` and etc, and putting this method there?

```go
package locking

const (
       idPrefix string = "*id*://"
)

type LockCache struct {
        kv *kv.Store
}

func (c *LockCache) CacheLock(l Lock) error {
        // ...
}

func (c *LockCache) encodeId(id string) string {
       if !c.isId(id) {
               return idPrefix+id
        }

       return id
}

func (c *LockCache) isId(key string) bool {
       return strings.HasPrefix(key, idPrefix)
}
```

ttaylorr: I like this approach a lot. I think it's a nice, clear line of abstraction that we can draw. The `*locking.Client` would own a `*LockCache` underneath and delegate into the cache's public methods from there.

locking/locks_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+	oldStore := config.LocalGitStorageDir
+	config.LocalGitStorageDir, err = ioutil.TempDir("", "testCacheLock")
+	assert.Nil(t, err)
+	defer func() {
+		os.RemoveAll(config.LocalGitStorageDir)
+		config.LocalGitStorageDir = oldStore
+	}()
+
+	client, err := NewClient(config.NewFrom(config.Values{}))
```
ttaylorr: Here's another place where I think the approach I proposed above would yield a win. This test is testing the behavior of caching a lock, but creates a locking client to do so. Moving that behavior down into a more isolated part of the system would mean that this test can be closer to it's subject.

locking/locks.go
```diff
@@ -225,33 +249,101 @@ func (c *Client) lockIdFromPath(path string) (string, error) {
...
 	return nil
 }
 
 // Remove a cached lock by path becuase it's been relinquished
-func (c *Client) cacheUnlock(filePath string) error {
-	// TODO
+func (c *Client) cacheUnlockByPath(filePath string) error {
+	ilock := c.cache.Get(filePath)
+	if ilock != nil {
+		lock := ilock.(*Lock)
```
ttaylorr: Not that this is likely behavior, but I think that using the two-tuple variant of this cast would be safer in that it'd prevent an unlikely panic:

```go
func (c *Client) cacheUnlockByPath(filePath string) error {
        ilock := c.cache.Get(filePath)
        if ilock == nil { // <- needed to prevent `(*Lock)(nil)`
                return nil
        }
        if l, ok := ilock.(*Lock); ok {
                // ...
        }

        return nil
}
```

locking/locks.go
```diff
@@ -225,33 +249,101 @@ func (c *Client) lockIdFromPath(path string) (string, error) {
...
-	// TODO
+	// Id as key is encoded
+	idkey := c.encodeIdKey(id)
+	ilock := c.cache.Get(idkey)
+	if ilock != nil {
+		lock := ilock.(*Lock)
+		c.cache.Remove(idkey)
+		c.cache.Remove(lock.Path)
+	}
 	return nil
```
ttaylorr: Do you think we should return an error if we couldn't find the lock? I think a sentinel value here would be fine:

```go
package locking

var ErrMissingLock error = errors.New("locking: could not find lock")
```

locking/locks.go
```diff
@@ -225,33 +249,101 @@ func (c *Client) lockIdFromPath(path string) (string, error) {
...
 }
 
 // Get the list of cached locked files
-func (c *Client) cachedLocks() []string {
-	// TODO
-	return nil
+func (c *Client) cachedLocks() []Lock {
+	var locks []Lock
+	c.cache.Visit(func(key string, val interface{}) bool {
```
ttaylorr: This API is awesome 👍

locking/locks.go
sinbad: Actually I remember this was deliberately by value now. Returning a non-pointer is important because I don't want to allow the client to modify the original lock, because it's cached and therefore they could corrupt the cache by doing that. This outweighs the zero value Lock concern IMO. Yes I could return a pointer and copy it but it's unnecessary complexity. A zero value in the event of an error is idiomatic anyway (nil is just a zero value pointer after all).

locking/locks.go
sinbad: I hate to be the optimisation nitpicker here but doing it my way avoids comparison operations in the case of `!filterByPath` or `!filterById` whereas your way requires the comparison all the time, so I prefer my way 😄  I think it's clear enough.

locking/locks.go
sinbad: Yeah that's a good shout.

locking/locks.go
sinbad: Cool, I've just collapsed this into:
```go
if lock, ok := ilock.(*Lock); ok && lock != nil {
```

Since the actual casting of nil is safe anyway so long as we check.

locking/locks.go
sinbad: I always think it's better to fail silently when requesting to remove a non-existent item from any collection; this is idiomatic with core collections and what I would expect from an API.

locking/locks_test.go
sinbad: Yep, it made it better :)

locking/cache.go
```diff
@@ -0,0 +1,101 @@
...
+		return key[len(idKeyPrefix):]
+	}
+	return key
+}
+func (c *LockCache) isIdKey(key string) bool {
+	return strings.HasPrefix(key, idKeyPrefix)
+}
+
+// Cache a successful lock for faster local lookup later
+func (c *LockCache) CacheLock(l Lock) error {
```
technoweenie: I think having functions with a `Cache` prefix is redundant. Especially in the `*Client` usage: `c.cache.CacheUnlockById(id)`. What do you think about method names like this:

* `CacheLock` -> `Set`
* `CacheUnlockByPath` -> `DeleteByPath`
* `CacheUnlockById` -> `DeleteById`
* `CachedLocks` -> `GetAll` or just `All`

locking/locks.go
technoweenie: I agree that the original is clear enough.

locking/locks.go
```diff
@@ -225,33 +248,29 @@ func (c *Client) lockIdFromPath(path string) (string, error) {
...
 }
 
-// Cache a successful lock for faster local lookup later
-func (c *Client) cacheLock(filePath, id string) error {
-	// TODO
-	return nil
-}
+// Fetch locked files for the current committer and cache them locally
+// This can be used to sync up locked files when moving machines
+func (c *Client) fetchLocksToCache() error {
```
technoweenie: I think something like `refreshLocksCache()` would be a better name.

locking/locks.go
ttaylorr: I wouldn't mind implementing a `(l *Lock) Copy() *Lock` function, but I don't feel strongly.

locking/locks.go
ttaylorr: 👍

locking/locks.go
technoweenie: That depends on callers doing the right thing (not modifying return values from the locking client. I think cache corruption is a good enough reason to return a value.

locking/cache.go
sinbad: Fair point; to keep it distinct from `kv` I'll make it `Add` / `RemoveByBlah` / `All`.

PR #1767, opened 2016-12-13
Manifest config attempt 2

technoweenie:
This is a slightly different approach to #1766. Instead of introducing a new `tq.WithGitEnv()` option, this assumes that the caller will configure the `*tq.Manifest`, which is now a required argument for `NewTransferQueue()`. I was able to remove all `config` references inside the `tq` package except for the internal uses (which is tied to internal api methods still using it). Nothing in the exported interface uses `config` anymore though.

I like this approach, it gets us closer to @sinbad's idea of a config object per package. Could even rename `*tq.Manifest` to `*tq.Config` if you think that'd make more sense.

Notes:

* <del>Had to export `BasicTransfersOnly` and `TusTransfersAllowed` allowed, because `Unmarshal()` can only set exported properties. I'd rather not export these though.</del> -- not the case, see below
* Discovered that I dropped the dry run setting from `NewDownloadCheckQueue` in https://github.com/git-lfs/git-lfs/pull/1746/files#diff-1b73a9b301d9b5f76392f5818dd2ab15L43, it was re-added.
* Dropped the `ConcurrentTransfers()` call too.

### DIFF
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -291,7 +291,7 @@ func fetchAndReportToChan(allpointers []*lfs.WrappedPointer, filter *filepathfil
 	}
 
 	ready, pointers, meter := readyAndMissingPointers(allpointers, filter)
-	q := lfs.NewDownloadQueue(tq.WithProgress(meter))
+	q := newDownloadQueue(tq.WithProgress(meter))
 
 	if out != nil {
 		// If we already have it, or it won't be fetched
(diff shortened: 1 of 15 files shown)


technoweenie:
@ttaylorr I got rid of the `Init*` functions on `*tq.Manifest`, in favor of a `tq.NewManifestFromGitConfig(git tq.Env)` constructor. Ready for review.

sinbad:
Great, isolating the config into a pure data-carrier per package that's populated by the reflection stuff feels way more robust than mixing that in other types.

PR #1770, opened 2016-12-14
Tq/extract endpoint

technoweenie:
Experiment in extracting the endpoint and url alias stuff from `config` to a new `endpoint` package. I'm starting to think it belongs in `api` instead though.


config/config.go
```diff
@@ -392,18 +333,23 @@ func (c *Configuration) Remotes() []string {
...
 
-// GitProtocol returns the protocol for the LFS API when converting from a
-// git:// remote url.
 func (c *Configuration) GitProtocol() string {
-	if value, ok := c.Git.Get("lfs.gitprotocol"); ok {
-		return value
+	return c.endpointConfig().GitProtocol()
+}
+
+func (c *Configuration) endpointConfig() *endpoint.Config {
```
ttaylorr: I really dig this. It's too bad that we have to keep something like it around in `*config.Configuration`, but it's nice that we can track all of the uses of it by checking out the callers of this function.

endpoint/config.go
```diff
@@ -0,0 +1,199 @@
...
+	"path"
+	"strings"
+	"sync"
+
+	"github.com/git-lfs/git-lfs/git"
+)
+
+const defaultRemote = "origin"
+
+type Config struct {
```
ttaylorr: I'm still not sold on having a per-package `*Config` type. I think that splitting out and removing stuff from the `*config.Configuration` type is really good, but I think that we should go one step further and figure out what an appropriate abstraction is.

What do you think about `*endpoint.Finder`?

technoweenie: Yeah, I dig that. Still not sure if this belongs in here or in `api`, though.

PR #1771, opened 2016-12-14
Checkout: Remove use of pointer channel

technoweenie:
Finally removed our first `[]*lfs.WrappedPointer` :metal:

* Removes the callback arg on `(*GitScanner) ScanTree(ref)`. Everything was simply passing `nil` to it. 62d0b51ada64411e0045d49c0865ef3c17dcad2e
* Rewrote `lfs.ConvertCwdFilesRelativeToRepo()` and `lfs.ConvertRepoFilesRelativeToCwd()` from functions working with channels, to a couple structs that implement a `PathConverter` interface. 68efd0536a2eb3e0514ee4d2da7ab9c7a084fcb8
* Extracted a `*gitIndexer` type to handle all the `git update-index` stuff. 5ef6b8a1bda07212b97133f2a7794cd6cbfa8283
* Extract a shared `checkout()` func that the `pull` and `checkout` commands can share. 0d5d391f92d5146ba90f8d7cbcd7852279275232
* Removed a reliance on `[]*lfs.WrappedPointer` in the `checkout` command 08dd0309b2aab31663cb6e62b75828ca622a086f
* Moved existing checkout code that still needs `[]*lfs.WrappedPointer` to the pull cmd. It's the only other spot that uses it now, and is next on my list.


technoweenie:
It was easy enough to convert the `pull` command, so I did that one too. I tried coming up with a unified type for both operations so we could handle the setup and teardown of all the little objects in one spot, but the best I could do was that `*singleCheckout` type. Otherwise, there are just too many subtle differences.

technoweenie:
Pretty sure this will cause some more merge conflicts, so I'm going to hold off on this until the new `tq` package is in master. I have a hunch that PR will have enough conflicts to deal with.

Issue #1764, opened 2016-12-13
Tracking: `tq` package


ttaylorr:
This issue serves as a place to track the remaining 👊-items for merging `tq-master` into `master`.

**Original proposal**: #1651.

## To-do

- [x] #1758: Priority Retries
- [x] #1767: De-couple `config` package, use interface instead.
- [x] #1774 Change `Begin` function to use a `*tq.Manifest`
- [x] #1772: Introduce a `tq.Object` to replace use of `*tq.Transfer`, `tq.Transferable` and `*api.ObjectResource`.
- [ ] Split adapter implementations into sub-package: `tq/transferadapter` (name pending).

---

/cc @git-lfs/core

technoweenie:
I ran into an issue with `Begin()`, the custom adapter can use a different setting if the implementation is not concurrent: https://github.com/git-lfs/git-lfs/blob/39badd3c74bca15afd1e1e4774094f17895088d7/tq/custom.go#L109-L121

One thought is to make it a single struct, and change the [caller](https://github.com/git-lfs/git-lfs/blob/39badd3c74bca15afd1e1e4774094f17895088d7/tq/transfer_queue.go#L532-L539) to pass in a new struct:

```diff
 func (q *TransferQueue) ensureAdapterBegun() error {
	q.adapterInitMutex.Lock()
	defer q.adapterInitMutex.Unlock()

	if q.adapterInProgress {
		return nil
	}

-	// Progress callback - receives byte updates
-	cb := func(name string, total, read int64, current int) error {
-		q.meter.TransferBytes(q.transferKind(), name, read, total, current)
-		return nil
-	}

	tracerx.Printf("tq: starting transfer adapter %q", q.adapter.Name())
-	err := q.adapter.Begin(config.Config.ConcurrentTransfers(), cb)
+	err := q.adapter.Begin(BeginArgs{
+		ConcurrentTransfers: config.Config.ConcurrentTransfers(),
+		ProgressCallback: func(name string, total, read int64, current int) error {
+			q.meter.TransferBytes(q.transferKind(), name, read, total, current)
+			return nil
+		},
+	})
```

This gives us the flexibility to add more to `BeginArgs` in the future. It also gives adapters the flexibility to set their own values without worrying about affecting other adapters. Finally, we could add `ConcurrentTransfers` to `*tq.Manifest` and get rid of `config.Config.ConcurrentTransfers()`.

technoweenie:
[...]

PR #1772, opened 2016-12-14
tq: simplify usage of `tq.Transfer`

ttaylorr:
This pull-request simplifies the `tq.Transfer` type, and removes the usage of the mutable `Transferable` interface. Additionally, it enables us to remove `*api.ObjectResource` from the `tq`'s exported API.

Previously, we used three main types:

1. `Transferable` - an `interface{}` to cover downloadable and uploadable LFS pointers.
2. `*api.ObjectResource` - A type to hold values from the API.
3. `*transfer.Transfer` - A combination of `Transferable()` and `*api.ObjectResource`. Used in adapter implementations to grab values both from the API, and values pertaining to the local repository.

The bummer is that the `Transferable` is cached in a `map[string]Transferable` on the `*TransferQueue` for the _entire transfer_, which can incur some pretty high memory consumption. unsafe.Sizeof suggests that the we're storing much more than [56 bytes][1] per Transfer, not including the size of the inner-structs. My calculations estimate about **~128bytes per object** (for two actions (32 bytes/action), 4 bytes for the pointer), and no error).

To simplify, let's store the API data when we need it, and ignore it when we don't. For a local cache, all we need is the Name, Path, Oid and Size, which compacts to [56 bytes][2] as a maximum (we can remove 16 bytes/transfer by dropping the `Name` field). This is, on average, __68.7% less memory consumption per unique Transfer__.

In addition, the code gets simpler, since we only need to care about the objects retuned back from the API when we're sending them to the adapters.

On top of all of that, we got rid of the mutability in `SetObject()`, narrowed down our internal use of the `*api.ObjectResource` type, and only have _one remaining external use of the `api` package in `tq`._

---

Since this PR is on the larger side, here are the important groups of commits:

1. e18044f: remove `Transferable` interface 
2. 7f26137: simplify `(q *TransferQueue) Add()` signature.
3. 94b2d0d...5fc2344: various 💅 and 🍬 to clean things up

---

/cc @git-lfs/core 

[1]: http://golang-sizeof.tips/?t=Ly8gU2FtcGxlIGNvZGUKc3RydWN0IHsKCU9pZCAgICAgICAgICAgc3RyaW5nCglTaXplICAgICAgICAgIGludDY0CglBdXRoZW50aWNhdGVkIGJvb2wKCUFjdGlvbnMgICAgICAgbWFwW3N0cmluZ10qc3RydWN0ewoJICAgIEhyZWYgc3RyaW5nCgkgICAgSGVhZGVyIG1hcFtzdHJpbmddc3RyaW5nCgkgICAgRXhwaXJlc0F0IHRpbWUuVGltZQoJfQoJTGlua3MgICAgICAgICBtYXBbc3RyaW5nXSpzdHJ1Y3R7CgkgICAgSHJlZiBzdHJpbmcKCSAgICBIZWFkZXIgbWFwW3N0cmluZ11zdHJpbmcKCSAgICBFeHBpcmVzQXQgdGltZS5UaW1lCgl9CglFcnJvciAgICAgICAgICpzdHJ1Y3R7CgkgICAgTWVzc2FnZSBzdHJpbmcKCSAgICBDb2RlIGludAoJfQp9Cg==
[2]: http://golang-sizeof.tips/?t=Ly8gU2FtcGxlIGNvZGUKc3RydWN0IHsKICAgIE5hbWUsIFBhdGgsIE9pZCBzdHJpbmcKICAgIFNpemUgaW50NjQKfQo=


ttaylorr:
Tracking: #1764.

PR #1776, opened 2016-12-15
lfs,cmd: remove lfs.New{Downloadable,Uploadable}

ttaylorr:
This pull-request moves the `lfs.NewDownloadable` and `lfs.NewUploadable` functions into the `commands` package and un-exports them.

As an aside, the `NewUploadable` alternative had to be copied into the `test` package, since it was inaccessible as a private member of the `commands` package. I think this follows the Go-ism: "A little copy is better than a little dependency."

---

/cc @git-lfs/core

### DIFF
--- a/test/git-lfs-test-server-api/main.go
+++ b/test/git-lfs-test-server-api/main.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


ttaylorr:
Tracking: #1764.

test/git-lfs-test-server-api/main.go
```diff
@@ -285,6 +286,68 @@ func interleaveTestData(slice1, slice2 []TestObject) []TestObject {
...
+		Name: filename,
+		Path: localMediaPath,
+		Oid:  oid,
+		Size: fi.Size(),
+	}, nil
+}
+
+// ensureFile makes sure that the cleanPath exists before pushing it.  If it
+// does not exist, it attempts to clean it by reading the file at smudgePath.
+func ensureFile(smudgePath, cleanPath string) error {
```
technoweenie: I think we can just get rid of `ensureFile()` in this test command. It's generating the data on the fly, so I'm not sure how it'd ever get into a weird state.

PR #1777, opened 2016-12-15
tq: un-export `NewTransfer` and `Batch`

ttaylorr:
This pull-request un-exports two identifiers, `tq.NewTransfer` and `tq.Batch`, since they are not needed in the public API.

Since we got un-exported `tq.NewTransfer`, this removes the last public use of the `api` package in `tq`, meaning callers don't need to depend on `api` directly to use `tq`.

---

/cc @git-lfs/core

### DIFF
--- a/tq/transfer.go
+++ b/tq/transfer.go
@@ -32,13 +32,8 @@ type ObjectError struct {
 	Message string `json:"message"`
 }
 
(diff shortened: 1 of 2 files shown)


ttaylorr:
Tracking: #1764/

Issue #1750, opened 2016-12-09
git lfs track is very slow


smithkl42:
Running a simple `git lfs track` command on our repo takes about 30 seconds, and the `git-lfs` process chews up nearly 100% of CPU during that time period. Our repo has about 3000 standard git files, and a hundred or so git-lfs tracked files (for a combined size of about 1 GB).

Not being clear on the internals of git-lfs, I'm not sure what you can do about it, but since SourceTree kicks off a `git lfs track` every time you stage a file, it makes for a pretty miserable experience. And given the output - it doesn't look like it's doing anything more complicated than reading from a text file, though I'm sure the reality is much more complex - it *seems* like this shouldn't take so long.

I should note that I'm running git-lfs 1.5.3.

smithkl42:
BTW, I see that this is largely a duplicate of https://github.com/git-lfs/git-lfs/issues/1611 (which I looked for but didn't see initially). However, since that one is closed, and it still seems to be an issue, I figure it's worth keeping this one open.

ttaylorr:
:wave: Hi @smithkl42, thanks for opening this.

> Our repo has about 3000 standard git files, and a hundred or so git-lfs tracked files (for a combined size of about 1 GB).

This is interesting. We recently [shipped][1] an optimization to the `git lfs track` command that skips walking directories that are `.gitignore`'d, and the `.git` directory itself. Even so, with only 3000 files (and hopefully a proportional amount of Git objects), the naive `filepath.Walk` implementation should be fast enough.

That all being said, I'm not sure what's going on here. If you're able to share the contents of your repository, a link would be immensely helpful :sparkles:. If not, some information about the directory structure should suffice: how many top-level directories are there in your repository, are the sub-directories deeply nested, etc. Just enough information to give us a picture of what your repository looks like.

I'm thinking that what's probably happening here is that we're leaking goroutines. I would need a little more time to look at the fast walk implementation to find the cause, but that's at least one theory on what could be going on here. Another theory: one (or more) of the alive goroutines could be spinning and thrashing the CPU. Going to :pager: @sinbad in on this one, since his help would be very valuable here. 🙇

[1]: https://github.com/git-lfs/git-lfs/pull/1616

smithkl42:
Unfortunately, can't share the repo - sorry. (I know that makes troubleshooting more difficult.)

[...]

PR #1782, opened 2016-12-16
backport filepathfilter to v1.5.x

technoweenie:
This backports the `filepathfilter` package, which should improve `git lfs track` times. /cc #1750

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -8,10 +8,10 @@ import (
 	"sync"
 
 	"github.com/git-lfs/git-lfs/errors"
+	"github.com/git-lfs/git-lfs/filepathfilter"
 	"github.com/git-lfs/git-lfs/git"
 	"github.com/git-lfs/git-lfs/lfs"
 	"github.com/git-lfs/git-lfs/progress"
-	"github.com/git-lfs/git-lfs/tools"
 	"github.com/rubyist/tracerx"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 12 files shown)


ttaylorr:
Looks awesome, and thanks for backporting this 👍 .

It looks like there's a transient failure on the clone test using the `.lfsconfig` file. It appears to be transient because it's passing AppVeyor on the branch build, but failing on the PR build.

I looked into the Git source code, since "Could not checkout" isn't an LFS error. I found two instances of that message in the Git v2.8.2 tree, but neither of them come from the clone command:

```sh
~/g/git (master) $ git checkout -q v2.8.2
~/g/git ((HEAD) $ git describe HEAD
v2.8.2
~/g/git ((HEAD) $ ack "Could not checkout"
git-filter-branch.sh
356:                    die "Could not checkout the index"

git-rebase--interactive.sh
1138:   die "Could not checkout $switch_to"
```

Not sure what's going on here... do you have any thoughts?

technoweenie:
Ah, I've seen intermittent clone failures before. I don't think this PR could affect `lfsconfig`, but this is something that we should investigate. I wonder if we should dump `git lfs logs last` after test failures?

ttaylorr:
> I wonder if we should dump `git lfs logs last` after test failures?

Sounds good to me. We'll have to `set +e` to catch the dirty exit code, but having some debugging commits in here will be really helpful. 👍

ttaylorr:
I went ahead and opened #1801 which runs `git lfs logs last` if that clone command fails again. This should be helpful in debugging what's going on in the future.

For now, I'm going to re-run the tests to get rid of the transient failure, and merge this into the `release-1.5` tree.

PR #1791, opened 2016-12-20
lfsapi: teach `(*Client) Do()` how to handle 307 redirections

technoweenie:


lfsapi/client.go
```diff
@@ -0,0 +1,149 @@
...
+	res, err := c.doWithRedirects(c.httpClient(req.Host), req, nil)
+	if err != nil {
+		return res, err
+	}
+
+	return res, c.handleResponse(res)
+}
+
+func (c *Client) doWithRedirects(cli *http.Client, req *http.Request, via []*http.Request) (*http.Response, error) {
+	if seeker, ok := req.Body.(io.Seeker); ok {
```
ttaylorr: Should this be an error if `len(via) > 0 && !ok`?

lfsapi/client.go
```diff
@@ -0,0 +1,149 @@
...
+		seeker.Seek(0, io.SeekStart)
+	}
+
+	res, err := cli.Do(req)
+	if err != nil {
+		return res, err
+	}
+
+	if res.StatusCode != 307 {
```
ttaylorr: One of our roadmap items is to respond to 301s as well. Do you think we could add an `|| res.StatusCode == 301`, and maintain a cache of permanent redirections? I'm thinking something like a `map[*url.URL]*url.URL` map of from -> to.

lfsapi/client.go
```diff
@@ -0,0 +1,149 @@
...
+	redirectedReq, err := http.NewRequest(req.Method, redirectTo, nil)
+	if err != nil {
+		return res, errors.Wrapf(err, err.Error())
+	}
+
+	redirectedReq.Body = req.Body
+	redirectedReq.ContentLength = req.ContentLength
+
+	if err = checkRedirect(redirectedReq, via); err != nil {
+		return res, errors.Wrapf(err, err.Error())
```
ttaylorr: What does `errors.Wrapf(err, err.Error())` do? I think this would just turn an error "foo" into "foo: foo".

lfsapi/client.go
```diff
@@ -0,0 +1,149 @@
...
+		CheckRedirect: checkRedirect,
+	}
+
+	c.hostClients[host] = httpClient
+
+	return httpClient
+}
+
+func checkRedirect(req *http.Request, via []*http.Request) error {
+	if len(via) >= 3 {
```
ttaylorr: Do you think we should make this a configurable value?

lfsapi/client.go
```diff
@@ -0,0 +1,149 @@
...
+		return errors.New("stopped after 3 redirects")
+	}
+
+	oldest := via[0]
```
ttaylorr: I think this line will cause a panic if `len(via) == 0`.

lfsapi/client.go
```diff
@@ -0,0 +1,149 @@
...
+	for key := range oldest.Header {
+		if key == "Authorization" {
+			if req.URL.Scheme != oldest.URL.Scheme || req.URL.Host != oldest.URL.Host {
+				continue
+			}
+		}
+		req.Header.Set(key, oldest.Header.Get(key))
```
ttaylorr: I'm not sure this behavior fits in a `check<T>` function. This seems side-effect-y, so I think we should either document it's behavior, or extract it out to another function.

lfsapi/client_test.go
```diff
@@ -1,12 +1,92 @@
...
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+type redirectTest struct {
+	Test string
+}
+
+func TestClientRedirect(t *testing.T) {
```
ttaylorr: Very comprehensive 👍

lfsapi/client.go
technoweenie: Yes, but not as part of this refactoring effort.

lfsapi/client.go
technoweenie: You're right, it's just a relic of our ancient error handling.

technoweenie:
I did clean up the redundant error wrapping, and removed `checkRedirect()`.

ttaylorr:
> I did clean up the redundant error wrapping, and removed `checkRedirect()`.

Thanks so much! ✨

PR #1794, opened 2016-12-21
Api/tracing

technoweenie:
This attempts to port over the tracing code from `httputil`. The main change is relying no tracing/debugging flags on `*lfsapi.Client` instead of a global `*config.Configuration`.

TODO

* [x] Tests
* [x] HTTP Request Stats


technoweenie:
Ready for review: This PR ports the tracing capabilities from [the `httputil` package][1].

[1]: https://github.com/git-lfs/git-lfs/blob/49b62826aa44eccf1b10e23fe8787e7e93ee64bb/httputil/http.go#L61-L106

It's duties, at a high level, are:

```go
func (c *HttpClient) Do(req *http.Request) (*http.Response, error) {
	// trace request details for GIT_CURL_VERBOSE
	traceHttpRequest(c.Config, req)

	// wrap request body for GIT_CURL_VERBOSE
	crc := countingRequest(c.Config, req)
	if req.Body != nil {
		// Only set the body if we have a body, but create the countingRequest
		// anyway to make using zeroed stats easier.
		req.Body = crc
	}

	// track request start time for GIT_LOG_STATS
	start := time.Now()
	
	// make the request
	res, err := c.Client.Do(req)
	if err != nil {
		return res, err
	}

	// trace response details for GIT_CURL_VERBOSE
	traceHttpResponse(c.Config, res)

	// wrap response body for GIT_CURL_VERBOSE and GIT_LOG_STATS
	cresp := countingResponse(c.Config, res)
	res.Body = cresp

	// track http response details for GIT_LOG_STATS
	if c.Config.IsLoggingStats {
		// snip
	}

	return res, err
}
```

Note that there are three different flags for tracing and debugging:

1. `GIT_TRACE` - Handled invisibly through `tracerx.Printf()`
1. `GIT_CURL_VERBOSE` - Normal HTTP tracing, with obscured `Authorization` Header values
2. `LFS_DEBUG_HTTP` - Turns on full `Authorization` header when tracing HTTP dumps
3. `GIT_LOG_STATS` - Dumps details http stats to disk when the command exits

I'm trying to keep the code structure similar, but am interested in other ideas for a future PR. Also, it's impossible to test `GIT_TRACE` without major changes to the tracerx package :(

PR #1799, opened 2016-12-23
Api/fix auth access

technoweenie:
This fixes some issues with the new API auth code:

1. Clarify and describe in comments the difference between a request URL, api endpoint URL, and the git remote URL. Using the credsURL for check and set the lfs endpoint access was causing endless loops.
2. Strip URL auth when building the `lfs.URL.access` git config key.
3. Better tests.

### DIFF
--- a/lfsapi/auth_test.go
+++ b/lfsapi/auth_test.go
@@ -297,6 +297,20 @@ func TestGetCreds(t *testing.T) {
 				Authorization: "custom",
 			},
 		},
+		"netrc": getCredsTest{
+			Remote: "origin",
+			Method: "GET",
+			Href:   "https://netrc-host.com/repo/lfs/locks",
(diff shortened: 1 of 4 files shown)


lfsapi/auth_test.go
```diff
@@ -223,211 +222,272 @@ func credsToKey(creds Creds) string {
...
+	Header   map[string]string
+	Config   map[string]string
+	Expected getCredsExpected
 }
 
-func TestGetCredentials(t *testing.T) {
-	checks := []*getCredentialCheck{
-		{
-			Desc: "simple",
+func TestGetCreds(t *testing.T) {
```
ttaylorr: ✨ awesome tests.

abirfo1 (commented):
3854677

PR #1801, opened 2016-12-26
clone: debug transient "clone" failure

ttaylorr:
See: https://github.com/git-lfs/git-lfs/pull/1782#issuecomment-267678319.

---


ttaylorr:
(Acking out of our internal first-responder queue.)

technoweenie:
I think this can just go into master and release-1.5.

Issue #1803, opened 2016-12-26
Specific sequence of commands leads to git-lfs error (backtrace) on one machine, works on other


JonasT:
When running the following sequence on my server automated by a python script, I get a git-lfs crash with backtrace:
```
git init .
git lfs install --skip-smudge
git remote
git remote add source https://gitlfsiauthenticatednotsupported:noneneeded@gitlab.wobble.ninja/JonasT/ChiCraft.git
git remote update
git remote
git remote add target https://JonasT:<RETRACTED>@github.com/ChiCraft/ChiCraft.git
git remote update
git fetch source
git checkout master
git-lfs fetch --all
```

This leads to the following error:
```
Scanning for all objects ever referenced...
panic: runtime error: makeslice: len out of range

goroutine 1 [running]:
panic(0x7a0540, 0xc4200143e0)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
strings.Repeat(0x800fa8, 0x1, 0xffffffffffffffef, 0x2, 0x2)
	/usr/local/go/src/strings/strings.go:420 +0x49
github.com/git-lfs/git-lfs/progress.(*Spinner).update(0xc420045a40, 0xbbfda0, 0xc4200d2a00, 0xc420014398, 0x1, 0xc4200143a0, 0xf)
	/tmp/docker_run/src/github.com/git-lfs/git-lfs/amd64/obj-x86_64-linux-gnu/src/github.com/git-lfs/git-lfs/progress/spinner.go:57 +0x1b0
github.com/git-lfs/git-lfs/progress.(*Spinner).Spin(0xc420045a40, 0xbbfda0, 0xc4200d2a00)
	/tmp/docker_run/src/github.com/git-lfs/git-lfs/amd64/obj-x86_64-linux-gnu/src/github.com/git-lfs/git-lfs/progress/spinner.go:29 +0xc8
github.com/git-lfs/git-lfs/progress.(*Spinner).Print(0xc420045a40, 0xbbfda0, 0xc4200d2a00, 0xc4200143a0, 0xf)
	/tmp/docker_run/src/github.com/git-lfs/git-lfs/amd64/obj-x86_64-linux-gnu/src/github.com/git-lfs/git-lfs/progress/spinner.go:23 +0x5f
github.com/git-lfs/git-lfs/commands.scanAll(0x10, 0x0, 0x0)
[...]

PR #1806, opened 2016-12-27
progress: guard negative padding width, panic in `strings.Repeat`

ttaylorr:
This pull-request prevents the `len(padding)` used in the progress meter from ever being less than zero. This prevents the panic from `strings.Repeat` trying to [allocate a slice][1] with a [`len < 0`][2] as seen in #1803.

Previously, LFS printed a number of number of padding characters to the terminal when displaying progress meter updates, so that the terminal's cursor would be in the correct position to fill the entire line after printing the CR character, `\r`, as such:

```
       |-----------------------padding---------------------|
message                                                    
                                                           ^- cursor position
```

However, if the message's length is greater than the terminal width, `width-len(str)` is a negative number, causing the panic as shown above. This check with `tools.MaxInt(0, width-len(str))` prevents that from happening by forcing the padding to be an empty string.

This is a safe change, since we'll be outside of the terminal's assumed width, leaving the cursor in an OK position to rewrite the entire line. This would leave a remaining number of characters trailing off the line, but this only happens when the `ts` library can't accurately determine the tty's width, and we default back to 80. 

---

/cc @git-lfs/core #1803

[1]: https://github.com/golang/go/blob/go1.7.3/src/strings/strings.go#L420
[2]: https://github.com/golang/go/blob/go1.7.3/src/runtime/slice.go#L47-L50

### DIFF
--- a/progress/meter.go
+++ b/progress/meter.go
@@ -189,12 +189,6 @@ func (p *ProgressMeter) update() {
 		return
 	}
 
-	width := 80 // default to 80 chars wide if ts.GetSize() fails
-	size, err := ts.GetSize()
-	if err == nil {
-		width = size.Col()
-	}
-
 	// (%d of %d files, %d skipped) %f B / %f B, %f B skipped
 	// skipped counts only show when > 0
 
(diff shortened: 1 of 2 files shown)


ttaylorr:
This check against negative padding widths already existed in [`func (*meter.ProgressMeter) update()`][1], so I extracted it out to a function and gave it a name: `pad()`. (See: b1e6825).

I chose to include the full contents of the line instead of truncating it to fit inside the terminal to allow for corner cases caused when resizing the terminal.

[1]: https://github.com/git-lfs/git-lfs/blob/39cbf99ede45dec4d241d423a64dd5b8211a20ba/progress/meter.go#L211-L213

PR #1812, opened 2016-12-29
commands: convert push, pre-push to use async gitscanner

ttaylorr:
This pull-request teaches the `push` and `pre-push` commands how to use the new `gitscanner` without accumulating potentially large slices of `*lfs.WrappedPointers`.

Previously, in order to push a set of LFS objects to a remote, the pusher needed to have explicit knowledge of all objects they were going to push. In many cases, this isn't a problem, since usually the number of objects is manageable. During initial or large pushes, however, the number of objects to push can grow prohibitively large, causing a problem.

This pull-request applies the work that @technoweenie has done with the `gitscanner` type and applies it to the `push` and `pre-push` commands. Here's a breakdown of what happened:

1. 4d4f808: since the commands will now call `uploadPointers()` more than once, re-use the `*tq.TransferQueue` instance across invocations to prevent needless re-allocation and to take full advantage of batching.
2. 64a643e: since the `uploadPointers()` calls can place objects across multiple batches and thus do not guarantee that all objects in the call have been uploaded, introduce `Await()` to ensure that the `*tq.TransferQueue` is done before exiting LFS.
3. [`a58e91f^1...f1380bd`][1]: teach calling code how to asynchronously call `uploadPointers`.
4. [`028facf^1...430f5a7`][2]: :nail_care:
5. 7af9281: fix small bug in `lfs.NewDownloadCheckQueue` which created `n` empty elements in the options array (where `n` is equal to the length of options provided)
6. 15673a4: since the download check queue is used many times instead of all at once, do some preparation work to make it reusable. In this commit, provide a mechanism for callers to find out about failed transfers so that calling code can accurately determine when all given items have been checked.
7. b169435: Wire-up the download check queue to be reused in `*commands.uploadContext`.

[1]: https://github.com/git-lfs/git-lfs/compare/a58e91f%5E1...f1380bd
[2]: https://github.com/git-lfs/git-lfs/compare/028facf%5E1...430f5a7

---

One thing I'd like to investigate before merging is the use of the download check queue in b169435. Prior to that commit, the queue was re-instantiated each time the scanner reported items. This would be OK if the scanner reported items in small batches, since we'd be able to take advantage of the batching functionality of the `*tq.TransferQueue`, however it does not. What this means is that we have to perform an expensive object allocation (not to mention all of the internal `chan`s, `map`s, etc.) _each time a pointer is read from the gitscanner_.

To mitigate this, I did work in [`15673a4^1...b169435`][3] to avoid this expensive and repeated allocation. However, as a consequence, the download check queue needs to report synchronously whether or not the server has items that are missing locally in `.git/lfs/objects`. What this means is a batch size of 1, so that we're not waiting indefinitely to hear of the status a set of items with cardinality less than the batch size.

[3]: https://github.com/git-lfs/git-lfs/compare/15673a4%5E1...b169435

This may be fine, since I think it's a safe assumption to assume the number of items that needs to go through this check queue `(i.e., the number of objects missing from `.git/lfs/objects` needed to push) is sufficiently small. If not, it may be worth pursuing alternatives that remove the need for single-length batches.

One option we could pursue is using a second transfer queue to keep track of the items that are not present in `.git/lfs/objects`, or using the `*lfs.Batcher` type to batch it ourselves. I'm not sure if this is advantageous or not, since it does carry with it additional complexity. I'd love to hear additional thoughts on this.

---

/cc @git-lfs/core

### DIFF
--- a/commands/command_pre_push.go
+++ b/commands/command_pre_push.go
@@ -76,7 +76,7 @@ func prePushCommand(cmd *cobra.Command, args []string) {
 			continue
 		}
 
-		if err := scanLeftOrAll(gitscanner, ctx, left); err != nil {
+		if err := uploadLeftOrAll(gitscanner, ctx, left); err != nil {
 			Print("Error scanning for Git LFS files in %q", left)
 			ExitWithError(err)
 		}
--- a/commands/command_pre_push.go
+++ b/commands/command_pre_push.go
@@ -84,6 +84,8 @@ func prePushCommand(cmd *cobra.Command, args []string) {
 		}
 		uploadPointers(ctx, pointers)
 	}
+
+	ctx.Await()
 }
 
 func scanLeft(g *lfs.GitScanner, ref string) ([]*lfs.WrappedPointer, error) {
--- a/commands/command_pre_push.go
+++ b/commands/command_pre_push.go
@@ -2,7 +2,6 @@ package commands
 
 import (
 	"bufio"
-	"fmt"
 	"os"
 	"strings"
 
@@ -84,29 +83,6 @@ func prePushCommand(cmd *cobra.Command, args []string) {
 	}
 
 	ctx.Await()
-}
-
-func scanLeft(g *lfs.GitScanner, ref string) ([]*lfs.WrappedPointer, error) {
-	var pointers []*lfs.WrappedPointer
-	var multiErr error
-	cb := func(p *lfs.WrappedPointer, err error) {
-		if err != nil {
-			if multiErr != nil {
-				multiErr = fmt.Errorf("%v\n%v", multiErr, err)
-			} else {
-				multiErr = err
-			}
-			return
-		}
-
-		pointers = append(pointers, p)
-	}
-
-	if err := g.ScanLeftToRemote(ref, cb); err != nil {
-		return pointers, err
-	}
-
-	return pointers, multiErr
 }
 
 // decodeRefs pulls the sha1s out of the line read from the pre-push
--- a/commands/command_pre_push.go
+++ b/commands/command_pre_push.go
@@ -77,12 +77,10 @@ func prePushCommand(cmd *cobra.Command, args []string) {
 			continue
 		}
 
-		pointers, err := scanLeftOrAll(gitscanner, left)
-		if err != nil {
+		if err := scanLeftOrAll(gitscanner, ctx, left); err != nil {
 			Print("Error scanning for Git LFS files in %q", left)
 			ExitWithError(err)
 		}
-		uploadPointers(ctx, pointers...)
 	}
 
 	ctx.Await()
--- a/commands/command_pre_push.go
+++ b/commands/command_pre_push.go
@@ -82,7 +82,7 @@ func prePushCommand(cmd *cobra.Command, args []string) {
 			Print("Error scanning for Git LFS files in %q", left)
 			ExitWithError(err)
 		}
-		uploadPointers(ctx, pointers)
+		uploadPointers(ctx, pointers...)
 	}
 
 	ctx.Await()
--- a/commands/command_push.go
+++ b/commands/command_push.go
@@ -74,11 +74,11 @@ func scanLeftOrAll(g *lfs.GitScanner, ref string) ([]*lfs.WrappedPointer, error)
 }
 
 func uploadsWithObjectIDs(ctx *uploadContext, oids []string) {
-	pointers := make([]*lfs.WrappedPointer, len(oids))
-	for idx, oid := range oids {
-		pointers[idx] = &lfs.WrappedPointer{Pointer: &lfs.Pointer{Oid: oid}}
+	for _, oid := range oids {
+		uploadPointers(ctx, &lfs.WrappedPointer{
+			Pointer: &lfs.Pointer{Oid: oid},
+		})
 	}
-	uploadPointers(ctx, pointers...)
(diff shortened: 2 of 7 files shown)


technoweenie:
Hmm, I don't like this approach. We should definitely not be making download check calls in batches of 1. Maybe we need to stake a step back and consider why there is a download check queue at all? How can we make it a single queue?

I have a hunch that the transfer queue checks to see if the object lives in `.git/lfs/objects` before pushing. The download queue lets us filter objects on the server out early. Why not add this check to the upload transfer adapters instead?

ttaylorr:
> We should definitely not be making download check calls in batches of 1.

I definitely agree 👍 Taking another look at this PR, I think that I can do something cleaner than having the `Notify(fn func(oid string, ok bool))` and a batch size of 1.

> I have a hunch that the transfer queue checks to see if the object lives in `.git/lfs/objects` before pushing.

I'm not sure that this is the case. When I read the code, I thought this was checking for objects that the gitscanner thought were needed in the push, but were _not_ present in `.git/lfs/objects`. When the `*commands.uploadContext` prepares an unfiltered set of `*lfs.WrappedPointer`s for upload, it checks all of them to see [whether `lfs.ObjectExistsOfSize` is true for that pointer][1]. If the object does exist, it is added to the `uploadables` slice and then subsequently, the `*tq.TransferQueue`. 

Otherwise, it's added to the `missingLocalObjects` set and then [`checkMissing` is called on that set][2]. That function adds it to the download check queue, which will tell us whether or not the server has that object (i.e., if it's OK to push even though the object is missing locally). Otherwise, if the server does not have that object present, it'll be added to the main `*tq.TransferQueue`, which will propagate an error back up when it tries to upload an object it does not have the data for.

Some of the solutions I proposed in the original PR I think are worth considering:

> One option we could pursue is using a second transfer queue to keep track of the items that are not present in `.git/lfs/objects`, or using the `*lfs.Batcher` type to batch it ourselves.

I originally thought that this problem would have a large enough effect to warrant considering these alternatives seriously. However, if a user tries to push a LFS repository with many objects and they completed their initial `clone`/`fetch` with an `-X` flag touching many objects, this could be a significant problem.

> Why not add this check to the upload transfer adapters instead?

I think this is an interesting approach. I like that it allows us to use a single transfer queue and greatly simplifies the implementation of the `uploadContext`. However, I'm not sure if this brings us as much benefit as we'd like in terms of network/API calls. I'm assuming that we'd do a check something like:

```go
func (a *basicUploadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb ProgressCallback, authOkFunc func()) error {
	rel, err := t.Actions.Get("upload")
	if err != nil {
		return err
		// return fmt.Errorf("No upload action for this object.")
	}

        if !lfs.ObjectExistsOfSize(t.Oid, t.Size) {
                if serverHasObject(t.Oid, t.Size) {
                        // OK, bail out early.
                }

                // Otherwise, we can't upload, and return a failure.
        }

        req, err := httputil.NewHttpRequest("PUT", rel.Href, rel.Header)
        if err != nil {
                return err
        }

        // ...
}
```

This seems clean to me, but we'd still be making one network call per each object missing locally.

[1]: https://github.com/git-lfs/git-lfs/blob/b169435f4dbb6a67c1db14d0e0a08b4395e1c864/commands/uploader.go#L93-L100
[2]: https://github.com/git-lfs/git-lfs/blob/b169435f4dbb6a67c1db14d0e0a08b4395e1c864/commands/uploader.go#L103-L104

sinbad:
Your reasoning about why it checks for missing objects locally is correct; this increasingly happens when you have multiple branches with some shared ancestry but where you've never actually fetched for some of the intermediate steps (which is fine if someone else pushed those objects). 

I would advocate for putting as few bits of duplicated logic in the transfer adapters as possible. Dealing with the situation around missing objects locally but present on the server is identical regardless of what transfer adapter you're using, so IMO should not need to be implemented in every adapter. The adapter should only be called when you know you want to upload something. Remember that custom adapters are implemented outside the team, and therefore a clean line of responsibility between lfs and the transfer adapter is very important; you can't refactor a responsibility like this to an adapter without forcing all other adapters to duplicate the logic. That's a bad move IMO.

ttaylorr:
> That's a bad move IMO.

I agree. @technoweenie reminded me yesterday in a video call that we actually don't need this check at all. If the server already has an object, it just wont send that object's OID (and thus, any upload action) in the batch response.

sinbad:
> that we actually don't need this check at all. If the server already has an object, it just wont send that object's OID (and thus, any upload action) in the batch response.

That's true as well, but the pre-check was necessary (at least it used to be) because the pre-push would fail before even calling the API if it thought that the OID was missing locally, since it checked that while collecting the list of things to upload. I guess if the sequence is different this could be eliminated:

1. Collect list of OIDs to push but don't check if they're present locally
2. Call API - this strips out those the server already has
3. Only upload missing, which now must fail if not local because these are now hard requirements

ttaylorr:
After some pairing with @technoweenie, I went ahead and removed the download check queue entirely (338ab40) ✨ . This change essentially ignores all pointer-related errors before the object is in the `*tq.TransferQueue`, and moves the `lfs.ObjectExistsOfSize` functionality there instead, no adapter changes required.

This works well, since the `*tq.TransferQueue` only performs this check if an object has an upload action associated with it (i.e., is needed for the push to complete).

PR #1814, opened 2016-12-30
locking: add `--json` flag

ttaylorr:
This pull-request backports an internal change made to the various locking commands adding the `--json` flag.

To allow 3rd party tools written in languages that do not interoperate with Go to integrate with LFS , a `--json` flag is used to share a standard encoding format. This allows other programs to shell out to LFS and expect an exact data format in return over stdout.

As a caveat, the `Error()`'d messages are not encoded in JSON. This behavior should be OK, since programs can expect that JSON parse errors should be treated as LFS errors and can treat the contents of stdout as the error message.

---

While working on this PR, I noticed that none of the three locking commands (`locks`, `lock`, and `unlock`) have manpage documentation. Before removing the `GITLFSLOCKSENABLED` environment flag, manpages should be written, and they should include notes about the `--json` flag.

---

/cc @git-lfs/core

### DIFF
--- a/commands/command_lock.go
+++ b/commands/command_lock.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -40,6 +41,13 @@ func lockCommand(cmd *cobra.Command, args []string) {
 	lock, err := lockClient.LockFile(path)
 	if err != nil {
 		Exit("Lock failed: %v", err)
+	}
+
+	if locksCmdFlags.JSON {
+		if err := json.NewEncoder(os.Stdout).Encode(lock); err != nil {
+			Error(err.Error())
+		}
+		return
 	}
 
 	Print("\n'%s' was locked (%s)", args[0], lock.Id)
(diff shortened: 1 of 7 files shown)


sinbad:
I support offering better support for automated consumption of output, but this raises some further issues:

1. This should be supported consistently in all other commands which have significant output like`track`, `prune`, `fsck`
1. To be consistent with `git`, we should probably offer a `--porcelain` option if we're doing this, i.e. basic linewise with formatting options. JSON is rich but has its own parsing overhead/complexity - this is not an issue if your language supports it easily, but a plainer format is more universal e.g. consumption from simple scripts

I feel that if we're going to offer machine-parseable output it should be a full commitment. I'm not against offering `--json` as well for convenience of environments that offer easy support for it, but `--porcelain` offers more global support and keeps us in line with `git` principles (even if that means we also have to support `--format` for cases like this where there are many fields).

shana:
:heart_eyes: on having a `--json` flag (and would :heart: having it everywhere). Same for `--porcelain/--format`, really, they serve the same purpose for different use cases (apps shelling out vs shell scripting support)

ttaylorr:
> I agree that `--json` support should be more universal, and that `--porcelain/--format` should be supported as well. I think this PR is good to go though.

Wholeheartedly agree 👍. Thanks all for your input.

PR #1815, opened 2016-12-30
locking: teach pre-push hook to check for locks

ttaylorr:
This pull-request teaches the `pre-push` hook to warn when you are pushing files that you own the lock on, and to error (preventing the push) when you are pushing files that you don't own the lock of.

Here's a breakdown of what happened:

1. 18faf7a...8eb3659: teach the `pre-push` hook to search for locks in range
2. b3707a6...18ec881: implement and use `findLocks` helper to make searching for locks more concise
3. aa21a3d...518aeb8: integration tests for pushing owned and unowned locks
4. f1b2f98: Preemptively use the `--no-verify` flag to avoid calling the `pre-commit` flag introduced in #1816.

Some things that I'd like to take a look at before merging: 

- Should the `pre-push` hook be configurable where to search for locks? It seems that callers should be able to decide between searching the local lock cache, or searching the remote's locks. A flag seems like the wrong approach, since you'd have to update `.git/hooks/pre-push` each time, but a config option seems reasonable.
- Should the `pre-push` hook warn about your own locks? Should this be configurable?
- Should the `commands` package own a singleton, private, instance of a `*locking.Client`?

---

/cc @git-lfs/core


commands/command_pre_push.go
```diff
@@ -83,8 +100,34 @@ func prePushCommand(cmd *cobra.Command, args []string) {
...
+				Error("* %s", file)
+			}
+			os.Exit(1)
+		}
+
 		uploadPointers(ctx, pointers)
 	}
+
+	if len(myLocks) > 0 {
+		Print("Pushing your locked files:")
```
sinbad: I think it would be good to explain more what this means in practice - i.e. that the user should consider unlocking the files if they have no further changes to publish after this push.

commands/command_pre_push.go
```diff
@@ -83,8 +100,34 @@ func prePushCommand(cmd *cobra.Command, args []string) {
...
+				if l.Name == name && l.Email == email {
+					myLocks = append(myLocks, l.Path)
+				} else {
+					lockConflicts = append(lockConflicts, p.Name)
+				}
+			}
+		}
+
+		if len(lockConflicts) > 0 {
+			Error("Some files are locked in %s...%s", left, cfg.CurrentRemote)
```
sinbad: Maybe we could be explicit like "You cannot push changes because someone else has these files locked"

commands/commands.go
```diff
@@ -346,6 +347,24 @@ func buildProgressMeter(dryRun bool) *progress.ProgressMeter {
...
 }
 
+// findLocks finds matching locks using the given "lc" *locking.Client, and all
+// other parameters for the lookup. If an error was encountered, it will be
+// returned immediately, otherise a map of lock.Path -> lock will be returned
+// instead.
+func findLocks(lc *locking.Client, filter map[string]string, limit int, localOnly bool) (map[string]locking.Lock, error) {
+	locks, err := lc.SearchLocks(filter, limit, localOnly)
+	if err != nil {
+		return nil, errors.Wrap(err, "error finding locks")
```
ttaylorr: My bad! Went ahead and cleaned this up in c0047f6.

commands/commands.go
```diff
@@ -346,6 +347,24 @@ func buildProgressMeter(dryRun bool) *progress.ProgressMeter {
 	)
 }
 
+// findLocks finds matching locks using the given "lc" *locking.Client, and all
+// other parameters for the lookup. If an error was encountered, it will be
+// returned immediately, otherise a map of lock.Path -> lock will be returned
+// instead.
+func findLocks(lc *locking.Client, filter map[string]string, limit int, localOnly bool) (map[string]locking.Lock, error) {
+	locks, err := lc.SearchLocks(filter, limit, localOnly)
```
ttaylorr: Thanks for brining this up 😄 . I went ahead and changed the implementation of the `Search()` method on the `*locking.Client` to only try and deserialize the response if it was a `200 OK`. If the response was a 404 or 410, the lack of deserialization would return an empty slice of locks, making the lock-related parts of the pre-push command effectively a no-op.

I don't think any changes are needed for the `Lock()` and `Unlock()` methods, since those are mutative operations and _should_ error out if the server doesn't implement them.

Changes are in: 6fb728a.

commands/command_pre_push.go
ttaylorr: Good call! Added that in f82c96d.

commands/command_pre_push.go
ttaylorr: Sounds good: c9aa19a.

ttaylorr:
OK, everything should be up-to-date and work with the latest changes from #1812 and #1863. I opted to put this implementation in the `*commands.uploadContext`, since it's an easy 🎯, but this should definitely receive some more attention in the future if/when we decide to get rid of `*uploadContext`.

commands/uploader.go
```diff
@@ -66,11 +75,42 @@ func (c *uploadContext) prepareUpload(unfiltered ...*lfs.WrappedPointer) (*tq.Tr
...
-		// we will call Skip() based on the results of the download check queue.
-		c.meter.Add(p.Size)
+		// canUpload determines whether the current pointer "p" can be
+		// uploaded through the TransferQueue below. It is set to false
+		// only when the file is locked by someone other than the
+		// current committer.
+		var canUpload bool = true
+
+		lockQuery := map[string]string{"path": p.Name}
+		locks, err := c.lockClient.SearchLocks(lockQuery, 1, false)
```
technoweenie: Making an API call per file seems like a lot of overhead. I think our early attempts queried all the locks up front, and checked each file locally. That does mean that we'll have to maintain a potentially huge list of locks in memory locally, so we may want to look at an improved interface for verifying groups of files in a single POST request.

ttaylorr: Really good call; I implemented this change in 25bd4a1.

ttaylorr:
Some things I'd like to follow up on, post-merge:

- Teach `pre-push` to be able to configure whether to use the local lock cache exclusively.
- Teach `pre-push` to stop pushing files if encountered locked file.

PR #1816, opened 2016-12-30
locking: add pre-commit hook to detect inappropriately modified files

ttaylorr:
This pull-request adds a `pre-commit` hook that detects and prevents commits changing locked files owned by someone other than the current committer.

This hook looks at the currently staged files when it is called (before the user is told to edit the `.git/COMMIT_EDITMSG`) and inspects each one of them to see if they are locked by someone other than the current committer. The current committer is used since it is the latest and most accurate information determinable about who will commit the file.

Here's a breakdown of the changes:

1. 4f40215...19eaad8: implement the `pre-commit` hook
2. c0843c7: install the `pre-commit` hook
3. 479f031...bf24886: add integration tests
4. 2958106: add a man-page entry for the `pre-commit` command

Some things that we should investigate before merging:

- How does this hook behave when changing history, i.e., with `filter-branch`, `rebase` or `commit --amend`?
- Should we stream the results of `git.StagedFiles()`, to better handle a case where there are _many_ staged files?

---

/cc @git-lfs/core

### DIFF
--- a/commands/command_pre_commit.go
+++ b/commands/command_pre_commit.go
@@ -20,7 +20,8 @@ func preCommitCommand(cmd *cobra.Command, args []string) {
 
 	lockSet, err := findLocks(lc, nil, 0, false)
 	if err != nil {
-		Exit("error finding locks: %s", err)
+		Print("could not find locks, proceeding anyway: %s", err)
+		os.Exit(0)
 	}
 
 	files, err := git.StagedFiles()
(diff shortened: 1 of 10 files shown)


technoweenie:
I think that's a great point. Let's hold off on this and see if it's actually needed in real use.

ttaylorr:
> I think that's a great point. Let's hold off on this and see if it's actually needed in real use.

I agree, let's close for now 👍

PR #1817, opened 2016-12-30
locking: add post-merge hook to unlock locks

ttaylorr:
This pull-request implements a post-merge hook to unlock all locked (owned) files merged back into the `master` branch of a repository.

Here's a breakdown of what happened:

1. 1598acb...111046d: implement the `post-merge` hook
2. decf27e: Install the `post-merge` hook
3. 63b5143...77e2311: add integration tests covering this behavior
4. 0fbfbba...213577e: fix `git-lfs update` command to use generic message
5. cd796b6: add a man-page entry for post-merge

Some things I'd like to consider before merging:

- Should there be a configurable option to specify a default branch (or set of branches) to check, instead of just `master`?
- ~~What should happen if the merge is done using the `--no-verify` option, skipping the `pre-commit` hook and merging a file locked by someone other than the current committer?~~

---

/cc @git-lfs/core

### DIFF
--- a/commands/command_post_merge.go
+++ b/commands/command_post_merge.go
@@ -1,0 +1,61 @@
+package commands
+
+import (
+	"strings"
+
+	"github.com/git-lfs/git-lfs/lfs"
+	"github.com/git-lfs/git-lfs/locking"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+func postMergeCommand(cmd *cobra.Command, args []string) {
+	gitscanner := lfs.NewGitScanner(nil)
+	if err := gitscanner.RemoteForPush("master@{1}"); err != nil {
+		ExitWithError(err)
+	}
+
+	defer gitscanner.Close()
+
+	pointers, err := scanLeftOrAll(gitscanner, "master")
+	if err != nil {
+		Exit("Error scanning Git LFS files: %+v", err)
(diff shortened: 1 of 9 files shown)


technoweenie:
I agree that we should hold off on automatic things like this until we're sure it's something users really want. I don't agree that only servers should handle auto-locking. We could add the ability to install and uninstall optional hooks like `post-merge` or `pre-commit`. But I'd like to be more sure that this is what users want before committing to it in git-lfs :)

sinbad:
> I don't agree that only servers should handle auto-locking.

I'm not suggesting that, only that *unlocking* on local merge will fundamentally break the premise that locking deters parallel edits, because it will make other users believe that it's ok to lock the file now when in fact it isn't (doing so would create a parallel edit because changes are only local to the user who just auto-unlocked)

technoweenie:
Oh I see what you're saying. We slipped this in for a low-effort demo of auto-unlocking, instead of implementing it on the server. I agree this is the wrong way to implement on the client, if that's what we want. Probably a better way is to look for locked files when pushing to `master` (or handling this on the server).

sinbad:
Yep, `post-push` would be the time to do it, if there was a hook for it :/

ttaylorr:
I'm in agreement, and I think we can close this for now 👍

Issue #865, opened 2015-11-26
Tracking with Cygwin


whitneyit:
I just tried setting up `git-lfs` in Cywin.

``` sh
$ git lfs install
Updated pre-push hook.
Git LFS initialized.
```

That seemed to work well. However, I ran into the following error.

``` sh
$ git lfs track "*.psd"
*.psd is outside repository
```

And then when I tried to push, it errord and produced the following log

> remote: error: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.
> remote: error: Trace: 9b4ba7dc5ebae58e52f4b6d53b4f6b3a
> remote: error: See http://git.io/iEPt8g for more information.

My machine

$  uname -a
CYGWIN_NT-6.1 whitneyit 2.3.1(0.291/5/3) 2015-11-14 12:44 x86_64 Cygwin

whitneyit:
I then tried the same in `CMD` and the install _worked_ but it produced the same error

whitneyit:
I just reset the commit and tried again and it worked \_(ツ)_/
> Git LFS: (1 of 6 files) 40.07 MB / 418.56 MB

strebl:
I have the same problem. What did you do exactly?

strebl:
If I run `git lfs env` I get following output:

```
git-lfs/1.1.0 (GitHub; windows 386; go 1.5.1; git 258acf1)
git version 2.6.2

Endpoint=https://bitbucket.org/gibb/viewclients.git/info/lfs (auth=none)
  SSH=git@bitbucket.org:gibb/viewclients.git
Endpoint (production)=https://spieu0311/~/00_private/viewclients_hub.git/info/lfs (auth=none)
  SSH=vu2004@spieu0311:~/00_private/viewclients_hub.git
LocalWorkingDir=C:\cygdrive\c\Users\strebel\Dropbox\gibb\viewclients
LocalGitDir=C:\Users\strebel\Dropbox\gibb\viewclients\.git
LocalGitStorageDir=C:\Users\strebel\Dropbox\gibb\viewclients\.git
LocalMediaDir=C:\Users\strebel\Dropbox\gibb\viewclients\.git\lfs\objects
TempDir=C:\Users\strebel\Dropbox\gibb\viewclients\.git\lfs\tmp
ConcurrentTransfers=3
BatchTransfer=true
git config filter.lfs.smudge = "git-lfs smudge %f"
[...]

PR #1820, opened 2016-12-31
Fixed file paths on cygwin.

creste:
Fixes git-lfs/git-lfs#865.  These are the changes that @marcrossinyol created.  Tested against v1.5.4 on cygwin.  Can't test against master because master doesn't build on cygwin:

```
$ ./script/bootstrap
Installing goversioninfo to embed resources into Windows executables...
Creating the resource.syso version information file...
git-lfs.go:1: running "goversioninfo": exec: "goversioninfo": executable file not found in %PATH%
```

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -576,12 +576,10 @@ func GetCommitSummary(commit string) (*CommitSummary, error) {
 func isCygwin() bool {
 	cmd := subprocess.ExecCommand("uname")
 	out, err := cmd.Output()
-	output := string(out)
-	if (err == nil) && (strings.Contains(output, "CYGWIN")) {
-		return true
-	} else {
+	if err != nil {
 		return false
 	}
+	return bytes.Contains(out, []byte("CYGWIN"))
 }
 
 func translateCygwinPath(path string) (string, error) {
@@ -589,8 +587,7 @@ func translateCygwinPath(path string) (string, error) {
 	buf := &bytes.Buffer{}
 	cmd.Stderr = buf
 	out, err := cmd.Output()
-	output := string(out)
-	output = strings.TrimSpace(output)
+	output := strings.TrimSpace(string(out))
 	if err != nil {
 		return path, fmt.Errorf("Failed to translate path from cygwin to windows: %s", buf.String())
 	}
--- a/git/git.go
+++ b/git/git.go
@@ -573,6 +573,32 @@ func GetCommitSummary(commit string) (*CommitSummary, error) {
 	}
 }
 
+func isCygwin() bool {
+	cmd := subprocess.ExecCommand("uname")
+	buf := &bytes.Buffer{}
+	cmd.Stderr = buf
+	out, err := cmd.Output()
+	output := string(out)
+	if (err == nil) && (strings.Contains(output, "CYGWIN")) {
+		return true
+	} else {
+		return false
+	}
+}
+
+func translateCygwinPath(path string) (string, error) {
+	cmd := subprocess.ExecCommand("cygpath", "-w", path)
+	buf := &bytes.Buffer{}
+	cmd.Stderr = buf
+	out, err := cmd.Output()
+	output := string(out)
+	output = strings.TrimSpace(output)
+	if err != nil {
+		return path, fmt.Errorf("Failed to translate path from cygwin to windows")
+	}
+	return output, nil
+}
+
 func GitAndRootDirs() (string, string, error) {
 	cmd := subprocess.ExecCommand("git", "rev-parse", "--git-dir", "--show-toplevel")
 	buf := &bytes.Buffer{}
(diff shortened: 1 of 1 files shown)


git/git.go
```diff
@@ -573,6 +573,32 @@ func GetCommitSummary(commit string) (*CommitSummary, error) {
 	}
 }
 
+func isCygwin() bool {
+	cmd := subprocess.ExecCommand("uname")
+	buf := &bytes.Buffer{}
+	cmd.Stderr = buf
```
technoweenie: We don't need to set stderr since it's never used

git/git.go
```diff
@@ -573,6 +573,32 @@ func GetCommitSummary(commit string) (*CommitSummary, error) {
...
+
+func translateCygwinPath(path string) (string, error) {
+	cmd := subprocess.ExecCommand("cygpath", "-w", path)
+	buf := &bytes.Buffer{}
+	cmd.Stderr = buf
+	out, err := cmd.Output()
+	output := string(out)
+	output = strings.TrimSpace(output)
+	if err != nil {
+		return path, fmt.Errorf("Failed to translate path from cygwin to windows")
```
technoweenie: This should show the stderr contents:

    return path, fmt.Errorf("Failed to translate path from cygwin to windows: %s", buf.String())

technoweenie:
> Can't test against master because master doesn't build on cygwin:
>
>     $ ./script/bootstrap
>     Installing goversioninfo to embed resources into Windows executables...
>     Creating the resource.syso version information file...
>     git-lfs.go:1: running "goversioninfo": exec: "goversioninfo": executable file not found in %PATH%

@creste I think you just need to add `$GOPATH/bin` to your `%PATH%`, so it can run `goversioninfo`.

creste:
Thank you for the review.  I pushed changes that should fix the problems you identified.

> How can we get integration tests to run on cygwin on appveyor?

Sorry, I am not familiar with appveyor.  Their documentation says they support cygwin on Windows but that is the extent of my knowledge.

>How can we cache isCygwin(), so it doesn't need to be called multiple times in a process? Probably using a package var and a sync.Once.

Sorry, I am not very familiar with Go.  This is the first Go project I have tried to contribute to, and the changes weren't even mine.  I could try to figure it out but, as you pointed out, it is only called twice in `git lfs lock`.

>@creste I think you just need to add $GOPATH/bin to your %PATH%, so it can run goversioninfo.

That fixed the problem.  Thank you!

git/git.go
```diff
@@ -573,6 +573,30 @@ func GetCommitSummary(commit string) (*CommitSummary, error) {
 	}
 }
 
+func isCygwin() bool {
+	cmd := subprocess.ExecCommand("uname")
+	out, err := cmd.Output()
+	output := string(out)
```
ttaylorr: No need to make a temporary variable here, I would just use the `bytes` equivalent:

```go
bytes.Contains(out, []byte("CYGWIN"))
```

git/git.go
```diff
@@ -573,6 +573,30 @@ func GetCommitSummary(commit string) (*CommitSummary, error) {
...
+	if (err == nil) && (strings.Contains(output, "CYGWIN")) {
```
ttaylorr: This can be in-lined into a single statement:

```go
return err == nil && bytes.Contains(out, []byte("CWGIN"))
```

git/git.go
```diff
@@ -573,6 +573,30 @@ func GetCommitSummary(commit string) (*CommitSummary, error) {
...
+		return false
+	}
+}
+
+func translateCygwinPath(path string) (string, error) {
+	cmd := subprocess.ExecCommand("cygpath", "-w", path)
+	buf := &bytes.Buffer{}
+	cmd.Stderr = buf
+	out, err := cmd.Output()
+	output := string(out)
```
ttaylorr: This can be made into a single statement:

```go
output := strings.TrimSpace(string(out))
```

git/git.go
ttaylorr: Actually, what might be more idiomatic is:

```go
if err != nil {
        return false
}

return bytes.Contains(out, []byte("CYGWIN"))
```

creste:
Thank you for the review, @ttaylorr!  I committed changes that should fix the problems you identified.

Unfortunately, the CI build appears to be failing due to a unit test added for https://github.com/git-lfs/git-lfs/pull/1814, which is unrelated to this PR.  I manually tested the latest version of this PR and it is still working for me on Cygwin.

ttaylorr:
> Thank you for the review, @ttaylorr! I committed changes that should fix the problems you identified.

Thanks so much!

> Unfortunately, the CI build appears to be failing due to a unit test added for #1814, which is unrelated to this PR. I manually tested the latest version of this PR and it is still working for me on Cygwin.

I just pushed #1835 which, when merged, will fix this 👍 . Sorry about that!

PR #1822, opened 2017-01-02
Define lockable files, make read-only in working copy when not locked

sinbad:
This PR addresses the following locking user story points:

1. We want to avoid accidentally editing unmergeable files which have not been locked first
2. The best way to do that before any changes have been made is to make files read-only in the working copy, suggesting to the user that they shouldn't just edit them
3. Not all LFS files are unmergeable. Therefore we want to identify which file patterns this rule should apply to
4. Locking / unlocking a file should flip the read-only flag locally
5. We need to maintain that read-only flag across other git operations

To this end, this PR adds the following features:

1. Adds a `--lockable` option to `git lfs track`. This enables an additional attribute flag `lockable` which means files matching this pattern will be read-only when not locked.
2. Adds a `post-checkout` hook to make files read-only when checked out if needed
3. Adds a `post-commit` hook which makes added files read-only after they're committed. New files won't have been checked out but the locking rules should apply to them as soon as they enter git; `post-commit` is the event that needs to trigger that transition
4. Adds a `post-merge` hook which checks & fixes the read-only flag, which can be reset during merge (`post-checkout` unfortunately does not get called)

On this basis we can be reasonably confident that files marked as `lockable` will be kept read-only in the working copy unless the user locks them, or overrides this flag themselves.

There are a couple of caveats:

1. Some apps blatantly override read-only flags without asking. Some Apple tools do this. There's not a great deal we can do about this; this feature is mainly an advisory / reminder. Luckily it's not many tools.
2. `git reset <file>` can override the read-only flag and there's no hook to catch it. However, the only reason to use that is to undo edits you've already made, so it's reasonable to assume you'd only have changes to undo if you'd already locked the file. Other `reset` routes involve a ref checkout so are caught by the `post-checkout` hook.

The behaviour can be disabled by setting env var `GIT_LFS_SET_LOCKABLE_READONLY` to false, in case there are environments where you don't want this (e.g. automated scratch copies)

Hope this all makes sense!

NB: This PR targets a new integration branch, `locking-master-2` since I can't fast-forward `locking-master` to `master` because of the review checks on it. I'd be fine with targeting `locking-master` as well if that could be brought forward to latest `master`, I just can't do that as I have no admin access.


shana:
This is great! The fact that it's opt-in and per file/file type and can be globally disabled is excellent, it means that if a certain tool/workflow just breaks because of the readonly flags, it can be turned off while still being useful for tools that can take advantage of it.

Can we have some sort of "reset" command for git-lfs that restores readonly flags to all files that should have them (given env variables and lockable flag attributes) but for some reason don't? This would be useful for the scenario where a reset happens and a tool that is lfs-locking-aware detects it and wants to set the correct readonly states back. A tool would need to verify the environment, lockable configuration and locked files list to figure out which files need the flag reset. Would a `git checkout [lfs file]` restore it?

locking/lockable.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+}
+
+// Internal function to repopulate lockable patterns
+// You must have locked the c.lockableMutex in the caller
+func (c *Client) refreshLockablePatterns() {
+	// Always make non-nil even if empty
+	c.lockablePatterns = make([]string, 0, 10)
+
+	paths := git.GetAttributePaths(config.LocalWorkingDir, config.LocalGitDir)
```
technoweenie: This brings back the `config` import to the `locking` package. I'd prefer `locking.Client` get `LocalWorkingDir` and `LocalGitDir` properties, set in `locking.NewClient(*config.Configuration)`. 

The API refactor will move that to [this `commands` helper](https://github.com/git-lfs/git-lfs/blob/api/locking/commands/commands.go#L49-L60).

locking/locks.go
```diff
@@ -98,7 +107,17 @@ func (c *Client) UnlockFile(path string, force bool) error {
...
 
-	return c.UnlockFileById(id, force)
+	err = c.UnlockFileById(id, force)
+	if err != nil {
+		return err
+	}
+
+	// Make non-writeable if required
+	if c.IsFileLockable(path) &&
+		config.Config.Os.Bool("GIT_LFS_SET_LOCKABLE_READONLY", true) {
```
technoweenie: I'd like this to be a property on `locking.Client` too, set with the default value in `locking.NewClient()`.

sinbad:
@shana glad you like it! I've based this on my own experience in game projects, hoping it pans out for others.

A `git checkout` will reset the readonly flags so I didn't think there was a need for a separate command (it's done by the `post-checkout` hook). If you checkout individual files then actually the whole repo has to be scanned, since git doesn't tell hooks what files were checked out (only when you checkout a ref can we optimise it to just files that changed between the refs, since we get that info). However the new repo file scanning tools are extremely fast so this shouldn't be an issue.

docs/man/git-lfs-config.5.ronn
```diff
@@ -223,6 +223,13 @@ lfs option can be scoped inside the configuration for a remote.
   * `total` The entire size of the file, in bytes.
   * `name` The name of the file.
 
+* `GIT_LFS_SET_LOCKABLE_READONLY`
+
+  This environment variable controls whether files marked as 'lockable' in 
+  `git lfs track` are made read-only in the working copy when not locked by 
+  the current user. The default is `true`, you can disable this behaviour and
+  have all files writeable by setting this variable to 0, 'no' or 'false'.
```
shana: Oh yes indeed, good point, having granularity at the repo level on whether this is enabled or not would be very welcome.

shana:
> One question: is it possible to add a lockable, but not LFS file? Should it be? I suppose the entry would just look something like: `*.yml -text lockable`

So this would be really interesting to have, for this: There are files that are technically hard to merge by git or by humans (due to unpredictable serialization behaviours requiring smart merge tools that can be brittle), but are not necessarily binary files (or they are binary files but very small ones). Putting these in LFS for the purposes of having a locking feature for them adds a lot of overhead that isn't really needed, and in the case of text files that are just hard to merge, is actual a detriment to the process (turning a text file into a binary blob unnecessarily).

sinbad:
> One question: is it possible to add a lockable, but not LFS file? Should it be?

It's not possible at the moment. I think you could probably find some cases where this would be useful but they'd be a minority; most of the time it's binary files that you'd want in LFS anyway I think. Adding this as an orthogonal concept would make things more complex (and may even suggest making lockable a separate tool to `git lfs`) and I'm don't think that's justified.

sinbad:
> I also second @shana's request for a command to check/reset the readonly settings. It may make sense to add this to git lfs checkout, though that could overload its purpose (which right now, is to replace any LFS pointer files with the content from .git/lfs/objects).

I mentioned you can trigger this via `git checkout` already but my comment may have overlapped with yours. Do you still think it needs a separate command?

docs/man/git-lfs-config.5.ronn
sinbad: Good shout, adding that now and a wrapper `Configuration` method to combine the two.

git/attribs.go
```diff
@@ -0,0 +1,87 @@
...
+				lockable := strings.Contains(line, LockableAttrib)
+				paths = append(paths, AttributePath{Path: pattern, Source: relfile, Lockable: lockable})
+			}
+		}
+	}
+
+	return paths
+}
+
+func findAttributeFiles(workingDir, gitDir string) []string {
```
sinbad: Unfortunately this is not possible; see the end of the function where the order of the files is reversed to ensure the correct precedence of attributes is maintained.

locking/lockable.go
sinbad: OK; even after doing this it will still require the `config` package, until the per-package config comes in, but at least it puts it in one place ready for that I guess.

git/attribs.go
```diff
@@ -0,0 +1,87 @@
...
+		for scanner.Scan() {
+			line := scanner.Text()
+			if strings.Contains(line, "filter=lfs") {
+				fields := strings.Fields(line)
+				relfile, _ := filepath.Rel(workingDir, path)
+				pattern := fields[0]
+				if reldir := filepath.Dir(relfile); len(reldir) > 0 {
+					pattern = filepath.Join(reldir, pattern)
+				}
+				lockable := strings.Contains(line, LockableAttrib)
```
ttaylorr: This function looks good, but I'm thinking that this line would lock a file named "lockable" (or one that contains the substring "lockable") as lockable on disk, even if the user didn't want it to be. This is sort of edgecase-y, but I think it's still behavior worth checking. Maybe a more proper `.giattributes` parser would be worth investigating?

errors/errors.go
```diff
@@ -100,3 +101,18 @@ func StackTrace(err error) []string {
 
 	return nil
 }
+
+func Combine(errs []error) error {
+	if len(errs) > 0 {
```
ttaylorr: Nit: I'd prefer a smaller `if` statement here:

```go
if len(errs) == 0 {
      return nil
}

// ...
```

It'd be nice if we didn't need the `if` statement at all, but I think this is at least a small step forward 👍

locking/lockable.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/git-lfs/git-lfs/filepathfilter"
+	"github.com/git-lfs/git-lfs/git"
+	"github.com/git-lfs/git-lfs/tools"
+)
+
+// GetLockablePatterns returns a list of patterns in .gitattributes which are
+// marked as lockable
+func (c *Client) GetLockablePatterns() []string {
+	c.ensureLockablesLoaded()
+	return c.lockablePatterns
```
ttaylorr: This read isn't guarded by a mutex-lock. I think using a `sync.RWMutex` would be appropriate here.

locking/lockable.go
ttaylorr: I'm with @technoweenie, let's keep this import light.

locking/lockable.go
```diff
@@ -0,0 +1,209 @@
...
+	if c.lockablePatterns == nil {
+		c.refreshLockablePatterns()
+	}
+}
+
+// Internal function to repopulate lockable patterns
+// You must have locked the c.lockableMutex in the caller
+func (c *Client) refreshLockablePatterns() {
+	// Always make non-nil even if empty
+	c.lockablePatterns = make([]string, 0, 10)
```
ttaylorr: This can be initialized _after_ calling `git.GetAttributePaths` to avoid resizing:

```go
paths := git.GetAttributePaths(...)
c.lockablePatterns = make([]string, 0, len(paths))

for _, p := range paths {
        if p.Lockable {
                c.lockablePatterns = append(c.lockablePatterns, p.Path)
        }
}
// ...
```

commands/command_track.go
```diff
@@ -41,50 +42,110 @@ func trackCommand(cmd *cobra.Command, args []string) {
...
+
+		Print("Tracking %s", pattern)
+
+	}
+
+	// Now read the whole local attributes file and iterate over the contents,
+	// replacing any lines where the values have changed, and appending new lines
+	// change this:
+
+	attribContents, err := ioutil.ReadFile(".gitattributes")
```
ttaylorr: Is it necessary to read `attribContents` up front? Perhaps you could just use `attributesFile` and pass that to `bufio.NewScanner` below on L119?

commands/command_track.go
```diff
@@ -41,50 +42,110 @@ func trackCommand(cmd *cobra.Command, args []string) {
...
+	}
+	// Re-generate the file with merge of old contents and new (to deal with changes)
+	attributesFile, err := os.OpenFile(".gitattributes", os.O_WRONLY|os.O_TRUNC|os.O_CREATE, 0660)
+	if err != nil {
+		Print("Error opening .gitattributes file")
+		return
+	}
+	defer attributesFile.Close()
+
+	if len(attribContents) > 0 {
```
ttaylorr: > Is it necessary to read `attribContents` up front? Perhaps you could just use `attriubutesFile` and pass that to `bufio.NewScanner` below on L119?

This would also allow you to get rid of this conditional 😄

test/test-track.sh
```diff
@@ -292,3 +292,136 @@ begin_test "track blocklisted files with glob"
...
+  set -e
+
+  repo="track_lockable"
+  mkdir "$repo"
+  cd "$repo"
+  git init
+
+  # track *.jpg once, lockable
+  git lfs track --lockable "*.jpg" | grep "Tracking \*.jpg"
+  numjpg=$(grep -e "\*.jpg.*lockable" .gitattributes | wc -l)
```
ttaylorr: `grep -c` should be used in favor of `grep | wc -l` for clarity.

```sh
if [ "1" -ne "$(grep -ce "\*.jpg.*\[lockable\]" .gitattributes)"  ]; then
  # ...
fi
```

ttaylorr: If you want to keep the echo/cat/exit pattern below, I'd extract to this to a function for clarity:

```sh
assert_attributes_count() {
  local ext="$1"
  local lockable="$2"
  local expected="$3"
}

# ...

assert_attributes_count "jpg" "lockable" 1
```

(You could use "$lockable" to help compose the pattern and make it a little easier for callers, or define two functions: `assert_lockable_attribs_count`, `assert_lockable_attribs_count`).

locking/locks.go
```diff
@@ -92,7 +93,8 @@ func (c *Client) LockFile(path string) (Lock, error) {
 		return Lock{}, fmt.Errorf("Error caching lock information: %v", err)
 	}
 
-	return lock, nil
+	// Ensure writeable on return
+	return lock, tools.SetFileWriteFlag(path, true)
```
ttaylorr: I dislike the pattern of returning a single error via a method call in the return signature, because I think this hides too much. I prefer:

```go
if err := tools.SetFileWriteFlag(path, true); err != nil {
        return Lock{}, err
}

return lock, nil

File: lfs/hook.go [hunk 13]
```diff
@@ -22,6 +27,15 @@ type Hook struct {
 	Upgradeables []string
 }
 
+// NewStandardHook creates a new hook using the template script calling 'git lfs theType'
+func NewStandardHook(theType string, upgradeables []string) *Hook {
+	return &Hook{
+		Type:         theType,
+		Contents:     strings.Replace(hookBaseContent, "__CMD__", theType, -1),
```
ttaylorr: I'd prefer to use a `text/template` instead of string replacement, since the later is a common Go convention and is well understood by looking at the `hookBaseContent` variable alone.

commands/command_post_checkout.go
```diff
@@ -0,0 +1,75 @@
...
+// up to 3 arguments on the command line:
+//
+//   1. SHA of previous commit before the checkout
+//   2. SHA of commit just checked out
+//   3. Flag ("0" or "1") - 1 if a branch/tag/SHA was checked out, 0 if a file was
+//      In the case of a file being checked out, the pre/post SHA are the same
+//
+// This hook checks that files which are lockable and not locked are made read-only,
+// optimising that as best it can based on the available information.
+func postCheckoutCommand(cmd *cobra.Command, args []string) {
```
ttaylorr: This command needs a manpage entry.

commands/command_post_checkout.go
```diff
@@ -0,0 +1,75 @@
...
+func postCheckoutRevChange(client *locking.Client, pre, post string) {
+	// We can speed things up by looking at the difference between previous HEAD
+	// and current HEAD, and only checking lockable files that are different
+	files, err := git.GetFilesChanged(pre, post)
+
+	if err != nil {
+		LoggedError(err, "Warning: post-checkout rev diff %v:%v failed: %v\nFalling back on full scan.", pre, post, err)
+		postCheckoutFileChange(client)
+	}
+	err = client.FixLockableFileWriteFlags(files)
```
ttaylorr: Very nice 👍

commands/command_post_merge.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/git-lfs/git-lfs/locking"
+	"github.com/rubyist/tracerx"
+	"github.com/spf13/cobra"
+)
+
+// postMergeCommand is run through Git's post-merge hook.
+//
+// This hook checks that files which are lockable and not locked are made read-only,
+// optimising that as best it can based on the available information.
+func postMergeCommand(cmd *cobra.Command, args []string) {
```
ttaylorr: This command needs a manpage entry.

git/attribs.go
sinbad: That's a good point, if the file entry in `.gitattributes` was specific enough that could be a problem - it's not likely but could be a nasty edge case in that 0.0001% and experience says that's worth addressing.

sinbad: Turns out there's a cheap way to do it, we're already splitting on whitespace so just need to check `fields[1:]` instead. `.gitattributes` always splits on plain whitespace (you have to use `[[:space:]]` in file patterns) so this works.

errors/errors.go
sinbad: 👍 I submit PRs to projects with opposing code styles for this case and sometimes I forget 😄

locking/lockable.go
sinbad: A read-only mutex here wouldn't do anything useful unless the caller also locked it.

`ensureLockablesLoaded` ensures that `c.lockablePatterns` is only updated by one thread, and `refreshLockablePatterns` deliberately replaces the entire map (rather than clearing and re-populating it) so that previous callers to `GetLockablePatterns` can continue to use the old map. Therefore a read-only mutex is not required provided the return from `GetLockablePatterns` is not modified. Unfortunately Go doesn't have a `const` marker for returns, I thought about copying the map each time but ultimately thought it wasn't worth it.

commands/command_track.go
sinbad: Yes, because I'm overwriting the same file. I can't progressively read and overwrite it at the same time 😄 I could write a new attributes file then delete & rename, but it's more complex and I prefer simpler. I recognise that this way is heavier on memory usage, but `.gitattributes` files are never going to be massive (even a 1MB attributes file would be unthinkable) so it's a not really an issue.

test/test-track.sh
sinbad: > `grep -c` should be used in favor of `grep | wc -l` for clarity.

Problem with that is that `grep -c` returns a failure code when there are no results (which we validly test for in many cases), which means you lose the gained clarity trying to ignore that. I think `wc -l` is clearer while still dealing with `0` result cases.

sinbad: I've made the `assert_attributes_count` method and it tidies things up, nice idea.

locking/locks.go
sinbad: I'm sure I've had the opposite feedback before 😆 Fair enough.

lfs/hook.go
sinbad: That would be a lot of cruft for what will boil down to a global find/replace even with a template.  If you like I can make the string look like a template and still use `strings.Replace` but I see no reason to clutter this with all the init/struct param/execute patterns of templates.

ttaylorr:
> That would be a lot of cruft for what will boil down to a global find/replace even with a template.  If you like I can make the string look like a template and still use `strings.Replace` but I see no reason to clutter this with all the init/struct param/execute patterns of templates.

I think that would be a great compromise 👍 .

ttaylorr:
> Problem with that is that `grep -c` returns a failure code when there are no results (which we validly test for in many cases), which means you lose the gained clarity trying to ignore that. I think `wc -l` is clearer while still dealing with `0` result cases.

I've had success with stringify-ing it, but either works for me :-)

```sh
[ "0" -eq "$(grep -c foo bar)" ]
```

ttaylorr:
> Yes, because I'm overwriting the same file. I can't progressively read and
> overwrite it at the same time 😄

My mistake, I totally missed this on my first read-through of the code.

> I could write a new attributes file then delete & rename, but it's more
> complex and I prefer simpler. I recognise that this way is heavier on memory
> usage, but `.gitattributes` files are never going to be massive (even a 1MB
> attributes file would be unthinkable) so it's a not really an issue.

I think this is a great analysis, and OK to leave as-is.

ttaylorr:
Fantastic!

ttaylorr:
> A read-only mutex here wouldn't do anything useful unless the caller also
> locked it.
>
> `ensureLockablesLoaded` ensures that `c.lockablePatterns` is only updated by
> one thread, and `refreshLockablePatterns` deliberately replaces the entire map
> (rather than clearing and re-populating it) so that previous callers to
> `GetLockablePatterns` can continue to use the old map. Therefore a read-only
> mutex is not required provided the return from `GetLockablePatterns` is not
> modified. Unfortunately Go doesn't have a `const` marker for returns, I
> thought about copying the map each time but ultimately thought it wasn't worth
> it.

Sounds good, and thanks for the detailed explanation :-).

docs/man/git-lfs-config.5.ronn
```diff
@@ -223,6 +223,15 @@ lfs option can be scoped inside the configuration for a remote.
...
   * `name` The name of the file.
 
+* `GIT_LFS_SET_LOCKABLE_READONLY`
+  `lfs.setlockablereadonly`
+
+  These settings, the first an environment variable and the second a gitconfig
+  setting, control whether files marked as 'lockable' in `git lfs track` are
+  made read-only in the working copy when not locked by the current user.
+  The default is `true`; you can disable this behaviour and have all files
+  writeable by setting either variable to 0, 'no' or 'false'.
```
larsxschneider: @ttaylorr If the lockable feature is on, then Git LFS cannot [skip](https://github.com/git-lfs/git-lfs/blob/a6a5ba2f38d364f60d4f1b4b5361888fa76575bd/commands/command_post_checkout.go#L29) the `post-checkout` command. Consequently, Git LFS needs to [search](https://github.com/git-lfs/git-lfs/blob/a6a5ba2f38d364f60d4f1b4b5361888fa76575bd/git/attribs.go#L149-L159) for all `.gitattributes` files to check them for lockable patterns.

My team is working with a large repo (>20k directories, >300k files) and just searching for the `.gitattributes` file in all directories takes ~6 sec on macOS with a fast SSD. Consider these timings:


```
$ export GIT_LFS_SET_LOCKABLE_READONLY=1; time git-lfs post-checkout a0c5174a19f2dece44c09bc76f91feb69799e66f a0c5174a19f2dece44c09bc76f91feb69799e66f 0
git-lfs post-checkout a0c5174a19f2dece44c09bc76f91feb69799e66f  0  

35.09s user 6.41s system 730% cpu 5.679 total


$ export GIT_LFS_SET_LOCKABLE_READONLY=0; time git-lfs post-checkout a0c5174a19f2dece44c09bc76f91feb69799e66f a0c5174a19f2dece44c09bc76f91feb69799e66f 0
git-lfs post-checkout a0c5174a19f2dece44c09bc76f91feb69799e66f  0  

0.01s user 0.02s system 89% cpu 0.030 total
```

I plan to distribute the setting `lfs.setlockablereadonly=false` in my company gitconfig but it took me a while to figure out the culprit of the slowness. I don't know how widely the lockable feature is used, but I wonder if `lfs.setlockablereadonly=false`is a more sensible default? Alternatively we could find ways to speed up the `.gitattributes` discovery.

ttaylorr: > I don't know how widely the lockable feature is used, but I wonder if `lfs.setlockablereadonly=false` is a more sensible default?

I think `=true` is an important default value to keep. Here are a few thoughts on how we can get the best of both worlds:

1. Speed up the time it takes to search for `.gitattributes`. This has been optimized by @sinbad in https://github.com/git-lfs/git-lfs/pull/1616. I think that this work could be augmented by first `stat()`-ing directories to search, and then seeing if the stat is out-of-date with the last time that the directory was traversed, effectively memoizing the search. I presume that this would improve the speed a great deal.

2. Warn when the search is taking too long. We can set a timer (similar to what Microsoft does with GVFS when the `--ahead-behind` check takes too long) and warn if Git LFS is unable to determine the set of lockable files quickly.

What do you think? Perhaps we should have this discussion in a new issue, as well.

commands/command_post_merge.go
alivalingad56: > 

> __

PR #1824, opened 2017-01-03
api refactor: locking

technoweenie:
This updates the locking package to use the new `lfsapi.Client` from the `api-master` branch. As a result, this removes the dependencies on the `api` and `config` packages from `locking`, while adding `lfsapi` of course.

### DIFF
--- a/lfsapi/client.go
+++ b/lfsapi/client.go
@@ -15,6 +15,30 @@ import (
 
 var UserAgent = "git-lfs"
 
+func (c *Client) NewRequest(method string, e Endpoint, suffix string, body interface{}) (*http.Request, error) {
+	req, err := http.NewRequest(method, joinURL(e.Url, suffix), nil)
+	if err != nil {
+		return req, err
+	}
+
+	if body != nil {
+		if merr := MarshalToRequest(req, body); merr != nil {
+			return req, merr
+		}
+	}
+
+	return req, err
+}
+
+const slash = "/"
+
+func joinURL(prefix, suffix string) string {
+	if strings.HasSuffix(prefix, slash) {
+		return prefix + suffix
+	}
+	return prefix + slash + suffix
+}
+
 func (c *Client) Do(req *http.Request) (*http.Response, error) {
 	req.Header.Set("User-Agent", UserAgent)
 
--- a/lfsapi/client.go
+++ b/lfsapi/client.go
@@ -94,11 +94,11 @@ func (c *Client) httpClient(host string) *http.Client {
 	defer c.clientMu.Unlock()
 
 	if c.gitEnv == nil {
-		c.gitEnv = make(testEnv)
+		c.gitEnv = make(Env)
 	}
 
 	if c.osEnv == nil {
-		c.osEnv = make(testEnv)
+		c.osEnv = make(Env)
 	}
 
 	if c.hostClients == nil {
--- a/lfsapi/client.go
+++ b/lfsapi/client.go
@@ -161,6 +161,12 @@ func (c *Client) httpClient(host string) *http.Client {
 	return httpClient
 }
 
+func (c *Client) CurrentUser() (string, string) {
+	userName, _ := c.gitEnv.Get("user.name")
+	userEmail, _ := c.gitEnv.Get("user.email")
+	return userName, userEmail
+}
+
 func newRequestForRetry(req *http.Request, location string) (*http.Request, error) {
 	newReq, err := http.NewRequest(req.Method, location, nil)
 	if err != nil {
--- a/lfsapi/client.go
+++ b/lfsapi/client.go
@@ -15,16 +15,21 @@ import (
 
 var UserAgent = "git-lfs"
 
+const MediaType = "application/vnd.git-lfs+json; charset=utf-8"
+
 func (c *Client) NewRequest(method string, e Endpoint, suffix string, body interface{}) (*http.Request, error) {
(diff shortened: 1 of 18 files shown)


test/cmd/lfstest-gitserver.go
```diff
@@ -76,6 +76,14 @@ func main() {
 			return
 		}
 
+		if strings.Contains(r.URL.Path, "/info/lfs/locks") {
```
ttaylorr: Does this handle unlocking? We should probably remove the two old handlers on L71-72 to make sure that we're hitting the right URL.

ttaylorr: I used these to make sure all requests were routed correctly: https://github.com/git-lfs/git-lfs/pull/1818/files#diff-a3b6c1ff24bf259401c990b87d408bb6R160.

locking/locks.go
```diff
@@ -22,31 +21,51 @@ var (
...
 // You must call the returned object's `Close` method when you are finished with
 // it
-func NewClient(cfg *config.Configuration) (*Client, error) {
-
-	apiClient := api.NewClient(api.NewHttpLifecycle(cfg))
+func NewClient(remote string, lfsClient *lfsapi.Client) (*Client, error) {
+	return &Client{
+		Remote: remote,
+		client: &lockClient{Client: lfsClient},
+		cache:  &nilLockCacher{},
```
ttaylorr: I'm not sure I'm a fan of this pattern in Go. I'd rather sacrifice some if-statements to have a more straightforward set of types than have gain an interface and another implementation. I don't feel strongly, so if you like this pattern, 👍.

locking/locks.go
```diff
@@ -22,31 +21,51 @@ var (
...
+	}, nil
+}
 
-	lockDir := filepath.Join(config.LocalGitStorageDir, "lfs")
-	err := os.MkdirAll(lockDir, 0755)
+func (c *Client) SetupFileCache(path string) error {
```
ttaylorr: What do you think about having an option func passed to the constructor instead of this? I would love to avoid having to construct and then call another function before using something. I was thinking:

```go
locking.NewClient(foo, bar, locking.WithFileCache(path))
```

I think this fits the bill of a good optional argument.

locking/locks.go
technoweenie: I prefer the null object pattern.

locking/locks.go
technoweenie: Option funcs are good if the values are needed in the constructor. Otherwise, setting properties seems to be closer to go conventions. You only have to call `SetupFileCache()` if you want a cacher.

lfsapi/lfsapi.go
```diff
@@ -89,10 +90,25 @@ func NewClient(osEnv env, gitEnv env) (*Client, error) {
 	return c, nil
 }
 
-func decodeResponse(res *http.Response, obj interface{}) error {
+func IsDecodeTypeError(err error) bool {
```
technoweenie: I don't think the `errors` package should know about _all_ lfs errors.

locking/locks.go
technoweenie: To expand on that, things like zero values, interfaces, and typed nils in Go make me think it's designed for the null object pattern. The added locking cacher interface gives us the option to plug in alternate caching types too.

lfsapi/client.go
```diff
@@ -15,6 +15,35 @@ import (
...
+		}
+		req.Header.Set("Content-Type", MediaType)
+	}
+
+	return req, err
+}
+
+const slash = "/"
+
+func joinURL(prefix, suffix string) string {
```
technoweenie: It does extra stuff we don't need (calling `path.Clean()`) in our limited uses of `(*Client) NewRequest()` that we don't need. `suffix` is always a simple string like `"objects/batch"` or `"locks"`. Though I suppose anyone can import the package and call it with a weird `suffix`. Though I'm not sure why anyone would do that.

test/cmd/lfstest-gitserver.go
technoweenie: I'm not going to worry about this since `master` diverged when you fixed the locking relative path issue.

locking/locks.go
ttaylorr: Sounds good.

locking/locks.go
ttaylorr: Works for me 😄

lfsapi/lfsapi.go
ttaylorr: I agree.

lfsapi/client.go
technoweenie: Doesn't work without parsing the URL:

```
	Error:		Not equal: "https://example.com/a" (expected)
			        != "https:///example.com/a" (actual)
```

locking/api_test.go
```diff
@@ -0,0 +1,142 @@
...
+	}))
+	require.Nil(t, err)
+
+	lc := &lockClient{Client: c}
+	locks, res, err := lc.Search("", &lockSearchRequest{
+		Filters: []lockFilter{
+			{Property: "a", Value: "A"},
+		},
+		Cursor: "cursor",
+		Limit:  5,
```
technoweenie: Pagination isn't handled here, it's in `locking.Client`.

technoweenie:
Thanks for the review. A `locking.Client` pagination test should be added, but not in this API PR (since that pagination code is in the locking client).

PR #1825, opened 2017-01-04
Add test for ssh auth

technoweenie:
This adds a test for accessing the batch api on an ssh endpoint. The test sets `GIT_SSH` to a custom `ssh-echo` command so we don't have to set up a real ssh server.


ttaylorr:
👍

PR #1826, opened 2017-01-04
Api/tq manifest

technoweenie:
This breaks the `tq.Manifest` constructor, so that it accepts an `*lfsapi.Client` instance. This is how transfer queue and adapters get access to the `*lfsapi.Client` to make API requests. It also removes some references to the `CurrentRemote` property on `*config.Configuration` objects. Since the `*lfsapi.Client` already has os and git env objects, there's no need to pass them in manually. 

I'm not totally sold on whether the manifest constructor should accept `operation` and `remote` strings, vs just an `access` string. I think a future change could remove the need for both values, leaving us with a single `NewManifest(c *lfsapi.Client)` constructor. 

TODO:

* [ ] Review, merge (locking) #1824
* [ ] Change base to `api-master`


technoweenie:
I hated the new `tq.Manifest` constructors, so I removed all the arguments except the `*lfsapi.Client` in #1827.

PR #1827, opened 2017-01-04
Api/batch

technoweenie:
This updates the `tq.TransferQueue` to use `lfsapi.Client` to make batch API requests, following the pattern set in #1824.

Changes:

* Removed the need for `access`, `operation`, or `remote` in `tq.Manifest`. I initially went with the approach described in https://github.com/git-lfs/git-lfs/pull/1826, but didn't like it. Now, a `*tq.Manifest` is fully immutable, and can be used across multiple transfer queues for different LFS servers. This was impossible before, since `access` was scoped to an LFS endpoint.
* The `tq.batchResponse` now stores the `lfsapi.Endpoint` used for the batch request. `tq.toAdapterCfg()` uses that Endpoint in conjunction with the transfer queue's `tq.Manifest` to provide the config for adapters. This is where the concurrency will be set to `1` for ntlm authenticated LFS services.
* `tq.NewTransferQueue()` now requires a `remote` argument, since it can't use `CurrentRemote` in `*config.Configuration`. 
* Updated all http transfer adapters to use `lfsapi.Client`.

TODO:

* [x] Review, merge (locking) #1824
* [x] Change base to `api-master`


tq/api.go
```diff
@@ -0,0 +1,66 @@
...
+	TransferAdapterNames []string              `json:"transfers,omitempty"`
+}
+
+type batchResponse struct {
+	Endpoint            lfsapi.Endpoint
+	TransferAdapterName string                `json:"transfer"`
+	Objects             []*api.ObjectResource `json:"objects"`
+}
+
+func (c *tqClient) Batch(remote string, bReq *batchRequest) (*batchResponse, *http.Response, error) {
```
ttaylorr: 😍  Really clean, this is awesome.

tq/adapterbase.go
```diff
@@ -172,6 +176,26 @@ func (a *adapterBase) worker(workerNum int, ctx interface{}) {
 	a.workerWait.Done()
 }
 
+func (a *adapterBase) newHTTPRequest(method string, rel *Action) (*http.Request, error) {
```
ttaylorr: Nice 👌, so cool that we got rid of the `httputil` dependency.

lfsapi/lfsapi.go
```diff
@@ -90,6 +90,14 @@ func NewClient(osEnv env, gitEnv env) (*Client, error) {
 	return c, nil
 }
 
+func (c *Client) GitEnv() env {
```
technoweenie: Not really, since you pass the os and git envs to `NewClient()`. If transfer adapters get extracted to a separate package, I think it will need to be public.  I'll unexport for now, and we can export later if necessary.

commands/commands.go
```diff
@@ -59,25 +85,22 @@ func newLockClient(remote string) *locking.Client {
...
-// cfg var.
-func TransferManifest() *tq.Manifest {
-	return lfs.TransferManifest(cfg)
-}
-
 // newDownloadCheckQueue builds a checking queue, checks that objects are there but doesn't download
-func newDownloadCheckQueue(options ...tq.Option) *tq.TransferQueue {
-	return lfs.NewDownloadCheckQueue(cfg, options...)
+func newDownloadCheckQueue(manifest *tq.Manifest, remote string, options ...tq.Option) *tq.TransferQueue {
+	allOptions := make([]tq.Option, len(options), len(options)+1)
```
technoweenie: Sure, better safe than sorry.

lfsapi/lfsapi.go
technoweenie: Gah, should've looked at the code. This is needed in `tq` :(

ttaylorr:
Pending a 🍏  test-suite, 👍

PR #1829, opened 2017-01-05
RFC: add option "--size" to ls-files to print the file size

larsxschneider:
This was useful for me to inspect the size of LFS files.
 
There are two somewhat controversial decisions in there and I would like to get your feedback on them: 

1. I print size in MB only. Everything else made the code more complicated for little gain IMHO.
2. I pad the size with 6 chars to align it nicely... this would of course break if someone tracks a crazy large file


technoweenie:
This seems fine to me. Why not make it default though?

larsxschneider:
I thought about making it the default. I didn't do it because I thought maybe someone is using the current `git lfs ls-files` output format? (I spent to much time on the git-core mailing list and adopted their backwards compatibility thinking, haha).

larsxschneider:
Up to you. Default would be fine with me!

technoweenie:
How about we merge this and backport to v1.5.x, but make it default in a follow-up PR for 2.x (master)?

ttaylorr:
I believe we have some helpers to convert to MB/GB etc. I'm just a few
moments away from my computer, would you mind holding off until I can find
that?

larsxschneider:
In the default version we wouldn't need the `-s` / `--size` switch anymore, right? In that case I would rather remove all this cruft. I don't think this thing requires back-porting...

@ttaylorr no rush. I thought about something like that. I wonder if this makes automated parsing harder (I use the output in scripts)? Maybe an automated MB/GB conversion could be the default and with a special argument we make it always print bytes without padding? Would that be too complicated?

technoweenie:
I say we keep size in MB as default. Anything special should be handled in some kind of `--format` flag, as proposed in https://github.com/git-lfs/git-lfs/pull/1814#issuecomment-269961922. That definitely does not need to happen in this PR though.

@ttaylorr Sure, I was hoping to get a third opinion before merging :)

ttaylorr:
[Here's][1] the piece of code that I was thinking about, but I'm not sure that it's relevant to the suggestion I'm going to make. I think a good compromise on this would be making `--size` take an argument, i.e.: `--size=mb`, `--size=b`, `--size=gb`, etc. The default `--size=` could just assume MBs. I think this would be a good start on our way towards full support of `--porcelain`/`--format`. @larsxschneider @technoweenie what are your thoughts on this?

The code looks 👍, but I think adding an integration test asserting that the sizes get printed out as expected would be good before merging. 

[1]: https://github.com/git-lfs/git-lfs/blob/85f581166bba18dce6484e34ab5ba6dc03387405/progress/meter.go#L207-L220.

technoweenie:
>  The default `--size=` could just assume MBs. I think this would be a good start on our way towards full support of `--porcelain/--format`. 

How so? A custom arg like `--size` seems the opposite of a general `--format` arg.

ttaylorr:
> How so? A custom arg like `--size` seems the opposite of a general `--format` arg.

I was thinking that it'd be strange for us to implement `--porcelain` /
`--format` in some commands but not others within a given release. I figured
that `--size` would be a good short-term solution that we could release in 2.0.0
without committing to `--porcelain` and `--format` package-wide.

Though if you don't think it'd be that big of a deal to implement this one
command at a time (potentially over a few releases), then I think we should go
right for `--format`, and skip `--size` altogether :-).

technoweenie:
Since there's not a huge need to backport to 1.5, I'd rather just display the size in MB by default, and  potentially add `--format` before a 2.0 release. I don't think incrementally adding `--format` support over several releases is a bad thing.

ttaylorr:
> I don't think incrementally adding `--format` support over several releases is
> a bad thing.

Sounds good to me :-). I think adding a `--format` flag here would be a good
call in that case.

larsxschneider:
Superseded by #2540

PR #1836, opened 2017-01-06
Ntlm test

technoweenie:
This adds an integration test for NTLM authentication.

### DIFF
--- a/test/cmd/lfstest-gitserver.go
+++ b/test/cmd/lfstest-gitserver.go
@@ -27,6 +27,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/ThomsonReutersEikon/go-ntlm/ntlm"
 )
 
 var (
@@ -59,6 +61,12 @@ func main() {
 	mux := http.NewServeMux()
 	server = httptest.NewServer(mux)
 	serverTLS = httptest.NewTLSServer(mux)
+	ntlmSession, err := ntlm.CreateServerSession(ntlm.Version2, ntlm.ConnectionOrientedMode)
+	if err != nil {
+		fmt.Println("Error creating ntlm session:", err)
+		os.Exit(1)
+	}
+	ntlmSession.SetUserInfo("ntlmuser", "ntlmpass", "NTLMDOMAIN")
 
 	stopch := make(chan bool)
 
@@ -75,7 +83,7 @@ func main() {
 		}
 
 		if strings.Contains(r.URL.Path, "/info/lfs") {
-			if !skipIfBadAuth(w, r, id) {
+			if !skipIfBadAuth(w, r, id, ntlmSession) {
 				lfsHandler(w, r, id)
 			}
 
@@ -138,7 +146,7 @@ type lfsLink struct {
 }
 
 type lfsError struct {
-	Code    int    `json:"code"`
+	Code    int    `json:"code,omitempty"`
 	Message string `json:"message"`
 }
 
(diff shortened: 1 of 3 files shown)


test/cmd/lfstest-gitserver.go
```diff
@@ -138,10 +146,24 @@ type lfsLink struct {
...
 type lfsError struct {
-	Code    int    `json:"code"`
+	Code    int    `json:"code,omitempty"`
 	Message string `json:"message"`
 }
 
+func writeLFSError(w http.ResponseWriter, code int, msg string) {
+	by, err := json.Marshal(&lfsError{Message: msg})
+	if err != nil {
+		w.Header().Set("Content-Type", "text/plain")
```
ttaylorr: You can use `http.Error()`, which I think would look very clean:

```go
by, err := json.Marshal(&lfsError{Message: msg})
if err != nil {
        http.Error(w, errors.Wrap(err, "json encoding error").Error(), 500)
        return
}

// ...
```

test/cmd/lfstest-gitserver.go
```diff
@@ -1116,6 +1133,49 @@ func skipIfBadAuth(w http.ResponseWriter, r *http.Request, id string) bool {
...
+		authHeader = ""
+	}
+
+	switch authHeader {
+	case "":
+		w.Header().Set("Www-Authenticate", "ntlm")
+		w.WriteHeader(401)
+
+	// ntlmNegotiateMessage from httputil pkg
+	case "NTLM TlRMTVNTUAABAAAAB7IIogwADAAzAAAACwALACgAAAAKAAAoAAAAD1dJTExISS1NQUlOTk9SVEhBTUVSSUNB":
```
ttaylorr: Could this be extracted to a `const` with some godoc?

test/test-batch-transfer.sh
```diff
@@ -7,6 +7,7 @@
 begin_test "batch transfer"
 (
   set -e
+  exit 0
```
ttaylorr: Did you mean to delete this locally? 😁

test/test-batch-transfer.sh
```diff
@@ -122,3 +123,24 @@ begin_test "batch transfers with ssh endpoint"
   git push origin master 2>&1
 )
 end_test
+
+begin_test "batch transfers with ntlm server"
```
ttaylorr: Nice, this looks really clean 👍

test/cmd/lfstest-gitserver.go
```diff
@@ -1116,6 +1133,49 @@ func skipIfBadAuth(w http.ResponseWriter, r *http.Request, id string) bool {
...
+		w.WriteHeader(401)
+
+	default:
+		if !strings.HasPrefix(strings.ToUpper(authHeader), "NTLM ") {
+			writeLFSError(w, 500, "bad authorization header: "+authHeader)
+			return
+		}
+
+		auth := authHeader[5:] // strip "ntlm " prefix
+		val, err := base64.StdEncoding.DecodeString(auth)
```
ttaylorr: Should we be using `base64.UrlEncoding` here since this is from an HTTP request?

test/cmd/lfstest-gitserver.go
technoweenie: It could, but it's already in `httputil` (and will be moved to `lfsapi`. Could just export it there, and use here.

test/cmd/lfstest-gitserver.go
technoweenie: Rad.

test/cmd/lfstest-gitserver.go
technoweenie: No, this is going in an http header and not a url parameter.

test/test-batch-transfer.sh
technoweenie: Doesn't look like anything to me.

technoweenie:
Closing in favor of #1840

PR #1837, opened 2017-01-06
Api/ssh auth

technoweenie:
Ports SSH auth from the old httputil package. Uses the integration test added in #1825.


test/test-batch-transfer.sh
```diff
@@ -100,3 +100,25 @@ begin_test "batch transfers occur in reverse order by size"
...
+
+  reponame="batch-ssh"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" "$reponame"
+
+  sshurl="${GITSERVER/http:\/\//ssh://git@}/$reponame"
+  git config lfs.url "$sshurl"
+  git lfs env
+
+  oid="$(calc_oid "test")"
```
ttaylorr: Nit: if you want, I thinking naming this with a `$contents` variable would be consistent with the rest of the test suite and bring some additional clarity:

```sh
contents="test"
oid="$(calc_oid "$contents")"

printf "$contents" > test.dat
```

PR #1841, opened 2017-01-06
tq: move api.ObjectResource to tq pkg

technoweenie:
This removes the last `api` dependency in `tq`.

### DIFF
--- a/lfsapi/errors.go
+++ b/lfsapi/errors.go
@@ -3,14 +3,32 @@ package lfsapi
 import (
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/git-lfs/git-lfs/errors"
 )
+
+type httpError interface {
+	Error() string
+	HTTPResponse() *http.Response
+}
+
+func IsHTTPError(err error) (*http.Response, bool) {
+	if httpErr, ok := err.(httpError); ok {
+		return httpErr.HTTPResponse(), true
+	}
+	return nil, false
+}
 
 type ClientError struct {
 	Message          string `json:"message"`
 	DocumentationUrl string `json:"documentation_url,omitempty"`
 	RequestId        string `json:"request_id,omitempty"`
+	response         *http.Response
+}
+
+func (e *ClientError) HTTPResponse() *http.Response {
+	return e.response
 }
 
 func (e *ClientError) Error() string {
@@ -29,7 +47,7 @@ func (c *Client) handleResponse(res *http.Response) error {
 		return nil
 	}
 
-	cliErr := &ClientError{}
+	cliErr := &ClientError{response: res}
 	err := DecodeJSON(res, cliErr)
 	if IsDecodeTypeError(err) {
 		err = nil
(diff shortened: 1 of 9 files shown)


technoweenie:
Sounds good. It is strange though since now a `tq.Transfer` is both a return value from the batch api, and a value sent to transfer adapters (with an extra name and path). I kept the existing `newTransfer()` where it returns a new `tq.Transfer` with `name` and `path` set, instead of just modifying it inline. I don't want to start sending name/path properties to LFS servers.

technoweenie:
I ripped the `api` package out of the `git-lfs-test-server-api`. This actually functions as a good use case for an external tool using git-lfs packages as libraries. I'm not at all happy with the commit, but it's a good first pass.

1. A static function that accepts a `tq.Manifest`, and then builds a `tq.tqClient` doesn't seem ideal. Should the function just be on the Manifest? Seems like too much responsibility for that type.
2. `tq.Batch()` hides some crucial fields from `tq.batchResponse`: `Endpoint` and `TransferAdapterName`. They seem useful [inside the transfer queue](https://github.com/git-lfs/git-lfs/blob/924befe2f088890be5102c740b7c84a09d9d6041/tq/transfer_queue.go#L291-L377), but not sure they're useful externally.
3. The eternal API client question: how do we return the `http.Response`? 3rd argument? Property on the return value?

technoweenie:
Based on some IRL discussion with @ttaylorr: I made the following changes:

* Reverted `tq.Direction` to an int, added a `String()` func. This gets rid of `transferKind()` :metal:
* Added `IsHTTPError(error) (*http.Response, bool)` to determine if an API error has an associated `*http.Response` with it. This saves us from having to return a third `*http.Response` value.
* Updated `tq.TransferQueue` code to use `tq.Batch()` internally too.

tq/transfer.go
```diff
@@ -10,13 +10,24 @@ import (
...
+type Direction int
 
 const (
-	Upload   = Direction("upload")
-	Download = Direction("download")
+	Upload   = Direction(iota)
+	Download = Direction(iota)
 )
 
+func (d Direction) String() string {
```
ttaylorr: Nice ✨

lfsapi/errors.go
```diff
@@ -3,14 +3,32 @@ package lfsapi
...
 
 	"github.com/git-lfs/git-lfs/errors"
 )
 
+type httpError interface {
+	Error() string
+	HTTPResponse() *http.Response
+}
+
+func IsHTTPError(err error) (*http.Response, bool) {
```
ttaylorr: I think we can call this `IsHTTPErr` for brevity, but I don't feel strongly.

technoweenie: Maybe `IsHTTP(error)` would be better? Goes along with the `os` package:

```
func IsExist(err error) bool
func IsNotExist(err error) bool
func IsPathSeparator(c uint8) bool
func IsPermission(err error) bool
```

ttaylorr:
Sounds like a great idea :-).

PR #1845, opened 2017-01-09
Update custom-transfers.md

cbartz:
Fixed wrong key in documentation.

### DIFF
--- a/docs/custom-transfers.md
+++ b/docs/custom-transfers.md
@@ -250,7 +250,7 @@ When all transfers have been processed, git-lfs will send the following message
 to the stdin of the transfer process:
 
 ```json
-{ "event":"terminate" }
+{ "type":"terminate" }
 ```
 
 On receiving this message the transfer process should clean up and terminate.


technoweenie:
@sinbad: Does this look right to you? That definitely looks like [what `git-lfs` sends](https://github.com/git-lfs/git-lfs/blob/a0d9487e64d8059410fe9d4aa83c1cfb06477002/tq/custom.go#L88-L94), even if the [test custom adapter is looking for `event`](https://github.com/git-lfs/git-lfs/blob/a0d9487e64d8059410fe9d4aa83c1cfb06477002/test/cmd/lfstest-customadapter.go#L53-L56).

ttaylorr:
This looks correct to me. Thanks @technoweenie for sending the relevant links
along.

LGTM, pending @cbartz signing the CLA 👍🏻.

cbartz:
Ok. I wrote a custom transfer agent, and it didnt work with _event_ , thats how I found the bug.
I'll change my adapter to listen for _event_ and _type_ to get the terminate message. With this, it won't break once the bug fix has been released. 

Btw: In the doc file there is a line stating:

> After the initiation exchange, git-lfs will send any number of transfer requests to the stdin of the transfer process. 

Does this mean that all the transfer requests are send in one bunch? Or can it be the case that the git-lfs process waits for replies of the custom transfer adapter, and then send more transfer requests? For me, this behaviour gets not clear out of the documentation.

sinbad:
Apologies & thanks for spotting this, I've written 2 custom transfer agents but by sheer luck this bug didn't make either of them fatally break, they just fell out of the I/O loop when the stream was closed. 

> Does this mean that all the transfer requests are send in one bunch? Or can it be the case that the git-lfs process waits for replies of the custom transfer adapter, and then send more transfer requests? For me, this behaviour gets not clear out of the documentation.

Each instance of the custom adapter will only have one transfer in flight at once. So no more requests will be sent until the adapter sends back a "complete" response for that OID (successful or not). After the "complete" response, LFS may send more requests to the same stdin. So the exchange within a single adapter instance is serial. 

Git LFS will instantiate `lfs.concurrenttransfers` instances of the adapter process and so requests will happen in parallel, just between processes.

I'll make this clearer in the docs.

cbartz:
@sinbad : Thanks. FYI, I published my custom transfer agent [here](https://github.com/cbartz/git-lfs-swift-transfer-agent).

sinbad:
@cbartz oh, nice!

Issue #1783, opened 2016-12-16
Proposal: `lfsapi` package for v2.0.0


ttaylorr:
# `lfsapi` Proposal

Today, in order to interact with an API inside git-lfs, you have to import the
`api` package, which depends on `httputil`, `auth`, and of course, a
`*config.Configuration`. That adds up to a pretty heavy dependency.

Packages should be able to interact with the LFS API without importing such
a large set of dependencies. To do this, we can evolve the same philosophy that
applied to the `config` package: by reducing the scope of domain knowledge that
the package (`api`) owns, we can distribute calling code that uses it further
as the dependency will loose weight.

This much more closely matches the philosophy of the `net/http` package.
Clients would be able to make their own `*http.Request` instances, passing
them into the `*api.Client.Do` function.

This package should allow easy access to the following domains, while still keeping things lightweight:

* Endpoint resolving (experimented with in https://github.com/git-lfs/git-lfs/pull/1770)
* Git Credentials
* Netrc
* NTLM
* SSH auth
* HTTP request logging and redirection handling

By reducing the set of primitives exposed by an API package, we can remove the
tight coupling that currently exists in the `locking` package between service
classes and API calls.

Right now, that looks something like:

```go
s, resp := c.apiClient.Locks.Lock(&api.LockRequest{
	Path:               path,
	Committer:          api.NewCommitter(c.cfg.CurrentCommitter()),
	LatestRemoteCommit: latest.Sha,
})

if _, err := c.apiClient.Do(s); err != nil {
	return Lock{}, fmt.Errorf("Error communicating with LFS API: %v", err)
}

if len(resp.Err) > 0 {
	return Lock{}, fmt.Errorf("Server unable to create lock: %v", resp.Err)
}

lock := c.newLockFromApi(*resp.Lock)
```

But it could be reduced down to:

```go
lockReq := &api.LockRequest{
  Path:               path,
  Committer:          api.NewCommitter(c.cfg.CurrentCommitter()),
  LatestRemoteCommit: latest.Sha,
}

endpoint := apiClient.Endpoint("master", "download")
req, _ := http.NewRequest("POST", "/locks", jsonToIOReader(lockReq))
res, _ := apiClient.Do(endpoint, req)
defer res.Body()

var lockRes LockResponse
unmarshalBody(res.Body, &lockRes)
```

This enables packages like `tq` to define their own API interfaces,
perhaps like:

```go
package tq

type Batcher struct {
	c *lfsapi.Client
}

[...]

PR #1846, opened 2017-01-09
Pkg cleanup

technoweenie:
Bye old code :) See #1783 for discussion on the new `lfsapi` package that replaced all this.


technoweenie:
Not in its current state. JSON validation in the lfsapi tests would be nice though. Let's just add that to the 2.0 milestone.

ttaylorr:
> Not in its current state.

I'm happy to make any changes to that package as the original author :-). What
can I do to bring it up to speed and get it ready for its new home in `lfsapi`?

technoweenie:
Following current conventions, I think the core schema validation code would live in `lfsapi`. Schemas themselves would live in the appropriate packages (`locking` for the locking endpoints, and `tq` for the batch endpoint). 

I think this is out of scope for this PR though.

ttaylorr:
> Following current conventions, I think the core schema validation code would live in `lfsapi`. Schemas themselves would live in the appropriate packages (`locking` for the locking endpoints, and `tq` for the batch endpoint).

Sounds good. I can follow up in a future PR to re-introduce the
`SchemaValidator` as a publicly exported type in the `lfsapi` package which
would have methods:

- `Assert(t *testing.T)`, and
- `Refute(t *testing.T)`.

Other packages can use that type to validate their own schemas.

> I think this is out of scope for this PR though.

I agree 👍🏻 .

technoweenie:
I think I'd rather have a static func like:

```go
lfsapi.VerifySchema(t, lockSchema, reader, &obj)
obj.Lock // now we know obj follows the lockSchema
```

ttaylorr:
Sounds good to me! 👍

PR #1849, opened 2017-01-10
lfsapi/schema: implement `schema` pkg

ttaylorr:
This pull-request implements @technoweenie's idea in https://github.com/git-lfs/git-lfs/pull/1846#issuecomment-271621665 to introduce an `io.Reader` implementation: `schema.Reader`, capable of validating the data it's read against a JSON schema.

This will change our testing methodology a little bit, since instead of creating valid/invalid instances of request and response types, _then_ marshaling them to JSON, we have to valid data as it is sent and received. In other words, we'd be doing validation in `*net/http/httptest.Server`s against `req.Body`, which I think is a more appropriate level to test at, anyway.

---

/cc @git-lfs/core

### DIFF
--- a/lfsapi/schema/reader.go
+++ b/lfsapi/schema/reader.go
@@ -11,9 +11,9 @@ import (
 )
 
 var (
-	// ErrValidationIncomplete is an error returned when `ValidationErr()`
+	// errValidationIncomplete is an error returned when `ValidationErr()`
 	// is called while the reader is still processing data.
-	ErrValidationIncomplete = errors.New("lfsapi/schema: validation incomplete")
+	errValidationIncomplete = errors.New("lfsapi/schema: validation incomplete")
 )
 
 // state represents the set of valid states a `*Reader` (see below) can be in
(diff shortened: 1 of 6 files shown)


lfsapi/schema/reader.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/xeipuuv/gojsonschema"
+)
+
+var (
+	// ErrNoValidationDone is an error returned when `ValidationErr()` is
+	// called before having processed any data.
+	ErrNoValidationDone = errors.New("lfsapi/schema: no validation done")
+	// ErrValidationIncomplete is an error returned when `ValidationErr()`
+	// is called while the reader is still processing data.
+	ErrValidationIncomplete = errors.New("lfsapi/schema: validation incomplete")
```
technoweenie: Do these need to be exported? Also do we need two separate types? I'm not sure an external user would care about the distinction between receiving _no data_ vs _not all of the data_.

ttaylorr: Great call here. I made `ErrValidationIncomplete` be returned in place of `ErrNoValidationDone` in https://github.com/git-lfs/git-lfs/pull/1849/commits/721824deaacd1d64c7a7ff3f4103e80d1fce83ab. Also went ahead and un-exported the `ErrValidationIncomplete` variable in https://github.com/git-lfs/git-lfs/pull/1849/commits/640ca2134129ec4280b03c2bc0f5180ec885e09c.

I think this is ready to ship! 👍

PR #1854, opened 2017-01-12
Backport #1820 for v1.5.x: Fixed file paths on cygwin.

ttaylorr:
This backports #1820.


ttaylorr:
Pulling back on this since semver doesn't permit us to add features/changes to a point release. This will be released in the next non-patch release: either v1.6.0 (unlikely to exist) or v2.0.0.

PR #1857, opened 2017-01-12
release: 1.5.5

ttaylorr:
cc @git-lfs/core


sschuberth:
Just wondering, what will be the first release that has my fixes for storing the version as a resource into the Windows executable?

ttaylorr:
> Just wondering, what will be the first release that has my fixes for storing the version as a resource into the Windows executable?

That would be the major or minor release cut from `master`, so either 1.6.0 or 2.0.0. The point releases are restricted to bug-fixes only.

Issue #1860, opened 2017-01-13
lfs prune hangs on old commits


monitorjbl:
When running `git lfs prune --dry-run` on a repo with commits old enough to exceed the default `lfs.pruneoffsetdays` setting, the command hangs forever. After running `GIT_TRACE git lfs prun --dry-run`, I can see that the last command run by git-lfs is this:

```
git for-each-ref --sort=-committerdate --format=%(refname) %(objectname) %(committerdate:iso) refs
```

Doing a `ps aux | grep git` shows this as well:

```
[tayjones@TAYJONES-M-X2P9:/Users/tayjones]$ ps aux | grep git
tayjones         36376   0.1  0.1 556628824   9364 s003  S+    9:28AM   0:00.10 git-lfs prune --dry-run
tayjones         36389   0.0  0.0  2434840    796 s008  S+    9:28AM   0:00.00 grep git
tayjones         36384   0.0  0.0  2508100   2500 s003  S+    9:28AM   0:00.02 git for-each-ref --sort=-committerdate --format=%(refname) %(objectname) %(committerdate:iso) refs
tayjones         36375   0.0  0.0  2436796   1036 s003  S+    9:28AM   0:00.00 git lfs prune --dry-run
```

I noticed that the `git for-each-ref` command still has stdout open:

```
[tayjones@TAYJONES-M-X2P9:/Users/tayjones]$ lsof -p 36384
COMMAND   PID     USER   FD   TYPE             DEVICE  SIZE/OFF     NODE NAME
git     36384 tayjones  cwd    DIR                1,4       612 33887097 /private/tmp/comp-11000429
git     36384 tayjones  txt    REG                1,4   1877284 27791096 /usr/local/Cellar/git/2.10.1/bin/git
git     36384 tayjones  txt    REG                1,4     98064 33887145 /private/tmp/comp-11000429/.git/objects/pack/pack-6c003df035337371ad9f80b87baf41470a81ced5.idx
git     36384 tayjones  txt    REG                1,4   6321866 33887143 /private/tmp/comp-11000429/.git/objects/pack/pack-6c003df035337371ad9f80b87baf41470a81ced5.pack
git     36384 tayjones  txt    REG                1,4    646624 27407494 /usr/lib/dyld
git     36384 tayjones  txt    REG                1,4 558300650 27414567 /private/var/db/dyld/dyld_shared_cache_x86_64h
git     36384 tayjones    0r   CHR                3,2       0t0      302 /dev/null
[...]

PR #1861, opened 2017-01-13
Fix for #1860: Closing stdout pipe before function return

monitorjbl:


technoweenie:
This looks good, but now I'd really like some tests for it :) We should be able to come up with a prune test to trigger #1860.

@monitorjbl: If you want to write a test, that'd be awesome. If not, we'll write one and then merge this. However, I do ask that you sign our CLA :)

monitorjbl:
@technoweenie I'll give writing a test a shot. I'm gonna have to figure out how to generate "old" commits in a repo if you want an integration test [like this](https://github.com/git-lfs/git-lfs/blob/master/test/test-prune.sh) though. 

I did sign the CLA, I'm not sure why the check didn't update. When I click on it again, it says I've already signed it:

image

technoweenie:
The prune test uses `lfstest-testutils` to create commits from a custom json document. You can create old commits like [this](https://github.com/git-lfs/git-lfs/blob/3572931d1ad6cecca2a94132a5ed248a740fb40f/test/test-prune.sh#L40-L45). 

Are you logging into the cla app as a different github user? That's the only thing I can think of. I'll see if I can peek at the server too.

monitorjbl:
I was on the right github account, but I made the commit on my work laptop so the email address in the commit history was wrong. Looks like changing that did the trick!

monitorjbl:
@technoweenie It doesn't look like your testing library actually sets the dates on the commits. I ran the pre-existing tests in `test/prune-verify.sh` with the following options:

```
KEEPTRASH=1 GIT_LFS_TEST_DIR=/tmp/git-lfs-tests ./test/test-prune.sh 
```

One of the tests in here ("prune verify") looks like its supposed to be making commits that are up to [50 days old](https://github.com/git-lfs/git-lfs/blob/master/test/test-prune.sh#L468). I verified that it outputs JSON that looks like this:

```
[
  {
    "CommitDate":"2016-11-27T15:48:28Z",
    "Files":[
      {"Filename":"file.dat","Size":28, "Data":"Content for commit 1 (prune)"}]
  },
  {
    "CommitDate":"2016-12-07T15:48:28Z",
    "Files":[
      {"Filename":"file.dat","Size":42, "Data":"Content for commit 2 (prune - fail verify)"}]
  },
  {
    "CommitDate":"2016-12-12T15:48:28Z",
    "Files":[
      {"Filename":"file.dat","Size":28, "Data":"Content for commit 3 (prune)"}]
  },
  {
    "CommitDate":"2016-12-22T15:48:28Z",
    "Files":[
      {"Filename":"file.dat","Size":12, "Data":"HEAD content"}]
  }
]
```

So, I went to the repo that was generated for one of the tests and looked at the real git log:

```
$ cd /tmp/git-lfs-tests/test-prune.sh-74530/clone_prune_verify
$ git log

commit 284e764360c0b538a723eb2af6f82e0de51744bb
Author: Git LFS Tests <git-lfs@example.com>
Date:   Mon Jan 16 10:41:10 2017 -0500

    Test commit 3

commit fee71f23920ff0a1ed402ed7f27afe68e45266d0
Author: Git LFS Tests <git-lfs@example.com>
Date:   Mon Jan 16 10:41:10 2017 -0500

    Test commit 2

commit ec75c6e3984033d58c63e1d7a34e5937784ce5de
Author: Git LFS Tests <git-lfs@example.com>
Date:   Mon Jan 16 10:41:10 2017 -0500

    Test commit 1

commit 0459cbd1dc89c3048e65f49cd7469e080858974d
Author: Git LFS Tests <git-lfs@example.com>
Date:   Mon Jan 16 10:41:10 2017 -0500

    Test commit 0
```

The dates are all the current date. I'm running this on OSX, but it looks like you have a bit of shell script in [testhelpers.sh](https://github.com/git-lfs/git-lfs/blob/master/test/testhelpers.sh#L480) to handle this case. 

I'd originally assumed that [bufio.Scanner](https://golang.org/pkg/bufio/#Scanner) was exhausting the stdout pipe on small repos even if the read loop broke early. That may still be the case, but I wonder if the commit dates not actually being set is why this issue was not found before?

monitorjbl:
Well, scratch that @technoweenie . TIL about GIT_AUTHOR_DATE and GIT_COMMITTER_DATE. Looks like the `for-each-ref` call that gets made only looks at the [commit date](https://github.com/git-lfs/git-lfs/blob/master/git/git.go#L441). Still, it would be nice if the commits actually looked correct from `git log`. If you don't mind, I'd like to add the following change to `test/testutils.go` as part of this PR:

```
diff --git a/test/testutils.go b/test/testutils.go
index c2bb3e21..bfae1260 100644
--- a/test/testutils.go
+++ b/test/testutils.go
@@ -298,8 +298,10 @@ func commitAtDate(atDate time.Time, committerName, committerEmail, msg string) e
        // set GIT_COMMITTER_DATE environment var e.g. "Fri Jun 21 20:26:41 2013 +0900"
        if atDate.IsZero() {
                env = append(env, "GIT_COMMITTER_DATE=")
+               env = append(env, "GIT_AUTHOR_DATE=")
        } else {
                env = append(env, fmt.Sprintf("GIT_COMMITTER_DATE=%v", git.FormatGitDate(atDate)))
+               env = append(env, fmt.Sprintf("GIT_AUTHOR_DATE=%v", git.FormatGitDate(atDate)))
        }
        cmd.Env = env
        out, err := cmd.CombinedOutput()
```

monitorjbl:
I would guess the hang happens at [line 434](https://golang.org/src/os/exec/exec.go#L434) before the `c.closeDescriptors(c.closeAfterWait)` gets called. The process can't complete while the pipe still has data. 

I'll get these fixes in, thanks for the pointers! Still somewhat new to Go, so any advice is appreciated.

technoweenie:
> I would guess the hang happens at line 434 before the c.closeDescriptors(c.closeAfterWait) gets called.

Great catch :metal: Ok carry on!

subprocess/cmd.go
```diff
@@ -0,0 +1,110 @@
...
+	Env  []string
+	Dir  string
+
+	Stdin  io.Reader
+	Stdout io.Writer
+	Stderr io.Writer
+
+	Process *os.Process
+
+	cmd        *exec.Cmd
```
monitorjbl: I actually tried this the first time since I'd read about Go's inheritance-like ability, but I got some pointer errors because I didn't realize you had to initialize the child struct like this. Thanks for the tip, I got it to work!

subprocess/cmd.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	if cw.Stdout != nil {
+		cw.cmd.Stdout = cw.Stdout
+	}
+	if cw.Stderr != nil {
+		cw.cmd.Stderr = cw.Stderr
+	}
+}
+
+func (cw *SafeCmd) Wait() error {
```
monitorjbl: I think the `*Pipe()` functions need to be overridden as well. There's no way to get ahold of the pipes created by `exec.Cmd`'s functions.

subprocess/cmd.go
```diff
@@ -0,0 +1,110 @@
...
+	Stdin  io.Reader
+	Stdout io.Writer
+	Stderr io.Writer
+
+	Process *os.Process
+
+	cmd        *exec.Cmd
+	stdoutPipe io.ReadCloser
+	stderrPipe io.ReadCloser
+	stdinPipe  io.WriteCloser
```
monitorjbl: Will do!

subprocess/cmd.go
monitorjbl: I spoke a bit too soon. I need to be able to override the `*Pipe()` functions and call back to the original `exec.Cmd` function to do so. Fortunately, I can still include a reference to the original `exec.Cmd` struct while embedding. Cuts the code down quite a bit.

subprocess/cmd.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	"io"
+	"os"
+	"os/exec"
+)
+
+// Thin wrapper around exec.Cmd. Takes care of pipe shutdown by
+// keeping an internal reference to any created pipes. Whenever
+// Cmd.Wait() is called, all created pipes are closed.
+type SafeCmd struct {
```
monitorjbl: Will do!

subprocess/cmd.go
monitorjbl: Well, that was a rollercoaster of Go education. I think I got it working the way you originally suggested now.

technoweenie:
Nice, this looks much better :metal: Thanks for tackling this. Hope your first Go experience was a good one :)

PR #1863, opened 2017-01-13
locking: reintroduce `Committer` field

ttaylorr:
This pull-request reintroduces the `Committer` field on the `*locking.Lock` type.

This was the [original interface][1] of that type, but was removed in #1732 in favor of `Name` and `Email` top-level fields on the `Lock` type. Instead, let's bring back the original API which:

- has parity with how the data is sent over the network, and...
- is more direct in it's meaning. A lock has a committer, not a name and email. 

This pull-request also comes with some additional benefits:

1. Fix a bug where the `gitserver` was unable to find the fields to de-serialize the committer's name and email, thus making all locks have `""` empty-string owners.
2. Implement `fmt.Stringer` on the `Committer` type, which yields the expected format: `First Last <email@example.com>`.

---

/cc @git-lfs/core 

[1]: https://github.com/git-lfs/git-lfs/blob/7f578a97cef74a2e7fef6723bfc6ce30e512e44b/api/lock_api.go#L113-L114

### DIFF
--- a/locking/api.go
+++ b/locking/api.go
@@ -22,7 +22,7 @@ type lockRequest struct {
 	// `.git/refs/origin/<name>`.
 	LatestRemoteCommit string `json:"latest_remote_commit"`
 	// Committer is the individual that wishes to obtain the lock.
-	Committer committer `json:"committer"`
+	Committer Committer `json:"committer"`
 }
 
 // LockResponse encapsulates the information sent over the API in response to
@@ -188,7 +188,7 @@ func (c *lockClient) Search(remote string, searchReq *lockSearchRequest) (*lockL
 }
 
 // Committer represents a "First Last <email@domain.com>" pair.
-type committer struct {
+type Committer struct {
 	// Name is the name of the individual who would like to obtain the
 	// lock, for instance: "Rick Olson".
 	Name string `json:"name"`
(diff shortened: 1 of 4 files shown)


locking/api.go
```diff
@@ -197,6 +197,12 @@ type committer struct {
 	Email string `json:"email"`
 }
 
-func newCommitter(name, email string) committer {
-	return committer{Name: name, Email: email}
+func NewCommitter(name, email string) Committer {
+	return Committer{Name: name, Email: email}
```
technoweenie: Should this be returning a `*Committer`, since that's the receiver for `String()` below? Or should the func change the receiver to the struct value?

ttaylorr: Good question. The constructor function returns a raw-value `Committer`, since that's how it's embedded in the `Lock` type. `Lock` prefers a `Committer` over a `*Committer`, since the extra data needed to store the pointer of an immutable value is unnecessary.

However, there's no danger of mutable behavior happening in the `(*Committer) String() string` method, so copying the value is unnecessary. Go will automatically take the address to call methods defined on the pointer type, i.e. in a type defined like:

```go
type T struct {}

func (t *T) Foo() {}
```

calling

```go
t := T{}

t.Foo()
```

is equivalent to calling:

```go
t := T{}

(&t).Foo()
```

PR #1868, opened 2017-01-16
Add `--skip-repo` option to `git lfs install` & use in tests

sinbad:
It's bugged me for a while that the integration tests keep creating the `pre-push` hook in my local `git-lfs` working copy repo whenever I run them; it gets much worse when you factor in the new hooks we're starting to add. This happens even if you use `GIT_LFS_TEST_DIR`, if you run the tests from within the working copy.

Seems useful to have a `--skip-repo` option in `git lfs install` which skips the local repo changes, for when you just want to install filters. This is how our tests need to use it. 

I could have just made the tests `cd` to a non-repo folder instead but being explicit seems better.

### DIFF
--- a/commands/command_install.go
+++ b/commands/command_install.go
@@ -13,6 +13,7 @@ var (
 	localInstall      = false
 	systemInstall     = false
 	skipSmudgeInstall = false
+	skipRepoInstall   = false
 )
 
 func installCommand(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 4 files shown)


sinbad:
Names are hard 😄 

I don't really like `--skip-all` because that sounds like it would do nothing, when we're still installing the filters. I had considered options like `--filters-only` but wanted to use `--skip-something` since that was already an established convention. I thought about `--skip-local` but that overlaps too much with the `--local` option. My other primary option was `--skip-hooks`, but I considered that what you really wanted was to leave the repo alone entirely, which includes not creating any new folders in `.git/lfs` etc. Hence `--skip-repo` was the best I could come up with :/

Issue #1884, opened 2017-01-23
Locking Pre-Push hook should be optional


technoweenie:
#1815 added a lock check to the pre-push hook. However, this causes problems for users that use LFS servers that don't implement the locking API (which is all of them since it's not officially shipped yet). This process should definitely be optional.

First, I think LFS should track whether to check locks during pre-push. This could be done through an `lfs.<url>.locks` key, similar to the `lfs.<url>.access` key.

* If `lfs.<url>.skiplocksverify` is NULL: Run the locking check.
  * If it's successful, set `lfs.<url>.skiplocksverify=true`.
  * If it's an `HTTP 501 Not Implemented` response, set `lfs.<url>.skiplocksverify=false`
  * If there's an error, print an error to STDERR, with instructions on disabling this check in case the user's LFS service doesn't implement locking.
* If `lfs.<url>.skiplocksverify=true`: run the locking check.
  * If it's successful, carry on.
  * If it's an `HTTP 501 Not Implemented` response, set `lfs.<url>.skiplocksverify=false`
  * If it errors, halt the push with a detailed error message.
* If `lfs.<url>.skiplocksverify=false`: skip the locking check completely.

The 501 response from an LFS Locking endpoint could be used to tell the client not to bother with a client-side locking check, because the LFS service is handling it itself on the server side.

ttaylorr:
I think this is an excellent approach, and I can take this on for v2.0.0. Starting `lfs.<url>.locks` at `NULL` is a good default, since that allows us to check (and gracefully fail) if locks aren't enabled.

ttaylorr:
Here's a quick stab to unblock pushes temporarily #1885.

technoweenie:
I just updated this proposal to support `501 Not Implemented`, which seems better than `410 Gone`, based on https://github.com/git-lfs/git-lfs/issues/1908#issuecomment-277634262.

ttaylorr:
> I just updated this proposal to support `501 Not Implemented`, which seems better than `410 Gone`, based on #1908 (comment).

Good call, thanks 👍 .

technoweenie:
Updated the key to `lfs.<url>.skiplocksverify`. `lfs.<url>.locks` is just too vague.

ttaylorr:
Closed as of #1951.

Issue #1500, opened 2016-09-02
Include http.extraHeader in requests


Tasteful:
Git-command have the possibility to pass additional [http-headers](https://github.com/git/git/blob/a42d7b6a5b5addcce01f483547ad96825c7a96e0/Documentation/config.txt#L1733) with the request. This is used by [VSTS](https://github.com/Microsoft/vsts-tasks) build agent (Team Foundation Server) to pass OAuth token in an `AUTHORIZATION: bearer`-header when fetching the files from the server instead of using cached credentials. 

Example how the agent is setting the value

```
git config http.https://fabrikam.visualstudio.com/DefaultCollection/Fabrikam/_git/Fabrikam.extraheader "AUTHORIZATION: bearer ********"
```

Is it possible that git-lfs can pass the extra headers with the request?

technoweenie:
Wow, I had no idea about that. Very possible. :+1:

technoweenie:
@Tasteful What is the git remote you're using? Is it `https://fabrikam.visualstudio.com/DefaultCollection/Fabrikam/_git/Fabrikam`? I'm unfamiliar with how VSO urls work.

It's not super clear how the LFS http client would pick out the correct headers. A typical git and lfs server setup should look something like this:

* Git: `https://git-server.com/user/repo.git`
* LFS: `https://git-server.com/user/repo.git/info/lfs`

LFS builds its default url by adding `info/lfs` to the end of the git url: `{git-remote}/info/lfs`. The LFS client uses this as a base to speak to various APIs: `{lfs-url}/objects/batch` or `{lfs-url}/locks`. What git config key should LFS use when talking to `{lfs-url}/objects/batch`?

1. The git remote: `https://git-server.com/user/repo.git`
2. The lfs url: `https://git-server.com/user/repo.git/info/lfs`
3. The full url: `https://git-server.com/user/repo.git/info/lfs/objects/batch`

If we go with option 1, extra headers for your git server will get applied to your LFS api requests. However, it breaks if the user configures this with custom `lfs.url` or `remote.origin.lfsurl` git config settings, since we probably don't want to send the same headers to an LFS service hosted separately from your Git service.

Option 2 works, but we may want to add lfs specific config values like so that Git doesn't try to use them.

I didn't see much in the git config docs that talk about how git uses it. May have to dig into Git code...

Tasteful:
Here is the some environment information, I replaced the actual domainname, project and repository name with placeholders.

git config -l --local
```
[...]

PR #1885, opened 2017-01-23
commands: unhalt locking pre-push checks 

ttaylorr:
This pull-request temporarily makes failed lock searches and pushes containing modifications to un-owned locks non-fatal.

---

/cc @git-lfs/core

### DIFF
--- a/test/test-pre-push.sh
+++ b/test/test-pre-push.sh
@@ -529,11 +529,6 @@ begin_test "pre-push with unowned lock"
     ok="${PIPESTATUS[0]}"
     set -e
 
-    if [ "0" -eq $ok ]; then
-      echo >&2 "ERR: expected push to fail, didn't..."
-      exit 1
-    fi
-
     grep "Unable to push 1 locked file(s)" push.log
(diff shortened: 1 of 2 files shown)


test/test-pre-push.sh
```diff
@@ -529,11 +529,6 @@ begin_test "pre-push with unowned lock"
     ok="${PIPESTATUS[0]}"
     set -e
 
-    if [ "0" -eq $ok ]; then
```
technoweenie: How about removing lines 527, 529, and 530 too? Based on the behavior outlined in https://github.com/git-lfs/git-lfs/issues/1884 for `lfs.<url>.locks` null values, this test should always require a successful push.

ttaylorr: Good catch, I totally missed that one 🤘 35b9525.

ttaylorr:
> Confirmed via live test:

Nice. I was thinking about adding a header that would "disable" the locking endpoints, but we don't have a great mechanism to send custom headers pending #1500.

PR #1886, opened 2017-01-24
Fix clash between progress meter and credential helper

### LINKED ISSUE #847: credential output hidden while transferring files
When Git LFS is uploading or downloading a batch of files, it shows some progress output. It hides any other output from the command, like `git-credentials` asking for your username and password. Here's what I see (scroll to the right to see the username prompt).

```
Git LFS: (1 of 4 files) 16.64 KB / 818.94 KB                                                                                                                                                                                                                                                                                                                               Username for 'https://github-cloud.s3.amazonaws.com':
Git LFS: (1 of 4 files) 101.27 KB / 818.94 KB
```

If you can't see the prompt, then git lfs appears to hang for no apparent reason. The only fix is to setup a credential cacher right now.

technoweenie:
This is an attempt to fix #847. It pauses the progress meter before making the batch api request, and then starts it when it passing the batch api response to the transfer adapters. This works because:

* `(progress.Meter) Start()`, `Finish()`, and `Pause()` internally use channels and `atomic/sync`. No need for that `sync.Once` in `tq.TransferQueue`.
* The `tq.TransferQueue` only transfers items from a single batch concurrently. The sequence (Batch API -> _n_ transfer adapter workers) runs sequentially, so that the next batch (if needed) can process the retries from the previous batch.

Unfortunately, there isn't really a way to test this, since it requires interactive input.  Here's an example of this PR working (with a tiny hack to change the batch size to 2):

```
$ git lfs fetch --all 2>&1
✔ 9 objects found
Fetching objects...
Username for 'https://github.com': technoweenie
Password for 'https://technoweenie@github.com':
Git LFS: (0 of 9 files) 0 B / 1.03 MB                                                                                                                                                                                                        Username for 'https://github.com': technoweenie
Password for 'https://technoweenie@github.com':
Git LFS: (3 of 9 files) 356.27 KB / 1.03 MB                                                                                                                                                                                                  Username for 'https://github.com': technoweenie
Password for 'https://technoweenie@github.com':
Git LFS: (5 of 9 files) 886.59 KB / 1.03 MB                                                                                                                                                                                                  Username for 'https://github.com': technoweenie
Password for 'https://technoweenie@github.com':
Git LFS: (7 of 9 files) 948.33 KB / 1.03 MB                                                                                                                                                                                                  Username for 'https://github.com': technoweenie
Password for 'https://technoweenie@github.com':
Git LFS: (9 of 9 files) 1000.70 KB / 1.03 MB
```

Here's the same, without pausing:

```
$ git lfs fetch --all 2>&1
✔ 9 objects found
Fetching objects...
Username for 'https://github.com': technoweenie
Password for 'https://technoweenie@github.com':
Git LFS: (1 of 9 files) 0 B / 1.03 MB                                                                                                                                                                                                        UGit LFS: (2 of 9 files) 0 B / 1.03 MB
```

### DIFF
--- a/progress/meter.go
+++ b/progress/meter.go
@@ -106,13 +106,14 @@ func NewMeter(options ...meterOption) *ProgressMeter {
 
 // Start begins sending status updates to the optional log file, and stdout.
 func (p *ProgressMeter) Start() {
-	if atomic.SwapInt32(&p.started, 1) == 0 {
+	if atomic.CompareAndSwapInt32(&p.started, 0, 1) {
 		go p.writer()
 	}
 }
 
+// Pause stops sending status updates temporarily, until Start() is called again.
 func (p *ProgressMeter) Pause() {
-	if atomic.SwapInt32(&p.started, 0) == 1 {
+	if atomic.CompareAndSwapInt32(&p.started, 1, 0) {
 		p.finished <- true
 	}
 }
--- a/progress/meter.go
+++ b/progress/meter.go
@@ -111,6 +111,12 @@ func (p *ProgressMeter) Start() {
 	}
 }
 
+func (p *ProgressMeter) Pause() {
+	if atomic.SwapInt32(&p.started, 0) == 1 {
+		p.finished <- true
+	}
+}
+
 // Add tells the progress meter that a single file of the given size will
 // possibly be transferred. If a file doesn't need to be transferred for some
 // reason, be sure to call Skip(int64) with the same size.
--- a/tq/transfer_queue.go
+++ b/tq/transfer_queue.go
@@ -105,7 +105,6 @@ type TransferQueue struct {
 	errorc        chan error // Channel for processing errors
 	watchers      []chan string
 	trMutex       *sync.Mutex
-	startProgress sync.Once
 	collectorWait sync.WaitGroup
 	errorwait     sync.WaitGroup
 	// wait is used to keep track of pending transfers. It is incremented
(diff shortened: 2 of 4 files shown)


progress/meter.go
```diff
@@ -111,6 +111,12 @@ func (p *ProgressMeter) Start() {
 	}
 }
 
+func (p *ProgressMeter) Pause() {
+	if atomic.SwapInt32(&p.started, 0) == 1 {
```
ttaylorr: Nit: Using `atomic.CompareAndSwapInt32(&p.started, 1, 0)` accomplishes the same behavior.

tq/transfer_queue.go
```diff
@@ -307,7 +307,7 @@ func (q *TransferQueue) enqueueAndCollectRetriesFor(batch batch) (batch, error)
 	}
 
 	q.useAdapter(bRes.TransferAdapterName)
-	q.startProgress.Do(q.meter.Start)
```
ttaylorr: Before we resume the progress meter, we have two opportunities to bail out early: on L302 and L306. I think both of those are fine, since we'd exit this function early, and the progress meter would get resumed the next time a new batch was enqueued. There'd be no issues with calling `Pause()` twice, since there's a `CompareAndSwap<T>` operation guarding the channel send into `p.finished`.

@technoweenie would you mind double checking my thinking on this?

progress/meter.go
technoweenie: Ah good call. This was copy pasta from `Start()`, but I'll tweak both.

tq/transfer_queue.go
technoweenie: Seems legit. I went ahead and added some tests in the `progress` package to assert that `Start()` and `Pause()` can be called multiple times.

technoweenie:
Well, since I made `Start()` and `Pause()` safe for multiple calls, I went ahead and did the same for `Finish()`. Also, calling `Finish()` means that `Start()` and `Pause()` can never be called again.

technoweenie:
Welp, those changes did not make things happier. I backed out of them in this PR.

ttaylorr:
> Welp, those changes did not make things happier. I backed out of them in this PR.

👍

PR #1887, opened 2017-01-24
Teach uninstall cmd about --local and --system

### LINKED ISSUE #1753: No system level 'uninstall' analogous to system level 'install'
You can run 
`git lfs install --system`

but

`git lfs uninstall --system`
yields
`Sorry, no usage text found for "uninstall"

Error: unknown flag: --system`

technoweenie:
Fixes #1753

### DIFF
--- a/lfs/attribute.go
+++ b/lfs/attribute.go
@@ -106,8 +106,6 @@ func (a *Attribute) Uninstall(opt InstallOptions) {
 	} else if opt.System {
 		git.Config.UnsetSystemSection(a.Section)
 	} else {
-		// uninstall from both system and global
-		git.Config.UnsetSystemSection(a.Section)
 		git.Config.UnsetGlobalSection(a.Section)
 	}
 }
--- a/lfs/attribute.go
+++ b/lfs/attribute.go
@@ -100,10 +100,16 @@ func (a *Attribute) set(key, value string, upgradeables []string, opt InstallOpt
 }
 
 // Uninstall removes all properties in the path of this property.
-func (a *Attribute) Uninstall() {
-	// uninstall from both system and global
-	git.Config.UnsetSystemSection(a.Section)
-	git.Config.UnsetGlobalSection(a.Section)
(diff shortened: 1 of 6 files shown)


lfs/attribute.go
```diff
@@ -100,10 +100,16 @@ func (a *Attribute) set(key, value string, upgradeables []string, opt InstallOpt
...
 
 // Uninstall removes all properties in the path of this property.
-func (a *Attribute) Uninstall() {
-	// uninstall from both system and global
-	git.Config.UnsetSystemSection(a.Section)
-	git.Config.UnsetGlobalSection(a.Section)
+func (a *Attribute) Uninstall(opt InstallOptions) {
+	if opt.Local {
+		git.Config.UnsetLocalSection(a.Section)
+	} else if opt.System {
```
ttaylorr: Could demote this to a bare `if` and remove the `else` block below, I think.

technoweenie: If the existing behavior can't change, you could rewrite this to:

```go
	if opt.Local {
		git.Config.UnsetLocalSection(a.Section)
		return
	} else if !opt.System {
		git.Config.UnsetGlobalSection(a.Section)
	}
	git.Config.UnsetSystemSection(a.Section)
```

It's just 1 fewer line (2 if you count the comment), but also harder to read, IMO.

Now, if we want this to affect the system config only if `--system` is set, then it'd change to:

```go
	if opt.Local {
		git.Config.UnsetLocalSection(a.Section)
		return
	} else if opt.System {
		git.Config.UnsetSystemSection(a.Section)
		return
	}
	git.Config.UnsetGlobalSection(a.Section)
```

However, I'd still prefer the current multi if block:

```go
	if opt.Local {
		git.Config.UnsetLocalSection(a.Section)
	} else if opt.System {
		git.Config.UnsetSystemSection(a.Section)
	} else {
		git.Config.UnsetGlobalSection(a.Section)
	}
```

PR #1890, opened 2017-01-25
Fix lfsapi inconsistencies

technoweenie:
I noticed a few inconsistencies with the completely rewritten API code:

1. Requests from Git LFS aren't sending a `Content-Length` request header.
2. ALL ssh auth requests send the `upload` operation, which interferes with the SSH command `ssh  user@githost.com git-lfs-authenticate user/repo {operation}`.

This PR fixes both.

### DIFF
--- a/lfsapi/endpoint.go
+++ b/lfsapi/endpoint.go
@@ -15,6 +15,20 @@ type Endpoint struct {
 	SshUserAndHost string
 	SshPath        string
 	SshPort        string
+	Operation      string
+}
+
+func endpointOperation(e Endpoint, method string) string {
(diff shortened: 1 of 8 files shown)


lfsapi/endpoint.go
```diff
@@ -15,6 +15,20 @@ type Endpoint struct {
 	SshUserAndHost string
 	SshPath        string
 	SshPort        string
+	Operation      string
+}
+
+func endpointOperation(e Endpoint, method string) string {
```
ttaylorr: I really like how concise this method turned out 👍

PR #1869, opened 2017-01-16
support for client certificates

schuCriX:
allow for sslKey and sslCert in gitconfig http subsection

PR #1893, opened 2017-01-25
support for client certificates pt 2

technoweenie:
This is a continuation of #1869. It ignores a common error from git users on the mac when trying to send ssl client certs. This also lets me run the circleci builds, since they didn't on #1869 for some reason.

### DIFF
--- a/test/test-clone.sh
+++ b/test/test-clone.sh
@@ -140,6 +140,61 @@ begin_test "cloneSSL"
 )
 end_test
 
+begin_test "clone ClientCert"
+(
+
+  set -e
+  reponame="test-cloneClientCert"
+  setup_remote_repo "$reponame"
+  clone_repo_clientcert "$reponame" "$reponame"
+
+  git lfs track "*.dat" 2>&1 | tee track.log
+  grep "Tracking \*.dat" track.log
+
+  # generate some test data & commits with random LFS data
+  echo "[
+  {
+    \"CommitDate\":\"$(get_date -5d)\",
+    \"Files\":[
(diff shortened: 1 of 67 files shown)


test/test-clone.sh
```diff
@@ -144,12 +144,14 @@ begin_test "clone ClientCert"
   set -e
   reponame="test-cloneClientCert"
   setup_remote_repo "$reponame"
+  pwd
   out="$(clone_repo_clientcert "$reponame" "$reponame")"
```
ttaylorr: My hunch is that the directory change isn't happening since `clone_repo_clientcert` is being called from a sub-shell, `$()`.

If `clone_repo_clientcert` doesn't generate it's own log files, I'd just `| tee out.log`, and change the grep to:

```sh
if [ "0" -ne "$(grep -c 'client-cert-mac-openssl' < out.log)" ]; then
  # ...
fi
```

technoweenie: Ah yes, the sub-shell explains it. Thanks.

technoweenie: I can't use `tee` exactly, since it swallows errors. If `clone_repo_clientcert()` returns an error, I want the test to halt.

ttaylorr: > I can't use `tee` exactly, since it swallows errors.

You should still be able to capture non-zero exit codes with `${PIPESTATUS[0]}`:

```sh
thing_that_might_fail 2>&1 | tee foo.log
ok="${PIPESTATUS[0]}"
```

But I don't mind either way which approach you take here 👍

technoweenie:
Forcing a merge here, since #1869 was approved.

Issue #1844, opened 2017-01-09
Proposal: locking next steps - serialising changes on lock


sinbad:
Assuming that #1822 (read-only unlocked files) is now ready to go, the next thing I planned to tackle in locking was this part of my original proposal:

> 2. User tries to lock a file. This is only allowed if:
>    * The file is not already locked by anyone else, AND
>    * One of the following are true:
>        * The user has, or agrees to check out, a descendant of the latest commit
>           that was made for that file, whatever branch that was on, OR
>        * The user stays on their current commit but resets the locked file to the
>           state of the latest commit (making it modified locally, and
>           also cherry-picking changes for that file in practice).

This has implications for the API, the tasks the server needs to do on lock, and some interactive prompts (together with options to drive them non-interactively) from the git-lfs client where decisions must be made.

My proposed approach is that:

1. The client must supply to the server, on requesting a lock:
    * `CommitSHA`: the git commit SHA of HEAD where the lock is requested
    * `ContentSHA`: the SHA-256 of the current *content* of the file which the user will be modifying
    * `OverrideLatest`: an boolean indicating that the user has decided to lock the file at non-latest content, thus creating a future merge conflict (important if the user wants to lock anyway and not serialise changes across all branches)
2. The server must implement the checks:
   * That no other user has this file locked currently
       * If they have, reject the lock with details of the current lock
   * If `OverrideLatest` is *not* set:
       * Check whether the latest change to this file on **any** branch is different to `ContentSHA`
           * If it is, reject the lock & return:
               * `CommitSHALatest`: the commit SHA of the latest change
               * `Branches`: list of remote branches which include that SHA (optional; could make client fetch & derive that for itself)
               * `ContentSHALatest`: the content SHA of the latest change, and 
               * `IsDescendant`: whether `CommitSHALatest` is a descendant of client's `CommitSHA`
           * Note that the server has to check this and not the client since the change is likely to have come from other users and client may not have all the necessary info until pulling
3. On receiving a lock rejection the client should:
    * If it's because the file is currently locked by someone else, just exit with an error & details
        * Later on that could be overridden with a `--force` option to break unattended locks
    * If it's because of a content conflict:
        * If the server's `IsDescendant` response is true, suggest pulling one of the `Branches` before locking & exit with error (or if `Branches` is cut from the response, suggest fetching & deriving that locally)
        * If `CommitSHALatest` is unknown locally, indicate that `git fetch` is needed before proceeding
[...]

PR #1896, opened 2017-01-30
Check that files are not uncommitted before unlock

sinbad:
A user should not have any uncommitted changes in their working copy when unlocking a file, since to relinquish the lock means that they've either finished making changes, or they don't want to keep their changes. It must be assumed that when relinquishing the lock other users are free to lock & begin making changes immediately which would clash with these outstanding changes.

`--force` overrides this check but with a warning.

This is only part of the overall solution; in future we'll also want to ensure they don't have any unpushed changes before allowing the unlock since those will have the same effect. However this will form part of #1844 which is a more complex proposal. This is an easy win in comparison and easy to understand.

### DIFF
--- a/commands/command_unlock.go
+++ b/commands/command_unlock.go
@@ -33,7 +33,8 @@ func unlockCommand(cmd *cobra.Command, args []string) {
 			Exit("Unable to determine path: %v", err.Error())
 		}
 
-		unlockCheckFileStatus(path)
+		// This call can early-out
+		unlockAbortIfFileModified(path)
 
 		err = lockClient.UnlockFile(path, unlockCmdFlags.Force)
 		if err != nil {
@@ -41,7 +42,8 @@ func unlockCommand(cmd *cobra.Command, args []string) {
 		}
 	} else if unlockCmdFlags.Id != "" {
 
-		unlockCheckFileStatusById(unlockCmdFlags.Id, lockClient)
+		// This call can early-out
+		unlockAbortIfFileModifiedById(unlockCmdFlags.Id, lockClient)
 
 		err := lockClient.UnlockFileById(unlockCmdFlags.Id, unlockCmdFlags.Force)
 		if err != nil {
@@ -62,7 +64,7 @@ func unlockCommand(cmd *cobra.Command, args []string) {
 	Print("'%s' was unlocked", args[0])
 }
 
-func unlockCheckFileStatus(path string) {
+func unlockAbortIfFileModified(path string) {
 	modified, err := git.IsFileModified(path)
 
 	if err != nil {
@@ -80,7 +82,7 @@ func unlockCheckFileStatus(path string) {
 	}
 }
 
-func unlockCheckFileStatusById(id string, lockClient *locking.Client) {
+func unlockAbortIfFileModifiedById(id string, lockClient *locking.Client) {
 	// Get the path so we can check the status
 	filter := map[string]string{"id": id}
 	// try local cache first
@@ -91,7 +93,7 @@ func unlockCheckFileStatusById(id string, lockClient *locking.Client) {
 	}
 
 	if len(locks) > 0 {
-		unlockCheckFileStatus(locks[0].Path)
+		unlockAbortIfFileModified(locks[0].Path)
 	}
 
 	// Don't block if we can't determine the path, may be cleaning up old data
--- a/commands/command_unlock.go
+++ b/commands/command_unlock.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"os"
 
+	"github.com/git-lfs/git-lfs/git"
+	"github.com/git-lfs/git-lfs/locking"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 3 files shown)


git/git.go
```diff
@@ -1092,3 +1092,43 @@ func GetFilesChanged(from, to string) ([]string, error) {
...
+	}
+	scanner := bufio.NewScanner(outp)
+	for scanner.Scan() {
+		line := scanner.Text()
+		// Porcelain format is "<I><W> <filename>"
+		// Where <I> = index status, <W> = working copy status
+		if len(line) > 3 {
+			// Double-check even though should be only match
+			if strings.TrimSpace(line[3:]) == filepath {
+				return true, nil
```
technoweenie: This won't call `cmd.Wait()`. I think that should go in a `defer` right before the scanner is initialized.

commands/command_unlock.go
```diff
@@ -31,11 +33,16 @@ func unlockCommand(cmd *cobra.Command, args []string) {
 			Exit("Unable to determine path: %v", err.Error())
 		}
 
+		unlockCheckFileStatus(path)
```
ttaylorr: I'm not a huge fan of the absence of a return value here. That's not to say that all functions need to return something, but the fact this function `Exit()`s from within forces the reader to inspect the body of the function to know that the calling code may not progress beyond this point.

I was originally going to suggest:

```go
if err := unlockCheckFileStatus(...); err != nil {
         // err -> nil if file unmodified, or ErrFileIsModified
}
```

but that turned into:

```go
if m, err := unlockCheckFileStatus(...); err != nil {
         // handle error
} else {
        if m { Exit(...) }
        else { Error(...) }
}
```

to handle differentiating between an error in getting the modified status of a file, and whether or not the file is actually modified. One way you can do that is `*bool`, but I think the greater truth that this function is unnecessary.

I'd suggest either:

- Removing the function and promoting it's body inline
- Adding a comment suggesting that calling this function _can_ bail out.

commands/command_unlock.go
```diff
@@ -55,6 +62,41 @@ func unlockCommand(cmd *cobra.Command, args []string) {
...
+
+	}
+}
+
+func unlockCheckFileStatusById(id string, lockClient *locking.Client) {
+	// Get the path so we can check the status
+	filter := map[string]string{"id": id}
+	// try local cache first
+	locks, _ := lockClient.SearchLocks(filter, 0, true)
+	if len(locks) == 0 {
```
ttaylorr: Noting for the future: if locking is disabled for this host (#1884) we shouldn't check.

commands/command_unlock.go
```diff
@@ -55,6 +62,41 @@ func unlockCommand(cmd *cobra.Command, args []string) {
...
+		// Fall back on calling server
+		locks, _ = lockClient.SearchLocks(filter, 0, false)
+	}
+
+	if len(locks) > 0 {
+		unlockCheckFileStatus(locks[0].Path)
+	}
+
+	// Don't block if we can't determine the path, may be cleaning up old data
```
ttaylorr: Nit: I'd prefer to change this into:

```go
if len(locks) == 0 {
        // Don't block if we can't determine the path, ...
        return
}

unlockCheckFileStatus(locks[0].Path)
```

git/git.go
```diff
@@ -1092,3 +1092,43 @@ func GetFilesChanged(from, to string) ([]string, error) {
...
+	cmd := subprocess.ExecCommand("git", args...)
+	outp, err := cmd.StdoutPipe()
+	if err != nil {
+		return false, fmt.Errorf("Failed to call git status: %v", err)
+	}
+	if err := cmd.Start(); err != nil {
+		return false, fmt.Errorf("Failed to start git status: %v", err)
+	}
+	scanner := bufio.NewScanner(outp)
+	for scanner.Scan() {
```
ttaylorr: We should be able to inline this and not leak `scanner` as:

```go
for scanner := bufio.NewScanner(outp); scanner.Scan(); {
        // ...
}
```

git/git.go
```diff
@@ -1092,3 +1092,43 @@ func GetFilesChanged(from, to string) ([]string, error) {
...
+		"-c", "core.quotepath=false", // handle special chars in filenames
+		"status",
+		"--porcelain",
+		"--", // separator in case filename ambiguous
+		filepath,
+	}
+	cmd := subprocess.ExecCommand("git", args...)
+	outp, err := cmd.StdoutPipe()
+	if err != nil {
+		return false, fmt.Errorf("Failed to call git status: %v", err)
```
ttaylorr: Nit: this behavior should now be accomplished from: `errors.Wrap(msg, err)`.

`errors` being: `github.com/git-lfs/git-lfs/errors`.

git/git.go
ttaylorr: Would it be possible to just terminate running the command here instead?

commands/command_unlock.go
technoweenie: #1884 is specifically about the upcoming `/locks/verify` endpoint. If your server doesn't support LFS, I'd expect running `lfs unlock` to error.

git/git.go
technoweenie: Sure, I suppose in this case we actually care about the error, vs most `defer foo.Close()` calls, where we're ok to ignore any errors.

sinbad: Right, I don't think this will actually do any harm but it is a bit messy. The problem with `defer` is that it won't catch & return errors which can sometimes only happen on `Wait` for commands. I'll tidy it more specifically.

commands/command_unlock.go
sinbad: Yeah I originally had this function returning but it meant duplicating the outcomes (it's needed in a derived form when checking by Id), and also duplicating the override of that with `--force`. We have form in the `commands` package of `Exit`ing inside functions so I thought this was the simplest in the round. I'll comment it though and maybe change the name to `unlockAbortIfFileUncommitted`.

git/git.go
sinbad: I've imported as `lfserrors` since `errors` is already imported from standard Go in this file elsewhere.

technoweenie: I don't think that's necessary, the `git-lfs/errors` package is a suitable replacement for stdlib `errors`.

technoweenie:
Thanks for the updates, this looks good :metal:

ttaylorr:
👍

PR #1904, opened 2017-02-01
Test per repo locks

technoweenie:
This updates the lfs test gitserver to store locks per repo, instead of in a global slice.

### DIFF
--- a/test/cmd/lfstest-gitserver.go
+++ b/test/cmd/lfstest-gitserver.go
@@ -105,14 +105,6 @@ func main() {
 			return
 		}
 
-		if strings.Contains(r.URL.Path, "/info/lfs/locks") {
-			if !skipIfBadAuth(w, r, id, ntlmSession) {
-				locksHandler(w, r)
-			}
-
-			return
-		}
-
 		if strings.Contains(r.URL.Path, "/info/lfs") {
 			if !skipIfBadAuth(w, r, id, ntlmSession) {
 				lfsHandler(w, r, id)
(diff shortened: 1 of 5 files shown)


test/cmd/lfstest-gitserver.go
```diff
@@ -838,6 +859,13 @@ func (c LocksByCreatedAt) Swap(i, j int)      { c[i], c[j] = c[j], c[i] }
 var lockRe = regexp.MustCompile(`/locks/?$`)
 
 func locksHandler(w http.ResponseWriter, r *http.Request) {
+	repo, err := repoFromLfsUrl(r.URL.Path)
```
ttaylorr: Nit: we already have this data in the `lfsHandler`, which is what calls this. I would tend to prefer passing that in as a last parameter, though this does make this method less of a "handler" in that it takes more than just a (`http.ResponseWriter`, `*http.Request`).

test/cmd/lfstest-gitserver.go
```diff
@@ -809,24 +809,45 @@ type LockList struct {
...
+}
+
+func getLocks(repo string) []Lock {
+	lmu.RLock()
+	defer lmu.RUnlock()
 
-	locks = append(locks, l...)
+	found := repoLocks[repo]
+	locks := make([]Lock, len(found))
+	for i, l := range found {
```
ttaylorr: Other nit: `copy` may be more explicit here.

test/cmd/lfstest-gitserver.go
technoweenie: Good call. There was one spot that `locksHandler()` was called outside of `lfsHandler()`, but that was the old global `/locks` route that's unneeded. 🔥

test/cmd/lfstest-gitserver.go
technoweenie: I renamed it to `cp` since `copy()` is a builtin.

ttaylorr: Sorry, I meant using the `copy` builtin would be appropriate:

```go
locks := make([]Lock, len(found)
copy(locks, found)
```

Issue #1891, opened 2017-01-25
Proposal: Custom API endpoint for client side lock verification


technoweenie:
Git LFS in `master` verifies that pushes are not updating locked files based on work in #1815.

1. Call [`GET /locks`](https://github.com/git-lfs/git-lfs/blob/e029224251e2d8b134ee92f3cc00e001fdcbfb40/docs/proposals/locking_api.md#get-locks) to get all locks from the server.
2. [Sort locks into `owned` and `unowned` buckets](https://github.com/git-lfs/git-lfs/pull/1815/files#diff-3cff4ed9c5b60b5b81240c88cd74e404R98) based on the `user.name` and `user.email` git config values.

While this is fine for a prototype, it's not good enough for production.

My proposal is to introduce a new endpoint specifically for these client side checks:

```
> POST /locks/verify
> Accept: application/vnd.git-lfs+json
> Authorization: Basic
> Content-Type: application/vnd.git-lfs+json
>
> {
> }

< HTTP/1.1 200 Ok
< Content-Type: application/vnd.git-lfs+json
<
< {
<   "ours": [
<     // full "lock" properties omitted to save space
<     { "id": "1", "path": "foo/bar" }
<   ],
<   "theirs": [
<     { "id": "2", "path": "foo/baz" }
<   ],
<   next_cursor: "optional-next-id",
<   error: "optional error"
< }
```

This is the bare minimum that could replace the existing code. It leaves us a lot of flexibility to add properties in the future. Maybe we need to send other details in the request to the server to support multi-branch workflows, for example.

The immediate benefit here is that it puts the burden of sorting the locks into owned/unowned buckets is on the server, which is able to verify user credentials.

Also, any LFS server that does this validation on the server can return a `410 Gone` response for just this `/locks/verify` endpoint. This would disable client side lock verification if it's unnecessary, based on this proposal: Locking Pre-Push hook should be optional #1884.

ttaylorr:
> The immediate benefit here is that it puts the burden of sorting the locks into owned/unowned buckets is on the server, which is able to verify user credentials.

Definitely on board with this -- offloading that work to an authoritative source of data is a huge win. Here are a few of my thoughts:

Some high-level questions:

1. Really happy that we have opportunity to grow with this endpoint. I think this is the right approach to be taking.

Some implementation questions:

[...]

PR #1905, opened 2017-02-01
Locking/verify api

technoweenie:
This implements https://github.com/git-lfs/git-lfs/issues/1891

TODO:

* [x] Ship https://github.com/git-lfs/git-lfs/pull/1904
* [x] Write integration tests

Hit me up with any review comments on the go code though. I'm not crazy about the name `VerifiableLocks()`, and am open to suggestions there.

### DIFF
--- a/commands/uploader.go
+++ b/commands/uploader.go
@@ -24,10 +24,17 @@ type uploadContext struct {
 	committerName  string
 	committerEmail string
 
-	locks          map[string]locking.Lock
 	trackedLocksMu *sync.Mutex
-	ownedLocks     []locking.Lock
-	unownedLocks   []locking.Lock
+
+	// ALL verifiable locks
+	ourLocks   map[string]locking.Lock
+	theirLocks map[string]locking.Lock
+
+	// locks from ourLocks that were modified in this push
+	ownedLocks []locking.Lock
+
+	// locks from theirLocks that were modified in this push
+	unownedLocks []locking.Lock
 }
 
 func newUploadContext(remote string, dryRun bool) *uploadContext {
@@ -38,7 +45,8 @@ func newUploadContext(remote string, dryRun bool) *uploadContext {
 		Manifest:       getTransferManifest(),
 		DryRun:         dryRun,
 		uploadedOids:   tools.NewStringSet(),
-		locks:          make(map[string]locking.Lock),
+		ourLocks:       make(map[string]locking.Lock),
+		theirLocks:     make(map[string]locking.Lock),
 		trackedLocksMu: new(sync.Mutex),
 	}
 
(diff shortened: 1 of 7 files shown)


locking/api.go
```diff
@@ -190,6 +190,61 @@ func (c *lockClient) Search(remote string, searchReq *lockSearchRequest) (*lockL
...
+	// Servers must return a list of locks sorted in reverse chronological
+	// order, so the Cursor provides a consistent method of viewing all
+	// locks, even if more were created between two requests.
+	Cursor string `json:"cursor,omitempty"`
+	// Limit is the maximum number of locks to return in a single page.
+	Limit int `json:"limit,omitempty"`
+}
+
+// lockVerifiableList encapsulates a set of Locks to verify a Git push.
+type lockVerifiableList struct {
```
ttaylorr: Perhaps: `lockVerifiableResponse` (to be consistent with `<T>Request`)?

technoweenie: I named it based on `lockList`. Most of the names are based on the existing names for the search api call :)

ttaylorr: Sounds good, thanks!

technoweenie:
Ripping the `wip` bandage off, this is ready to go.

The test server's `locksHandler` now implements the `/locks/verify` endpoint. It considers any lock with `theirs` in the path to be one of "their" locks, while all others are "ours". It also uses some magic repo names to return 404 and 410 already, once https://github.com/git-lfs/git-lfs/issues/1884 is implemented.

I'm not crazy about using magic strings like this, but it's only in the test server. Our LFS client doesn't care why a server decides to include a lock in `ours` or `theirs`, just that it does.

commands/uploader.go
```diff
@@ -96,17 +107,16 @@ func (c *uploadContext) prepareUpload(unfiltered ...*lfs.WrappedPointer) (*tq.Tr
 		// current committer.
 		var canUpload bool = true
 
-		if lock, ok := c.locks[p.Name]; ok {
-			owned := lock.Committer.Name == c.committerName &&
-				lock.Committer.Email == c.committerEmail
+		if lock, ok := c.theirLocks[p.Name]; ok {
```
ttaylorr: Nit: I would call this:

```go
if theirs, ok := c.theirLocks[p.Name]; ok {
        // ...
}
```

to be consistent with how we refer to it in the API.

commands/uploader.go
```diff
@@ -24,10 +24,17 @@ type uploadContext struct {
...
 	trackedLocksMu *sync.Mutex
-	ownedLocks     []locking.Lock
-	unownedLocks   []locking.Lock
+
+	// ALL verifiable locks
+	ourLocks   map[string]locking.Lock
+	theirLocks map[string]locking.Lock
+
+	// locks from ourLocks that were modified in this push
+	ownedLocks []locking.Lock
```
ttaylorr: Since this (and `unownedLocks`) are always subsets of `ourLocks` and `theirLocks` respectively, do you think that we could wrap the `locking.Lock` type here and store less copies of it in memory?

I was thinking:

```go
type owner byte

const (
        ownerAbsent owner = iota
        ownerOwned
        ownerUnowned
)

type lockWithOwner struct {
        locking.Lock

        State owner
}

ourLocks map[string]lockWithOwner
```

technoweenie: I'd rather keep it simple. I'm not sure users will be creating enough locks to matter.

PR #1909, opened 2017-02-02
Locking/docs

technoweenie:
Here's the first draft of the locking api docs. I made a few drastic changes, but hopefully before the pitchforks need to come out :) 

1. Killed `lock.remote`, which is completely unnecessary.
2. Killed `lock.latest_remote_commit` and `lock.commit_sha` until we completely figure out how they should be used.
3. No more `commit_needed` response when trying to create a lock, since `lock.latest_remote_commit` is no longer part of the request.
4. The unlock request body no longer sends an `id` property, since it's already in the URL.
5. The big one: the `error` property was renamed to `message` to be more consistent with the Batch API error responses.
6. Remove `committer` from lock creations
7. Rename `committer` to `owner` when fetching locks. Remove `committer.email` too.

In addition to renaming `error` to `message`, this adds the optional `request_id` and `documentation_url` error properties that the Batch API error responses use. This extra info can help with troubleshooting. Hopefully a locking service will include their own `documentation_url` for user assistance, before users try posting an issue here. It's very possible that locking issues can be service related, and not client bugs.

If these changes look fine, I'll make the tweaks in the code before waiting on final reviews and merging.

* [x] Finish and merge #1905 (this PR documents that API)
* [x] Finalize API details
* [x] Update client code for the changes

### DIFF
--- a/docs/api/batch.md
+++ b/docs/api/batch.md
@@ -1,6 +1,6 @@
 # Git LFS Batch API
 
-Added: v1.0
+Added: v0.6
 
 The Batch API is used to request the ability to transfer LFS objects with the
 LFS server. The Batch URL is built by adding `/objects/batch` to the LFS server
--- a/docs/api/batch.md
+++ b/docs/api/batch.md
@@ -1,5 +1,7 @@
 # Git LFS Batch API
 
+Added: v1.0
+
 The Batch API is used to request the ability to transfer LFS objects with the
 LFS server. The Batch URL is built by adding `/objects/batch` to the LFS server
 URL.
@@ -172,7 +174,7 @@ errors.
 
 {
   "message": "Not found",
-  "documentation_url": "https://git-lfs-server.com/docs/errors",
+  "documentation_url": "https://lfs-server.com/docs/errors",
   "request_id": "123"
 }
 ```
@@ -189,7 +191,7 @@ a custom header key so it does not trigger password prompts in browsers.
 
 {
   "message": "Credentials needed",
-  "documentation_url": "https://git-lfs-server.com/docs/errors",
+  "documentation_url": "https://lfs-server.com/docs/errors",
   "request_id": "123"
 }
 ```
--- a/docs/api/locking.md
+++ b/docs/api/locking.md
@@ -1,0 +1,449 @@
+# Git LFS File Locking API
+
+Added: v2.0
+
+The File Locking API is used to create, list, and delete locks, as well as
+verify that locks are respected in Git pushes. The locking URLs are built
+by adding a suffix to the LFS Server URL.
+
+Git remote: https://git-server.com/foo/bar  
+LFS server: https://git-server.com/foo/bar.git/info/lfs  
+Locks API: https://git-server.com/foo/bar.git/info/lfs/locks
+
+See the [Server Discovery doc](./server-discovery.md) for more info on how LFS
+builds the LFS server URL.
+
+All File Locking requests require the following HTTP headers:
+
+    Accept: application/vnd.git-lfs+json
+    Content-Type: application/vnd.git-lfs+json
+
+See the [Authentication doc](./authentication.md) for more info on how LFS
+gets authorizes Batch API requests.
+
+Note: This is the first version of the File Locking API, supporting only the
+simplest use case: single branch locking. The API is designed to be extensible
+as we experiment with more advanced locking scenarios, as defined in the
+[original proposal](/docs/proposals/locking.md).
+
+## Create Lock
+
+The client sends the following to create a lock by sending a `POST` to `/locks`
+(appended to the LFS server url, as described above).
+
+* `path` - String path name of the file that is locked. This should be
+relative to the root of the repository working directory.
+* `committer` - The optional values of `user.name` and `user.email` from the
+git config.
+
+Lock services are not required to use or store the committer details, since most
+lock services should already be able to identify the committer according to
+their credentials. The `committer` values are used for display purposes only.
+
+```js
+// POST https://lfs-server.com/locks
+// Accept: application/vnd.git-lfs+json
+// Content-Type: application/vnd.git-lfs+json
+// Authorization: Basic ...
+{
+  "path": "foo/bar.zip",
+  "committer": {
+    "name": "Jane Doe",
+    "email": "janedoe@email.com"
+  }
+}
+```
+
+### Successful Response
+
+Successful responses return the created lock:
+
+* `id` - String ID of the Lock. Git LFS doesn't enforce what type of ID is used,
+as long as it's returned a string.
+* `path` - String path name of the locked file.
+* `locked_at` - The String timestamp the lock was created.
+* `committer` - The optional name/email of the user that created the Lock. This
+could be different from the `committer` values in the request if the lock
+service already knows the committer's identity based on the API credentials.
+
+```js
+// HTTP/1.1 201 Created
+// Content-Type: application/vnd.git-lfs+json
+{
+  "lock": {
+    "id": "some-uuid",
+    "path": "/path/to/file",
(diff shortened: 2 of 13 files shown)


docs/api/batch.md
```diff
@@ -1,5 +1,7 @@
 # Git LFS Batch API
 
+Added: v1.0
```
ttaylorr: Was this released in v1.0? The first mention of the batch API I have seen in our release notes comes from [v0.6.0][1].

[1]: https://github.com/git-lfs/git-lfs/releases/tag/v0.6.0

docs/api/locking.md
```diff
@@ -0,0 +1,449 @@
+# Git LFS File Locking API
+
+Added: v2.0
+
+The File Locking API is used to create, list, and delete locks, as well as
+verify that locks are respected in Git pushes. The locking URLs are built
+by adding a suffix to the LFS Server URL.
+
+Git remote: https://git-server.com/foo/bar
```
ttaylorr: Nit: I think putting this data in a markdown table would look nice 👍 .

docs/api/locking.md
```diff
@@ -0,0 +1,449 @@
...
+
+All File Locking requests require the following HTTP headers:
+
+    Accept: application/vnd.git-lfs+json
+    Content-Type: application/vnd.git-lfs+json
+
+See the [Authentication doc](./authentication.md) for more info on how LFS
+gets authorizes Batch API requests.
+
+Note: This is the first version of the File Locking API, supporting only the
```
ttaylorr: Perhaps we should `>` quote this section since it's a `Note:` and temporary. Would also be nice to distinguish this section from the rest of the documentation.

docs/api/locking.md
```diff
@@ -0,0 +1,449 @@
...
+}
+```
+
+### Successful Response
+
+Successful responses return the created lock:
+
+* `id` - String ID of the Lock. Git LFS doesn't enforce what type of ID is used,
+as long as it's returned a string.
+* `path` - String path name of the locked file.
```
ttaylorr: We should state that this is relative to the root of the repository.

docs/api/locking.md
```diff
@@ -0,0 +1,449 @@
...
+* `locked_at` - The String timestamp the lock was created.
```
ttaylorr: We should dictate in what time format this should be.

docs/api/locking.md
```diff
@@ -0,0 +1,449 @@
...
+{
+  "message": "You must have pull access to list locks",
+  "documentation_url": "https://lfs-server.com/docs/errors",
+  "request_id": "123"
+}
+```
+
+### Error Response
+
+* `message` - String error message.
```
ttaylorr: This may be out of scope of the current PR, but we could probably format a lot of these request/response fields lists with a table:

| Name | Type | Description |
| --- | ---- | --- |
| `message` | String | Error message. |

docs/api/locking.md
technoweenie: You're killing the semantic HTMList in me :)

Seriously though, I don't have a strong opinion either way.

docs/api/locking.md
technoweenie: Markdown tables are so hard to read and write in raw text, I prefer to use them only when necessary. A two column table isn't necessary.

docs/api/locking.md
technoweenie: I think that's more complexity than a list of property needs.

technoweenie:
I've updated the code and tests to match the changes. @sinbad: I'm just waiting on your approval. I'd like to get this in next week if possible.

docs/api/locking.md
```diff
@@ -0,0 +1,449 @@
...
+[original proposal](/docs/proposals/locking.md).
+
+## Create Lock
+
+The client sends the following to create a lock by sending a `POST` to `/locks`
+(appended to the LFS server url, as described above).
+
+* `path` - String path name of the file that is locked. This should be
+relative to the root of the repository working directory.
+* `committer` - The optional values of `user.name` and `user.email` from the
```
sinbad: I get that the committer info is optional in the request, but I think we should nix it completely in the interests of committing to the minimum API. I don't believe any production server implementation is going to trust these values anyway. How about we say directly that auth & identity should be handled by the implementation in the same way it's done for the batch API?

docs/api/locking.md
```diff
@@ -0,0 +1,449 @@
...
+### Successful Response
+
+Successful responses return the created lock:
+
+* `id` - String ID of the Lock. Git LFS doesn't enforce what type of ID is used,
+as long as it's returned a string.
+* `path` - String path name of the locked file. This should be relative to the
+root of the repository working directory.
+* `locked_at` - The string ISO 8601 formatted timestamp the lock was created.
+* `committer` - The optional name/email of the user that created the Lock. This
```
sinbad: As a server response I think we should either require the server to provide the user information or remove it from the API. Leaving it optional means it's essentially useless to us because we can't rely on it.

docs/api/locking.md
```diff
@@ -0,0 +1,449 @@
...
+Note: As more advanced locking workflows are implemented, more details will
+likely be added to this request body in future iterations.
+
+### Successful Response
+
+A successful response will list the relevant locks:
+
+* `ours` - Array of Lock objects matching files that the user is allowed to
+modify.
+* `theirs` - Array of Lock objects matching files that the user is not allowed
```
sinbad: I think "allowed to modify" is ambiguous. Can we be more specific and say:

* ours - Array of lock objects currently owned by the authenticated user
* theirs - Array of lock objects currently owned by other users

docs/api/locking.md
technoweenie: Sounds good to me :+1: i wasn't sure how far I should go in the API purge :)

docs/api/locking.md
technoweenie: Good call, I'll make it more explicit.

docs/api/locking.md
technoweenie: How about we keep just `committer.name`? That way the client has something to show next to locks. Also, I think I'm going to rename it to `owner`. If the value is readonly (never sent from the client), I think this name makes more sense than `committer`. The name `committer` was chosen in a misguided (hindsight 20/20) attempt to be consistent with Git.

locking/locks.go
```diff
@@ -318,33 +324,26 @@ func (c *Client) lockIdFromPath(path string) (string, error) {
...
 	}
 
 	// We're going to overwrite the entire local cache
 	c.cache.Clear()
-
-	_, email := c.client.CurrentUser()
-	for _, l := range locks {
-		if l.Committer.Email == email {
-			c.cache.Add(l)
-		}
```
technoweenie: @sinbad: I updated this to use `VerifiableLocks()`. Though now I'm wondering if the main `/locks` endpoint should also partition locks into `ours` and `theirs` slices?

IMO if we did that, we'd still want `VerifiableLocks()` specifically for `pre-push`, since it'll be where the server works out the more advanced lock workflows that we've been discussion.

technoweenie:
Okay, final call for reviews. I think I've addressed every one's feedback.

docs/api/locking.md
sinbad: Yep that's good - and a required response 👍

locking/locks.go
sinbad: I think it's still good to have an endpoint that just lists all locks in a flat list, regardless of ownership - it's a fairly fundamental building block and having to combine 2 lists if you're doing something that requires them all would be a pain. It's also more intuitive for `locks` to just list all locks I think.

I agree with using `VerifiableLocks` here though.

sinbad:
Great work! 👍 Sorry for the delay in re-review, I've been very ill :/

PR #1912, opened 2017-02-03
WIP lfsapi: initial httpconfig type for getting `http.{url}.*` git config values

technoweenie:
This is an exploration of what it'd take to support `http.{url}.*` support (/cc #1500). The trick here is that [git-config's rules for `http.{url}.*` keys](https://git-scm.com/docs/git-config#git-config-httplturlgt) support a lot of rules for matching keys.

Given a url like `https://git-server.com/user/repo.git/info/lfs/objects/batch`, this code will look for these keys, where `suffix` is one of the [`http` config keys](https://git-scm.com/docs/git-config#git-config-httpproxy)

```
http.https://git-server.com/user/repo.git/info/lfs/objects/batch/.suffix
http.https://git-server.com/user/repo.git/info/lfs/objects/batch.suffix
http.https://git-server.com/user/repo.git/info/lfs/objects/.suffix
http.https://git-server.com/user/repo.git/info/lfs/objects.suffix
http.https://git-server.com/user/repo.git/info/lfs/.suffix
# ...
http.https://git-server.com/.suffix
http.https://git-server.com.suffix
http.suffix
```

One concern I have is that this is a lot of checks, even if it's mostly just building lots of strings, and doing lots of map lookups. Though, that's likely not much compared to actually making an http request.

Is my interpretation of the config rules correct? Should we use this as a basis for supporting `http.extraheader`?

TODO:

* [x] rename type from `httpconfig` to `urlconfig`
* [x] include property to specify the key prefix (typically `http` or `lfs`)
* [ ] Use when accessing `http.<url>.extraheader`
* [ ] Use when accessing `http.<url>.*`
* [ ] Use when accessing `lfs.<url>.access`
* [ ] Use when accessing `lfs.<url>.locksverify`


lfsapi/httpconfig.go
```diff
@@ -0,0 +1,68 @@
...
+
+	hosts := make([]string, 0)
+	if u.User != nil {
+		hosts = append(hosts, fmt.Sprintf("%s://%s@%s", u.Scheme, u.User.Username(), u.Host))
+	}
+	hosts = append(hosts, fmt.Sprintf("%s://%s", u.Scheme, u.Host))
+
+	pLen := len(u.Path)
+	if pLen > 2 {
+		end := pLen
```
ttaylorr: Small nit: you could turn this into a one-liner (which I think is still clean)

```go
end := len(strings.TrimSuffix(u.Path, slash))
```

Tasteful:
If you can provide a build artifact of this I can try it out with my tfs build agent.

ttaylorr:
Taking another look at this since I'd love to see this make it into v2.1.0. I read through the order of precedent that Git defines in https://git-scm.com/docs/git-config#git-config-httplturlgt, and was wondering if it would be easier to parse all of those keys into `*net/url.URL` instances and compare against that the given key.

I think the biggest win we'd get out of that approach would be that we have easy access to variables like `Scheme` and `Path`. This would allow us to do direct comparisons against the key and set of URL config values that we have to choose from instead of the going the other way around by building a set of matching keys and seeing if those keys have corresponding values.

On the other side of the coin, I can think of two downsides:

1. Direct comparison isn't a perfect fit since the git-config specification allows for a priority match among multiple "good" matches:

> The list above is ordered by decreasing precedence; a URL that matches a config key’s path is preferred to one that matches its user name. For example, if the URL is `https://user@example.com/foo/bar` a config key match of `https://example.com/foo` will be preferred over a config key match of `https://user@example.com`.

I think partitioning this as:

```
m map[string][]*net.URL
```

where the `[string]` key is the domain of the URL would allow us to narrow down the set of all possible matches into least qualifying ones and then check each one in that set. My thought there is that we know that a key can only match a config value iff at least the scheme, domain/host and port (or implicit port for that scheme) match. If we can partition the results based on the minimum qualifying terms, then we can narrow the search set for a given key lookup.

2. Having lots of `http.<url>` keys or `lfs.<url>` keys could be more expensive to parse than the time we'd save instead of generating keys. I think this largely depends on the size of the keyspace that you're searching, and for most cases I don't think it matters since the number of config keys will be small. That being said, this shouldn't matter a great deal when compared against the time it takes to do a network roundtrip, unless the keyspace is huge.

Does this seem like a good approach to take?

technoweenie:
That approach could work too. I went with the current approach since it doesn't require any upfront config caching. This PR does make a lot of `Get()` calls potentially, but they're all just map accesses.

I think having a map like `map[string][]*net.URL` could work though. Sort them by the length of the path, so longer path names are tried first.

```
http.https://github.com/abc/def.foo
http.https://github.com/abc.foo
http.https://github.com.foo
```

> Having lots of `http.<url>` keys or `lfs.<url>` keys could be more expensive to parse than the time we'd save instead of generating keys.

IMO this isn't a big concern in real usage.

ttaylorr:
> Sort them by the length of the path, so longer path names are tried first.

Rad, good call. This would eliminate a whole class of redundant URL comparison checks where a longer key exists. I think we could take this a step further and sort each URL lexicographically by length, but that may be too much/not needed.

> > Having lots of `http.<url>` keys or `lfs.<url>` keys could be more expensive to parse than the time we'd save instead of generating keys.
> 
> IMO this isn't a big concern in real usage.

I agree. I think it makes sense to go down this route.

ttaylorr:
@technoweenie I gave this another look today, and I think it makes more sense to build out a set of keys to lookup by rather than the other way around. The challenge is that, during parsing, it's unknown whether the config suffixes (".locksverify", ".extraHeader") are a part of the URL, or are a part of the config key.

That being said, it's certainly possible to account for that, by parsing the URL several times and chopping off another config section each time, but I think it's easier and clearer to follow the way that you have it.

I'd like to merge soon, before changing the `lfsapi` package and `commands/uploader.go` code individually in future PRs. That being said, I'm struggling with where to put this type. It definitely has shared functionality, and the `config` package seems to make the most sense. Should each package that wants to use it construct its own instance of `URLConfig`? Should we expand the `config.Environment` interface? Add a function to `*config.Config`?

I can't decide. What do you think makes most sense?

technoweenie:
> The challenge is that, during parsing, it's unknown whether the config suffixes (".locksverify", ".extraHeader") are a part of the URL, or are a part of the config key.

I think it's safe to assume that the last `.` marks the end of the url. I don't think any git config keys end in a url.

> I can't decide. What do you think makes most sense?

I put it in `lfsapi` because that's where I thought it made most sense. But that's because I'd only care about getting `http.<url>.extraheader` for the http client. I imagine `lfs.<url>.locksverify` should be accessed from the `locking` package or something. So, I think exposing `config.URLConfig` works :+1:

> Should each package that wants to use it construct its own instance of URLConfig? Should we expand the config.Environment interface? Add a function to *config.Config?

"no" for all 3 of those, imo :)

ttaylorr:
> > The challenge is that, during parsing, it's unknown whether the config suffixes (".locksverify", ".extraHeader") are a part of the URL, or are a part of the config key.
> 
> I think it's safe to assume that the last `.` marks the end of the url. I don't think any git config keys end in a url.
> 
> >

I was thinking about a case where you have a URL like:

```
http://example.com/index.php
```

Should the URL be treated as `http://example.com/index` and the key "php", or as the full URL with no URL key? With multiple dots, I don't think it's decide-able which way to parse this.


> > Should each package that wants to use it construct its own instance of URLConfig? Should we expand the config.Environment interface? Add a function to *config.Config?
> 
> "no" for all 3 of those, imo :)

Glad you think "no" for these, I tend to agree that each package should manage its own instance. That being said, I think this is ready to go, and the package-specific PRs cand come after this.

technoweenie:
Right, but the git config key would be `http.http://example.com/index.php.foo`, right? If you're parsing that, you know that the first and last `.`'s are the boundaries around the url. There are no git config keys like `http.http://example.com/index.php` (`http.<url>.php`).

ttaylorr:
> There are no git config keys like `http.http://example.com/index.php` (`http.<url>.php`).

Ah, I get what you're saying now. I guess technically we could have a key like that, but we don't (and I can't foresee a case where we would), so I think the URL boundaries are the first and last `.`.

I'd like to come back to this implementation later, but since it works for now I think I can get going on the http.<url>.extraHeader and then refactor this after.

ttaylorr:
>  Use when accessing http.<url>.extraheader

cc: #2159

ttaylorr:
> [ ] Use when accessing `http.<url>.*`

cc: #2160

ttaylorr:
> [ ] Use when accessing `lfs.<url>.access`

cc: https://github.com/git-lfs/git-lfs/pull/2161

ttaylorr:
> * [ ] Use when accessing `lfs.<url>.locksverify`

cc: https://github.com/git-lfs/git-lfs/pull/2162

PR #1921, opened 2017-02-07
commands: teach track how to handle empty lines in .gitattributes

technoweenie:
Just got a report from a Visual Studio user that `git lfs track` blows up when parsing the default `.gitattributes` that VS sets up. This fixes a bug with handling blank lines.


ttaylorr:
Sweet, thanks for grabbing this one 👍 .

PR #1922, opened 2017-02-09
malformed pointer bugs

larsxschneider:
Hi folks,

I discovered two bugs and added a test case to demonstrate them.

Cheers,
Lars

--

**BUG 1**
Malformed pointers >200160 bytes make Git LFS hang.
Reproducible with latest master and v1.5.5 on macOS and Windows using the filter-protocol.

@ttaylorr any idea? Pretty weird number 🤔 

**BUG 2**
Empty pointers are listed as malformed pointers.

### DIFF
--- a/test/test-malformed-pointers.sh
+++ b/test/test-malformed-pointers.sh
@@ -14,6 +14,7 @@ begin_test "malformed pointers"
   git add .gitattributes
   git commit -m "initial commit"
 
+                                     > malformed_empty.dat
   base64 /dev/urandom | head -c 1023 > malformed_small.dat
   base64 /dev/urandom | head -c 1024 > malformed_exact.dat
   base64 /dev/urandom | head -c 1025 > malformed_large.dat
@@ -33,6 +34,7 @@ begin_test "malformed pointers"
   pushd .. >/dev/null
     clone_repo "$reponame" "$reponame-assert"
 
+    grep "malformed_empty.dat" clone.log && false
     grep "malformed_small.dat" clone.log
     grep "malformed_exact.dat" clone.log
     grep "malformed_large.dat" clone.log
--- a/test/test-malformed-pointers.sh
+++ b/test/test-malformed-pointers.sh
@@ -18,6 +18,9 @@ begin_test "malformed pointers"
   base64 /dev/urandom | head -c 1024 > malformed_exact.dat
   base64 /dev/urandom | head -c 1025 > malformed_large.dat
 
+  base64 /dev/urandom | head -c 200160 > malformed_xxl_works.dat
+  base64 /dev/urandom | head -c 200161 > malformed_xxl_fails.dat
+
   git \
     -c "filter.lfs.process=" \
     -c "filter.lfs.clean=cat" \


technoweenie:
Awesome, thanks for opening this with reproducible tests. Should be able to get this into v2.0 :)

ttaylorr:
> **BUG 1**
> Malformed pointers >200160 bytes make Git LFS hang.
> Reproducible with latest master and v1.5.5 on macOS and Windows using the filter-protocol.

I think @peff and I arrived at what might be causing this while at Git Merge. When Git sends data from itself into LFS, it does all of the pipe `write()`s, before reading any of the data back. This breaks things down into a few cases:

1. Git sends a large file from itself to LFS, which gets consumed in chunks of 1024 bytes (see: [`io.Copy`][1]). LFS responds with a small pointer, which [Git consumes continually][2], since it is done writing the large file. The pipe buffer is never exceeded, since either side is continually draining from the other.
2. Git sends a large amount of (malformed) data, which LFS expected to be a pointer. This is fine if the file's length `n`, is less than half of the pipe buffer's maximum size. However, if LFS detects it's a malformed pointer, [it'll start writing it back out verbatim][3], _while Git is still writing data to LFS_. This is necessary, since we can't buffer more than a small amount of data for these potentially large files. The problem, though, is that [Git isn't reading any of the data that LFS writes back to it][4], so the pipe buffer fills up, causing it to hang. @peff and I confirmed this to be the case.

I think we can do a couple of things about this:

- Spool the data to disk and write it back to Git from LFS once Git is done `write()`-ing. This would ideally be a temporary measure since it involves more I/O than is necessary, and forces other users of the `process` filter to have to do the same.
- Much preferable, would be that Git use some sort of `select()`-based I/O to prevent the problem from occurring entirely.

I think a reasonable thing to do given that Git is close to releasing 2.12, is to temporarily implement the first option in LFS, and wait on a fix in upstream `git.git` as described in the second option. If this seems weird or I wrote it confusingly, please let me know (cc: @peff, @larsxschneider).

---

> **BUG 2**
> Empty pointers are listed as malformed pointers.

I think for this case, LFS can just ignore it and move on.

[1]: https://github.com/golang/go/blob/2a74b9e81405e34c67880866552b5d7bcab74de1/src/io/io.go#L387
[2]: https://github.com/git/git/blob/6e3a7b3398559305c7a239a42e447c21a8f39ff8/convert.c#L745
[3]: https://github.com/git-lfs/git-lfs/blob/5a55c028a88748c94ea4c40e031de5a3d9901bba/commands/command_smudge.go#L33-L35
[4]: https://github.com/git/git/blob/6e3a7b3398559305c7a239a42e447c21a8f39ff8/convert.c#L733-L750

larsxschneider:
Thanks for this great explanation @ttaylorr !
 
The problem would usually not happen because on *clean* Git LFS needs to read the full content until it can respond with the pointer and on *smudge* Git LFS needs to read the full pointer until it can respond with the content, right? 

That means a `select()`-based streaming interface wouldn't be necessary in the *usual* case. Therefore, I would vote against implementing it because it is certainly more complicated and consequently more error prone. I agree with your assessment about spooling.

*How about this?*
If Git LFS detects a malformed pointer then we can just return an empty response with the status [`error`](https://github.com/git/git/blob/5588dbffbd61e4906e453808c6ad32f792fea521/convert.c#L756-L757). This would cause Git to use the file "as-is" and we wouldn't even waste any time to transfer this file over the pipe from Git LFS to Git, again. However, there is a catch: This would only work if we set `filter.lfs.required=false` which would be semantically correct from my point of view. Could that work? Why do we set `filter.lfs.required=true` by default?

ttaylorr:
@larsxschneider I opened up #1932 to address the first bug 👍.

> Therefore, I would vote against implementing it because it is certainly more complicated and consequently more error prone. I agree with your assessment about spooling.

This seems fair to me, though I think eventually it'd be nice to see an upstream patch for this in `git.git`. Since the change in LFS is simple enough, I think we can just roll with that for now.

> However, there is a catch: This would only work if we set `filter.lfs.required=false` which would be semantically correct from my point of view. Could that work? Why do we set `filter.lfs.required=true` by default?

`filter.lfs.required` is `true` so LFS has the opportunity to halt invalid checkouts. I agree with your solution in that it would work, but Git would no longer be able to distinguish between LFS saying "just `cat` this file" and "please stop the checkout, something is wrong."

peff:
> That means a select()-based streaming interface wouldn't be necessary in the usual case. Therefore, I would vote against implementing it because it is certainly more complicated and consequently more error prone. I agree with your assessment about spooling.

I think this is true for LFS, which is the only user of the process-filter interface for now. But it is a gotcha waiting to happen for anybody else who tries to write a process-filter script.

I gave a little thought to the `select` interface, and it's actually kind of nasty, because the basic unit of work is a pktline. So doing it right would require possibly sending half a pktline, waiting for `select` to tell us we can write again, and then writing the rest, and so on. Not impossible, but it muddies a lot of the nice "copy X to Y via pktlines" functions you wrote.

It would be much nicer if we could use an async routine that just writes, while the main thread reads back.
The single-invocation filters use this approach. The code is simple, but it does make it a little harder to detect write errors from the main thread.

PR #1924, opened 2017-02-09
config: bump version to '2.0-pre' 

technoweenie:
This makes it super clear that you're not on an official LFS release.

```
$ script/run version
git-lfs/2.0-pre (GitHub; darwin amd64; go 1.7.5; git a699d202)
```


ttaylorr:
A post-merge nitpick, which you're totally free to ignore:

Taking a look at the relevant [semver.org][1] documentation:

> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version.

This makes me think that we should have the `.0` patch version in there as well, making this `2.0.0-pre`, since I think semver would have us release the next MAJOR version of LFS as `2.0.0`, not `2.0`. Either way, just a thought 👍 .

[1]: http://semver.org/#spec-item-9

technoweenie:
Yeah, that makes sense. This isn't really meant for actual releases. I just don't want people to confuse custom builds on master with our official releases. That said, we should totally change it to `2.1.0-pre` when we ship 2.0 :+1:

PR #1927, opened 2017-02-09
Fix `lfs unlock --force` on a missing file

### LINKED ISSUE #1925: lfs unlock requires the file to exist locally
I think passing `--force` should remove the lock regardless.

technoweenie:
Fixes #1925

### DIFF
--- a/commands/command_lock.go
+++ b/commands/command_lock.go
@@ -72,7 +72,7 @@ func lockPath(file string) (string, error) {
 
 	abs := filepath.Join(wd, file)
 	path := strings.TrimPrefix(abs, repo)
-	path = strings.TrimPrefix(path, os.PathSeparator)
+	path = strings.TrimPrefix(path, string(os.PathSeparator))
 	if stat, err := os.Stat(abs); err != nil {
 		return path, err
 	} else {
--- a/commands/command_lock.go
+++ b/commands/command_lock.go
@@ -72,15 +72,18 @@ func lockPath(file string) (string, error) {
 
 	abs := filepath.Join(wd, file)
 	path := strings.TrimPrefix(abs, repo)
+	if len(path) > 1 {
+		path = path[1:]
+	}
 
 	if stat, err := os.Stat(abs); err != nil {
-		return "", err
+		return path, err
 	} else {
 		if stat.IsDir() {
-			return "", fmt.Errorf("lfs: cannot lock directory: %s", file)
+			return path, fmt.Errorf("lfs: cannot lock directory: %s", file)
 		}
 
-		return path[1:], nil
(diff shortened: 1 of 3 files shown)


commands/command_lock.go
```diff
@@ -72,15 +72,18 @@ func lockPath(file string) (string, error) {
 
 	abs := filepath.Join(wd, file)
 	path := strings.TrimPrefix(abs, repo)
+	if len(path) > 1 {
```
ttaylorr: Is this to trim off the leading `/`? I think this should be fine to change since we don't have any production implementations in the wild right now, but maybe a clarifying comment, or something like:

```go
path = strings.TrimPefix("/", path)
```

would be good.

test/test-unlock.sh
```diff
@@ -19,6 +19,32 @@ begin_test "unlocking a lock by path"
 )
 end_test
 
+begin_test "force unlocking lock with missing file"
+(
+  set -e
+
+  reponame="force-unlock-missing-file"
+  setup_remote_repo_with_file "$reponame" "a.dat"
```
ttaylorr: Ha, so glad these are repo-local now 🙌 .

test/test-unlock.sh
```diff
@@ -19,6 +19,32 @@ begin_test "unlocking a lock by path"
...
+
+  GITLFSLOCKSENABLED=1 git lfs lock "a.dat" | tee lock.log
+  id=$(grep -oh "\((.*)\)" lock.log | tr -d "()")
```
ttaylorr: Definitely outside the scope of this PR, but we should perhaps extract this out to a function, since this `grep | tr` business is a tinge confusing. This one's on me, since I think I wrote this a while ago.

test/test-unlock.sh
```diff
@@ -19,6 +19,32 @@ begin_test "unlocking a lock by path"
...
+  assert_server_lock "$reponame" "$id"
+
+  git rm a.dat
+  git commit -m "a.dat"
+  rm *.log *.json # ensure clean git status
+  git status
```
ttaylorr: Does `git status` exit with a non-zero code if there are modified files? If not, we should probably `grep` through the output of this.

technoweenie: No, I just left this here for a bit of additional debugging in case there's a failure.

test/test-unlock.sh
technoweenie: Sure, sounds good.

test/test-unlock.sh
ttaylorr: Ah, cool -- thanks for explaining 👍 .

PR #1932, opened 2017-02-11
Spool malformed pointers to avoid deadlock

ttaylorr:
This pull-request mitigates the pipe buffer problem described in https://github.com/git-lfs/git-lfs/pull/1922#issuecomment-279024324 by spooling the contents of malformed pointers to a temporary file on disk before writing them back out to Git.

# Background

LFS and Git have essentially three scenarios under which they exchange files:

1. Git sends a (valid) pointer file, LFS responds with a large object.
2. Git sends a large object (tracked with LFS), LFS responds with a (valid) pointer.
3. Git sends a (malformed) pointer file, LFS responds with the same contents.

The snag is that Git does all of its `write()`s before any `read()` occurs on the shared pipe. This is fine in scenarios 1 & 2, since LFS drains the incoming data from Git continually, allowing the outgoing `write()`s to not exceed the pipe buffer.

Prior to this pull-request, scenario three would cause Git/LFS to hang when the size of the malformed pointer was at least half of the maximum pipe buffer size. Since LFS is essentially `tail -f`-ing the incoming contents while Git is still `write()`-ing, so there are no corresponding `read()`s, and the pipe buffer fills up.

# Solution

A temporary solution for LFS is to [spool][1] the contents of the malformed pointer entirely _before_ writing any data back out to Git. This means that while Git is `write()`-ing to the pipe, LFS is continually `read()`-ing that data, and LFS delays it's `write()`s until after Git is done, thus unblocking the pipe.

I'd eventually like to see an upstream patch in `git.git` (cc @larsxschneider @peff) using `select()`-based I/O, which would prevent other implementations of the process filter from having to adopt this same behavior.

One other approach proposed by @larsxschneider was:

> If Git LFS detects a malformed pointer then we can just return an empty response with the status `error`. This would cause Git to use the file "as-is" and we wouldn't even waste any time to transfer this file over the pipe from Git LFS to Git, again.

This would work, but requires us to set `filter.lfs.required` to `false`, which prevents us from responding with "real" errors that are intended to halt the checkout.

## Implementation

I considered a few other approaches to the Go implementation:

1. An `io.Writer` that spooled internally, making it possible to write:

```go
io.Copy(Spool(to), pbuf)
```

but this requires the implementation of the writing end of "to" to know when it is done being written to, of which there is no reliable way to detect.

2. Creating a temporary buffer in memory of a fixed size that would be written to _before_ spooling out to disk.

This would prevent slower disk I/O in a few cases where the malformed pointer's size is less than half of the pipe buffer, but introduces complexity not only in implementation, but platform-specific bindings. The pipe buffer size varies among operating systems _and_ platforms, so the tradeoff didn't seem worthwhile to me, but if others feel differently I'm happy to implement this.  

---

/cc @git-lfs/core for review

[1]: https://en.wikipedia.org/wiki/Spooling


tools/iotools.go
```diff
@@ -89,3 +93,42 @@ func (r *RetriableReader) Read(b []byte) (int, error) {
...
+// The temporary file is cleaned up after the copy is complete.
+//
+// The number of bytes written to "to", as well as any error encountered are
+// returned.
+func Spool(to io.Writer, from io.Reader) (n int64, err error) {
+	return SpoolName(to, from, fmt.Sprintf("lfs-%d", time.Now().Unix()))
+}
+
+// Spool spools the contents from 'from' to 'to' by buffering the entire
+// contents of 'from' into a temprorary file named "name", first.
```
larsxschneider: minor nit: the temporary file is _prefixed_ with "name", right? Because `lfs-1486823611` could collide if two malformed files would be processed within one second. Maybe we should rename `name` to `prefix`?

tools/iotools.go
```diff
@@ -89,3 +93,42 @@ func (r *RetriableReader) Read(b []byte) (int, error) {
...
+//
+// The temporary file is cleaned up after the copy is complete.
+//
+// The number of bytes written to "to", as well as any error encountered are
+// returned.
+func SpoolName(to io.Writer, from io.Reader, name string) (n int64, err error) {
+	tmp, err := ioutil.TempFile("", name)
+	if err != nil {
+		return 0, errors.Wrap(err, "spool tmp.")
```
larsxschneider: `.` after `tmp` intentional?

tools/iotools.go
```diff
@@ -89,3 +93,42 @@ func (r *RetriableReader) Read(b []byte) (int, error) {
...
+	}
+	defer os.Remove(tmp.Name())
+
+	if n, err = io.Copy(tmp, from); err != nil {
```
larsxschneider: I would imagine that this usually happens to small files because if you add a 1GB file directly to Git instead of LFS then you would realize that pretty quickly.

Therefore I wonder if we should indeed spool the first 10MB in memory to avoid any slow down due to disk I/O?!

larsxschneider: ... but maybe this is just premature optimization 😈

tools/iotools.go
sinbad: Agree with the collision issue. Why generate our own temporary file names at all, there are stdlib functions for this that would be better.

sinbad:
Always spooling will slow down the filters; even the small input smudge filters since the cost of the round trip just for creating a small file is much higher than memory. I suggest reading the first, say, 1k into an in-memory buffer, at which point you already know whether it's problem case 3. or not and can only spool in that case?

tools/iotools.go
ttaylorr: > Because `lfs-1486823611` could collide if two malformed files would be processed within one second.

I don't think this is the case since we [remove the file][1] before moving onto the next file from the process filter.

Either way, it shouldn't matter, since as @sinbad notes:

> I see it's just a seed for `TempFile` in `SpoolName` anyway so no problem in practice.

tools/iotools.go
ttaylorr: It was, but I don't think it looks great. Removed in: 8a6f60d4.

ttaylorr:
I just pushed 84af187 to address the comments in https://github.com/git-lfs/git-lfs/pull/1932#discussion_r100668708 and https://github.com/git-lfs/git-lfs/pull/1932#issuecomment-279368675. I was a bit off in my original thinking when I said:

> 2. Creating a temporary buffer in memory of a fixed size that would be written to before spooling out to disk.
>
> This would prevent slower disk I/O in a few cases where the malformed pointer's size is less than half of the pipe buffer, but introduces complexity not only in implementation, but platform-specific bindings. The pipe buffer size varies among operating systems and platforms, so the tradeoff didn't seem worthwhile to me, but if others feel differently I'm happy to implement this.

This is correct, but not in the context that I thought it was. I was thinking about sending data back directly to Git instead of buffering it in memory, in which case the amount that we buffer doesn't matter. The code did get a little more complex, but I added some comments to clear things up 👍 .

I'm not super happy with the fact that we buffer internally in `lfs/pointer_smudge.go` and `tools/iotools.go`, but I think for having a generic `tools.Spool` function, this works.

This should be ready for a final merge now.

peff:
Taking a step back on the problem: is there a way to avoid this pass-through entirely? IOW, is there a way that the filter process can tell Git to just use the contents it already has, without sending them back? Then you could punt on the spooling issue entirely (if we assume that pointer files are always small enough not to cause pipe deadlock, which seems reasonable).

In an ideal world, there would be a `status=declined` response for this. But since we're trying to fix problems with existing deployed Git, we can't rely on that.

I _think_ if you return `status=error` and the filter is not marked as `required` in Git's config, then it would just work. But I didn't test. And I'm not sure if the latter is a reasonable assumption to make.

So probably this is a dead end line of thinking, but I thought I'd raise it just in case.

ttaylorr:
> I _think_ if you return `status=error` and the filter is not marked as `required` in Git's config, then it would just work.

I think you're right, but I'd like to retain the ability for LFS to send valid, checkout-halting errors:

> One other approach proposed by @larsxschneider was:
> 
> > If Git LFS detects a malformed pointer then we can just return an empty response with the status `error`. This would cause Git to use the file "as-is" and we wouldn't even waste any time to transfer this file over the pipe from Git LFS to Git, again.
> 
> This would work, but requires us to set `filter.lfs.required` to `false`, which prevents us from responding with "real" errors that are intended to halt the checkout.

peff:
@ttaylor Thanks. I don't know how I totally missed that in your original PR description. :sweat: Sorry for the noise.

ttaylorr:
@peff No worries! I do like the idea of `status=declined`, but I think the `select()`-based I/O is more generally applicable.

ttaylorr:
👋  thanks, everybody!

PR #1936, opened 2017-02-13
documentation changes for 2.0

technoweenie:
* Removes `lfs.batch` completely. I noticed it was listed in `git-config.5.ronn`.
* Adds the bare minimum docs for `lock/unlock/locks`.

I'll add more for the comments:

* [x] `lock` requires the file to exist
* [x] `unlock` requires a clean git history, and for the file to exist
* [x] `unlock --force` skips those requirements

Anything else?


docs/man/git-lfs-config.5.ronn
```diff
@@ -190,6 +184,22 @@ lfs option can be scoped inside the configuration for a remote.
   If set to "basic" then credentials will be requested before making batch
   requests to this url, otherwise a public request will initially be attempted.
 
+* `lfs.<url>.skiplocksverify`
```
ttaylorr: Per our discussion in 🍖 -space, I think calling this `lfs.<url>.locks_verify` would be more appropriate, since it's one less cycle to figure out whether or not this performs the lock verification step.

docs/man/git-lfs-config.5.ronn
```diff
@@ -190,6 +184,22 @@ lfs option can be scoped inside the configuration for a remote.
...
+
+  Determines whether locks are checked before Git pushes. This prevents you from
+  pushing changes to files that other users have locked. The Git LFS pre-push
+  hook varies its behavior based on the value of this config key.
+
+  * `null` - In the absence of a value, Git LFS will attempt the call, and warn
```
ttaylorr: These will need an update pending the change above.

docs/man/git-lfs-lock.1.ronn
```diff
@@ -0,0 +1,27 @@
...
+===============================================================
+
+## SYNOPSIS
+
+`git lfs lock` [options] <path>
+
+## DESCRIPTION
+
+Sets the given file path as "locked" against the Git LFS server, with the
+intention of blocking attempts by other users to update the given path. Locking
```
ttaylorr: I would re-phrase this in terms of what locking a file actually does, not it's intention. In practice, locking a file causes the pre-push command to prevent the push client-side, so we should make that explicit in the command docs.

docs/man/git-lfs-lock.1.ronn
```diff
@@ -0,0 +1,27 @@
...
+a file requires the file to exist in the working copy.
+
+## OPTIONS
+
+* `-r` <name> `--remote=`<name>:
+  Specify the Git LFS server to use. Ignored if the `lfs.url` config key is set.
+
+* `--json`:
```
ttaylorr: A caveat of this flag is that `Error()` calls don't format their output as JSON, mostly as a function of us not having a standardized way to report errors back out to external programs.

I think it would be good to note here that any un-parseable JSON (+ a non-zero exit code) should be treated as the error message.

docs/man/git-lfs-locks.1.ronn
```diff
@@ -0,0 +1,37 @@
...
+
+Lists current locks from the Git LFS server.
+
+## OPTIONS
+
+* `-r` <name> `--remote=`<name>:
+  Specify the Git LFS server to use. Ignored if the `lfs.url` config key is set.
+
+* `-i <id>` `--id=<id>`:
+  Specifies a lock by its ID.
```
ttaylorr: Would be good to note that these flags (`--id`, `--path`) limit to one result.

docs/man/git-lfs-locks.1.ronn
```diff
@@ -0,0 +1,37 @@
...
+* `-p <path>` `--path=<path>`:
+  Specifies a lock by its path.
+
+* `--local`:
+  Lists only the locks cached locally. Skips a remote call.
+
+* `-l <num>` `--limit=<num>`:
+  Specifies number of results to return.
+
+* `--json`:
```
ttaylorr: Same note here about the `--json` output with errors.

docs/man/git-lfs-unlock.1.ronn
```diff
@@ -0,0 +1,33 @@
+git-lfs-unlock(1) -- Remove "locked" setting for a file on the Git LFS server
+=============================================================================
+
+## SYNOPSIS
+
+`git lfs unlock` [options] <path>
```
ttaylorr: I think this should technically be `[OPTIONS]`.

docs/man/git-lfs-config.5.ronn
technoweenie: Actually, I think the values are correct. I think I just forgot to switch the true/false values when I changed the name to `skipverifylocks`. Now that it's `locksverify`, I think it's okay now :)

PR #1940, opened 2017-02-14
Backport #1932 for v1.5.x: Spool malformed pointers to avoid deadlock

technoweenie:
This backports #1932.

Conflicting files:
- commands/command_smudge.go
- tools/iotools.go


ttaylorr:
Looks good.

PR #1951, opened 2017-02-16
pre-push: make lock verification optional for servers that don't support it

ttaylorr:
This pull-request allows the LFS client to be configured in such a way that makes lock verification a no-op for servers that don't yet support the new locking API.

At face value, this also makes it very easy for users to disable lock verification to push locked files to servers that don't block pushes not owned by the lock owner. However, there's really no change before/after this PR, since all of this logic is wrapped up in a Git hook anyway, you can still just delete the hook.

Here are the rules:

- If the `.locksverify` rule is unset, make the verify API call...
  - If it was successful, notify the user they should "enable" lock verification for that remote
  - If it failed...
    - AND the failure was a 501, disable the lock verification and move on
    - OR print a warning, but still push
- If the `.locksverify` config is true, make the verify API call...
  - If it was successful, move on
  - If it failed...
    - AND the failure was a 501, disable the lock verification and move on
    - OR print a warning and halt the push  
- If the `.locksverify` config was false, skip the verify API call.

Pushing to a sever that doesn't support locking:

```
~/g/git-lfs (optional-pre-push) $ git push origin master
[master (root-commit) edbdd87] initial commit
 2 files changed, 4 insertions(+)
 create mode 100644 a.dat
Remote "origin" does not support the LFS locking API. Consider disabling it with:
  $ git config lfs.http://127.0.0.1:61354/lock-unset-verify-5xx.git/info/lfs.locksverify false
http: Fatal error: Unable to parse HTTP response for POST http://127.0.0.1:61354/lock-unset-verify-5xx.git/info/lfs/locks/verify: EOF
```

Pushing to a server that does support locking:
```
~/g/git-lfs (optional-pre-push) $ git push origin master
[master (root-commit) a4f4c34] initial commit
 2 files changed, 4 insertions(+)
 create mode 100644 a.dat
  $ git config lfs.http://127.0.0.1:61383/lock-unset-verify.git/info/lfs.locksverify true
Git LFS: (1 of 1 files) 7 B / 7 B
To http://127.0.0.1:61383/lock-unset-verify
```

---

/cc @git-lfs/core

### DIFF
--- a/test/test-pre-push.sh
+++ b/test/test-pre-push.sh
@@ -530,3 +530,163 @@ begin_test "pre-push with their lock"
   popd >/dev/null
 )
 end_test
+
+begin_test "pre-push locks verify 5xx with verification enabled"
+(
+  set -e
+
+  reponame="lock-enabled-verify-5xx"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" "$reponame"
+
+  endpoint="$(repo_endpoint $GITSERVER $reponame)"
+
+  printf "example" > a.dat
+  git lfs track "*.dat"
+  git add .gitattributes a.dat
+  git commit --message "initial commit"
+
+  git config "lfs.$endpoint.locksverify" true
+
+  git push origin master 2>&1 | tee push.log
+  grep "\"origin\" does not support the LFS locking API" push.log
+  grep "git config lfs.$endpoint.locksverify false" push.log
+)
+end_test
+
+
+begin_test "pre-push locks verify 5xx with verification disabled"
+(
+  set -e
+
+  reponame="lock-disabled-verify-5xx"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" "$reponame"
+
+  endpoint="$(repo_endpoint $GITSERVER $reponame)"
+
+  printf "example" > a.dat
+  git lfs track "*.dat"
+  git add .gitattributes a.dat
+  git commit --message "initial commit"
+
+  git config "lfs.$endpoint.locksverify" false
+
+  git push origin master 2>&1 | tee push.log
+  [ "0" -eq "$(grep -c "\"origin\" does not support the LFS locking API" push.log)" ]
+)
+end_test
+
+begin_test "pre-push locks verify 5xx with verification unset"
+(
+  set -e
+
+  reponame="lock-unset-verify-5xx"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" "$reponame"
+
+  endpoint="$(repo_endpoint $GITSERVER $reponame)"
+
+  printf "example" > a.dat
+  git lfs track "*.dat"
+  git add .gitattributes a.dat
+  git commit --message "initial commit"
+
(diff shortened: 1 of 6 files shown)


test/test-pre-push.sh
```diff
@@ -563,3 +530,183 @@ begin_test "pre-push with their lock"
...
+  contents="example"
+  contents_oid="$(calc_oid "$contents")"
+  printf "$contents" > a.dat
+  git lfs track "*.dat"
+  git add .gitattributes a.dat
+  git commit --message "initial commit"
+
+  git config "lfs.$endpoint.locksverify" true
+
+  git push origin master 2>&1 | tee push.log
```
technoweenie: I think this should confirm the exit code is non-zero (through `PIPE_STATUS`?)

test/test-pre-push.sh
```diff
@@ -563,3 +530,183 @@ begin_test "pre-push with their lock"
...
+  git lfs track "*.dat"
+  git add .gitattributes a.dat
+  git commit --message "initial commit"
+
+  git config "lfs.$endpoint.locksverify" false
+
+  git push origin master 2>&1 | tee push.log
+  [ "0" -eq "$(grep -c "\"origin\" does not support the LFS locking API" push.log)" ]
+
+  refute_server_object "$reponame" "$contents_oid"
```
technoweenie: Shouldn't the push succeed? Also, `contents_oid` isn't defined.

test/test-pre-push.sh
```diff
@@ -563,3 +530,183 @@ begin_test "pre-push with their lock"
...
+  git lfs track "*.dat"
+  git add .gitattributes a.dat
+  git commit --message "initial commit"
+
+  [ -z "$(git config "lfs.$endpoint.locksverify")" ]
+
+  git push origin master 2>&1 | tee push.log
+  grep "\"origin\" does not support the LFS locking API" push.log
+
+  refute_server_object "$reponame" "$contents_oid"
```
technoweenie: This push should succeed too, despite being warned. Maybe the `refute` call is passing because `contents_oid` is unset here too.

test/test-pre-push.sh
ttaylorr: Hm, I think that the `refute_server_object` below should provide the same assertion.

test/test-pre-push.sh
ttaylorr: Yes! You were right in both of your comments (this and https://github.com/git-lfs/git-lfs/pull/1951#discussion_r101626607), but I made an error in writing these tests. Interestingly, this passed when changed from `assert_server_object` to `refute_server_object` since `$contents_oid` was undefined.

I fixed the tests in fbfe3f6 and made the `$oid` and `$id` params of some of our helpers required in 9f38f41.

ttaylorr:
Backing out of the changes in 642e663, since it looks like there are a few places where those functions are called incorrectly throughout our integration suite.

test/test-pre-push.sh
technoweenie: Sure, I can't think of a case where `git push` would succeed, but fail to push all the objects.

PR #1952, opened 2017-02-16
all: build using Go 1.8

ttaylorr:
This pull request upgrades our testing infrastructure to use Go 1.8.

- Travis: bumped in the `.travis.yml` to use 1.8.
- AppVeyor: Go 1.8 isn't in on `cinst` yet, so let's install it ourselves.
- CircleCI: upgrade through https://github.com/Homebrew/homebrew-core/pull/10048

---

/cc @git-lfs/core

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 # http://docs.travis-ci.com/user/languages/go/
 language: go
 
-go: 1.7.4
+go: 1.8
 
 os:
   - linux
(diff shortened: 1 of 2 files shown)


sschuberth:
Does it really make sense to jump at new Go versions immediately after their release? Waiting for the next minor release (like 1.8.1) would have the benefit of getting additional fixes for bugs that were rushed out with the release, and we'd have a higher chance of the release being available as a Chocolatey package.

technoweenie:
I think _testing_ with the new version is good. Though we probably could've held off on updating the appveyor build until it's available on chocolatey.

We're not releasing LFS 2.0 for at least a week, so we still have time to evaluate go 1.8.

PR #1955, opened 2017-02-17
progress: fix never reading bytes with sufficiently small files

### LINKED ISSUE #1926: progress meter doesn't update bytes transferred
```
$ git lfs clone https://github.com/larsxschneider/lfstest-manyfiles
Cloning into 'lfstest-manyfiles'...
Receiving objects: 100% (15012/15012), 2.02 MiB | 1.87 MiB/s, done.
Git LFS: (1798 of 1901 files) 0 B / 10.16 KB
```

Why is the first part stuck at `0 B`? 

I've also noticed that sometimes the command exits with an incorrect value:

```
$ git lfs clone https://github.com/github/git-lfs-test
Cloning into 'git-lfs-test'...
remote: Counting objects: 46, done.
remote: Total 46 (delta 0), reused 0 (delta 0), pack-reused 46
Unpacking objects: 100% (46/46), done.
Git LFS: (6 of 6 files) 780.54 KB / 818.95 KB
```

I think this is just a race condition. The command could be exiting before the progress meter has another chance to write to stdout.

ttaylorr:
This pull-requests fixes #1926, a bug which would cause push/pull operations with small objects to look like this:

```
~/Desktop $ git lfs clone git@github.com:larsxschneider/lfstest-manyfiles.git
Cloning into 'lfstest-manyfiles'...
Receiving objects: 100% (15012/15012), 2.02 MiB | 2.25 MiB/s, done.
Checking out files: 100% (15001/15001), done.
Git LFS: (87 of 201 files) 0 B / 1.15 KB
                           ^
```

The issue lies in the fact that `io.Reader`s, when at the end of a stream, can either return 
- `([]byte, nil)` then `(nil, io.EOF)`, or
- `([]byte, io.EOF)`

From the Go documentation<sup>[[1](https://golang.org/pkg/io/#Reader)]</sup>:

> An instance of this general case is that a Reader returning a non-zero number of bytes at the end of the input stream may return either err == EOF or err == nil. The next Read should return 0, EOF.

Particularly, the problem lied [here][1]:

```go
if err == nil && w.C != nil {
	err = w.C(w.TotalSize, w.ReadSize, n)
}
```

`err` could be `io.EOF` even when there was a positive number of bytes `n` to progress upon. This happens when the buffer "p" being read into is larger than the available data on the stream. In other words, small files make this incredibly common.

However, changing `err == nil` to `err == nil || err == io.EOF` isn't quite the right fix, since `(nil, io.EOF)` is a valid return value from io.Readers. Instead, let's check for positive values of `n`, and increment based on that. This is semantically equivalent to `err == nil || (err == io.EOF && n > 0)`, since the interface documentation specifies that non-nil errors should indicate zero bytes read.

I added an integration test to verify that we counted the bytes correctly for small files, and a unit test to prevent further regressions.

```
~/Desktop $ git lfs clone git@github.com:larsxschneider/lfstest-manyfiles.git
Cloning into 'lfstest-manyfiles'...
Receiving objects: 100% (15012/15012), 2.02 MiB | 2.40 MiB/s, done.
Git LFS: (61 of 201 files) 366 B / 1.15 KB
```

---

/cc @git-lfs/core 

[1]: https://github.com/git-lfs/git-lfs/blob/c22c48bb6471499f57ffbbe5dbfbaa6c1a1e9e35/progress/copycallback.go#L57

### DIFF
--- a/progress/copycallback.go
+++ b/progress/copycallback.go
@@ -31,12 +31,11 @@ func (r *bodyWithCallback) Read(p []byte) (int, error) {
 
 	if n > 0 {
 		r.readSize += int64(n)
-	}
 
-	if err == nil && r.c != nil {
-		err = r.c(r.totalSize, r.readSize, n)
+		if (err == nil || err == io.EOF) && r.c != nil {
+			err = r.c(r.totalSize, r.readSize, n)
+		}
 	}
-
 	return n, err
 }
 
@@ -52,12 +51,11 @@ func (w *CallbackReader) Read(p []byte) (int, error) {
 
 	if n > 0 {
 		w.ReadSize += int64(n)
-	}
 
-	if err == nil && w.C != nil {
-		err = w.C(w.TotalSize, w.ReadSize, n)
+		if (err == nil || err == io.EOF) && w.C != nil {
+			err = w.C(w.TotalSize, w.ReadSize, n)
+		}
 	}
-
 	return n, err
 }
 
--- a/progress/copycallback_test.go
+++ b/progress/copycallback_test.go
@@ -1,7 +1,7 @@
 package progress
 
 import (
-	"bytes"
+	"io"
 	"sync/atomic"
 	"testing"
 
@@ -31,7 +31,7 @@ func TestCopyCallbackReaderCallsCallbackUnderfilledBuffer(t *testing.T) {
 		C:         cb,
 		TotalSize: 3,
 		ReadSize:  2,
-		Reader:    bytes.NewReader(buf),
+		Reader:    &EOFReader{b: buf},
 	}
 
 	p := make([]byte, len(buf)+1)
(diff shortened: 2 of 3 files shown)


ttaylorr:
> It may make sense to make a custom reader implementation to trigger the bug.

Good call. I was thinking of doing this last night, but I'm glad you made the recommendation. Here's a failing test on `master` that passes on my branch:

```
~/g/git-lfs (master!) $ go test ./progress
--- FAIL: TestCopyCallbackReaderCallsCallbackUnderfilledBuffer (0.00s)
        Error Trace:    copycallback_test.go:41
        Error:          Expected nil, but got: &errors.errorString{s:"EOF"}

        Error Trace:    copycallback_test.go:43
        Error:          Not equal: 1 (expected)
                                != 0x0 (actual)
        Messages:       progress: expected 1 call to callback, got 0

        Error Trace:    copycallback_test.go:44
        Error:          Not equal: 3 (expected)
                                != 0 (actual)

        Error Trace:    copycallback_test.go:45
        Error:          Not equal: 3 (expected)
                                != 0 (actual)

        Error Trace:    copycallback_test.go:46
        Error:          Not equal: 1 (expected)
                                != 0 (actual)

FAIL
FAIL    github.com/git-lfs/git-lfs/progress     0.017s
```

and on `zero-bytes-forever`:

```
~/g/git-lfs (zero-bytes-forever!) $ go test ./progress
ok      github.com/git-lfs/git-lfs/progress     0.018s
```

xref: 591307d1.

PR #1958, opened 2017-02-17
tq: decrement uploaded bytes in basic_upload before retry

ttaylorr:
This pull-request fixes a bug that caused the progress meter to display that it had transferred more bytes than were included in the total push.

To set things up, I'm running a basic LFS server on my laptop that is configured to return an HTTP 400 when a client tries to upload more than 10 bytes. I've configured my repo to use that remote, and to push 502 files, the first 501 are less than 10 bytes.

```
~/D/pull-bug (master) $ git config lfs.transfer.maxretries
100
~/D/pull-bug (master) $ git lfs push local master
Git LFS: (501 of 502 files) 102.85 KB / 2.85 KB
LFS: Client error: http://[::]:55994/storage
```

The push is comprised of:

- 501 (less than 10b) files totaling a size of 1.85 KB
- 1 (1k) file

LFS uploads the first 1.85 KB successfully, and then tries 100 times (unsuccessfully) to upload the last file, but fails to decrement before it retries again, which is why the progress meter goes "over" the total count.

A more robust fix would expand the `tq.Adapter` interface to provide a flexible way for _all_ transfer adapter implementations to report if they "lost" progress, but I think that fixing the most common case with relatively low effort is a good first step.

I originally wanted to call `Seek(0, io.SeekCurrent)` on the open file descriptor, but I decided against this for two reasons:
1. This forces us to keep the file descriptor open.
2. It binds the underlying reader to implement `io.Seeker`, which may not always be the case.

Instead, I wrote a new type in the `tools` package, to implement `io.ReadSeeker` (and `io.Closer`, for usage with the `lfsapi` package) which atomically keeps track of the number of bytes read. Importantly, it also "overrides" the `Seek()` method, so any sort of seeking that we do will still yield correct updates to the total number of bytes read.

In conclusion:

```
~/D/pull-bug (master) $ git lfs push local master
Git LFS: (501 of 502 files) 1.85 KB / 2.85 KB
LFS: Client error: http://[::]:55994/storage
```

works as expected.

---

/cc @git-lfs/core

### DIFF
--- a/progress/copycallback.go
+++ b/progress/copycallback.go
@@ -14,8 +14,6 @@ type BodyWithCallback struct {
 	ReadSeekCloser
 }
 
-var _ ReadSeekCloser = (*BodyWithCallback)(nil)
-
 func NewByteBodyWithCallback(by []byte, totalSize int64, cb CopyCallback) *BodyWithCallback {
 	return NewBodyWithCallback(NewByteBody(by), totalSize, cb)
 }
--- a/progress/copycallback.go
+++ b/progress/copycallback.go
@@ -7,26 +7,30 @@ import (
 
 type CopyCallback func(totalSize int64, readSoFar int64, readSinceLast int) error
 
-type bodyWithCallback struct {
+type BodyWithCallback struct {
 	c         CopyCallback
 	totalSize int64
 	readSize  int64
 	ReadSeekCloser
 }
 
-func NewByteBodyWithCallback(by []byte, totalSize int64, cb CopyCallback) ReadSeekCloser {
+var _ ReadSeekCloser = (*BodyWithCallback)(nil)
+
+func NewByteBodyWithCallback(by []byte, totalSize int64, cb CopyCallback) *BodyWithCallback {
 	return NewBodyWithCallback(NewByteBody(by), totalSize, cb)
 }
 
-func NewBodyWithCallback(body ReadSeekCloser, totalSize int64, cb CopyCallback) ReadSeekCloser {
-	return &bodyWithCallback{
+func NewBodyWithCallback(body ReadSeekCloser, totalSize int64, cb CopyCallback) *BodyWithCallback {
+	return &BodyWithCallback{
 		c:              cb,
 		totalSize:      totalSize,
 		ReadSeekCloser: body,
 	}
 }
 
-func (r *bodyWithCallback) Read(p []byte) (int, error) {
+// Read wraps the underlying Reader's "Read" method. It also captures the number
+// of bytes read, and calls the callback.
+func (r *BodyWithCallback) Read(p []byte) (int, error) {
 	n, err := r.ReadSeekCloser.Read(p)
 
 	if n > 0 {
@@ -37,6 +41,27 @@ func (r *bodyWithCallback) Read(p []byte) (int, error) {
 		}
 	}
 	return n, err
+}
+
+// Seek wraps the underlying Seeker's "Seek" method, updating the number of
+// bytes that have been consumed by this reader.
+func (r *BodyWithCallback) Seek(offset int64, whence int) (int64, error) {
+	switch whence {
+	case io.SeekStart:
+		r.readSize = offset
+	case io.SeekCurrent:
+		r.readSize += offset
+	case io.SeekEnd:
+		r.readSize = r.totalSize + offset
+	}
+
+	return r.ReadSeekCloser.Seek(offset, whence)
+}
+
+// ResetProgress calls the callback with a negative read size equal to the
+// total number of bytes read so far, effectively "resetting" the progress.
+func (r *BodyWithCallback) ResetProgress() error {
+	return r.c(r.totalSize, r.readSize, -int(r.readSize))
 }
 
 type CallbackReader struct {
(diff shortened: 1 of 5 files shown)


tools/iotools.go
```diff
@@ -144,3 +145,71 @@ func Spool(to io.Writer, from io.Reader) (n int64, err error) {
...
+// of bytes read, and atomically updates the running count, making this method
+// safe to call across multiple goroutines.
+func (r *CountingReadSeekCloser) Read(p []byte) (n int, err error) {
+	n, err = r.ReadSeekCloser.Read(p)
+	atomic.AddInt64(&r.n, int64(n))
+
+	return
+}
+
+// Seek wraps the underlying Seeker's "Seak" method, atomically updating the
```
technoweenie: "Seak"

tq/basic_upload.go
```diff
@@ -80,7 +86,9 @@ func (a *basicUploadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb Progres
 		}
 		return nil
 	}
-	var reader lfsapi.ReadSeekCloser = progress.NewBodyWithCallback(f, t.Size, ccb)
+
+	fcount := tools.NewCountingReadSeekCloser(f, stat.Size())
+	var reader lfsapi.ReadSeekCloser = progress.NewBodyWithCallback(fcount, t.Size, ccb)
```
technoweenie: Why not just do the counting in the `*bodyWithCallback` that's returned? Why is a new type needed?

tools/iotools.go
```diff
@@ -144,3 +145,71 @@ func Spool(to io.Writer, from io.Reader) (n int64, err error) {
...
+// is atomic, this method is safe to call across multiple goroutines.
+func (r *CountingReadSeekCloser) Seek(offset int64, whence int) (int64, error) {
+	switch whence {
+	case io.SeekStart:
+		atomic.SwapInt64(&r.n, offset)
+	case io.SeekCurrent:
+		atomic.AddInt64(&r.n, offset)
+	case io.SeekEnd:
+		atomic.SwapInt64(&r.n, r.totalSize+offset)
+	}
```
technoweenie: Does this really need to use atomic? It looks like the reader gets passed in through an `http.Request` body, which is not accessed from multiple goroutines.

tq/basic_upload.go
```diff
@@ -94,6 +102,15 @@ func (a *basicUploadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb Progres
...
 	res, err := a.doHTTP(t, req)
 	if err != nil {
+		// We're about to return a retriable error, meaning that this
+		// transfer will either be retried, or it will fail.
+		//
+		// Either way, let's decrement the number of bytes that we've
+		// read _so far_, so that the next iteration doesn't re-transfer
+		// those bytes, according to the progress meter.
+		offset := fcount.N()
+		ccb(t.Size, offset, -int(offset))
```
technoweenie: If this was using the progress package's `*bodyWithCallback`, it could just be a single method like `ResetProgress()`.

tq/basic_upload.go
```diff
@@ -80,7 +86,9 @@ func (a *basicUploadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb Progres
 		}
 		return nil
 	}
-	var reader lfsapi.ReadSeekCloser = progress.NewBodyWithCallback(f, t.Size, ccb)
+
+	fcount := tools.NewCountingReadSeekCloser(f, stat.Size())
```
technoweenie: Wouldn't `t.Size` work here too?

tools/iotools.go
ttaylorr: I was thinking that this would be a good insurance policy to have if the reader ever got used in some crazy circumstances. I think the additional overhead required to use `sync/atomic` is low relative to the network operations being used in this context.

tq/basic_upload.go
ttaylorr: Ah, great call.

tools/iotools.go
ttaylorr: Nope 👍

ttaylorr:
> Why not just do the counting in the `*bodyWithCallback` that's returned? Why is a new type needed?

Good question, and great call. I originally considered implementing this functionality on the wrapped callback reader that gets created, but that only happens conditionally, so I think I skipped over the body callback reader.

I went ahead and re-implemented this on the callback reader in https://github.com/git-lfs/git-lfs/commit/589dffbefadf8ef28e03241ff6cf413a1e0c1962.

progress/copycallback.go
```diff
@@ -7,26 +7,30 @@ import (
...
-type bodyWithCallback struct {
+type BodyWithCallback struct {
 	c         CopyCallback
 	totalSize int64
 	readSize  int64
 	ReadSeekCloser
 }
 
-func NewByteBodyWithCallback(by []byte, totalSize int64, cb CopyCallback) ReadSeekCloser {
+var _ ReadSeekCloser = (*BodyWithCallback)(nil)
```
technoweenie: These declarations are unnecessary. It wouldn't compile if this wasn't the case.

PR #1959, opened 2017-02-18
AppVeyor: Again use a Chocolatey package for installing Golang

sschuberth:
This time for version 1.8.0, which is available now.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,9 +8,7 @@ clone_folder: $(GOPATH)\src\github.com\git-lfs\git-lfs
 
 install:
   - rd C:\Go /s /q
-  - appveyor DownloadFile https://storage.googleapis.com/golang/go1.8.windows-amd64.zip
-  - 7z x go1.8.windows-amd64.zip -oC:\ >nul
-  - C:\go\bin\go version
+  - cinst golang --version 1.8.0 -y
   - cinst InnoSetup -y
   - refreshenv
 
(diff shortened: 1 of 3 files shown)


ttaylorr:
Looks like there are some test failures on Windows. Would you mind taking a look at thos before we merge?

sschuberth:
@ttaylorr I have no idea what these error are about:

    ..\..\..\..\..\AppData\Local\Temp\go-build605275998\github.com\git-lfs\git-lfs\commands\_test\_testmain.go:56: cannot use matchString (type func(string, string) (bool, error)) as type testing.testDeps in argument to testing.MainStart:
        func(string, string) (bool, error) does not implement testing.testDeps (missing MatchString method)

As the `_testmain.go` file doe snot even exist in the repo, I guess this is some auto-generated test framework file, and further Googling seems to indicate there was a [testing.testmain API change](https://github.com/constabulary/gb/pull/657/commits/24e6a87e8e469fccbc15b33d740bb8c0049ad777). But if that's the case, why does Go 1.8 work fine on the other platforms?

ttaylorr:
If there were testmain API changes in Go 1.8, it's possible that there is somehow multiple versions of Go installed on the Windows VMs, and it's picking one less than 1.8 to run the tests with. Maybe a well placed `go version` would be helpful?

sschuberth:
> Maybe a well placed `go version` would be helpful?

Good idea, that indeed did the trick. [This](https://ci.appveyor.com/project/git-lfs/git-lfs/build/1.0.1390#L135) says "go1.7.1" just before running the tests, although during the build I see the expected "Using go1.8" statements. Need to look further what's going on.

Issue #1781, opened 2016-12-16
Proposal: resumable transfers for v2.0.0


ttaylorr:
After closing #1651, this issue remains open, but is out of scope of the original issue.

> **Resumable transfers, by default.** We should be able to resume uploads and downloads as the default if a transfer fails part-way through. This, combined with better retry behavior will yield for much more resilient and efficient transfers.

Opening this issue to track it:

---

/cc @git-lfs/core

ttaylorr:
Commenting here to ack out of our internal first-responder scripts! 👋

AndrewJDR:
I feel like #802 is related to this because both would be using range requests, presumably.
It would be great to have multi-threaded downloading integrated. git-lfs delivers a fraction of the download speed that I know I'm capable of getting on my connection. In certain cases, it's downright abysmal compared to what I know is possible (i.e. if I use a multi-threaded download manager along with "View Raw" on github.com, I can download the file in a fraction of the time it takes for git-lfs to download the same file).

technoweenie:
Git LFS already supports parallel downloads through `git lfs fetch/pull/clone`. Just not in the internal git `smudge` and `process` filters though. We've talked about improving the new `process` filter in a way that lets it download files in parallel too, but it requires changes to Git.

It still doesn't cover the case you mentioned, downloading a single file in parts. The `basic_download` adapter already supports retries with the `Range` header, so it's definitely _possible_. Git LFS just needs to communicate to the `basic_download` adapter when it should download files in parallel (ie: when the current batch has fewer files than `lfs.concurrenttransfers`).

PR #1961, opened 2017-02-18
tq: bump 'defaultMaxRetries' to 8

ttaylorr:
This pull request bumps the default maximum retries LFS will make before failing an object transfer to eight, in conjunction with #1781.

---

/cc @git-lfs/core


ttaylorr:
🙌

PR #1962, opened 2017-02-18
tools: fix truncating string fields between balanced quotes

### LINKED ISSUE #1895: Quotes in GIT_SSH_COMMAND are not properly handled
If you set `GIT_SSH_COMMAND="ssh -i '/a path/with spaces'"` lfs commands will complain about being unable to read `'/a path/with spaces'` (including quotes) where file `/a path/with spaces` actually exists.

ttaylorr:
This pull-request fixes #1895 by implementing `tools.QuotedFields`, an alternative to `strings.Fields` that respects balanced pairs of quotes.

The problem in #1895 is that a `GIT_SSH_COMMAND` with spaces inside a balanced pair of quotes would be truncated across multiple "fields", causing the arguments to be passed incorrectly.

For instance, the following string:

```
foo bar 'baz "etc"'
```

given to `strings.Fields` would return:

```go
[]string{"foo", "bar", "'baz", "\"etc\"'"}
```

whereas given to `tools.QuotedFields`, it instead returns:

```go
[]string{"foo", "bar", "'baz \"etc\"'"}
```

When trying to figure out the best way to implement this, I considered a few options:

- Instantiate the `exec.Cmd` in such a way that we wouldn't need to de-quote the input string.
- Write a `func(c rune) bool` that could be given to `strings.FieldsFunc`
- Use a regexp.

My findings were that the first and second options were both not possible. In particular, the second option seemed more attractive to me, since it would be a more easily approachable implementation. However, according to the docs of `strings.FieldsFunc`<sup>[[1]](https://golang.org/pkg/strings/#FieldsFunc)</sup>:

> FieldsFunc makes no guarantees about the order in which it calls f(c). If f does not return consistent results for a given c, FieldsFunc may crash.

In my testing, I was not able to make `FieldsFunc` crash, but since we can't rely on that behavior in the future, I chose to implement this with a regular expression instead. It turns out the regular expression isn't that complicated, so I'm fine going forward with this approach.

I also tried my hand at implementing this in a loop that scans each rune in the string and composes the string slice itself. I found this to be much more complicated than the regex, for not a significant performance gain:

```
~/g/git-lfs (fix-fields-with-quotes!) $ go test -bench='BenchmarkQuote' ./tools
BenchmarkQuoteWithRegex-4         500000              3482 ns/op
BenchmarkQuoteWithLoop-4         1000000              1025 ns/op
PASS
ok      github.com/git-lfs/git-lfs/tools        2.977s
```

As a side note, I used this PR as an opportunity to try out the new `testing.T.Run` method in Go 1.8.

---

/cc @git-lfs/core

### DIFF
--- a/lfsapi/ssh.go
+++ b/lfsapi/ssh.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/git-lfs/git-lfs/tools"
 	"github.com/rubyist/tracerx"
 )
 
@@ -66,7 +67,7 @@ func sshGetExeAndArgs(osEnv Env, e Endpoint) (exe string, baseargs []string) {
 
 	ssh, _ := osEnv.Get("GIT_SSH")
 	sshCmd, _ := osEnv.Get("GIT_SSH_COMMAND")
-	cmdArgs := strings.Fields(sshCmd)
+	cmdArgs := tools.QuotedFields(sshCmd)
 	if len(cmdArgs) > 0 {
 		ssh = cmdArgs[0]
 		cmdArgs = cmdArgs[1:]
--- a/lfsapi/ssh_test.go
+++ b/lfsapi/ssh_test.go
@@ -177,6 +177,20 @@ func TestSSHGetExeAndArgsSshCommandArgs(t *testing.T) {
 	assert.Equal(t, []string{"--args", "1", "user@foo.com"}, args)
 }
 
+func TestSSHGetExeAndArgsSshCommandArgsWithMixedQuotes(t *testing.T) {
+	cli, err := NewClient(TestEnv(map[string]string{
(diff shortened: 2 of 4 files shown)


ttaylorr:
> > whereas given to tools.QuotedFields, it instead returns:
> > ```
> > []string{"foo", "bar", "'baz \"etc\"'"}
> > ```
> 
> I think this is incorrect. We don't want the single quote, since that's being used to escape the bash argument. The argument is `baz "etc"`, or `"baz \"etc\""` as a go string.

Good call. I fixed this right up in https://github.com/git-lfs/git-lfs/pull/1962/commits/7ec512c72f59db13711e0c782d75ab8801822639 and https://github.com/git-lfs/git-lfs/pull/1962/commits/a35f0759de05df51ca69a5a76d6c3577ff2397cf.

Should be ready for another review now.

Issue #1964, opened 2017-02-20
Install without admin access on Windows


bbodenmiller:
It appears Git LFS cannot be installed without administrator access on Windows. Git for Windows can be however. Please add support or details on how to install Git LFS without admin access.

![image](https://cloud.githubusercontent.com/assets/1192780/23142265/e721c730-f76f-11e6-9a11-866f6738b819.png)

ttaylorr:
@bbodenmiller thanks for opening this issue, and sorry that you're having trouble. I took a quick look to refresh my memory on how our Windows installation process works, and I don't see anything that would necessitate administrator access to successfully install LFS.

That said, I imagine that there are Windows configurations that forbid non-admin users from installing executables on the computer. Would you mind double checking that that isn't the case?

bbodenmiller:
That doesn't appear to be the case here. Do you know where in the LFS installer code I could see where it might be causing this alert?

Here's some screenshots comparing Git for Windows and Git LFS installation on the same machine running as a non-admin user. As you can see Git for Windows installs but Git LFS doesn't make it anywhere. 😞 

![image](https://cloud.githubusercontent.com/assets/1192780/23147041/d574fc52-f790-11e6-8587-8ced8e1a168a.png)
![image](https://cloud.githubusercontent.com/assets/1192780/23147046/df8130bc-f790-11e6-8e1f-fef29976e1d9.png)
![image](https://cloud.githubusercontent.com/assets/1192780/23147064/fdc00152-f790-11e6-8a8d-445e0ed31c51.png)

ttaylorr:
@bbodenmiller thanks for posting. The contents of our Windows Installer is [here](https://github.com/git-lfs/git-lfs/blob/master/script/windows-installer/inno-setup-git-lfs-installer.iss), but I was unable to find anything that would indicate a need for administrative privileges.

@sschuberth is more of an expert in this area than I am, so perhaps he could help.

sschuberth:
@bbodenmiller Would you mind trying the installer from https://ci.appveyor.com/api/buildjobs/23511y185oxf3gv6/artifacts/git-lfs-windows-1.5.0.exe?

dcorbin-wintellect:
Git-2.33.0.2-64-bit.exe  -- prompts for Administrator, and can not go past it....

KalleOlaviNiemitalo:
[...]

PR #1966, opened 2017-02-21
Allow non-admin installations on Windows

sschuberth:
Please have a look at the individual commit messages for the details.


ttaylorr:
Looks good to me. Would you mind posting some screenshots of this in action on Windows? I'd do it myself, but my VM is M.I.A.

sschuberth:
I just realized we should probably use a different default install path when running as non-admin. I'll add another commit for this and then post some screenshots.

sschuberth:
@ttaylorr @strich On a somewhat related note, do you have any idea why the choice was made to set the non-default `UsePreviousAppDir=no` option?

ttaylorr:
> @ttaylorr @strich On a somewhat related note, do you have any idea why the choice was made to set the non-default `UsePreviousAppDir=no` option?

Not sure from my end. @strich any thoughts?

sschuberth:
> Would you mind posting some screenshots of this in action on Windows?

Here are screenshots of admin vs. non-admin install, i.e. it "just works" and does not show the "You must be logged in as an administrator" error message anymore, but the default install paths differs:

asadmin

asuser

But I just found another issue with Registry access. Will do yet another commit.

ttaylorr:
> Here are screenshots of admin vs. non-admin install, i.e. it "just works" and does not show the "You must be logged in as an administrator" error message anymore, but the default install paths differs:

Looking great. Ping me when you're ready for a merge 👍

sschuberth:
@ttaylorr Unfortunately there was no feedback from @bbodenmiller for issue #1964, but I believe the PR to be good to merge now.

strich:
> @ttaylorr @strich On a somewhat related note, do you have any idea why the choice was made to set the non-default UsePreviousAppDir=no option?

I believe it was to work around the previous installer and its location. It may no longer be necessary.

As for these changes - I'm a bit confused. The Git for Windows installer, by default, asks for admin elevation and if you deny it, the installer exits. By default it installs to `Program Files`. I don't see the screenshots @bbodenmiller [posted here](https://github.com/git-lfs/git-lfs/issues/1964#issuecomment-281220543).

As far as I know, off the top of my head, you require admin privs if you want to modify environment variables like the `PATH` var, which we need to do to ensure LFS is available at the commandline.

That's not to say that there shouldn't be some option to install LFS without admin permissions - Maybe a portable installer version, as Git for Windows also has?

sschuberth:
> The Git for Windows installer, by default, asks for admin elevation and if you deny it, the installer exits.

That's not correct, at least not anymore as of https://github.com/git-for-windows/build-extra/commit/23672af723da18e5bc3c679e52106de3c2dec55a. See the documentation for [PrivilegesRequired](http://www.jrsoftware.org/ishelp/index.php?topic=setup_privilegesrequired), but note that [the very useful "none" value is actually undocumented](http://news.jrsoftware.org/read/article.php?id=104844&group=jrsoftware.innosetup#104844).

> I don't see the screenshots @bbodenmiller [posted here](https://github.com/git-lfs/git-lfs/issues/1964#issuecomment-281220543).

I guess the screenshots in the left column come from the Git for Windows installer, and the screenshots in the right column come form the Git LFS installer. I.e. they are not supposed to be read left-ti-right, which is a bit misleading. So the error messages come from the Git LFS installer, not from the Git for Windows one.

> you require admin privs if you want to modify environment variables like the `PATH` var

Again not correct ;-) Windows has system- and user-scope variables. Admin privileges are only required to modify systems variables, but any user can modify his / her own variables. In case of `PATH`, if present at both system and user scope, their values are simply concatenated. System values come first, and thus take precedence.

> Maybe a portable installer version, as Git for Windows also has?

Why? What's wrong with the solution in this PR, esp. as Git for Windows does it the same way?

strich:
I just downloaded the Git for Windows installer and it forces admin elevation.

Tell me, with this patch, what is the default behavior? Does the installer request elevation and if denied fall back to `*/AppData/*`?

bbodenmiller:
> I just downloaded the Git for Windows installer and it forces admin elevation.

Still not true, I've installed on hundreds of machines without admin access. It does do a UAC prompt which is confusing but a nonadmin user can say yes on the UAC prompt then the installer detects user is non-admin and installs to AppData.

> @ttaylorr Unfortunately there was no feedback from @bbodenmiller for issue #1964, but I believe the PR to be good to merge now.

I will test today.

strich:
>Still not true, I've installed on hundreds of machines without admin access. It does do a UAC prompt which is confusing but a nonadmin user can say yes on the UAC prompt then the installer detects user is non-admin and installs to AppData.

Ah! Okay I misunderstood how it reacts to UAC. Fair enough. I think this also answers my previous question too - Can you confirm `PrivilegesRequired=none` still requests UAC when the installer is launched?

sschuberth:
> Can you confirm `PrivilegesRequired=none` still requests UAC when the installer is launched?

@strich Just try for yourself, the installer is available as a build artifact from CI: https://ci.appveyor.com/api/buildjobs/0swx6y4nn0g2l372/artifacts/git-lfs-windows-1.5.0.exe

ttaylorr:
@strich @sschuberth @bbodenmiller thanks for figuring this all out. Seems like we're in a good place to merge now.

@bbodenmiller -- let me know if you're able to make this work with the latest changes. If so, I'll merge it up.

script/windows-installer/inno-setup-git-lfs-installer.iss
```diff
@@ -1,4 +1,5 @@
 #define MyAppName "Git LFS"
+#define MyAppName "Git LFS"
```
bbodenmiller: Appears this is defined twice?

bbodenmiller:
As far as I can tell from my testing, installation, use, and uninstall work for nonadmin using https://ci.appveyor.com/api/buildjobs/gk3tihx6hxguyd6c/artifacts/git-lfs-windows-1.5.0.exe.

script/windows-installer/inno-setup-git-lfs-installer.iss
sschuberth: Thanks for spotting, fixed.

technoweenie:
Thanks for confirming, @bbodenmiller! This looks ready to go.

bbodenmiller:
Any idea on timeline for next release?

technoweenie:
I'm hoping for next week 🤞

PR #1983, opened 2017-02-24
Print error causes to stderr

technoweenie:
This adds `error.Cause()`, which removes all those error wrap prefixes when printing messages to stderr.

```
# BEFORE
$ git lfs lock path/to/file
Lock failed: api: http: http: Lock exists

# AFTER
$ git lfs lock path/to/file
Lock failed: Lock exists
```

### DIFF
--- a/commands/command_locks.go
+++ b/commands/command_locks.go
@@ -6,6 +6,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/git-lfs/git-lfs/errors"
 	"github.com/git-lfs/git-lfs/locking"
 	"github.com/git-lfs/git-lfs/tools"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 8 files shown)


ttaylorr:
@technoweenie I pushed 9e1b374, which checks whether there are arguments XOR the `--id` flag before printing the usage. This should make the build pass again.

technoweenie:
Thanks for tackling that :+1:

Sanny388cry (commented):
# BEFORE
$ git lfs lock path/to/file
Lock failed: api: http: http: Lock exists

# AFTER
$ git lfs lock path/to/file
Lock failed: Lock exists

commands/command_locks.go
Sanny388cry, line 1: commands/command_locks.go

PR #1987, opened 2017-02-27
More msg tweaks

technoweenie:
Three small things:

1. I thought printing the doc url and request id for every error was too noisy. Especially, when some are common errors like attempting to unlock someone else's file.
2. Added IDs to the `git lfs locks` output.

        $ git lfs locks
        collabocats.txt	ttaylorr	ID:1
        supportocat.txt	technoweenie   	ID:2

3. `git lfs unlock --id={id}` was causing Go to panic since `args` is nil. Fixed that, and updated the test that let the error slide by piping the contents to `tee`.

### DIFF
--- a/commands/command_locks.go
+++ b/commands/command_locks.go
@@ -35,20 +35,33 @@ func locksCommand(cmd *cobra.Command, args []string) {
 		return
 	}
 
-	var maxlen int
+	var maxPathLen int
+	var maxNameLen int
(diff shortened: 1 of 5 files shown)


technoweenie:
> Is the `id` field something that users want to know about in the majority case?

The majority case? Probably not. But it is important for `git lfs unlock --id={id}`. Also, the output is pretty sparse right now with just `{path} <tab> {name}`.

PR #1999, opened 2017-03-03
Mention `.lfsconfig` in the `git-lfs-config` documentation

terrorobe:
Seems like this isn't mentioned anywhere else except in the git-lfs tutorial. Feel free to nit on the wording.

### DIFF
--- a/docs/man/git-lfs-config.5.ronn
+++ b/docs/man/git-lfs-config.5.ronn
@@ -8,6 +8,11 @@ precedence. Most options pertaining to git-lfs are contained in the `[lfs]`
 section, meaning they all named `lfs.foo` or similar, although occasionally an
 lfs option can be scoped inside the configuration for a remote.
 
+These settings are read from a file called `.lfsconfig` found at the root of the
+repository, though legacy support exists to read from `.gitconfig` as well.
+
+The `.lfsconfig` file uses the same format as `.gitconfig`.
+
 ## LIST OF OPTIONS
 
 ### General settings


terrorobe:
I already did so but apparently, that didn't do a new status push. Should I do no-op commits to refresh the status or will you just merge it as-is?

screen shot 2017-03-03 at 19 12 44

ttaylorr:
> Should I do no-op commits to refresh the status or will you just merge it as-is?

I'll just merge it in as-is. 👍

docs/man/git-lfs-config.5.ronn
```diff
@@ -8,6 +8,11 @@ precedence. Most options pertaining to git-lfs are contained in the `[lfs]`
 section, meaning they all named `lfs.foo` or similar, although occasionally an
 lfs option can be scoped inside the configuration for a remote.
 
+These settings are read from a file called `.lfsconfig` found at the root of the
+repository, though legacy support exists to read from `.gitconfig` as well.
+
+The `.lfsconfig` file uses the same format as `.gitconfig`.
```
technoweenie: This isn't totally accurate. These settings are read from the normal git config, but can _also_ be read from a `.lfsconfig` file found at the root of the repository. Support for `.gitconfig` at the root of the repository was removed in LFS 2.0.

terrorobe: How would you word this? Something along these lines?

> These settings are read from a file called `.lfsconfig` found at the root of the repository, though legacy support exists to read from the Git configuration as returned by `git config -l` as well.

And should I open a new PR for that or edit on top of this branch?

technoweenie: LFS reads configs from a `.lfsconfig` in the root of the working directory, but also overrides any of those settings from `git config -l`. This allows you to override something like `lfs.url` on your local repository if you don't like the setting in `.lfsconfig`.

When I talk about `.gitconfig` support being removed in 2.0, I'm specifically referring to reading `.gitconfig` files in the root of the repository. This is something git does not support anywhere, and was specifically added for Git LFS.

> And should I open a new PR for that or edit on top of this branch?

New PR please, since this is already merged. Thanks :)

Issue #2003, opened 2017-03-06
git-lfs 2.0.0 skips upload when locking API is not supported


derrickstolee:
Not all services that support Git LFS support the new locking protocol (such as VSTS or older versions of the on-prem GitHub offering). The LFS client should ignore the locking protocol when not supported, but currently skips file uploads.

**Repro Steps:**
1. Create an LFS repo using 2.0.0 and add one tracked LFS file (I added a jpg)
2. Create a new VSTS repo and set the clone URL as the origin.
3. `git push origin --all`.

**Expected Behavior:**
The push should succeed and the LFS object should be pushed along with the Git objects.

**Actual Behavior:**
The LFS client "skips" the object, but the rest of the push appears to succeed. The skipped file message does not look like an error, so the user assumes the push succeeded.

See the following log of the commands that I ran first with 2.0.0 and then with 1.5.6:
[LFS-2.0.0-IssueLog.txt](https://github.com/git-lfs/git-lfs/files/821525/LFS-2.0.0-IssueLog.txt)

technoweenie:
Git LFS does detect that locking is not supported, and skips it correctly. The problem is VSTS doesn't support the batch api correctly. Responses put the actions in a `_links` property, but since LFS looks for `actions`, it assumes nothing needs to be done on the server and counts it as skipped.

Here's the current API spec:

* [Batch doc](https://github.com/git-lfs/git-lfs/blob/master/docs/api/batch.md)
* [Batch response schema](https://github.com/git-lfs/git-lfs/blob/master/tq/schemas/http-batch-response-schema.json)

Here's the versions from LFS v0.6.0, when the batch api was originally released:

* [Batch doc](https://github.com/git-lfs/git-lfs/blob/v0.6.0/docs/api/http-v1-batch.md)
* [Batch response schema](https://github.com/git-lfs/git-lfs/blob/v0.6.0/docs/api/http-v1-batch-response-schema.json)

The `_links` property is held over from the original API that we deprecated in v0.6.0, and finally removed all support for in v2.0.0. I think something in the code inadvertently supported your hybrid API response, so no one noticed it was wrong. We can provide a fix for this in v2.0.1.

derrickstolee:
Thanks for the details about the LFS API deprecation. We will work to update our API usage on the live site, but this will also affect our on-prem customers, so we appreciate the work to do a client-side fix for 2.0.1.

derrickstolee:
[...]

PR #2004, opened 2017-03-06
tq: fallback to _links if present

ttaylorr:
This pull-request teaches the `tq` package how to fallback to the deprecated `_links` property if it is given instead of `actions` in a batch response. It is intended to address https://github.com/git-lfs/git-lfs/issues/2003.

This functionality is implemented through a new, private type: `multiActionSet`, which behaves like `io.Multi{Reader,Writer}`, except that it instead implements the `ActionSet` interface. When the `HybridActions()` function is called, a `multiActionSet` is composed together, making both `actions` and `_links` accessible.

Unfortunately, the public interface for `tq.Transfer` now includes:

- `tq.Transfer.Actions`
- `tq.Transfer.Links`
- `tq.Transfer.HybridActions()`

Which, to me, feels bloated. `Actions` and `Links` are necessarily left public so as to allow the `encoding/json` package to modify those fields via the `reflect` package.  

EDIT: One thing we could do is make the `HybridActions()` function private in the interim, until this behavior is no longer supported.

Once this merges, I'll go ahead and backport this into the `release-2.0` branch, and schedule a release of v2.0.1.

---

/cc @git-lfs/core 
/xref https://github.com/git-lfs/git-lfs/issues/2003

### DIFF
--- a/tq/basic_download.go
+++ b/tq/basic_download.go
@@ -88,7 +88,7 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
 		defer dlFile.Close()
 	}
 
-	rel, err := t.Actions.Get("download")
+	rel, err := t.HybridActions().Get("download")
 	if err != nil {
 		return err
 	}
--- a/tq/basic_download.go
+++ b/tq/basic_download.go
@@ -91,7 +91,9 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
 	rel, err := t.Actions.Get("download")
 	if err != nil {
 		return err
-		// return errors.New("Object not found on the server.")
+	}
+	if rel == nil {
+		return errors.New("Object not found on the server.")
 	}
 
 	req, err := a.newHTTPRequest("GET", rel)
--- a/tq/basic_download.go
+++ b/tq/basic_download.go
@@ -93,7 +93,7 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
 		return err
 	}
 	if rel == nil {
-		return errors.Errorf("Object %s not found on the server.", t.Oid[:7])
+		return errors.Errorf("Object %s not found on the server.", t.Oid)
 	}
 
 	req, err := a.newHTTPRequest("GET", rel)
--- a/tq/basic_download.go
+++ b/tq/basic_download.go
@@ -88,7 +88,7 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
 		defer dlFile.Close()
 	}
 
-	rel, err := t.HybridActions().Get("download")
+	rel, err := t.Rel("download")
 	if err != nil {
 		return err
 	}
--- a/tq/basic_download.go
+++ b/tq/basic_download.go
@@ -93,7 +93,7 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
 		return err
 	}
 	if rel == nil {
-		return errors.New("Object not found on the server.")
+		return errors.Errorf("Object %s not found on the server.", t.Oid[:7])
 	}
 
 	req, err := a.newHTTPRequest("GET", rel)
--- a/tq/transfer.go
+++ b/tq/transfer.go
@@ -104,7 +104,7 @@ const (
 func (as ActionSet) Get(rel string) (*Action, error) {
 	a, ok := as[rel]
 	if !ok {
-		return nil, &ActionMissingError{Rel: rel}
+		return nil, nil
 	}
 
 	if !a.ExpiresAt.IsZero() && a.ExpiresAt.Before(time.Now().Add(objectExpirationToTransfer)) {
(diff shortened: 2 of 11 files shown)


tq/transfer.go
```diff
@@ -33,11 +33,21 @@ type Transfer struct {
...
+	Actions       mapActionSet `json:"actions,omitempty"`
+	Links         mapActionSet `json:"_links,omitempty"`
 	Error         *ObjectError `json:"error,omitempty"`
 	Path          string       `json:"path,omitempty"`
 }
 
+// HybridActions returns an action set composed of the "actions" property, as
+// well as the "_links" property. It is implemented by a multiActionSet, and
+// therefore retains that ordering.
+func (t *Transfer) HybridActions() ActionSet {
```
technoweenie: What do you think about having a `Action()` function that checks `Actions` and `Links`? I'm not sure `multiActionSet` is necessary. Basically: `t.Actions.Get(string)` => `t.Action(string)`. I think the wrapped types and interface are a bit much for this.

tq/transfer_queue.go
```diff
@@ -344,14 +344,14 @@ func (q *TransferQueue) enqueueAndCollectRetriesFor(batch batch) (batch, error)
 					tracerx.Printf("tq: enqueue retry #%d for %q (size: %d)", count, tr.Oid, tr.Size)
 					next = append(next, t)
 				} else {
-					if !IsActionMissingError(err) {
-						q.errorc <- errors.Errorf("[%v] %v", tr.Name, err)
-					}
+					q.errorc <- errors.Errorf("[%v] %v", tr.Name, err)
```
technoweenie: I don't think this is the change we want. If a request to upload an object returns an object with no action, it will now be treated as an error, when previously it just meant that the server didn't need it anymore.

ttaylorr: Good question, but I don't think that this is the behavior that's exhibited. A missing action now returns `(nil, nil)`, which would skip this block entirely, and only execute:

```go
} else if a == nil {
      q.Skip(o.Size)
      q.wait.Done()
}
```

I should've discussed this change more in the original post, but when I implemented the `multiActionSet`, I found juggling the `IsActionMissingErr` calls to be awkward, and it made sense to just remove it entirely, since "no action, no error" corresponds to a `(nil, nil)` return value well.

technoweenie: Ohh I see. Basically substituting the `a == nil` check to see if the action exists. Carry on :+1:

tq/transfer.go
ttaylorr: _Great_ call. I suppose the general rule is that composing single-functioned interfaces together can usually be accomplished by another function, instead of another interface. That all being said, I found that we already had `Rel(name string) (a *Action, ok bool)`, which I re-purposed to be `Rel(name string) (a *Action, err error)` and take on this functionality in 3180a2d.

tq/basic_download.go
```diff
@@ -88,10 +88,12 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
...
 	}
 
-	rel, err := t.Actions.Get("download")
+	rel, err := t.Rel("download")
 	if err != nil {
 		return err
-		// return errors.New("Object not found on the server.")
+	}
+	if rel == nil {
+		return errors.New("Object not found on the server.")
```
technoweenie: Should this mention the affected `oid`? What does the full output look like from the command?

tq/transfer.go
```diff
@@ -34,17 +34,25 @@ type Transfer struct {
...
-	rel, ok := t.Actions[name]
-	return rel, ok
+	if t.Links != nil {
+		a, err := t.Links.Get(name)
+		if a != nil || err != nil {
+			return a, err
+		}
+	}
+
+	return nil, nil
```
technoweenie: I think this func could be simplified to:

```go
a, err := t.Actions.Get(name)
if a != nil || err != nil {
  return a, err
}

if t.Links == nil {
  return nil, nil
}

return t.Links.Get(name)
```

ttaylorr: I might pass on this one, if that's alright. Though a little more verbose, I like the flow of:

1. `t.Actions.Get()`
2. Check if we have `t.Links`
  - `t.Links.Get()`
3. Otherwise, `(nil, nil)`

If you feel strongly, though, I don't really mind either way 👍

tq/basic_download.go
ttaylorr: Good call, just pushed up 452bb6fb7742ec6d2e7b4d6183eea5b9815e0c56 to address. The errors are printed as-is, so before this would have just said "Object not found on the server" and then given up.

ttaylorr:
Should be ready for a final merge. 👍

tq/transfer.go
technoweenie: Sure, it works. I just like early returns in Go code to reduce extra nesting elsewhere in the function. I can't guarantee I won't point out when you can use early returns to reduce extra nesting the next time this comes up, or that I won't make this change myself the next time I have to change this function (which isn't often of course).

tq/basic_download.go
```diff
@@ -88,10 +88,12 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
...
 	}
 
-	rel, err := t.Actions.Get("download")
+	rel, err := t.Rel("download")
 	if err != nil {
 		return err
-		// return errors.New("Object not found on the server.")
+	}
+	if rel == nil {
+		return errors.Errorf("Object %s not found on the server.", t.Oid[:7])
```
technoweenie: This allows bad servers to trigger a panic by returning OID strings with 6 or fewer characters.

ttaylorr: Good catch! I just pushed up https://github.com/git-lfs/git-lfs/pull/2004/commits/2ff28591073fc8f04cb235c5d3a57228a7be6a79 to address. 👍

PR #2008, opened 2017-03-06
release: v2.0.1

ttaylorr:
/cc @git-lfs/core

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Git LFS Changelog
 
+## 2.0.1 (6 March, 2017)
+
+### Misc
+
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Git LFS Changelog
 
+## 2.0.1 (6 March, 2017)
+
+### Misc
+
+* tq: fallback to \_links if present #2007 (@ttaylorr)
```
technoweenie: In the future, I'd just wrap backticks around it like:

```
`_links`
```

Escaping with a `\` is technically correct markdown, but backticks are more readable in plain markdown, and in rendered html.

Issue #953, opened 2016-01-26
git-lfs fsck checks full history from current commit.  Should use SkipDeletedBlobs to only check current HEAD.


wmwv:
`git-lfs fsck` says that it just checks files in current HEAD.  But it instead ends up walking back through all of history.

`git-lfs fsck` ends up calling
`git rev-list --objects --do-walk <commit>`
(which walks back through history)

instead of
`git rev-list --objects --no-walk <commit>`
(which would just give the files in the current commit)

The `SkipDeletedBlogs` option in `revListShas` looks like the right option to pass to get the `--no-walk`.  But `doFsck` calls `ScanRefs` with no options.

`pointers, err := lfs.ScanRefs(ref.Sha, "", nil)`

`ScanRefs` calls `ScanRefsToChan`, which loads default options.  `ScanRefsToChan` then calls `refListShas`.  `refListShas` then calls

`git rev-list --objects --do-walk`

technoweenie:
Ah, I meant to reply to this last week. I think `git lfs ls-files` is more what you want. `fsck` is an older command, and should probably be changed. Since `ls-files` checks all the files in the working directory, maybe `fsck` should only look at files in `.git/lfs/objects`.

wmwv:
My use case is to determining when a git-lfs download did not succeed.

`git lfs ls-files` doesn't do a SHA check on the files, it just shows you the first part of the objid and checks that the path to the file works and then shows that path.

dakotahawkins:
It would be great to have this fixed, perhaps as an option. I've just been playing around with the bfg repo cleaner to make a lfs-only copy of our repo and the fsck output scared me.

jjgod:
I mean, sounds like we should just do this:

    diff --git a/commands/command_fsck.go b/commands/command_fsck.go
    index ee9e4b5..29947a2 100644
    --- a/commands/command_fsck.go
    +++ b/commands/command_fsck.go
    @@ -47,7 +47,7 @@ func fsckCommand(cmd *cobra.Command, args []string) {
             }
         })
     
    -    if err := gitscanner.ScanRefWithDeleted(ref.Sha, nil); err != nil {
    +    if err := gitscanner.ScanRef(ref.Sha, nil); err != nil {
             ExitWithError(err)
         }

ttaylorr:
👍

PR #2022, opened 2017-03-10
Fix a number of issues with locking and fsck

TheJare:
- fsck only scans current version of objects
- Files are made readonly after unlocking them, by path or id
- Locking paths are relative to current directory, not to repo root
- Fixed errors with unable to stat files if cwd is not the repo root
- Fixed error on OSX when filesystem path to repo contains uppercase letters. Sometimes $PWD would have them in lowercase, OSX filesystem preserves case but is case insensitive, however the various path functions in Go do not do this. Changed to ask Git itself for the repo-relative paths.

### DIFF
--- a/commands/command_fsck.go
+++ b/commands/command_fsck.go
@@ -47,7 +47,7 @@ func fsckCommand(cmd *cobra.Command, args []string) {
 		}
 	})
 
-	if err := gitscanner.ScanRefWithDeleted(ref.Sha, nil); err != nil {
+	if err := gitscanner.ScanRef(ref.Sha, nil); err != nil {
 		ExitWithError(err)
 	}
 
--- a/git/git.go
+++ b/git/git.go
@@ -615,6 +615,32 @@ func GitAndRootDirs() (string, string, error) {
 	return absGitDir, absRootDir, nil
 }
 
+func PathFromRoot(pattern string) ([]string, error) {
+	safePattern := sanitizePattern(pattern)
+
+	var ret []string
+	cmd := subprocess.ExecCommand("git",
+		"-c", "core.quotepath=false", // handle special chars in filenames
+		"ls-files",
+		"--full-name",
+		"-cdmok", // Admit files in any state, let the lock fail afterwards if not valid
+		"--",       // no ambiguous patterns
+		safePattern)
+
+	outp, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to call git ls-files: %v", err)
+	}
+	cmd.Start()
+	scanner := bufio.NewScanner(outp)
+	for scanner.Scan() {
+		line := scanner.Text()
+		ret = append(ret, strings.TrimSpace(line))
+	}
+	return ret, cmd.Wait()
+
+}
+
 func RootDir() (string, error) {
 	cmd := subprocess.ExecCommand("git", "rev-parse", "--show-toplevel")
 	out, err := cmd.Output()
--- a/locking/locks.go
+++ b/locking/locks.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/git-lfs/git-lfs/errors"
 	"github.com/git-lfs/git-lfs/filepathfilter"
+	"github.com/git-lfs/git-lfs/git"
 	"github.com/git-lfs/git-lfs/lfsapi"
 	"github.com/git-lfs/git-lfs/tools"
 	"github.com/git-lfs/git-lfs/tools/kv"
@@ -107,8 +108,13 @@ func (c *Client) LockFile(path string) (Lock, error) {
 		return Lock{}, errors.Wrap(err, "lock cache")
 	}
 
+	gitRoot, err := git.RootDir()
+	if err != nil {
+		return Lock{}, err
+	}
+	absPath := filepath.Join(gitRoot, path)
 	// Ensure writeable on return
-	if err := tools.SetFileWriteFlag(path, true); err != nil {
+	if err := tools.SetFileWriteFlag(absPath, true); err != nil {
 		return Lock{}, err
 	}
 
@@ -129,10 +135,6 @@ func (c *Client) UnlockFile(path string, force bool) error {
 		return err
 	}
 
-	// Make non-writeable if required
-	if c.SetLockableFilesReadOnly && c.IsFileLockable(path) {
-		return tools.SetFileWriteFlag(path, false)
-	}
 	return nil
 
 }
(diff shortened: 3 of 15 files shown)


commands/command_fsck.go
```diff
@@ -47,7 +47,7 @@ func fsckCommand(cmd *cobra.Command, args []string) {
 		}
 	})
 
-	if err := gitscanner.ScanRefWithDeleted(ref.Sha, nil); err != nil {
+	if err := gitscanner.ScanRef(ref.Sha, nil); err != nil {
```
ttaylorr: Could you provide a little more context why this method call was changed to no longer scan through deleted objects?

My thought was that the idea of a fsck in this context was to find and remove the LFS pointers in deleted Git objects.

locking/locks.go
```diff
@@ -154,6 +157,17 @@ func (c *Client) UnlockFileById(id string, force bool) error {
 	if err := c.cache.RemoveById(id); err != nil {
 		return fmt.Errorf("Error caching unlock information: %v", err)
 	}
+	path := unlockRes.Lock.Path
+	gitRoot, err := git.RootDir()
+	if err != nil {
+		return err
+	}
+	absPath := filepath.Join(gitRoot, path)
```
ttaylorr: The locking API guarantees that paths returned from a spec-compliant server will be absolute with the root being the root directory of the git repository.

That all being said, I think this might still be a safe (and good) thing to do, _if_ we first:

- Checked to see whether or not the `unlockRes.Lock.Path` is absolute (i.e., starts with a "/").
- If so, use that as is.
- If not, make a call to `filepath.Join` with the first argument being the `gitRoot`.

I think that this can be implemented fairly cleanly as:

```go
var lockPath = unlockRes.Lock.Path
if !path.IsAbs(lockPath) {
        gitRoot, err := git.RootDir()
        if err != nil {
                return err
        }

        lockPath = filepath.Join(gitRoot, lockPath)
}

// ...
```

locking/locks.go
```diff
@@ -154,6 +157,17 @@ func (c *Client) UnlockFileById(id string, force bool) error {
...
+
+	// Make non-writeable if required
+	if c.SetLockableFilesReadOnly && c.IsFileLockable(path) {
```
ttaylorr: Good catch.

Can you add a test in the `test/test-unlock.sh` file to prevent this functionality from regressing in the future and add a note to the manpage to document this behavior?

locking/locks.go
```diff
@@ -106,8 +108,13 @@ func (c *Client) LockFile(path string) (Lock, error) {
 		return Lock{}, errors.Wrap(err, "lock cache")
 	}
 
+	gitRoot, err := git.RootDir()
+	if err != nil {
+		return Lock{}, err
+	}
+	absPath := filepath.Join(gitRoot, path)
```
ttaylorr: See same note about making relative paths absolute.

git/git.go
```diff
@@ -615,6 +615,32 @@ func GitAndRootDirs() (string, string, error) {
 	return absGitDir, absRootDir, nil
 }
 
+func PathFromRoot(pattern string) ([]string, error) {
```
ttaylorr: Awesome function: this makes the call site much simpler. I have two suggestions:

1. What do you think about changing the name of this function to be `PathsFromRoot` (plural) since it returns multiple paths?
2. What do you think about adding a `PathFromRoot` (singular) that returns `(string, error)` and additionally returns an error if more than one path matched?

In suggestion 2, I think the idea would be to allow `("", nil)` a valid return value and leave that check (`if len(path) == 0`) at the original location in commands/command_lock.go, but move the `if len(paths) > 1` check to this new function.

commands/command_fsck.go
TheJare: I am not clear on the goal of fsck, but when I ran it on my repo it complained about missing objects that were old revisions of lfs-stored files. It then tried to move the offending objects' files to the "bad" folder and, since the files of course did not exist, it complained again.

I found this looked like a reasonable fix, and then saw #953 which seems to agree. I should probably have left this in another pull request, but I'm not very well versed in proper pull requesting. :)

git/git.go
TheJare: I considered making the function just return one path or error if there are more, but it seems to me the evolution of the lock commands will/should include supporting multiple files at a time (e.g. `git lfs lock a.jpg b.jpg`, or `git lfs lock *.zip` so returning multiple paths directly and letting be up to the caller to decide about supporting multiple files or just one.

On the other hand, the function as it stands will match many more files than one would initially guess, so if one passes "a.zip" as argument I think the function will return "a.zip" files in folder and subfolders. This will only cause more trouble if git-lfs starts supporting wildcard arguments ("*" would lock the entire repo!). So this needs more design attention, but right now seems like an improvement over the previous functionality and I didn't want to complicate it with more API points.

locking/locks.go
TheJare: Not sure if this comment refers to initializing the Config.SetLockableFilesReadOnly to true by default? If so, I believe the docs already say that's how it works. I assumed the config variable (and env var) were added as part of a future method for configuring this behaviour. I noticed other places that do not test the flag and were already setting the readonly flag (despite the config defaulting to false), so there's more fixes to be done.

I'll look at adding the test.

locking/locks.go
TheJare: At the time I did this I was testing with my own fork of lfs-test-server since there were no other servers (not even github's own) that were supporting locking. I noticed yesterday as I was working on this that Github locks were starting to work (and could see how they were adding/fixing stuff on the fly as I tested my git-lfs work on Github as well as my fork). I'll review this because my fork may not be adhering to that aspect of the API.

commands/command_fsck.go
ttaylorr: Ah, thanks for linking that -- I forgot about that issue. I agree that the fsck changes could come in a separate pull request, since it seems there are more options re: different ways that that could work.

How do you feel about backing out of this change for now and moving it into a different PR?

locking/locks.go
ttaylorr: > I'll review this because my fork may not be adhering to that aspect of the API.

Sounds good. I think my preference would be to back out of this change as well, but it may be good to keep anyway, so that LFS can better handle servers that implement the locking API incorrectly.

What are your thoughts?

locking/locks.go
ttaylorr: > I noticed other places that do not test the flag and were already setting the readonly flag (despite the config defaulting to false), so there's more fixes to be done.
> 
> I'll look at adding the test.

That sounds good, this was my main area of concern.

git/git.go
ttaylorr: > I considered making the function just return one path or error if there are more, but it seems to me the evolution of the lock commands will/should include supporting multiple files at a time

I think that it should be the caller's (in this case, the `commands` package, and particularly, `commands/commands_lock.go`) to break up multiple arguments into multiple calls. I think this could look something like:

```go
for _, lpath := range args {
        path, err := git.PathFromRoot(lpath)
        // ...
}
```

That being said, I think I still prefer my original suggestion of having both:

- `PathFromRoot(pattern string) (path string, err error)`
- `PathsFromRoot(pattern string) (paths []string, err error)`

Would you be opposed to going forward with this?

PR #2023, opened 2017-03-10
Add lfs.storage parameter for overriding LFS storage location

bozaro:
**Motivation**: use single shared directory for multiple working copies on build farm. Sharing LFS storage directory allow reduce disk space usage and downloading time.


ttaylorr:
This seems like something that should definitely be worthwhile to have in LFS. Before a more in-depth review, I think it'd be good to take a look at some of the test failures that you're seeing in TravisCI.

As well, I can't think of any existing functionality in Git LFS that would be broken by having more (or already existing objects) in the `.git/lfs/objects` (or wherever `lfs.storage` is set to). That being said, I think it would be good to prove this by writing some integration tests (see the `tests/` directory) that cover a few different cases of this.

Our integration tests are just shell scripts, but they're output was certainly daunting to me when I wrote my first few. I'm happy to help guide you through the process, or answer any questions that you might have along the way.

/cc @git-lfs/core for early visibility

technoweenie:
One thing that needs to be tested is how this affects `git lfs prune`. If you're setting `lfs.storage` to have a central object directory for multiple repos, it's very possible that `git lfs prune` could delete objects that other repositories depend on. The easiest thing is to probably bail with an error if `lfs.storage` is set.

Also, how does this affect git worktrees, and reference clones (which I think set a custom git alternate)?

I think my preference would be to piggy back on git alternates instead of introducing a separate lfs config value. Though, I'd have to do some research on the feature to have more confidence in this stance :)

ttaylorr:
> One thing that needs to be tested is how this affects `git lfs prune`. If you're setting `lfs.storage` to have a central object directory for multiple repos, it's very possible that `git lfs prune` could delete objects that other repositories depend on. The easiest thing is to probably bail with an error if `lfs.storage` is set.

Ah, I didn't consider the `prune` command, that's a good call. My preference is to lean into worktrees and have include better support for that, but I would also need to do more research to know for sure.

larsxschneider:
@bozaro Are you talking about "multiple working copies" of the same repo? We use `git worktree` to have all `.git/*` (including LFS) data only once on a machine. We did not run into any issues, yet. Wouldn't that solve your problem?

sinbad:
Yeah you can already use worktrees although that only solves the problem for multiple clones of the same repo. It depends how many shared assets you have across different repos - if you're in the habit of committing large binary dependencies then that's more likely than if you're committing normal project-specific content. 

`git lfs prune` specifically caters for worktrees, which is necessary whenever you use shared storage locations as Rick pointed out. It can do because the link is bi-directional in worktrees. Any cross-repo solution would have to address the same issue to avoid destroying shared content. Alternates are a bit different because they're just a convenient fallback rather than a primary shared store (they're only ever read from). 

Unless there's a really strong case for cross-unrelated-repo shared storage I agree that it's better to rely on worktrees for storage saving.

bozaro:
We use many working copies of a small number of repositories.

We don't use `worktrees` on out build farm, because:

 * git worktrees don't supported on out CI server (Jenkins);
 * git worktrees don't allow for checkout same branch in multiple working copies;
 * git worktrees require additional steps on removing working copy.

Looks like `git lfs prune` can't work with lfs.storage by design. But in out case we can simply wipe local LFS storage every three-six months :)

bozaro:
I also use lfs.storage on my workstation to move LFS on HDD drive.
Before this option I store whole .git directory with LFS files on HDD. Now I moved .git directory to SSD drive and keep LFS files on HDD drive.

So, my current directory layout on my workstation:

 * SSD: main worktree (~203GB) and .git directory (~46GB);
 * HDD: second worktree (~187GB) and LFS storage (~218GB).

ttaylorr:
I think not having support for worktrees is a fair constraint. I'm 👍  for giving users to choose either `lfs.storage`, alternates, or worktrees.

bozaro:
> I think not having support for worktrees is a fair constraint. I'm 👍 for giving users to choose either lfs.storage, alternates, or worktrees.

In my case I use worktrees and lfs.storage on the same repository (worktrees for sharing .git content and lfs.storage for move LFS files from SSD to HDD).

technoweenie:
If this is all for the same repository, would using git alternates work? I'd prefer to lean on an existing Git feature, instead of introducing an LFS specific option like `lfs.storage`. This could even save space used up by git on your ssd and hdd. Or do you specifically want `lfs.storage`, but NOT git alternates?

bozaro:
This option fix two issues:

 * Use shared LFS storage for saving download time and disk space on build farm (and define LFS storage before "git clone");
 * Move LFS storage outside .git directory for saving disk space without and keep fast git commands (store .git on SDD).

As alternative theoretically I can use:

 * symlink for .git/lfs directory - this solution require many manual steps and minor git-lfs modifications for store lfs/tmp near lfs/objects;
 * git worktree - this solutions requre dancing for add/remove worktree. Also it don't allow checkout single branch on multiple worktrees.

In any case, I do not see a beautiful solution to the problem without option like `lfs.storage`.

technoweenie:
> Use shared LFS storage for saving download time and disk space on build farm (and define LFS storage before "git clone");

I think you can do this with [`git lfs clone --reference`](https://git-scm.com/docs/git-clone#git-clone---reference-if-ableltrepositorygt) (which uses git alternates)

> Move LFS storage outside .git directory for saving disk space without and keep fast git commands (store .git on SDD).

This we can't do with git alternates, and requires `lfs.storage`.

Back to my original reply:

> One thing that needs to be tested is how this affects `git lfs prune`. If you're setting `lfs.storage` to have a central object directory for multiple repos, it's very possible that git lfs prune could delete objects that other repositories depend on.

I misunderstood the intention of this setting. I thought you wanted a single `lfs.storage` setting to apply to multiple repositories, not multiple clones of the same repository. So, I think `git lfs prune` will continue to work fine. 

I'm okay with adding this feature. Let me do a more in-depth review of code. There are some things I'm definitely not happy about, but I think they're more issues with the design of the `config` and `localstorage` packages, and not necessarily this PR.

mudlo:
Useful improvement

Wingjam:
+1

bozaro:
Rebased to master

docs/man/git-lfs-config.5.ronn
```diff
@@ -56,6 +56,10 @@ be scoped inside the configuration for a remote.
   Enables in-memory SSH and Git Credential caching for a single 'git lfs'
   command. Default: false. This will default to true in v2.1.0.
 
+* `lfs.storage`
+
+  Allow override LFS storage directory. Default: lfs in Git repository directory.
```
ttaylorr: Do you think it would be beneficial to note that the path is relativized to inside of `.git/` if it's non-absolute?

technoweenie: It may also be worth mentioning that you should not run `git lfs prune` if you have different repositories sharing the same directory.

technoweenie:
@ttaylorr I concur, this is good for v2.3 once the doc changes make it in.

docs/man/git-lfs-config.5.ronn
bozaro: I added more information to man pages

PR #2035, opened 2017-03-15
lfs: extract `DiffIndexScanner`

ttaylorr:
This pull request extracts a `DiffIndexScanner` type that will be used by the `status` command to yield more helpful information about why certain files are modified.

Since both the proposed changes to the status command _and_ the `ScanIndex()` operation of the new gitscanner require scanning the output of the `diff-index` command, this PR is only mean to extract that common functionality.

---

/cc @git-lfs/core

### DIFF
--- a/lfs/diff_index_scanner.go
+++ b/lfs/diff_index_scanner.go
@@ -135,6 +135,7 @@ func (s *DiffIndexScanner) Scan() bool {
 	}
 
 	s.next, s.err = s.scan(s.from.Text())
+	s.err = errors.Wrap(s.err, "diff-index scan")
 
 	return s.err == nil
 }
--- a/lfs/diff_index_scanner.go
+++ b/lfs/diff_index_scanner.go
@@ -99,19 +99,19 @@ type DiffIndexScanner struct {
 // If any error was encountered in starting the command or closing its `stdin`,
 // that error will be returned immediately. Otherwise, a `*DiffIndexScanner`
 // will be returned with a `nil` error.
-func NewDiffIndexScanner(ref string, cached bool) (*DiffIndexScanner, *wrappedCmd, error) {
+func NewDiffIndexScanner(ref string, cached bool) (*DiffIndexScanner, error) {
 	cmd, err := startCommand("git", diffIndexCmdArgs(ref, cached)...)
 	if err != nil {
-		return nil, cmd, errors.Wrap(err, "diff-index")
+		return nil, errors.Wrap(err, "diff-index")
 	}
 
 	if err = cmd.Stdin.Close(); err != nil {
-		return nil, cmd, errors.Wrap(err, "diff-index: close")
+		return nil, errors.Wrap(err, "diff-index: close")
 	}
 
 	return &DiffIndexScanner{
 		from: bufio.NewScanner(cmd.Stdout),
-	}, cmd, nil
+	}, nil
 }
 
 // diffIndexCmdArgs returns a string slice containing the arguments necessary
--- a/lfs/diff_index_scanner.go
+++ b/lfs/diff_index_scanner.go
@@ -8,6 +8,11 @@ import (
 	"github.com/git-lfs/git-lfs/errors"
 )
 
+// Status represents the status of a file that appears in the output of `git
+// diff-index`.
+//
+// More information about each of its valid instances can be found:
+// https://git-scm.com/docs/git-diff-index
 type Status rune
 
 const (
@@ -21,6 +26,8 @@ const (
 	StatusUnknown      Status = 'X'
 )
 
+// String implements fmt.Stringer by returning a huamn-readable name for each
+// status.
 func (s Status) String() string {
 	switch s {
 	case StatusAddition:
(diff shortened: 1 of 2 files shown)


lfs/diff_index_scanner.go
```diff
@@ -0,0 +1,191 @@
...
+
+	"github.com/git-lfs/git-lfs/errors"
+)
+
+// Status represents the status of a file that appears in the output of `git
+// diff-index`.
+//
+// More information about each of its valid instances can be found:
+// https://git-scm.com/docs/git-diff-index
+type Status rune
```
ttaylorr: @technoweenie I think this needs a better name, but I'm not sure what to call it. Exporting this field is necessary, since it'll be used from the `commands` package, but I think adding `lfs.Status` to the public API is too generic.

I was originally thinking of renaming this to be called `lfs.DiffIndexStatus` or `lfs.DiffIndexEntryStatus`. That got me thinking, what about moving these diff index changes to the `git` package, or a new `git/diffindex` package?

There are a few helpers that we'd have to move, like `lfs.startCommand`, but I don't think that'd be a problem.

What are your thoughts?

lfs/gitscanner_index.go
```diff
@@ -106,67 +105,46 @@ func scanIndex(cb GitScannerFoundPointer, ref string) error {
...
+				sha = scanner.Entry().SrcSha
+			}
 
-			if len(description) >= 5 {
-				status := description[4][0:1]
-				sha1 := description[3]
-				if status == "M" {
-					sha1 = description[2] // This one is modified but not added
-				}
+			indexMap.Add(sha, &indexFile{
```
ttaylorr: This `*indexFile` is kept over the new `DiffIndexEntry` type, since the `Name` field (and `sha` value) aren't entirely the output from the `diff-index` command.

- `Name` refers to either the new name of the file (if renamed), or the `SrcName` field.
- `Sha` refers to either the `DstSha`, or the `SrcSha`, if the status is 'M' for modified.

lfs/diff_index_scanner.go
```diff
@@ -0,0 +1,191 @@
...
+
+	return true
+}
+
+// scan parses the given line and returns a `*DiffIndexEntry` or an error,
+// depending on whether or not the parse was successful.
+func (s *DiffIndexScanner) scan(line string) (*DiffIndexEntry, error) {
+	parts := strings.Split(line, "\t")
+	if len(parts) < 2 {
+		return nil, errors.Errorf("invalid line: %s", line)
```
ttaylorr: These errors are also generic, and could potentially be extracted into types. I think this is a bit overkill for the `lfs` package (think: `lfs.DiffIndexScannerInvalidLineErr` 😅 ), but I think would be appropriate for a new package entirely.

That being said, their messages are too generic, and should at least be wrapped with `diff-index: `. I think the most appropriate place to do this would be in the public `Scan()` function from which this function is called.

lfs/diff_index_scanner.go
```diff
@@ -0,0 +1,191 @@
...
+//
+// If "cache" is given, the DiffIndexScanner will scan for differences between
+// the given ref and the index. If "cache" is _not_ given, DiffIndexScanner will
+// scan for differences between the given ref and the currently checked out
+// tree.
+//
+// If any error was encountered in starting the command or closing its `stdin`,
+// that error will be returned immediately. Otherwise, a `*DiffIndexScanner`
+// will be returned with a `nil` error.
+func NewDiffIndexScanner(ref string, cached bool) (*DiffIndexScanner, *wrappedCmd, error) {
```
ttaylorr: Last thing: the old `ScanIndex()` function had a `cmd.Wait()` after the `bufio.NewScanner(cmd.Stdout).Scan()` loop. I don't think that's necessary, since the cmd [will have already closed `exec/cmd.Stdout`][1] by the time that `Wait()` will have returned.

[1]: https://github.com/golang/go/blob/go1.8/src/os/exec/exec.go#L448

lfs/diff_index_scanner.go
technoweenie: I'm cool with `lfs.DiffIndexStatus` for now. My rough plan was to move all gitscanner stuff to a separate package once all the refactoring is done. I don't think a `git/diffindex` package is necessary right now.

ttaylorr:
Ack, I lied in https://github.com/git-lfs/git-lfs/pull/2035#discussion_r106277999: removing `cmd.Wait()` makes the tests in `test/test-status.sh` race-y. I'll take a look now 👍 .

ttaylorr:
> Ack, I lied in #2035 (comment): removing `cmd.Wait()` makes the tests in `test/test-status.sh` race-y. I'll take a look now.

Two lies! What was causing the tests to fail was wrapping a `nil` error, causing the `errs` channel to be full and causing the command to exit without any information. I went ahead and fixed that in 42407fb.

_Now_ this should be ready to merge 😄

lfs/diff_index_scanner.go
technoweenie: What about: `Invalid diff-index output: %s`

I don't think a new error type is necessary just have one. Introduce a common error type or interface if the caller needs to do something special with these errors.

lfs/diff_index_scanner.go
```diff
@@ -0,0 +1,194 @@
...
+// depending on whether or not the parse was successful.
+func (s *DiffIndexScanner) scan(line string) (*DiffIndexEntry, error) {
+	parts := strings.Split(line, "\t")
+	if len(parts) < 2 {
+		return nil, errors.Errorf("invalid line: %s", line)
+	}
+
+	desc := strings.Fields(parts[0])
+	if len(desc) < 5 {
+		return nil, errors.Errorf("invalid description: %s", parts[0])
```
technoweenie: How about `invalid diff-index description: %s`?

lfs/diff_index_scanner.go
```diff
@@ -0,0 +1,194 @@
...
+	entry := &DiffIndexEntry{
+		SrcMode: strings.TrimPrefix(desc[0], ":"),
+		DstMode: desc[1],
+		SrcSha:  desc[2],
+		DstSha:  desc[3],
+		Status:  DiffIndexStatus(rune(desc[4][0])),
+		SrcName: parts[1],
+	}
+
+	if score, err := strconv.Atoi(desc[4][1:]); err != nil {
```
technoweenie: I'm not sure of the chances, but nothing is really confirming that `desc[4]` has at least 1 character.

lfs/gitscanner_index.go
```diff
@@ -106,67 +105,44 @@ func scanIndex(cb GitScannerFoundPointer, ref string) error {
...
 	revs := make(chan string, chanBufSize)
-	errchan := make(chan error, 1)
+	errs := make(chan error, 1)
 
 	go func() {
-		scanner := bufio.NewScanner(cmd.Stdout)
 		for scanner.Scan() {
-			// Format is:
-			// :100644 100644 c5b3d83a7542255ec7856487baa5e83d65b1624c 9e82ac1b514be060945392291b5b3108c22f6fe3 M foo.gif
-			// :<old mode> <new mode> <old sha1> <new sha1> <status>\t<file name>[\t<file name>]
```
technoweenie: I like these old comments. Can you add them to the new `scan()` func somewhere?

lfs/diff_index_scanner.go
ttaylorr: It's subtle, but using `strings.Fields` does guarantee that at each field is at least one, non-whitespace character long.

Since we check that the `len(desc)` is at least 5, we're guaranteed to have at least 5 non-empty slice elements.

[Here](https://play.golang.org/p/uvyIM5YDnt) is an example.

lfs/gitscanner_index.go
ttaylorr: Good call, I added those back in e16214c.

ttaylorr:
Thanks for the review, @technoweenie. I addressed your concern about `desc[4]` having no characters, and added the comments back from https://github.com/git-lfs/git-lfs/pull/2035#discussion_r106455809.

I think I may pass on rewording the error messages. I wrote them assuming that the caller would re-wrap them with `errors.Wrap(scanner.Error(), "diff-index")`, so the output would be:

- `diff-index: invalid line: ...`
- `diff-index: invalid description: ...`

Which I think is clear and not overly specific. Do you think this is ready to merge?

ttaylorr:
:wave: Thanks for the review!

PR #2042, opened 2017-03-16
status: use DiffIndexScanner to populate results

ttaylorr:
This pull-request brings us another step closer to adding additional information to the output of the `status` command by using the new `*lfs.DiffIndexScanner` instead of the `gitscanner`.

By doing so, we give the status command direct access to the `*lfs.DiffIndexScanner` instance, which will eventually receive methods that can answer and provide additional information about why certain files are modified. 

---

/cc @git-lfs/core

### DIFF
--- a/commands/command_status.go
+++ b/commands/command_status.go
@@ -30,33 +30,9 @@ func statusCommand(cmd *cobra.Command, args []string) {
 
 	statusScanRefRange(ref)
 
-	entries, errs, err := scanIndex(scanIndexAt)
+	staged, unstaged, err := scanIndex(scanIndexAt)
 	if err != nil {
 		ExitWithError(err)
-	}
-
-	staged := make([]*lfs.DiffIndexEntry, 0)
-	unstaged := make([]*lfs.DiffIndexEntry, 0)
-
-L:
-	for {
-		select {
-		case entry, ok := <-entries:
-			if !ok {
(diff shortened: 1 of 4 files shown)


commands/command_status.go
```diff
@@ -30,41 +30,99 @@ func statusCommand(cmd *cobra.Command, args []string) {
...
-			Print("\t%s", p.Name)
-		}
+	for _, entry := range unstaged {
+		Print("\t%s", entry.SrcName)
 	}
 
 	Print("")
 }
 
+func scanIndex(ref string) (<-chan *lfs.DiffIndexEntry, <-chan error, error) {
```
technoweenie: Does this really need to return 2 channels? It means every caller has to setup a complex `select` with a goto to use it. What do you think about returning two slices of diff entries instead? The non-porcelain status calls are collecting them into two slices anyway. Also, `scanIndex()` would be a convenient spot to do any post-scan operations, such as sorting, or bulk `cat-file` calls if needed.

ttaylorr: @technoweenie good call. My original intent was that this could be consumed asynchronously in case the number of `diff-index` entries was large, but since they get collected into a slice anyway, I don't think it matters.

I went ahead and made this change in 0fb059d, and I think that this PR is now ready to merge. 👍

Issue #2018, opened 2017-03-09
On checkout git-lfs process execute git-lfs-authenticate for each downloading file


bozaro:
I use Git v2.12.0 and Git LFS v2.0.0.
In this configuration filter-process used for git-lfs intergration (single git-lfs for one git command).

In this case git-lfs can store git-lfs-authenticate result in process state before it expired. But git-lfs run git-lfs-authenticate for every downloading file. On Windows this command taked too long time even on local environment (~0.8 sec in my case).

[git-lfs.trace.txt](https://github.com/git-lfs/git-lfs/files/830234/git-lfs.trace.txt)

technoweenie:
It looks like it's happening on each batch, not each file. For some reason, your batch size is 1 though. Are you scanning multiple refs with just 1 object? By default, LFS has a batch size of 100 objects. But it will send the remainder when the transfer queue is shut down. The transfer queue should emit a trace message like "tq: running as batched queue, batch size of 100", which should tell you its batch size.

Here's an example trace from `git lfs fetch --all` on a small repository:

```
trace git-lfs: tq: running as batched queue, batch size of 100
trace git-lfs: fetch bin/b.bin [0263829989b6fd954f72baaf2fc64bc2e2f01d692d4de72986ea808f6e99813f]
trace git-lfs: fetch bin/hi.bin [98ea6e4f216f2fb4b69fff9b3a44842c38686ca685f3f55dc48c5d3fb1107be4]
trace git-lfs: fetch gif/atom-undo.gif [0350ec3d1c0bd3e909010ed6056fc897ffbb199d22087bd53b0dac96f9be1fc3]
trace git-lfs: fetch gif/droidtocat.gif [d1c8fab51418ef587fcf5dd8e73c60b9700704e2f8f5292ea12ec27c285b23a3]
trace git-lfs: fetch jpg/kabuki-tattoo-left.JPG [e1732d149d262963996ca77a47a5d3c51c85c6e97187afad370a8bef2828b6b7]
trace git-lfs: fetch png/render.png [55d51edb307ed7bf4a1bbca3867d132a07f76828d57f37e0f31f712c02ceafe0]
trace git-lfs: fetch bar.txt [7d865e959b2466918c9863afca942d0fb89d7c9ac0c99bafc3749504ded97730]
trace git-lfs: fetch gif/atom-undo.gif [b9f86fab477109565871ced361ba69f2425a91fbe6057fa7a9629a8d536d7c71]
[...]

PR #2045, opened 2017-03-17
Cache git-lfs-authenticate repsonse

bozaro:
Changes summary:

 * lfsapi.NewRequest now returns RequestFactory instead of (*http.Request, err);
 * RequestFactory interface have two methods:
   * NewRequest() (*http.Request, error) - generate request;
   * InvalidateAuthorization() bool - invalidate cached authentication data.
 * For simple HTTP requests RequestFactory simply wraps http.Request.
 * For SSH based HTTP requests RequestFactory generate http.Request based on
   last `git-lfs-authenticate` response.

This change for fix issue #2018

### DIFF
--- a/lfsapi/auth.go
+++ b/lfsapi/auth.go
@@ -20,10 +20,15 @@ var (
 	defaultEndpointFinder   = NewEndpointFinder(nil)
 )
 
-func (c *Client) DoWithAuth(remote string, req *http.Request) (*http.Response, error) {
+func (c *Client) DoWithAuth(remote string, factory RequestFactory) (*http.Response, error) {
 	credHelper := c.Credentials
 	if credHelper == nil {
 		credHelper = defaultCredentialHelper
+	}
+
+	req, err := factory.NewRequest()
+	if err != nil {
+		return nil, err
 	}
 
 	netrcFinder := c.Netrc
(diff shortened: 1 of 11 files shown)


technoweenie:
Interesting approach, but I think a better approach would be wrapping the credential and ssh code with a write-through caching layer. For example, there's already a [`CredentialHelper` interface](https://github.com/git-lfs/git-lfs/blob/master/lfsapi/creds.go#L10-L14) that's used for tests. That way, the interfaces to making HTTP requests doesn't have to change everywhere. What do you think about that?

Also, the SSH auth cache doesn't take `expires_at` into account. So it's possible that requests made after the expiration would be stuck using a bad token.

bozaro:
> Also, the SSH auth cache doesn't take expires_at into account. So it's possible that requests made after the expiration would be stuck using a bad token.

It not a problem: `expire_at` is useless, because nobody guaranty time syncronization between client and lfs-server hosts.
On expired token usage, server should return 401 error and client reinvoke `git-lfs-authenticate` command for new token.

bozaro:
> Interesting approach, but I think a better approach would be wrapping the credential and ssh code with a write-through caching layer. For example, there's already a CredentialHelper interface that's used for tests. That way, the interfaces to making HTTP requests doesn't have to change everywhere. What do you think about that?

I wanted to use CredentialHelper, but `git-lfs-authenticate` returns not credentional, but request URL and headers. And there are no stable paths in `git-lfs-authenticate` response. It can change any information.

ttaylorr:
> > Also, the SSH auth cache doesn't take expires_at into account. So it's possible that requests made after the expiration would be stuck using a bad token.
> 
> It not a problem: `expire_at` is useless, because nobody guaranty time syncronization between client and lfs-server hosts.

LFS exhibits this behavior in other places where it will retry requests that have actions that are already expired instead of receiving a 401 and retrying the request later down the pipeline. If the client and server have drastically different times, the request retry budget will be exceeded and the request will fail.

I think it would be worthwhile to keep that behavior consistent here and respect the `expires_at` property.

> I wanted to use CredentialHelper, but `git-lfs-authenticate` returns not credentional, but request URL and headers.

I _think_ it would be fine to also return credentials from this function.

technoweenie:
I think the ssh auth can have its own custom interface that returns a (cached) `href` + `header` + `expires_at`. It doesn't have to use the credential helper interface.

ttaylorr:
Closing since we merged #2080./

Issue #2046, opened 2017-03-17
Git LFS v2.0.x fails when pushing video files to bitbucket lfs origin


andronocean:
I'm experiencing an issue with pushing several video files tracked with LFS to a Bitbucket origin under versions 2.0.0 and 2.0.1. Everything works fine when I downgrade to v1.5.6.

In detail, I have a `videos/` directory in my repo with a number of files, ranging from 7 to 90 MB, in .webm, .mp4, .swf, .ogg, and .flv formats. I can commit locally without issue, and git lfs is happily tracking `videos/*`. On `git push` with 2.0.x I get the following printout:

```
$ git push
Remote "origin" does not support the LFS locking API. Consider disabling it with:
  $ git config 'lfs.https://bitbucket/telstarevents/saratoga.git/info/lfs.locksverify' false
Git LFS: (0 of 10 files) 85.69 MB / 338.78 MB                                  panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x13017bc]

goroutine 85 [running]:
github.com/git-lfs/git-lfs/tq.verifyUpload(0xc42010e300, 0xc4203ec900, 0xc420412440, 0x40)
	/private/tmp/git-lfs-20170307-63219-522lj3/git-lfs-2.0.1/src/github.com/git-lfs/git-lfs/tq/verify.go:43 +0x28c
github.com/git-lfs/git-lfs/tq.(*customAdapter).DoTransfer(0xc4204128c0, 0x13e1a40, 0xc420412ac0, 0xc4203ec900, 0xc420480d20, 0xc42050a8a0, 0x0, 0x0)
	/private/tmp/git-lfs-20170307-63219-522lj3/git-lfs-2.0.1/src/github.com/git-lfs/git-lfs/tq/custom.go:320 +0x71e
github.com/git-lfs/git-lfs/tq.(*adapterBase).worker(0xc42011ed00, 0x0, 0x13e1a40, 0xc420412ac0)
	/private/tmp/git-lfs-20170307-63219-522lj3/git-lfs-2.0.1/src/github.com/git-lfs/git-lfs/tq/adapterbase.go:162 +0x5cb
created by github.com/git-lfs/git-lfs/tq.(*adapterBase).Begin
[...]

PR #1971, opened 2017-02-22
tq: retry verify requests a configurable number of times

ttaylorr:
This pull-request implements `lfs.transfer.maxverifies`, a `.gitconfig` setting to retry verify requests at the end of an upload a configurable number of times.

Closes: #1781.

---

/cc @git-lfs/core

PR #2047, opened 2017-03-17
tq: use initialized lfsapi.Client instances in transfer adapters

ttaylorr:
This pull request fixes a bug introduced in https://github.com/git-lfs/git-lfs/commit/0677bdd06a97b60455dac2cfa19d54ff697a1e17 (via: [this blame view][1]) where un-initialized `lfsapi.Client` instances were used in place of the [real one][2].

Previously, this wasn't a problem, but since I added #1971 (which uses the API client), this caused a nil pointer de-reference `panic()` on pushes with verify actions using the Tus and custom transfer adapters.

@git-lfs/releases I think that this warrants being backported and released as v2.0.2. In the interim, or if we decided that this isn't worth it, we can always have users set the `lfs.basictransfersonly`. What do you think?

---

/cc @git-lfs/core https://github.com/git-lfs/git-lfs/issues/2046

[1]: https://github.com/git-lfs/git-lfs/blame/v2.0.1/tq/custom.go#L318-L323
[2]: https://github.com/git-lfs/git-lfs/blob/v2.0.1/tq/adapterbase.go#L67-L67

### DIFF
--- a/test/test-custom-transfers.sh
+++ b/test/test-custom-transfers.sh
@@ -61,6 +61,7 @@ begin_test "custom-transfer-upload-download"
   {
     \"CommitDate\":\"$(get_date -10d)\",
     \"Files\":[
+      {\"Filename\":\"verify.dat\",\"Size\":18,\"Data\":\"send-verify-action\"},
       {\"Filename\":\"file1.dat\",\"Size\":1024},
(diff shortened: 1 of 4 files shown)


ttaylorr:
Added 43f1109 & 20592ed to ensure that object verification is tested as a part of the integration test for the `custom` and `tus` adapters.

PR #2049, opened 2017-03-19
fsck only scans current version of objects

TheJare:
I am not clear on the goal of fsck, but when I ran it on my repo it complained about missing objects that were old revisions of lfs-stored files. It then tried to move the offending objects' files to the "bad" folder and, since the files of course did not exist, it complained again.

I found this looked like a reasonable fix, and then saw #953 which seems to agree.

### DIFF
--- a/commands/command_fsck.go
+++ b/commands/command_fsck.go
@@ -47,7 +47,7 @@ func fsckCommand(cmd *cobra.Command, args []string) {
 		}
 	})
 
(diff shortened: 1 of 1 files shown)


amed299:
[git-lfs-darwin-386-2.0.2.tar.gz](https://github.com/git-lfs/git-lfs/files/948376/git-lfs-darwin-386-2.0.2.tar.gz)

PR #2070, opened 2017-03-23
commands/status: add blob info to each entry

ttaylorr:
This pull-request makes it easier to diagnose why certain files are modified by LFS by adding blob info to each status line entry in the `git lfs status` command.

Here's what the output looks like:

```
On branch master

Git LFS objects to be committed:

    file2.dat (LFS: eeaf82b)
    file3.dat (LFS: a12942e)

Git LFS objects not staged for commit:

    file1.dat (LFS: 1307990 -> Git: 8735179)
    file3.dat (Git: 0dc8537)
```

The `()` shows some meta information about each entry's source and destination. For each end, it shows whether the file is tracked by Git or LFS, and the contents SHA of each. The contents SHA is either the `shasum -a 256` of the blob's contents (NB: _not_ the blob SHA itself), or the pointer's OID (if the file is tracked by LFS).

The extra info contained between the `()` comes from the following rules:

1. If the file is new (marked as `A` for addition), show a) the source (either `Git:` or `LFS:`) & the blob contents SHA.
2. If the file is not new, is a copy, a modification, or etc., show both the source and destination (either: `Git:` or `LFS:`) as well as the blob contents SHA for each.

The blob contents SHA is defined as:

- If the file is handled by LFS and is _currently_ parse-able as an LFS pointer, mark it as LFS. A file is parse-able as an LFS pointer if it wasn't checked out (using the `-I` or `-X` flags), or is staged into the index and tracked with the LFS `filter` in your repository's `.gitattributes`.
- If the file is not tracked by LFS, or is tracked by LFS and is not yet staged into the index (meaning the LFS has not yet `clean`-ed the file) the file is tracked by Git. If the file is tracked by LFS, but isn't parse-able as an LFS pointer (meaning the `clean`/`smudge` filters were skipped, or the file was already committed as a large object into the Git history), show the file as having a `Git:` source. 

A few things that I'd like to address beyond this pull-request:

1. Since #2042, we're showing _all_ files (not just those tracked by LFS) in the status output. I think this makes sense since we want to show files that are supposed to be tracked by LFS but aren't. One thing we could do is only show files which match the patterns in `.gitattributes`, but I like the complete-ness of what we have right now.
2. Files are shown as coming from `Git:` if they are tracked by LFS and not yet staged. This is technically correct, but we could show a better prefix, or try and figure out if they _would_ be tracked by LFS. I don't know offhand of a complete way to do this, since we not only have to check the contents of the repository's `.gitattributes` file(s), but also whether or not the user has a valid LFS clean/smudge/process filter installed.
3. `diff-index` doesn't show new, untracked files. This may have always been a problem with the status command, but deserves investigation nonetheless. 

---

/cc @git-lfs/core

### DIFF
--- a/commands/command_status.go
+++ b/commands/command_status.go
@@ -135,7 +135,7 @@ func blobInfo(s *lfs.CatFileBatchScanner, blobSha, name string) (sha, from strin
 		return "", "", err
 	}
 
-	return fmt.Sprintf("%x", shasum.Sum(nil)), "Git", nil
+	return fmt.Sprintf("%x", shasum.Sum(nil)), "File", nil
 }
 
 func scanIndex(ref string) (staged, unstaged []*lfs.DiffIndexEntry, err error) {
--- a/commands/command_status.go
+++ b/commands/command_status.go
@@ -1,7 +1,11 @@
 package commands
 
 import (
+	"crypto/sha256"
 	"fmt"
+	"io"
+	"os"
+	"regexp"
 	"strings"
 
 	"github.com/git-lfs/git-lfs/git"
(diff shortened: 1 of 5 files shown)


lfs/gitscanner_catfilebatch.go
```diff
@@ -65,29 +60,73 @@ func runCatFileBatch(pointerCh chan *WrappedPointer, lockableCh chan string, loc
...
 	return s.pointer
 }
 
-func (s *catFileBatchScanner) Err() error {
+func (s *CatFileBatchScanner) Err() error {
 	return s.err
 }
 
-func (s *catFileBatchScanner) Scan() bool {
+func (s *CatFileBatchScanner) Scan(sha []byte) bool {
```
technoweenie: I like this improvement of calling `Scan(oid)`, instead of manually writing to stdin. Why does it take a `[]byte` though? The callers I saw get a sha as a string.

lfs/gitscanner_catfilebatch.go
```diff
@@ -100,40 +139,53 @@ func (s *catFileBatchScanner) Scan() bool {
...
-		return "", nil, errors.Wrap(fmt.Errorf("Invalid: %q", string(l)), "git cat-file --batch")
+		return "", "", nil, errors.Wrap(fmt.Errorf("Invalid: %q", string(l)), "git cat-file --batch")
 	}
 
 	blobSha := string(fields[0])
 	size, _ := strconv.Atoi(string(fields[2]))
+	sha := sha256.New()
 	buf := make([]byte, size)
-	read, err := io.ReadFull(s.r, buf)
+	read, err := io.ReadFull(io.TeeReader(s.r, sha), buf)
```
technoweenie: Using the CatFileBatchScanner in `git lfs status` will cause it to load potentially huge blobs in memory. This doesn't happen in normal git scanning because the `CatFileBatchCheckScanner` filters any blobs over `lfs.blobSizeCutoff = 1024`.

Is it possible to skip the `buf` init if the the blob size here is also over that threshold?

commands/command_status.go
```diff
@@ -36,22 +40,102 @@ func statusCommand(cmd *cobra.Command, args []string) {
...
+		return "", "", err
+	}
+	defer f.Close()
+
+	shasum := sha256.New()
+	if _, err = io.Copy(shasum, f); err != nil {
+		return "", "", err
+	}
+
+	return fmt.Sprintf("%x", shasum.Sum(nil)), "Git", nil
```
technoweenie: Since this is reading from the local fs, and not git, what about `"New"` or even `"File"` as a better prefix?

lfs/gitscanner_catfilebatch.go
ttaylorr: Good call. Since all of the callers have `string`s when they call `Scan()` I went ahead and made this change in: https://github.com/git-lfs/git-lfs/pull/2070/commits/b51d64941b3134c1557f8f7b1adc2430a14de43a.

commands/command_status.go
ttaylorr: I dig it: https://github.com/git-lfs/git-lfs/pull/2070/commits/3a858da31610423b022784f9f41509d156189167

lfs/gitscanner_catfilebatch.go
ttaylorr: Great catch -- I made sure that we only read into `buf` is `size <= blobSizeCutoff` in https://github.com/git-lfs/git-lfs/pull/2070/commits/844c0b0db2bf3d39f52773e5f63ee10683a28342.

PR #2077, opened 2017-03-23
commands/{,un}track: extract {,un}EscapeTrackPattern()

ttaylorr:
This pull request extracts two functions: `escapeTrackPattern()` and `unescapeTrackPattern()` which correctly handles both spaces (`" "` -> `"[[:space]]"`) and the pound symbol (`"#"` -> `"\#"`).

It also fixes an incorrect regex in `assert_attributes_count` and un-escapes patterns when calling `git lfs untrack`, so that instead of:

```bash
$ git lfs untrack " "
Untracked [[:space:]]
```

you get instead:

```bash
$ git lfs untrack " "
Untracked  
```

I'm thinking that we should quote the pattern that comes out to make it clear what's going on with whitespace-only patterns.

This PR also fixes https://github.com/git-lfs/git-lfs/issues/2041.

---

/cc @git-lfs/core https://github.com/git-lfs/git-lfs/issues/2041

### DIFF
--- a/commands/command_untrack.go
+++ b/commands/command_untrack.go
@@ -57,7 +57,7 @@ func untrackCommand(cmd *cobra.Command, args []string) {
 
 		path := strings.Fields(line)[0]
 		if removePath(path, args) {
-			Print("Untracking %s", path)
+			Print("Untracking %s", unescapeTrackPattern(path))
 		} else {
 			attributesFile.WriteString(line + "\n")
 		}
@@ -66,7 +66,7 @@ func untrackCommand(cmd *cobra.Command, args []string) {
 
 func removePath(path string, args []string) bool {
 	for _, t := range args {
-		if path == t {
+		if path == escapeTrackPattern(t) {
 			return true
 		}
 	}
--- a/commands/command_untrack.go
+++ b/commands/command_untrack.go
@@ -57,7 +57,7 @@ func untrackCommand(cmd *cobra.Command, args []string) {
 
 		path := strings.Fields(line)[0]
 		if removePath(path, args) {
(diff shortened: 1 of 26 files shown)


ttaylorr:
Incidentally, Git doesn't care whether a `#` appears in the middle of a pattern escaped or unescaped, so we're OK to do global replaces (`strings.Replace(str, from, to, -1)`) on all patterns so far.

commands/command_untrack.go
```diff
@@ -57,7 +57,7 @@ func untrackCommand(cmd *cobra.Command, args []string) {
 
 		path := strings.Fields(line)[0]
 		if removePath(path, args) {
-			Print("Untracking %s", path)
+			Print("Untracking %s", unescapeTrackPattern(path))
```
technoweenie: > I'm thinking that we should quote the pattern that comes out to make it clear what's going on with whitespace-only patterns.

I agree, even if a whitespace-only pattern is probably not that useful :) Can you change this and the "tracking" message to use `%q`?

ttaylorr:
> Can you change this and the "tracking" message to use `%q`?

Ha, turns out we have a _lot_ of tests that assert on the output of the `track` command:

```bash
~/g/git-lfs (track-escape-patterns) $ find test | grep -e "^test/test-.*\.sh$" | xargs ack "(Unt|T)racking" | wc -l
      68
```

Anyway, using the quoted output in https://github.com/git-lfs/git-lfs/pull/2077/commits/624d85684d232ba98c3418dcd5c66bbf61118ea1 & https://github.com/git-lfs/git-lfs/pull/2077/commits/b61a8d4524a625b506638f806edeeae6ca9ab187.

Issue #2063, opened 2017-03-21
Be more forgiving of missing objects


technoweenie:
Currently, if a repository is missing LFS data for older versions of files, it's impossible to push it to a new endpoint. I think the `git lfs push` and `pre-push` commands should spew errors if it can't find an object locally, without halting the push operation early. The commands should still return with a non-zero exit code, but at least the other objects have a chance to be uploaded. This allows repositories with missing objects to be easily fixed without resorting to `git filter-branch`.

This is also a small baby step towards Object Expiration (as described in https://github.com/git-lfs/git-lfs/issues/1101#issuecomment-260769632), which would let repository owners mark certain OIDs as known missing objects. By marking an object as expired, the `push` and `pre-push` commands will just skip the upload attempt, so the commands can return successfully.

Finally, double check that upload retries reset the progress meter properly.

ttaylorr:
👋  (ack-ing out of our internal first responder rotation).

ttaylorr:
Closed via #2078.

technoweenie:
I don't think the push issues were ever addressed. We just decided against an explicit `--strict-mode`.

This is what I'd like for v2.2.0:

* `push` and `pre-push` attempt to upload everything.
* If `lfs.allowIncomplePush` is enabled, return exit code of 0. Otherwise, exit code of 1 if all the objects are not uploaded. This makes it easy to set on a per-repo basis, even in `.lfsconfig`.

ttaylorr:
@technoweenie my mistake on closing this one out early. I think what you're proposing is a good idea, and I opened up #2199 to address/implement. Let me know what you think.

PR #2078, opened 2017-03-24
commands/{pre-,}push: allow missing objects

ttaylorr:
This pull request teaches the `git lfs push` & `git lfs pre-push` commands to allow missing objects by default, and enforce that they exist with the `--strict-mode` flag.

From https://github.com/git-lfs/git-lfs/issues/2063:

> The commands should still return with a non-zero exit code, but at least the other objects have a chance to be uploaded. 

I think printing a warning and exiting cleanly is OK in this scenario, where `--strict-mode` is not given, because the command did what it was supposed to do given the expectations set by the caller.

> Finally, double check that upload retries reset the progress meter properly.

I verified that this is what happens, but I'm unsure if it's required that I `q.meter.Pause()` and `q.meter.Start()` in order to log to stderr. I think that it is, but perhaps not? (cc @technoweenie)

Lastly, I think that this was originally targeted for v2.0.2 (from #2063), but perhaps it should just ship in 2.1.0 since `--strict-mode` is a new flag and this is technically a breaking change?

---

/cc @git-lfs/core #2063 
Closes: #2063

### DIFF
--- a/commands/command_pre_push.go
+++ b/commands/command_pre_push.go
@@ -12,6 +12,7 @@ import (
 
 var (
 	prePushDryRun       = false
+	prePushRequireAll   = false
 	prePushDeleteBranch = strings.Repeat("0", 40)
 )
 
@@ -50,7 +51,7 @@ func prePushCommand(cmd *cobra.Command, args []string) {
 		Exit("Invalid remote name %q", args[0])
 	}
 
-	ctx := newUploadContext(args[0], prePushDryRun)
+	ctx := newUploadContext(args[0], prePushDryRun, prePushRequireAll)
 
 	gitscanner, err := ctx.buildGitScanner()
 	if err != nil {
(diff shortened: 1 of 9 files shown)


technoweenie:
I don't think we need `--strict-mode`. The push commands should always be strict. They should just follow through the entire upload, uploading everything they can, before exiting with an error. By adding `--strict-mode`, the unsafe behavior (ignoring missing objects) is default.

ttaylorr:
> I don't think we need `--strict-mode`. The push commands should always be strict. They should just follow through the entire upload, uploading everything they can, before exiting with an error. By adding `--strict-mode`, the unsafe behavior (ignoring missing objects) is default.

Good point, what do you think about changing `--strict-mode` to `--allow-missing` and leaving it to `false` by default?

ttaylorr:
Re-reading this, I just saw:

> They should just follow through the entire upload, uploading everything they can, before exiting with an error.

I think keeping in `--allow-missing` is a safe idea, so long as we continue to let the transfer queue operate until it's processed all objects _and_ `Wait()` has been called, only then can we exit with a non-zero code if `--allow-missing` is not given.

ttaylorr:
Third take -- let's remove `--strict-mode` (and the idea of `--allow-missing`) all together, and still allow `lfs push` and `lfs pre-push` to upload all objects that it _can_ and exit in a dirty state if any of the objects in the pushset were missing.

This will allow

- All of the LFS objects that do exist to travel to the remote as early as possible
- A user to see all of the missing objects in one push instead of getting one error at a time
- A user to "fix" the missing object later, and have a smaller push size when they complete the fix.

PR #2082, opened 2017-03-24
commands,tq: specialized logging for missing/corrupt objects

ttaylorr:
This pull request is a continuation of https://github.com/git-lfs/git-lfs/pull/2078 following up on https://github.com/git-lfs/git-lfs/pull/2078#issuecomment-289071404 with https://github.com/git-lfs/git-lfs/pull/2078#issuecomment-289078761.

Turns out that our `*tq.TransferQueue` already had some early returns built in for objects that are a) missing or b) corrupt. I taught the `*tq.TransferQueue` and the `*commands.uploader` types to recognize and log them appropriately.

From https://github.com/git-lfs/git-lfs/pull/2078#issuecomment-289078761, this allows:

> - All of the LFS objects that do exist to travel to the remote as early as possible
> - A user to see all of the missing objects in one push instead of getting one error at a time
> - A user to "fix" the missing object later, and have a smaller push size when they complete the fix.

With three files, `missing.dat`, `corrupted.dat`, and `present.dat`, the output looks like this:

```bash
~/g/git-lfs-test (master) $ git push origin master
  $ git config 'lfs.http://127.0.0.1:61507/push-missing-objects.git/info/lfs.locksverify' true
Git LFS: (1 of 3 files) 7 B / 21 B
Push completed with missing objects:
  (missing) missing.dat (ffa63583dfa6706b87d284b86b0d693a161e4840aad2c5cf6b5d27c3b9621f7d)
  (corrupt) corrupt.dat (11d510e067d2cdcd7559bd86d27a2f4c20babd43670346b97af99b522c1f0075)
error: failed to push some refs to 'http://127.0.0.1:61507/push-missing-objects'
```

- `missing.dat` & `corrupted.dat` are not uploaded
- `present.dat` is uploaded to the LFS server
- No Git blobs are uploaded to the Git server

---

/cc @git-lfs/core @shana

### DIFF
--- a/commands/uploader.go
+++ b/commands/uploader.go
@@ -270,8 +270,31 @@ func uploadPointers(c *uploadContext, unfiltered ...*lfs.WrappedPointer) {
 func (c *uploadContext) Await() {
 	c.tq.Wait()
 
+	var missing = make(map[string]string)
+	var corrupt = make(map[string]string)
+
 	for _, err := range c.tq.Errors() {
-		FullError(err)
+		if malformed, ok := err.(*tq.MalformedObjectError); ok {
+			if malformed.Missing() {
+				missing[malformed.Name] = malformed.Oid
+			} else if malformed.Corrupt() {
+				corrupt[malformed.Name] = malformed.Oid
+			}
+		} else {
+			FullError(err)
+		}
+	}
+
+	if len(missing) > 0 || len(corrupt) > 0 {
+		Print("Push completed with missing objects:")
+		for name, oid := range missing {
+			Print("  (missing) %s (%s)", name, oid)
+		}
+		for name, oid := range corrupt {
+			Print("  (corrupt) %s (%s)", name, oid)
+		}
+
+		os.Exit(2)
 	}
 
 	if len(c.tq.Errors()) > 0 {
--- a/commands/uploader.go
+++ b/commands/uploader.go
@@ -272,6 +272,7 @@ func (c *uploadContext) Await() {
 
 	var missing = make(map[string]string)
 	var corrupt = make(map[string]string)
+	var others = make([]error, 0, len(c.tq.Errors()))
 
 	for _, err := range c.tq.Errors() {
 		if malformed, ok := err.(*tq.MalformedObjectError); ok {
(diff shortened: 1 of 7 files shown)


ttaylorr:
The `continuous-integration/appveyor/branch` build is failing because I pushed this branch with a different name and then deleted the remote ref before the build started.

commands/uploader.go
```diff
@@ -270,8 +270,31 @@ func uploadPointers(c *uploadContext, unfiltered ...*lfs.WrappedPointer) {
...
+			} else if malformed.Corrupt() {
+				corrupt[malformed.Name] = malformed.Oid
+			}
+		} else {
+			FullError(err)
+		}
+	}
+
+	if len(missing) > 0 || len(corrupt) > 0 {
+		Print("Push completed with missing objects:")
```
technoweenie: I think this message makes it seem the upload was successful. What about something like `LFS Upload failed:`?

commands/uploader.go
```diff
@@ -270,8 +270,31 @@ func uploadPointers(c *uploadContext, unfiltered ...*lfs.WrappedPointer) {
...
 	for _, err := range c.tq.Errors() {
-		FullError(err)
+		if malformed, ok := err.(*tq.MalformedObjectError); ok {
+			if malformed.Missing() {
+				missing[malformed.Name] = malformed.Oid
+			} else if malformed.Corrupt() {
+				corrupt[malformed.Name] = malformed.Oid
+			}
+		} else {
+			FullError(err)
```
technoweenie: What do you think about just collecting all the errors here, and not calling `FullError()`? It looks like `partitionTransfers()` can only send one other potential error:

    errors.Errorf("Git LFS: object %q has invalid size (got: %d)", t.Oid, t.Size)

Even though that's technically a server issue, I don't think it's worth halting the `push` command over. It should just be listed below.

ttaylorr:
@technoweenie thanks for the review: I pushed up https://github.com/git-lfs/git-lfs/pull/2082/commits/0be9addf8857a1ea27c60d5a4710836fa8ebef9c & https://github.com/git-lfs/git-lfs/pull/2082/commits/52f04a9b73f23ec1a3204dd2c0578f865788d14b to address.

I think going forward we can do some sort of "streaming" error reporting, where server/filesystem errors can be reported as they happen. We'd have to re-draw the progress meter below each error as we receive them so as not to create phantom progress meters, but this is doable.

I think this is an OK stopping point for now 👍 .

PR #2122, opened 2017-03-31
[WIP] git lfs migrate

ttaylorr:
This pull request spikes out a simple implementation of what a `git lfs migrate` command might look like.

PR #2127, opened 2017-04-04
git: implement `*git.ObjectScanner` type 

ttaylorr:
This pull request makes the `*lfs.CatFileBatchScanner` type responsible for only parsing LFS-specific data, and defers reading Git object data to a new type called `*git.ObjectScanner.`

In particular, this is required for #2122 in order to read object contents without running the Git plumbing command: `git-checkout-index.1`.

---

/cc @git-lfs/core for review 👀

### DIFF
--- a/lfs/gitscanner_catfilebatch.go
+++ b/lfs/gitscanner_catfilebatch.go
@@ -17,7 +17,7 @@ import (
 // if that blob is for a locked file. Any errors are sent to errCh. An error is
 // returned if the 'git cat-file' command fails to start.
 func runCatFileBatch(pointerCh chan *WrappedPointer, lockableCh chan string, lockableSet *lockableNameSet, revs *StringChannelWrapper, errCh chan error) error {
-	scanner, err := NewCatFileBatchScanner()
+	scanner, err := NewPointerScanner()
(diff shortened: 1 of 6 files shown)


lfs/gitscanner_catfilebatch.go
```diff
@@ -136,31 +107,25 @@ func (s *CatFileBatchScanner) Scan(sha string) bool {
 		return false
 	}
 
+	fmt.Println("returning true with blobsha:", b, "contents:", c, err)
```
technoweenie: Did you mean to leave this here? Should probably use `tracerx.Printf()` for any debug messages that you want to leave in...

ttaylorr:
> * `lfs.CatFileBatchScanner` -> `lfs.LFSPointerScanner`

Good call, though I think we can avoid the `LFS` prefix, since it's already in the `lfs` package.

> * `git.ObjectScanner` -> `git.CatFileBatchScanner`

I think I may leave this one as-is, since even though it does use the `git cat-file --batch` plumbing command, its intended purpose is to scan Git objects. I like that the name doesn't bind us to the implementation, even though it's not likely to change.

PR #2130, opened 2017-04-05
lfsapi,tq: relative expiration support

### LINKED ISSUE #2125: Relative expiration times
The [Batch API](https://github.com/git-lfs/git-lfs/blob/master/docs/api/batch.md#successful-responses) and [`git-lfs-authenticate` ssh call](https://github.com/git-lfs/git-lfs/blob/master/docs/api/authentication.md#ssh) both can optionally return an `expires_at` marking when an authentication token expires. The client uses this hint to know if it should attempt the request, or get a fresh token. However, this causes problems for users with incorrect times.

Proposal: Introduce `expires_in`, which returns the number of seconds the token is valid. This should take precedence over an existing `expires_at` property.

ttaylorr:
This pull-request implements the idea proposed in #2125 to allow an optional `expires_in` property in both the SSH auth API and the batch API. It is introduced as a non-breaking change.

Both the SSH auth response and batch API response types have been expanded as described in https://github.com/git-lfs/git-lfs/issues/2125#issuecomment-291539131:

```go
type T struct {
        // ...
        ExpiresAt time.Time `json:"expires_at,omitempty"`
        ExpiresIn time.Duration `json:"expires_in,omitempty"` 
}
```

and have gained a new helper method:

```go
func (t *T) IsExpiredWithin(d time.Duration) (time.Time, bool) { ... }
```

which returns whether or not the action/SSH auth response is expired, and if so, when.

The behavior for the batch API remains the same, and the SSH auth behavior only changes when `lfs.cachecredentials` which will cause a cached SSH auth response to be invalidated if it is expired.

Closes #2125.

---

/cc @git-lfs/core #2125

### DIFF
--- a/docs/api/authentication.md
+++ b/docs/api/authentication.md
@@ -8,7 +8,8 @@ can come from the following places:
 
 Git LFS will add any HTTP headers returned from the `git-lfs-authenticate`
 command to any Batch API requests. If servers are returning expiring tokens,
-they can add an `expires_at` property to hint when the token will expire.
+they can add an `expires_in` (or `expires_at`) property to hint when the token
+will expire.
 
 ```bash
 # Called for remotes like:
@@ -19,6 +20,9 @@ $ ssh git@git-server.com git-lfs-authenticate foo/bar.git download
   "header": {
     "Authorization": "RemoteAuth some-token"
   },
+
+  # optional, for expiring tokens, preferred over expires_at
+  "expires_in": 86400000000000
 
   # optional, for expiring tokens
   "expires_at": "2016-11-10T15:29:07Z"
--- a/docs/api/authentication.md
+++ b/docs/api/authentication.md
@@ -22,7 +22,7 @@ $ ssh git@git-server.com git-lfs-authenticate foo/bar.git download
   },
 
   # optional, for expiring tokens, preferred over expires_at
-  "expires_in": 86400000000000
+  "expires_in": 86400
 
   # optional, for expiring tokens
   "expires_at": "2016-11-10T15:29:07Z"
--- a/docs/api/batch.md
+++ b/docs/api/batch.md
@@ -6,8 +6,8 @@ The Batch API is used to request the ability to transfer LFS objects with the
 LFS server. The Batch URL is built by adding `/objects/batch` to the LFS server
 URL.
 
-Git remote: https://git-server.com/foo/bar
-LFS server: https://git-server.com/foo/bar.git/info/lfs
+Git remote: https://git-server.com/foo/bar</br>
+LFS server: https://git-server.com/foo/bar.git/info/lfs<br>
 Batch API: https://git-server.com/foo/bar.git/info/lfs/objects/batch
 
 See the [Server Discovery doc](./server-discovery.md) for more info on how LFS
--- a/docs/api/batch.md
+++ b/docs/api/batch.md
@@ -6,8 +6,8 @@ The Batch API is used to request the ability to transfer LFS objects with the
 LFS server. The Batch URL is built by adding `/objects/batch` to the LFS server
 URL.
 
-Git remote: https://git-server.com/foo/bar  
-LFS server: https://git-server.com/foo/bar.git/info/lfs  
+Git remote: https://git-server.com/foo/bar
+LFS server: https://git-server.com/foo/bar.git/info/lfs
 Batch API: https://git-server.com/foo/bar.git/info/lfs/objects/batch
 
 See the [Server Discovery doc](./server-discovery.md) for more info on how LFS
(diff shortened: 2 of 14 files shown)


docs/api/batch.md
```diff
@@ -6,8 +6,8 @@ The Batch API is used to request the ability to transfer LFS objects with the
 LFS server. The Batch URL is built by adding `/objects/batch` to the LFS server
 URL.
 
-Git remote: https://git-server.com/foo/bar  
-LFS server: https://git-server.com/foo/bar.git/info/lfs  
+Git remote: https://git-server.com/foo/bar
+LFS server: https://git-server.com/foo/bar.git/info/lfs
```
technoweenie: Gah, small markdown issue. The 2 trailing spaces are a line break. I suppose this could either be a markdown list, or lines could end with an explicit `<br>`.

docs/api/authentication.md
```diff
@@ -20,6 +21,9 @@ $ ssh git@git-server.com git-lfs-authenticate foo/bar.git download
     "Authorization": "RemoteAuth some-token"
   },
 
+  # optional, for expiring tokens, preferred over expires_at
+  "expires_in": 86400000000000
```
technoweenie: Super minor, but the example values in this PR still show the values in nanoseconds :)

docs/api/server-discovery.md
```diff
@@ -14,16 +14,16 @@ servers.
...
-Git Remote: `https://git-server.com/foo/bar.git`  
+Git Remote: `https://git-server.com/foo/bar.git`
 LFS Server: `https://git-server.com/foo/bar.git/info/lfs`
 
-Git Remote: `git@git-server.com:foo/bar.git`  
+Git Remote: `git@git-server.com:foo/bar.git`
 LFS Server: `https://git-server.com/foo/bar.git/info/lfs`
 
-Git Remote: `ssh://git-server.com/foo/bar.git`  
+Git Remote: `ssh://git-server.com/foo/bar.git`
```
technoweenie: Same markdown line ending issue. 

I suppose this could be table too, like:

```
| Git Remote | LFS Server |
|------------|------------|
| `https://git-server.com/foo/bar` | `https://git-server.com/foo/bar.git/info/lfs` |
```

tools/time_tools.go
```diff
@@ -0,0 +1,24 @@
...
+// IsExpiredAtOrIn returns whether or not the result of calling TimeAtOrIn is
+// "expired" within "until" units of time from now.
+func IsExpiredAtOrIn(until time.Duration, at time.Time, in time.Duration) (time.Time, bool) {
+	expiration := TimeAtOrIn(at, in)
+	if expiration.IsZero() {
+		return expiration, false
+	}
+
+	return expiration, expiration.Before(time.Now().Add(until))
+}
```
technoweenie: I don't think this is correct, since it's always comparing with `time.Now()`. I think the structs would have to store `time.Now()` at the time they're deserialized from the json response. Otherwise a positive `expires_in` value would never expire, right?

docs/api/authentication.md
ttaylorr: Ha, nice catch -- I fixed those up in https://github.com/git-lfs/git-lfs/pull/2130/commits/2572163f392890d02ec02654d186b65343366bdd.

docs/api/batch.md
ttaylorr: Ah, I remember thinking that I needed to not stage those changes and then totally forgot. I think we should go with the `<br>`, since I think it's fairly common for editors to trim trailing whitespace off the end of a line.

I went with that in 2a1e63e, but was thinking about adding markdown tables the whole time.

tools/time_tools.go
ttaylorr: Good call, I think I addressed this in https://github.com/git-lfs/git-lfs/pull/2130/commits/0d66374f4f2da583bbbed1da621b778547175e60. I'm not a super huge fan of embedding a `createdAt` field on both of those structs, but I think it should work alright.

tools/time_tools.go
```diff
@@ -0,0 +1,24 @@
+package tools
+
+import "time"
+
+// IsExpiredAtOrIn returns whether or not the result of calling TimeAtOrIn is
+// "expired" within "until" units of time from now.
+func IsExpiredAtOrIn(now time.Time, until time.Duration, at time.Time, in time.Duration) (time.Time, bool) {
```
technoweenie: I'd like to see some golang tests here. Time arithmetic and comparisons always trip me up.

ttaylorr:
Tests written in https://github.com/git-lfs/git-lfs/pull/2130/commits/a146cb8f92a5526718bae5e7dbc6db6acce33847 & https://github.com/git-lfs/git-lfs/pull/2130/commits/d115a1a1e1c1acf5f5aaa46ad3eed1ab4751908f, going to push this out once they're green. Thanks again 🤘 !

Issue #2141, opened 2017-04-07
all: build against Go 1.8.1


ttaylorr:
Go 1.8.1 was [recently released](https://groups.google.com/forum/#!topic/golang-nuts/KO2t50AYo9I). In the spirit of staying on the latest and greatest Golang, I'd like to build LFS v2.1.0 on this release of Go.

I'd like to give this release a few days to get support in AppVeyor, on Hombrew, and to make sure that there are no new regressions.

Thoughts?

---

/cc @git-lfs/core

technoweenie:
Sounds good :+1:

sschuberth:
Is there any real benefit in upgrading to 1.8.1, or is it only for the sake of being a cool kid and using the latest? If there's no real benefit I don't believe it's worth to switch to 1.8.1 if that means we again have to manually install it in AppVeyor instead of being able to use using Chocolatey.

technoweenie:
Bug fixes in 1.8.1: https://github.com/golang/go/issues?q=is%3Aissue+milestone%3AGo1.8.1+is%3Aclosed

I don't see any reason to mess with the chocolatey code, just to revert it once they update Go. Hopefully Chocolatey is on the latest Go by the time LFS v2.1.0 is ready.

sschuberth:
> Bug fixes in 1.8.1: https://github.com/golang/go/issues?q=is%3Aissue+milestone%3AGo1.8.1+is%3Aclosed

I know the list of fixes. What I meant was if someone checked that we actually do benefit of these in this project. But I've now checked myself and there seem to be some pretty generic compiler / runtime fixes.

Anyway, Chocolately already has [Go 1.8.1](https://chocolatey.org/packages/golang), so from that perspective I have no objections anymore.

ttaylorr:
> What I meant was if someone checked that we actually do benefit of these in this project.

These are good questions. My reasoning for opening the proposal and getting on the latest version of Go was to minimize the gap between current and "next" release. Even if these bug fixes don't benefit us directly, they bring us closer to Go 1.8.x (x > 1) which could have a critical fix that we need to be on. For those necessary upgrades, I want the change in Go version to be as small as possible so that there's less chance for unexpected behavior/interaction with the Go compiler.

PR #2145, opened 2017-04-10
all: build on Go 1.8.1

ttaylorr:
This pull request resolves https://github.com/git-lfs/git-lfs/issues/2141 by building Git LFS on Go 1.8.1.

In terms of CI, we have:

1. Travis, which was upgraded in the `.travis.yml`.
2. CircleCI, which receives whatever the latest `golang` bottle is on Homebrew, which was updated in https://github.com/Homebrew/homebrew-core/pull/12201.
3. AppVeyor, which @sschuberth pointed out is 👍 to go via https://github.com/git-lfs/git-lfs/issues/2141#issuecomment-292698702.

I also noticed that our README recommends that you have a Go higher than 1.7.3. I think this is technically correct, since we don't use any of the latest Go APIs, but I think now is as good a time as any to bump that version.

Closes https://github.com/git-lfs/git-lfs/issues/2141.

---

/cc @git-lfs/core https://github.com/git-lfs/git-lfs/issues/2141

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 # http://docs.travis-ci.com/user/languages/go/
 language: go
 
-go: 1.8
+go: 1.8.1
 
 os:
   - linux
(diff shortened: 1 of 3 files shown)


sschuberth:
Well, we still need to update AppVeyor to 1.8.1 by basically reverting d93e1443ad66fb0d690191cf1f285ddec132d476 and changing `1.8.0` in there to `1.8.1`. I wasn't expecting us to change the Go version again that soon, which is why I thought it would be a nice clean-up to rely on AppVeyor's system Go. But as we seem to often need a custom version, the `cinst golang --version` line should stay.

ttaylorr:
> Well, we still need to update AppVeyor to 1.8.1

Thanks, I think I had misread your comment to mean that AppVeyor's latest registered version of Golang was 1.8.1.

PR #2152, opened 2017-04-12
config: support multi-valued config entries

ttaylorr:
This pull request expands the the `config.Environment` and `config.Fetcher` types to support multiple values being assigned to a single config key. As per the `.gitconfig` documentation, multiple config values are allowed<sup>[[1](https://git-scm.com/docs/git-config#_description)]</sup>:

> Multiple lines can be added to an option by using the `--add` option.

And as evidenced by:

```bash
~/g/git-lfs (fetcher-get-all) $ git --version
git version 2.12.0
~/g/git-lfs (fetcher-get-all) $ git \
  -c "foo=bar" \
  -c "foo=baz" \
  config -l | tail -n 2
foo=bar
foo=baz
```

This is needed to support multiple values for the `http.<url>.extraHeaders` config key. The change in the `*config.URLFetcher` will come in a subsequent PR, since this one is already on the large side.

This change is pretty gnarly, as I couldn't do it across multiple commits without breaking the compiler. Here's a quick overview:

1. The `config.Fetcher` type got a few behavior changes:

```diff
diff --git a/config/fetcher.go b/config/fetcher.go
index 24e0dfad..2d3bf660 100644
--- a/config/fetcher.go
+++ b/config/fetcher.go
@@ -6,8 +6,17 @@ package config
 type Fetcher interface {
 	// Get returns the string value associated with a given key and a bool
 	// determining if the key exists.
+	//
+	// If multiple entries match the given key, the first one will be
+	// returned.
 	Get(key string) (val string, ok bool)
 
-	// All returns a copy of all the key/value pairs for the current environment.
-	All() map[string]string
+	// GetAll returns the a set of string values associated with a given
+	// key. If no entries matched the given key, an empty slice will be
+	// returned instead.
+	GetAll(key string) (vals []string)
+
+	// All returns a copy of all the key/value pairs for the current
+	// environment.
+	All() map[string][]string
 }
```

which break down as:

- The `All()` function now returns a "multi-map" `map[string][]string`, instead of a `map[string]string`.
- The `Get()` function retains its original behavior by returning the first matching config entry.
- A new function, `GetAll()` was added to return a set of values that match the given key, if the caller wants multiple values.

I figured that this would be the least invasive way that we could make the change. `All()` _has_ to change, but retaining the original behavior of `Get()` seems to offer more flexibility, and is less annoying than `Get(key)[0]` everywhere.

2. The `GitFetcher` type was updated to support multiple keys, and the `OsFetcher` type was updated to return a 1-ary set of `[]string{os.Getenv(k)}`.
3. All `Env` "copies" changed to support the above interface changes.
4. All `MapEnvironment`-like helpers got a `UniqMapEnvironment` counter-part, which translates up from a `map[string]string` to a `map[string][]string`, where `n[k]=[]string{m[k]}`, like:

```go
type UniqMapFetcher map[string]string

func (u UniqMapFetcher) All() map[string][]string {
        m := make(map[string][]string, len(u))
        for k, v := range u {
                m[k] = []string{v}
        }

        return m
}
```

and so on. This made the unit test changes much more ergonomic for scenarios where a 1->1 config mapping was being tested.

5. Integration tests were updated where new `.git/config` values were being `--add`'ed to old ones.

---

/cc @git-lfs/core

### DIFF
--- a/config/git_fetcher.go
+++ b/config/git_fetcher.go
@@ -13,7 +13,7 @@ import (
 
 type GitFetcher struct {
 	vmu  sync.RWMutex
-	vals map[string]string
+	vals map[string][]string
 }
 
 type GitConfig struct {
@@ -29,7 +29,7 @@ func NewGitConfig(gitconfiglines string, onlysafe bool) *GitConfig {
 }
 
 func ReadGitConfig(configs ...*GitConfig) (gf *GitFetcher, extensions map[string]Extension, uniqRemotes map[string]bool) {
-	vals := make(map[string]string)
+	vals := make(map[string][]string)
 
 	extensions = make(map[string]Extension)
 	uniqRemotes = make(map[string]bool)
@@ -47,7 +47,7 @@ func ReadGitConfig(configs ...*GitConfig) (gf *GitFetcher, extensions map[string
 			key, val := strings.ToLower(pieces[0]), pieces[1]
 
 			if origKey, ok := uniqKeys[key]; ok {
-				if ShowConfigWarnings && vals[key] != val && strings.HasPrefix(key, gitConfigWarningPrefix) {
+				if ShowConfigWarnings && len(vals[key]) > 0 && vals[key][0] != val && strings.HasPrefix(key, gitConfigWarningPrefix) {
 					fmt.Fprintf(os.Stderr, "WARNING: These git config values clash:\n")
 					fmt.Fprintf(os.Stderr, "  git config %q = %q\n", origKey, vals[key])
 					fmt.Fprintf(os.Stderr, "  git config %q = %q\n", pieces[0], val)
@@ -101,7 +101,7 @@ func ReadGitConfig(configs ...*GitConfig) (gf *GitFetcher, extensions map[string
 				continue
 			}
 
-			vals[key] = val
+			vals[key] = append(vals[key], val)
 		}
 	}
 
@@ -119,21 +119,31 @@ func ReadGitConfig(configs ...*GitConfig) (gf *GitFetcher, extensions map[string
 //
 // Get is safe to call across multiple goroutines.
 func (g *GitFetcher) Get(key string) (val string, ok bool) {
-	g.vmu.RLock()
-	defer g.vmu.RUnlock()
+	all := g.GetAll(key)
 
-	val, ok = g.vals[strings.ToLower(key)]
-	return
+	if len(all) == 0 {
+		return "", false
+	}
+	return all[0], true
 }
 
-func (g *GitFetcher) All() map[string]string {
-	newmap := make(map[string]string)
+func (g *GitFetcher) GetAll(key string) []string {
+	g.vmu.RLock()
+	defer g.vmu.RUnlock()
+
+	return g.vals[strings.ToLower(key)]
+}
+
+func (g *GitFetcher) All() map[string][]string {
+	newmap := make(map[string][]string)
 
 	g.vmu.RLock()
 	defer g.vmu.RUnlock()
 
-	for key, value := range g.vals {
-		newmap[key] = value
+	for key, values := range g.vals {
+		for _, value := range values {
+			newmap[key] = append(newmap[key], value)
+		}
 	}
 
 	return newmap
--- a/config/git_fetcher.go
+++ b/config/git_fetcher.go
@@ -47,7 +47,7 @@ func ReadGitConfig(configs ...*GitConfig) (gf *GitFetcher, extensions map[string
 			key, val := strings.ToLower(pieces[0]), pieces[1]
 
 			if origKey, ok := uniqKeys[key]; ok {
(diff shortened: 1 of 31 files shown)


config/git_fetcher.go
```diff
@@ -47,7 +47,7 @@ func ReadGitConfig(configs ...*GitConfig) (gf *GitFetcher, extensions map[string
 			key, val := strings.ToLower(pieces[0]), pieces[1]
 
 			if origKey, ok := uniqKeys[key]; ok {
-				if ShowConfigWarnings && vals[key] != val && strings.HasPrefix(key, gitConfigWarningPrefix) {
+				if ShowConfigWarnings && len(vals[key]) > 0 && vals[key][0] != val && strings.HasPrefix(key, gitConfigWarningPrefix) {
```
technoweenie: This should check the last value, not the first.

config/git_fetcher.go
```diff
@@ -119,21 +119,31 @@ func ReadGitConfig(configs ...*GitConfig) (gf *GitFetcher, extensions map[string
 //
 // Get is safe to call across multiple goroutines.
 func (g *GitFetcher) Get(key string) (val string, ok bool) {
+	all := g.GetAll(key)
+
+	if len(all) == 0 {
+		return "", false
+	}
+	return all[0], true
```
technoweenie: This should check the last value, not the first.

config/map_fetcher.go
```diff
@@ -2,22 +2,42 @@ package config
...
 // Get implements the func `Fetcher.Get`.
 func (m mapFetcher) Get(key string) (val string, ok bool) {
-	val, ok = m[key]
-	return
+	all := m.GetAll(key)
+
+	if len(all) == 0 {
+		return "", false
+	}
+	return all[0], true
```
technoweenie: This should check the last value, not the first.

lfsapi/endpoint_finder.go
```diff
@@ -279,12 +279,12 @@ func initAliases(e *endpointGitFinder, git Env) {
 	prefix := "url."
 	suffix := ".insteadof"
 	for gitkey, gitval := range git.All() {
-		if !(strings.HasPrefix(gitkey, prefix) && strings.HasSuffix(gitkey, suffix)) {
+		if len(gitval) == 0 || !(strings.HasPrefix(gitkey, prefix) && strings.HasSuffix(gitkey, suffix)) {
 			continue
 		}
-		if _, ok := e.aliases[gitval]; ok {
+		if _, ok := e.aliases[gitval[0]]; ok {
```
technoweenie: This should check the last value, not the first.

lfsapi/endpoint_finder.go
```diff
@@ -279,12 +279,12 @@ func initAliases(e *endpointGitFinder, git Env) {
...
 			fmt.Fprintf(os.Stderr, "WARNING: Multiple 'url.*.insteadof' keys with the same alias: %q\n", gitval)
 		}
-		e.aliases[gitval] = gitkey[len(prefix) : len(gitkey)-len(suffix)]
+		e.aliases[gitval[0]] = gitkey[len(prefix) : len(gitkey)-len(suffix)]
```
technoweenie: This should check the last value, not the first.

lfsapi/lfsapi.go
```diff
@@ -147,18 +147,79 @@ func DecodeJSON(res *http.Response, obj interface{}) error {
...
 
 func (e TestEnv) Get(key string) (string, bool) {
-	v, ok := e[key]
-	return v, ok
+	all := e.GetAll(key)
+
+	if len(all) == 0 {
+		return "", false
+	}
+	return all[0], true
```
technoweenie: This should check the last value, not the first.

config/os_fetcher.go
```diff
@@ -54,6 +54,15 @@ func (o *OsFetcher) Get(key string) (val string, ok bool) {
...
 }
 
-func (o *OsFetcher) All() map[string]string {
+// GetAll implements the `config.Fetcher.GetAll` method by returning, at most, a
+// 1-ary set containing the result of `config.OsFetcher.Get()`.
+func (o *OsFetcher) GetAll(key string) []string {
+	if v, ok := o.Get(key); ok {
+		return []string{v}
+	}
+	return make([]string, 0)
```
ttaylorr: I left this as a non-nil slice, assuming the caller may try and `append` to it we could avoid a panic by having a non-nil slice.

ttaylorr:
@technoweenie thanks for the review! Good call on returning the last gitconfig element, not the first. I made one larger commit to grab everything and fixed it all up in https://github.com/git-lfs/git-lfs/pull/2152/commits/34d590c52f00e6953d5958ae96cfe8668a6b84b2.

config/os_fetcher.go
technoweenie: `append` should work fine with a nil slice: https://play.golang.org/p/d1iLXa20J2

PR #2154, opened 2017-04-13
config: support multi-value lookup on URLConfig

ttaylorr:
This pull request teaches the `*config.URLConfig` type to fetch multiple values matching a given `prefix.<url>.key`.

This is supporting work for the http.extraHeaders support as defined in #1500.

Right now, it's based off the work done in https://github.com/git-lfs/git-lfs/pull/2152. Once that branch gets merged, this diff should become more readable. In the meantime, **here's an inter-diff**: https://github.com/git-lfs/git-lfs/compare/fetcher-get-all...url-config-many-headers.

To implement this, I changed the `get` function to `getAll`, which replaces the internal `c.git.Get()` with `c.git.GetAll()` calls, which will return _all_ matching values for a given config key, if any were found. What this means for behavior, is that the public `config.URLConfig.GetAll()` function will return more than one value _only_ if it's in the same keyspace.

For instance, given a git config:

```
http.http://example.com/foo.bar=baz
http.http://example.com/.bar=baz2
```

Only a value of `[]string{"baz"}` will be returned. I _think_ this is the right behavior, but we could turn those `return`s into `append`s and roll up the values of `baz` and `baz2`. Either would work.

1. https://github.com/git-lfs/git-lfs/commit/fdbf3041f965275419cee6a93063f121d81a3596, https://github.com/git-lfs/git-lfs/commit/0b95e7c8b52ceca93661babf6995aaf0455987b5 & https://github.com/git-lfs/git-lfs/commit/473ccb7a17634bfd3225219d4250c005d3852643: some light refactoring
2. https://github.com/git-lfs/git-lfs/commit/5936cf4f29a07272ede29ed8578097e56582b294: rename `get()` to `getAll()` and use that for multi-valued lookups.

---

/cc @git-lfs/core

### DIFF
--- a/config/url_config.go
+++ b/config/url_config.go
@@ -35,11 +35,7 @@ func (c *URLConfig) get(key, rawurl string) (string, bool) {
 		return "", false
 	}
 
-	hosts := make([]string, 0)
-	if u.User != nil {
-		hosts = append(hosts, fmt.Sprintf("%s://%s@%s", u.Scheme, u.User.Username(), u.Host))
-	}
-	hosts = append(hosts, fmt.Sprintf("%s://%s", u.Scheme, u.Host))
+	hosts := c.hosts(u)
 
 	pLen := len(u.Path)
 	if pLen > 2 {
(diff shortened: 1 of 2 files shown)


config/url_config.go
```diff
@@ -23,53 +23,76 @@ func NewURLConfig(git Environment) *URLConfig {
 func (c *URLConfig) Get(prefix, key string, rawurl string) (string, bool) {
 	key = strings.ToLower(key)
 	prefix = strings.ToLower(prefix)
-	if v, ok := c.get(key, rawurl); ok {
-		return v, ok
+	if v := c.getAll(key, rawurl); len(v) > 0 {
+		return v[0], true
```
technoweenie: This should return the last value.

ttaylorr:
@technoweenie thanks for the review! I didn't see any other spots beyond the one that you sent me, which I fixed up in https://github.com/git-lfs/git-lfs/pull/2154/commits/a8098577e12a5bd14959e757c379d54ee41628fb. I think this is ready to merge now 👍 .

PR #2156, opened 2017-04-14
config: remove dependency on lfsapi package

ttaylorr:
This pull request removes a dependency drawn from the `github.com/git-lfs/git-lfs/config` package to the `github.com/git-lfs/git-lfs/lfsapi` package, where several of the `Endpoint()`-related functions returned `lfsapi.<T>` types.

This work supports the proposal in #1500, where I need to import the `config` package from `lfsapi` in order to maintain an instance of the `*config.URLConfig` type.

---

/cc @git-lfs/core

### DIFF
--- a/lfsapi/client.go
+++ b/lfsapi/client.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/git-lfs/git-lfs/config"
 	"github.com/git-lfs/git-lfs/errors"
 	"github.com/rubyist/tracerx"
 )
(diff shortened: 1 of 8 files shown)


lfsapi/client.go
```diff
@@ -222,3 +223,7 @@ func newRequestForRetry(req *http.Request, location string) (*http.Request, erro
 	newReq.ContentLength = req.ContentLength
 	return newReq, nil
 }
+
+func init() {
+	UserAgent = config.VersionDesc
```
technoweenie: This may be better off in the commands package. That way config and lfsapi are completely decoupled.

PR #2158, opened 2017-04-15
config: add prefix to URLConfig type

ttaylorr:
This pull request fixes an oversight I made in https://github.com/git-lfs/git-lfs/pull/1912 where the prefix of a `*config.URLConfig` lookup wasn't passed correctly into the `getAll()` function.

This pull request is based off of #2156, which should be merged into master before this branch. __In the meantime, here's an inter-diff__: https://github.com/git-lfs/git-lfs/compare/lfsapi-config-cycle...url-config-prefix.

I snuck a second change in here, to reorder the arguments from:

```diff
-func (c *URLConfig) Get(prefix, key string, rawurl string) (string, bool) {
+func (c *URLConfig) Get(prefix, rawurl, key string) (string, bool) {
```

Which more closely matches the `{prefix}.{url}.{key}` ordering that is actually present in the config entry.

---

/cc @git-lfs/core

### DIFF
--- a/config/url_config.go
+++ b/config/url_config.go
@@ -11,6 +11,10 @@ type URLConfig struct {
 }
 
 func NewURLConfig(git Environment) *URLConfig {
+	if git == nil {
+		git = EnvironmentOf(make(mapFetcher))
+	}
+
 	return &URLConfig{
 		git: git,
 	}
--- a/config/url_config.go
+++ b/config/url_config.go
@@ -20,7 +20,7 @@ func NewURLConfig(git Environment) *URLConfig {
 // rules in https://git-scm.com/docs/git-config#git-config-httplturlgt.
 // The value for `http.{key}` is returned as a fallback if no config keys are
 // set for the given urls.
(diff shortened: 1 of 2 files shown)


lfsapi/client.go
```diff
@@ -222,3 +223,7 @@ func newRequestForRetry(req *http.Request, location string) (*http.Request, erro
 	newReq.ContentLength = req.ContentLength
 	return newReq, nil
 }
+
+func init() {
+	UserAgent = config.VersionDesc
```
technoweenie: This may be better off in the `commands` package. That way `config` and `lfsapi` are completely decoupled.

config/config.go
```diff
@@ -54,12 +52,9 @@ type Configuration struct {
 
 	CurrentRemote string
 
-	loading        sync.Mutex // guards initialization of gitConfig and remotes
-	remotes        []string
-	extensions     map[string]Extension
-	manualEndpoint *lfsapi.Endpoint
-	endpointFinder lfsapi.EndpointFinder
-	endpointMu     sync.Mutex
```
technoweenie: 😍

PR #2159, opened 2017-04-15
lfsapi: support for http.<url>.extraHeader

ttaylorr:
This pull request resolves the proposal made in #1500 by implementing support for `http.<url>.extraHeader`.

This pull request is based off of https://github.com/git-lfs/git-lfs/pull/2158, and that should be merged before this one. In the meantime, __here's an inter-diff__: https://github.com/git-lfs/git-lfs/compare/url-config-prefix...lfsapi-extra-headers.

When making any HTTP(S) request from the lfsapi package, LFS will first look for matching `http.{url}.extraHeader` config entries, where `url` is the request URL. LFS currently adds one extra header to each request it makes, which is the `User-Agent: ` header. The `lfsapi` package looks for `extraHeaders` _before_ adding the User-Agent header, so it cannot be overridden.

As per https://github.com/git-lfs/git-lfs/pull/2152 & https://github.com/git-lfs/git-lfs/pull/2154, multiple extraHeaders can be added to a particular request. The following configuration is valid:

```bash
$ git config http.https://github.com/user/repo.git/info/lfs.extraHeader "X-Foo: bar"
$ git config http.https://github.com/user/repo.git/info/lfs.extraHeader "X-Non-Unique: bar2"
$ git config http.https://github.com/user/repo.git/info/lfs.extraHeader "X-Non-Unique: bar3"
```

would send a request with the headers equal to:

```curl
> GET /user/repo.git/info/lfs HTTP/1.1
> Host: github.com
> X-Foo: bar
> X-Non-Unique: bar2
> X-Non-Unique: bar3
```

Closes: #1500.

---

/cc @git-lfs/core #1500

### DIFF
--- a/lfsapi/client.go
+++ b/lfsapi/client.go
@@ -76,7 +76,25 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
 	return res, c.handleResponse(res)
 }
 
+func (c *Client) setExtraHeaders(req *http.Request) http.Header {
+	var copy http.Header = make(http.Header)
+	for k, vs := range req.Header {
+		copy[k] = vs
+	}
+
+	for k, vs := range c.extraHeaders(req.URL) {
+		for _, v := range vs {
+			copy[k] = append(copy[k], v)
+		}
+	}
+	return copy
+}
+
 func (c *Client) extraHeaders(u *url.URL) map[string][]string {
+	if c.uc == nil {
+		c.uc = config.NewURLConfig(c.GitEnv())
+	}
+
 	hdrs := c.uc.GetAll("http", u.String(), "extraHeader")
 	m := make(map[string][]string, len(hdrs))
 
--- a/lfsapi/client.go
+++ b/lfsapi/client.go
@@ -66,7 +66,7 @@ func joinURL(prefix, suffix string) string {
 }
 
 func (c *Client) Do(req *http.Request) (*http.Response, error) {
-	req.Header = c.setExtraHeaders(req)
+	req.Header = c.extraHeadersFor(req)
 	req.Header.Set("User-Agent", UserAgent)
 
 	res, err := c.doWithRedirects(c.httpClient(req.Host), req, nil)
@@ -77,7 +77,7 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
 	return res, c.handleResponse(res)
 }
 
-func (c *Client) setExtraHeaders(req *http.Request) http.Header {
+func (c *Client) extraHeadersFor(req *http.Request) http.Header {
 	copy := make(http.Header, len(req.Header))
 	for k, vs := range req.Header {
 		copy[k] = vs
--- a/lfsapi/client.go
+++ b/lfsapi/client.go
@@ -66,6 +66,7 @@ func joinURL(prefix, suffix string) string {
 }
 
 func (c *Client) Do(req *http.Request) (*http.Response, error) {
+	req.Header = c.setExtraHeaders(req)
 	req.Header.Set("User-Agent", UserAgent)
 
 	res, err := c.doWithRedirects(c.httpClient(req.Host), req, nil)
--- a/lfsapi/client.go
+++ b/lfsapi/client.go
@@ -100,12 +100,12 @@ func (c *Client) extraHeaders(u *url.URL) map[string][]string {
 	m := make(map[string][]string, len(hdrs))
 
 	for _, hdr := range hdrs {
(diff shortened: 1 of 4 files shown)


lfsapi/client.go
```diff
@@ -75,6 +77,41 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
 	return res, c.handleResponse(res)
 }
 
+func (c *Client) setExtraHeaders(req *http.Request) http.Header {
+	var copy http.Header = make(http.Header)
+	for k, vs := range req.Header {
+		copy[k] = vs
+	}
```
technoweenie: Why bother making a copy? Can't you just modify the req headers in place?

Also, why are you using the longer `var` syntax to init the var, instead of `copy := make(http.Header)`?

lfsapi/client.go
```diff
@@ -75,6 +77,41 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
...
+		for _, v := range vs {
+			copy[k] = append(copy[k], v)
+		}
+	}
+	return copy
+}
+
+func (c *Client) extraHeaders(u *url.URL) map[string][]string {
+	if c.uc == nil {
+		c.uc = config.NewURLConfig(c.GitEnv())
```
technoweenie: I'd probably move this to `lfsapi.NewClient`, or at least add a mutex.

lfsapi/client.go
```diff
@@ -75,6 +77,41 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
...
+	}
+
+	hdrs := c.uc.GetAll("http", u.String(), "extraHeader")
+	m := make(map[string][]string, len(hdrs))
+
+	for _, hdr := range hdrs {
+		parts := strings.SplitN(hdr, ": ", 2)
```
technoweenie: Relying on just a single space after the `:` seems kind of fragile. What do you think about splitting on `:`, and then trimming any whitespace in the value?

lfsapi/client.go
ttaylorr: > Why bother making a copy? Can't you just modify the req headers in place?

Good question. I think it would use less memory to modify the headers in place (though not much). My thought with making the copy was to localize mutating the request to outside of this function so that this could be more easily tested/changed in the future. I'm not tied at all to this implementation, so if you feel strongly, I'm happy to change it.

I pushed https://github.com/git-lfs/git-lfs/pull/2159/commits/d4aec737b2e3a5c8f33517ec9f3a04137ab61a5e to clarify the name of this method, and https://github.com/git-lfs/git-lfs/pull/2159/commits/961456e88f831608b3c43342ea536d7ebd5e1e68 to remove the extraneous `var` decl. I prefer those for initializing values that are going to change immediately, but I don't mind either way.

lfsapi/client.go
ttaylorr: Good call. I had this here originally to prevent a case during testing where the `*lfsapi.Client` was initialized as `&lfsapi.Client{}` (or `new(lfsapi.Client)`) and `uc` was un-initialized. I moved this initialization to the constructor (since the value returned by `c.GitEnv()` is private and doesn't mutate), and taught `*config.URLConfig` to respond to method calls against the `nil` receiver.

x-ref: https://github.com/git-lfs/git-lfs/pull/2159/commits/fcb9fb2ba8747ee91764fa23325c089f78826c46, https://github.com/git-lfs/git-lfs/pull/2159/commits/9aa4be7d292b301b2be678347683dc4bf5535735 & https://github.com/git-lfs/git-lfs/pull/2159/commits/20cbfbabedf7876b3f5c9798c07d80e19ee1f8fa, respectively.

lfsapi/client.go
ttaylorr: Great call. Made this change in https://github.com/git-lfs/git-lfs/pull/2159/commits/442dc02f7d71378af84ecdcefa8012e932c9cf22.

ttaylorr:
@technoweenie thanks for the review! I addressed your comments, and I think this should be ready to merge.

lfsapi/client.go
```diff
@@ -76,6 +77,37 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
 	return res, c.handleResponse(res)
 }
 
+func (c *Client) extraHeadersFor(req *http.Request) http.Header {
+	copy := make(http.Header, len(req.Header))
```
technoweenie: You could even get the values from `c.extraHeaders()` to get a more accurate initial size :)

Issue #2093, opened 2017-03-28
Set `locksverify` based on the host


rjbell4:
It appears that enabling/disabling `locksverify` is done on a per URL basis.  We have a lot of repositories in our organization.  Managing a full set of `locksverify=false` settings (in order to prevent warning messages and lots of questions) is going to be a bear.  It would be *highly* preferable instead to disable it for *ALL* repos on a server, in one fell swoop.

ttaylorr:
@rjbell4 thanks for opening this, I think this would be useful to have in the LFS client for situations like yours.

If you have access to the server, you could respond with an HTTP 501 to the `locks/verify` endpoint and the client will [know to ignore that][1].

[1]: https://github.com/git-lfs/git-lfs/blob/v2.0.1/commands/uploader.go#L125-L126

rjbell4:
I am using Artifactory as a server, so I am at their mercy for the server implementation.

ttaylorr:
@rjbell4 I assume that other's are in the same boat as well. Going to milestone this into the 2.1.0 release.

technoweenie:
I think the approach in https://github.com/git-lfs/git-lfs/pull/1912 could be adapted to work with `http.<url>.*` and `lfs.<url>.*` keys.

ttaylorr:
I think getting #1912 merged and then updating the changes in #1951 to work with that would be 👍.

ttaylorr:
👋 @rjbell4, I just pushed up https://github.com/git-lfs/git-lfs/pull/2162 which adds support for this as you originally described. Pending a merge of that pull request into master, this will ship with the next release of Git LFS, v2.1.0.

PR #2162, opened 2017-04-17
commands: support URL-style lookups for `lfs.{url}.locksverify`

ttaylorr:
This pull request implements support for URL-style lookups for the `lfs.{url}.locksverify` config value and resolves https://github.com/git-lfs/git-lfs/issues/2093.

When writing this, I was wondering if we should change the suggested `git config` commands to use a site's host, instead of the fully qualified path. @technoweenie wisely points out that we can't accurately determine whether or not the hostname is the "root" of the LFS API, so the messages were left unchanged.

To partially mitigate this, I added a `hostsWithKnownLockingSupport`, which skips printing out the suggested `git config` message if it's known that a host supports the LFS locking API. I skipped testing this in the integration suite since there's not an accurate way to do this without making a network call to one of the sites in the list.

At the time of writing, GitHub is the only host I know of that completely supports the File Locking API. As additional hosts add support, they should feel free to file a PR to add themselves to that list.

Closes: https://github.com/git-lfs/git-lfs/issues/2093.

---

/cc @git-lfs/core https://github.com/git-lfs/git-lfs/issues/2093

### DIFF
--- a/commands/uploader.go
+++ b/commands/uploader.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"strconv"
 	"strings"
@@ -135,12 +136,41 @@ func verifyLocks(remote string) (ours, theirs []locking.Lock) {
 		} else {
 			ExitWithError(err)
 		}
-	} else if state == verifyStateUnknown {
+	} else if state == verifyStateUnknown && !supportsLockingAPI(endpoint) {
 		Print("Locking support detected on remote %q. Consider enabling it with:", remote)
 		Print("  $ git config 'lfs.%s.locksverify' true", endpoint.Url)
 	}
 
 	return ours, theirs
+}
+
+var (
+	// hostsWithKnownLockingSupport is a list of scheme-less hostnames
+	// (without port numbers) that are known to implement the LFS locking
+	// API.
+	//
+	// Additions are welcome.
+	hostsWithKnownLockingSupport = []string{
+		"github.com",
+	}
+)
+
+// supportsLockingAPI returns whether or not a given lfsapi.Endpoint "e"
+// is known to support the LFS locking API by whether or not its hostname is
+// included in the list above.
+func supportsLockingAPI(e lfsapi.Endpoint) bool {
+	u, err := url.Parse(e.Url)
+	if err != nil {
+		tracerx.Printf("commands: unable to parse %q to determine locking support: %v", e.Url, err)
+		return false
+	}
+
+	for _, host := range hostsWithKnownLockingSupport {
+		if u.Hostname() == host {
+			return true
+		}
+	}
+	return false
 }
 
 func (c *uploadContext) scannerError() error {
--- a/commands/uploader.go
+++ b/commands/uploader.go
@@ -136,7 +136,7 @@ func verifyLocks(remote string) (ours, theirs []locking.Lock) {
 		} else {
 			ExitWithError(err)
 		}
-	} else if state == verifyStateUnknown && !supportsLockingAPI(endpoint) {
+	} else if state == verifyStateUnknown {
 		Print("Locking support detected on remote %q. Consider enabling it with:", remote)
 		Print("  $ git config 'lfs.%s.locksverify' true", endpoint.Url)
 	}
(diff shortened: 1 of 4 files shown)


commands/uploader.go
```diff
@@ -333,9 +364,9 @@ func (c *uploadContext) Await() {
 // given "endpoint". If no state has been explicitly set, an "unknown" state
 // will be returned instead.
 func getVerifyStateFor(endpoint lfsapi.Endpoint) verifyState {
-	key := strings.Join([]string{"lfs", endpoint.Url, "locksverify"}, ".")
+	uc := config.NewURLConfig(cfg.Git)
 
-	v, ok := cfg.Git.Get(key)
+	v, ok := uc.Get("lfs", endpoint.Url, "locksverify")
```
technoweenie: I think this is where `hostsWithKnownLockingSupport` should be used. The way it is now, pushing to github won't warn about the lockVerify step. But it also won't halt a push if our locking services are down. Since our locking service is open to all repositories, it should assume that `locksVerify` is set to `"1"` instead of `""`.

test/test-pre-push.sh
```diff
@@ -762,6 +762,7 @@ begin_test "pre-push locks verify 200"
   grep "git config 'lfs.$endpoint.locksverify' true" push.log
   assert_server_object "$reponame" "$contents_oid"
 )
+end_test
```
technoweenie: Oof. Nice catch :/

commands/uploader.go
ttaylorr: Good call, I think I got this in https://github.com/git-lfs/git-lfs/pull/2162/commits/aaa09af5cc859085dc3fe3ee8f78a5078b968864.

commands/uploader.go
```diff
@@ -377,6 +362,24 @@ func getVerifyStateFor(endpoint lfsapi.Endpoint) verifyState {
...
+// is known to support the LFS locking API by whether or not its hostname is
+// included in the list above.
+func supportsLockingAPI(e lfsapi.Endpoint) bool {
+	u, err := url.Parse(e.Url)
+	if err != nil {
+		tracerx.Printf("commands: unable to parse %q to determine locking support: %v", e.Url, err)
+		return false
+	}
+
+	for _, host := range hostsWithKnownLockingSupport {
```
technoweenie: I think this step should be scipped if the incoming `e.Url` does not have an `https` scheme or does not have port 443. Any exception we add to `hostsWithKnownLockingSupport` will likely be some public site with ssl everywhere, and standard ports. 

This also doesn't stop end users from configuring their own internal lfs servers with locksVerify enabled.

ttaylorr: Good call. I think as we add/get support among more hosts, we should have the flexibility to specify by scheme, path prefix, and etc. This may be premature, but I changed the `hostsWithKnownLockingSupport` to be a `[]*url.URL`, where we can hold onto and compare against fields like `Host`, `Scheme`, Port()`, and etc.

That change is in https://github.com/git-lfs/git-lfs/pull/2162/commits/739bd779b35aa1a13b6572219c4ed3fc6dc9738e. I wrote a test case on the Playground [here](https://play.golang.org/p/o6biD__ny6), which I turned into a unit test in https://github.com/git-lfs/git-lfs/pull/2162/commits/a6c7b965c76147f91718aa0448359507b2142301.

commands/uploader.go
```diff
@@ -332,13 +332,16 @@ func (c *uploadContext) Await() {
 }
 
 var (
+	githubHttps, _ = url.Parse("https://github.com")
+	githubSsh, _   = url.Parse("ssh://github.com")
```
technoweenie: If you add a `/` at the end of these, `strings.HasPrefix(u.Path, supported.Path)` will protect against urls like `https://github.com:123/...`.

Also, does LFS add the `ssh://` prefix for git remotes like `git@github.com:...`?

ttaylorr: > If you add a `/` at the end of these, `strings.HasPrefix(u.Path, supported.Path)` will protect against urls like `https://github.com:123/...`.

Not totally sure on this one, since I think the `:123` port number would be a part of the `u.Host` and `u.Port()` member/function, but wouldn't appear in the `u.Path`.

> Also, does LFS add the `ssh://` prefix for git remotes like `git@github.com:...`?

Good question, the `*lfsapi.endpointGitFinder` looks at empty schemes in [`lfsapi/endpoint_finder.go:178-9`](https://github.com/git-lfs/git-lfs/blob/v2.0.2/lfsapi/endpoint_finder.go#L178-L179) and automatically adds the `ssh://` scheme. Some related work also in #2090.

technoweenie: > Not totally sure on this one, since I think the :123 port number would be a part of the u.Host and u.Port() member/function, but wouldn't appear in the u.Path.

Correct you are! Ignoooore meeeee.

PR #2168, opened 2017-04-21
Fail push if locks verify is enabled, and any locked files are updated

### LINKED ISSUE #2166: LFS does not halt push when lock is held on object
If someone holds a lock on a file, LFS currently only provides a text warning that the path is locked. The git push succeeds and thus the remote gets in a bad state (it has the LFS pointer object but the LFS object was not pushed). It requires using a `git lfs push [remote] --all` or `git lfs push [remote] --object-id [id]`.

[Here is the line that only warns](https://github.com/git-lfs/git-lfs/blame/13719e4c9387a3c7a07b921dac76f8374f33c064/commands/uploader.go#L330)

It was introduced by #1885 
Due to issue #1884 
And issue #1878 
And adding the failed push apparently was waiting on #1951 and #2093 

Based on the above history, LFS should be ready to fail when pushing a path that is locked by someone else.

In addition: It would probably be good to start warning when pushing an object without owning the lock.

technoweenie:
Fixes #2166, and reverses #1885 (which was needed because users were inadvertently running on a pre-release version of LFS, which blocked their pushes because no File Locking API implementations existed in the wild yet.

Question: should `locksverify` [automatically be enabled](https://github.com/git-lfs/git-lfs/blob/ab50944e2632844770c4cbf4ef8b79d5ffcfaa81/commands/uploader.go#L141-L145) if it's currently `NULL`, and the locks verify api call returned a valid result? I'm leaning towards "yes." I think our approach in LFS v2.0 is a bit too cautious. I think #2162 is a good step towards improving this situation for public LFS servers that implement the file locking API. But I think auto-enabling after successful results will remove an explicit opt-in step when using internal or custom LFS servers.

### DIFF
--- a/commands/uploader.go
+++ b/commands/uploader.go
@@ -125,15 +125,15 @@ func verifyLocks(remote string) (ours, theirs []locking.Lock, st verifyState) {
 		if errors.IsNotImplementedError(err) {
 			disableFor(endpoint)
 		} else if state == verifyStateUnknown || state == verifyStateEnabled {
-			if !errors.IsAuthError(err) {
+			if errors.IsAuthError(err) {
+				ExitWithError(err)
+			} else {
 				Print("Remote %q does not support the LFS locking API. Consider disabling it with:", remote)
 				Print("  $ git config 'lfs.%s.locksverify' false", endpoint.Url)
 
 				if state == verifyStateEnabled {
 					ExitWithError(err)
 				}
-			} else {
-				ExitWithError(err)
 			}
 		}
 	} else if state == verifyStateUnknown {
--- a/commands/uploader.go
+++ b/commands/uploader.go
@@ -125,11 +125,12 @@ func verifyLocks(remote string) (ours, theirs []locking.Lock, st verifyState) {
 		if errors.IsNotImplementedError(err) {
 			disableFor(endpoint)
 		} else if state == verifyStateUnknown || state == verifyStateEnabled {
-			if state == verifyStateEnabled {
-				ExitWithError(err)
-			} else if !errors.IsAuthError(err) {
+			if !errors.IsAuthError(err) {
 				Print("Remote %q does not support the LFS locking API. Consider disabling it with:", remote)
 				Print("  $ git config 'lfs.%s.locksverify' false", endpoint.Url)
+				if state == verifyStateEnabled {
+					ExitWithError(err)
+				}
 			}
 		}
 	} else if state == verifyStateUnknown {
--- a/commands/uploader.go
+++ b/commands/uploader.go
@@ -116,11 +116,7 @@ func newUploadContext(remote string, dryRun bool) *uploadContext {
 
 func verifyLocks(remote string) (ours, theirs []locking.Lock, st verifyState) {
 	endpoint := getAPIClient().Endpoints.Endpoint("upload", remote)
-
 	state := getVerifyStateFor(endpoint)
-	if state == verifyStateDisabled {
-		return
-	}
 
 	lockClient := newLockClient(remote)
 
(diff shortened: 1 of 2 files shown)


commands/uploader.go
```diff
@@ -141,7 +143,7 @@ func verifyLocks(remote string) (ours, theirs []locking.Lock) {
 		Print("  $ git config 'lfs.%s.locksverify' true", endpoint.Url)
 	}
 
-	return ours, theirs
+	return ours, theirs, state
```
ttaylorr: 👍. I'm not in love with the fact that `verifyLocks` returns a state, rather than taking one, but I think the refactoring to invert this isn't worth it.

commands/uploader.go
```diff
@@ -327,7 +329,11 @@ func (c *uploadContext) Await() {
 	c.trackedLocksMu.Unlock()
 
 	if avoidPush {
-		Error("WARNING: The above files would have halted this push.")
+		if c.lockVerifyState == verifyStateEnabled {
+			Exit("ERROR: Cannot update locked files.")
+		} else {
+			Error("WARNING: The above files would have halted this push.")
```
ttaylorr: Nit: this could be de-dented since `Exit()` halts execution. Not sure what your preference is here, since this isn't as obvious as a `return` or `panic()` instead of a conditional.

More generally speaking, I'm glad that this warning is still being kept. I originally added it to warn about locks that _would_ have halted the push had lock verification been binding. This is rad, since if the state of lock verification is undefined, we can pretend like it is and _still_ warn about illegally modified files.

That being said, the behavior of `verifyLocks()` is to make an early-return if the lock verification state is disabled. That would cause this message not to be printed at all (even if the server _did_ support the locking API _and_ there were illegally modified files in the push). I think we should change the behavior of `verifyLocks()` to _always_ try and make the API call, and only propagate the error if one occurs and the state is enabled or undefined.

That would allow us to always present this message even if lock verification is turned off. I think this hardens the requirement of owning your locked files and would be a positive improvement 👍 .

technoweenie: To make sure I'm understanding properly, does this table describe what you're thinking?

| `locksverify` | API request | verified | not verified | error |
|--------------|-------------|--------|-------------|------|
| `null` | yes | silent | warning | warning |
| `true` | yes | silent | error | error |
| `false` | yes | silent | warning | silent |

ttaylorr: This is exactly what I was thinking ✨ .

technoweenie: Cool, sounds good to me :+1:

ttaylorr:
@technoweenie I updated this branch to exhibit the behavior we talked about in https://github.com/git-lfs/git-lfs/pull/2168#discussion_r112798604 & https://github.com/git-lfs/git-lfs/pull/2168#discussion_r112969188 in https://github.com/git-lfs/git-lfs/pull/2168/commits/81df03780b499257bf0d678e6efefdc866e6a08e.

commands/uploader.go
```diff
@@ -112,36 +114,34 @@ func newUploadContext(remote string, dryRun bool) *uploadContext {
...
+			if !errors.IsAuthError(err) {
+				Print("Remote %q does not support the LFS locking API. Consider disabling it with:", remote)
+				Print("  $ git config 'lfs.%s.locksverify' false", endpoint.Url)
+
+				if state == verifyStateEnabled {
+					ExitWithError(err)
+				}
+			} else {
 				ExitWithError(err)
 			}
```
technoweenie: This `ExitWithError` will raise for `verifyStateUnknown`, while the [table](https://github.com/git-lfs/git-lfs/pull/2168#discussion_r112969188) says it should be silent. I think the `if state == verifyStateEnabled` check could be moved outside the `errors.IsAuthError()` check.

commands/uploader.go
```diff
@@ -223,7 +223,14 @@ func (c *uploadContext) prepareUpload(unfiltered ...*lfs.WrappedPointer) (*tq.Tr
 			c.trackedLocksMu.Lock()
 			c.unownedLocks = append(c.unownedLocks, lock)
 			c.trackedLocksMu.Unlock()
-			canUpload = false
+
+			// If the verification state is enabled or undefined,
+			// this failed locks verification means that the push
+			// should fail.
```
technoweenie: This also doesn't fit the [table](https://github.com/git-lfs/git-lfs/pull/2168#discussion_r112969188), which states that pushes that aren't verified should just warn. Though I think being safer by default, and requiring the user to opt-out is a good move :+1:

commands/uploader.go
ttaylorr: OK, I think I got what you're saying here in: https://github.com/git-lfs/git-lfs/pull/2168/commits/1dc2b4730fd14dffc12e12f17fa91ca6ca88e0e3.

ttaylorr:
@technoweenie thanks for the review! I think I covered everything and this branch should be ready to merge.

technoweenie:
Looks good :+1: I can't approve since I created the PR though :)

Issue #2065, opened 2017-03-21
track command should be able to scan all filters to convert objects to LFS


technoweenie:
When you run `git lfs track <pattern>`, LFS will touch any matching files, and convert them to LFS objects in the next commit. However, there is no way to trigger this operation manually.

This is something that could also be run after a successful `git lfs clone` command.

ttaylorr:
I like this idea. I think calling this command `git lfs touch` would be good from my end.

shana:
Would `git lfs import` be more obvious of what it's doing, perhaps?

ttaylorr:
Hm, my initial thought with `import` would be that it modifies history to import all revisions of the given file, not just starting at the next commit.

shana:
Oh, good point. That's a highly destructive operation though, and I probably wouldn't recommend doing it on a file by file basis (unless the user really wants to?). The thing with `git lfs touch` is that `touch` is an implementation detail, not a description of what the operation is actually doing, which is "converting" a file to lfs.

ttaylorr:
@shana I think that's a great point re: `touch` being an implementation detail. There has been some discussion about a `migrate` command that would rewrite history and move files into LFS from Git.

Do you think it makes sense to punt on this in favor of that?

technoweenie:
> Do you think it makes sense to punt on this in favor of that?

Some users may want to just transition to LFS with `git lfs track`, while others may opt for a full repository rewrite. I think having both options is important. Also, `git lfs track <pattern>` already does this, just for a single pattern.

shana:
What @technoweenie said. Both are important, but require different "commitments" from the user and reflect slightly different needs:

- rewriting all history is a heavy operation that affects everyone. It would likely be done by a designated repo owner or admin in a coordinated action, since it will require a force push and perhaps disabling branch protection to do so. It can't be incorporated into a day-to-day routine, it would likely be done either as part of a specific migration from another source control or provider, or as part of a repo cleanup. `migrate` sounds like a good verb for this.

- setting up files for lfs on the next commit is light. It can be done by anyone, and can be part of a day-to-day routine. Given that `track` already does this, then it's practically done already, just need a good name for it 😄

ttaylorr:
@technoweenie @shana thanks for providing some insight. I agree 👍

Issue #2146, opened 2017-04-10
proposal: `migrate` sub-command


ttaylorr:
## Objective

Make it easier for people to start using Git LFS when they already have large
objects committed into Git repositories.

## Background

One of the common issues that people run into is going along the following path:

1. Start with a Git repository that has large objects in the history
2. `git lfs track *.dat`
3. Attempt to push to a remote, have the push fail a `post-receive` hook, and
stop using Git LFS.

The underlying problem is an expectation that `git lfs track` will convert
previously existing large Git blobs to LFS objects, which it does not do. `git
lfs track` only tracks object within LFS _after_ those changes to the
`.gitattributes` file have been cached.

What we have recommended thus far is to either run `git-filter-branch(1)` or
download an external tool with the same purpose (one example of this is BFG:
https://github.com/rtyley/bfg-repo-cleaner).

## Solution

I would like to make it possible for Git LFS to do this conversion without
needing to run a complicated `git-filter-branch(1)` command or use an external
tool. By implementing this in Git LFS ourselves, we have control of the behavior
of the migration tool, which would allow us to implement features like:

1. Forward and backward (to and from LFS objects) migrations
2. Specific file-path matching
3. Fast, LFS-specific object caching

If we could bundle a migration tool to convert large blobs in Git history to LFS
objects within Git LFS itself, we would also be able to:

1. Make it easier for users to adopt Git LFS
2. Cut down on a class of bug reports where `git lfs track` is run against large
objects that already exist

### Flags

I'd like the `git-lfs-migrate(1)` command to have the following options:

1. `-I`, `-X` (or `--include` and `--exclude`): our standard pattern matching
flags to dictate which objects should and shouldn't be tracked by Git LFS

A non-goal here is `--greater-than=` or `--less-than=` size flags, since this
type of pattern matching is not support in the attributes specification at
https://git-scm.com/docs/gitattributes.

### Sub-commands

At first I thought it would be unnecessary to supply a "reverse" (from LFS
object to Git large object) migration, because the migration command would be
able to build a parallel DAG of commit objects and then do a `git-update-ref(1)`
at the end, producing a `git reflog` like:

```bash
~/example-repo (master) $ git reflog
dc0b3c60 HEAD@{1}: checkout: moving from master-lfs-migrate to master
115a1a1 HEAD@{2}: commit: tools/time_tools: test tools.IsExpiredAtOrIn
...
```

[...]

PR #2175, opened 2017-04-24
commands/track: teach --no-modify-attrs

ttaylorr:
This pull request adds a `--no-modify-attrs` flag to the `git lfs track` command to "touch" matched tree entries while leaving `.gitattributes` unchanged.

The motivation (which has additional background in https://github.com/git-lfs/git-lfs/issues/2065) is that for files that you wish to "convert" to LFS, this should be a lightweight alternative to the destructive `migrate` command proposed in #2146.

Closes: https://github.com/git-lfs/git-lfs/issues/2065

---

/cc @git-lfs/core https://github.com/git-lfs/git-lfs/issues/2065

### DIFF
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -26,6 +26,7 @@ var (
 	trackNotLockableFlag    bool
 	trackVerboseLoggingFlag bool
 	trackDryRunFlag         bool
+	trackNoModifyAttrsFlag  bool
 )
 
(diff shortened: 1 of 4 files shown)


ttaylorr:
> Not thrilled about the extra nesting.

Agree 👍 

> But any refactoring should be done in an an isolated PR.

Agree 👍

PR #2177, opened 2017-04-24
debian/control: we need golang 1.8

### LINKED ISSUE #2176: debian/control should refer to golang 2:1.8, not golang 1.3
#1587 is about building inside git-lfs's project's servers; this is different.

On Ubuntu 17.04, a debian-workflow-oriented developer should be able to build git-lfs packages as follows:
```
git clone https://github.com/github/git-lfs.git
cd git-lfs
sudo mk-build-deps -i
debuild
```
but in the case where you install golang before enabling the universe repository,
debuild will fail to notice your golang is too old, and will try to build anyway, failing on
the golang 1.8isms.

The popular debian-based distros seem to all use epoch 2 for the version number,
so you need to change the required version to 2:1.8.

(The ubuntu 16.04 ppa isn't quite good enough to test the fix because it provides a version
which compares to less than 1.8, even though it's really 1.8.  To test there 
one has to remove the :native from debian/control to work around https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=819596, enable ppa:longsleep/golang-backports instead of universe, and fudge the minimum golang version number in debian/control to be 2:1.7.99 or something.)

dankegel:
Little tweak to match distro versioning for golang.  Should fix #2176 
Only tested on ubuntu 16.04 (see notes in commit log), but the real target is ubuntu 17.04.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: git-lfs
 Section: vcs
 Priority: optional
 Maintainer: Stephen Gelman <gelman@getbraintree.com>
-Build-Depends: debhelper (>= 9), dh-golang, golang-go:native (>= 1.3.0), git (>= 1.8.2), ruby-ronn
+Build-Depends: debhelper (>= 9), dh-golang, golang-go:native (>= 2:1.8), git (>= 1.8.2), ruby-ronn
 Standards-Version: 3.9.6
 
 Package: git-lfs


ttaylorr:
Hi @dankegel, thanks for opening this, and providing some background info in #2176. That was very helpful to me, since I didn't have anything to do with packaging LFS :-).

I booted a Docker image on `ubuntu:17.04` and was unable to build this correctly:

```
$ pwd
/git-lfs
$ debuild # ...
Correcting dependencies...Starting pkgProblemResolver with broken count: 1
Starting 2 pkgProblemResolver with broken count: 1
Investigating (0) git-lfs-build-deps:amd64 < 2.0.0 @iU mK Nb Ib >
Broken git-lfs-build-deps:amd64 Depends on golang-go:amd64 < none | 2:1.7~1ubuntu1 @un uH >(>= 2:1.8)
  Removing git-lfs-build-deps:amd64 because I can't find golang-go:amd64
Done
 Done
Starting pkgProblemResolver with broken count: 0
Starting 2 pkgProblemResolver with broken count: 0
Done
The following packages will be REMOVED:
  git-lfs-build-deps
0 upgraded, 0 newly installed, 1 to remove and 2 not upgraded.
1 not fully installed or removed.
```

Could you help?

peshidoe:
[](url)
[](124qsexqahqfnf5qnym8nwlbgnduxyu7qk
124qsexqahqfnf5qnym8nwlbgnduxyu7qk

/28265195/25785601/e4d4429a-3352-11e7-8c83-ac34baf013d1.png)

ttaylorr:
Hi @dankegel, I'd love to follow up sometime this week about the above (https://github.com/git-lfs/git-lfs/pull/2177#issuecomment-296804603).

ttaylorr:
@dankegel, I'm going to close this for now. Please let me know if you get a chance to look at this later -- I would love to fix this!

PR #2184, opened 2017-04-27
Streaming log stats

technoweenie:
This updates the http logging enabled by `GIT_LOG_STATS=1`:

* Updated the format to include more relevant info, including connection, dns, and tls timings. Uninteresting things like header sizes are left out.
* Instead of tracking requests in per-client slices protected by a mutex, the `*httpTransfer` types are passed around in the request context.
* The log is written to in real time, instead of at the end of the command.

### DIFF
--- a/commands/run.go
+++ b/commands/run.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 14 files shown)


technoweenie:
Since reviewing, I worked a couple more things in:

* Introduced timing for just downloading the response, based on the `GotFirstResponseByte` hook. Now we can see request/connection latency vs download time.
* Introduced an activity timeout. By default, the connection will time out if a read or write has not occurred in the last 10s. This should raise issues sooner for extremely bad or terminated connections.

ttaylorr:
👍

PR #2187, opened 2017-04-28
release: v2.1.0

ttaylorr:
---

/cc @git-lfs/core @git-lfs/releases


ttaylorr:
Both AppVeyor builds are failing because the `golang` package cannot be downloaded due to Chocolatey being offline. #2184 has an identical tree, so let's assume that those builds passed.

Issue #2190, opened 2017-04-30
Authentication required to push to a contributor's fork


jonasmalacofilho:
I can't seem to push to a contributor's fork anymore:

```
ERROR: Authentication error: Authentication required: You must have push access to verify locks
error: failed to push some refs to 'git@github.com:/lehonma/the-online-brt-planning-guide'
```

<details><summary>Context</summary>

Repository: https://github.com/ITDP/the-online-brt-planning-guide  
Maintainer fork: https://github.com/lehonma/the-online-brt-planning-guide  
PR/fork branch: ITDP/the-online-brt-planning-guide#158 (Chapter_17)
</details>

<details><summary><code>GIT_TRACE=1 GIT_CURL_VERBOSE=1 git push</code></summary>

```
03:57:43.271542 git.c:371               trace: built-in: git 'push'
03:57:43.275577 run-command.c:369       trace: run_command: 'ssh' 'git@github.com' 'git-receive-pack '\''/lehonma/the-online-brt-planning-guide'\'''
Host key fingerprint is SHA256:nThbg6kXUpJWGl7E1IGOCspRomTxdCARLviKw6E5SY8
+---[RSA 2048]----+
| =+o...+=o..     |
|o++... *o .      |
|*.o.  *o.        |
|oo.  ..o.= .     |
|.+o. .. S =      |
|*=+ .  o = .     |
|OE .  . o        |
| o     .         |
|                 |
+----[SHA256]-----+
03:57:44.841000 run-command.c:369       trace: run_command: '.git/hooks/pre-push' 'lehonma' 'git@github.com:/lehonma/the-online-brt-planning-guide'
03:57:44.853271 git.c:595               trace: exec: 'git-lfs' 'pre-push' 'lehonma' 'git@github.com:/lehonma/the-online-brt-planning-guide'
03:57:44.853361 run-command.c:369       trace: run_command: 'git-lfs' 'pre-push' 'lehonma' 'git@github.com:/lehonma/the-online-brt-planning-guide'
trace git-lfs: run_command: 'git' config -l
trace git-lfs: run_command: 'git' version
trace git-lfs: tq: running as batched queue, batch size of 100
trace git-lfs: ssh: git@github.com git-lfs-authenticate /lehonma/the-online-brt-planning-guide upload
trace git-lfs: HTTP: POST https://lfs.github.com/lehonma/the-online-brt-planning-guide/locks/verify
[...]

PR #2192, opened 2017-05-01
support lfs.<url>.* values where url does not include .git

technoweenie:
LFS 2.1.0 added the ability to specify some `lfs.*` and `http.*` values with a URL. You can set `lfs.*.access` based on the Endpoint URL:

```sh
$ git lfs env
git-lfs/2.2.0-pre (GitHub; darwin amd64; go 1.8.1; git 68340281)
git version 2.11.0

# enables basic auth for this repo
$ git config lfs.https://github.com/git-lfs/git-lfs.git/info/lfs.access basic
```

Most Git hosts don't require the `.git` extension, so you'd _think_ you could use that url in the config:

```sh
$ git remote get-url origin
https://github.com/git-lfs/git-lfs

$ git config lfs.https://github.com/git-lfs/git-lfs.access basic

$ git lfs env
git-lfs/2.2.0-pre (GitHub; darwin amd64; go 1.8.1; git 68340281)
git version 2.11.0

```

It fails here because it's checking the following git config keys:

* `lfs.https://github.com/git-lfs/git-lfs.git/info/lfs.access`
* `lfs.https://github.com/git-lfs/git-lfs.git/info.access`
* `lfs.https://github.com/git-lfs/git-lfs.git.access`
* `lfs.https://github.com/git-lfs.access`

It totally skips `https://github.com/git-lfs/git-lfs`! 

This PR adds an extra check if the url has `.git/info/lfs` in it.

* `lfs.https://github.com/git-lfs/git-lfs.git/info/lfs.access`
* `lfs.https://github.com/git-lfs/git-lfs.git/info.access`
* `lfs.https://github.com/git-lfs/git-lfs.git.access`
* `lfs.https://github.com/git-lfs/git-lfs.access`
* `lfs.https://github.com/git-lfs.access`

/cc https://github.com/git-lfs/git-lfs/issues/2190

### DIFF
--- a/config/url_config.go
+++ b/config/url_config.go
@@ -55,12 +55,19 @@ func (c *URLConfig) getAll(prefix, rawurl, key string) []string {
 
 	for i := len(paths); i > 0; i-- {
 		for _, host := range hosts {
-			path := strings.Join(paths[:i], "/")
+			path := strings.Join(paths[:i], slash)
 			if v := c.git.GetAll(fmt.Sprintf("%s.%s/%s.%s", prefix, host, path, key)); len(v) > 0 {
 				return v
 			}
 			if v := c.git.GetAll(fmt.Sprintf("%s.%s/%s/.%s", prefix, host, path, key)); len(v) > 0 {
 				return v
+			}
+
+			if isDefaultLFSUrl(path, paths, i) {
+				path = path[0 : len(path)-4]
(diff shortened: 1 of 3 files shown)


technoweenie:
It should be in 2.1.1, but we should wait a few days to see if anything else comes up.

PR #2198, opened 2017-05-02
Bugfix: cannot push to scp style URL

jiangxin:
When push to a scp style url, like:

    git push git@github.com:git-lfs/git-lfs.git HEAD

Git-lfs will complain about bad url.  This is because golang 1.8 has
changed the implement of `url.Parse` method.  It will fail with a nil
pointer and an error if the remote url is a scp style URL.

Make a golang compatible test, I.E. check null pointer, then check u.Scheme.

Signed-off-by: Jiang Xin <xin.jiang@huawei.com>

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -251,15 +251,8 @@ func ValidateRemote(remote string) error {
 
 // ValidateRemoteURL checks that a string is a valid Git remote URL
 func ValidateRemoteURL(remote string) error {
-	u, err := url.Parse(remote)
-	if err != nil {
-		return err
-	}
-
-	switch u.Scheme {
-	case "ssh", "http", "https", "git":
-		return nil
-	case "":
+	u, _ := url.Parse(remote)
+	if u == nil || u.Scheme == "" {
 		// This is either an invalid remote name (maybe the user made a typo
 		// when selecting a named remote) or a bare SSH URL like
 		// "x@y.com:path/to/resource.git". Guess that this is a URL in the latter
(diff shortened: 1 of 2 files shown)


technoweenie:
Hey @jiangxin, this change looks good. Would you mind adding a test to `git/git_test.go`? If we had written them to start, we would've noticed them immediately after upgrading to go 1.8 :/

jiangxin:
Add testcase TestValidateRemoteURL in a separate  commit, and it will fail without this patch if is compiled using golang 1.8+.

ttaylorr:
@jiangxin thanks for adding this test case. I pulled your branch down locally and was able to observe this test pass under 1.7.x:

```
~/g/git-lfs (pull-2198) $ brew switch go 1.7.4_1
# ...
3 links created for /usr/local/Cellar/go/1.7.4_1
~/g/git-lfs (pull-2198) $ go version
go version go1.7.4 darwin/amd64
~/g/git-lfs (pull-2198) $ go test -v ./git -run TestValidateRemoteURL
=== RUN   TestValidateRemoteURL
--- PASS: TestValidateRemoteURL (0.00s)
PASS
ok      github.com/git-lfs/git-lfs/git  0.011s
```

Though with this diff applied, it fails under 1.7.x, as expected:

```diff
diff --git a/git/git_test.go b/git/git_test.go
index 0f3f95d9..5e62653e 100644
--- a/git/git_test.go
+++ b/git/git_test.go
@@ -540,5 +540,5 @@ func TestValidateRemoteURL(t *testing.T) {
        assert.Nil(t, ValidateRemoteURL("ssh://git@github.com:22/git-lfs/git-lfs"))
        assert.Nil(t, ValidateRemoteURL("git@github.com:git-lfs/git-lfs"))
        assert.Nil(t, ValidateRemoteURL("git@server:/absolute/path.git"))
-       assert.NotNil(t, ValidateRemoteURL("ftp://git@github.com/git-lfs/git-lfs"))
+       assert.NotNil(t, ValidateRemoteURL("git@github.com:git-lfs/git-lfs.git"))
 }
```

Your change looks good to me, but I'm curious if there's a reason you went with `ftp://git@...` over the example you provided in the pull request body.

Issue #2193, opened 2017-05-01
Need documentation for repository repair or way to actually force a push


allaryin:
I have a client with a 30gb repository stored primarily in LFS (via GitLab). If I do a fetch --all, I get 30 gigs of files, plus 404's on 6 of them. But the 6 problematic files all exist on disk and contain valid data - so I am assuming that these 404's are merely for ancient historical versions of the files.

I need to create a full mirror of the repository, but cannot because git bails with "LFS upload failed" on the six files in question (after having successfully uploaded the other 30gb of data to our server).

After fighting this issue for several days, upgrading git-lfs to 2.1.0, and continuing to fight it for two more, I am at a loss. I cannot find any comprehensive documentation on how to simply make this push happen so I can move on to actually using the data. I don't care about losing a little history, just as long as the bulk remains intact.

ttaylorr:
Hi @allaryin, thanks for opening this issue and giving me a detailed look at what you're trying to accomplish. Here are some thoughts:

The behavior that you're encountering was formalized in https://github.com/git-lfs/git-lfs/pull/2082, and is meant to warn users about all missing objects in a push, so they can fix them at once for the next push. This doesn't quite match the behavior you're looking for, which I believe is to just ignore the missing files and go on.

Right now, there's no "blessed" way to do this. We had discussed adding a `--strict-mode` flag (in https://github.com/git-lfs/git-lfs/pull/2078), where the absence of such a flag would allow pushes like the ones that you're describing. Ultimately, this did not make it into master.

The best way to get around the missing data right now is to:

- rewrite the points in history that reference those missing pointers, or
- add the file back to `.git/lfs/objects`

allaryin:
(I'd love to have asked this elsewhere, but the chat room linked from the project page is dead and this appears to be the only place questions get answered)

The man files are spartan, and since you chose recently not to implement a feature that might have actually solved this problem, I stand by my original request for documentation. As it is, most of the leads I've followed have been in the form of ticket history like this... which is frankly kind of sad. A project with this level of visibility needs better formal documentation - especially when issues like this are complete showstoppers for people.

I can't simply replace the files, as I am apparently missing the versions of the files that are missing. But even knowing what the hash you output meant and where it was stored on disk involved reading the protocol specification.

[...]

PR #2199, opened 2017-05-02
commands/uploader: allow incomplete pushes

ttaylorr:
This pull request allows the uploader to push ranges with incomplete or missing/corrupt LFS objects as given by the `lfs.allowincompletepush` configuration value.

Closes: #2063.

---

/cc @git-lfs/core @allaryin https://github.com/git-lfs/git-lfs/issues/2193 https://github.com/git-lfs/git-lfs/issues/2063

### DIFF
--- a/test/test-pre-push.sh
+++ b/test/test-pre-push.sh
@@ -200,6 +200,36 @@ begin_test "pre-push with missing pointer not on server"
 )
 end_test
 
+begin_test "pre-push with missing object (lfs.allowincompletepush)"
+(
+  set -e
+
+  reponame="$(basename "$0" ".sh")-missing-pointer-allow-incomplete"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" "$reponame"
+
+  oid="7aa7a5359173d05b63cfd682e3c38487f3cb4f7f1d60659fe59fab1505977d4c"
+
+  echo "$(pointer "$oid" 4)" > new.dat
+  git add new.dat
+  git commit -m "add new pointer"
+
+  git config "lfs.allowincompletepush" "true"
+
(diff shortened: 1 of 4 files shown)


ttaylorr:
This appears to be working: https://github.com/git-lfs/git-lfs/issues/2193#issuecomment-298963675.

test/test-pre-push.sh
```diff
@@ -200,6 +200,36 @@ begin_test "pre-push with missing pointer not on server"
...
+    git lfs pre-push origin "$GITSERVER/$reponame" 2>&1 |
+    tee push.log
+
+  if [ "0" -ne "${PIPESTATUS[1]}" ]; then
+    echo >&2 "fatal: expected 'git lfs pre-push origin $GITSERVER/$reponame' to succeed"
+    exit 1
+  fi
+
+  grep "LFS upload missing objects" push.log
+  grep "  (missing) new.dat ($oid)" push.log
```
ttaylorr: @technoweenie thanks for the feedback. I think testing this most directly is the best approach, so I nuked the `echo | git lfs pre-push` entirely and moved this to its own singular test in `test/test-push.sh`.

Can you take a look at https://github.com/git-lfs/git-lfs/pull/2199/commits/b9539edea75ca3916631b0bb87ae2d4c7ec490df and let me know if that works?

technoweenie: I would've left the `pre-push` test too. IMO testing both in integration tests is important. It only happens to work in both because they both now use the Uploader type.

ttaylorr: @technoweenie, ah -- this makes sense. I brought back the `pre-push` test that covers the case with missing and present objects. I think that test and the one in `test/test-push.sh` should be enough to fully cover this behavior.

PR #2203, opened 2017-05-07
Fix debian 9 (stretch) builds

ssgelm:
Sorry for the delay on this - I've had it on my todo list for months but hadn't gotten around to it.

The fix was to skip the `dh_golang` step which is used to add the `Built-using` field to the resulting deb.  The field is only used to track the version of go used to build a package so omitting it shouldn't be a problem.  In the commit at https://anonscm.debian.org/cgit/collab-maint/dh-golang.git/commit/script/dh_golang?id=7c3fbec6ea92294477fa8910264fe9bd823f21c3 the dh-golang maintainers changed the logic used to determine the version of go used to check the installed package.  This fails because git-lfs gets built using a manually installed version of go.

This fixes #1472.

PR #2205, opened 2017-05-08
build: publish Debian 9-compat packages

ttaylorr:
This pull request expands on #2203 to include Debian 9 in the build matrix, and distribute packages to Linux distributions built on/with Debian 9.

---

/cc @git-lfs/core

### DIFF
--- a/docker/run_dockers.bsh
+++ b/docker/run_dockers.bsh
@@ -54,7 +54,7 @@ while [[ $# > 0 ]]; do
 done
 
 if [[ ${#IMAGES[@]} == 0 ]]; then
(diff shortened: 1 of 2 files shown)


andyneff:
@ttaylorr Mint 18 and 18.1 (sarah and serena) are not actually based on debian 9

They are based on Ubuntu 16.04 (Xenial). This looks like an oversight and they should have been added previously.

ttaylorr:
@andyneff great catch, thanks! I went ahead and fixed this in https://github.com/git-lfs/git-lfs/pull/2205/commits/cbf8c03711f9fd57df3f3f59d74a522cef93373d.

I think this is ready to merge. @technoweenie, could you 👍 ?

Issue #2204, opened 2017-05-08
Mismatch of LFS version causing trouble (?)


FredyC:
I don't want to fill #2202 with irrelevant information. The whole deal is that my coworker is rather inexperienced with Git, not even mentioning LFS (not that I know much about LFS). I went on vacation and he was working alone, pushing stuff.

When I came back I found out that I cannot pull changes. There is a bunch of errors like this for files he changed.

> LFS: Repository or object not found: https://gitlab.com/group/project.git/gitlab-lfs/objects/2dbb7e94eec56de3ec380469bfb7e6bb335cf78dee658dd22fab842c380e07ef
> Check that it exists and that you have proper access to it

So I asked what he did, but he isn't really sure that he did something special except at some point he updated lfs to 2.x. Apparently, those files were not ever pushed to LFS from his side and it's not possible now.

There is also some strange issue regarding locking...

> git lfs push origin master --all
> Remote "origin" does not support the LFS locking API. Consider disabling it with:
>  $ git config 'lfs.https://gitlab.com/group/project.git/info/lfs.locksverify' false
> Git LFS: (0 of 193 files, 1817 skipped) 0 B / 139.44 MB, 909.54 MB skipped 

> git config 'lfs.https://gitlab.com/group/project.git/info/lfs.locksverify' false
> error: invalid key: 'lfs.https://gitlab.com/group/project.git/info/lfs.locksverify'

He has sent me copy of his repo (.git folder including), but I cannot seem to find a way to fix it even on my older `git-lfs/1.5.2 (GitHub; windows amd64; go 1.7.3; git 547a06aa)`. Everything is skipped during push.

Also running `git lfs status` actually shows these missing files in `Git LFS objects to be pushed to origin/master:` and yet running push skips them all.

technoweenie:
I don't think the LFS version mismatch is causing any issues here. Or if so, the issues are likely better described by #2202 or #2201. We'll look into those first, and revisit this if necessary later.

technoweenie:
> There is also some strange issue regarding locking...

```
git lfs push origin master --all
Remote "origin" does not support the LFS locking API. Consider disabling it with:
$ git config 'lfs.https://gitlab.com/group/project.git/info/lfs.locksverify' false
Git LFS: (0 of 193 files, 1817 skipped) 0 B / 139.44 MB, 909.54 MB skipped
git config 'lfs.https://gitlab.com/group/project.git/info/lfs.locksverify' false
[...]

PR #2206, opened 2017-05-08
remove quotes from git config syntax

technoweenie:
The quoted syntax doesn't work on Windows. Probably a bash vs Windows cmd thing.

/cc https://github.com/git-lfs/git-lfs/issues/2204#issuecomment-299907812


ttaylorr:
Sorry, approved this a little early. I think some of the tests in `test/test-pre-push.sh` will need to be updated to not expect the `'` characters.

PR #2216, opened 2017-05-12
config: allow multiple environments when calling config.Unmarshal

ttaylorr:
This pull request allows the ability to use multiple environments given an ordering when calling `config.Configuration.Unmarshal`.

For instance, the following code _would_ have `panic()`'d in previous versions of LFS:

```go
type T struct {
        X string `os:"X" git:"lfs.x"`
}
```

In this pull request, calling `config.Configuration.Unmarshal()` on the above code will produce the following behavior:

1. Try and find a value of `X` in the OS environment.
2. If the above value could not be found, try and find it in the Git environment.
3. If the above value is still missing, use the value of X itself (to allow for defaults).

Unfortunately, this required parsing struct tags ourselves, because the `reflect.StructTag` type does not specify an ordering, and rather is defined as: `type reflect.StructTag string`. A simple regexp was written to accomplish this goal.

This is used for work done in a following PR to satisfy looking for the `GIT_SSH_COMMAND` variable in either the OS environment, or Git's own `core.sshCommand`.

---

/cc @git-lfs/core

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -3,9 +3,11 @@
 package config
 
 import (
-	"errors"
 	"fmt"
 	"reflect"
+	"regexp"
+	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/git-lfs/git-lfs/tools"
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -156,27 +153,59 @@ func (c *Configuration) Unmarshal(v interface{}) error {
...
+
 // parseTag returns the key, environment, and optional error assosciated with a
 // given tag. It will return the XOR of either the `git` or `os` tag. That is to
 // say, a field tagged with EITHER `git` OR `os` is valid, but pone tagged with
 // both is not.
 //
 // If neither field was found, then a nil environment will be returned.
-func (c *Configuration) parseTag(tag reflect.StructTag) (key string, env Environment, err error) {
-	git, os := tag.Get("git"), tag.Get("os")
+func (c *Configuration) parseTag(tag reflect.StructTag) []*lookup {
```
technoweenie: What if this returned `map[string]Environment, error`? Then you could drop the `lookup` type and call `(Environment) Get(key)` or whatever. Then, it could directly pass the error to `Unmarshal()`, instead of calling `panic`.

ttaylorr: I think changing the signature to `([]*lookup, error)` would be good. I originally implemented this with a `map[string]Environment` but the ordering of struct tags is important so an ordered data type was a necessity.

PR #2220, opened 2017-05-15
git/odb: loose object read/write

ttaylorr:
This pull request introduces a new package, `git/odb` and implements loose object reading and writing necessary for the work described in #2146.

Here's a breakdown of the core concepts:

- An `*odb.ObjectDatabase` capable of reading and writing `Object`s from a `Storer`.
- An `Object` interface which specifies a type, and a encode/decode method.
- A `Storer` interface, which abstracts the filesystem away from the object database.
- Several implementations of `Object` for `Blob`, `Tree` and `Commit`.

The documentation in this package is authoritative, and it will be a good place to look for details beyond the broad overview provided above.

---

/cc @git-lfs/core for review
/cc @peff for thoughts on Git-specific details


ttaylorr:
Looks like there's a timezone issue in the `TestCommitDecoding` test. I'll look into that shortly.

git/odb/commit.go
```diff
@@ -0,0 +1,216 @@
...
+// ParseSignature parses a given string into a signature instance, returning any
+// error that it encounters along the way.
+//
+// ParseSignature expects the signature encoded in the given string to be
+// formatted correctly, and reproduce-able by the Signature.String() function
+// above.
+func ParseSignature(str string) (*Signature, error) {
+	name := signatureNameRe.FindString(str)
+	emailParts := signatureEmailRe.FindStringSubmatch(str)
+	timeParts := signatureTimeRe.FindAllStringSubmatch(str, 2)
```
peff: In git.git we have some gross heuristics to do our best with malformed entries. E.g., we parse the name/email left-to-right, but then parse the timestamp from the end, which leaves malformed or duplicated email bits in the middle as missing cruft. I don't know how thorough you'd want to be here. In fact, I'm not quite sure why you need to parse these ident lines in the first place. An LFS migration should be able to just treat them as opaque.

git/odb/commit.go
```diff
@@ -0,0 +1,216 @@
...
+	// ParentIds are the IDs of all parents for which this commit is a
+	// linear child.
+	ParentIds [][]byte
+	// TreeId is the root Tree associated with this commit.
+	TreeId []byte
+	// Message is the commit message, including any signing information
+	// associated with this commit.
+	Message string
+}
+
```
peff: You may encounter other headers besides these. At the very least, Git may generate `encoding`, `gpgsig`, and `mergetag` headers. It's not encouraged, but you could also see other random headers added by other implementations.

You'd want to preserve these in general (though for something like `gpgsig`, you'd eventually want to support re-signing). And you'd probably want to do so as byte-for-byte identically as you can, just replacing the parent/tree sha1 pointers as needed.

It looks like your approach is to do a Decode/Modify/Encode pair for each object. While that's pleasant to work with during the modify step, I think you may want to consider instead doing a line-by-line filter, where you rewrite the lines whose sha1s are modified, and otherwise try to keep everything byte-for-byte identical.

git/odb/object_reader.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		return UnknownObjectType, 0, err
+	}
+	sizeStr = strings.TrimSuffix(sizeStr, "\x00")
+
+	size, err = strconv.ParseInt(sizeStr, 10, 64)
+	if err != nil {
+		return UnknownObjectType, 0, err
+	}
+
```
peff: The object reading all looks sane from the Git side (bearing in mind that I don't know go very well, so I'm just assuming that functions do what I'd expect from the context).

git/odb/object_type.go
```diff
@@ -0,0 +1,45 @@
...
+	switch strings.ToLower(s) {
+	case "blob":
+		return BlobObjectType
+	case "tree":
+		return TreeObjectType
+	case "commit":
+		return CommitObjectType
+	default:
+		return UnknownObjectType
+	}
```
peff: You may also encounter tags in the wild. It probably makes sense to ignore them for now, though, and get the basic algorithm working.

git/odb/file_storer.go
```diff
@@ -0,0 +1,65 @@
...
+// is complete.
+func (fs *FileStorer) Create(sha []byte) (f io.ReadWriteCloser, err error) {
+	path := fs.path(sha)
+	dir := filepath.Dir(path)
+
+	if err = os.MkdirAll(dir, 0755); err != nil {
+		return nil, err
+	}
+	return fs.open(path, os.O_RDWR|os.O_CREATE|os.O_EXCL)
+}
```
peff: You should avoid opening the object path directly. While you're writing, simultaneous readers will see a corrupted object. And if the program is interrupted for any reason, you'll leave a half-written object on disk.  The usual git.git way is to open a temporary file in the `objects` directory, and then `rename()` or `link()` the file into place. Those syscalls are atomic with respect to other processes trying to read the object; they either see the whole thing or nothing at all.

As a bonus, this makes it easy to stream files into the object database; you open the tempfile, stream the contents into place, doing zlib compression and computing the sha1 on the fly, and then rename the file based on the final sha1 you computed.

git/odb/tree.go
```diff
@@ -0,0 +1,123 @@
...
+		modes, err := buf.ReadString(' ')
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return n, err
+		}
+		n += len(modes)
+		modes = strings.TrimSuffix(modes, " ")
+		modes = fmt.Sprintf("%06s", modes)
```
peff: I'm confused what this `sprintf` is doing. Especially with a precision specifier. We're on the reading side, so it's not formatting the mode. Why can't we just skip straight to `ParseInt`? (I have a feeling I'm missing something go-specific here).

git/odb/tree.go
```diff
@@ -0,0 +1,123 @@
...
+		switch mode & syscall.S_IFMT {
+		case syscall.S_IFREG:
+			typ = BlobObjectType
+		case syscall.S_IFDIR:
+			typ = TreeObjectType
+		case syscall.S_IFLNK:
+			typ = BlobObjectType
+		default:
+			panic(fmt.Sprintf("git/odb: unknown object type: %q %q", modes, fname))
+		}
```
peff: You may run into gitlinks for submodules here, too (mode 0160000). That's probably something you can ignore for the initial run, but you'll see in the wild.

git/odb/tree.go
```diff
@@ -0,0 +1,123 @@
...
+
+// Encode encodes the tree's contents to the given io.Writer, "w". If there was
+// any error copying the tree's contents, that error will be returned.
+//
+// Otherwise, the number of bytes written will be returned.
+func (t *Tree) Encode(to io.Writer) (n int, err error) {
+	const entryTmpl = "%s %s\x00%s"
+
+	for _, entry := range t.Entries {
+		fmode := strconv.FormatInt(int64(entry.Filemode), 8)
```
peff: Good, you didn't make the classic 0-padding mistake. :+1:

Tree-formatting looks good overall.

git/odb/tree.go
ttaylorr: @peff good call, and I hadn't thought of this while writing the initial Tree `encode`/`decode` implementation. I did a series of `git cat-file -p` traversals into a directory in my dotfiles that had submodules, and found that the representation is `commit`:

```
~/.dotfiles (work-gh) $ g cat-file -p 8817b34ec1f834148150c046cca6f72c5c8c8b23
160000 commit a16a9b63eb85cc0960a7f25c54647ac1f99f3360  ack.vim
# ...
```

So I've implemented the same behavior in ec4c9005.

peff: > and found that the representation is commit:

Sort of. Gitlink entries do point to commits, but note that this is the one object pointer in Git that doesn't require that the object be present (because it's probably in a submodule). So you wouldn't want to recurse into it the same way as the other tree entries.

ttaylorr: > So you wouldn't want to recurse into it the same way as the other tree entries.

That's fine, there's no tree recursion implemented yet, but I'll make a note of this for when I do tree rewriting.

git/odb/tree.go
ttaylorr: Good catch, this is a misconception on my part. Since Git removes leading zeros on the left, I assumed the wrong byte order and thought that adding those zeroes back would be significant (this is what "%06s" does, add `len(s)-6` `0`'s to the front of `%s`).

My byte-order assumption was wrong, so I removed this line in 39d000c7.

git/odb/object_type.go
ttaylorr: Good call. I'll make a note of this once the initial algorithm is functional.

git/odb/commit.go
ttaylorr: Interesting. Though not ideal, I _believe_ that extra header data under this implementation would be instead added to the commit's "message". This wouldn't produce a byte-for-byte representation, though, since a `LF` would be inserted before the commit "message".

That being said, I think since other implementations can add additional headers a `headers []*struct{K, V string}` would be ideal. A `map[string]string` is what we want behaviorally, but Go's map implementation does not guarantee ordering, so we wouldn't for sure be able to generate a byte-for-byte copy.

git/odb/commit.go
ttaylorr: I think the implementation should be fine as is, since the `signatureTimeRe` is anchored to the end of the string `$`, so between that, the name regexp, and the email regexp, any cruft between the last `>` and first digit of the timestamp would be ignored.

git/odb/file_storer.go
ttaylorr: Good call. The hashing and zlib compression is handled by the `*odb.ObjectWriter`, but I removed the `Create(sha []byte) (*os.File, error)` method and replaced it with `Store(sha []byte, r io.Reader) (n int64, err error)`, which allows to stream in the data, and then store it in the object database atomically.

ttaylorr: (That was done in 83f5bea8).

git/odb/storer.go
```diff
@@ -0,0 +1,16 @@
+package odb
+
+import "io"
+
+// Storer implements a storage engine for reading, writing, and creating
+// io.ReadWriters that can store information about loose objects
+type Storer interface {
```
technoweenie: This doesn't need to be exported, since it's only used on a private `ObjectDatabase` property.

git/odb/object_writer.go
```diff
@@ -0,0 +1,123 @@
...
+// writes into the io.Writer given, "w".
+func NewObjectWriter(w io.Writer) *ObjectWriter {
+	return NewObjectWriteCloser(&nopCloser{w})
+}
+
+// NewObjectWriter returns a new *ObjectWriter instance that drains incoming
+// writes into the io.Writer given, "w".
+//
+// Upon closing, it calls the given Close() function of the io.WriteCloser.
+func NewObjectWriteCloser(w io.WriteCloser) *ObjectWriter {
```
technoweenie: I think these should be `NewWriter()` and `NewObjectWriteCloser()`, to be consistent with their `read` counterparts. I think having `Object` in the name is redundant when you consider they're from the `odb` package. Though if you think `Object` is preferred, then I think both the reader and writer functions should be consistent at least.

git/odb/storer.go
ttaylorr: @technoweenie good call. I made the `Storer` and all of the concrete implementing types (`*FileStorer`, `*MemoryStorer`) private in c72121dc.

git/odb/object_writer.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+}
+
+// WriteHeader writes object header information and returns the number of
+// uncompressed bytes written, or any error that was encountered along the way.
+//
+// WriteHeader MUST be called only once, or a panic() will occur.
+func (w *ObjectWriter) WriteHeader(typ ObjectType, len int64) (n int, err error) {
+	if !atomic.CompareAndSwapUint32(&w.wroteHeader, 0, 1) {
+		panic("git/odb: cannot write headers more than once")
```
ttaylorr: @technoweenie interesting question. I think I could go either way on this. My thought when implementing this was that if a program calls `Write()` before `WriteHeader()`, something is wrong at a higher level and any error returned from `Write()` or `WriteHeader()` would not be respected, so we should just terminate.

With respect to LFS specifically, all uses of the `*ObjectWriter` from the `*ObjectDatabase` type call these two in the correct order, so I think I'm inclined to leave this as-is.

git/odb/object_writer.go
ttaylorr: @technoweenie good catch. I think having these be called `NewObjectReader` and `NewObjectReadCloser` (with identical `Writer` equivalents) is preferable to dropping "Object" from the signature, and I've gone ahead and made that change in 2c6ddd6c.

I think keeping "Object" is important, because these are not Object Database readers (as odb.Reader` would suggest). Rather, they are Object Readers, reading individual objects, not the whole database.

git/odb/object_reader.go
```diff
@@ -0,0 +1,141 @@
...
+	return r.header.typ, r.header.size, nil
+}
+
+// Read reads uncompressed bytes into the buffer "p", and returns the number of
+// uncompressed bytes read. Otherwise, it returns any error encountered along
+// the way.
+//
+// This function is safe to call before reading the Header information, as any
+// call to Read() will ensure that read has been called at least once.
+func (r *ObjectReader) Read(p []byte) (n int, err error) {
```
ttaylorr: @technoweenie this function is needed, since `io.Copy()` expects it to be there when consuming data off of an `io.Reader` implementation.

git/odb/commit.go
```diff
@@ -0,0 +1,216 @@
...
+	// contents.
+	Author *Signature
+	// Committer is the individual or entity that added this commit to the
+	// history.
+	Committer *Signature
+	// ParentIds are the IDs of all parents for which this commit is a
+	// linear child.
+	ParentIds [][]byte
+	// TreeId is the root Tree associated with this commit.
+	TreeId []byte
```
ttaylorr: Good catch, I renamed both of those in cd23172f37dddb22ccd1625ef3dd859405997fb2 and 8e7b810b7802485c567fab89e8aa8a6a03a99590, respectively.

git/odb/commit.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	email := emailParts[1]
+
+	if len(timeParts) < 1 {
+		return nil, errors.Errorf("git/odb: expected time in signature: %q", str)
+	}
+
+	epoch, err := strconv.ParseInt(timeParts[0][0], 10, 64)
+	if err != nil {
+		return nil, err
```
ttaylorr: Good call, I fixed that up in 609ca00d.

git/odb/blob.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Encode encodes the blob's contents to the given io.Writer, "w". If there was
+// any error copying the blob's contents, that error will be returned.
+//
+// Otherwise, the number of bytes written will be returned.
+func (b *Blob) Encode(to io.Writer) (n int, err error) {
+	nn, err := io.Copy(to, b.Contents)
+	if err != nil {
+		return 0, err
```
ttaylorr: Good call, I fixed this in: 8d232a4f.

git/odb/blob.go
```diff
@@ -0,0 +1,64 @@
...
+	// may only be read once. It may or may not implement io.ReadSeeker.
+	Contents io.Reader
+
+	// closeFn is a function that is called to free any resources held by
+	// the Blob.  In particular, this will close a file, if the Blob is
+	// being read from a file on disk.
+	closeFn func() error
+}
+
+var _ Object = (*Blob)(nil)
```
ttaylorr: I definitely go back and forth on whether or not they are useful. I agree that they are not useful for outside consumers, or for compile-time checks, since these types are being used in the `*ObjectDatabase` as fillers for their interface type.

That being said, I do find them very useful when looking at the code myself as a reminder of which interfaces a type is supposed to implement.

git/odb/commit.go
peff: Yeah, I think a Decode/Encode round-trip would corrupt the object. That might be an interesting test case. You _should_ be able to go into a real repo, do a Decode/Encode round-trip on every object, and end up with a byte-for-byte identical copy.

git/odb/commit.go
peff: > between that, the name regexp, and the email regexp, any cruft between the last `>` and first digit of the timestamp would be ignored.

One issue, though, is that decoding for display and decoding to do Decode/Encode round-trip are two different things. For display, you want to ignore cruft and show the user the best thing you can manage to parse out the mess.

For a rewrite tool, though, you probably want to preserve any bytes that you aren't explicitly meaning to change. The existing bytes may be wrong (and the rewrite may be an opportunity to fix them), but it should be something the user explicitly asks for.

ttaylorr: > For a rewrite tool, though, you probably want to preserve any bytes that you aren't explicitly meaning to change.

I think this is a great point. From the migrator's point of view, the signature should essentially be treated as opaque. I changed the signature from:

```diff
type Commit struct {
-        Author *Signature
-        Committer *Signature
+        Author string
+        Committer string
}
```

but kept the `*Signature` type as it makes it easy to drive out tests. This change occurred in: b48f4026.

git/odb/commit.go
ttaylorr: @peff good call, I added an `ExtraHeaders` array to the `*Commit` type in cb5ab7c6. This should make byte-for-byte round trips possible.

git/odb/blob.go
technoweenie: If that's all, why not just throw it in the comments for each type? These var declarations get flagged one very review that  you add them :)

ttaylorr:
Thanks all for the review 🙇 .

PR #2228, opened 2017-05-17
commands: fix logged error not interpolating format qualifiers

ttaylorr:
This pull request fixes an error I noticed in our `LoggedError` and `logPanic` functions that format variables like `"%+v"` weren't being formatted correctly when viewing the output of `git lfs logs last`.

This problem was introduced in https://github.com/git-lfs/git-lfs/pull/2178.

---

/cc @git-lfs/core


ttaylorr:
> LOL how long has this been here??? 👍

Just since 2.1.0, but still not great 😅 .

PR #2231, opened 2017-05-17
Filepathfilter top level fix

### LINKED ISSUE #2221: git-lfs handling of gitignore for directories leads to git operation slowdowns
I'm tracking down performance degradations that match some observations from #1750.

We have `.gitignore` entries at the base for output directories that can exist at any stage of the tree - ie `obj/...`, `a/b/obj/...` `c/obj/...` , which are ignored via `.gitignore` rule at the base of the tree

```
# Output directories
obj/
```

Running

```
$ strace -f -o log git lfs track
```

Shows that the obj-directories are being visited (hint some of these directories have absurdly big contents - think full linux kernel tree)

From https://github.com/git-lfs/git-lfs/issues/1750#issuecomment-267281394 the suggestion is to use `**/obj/**` which does indeed properly exclude these directories from git-lfs thus dropping track time from 9 seconds to 0.048 seconds.  (note 9 seconds is more like 30 seconds when the disk-cache isn't warm).  That 9 second duration is incurred on every step of a rebase.

Now the comment about using `**/dir/**` was made "in-passing", but it seems having different interpretation of the `.gitignore` file compared to git is a bug of some sort, that either needs:

1. Justification that git its not an error
  1. and thus potentially a programmatic warning in git-lfs 
  1. but definitely documentation or best practices
2. Deferring
  1. and again documentation
3. Fixing

I couldn't find anywhere that this issue was tracked, hence I've logged this.

```
$ git version
git version 1.9.1

$ git lfs version
git-lfs/2.1.0 (GitHub; linux amd64; go 1.8.1)

$ lsb_release -a
...
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.5 LTS
Release:	14.04
Codename:	trusty
```

technoweenie:
Attempts to fix #2221:

> We have `.gitignore` entries at the base for output directories that can exist at any stage of the tree - ie `obj/...`, `a/b/obj/...`, `c/obj/...` , which are ignored via `.gitignore` rule at the base of the tree

```
# Output directories
obj/
```

However, when `tools.FastWalkGitRepo()` builds the exclude pattern from `.gitignore`, it only 
matches the top level `obj/...` paths, and not any sub paths.

The core fix to `filepathfilter` is basically:

```diff
# name = "a/b/obj/..."
# p.rawpattern = "obj"
 matched, _ := filepath.Match(p.rawPattern, name)
-return matched || strings.HasPrefix(name, p.rawPattern+string(filepath.Separator))
+regpattern = fmt.Sprintf(`(\A|%s)%s(%s|\z)`,
+	string(filepath.Separator),
+	p.rawPattern,
+	string(filepath.Separator),
+)
+regexp.MustCompile(regpattern).MatchString(name)
```

Instead of using `strings.HasPrefix`, it builds a regex like `(\A|/)obj(/|\z)`.

If only it was that simple :feelsgood:   The commits came in 3 stages:

1. `FastWalkGitRepo` was building the patterns based on the absolute path to the repository files. So, instead of matching `(\A|/)obj(/|\z)` against `a/b/obj/...`, it was matching `(\A|/)/Users/rick/foo/obj(/|\z)` against `/Users/rick/foo/a/b/obj/...`. So it had to be updated to pass the absolute root dir and the relative repository working dir around in the private functions. 
2. Next, `filepathfilter` was updated to use the regex pattern mentioned above. In doing so, I noticed that the old `(*basicPattern) Match()` was exactly the same as `(*doubleWildcardPattern) Match()`. So, I removed the redundant `basicPattern` type, and gave the remaining pattern types more generic names.
3. Update `FastWalkGitRepo` tests to ensure it's matching the sub directory.


technoweenie:
Yikes, using the regex is much slower. First, I ran this test on master:

```diff
diff --git a/filepathfilter/bench_test.go b/filepathfilter/bench_test.go
index 83899ecc..b52b6a1d 100644
--- a/filepathfilter/bench_test.go
+++ b/filepathfilter/bench_test.go
@@ -10,6 +10,16 @@ import (
        "github.com/git-lfs/git-lfs/tools"
 )

+func BenchmarkFilterNoWildcard(b *testing.B) {
+       files := benchmarkTree(b)
+       filter := filepathfilter.New([]string{"lfs"}, nil)
+       for i := 0; i < b.N; i++ {
+               for _, f := range files {
+                       filter.Allows(f)
+               }
+       }
+}
+
 func BenchmarkFilterIncludeWildcardOnly(b *testing.B) {
        files := benchmarkTree(b)
        filter := filepathfilter.New([]string{"*.go"}, nil)
```

```sh
$ go test ./filepathfilter -bench BenchmarkFilter
BenchmarkFilterNoWildcard-8              	    3000	    335559 ns/op
BenchmarkFilterIncludeWildcardOnly-8     	    2000	    881129 ns/op
BenchmarkFilterIncludeDoubleAsterisk-8   	     500	   3683718 ns/op
```

Here it is on this branch:

```sh
$ go test ./filepathfilter -bench BenchmarkFilter
BenchmarkFilterNoWildcard-8              	     200	   7115666 ns/op
BenchmarkFilterIncludeWildcardOnly-8     	    2000	    845644 ns/op
BenchmarkFilterIncludeDoubleAsterisk-8   	     300	   4115009 ns/op
```

Don't bother reviewing this yet, I'll try another approach.

Issue #2185, opened 2017-04-28
HTTP --> HTTPS redirects do not work


larsxschneider:
It looks like as if Git LFS does not handle redirects correctly. If I clone a Git repo via HTTP that is redirected to HTTP*S* then I observe the following error using `git lfs clone`:

```
$ git lfs clone http://github.com/larsxschneider/lfstest-individual-files.git bb
Cloning into 'bb'...
warning: redirecting to https://github.com/larsxschneider/lfstest-individual-files.git/
remote: Counting objects: 52, done.
remote: Total 52 (delta 0), reused 0 (delta 0), pack-reused 52
Unpacking objects: 100% (52/52), done.
Git LFS: (0 of 1 files) 0 B / 3.00 MB
batch response: Expected json type, got: ""
```

It also fails with `git clone`:
```
$ git clone http://github.com/larsxschneider/lfstest-individual-files.git aa
Cloning into 'aa'...
warning: redirecting to https://github.com/larsxschneider/lfstest-individual-files.git/
remote: Counting objects: 52, done.
remote: Total 52 (delta 0), reused 0 (delta 0), pack-reused 52
Unpacking objects: 100% (52/52), done.
Downloading folder/demo.file (3.00 MB)
Error downloading object: folder/demo.file (ae9d512afd1d78655599cd7c5c40d0f79fe9720ab9e02810d24fa34f093a2384)

Errors logged to /xyz/.git/lfs/objects/logs/20170428T105225.022645102.log
Use `git lfs logs last` to view the log.
error: external filter 'git-lfs filter-process' failed
fatal: folder/demo.file: smudge filter lfs failed
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry the checkout with 'git checkout -f HEAD'
```

Log file:
```
git-lfs/2.0.2 (GitHub; darwin amd64; go 1.8)
git version 2.12.2

$ git-lfs filter-process
Error downloading object: folder/demo.file (ae9d512afd1d78655599cd7c5c40d0f79fe9720ab9e02810d24fa34f093a2384)

Smudge error: Error downloading folder/demo.file (ae9d512afd1d78655599cd7c5c40d0f79fe9720ab9e02810d24fa34f093a2384): batch response: Expected json type, got: ""
github.com/git-lfs/git-lfs/errors.newWrappedError
[...]

PR #2235, opened 2017-05-18
lfsapi: support cross-scheme redirection

ttaylorr:
This pull request addresses https://github.com/git-lfs/git-lfs/issues/2185, and supports cross-scheme redirects in the `*lfsapi.Client` type.

The issue is that, even though Go's `*net/http.Client` supports redirects (in response to a 3xx status code) directly, we explicitly disable that as<sup>[[1](https://github.com/git-lfs/git-lfs/blob/v2.1.0/lfsapi/client.go#L251-L253)]</sup>:

```go
httpClient := &http.Client{
  Transport: tr,
  CheckRedirect: func(*http.Request, []*http.Request) error {
    return http.ErrUseLastResponse
  },
}
```

So we're forced to handle HTTP requests ourselves in the `lfsapi.Client.doWithRedirects` function. This is fine, since we already have logic to rewind the body and copy the headers over.

As an aside, ideally we could hook into the `net/http.Client.CheckRedirect` field to hand this responsibility over to the `net/http` package, but the HTTP client follows an older RFC (which it maintains for backwards compatibility reasons) that _prevents_ the client from copying the HTTP verb the request was sent with when a 301, 302, or 303 response code is given: (see: https://github.com/golang/go/blob/go1.8.1/src/net/http/client.go#L423-L429). We cannot work around this, so we must handle it ourselves.

This pull request expands the list of 3xx response codes that we will redirect from and increases the permissiveness of HTTP header copying to only check that the from->to host transition is the same, without checking the scheme.

Closes: https://github.com/git-lfs/git-lfs/issues/2185

---


lfsapi/client.go
```diff
@@ -275,7 +284,7 @@ func newRequestForRetry(req *http.Request, location string) (*http.Request, erro
 
 	for key := range req.Header {
 		if key == "Authorization" {
-			if req.URL.Scheme != newReq.URL.Scheme || req.URL.Host != newReq.URL.Host {
```
technoweenie: What do you think about preventing https => http redirects? That'd prevent the client from inadvertently sending creds over an insecure channel.

lfsapi/client.go
```diff
@@ -130,7 +130,16 @@ func (c *Client) doWithRedirects(cli *http.Client, req *http.Request, via []*htt
 
 	c.traceResponse(req, tracedReq, res)
 
-	if res.StatusCode != 307 {
+	if res.StatusCode != 301 &&
+		res.StatusCode != 302 &&
+		res.StatusCode != 303 &&
+		res.StatusCode != 307 &&
+		res.StatusCode != 308 {
```
technoweenie: This is kind of gross. I'd prefer a map or switch statement.

```go
if _, ok := redirectCodes[res.StatusCode]; ok {
  // ...
}
```

lfsapi/client.go
larsxschneider: That sounds sensible to me. I wonder if there could ever be a legitimate use case for a https ==> http redirect?!

lfsapi/client.go
ttaylorr: Thanks 👍. I don't have strong feelings either way and don't want to reach for a map/func until we need to share this check somewhere else.

lfsapi/client.go
ttaylorr: I added some code in 36944dc in order to prevent insecure redirects. If a redirect is sent from HTTPS down to HTTP, the following error will be given:

```
lfsapi/client: refusing insecure redirect, https->http
```

ttaylorr:
@technoweenie @larsxschneider thanks for the review! I added some tests in https://github.com/git-lfs/git-lfs/pull/2235/commits/38417f3d974fa18e297a718f72d95cd9c9e063ff to cover this new behavior, and I think this is ready to merge.
 
A note about the tests: I initially wanted to test same _and_ cross-host redirects, but I can't listen on the same serve in both HTTPs and HTTP in order to do a same-host secure upgrade, so the test coverage is as complete as possible.

Issue #2236, opened 2017-05-19
Pull command suppress download errors occurred during it's work and exits with zero code


seth2810:
Similar to #731 when `git lfs` can't download some of my files during it work, for example storage host died, `pull` command will return zero code rather than return an error code:

```bash
> git-lfs pull
Git LFS: (0 of 25063 files) 0 B / 621.71 MB                                                                                                                                         
batch request: ssh_exchange_identification: Connection closed by remote host: exit status 255
...
batch request: ssh_exchange_identification: Connection closed by remote host: exit status 255
> echo $?
0
```

In the same time `fetch` command catch this case:

```bash
> git lfs fetch
Fetching branch
Git LFS: (0 of 25063 files) 0 B / 621.71 MB                                                                                                                                         
batch request: exit status 255
...
batch request: exit status 255
Warning: errors occurred
> echo $?
2
```

https://github.com/git-lfs/git-lfs/blob/master/commands/command_fetch.go#L325-L330
https://github.com/git-lfs/git-lfs/blob/master/commands/command_fetch.go#L105-L107

https://github.com/git-lfs/git-lfs/blob/master/commands/command_pull.go#L107-L109

I would think it makes sense that `git lfs pull` should return non-zero in this case. If not, how can I detect that errors occurred and fix them/try again?

seth2810:
Maybe somebody can help with writing a test covering my case, now I do not know how to simulate a connection break in the test? Locally reproduced by redirecting traffic to a dummy host.

ttaylorr:
👋 @seth2810, thanks for opening this up! I see that you started work on this in https://github.com/git-lfs/git-lfs/pull/2237, so I'm going to redirect discussion about this PR to there, if you don't mind.

PR #2237, opened 2017-05-19
Make pull return non-zero error code when some downloads failed

seth2810:
https://github.com/git-lfs/git-lfs/issues/2236


ttaylorr:
@seth2810 this looks good. In terms of forcing an error to occur during command execution, you can instruct the Git server we run during tests to return an error by using a special "`contentHandler`".

- You can add the content handler [here](https://github.com/git-lfs/git-lfs/blob/master/test/cmd/lfstest-gitserver.go#L51-L63), and then
- Check for presence of the content handler [here](https://github.com/git-lfs/git-lfs/blob/master/test/cmd/lfstest-gitserver.go#L382-L391).

technoweenie:
@ttaylorr I'm tweaking the error message and fixing this up to get this out in today's release. 

@seth2810: Thanks for getting this started!

ttaylorr:
> @ttaylorr I'm tweaking the error message and fixing this up to get this out in today's release.

@technoweenie sounds great, thanks!

technoweenie:
@seth2810: your pull test had the right mindset, but it was using some undefined local variables: `$reponame`, `$b_oid`, and `$contents_oid`. I fixed the test by declaring them in the individual test, and fixing the `cd ../clone` line too. 

@ttaylorr's idea would've worked too, but I prefer tests that don't rely on special test server behavior if we can avoid it.

PR #2241, opened 2017-05-19
sanitize ssh options parsed from ssh:// url

technoweenie:
This fixes an issue where SSH options can be parsed out of `ssh://` urls. For example, the url `ssh://-oProxyCommand=gnome-calculator` will exec the following command:

```sh
$ ssh -p 12345 -oProxyCommand=gnome-calculator git-lfs-authenticate ...
```

SSH will run the command from the `-oProxyCommand` flag, which is not what we want.

This PR fixes it by inserting `--`, which disables SSH options after it.

```sh
$ ssh -p 12345 -- -oProxyCommand=gnome-calculator git-lfs-authenticate ...
```

LFS does support `plink` and `tortoiseplink`, which don't have anything like `--`. So instead, LFS will remove any leading `-`'s from the `user@host` components of the `ssh://` url.

### DIFF
--- a/lfsapi/ssh.go
+++ b/lfsapi/ssh.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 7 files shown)


larsxschneider:
I assume v1.5.6 is affected, too?
Is v1.5.6 considered end of life and therefore it will not receive any security patches?
I assume yes, I just want to clarify.

ttaylorr:
> I assume v1.5.6 is affected, too?

You are correct in your assumption, 1.5.6 is affected.

> Is v1.5.6 considered end of life and therefore it will not receive any security patches?

All v1.x.x releases are outside of the MAJOR semver indicator, so are considered EOL'd. I'm happy to cut a 2.0.x release containing this patch, if you'd like.

larsxschneider:
Thanks for the clarification! No v2.0.x patch necessary! 😄

Issue #2249, opened 2017-05-22
git lfs clone -v not work


linquize:
`git lfs clone -v` does not work

```
Error: unknown shorthand flag: 'v' in -v
```

From `git clone`, `-v` should be the same as `--verbose`

ttaylorr:
👋 @linquize, thanks for opening this issue! I went ahead and opened https://github.com/git-lfs/git-lfs/issues/2251 to address. It's in the `v2.2.0` milestone, so this will released as a part of that series.

In the meantime, here are some builds of 6d62dc9030a28c6a252963495c88861a18bb4770 that you can use: 

[git-lfs-darwin-386-2.2.0-pre.tar.gz](https://github.com/git-lfs/git-lfs/files/1019272/git-lfs-darwin-386-2.2.0-pre.tar.gz)
[git-lfs-darwin-amd64-2.2.0-pre.tar.gz](https://github.com/git-lfs/git-lfs/files/1019273/git-lfs-darwin-amd64-2.2.0-pre.tar.gz)
[git-lfs-freebsd-386-2.2.0-pre.tar.gz](https://github.com/git-lfs/git-lfs/files/1019275/git-lfs-freebsd-386-2.2.0-pre.tar.gz)
[git-lfs-freebsd-amd64-2.2.0-pre.tar.gz](https://github.com/git-lfs/git-lfs/files/1019277/git-lfs-freebsd-amd64-2.2.0-pre.tar.gz)
[git-lfs-linux-386-2.2.0-pre.tar.gz](https://github.com/git-lfs/git-lfs/files/1019276/git-lfs-linux-386-2.2.0-pre.tar.gz)
[git-lfs-linux-amd64-2.2.0-pre.tar.gz](https://github.com/git-lfs/git-lfs/files/1019278/git-lfs-linux-amd64-2.2.0-pre.tar.gz)
[git-lfs-windows-386-2.2.0-pre.zip](https://github.com/git-lfs/git-lfs/files/1019274/git-lfs-windows-386-2.2.0-pre.zip)
[git-lfs-windows-amd64-2.2.0-pre.zip](https://github.com/git-lfs/git-lfs/files/1019279/git-lfs-windows-amd64-2.2.0-pre.zip)

PR #2251, opened 2017-05-22
commands/clone: support shorthand --verbose '-v' flag

ttaylorr:
This pull request closes https://github.com/git-lfs/git-lfs/issues/2249 by adding the shorthand `-v` flag to be the equivelant of `--verbose` when calling `git lfs clone`.

This more closely matches the behavior of the builtin `git-clone(1)`:

> ```
> --verbose, -v
>    Run verbosely. Does not affect the reporting of progress status to the standard error stream.
> ```

While looking through the man pages for `git-clone(1)`, I double checked to make sure that none of the other wrapped clone flags that we provide were missing shorthand, single `-` flags.

Closes https://github.com/git-lfs/git-lfs/issues/2249.

---

/cc @git-lfs/core


linquize:
in `test-clone.sh`, it did not remove "$newclonedir", so the test failed

ttaylorr:
> in `test-clone.sh`, it did not remove "$newclonedir", so the test failed

Thanks, I just pushed up 080683b which should fix that 👍.

PR #2261, opened 2017-05-24
git,lfs: extract git.RevListScanner

ttaylorr:
This pull request extracts the internal `lfs.revListShas()` function into a new type, `*git.RevListScanner` capable of scanning revision ranges for objects and commits.

This work is required for the migrate command (proposed in #2146) in order to avoid implementing a topological sorting algorithm in Go that is functionally equivalent to a [`sort_in_topological_order` call](https://github.com/git/git/blob/v2.13.0/commit.c#L646-L758). Instead, we will opt to use the builtin `git-rev-list(1)` given a `--topo-order` flag and a left/right refspect.

As a first step, we extract that type from its original home in `lfs.revListShas()` so that it can be expanded in a follow-up pull request in order to support the use-case above.

---

/cc @git-lfs/core

### DIFF
--- a/git/rev_list_scanner.go
+++ b/git/rev_list_scanner.go
@@ -74,15 +74,6 @@ func (o *ScanRefsOptions) SetName(sha, name string) {
 	o.Names[sha] = name
 }
 
-// RevListObject is a tuple type returning results from the RevListScanner.
-type RevListObject struct {
-	// Name is an optional field that gives the name of the object (if the
-	// object is a tree, blob).
-	Name string
-	// Oid is the hex-decoded bytes of the object's ID.
-	Oid []byte
-}
-
 // RevListScanner is a Scanner type that parses through results of the `git
 // rev-list` command.
 type RevListScanner struct {
@@ -94,8 +85,10 @@ type RevListScanner struct {
 	// type.
 	closeFn func() error
 
-	// obj is the most recently read object.
-	obj *RevListObject
+	// name is the name of the most recently read object.
+	name string
+	// oid is the oid of the most recently read object.
+	oid []byte
 	// err is the most recently encountered error.
 	err error
 }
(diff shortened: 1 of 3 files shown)


git/rev_list_scanner.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// RevListObject is a tuple type returning results from the RevListScanner.
+type RevListObject struct {
+	// Name is an optional field that gives the name of the object (if the
+	// object is a tree, blob).
+	Name string
+	// Oid is the hex-decoded bytes of the object's ID.
+	Oid []byte
+}
```
technoweenie: I don't know if this type is necessary. What do you think about just adding `Name()` and `OID()` funcs on the `RevListScanner`?

lfs/gitscanner_refs.go
```diff
@@ -93,84 +91,51 @@ func scanRefsToChan(scanner *GitScanner, pointerCb GitScannerFoundPointer, refLe
...
+			rev := scanner.Object()
+			if rev == nil {
+				break
 			}
 
-			sha1 := line[0:40]
-			if len(line) > 40 {
-				opt.SetName(sha1, line[41:len(line)])
+			sha := hex.EncodeToString(rev.Oid)
+			tracerx.Printf(sha)
```
technoweenie: This should have more details, or be removed IMO.

ttaylorr:
@technoweenie thanks for the review! I pushed up some changes to address your comments, and I think this is ready to go. Mind taking another look?

lfs/gitscanner_refs.go
ttaylorr: Removed in f712771, thanks!

git/rev_list_scanner.go
ttaylorr: Good call. I got rid of that type and opted for the `Name()` and `OID()` funcs instead in https://github.com/git-lfs/git-lfs/pull/2261/commits/f7127717d60dd7d8bdc7da92c7e5fe7dce07b22e.

PR #2283, opened 2017-05-31
git/odb: read packed objects through *git.ObjectScanner

ttaylorr:
This pull request teaches the `*git/odb.ObjectDatabase` type to read "missing," or (more accurately) packed objects from the Git object database.

When an object becomes packed by Git (before a push, after a `git-pack-objects(1)`, or `git-gc(1)`) they are no longer accessible inside of `.git/objects/xx/yyyy`. Instead, they are delta compressed and stored inside of `.git/objects/pack/pack-nnn.pack` (and the corresponding `.git/objects/pack/pack-nnn.idx`).

Since the `git/odb` package doesn't currently know how to read packed objects, this pull request teaches it to fallback to the `*git.ObjectScanner` (which internally uses `git-cat-file(1) --batch` to "unpack" the objects before streaming their contents back to `git/odb`.

- 70818eb: Introduce an `open()` function which will be the home for reassembling the object from `git-cat-file` output.
- e8b24aa: Since `git-cat-file(1) --batch` yields objects without zlib compression applied, a new set of "uncompressed" `*git/odb.ObjectReader`'s are introduced to support reading uncompressed data. This is done to avoid recompressing the data just to immediately decompress it. 
- 4ac2ece: Prepare to test this new behavior by returning an error which satisfies `os.IsNotExist`.
- aaf8a45: Store and provide utilities to `Close()` a `*git.ObjectScanner` on the `*git/odb.ObjectDatabase`.
- 39d387a: Fallback to the `*git.ObjectScanner` when objects are missing by reassembling their object contents.

Some other thoughts to keep in mind after merging: Increasing the number of code paths for reading an object is intentional for two reasons.

I'm primarily concerned about the speed of reading large blobs through a pipe on Windows, which I think is kept faster by opening a file descriptor instead of a pipe.

The second is that I'd eventually like to teach the migrator how to parallelize history recreation. Right now, the cost of parallelization scales linearly with the number of parallel "workers" as a function of the cost of open file descriptors. If we moved the primary object reading path to go through `git-cat-file(1)`, we'd have to run multiple invocations of the command in order to read multiple objects at once. In addition to more pipe throughput (which is already slower), we also have to pay the cost as a linear function of the overhead caused by spawning a new process. 

---

/cc @git-lfs/core

### DIFF
--- a/git/odb/object_db.go
+++ b/git/odb/object_db.go
@@ -2,8 +2,12 @@ package odb
 
 import (
 	"bytes"
+	"encoding/hex"
+	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
+	"strings"
 	"sync/atomic"
 
 	"github.com/git-lfs/git-lfs/git"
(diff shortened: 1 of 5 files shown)


git/odb/object_db.go
```diff
@@ -2,23 +2,62 @@ package odb
...
+}
+
+// Close closes the *ObjectDatabase, freeing any open resources (namely: the
+// `*git.ObjectScanner instance), and returning any errors encountered in
+// closing them.
+//
+// If Close() has already been called, this function will panic().
+func (o *ObjectDatabase) Close() error {
+	if !atomic.CompareAndSwapUint32(&o.closed, 0, 1) {
+		panic("git/odb: *ObjectDatabase already closed")
```
technoweenie: Why a panic? Wouldn't returning an error be better?

ttaylorr: @technoweenie good call, I agree that an error is more appropriate. I found two places in this type that `panic()` and replaced both of them with errors: https://github.com/git-lfs/git-lfs/pull/2283/commits/f0dddc2301bdcde0807ae0c072298f454d835e65.

technoweenie:
Nice work :metal:

PR #2292, opened 2017-06-02
git/odb: introduce *HistoryRewriter

ttaylorr:
This pull request introduces a new type, `*git/odb.HistoryRewriter` that is capable of assembling a topologically equivalent DAG by remapping unique blobs through a `BlobRewriteFn`.

Essentially, the API that I'd like to provide, and the one needed by the migrator command (see: #2146 discussion) is:

```go
history := odb.NewHistoryRewriter(odb.FromFilesystem("..."))

tip, err := history.Rewrite(&odb.HistoryRewriteOptions{Left: "master",
        BlobFn: func(path string, b *odb.Blob) (*odb.Blob, error) {
                // ...
        },
})
```

Where:

- `Left`: is the range of commits to include
- `BlobFn`: is the map function to apply to each blob
- and `tip` is the new tip of that revision set.

To accomplish this, the `*HistoryRewriter` has three functions, `Rewrite`, `rewriteTree`, and `rewriteBlob`:

- [`rewriteBlob`][1]: at the lowest level, takes a given blob and applies the mapping function to it, propagating errors or saving the new blob to the object database.
- [`rewriteTree`][2]: at a higher level, enumerates each entry in a tree, calling `rewriteBlob` if the entry is a blob, or recursing if a subtree exists.
- [`Rewrite()`][3]: walks the commits (output from `*git.RevListScanner` via `git-rev-list(1)`, reassembling history by linking commit parents and moving the old trees to the new trees by first calling `rewriteTree()`, essentially performing:

[1]: https://github.com/git-lfs/git-lfs/pull/2292/files#diff-e438306fed436a78cbb097311f55f909R188
[2]: https://github.com/git-lfs/git-lfs/pull/2292/files#diff-e438306fed436a78cbb097311f55f909R149
[3]: https://github.com/git-lfs/git-lfs/pull/2292/files#diff-e438306fed436a78cbb097311f55f909R75

```go
for scanner.Scan() {
        commit, err := odb.Commit(scanner.OID())
        if err != nil {
                return err
        }

        commit.TreeID = rewriteTree(commit.TreeID)
        for i := 0; i < len(commit.Parents); i ++ {
                commit.Parents[i] = uncache(commit.Parents[i])
        }

        return odb.WriteCommit(commit)
}
```

To test this behavior, I first approached the problem by creating an in-memory "repository" with blobs and trees and commits that linked against one another. I found that while in theory this seemed like a good idea for immutability and speed concerns, in practice this required a lot of hard-to-read setup code. The `git/odb` package wasn't designed for this sort of level of repository stubbing.

Instead, I added four bare `.git` repositories that have the state that I'd like within them. When loaded using the `FixtureDatabase` helper, their contents are copied to `/var/tmp` where they can be mutated and asserted on without setup. This has a cost of reduced transparency, but ends up making the tests a lot easier to grok.

Here's a summary of what happened in this pull request:

- https://github.com/git-lfs/git-lfs/compare/0f2b8ea...4094ea6: introduce four fixture repositories.
- Scaffolding to be able to test the `*git/odb.HistoryRewriter`:
  - c91865f: teach `DatabaseFixture` helper, to create a `*git/odb.ObjectDatabase` against a copied version of a given fixture repository.
  - b1a5a1f: teach `HexDecode(t, sha)` to convert ASCII-encoded hexadecimal `string`s to `[]byte`s, or call `t.Fatal()`.
  - 5014390: teach `AssertBlobContents` to assert that a given tree entry has the expected blob contents.
  - 84a7d2b: teach `AssertCommitParent` to assert that a commit has a given parent.
  - 472b87c: teach `AssertCommitTree` to assert that a commit has a given tree.
- 748378b: introduce and test the `*git/odb.HistoryRewriter`.
- https://github.com/git-lfs/git-lfs/compare/748378b...4023333: cleanup unneeded parts of bare repository (`COMMIT_EDITMSG`, `info`, `hooks` examples, etc).

And some things that I'd like to think about going forward:

1. Teach the `*HistoryRewriter` how to use a `filepathfilter.Filter` to cull out subtrees that wont be rewritten.
2. Teach the `*HistoryRewriter` to accept a `TreeCallbackFn` which will visit a whole subtree once before writing it (e.g., to add an entry to `.gitattributes`).

---

/cc @git-lfs/core


git/odb/history_rewriter.go
```diff
@@ -0,0 +1,271 @@
...
+// new, modified blob. If it returns an error, the new blob will not be written
+// and instead the error will be returned from the Rewrite() function.
+//
+// Invocations of an instance of BlobRewriteFn are not expected to store the
+// returned blobs in the *git/odb.ObjectDatabase.
+//
+// The path argument is given to be an absolute path to the tree entry being
+// rewritten, where the repository root is the root of the path given. For
+// instance, a file "b.txt" in directory "dir" would be given as "dir/b.txt",
+// where as a file "a.txt" in the root would be given as "a.txt".
```
technoweenie: I think this is a future point to optimize. We shouldn't need to re-process identical blobs multiple times. It looks like there are a couple results of this callback:

1. File path matches `.gitattributes`, and the big file is converted to an LFS pointer.
2. File path does not match `.gitattributes`, but the git blob does exist somewhere else in a path matching `.gitattributes`.
3. No change needed, just return the blob.

I suppose the caller can just implement a callback that saves blob contents or something. Is that the intent? That could make sense given the various ways `HistoryRewriter` can be used.

ttaylorr: This makes sense to me. Since we the rewritten blobs are written to disk and their OIDs are kept in a cache, we can rely on that cache when the filter given to the `*git/odb.HistoryRewriter` matches the pathspec.

My next pull request on this topic is to add `filepathfilter.Filter` support to the `HistoryRewriter`, so I think this can be addressed then.

ttaylorr:
Thanks for the review 😄

git/odb/history_rewriter.go
technoweenie: Is there going to be an original blob OID mapping to the rewritten blob/lfs OIDs? How is the addition of `filepathfilter.Filter` going to affect the exported `HistoryRewriter` interface?

ttaylorr: > Is there going to be an original blob OID mapping to the rewritten blob/lfs OIDs?

We store that information [here](https://github.com/git-lfs/git-lfs/blob/3346d8720fc85bcabfc3b3d4731c29a018e5b0a1/git/odb/history_rewriter.go#L17-L20).

> How is the addition of `filepathfilter.Filter` going to affect the exported `HistoryRewriter` interface?

I was thinking of adding a new option func argument to the `NewHistoryRewriter` constructor, like so:

```go
filter := filepathfilter.New(...)

rewriter := odb.NewHistoryRewriter(db, odb.WithFilter(filter))
```

Where calling `rewriter.Rewrite()` with a `BlobRewriteFn` "fn", would only call "fn" for tree entries that whose path matches the given filter "filter".

Issue #2289, opened 2017-06-02
Breaking change between 2.0.2 <=> 2.1.0 with `git lfs status --porcelain`


asottile:
Here's a small setup test bed repository:

```
rm -rf repo
git init repo
cd repo
git lfs install
git lfs track a.bin b.bin
echo hi > a.bin
git add -- .
git commit -am foo
git mv a.bin b.bin
```

Under 2.0.2:

```console
$ git-lfs-2.0.2 status --porcelain
R  a.bin -> b.bin 3
```

Under 2.1.0:

```console
$ git-lfs-2.1.0 status --porcelain
R  a.bin -> b.bin
```

I'm not sure if this is intentional, it's certainly a breaking change though :) (and in a `--porcelain` output!)

This is breaking our tests in pre-commit/pre-commit-hooks: https://github.com/pre-commit/pre-commit-hooks/issues/195

I can try and bisect this if you'd like

asottile:
@ttaylorr This appears to be the culprit: #2042

asottile:
Now that I look at it, the change also breaks `--porcelain` output for not moves too (I just happened to notice / debug my test which exercises moves)

technoweenie:
Would it help if the status command had some low level format that your tests could depend on? Either `--json` (which you can check with `jq`), or even just `--format`, which is common with Git.

asottile:
At least in traditional git, that's what porcelain is supposed to be. json would be great though!

asottile:
(I also don't really care about the size field (I discard it), but I'm unable to consistently parse the two formats)

ttaylorr:
@asottile thanks for pointing this out. There was a breaking change to the `--porcelain` and normal output in https://github.com/git-lfs/git-lfs/pull/2042, and again in https://github.com/git-lfs/git-lfs/pull/2070.

If I remember correctly, we shipped these in 2.1.0 across a `MINOR` version boundary, but the change technically should have warranted a 3.0 release. I can add the information back to `--porcelain` and release those changes in 2.1.2 or 2.2.0, but I think it's possible in shell to parse either version correctly with `cut`.

asottile:
It would be, if files could not contain the space character :)

asottile:
I'm going to take a stab at implementing `--json`

ttaylorr:
> I'm going to take a stab at implementing `--json`

[...]

PR #2311, opened 2017-06-08
Implement status --json

asottile:
Admittedly, I'm a total golang noob :) (my only other real project being a [bridge to python](https://github.com/asottile/setuptools-golang))

This is for #2289, I'd love some feedback.

### DIFF
--- a/commands/command_status.go
+++ b/commands/command_status.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"crypto/sha256"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -14,7 +15,8 @@ import (
 )
 
 var (
-	porcelain = false
+	porcelain  = false
+	statusJson = false
 )
 
 func statusCommand(cmd *cobra.Command, args []string) {
@@ -30,6 +32,9 @@ func statusCommand(cmd *cobra.Command, args []string) {
 
 	if porcelain {
 		porcelainStagedPointers(scanIndexAt)
+		return
+	} else if statusJson {
+		jsonStagedPointers(scanIndexAt)
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


commands/command_status.go
```diff
@@ -14,7 +15,8 @@ import (
 )
 
 var (
-	porcelain = false
+	porcelain   = false
+	status_json = false
```
ttaylorr: Go prefers `camelCase` variables, so this could be renamed to `statusJson`.

commands/command_status.go
```diff
@@ -31,6 +33,9 @@ func statusCommand(cmd *cobra.Command, args []string) {
 	if porcelain {
 		porcelainStagedPointers(scanIndexAt)
 		return
+	} else if status_json {
```
ttaylorr: This `else` is redundant, since the above `if` branch terminates execution via a return. What do you think of:

```go
if porcelain {
        porcelainStagedPointers(scanIndexAt)
        return
}

if statusJson {
        jsonStagedPointers(scanIndexAt)
        return
}

// ...
```

commands/command_status.go
```diff
@@ -224,6 +229,39 @@ func statusScanRefRange(ref *git.Ref) {
 
 }
 
+type JsonStatus struct {
+	Files map[string]interface{}
```
ttaylorr: Do you think it would be worthwhile to make this more type-safe by adding a `JsonStatusEntry` type? For instance:

```go
type JSONStatusEntry struct {
        Status string `json:"status"`
        From string `json:"from"`
}

type JSONStatus struct {
        Files map[string]JSONStatusEntry `json:"files"`
}
```

You can also use the backticks `json:""` after a struct field to apply a custom name to it during formatting, for instance:

```go
type T struct {
        X string `json:"x"`
}
```

would be encoded as `{"x":"..."}` instead of `{"X":"..."}`.

commands/command_status.go
asottile: ah indeed, I got this mostly right elsewhere :)

commands/command_status.go
asottile: yeah that's fair, I kept it as `if ... else if` so it's more obvious that they are disparate branches at a glance (without having to read the code inside the blocks)

commands/command_status.go
asottile: That was my initial thought, but I wanted to avoid `"from": ""`

asottile: seems I want 
```
`json:"from,omitempty"`
```
!!!

ttaylorr: Ah, `,omitempty` is what you're looking for 👍.

commands/command_status.go
```diff
@@ -224,6 +229,44 @@ func statusScanRefRange(ref *git.Ref) {
...
+
+	status := JSONStatus{Files: make(map[string]JSONStatusEntry)}
+
+	for _, entry := range append(unstaged, staged...) {
+		switch entry.Status {
+		case lfs.StatusRename, lfs.StatusCopy:
+			status.Files[entry.DstName] = JSONStatusEntry{
+				Status: string(entry.Status), From: entry.SrcName,
+			}
+			break
```
ttaylorr: `break` is not required in Go to prevent from falling through into the `default` branch. If, however, you want to skip out of the outer `for` loop on L249, you'll need to label it and `break <label>`, i.e.,

```go
L:
        for _, entry := range append(unstaged, staged...) {
                switch entry.Status {
                case lfs.StatusRename, lfs.StatusCopy:
                        // ...
                        break L // <- will move to "HERE" below.
        }

        // HERE
```

asottile: sweet, fixed :)

PR #2312, opened 2017-06-09
Retry on timeout or temporary errors

jakub-m:
git-lfs could fail randomly on "i/o timeout" error. This is especially
annoying with very long running transfers. This change recognises
Timeout or Temporary errors and retries them.

---

Tested by doing git lfs smudge on a file in lfs repository, where `lfs.url=http://localhost:9000` and  `nc -v -l 9000` was listening locally.

Without the fix git-lfs client aborted instantly, with the fix it retried.

### DIFF
--- a/tq/transfer_queue.go
+++ b/tq/transfer_queue.go
@@ -1,7 +1,6 @@
 package tq
 
 import (
-	"net/url"
 	"os"
 	"sort"
 	"sync"
@@ -634,12 +633,7 @@ func (q *TransferQueue) run() {
 
 // canRetry returns whether or not the given error "err" is retriable.
 func (q *TransferQueue) canRetry(err error) bool {
-	switch cause := errors.Cause(err).(type) {
-	case *url.Error:
-		return cause.Temporary() || cause.Timeout()
-	default:
-		return errors.IsRetriableError(err)
-	}
+	return errors.IsRetriableError(err)
 }
 
 // canRetryObject returns whether the given error is retriable for the object
--- a/tq/transfer_queue.go
+++ b/tq/transfer_queue.go
@@ -1,6 +1,7 @@
 package tq
 
 import (
+	"net/url"
 	"os"
 	"sort"
 	"sync"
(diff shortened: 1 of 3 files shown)


tq/transfer_queue.go
```diff
@@ -633,7 +634,12 @@ func (q *TransferQueue) run() {
 
 // canRetry returns whether or not the given error "err" is retriable.
 func (q *TransferQueue) canRetry(err error) bool {
-	return errors.IsRetriableError(err)
+	switch cause := errors.Cause(err).(type) {
```
ttaylorr: Rad ⚡️. I know that both the `switch (...).(type)` is a legit way to do this, but I think the convention in the majority of LFS looks something like:

```go
func (q *TransferQueue) canRetry(err error) bool {
        err = errors.Cause(err)

        if cause, ok := err.(*url.Error); ok {
                return cause.Temporary() || cause.Timeout()
        }
        return errors.IsRetriableError(err)
}
```

technoweenie: What would you think about just putting that code in [`errors.IsRetriableError`](https://github.com/git-lfs/git-lfs/blob/b7b1f2b7985c8f7a18c563c70898f669ffb8c13c/errors/types.go#L129-L141) so that all callers recognize `*url.Error`?

ttaylorr: Ah, I dig that a bunch. @jakub-m what do you think?

jakub-m: @ttaylorr sure will rewrite it and write the tests.

jakub-m: @technoweenie Initially I had it in `errors.IsRetriableError` but then I moved it to `canRetry`, because it seems that `errors.IsRetriableError`, together with the other functions `types.go`, are about type inference, and not about flow control. I don't know the codebase and the conventions there well so I am ok to move it there if you are sure this is the place.

technoweenie: That's a good point. Though, I don't think the `errors` package strictly has to be about type inference. The `Temporary()` and `Timeout()` methods didn't exist in Go, otherwise we would've used them. I'd prefer to move this to `errors.IsRetriableError()`, but don't have strong feelings. @ttaylorr any thoughts?

ttaylorr: > I'd prefer to move this to `errors.IsRetriableError()`, but don't have strong feelings. @ttaylorr any thoughts?

I agree; I think this is most well-suited in `errors.IsRetriableError()`.

jakub-m: Added tests, moved the logic to `errors.isRetriableError()`

PR #2313, opened 2017-06-10
commands: add git-lfs-migrate(1) 'info' subcommand

ttaylorr:
This pull request adds documentation and an implementation for the 'info' subcommand of the `git-lfs-migrate(1)` suite.

The 'info' subcommand is designed to gather statistics regarding the size of data stored in Git per filepath extension. The output of the command looks as follows:

```
~/g/git-lfs (migrate-subcommand-info) $ git-lfs migrate info \
  --include-ref=refs/heads/migrate-subcommand-info \
  --exclude-ref=refs/heads/master \
  --top=3
*.go 2.0 MB
*.sh 325.3 KB
*.md 225.6 KB
``` 

The following arguments are permitted:

- `--include-ref`: a set of references of which commits reachable by those refs are included in the traversal.
- `--exclude-ref`: a set of references of which commits reachable by those refs are excluded in the traversal.
- `--include`: a filepath filter include set
- `--exclude`: a filepath filter exclude set
- `<branch ...>`: an optional set of branches to include, defaulting to the currently checked out branch, or those references given in `--{include,exclude}-ref`.

Here's a summary of the changes:

- fd837f4...11b8d1f: document and specify command arguments in the man pages
- b47255d: add scaffolding for subcommands to `git-lfs-migrate(1)`.
- 58fda2d...1a67cec: helper functions necessary to complete a migration.
- d2c6b0c: teach `tools.ClampInt`.
- 1612c44: add test fixtures for the `info` subcommand
- af3871b: implement and test the `info` subcommand

---

This is the first command to be implemented as described in the `git-lfs-migrate(1)` proposal issue: (see discussion: #2146).

---

/cc @git-lfs/core

### DIFF
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -20,6 +20,27 @@ var (
 	migrateExcludeRefs []string
 )
 
+// migrate takes the given command and arguments, as well as a BlobRewriteFn to
+// apply, and performs a migration.
+func migrate(cmd *cobra.Command, args []string, fn githistory.BlobRewriteFn) {
+	requireInRepo()
+
+	db, err := getObjectDatabase()
+	if err != nil {
+		ExitWithError(err)
+	}
+
+	opts, err := rewriteOptions(args, fn)
+	if err != nil {
+		ExitWithError(err)
+	}
+
+	_, err = getHistoryRewriter(cmd, db).Rewrite(opts)
+	if err != nil {
+		ExitWithError(err)
+	}
+}
+
 // getObjectDatabase creates a *git.ObjectDatabase from the filesystem pointed
 // at the .git directory of the currently checked-out repository.
 func getObjectDatabase() (*odb.ObjectDatabase, error) {
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -175,6 +175,15 @@ func currentRefToMigrate() (*git.Ref, error) {
 	return current, nil
 }
 
+// getHistoryRewriter returns a history rewriter that includes the filepath
+// filter given by the --include and --exclude arguments.
+func getHistoryRewriter(cmd *cobra.Command, db *odb.ObjectDatabase) *githistory.Rewriter {
+	include, exclude := getIncludeExcludeArgs(cmd)
+	filter := buildFilepathFilter(cfg, include, exclude)
+
+	return githistory.NewRewriter(db, githistory.WithFilter(filter))
+}
+
 func init() {
 	RegisterCommand("migrate", nil, func(cmd *cobra.Command) {
 		// Adding flags directly to cmd.Flags() doesn't apply those
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"path/filepath"
+	"strings"
 
 	"github.com/git-lfs/git-lfs/errors"
 	"github.com/git-lfs/git-lfs/git"
@@ -26,6 +27,23 @@ func getObjectDatabase() (*odb.ObjectDatabase, error) {
 		return nil, errors.Wrap(err, "cannot open root")
 	}
 	return odb.FromFilesystem(filepath.Join(dir, "objects"))
+}
+
+// formatRefName returns the fully-qualified name for the given Git reference
+// "ref".
+func formatRefName(ref *git.Ref, remote string) string {
+	var name []string
+
+	switch ref.Type {
+	case git.RefTypeRemoteBranch:
+		name = []string{"refs", "remotes", remote, ref.Name}
+	case git.RefTypeRemoteTag:
+		name = []string{"refs", "tags", ref.Name}
+	default:
+		return ref.Name
+	}
+	return strings.Join(name, "/")
+
 }
 
 // currentRefToMigrate returns the fully-qualified name of the currently
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -29,6 +29,64 @@ func getObjectDatabase() (*odb.ObjectDatabase, error) {
 	return odb.FromFilesystem(filepath.Join(dir, "objects"))
 }
 
+// includeExcludeRefs returns fully-qualified sets of references to include, and
+// exclude, or an error if those could not be determined.
+//
+// They are determined based on the following rules:
+//
+//   - Include all local refs/heads/<branch> references for each branch
+//     specified as an argument.
+//   - Include the currently checked out branch if no branches are given as
+//     arguments and the --include-ref= or --exclude-ref= flag(s) aren't given.
+//   - Include all references given in --include-ref=<ref>.
+//   - Exclude all references given in --exclude-ref=<ref>.
+func includeExcludeRefs(args []string) (include, exclude []string, err error) {
+	hardcore := len(migrateIncludeRefs) > 0 || len(migrateExcludeRefs) > 0
+
+	if len(args) == 0 && !hardcore {
+		// If no branches were given explicitly AND neither
+		// --include-ref or --exclude-ref flags were given, then add the
+		// currently checked out reference.
+		current, err := currentRefToMigrate()
+		if err != nil {
+			return nil, nil, err
+		}
+		args = append(args, current.Name)
+	}
+
+	for _, name := range args {
+		// Then, loop through each branch given, resolve that reference,
+		// and include it.
+		ref, err := git.ResolveRef(name)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		include = append(include, ref.Name)
+	}
+
+	if hardcore {
+		// If either --include-ref=<ref> or --exclude-ref=<ref> were
+		// given, append those to the include and excluded reference
+		// set, respectively.
+		include = append(include, migrateIncludeRefs...)
+		exclude = append(exclude, migrateExcludeRefs...)
+	} else {
+		// Otherwise, if neither --include-ref=<ref> or
+		// --exclude-ref=<ref> were given, include no additional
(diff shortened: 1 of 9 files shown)


commands/command_migrate.go
```diff
@@ -0,0 +1,230 @@
...
+		// Adding flags directly to cmd.Flags() doesn't apply those
+		// flags to any subcommands of the root. Therefore, loop through
+		// each subcommand specifically, and include common arguments to
+		// each.
+		//
+		// Once done, link each orphaned command to the
+		// `git-lfs-migrate(1)` command as a subcommand (child).
+
+		for _, subcommand := range []*cobra.Command{
+			MigrateInfoCommand,
```
technoweenie: I don't think it makes sense to declare this `MigrateInfoCommand` var in `commands/command_migrate_info.go` if some of the flags are added here. They should all be set up in _this_ `init()` func. What do you think?

docs/man/git-lfs-migrate.1.ronn
```diff
@@ -0,0 +1,72 @@
...
+=======================================================
+
+## SYNOPSIS
+
+`git lfs migrate` <mode> [options] [--] [branch ...]
+
+## MODES
+
+* `info`
+    Perform no migration, instead show information about repository size.
```
technoweenie: I think you trim the first part of this sentence. Talk about what it _does_.

docs/man/git-lfs-migrate.1.ronn
```diff
@@ -0,0 +1,72 @@
...
+
+* `--include-ref`=<refname>:
+    See [INCLUDE AND EXCLUDE (REFS)].
+
+* `--exclude-ref`=<refname>:
+    See [INCLUDE AND EXCLUDE (REFS)].
+
+* [branch ...]:
+    Migrate only the set of branches listed. If not given, `git-lfs-migrate(1)`
+    will the currently checked out branch.
```
technoweenie: Looks like you're missing a word here :)

commands/command_migrate.go
ttaylorr: I think that's a great idea; it should make readers more easily aware of the various `migrate` subcommands. I made this change via: 1f443d0.

docs/man/git-lfs-migrate.1.ronn
ttaylorr: Thanks for the suggestion, I much prefer your wording: d453d9d5.

docs/man/git-lfs-migrate.1.ronn
ttaylorr: 😅: 471fc9e.

ttaylorr:
@technoweenie good call re: surfacing more relevant information. Here's an updated format I implemented in https://github.com/git-lfs/git-lfs/pull/2313/commits/f3e0af6b5d3723c71107cc049f064b59d6e5541a. What do you think?

```
~/g/git-lfs (migrate-subcommand-info) $ ./bin/git-lfs migrate info \
  --top=5 --above=10kb \
  --include-ref=refs/heads/migrate-subcommand-info
  --exclude-ref=refs/heads/master
Files above 10 KB:
*.go    906 KB, 48/479 files(s)  10%
*.bmp      154 KB, 1/2 files(s)  50%
*.sh      149 KB, 8/75 files(s)  11%
*.md      114 KB, 7/45 files(s)  16%
*.ico       34 KB, 1/1 files(s) 100%
```

commands/command_migrate_info.go
```diff
@@ -0,0 +1,161 @@
...
+	files = tools.Rjust(files)
+
+	percentages := make([]string, 0, len(e))
+	for _, entry := range e {
+		percentAbove := 100 * (float64(entry.TotalAbove) / float64(entry.Total))
+
+		percentage := fmt.Sprintf("%.0f%%", percentAbove)
+
+		percentages = append(percentages, percentage)
+	}
```
technoweenie: Two minor things about this: 

First, it's given an `io.Writer`, but the output is buffered. It's not exactly a lot of data, but I'm just used to writing it out with `fmt.Fprintf` if the output doesn't need post-processing. However, having to deal with just 1 `err` from `fmt.Fprintln` is kinda nice.

Second: You could reduce how often `e` is looped by building `extensions`, `files`, and `percentages` at the same time. At a high level, the structure of this func would be something like:

1. Init `extensions`, `files`, `percentages`
2. Loop through `e`, appending to those slices
3. `Rjust` or `Ljust` them
4. Write header
5. Write each line with the justified `extensions`, `files`, `percentages`

I think structurally it makes more sense, but 36179dee6c1cc9eb2d1206d54360d39f306ba3e4 saves a whopping 3 LOC (7 if I didn't add comments).

commands/command_migrate_info.go
```diff
@@ -0,0 +1,161 @@
...
+			entry.Total++
+			entry.BytesTotal += b.Size
+
+			if b.Size > int64(migrateInfoAbove) {
+				entry.TotalAbove++
+				entry.BytesAbove += b.Size
+			}
+
+			// TODO(@ttaylorr): needed?
+			exts[ext] = entry
```
ttaylorr: I think either works, but I dig the explicitness of `exts[ext] = entry`. Removed the `TODO` comment in https://github.com/git-lfs/git-lfs/pull/2313/commits/6b94d8ba9c12bb7dfe99ebe85ca1285b14573325.

technoweenie:
I think this output looks great, but the file sizes should be aligned too:

```
Files above 10 KB:
*.go    906 KB, 48/479 files(s)  10%
*.bmp      154 KB, 1/2 files(s)  50%
*.sh      149 KB, 8/75 files(s)  11%
*.md      114 KB, 7/45 files(s)  16%
*.ico       34 KB, 1/1 files(s) 100%
```

Aligned, it'd look something like:

```
*.go 	906 KB	48/480 files(s)	 10%
*.bmp	154 KB	   1/2 files(s)	 50%
*.sh 	149 KB	  8/75 files(s)	 11%
*.md 	114 KB	  7/45 files(s)	 16%
*.ico	34 KB 	   1/1 files(s)	100%
```

In chat, you questioned out loud about how to parse this output with scripts:

* I know I suggested moving it to a header, but maybe `Files above 10 KB:` is not necessary? I don't think I need the command to repeat the `--above` flag to me.
* While humanized sizes are definitely the way to go, IMO, what about introducing a flag to specify the unit?  `--size-unit=[ bytes | M | etc ]`.

commands/command_migrate_info.go
```diff
@@ -0,0 +1,161 @@
...
+	// 'info' which specifies how many info entries to show by default.
+	migrateInfoTopN int
+
+	// migrateInfoAboveFmt is a flag given to the git-lfs-migrate(1)
+	// subcommand 'info' specifying a human-readable string threshold of
+	// filesize before entries are counted.
+	migrateInfoAboveFmt string
+	// migrateInfoAbove is the number of bytes parsed from the above
+	// migrateInfoAboveFmt flag.
+	migrateInfoAbove uint64
```
technoweenie: I think `migrateInfoAbove` should have a sensible default, like 1MB.

commands/command_migrate_info.go
```diff
@@ -0,0 +1,161 @@
...
+
+	migrateInfoTopN = tools.ClampInt(migrateInfoTopN, len(entries), 0)
+
+	entries = entries[:tools.MaxInt(0, migrateInfoTopN)]
+
+	entries.Print(os.Stderr)
+}
+
+// MigrateInfoEntry represents a tuple of filetype to total size taken by that
+// file type.
```
ttaylorr: Nice catch! Updated those docs and added some godoc to the fields on this type in 7c77064.

commands/command_migrate_info.go
ttaylorr: > However, having to deal with just 1 `err` from `fmt.Fprintln` is kinda nice.

Totally agree.

> Second: You could reduce how often e is looped by building `extensions`, `files`, and `percentages` at the same time.

Thank you! This is such an awesome suggestion. I changed this in 0dbc358 and it made the code much cleaner.

commands/command_migrate_info.go
ttaylorr: > I think this output looks great, but the file sizes should be aligned too:

Good call: I built up a `size` array as well, and aligned it in 7d2dbd6. I think it looks much better this way!

ttaylorr:
> I know I suggested moving it to a header, but maybe `Files above 10 KB:` is not necessary? I don't think I need the command to repeat the `--above` flag to me.

I agree, and I really dig this reasoning. I think it a) makes the output easier to look at, b) doesn't repeat already known information, and c) makes the output easier to parse with Unix-y tools. Removed the header in ae76e72.

> * While humanized sizes are definitely the way to go, IMO, what about introducing a flag to specify the unit? `--size-unit=[ bytes | M | etc ]`.

Totally possible, and I think this would make sense for a lot of command-line tools that want to parse the output.

ttaylorr:
> I think `migrateInfoAbove` should have a sensible default, like 1MB.

Good call, I made the default 1MB in 97b0f36. This can be easily changed in the future if 1MB is too high/low.

> > * While humanized sizes are definitely the way to go, IMO, what about introducing a flag to specify the unit? `--size-unit=[ bytes | M | etc ]`.
> 
> Totally possible, and I think this would make sense for a lot of command-line tools that want to parse the output.

Started looking into this, but I think this may be best done in a separate PR after this one, since it will involve changes in both the humanize package and the `info` command.

ttaylorr:
@technoweenie so long as I follow up on the `--size-unit` flag in a separate PR, I think this is ready to go.

ttaylorr:
Thanks for the review!

PR #2324, opened 2017-06-13
tools: implement and use 'humanize' package

ttaylorr:
This pull request introduces a new package: `github.com/git-lfs/git-lfs/tools/humanize` designed to format and parse byte amounts as human readable text.

This will be used in #2313 in order to parse the `--above=<size>` flag, as well as the output of the command (ref: #2146).

---

/cc @git-lfs/core


technoweenie:
Why isn't it using github.com/dustin/go-humanize instead?

ttaylorr:
`github.com/dustin/go-humanize` offers more functions than we need. If you'd rather vendor it I don't mind either way.

technoweenie:
It's probably fine, math isn't changing any time soon :)

PR #2327, opened 2017-06-14
commands: typo fix for fetch errors messages

gurugray:
the are some strange messages when we fetching objects (not pushing they).
fixed this misunderstanding


gurugray:
thanks!

PR #2329, opened 2017-06-15
commands: add logging to git-lfs-migrate-info(1)

ttaylorr:
This pull request introduces a new package, `github.com/git-lfs/git-lfs/git/githistory/log` for Git-style progress based logging, like this:

This pull request is based off the `migrate-subcommand-info` branch, since I wanted to use the `info(1)` subcommand to test out the logger in a real-world setting. Once #2313 is merged, I'll change the base of this pull request to `master`.

The API works as follows: given a `*log.Logger` with a `sink io.Writer`, ask for one of two types of logging tasks:

- `*WaitingTask` (via `(*log.Logger) Waiter()`): a task for which the upper bound of items to process is not known. Logged as `message: ...` and `message: ..., done`. 
- `*PercentageTask` (via `(*log.Logger) Percentage()`): a task for which the upper bound of items is known. Logged as `message:   1% (1/100)` and `message: 100% (100/100), done`.

When a new task is created (via either `Waiter()` or `Percentage()`) that task is queued by the logger and becomes the primary task being logged. Only one task may be logged at a time, though new tasks can be created before the existing task is completed. This, together, creates the two axioms of `*log.Logger`:

1. Only one task may be logged at once, and that task is logged until completion.
2. "Infinite" tasks may be created before the current task is finished logging, creating a pseudo-infinite channel buffer.

The finer details of how this works are described in the `consume()` logger function.

One other note: this was originally designed as a utility for the `githistory` package, but I think that it could have other uses throughout LFS. I'd like to eventually consider what making this a top-level `github.com/git-lfs/git-lfs/log` package could look like.

---

/cc @git-lfs/core 
/ref #2146, #2313

### DIFF
--- a/git/githistory/log/log.go
+++ b/git/githistory/log/log.go
@@ -114,9 +114,10 @@ func (l *Logger) consume() {
 		}
 	}()
 
+	defer close(l.tasks)
+
 	pending := make([]Task, 0)
 
-L:
 	for {
 		// If there is a pending task, "peek" it off of the set of
 		// pending tasks.
@@ -132,7 +133,7 @@ L:
 			if !ok {
 				// If the queue is closed, no more new tasks may
 				// be added.
-				break L
+				return
 			}
 
 			// Otherwise, add a new task to the set of tasks to
@@ -148,7 +149,7 @@ L:
 				if !ok {
 					// If the queue is closed, no more tasks
 					// may be added.
-					break L
+					return
 				}
 				// Otherwise, add the next task to the set of
 				// pending, active tasks.
@@ -160,8 +161,6 @@ L:
 			}
 		}
 	}
-
-	close(l.tasks)
 }
 
 // logTask logs the set of updates from a given task to the sink, then logs a
--- a/git/githistory/log/log.go
+++ b/git/githistory/log/log.go
@@ -6,9 +6,14 @@ import (
 	"io/ioutil"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/git-lfs/git-lfs/tools"
 	"github.com/olekukonko/ts"
+)
+
+const (
+	DefaultLoggingThrottle = 200 * time.Millisecond
 )
 
 // Logger logs a series of tasks to an io.Writer, processing each task in order
@@ -20,6 +25,10 @@ type Logger struct {
 	// widthFn is a function that returns the width of the terminal that
 	// this logger is running within.
 	widthFn func() int
+
+	// throttle is the minimum amount of time that must pass between each
+	// instant data is logged.
+	throttle time.Duration
 
 	// queue is the incoming, unbuffered queue of tasks to enqueue.
 	queue chan Task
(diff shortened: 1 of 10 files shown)


git/githistory/log/log.go
```diff
@@ -0,0 +1,197 @@
...
+	sink io.Writer
+
+	// widthFn is a function that returns the width of the terminal that
+	// this logger is running within.
+	widthFn func() int
+
+	// queue is the incoming, unbuffered queue of tasks to enqueue.
+	queue chan Task
+	// tasks is the set of tasks to process.
+	tasks chan Task
```
technoweenie: If you skip `queue` and just use `tasks`, it should remove a lot of of complexity from `consume()`. The comment sounds like you intended for `tasks` to be buffered, but it's not.

git/githistory/log/log.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// logTask logs the set of updates from a given task to the sink, then logs a
+// "done" message, and then marks the task as done.
+func (l *Logger) logTask(task Task) {
+	defer l.wg.Done()
+
+	var last string
+	for last = range task.Updates() {
+		l.logLine(last)
```
technoweenie: It may be worth watching your CPU on an intense write load. The existing spinner that `fetch` updates the terminal no more frequently than every 200ms. It's plenty time to represent progress to humans, and doesn't write to the terminal nearly as frequently.

git/githistory/log/log.go
ttaylorr: Thanks for pointing this out. I added the `queue` and `tasks` channel so that new tasks could be enqueued before the current task is finished processing.

Looking back through this, I realized that I removed the callsite in the code that originally necessitated this, but I think if we want to use this package as a general purpose logging utility for LFS, it's worthwhile to keep the extra complexity.

I added a test to ensure that this non-blocking behavior is in place. Before 930c151:

```
~/g/git-lfs (migrate-logger) $ go test -v ./git/githistory/log -run TestLoggerLogsMultipleTasksWithoutBlocking
=== RUN   TestLoggerLogsMultipleTasksWithoutBlocking
# ...
```

and with 930c151:

```
~/g/git-lfs (migrate-logger) $ go test -v ./git/githistory/log -run TestLoggerLogsMultipleTasksWithoutBlocking
=== RUN   TestLoggerLogsMultipleTasksWithoutBlocking
--- PASS: TestLoggerLogsMultipleTasksWithoutBlocking (0.00s)
PASS
ok      github.com/git-lfs/git-lfs/git/githistory/log   0.013s
```

git/githistory/log/log.go
ttaylorr: Good call. I tried to implement this with channels, but the code ended up being simpler with `time` arithmetic. As of fc1faca, you can specify a throttle which is the minimum amount of time that must pass between logs. Additional log updates will be dropped on the ground.

Even if updates aren't logged, however, the logger will still remember them, so if a `PercentageTask` only gets to 95% within the throttle window, it will always finish as: `msg: 100% ..., done`.

PR #2333, opened 2017-06-15
commands: teach --unit to 'info' subcommand of 'git-lfs-migrate(1)'

ttaylorr:
This pull request teaches the `--unit` flag to the 'info' subcommand of `git-lfs-migrate(1)`.

Previously, the behavior of the command was to find the largest storage unit that could represent the number of bytes counted as a whole number. Now, if instead the `--unit` flag is given, that unit is used to format the bytes with at most one decimal place.

Here's a summary of what happened:

- 06fde60 & 40e833d: implement and use `ParseByteUnit` so that we can parse the string `--unit` given.
- cb5baef & 8d09c7e: implement and use `ForamtBytesUnit` so that we can format storage quantities with arbitrary units.
- d1037fc: use the above two utilities to format the bytes accordingly.

Here's a sample from the manpage:

> ```
> * `--unit=<unit>`
>     Format the number of bytes in each entry as a quantity of the storage unit
>     provided. Valid units include:
>       * b, kib, mib, gib, tib, pib - for IEC storage units
>       * b, kb, mb, gb, tb, pb - for SI storage units
> 
>     If a --unit is not specified, the largest unit that can fit the number of
>     counted bytes as a whole number quantity is chosen.
> ```

---

/cc @git-lfs/core 
/cc #2146

### DIFF
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -209,6 +209,7 @@ func init() {
 	info := NewCommand("info", migrateInfoCommand)
 	info.Flags().IntVar(&migrateInfoTopN, "top", 5, "--top=<n>")
 	info.Flags().StringVar(&migrateInfoAboveFmt, "above", "1mb", "--above=<n>")
+	info.Flags().StringVar(&migrateInfoUnitFmt, "unit", "", "--unit=<unit>")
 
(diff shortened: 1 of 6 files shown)


marlazinger200:
1

PR #2334, opened 2017-06-16
git/githistory/log: teach Durable() to Task

ttaylorr:
This pull request teaches a new `Task` type to the `github.com/git-lfs/git-lfs/git/githistory/log` package: `DurableTask`.

A `DurableTask` ensures that all log `<-Update()` reads are logged to the `sink io.Writer`, regardless of throttling conditions. This required for multi-line log updates (like output from `git-ref-update(1)`) where all log lines are important.

---

/cc @git-lfs/core 
/refs #2146 #2329

### DIFF
--- a/git/githistory/log/log.go
+++ b/git/githistory/log/log.go
@@ -185,7 +185,7 @@ func (l *Logger) consume() {
 func (l *Logger) logTask(task Task) {
 	defer l.wg.Done()
 
-	logAll := task.Durable()
+	logAll := !task.Throttled()
 	var last time.Time
 
 	var msg string
--- a/git/githistory/log/log.go
+++ b/git/githistory/log/log.go
@@ -175,8 +175,20 @@ func (l *Logger) consume() {
 
 // logTask logs the set of updates from a given task to the sink, then logs a
 // "done" message, and then marks the task as done.
+//
+// By default, the *Logger throttles log entry updates to once per the duration
+// of time specified by `l.throttle time.Duration`.
+//
+// If the duration if 0, or the task is "durable" (by implementing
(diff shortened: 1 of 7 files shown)


technoweenie:
Why is this needed? If the tasks need to be durable, then they probably don't need to be throttled in the first place. Do only some of the tasks need this?

ttaylorr:
> Why is this needed? If the tasks need to be durable, then they probably don't need to be throttled in the first place. Do only some of the tasks need this?

Good question 😄 Only some of the tasks need this. For instance, the `*PercentageTask` writes updates every time `Count()` is called, and usually gets throttled. The `*ListTask` type (see: https://github.com/git-lfs/git-lfs/pull/2335) yields updates at an unknown rate, but it's important that all updates are logged. In other words, `ListTask` satisfies `DurableTask`, but `*PercentageTask` does not,.

technoweenie:
Ah, that makes sense after looking at that PR. I would just add `Durable()` to the base Task interface, since there are only a few task types. What do you think about `Throttled()` for a name instead of `Durable()`?

ttaylorr:
> Ah, that makes sense after looking at that PR.

Thanks for taking a look, and sorry for not explaining things more thoroughly in this PR.

> I would just add `Durable()` to the base Task interface, since there are only a few task types.

Good call, I like the explicitness of being able to declare that a task is `Durable()` or not per task. I added that in a7a99e5, 8c64a43 & 5223586.

> What do you think about `Throttled()` for a name instead of `Durable()`?

I think I prefer the `Durable()` name, since `Throttled()` feels like a negated conditional.

technoweenie:
I just don't think `Durable` accurately describes what is happening. Maybe I'm just used to seeing that word in the context of a data store.

PR #2335, opened 2017-06-16
git/githistory/log: implement ListTask type

ttaylorr:
This pull request provides a new `Task` implementation, `ListTask`, which is used to log line-delimited lists of data using the `git/githistory/log.Logger` type.

This is required work for the `import` subcommand of `git-lfs-migrate(1)` which will print out one line for each ref updated by the migrator.

--- 

/cc @git-lfs/core 
/refs #2146 #2329 #2334

### DIFF
--- a/git/githistory/log/list_task.go
+++ b/git/githistory/log/list_task.go
@@ -32,9 +32,9 @@ func (l *ListTask) Complete() {
 	close(l.ch)
 }
 
-// Durable implements the Task.Durable function and ensures that all log updates
-// are printed to the sink.
-func (l *ListTask) Durable() bool { return true }
+// Throttled implements the Task.Throttled function and ensures that all log
+// updates are printed to the sink.
+func (l *ListTask) Throttled() bool { return false }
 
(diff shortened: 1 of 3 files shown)


ttaylorr:
Updated from the feedback in https://github.com/git-lfs/git-lfs/pull/2334#issuecomment-309558542, so this should be ready for :eyes:.

git/githistory/log/list_task.go
```diff
@@ -0,0 +1,41 @@
...
+	l.ch <- fmt.Sprintf("%s\n", update)
+}
+
+func (l *ListTask) Complete() {
+	l.ch <- fmt.Sprintf("%s: ...", l.msg)
+	close(l.ch)
+}
+
+// Durable implements the Task.Durable function and ensures that all log updates
+// are printed to the sink.
```
technoweenie: :+1: This is a better comment than the other `Durable()` ones in #2334 like:

> Durable implements Task.Durable and returns false, indicating that this task is not durable.

PR #2338, opened 2017-06-19
git: teach AllRefs()

ttaylorr:
This pull request teaches a new function to the `git` package, `AllRefs()` which returns a listing of all references in a repository.

This is required for the `import` subcommand of `git-lfs-migrate(1)` which needs a listing of all references in order to move references from the old graph onto the migrated one. Ideally this could come from `git-rev-list(1)` (which we already run to generate a listing of commits to migrate), but tricking the output to be machine readable is difficult to do without switching to `git-log`, which doesn't have great machine-readable output either.

---

/cc @git-lfs/core 
/cc #2146

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -1080,6 +1080,50 @@ func RemoteRefs(remoteName string) ([]*Ref, error) {
 	return ret, cmd.Wait()
 }
 
+// AllRefs returns a slice of all references in a Git repository, or an error if
+// those references could not be loaded.
+func AllRefs() ([]*Ref, error) {
+	cmd := subprocess.ExecCommand("git", "for-each-ref")
+
+	outp, err := cmd.StdoutPipe()
+	if err != nil {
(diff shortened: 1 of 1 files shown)


git/git.go
```diff
@@ -1080,6 +1080,45 @@ func RemoteRefs(remoteName string) ([]*Ref, error) {
 	return ret, cmd.Wait()
 }
 
+// AllRefs returns a slice of all references in a Git repository, or an error if
+// those references could not be loaded.
+func AllRefs() ([]*Ref, error) {
+	cmd := subprocess.ExecCommand("git",
+		"for-each-ref", "--format='%(objectname)%00%(refname)'")
```
peff: Using a NUL here is probably overkill. The objectname can't have any whitespace in it. It looks like splitting on NUL isn't any harder, so it's fine as an extra precaution.

The entries are still delimited by newlines, but that's OK, as newlines are forbidden in refnames.

PR #2341, opened 2017-06-21
filepathfilter: teach AllowsPattern()

ttaylorr:
This pull request teaches a new function to the `filepathfilter.Filter` type: `AllowsPattern()`. `AllowsPattern()` has the same semantics as `Allows()`, but also returns the pattern that caused the given filename to be accepted or rejected.

The `git-lfs-migrate(1)` 'import' subcommand needs to modify `.gitattributes` with the pattern(s) that matched a given filename, i.e.: if we cleaned `a.dat` and `*.dat` was given as a pattern, make sure we add `*.dat` to the `.gitattributes`, not `a.dat`.

I'm not totally sold on the name `AllowsPattern()`: @technoweenie do you have ideas here?

---

/cc @git-lfs/core 
/refs #2146

### DIFF
--- a/filepathfilter/filepathfilter.go
+++ b/filepathfilter/filepathfilter.go
@@ -148,6 +148,12 @@ func (p *pathPrefixPattern) Match(name string) bool {
 	return matched
 }
 
+// String returns a string representation of the underlying pattern for which
+// this *pathPrefixPattern is matching.
+func (p *pathPrefixPattern) String() string {
+	return p.rawPattern
+}
+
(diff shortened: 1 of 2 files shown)


ttaylorr:
> I don't like `AllowsPattern` either. Best I can come up with is `MatchPattern(filename string) (pattern string, allowed bool)`

I think my vote would be to stick with `AllowsPattern()` (for consistency with Allows()`) until either one of us can come up with something we like more.

PR #2344, opened 2017-06-21
git/odb: teach SubtreeOrder to sort tree entries

ttaylorr:
This pull request provides an implementation of `sort.Interface` that sorts `[]*TreeEntry`'s in subtree order, required to write valid trees to the object database.

Right now, there are no call sites in Git LFS that modify (append/delete/re-order) the set of `Entries` on the `*git/odb.Tree` instance.

However, the 'import' subcommand of the `git-lfs-migrate(1)` suite will conditionally append a `.gitattributes` tree entry into the root tree. In doing so, we may potentially create an out of order tree, which causes Git to behave in undefined ways. (I found this out during some testing today where Git would persistently mark a file as modified even though it wasn't. `git-fsck(1)` gave further details).

The following pull request will introduce an `Append(entries ...*TreeEntry)` method to append arbitrary tree entries. This has the potential to create out-of-order tree entries. To mitigate this, introduce a way to sort that will be called after appending new tree entries to guarantee that we always write valid (sub-)trees.

---

/cc @git-lfs/core 
/cc #2146

### DIFF
--- a/git/odb/tree.go
+++ b/git/odb/tree.go
@@ -163,6 +163,10 @@ func (s SubtreeOrder) Less(i, j int) bool {
 // This is done because '/' sorts ahead of '\0', and is compatible with the
 // tree order in upstream Git.
 func (s SubtreeOrder) Name(i int) string {
+	if i < 0 || i >= len(s) {
+		return ""
+	}
+
 	entry := s[i]
 
 	if entry.Type() == TreeObjectType {
--- a/git/odb/tree.go
+++ b/git/odb/tree.go
@@ -126,3 +126,47 @@ func (e *TreeEntry) Type() ObjectType {
 		}
 	}
 }
+
+// SubtreeOrder is an implementation of sort.Interface that sorts a set of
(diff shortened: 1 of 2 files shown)


git/odb/tree.go
```diff
@@ -126,3 +126,47 @@ func (e *TreeEntry) Type() ObjectType {
...
+}
+
+// Name returns the name for a given entry indexed at "i", which is a C-style
+// string ('\0' terminated unless it's a subtree), optionally terminated with
+// '/' if it's a subtree.
+//
+// This is done because '/' sorts ahead of '\0', and is compatible with the
+// tree order in upstream Git.
+func (s SubtreeOrder) Name(i int) string {
+	entry := s[i]
```
technoweenie: Is it worth protecting this against potential panics? I doubt it'll be an issue when sorting, but could be if someone imported this and used it completely incorrectly.

ttaylorr: Good call: I added checks for both potential panics in 69cc7a8 & a1874f5.

PR #2346, opened 2017-06-21
tools: introduce tools.OrderedSet

ttaylorr:
This pull request introduces a new type in the `tools` package: `*tools.OrderedSet`.

An `*OrderedSet` is identical in interface to the `*tools.StringSet`, but maintains insertion order when asked for elements in the set.

This is required by the 'import' subcommand of the `git-lfs-migrate(1)` suite, which builds up a listing of tracked patterns over time and adds them to the root tree's `.gitattributes` entry. `tools.StringSet` does not allow for this behavior, since order is not guaranteed, therefore random diffs may appear throughout history on the `/.gitattributes` file.

---

/cc @git-lfs/core 
/refs #2146

### DIFF
--- a/tools/ordered_set.go
+++ b/tools/ordered_set.go
@@ -1,0 +1,217 @@
+package tools
+
+// OrderedSet is a unique set of strings that maintains insertion order.
+type OrderedSet struct {
+	// s is the set of strings that we're keeping track of.
+	s []string
+	// m is a mapping of string value "s" into the index "i" that that
+	// string is present in in the given "s".
+	m map[string]int
+}
+
+// NewOrderedSet creates an ordered set with no values.
+func NewOrderedSet() *OrderedSet {
+	return NewOrderedSetWithCapacity(0)
+}
+
+// NewOrderedSetWithCapacity creates a new ordered set with no values. The
+// returned ordered set can be appended to "capacity" number of times before it
+// grows internally.
+func NewOrderedSetWithCapacity(capacity int) *OrderedSet {
+	return &OrderedSet{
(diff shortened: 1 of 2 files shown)


ttaylorr:
Not sure what happened in that Windows test, it appears unrelated. @technoweenie I don't know if I have access to AppVeyor anymore. Would you mind re-running that test, please?

tools/ordered_set.go
```diff
@@ -0,0 +1,217 @@
...
+	return true
+}
+
+// Contains returns whether or not the given "i" is contained in this ordered
+// set. It is a constant-time operation.
+func (s *OrderedSet) Contains(i string) bool {
+	if _, ok := s.m[i]; ok {
+		return true
+	}
+	return false
```
technoweenie: You could shorten this to 2 lines if you just return `ok`.

tools/ordered_set.go
```diff
@@ -0,0 +1,217 @@
+package tools
+
+// OrderedSet is a unique set of strings that maintains insertion order.
+type OrderedSet struct {
+	// s is the set of strings that we're keeping track of.
+	s []string
+	// m is a mapping of string value "s" into the index "i" that that
+	// string is present in in the given "s".
+	m map[string]int
```
technoweenie: If `s` is always an ordered set of strings, why bother tracking each string's index in `m`?

ttaylorr: Yeah, this one can be a little non-obvious. In most cases we care about whether or not the key exists in `m`, but `Remove()` needs the index as you noted below 👍 .

PR #2353, opened 2017-06-23
commands: add git-lfs-migrate(1) 'import' subcommand

ttaylorr:
This pull request implements the `git-lfs-migrate(1)` 'import' subcommand.

The 'import' subcommand is designed to convert large blobs stored in Git history to LFS pointer files based on the `--include`, `--exclude`, and `--include-ref`, `--exclude-ref` flags. It works by calling the `commands.clean()` function with the blob loaded in memory, and then a) storing that blob's contents in `.git/lfs/objects` and b) writing out an LFS pointer file in its place in the githistory.

Regarding some of the notes from #2146 about when/where to insert .gitattributes changes @technoweenie:

> > When is `.gitattributes` built? I think it'd be cool to slip into the first commit or something.
> 
> Right now I'm adding a `.gitattributes` entry into the first commit that has any paths matching those in the `-I` or `-X` filter, but this could easily be changed to go anywhere before that point in history. I think my preference would be to delay adding the `.gitattributes` entry until the first commit at which that object appears, since that brings us closer to the 1:1 mapping of old history to new, and to me, is clearer in terms of not having a (potentially) huge `.gitattributes` entry up front.
> 

and @andyneff:

> I am actually more inclined to lean towards @ttaylorr's first implementation. I like the idea that the commits before the LFS files are still the same shas, and thus the graphs are topologically connected to each other. I don't know. I just feels good to me ;)

I initially implemented this according to my original suggestion of appending patterns to the `.gitattributes` file when a file of that kind was first rewritten. This is problematic for two reasons:

1. It involves duplicate work to figure out which pattern included/excluded a file (see: #2341).
2. It does _not_ work for `-X`, `--exclude` flags. The `BlobFn` (used to rewrite blobs) is only called on blobs that _do_ match the filter, not ones that don't. This prevents us from ever seeing tree entries that are excluded from the filterset, thus never presenting us an opportunity to add the negative matches to the `.gitattributes`.

Instead, I add the `.gitattributes` changes to the first commit that we migrate by writing lines like:

```gitattributes
pattern merge=lfs filter=lfs diff=lfs -text
```

and adding negative entries for `--exclude`'d patterns like:

```gitattributes
pattern text -merge -filter -diff
```

We keep track of these in a set of patterns that the LFS migrator has tracked, and merge them each time with the .gitattributes in the root tree therefore persisting any .gitattributes changes that exist in the original history. See:

```go
// Create a blob of the attributes that are optionally
// present in the "t" tree's .gitattributes blob, and
// union in the patterns that we've tracked.
//
// Perform this Union() operation each time we visit a
// root tree such that if the underlying .gitattributes
// is present and has a diff between commits in the
// range of commits to migrate, those changes are
// preserved.
blob, err := trackedToBlob(db, theirs.Clone().Union(ours))
```

Here's some example output:

1. Generate some commits ahead of the remote ref that contain data to be migrated to LFS:

```sh
~/g/git-lfs (migrate-demo) $ for i in $(seq 1 10); do
  base64 < /dev/urandom | head -c 64 > a.dat
  git add a.dat
  git commit -m "a.dat: $i"
  done
# ...
```

2. Run `git lfs migrate info` on that data:

```
~/g/git-lfs (migrate-demo) $ git lfs migrate info --above=0b --include="*.dat"
migrate: Rewriting commits: 100% (10/10), done
*.dat   640 B   10/10 files(s)  100%
```

3. Run `git lfs migrate import` to rewrite those commits such that the large files are tracked with LFS:

```
~/g/git-lfs (migrate-demo) $ git lfs migrate import --include="*.dat"
migrate: Rewriting commits: 100% (11/11), done
  migrate-demo  f18bb746d44e8ea5065fc779bb1acdf3cdae7ed8 -> 35b0fe0a7bf3ae6952ec9584895a7fb6ebcd498b
migrate: Updating refs: ..., done
```

4. Observe the initial commit to show that a) original `.gitattributes` contents are persisted, b) new `.gitattributes` patterns are added accordingly, and c) the `*.dat` file was added to Git LFS.

```diff
~/g/git-lfs (migrate-demo) $ git show 3f65c4db7313e9f4fe0c546fcbba8d10593f2025
commit 3f65c4db7313e9f4fe0c546fcbba8d10593f2025
Author: Taylor Blau <me@ttaylorr.com>
Date:   Fri Jun 23 13:07:12 2017 -0600

    a.dat: 1

diff --git a/.gitattributes b/.gitattributes
index e4c71dbf..3716304e 100644
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
 * text=auto
 * eol=lf
 *.bat eol=crlf
+*.dat filter=lfs diff=lfs merge=lfs -text
diff --git a/a.dat b/a.dat
new file mode 100644
index 00000000..ec26fbcd
--- /dev/null
+++ b/a.dat
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:86df498e7a88666447e0c540defbef3f2b2091eecba3c0b7c9ac21962aaef576
+size 64
```

5. Observe an interdiff to show that changes are persisted through the migration and the diffs apply cleanly:

```diff
~/g/git-lfs (migrate-demo) $ git show 94a7b26a4260b04af84c7ce9c7fae6e6586345a6
commit 94a7b26a4260b04af84c7ce9c7fae6e6586345a6
Author: Taylor Blau <me@ttaylorr.com>
Date:   Fri Jun 23 13:07:12 2017 -0600

    a.dat: 2

diff --git a/a.dat b/a.dat
index ec26fbcd..b7fa3fba 100644
--- a/a.dat
+++ b/a.dat
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:86df498e7a88666447e0c540defbef3f2b2091eecba3c0b7c9ac21962aaef576
+oid sha256:bb1df455c4d90dbc0c6903500c036096eeec78f1abc151997a80387c81ccd7e0
 size 64
```

6. Observe that `HEAD` contains a checked-out LFS object, a pointer file in the index, and a corresponding entry in `.git/lfs/objects`:

```
~/g/git-lfs (migrate-demo) $ git cat-file -p :a.dat
oid sha256:b91016d0851cd677a9209a0c0d11c7e9b2d5654f2baeb6485579b899cf6ffa01
size 64
~/g/git-lfs (migrate-demo) $ cat a.dat
yVnUDK0PSDqmoSvY4W/zD/a3GD/rEuPEx/kbwnjFK4E9jMLp0xRiLzCqmwwkPROO%                                                          
~/g/git-lfs (migrate-demo) $ cat .git/lfs/objects/b9/10/b91016d0851cd677a9209a0c0d11c7e9b2d5654f2baeb6485579b899cf6ffa01 
yVnUDK0PSDqmoSvY4W/zD/a3GD/rEuPEx/kbwnjFK4E9jMLp0xRiLzCqmwwkPROO%                                                          
```

Closes: #2146.

---

/cc @git-lfs/core 
/refs #2146

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
   global:
     - GIT_LFS_TEST_DIR="$HOME/git-lfs-tests"
     - GIT_SOURCE_REPO="https://github.com/git/git.git"
-    - GIT_EARLIEST_SUPPORTED_VERSION="v1.8.5"
+    - GIT_EARLIEST_SUPPORTED_VERSION="v1.9.0"
     - GIT_LATEST_SOURCE_BRANCH="master"
 
 matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
   global:
     - GIT_LFS_TEST_DIR="$HOME/git-lfs-tests"
     - GIT_SOURCE_REPO="https://github.com/git/git.git"
-    - GIT_EARLIEST_SUPPORTED_VERSION="v1.9.0"
+    - GIT_EARLIEST_SUPPORTED_VERSION="v2.0.0
     - GIT_LATEST_SOURCE_BRANCH="master"
 
 matrix:
--- a/circle.yml
+++ b/circle.yml
@@ -2,7 +2,7 @@ machine:
   environment:
     GIT_LFS_TEST_DIR: $HOME/git-lfs-tests
     GIT_SOURCE_REPO: https://github.com/git/git.git
-    GIT_EARLIEST_SUPPORTED_VERSION: v1.8.5
+    GIT_EARLIEST_SUPPORTED_VERSION: v1.9.0
     GIT_LATEST_SOURCE_BRANCH: master
     XCODE_SCHEME: test
     XCODE_WORKSPACE: test
--- a/circle.yml
+++ b/circle.yml
@@ -2,7 +2,7 @@ machine:
   environment:
     GIT_LFS_TEST_DIR: $HOME/git-lfs-tests
     GIT_SOURCE_REPO: https://github.com/git/git.git
-    GIT_EARLIEST_SUPPORTED_VERSION: v1.9.0
+    GIT_EARLIEST_SUPPORTED_VERSION: v2.0.0
     GIT_LATEST_SOURCE_BRANCH: master
     XCODE_SCHEME: test
     XCODE_WORKSPACE: test
--- a/commands/command_clean.go
+++ b/commands/command_clean.go
@@ -16,17 +16,22 @@ import (
 // error is non-fatal, otherwise they will halt using the built in
 // `commands.Panic`.
 //
+// If fileSize is given as a non-negative (>= 0) integer, that value is used
+// with preference to os.Stat(fileName).Size(). If it is given as negative, the
+// value from the `stat(1)` call will be used instead.
+//
 // If the object read from "from" is _already_ a clean pointer, then it will be
 // written out verbatim to "to", without trying to make it a pointer again.
-func clean(to io.Writer, from io.Reader, fileName string) error {
+func clean(to io.Writer, from io.Reader, fileName string, fileSize int64) error {
 	var cb progress.CopyCallback
 	var file *os.File
-	var fileSize int64
 
 	if len(fileName) > 0 {
 		stat, err := os.Stat(fileName)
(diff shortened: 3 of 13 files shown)


ttaylorr:
Opened up #2358, which should cause this branch to pass on CircleCI after it's merged in.

PR #2369, opened 2017-06-29
commands: only pass --jobs to 'git clone' if set

### LINKED ISSUE #2366: Unable to clone repo with git lfs after 2.2.0 update
## Behaviour

As of the 2.2.0 update, calling `git lfs clone` returns the following:
```
error: unknown option `jobs'
usage: git clone [<options>] [--] <repo> [<dir>]

    ...
```
Calling the same command with 2.1.1 returns the expected result:
```
You must specify a repository to clone.

usage: git clone [<options>] [--] <repo> [<dir>]

    ...
```

## Environment
Both versions were tested inside a docker container using the following dockerfile (the `apt-get install git-lfs=VERSION`) being changed accordingly:
```
FROM ubuntu:16.04

RUN apt-get update
RUN apt-get install curl git -y

RUN curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash
RUN apt-get install git-lfs=2.2.0 -y

RUN git lfs install
```
Calling:
```
> docker build . -t git-lfs-test
> docker run --rm git-lfs-test git lfs clone
```
Also tried ubuntu:14.04 base image with same result.

Alternatively, without docker I was able to clone again with `git lfs clone` after installing 2.1.1 again. Just wanted to make sure it wasn't an environment issue before submitting an issue.

technoweenie:
Running `git lfs clone` always passes the `--jobs` option to `git clone`. This won't fly if you have git < 2.9:

```
error: unknown option `jobs'
usage: git clone [<options>] [--] <repo> [<dir>]
```

Fixes #2366 by setting the flag default to -1.

### DIFF
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -153,7 +153,7 @@ func init() {
 		cmd.Flags().StringVarP(&cloneFlags.ShallowExclude, "shallow-exclude", "", "", "See 'git clone --help'")
 		cmd.Flags().BoolVarP(&cloneFlags.ShallowSubmodules, "shallow-submodules", "", false, "See 'git clone --help'")
 		cmd.Flags().BoolVarP(&cloneFlags.NoShallowSubmodules, "no-shallow-submodules", "", false, "See 'git clone --help'")
-		cmd.Flags().Int64VarP(&cloneFlags.Jobs, "jobs", "j", 1, "See 'git clone --help'")
+		cmd.Flags().Int64VarP(&cloneFlags.Jobs, "jobs", "j", -1, "See 'git clone --help'")
 
 		cmd.Flags().StringVarP(&includeArg, "include", "I", "", "Include a list of paths")
 		cmd.Flags().StringVarP(&excludeArg, "exclude", "X", "", "Exclude a list of paths")


technoweenie:
Ah, the reason our ci didn't catch this is because typically test against 3 git versions (on circle and travis):

* Latest release
* `GIT_EARLIEST_SUPPORTED_VERSION` which is 2.0.0.
* Current `master` branch.

The `GIT_EARLIEST_SUPPORTED_VERSION` test _should_ have caught this, but [the clone test requires git 2.2.0+](https://github.com/git-lfs/git-lfs/blob/d65f80119281b1b7f7f650c44324d25087a00484/test/test-clone.sh#L5). Doh. I don't know if it's worth adding to the test matrix though.

ttaylorr:
> The `GIT_EARLIEST_SUPPORTED_VERSION` test _should_ have caught this, but [the clone test requires git 2.2.0+](https://github.com/git-lfs/git-lfs/blob/d65f80119281b1b7f7f650c44324d25087a00484/test/test-clone.sh#L5). Doh. I don't know if it's worth adding to the test matrix though.

Ah, thanks for the explanation. This was very helpful. I think it's certainly worth poking around at to split out the tests that require Git 2.2.0+ and those that don't into separate test files. I don't have time to get to this immediately, but I've opened up https://github.com/git-lfs/git-lfs/issues/2391 to track this for the 2.3.0 release.

Issue #2380, opened 2017-07-04
`git lfs migrate info` hangs at `Sorting commits: ...`


sentience:
```
$ git version
git version 2.12.2

$ git lfs version
git-lfs/2.2.0 (GitHub; darwin amd64; go 1.8.3)

$ GIT_TRACE=1 GIT_CURL_VERBOSE=1 git lfs migrate info
19:49:05.786928 git.c:595               trace: exec: 'git-lfs' 'migrate' 'info'
19:49:05.787711 run-command.c:369       trace: run_command: 'git-lfs' 'migrate' 'info'
trace git-lfs: run_command: 'git' config -l
trace git-lfs: run_command: git cat-file --batch
trace git-lfs: run_command: 'git' rev-parse HEAD --symbolic-full-name HEAD
trace git-lfs: run_command: 'git' rev-parse master --symbolic-full-name master
trace git-lfs: run_command: git rev-list --reverse --topo-order --do-walk master ^refs/remotes/origin/change-font ^refs/remotes/origin/master --
migrate: Sorting commits: ...
```

That last command hangs forever at that point. Any tips for troubleshooting?

ltrzesniewski:
FWIW, I had this issue when my remote URL was a local directory.
Switching to a LFS-enabled https URL or removing the remote altogether fixed the issue for me.

sentience:
@ltrzesniewski My remote was an ordinary `git@github.com:organisation/repository-name.git`, but removing it did successfully work around the hang. 👍

ashumkin:
the same issue
removing remote resolves THIS issue
but the issue #2381 occurs

ttaylorr:
Hi everybody, thanks for opening and commenting on this issue. I think what's going on here is a hang that occurs when there are no commits to migrate.

I took a shot at fixing this in https://github.com/git-lfs/git-lfs/pull/2383, and I'd love if you tried it out. If it works for everybody, I think it would be rad to ship this in v2.2.1 ⚡️.

Here's a list of builds for LFS's default distributions (OS, arch combos) that includes #2383. Can you give it a shot using the same remotes and command that caused it to hang earlier and let me know if it works? I'd expect the output to be blank, since there shouldn't be any commits to migrate (therefore, no files, etc).

[git-lfs-darwin-386-2.2.0.tar.gz](https://github.com/git-lfs/git-lfs/files/1124683/git-lfs-darwin-386-2.2.0.tar.gz)
[...]

PR #2383, opened 2017-07-05
git/githistory: fix hanging on empty set of commits

ttaylorr:
This pull request fixes a hang that occurs when the set of commits to migrate is empty, as pointed out in #2380.

In `*git/odb.Rewriter.commitsToMigrate()`, we store a set of `[][]byte`, where each `[]byte` entry is a commit's SHA1 which should be migrated. Our logic looks something like this:

```go
 func (r *Rewriter) commitsToMigrate(opt *RewriteOptions) ([][]byte, error) {
 	waiter := r.l.Waiter("migrate: Sorting commits")

 	scanner, err := git.NewRevListScanner(...)

 	var commits [][]byte
 	for scanner.Scan() {
 		if len(commits) == 0 {
 			waiter.Complete()
 		}

 		commits = append(commits, scanner.OID())
 	}

 	return commits
}
``` 

This has the behavior of flashing a `"migrate: Sorting commits"` message to the terminal, appending `"..., done"` when the first result was received from `git-rev-list(1)`. If, however, the loop inside of `for scanner.Scan()` wasn't entered, the `Waiter` never receives a `Complete()` call.

The `*git/githistory/log` package [picks up the next task][1] (`r.l.Waiter()`, `r.l.Counter()`, etc.) without blocking, but [cannot receive any updates on it][2] causing any channel write whose buffer is full to block. This results in the subsequent [rewriting commits log task][3] to block infinitely, for the waiter is never closed.

There are two cases where an early return would cause such a block:

1. Errors, like `scanner.Err() != nil`, `git.RevListScanner()` returning a non-nil error, etc. These don't require the `"migrate: Sorting commits..."` message to print `"..., done"`, since they cause LFS to exit before spinning up the next logging task.
2. Empty result sets from `git-rev-list(1)`, which is fixed in this PR.

Closes: #2380.

---

/cc @git-lfs/core 
/cc @shana 

[1]: https://github.com/git-lfs/git-lfs/blob/v2.2.0/git/githistory/log/log.go#L162-L179
[2]: https://github.com/git-lfs/git-lfs/blob/v2.2.0/git/githistory/log/log.go#L128-L132
[3]: https://github.com/git-lfs/git-lfs/blob/v2.2.0/git/githistory/rewriter.go#L171

### DIFF
--- a/git/githistory/rewriter.go
+++ b/git/githistory/rewriter.go
@@ -382,6 +382,12 @@ func (r *Rewriter) commitsToMigrate(opt *RewriteOptions) ([][]byte, error) {
 		commits = append(commits, scanner.OID())
 	}
 
+	if len(commits) == 0 {
+		// If git-rev-list(1) returned an empty set, mark the log waiter
+		// (see above) as complete.
+		waiter.Complete()
+	}
+
 	if err = scanner.Err(); err != nil {
 		return nil, err
 	}
--- a/git/githistory/rewriter.go
+++ b/git/githistory/rewriter.go
@@ -375,18 +375,10 @@ func (r *Rewriter) commitsToMigrate(opt *RewriteOptions) ([][]byte, error) {
 
 	var commits [][]byte
 	for scanner.Scan() {
-		if len(commits) == 0 {
-			waiter.Complete()
-		}
-
 		commits = append(commits, scanner.OID())
 	}
 
-	if len(commits) == 0 {
-		// If git-rev-list(1) returned an empty set, mark the log waiter
-		// (see above) as complete.
-		waiter.Complete()
-	}
+	waiter.Complete()
 
 	if err = scanner.Err(); err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


technoweenie:
It looks like `commitsToMigrate()` has a `len(commits) == 0` check twice. Does it need that? It looks like you want the waiter to complete either after the scanner returns empty, or the first time `Scan()` is called. Is there some reason the waiter can't be `completed` after `scanner.Scan()` breaks out of the `for`?

ttaylorr:
> It looks like `commitsToMigrate()` has a `len(commits) == 0` check twice. Does it need that?

Great call! I double checked this behavior before and after 0fd2fb9, and it had no measurable effect. I think my original intent was to keep the "sorting commits" message's timing close to what Git is doing (i.e., `git-rev-list(1)` is sorting commits from its invocation until the first commit SHA1 is sent over stdout), but I think trading this accuracy for simplicity is a worthy exchange.

ttaylorr:
Thanks for the review! 🙇

PR #2387, opened 2017-07-05
Allow to set batch size in LFS config

mlajszczak:
Adjusting the batch size might increase download/upload performance
significantly.

E.g. in cases where concurrent download/upload of objects is very fast
but the server has high latency the latter becomes a bottleneck and
increasing batch size might help.

### DIFF
--- a/docs/man/git-lfs-config.5.ronn
+++ b/docs/man/git-lfs-config.5.ronn
@@ -60,6 +60,10 @@ be scoped inside the configuration for a remote.
 
   These settings control how the upload and download of LFS content occurs.
 
+* `lfs.batchsize`
+
+  The number of objects in a batch. Default 100.
+
 * `lfs.concurrenttransfers`
 
   The number of concurrent uploads/downloads. Default 3.
--- a/tq/manifest.go
+++ b/tq/manifest.go
@@ -8,11 +8,13 @@ import (
 )
 
 const (
+	defaultBatchSize           = 100
 	defaultMaxRetries          = 8
 	defaultConcurrentTransfers = 3
 )
 
 type Manifest struct {
+	batchSize int
 	// MaxRetries is the maximum number of retries a single object can
 	// attempt to make before it will be dropped.
 	maxRetries           int
@@ -62,6 +64,9 @@ func NewManifestWithClient(apiClient *lfsapi.Client) *Manifest {
 
 	var tusAllowed bool
 	if git := apiClient.GitEnv(); git != nil {
+		if v := git.Int("lfs.batchsize", 0); v > 0 {
+			m.batchSize = v
+		}
 		if v := git.Int("lfs.transfer.maxretries", 0); v > 0 {
 			m.maxRetries = v
 		}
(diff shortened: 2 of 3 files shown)


technoweenie:
What do you think about adding an upper and lower bound? Obviously anything < 1 should be invalid. Not sure what a good upper bound should be.... 1000? Setting an upper bound will prevent the client from delivering massive JSON blobs to LFS servers.

Your use case sounds good, but have you run any benchmarks confirming that it does in fact improve things?

Also, keep in mind that servers may reject large batch sizes. It's important to be able to tune things for performance, but hopefully servers are protecting themselves from attempting to process batches of objects that are too large.

92161898515:
Muito bom

mlajszczak:
Hi,

Bounds seem good to me but I find 1K upper bound too defensive. I'd go with something like 10K. As for lower bound, I don't have a strong opinion.

I've done some measurements but the results are less satisfactory than I'd expect.

I've tested different batch sizes on a repository with 10k objects (of size 200KB-1MB) in its head. These are my findings:
- For batch size 10k download takes significantly less time than in default case (100), ~50s vs 4m45s. However, after downloading the big batch, git-lfs hangs for a long while checking out objects. As a result, big batch saves only 30s comparing to small one (end-to-end git-lfs pull: 5m for 100 batch size, 4m30 for the big one)
- I've tried some smaller batches (1k, 2.5K) but I only managed to save 30s tops. git-lfs seemed to hang for some time after downloading few batches.

ttaylorr:
> Upper bounds seems good to me (I'd go with something more aggressive than 1k, 10k?).

How about 1024, as a power of two.

technoweenie:
I don't know if a power of 2 is any better than a power of 10 in this case. 

@mlajszczak Two questions: how many concurrent transfers are you using?
 `lfs.concurrenttransfers` defaults to 3.

Also, what version of Git are you using? What git/lfs commands are you running to test this? If you're using the latest LFS version, and Git v2.11+, you can take advantage of Git filter processes, which should greatly improve the checkout time over Git v2.10 and below.

mlajszczak:
I set ```lfs.concurrentransfers``` to 30.
I was using Git in some older version (2.7.x). I assume that git-lfs forks to Git (i.e. ```git``` from ```PATH```)? When I upgrade Git, should it help out of the box or I need to configure something?

technoweenie:
@mlajszczak Depends on what command you're running. If you're running `git lfs [ fetch | pull | clone ]`, upgrading Git will have no effect. But increasing `lfs.concurrenttransfers` will help.

If you're running `git [ pull | clone ]`, then newer versions of Git will immediately start using the `filter.lfs.process` config value, and performance will be improved. Unfortunately, `lfs.concurrenttransfers` will not have much effect here _yet_. There are future improvements coming to Git and Git LFS to improve this.

technoweenie:
I think this PR should be closed. `lfs.concurrenttransfers` lets you improve the concurrency without affecting the batch size. Increasing the batch size is only useful if `lfs.concurrenttransfers` is over 100 (which it doesn't need to be at all).

I would be interested in a better `lfs.concurrenttransfers` default though. Maybe `3 * {cpu-cores}` or something?

mlajszczak:
I'm fine with closing this PR.
I not sure about ```lfs.concurrenttransfers``` being a function of CPU number as it's IO bound.

PR #2389, opened 2017-07-05
migrate: don't checkout HEAD on bare repositories

ttaylorr:
This pull request skips the final `git-checkout(1)` that we do on repositories after migrating them via `git-lfs-migrate(1) import`.

This was pointed out in https://github.com/git-lfs/git-lfs/issues/2380#issuecomment-313148926.

Before this PR:

```
~/D/foo.git (master) $ git rev-parse --is-bare-repository
true

~/D/foo.git (master) $ git lfs migrate import
migrate: Rewriting commits: 100% (1/1), done
  master        85d2a9aa2e1fa73548092096f563688e6cd3f2c6 -> a65ddcf4418a78b9e414ead8a5c86c4aebfbb0e1
migrate: Updating refs: ..., done
Error running git [checkout --force]: 'fatal: This operation must be run in a work tree' 'exit status 128'
```

Here's a quick look at how things went down:

1. 2a2a76c: teach `git.IsBare()` to determine if a repository is bare (via `git-rev-list --is-bare-repository`).
2. ca61a20: only call `git.Checkout()` if `!git.IsBare()`.
3. f3fc56c: teach `make_bare` to convert a full repository into a bare one
4. 9b0a2e8: call `git-lfs-migrate(1)` on a bare repository to ensure it doesn't fail.

---

/cc @git-lfs/core 
/cc #2380

### DIFF
--- a/commands/command_migrate_import.go
+++ b/commands/command_migrate_import.go
@@ -94,8 +94,12 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 		UpdateRefs: true,
 	})
 
-	if err := git.Checkout("", nil, true); err != nil {
-		ExitWithError(err)
+	if bare, _ := git.IsBare(); !bare {
+		// Only perform `git-checkout(1) -f` if the repository is
+		// non-bare.
+		if err := git.Checkout("", nil, true); err != nil {
+			ExitWithError(err)
+		}
 	}
 }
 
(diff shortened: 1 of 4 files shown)


ttaylorr:
@technoweenie thanks for the review and the suggestion! I appreciate it. I dig the explicitness of `git.IsBare()` and `git.Checkout()` as well, but I think inverting the arguments could be cool, too: returning an `error` from `git.Checkout()` and only exiting if it doesn't satisfy `git.IsBareRepo(err)`.

I will play around with this locally and push a separate PR out 👍 .

Issue #2363, opened 2017-06-28
"git lfs install" won't install the pre-push hook


schwaerz:
After updating to the newer git & git lfs version(s) I've to explicitly call `git lfs push origin master` to get the files pushed to our Artifactory.
I found that for some reason there's no pre-push hook installed (neither in the repository nor globally) - although I get following messages when running `git lfs install`:

    $ git lfs install
    Updated git hooks.
    Git LFS initialized.

I'm running git & git lfs on Windows7 (64bit).
Tried using git lfs 2.1.1 which is bundled with the latest git 2.13.2 as well as the latest git lfs available for windows (2.2.0).

Probably I'm doing something wrong. Unfortunately I found very less documentation about git lfs - and nothing about the hooks.

Hope somebody can clarify.

technoweenie:
Did you run that in your git repository directory? It only works when run from the working directory of a repository. There are a number of commands that will install it for you (such as `lfs track`) too, so you shouldn't have to run it manually too.

However, `git lfs push` does not need the `pre-push` hook, as you're already calling LFS. LFS tries to be optimal about what it tries to upload. If your local repo is caught up with the remote, then it may not see anything to push. You can run `git lfs push --all origin master` to force a full history scan across all refs. See more at the [git-lfs-push man page](https://github.com/git-lfs/git-lfs/blob/d65f80119281b1b7f7f650c44324d25087a00484/docs/man/git-lfs-push.1.ronn).

schwaerz:
Thanks for the answer.

Sorry but forgot to mention that I was running the install from within the repository root.

What I tried was using a plain `git push` which worked well before when using git lfs version 1.5.6 or so. I'd assume that this would require the pre-push hook.
If I call the git push I later on have to do what you proposed to get the lfs files pushed to the server: `git lfs push --all origin master`.
However I'd have expected that a plain `git push` should do the job so I - and all the other colleagues around - won't forget to push the lfs's.

technoweenie:
Ah, that makes more sense. Git LFS should definitely installing the hooks for you. You shouldn't have to run `git lfs push` after every `git push`.

I confirmed in the code hooks are being installed through a number of LFS commands, including `install`. However, errors are silently ignored, which is just _fantastic_. I made a custom build for you that dumps this output to `GIT_TRACE`:

```
$ GIT_TRACE=1 ~p/git-lfs/bin/git-lfs install
trace git-lfs: run_command: 'git' config -l
[...]

PR #2393, opened 2017-07-07
lfs: trace hook install, uninstall, upgrade

ttaylorr:
This pull request adds tracing output to the `git lfs install`, `git lfs uninstall`, and `git lfs upgrade` commands when installing `.git/hooks`, similar to that suggested in: https://github.com/git-lfs/git-lfs/issues/2363#issuecomment-311985990 & https://github.com/git-lfs/git-lfs/issues/2363#issuecomment-311994628.

This should make it easier to diagnose situations where hooks aren't being installed correctly, as we'll now have information about:

1. Which files are being `stat(1)`-ed
2. Which directories are being created
3. Which files are being written
4. Whether files that do exist are marked as upgrade-able.

---

/cc @git-lfs/core 
/cc https://github.com/git-lfs/git-lfs/issues/2363

### DIFF
--- a/lfs/hook.go
+++ b/lfs/hook.go
@@ -67,14 +67,18 @@ func (h *Hook) Dir() string {
 // directory. It returns and halts at any errors, and returns nil if the
 // operation was a success.
 func (h *Hook) Install(force bool) error {
+	msg := fmt.Sprintf("Install hook: %s, force=%s, path=%s", h.Type, force, h.Path())
+
 	if err := os.MkdirAll(h.Dir(), 0755); err != nil {
 		return err
 	}
 
 	if h.Exists() && !force {
+		tracerx.Printf(msg + ", upgrading...")
 		return h.Upgrade()
 	}
 
+	tracerx.Printf(msg)
 	return h.write()
 }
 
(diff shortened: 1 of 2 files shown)


ttaylorr:
@technoweenie thanks for the feedback! I totally agree. I just pushed up some commits to take the output down to:

```
~/g/example-repo $ GIT_TRACE=1 git lfs install
# ...
race git-lfs: Install hook: pre-push, force=false, path=/Users/ttaylorr/github/example-repo/.git/hooks/pre-push, upgrading...
trace git-lfs: Install hook: post-checkout, force=false, path=/Users/ttaylorr/github/example-repo/.git/hooks/post-checkout, upgrading...
trace git-lfs: Install hook: post-commit, force=false, path=/Users/ttaylorr/github/example-repo/.git/hooks/post-commit, upgrading...
trace git-lfs: Install hook: post-merge, force=false, path=/Users/ttaylorr/github/example-repo/.git/hooks/post-merge, upgrading...
Updated git hooks.
Git LFS initialized.
```

With that, I think that this should be ready to go.

technoweenie:
Better, but not quite there. Looks like a bad format code:

```
trace git-lfs: Install hook: pre-push, force=%!!(MISSING)s(bool=false), path=/Users/rick/go/src/github.com/git-lfs/git-lfs/.git/hooks/pre-push, upgrading...
```

ttaylorr:
> Better, but not quite there. Looks like a bad format code:

All fixed!

Issue #2415, opened 2017-07-12
git-lfs-migrate(1): packed object reads 


ttaylorr:
# Goal

Increase the speed at which `git-lfs-migrate(1)` can read packed objects.

## Background

To rewrite history, the `git-lfs-migrate(1)` command has to read blob,
commit, and tree objects from the object database of a repository, usually
inside of `.git/objects`.

This functionality is implemented in the `git/odb` package which can (as of
v2.2.0) read and write loose objects. This works fine, but becomes an issue when
trying to read an object that is packed.

When a packed object is encountered in v2.2.0, we write to a long-running
invocation of `git-cat-file(1) --batch`, which then writes to stdout a (mostly)
correct loose representation of the object whose SHA1 we wrote previously.

This too works fine, but can be improved in two ways:

1. **Parallelization cost**. In order to have more "workers" traversing the
   history graph, we must linearly increase the number of `git-cat-file(1)`
   invocations we have running. Since the output of `git-cat-file(1)` is not
   multiplexed, we can only read one object out of the program at a time.
   Therefore, if we have more than one worker(s) trying to read an object at a
   time, we must either a) synchronize and wait, or b) run more
   `git-cat-file(1)`'s.
2. **Pipe throughput**. As we have experienced with `filter.lfs.process` (and
   friends), the pipe data transfer overhead is non-zero. Moreover, this is
   effect is especially noticeable on the shell that ships with Git for Windows
   (a project which distributes LFS). If we can avoid the pipe data transfer
   overhead, we will increase the speed at which `git-lfs-migrate(1)` can
   rewrite history.

## Proposition(s)

Teach the `git/odb` package how to read packed objects. This will *eliminate
the parallelization costs* (it is cheaper to open a file descriptor than it is
to run a program), and *remove pipe data throughput from the equation*.

I'd like the API to look something like this:

```go
package git/odb/pack

type Index struct { ... }

type IndexEntry struct {
  Name       [20]byte
  PackOffset uint64
  CRC        uint32
}

// Entry returns the a packed object entry for the given name in O(log(n)) time.
func (i *Index) Entry(name [20]byte) *IndexEntry { ... }
```

and:

```go
package git/odb/pack

type Packfile struct { ... }

// Resolve uses the packfile index to find and apply the delta COPY/ADD
// instructions to resolve the object contents for the given name "name".
func (p *Packfile) Resolve(name [20]byte) (io.Reader, error) { ... }
```

lastly:

```go
package git/odb

[...]

PR #2420, opened 2017-07-15
git/odb/pack: introduce '*Index' type

ttaylorr:
This pull request introduces the `*Index` type, and creates a new package `git/odb/pack` to hold it.

The goal of this new package is twofold:

1. Parse index files to find packfile offsets.
2. Parse and assemble objects within packfiles.

This pull request is the first in a series of four aimed at accomplishing the first goal. Contained in this pull request is the scaffolding necessary to get us going for parsing actual index files. That means:

- An `*Index` type, which will eventually get a `Entry([]byte) uint64, error)` method to resolve object names to their offsets in a packfile.
- An `IndexVersion` type, which will learn how to parse index entries in later pull requests in this series.
- A `DecodeIndex()` function, which will learn how to use the above `IndexVersion` instances to parse the index header, and eventually find objects in the index.

To address a point raised in #2415:

> @peff wisely points out that I can use [`mmap(2)`](https://en.wikipedia.org/wiki/Mmap) in order to avoid the `seek(1)` cost of bouncing around a file on disk. To me, this means that the 2nd option of lazily loading each entry makes more sense, since:

This is the implementation that I pursued in this series. The idea is that, in order to save memory (and to avoid parsing objects that wont be searched for) we delay parsing index entires until they are asked for.

The next two pull requests in the series will introduce implementations of `IndexVersion.Search()` which have the following signature:

```go
func (i IndexVersion) Search(idx *Index, name []byte, at uint64) (*IndexEntry, int, error) {
        // ...
}
```

Or, in other words: given an index, an object to search for, and a location to search in, return either an entry (match), a comparison value (no match), or an error otherwise.

We can use this comparison value (see: `bytes.Compare()`) in order to direct the next window of our binary search through the index file in order to produce an entry in `O(log(n))` time.

---

/cc @git-lfs/core 
/cc #2415

### DIFF
--- a/git/odb/pack/errors.go
+++ b/git/odb/pack/errors.go
@@ -1,0 +1,15 @@
+package pack
+
+import "fmt"
+
+// UnsupportedVersionErr is a type implementing 'error' which indicates a
+// the presence of an unsupported packfile version.
+type UnsupportedVersionErr struct {
+	// Got is the unsupported version that was detected.
+	Got uint32
+}
+
+// Error implements 'error.Error()'.
+func (u *UnsupportedVersionErr) Error() string {
+	return fmt.Sprintf("git/odb/pack: unsupported version: %d", u.Got)
+}
--- a/git/odb/pack/errors_test.go
+++ b/git/odb/pack/errors_test.go
@@ -1,0 +1,13 @@
+package pack
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUnsupportedVersionErr(t *testing.T) {
+	u := &UnsupportedVersionErr{Got: 3}
+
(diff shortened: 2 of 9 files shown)


ttaylorr:
A future point of consideration for the `IndexVersion` interface: perhaps it could be expanded to something like:

```go
package git/odb/pack

type IndexVersion interface {
        Name(r io.ReaderAt, at uint64) ([]byte, error)
        Entry(r io.ReaderAt, at uint64) (*IndexEntry, error)
}
```

which would remove the `comparison int` return value from the existing `Search` function. Instead, the `*Index` type would be responsible for calling `bytes.Compare()`.

I think that might be a more logical separation of concerns, but I propose we hold off on implementing this change until after this series is reviewed.

PR #2423, opened 2017-07-15
git/odb/pack: implement Index.Entry() to find object offsets

ttaylorr:
This pull request uses https://github.com/git-lfs/git-lfs/pull/2421 & https://github.com/git-lfs/git-lfs/pull/2422 to implement a [binary search][1] and entry parse operation over V1 and V2 index files. This is used to determine a given object's offset in the corresponding packfile.

In https://github.com/git-lfs/git-lfs/commit/b761e15ee4eac7255721f88a21d46ef80f9fcab4, we implement a `*bounds` type, which stores a `left` and `right` index used for the binary search. This is done in order to implement the `Equal()` method on `*bounds` in order to detect infinite loops, as:

```go
var last *bounds

for bounds.Left() < bounds.Right() {
  if last.Equal(bounds) {
    // If the bounds are unchanged, that means either that the object does not
    // exist in the packfile, or the fanout table is corrupt.
    //
    // Either way, we won't be able to find the object.  Return immediately to
    // prevent infinite looping.
    return nil, nil
  }
  last = bounds

  // ...
}
```

In https://github.com/git-lfs/git-lfs/commit/0c15b26c215f4b6372019906daed6bbb1fa3938c, we use the `*bounds` type implement a binary search through the index files in order to find the packfile offset for a given object. This works like the following:

1. Use the fanout table to figure out a lower and upper bound for where the object could be within the index file.
  a. If the bounds are unchanged from the last search, return immediately: we have a corrupt index.
  b. If `left >= right`, we don't have the object, so return `(nil, nil)` immediately.
2. For those offsets, pick a midpoint and try and parse the entry at that location.
3. If the entry at that location has a SHA-1 greater than the one we're looking for, narrow our search to the lower half of our sorted list.
4. If the entry at that location has a SHA-1 lesser than the one we're looking for, narrow our search to the upper half of our sorted list.
5. If the entry has a SHA-1 equal to the one that we were looking for, return it immediately.
6. Otherwise, with the modified bounds, go to step 2.

---

/cc @git-lfs/core @peff
/cc #2415 

[1]: https://en.wikipedia.org/wiki/Binary_search_algorithm

### DIFF
--- a/git/odb/pack/index.go
+++ b/git/odb/pack/index.go
@@ -28,8 +28,97 @@ func (i *Index) Count() int {
 	return int(i.fanout[255])
 }
 
+// Entry returns an entry containing the offset of a given SHA1 "name".
+//
+// Entry operates in O(log(n))-time in the worst case, where "n" is the number
+// of objects that begin with the first byte of "name".
+//
+// If the entry cannot be found, (nil, nil) will be returned. If there was an
+// error searching for or parsing an entry, it will be returned as (nil, err).
+//
+// Otherwise, (entry, nil) will be returned.
+func (i *Index) Entry(name []byte) (*IndexEntry, error) {
+	var last *bounds
+	bounds := i.bounds(name)
+
+	for bounds.Left() < bounds.Right() {
+		if last.Equal(bounds) {
+			// If the bounds are unchanged, that means either that
+			// the object does not exist in the packfile, or the
+			// fanout table is corrupt.
+			//
+			// Either way, we won't be able to find the object.
+			// Return immediately to prevent infinite looping.
+			return nil, nil
+		}
+		last = bounds
+
+		// Find the midpoint between the upper and lower bounds.
+		mid := (bounds.Left() + bounds.Right()) / 2
+
+		// Search for the given object at that midpoint.
+		entry, cmp, err := i.version.Search(i, name, mid)
+		if err != nil {
+			return nil, err
+		}
+
+		if cmp == 0 {
+			// If "cmp" is zero, that means the object at that index
+			// "at" had a SHA equal to the one given by name, and we
+			// are done.
+			return entry, nil
+		} else if cmp < 0 {
+			// If the comparison is less than 0, we searched past
+			// the desired object, so limit the upper bound of the
+			// search to the midpoint.
+			bounds = bounds.WithRight(mid)
+		} else if cmp > 0 {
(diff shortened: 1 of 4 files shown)


git/odb/pack/index.go
```diff
@@ -28,8 +28,97 @@ func (i *Index) Count() int {
...
+			// fanout table is corrupt.
+			//
+			// Either way, we won't be able to find the object.
+			// Return immediately to prevent infinite looping.
+			return nil, nil
+		}
+		last = bounds
+
+		// Find the midpoint between the upper and lower bounds.
+		mid := (bounds.Left() + bounds.Right()) / 2
```
peff: I don't know how Go handles integer overflow, but this is subject to overflow when the size of the bounds approaches the capacity of your left/right integer types. One overflow-safe way is `bounds.Left() + (bounds.Right() - bounds.Left()) / 2`. Since you're using 64-bit integers it seems pretty unlikely in practice, though, and you can probably ignore it.

peff:
This (and the PRs leading up to it) look OK from my cursory look. The nice thing about this area is that most bugs will be utterly obvious when you try to run them on a real packfile. I see most of your tests are unit tests; I don't know if you have tests that run against a particular fixture created by Git.

If you do and you want to test each index version individually, you should be able to do:

```
... create a repo or use some fixture ...
# everything in one pack for simplicity (though probably you'd want to test the multi-pack behavior eventually, too)
git repack -ad
pack=$(ls .git/objects/pack/*.pack)
git index-pack --index-version=1 $pack
... do some lfs stuff ...
git index-pack --index-version=2 $pack
... same lfs stuff ...
```

ttaylorr:
@peff thanks for taking the time to look everything over, and passing along some shell-fu. Here's something:

Packing all of the objects in git-lfs/git-lfs with `--index-version=1`, and --index-version=2`:

```
~/g/verify-pack $ PACK=~/github/git-lfs/.git/objects/pack/pack-5b2c32d3c1d5708b8edc9abfa5f34c7e2e431c2f.idx
~/g/verify-pack $ git verify-pack -v "$PACK" | cut -d ' ' -f 1 | grep -E "^[[:alnum:]]{40}$" | go run cmd/verify-pack/main.go "$PACK"
# ...
OK
```

Using: https://github.com/git-lfs/verify-pack.

PR #2429, opened 2017-07-20
Allow using custom transfer agents directly

aleb:
Some custom transfer agents figure out everything by themselves and don't need
any authentication or transfer details from the API server.

Added the `lfs.standalonetransferagent` config option for specifying which
custom agent should be used directly.

### DIFF
--- a/tq/transfer_queue.go
+++ b/tq/transfer_queue.go
@@ -283,23 +283,39 @@ func (q *TransferQueue) enqueueAndCollectRetriesFor(batch batch) (batch, error) 
 	tracerx.Printf("tq: sending batch of size %d", len(batch))
 
 	q.meter.Pause()
-	bRes, err := Batch(q.manifest, q.direction, q.remote, batch.ToTransfers())
-	if err != nil {
-		// If there was an error making the batch API call, mark all of
-		// the objects for retry, and return them along with the error
-		// that was encountered. If any of the objects couldn't be
-		// retried, they will be marked as failed.
+	var bRes *BatchResponse
+	if q.manifest.standaloneTransferAgent != "" {
+		// Trust the external transfer agent can do everything by itself.
+		objects := make([]*Transfer, 0, len(batch))
 		for _, t := range batch {
-			if q.canRetryObject(t.Oid, err) {
-				q.rc.Increment(t.Oid)
-
-				next = append(next, t)
-			} else {
-				q.wait.Done()
-			}
+			objects = append(objects, &Transfer{Oid: t.Oid, Size: t.Size, Path: t.Path})
 		}
+		bRes = &BatchResponse{
+			Objects:             objects,
+			TransferAdapterName: q.manifest.standaloneTransferAgent,
+		}
+	} else {
+		// Query the Git LFS server for what transfer method to use and
+		// details such as URLs, authentication, etc.
+		var err error
+		bRes, err = Batch(q.manifest, q.direction, q.remote, batch.ToTransfers())
+		if err != nil {
+			// If there was an error making the batch API call, mark all of
+			// the objects for retry, and return them along with the error
+			// that was encountered. If any of the objects couldn't be
+			// retried, they will be marked as failed.
+			for _, t := range batch {
+				if q.canRetryObject(t.Oid, err) {
+					q.rc.Increment(t.Oid)
 
-		return next, err
+					next = append(next, t)
+				} else {
+					q.wait.Done()
+				}
+			}
+
+			return next, err
+		}
 	}
 
 	if len(bRes.Objects) == 0 {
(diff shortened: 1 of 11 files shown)


aleb:
If this is the right approach, I can add some tests.

tq/transfer_queue.go
```diff
@@ -283,23 +283,44 @@ func (q *TransferQueue) enqueueAndCollectRetriesFor(batch batch) (batch, error)
...
-		for _, t := range batch {
-			if q.canRetryObject(t.Oid, err) {
-				q.rc.Increment(t.Oid)
-
-				next = append(next, t)
-			} else {
-				q.wait.Done()
-			}
+	var bRes *BatchResponse
+	if q.manifest.standaloneTransferAgent != "" {
```
technoweenie: What do you think about moving this check into the `Batch()` function? That way `standaloneTransferAgent` just works for all `tq.Batch()` callers.

tq/transfer_queue.go
```diff
@@ -283,23 +283,44 @@ func (q *TransferQueue) enqueueAndCollectRetriesFor(batch batch) (batch, error)
...
+		// Trust the external transfer agent can do everything by itself.
+		objects := make([]*Transfer, 0, len(batch))
+		for _, object := range batch {
+			objects = append(objects, &Transfer{
+				Name: object.Name,
```
technoweenie: `Name` isn't used in the custom transfer adapter protocol and can probably be omitted.

tq/transfer_queue.go
```diff
@@ -283,23 +283,44 @@ func (q *TransferQueue) enqueueAndCollectRetriesFor(batch batch) (batch, error)
...
+				Oid:  object.Oid,
+				Size: object.Size,
+				Path: object.Path,
```
technoweenie: I wonder if `batch.ToTransfers()` should just pass the `Path` along too, so it could be used here.

tq/transfer_queue.go
aleb: Indeed this could be moved in the `Batch()` function, but I think it might be better to get rid of that method altogether. All it does is to check for len(objects) which is already checked in `tq.Batch()`, and make a very simple call .. and on top of that, it's used in a single place. 

What other callers do you think there would be?

tq/transfer_queue.go
aleb: You are right, the name is not used in the protocol. Removed.

tq/transfer_queue.go
aleb: Looks much slimmer now.

tq/transfer_queue.go
technoweenie: `tq.Batch()` is currently also used in [`test/git-lfs-test-server-api/main.go`](https://github.com/git-lfs/git-lfs/blob/00c2f1ef2c7387725d64b5efcd1e24c7254074e8/test/git-lfs-test-server-api/main.go#L281-L292). I'm open to refactoring the `tq` package, but I think it should happen separately from this PR.

aleb:
I pushed an update, the test passes but not if I shut down the lfs server:
```
  # make sure git-lfs does not interact with the server when a standalone
  # transfer agent is set.
  curl -s "$(cat "$LFS_URL_FILE")/shutdown"
```
The error is this:
```
    17:08:06.236406 git.c:322               trace: built-in: git 'push' 'origin' 'master'
    17:08:06.237605 run-command.c:626       trace: run_command: 'git-remote-http' 'origin' 'http://127.0.0.1:61154/test-custom-transfer-standalone'
    fatal: unable to access 'http://127.0.0.1:61154/test-custom-transfer-standalone/': Failed to connect to 127.0.0.1 port 61154: Connection refused
```
Where should I look to find where it tries to connect to the server?

aleb:
This is mergeable IMO. Please have a look.

aleb:
I figured that I should not stop the server used by the test. Clarified a bit the documentation about the setup/shutdown methods.

ttaylorr:
@aleb this looks great, and thanks for your contribution to LFS! I've schedule this to be released in 2.3.0.

aleb:
There is a minor problem when using a standalone custom-transfer agent. When pushing a branch to a remote, the git-lfs fails because it tries to connect to the inexistent remote server through HTTPS to check some locking thing. Before dying it suggests to run something like:
```
git config --replace-all lfs.https://SERVER/.locksverify false 
```
This solves the problem, but I was wondering if the check should even be made when using a standalone custom-transfer agent. Does it makes sense to deactivate it when `lfs.standalonetransferagent` is set?

ttaylorr:
> Does it makes sense to deactivate it when `lfs.standalonetransferagent` is set?

I think it makes sense to deactivate this. The code you're looking for is here:

https://github.com/git-lfs/git-lfs/blob/76e82ac07c27710c01c9a7ca7b295800f46eb34e/commands/uploader.go#L124-L127

PR #2443, opened 2017-07-26
script/packagecloud: release LFS on fedora/26

rudineirk:
This pull-request adds fedora/26 to the list of targets for Packagecloud, which will make all future builds available on the latest release of Fedora.


ttaylorr:
@rudineirk thanks for your contribution!

PR #2453, opened 2017-07-30
add function to invoke Git with disabled LFS filters

larsxschneider:
Git LFS invokes Git for various actions internally. A new Git instance is started with the default Git config and therefore might invoke another Git LFS process via the filter machinery. I have noticed that in #2439.

Teach Git LFS to invoke Git with disabled LFS filters to prevent unnecessary Git LFS processes.

I recommend to review the changes by commit.

---

**Attention: I haven't performed extensive testing on this, yet.**


larsxschneider:
_Unrelated to this PR:_
It looks like as if GitHub orders commits by date as of late.  aa4c822 is the actual last commit of this PR but GitHub shows me c7c99eb as the last commit.

larsxschneider:
On my local machine all tests pass with OK. However, the final exit code is 1. 🤔 
```
$ ./script/cibuild
commands/mancontent_gen.go
?   	github.com/git-lfs/git-lfs	[no test files]
ok  	github.com/git-lfs/git-lfs/commands	0.022s
ok  	github.com/git-lfs/git-lfs/config	0.020s
?   	github.com/git-lfs/git-lfs/docs/man	[no test files]
ok  	github.com/git-lfs/git-lfs/errors	0.019s
ok  	github.com/git-lfs/git-lfs/filepathfilter	0.031s
ok  	github.com/git-lfs/git-lfs/git	3.953s
ok  	github.com/git-lfs/git-lfs/git/githistory	0.366s
ok  	github.com/git-lfs/git-lfs/git/githistory/log	0.055s
ok  	github.com/git-lfs/git-lfs/git/odb	0.050s
ok  	github.com/git-lfs/git-lfs/git/odb/pack	0.020s
ok  	github.com/git-lfs/git-lfs/lfs	1.880s
ok  	github.com/git-lfs/git-lfs/lfsapi	2.441s
?   	github.com/git-lfs/git-lfs/localstorage	[no test files]
ok  	github.com/git-lfs/git-lfs/locking	0.552s
ok  	github.com/git-lfs/git-lfs/progress	0.032s
?   	github.com/git-lfs/git-lfs/script	[no test files]
?   	github.com/git-lfs/git-lfs/script/genmakefile	[no test files]
?   	github.com/git-lfs/git-lfs/subprocess	[no test files]
?   	github.com/git-lfs/git-lfs/test	[no test files]
?   	github.com/git-lfs/git-lfs/test/git-lfs-test-server-api	[no test files]
ok  	github.com/git-lfs/git-lfs/tools	0.177s
ok  	github.com/git-lfs/git-lfs/tools/humanize	0.030s
ok  	github.com/git-lfs/git-lfs/tools/kv	0.023s
ok  	github.com/git-lfs/git-lfs/tq	0.287s
ok  	github.com/git-lfs/git-lfs/vendor/github.com/ThomsonReutersEikon/go-ntlm/ntlm	0.017s
ok  	github.com/git-lfs/git-lfs/vendor/github.com/ThomsonReutersEikon/go-ntlm/ntlm/md4	0.010s
ok  	github.com/git-lfs/git-lfs/vendor/github.com/bgentry/go-netrc/netrc	0.011s
?   	github.com/git-lfs/git-lfs/vendor/github.com/inconshreveable/mousetrap	[no test files]
ok  	github.com/git-lfs/git-lfs/vendor/github.com/kr/pretty	0.011s
?   	github.com/git-lfs/git-lfs/vendor/github.com/rubyist/tracerx	[no test files]
ok  	github.com/git-lfs/git-lfs/vendor/github.com/spf13/cobra	0.028s
ok  	github.com/git-lfs/git-lfs/vendor/github.com/spf13/pflag	0.017s
ok  	github.com/git-lfs/git-lfs/vendor/github.com/xeipuuv/gojsonpointer	0.007s
ok  	github.com/git-lfs/git-lfs/vendor/golang.org/x/exp/mmap	0.010s
ok  	github.com/git-lfs/git-lfs/git	3.839s
Using go1.8.3
Git LFS: /Users/lars/Code/go-Workspace/src/github.com/git-lfs/git-lfs/bin/git-lfs
git-lfs/2.3.0-pre (GitHub; darwin amd64; go 1.8.3; git 8bb86141)
git version 2.13.2
Git LFS initialized.

HOME: /Users/lars/Code/go-Workspace/src/github.com/git-lfs/git-lfs/test/remote/home
TMP: /private/var/folders/2w/khgj5gss4nz9nclyxmx136pm0000gn/T/git-lfs_TEMP.XXXXXX.XJgiVWPr
CREDS: /Users/lars/Code/go-Workspace/src/github.com/git-lfs/git-lfs/test/remote/creds
lfstest-gitserver:
  LFSTEST_URL=/Users/lars/Code/go-Workspace/src/github.com/git-lfs/git-lfs/test/remote/url
  LFSTEST_SSL_URL=/Users/lars/Code/go-Workspace/src/github.com/git-lfs/git-lfs/test/remote/sslurl
  LFSTEST_CLIENT_CERT_URL=/Users/lars/Code/go-Workspace/src/github.com/git-lfs/git-lfs/test/remote/clientcerturl (https://127.0.0.1:63317)
  LFSTEST_CERT=/Users/lars/Code/go-Workspace/src/github.com/git-lfs/git-lfs/test/remote/cert
  LFSTEST_CLIENT_CERT=/Users/lars/Code/go-Workspace/src/github.com/git-lfs/git-lfs/test/remote/client.crt
  LFSTEST_CLIENT_KEY=/Users/lars/Code/go-Workspace/src/github.com/git-lfs/git-lfs/test/remote/client.key
  LFSTEST_DIR=/Users/lars/Code/go-Workspace/src/github.com/git-lfs/git-lfs/test/remote
GIT:
filter.lfs.clean git-lfs clean -- %f
filter.lfs.smudge git-lfs smudge -- %f
filter.lfs.process git-lfs filter-process
filter.lfs.required true
credential.usehttppath true
credential.helper lfstest
user.name Git LFS Tests
user.email git-lfs@example.com

Running this maxprocs 4
test: transfer queue rejects unknown OIDs ...                      OK
test: batch error handling ...                                     OK
test: chunked transfer encoding ...                                OK
test: batch storage upload causes retries ...                      OK
test: batch storage download causes retries ...                    OK
test: checkout ...                                                 OK
test: checkout: outside git repository ...                         OK
test: clean simple file ...                                        OK
test: clean a pointer ...                                          OK
test: clean pseudo pointer ...                                     OK
test: clean pseudo pointer with extra data ...                     OK
test: default config ...                                           OK
test: extension config ...                                         OK
test: url alias config ...                                         OK
test: ambiguous url alias ...                                      OK
test: url alias must be prefix ...                                 OK
test: batch transfer ...                                           OK
test: batch transfers occur in reverse order by size ...           OK
test: batch transfers with ssh endpoint ...                        OK
test: batch transfers with ntlm server ...                         OK
test: attempt private access without credential helper ...         OK
test: commit, delete, then push ...                                OK
test: multiple revs with same OID get pushed once ...              OK
test: env with no remote ...                                       OK
test: env with origin remote ...                                   OK
test: env with multiple remotes ...                                OK
test: env with other remote ...                                    OK
test: env with multiple remotes and lfs.url config ...             OK
test: env with multiple remotes and lfs configs ...                OK
test: env with multiple remotes and lfs url and batch configs ...  OK
test: env with .lfsconfig ...                                      OK
test: env with environment variables ...                           OK
test: env with bare repo ...                                       OK
test: env with multiple ssh remotes ...                            OK
test: env with skip download errors ...                            OK
test: env with extra transfer methods ...                          OK
test: custom-transfer-wrong-path ...                               OK
test: custom-transfer-upload-download ...                          OK
test: ext ...                                                      OK
test: http.<url>.extraHeader ...                                   OK
test: credentials without useHttpPath, with bad path password ...  OK
test: credentials with useHttpPath, with wrong password ...        OK
test: credentials with useHttpPath, with correct password ...      OK
test: git credential ...                                           OK
test: credentials from netrc ...                                   OK
test: credentials from netrc with bad password ...                 OK
test: credentials from lfs.url ...                                 OK
test: credentials from remote.origin.url ...                       OK
test: fetch: setup for include test ...                            OK
test: fetch: include first matching file ...                       OK
test: fetch: include second matching file ...                      OK
test: init fetch unclean paths ...                                 OK
test: fetch unclean paths with include filter in gitconfig ...     OK
test: fetch unclean paths with exclude filter in gitconfig ...     OK
test: fetch unclean paths with include filter in cli ...           OK
test: fetch unclean paths with exclude filter in cli ...           OK
test: expired action (absolute time) ...                           OK
test: expired action (relative time) ...                           OK
test: expired action (both time) ...                               OK
test: ssh expired (absolute time) ...                              OK
test: ssh expired (relative time) ...                              OK
test: ssh expired (both time) ...                                  OK
test: filter-branch (git-lfs/git-lfs#1773) ...                     OK
test: filter process: checking out a branch ...                    OK
test: filter process: adding a file ...                            OK
test: filter process: add a file with 1024 bytes ...               OK
test: init fetch-recent ...                                        OK
test: fetch-recent normal ...                                      OK
test: fetch-recent commits ...                                     OK
test: fetch-recent days ...                                        OK
test: fetch-recent older commits ...                               OK
test: fetch-recent remote branch ...                               OK
test: fetch-recent remote refs ...                                 OK
test: fsck default ...                                             OK
test: fsck dry run ...                                             OK
test: fsck: outside git repository ...                             OK
skip: test/test-install-custom-hooks-path-unsupported.sh (git version > 2.9.0)
test: install with supported core.hooksPath ...                    OK
test: happy path ...                                               OK
test: clears local temp objects ...                                OK
test: clone ...                                                    OK
test: cloneSSL ...                                                 OK
test: clone ClientCert ...                                         OK
test: clone with flags ...                                         OK
test: clone (with include/exclude args) ...                        OK
test: clone (with .lfsconfig) ...                                  OK
test: clone with submodules ...                                    OK
test: clone in current directory ...                               OK
test: install again ...                                            OK
test: install with old (non-upgradeable) settings ...              OK
test: install with upgradeable settings ...                        OK
test: install updates repo hooks ...                               OK
test: install outside repository directory ...                     OK
test: install --skip-smudge ...                                    OK
test: install --local ...                                          OK
test: install --local outside repository ...                       OK
test: install in directory without access to .git/lfs ...          OK
test: install in repo without changing hooks ...                   OK
test: logs ...                                                     OK
test: ls-files ...                                                 OK
test: ls-files: outside git repository ...                         OK
test: ls-files: with zero files ...                                OK
test: ls-files: show duplicate files ...                           OK
test: ls-files: show duplicate files with long OID ...             OK
test: malformed pointers ...                                       OK
test: empty pointers ...                                           OK
<no output for test/test-migrate-fixtures.sh>
test: creating a lock ...                                          OK
test: create lock with server using client cert ...                OK
test: creating a lock (with output) ...                            OK
test: locking a previously locked file ...                         OK
test: locking a directory ...                                      OK
test: locking a nested file ...                                    OK
test: list a single lock ...                                       OK
test: list a single lock (--json) ...                              OK
test: list locks with a limit ...                                  OK
test: list locks with pagination ...                               OK
test: cached locks ...                                             OK
test: download authenticated object ...                            OK
test: pointer --file --stdin ...                                   OK
test: pointer --file --stdin mismatch ...                          OK
test: pointer --stdin ...                                          OK
test: pointer --stdin without stdin ...                            OK
test: pointer --stdin with bad pointer ...                         OK
test: pointer --file --pointer mismatch ...                        OK
test: pointer --file --pointer ...                                 OK
test: pointer --pointer ...                                        OK
test: pointer missing --pointer ...                                OK
test: migrate info (default branch) ...                            OK
test: migrate info (given branch) ...                              OK
test: migrate info (default branch with filter) ...                OK
test: migrate info (given branch with filter) ...                  OK
test: migrate info (default branch, exclude remote refs) ...       OK
test: migrate info (given branch, exclude remote refs) ...         OK
test: migrate info (include/exclude ref) ...                       OK
test: migrate info (include/exclude ref with filter) ...           OK
test: migrate info (above threshold) ...                           OK
test: migrate info (above threshold, top) ...                      OK
test: migrate info (given unit) ...                                OK
test: migrate info (doesn't show empty info entries) ...           OK
test: migrate info (empty set) ...                                 OK
test: post-commit ...                                              OK
test: init for fetch tests ...                                     OK
test: fetch ...                                                    OK
test: fetch with remote ...                                        OK
test: fetch with remote and branches ...                           OK
test: fetch with master commit sha1 ...                            OK
test: fetch with newbranch commit sha1 ...                         OK
test: fetch with include filters in gitconfig ...                  OK
test: fetch with exclude filters in gitconfig ...                  OK
test: fetch with include/exclude filters in gitconfig ...          OK
test: fetch with include filter in cli ...                         OK
test: fetch with exclude filter in cli ...                         OK
test: fetch with include/exclude filters in cli ...                OK
test: fetch with include filter overriding exclude filter ...      OK
test: fetch with missing object ...                                OK
test: fetch-all ...                                                OK
test: fetch: outside git repository ...                            OK
test: fetch with no origin remote ...                              OK
test: fetch --prune ...                                            OK
test: fetch raw remote url ...                                     OK
test: fetch with invalid remote ...                                OK
test: post-checkout ...                                            OK
test: migrate import (default branch) ...                          OK
test: migrate import (given branch) ...                            OK
test: migrate import (default branch with filter) ...              OK
test: migrate import (given branch with filter) ...                OK
test: migrate import (default branch, exclude remote refs) ...     OK
test: migrate import (given branch, exclude remote refs) ...       OK
test: migrate import (include/exclude ref) ...                     OK
test: migrate import (include/exclude ref with filter) ...         OK
test: migrate import (existing .gitattributes) ...                 OK
test: migrate import (bare repository) ...                         OK
test: post-merge ...                                               OK
test: prune worktree ...                                           OK
test: progress meter displays positive progress ...                OK
test: push: upload to bad dns ...                                  OK
test: pull ...                                                     OK
test: pull with raw remote url ...                                 OK
test: pull: with missing object ...                                OK
test: pull: outside git repository ...                             OK
test: push a file with the same name as a branch ...               OK
test: push missing objects ...                                     OK
test: push: upload file with storage 403 ...                       OK
test: push: upload file with storage 404 ...                       OK
test: push: upload file with storage 410 ...                       OK
test: push: upload file with storage 422 ...                       OK
test: push: upload file with storage 500 ...                       OK
test: push: upload file with storage 503 ...                       OK
test: push: upload file with api 403 ...                           OK
test: push: upload file with api 404 ...                           OK
test: push: upload file with api 410 ...                           OK
test: push: upload file with api 422 ...                           OK
test: push: upload file with api 500 ...                           OK
test: clone with reference ...                                     OK
test: fetch from clone reference ...                               OK
test: resume-http-range ...                                        OK
test: resume-http-range-fallback ...                               OK
test: tus-upload-uninterrupted ...                                 OK
test: tus-upload-interrupted-resume ...                            OK
test: push ...                                                     OK
test: push --all (no ref args) ...                                 OK
test: push --all (1 ref arg) ...                                   OK
test: push --all (multiple ref args) ...                           OK
test: push --all (ref with deleted files) ...                      OK
test: push object id(s) ...                                        OK
test: push modified files ...                                      OK
test: push with invalid remote ...                                 OK
test: push ambiguous branch name ...                               OK
test: push (retry with expired actions) ...                        OK
test: push to raw remote url ...                                   OK
test: push (with invalid object size) ...                          OK
test: push with deprecated _links ...                              OK
test: push with missing objects (lfs.allowincompletepush) ...      OK
test: pre-push ...                                                 OK
test: pre-push dry-run ...                                         OK
test: pre-push 307 redirects ...                                   OK
test: pre-push with existing file ...                              OK
test: pre-push with existing pointer ...                           OK
test: pre-push with missing pointer not on server ...              OK
test: pre-push with missing pointer which is on server ...         OK
test: pre-push with missing and present pointers ...               OK
test: pre-push multiple branches ...                               OK
test: pre-push with bad remote ...                                 OK
test: pre-push unfetched deleted remote branch & server GC ...     OK
test: pre-push delete branch ...                                   OK
test: pre-push with our lock ...                                   OK
test: pre-push with their lock on lfs file ...                     OK
test: pre-push with their lock on non-lfs lockable file ...        OK
test: pre-push locks verify 5xx with verification enabled ...      OK
test: pre-push disable locks verify on exact url ...               OK
test: pre-push disable locks verify on partial url ...             OK
test: pre-push locks verify 5xx with verification unset ...        OK
test: pre-push locks verify 501 with verification enabled ...      OK
test: pre-push locks verify 501 with verification disabled ...     OK
test: pre-push locks verify 501 with verification unset ...        OK
test: pre-push locks verify 200 ...                                OK
test: pre-push locks verify 403 with verification enabled ...      OK
test: pre-push locks verify 403 with verification disabled ...     OK
test: pre-push locks verify 403 with verification unset ...        OK
test: ssh with proxy command in lfs.url ...                        OK
test: prune unreferenced and old ...                               OK
test: prune keep unpushed ...                                      OK
test: prune keep recent ...                                        OK
test: prune remote tests ...                                       OK
test: prune verify ...                                             OK
test: prune verify large numbers of refs ...                       OK
test: smudge ...                                                   OK
test: smudge with temp file ...                                    OK
test: smudge with invalid pointer ...                              OK
test: smudge include/exclude ...                                   OK
test: smudge with skip ...                                         OK
test: smudge clone with include/exclude ...                        OK
test: smudge skip download failure ...                             OK
test: submodule env with .lfsconfig ...                            OK
test: submodule update --init --remote with .lfsconfig ...         OK
test: status ...                                                   OK
test: status --porcelain ...                                       OK
test: status --json ...                                            OK
test: status: outside git repository ...                           OK
test: status - before initial commit ...                           OK
test: status shows multiple files with identical contents ...      OK
test: status shows multiple copies of partially staged files ...   OK
test: status: LFS to LFS change ...                                OK
test: status: Git to LFS change ...                                OK
test: status: Git to LFS conversion ...                            OK
test: submodule local git dir ...                                  OK
test: submodule env ...                                            OK
test: track (--no-modify-attrs) ...                                OK
test: uninstall outside repository ...                             OK
test: uninstall outside repository without access to .git/lfs ...  OK
test: uninstall inside repository with default pre-push hook ...   OK
test: uninstall inside repository without lfs pre-push hook ...    OK
test: uninstall hooks inside repository ...                        OK
test: uninstall --local ...                                        OK
test: untrack ...                                                  OK
test: untrack outside git repo ...                                 OK
test: untrack removes escape sequences ...                         OK
test: track ...                                                    OK
test: track --verbose ...                                          OK
test: track --dry-run ...                                          OK
test: track directory ...                                          OK
test: track without trailing linebreak ...                         OK
test: track with existing crlf ...                                 OK
test: track with autocrlf=true ...                                 OK
test: track with autocrlf=input ...                                OK
test: track outside git repo ...                                   OK
test: track representation ...                                     OK
test: track absolute ...                                           OK
test: track in gitDir ...                                          OK
test: track in symlinked dir ...                                   OK
test: track blocklisted files by name ...                          OK
test: track blocklisted files with glob ...                        OK
test: track lockable ...                                           OK
test: track lockable read-only/read-write ...                      OK
test: track escaped pattern ...                                    OK
test: push unusually named files ...                               OK
test: pull unusually named files ...                               OK
test: update ...                                                   OK
test: update with leading spaces ...                               OK
test: update lfs.{url}.access ...                                  OK
test: update: outside git repository ...                           OK
test: git worktree ...                                             OK
test: track files using wildcard pattern with leading slash ...    OK
test: track files using filename pattern with leading slash ...    OK
test: push zero len file ...                                       OK
test: pull zero len file ...                                       OK
test: verify with retries ...                                      OK
test: verify with retries (success without retry) ...              OK
test: verify with retries (insufficient retries) ...               OK
test: verify with retries (bad .gitconfig) ...                     OK
exit status 1
```

git/git.go
```diff
@@ -490,13 +490,45 @@ func (c *gitConfig) IsGitVersionAtLeast(ver string) bool {
...
+	return append([]string{
+		"-c", fmt.Sprintf("filter.lfs.smudge=%v", filterOverride),
+		"-c", "filter.lfs.clean=",
+		"-c", "filter.lfs.process=",
+		"-c", "filter.lfs.required=false",
+	}, args...)
+}
+
+// Invoke Git with disabled LFS filters
+func Exec(args ...string) *subprocess.Cmd {
```
ttaylorr: I would like to unexport this and `SimpleExec` below, so that outside callers cannot `git.Exec(...)`. This type of code allows arbitrary usage of Git, which I would like to prevent. If we find ourselves needing a new `git` process that isn't already covered in this package, let's add it here where we can track it using static analysis tools, etc., instead of searching through the code.

git/git.go
```diff
@@ -490,13 +490,45 @@ func (c *gitConfig) IsGitVersionAtLeast(ver string) bool {
...
+	return subprocess.ExecCommand("git", PrependConfigToDisableLFS(args...)...)
+}
+
+// Invoke Git with disabled LFS filters
+func SimpleExec(args ...string) (string, error) {
```
ttaylorr: Same note here as above.

git/git.go
```diff
@@ -490,13 +490,45 @@ func (c *gitConfig) IsGitVersionAtLeast(ver string) bool {
 	return IsVersionAtLeast(gitver, ver)
 }
 
+// Prepend Git config instructions to disable Git LFS filter
+func PrependConfigToDisableLFS(args ...string) []string {
```
ttaylorr: Do outside callers need this? If not, I'd like to un-export this.

ttaylorr: Additionally, I'm not sure that disabling LFS in all uses of `Exec()`, and `SimpleExec()` is safe. If you want to keep this function, I'd recommend that we instead change the calls from:

```go
// Exec runs Git, disabling LFS, with the given arguments.

func Something() ([]byte, error) {
        return Exec("something", ...)
}
```

to:

```diff
@@ -1,5 +1,5 @@
-// Exec runs Git, disabling LFS, with the given arguments.
+// Exec runs Git with the given arguments.
 
func Something() ([]byte, error) {
-        return Exec("something", ...)
+        return Exec(PrependConfigToDisableLFS("something")...)
}
```

larsxschneider: No outside caller needs this. Un-export is done by a lower case function name?

ttaylorr: Yes!

ttaylorr:
> On my local machine all tests pass with OK. However, the final exit code is 1. 🤔

I've seen this intermittently before; usually I think this has to do with [shutting down the Git server we use during the tests][1].

[1]: https://github.com/git-lfs/git-lfs/blob/v2.2.1/test/testhelpers.sh#L517-L521

git/git.go
larsxschneider: I am not sure I understand what you're saying here.
Would you prefer it if I don't use the additional Git Exec wrapper functions?

In case of config changes disabling LFS wouldn't be safe. But everything else should be, no?

git/git.go
larsxschneider: You mean these kind of calls:
https://github.com/larsxschneider/git-lfs/blob/7418ab2974609e0fd2ec0d4e131126595feb31ed/commands/command_pointer.go#L120
... should go into the `git` package?

ttaylorr: Yes, ideally I'd like that to turn into something like:

```go
h, err := git.HashObject(someReader)
```

Where:

1. The `exec("hash-object", ...)` would live inside of the `git` package, and
2. It would be exposed with a function exported from package `git` called `HashObject()`.

git/git.go
ttaylorr: I am in favor of the wrapper functions, with two recommendations:

1. Un-export `Exec()` and `SimpleExec()`.
1. Make `exec()` and `simpleExec()` execute their arguments literally, without implicitly prepending the LFS-disabling arguments to Git. If a caller wants to run a Git command that doesn't then go back and run LFS, they should call `exec` like: `exec(disableLfs(args...))`.

larsxschneider:
> I've seen this intermittently before; usually I think this has to do with shutting down the Git server we use during the tests.

This is very consistent on my local machine and on the TravisCI/CircleCI builds.
E.g. https://circleci.com/gh/git-lfs/git-lfs/1973

(only builds with "$GIT_EARLIEST_SUPPORTED_VERSION" print actually failing tests)

git/git.go
larsxschneider: re 2) My thinking was that "disablingLFS" should be the "default" case, no? Although it's a bit verbose I agree with the explicit prepending!

ttaylorr: > re 2) My thinking was that "disablingLFS" should be the "default" case, no?

I would prefer that this not be the default case. I think that this approach will make it clearer that by calling `exec()`/`simpleExec()` you have to explicitly opt-in to not using LFS.

I agree though that this approach makes it harder for new callers to know when to add this and when not to. What do you think about `execWithLFS` and `execWithoutLfs` as another alternative?

larsxschneider:
@ttaylorr [subprocess/cmd](https://github.com/git-lfs/git-lfs/blob/master/subprocess/cmd.go#L11-L15) and [lfs/gitscanner_cmd](https://github.com/git-lfs/git-lfs/blame/master/lfs/gitscanner_cmd.go#L12-L17) look pretty similar. Would it make sense to extend `subprocess` and use it the `lfs` package? Or did you you introduce the duplication to keep the `lfs` package as independent as possible?

ttaylorr:
> @ttaylorr [subprocess/cmd](https://github.com/git-lfs/git-lfs/blob/master/subprocess/cmd.go#L11-L15) and [lfs/gitscanner_cmd](https://github.com/git-lfs/git-lfs/blame/master/lfs/gitscanner_cmd.go#L12-L17) look pretty similar. Would it make sense to extend `subprocess` and use it the `lfs` package?

If `lfs` could import `subprocess` I think that would be a good idea 👍 .

larsxschneider:
@ttaylorr I am pretty happy with the current state. If you have time, please give is thorough review 😄

larsxschneider:
`test: pre-push with their lock on non-lfs lockable file` seems to be flaky 😢

git/git.go
```diff
@@ -84,19 +85,116 @@ type CommitSummary struct {
...
+		return nil, err
+	}
+	if err = cmd.Stdin.Close(); err != nil {
+		return nil, err
+	}
+
+	return bufio.NewScanner(cmd.Stdout), nil
+}
+
+func HashObject(by []byte) (string, error) {
```
technoweenie: What do you think about changing this to accept a reader? This will allow it to take larger objects without being forced to read it into memory first. Then you can get rid of the `bytes.NewReader()` in here.

I realize that the usage in `commands/command_pointer.go` will be just fine buffering in memory. But, I think it makes sense to change this function since it's now an exported `git` package function.

git/git.go
```diff
@@ -360,9 +458,16 @@ func DefaultRemote() (string, error) {
 	return "", errors.New("Unable to pick default remote, too ambiguous")
 }
 
-func UpdateIndex(file string) error {
-	_, err := subprocess.SimpleExec("git", "update-index", "-q", "--refresh", file)
-	return err
+func StartUpdateIndexFromStdin(output *bytes.Buffer) (io.WriteCloser, error) {
```
technoweenie: In the same vein as my `git.HashObject()` comment, what do you think about changing this to accept an `io.Writer`? It'll give the caller some flexibility in how the output is captured.

larsxschneider: Both done! Thanks for the suggestion 👍

Issue #2434, opened 2017-07-24
Git on Windows client corrupts files > 4Gb


obe1line:
When cloning a file larger than 4Gbyte from a BitBucket server repository (LFS enabled), the file is not reconstructed correctly E.g. a 6Gb file results in a 700Mb file. The lfs/objects folder contains the correct file however.

Steps to reproduce:

Server:  Basic install of Debian with Bitbucket Server 4.6, Git 2.13 (64-bit)
Client 1:  Ubuntu 16.02 64 bit, Git 2.13 and Git-LFS 2.2.1 (both 64-bit)
Client 2:  Windows 64 bit (2012), Git 2.13 and Git-LFS 2.2.1 (both 64-bit)

1. Create a repository on the server with LFS support enabled
2. Clone the repository (git clone ssh://git@server:7999/tst/test-git.git)
3. git track '*.iso' , commit and push to the remote
4. add an iso file larger than 4Gb (I used Visual Studio 2013 Update 4 which is 5.82Gb), commit and push to the remote
5. Clone the repository into a clean folder

Client 1 works correctly
Client 2 pulls down the file correctly into .git/lfs/objects/aa/6d/aa6d2a8e9acbb78895b3d2c6ae3cb0db737344aa82b2859d31f757deec931049 but does not reconstruct/copy it correctly to the destination folder (it results in a 1.82Gb file).

Atlassian have looked into the problem and believe that the BitBucket server is working correctly, due to the fact that the correct content is retrieved over the network into the temporary object file (the CRCs match the original file).

Note that no Git configuration has changed (smudge filters etc are the default).

If the file is removed and "git lfs pull" performed, the file is created correctly. Using "git lfs clone" also works.

technoweenie:
First, you can confirm that BitBucket is sending the data correctly by checking the file in `.git/lfs/objects`. If this checks out, the server is doing things correctly:

```sh
$ cd .git/lfs/objects/aa/6d
$ shasum -a 256 aa6d2a8e9acbb78895b3d2c6ae3cb0db737344aa82b2859d31f757deec931049
aa6d2a8e9acbb78895b3d2c6ae3cb0db737344aa82b2859d31f757deec931049  aa6d2a8e9acbb78895b3d2c6ae3cb0db737344aa82b2859d31f757deec931049
```

[...]

PR #2459, opened 2017-07-31
commands: teach smudge, process-filter to warn about Windows size bug

ttaylorr:
This pull request teaches the `git lfs {smudge,filter-process}` commands to warn when smudging blobs larger than 4 GB on Windows.

From https://github.com/git-lfs/git-lfs/issues/2434#issuecomment-317582466:

> [...] I ran this by a Git core dev, and he mentioned that Git on Windows does not support files over 4GB. Unfortunately, this is not something we can fix in LFS. The best we can do now add a warning when large objects are added.

This pull request adds that warning.

Here's a quick breakdown of how things happened:

1. 3a080f9: teach `smudge()` (et. al.) to return `(int64, error)` as `io.Copy` does, to indicate the number of bytes smudged by a call to that function.
2. f6beec6: use the return value from that function to determine and potentially warn about files larger than 4gb being copied into the worktree.

@technoweenie also mentioned:

> [...] I think we should add an early warning in the filter smudge and process code, perhaps linking to a page offering this workaround.

I originally added a link back to https://github.com/git-lfs/git-lfs/issues/2434, but thought that adding a section to the manpage would be sufficient, and perhaps a little cleaner, since the manpages are distributed in the binary and don't require you to be online to view them.

Closes: https://github.com/git-lfs/git-lfs/issues/2434.

---

/cc @git-lfs/core 
/cc https://github.com/git-lfs/git-lfs/issues/2434

### DIFF
--- a/commands/command_filter_process.go
+++ b/commands/command_filter_process.go
@@ -62,7 +62,7 @@ func filterCommand(cmd *cobra.Command, args []string) {
 			err = clean(w, req.Payload, req.Header["pathname"], -1)
 		case "smudge":
 			w = git.NewPktlineWriter(os.Stdout, smudgeFilterBufferCapacity)
-			err = smudge(w, req.Payload, req.Header["pathname"], skip, filter)
+			_, err = smudge(w, req.Payload, req.Header["pathname"], skip, filter)
 		default:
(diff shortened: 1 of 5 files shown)


ThoonsanEain:
#4940

ThoonsanEain:
#4840

Issue #2466, opened 2017-08-02
'delay' filter-process capability


ttaylorr:
# Goal

Improve the speed at which Git LFS performs checkouts with a majority of
uncached objects.

## Background

Late last year, we added support for the new `process` filter, added in Git
v2.11.0, as a means to convert files during check-in/out operations. This
greatly improved the performance of both operations, most notably on Windows,
where the overhead of spawning `git-lfs-clean(1)` or `git-lfs-smudge(1)` for
each file in the operation incurred significant overhead. (See:
https://github.com/git-lfs/git-lfs/issues/1329#issuecomment-231584553).

We noted, however, that this protocol (as is the case with 'smudge' and 'clean'
filters) has an opportunity to improve its performance with uncached items.
Consider the following:

1. Git does a checkout operation after a `fetch` or `clone` on a repository with
   many LFS objects.
2. Git asks LFS for the first remaining object in the checkout.
3. LFS sees that the object is not in the cache, since this is a clone and no
   objects are cached yet.
4. LFS downloads the file in whole, and then responds to Git.
5. If the checkout is done, see step 6, else see step 2.
6. Done

In the above, LFS has to download each large file one at a time before it can
move onto the next. While thinking about ways to improve this, I summarized some
discussion that @larsxschneider and I had been having in
https://github.com/git-lfs/git-lfs/issues/1632#issuecomment-259756814:

> @larsxschneider and I have already begun talking about adding a new capability
> to the protocol to allow the `git-filter-server` (LFS) to "promise" that it
> will convert a file and yield the path on disk where it will dump the result
> of that conversion. [...]
>
> This will allow us to process many transfers concurrently, taking full
> advantage of the proposed new transfer queue design, and yielding even greater
> speeds for the new filter protocol implementation. Perhaps more importantly,
> it will allow us to do normal Git operations (`git clone`, `git push`,
> `git pull`, and etc.) without needing to temporarily disable the smudge filter
> to parallelize the transfers. [...]

## Propositions

Lars has already implemented a spike of what support for what this would look
like in https://github.com/git-lfs/git-lfs/pull/1646.

Since Lars' patch (see: https://github.com/git/git/commit/487fe1ffcd3b3a38477b7e564f235bb7d1b89ecc) is on 'next' and 'pu', I propose that we
expand on his spike above and add support for the 'delay' capability into Git LFS.

## Technical Overview

See: the documentation in [gitattributes.txt][1].

## Non-goals

[...]

PR #2468, opened 2017-08-02
commands,git: teach NegotiateCapabilities() to to return 'supcaps'

ttaylorr:
This pull request teaches the NegotiateCapabilities() method to return a string slice of capabilities supported by the `git-filter-client`.

Previously, we relied on an assumption that the list of required and supported capabilities negotiated between the `git-filter-client` and `git-filter-server` were constant (specifically: `capability=clean` and `capability=smudge`). Since `git` now supports the 'delay' capability, we need to be able to search through the list of capabilities supported by the `git-filter-client` to determine whether or not it is legal for `git lfs filter-process` to return a delayed checkout entry.

This is required work for: https://github.com/git-lfs/git-lfs/issues/2466.

---

/cc @git-lfs/core 
/cc https://github.com/git-lfs/git-lfs/issues/2466

### DIFF
--- a/git/filter_process_scanner.go
+++ b/git/filter_process_scanner.go
@@ -98,25 +98,25 @@ func (o *FilterProcessScanner) Init() error {
 // capabilities given to LFS by the parent, an error will be returned. If there
 // was an error reading or writing capabilities between the two, an error will
 // be returned.
-func (o *FilterProcessScanner) NegotiateCapabilities() error {
+func (o *FilterProcessScanner) NegotiateCapabilities() ([]string, error) {
 	reqCaps := []string{"capability=clean", "capability=smudge"}
 
 	supCaps, err := o.pl.readPacketList()
 	if err != nil {
-		return fmt.Errorf("reading filter-process capabilities failed with %s", err)
+		return nil, fmt.Errorf("reading filter-process capabilities failed with %s", err)
 	}
 	for _, reqCap := range reqCaps {
 		if !isStringInSlice(supCaps, reqCap) {
-			return fmt.Errorf("filter '%s' not supported (your Git supports: %s)", reqCap, supCaps)
+			return nil, fmt.Errorf("filter '%s' not supported (your Git supports: %s)", reqCap, supCaps)
 		}
 	}
 
 	err = o.pl.writePacketList(reqCaps)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


git/filter_process_scanner.go
```diff
@@ -98,25 +98,25 @@ func (o *FilterProcessScanner) Init() error {
 // capabilities given to LFS by the parent, an error will be returned. If there
 // was an error reading or writing capabilities between the two, an error will
 // be returned.
-func (o *FilterProcessScanner) NegotiateCapabilities() error {
```
larsxschneider: The comment above reads a bit strange. What do you mean with `parent`? Is the last sentence correct?

git/filter_process_scanner.go
```diff
@@ -98,25 +98,25 @@ func (o *FilterProcessScanner) Init() error {
...
+		return nil, fmt.Errorf("reading filter-process capabilities failed with %s", err)
 	}
 	for _, reqCap := range reqCaps {
 		if !isStringInSlice(supCaps, reqCap) {
-			return fmt.Errorf("filter '%s' not supported (your Git supports: %s)", reqCap, supCaps)
+			return nil, fmt.Errorf("filter '%s' not supported (your Git supports: %s)", reqCap, supCaps)
 		}
 	}
 
 	err = o.pl.writePacketList(reqCaps)
```
larsxschneider: Git expects you to answer with your supported caps. That means if Git offers you `delay` then you need to answer here with `delay` to be able to use it. I am surprised it worked without this. That might indicate a bug in Git 🤔

git/filter_process_scanner.go
ttaylorr: I probably thought that `parent` was good terminology when I wrote this 😆. I removed it from this godoc in 7864e70.

git/filter_process_scanner.go
ttaylorr: Since we're not sending the `capability=delay` entry here yet, would this change to:

```go
if supCaps.Has("delay") {
        reqCaps = append(reqCaps, "capability=delay")
}

err = o.pl.writePacketList(reqCaps)
```

Since this pull request only sends `capability=clean` and `capability=smudge`, I think this should be fine.

PR #2470, opened 2017-08-02
ci: run builds on 'next' branch of git.git

ttaylorr:
This pull request teaches CircleCI and Travis how to run CI with the version of Git present at the tip of branch 'next'.

This is useful for adding the 'delay' capability to `git lfs filter-process` (see: https://github.com/git-lfs/git-lfs/issues/2466), since the 'delay' capability is not yet in master, or a released version of Git. This will give us a chance to test the 'delay' functionality out in CI before it's released in Git.

Once Git releases a version that contains https://github.com/git/git/commit/2841e8f81cb2820024804b9341577be1d0ce1240, we can revert this pull request.

This is required work for: https://github.com/git-lfs/git-lfs/issues/2466.

---

/cc @git-lfs/core 
/cc https://github.com/git-lfs/git-lfs/issues/2466

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ env:
     - GIT_SOURCE_REPO="https://github.com/git/git.git"
     - GIT_EARLIEST_SUPPORTED_VERSION="v2.0.0
     - GIT_LATEST_SOURCE_BRANCH="master"
+    - GIT_NEXT_SOURCE_BRANCH="next"
 
 matrix:
   fast_finish: true
@@ -23,6 +24,16 @@ matrix:
           git clone $GIT_SOURCE_REPO git-source;
           cd git-source;
           git checkout $GIT_LATEST_SOURCE_BRANCH;
+          make --jobs=2;
+          make install;
+          cd ..;
+    - env: git-latest-next-from-source
+      os: linux
+      before_script:
+        - >
+          git clone $GIT_SOURCE_REPO git-source;
+          cd git-source;
+          git checkout $GIT_NEXT_SOURCE_BRANCH;
           make --jobs=2;
           make install;
           cd ..;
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -4,6 +4,7 @@ machine:
     GIT_SOURCE_REPO: https://github.com/git/git.git
     GIT_EARLIEST_SUPPORTED_VERSION: v2.0.0
     GIT_LATEST_SOURCE_BRANCH: master
+    GIT_NEXT_SOURCE_BRANCH: master
```
larsxschneider: Shouldn't that be `next`?

ttaylorr: @larsxschneider great catch! I fixed that up in: 
https://github.com/git-lfs/git-lfs/pull/2470/commits/a2e6e2d21cb904e09963715693fa64dc367c17e1.

.travis.yml
```diff
@@ -12,6 +12,7 @@ env:
     - GIT_SOURCE_REPO="https://github.com/git/git.git"
     - GIT_EARLIEST_SUPPORTED_VERSION="v2.0.0
     - GIT_LATEST_SOURCE_BRANCH="master"
+    - GIT_NEXT_SOURCE_BRANCH="next"
```
larsxschneider: You could also defined `GIT_LATEST_SOURCE_BRANCH="next"` for the time being.
That's why I called it `LATEST` and not `MASTER` back then 😄

circle.yml
```diff
@@ -4,6 +4,7 @@ machine:
     GIT_SOURCE_REPO: https://github.com/git/git.git
     GIT_EARLIEST_SUPPORTED_VERSION: v2.0.0
     GIT_LATEST_SOURCE_BRANCH: master
+    GIT_NEXT_SOURCE_BRANCH: next
```
larsxschneider: Same here

.travis.yml
ttaylorr: I think running on `master` & `next` for the time being will be fine.

circle.yml
ttaylorr: Is this the same line as https://github.com/git-lfs/git-lfs/pull/2470#discussion_r130989750?

PR #2499, opened 2017-08-10
commands: Disable lock verification when using a standalone custom-tr…

aleb:
…ansfer agent

### DIFF
--- a/commands/uploader.go
+++ b/commands/uploader.go
@@ -378,6 +378,11 @@ var (
 // given "endpoint". If no state has been explicitly set, an "unknown" state
 // will be returned instead.
 func getVerifyStateFor(endpoint lfsapi.Endpoint) verifyState {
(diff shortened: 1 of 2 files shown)


aleb:
Follow-up of https://github.com/git-lfs/git-lfs/pull/2429

commands/uploader.go
```diff
@@ -378,6 +378,11 @@ var (
 // given "endpoint". If no state has been explicitly set, an "unknown" state
 // will be returned instead.
 func getVerifyStateFor(endpoint lfsapi.Endpoint) verifyState {
+	if v, ok := cfg.Git.Get("lfs.standalonetransferagent"); ok && v != "" {
```
ttaylorr: `len(v) > 0` is enough to cover this, since we will never return `("non empty string", false)`.

aleb:
Done. I can move the test to test-push if you think it's better.

PR #2506, opened 2017-08-12
tq: increase default lfs.concurrenttransfers to 8

ttaylorr:
This pull request increases the default number of `lfs.concurrenttransfers` to 8.

---

/cc @git-lfs/core 
/cc #2466

### DIFF
--- a/docs/man/git-lfs-config.5.ronn
+++ b/docs/man/git-lfs-config.5.ronn
@@ -78,7 +78,7 @@ be scoped inside the configuration for a remote.
 
 * `lfs.concurrenttransfers`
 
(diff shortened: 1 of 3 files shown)


larsxschneider:
I have configured it to 10 in the Git config that I distribute to my coworkers. I picked 10 out of a whim. Do you have any greater reasoning for 8? 🤔 😄

ttaylorr:
> Do you have any greater reasoning for 8?

Since 8 is a power of 2 and CPU cores frequently come in powers of 2, the scheduler may have an easier time 8 goroutines than 3. It probably makes no discernible difference, so I don't mind either way.

PR #2507, opened 2017-08-14
commands: make 4gb smudge warning show on all platforms

ttaylorr:
This pull request teaches the `git-lfs-smudge(1)` and `git-lfs-filter-process(1)` commands to display the 4gb object warning on all platforms, not just `runtime.GOOS == "windows"`.

In situations where multiple collaborators are working on a repository among different operating systems, it is useful for a contributor running macOS to know that he/she/they can potentially cause an issue for another collaborator running Windows when they commit a file larger than 4gb. 

---

/cc @git-lfs/core 
/cc https://github.com/git-lfs/git-lfs/pull/2459

### DIFF
--- a/commands/command_smudge.go
+++ b/commands/command_smudge.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"runtime"
 
 	"github.com/git-lfs/git-lfs/errors"
 	"github.com/git-lfs/git-lfs/filepathfilter"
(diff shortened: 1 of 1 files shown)


ttaylorr:
> What do you think?

That sounds like a great idea. I have a patch ready, but I'd rather put it in a separate PR so it makes it easier to generate the changelog. I gave this a shot in: https://github.com/git-lfs/git-lfs/pull/2510.

PR #2510, opened 2017-08-14
commands,lfs: warn on 4gb size conversion during clean

ttaylorr:
This pull request follows up from a comment left by @technoweenie in https://github.com/git-lfs/git-lfs/pull/2507#pullrequestreview-56176535:

> I think including the warning when adding files would be good too. We can allow the commit, but I think it'd be better to warn them sooner (before push) rather than later (during fetch/clone).

To accomplish this, we teach `clean()` (and friends) to return the number of bytes read and written. We then perform the same check using the `possiblyMalformedSmudge` function, and optionally warn.

---

/cc @git-lfs/core

### DIFF
--- a/commands/command_clean.go
+++ b/commands/command_clean.go
@@ -1,7 +1,6 @@
 package commands
 
 import (
-	"fmt"
 	"io"
 	"os"
 
@@ -23,7 +22,7 @@ import (
 //
 // If the object read from "from" is _already_ a clean pointer, then it will be
 // written out verbatim to "to", without trying to make it a pointer again.
-func clean(to io.Writer, from io.Reader, fileName string, fileSize int64) (nr, nw int64, err error) {
+func clean(to io.Writer, from io.Reader, fileName string, fileSize int64) (*lfs.Pointer, error) {
 	var cb progress.CopyCallback
 	var file *os.File
 
@@ -44,7 +43,7 @@ func clean(to io.Writer, from io.Reader, fileName string, fileSize int64) (nr, n
 		}
 	}
 
-	cleaned, nr, err := lfs.PointerClean(from, fileName, fileSize, cb)
+	cleaned, err := lfs.PointerClean(from, fileName, fileSize, cb)
 	if file != nil {
 		file.Close()
 	}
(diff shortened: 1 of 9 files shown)


lfs/pointer_clean.go
```diff
@@ -18,10 +18,10 @@ type cleanedAsset struct {
 	*Pointer
 }
 
-func PointerClean(reader io.Reader, fileName string, fileSize int64, cb progress.CopyCallback) (*cleanedAsset, error) {
+func PointerClean(reader io.Reader, fileName string, fileSize int64, cb progress.CopyCallback) (*cleanedAsset, int64, error) {
```
technoweenie: Does this need to return another `int64`? You should be able to get the actual blob size from `*cleanedAsset`.

commands/command_clean.go
```diff
@@ -22,7 +23,7 @@ import (
 //
 // If the object read from "from" is _already_ a clean pointer, then it will be
 // written out verbatim to "to", without trying to make it a pointer again.
-func clean(to io.Writer, from io.Reader, fileName string, fileSize int64) error {
+func clean(to io.Writer, from io.Reader, fileName string, fileSize int64) (nr, nw int64, err error) {
```
technoweenie: The `nr` and `nw` values aren't very clear to me. It looks like the sizes are the blob size and the pointer size. What about something like `bSize` and `pSize`? Same goes for any `clean()` callers.

commands/command_clean.go
```diff
@@ -97,8 +98,10 @@ func cleanCommand(cmd *cobra.Command, args []string) {
 		fileName = args[0]
 	}
 
-	if err := clean(os.Stdout, os.Stdin, fileName, -1); err != nil {
+	if nr, _, err := clean(os.Stdout, os.Stdin, fileName, -1); err != nil {
 		Error(err.Error())
+	} else if possiblyMalformedSmudge(nr) {
```
technoweenie: I think this func name should be changed since this is not a smudge. What about something like `gitObjectTooBig()`?

ttaylorr: Good call, I changed this to `possiblyMalformedObjectSize()` in 72c15d2.

commands/command_clean.go
ttaylorr: I agree that these are confusing. I rewrote this in a way that those return values aren't needed, instead `commands.clean()` returns an `(*lfs.Pointer, error)`, which we can access the `Size` member of.

ttaylorr:
@technoweenie thanks for the review, I think that this is ready to merge.

PR #2524, opened 2017-08-25
ci: disable integration tests on 'next' branch of git

ttaylorr:
This pull request disables running our integration tests on the 'next' branch of Git.

This change was originally introduced in order to test @larsxschneider's 'delay' patch in #2511, but that has since been merged into the master branch of Git, and is already covered under an existing test runner.

To save time on CI and avoid redundancy, let's remove the extraneous branch.

---

/cc @git-lfs/core 
/cc #2466


larsxschneider:
You could delay the merge until https://github.com/git/git/commit/52f1d62eb44faf569edca360ec9af9ddd4045fe0 hits core git `master` (the patch is not even in `next`, yet). But I don't expect any trouble with this change...

ttaylorr:
@larsxschneider good call, though I don't anticipate there will be problems.

PR #2526, opened 2017-08-25
commands: mark 'git lfs clone' as deprecated

ttaylorr:
This pull request marks the `git-lfs(1) 'clone'` command as deprecated, and puts that deprecation behind a "feature flag".

Since merging #2511, `git clone` is as fast as `git lfs clone` in the majority of cases. To avoid the maintenance cost of the `git lfs clone` command, we mark it as deprecated here, and point callers in the direction of `git clone`. @technoweenie points out a great reason for doing this in https://github.com/git-lfs/git-lfs/issues/2466#issuecomment-322290355:

> > git lfs clone could certainly go 😄 ! Maybe we can mark it as deprecated with a warning on invocation in the 2.x series already?
> 
> I'm all for deprecating it. We should recommend users update Git, but keep the command around for any stragglers. Also, reject PRs for support of any _new_ `clone` flags, since new flags would mean you have a new enough Git with delay support :) 👍

Removing this command also decreases the size of LFS's surface area. Less commands that LFS provides on top of Git make the experience of using LFS more seamless and familiar. I think that's a good thing 👍 .

---

/cc @git-lfs/core 
/cc #2466

### DIFF
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -20,17 +20,12 @@ var (
 	cloneFlags git.CloneFlags
 
 	cloneSkipRepoInstall bool
-
-	// cloneIsDeprecated marks whether or not the clone command is
-	// deprecated. It is false until Git v2.15.0 is released, including the
-	// 'delay' capability.
-	cloneIsDeprecated = false
 )
 
 func cloneCommand(cmd *cobra.Command, args []string) {
 	requireGitVersion()
 
-	if cloneIsDeprecated {
+	if git.Config.IsGitVersionAtLeast("2.15.0") {
 		msg := []string{
 			"WARNING: 'git lfs clone' is deprecated and will not be updated",
 			"          with new flags from 'git clone'",
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -39,7 +39,7 @@ func cloneCommand(cmd *cobra.Command, args []string) {
 			"speeds to 'git lfs clone'.",
 		}
 
-		fmt.Fprintln(os.Stderr, strings.Join(msg, "\n")+"\n")
+		fmt.Fprintln(os.Stderr, strings.Join(msg, "\n"))
 	}
 
 	// We pass all args to git clone
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"fmt"
 	"os"
 	"path"
 	"path/filepath"
@@ -19,10 +20,27 @@ var (
 	cloneFlags git.CloneFlags
 
 	cloneSkipRepoInstall bool
+
+	// cloneIsDeprecated marks whether or not the clone command is
+	// deprecated. It is false until Git v2.15.0 is released, including the
+	// 'delay' capability.
+	cloneIsDeprecated = false
 )
 
 func cloneCommand(cmd *cobra.Command, args []string) {
 	requireGitVersion()
+
+	if cloneIsDeprecated {
+		msg := []string{
+			"WARNING: 'git-lfs(1) clone' is deprecated and will not be updated",
+			"          with new flags from 'git clone'",
+			"",
+			"'git-clone(1)' has been updated in upstream Git to have comparable",
+			"speeds to 'git-lfs(1) clone'.",
+		}
+
+		fmt.Fprintln(os.Stderr, strings.Join(msg, "\n")+"\n")
+	}
 
 	// We pass all args to git clone
 	err := git.CloneWithoutFilters(cloneFlags, args)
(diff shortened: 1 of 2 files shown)


commands/command_clone.go
```diff
@@ -19,11 +20,28 @@ var (
...
 func cloneCommand(cmd *cobra.Command, args []string) {
 	requireGitVersion()
 
+	if cloneIsDeprecated {
+		msg := []string{
+			"WARNING: 'git-lfs(1) clone' is deprecated and will not be updated",
+			"          with new flags from 'git clone'",
+			"",
+			"'git-clone(1)' has been updated in upstream Git to have comparable",
+			"speeds to 'git-lfs(1) clone'.",
```
larsxschneider: I would replace `git-lfs(1)` with just `git-lfs` as I imagine `(1)` would confuse most users (just a gut feeling, though).

ttaylorr: @larsxschneider good call, I removed the `(1)` suffix in 44c230c.

commands/command_clone.go
```diff
@@ -19,11 +20,28 @@ var (
...
+	// deprecated. It is false until Git v2.15.0 is released, including the
+	// 'delay' capability.
+	cloneIsDeprecated = false
 )
 
 func cloneCommand(cmd *cobra.Command, args []string) {
 	requireGitVersion()
 
+	if cloneIsDeprecated {
+		msg := []string{
```
technoweenie: Should this check the git version, only printing the warning if they have 2.15.0+?

ttaylorr: Great idea, I added this (and an integration test for it) in: 
https://github.com/git-lfs/git-lfs/pull/2526/commits/18a3ec2041c898141b4a58ad63c6ee201179a7af

test/test-clone-deprecated.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+. "test/testlib.sh"
+
+ensure_git_version_isnt $VERISON_LOWER "2.15.0"
```
technoweenie: Should be `$VERSION_LOWER`.

ryran:
This makes RHEL/Centos 7 people sad...

larsxschneider:
@ryran: Why is that?

ryran:
@larsxschneider Because RHEL7/Centos7 will never see git v2.14+ ... and when you remove the `lfs-clone` command, users of RHEL7/Centos7 will have to go back to wicked-slow simple `git clone` commands.

That said, I understand -- I shouldn't have complained about it publicly. It's not your fault that RHEL standardizes on a particular version for 10 years. Maybe I'll see if the changes are discrete enough to make a backport request possible. Cheers.

larsxschneider:
Wouldn't be the consequent solution to use an older version of Git LFS, too?

ttaylorr:
> @larsxschneider Because RHEL7/Centos7 will never see git v2.14+ ... and when you remove the `lfs-clone` command, users of RHEL7/Centos7 will have to go back to wicked-slow simple `git clone` commands.

One thing we've been talking about is un-deprecating this command before v2.4.0, since we'll need to support older Git LFS <=> newer Git combinations (if Git ever adds more flags to the `git-clone(1)` builtin). I don't think that would provide any direct benefit on RHEL, but if we un-deprecate this for other platforms, I think i would solve both problems.

@git-lfs/core thoughts?

larsxschneider:
Hmm... can you share some specific reasons/arguments for "un-deprecating"? Why would an older Git LFS not work with a newer Git? I would be very surprised if Git would change the `clone` command in way that would break Git LFS.

technoweenie:
I think the rationale is that users on newer LFS versions and old pre-filter-process Git versions can retain the speed benefits of `git lfs clone`. Another thing is that `git lfs clone` supports the `--include` and `--exclude` flags, allowing you to control what LFS objects are downloaded after cloning. With `git clone`, you can only change this through global git config or a `.lfsconfig` in the repo.

technoweenie:
Ah, I always forget about the `--config` thing. Let's keep the deprecation then :+1:

> TBH I don't buy the "newer LFS", "older Git" argument. If you have to stick to an old Git version, then it is likely/expected that you have to stick to an older Git LFS version too, no?

It is, but people do weird things sometimes :)

ttaylorr:
> Ah, I always forget about the `--config` thing. Let's keep the deprecation then 👍
> 
> >

Agreed.

keiko713:
I was playing around with how we can retire `git lfs clone`, and noticed that behavior of `git clone` and `git lfs clone` are different when there is no `.gitattributes` file.
`git lfs clone` still download original files whereas `git clone` don't. I'm wondering if this is some expected difference?

Issue #2497, opened 2017-08-09
Files marked as "lockable" but not tracked in lfs don't get their readonly file system attribute set.


mrbobbybobberson:
In https://github.com/git-lfs/git-lfs/wiki/File-Locking, it says:

> Once file patterns in .gitattributes are lockable, Git LFS will make them readonly on the local file system automatically. This prevents users from accidentally editing a file without locking it first.

But I'm not getting the readonly file system attribute set.  Is it working for others?

I'm on Windows 10 Version 1703 and using git-lfs/2.2.1 (GitHub; windows amd64; go 1.8.3; git 621d1f8).

mrbobbybobberson:
Something else that's interesting: the "git lfs lock" command works with any path to a file that exists on disk regardless of whether it's marked as lockable in my .gitattributes.  Is that by design?

ttaylorr:
Hi @mrbobbybobberson, thanks for opening this issue. Could you post the steps that you took in order to observe a file that was not marked as read-only? Specifically, I'm looking for something similar to [this](https://github.com/git-lfs/git-lfs/blob/211e8862901a758f9271d411a8d5cc224c9e0746/test/test-track.sh#L427) test case that would indicate the behavior of `git lfs track` or the post-checkout, post-commit (et. al.) hooks would be broken.

mrbobbybobberson:
Sure thing, @ttaylorr . These steps reproduce the issue for me:

1. Manually add the line "*.hardToMergeFile lockable" to your root .gitattributes file, commit, and push the change.
2. Add a new file named "some.hardToMergeFile", commit, and push the change.
3. Check the readonly attribute on "some.hardToMergeFile" in your file system.

At this point, my system shows the file as writeable, but I expected it to be readonly, as [the wiki](https://github.com/git-lfs/git-lfs/wiki/File-Locking) says:

> If you'd like to register a file type as lockable, without using LFS, you can edit the .gitattributes file directly:

> `*.yml lockable`

> Once file patterns in .gitattributes are lockable, Git LFS will make them readonly on the local file system automatically. This prevents users from accidentally editing a file without locking it first.

ttaylorr:
@mrbobbybobberson ⚡️ thanks for pointing this out: you're totally right. I took a stab at fixing this in https://github.com/git-lfs/git-lfs/pull/2528, which I have scheduled to be included in the next release./

ysakai2:
[...]

PR #2528, opened 2017-08-25
git: expand 'GetAttributePaths' check to include non-LFS lockables

ttaylorr:
This pull request fixes a bug pointed out by @mrbobbybobberson in https://github.com/git-lfs/git-lfs/issues/2497#issuecomment-324173879.

According to our wiki:

> If you'd like to register a file type as lockable, without using LFS, you can edit the .gitattributes file directly:
>
> `*.yml lockable`
>
> Once file patterns in .gitattributes are lockable, Git LFS will make them readonly on the local file system automatically. This prevents users from accidentally editing a file without locking it first.

The post-commit hook should mark those files as read-only. This is correct (and verified by the `test/test-post-commit.sh` test) when files are _also_ tracked using Git LFS. When the `post-commit` hook runs, Git LFS tries to determine a filepathfilter that matches all of the files that are marked as lockable in any of the repository's .gitattributes file(s):

https://github.com/git-lfs/git-lfs/blob/200b8f08cac8f28c46dbf59ae4ca4cf9bcbb9c7d/locking/lockable.go#L39-L52

it does so by scanning the repository's .gitattributes(s) for any paths that are lockable, using this function:

https://github.com/git-lfs/git-lfs/blob/200b8f08cac8f28c46dbf59ae4ca4cf9bcbb9c7d/git/attribs.go#L27-L31

This function scans through each line of each `.gitattributes` file that it finds and returns some information about each one in a slice. Before inspecting each line, however, it first tries to determine if it's tracked by LFS:

https://github.com/git-lfs/git-lfs/blob/200b8f08cac8f28c46dbf59ae4ca4cf9bcbb9c7d/git/attribs.go#L44

For entries like:

```
*.dat lockable
```

this is not the case, as the string `"filter=lfs"` does not appear in that line. Since `lockable` is a part of the scope of Git LFS, include that check as a fallback when scanning each line.

This can go in the list of future improvements to make if we ever write a `.gitattributes` parser, but for now I think this will do.

Closes: https://github.com/git-lfs/git-lfs/issues/2497.

---

/cc @git-lfs/core 
/cc @mrbobbybobberson

### DIFF
--- a/git/attribs.go
+++ b/git/attribs.go
@@ -57,7 +57,14 @@ func GetAttributePaths(workingDir, gitDir string) []AttributePath {
 
 		for scanner.Scan() {
 			line := scanner.Text()
-			if strings.Contains(line, "filter=lfs") {
+
+			// Check for filter=lfs (signifying that LFS is tracking
+			// this file) or "lockable", which indicates that the
+			// file is lockable (and may or may not be tracked by
+			// Git LFS).
+			if strings.Contains(line, "filter=lfs") ||
+				strings.HasSuffix(line, "lockable") {
+
 				fields := strings.Fields(line)
 				pattern := fields[0]
 				if len(reldir) > 0 {
--- a/test/test-post-commit.sh
+++ b/test/test-post-commit.sh
@@ -48,3 +48,29 @@ begin_test "post-commit"
 
 )
 end_test
+
+begin_test "post-commit (locked file outside of LFS)"
+(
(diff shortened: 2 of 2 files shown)


ttaylorr:
@technoweenie good call -- I added an integration test for this behavior in bc5a689, which pointed out an unrelated bug in the post-commit hook. I added a note in that commit pointing it out.

Issue #2519, opened 2017-08-20
Investigate SSH auth timeout issues


technoweenie:
See: https://github.com/git-lfs/git-lfs/pull/2511#issuecomment-323599680

ttaylorr:
Thanks for opening this, @technoweenie! I took a quick look through `lfsapi/ssh.go` to refresh my memory on how that code works, and on the surface it appears that the SSH cacher, when enabled, should be caching and appropriately dropping entries that have expired.

@larsxschneider can you post the output of the command you ran in https://github.com/git-lfs/git-lfs/pull/2511#issuecomment-323599680 with `GIT_TRACE=1` here? The tracerx output in there should be enough to figure out any subtly that I've missed so far. Thanks!

larsxschneider:
@ttaylorr The repo is public: https://github.com/larsxschneider/lfstest-manyfiles 😄 
Are you not able to reproduce the error with this call?
```
git clone git@github.com:larsxschneider/lfstest-manyfiles.git
```

larsxschneider:
You might need to slow down your Internet connection a bit to reproduce it https://apple.stackexchange.com/a/24078 😉

PR #2530, opened 2017-08-25
tools: compare expiration to "time.Now" instead of "from"

ttaylorr:
This pull request fixes a bug that @larsxschneider noticed in https://github.com/git-lfs/git-lfs/pull/2511#issuecomment-323599680 (and that @technoweenie explained in https://github.com/git-lfs/git-lfs/pull/2511#issuecomment-323601319):

> As for the SSH thing, my hunch is that the expiration timeout is not being respected. Is this happening at around the 10 minute mark?
> 
> [...]
> 
> - Does the Batch API try to refresh the Authorization header after it expires? I thought it recognizes that and automatically retries. Maybe there's a race condition, and it should refresh the token if expires_in is < 5 or something.
> - Does the SSH cred cacher respect the expires_in? It should be falling back to a real ssh call after expires_in seconds have passed.

I incorrectly noted that the behavior looks correct in https://github.com/git-lfs/git-lfs/issues/2519#issuecomment-324730475, but upon further investigation was able to find the bug.

To explain, here's how the `tools.IsExpiredAtOrIn()` function works:

1. Receive a starting time, a relative duration to expire, an absolute time to expire, and the current time.
2. If the relative duration is non-zero (i.e., the SSH `git-lfs-authenticate` call supports `ExpiresIn`), add that duration to the starting time "from", and mark that as the expiration time.
3. If now, mark the absolute "at" time as the expiration time.
4. Figure out if the current time is less than "in" units of time away from the expiration time as determined above.

The problem was in step four: previously, we were taking the current time to be the "now" parameter, where it actually should be `time.Now()`. The from parameter is given as the creation timestamp of the SSH credential cache entry, which will always be earlier (and therefore, non-expired) than the expiration timestamp.

Fix this by using `time.Now()` instead of `now` in order to determine if the time has passed.

Closes: https://github.com/git-lfs/git-lfs/issues/2519.

---

/cc @git-lfs/core 
/cc @larsxschneider 
/cc https://github.com/git-lfs/git-lfs/issues/2519

### DIFF
--- a/tools/time_tools.go
+++ b/tools/time_tools.go
@@ -1,24 +1,26 @@
 package tools
 
-import "time"
+import (
+	"time"
+)
 
 // IsExpiredAtOrIn returns whether or not the result of calling TimeAtOrIn is
 // "expired" within "until" units of time from now.
-func IsExpiredAtOrIn(now time.Time, until time.Duration, at time.Time, in time.Duration) (time.Time, bool) {
-	expiration := TimeAtOrIn(now, at, in)
+func IsExpiredAtOrIn(from time.Time, until time.Duration, at time.Time, in time.Duration) (time.Time, bool) {
+	expiration := TimeAtOrIn(from, at, in)
 	if expiration.IsZero() {
 		return expiration, false
 	}
 
-	return expiration, expiration.Before(now.Add(until))
+	return expiration, expiration.Before(time.Now().Add(until))
 }
 
 // TimeAtOrIn returns either "at", or the "in" duration added to the current
 // time. TimeAtOrIn prefers to add a duration rather than return the "at"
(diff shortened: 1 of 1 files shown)


larsxschneider:
The fix seems to work. This recreates your initial implementation that you've [changed](https://github.com/git-lfs/git-lfs/commit/0d66374f4f2da583bbbed1da621b778547175e60#diff-b30090b5e9a4d6bc99bff85cb30ea499R7) as part of #2130. I haven't looked further, but can we be sure we don't break any other intended use of this function?

@technoweenie 's comment is [spot on](https://github.com/git-lfs/git-lfs/pull/2130/files#r110219458) 😄

PR #2540, opened 2017-08-29
commands: teach 'git lfs ls-files' a '--debug' option

larsxschneider:
There was no way to conveniently display LFS pointer file properties for
debugging. Teach 'git lfs ls-files' a '--debug' option that prints all
available information following the 'git ls-files' interface [1].

[1] https://github.com/git/git/commit/84974217151a66b65ae5b173045483ce5eeb5142

--

If you like the approach of this PR, then this would this would replace #1829.
My goal here is a quick way to check the sizes of the Git LFS files in a repo.

### DIFF
--- a/commands/command_ls_files.go
+++ b/commands/command_ls_files.go
@@ -10,6 +10,7 @@ import (
 
 var (
 	longOIDs = false
+	debug    = false
 )
 
 func lsFilesCommand(cmd *cobra.Command, args []string) {
@@ -38,7 +39,24 @@ func lsFilesCommand(cmd *cobra.Command, args []string) {
 			return
 		}
 
-		Print("%s %s %s", p.Oid[0:showOidLen], lsFilesMarker(p), p.Name)
+		if debug {
+			Print(
+				"filepath: %s\n"+
+					"    size: %d\n"+
+					"checkout: %v\n"+
+					"download: %v\n"+
+					"     oid: %s %s\n"+
+					" version: %s\n",
+				p.Name,
+				p.Size,
+				fileExistsOfSize(p),
+				lfs.ObjectExistsOfSize(p.Oid, p.Size),
+				p.OidType,
+				p.Oid,
+				p.Version)
+		} else {
+			Print("%s %s %s", p.Oid[0:showOidLen], lsFilesMarker(p), p.Name)
+		}
 	})
 	defer gitscanner.Close()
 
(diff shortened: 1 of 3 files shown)


docs/man/git-lfs-ls-files.1.ronn
```diff
@@ -15,6 +15,10 @@ reference.  If no reference is given, scan the currently checked-out branch.
 * `-l` `--long`:
   Show the entire 64 character OID, instead of just first 10.
 
+* -d --debug:
```
ttaylorr: Can this be `--extra`?

larsxschneider: It could although my intention was to match the `git ls-files` option: git/git@8497421
Up to you. Either way is fine with me 🙂

commands/command_ls_files.go
```diff
@@ -38,7 +39,25 @@ func lsFilesCommand(cmd *cobra.Command, args []string) {
...
 
-		Print("%s %s %s", p.Oid[0:showOidLen], lsFilesMarker(p), p.Name)
+		if debug {
+			Print(
+				"  filepath: %s\n"+
+					"      size: %d\n"+
+					"  checkout: %v\n"+
+					"downloaded: %v\n"+
+					"       oid: %s\n"+
+					"   oidtype: %s\n"+
```
technoweenie: I'd probably smash the oid into one line:

    oid: %s %s
    oid: sha256 abcdef...

docs/man/git-lfs-ls-files.1.ronn
ttaylorr: Ah, I did not know that `git ls-files` has `--debug`. I agree that we should continue to use that for consistency.

commands/command_ls_files.go
larsxschneider: Agreed 👍

test/test-ls-files.sh
```diff
@@ -24,6 +24,16 @@ begin_test "ls-files"
...
+
+  diff -u <(git lfs ls-files --debug) <(cat <<-EOF
+filepath: some.dat
+    size: 10
+checkout: true
+download: true
+     oid: sha256 5aa03f96c77536579166fba147929626cc3a97960e994057a9d80271a736d10f
+ version: https://git-lfs.github.com/spec/v1
+
+EOF)
```
larsxschneider: @ttaylorr Is this sufficient?

ttaylorr: Looks great, thanks!

Issue #2359, opened 2017-06-26
migrate: teach option to not save rewritten objects


ttaylorr:
This issue affects the migrator running in `info` mode. If the majority of objects being scanned by the `*git/githistory.HistoryRewriter` are packed, and the migration is a no-op, the `info` migrator will write loose objects, effectively unpacking a potentially large repository.

Teach either:

1. `NoWriteObjects` to skip writing objects all together
2. Packfile APIs in order to write either a) packed objects, or b) skip writing packed objects if the object already exists in the database (loose or packed).

---

/cc @git-lfs/core

ttaylorr:
Acking out of our internal first responder queue.

PR #2541, opened 2017-08-30
git/{odb,githistory}: don't write unchanged objects

ttaylorr:
This pull request teaches a typesafe `Equal()` to `*Blob`'s, `*Commit`'s, and `*Tree`'s, and uses that function in the `git/githistory` package to avoid writing unchanged objects to disk.

Previously, the `git lfs migrate` command would write the contents of every object that it examined in the migration, regardless of whether that object's contents changed. This is sub-optimal, because it involves not only marshaling the contents of unchanged objects to a buffer and performing an expensive save, but also because it effectively has the behavior of unpacking all objects in a repository. That's no good.

To address this, this pull request opts for the following option that I originally proposed in https://github.com/git-lfs/git-lfs/issues/2359:

> Teach either:
> 
> 1. [...]
> 2. Packfile APIs in order to write either a) packed objects, or b) _skip writing packed objects if the object already exists in the database (loose or packed)_.

We can detect whether objects already exist in the object database in either a loose or packed format by determining whether or not the contents of the blob/tree/commit changed when rewriting it. In order to open the object, it must exist, so therefore, if it is unchanged, the object is already saved to the database.

By defining `(t *<T>) Equal(<T> other)` on each object type in package `git/odb`, and comparing the "rewritten" instances of each object against the original, we can avoid writing objects to the database (which involves several filesystem-level operations, `open()`, `write()`, a syscall to move, etc.) and instead return early, avoiding all of the previously mentioned operations.

This yields a significant performance boost. When examining the `git-lfs/git-lfs` repository prior to this change, it takes about 13.3 seconds to examing all ~6,000 commits:

```
~/g/git-lfs (master) $ time git lfs migrate info --include-ref=refs/heads/master
migrate: Rewriting commits: 100% (5714/5714), done
*.md    1.1 MB  191/191 files(s)        100%
*.go    106 KB    86/86 files(s)        100%
*.yml   89 KB     87/87 files(s)        100%
*.toml  35 KB     28/28 files(s)        100%
*.lock  30 KB     17/17 files(s)        100%
git lfs migrate info --include-ref=refs/heads/master  10.77s user 7.31s system 135% cpu 13.318 tota
```

Including this change, however, the operation takes only 0.7 seconds:

```
~/g/git-lfs (master!) $ time git lfs migrate info --include-ref=refs/heads/master
migrate: Rewriting commits: 100% (5714/5714), done
*.md    1.1 MB  191/191 files(s)        100%
*.go    106 KB    86/86 files(s)        100%
*.yml   89 KB     87/87 files(s)        100%
*.toml  35 KB     28/28 files(s)        100%
*.lock  30 KB     17/17 files(s)        100%
git lfs migrate info --include-ref=refs/heads/master  0.49s user 0.18s system 101% cpu 0.662 total
```

Closes: https://github.com/git-lfs/git-lfs/issues/2359.

---

/cc @git-lfs/core

### DIFF
--- a/git/odb/blob.go
+++ b/git/odb/blob.go
@@ -68,3 +68,18 @@ func (b *Blob) Close() error {
 	}
 	return b.closeFn()
 }
+
+// Equal returns whether the receiving and given blobs are equal, or in other
+// words, whether they are represented by the same SHA-1 when saved to the
+// object database.
+func (b *Blob) Equal(other *Blob) bool {
+	if (b == nil) != (other == nil) {
+		return false
+	}
+
+	if b != nil {
+		return b.Contents == other.Contents &&
+			b.Size == other.Size
+	}
+	return true
+}
--- a/git/githistory/rewriter.go
+++ b/git/githistory/rewriter.go
@@ -321,6 +321,10 @@ func (r *Rewriter) rewriteTree(sha []byte, path string, fn BlobRewriteFn, tfn Tr
 	if err != nil {
 		return nil, err
 	}
+
+	if tree.Equal(rewritten) {
+		return sha, nil
+	}
 	return r.db.WriteTree(rewritten)
 }
 
--- a/git/githistory/rewriter.go
+++ b/git/githistory/rewriter.go
@@ -215,7 +215,7 @@ func (r *Rewriter) Rewrite(opt *RewriteOptions) ([]byte, error) {
 
 		// Construct a new commit using the original header information,
 		// but the rewritten set of parents as well as root tree.
-		rewrittenCommit, err := r.db.WriteCommit(&odb.Commit{
+		rewrittenCommit := &odb.Commit{
 			Author:       original.Author,
 			Committer:    original.Committer,
 			ExtraHeaders: original.ExtraHeaders,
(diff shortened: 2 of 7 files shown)


git/odb/blob.go
```diff
@@ -68,3 +68,18 @@ func (b *Blob) Close() error {
...
+// Equal returns whether the receiving and given blobs are equal, or in other
+// words, whether they are represented by the same SHA-1 when saved to the
+// object database.
+func (b *Blob) Equal(other *Blob) bool {
+	if (b == nil) != (other == nil) {
+		return false
+	}
+
+	if b != nil {
+		return b.Contents == other.Contents &&
```
technoweenie: I'm not sure comparing `io.Reader` is the right solution here. I'd expect a test like this to pass:

```go
func TestBlobEqualReturnsTrueWithUnchangedButReadContents(t *testing.T) {
	c1 := strings.NewReader("Hello, world!")
	c2 := strings.NewReader("Hello, world!")

	b1 := &Blob{Size: int64(c1.Len()), Contents: c1}
	b2 := &Blob{Size: int64(c2.Len()), Contents: c2}

	assert.True(t, b1.Equal(b2))
}
```

Based on the way that the blob rewrite function usually returns a new `*Blob`, I think this will always fail the `Equal()` check, unless both `Contents` are pointers to the same reader.

https://github.com/git-lfs/git-lfs/blob/7aeeec245c69984279fe2531b077e41e2d00ad74/commands/command_migrate_import.go#L31-L49

I suppose this works out in practice, since any call to `BlobFn` means that blob is definitely being converted.

ttaylorr: Great question -- you're absolutely right. Two `io.Reader`'s that yield the same content but are instantiated differently will fail the equality check.

That's fine in practice here, since as you pointed out, all calls to the `BlobFn` mean the blob is definitely being converted. That said, there are cases in the 'info' migrator where the contents of the blob does not change:

https://github.com/git-lfs/git-lfs/blob/cdd0cb70a45db94a93741013e7b351c003784be5/commands/command_migrate_info.go#L70-L91

A more complete implementation might involve some buffering elsewhere, and keeping track of the SHA as it changes, but I think that this will do for now.

Issue #2532, opened 2017-08-28
Adding origins with missing LFS info won't lead to any complaints, warnings, . and git/git-lfs will happily pull commits with missing LFS storage objects


JonasT:
I just ran into the following situation where I freshly cloned a repo, then added an additional remote (with same repo but slightly different state) and it seems LFS completely borked to fetch the LFS objects on its own for some reason:
```
jonas@cyberman:~/Develop$ git clone https://github.com/tethercode/stdlib ./stdlib-delete
Cloning into './stdlib-delete'...
remote: Counting objects: 5669, done.
remote: Compressing objects: 100% (18/18), done.
remote: Total 5669 (delta 6), reused 9 (delta 3), pack-reused 5648
Receiving objects: 100% (5669/5669), 1.28 MiB | 451.00 KiB/s, done.
Resolving deltas: 100% (3886/3886), done.
jonas@cyberman:~/Develop$ cd stdlib-delete/
jonas@cyberman:~/Develop/stdlib-delete$ vim .git/config 
jonas@cyberman:~/Develop/stdlib-delete$ git pull gitlab
warning: redirecting to https://gitlab.com/tethercode/stdlib.git/
remote: Counting objects: 124, done.
remote: Compressing objects: 100% (121/121), done.
remote: Total 124 (delta 75), reused 0 (delta 0)
Receiving objects: 100% (124/124), 22.58 KiB | 7.53 MiB/s, done.
Resolving deltas: 100% (75/75), completed with 16 local objects.
From https://gitlab.com/tethercode/stdlib
 + 767ce25...e3fc7d9 master     -> origin/master  (forced update)
You asked to pull from the remote 'gitlab', but did not specify
a branch. Because this is not the default configured remote
for your current branch, you must specify a branch on the command line.
jonas@cyberman:~/Develop/stdlib-delete$ git pull gitlab master
warning: redirecting to https://gitlab.com/tethercode/stdlib.git/
From https://gitlab.com/tethercode/stdlib
 * branch            master     -> FETCH_HEAD
Merge made by the 'recursive' strategy.
 .gitignore                                   |   1 +
 Makefile                                     |   2 +-
 compiler/errorprint.tc                       |  90 ++++
 compiler/helpers.tc                          |  87 ++++
 compiler/run.tc                              |  31 ++
 compiler/semantics.tc                        |   9 +-
 compiler/tetherasm/instruction.tc            | 623 ++++-----------------------
[...]

PR #2544, opened 2017-09-01
commands: add --all flag to 'git lfs clone'

technoweenie:
Adds a `--all` flag to `git lfs clone` that runs `git lfs fetch --all` right after the `git clone` command.

I realize the command is deprecated as of https://github.com/git-lfs/git-lfs/pull/2526. This is a simple change, and a little useful (see https://github.com/git-lfs/git-lfs/issues/2532#issuecomment-326634994 for use case). Is it worth keeping the `git lfs clone` command around, or should we continue on with the deprecation?

IMO, with the upcoming `delay` filter capabilities, the main need for `git lfs clone` is gone, so we should reject this PR.

### DIFF
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -92,6 +92,14 @@ func cloneCommand(cmd *cobra.Command, args []string) {
 		cfg.CurrentRemote = remote
 		fetchRef("HEAD", filter)
 	} else {
+		if fetchAllArg {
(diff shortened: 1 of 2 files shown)


ttaylorr:
> IMO, with the upcoming `delay` filter capabilities, the main need for `git lfs clone` is gone, so we should reject this PR.

Agreed. I think this is a good change, but should be covered by `git clone`.

Issue #2381, opened 2017-07-05
`git lfs migrate info` fails with "invalid cross-device link" error


ashumkin:
```
$ GIT_TRACE=2 git lfs migrate info
14:00:04.174380 git.c:557               trace: exec: 'git-lfs' 'migrate' 'info'
14:00:04.174469 run-command.c:347       trace: run_command: 'git-lfs' 'migrate' 'info'
trace git-lfs: run_command: 'git' config -l
trace git-lfs: run_command: git cat-file --batch
trace git-lfs: run_command: 'git' rev-parse HEAD --symbolic-full-name HEAD
trace git-lfs: run_command: 'git' rev-parse master --symbolic-full-name master
trace git-lfs: run_command: git rev-list --reverse --topo-order --do-walk master --
migrate: Sorting commits: ..., done                                                                                                                                                                                
rename /tmp/040452966 /data/projects/MobileApplicationsControlServer-v1/.git/objects/c8/d49f5b5760e17701079038132a32986449b51d: invalid cross-device link 
```
```
$ git lfs version
git-lfs/2.2.0 (GitHub; linux amd64; go 1.8.3; git 2fc3f7f)

$ git --version
git version 2.4.5.21.g9c7f10a

$ df -h
Filesystem                    Size  Used Avail Use% Mounted on
...
tmpfs                         7.6G  148K  7.6G   1% /tmp
...
/dev/mapper/vg_data      266G  239G   27G  90% /data
```
any suggestions?

ttaylorr:
Hi @ashumkin, thanks for opening this issue and sorry that you're having trouble!

I think this issue occurs because of a step taken in the object saving process, wherein the object is first written to `/tmp` and then linked (via `os.Rename()`) into `.git/objects` inside of your repository.

I opened up #2382 to try and fix this, and if it works, I'd love to ship it in the v2.2.1 release.

I built #2382 into a binary for Linux on amd64 and x86-64 architectures. Would you mind giving it a shot and seeing if it works?

[git-lfs-linux-386-2.2.0.tar.gz](https://github.com/git-lfs/git-lfs/files/1124631/git-lfs-linux-386-2.2.0.tar.gz)
[git-lfs-linux-amd64-2.2.0.tar.gz](https://github.com/git-lfs/git-lfs/files/1124632/git-lfs-linux-amd64-2.2.0.tar.gz)

ashumkin:
I confirm fix

ioquatix:
[...]

PR #2566, opened 2017-09-08
commands/migrate: persistently initialize localstorage

ttaylorr:
This pull request prevents a failure from hardlinking new objects created during migration when a different filesystem is mounted at `/tmp`.

As pointed out by @ioquatix in https://github.com/git-lfs/git-lfs/issues/2381#issuecomment-327677404:

> I'm using 2.2.1 and I had to do the following:
> 
> ```
> sudo -u http TMPDIR=`pwd`/.git/lfs/tmp git lfs pull
> ```
>
> Otherwise I got the same errors trying to copy files from /tmp to /srv/http/...

Using `tmpfs` mounted at `/tmp` can still cause a cross-device link issue when trying to create a hardlink via `os.Rename()`.

I was curious why this was happening, since we use `lfs.TempDir` and `lfs.TempFile()` for creating temporary files to store objects when migrating. It turns out, however, that failing to call `lfs.InitializeLocalStorage()` will let `lfs.TempDir` be equal to `os.Gettemp()`:

https://github.com/git-lfs/git-lfs/blob/a2981c1b39353858e08f498e9141ad118c9d1e35/localstorage/currentstore.go#L22

and:

https://github.com/git-lfs/git-lfs/blob/a2981c1b39353858e08f498e9141ad118c9d1e35/localstorage/currentstore.go#L37

Thus, failing to call `lfs.InitLocalStorage()` will hardlink files from `/tmp` into `.git/lfs/objects` (and `.git/objects`), potentially causing hardlink errors as above.

Closes: https://github.com/git-lfs/git-lfs/issues/2381

---

/cc @git-lfs/core

### DIFF
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -9,6 +9,7 @@ import (
 	"github.com/git-lfs/git-lfs/git/githistory"
 	"github.com/git-lfs/git-lfs/git/githistory/log"
 	"github.com/git-lfs/git-lfs/git/odb"
+	"github.com/git-lfs/git-lfs/localstorage"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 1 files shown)


ioquatix:
Thanks for sorting this out so quickly.

PR #2570, opened 2017-09-11
commands: print IP addresses in error logs

larsxschneider:
It can be hard to trace Git LFS errors on the server without client IP
address. Fix this by adding the IP address to the error log output.

#### Attention
Is this OK from a privacy perspective? I think it is OK as we log the file to the local machine that knows the IP address anyways. A user that is concerned about the IP address could remove it from an error report. The same is already the case if the error output contains sensitive filenames of a repo.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -231,7 +231,7 @@ func Debug(format string, args ...interface{}) {
 }
 
 // LoggedError prints the given message formatted with its arguments (if any) to
-// Stderr. If an empty string is passed as the "format" arguemnt, only the
+// Stderr. If an empty string is passed as the "format" argument, only the
 // standard error logging message will be printed, and the error's body will be
 // omitted.
 //
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


commands/commands.go
```diff
@@ -334,6 +335,46 @@ func logPanic(loggedError error) string {
...
+			case *net.IPNet:
+				ip = v.IP
+			case *net.IPAddr:
+				ip = v.IP
+			}
+			if ip == nil || ip.IsLoopback() {
+				continue
+			}
+			s = s + ip.String() + " "
+		}
```
larsxschneider: I got heavily inspired here: https://stackoverflow.com/a/23558495 ;-)

commands/commands.go
```diff
@@ -334,6 +335,46 @@ func logPanic(loggedError error) string {
...
+		for _, addr := range addrs {
+			var ip net.IP
+			switch v := addr.(type) {
+			case *net.IPNet:
+				ip = v.IP
+			case *net.IPAddr:
+				ip = v.IP
+			}
+			if ip == nil || ip.IsLoopback() {
```
ttaylorr: Should this check go before the `switch` above? You could have something like `(*net.IPNet)(nil)`, which would panic on L362 after trying to dereference it.

commands/commands.go
```diff
@@ -334,6 +335,46 @@ func logPanic(loggedError error) string {
...
+				continue
+			}
+			s = s + ip.String() + " "
```
ttaylorr: Instead of adding a `" "` on the end of `s` here, could you use a `[]string` and call `strings.Join(slice, " ")` to prevent the trailing whitespace?

larsxschneider: Good idea!

commands/commands.go
larsxschneider: I am not sure if I understand you right, but this doesn't panic:
```
                        addr = nil
			var ip net.IP
			switch v := addr.(type) {
			case *net.IPNet:
				ip = v.IP
			case *net.IPAddr:
				ip = v.IP
			}
```

ttaylorr: 😱 totally disregard my suggestion -- I misread your code as `if addr == nill || addr.IsLoopback()`, instead of `ip`. This is all good 👍

larsxschneider:
@ttaylorr I fixed the string issue in a rebase!

PR #2571, opened 2017-09-11
commands: print current time in UTC to error logs

larsxschneider:
It can be hard to trace Git LFS errors on the server without the exact
time. Fix this by adding the time in UTC to the error log output.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -362,6 +362,9 @@ func logPanicToWriter(w io.Writer, loggedError error, le string) {
 		fmt.Fprintf(w, "%s=%v"+le, key, val)
 	}
 
+	fmt.Fprint(w, le+"Current time in UTC: "+le)
+	fmt.Fprint(w, time.Now().UTC().Format("2006-01-02 15:04:05")+le)
+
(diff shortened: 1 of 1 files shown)


commands/commands.go
```diff
@@ -362,6 +362,9 @@ func logPanicToWriter(w io.Writer, loggedError error, le string) {
 		fmt.Fprintf(w, "%s=%v"+le, key, val)
 	}
 
+	fmt.Fprint(w, le+"Current time in UTC: "+le)
```
ttaylorr: Could this be moved onto the same line:

```go
fmt.Fprintf(w, le+"Current time in UTC: %s"+le, time.Now().UTC().Format("2006-01-02 15:04:05"))
```

larsxschneider: Ha, this is how I had it initially. I changed it because the line was too long and the other two data sources also have a "header" and a "content" print line.

Does this convince you? If not, I am happy to merge the lines, too. Either way is fine for me.

ttaylorr: > Does this convince you?

It does :)

PR #2578, opened 2017-09-13
lfsapi: do not accept arguments to GIT_ASKPASS, core.askpass

ttaylorr:
This pull request fixes a bug where program arguments would be incorrectly accepted by the `GIT_ASKPASS` parser, therefore breaking `GIT_ASKPASS` and `core.askpass` exec paths that have spaces or quotes in them.

---

/cc @git-lfs/core @joshaber

### DIFF
--- a/lfsapi/creds.go
+++ b/lfsapi/creds.go
@@ -3,6 +3,7 @@ package lfsapi
 import (
 	"bytes"
 	"fmt"
+	"net/url"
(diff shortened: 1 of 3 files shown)


ttaylorr:
Rebased this off of master and added df84691, which executes `core.askpass` (or `$GIT_ASKPASS`) twice, once for the username and once for the password.

PR #2580, opened 2017-09-13
Migrate doc update

technoweenie:
This updates the man page:

* Lists the `import` mode
* Reorders the sections so that `info` is described before `import`
* Add some examples of how to use `info` and `import`.

### DIFF
--- a/docs/man/git-lfs-migrate.1.ronn
+++ b/docs/man/git-lfs-migrate.1.ronn
@@ -111,8 +111,66 @@ The following configuration:
 
 Would, therefore, include commits: F, E, D, C, B, but exclude commit A.
 
-The presence of flag --everything indicates that all local references should be
+The presence of flag `--everything` indicates that all local references should be
 migrated.
+
+## EXAMPLES
+
+### Migrate unpushed commits
+
+The migrate command's most common use case is to convert large git objects to
+LFS before pushing your commits. By default, it only scans commits that don't
+exist on any remote.
+
+First, run `git lfs migrate info` to list the file types taking up the most
+space in your repository.
+
+```
+$ git lfs migrate info
+migrate: Fetching remote refs: ..., done
+migrate: Sorting commits: ..., done
+migrate: Examining commits: 100% (1/1), done
+*.mp3  	284 MB	  1/1 files(s)	100%
+*.pdf  	42 MB 	  8/8 files(s)	100%
+*.psd  	9.8 MB	15/15 files(s)	100%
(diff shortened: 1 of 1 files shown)


docs/man/git-lfs-migrate.1.ronn
```diff
@@ -10,6 +10,9 @@ git-lfs-migrate(1) - Migrate history to or from git-lfs
 * `info`
     Show information about repository size.
 
+* `import`
```
ttaylorr: ❇️ Can we put this above the `info` mode, since `import` sorts ahead of `info` lexicographically?

docs/man/git-lfs-migrate.1.ronn
```diff
@@ -72,6 +63,18 @@ The 'info' mode has these additional options:
     If a --unit is not specified, the largest unit that can fit the number of
     counted bytes as a whole number quantity is chosen.
 
+### IMPORT
```
ttaylorr: Same note here as above.

docs/man/git-lfs-migrate.1.ronn
```diff
@@ -108,9 +111,67 @@ The following configuration:
...
+```
+
+Now, you can run `git lfs migrate import` to convert some file types to LFS:
+
+```
+$ git lfs migrate import --include="*.mp3" --include="*.psd"
+migrate: Fetching remote refs: ..., done
+migrate: Sorting commits: ..., done
+migrate: Rewriting commits: 100% (1/1), done
+  mp3	d2b959babd099fe70da1c1512e2475e8a24de163 -> 136e706bf1ae79643915c134e17a6c933fd53c61
```
ttaylorr: Is `mp3` a local ref? This may be confusing as `mp3` is an include pattern given to this command. Can we change this to `master`?

docs/man/git-lfs-migrate.1.ronn
technoweenie: I was thinking it'd be better to sort _logically_. `info` is the first command you run. 

I imagine we'll add an `export` command at some point. What looks better?

* lexicographically
  * export
  * import
  * info
* technologic
  * info
  * import
  * export

docs/man/git-lfs-migrate.1.ronn
technoweenie: Ah yup!

docs/man/git-lfs-migrate.1.ronn
ttaylorr: Ah, great call! I really dig the technological ordering :)

PR #2584, opened 2017-09-14
commands: allow $GIT_LFS_TRACK_NO_INSTALL_HOOKS to disable hook installation

ttaylorr:
This pull request allows callers to disable the hook installation component of running `git lfs track` via an environment variable `GIT_LFS_TRACK_NO_INSTALL_HOOKS`.

If set to a truth-y value, this skips the `lfs.InstallHooks(false)` call made by `git lfs track`. It is implemented as an environment flag since this is not functionality that we are intending to expose to users, rather, this is intended to be used by other programs leveraging Git LFS. It is absent from the manpage for the same reason.

---

/cc @git-lfs/core @joshaber

### DIFF
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -42,7 +42,9 @@ func trackCommand(cmd *cobra.Command, args []string) {
 		os.Exit(128)
 	}
 
-	lfs.InstallHooks(false)
+	if !cfg.Os.Bool("GIT_LFS_TRACK_NO_INSTALL_HOOKS", false) {
+		lfs.InstallHooks(false)
+	}
 
 	if len(args) == 0 {
 		listPatterns()
--- a/test/test-track.sh
+++ b/test/test-track.sh
@@ -488,3 +488,25 @@ begin_test "track (symlinked repository)"
   popd > /dev/null
 )
 end_test
+
+begin_test "track (\$GIT_LFS_TRACK_NO_INSTALL_HOOKS)"
+(
+  set -e
+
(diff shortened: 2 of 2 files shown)


ttaylorr:
@technoweenie good points:

> Other commands call `lfs.InstallHooks()` too. Should they respect this env flag?

We have other callers of this command in:

```
~/g/git-lfs (git-track-no-install) $ rg --no-heading 'lfs\.InstallHooks'
commands/command_clean.go:	lfs.InstallHooks(false)
commands/command_clone.go:		if err := lfs.InstallHooks(false); err != nil {
commands/command_filter_process.go:	lfs.InstallHooks(false)
commands/command_fsck.go:	lfs.InstallHooks(false)
commands/command_smudge.go:	lfs.InstallHooks(false)
commands/command_track.go:		lfs.InstallHooks(false)
commands/command_untrack.go:	lfs.InstallHooks(false)
commands/command_update.go:		if err := lfs.InstallHooks(updateForce); err != nil {
```

But I don't really want this sort of thing to leak out over the codebase. If we were going to do that, I think adding a `--no-install-hooks` global flag would be best. For now, I think this will satisfy the main use-case.

> Probably worth noting in the man pages for each command that this env flag is available.

Fair point, though I think if we stick to just updating the `git lfs track` command, we should leave it out since this is deliberately incomplete. From my original post:

>> It is implemented as an environment flag since this is not functionality that we are intending to expose to users, rather, this is intended to be used by other programs leveraging Git LFS. It is absent from the manpage for the same reason.

PR #2585, opened 2017-09-15
show proper icon in add/remove programs list

shiftkey:
After installing Git LFS you'll see an entry in Add/Remove Programs that doesn't have an icon:

I'm _pretty sure_ this is the fix, as it autocompletes in the GUI but beyond that I can't point to any docs around what this represents:

I also couldn't validate this locally with Inno Setup 5.5.9 as it returns this error (same thing when you try and compile it in the GUI):

```
C:\Users\shiftkey\go\src\github.com\git-lfs\git-lfs\script>compile-win-installer-unsigned.bat

C:\Users\shiftkey\go\src\github.com\git-lfs\git-lfs\script>"C:\Program Files (x86)\Inno Setup 5\iscc.exe" /Qp "windows-installer\\inno-setup-git-lfs-installer.iss"
[ISPP] Preprocessing.
Error on line 7 in C:\Users\shiftkey\go\src\github.com\git-lfs\git-lfs\script\windows-installer\inno-setup-git-lfs-installer.iss: [ISPP] Actual parameter FileName is not of the declared type.
Compile aborted.
```


ttaylorr:
@shiftkey thanks for taking a look into this! I think that @sschuberth has helped with this sort of error in InnoSetup before... perhaps you could install the 64bit version instead and see if that clears things up?

script/windows-installer/inno-setup-git-lfs-installer.iss
```diff
@@ -33,6 +33,7 @@ OutputBaseFilename={#MyAppFilePrefix}-{#MyAppVersion}
 OutputDir=..\..\
 PrivilegesRequired=none
 SetupIconFile=git-lfs-logo.ico
+UninstallDisplayIcon=git-lfs-logo.ico
```
sschuberth: Mind sorting this in alphabetically for quicker lookup, like the other lines?

sschuberth:
The `[ISPP] Actual parameter FileName is not of the declared type` error occurs because the installer is trying to extract the Git LFS version from the Git LFS binary, which needs to be built first. As I just did some minor improvements to the script in #2586 anyway, I could add actually compiling Git LFS to that script, too, if desired.

ttaylorr:
> The `[ISPP] Actual parameter FileName is not of the declared type` error occurs because the installer is trying to extract the Git LFS version from the Git LFS binary, which needs to be built first.

Great catch, and thank you for digging into this. I have a hard time understanding those InnoSetup errors anyhow :)

> I could add actually compiling Git LFS to that script, too, if desired.

That's OK, I think compiling that outside of the installer setup would be fine... perhaps we could add a comment in the script saying that LFS needs to be built first before building the installer?

sschuberth:
> perhaps we could add a comment in the script saying that LFS needs to be built first before building the installer?

My thinking was that it's quite obvious that the binary needs to be built before you build the installer that packages it 😉 However, I agree that the error message could be clearer. I'll think about improving it, either as part of the Inno Setup script itself, or the batch file calling the Inno Setup compiler.

sschuberth:
> I'll think about improving it, either as part of the Inno Setup script itself, or the batch file calling the Inno Setup compiler.

Done that in #2588.

shiftkey:
Rebased on top of `master`, here's what it now looks like:

script/windows-installer/inno-setup-git-lfs-installer.iss
```diff
@@ -45,6 +45,7 @@ PrivilegesRequired=none
 SetupIconFile=git-lfs-logo.ico
 SolidCompression=yes
 UsePreviousAppDir=no
+UninstallDisplayIcon={app}\git-lfs.exe
```
sschuberth: Sorry to be a nitpicker here, but that's still not sorted alphabetically, as it should come *before* `UsePreviousAppDir` :wink:

shiftkey: @sschuberth not at all - my bad!

PR #2589, opened 2017-09-15
installer: Insert the app directory at the beginning of PATH

### LINKED ISSUE #2587: Git LFS Windows installer does not update Git LFS installed by Git for Windows
Git for Windows installs (by default, but optionally) `git-lfs.exe` into `C:\Program Files\Git\mingw64\bin`. Even if Git LFS is installed by Git for Windows, Git LFS will install itself to `C:\Program Files\Git LFS`.

This can lead to two different versions of Git LFS on a single machine. Git will use the version that "earlier" in the PATH I think.

Is this expected/known?

//cc @sschuberth

sschuberth:
This way binaries that come with this installer are given precedence
over any previously installed Git LFS binaries shipping with Git for
Windows. Users can then run this installer to get a more recent version
of Git LFS than the one coming with Git for Windows without the need to
wait for a new Git for Windows release with updated bundled Git LFS
binaries.

Closes #2587.

### DIFF
--- a/script/windows-installer/inno-setup-git-lfs-installer.iss
+++ b/script/windows-installer/inno-setup-git-lfs-installer.iss
@@ -52,9 +52,9 @@ Source: ..\..\git-lfs-x64.exe; DestDir: "{app}"; Flags: ignoreversion; DestName:
 Source: ..\..\git-lfs-x86.exe; DestDir: "{app}"; Flags: ignoreversion; DestName: "git-lfs.exe"; AfterInstall: InstallGitLFS; Check: not Is64BitInstallMode
 
 [Registry]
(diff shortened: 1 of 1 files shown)


sschuberth:
@larsxschneider could you please also verify this?

sschuberth:
Please note that this will only be effective for new installs, not for upgrades.

larsxschneider:
@sschuberth Can you upload an installer for me to test? 😄

sschuberth:
Just take the one from CI like I do :wink:

sschuberth:
I just tried to verify it, but it does not work. Seems like Git for Windows always searches its own binary paths first. So we probably need a whole different approach. I'm closing this unmerged.

PR #2590, opened 2017-09-17
Support standalone custom transfer based on API URL prefix match

sprohaska:
PR-2429 has implemented a mechanism to unconditionally specify
standalone custom transfer agents.

This commit addresses two limitations, so that standalone custom
transfer can be configured per remote.

* Custom transfer agents may now be specified based on an URL prefix
  match on the API URL.

* git-lfs passes the Git remote in the stage 1 initiation message.
  The standalone custom transfer agent can use it to determine the
  remote file location.

One remote can, for example, point to a server that supports ssh login
and use rsync.  Another remote can point to GitHub.  LFS works for both.
Example Git config:

```
remote.github.url=...
remote.origin.url=ssh://gitssh.example.com/git/some/repo
lfs.customtransfer.rsync.path=git-lfs-rsync-agent
lfs.https://gitssh.example.com/git/.standalonetransfer=rsync
```

The config assumes that `git-lfs-rsync-agent` determines the remote from
the stage 1 init message and then inspecting `remote.origin.url` to
infer the rsync host.  <https://github.com/aleb/git-lfs-rsync-agent>,
2017-09-17, does not yet do that.

[PR-2429] 09b7c5381bc3fa4ba7ca3564681a5ae5b7e6f3dc 'Allow using custom transfer agents directly', <https://github.com/git-lfs/git-lfs/pull/2429>.

CC: Alexandru Băluț <ab@daedalean.ai>
Signed-off-by: Steffen Prohaska <prohaska@zib.de>


commands/uploader.go
```diff
@@ -436,12 +436,15 @@ func ensureFile(smudgePath, cleanPath string, allowMissing bool) error {
...
+
+	// When using a standalone custom transfer agent there is no LFS server.
 	if v, _ := cfg.Git.Get("lfs.standalonetransferagent"); len(v) > 0 {
-		// When using a standalone custom transfer agent there is no LFS server.
 		return verifyStateDisabled
 	}
-
-	uc := config.NewURLConfig(cfg.Git)
+	if _, ok := uc.Get("lfs", endpoint.Url, "standalonetransfer"); ok {
+		return verifyStateDisabled
```
technoweenie: This should negate the need to run `cfg.Git.Get()` above, as the URL Config object will fall back to `lfs.standalonetransfer` if no match for the given endpoint url is in your config.

lfsapi/endpoint.go
```diff
@@ -11,11 +11,12 @@ const UrlUnknown = "<unknown>"
...
-	SshUserAndHost string
-	SshPath        string
-	SshPort        string
-	Operation      string
+	Url                string
+	SshUserAndHost     string
+	SshPath            string
+	SshPort            string
+	Operation          string
+	StandaloneTransfer string
```
technoweenie: I don't think this extra `StandaloneTransfer` field is needed. It looks like it's used below to return fake Batch API responses, when that case is already handled in the `TransferQueue`:

https://github.com/git-lfs/git-lfs/blob/5703b808220103063ededf5c6149c93201be9e07/tq/transfer_queue.go#L412-L421

I think you can use the UrlConfig object during the initialization of the `tq.Manifest` instead, like the check you added in `getVerifyStateFor()`.

https://github.com/aleb/git-lfs/blob/4949d921400da0b7669dc3467e83a5d25042ac29/tq/manifest.go#L73

test/test-custom-transfers.sh
```diff
@@ -192,3 +192,88 @@ begin_test "custom-transfer-standalone"
...
+  setup_remote_repo "$reponame"
+
+  # clone directly, not through lfstest-gitserver
+  clone_repo_url "$REMOTEDIR/$reponame.git" $reponame
+
+  # set up custom transfer adapter to use a specific transfer agent, using a URL prefix match
+  git config lfs.customtransfer.testcustom.path lfstest-standalonecustomadapter
+  git config lfs.customtransfer.testcustom.concurrent false
+  git config remote.origin.lfsurl https://git.example.com/example/path/to/repo
+  git config lfs.https://git.example.com/example/path/.standalonetransfer testcustom
```
technoweenie: It may make sense to add a nonsense `lfs.standalonetransfer` value to ensure that LFS is preferring the specific `lfs.https://git.example.com/example/path/.standalonetransfer` key.

sprohaska:
I've changed the implementation to determine the `standalonetransferagent` in `tq.Manifest` and leave `Endpoint` alone. `operation` and `remote` are now passed to the `tq.Manifest` constructor, so that it can use it to determine the endpoint URL and perform the URL prefix match.

Further changes:

* split into 4 commits.
* `tq.Manifest.IsStandaloneTransfer()` to decide whether to skip lock check.
* A single paragraph in `custom-transfers.md`.

PR #2598, opened 2017-09-19
don't fetch for unborn repositories

### LINKED ISSUE #2592: git-lfs clone with blank repository tries to pull HEAD
Version: `2.3.0`
Git Version: `2.14.1.windows.1`
OS: Windows `10.0.15063`

Log file:

```
$ git-lfs clone --recursive --progress -- https://github.com/shiftkey/blank-repo.git C:\Users\shiftkey\Documents\GitHub\blank-repo
Could not pull

Git can't resolve ref: "HEAD"

Current time in UTC:
2017-09-18 02:40:00

ENV:
LocalWorkingDir=C:\Users\shiftkey\Documents\GitHub\blank-repo
LocalGitDir=C:\Users\shiftkey\Documents\GitHub\blank-repo\.git
LocalGitStorageDir=C:\Users\shiftkey\Documents\GitHub\blank-repo\.git
LocalMediaDir=C:\Users\shiftkey\Documents\GitHub\blank-repo\.git\lfs\objects
LocalReferenceDir=
TempDir=C:\Users\shiftkey\Documents\GitHub\blank-repo\.git\lfs\tmp
ConcurrentTransfers=3
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=C:\Users\shiftkey\Documents\GitHub\blank-repo\.git\lfs
AccessDownload=none
AccessUpload=none
DownloadTransfers=basic
UploadTransfers=basic
GIT_ASKPASS=C:\Users\shiftkey\src\desktop\dist\GitHubDesktop-dev-win32-x64\resources\app\static\ask-pass-trampoline.bat
GIT_CONFIG_PARAMETERS='credential.helper='
GIT_EXEC_PATH=C:\Users\shiftkey\src\desktop\dist\GitHubDesktop-dev-win32-x64\resources\app\git\mingw64\libexec\git-core
GIT_TERMINAL_PROMPT=0
LOCAL_GIT_DIRECTORY=C:\Users\shiftkey\src\desktop\dist\GitHubDesktop-dev-win32-x64\resources\app\git

Client IP addresses:
fdb2:2c26:f4e4:0:24ae:5968:eb4f:e47 fdb2:2c26:f4e4:0:3440:8222:75f8:51cd fe80::24ae:5968:eb4f:e47 10.211.55.20
```

I'd like to understand the behaviour here:

 - the first error is coming from [this panic](https://github.com/git-lfs/git-lfs/blob/master/commands/command_pull.go#L46) which seems deliberate
 - the inner error is [from here](https://github.com/git-lfs/git-lfs/blob/1a1c679186cdd2c60bc043f6132cc87db001bddc/git/git.go#L197-L203), which is also fine
 - it seems like we should check that `HEAD` exists before doing [this `pull`](https://github.com/git-lfs/git-lfs/blob/master/commands/command_clone.go#L95) and maybe log a different message to indicate that we've skipped any post-clone network operations

If that all makes sense, I'd be happy to propose a patch. Alternatively, if this should be addressed further down the stack (to make `pull` a no-op in this situation) I can dig deeper into that.

shiftkey:
Fixes #2592 

This currently works by checking for `HEAD` ref after a clone and quietly moving on, but I'm open to different approaches to tackling this if we want to make the actions here more resilient and less noisy for this error case.

Any pointers for adding new tests to catch this regression would also be welcome!

### DIFF
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -101,8 +101,6 @@ func cloneCommand(cmd *cobra.Command, args []string) {
 				Exit("Error performing 'git lfs pull' for submodules: %v", err)
 			}
 		}
-	} else {
-		Print("Repository is unborn, unable to fetch any LFS assets")
 	}
 
 	if !cloneSkipRepoInstall {
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -85,9 +85,7 @@ func cloneCommand(cmd *cobra.Command, args []string) {
 		remote = "origin"
 	}
 
-	ref, err := git.CurrentRef()
-
-	if ref != nil {
+	if ref, err := git.CurrentRef(); err == nil {
 		includeArg, excludeArg := getIncludeExcludeArgs(cmd)
 		filter := buildFilepathFilter(cfg, includeArg, excludeArg)
 		if cloneFlags.NoCheckout || cloneFlags.Bare {
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -101,6 +101,8 @@ func cloneCommand(cmd *cobra.Command, args []string) {
 				Exit("Error performing 'git lfs pull' for submodules: %v", err)
 			}
 		}
+	} else {
+		Print("Repository is unborn, unable to fetch any LFS assets")
 	}
 
 	if !cloneSkipRepoInstall {
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -85,17 +85,21 @@ func cloneCommand(cmd *cobra.Command, args []string) {
 		remote = "origin"
 	}
 
-	includeArg, excludeArg := getIncludeExcludeArgs(cmd)
-	filter := buildFilepathFilter(cfg, includeArg, excludeArg)
-	if cloneFlags.NoCheckout || cloneFlags.Bare {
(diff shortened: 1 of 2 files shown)


commands/command_clone.go
```diff
@@ -85,18 +85,24 @@ func cloneCommand(cmd *cobra.Command, args []string) {
...
-		cfg.CurrentRemote = remote
-		fetchRef("HEAD", filter)
-	} else {
-		pull(remote, filter)
-		err := postCloneSubmodules(args)
-		if err != nil {
-			Exit("Error performing 'git lfs pull' for submodules: %v", err)
+	ref, err := git.CurrentRef()
+
+	if ref != nil {
```
technoweenie: It's a common Go idiom to check for `err == nil` instead of `ref != nil`. You could even use `if ref, err := git.CurrentRef(); err == nil {`.

shiftkey:
```sh
    echo >&2 "fatal: expected clone to succeed ..."
    exit 1
  fi
  exit 1
```

i see what you did there @technoweenie 😈

shiftkey:
:apple:

PR #2626, opened 2017-09-27
don't replace pointers with objects if clean filter is not configured

### LINKED ISSUE #2472: git lfs usage doesn't fail if git lfs hasn't been installed
`git lfs pull` will sort of work even if you haven't run `git lfs install`, leading to some confusing behavior. Specifically, it will pull, but the changes will show up as staged changes in your usual git repo.

Instead, running any lfs commands prior to running `git lfs install` should print some sort of warning, telling you that you need to install it first.

technoweenie:
This prevents LFS from replacing LFS pointers with the actual content if the `clean` filter is not configured. Otherwise, this leads to a dirty `git status`. Fixes #2472

### DIFF
--- a/commands/pull.go
+++ b/commands/pull.go
@@ -20,12 +20,10 @@ import (
 func newSingleCheckout(gitEnv config.Environment, remote string) abstractCheckout {
 	manifest := getTransferManifestOperationRemote("download", remote)
 
-	//
 	clean, ok := gitEnv.Get("filter.lfs.clean")
 	if !ok || len(clean) == 0 {
 		return &noOpCheckout{manifest: manifest}
 	}
-	// */
 
 	// Get a converter from repo-relative to cwd-relative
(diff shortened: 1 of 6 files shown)


commands/pull.go
```diff
@@ -16,7 +17,16 @@ import (
...
-func newSingleCheckout() *singleCheckout {
+func newSingleCheckout(gitEnv config.Environment, remote string) abstractCheckout {
+	manifest := getTransferManifestOperationRemote("download", remote)
+
+	//
+	clean, ok := gitEnv.Get("filter.lfs.clean")
+	if !ok || len(clean) == 0 {
+		return &noOpCheckout{manifest: manifest}
+	}
+	// */
```
rubyist: Should these empty comments be removed?

technoweenie: Ahh yes, thanks. I was using that to ensure the tests were testing things correctly :)

PR #2647, opened 2017-10-04
potentially invoke lfs for 'git update-index'

technoweenie:
Brings LFS back for `git update-index` calls. This fixes issues where `git status` is unclean after a `git lfs pull` or `git lfs clone`.

/cc https://github.com/git-lfs/git-lfs/issues/2646#issuecomment-334204675

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -459,7 +459,7 @@ func DefaultRemote() (string, error) {
 }
 
 func UpdateIndexFromStdin() *subprocess.Cmd {
-	return gitNoLFS("update-index", "-q", "--refresh", "--stdin")
+	return git("update-index", "-q", "--refresh", "--stdin")
 }
 
 type gitConfig struct {
--- a/test/test-clone.sh
+++ b/test/test-clone.sh
@@ -123,7 +123,7 @@ begin_test "cloneSSL"
 
   newclonedir="testcloneSSL1"
   git lfs clone "$SSLGITSERVER/$reponame" "$newclonedir" 2>&1 | tee lfsclone.log
-  #assert_clean_status
+  assert_clean_status
   grep "Cloning into" lfsclone.log
   grep "Git LFS:" lfsclone.log
   # should be no filter errors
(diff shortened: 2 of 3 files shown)


test/test-clone.sh
```diff
@@ -119,6 +123,7 @@ begin_test "cloneSSL"
 
   newclonedir="testcloneSSL1"
   git lfs clone "$SSLGITSERVER/$reponame" "$newclonedir" 2>&1 | tee lfsclone.log
+  #assert_clean_status
```
larsxschneider: Is this commented intentionally?

git/git.go
```diff
@@ -459,7 +459,7 @@ func DefaultRemote() (string, error) {
 }
 
 func UpdateIndexFromStdin() *subprocess.Cmd {
-	return gitNoLFS("update-index", "-q", "--refresh", "--stdin")
+	return git("update-index", "-q", "--refresh", "--stdin")
```
larsxschneider: OK, here is my understanding:

We hit this code path *only* if we manually resolve the Git LFS pointers. Manually here means: we don't use the Git clean/smudge mechanism.  This happens in the `git lfs clone ...` and `git lfs pull` calls. In these calls the following happens:

1. Git checks out all files without running the Git LFS filters. We have the Git LFS pointer files checked out as a result.
2. Git LFS iterates over all pointer files and replaces them with the actual content
3. Git LFS asks Git to re-index the changed content (this happens with the `update-index` call above). During the re-index Git needs to `clean` the real content to calculate the pointer file and compare that against the version stored in the repository. This way `git status` can tell us if the file was changed or not.

--> I think it makes sense to run Git with the clean filter here.

technoweenie:
Thanks for the review :metal:

larsxschneider:
Your welcome! Do you know when 2.3.3 will be released with the fix?

technoweenie:
I have some stuff this morning, but I could release it this afternoon. Though, I don't think a friday afternoon release is a good idea, so I'm leaning towards Monday. What do you think?

larsxschneider:
I trust your judgement here. I agree with the Friday argument. However, this is a pretty nasty bug, too.

PR #2659, opened 2017-10-10
handle scenario where multiple configuration values exist in ~/.gitconfig

shiftkey:
Downstream issue: https://github.com/desktop/desktop/issues/2984

```
Error running git [config --global filter.lfs.clean git-lfs clean -- %f]: 'warning: filter.lfs.clean has multiple values
error: cannot overwrite multiple values with a single value
       Use a regexp, --add or --replace-all to change filter.lfs.clean.' 'exit status 5'
Run `git lfs install --force` to reset git config.
```

`--add` and `--replace-all` are vintage flags for `git config`, but it seems like something in a recent Git update (2.14.2? need to confirm) is enforcing this check check better to ensure changes are lax (this bit us in a similar way when we were updating global config in https://github.com/desktop/desktop/issues/2945).

Wrapping my head around how flow `--replace-all` through when `--force` is set in a way that makes sense - shout if you have a better way to do this.

### DIFF
--- a/commands/command_update.go
+++ b/commands/command_update.go
@@ -31,7 +31,7 @@ func updateCommand(cmd *cobra.Command, args []string) {
 		switch value {
 		case "basic":
 		case "private":
-			git.Config.SetLocal("", key, "basic", false)
+			git.Config.SetLocal("", key, "basic")
 			Print("Updated %s access from %s to %s.", matches[1], value, "basic")
 		default:
 			git.Config.UnsetLocalKey("", key)
--- a/commands/command_update.go
+++ b/commands/command_update.go
@@ -31,7 +31,7 @@ func updateCommand(cmd *cobra.Command, args []string) {
 		switch value {
 		case "basic":
 		case "private":
-			git.Config.SetLocal("", key, "basic")
+			git.Config.SetLocal("", key, "basic", false)
 			Print("Updated %s access from %s to %s.", matches[1], value, "basic")
 		default:
 			git.Config.UnsetLocalKey("", key)
--- a/commands/uploader.go
+++ b/commands/uploader.go
@@ -486,6 +486,6 @@ func disableFor(endpoint lfsapi.Endpoint) error {
 
 	key := strings.Join([]string{"lfs", endpoint.Url, "locksverify"}, ".")
 
-	_, err := git.Config.SetLocal("", key, "false")
+	_, err := git.Config.SetLocal("", key, "false", false)
 	return err
 }
--- a/commands/uploader.go
+++ b/commands/uploader.go
@@ -486,6 +486,6 @@ func disableFor(endpoint lfsapi.Endpoint) error {
 
 	key := strings.Join([]string{"lfs", endpoint.Url, "locksverify"}, ".")
 
-	_, err := git.Config.SetLocal("", key, "false", false)
+	_, err := git.Config.SetLocal("", key, "false")
 	return err
 }
--- a/lfs/attribute.go
+++ b/lfs/attribute.go
@@ -80,15 +80,15 @@ func (a *Attribute) set(key, value string, upgradeables []string, opt InstallOpt
 		if opt.Local {
 			// ignore error for unset, git returns non-zero if missing
 			git.Config.UnsetLocalKey("", key)
-			_, err = git.Config.SetLocal("", key, value, opt.Force)
+			_, err = git.Config.SetLocal("", key, value)
 		} else if opt.System {
 			// ignore error for unset, git returns non-zero if missing
 			git.Config.UnsetSystem(key)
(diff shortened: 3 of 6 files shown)


xt0rted:
I was running 2.14.1 when the error in https://github.com/desktop/desktop/issues/2984 occurred. I'm not sure if it started in a prior version though, but I did have 2.14.1 installed and then updated to 2.14.2 (running `git --version` still showed 2.14.1 so I'm not sure what happened there).

commands/command_update.go
```diff
@@ -31,7 +31,7 @@ func updateCommand(cmd *cobra.Command, args []string) {
 		switch value {
 		case "basic":
 		case "private":
-			git.Config.SetLocal("", key, "basic")
+			git.Config.SetLocal("", key, "basic", false)
```
technoweenie: This sets `lfs.{url}.access=basic`.

commands/uploader.go
```diff
@@ -486,6 +486,6 @@ func disableFor(endpoint lfsapi.Endpoint) error {
 
 	key := strings.Join([]string{"lfs", endpoint.Url, "locksverify"}, ".")
 
-	_, err := git.Config.SetLocal("", key, "false")
+	_, err := git.Config.SetLocal("", key, "false", false)
```
technoweenie: `lfs.{url}.locksverify=false`

lfsapi/endpoint_finder.go
```diff
@@ -217,7 +217,7 @@ func (e *endpointGitFinder) SetAccess(rawurl string, access Access) {
 		git.Config.UnsetLocalKey("", key)
 		e.urlAccess[accessurl] = NoneAccess
 	default:
-		git.Config.SetLocal("", key, string(access))
+		git.Config.SetLocal("", key, string(access), false)
```
technoweenie: `lfs.{url}.access=?`

lfs/attribute.go
```diff
@@ -80,15 +80,15 @@ func (a *Attribute) set(key, value string, upgradeables []string, opt InstallOpt
...
 		} else if opt.System {
 			// ignore error for unset, git returns non-zero if missing
 			git.Config.UnsetSystem(key)
-			_, err = git.Config.SetSystem(key, value)
+			_, err = git.Config.SetSystem(key, value, opt.Force)
 		} else {
 			// ignore error for unset, git returns non-zero if missing
 			git.Config.UnsetGlobal(key)
-			_, err = git.Config.SetGlobal(key, value)
+			_, err = git.Config.SetGlobal(key, value, opt.Force)
```
technoweenie: These set the `filter.lfs.{type}` keys, which are not LFS specific. However, I can't imagine what Git would even do with multiple values here.

shiftkey:
@xt0rted the version of Git we use is separate to the one you're using on the command line.

This [recent mailing list thread](https://marc.info/?l=git&m=150728822930787&w=2) seems eerily related but I still haven't traced back where this behaviour might have been introduced. Guess I need to install some older versions of Git, but that's more for my curiosity than anything important.

@technoweenie 

> Though, is there any reason that LFS shouldn't always use `--replace-all`? 

I tied it to the presence of `--force` mostly for compatibility reasons, and because I'm generally nervous about overwriting config values, but if we're only touching LFS config values here then I'm fine with making it the default behaviour.

lfs/attribute.go
shiftkey: Yeah, the more I think about it the more I think just overwriting is a safer bet for things that Git LFS relies on.

shiftkey:
:apple:

technoweenie:
> I tied it to the presence of --force mostly for compatibility reasons, and because I'm generally nervous about overwriting config values, but if we're only touching LFS config values here then I'm fine with making it the default behaviour.

I think that's all legit. Though in this specific case, the LFS `git` package is not meant to be a general purpose thing. So I think this will work fine.

sactiw13:
I have git lfs 2.3.4 version installed on a Ubuntu 16.04 .5 LTS machine, but I am still getting this issue while applying git patch: 

admin@ub_16_client:~$ **git lfs env**
git-lfs/2.3.4 (GitHub; linux amd64; go 1.8.1)
git version 2.7.4

**Error description:**
```
Checking patch common/container.img...
Applied patch common/container.img cleanly.
warning: filter.lfs.smudge has multiple values
error: cannot overwrite multiple values with a single value
       Use a regexp, --add or --replace-all to change filter.lfs.smudge.
```

Note: git version on the machine is pretty old 2.7.4, could that be a issue i.e. needs to be updated?

bk2204:
Hey, if you're still seeing this problem, could you please open a new issue?  That way we can help identify any issues that might still be occurring that are different from what's been fixed here.

PR #2676, opened 2017-10-18
Git config refactor

technoweenie:
Started working on something, got disgusted by the global config objects, so I started some refactoring.


technoweenie:
Another step towards refactoring the `config` package. The ultimate goal is to remove references to global values like `config.Config` and `git.Config`. The only exception is the `commands` package can have a global `cfg *config.Configuration` var.

* `git.Config` is now part of `*config.Configuration`, with helpers for comparing the git version and updating the git config.
* The `FetchPruneConfig` was moved to the `lfs` package.
* `localstorage` config no longer depends on the global `config.Config`.
* `(*config.Configuration) unmarshal()` was removed. It was only used in a few spots, so it didn't seem worth the LOC for the implementation.
* hooks and filters are loaded on demand, no longer depend on the global `config.Config`.

Issue #2690, opened 2017-10-25
remove unnecessary 'git credential approve' calls


technoweenie:
> It is still true, though, that git credential approve is executed in a new process for each transferred file (on checkout), correct? Note that it is not re-prompting for credentials--it is getting cached credentials, but it seems to still re-authorize for each file transfer, and after each successful authorization, it calls 'credential approve' to re-cache the approved credentials. On Windows, this overhead raises the bar for which types of files are worth tracking with LFS. For example, it was not worth using LFS to track DLLs in a particular repo of mine, because (1) the DLLs often compressed moderately well, and (2) the file sizes varied widely enough that transferring many smaller/medium DLLs via LFS took longer, in large part due to the credential approve calls.

/cc https://github.com/git-lfs/git-lfs/issues/2328#issuecomment-339361087

technoweenie:
I was able to replicate this with a bit of extra tracing. This patch was applied to the current `master` (which is beyond LFS v2.3.4), and run with Git v2.14.3. Minus the custom patches, this should match behavior with any recent LFS that has the credential cache enabled. It was enabled by default in LFS v2.3.3+.

```diff
diff --git a/lfsapi/creds.go b/lfsapi/creds.go
index 187043b1..c6839e83 100644
--- a/lfsapi/creds.go
+++ b/lfsapi/creds.go
@@ -307,6 +307,8 @@ func (c *credentialCacher) Reject(creds Creds) error {
 }

 func (c *credentialCacher) Approve(creds Creds) error {
+       tracerx.Printf("creds: git credential cache approve (%q, %q, %q)",
+               creds["protocol"], creds["host"], creds["path"])
        err := c.helper.Approve(creds)
        if err == nil {
                c.cmu.Lock()
@@ -332,6 +334,8 @@ func (h *commandCredentialHelper) Reject(creds Creds) error {
 }

 func (h *commandCredentialHelper) Approve(creds Creds) error {
+       tracerx.Printf("creds: git credential helper approve (%q, %q, %q)",
+               creds["protocol"], creds["host"], creds["path"])
        _, err := h.exec("approve", creds)
        return err
 }
```

## Git Credential Behavior

First, it seems that Git itself always calls `store` on the helper itself. I'm not sure if Git is interacting directly with the credential helper, or if it's going through `git credential` and it's just not showing up in the trace. My hunch is the latter.

```sh
$ GIT_TRACE=1 git clone https://my-git-server/technoweenie/lfs-pull-bug
[...]

Issue #2686, opened 2017-10-24
Add support for per-url credential helpers


technoweenie:
LFS doesn't detect credential helpers set on a url:

```
[credential "https://my-git-server.com/"]
        helper = /path/to/helper
```

The problem is that LFS loads the cred helpers when an `*lfsapi.Client` is initialized:

https://github.com/git-lfs/git-lfs/blob/3480fa480e135a34166ebf4b78f5c8f267f0e9d7/lfsapi/lfsapi.go#L69-L72

Instead, it needs to happen at request time here, using `*config.URLConfig`:

https://github.com/git-lfs/git-lfs/blob/3480fa480e135a34166ebf4b78f5c8f267f0e9d7/lfsapi/auth.go#L23-L27

Issue #2634, opened 2017-09-30
lfs 2.3.x seems to prefer askpass to credential helper


zvinless:
When pushing/pulling lfs files with both a credential helper (`osxkeychain`) and a GIT_ASKPASS environment variable set, it seems like lfs tries to use the askpass program. From what I understand this is [incongruent with git](https://git-scm.com/docs/gitcredentials#_requesting_credentials) (though I am honestly not very familiar with how git deals with credentials).

Steps:
1. Have valid credentials in macOS keychain such that you can clone your own GitHub repo
1. Execute `git config --global credential.helper osxkeychain` to set global credential helper
1. Clone one of your GitHub repos with lfs files
1. Set `GIT_ASKPASS` environment variable in ~/.bash_profile to an empty shell script so that it will fail
1. Commit an lfs tracked file to repo
1. Try to execute `git push`
1. Observe something along the lines of: `Git credentials for [repo].git not found:
fork/exec [path/to/askpass/program]: exec format error` (indicating that the askpass program is getting used; don't worry that it fails)
1. Remove GIT_ASKPASS variable from ~/.bash_profile and observe that you can now push the commit
1. Redo these steps with a plain GitHub repo (no lfs) and observe everything works fine the first time

I first noticed this because Visual Studio Code's git integration sets the GIT_ASKPASS environment variable and provides its own askpass program, which started getting used instead of my keychain credentials exclusively since updating to git lfs 2.3.0.

env:
```
macOS 10.13
git-lfs/2.3.1 (GitHub; darwin amd64; go 1.9)
git version 2.14.2

Endpoint=https://github.com/zvinless/large-files.git/info/lfs (auth=basic)
LocalWorkingDir=/Users/zvinless/Desktop/large-files
LocalGitDir=/Users/zvinless/Desktop/large-files/.git
LocalGitStorageDir=/Users/zvinless/Desktop/large-files/.git
LocalMediaDir=/Users/zvinless/Desktop/large-files/.git/lfs/objects
LocalReferenceDir=
TempDir=/Users/zvinless/Desktop/large-files/.git/lfs/tmp
ConcurrentTransfers=3
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
FetchRecentRefsIncludeRemotes=true
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
[...]

PR #2695, opened 2017-10-26
Cred helper improvements

technoweenie:
This PR adds support for per-url credential helpers (#2686) and fixes performance issues discussed in #2690.

* [x] Teach the LFS credential cacher to only call approve the first time. LFS should only ever call git credential approve for a given url/path once in the lifetime of a single LFS command.
* [x] Teach the LFS HTTP client to only call Approve() on the credential helper if it was used for a request. I don't know how other LFS servers work, but GitHub's doesn't require it for transferring specific objects.
* [x] Add approve tracing to LFS so we can confirm a fix.
* [x] Teach LFS to only send `path` to credential helpers if `credential.useHttpPath` is enabled.
* [x] If askpass is set, fall back to credential helper if it's misconfigured.
* [x] Check for `credential.<url>.usehttppath` and `credential.<url>.helper`

By creating the credential helper per request, any notion of a cache is lost. So, there's still more work to be done in `(*Client) getCredentialHelper`.


technoweenie:
Here are some sample builds in case anyone wants to try this out:

[git-lfs-darwin-amd64-2.3.4.tar.gz](https://github.com/git-lfs/git-lfs/files/1422775/git-lfs-darwin-amd64-2.3.4.tar.gz)
[git-lfs-linux-amd64-2.3.4.tar.gz](https://github.com/git-lfs/git-lfs/files/1422776/git-lfs-linux-amd64-2.3.4.tar.gz)
[git-lfs-windows-amd64-2.3.4.zip](https://github.com/git-lfs/git-lfs/files/1422777/git-lfs-windows-amd64-2.3.4.zip)

wouterhund:
As mentioned in https://github.com/git-lfs/git-lfs/issues/2634#issuecomment-340069678 I tested the linux build and confirmed they resolve https://github.com/git-lfs/git-lfs/pull/2695

PR #2704, opened 2017-10-30
release lfs for ubuntu/artful too

technoweenie:
This adds ubuntu/artful to the list of distros for future LFS releases.


technoweenie:
Considering debian 10 isn't released yet, I'm going to assume that ubuntu artful is built on 9, the current stable debian release.

PR #2706, opened 2017-10-31
verify locks against each ref being pushed

technoweenie:
Implements 2 aspects of #2712:

* basic `push.default` support, which governs how Git chooses an implicit remote ref based on the remote and local ref being pushed.
* Verify locks on each ref being pushed.

Bonus: locks are not verified for `git lfs push --object-id` anymore. Since this push doesn't affect any Git code, there's no need to verify anything.

This also extracts a `*lockVerifier` and `*refUpdate`. There is still too much coupling, but it's better than everything being in `*uploadContext`. Baby steps....

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -166,35 +166,6 @@ func (c *Configuration) CurrentRef() *git.Ref {
 	return c.ref
 }
 
-func (c *Configuration) RemoteRef() *git.Ref {
-	r := c.CurrentRef()
-
-	c.loading.Lock()
-	defer c.loading.Unlock()
-
-	if c.remoteRef != nil {
-		return c.remoteRef
-	}
-
-	if r != nil {
-		merge, _ := c.Git.Get(fmt.Sprintf("branch.%s.merge", r.Name))
-		if len(merge) > 0 {
-			c.remoteRef = git.ParseRef(merge, "")
-		} else {
-			c.remoteRef = r
-		}
-	}
-
-	return c.remoteRef
-}
-
-func (c *Configuration) RemoteRefName() string {
-	if r := c.RemoteRef(); r != nil {
-		return r.Name
-	}
-	return ""
-}
-
 func (c *Configuration) IsDefaultRemote() bool {
 	return c.Remote() == defaultRemote
 }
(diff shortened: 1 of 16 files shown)


technoweenie:
I ran into a snag supporting the locks/verify endpoint. Pushes can either specify a remote ref, or you can push all the refs.

```sh
$ git push origin local-branch:remote-branch

# calls the pre-push hook by passing this to STDIN:

echo "refs/heads/local-branch AFTER-SHA refs/heads/remote-branch BEFORE-SHA" | git lfs pre-push
```

That leaves us with some options:

1. Call the `locks/verify` endpoint all the refs as a json array. This leads to some small inconsistencies with the API, with some endpoints sending a `ref` string, and others sending a `refs` array.
2. Call the `locks/verify` endpoint for each ref. It adds more locks calls, but also keeps the API consistent and gives the server implementation the ability to return filtered locked files per ref, if desired.

I'm currently leaning towards option 2.

technoweenie:
#2700 implements this portion of the [`git push` docs](https://git-scm.com/docs/git-push):

> When the command line does not specify where to push with the `<repository>` argument, `branch.*.remote` configuration for the current branch is consulted to determine where to push. If the configuration is missing, it defaults to origin.

Now, it needs to implement the following:

> When the command line does not specify what to push with `<refspec>...` arguments or `--all`, `--mirror`, `--tags` options, the command finds the default `<refspec>` by consulting `remote.*.push` configuration, and if it is not found, honors `push.default` configuration to decide what to push (See [git-config](https://git-scm.com/docs/git-config) for the meaning of `push.default`).
>
> When neither the command-line nor the configuration specify what to push, the default behavior is used, which corresponds to the simple value for `push.default`: the current branch is pushed to the corresponding upstream branch, but as a safety measure, the push is aborted if the upstream branch does not have the same name as the local one.

config/config.go
```diff
@@ -165,6 +166,38 @@ func (c *Configuration) CurrentRef() *git.Ref {
...
+
+	if c.remoteRef != nil {
+		return c.remoteRef
+	}
+
+	if r != nil {
+		merge, _ := c.Git.Get(fmt.Sprintf("branch.%s.merge", r.Name))
+		if strings.HasPrefix(merge, "refs/heads/") {
+			c.remoteRef = &git.Ref{
+				Name: merge[11:],
```
lonelone85: Caspernaing

Issue #2413, opened 2017-07-12
Comment `# filter=lfs` in .gitattributes is interpreted by Git LFS


jokram:
### Steps to reproduce
1. In the `.gitattributes` put the following lines:
```
# filter=lfs
*.zip	filter=lfs diff=lfs merge=lfs -text
```
2. Call `git lfs track` to list the patterns:
```
$ git lfs track
Listing tracked patterns
    # (.gitattributes)
    *.zip (.gitattributes)
```
3. If you slighlty change the comment in `.gitattributes` to:
```
# filter=lf
*.zip	filter=lfs diff=lfs merge=lfs -text
```
then it works properly:
```
$ git lfs track
Listing tracked patterns
    *.zip (.gitattributes)
```

### Conclusion
It seems that the line starting with the hash will be interpreted by Git LFS, but should be ignored as comment

### Environment
```
git-lfs/2.1.1 (GitHub; windows amd64; go 1.8.1; git 3314e287)
git version 2.13.2.windows.1
```

ttaylorr:
Hi @jokram, thanks for opening this. Good catch pointing out this bug. I think the issue lies [here](https://github.com/git-lfs/git-lfs/blob/v2.2.1/git/attribs.go#L60) (via [this](https://github.com/git-lfs/git-lfs/blob/v2.2.1/commands/command_track.go#L52) call). The easiest change that I can think to make is to apply a diff like:

```diff
diff --git a/git/attribs.go b/git/attribs.go
index 16321b9f..4806c903 100644
--- a/git/attribs.go
+++ b/git/attribs.go
@@ -57,7 +57,7 @@ func GetAttributePaths(workingDir, gitDir string) []AttributePath {
 
 		for scanner.Scan() {
 			line := scanner.Text()
-			if strings.Contains(line, "filter=lfs") {
+			if !strings.HasPrefix(line, "#") && strings.Contains(line, "filter=lfs") {
 				fields := strings.Fields(line)
 				pattern := fields[0]
 				if len(reldir) > 0 {
```

But I wonder if LFS could do something more robust. Given:

```gitattributes
# *.dat filter=lfs diff=lfs merge=lfs -text
```

and:

```bash
$ git lfs track "*.dat"
```

The `.gitattributes` should be rewritten as uncommenting the commented tracked path:

```gitattributes
*.dat filter=lfs diff=lfs merge=lfs -text
```

What do you think we should do in this case?

technoweenie:
[...]

PR #2708, opened 2017-11-01
Ignore comment attr lines

mathstuf:
---
This only addresses the initial description of #2413. It does not fix `[attr]` behavior or checking for `-filter=lfs`.

### DIFF
--- a/git/attribs.go
+++ b/git/attribs.go
@@ -56,7 +56,11 @@ func GetAttributePaths(workingDir, gitDir string) []AttributePath {
 		scanner.Split(le.ScanLines)
 
 		for scanner.Scan() {
-			line := scanner.Text()
+			line := strings.TrimSpace(scanner.Text())
+
+			if strings.HasPrefix(line, "#") {
+				continue
+			}
 
 			// Check for filter=lfs (signifying that LFS is tracking
 			// this file) or "lockable", which indicates that the
--- a/git/attribs.go
+++ b/git/attribs.go
@@ -36,7 +36,7 @@ func (s *AttributeSource) String() string {
 
 // GetAttributePaths returns a list of entries in .gitattributes which are
 // configured with the filter=lfs attribute
-// workingDIr is the root of the working copy
+// workingDir is the root of the working copy
 // gitDir is the root of the git repo
 func GetAttributePaths(workingDir, gitDir string) []AttributePath {
 	paths := make([]AttributePath, 0)


git/attribs.go
```diff
@@ -58,6 +58,12 @@ func GetAttributePaths(workingDir, gitDir string) []AttributePath {
 		for scanner.Scan() {
 			line := scanner.Text()
 
+			line := strings.TrimSpace(line)
```
technoweenie: The [tests](https://ci.appveyor.com/project/lfs-build/git-lfs/build/1.0.2361) show a compilation error here:

```
git\attribs.go:61: no new variables on left side of :=
git\attribs.go:63: not enough arguments in call to strings.HasPrefix
	have (string)
	want (string, string)
```

The `line` var is being initialized on line 59, and Go won't let you re-initialize it with `:=` on line 61. One of these should fix it:

```go
line := scanner.Text()
line = strings.TrimSpace(line)

// 1 liner is fine too
line := strings.TrimSpace(scanner.Text())
```

git/attribs.go
```diff
@@ -58,6 +58,12 @@ func GetAttributePaths(workingDir, gitDir string) []AttributePath {
...
+
+			if strings.HasPrefix("#") {
+				continue
```
technoweenie: The [tests](https://ci.appveyor.com/project/lfs-build/git-lfs/build/1.0.2361) show a compilation error here:

```
git\attribs.go:63: not enough arguments in call to strings.HasPrefix
	have (string)
	want (string, string)
```

The [docs](https://golang.org/pkg/strings/#HasPrefix) show a function signature of: `strings.HasPrefix(s, prefix string) bool`. In this case, `line` is the string, and `"#"` is the prefix you're checking for:

```go
if strings.HasPrefix(line, "#") {
```

mathstuf:
Fixed.

Sorry about the compile errors, I'm not that familiar with Go and `go build` is complaining about things like:

```
git-lfs.go:12:2: cannot find package "github.com/git-lfs/git-lfs/commands" in any of:
        /usr/lib/golang/src/github.com/git-lfs/git-lfs/commands (from $GOROOT)                                                                                                                                                                                                                                                
        /home/boeckb/go/src/github.com/git-lfs/git-lfs/commands (from $GOPATH) 
```

which makes no sense to me…

technoweenie:
@mathstuf Ah, the problem is that you're not working in a GOPATH, so Go doesn't know where to find the package. What I do for go repos, is run `go get github.com/git-lfs/git-lfs`, which clones it to `$GOPATH/src/github.com/git-lfs/git-lfs`. Then I symlink that directory to my preferred "project" directory to do my regular work. I believe Go will set your default GOPATH to `~/go`.

We really should have a little primer like this for contributors new to GO.

mathstuf:
Hmm, my setup is usually more complicated than that even since I try to put build artifacts on a different partition (to keep the main one uncluttered and an easy way to exclude them from backups with `--one-filesystem` or similar support).

My simple symlinking isn't working though, so there's probably other assumptions I'm breaking with my setup:

`$HOME/code/depot/lang-go` (`$GOPATH`)
`$HOME/code/depot/lang-go/src` (symlink to `$GOPATH`)
`$HOME/code/depot/group-vcs/git-lfs/src` (clone of this repo)

And now `go build` complains: `go install _/home/boeckb/code/depot/group-vcs/git-lfs/src: build output "src" already exists and is a directory`.

technoweenie:
I don't think you want to symlink `$HOME/code/depot/lang-go/src` to `$GOPATH`. Technically, `$HOME/code/depot/lang-go` should be your GOPATH. Go will put some compiled artifacts in `$GOPATH/bin` and `$GOPATH/pkg` too.

mathstuf:
I moved it, but that doesn't help though: same error message.

ttaylorr:
@mathstuf I gave this a whirl locally and it worked great! ✨. If you're still having trouble with your `$GOPATH`, give me a ping and I'll be happy to help!

mathstuf:
Thanks. I may have changes in the future (that probably won't be as trivial); I'll ping you there if I'm still having issues then.

PR #2709, opened 2017-11-01
Testlib no global config

mathstuf:


technoweenie:
Hey @mathstuf, what is the goal here? Is this a fix for anyone that uses `XDG_CONFIG_HOME` or has `$HOME/.config/git/config` already? 

Instead of copying `$TESTHOME/.gitconfig` to `$HOME/.config/git/config` AND `$HOME/.gitconfig`, what do you think about adding `unset XDG_CONFIG_HOME` instead?

mathstuf:
That should work too. Updated.

technoweenie:
Ah can you accept the CLA please?

mathstuf:
Where can I get a copy of the CLA that doesn't need authentication to send to the right folks here?

mathstuf:
OK, legal here OK'd the CLA. Closed and reopened to trigger checks again.

PR #2716, opened 2017-11-04
Wait longer for test lfs server to start.

QuLogic:
Also, back off on polling a bit after the first check to reduce any I/O contention that might slow things down even more. Instead of 5 seconds, this will wait up to about 60 seconds for files created by the test server to appear. This allows very slow systems to be able to pass tests.

Finally, add a message if the file doesn't appear to make it clear what went wrong.

For example, slow builders like [aarch64](https://kojipkgs.fedoraproject.org//work/tasks/291/22880291/build.log) or [armv7hl](https://kojipkgs.fedoraproject.org//work/tasks/294/22880294/build.log) mysteriously fail right after printing git(lfs) versions without any indication why.

### DIFF
--- a/test/testhelpers.sh
+++ b/test/testhelpers.sh
@@ -266,15 +266,20 @@ size %s
 wait_for_file() {
   local filename="$1"
   n=0
-  while [ $n -lt 10 ]; do
+  wait_time=1
+  while [ $n -lt 17 ]; do
     if [ -s $filename ]; then
       return 0
     fi
 
-    sleep 0.5
+    sleep $wait_time
     n=`expr $n + 1`
+    if [ $wait_time -lt 4 ]; then
+      wait_time=`expr $wait_time \* 2`
+    fi
   done
 
+  echo "$filename did not appear after 60 seconds."
   return 1
 }
 


QuLogic:
Why are there two git-lfs CLAs? The one in `CONTRIBUTING.md` does not seem to be the one checked by the CLA bot.

technoweenie:
Thanks, this looks like a good update. I am a little concerned about the gradually increasing wait time. I suppose if compilation can take 32s, it doesn't matter if the total wait time is 35s, around 10% extra. But, something taking 4s will have to wait a total of 7s, about 75% extra. 

Did you see any kind of problem checking `[ -s $filename ]` every 0.5s, or did you bump the default to 1s since bash only understands integers?

I created #2721, with an extra debug message to see how long CI waits for these files now. My macbook pro's output looks like: 

```sh
Git LFS: /Users/rick/go/src/github.com/git-lfs/git-lfs/bin/git-lfs
git-lfs/2.3.4 (GitHub; darwin amd64; go 1.9.2; git 87b53b36)
git version 2.15.0
/Users/rick/go/src/github.com/git-lfs/git-lfs/test/remote/url: attempt: 1, wait_time: 2
/Users/rick/go/src/github.com/git-lfs/git-lfs/test/remote/sslurl: attempt: 0, wait_time: 1
/Users/rick/go/src/github.com/git-lfs/git-lfs/test/remote/clientcerturl: attempt: 0, wait_time: 1
/Users/rick/go/src/github.com/git-lfs/git-lfs/test/remote/cert: attempt: 0, wait_time: 1
/Users/rick/go/src/github.com/git-lfs/git-lfs/test/remote/client.crt: attempt: 0, wait_time: 1
/Users/rick/go/src/github.com/git-lfs/git-lfs/test/remote/client.key: attempt: 0, wait_time: 1
Git LFS initialized.
```

Also, I'm really curious why you're running this on those architectures. I wouldn't expect LFS to be helpful, but my only experience with arm processors is some raspberry pi.

technoweenie:
The CI results ([circle](https://circleci.com/gh/git-lfs/git-lfs/2533?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), [appveyor](https://ci.appveyor.com/project/lfs-build/git-lfs/build/1.0.2399), [travis](https://travis-ci.org/git-lfs/git-lfs/jobs/298729415)) matched my laptop:

> /test/remote/url: attempt: 1, wait_time: 2

I didn't do any fancy bash math, so that's saying it's gone through the loop once, and the _next_ wait time would be 2s. 

Next, I wanted to see what would happen if I set the wait time to 0.5s?

> macbook pro: attempt: 1, wait_time: 0.5
> [travis](https://travis-ci.org/git-lfs/git-lfs/jobs/298736881): attempt: 2, wait_time: 0.5
> [appveyor](https://ci.appveyor.com/project/lfs-build/git-lfs/build/1.0.2403): attempt: 2, wait_time: 0.5
> [circle](https://circleci.com/gh/git-lfs/git-lfs/2534?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link): attempt: 2, wait_time: 0.5

That all makes sense, since it's only waiting for a go command to spin up an http server. I'd prefer the wait_time at 0.5s, unless checking `[ -s $filename ]` that frequently impacts your I/O too much.

Also, I think the longest is 63s, not 60. While I don't think it matters, I'd probably change it to `~60` if I had to make a change to this file anyway.

technoweenie:
> Why are there two git-lfs CLAs? 

Oh no, I think one is pointing to the original `github/git-lfs` repository. We transferred it to the `git-lfs` org, but I guess we never completed the CLA app's transfer.

QuLogic:
I don't have any particular use for git-lfs on those arches; they're just the default arches in Fedora. And if this small tweak is all that's necessary to build there, I don't see why not.

I only switched 0.5 to 1 because bash doesn't support floating point math. It's a semi-common thing to increase polling times a bit when waiting a long time, but I'm not really married to that change. The main problem is that 5 seconds is simply too short to wait. I don't recall how long it took for these to work, but I've seen some servers take up to 30 seconds at least (but this is due to waiting for certain network responses.)

technoweenie:
@QuLogic thanks, just checking.

PR #2721, opened 2017-11-07
Qu logic wait longer

technoweenie:
Running #2716, with some extra output to see how long CI waits on these files.


technoweenie:
Just grabbing some measurements, see #2716.

Issue #2712, opened 2017-11-02
Send remote ref with API calls


technoweenie:
We need to update all LFS API requests to send the remote ref. This gives the servers opportunities to tweak how locks are stored or authorization is checked for repositories based on the ref being updated.

As of LFS v2.4.0, the client will always send a ref with API calls. Servers that want to support these fields in a future proof manner can look for the following requests:

```json
// LFS 2.3 and below
// absent 'ref' property
// servers should retain existing behavior
// POST locks/verify
{ "limit": 100 }

// LFS 2.4.0
// valid 'ref'
// servers should use ref in authz check
// POST locks/verify
{ "ref": { "name": "master" }, "limit": 100 }

// LFS 2.4.0
// empty 'ref', client had problems finding it
// servers should retain existing behavior
// POST locks/verify
{ "ref": null, "limit": 100 }
```

**Open question:** This is currently sending a relative ref name, like "master" or "feature-branch." Should it instead try to send an absolute ref name, like "refs/heads/feature-branch"?

EDIT: Yes, the client should send a full refspec to the LFS server :+1:

TODO:

* Default remote ref:
  * [x] Check `push.default` #2706
* [x] Improved remote tracking #2715 #2711
* API calls:
  * [x] locks verify #2706
    * [ ] add retries
  * [x] lock https://github.com/git-lfs/git-lfs/pull/2773
  * [x] unlock https://github.com/git-lfs/git-lfs/pull/2773
  * [x] list locks https://github.com/git-lfs/git-lfs/pull/2773
  * [x] objects/batch <del>https://github.com/git-lfs/git-lfs/pull/2774</del> https://github.com/git-lfs/git-lfs/pull/2806 #2809
* [x] Docs https://github.com/git-lfs/git-lfs/pull/2820

abdurajak:
#2722

ttaylorr:
I have a few questions that I think would be good to discuss before I work on fleshing the rest of this project out:

1. Should we expect implementations of the specification to trust the references that the client is sending with their API request? It is my understanding that if a malicious actor were to intercept the communication between Git LFS and a server (via a reverse proxy/modified client) they could send refs that they do/don't own and upload objects that the owner of a repository would otherwise not want.

[...]

PR #2722, opened 2017-11-08
Locks/verify: full refspec

technoweenie:
Updates the `locks/verify` calls to send the full refspec. /cc #2712.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -54,12 +54,8 @@ func (t RefType) Prefix() (string, bool) {
 		return "refs/tags", true
 	case RefTypeRemoteTag:
 		return "refs/remotes/tags", true
-	case RefTypeHEAD:
-		return "", false
-	case RefTypeOther:
-		return "", false
 	default:
-		panic(fmt.Sprintf("git: unknown RefType %d", t))
+		return "", false
 	}
 }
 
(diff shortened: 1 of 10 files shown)


git/git.go
```diff
@@ -95,6 +91,16 @@ type Ref struct {
 	Sha  string
 }
 
+func (r *Ref) Refspec() string {
```
ttaylorr: 👍 I agree with your comment that `Refspec` is a better name here. I think we can merge this as-is and I'll update my branch to pull in `Refspec()` over `String()`.

script/integration.go
```diff
@@ -50,17 +50,31 @@ func mainIntegration() {
 	for _, file := range files {
 		tests <- file
 	}
+	close(tests)
+
+	outputDone := make(chan bool)
+	go func() {
```
ttaylorr: Nit: this could be replaced with:

```go
go func() {
        for out := range output {
                fmt.Println(out)
        }
        outputDone <- true
}()
```

Though I am curious why this change is included in this PR?

technoweenie: I was running into an issue where `script/integration` wouldn't dump out the `test-pre-push.sh` errors for this PR. This keeps the process alive until the tests have run AND the output has been sent STDOUT.

technoweenie: I had planned on pushing that out as a separate PR. I must've accidentally merged to my local master, so it was still there when I branched off for this PR :)

I'll port the `Refspec()` change to #2734, and push out a PR for the test change, and another for this change.

ttaylorr: Rad! Thanks for porting this into #2734, I think that should be ready to go now 👍 .

technoweenie:
Don't merge. Will re-apply these changes once #2734 has merged to master.

PR #2732, opened 2017-11-16
progress: use git/githistory/log package for formatting

ttaylorr:
This pull request extends the `progress.Meter` interface to implement `git/githistory/log.Task` so that the `progress.Meter` can be driven from a `*git/githistory/log.Logger`.

Being able to use the various implementations of `progress.Meter` within the `git/githistory/log` sub-package has the following benefits:

- Separating the concept of generating a message and printing it.
- Eliminating IO-races, where two things want to log at the same time and the message is garbled, or out of order.
- Consistent use of the `git/githistory/log` package, which makes changing IO behavior in this context easier.

Since it's been awhile for me, I wanted to write down some of the ideas in the `git/githistory/log` package here:

1. The `*Logger` logs `Task`s, which provide a channel of updates, and a `Throttled()` method, telling the logger whether or not they should be throttled.
2. The `*Logger` logs a `Task` until completion, enqueuing and subsequently printing messages from other `Task`s only after the currently running Task is complete. It buffers further tasks in a pseudo-infinite buffer.

This pull request seeks to achieve the above benefits by implementing `log.Task` in `progress.Meter`, making it possible to `log.NewLogger(...).Enqueue(progress.NewMeter())`. The approach towards doing this was done in a minimal fashion, keeping the majority of the formatting code as-is, but adding a new `update()` method, which sends a new update after each time a method that would change the output of the progress meter is called.

For instance:

```go
func (p *ProgressMeter) Add(size int64) {
	defer p.update()
 	atomic.AddInt32(&p.estimatedFiles, 1)
 	atomic.AddInt64(&p.estimatedBytes, size)
}
```

will modify the `*ProgressMeter`'s state, then issue a deferred `p.update()` call, which will construct a new update, and send it. The `*log.Logger` will either accept it, or drop it on the floor if the `*ProgressMeter` is generating too many updates.

This had some other unexpected benefits like removing the `for _ = range time.NewTicker(...).C` idiom, and simplifying the `Start()` and `Pause()` implementations.

## 

/cc @git-lfs/core @technoweenie

### DIFF
--- a/git/githistory/log/log.go
+++ b/git/githistory/log/log.go
@@ -80,7 +80,7 @@ func (l *Logger) Close() {
 // Waitier creates and enqueues a new *WaitingTask.
 func (l *Logger) Waiter(msg string) *WaitingTask {
 	t := NewWaitingTask(msg)
-	l.enqueue(t)
+	l.Enqueue(t)
 
 	return t
 }
@@ -88,7 +88,7 @@ func (l *Logger) Waiter(msg string) *WaitingTask {
 // Percentage creates and enqueues a new *PercentageTask.
 func (l *Logger) Percentage(msg string, total uint64) *PercentageTask {
 	t := NewPercentageTask(msg, total)
-	l.enqueue(t)
+	l.Enqueue(t)
 
 	return t
 }
@@ -96,13 +96,13 @@ func (l *Logger) Percentage(msg string, total uint64) *PercentageTask {
 // List creates and enqueues a new *ListTask.
 func (l *Logger) List(msg string) *ListTask {
 	t := NewListTask(msg)
-	l.enqueue(t)
+	l.Enqueue(t)
 
 	return t
 }
 
-// enqueue enqueues the given Tasks "ts".
-func (l *Logger) enqueue(ts ...Task) {
+// Enqueue enqueues the given Tasks "ts".
+func (l *Logger) Enqueue(ts ...Task) {
 	if l == nil {
 		for _, t := range ts {
 			go func(t Task) {
--- a/git/githistory/log/log.go
+++ b/git/githistory/log/log.go
@@ -8,7 +8,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/git-lfs/git-lfs/tools"
 	"github.com/olekukonko/ts"
 )
 
@@ -213,7 +212,7 @@ func (l *Logger) logTask(task Task) {
 // It returns the number of bytes "n" written to the sink and the error "err",
 // if one was encountered.
 func (l *Logger) logLine(str string) (n int, err error) {
-	padding := strings.Repeat(" ", tools.MaxInt(0, l.widthFn()-len(str)))
+	padding := strings.Repeat(" ", maxInt(0, l.widthFn()-len(str)))
 
 	return l.log(str + padding + "\r")
 }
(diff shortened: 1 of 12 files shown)


git/githistory/log/log.go
```diff
@@ -225,3 +229,10 @@ func (l *Logger) logLine(str string) (n int, err error) {
 func (l *Logger) log(str string) (n int, err error) {
 	return fmt.Fprint(l.sink, str)
 }
+
+func maxInt(a, b int) int {
+	if a > b {
+		return a
+	}
+	return b
+}
```
technoweenie: Why'd you pull this out? Is there a circular dependency now?

ttaylorr: Great question, this happened when importing the `git/githistory/log` package from `progress`, since both of those packages import `tools`. This is the only reference to the `tools` package inside of `git/githistory/log`, so this was sufficient to remove the cyclic dependency.

progress/meter.go
```diff
@@ -106,29 +104,27 @@ func NewMeter(options ...meterOption) *ProgressMeter {
...
-		p.finished <- true
-	}
+	atomic.StoreUint32(&p.paused, 1)
 }
 
 // Add tells the progress meter that a single file of the given size will
 // possibly be transferred. If a file doesn't need to be transferred for some
 // reason, be sure to call Skip(int64) with the same size.
 func (p *ProgressMeter) Add(size int64) {
+	defer p.update()
```
technoweenie: What's the benefit of using `defer` here?  Couldn't you just put `p.update()` at the end of the function?

progress/spinner.go
```diff
@@ -2,50 +2,68 @@ package progress
...
 // Finish the spinner with a completion message & newline
-func (s *Spinner) Finish(out io.Writer, finishMsg string) {
-	s.msg = finishMsg
-	s.stage = 0
+func (s *Spinner) Finish(fstr string, vs ...interface{}) {
 	var sym string
 	if runtime.GOOS == "windows" {
-		// Windows console sucks, can't do nice check mark except in ConEmu (not cmd or git bash)
-		// So play it safe & boring
+		// Windows' console can't render UTF-8 check marks outside of
```
technoweenie: Ha, thanks for replacing this with a more positive comment ;)

progress/meter.go
ttaylorr: I don't think there's any semantic benefit to having this be a deferred function call, but I like that the `defer` keyword allows us to put the `p.update()` call near the top of the method as a sort of way to annotate "hey, this is going to produce some updates that will change the output of this on the terminal". Deferred calls are pretty fast nowadays (and this code is running in the context of uploading/downloading large files over the network, so I think that the performance concern would be negligible.

technoweenie: Why not just leave the `defer` off then and keep it at the top? It looks like the only reason it'd hang is if the channel is full, in which case it's okay to delay the counter update.

ttaylorr: When `update()` is deferred, it reflects an update that includes the changes made by calling `Add()`, `Skip()`, etc. I think that it would be fine to run it before, we would just get something that says `Git LFS: 0 of 0 files ...`, and would have to `update()` once more when calling `Finish()`.

I don't feel strongly either way, so am happy to change it either way!

Issue #2729, opened 2017-11-15
git lfs push failed even with -c http.extraheader="Authorization .."


TingluoHuang:
git version 2.15.0.windows.1
git-lfs/2.3.4 (GitHub; windows amd64; go 1.8.3; git d2f6752f)

I don't have git credential manager installed on my box, and i don't plan to install it.

on a repository that doesn't have any LFS object, i can use following command to do `git push`
`git -c http.extraheader="Authorization: Basic Base64EncodedUserPass" push`

however, on a repository that has LFS objects, the same command will fail with error
`LFS: Git credentials fro https://xxxxxxx not found`

I verified `git -c http.extraheader="Authorization: Basic Base64EncodedUserPass" lfs fetch` works fine.

I think this is a bug of how we handle http.extraheader option on `lfs push`.

ttaylorr:
Hi @TingluoHuang, thanks for opening this and sorry that you're having problems!

This looks like a bug to me, so I went ahead and opened up https://github.com/git-lfs/git-lfs/pull/2733 to fix it. If approved and merged, I've scheduled this to be released in 2.4.0, but here are some binaries that you can try in the meantime.

[git-lfs-windows-386-2.3.4.zip](https://github.com/git-lfs/git-lfs/files/1480092/git-lfs-windows-386-2.3.4.zip)
[git-lfs-windows-amd64-2.3.4.zip](https://github.com/git-lfs/git-lfs/files/1480093/git-lfs-windows-amd64-2.3.4.zip)

PR #2733, opened 2017-11-16
lfsapi/auth: teach DoWithAuth to respect http.extraHeaders

ttaylorr:
This pull request fixes a bug pointed out in https://github.com/git-lfs/git-lfs/issues/2729 where @TingluoHuang noticed that setting `http.$url.extraHeader` in your Git config would still cause API requests needing authorization to contact your credential helper, even if the authorization header was set.

Introduced here is a mechanism to avoid contacting the credential helper if any matching `http.extraHeaders` entry contains a header with the key `Authorization: `. This intentionally does not fall back to the credential cache or other credential helpers, since we assume that if a user explicitly provided an `Authorization` header, they would like to use that.

Closes: https://github.com/git-lfs/git-lfs/issues/2729

##

/cc @git-lfs/core https://github.com/git-lfs/git-lfs/issues/2729

### DIFF
--- a/lfsapi/auth.go
+++ b/lfsapi/auth.go
@@ -61,7 +61,7 @@ func (c *Client) doWithCreds(req *http.Request, credHelper CredentialHelper, cre
 	if access == NTLMAccess {
 		return c.doWithNTLM(req, credHelper, creds, credsURL)
 	}
-	return c.Do(req)
+	return c.do(req)
 }
 
 // getCreds fills the authorization header for the given request if possible,
--- a/lfsapi/auth.go
+++ b/lfsapi/auth.go
@@ -25,17 +25,6 @@ var (
 // supporting basic and ntlm authentication.
 func (c *Client) DoWithAuth(remote string, req *http.Request) (*http.Response, error) {
 	req.Header = c.extraHeadersFor(req)
-	if len(req.Header.Get("Authorization")) > 0 {
-		// If the extra headers that were configured contained
-		// credentials, assume it is preferred to use those over
-		// contacting the credential helper. Fallback to c.Do.
-		res, err := c.doWithRedirects(c.httpClient(req.Host), req, nil)
-		if err != nil {
-			return res, err
-		}
-
-		return res, c.handleResponse(res)
-	}
 
 	apiEndpoint, access, credHelper, credsURL, creds, err := c.getCreds(remote, req)
 	if err != nil {
--- a/lfsapi/auth.go
+++ b/lfsapi/auth.go
@@ -24,6 +24,19 @@ var (
 // authentication from netrc or git's credential helpers if necessary,
 // supporting basic and ntlm authentication.
 func (c *Client) DoWithAuth(remote string, req *http.Request) (*http.Response, error) {
+	req.Header = c.extraHeadersFor(req)
+	if len(req.Header.Get("Authorization")) > 0 {
+		// If the extra headers that were configured contained
+		// credentials, assume it is preferred to use those over
+		// contacting the credential helper. Fallback to c.Do.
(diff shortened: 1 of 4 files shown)


lfsapi/auth.go
```diff
@@ -24,6 +24,19 @@ var (
 // authentication from netrc or git's credential helpers if necessary,
 // supporting basic and ntlm authentication.
 func (c *Client) DoWithAuth(remote string, req *http.Request) (*http.Response, error) {
+	req.Header = c.extraHeadersFor(req)
```
technoweenie: The Client type is supposed to have `Do()` and `DoWithAuth()` as the two main ways to make a request. However, `DoWithAuth()` eventually calls `Do()` in `doWithCreds()`:

https://github.com/git-lfs/git-lfs/blob/5427600232727855328b7c40021b5bb20a5424b4/lfsapi/auth.go#L58-L63

One easy fix is to pull out a `do()` function from `Do()`:

https://github.com/git-lfs/git-lfs/blob/5427600232727855328b7c40021b5bb20a5424b4/lfsapi/client.go#L86-L94

Then, make sure any `lfsapi.Client` funcs that call `Do()`, call this new `do()` instead. Does that make sense?

There's likely a better extraction hiding in here too. I don't like having `Do()` and `DoWithAuth()`. But this would require a lot more thought and work, where a quick fix might be better for the extra headers fix.

ttaylorr: Great call! I think we might even not have to deal with this based on your comment in https://github.com/git-lfs/git-lfs/pull/2733#discussion_r151534714, since the point of this `if` was to avoid calling `getCreds()`, since my original assumption was that `getCreds()` _always_ contacted the credential helper.

Since it double checks for authorization, I think the fix here is only ensuring that we call `c.extraHeadersFor`.

technoweenie: > since my original assumption was that getCreds() always contacted the credential helper.

It used to, but I fixed it: https://github.com/git-lfs/git-lfs/pull/2695

> Since it double checks for authorization, I think the fix here is only ensuring that we call c.extraHeadersFor.

Yeah, I suppose `extraHeadersFor()` could be called in `DoWithAuth()`, and again when `Do()` is called by `doWithCreds()`. It is rebuilding the `http.Header`, but should _retain_ the existing header without causing problems. 

Though I think if `DoWithAuth()` somehow changes one of the headers that is one of the git-configured extra headers, and is not `Authorization`, then the changes would be clobbered when `extraHeadersFor()` is called again inside `Do()`.

ttaylorr: @technoweenie -- ah, I think I understand now, and thanks for (double) explaining this to me! I agree that some larger changes may be useful to think about here, but for now I opted for `do()`, which handles redirects and errors, but does _not_ add extra headers.

After changing the behavior of `lfsapi` to no longer itself call `Do()`, I believe that this is ready to go.

lfsapi/client.go
```diff
@@ -85,6 +85,13 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
 	req.Header = c.extraHeadersFor(req)
 	req.Header.Set("User-Agent", UserAgent)
 
+	return c.do(req)
+}
+
+// do performs an *http.Request respecting redirects, and handles the response
+// as defined in c.handleResponse. Notably, it does not alter the headers for
+// the request argument in any way.
+func (c *Client) do(req *http.Request) (*http.Response, error) {
```
technoweenie: I think the `User-Agent` bit should move to `do()` too. This ensures a consistent user agent for all outgoing http requests.

ttaylorr: Nice -- I picked this up in https://github.com/git-lfs/git-lfs/pull/2733/commits/bda2fe845fd7b1b4917ac266c17cea8ec055261a.

Issue #2654, opened 2017-10-07
Ambiguous refs shouldn't error out


phord:
I ran `git-lfs migrate info` on a repo with an ambiguous reference.  Instead of producing anything useful, it only reported "ref ... is ambiguous".

    git lfs migrate info --everything
    ref '66df2eb' is ambiguous...

The problem is that my repo has a ref named `refs/tags/66df2eb`
This seems like it shouldn't be a fatal error.

technoweenie:
Ah, I think this would be solved by including the ref prefixes, disambiguating the ref names.

PR #2734, opened 2017-11-16
commands/migrate: allow for ambiguous references in migrations

ttaylorr:
This pull request fixes a bug pointed out by @phord in https://github.com/git-lfs/git-lfs/issues/2654 where the `git lfs migrate` suite would refuse to disambiguate confusing reference names when asked to perform migrations.

@technoweenie points out a fix that I implemented in this pull request via https://github.com/git-lfs/git-lfs/issues/2654#issuecomment-335191900:

> Ah, I think this would be solved by including the ref prefixes, disambiguating the ref names.

To accomplish this, I taught the `*git.Ref` type to `String()` itself, returning a fully-qualified reference, and collect those instead of the local reference names. This allows for references to have existing SHA-1's as their local name, and migrate will still work correctly.

Closes: https://github.com/git-lfs/git-lfs/issues/2654.

##

/cc @git-lfs/core https://github.com/git-lfs/git-lfs/issues/2654

### DIFF
--- a/test/test-migrate-import.sh
+++ b/test/test-migrate-import.sh
@@ -388,6 +388,7 @@ begin_test "migrate import (ambiguous reference)"
 
   # Create an ambiguously named reference sharing the name as the SHA-1 of
   # "HEAD".
+  sha="$(git rev-parse HEAD)"
   git tag "$sha"
 
   git lfs migrate import --everything
--- a/test/test-migrate-import.sh
+++ b/test/test-migrate-import.sh
@@ -403,8 +403,6 @@ begin_test "migrate import (--everything with --include-ref)"
 )
 end_test
 
-exit 0
-
 begin_test "migrate import (--everything with --exclude-ref)"
 (
   set -e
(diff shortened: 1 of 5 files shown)


technoweenie:
Drive by review: FWIW #2722 adds a `Refspec()` func to `*git.Ref`. I prefer that over `String()`, unless you're trying to satisfy a Stringer interface.

ttaylorr:
> I prefer that over `String()`, unless you're trying to satisfy a Stringer interface.

I like the idea of satisfying the `fmt.Stringer` interface, but we're not using this in any formatting context. Want to kick it over to `Refspec` in this PR instead? I figure we'll deal with some merge conflicts anyway, so perhaps your branch can merge first and I'll update and pull in `Refspec` from there?

test/test-migrate-import.sh
```diff
@@ -403,8 +416,6 @@ begin_test "migrate import (--everything with --include-ref)"
 )
 end_test
 
-exit 0
```
technoweenie: 🥇

PR #2735, opened 2017-11-17
Fix integration test early exit

technoweenie:
This updates `script/integration` so it waits until it sends all output from the channel to STDOUT before exiting the command. It prevents cases where the _last_ test script fails, but no output is shown for that test script.

### DIFF
--- a/script/integration.go
+++ b/script/integration.go
@@ -50,17 +50,24 @@ func mainIntegration() {
 	for _, file := range files {
 		tests <- file
 	}
+	close(tests)
 
-	go printOutput(output)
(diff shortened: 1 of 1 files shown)


technoweenie:
Hey, I'll let you deal with merging the PRs in the right order. I don't want to update branch here if it'll interfere with you trying to get something merged. Other than the changes I have to make for https://github.com/git-lfs/git-lfs/pull/2722, I don't think there is any merge order dependency, so I don't anticipate a problem. Does that work?

ttaylorr:
@technoweenie thanks! All merged.

Issue #2656, opened 2017-10-08
migrate shouldn't fetch or reach out to the remote at all without permission


phord:
lfs migrate first tries to `git fetch origin`.  This means it doesn't work at all if I'm offline.

    $ GIT_TRACE=2 git -C pure_tools.git lfs migrate info
     'git' version
     'git' config -l
     'git' ... rev-parse HEAD --symbolic-full-name HEAD
     'git' ... rev-parse master --symbolic-full-name master
     'git' ... fetch origin
     migrate: Fetching remote refs: ..., done

I should be able to run `git lfs migrate info` locally without bothering my remote server.

technoweenie:
Without args, `info` and `import` are processing _unpushed_ commits. The `fetch` is used to ensure you don't rewrite anything already pushed, requiring a force push on your remote.

You can use `--everything` to scan all local refs, or `--include-rev` and `--exclude-ref` to control what refs are analyzed or rewritten.

That said, I'd accept a PR for a flag that disables this check.

phord:
How can I have something in my repo I already pushed if the remote ref is not also updated? Are you worried I pushed it from a different local copy of the repo, or maybe I have something in my repo I imported with git-apply?

Ok, I can see that happening.  But it seems like the unusual use case.

PR #2738, opened 2017-11-17
commands/migrate: add '--skip-fetch' for offline migrations

ttaylorr:
This pull request addresses an enhancement pointed out @phord in https://github.com/git-lfs/git-lfs/issues/2656 where the `git-lfs-migrate(1)` command's use of `git-fetch(1)` would make the command un-useable offline.

Since it's been a little while since I've thought about how this works, here's a refresher for my own edification. When determining what to migrate, the tool considers first explicitly given references, or, if none were provided, the unpushed changes on the current branch (i.e., `git log <branch>..<remote-tracking-branch>`). To ensure that we get an up-to-date result and don't accidentally require a force push when the user wasn't expecting one, we always `git fetch` to grab the latest remote refs, if any of them have moved.

Prior to this pull request, there was no way to explicitly disable that, meaning if you don't have an internet connection, you have to (at least) remove remotes, which means the migration won't update any remote refs (since they will be deleted when calling `git remote remove`).

This pull request addresses this problem by providing a `--assume-current-remote-refs` that explicitly tells Git LFS to _not_ contact the remote, and assume the copy of remote references it currently has is up-to-date.

I decided against the original issue title:

> [...] shouldn't fetch or reach out to the remote at all without permission

Since I think the default behavior is safe, if not at least good. I would prefer to explicitly opt-_out_ of the safe behavior rather than into it.

Closes: https://github.com/git-lfs/git-lfs/issues/2656.

##

/cc @git-lfs/core https://github.com/git-lfs/git-lfs/issues/2656

### DIFF
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -20,6 +20,11 @@ var (
 	// in the migration.
 	migrateExcludeRefs []string
 
+	// migrateAssumeCurrentRemoteReferences assumes that the client has the
+	// latest copy of remote references, and thus should not contact the
+	// remote for a set of updated references.
+	migrateAssumeCurrentRemoteReferences bool
+
 	// migrateEverything indicates the presence of the --everything flag,
 	// and instructs 'git lfs migrate' to migrate all local references.
 	migrateEverything bool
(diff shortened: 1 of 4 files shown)


ttaylorr:
> What do you think about `--skip-remote-fetch` or even just `--skip-fetch`?

Great call! I went with `--skip-fetch`, and think that with these changes, this PR is good to go. Thanks for the review! 🙇 .

Issue #2743, opened 2017-11-20
Absorb `progress` functionality into `git/githistory/log`


technoweenie:
I like LFS' current output, but think it could be improved. For example:

```sh
$ git lfs fetch --all
Scanning for all objects ever referenced...
Fetching objects...
✔ 8 objects found, done
# why does a Git LFS command remind us what it is?
# an "LFS" prefix makes sense for smudge/clean filters though.
Git LFS: (8 of 8 files) 900 KB / 900 KB, done
```

Compare that to the `migrate` output:

```sh
$ git lfs migrate info --everything
migrate: Sorting commits: ..., done
migrate: Examining commits: 100% (14/14), done
*.gitattributes	420 B	4/4 files(s)	100%
*.gif          	391 B	3/3 files(s)	100%
*.bin          	378 B	3/4 files(s)	 75%
*.png          	260 B	2/2 files(s)	100%
*.JPG          	130 B	1/1 files(s)	100%
```

Each task is a single line, with a running counter until that task is done. It just looks more consistent with typical Git output:

```sh
$ git pull
remote: Counting objects: 63, done.
remote: Compressing objects: 100% (29/29), done.
remote: Total 63 (delta 50), reused 40 (delta 34), pack-reused 0
Unpacking objects: 100% (63/63), done.
```

I like how https://github.com/git-lfs/git-lfs/pull/2732 merges the two packages a bit, but I think more can be done here too. Why do we need a `progress` package at this point? Can we update the callers to use a `log.Task` implementation instead?

ttaylorr:
> Each task is a single line, with a running counter until that task is done. It just looks more consistent with typical Git output:

Totally agree -- I think that making LFS output self-consistent is our most important goal, but if we can do so while _also_ making that output consistent with Git, all the better. This was an explicit goal of mine when designing the output of `git lfs migrate`.

> Why do we need a `progress` package at this point? Can we update the callers to use a `log.Task` implementation instead?

I think that this is what is currently happening (`*progress.Meter` and `*progress.Spinner` are implementations of the `log.Task` interface), but we could perhaps go a bit further. For instance, I'm not sure that the `progress` package needs to exist any more, since package `tq` could own a single implementation of that instead. I think that the spinner could live in the `git/githistory/log` package, since it is more generally applicable than the implementation of `*progress.Meter`.

In fewer words, here are some things that I'd like to do:

1. Move generally-applicable logging things into the `git/githistory/log` package.
[...]

PR #2747, opened 2017-11-22
*: introduce package 'tlog'

ttaylorr:
This pull request promotes the implementation of a task-based logging system found originally in `git/githistory/log` to a new, top-level package `tlog`.

This accomplishes one of the goals that I proposed in https://github.com/git-lfs/git-lfs/issues/2743#issuecomment-345873955:

> In fewer words, here are some things that I'd like to do:
> 
> 1. [...]
> 2. Move the git/githistory/log package to package log (no longer nested in git/githistory).

However, instead of calling it `log` (confusing with Go's standard library package `log`) or `gitlog` (confusing with `git-log(1)`-related utilities), chosen here is `tlog` for **task**-log. I think that this is a good name since it:

- Isn't easily confused with other functions.
- Makes it a part of the name that this package is "task"-based, and therefore, provides some synchronicity guarantees.

Since this package contains a mostly-stable API, I feel comfortable promoting this package to live in the `github.com/git-lfs/git-lfs`-namespace. Future goals in https://github.com/git-lfs/git-lfs/issues/2743 support adding new functionality to the `log` package, which is something that I'd like to do after moving the code, and not before.

Tracking: https://github.com/git-lfs/git-lfs/issues/2743

/cc @git-lfs/core https://github.com/git-lfs/git-lfs/issues/2743

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -6,9 +6,9 @@ import (
 
 	"github.com/git-lfs/git-lfs/filepathfilter"
 	"github.com/git-lfs/git-lfs/git"
-	"github.com/git-lfs/git-lfs/git/githistory/log"
 	"github.com/git-lfs/git-lfs/lfs"
 	"github.com/git-lfs/git-lfs/progress"
+	"github.com/git-lfs/git-lfs/tlog"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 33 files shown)


ttaylorr:
> What do you think of `tasklog` instead of `tlog`?

🤘 I like `tasklog` much better than `tlog` -- I renamed this in 45c580e.

PR #2749, opened 2017-11-22
progress: move CopyCallback (& related) to package 'tools'

ttaylorr:
This pull request moves CopyCallback (and its friends) out of the `progress` package and into the `tools` package. Since this is based on https://github.com/git-lfs/git-lfs/pull/2747, an inter-diff can be summarized as a4ff7b8.

EDIT: This is the next step in removing the `progress` package, as outlined in https://github.com/git-lfs/git-lfs/issues/2743. Prior to this pull request, `progress` consisted of:

1. The `Progress` interface
2. Implementations of the above as `*progress.Spinner` and `*progress.Meter`.
3. `CopyCallback` (and related utilities).

As it has grown, `CopyCallback`'s use has become less related to the rest of the `progress` package. Since it is used in many locations within Git LFS, this pull request proposes moving it to the `tools` package. I think that since its surface area is small, `CopyCallback`-esque functionality does not warrant an entire package of its own, and it would fit well in its current state within the `tools` package.

Following this pull request, the remainder of `progress` will be moved into the `tasklog` and `tq` packages, thus enabling us to remove the `progress` package entirely. 

##

/cc @git-lfs/core https://github.com/git-lfs/git-lfs/issues/2743

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -6,9 +6,9 @@ import (
 
 	"github.com/git-lfs/git-lfs/filepathfilter"
 	"github.com/git-lfs/git-lfs/git"
-	"github.com/git-lfs/git-lfs/git/githistory/log"
 	"github.com/git-lfs/git-lfs/lfs"
 	"github.com/git-lfs/git-lfs/progress"
+	"github.com/git-lfs/git-lfs/tlog"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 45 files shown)


ttaylorr:
@git-lfs/core I mistakingly merged this PR thinking that it was approved! Please don't hesitate to let me know if there's anything that looks less than 👍 in this PR, and I'd be more than happy to follow up on it.

PR #2757, opened 2017-11-28
commands/prune: unify formatting

ttaylorr:
This pull request removes the use of the `*progress.Spinner` type in `git-lfs-prune(1)` in favor of a more consistent logging scheme.

Before:

```
~/example (master) $ git lfs prune
Pruning 1 files, (18 B)tained
✔ 3 local objects, 2 retained, done
✔ Deleted 1 files, done
```

After:

```
~/example (master) $ git lfs prune
prune: 3 local object(s), 2 retained, done
prune: Deleting objects: 100% (1/1), done
```

##

This removes one of two existing uses of the `*progress.Spinner` implementation, and the next pull-request in this series will remove the other use-case, which will allow `*progress.Spinner` to be deleted.

Please let me know if you have any thoughts on the formatting of these messages. I tried to make them as similar as I would imagine Git writing them, with these thoughts in mind:

1. Lowercase prefix of the command that is being run.
2. ", done\n" on the end (courtesy of `*tasklog.Logger`).
3. Use percentages where appropriate, text where otherwise.

##

/cc @git-lfs/core


ttaylorr:
TODO: looks like there are some tests failing here that I have to look into

ttaylorr:
All fixed. Here's a quick breakdown of what was going wrong --

1. The final update sent from `*SimpleTask`s were sometimes being dropped since `close()`-ing the updates channel does not wait for acknowledgement from the `tasklog.Logger`. By adding a "secret" `OnComplete()` method that is called if the `Task` implements it, we can wait for the logger to close.

2. Some of the final updates were being dropped in a race-y fashion because we were failing to call `logger.Close()` (fixed in https://github.com/git-lfs/git-lfs/pull/2757/commits/4b33121c9648b73570acd3e2312c0939c43fea0b). Adding this call waits until not only all active tasks have closed, but also until the logger has logged all messages.

commands/command_prune.go
```diff
@@ -176,28 +178,36 @@ func prune(fetchPruneConfig lfs.FetchPruneConfig, verifyRemote, dryRun, verbose
...
-		logger.Enqueue(spinner)
+		// enqueue a _new_ percentage task to track deletion progress.
+		percentage := tasklog.NewPercentageTask("prune: Deleting objects", uint64(len(prunableObjects)))
+		logger.Enqueue(percentage)
 
-		pruneDeleteFiles(prunableObjects, spinner)
+		pruneDeleteFiles(prunableObjects, percentage)
 	}
 
+	logger.Close()
```
technoweenie: Should this be moved to a `defer` right after the logger is initialized? That ensures that it's closed if this func returns early, such as when `prunableObjects` is empty.

commands/command_prune.go
```diff
@@ -232,7 +242,7 @@ func pruneCheckErrors(taskErrors []error) {
 	}
 }
 
-func pruneTaskDisplayProgress(progressChan PruneProgressChan, waitg *sync.WaitGroup, spinner *progress.Spinner) {
+func pruneTaskDisplayProgress(progressChan PruneProgressChan, waitg *sync.WaitGroup, task *tasklog.SimpleTask) {
```
technoweenie: I'm not crazy about this function receiving a task, and then closing it. What do you think about sending it the `logger` instead, so that the task is created, enqueued, used, and completed all in this func?

commands/command_prune.go
```diff
@@ -176,28 +178,36 @@ func prune(fetchPruneConfig lfs.FetchPruneConfig, verifyRemote, dryRun, verbose
...
 		}
+		info.Complete()
 
 		// Since the above progress will have completed, create and
-		// enqueue a _new_ spinner to track deletion progress.
-		spinner := progress.NewSpinner()
-		logger.Enqueue(spinner)
+		// enqueue a _new_ percentage task to track deletion progress.
+		percentage := tasklog.NewPercentageTask("prune: Deleting objects", uint64(len(prunableObjects)))
+		logger.Enqueue(percentage)
```
technoweenie: Since this `percentage` is only used inside `pruneDeleteFiles()`, why not pass `logger` to it instead. Same justification as my note on `pruneTaskDisplayProgress()`.

tasklog/percentage_task_test.go
```diff
@@ -32,6 +32,7 @@ func TestPercentageTaskCalculatesPercentWithoutTotal(t *testing.T) {
 }
 
 func TestPercentageTaskCallsDoneWhenComplete(t *testing.T) {
+	t.Skip()
```
technoweenie: Is this skip temporary, or should the test be removed?

ttaylorr:
@technoweenie thanks for the feedback, I agree with everything you said 😄. I un-skipped the test in 088b2ab, and changed the above methods to take a `*tasklog.Logger`, not a `tasklog.Task` in cde392b.

Can you take another look? 🙇

commands/command_prune.go
```diff
@@ -288,7 +284,10 @@ func pruneTaskCollectErrors(outtaskErrors *[]error, errorChan chan error, errorw
 	}
 }
 
-func pruneDeleteFiles(prunableObjects []string, task *tasklog.PercentageTask) {
+func pruneDeleteFiles(prunableObjects []string, logger *tasklog.Logger) {
+	task := tasklog.NewPercentageTask("prune: Deleting objects", uint64(len(prunableObjects)))
+	logger.Enqueue(task)
```
technoweenie: Small idea for anther PR:

```go
task := logger.NewPercentageTask(...) // create and enqueue task
defer task.Complete()
```

ttaylorr: 🤦‍♂️  I'm silly -- this exists! Changed in 65b4f34e.

commands/command_prune.go
```diff
@@ -176,28 +177,29 @@ func prune(fetchPruneConfig lfs.FetchPruneConfig, verifyRemote, dryRun, verbose
...
 	}
+
+	info := tasklog.NewSimpleTask()
+	logger.Enqueue(info)
 	if dryRun {
-		Print("%d files would be pruned (%v)", len(prunableObjects), humanize.FormatBytes(uint64(totalSize)))
+		info.Logf("prune: %d file(s) would be pruned (%s)", len(prunableObjects), humanize.FormatBytes(uint64(totalSize)))
 		if verbose {
-			Print(verboseOutput.String())
+			for _, item := range verboseOutput {
```
technoweenie: I think you can take out the two extra `if verbose` (here and L194), since `verboseOutput` will be empty in that case.

commands/command_prune.go
```diff
@@ -176,28 +177,29 @@ func prune(fetchPruneConfig lfs.FetchPruneConfig, verifyRemote, dryRun, verbose
...
+		info.Complete()
 	} else {
-		Print("Pruning %d files, (%v)", len(prunableObjects), humanize.FormatBytes(uint64(totalSize)))
 		if verbose {
-			Print(verboseOutput.String())
+			for _, item := range verboseOutput {
+				info.Logf("\n%s", item)
+			}
 		}
+		info.Complete()
```
technoweenie: This could be moved to a `defer`, allowing the one on L192 to be removed too. I point it out because I think having a consistent pattern for dealing with loggers and tasks is important. I really like how the `tasklog` package is shaping up :)

commands/command_prune.go
ttaylorr: Good, call -- I changed this in 782c58c3.

commands/command_prune.go
ttaylorr: Hm, I don't think that these calls to `Complete()` can be deferred, since `pruneDeleteFiles` creates a new task, which (for the `*tasklog.Logger` to output), requires there to be no actively running task.

PR #2758, opened 2017-11-29
commands/fetch: unify formatting

ttaylorr:
This pull request unifies the formatting of `git-lfs-fetch(1)`'s output, similarly to https://github.com/git-lfs/git-lfs/pull/2757.

Before:

```
~/repository (master) $ git lfs fetch --all
Scanning for all objects ever referenced...
Fetching objects...
✔ 501 objects found, done
Git LFS: (291 of 291 files) 1.1 KB / 1.1 KB, done
```

After:

```
~/repository (master) $ git lfs fetch --all
fetch: Fetching all references...
fetch: 501 object(s) found, done
Git LFS: (501 of 501 files) 1.9 KB / 1.9 KB, done
```

## 

This removes the later of the two remaining uses of the `*progress.Spinner` type, which will allow it to be removed entirely in a subsequent pull request.

Similar to https://github.com/git-lfs/git-lfs/pull/2757:

> Please let me know if you have any thoughts on the formatting of these messages. I tried to make them as similar as I would imagine Git writing them, with these thoughts in mind:
> 
> 1. Lowercase prefix of the command that is being run.
> 2. ", done\n" on the end (courtesy of `*tasklog.Logger`).
> 3. ~~Use percentages where appropriate, text where otherwise.~~

##

/cc @git-lfs/core

PR #2759, opened 2017-11-29
progress: remove spinner.go

ttaylorr:
This pull request removes `spinner.go` from package `progress` since it is no longer used within Git LFS (see: https://github.com/git-lfs/git-lfs/pull/2757, https://github.com/git-lfs/git-lfs/pull/2758).

Looking at usage of the exported identifiers in `meter.go` is the only remaining thing to look at in the progress package before it can be removed entirely 🎉 .

##

/cc @git-lfs/core


ttaylorr:
> Double check my merge commit. 

Looks good to me, thanks! :eyes:.

PR #2762, opened 2017-11-29
progress: remove *progress.Meter

ttaylorr:
This pull request does the scaffolding necessary to, and removes the `*progress.Meter` implementation from package `progress`.

My initial approach with this branch was to copy verbatim the `*progress.ProgressMeter` type from `progress` to package `tq`. This worked fine, but I noticed that it expanded package `tq`'s API significantly, with the majority of API changes coming from `progress.DryRun`, `progress.WithOSEnv`, (etc) options. To mitigate this, my first inclination was to create a sub-package `tq/progress`, which would house these types/functions. Wanting not to create a new package, I removed the progress meter option functions in favor of an options struct, and moved that into the `tq` package directly.

1. f411d01: merge in the changes from `progress-remove-logger` to avoid a merge conflict later when merging this branch into master.
2. 145d0de, 622c7d3: teach `*ProgressMeter` to respond to the `nil`-receiver, to make it possible to remove `progress.Noop()`, and instead use `(nil)(*progress.Meter`).
3. a549105: remove `progress.Noop()`.
4. fa1d512: since `*ProgressMeter` is the only remaining implementation of `progress.Meter`, remove the `progress.Meter` interface.
5. e35d20f: rename `*progress.ProgressMeter` -> `*progress.Meter` to [avoid stuttering][1].
6. 6521683: to reduce the scope of the `*progress.Meter`-related API, remove option functions in favor of a struct-style instantiation, i,e.,

```go
m := progress.NewMeter(&progress.MeterOption{
        DryRun: false,
        // ...
})
```

7. 54fb383: finally, move `*progress.Meter` to `*tq.Meter`, thus enabling the `progress` package to be removed.

##

/cc @git-lfs/core  

[1]: https://blog.golang.org/package-names#TOC_3.


tq/meter.go
```diff
@@ -86,44 +80,108 @@ func WithOSEnv(os env) meterOption {
...
+	if err != nil {
+		printErr(err.Error())
+		return nil, false
+	}
+
+	return file, true
+}
+
+// NewMeter creates a new Meter.
+func NewMeter(opt *MeterOption) *Meter {
```
technoweenie: Do we really need the `MeterOption` type? The only two things to change are whether dry run is enabled, and what logger is used.

```go
m := NewMeter()
m.DryRun = true
m.Logger = tq.LoggerToFile(filename)
m.Logger = tq.LoggerFromEnv(osEnv)
m.Start()
```

My hunch is that we didn't use fields like this to prevent race conditions. Should we continue trying to protect the caller, or can we simplify things and just document that those fields should be set before `Start()` is called? FWIW I do prefer the `MeterOptions` type to the option func args :+1:

tq/meter.go
```diff
@@ -86,44 +80,108 @@ func WithOSEnv(os env) meterOption {
...
-		opt(m)
-	}
+	opt.configure(m)
 
 	return m
 }
 
 // Start begins sending status updates to the optional log file, and stdout.
-func (p *ProgressMeter) Start() {
+func (p *Meter) Start() {
```
technoweenie: Super nitpick: Can you change this to `m *Meter`?

technoweenie:
I made the suggested fixes so I could see if it worked in #2774. If you really want to keep `MeterOptions`, we can force push it away.

ttaylorr:
Those changes look great, thanks!

Issue #2763, opened 2017-11-30
adding the file size to git lfs ls-files output


kforner:
It would be nice to also have the (actual) file size in the output of `git lfs ls-files`.
We would not need to inspect a pointer file anymore. 
It could be activated via an option if you don't want to modify the output (for downstream parsers)...

ttaylorr:
@kforner Great idea -- I opened https://github.com/git-lfs/git-lfs/pull/2764 to add just this.

kforner:
great !

PR #2764, opened 2017-12-01
commands/ls-files: add '--size' flag

ttaylorr:
This pull request resolves https://github.com/git-lfs/git-lfs/issues/2763 by adding a `--size` flag to the `git-lfs-ls-files(1)` command, which works as follows:

```
~/repository (master) $ git lfs ls-files --size
d1b2a59fbe * a.dat (8)
```

I'm not sure why we didn't have this feature previously, but to the best of my recollection, we were thinking of adding more complex options to `--size` like `--size=bytes`, `--size=kilobytes`, etc. My thought is that `--size` alone is sufficient (since downstream parsers can understand it easily), but perhaps we'd like to add more than this for command line-viewers to see something more reasonable.

I think that now is a good time to have that discussion, and I'll be more than happy to implement whatever we go with.

Closes: https://github.com/git-lfs/git-lfs/issues/2763.

##

/cc @git-lfs/core https://github.com/git-lfs/git-lfs/issues/2763


docs/man/git-lfs-ls-files.1.ronn
```diff
@@ -16,6 +16,10 @@ An asterisk (*) after the OID indicates a LFS pointer, a minus (-) a full object
 * `-l` `--long`:
   Show the entire 64 character OID, instead of just first 10.
 
+* `-s` `--size`:
+  Show the size in bytes of the LFS object between parenthesis at the end of a
+  line.
```
larsxschneider: I very much like this option 👍 

However, I feel this is an option for humans and therefore the byte size might be bit odd. I think MB as a default would be much more readable. I do agree that printing bytes is "technically more correct", though. However, robots can use the `--debug` option to get the exact byte size, no?

ttaylorr: Great call -- I pushed up something I think is somewhere down the middle in 11f6834. It uses the `humanize` package to find the biggest unit of storage that preserves accuracy for a given size. `git-lfs-ls-files(1)` uses that to print the biggest unit of storage it can, and if users/parser want something more specific they can use `--debug`.

larsxschneider: That might print different units in a single call, right?
e.g.:
```
d1b2a59fbe * a.dat (8 B)
a1b2a59fbe * big.dat (7 MB)
c1b2a59fbe * a.dat (1 GB)
```
I wonder if this makes it harder to less hard to read for humans.
Sorry for the 🚲 -shedding 😄

ttaylorr: It could, but I'm not sure what a better alternative would be. It might be inaccurate to assume for all objects in all repositories that "MB" is the most accuracy-preserving unit. What do you think?

larsxschneider: After thinking about it I believe `humanize` is the best approach here 👍

larsxschneider: Nit in the docs:
```
s/in bytes//
```

Issue #2688, opened 2017-10-25
Migrate for bare repositories


sgoo:
I'm a bit confused what the desired use case for LFS migrate is.
In my experience of helping people migrate to LFS they are wanting to reduce the overall size of there repository and re-write all branch branches/tags/refs into LFS.

As the LFS migrate tool stands right now it seems like its designed to help you migrate your default branch into LFS, and with the --everything flag other branches you have checked out locally.
But often I find my users are working with teams who have lots of branches and most of them are going to be remote branches worked on by other team member.
To fully embrace LFS they need to re-write all refs all at once.

Right now when helping people migrate to LFS at Bitbucket we recommend them to create a bare clone of the repo. This ensures they have every ref locally and its easy to push all the refs back to the remote at the end. Then our current tool of choice fro migrations is the BFG repo cleaner.

BFG is way over complicated though and it would be a lot better if we could tell them to use the built in migration functions of the core lfs client.

It would make sense to me if the migrate function built into LFS supported bare repos so that people can easily migrate an entire repository include all branches all at once.

technoweenie:
The `migrate` command should work on bare repositories:

```sh
$ git clone --bare https://github.com/git-lfs/git-lfs git-lfs-bare
$ cd git-lfs-bare
$ ls -al
total 40
drwxr-xr-x   11 rick  staff   374 Oct 30 10:45 .
drwxr-xr-x   58 rick  staff  1972 Oct 30 10:44 ..
-rw-r--r--    1 rick  staff    23 Oct 30 10:44 HEAD
-rw-r--r--    1 rick  staff   171 Oct 30 10:44 config
-rw-r--r--    1 rick  staff    73 Oct 30 10:44 description
drwxr-xr-x   12 rick  staff   408 Oct 30 10:44 hooks
drwxr-xr-x    3 rick  staff   102 Oct 30 10:44 info
drwxr-xr-x    4 rick  staff   136 Oct 30 10:45 lfs
drwxr-xr-x  260 rick  staff  8840 Oct 30 10:44 objects
-rw-r--r--    1 rick  staff  6855 Oct 30 10:44 packed-refs
drwxr-xr-x    4 rick  staff   136 Oct 30 10:44 refs

$ git lfs migrate info --everything
migrate: Sorting commits: ..., done
migrate: Examining commits: 100% (6358/6358), done
*.go  	48 MB 	7996/7996 files(s)	100%
[...]

PR #2769, opened 2017-12-05
commands/migrate: don't migrate remote references in bare repositories

ttaylorr:
This pull request fixes a bug pointed out by @sgoo in https://github.com/git-lfs/git-lfs/issues/2688, where `git lfs migrate info --everything` would fail when run under a bare repository.

The issue occurred because `git-lfs-migrate(1)` assumed that remote references do always exist, which is not the case when operating in a bare repository. (Interestingly, this is not the first time that I have forgotten about bare repositories: c.f., https://github.com/git-lfs/git-lfs/pull/2389, https://github.com/git-lfs/git-lfs/pull/1310).

Normally, this assumption is not error-prone, when the user provides an explicit list of reference to include and not include in a migration. However, when this list is not provided, _and_ the migration is run in a bare repository, the migrator would previously (and incorrectly) include remote references that do not exist.

From the Git documentation<sup>[[1](https://git-scm.com/docs/git-clone#git-clone---bare)]</sup>: 

> `--bare` [...] Also the branch heads at the remote are copied directly to corresponding local branch heads, without mapping them to refs/remotes/origin/.

To fix this, we skip the exclusion of remote references when the repository is bare.

Closes https://github.com/git-lfs/git-lfs/issues/2688.

##

/cc @git-lfs/core @sgoo

### DIFF
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -151,17 +151,24 @@ func includeExcludeRefs(l *tasklog.Logger, args []string) (include, exclude []st
 			include = append(include, ref.Refspec())
 		}
 	} else {
-		// Otherwise, if neither --include-ref=<ref> or
-		// --exclude-ref=<ref> were given, include no additional
-		// references, and exclude all remote references that are remote
-		// branches or remote tags.
-		remoteRefs, err := getRemoteRefs(l)
+		bare, err := git.IsBare()
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, errors.Wrap(err, "fatal: unable to determine bareness")
 		}
 
-		for _, rr := range remoteRefs {
-			exclude = append(exclude, rr.Refspec())
+		if !bare {
+			// Otherwise, if neither --include-ref=<ref> or
+			// --exclude-ref=<ref> were given, include no additional
+			// references, and exclude all remote references that
+			// are remote branches or remote tags.
+			remoteRefs, err := getRemoteRefs(l)
+			if err != nil {
+				return nil, nil, err
+			}
+
+			for _, rr := range remoteRefs {
+				exclude = append(exclude, rr.Refspec())
+			}
 		}
 	}
 
--- a/docs/man/git-lfs-migrate.1.ronn
+++ b/docs/man/git-lfs-migrate.1.ronn
@@ -130,7 +130,7 @@ migrated.
 
 The migrate command's most common use case is to convert large git objects to
 LFS before pushing your commits. By default, it only scans commits that don't
-exist on any remote.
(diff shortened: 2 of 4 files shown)


sgoo:
Awesome thanks for fix this, I checked it out locally and it seems to work well.

commands/command_migrate.go
```diff
@@ -151,17 +151,24 @@ func includeExcludeRefs(l *tasklog.Logger, args []string) (include, exclude []st
 			include = append(include, ref.Refspec())
 		}
 	} else {
-		// Otherwise, if neither --include-ref=<ref> or
-		// --exclude-ref=<ref> were given, include no additional
-		// references, and exclude all remote references that are remote
-		// branches or remote tags.
-		remoteRefs, err := getRemoteRefs(l)
+		bare, err := git.IsBare()
```
technoweenie: Would it make sense to do this inside `getRemoteRefs()`? If it returns true, then return a nil `remoteRefs` slice. Not a big deal, but I think it could improve the flow in this code a tad.

commands/command_migrate.go
```diff
@@ -151,17 +151,24 @@ func includeExcludeRefs(l *tasklog.Logger, args []string) (include, exclude []st
...
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, errors.Wrap(err, "fatal: unable to determine bareness")
```
technoweenie: I think this error wrapping is unnecessary, as any error will be incredibly obvious. The error from `getRemoteRefs()` isn't wrapped.

ttaylorr: Hm, I think that keeping the error wrapping might make sense, since `git rev-parse --is-bare-repository` can fail in weird ways, like:

```
~/not-a-real-dir $ git rev-parse --is-bare-repository
fatal: Unable to read current working directory: No such file or directory
```

commands/command_migrate.go
ttaylorr: Good call, I could go either way on this. I like that `getRemoteRefs()` _always_ returns remote refs, which to me feels "true" even in the bare case, since the remote refs _exist_, they just don't carry any meaning. I like pushing the decision of "do we care about remote refs" out of `getRemoteRefs()` itself.

commands/command_migrate.go
technoweenie: How do you get that though? If the `git` binary is missing?

ttaylorr: This one's tricky -- I am able to reproduce it when the current directory doesn't exist, i.e., if you change branches from B -> A where B has directory D but A does not.

PR #2773, opened 2017-12-07
send remote refspec for the other lock commands

technoweenie:
Implements #2712 for `git lfs [ lock | unlock | locks ]`.


technoweenie:
API docs are on the checklist: https://github.com/git-lfs/git-lfs/issues/2712

PR #2774, opened 2017-12-07
tq: send refspec with batch requests

technoweenie:
Implements #2712 for Batch API calls.

There is a problem here though. Currently, Transfer Queues and Git Scanners are scoped for a single remote. The updated `newUploadQueue()` builds a new Transfer Queue with the current remote ref.

That's fine, except the `pre-push` hook is called with the correct remote ref, and it supports multiple refs.

```
echo "refs/heads/master master refs/heads/tracked 0000000000000000000000000000000000000000" \
 | git lfs pre-push origin master
```

This should be calling `objects/batch` at least once for each ref. I made some changes to `commands/uploader.go` to initialize a new Transfer Queue and Git Scanner, but ran into a deadlock issue deep in the Git Scanner.

### DIFF
--- a/commands/command_push.go
+++ b/commands/command_push.go
@@ -96,7 +96,8 @@ func uploadsWithObjectIDs(ctx *uploadContext, oids []string) {
 	}
 
 	uploadPointers(ctx, pointers...)
-	ctx.Await()
+	ctx.CollectErrors(ctx.tq)
+	ctx.ReportErrors()
 }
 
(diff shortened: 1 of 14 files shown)


ttaylorr:
❄️ per https://github.com/git-lfs/git-lfs/issues/2712#issuecomment-352528973.

technoweenie:
Closing this due to conflicts. See #2806.

Issue #2575, opened 2017-09-12
Get the whole LFS object list from a repo


emericv:
There is a way to have the whole LFS object sha256 list from a git repo? (like a git-lfs ls-files, but including modified/deleted objects refs of all branchs/tags).

I want to build a LFS object garbage collector system to clean orphan object of my LFS server

ttaylorr:
@emericv thanks for opening this. `git lfs ls-files` takes a `<ref>` argument, so you can do something like:

```sh
git for-each-ref --format='%(refname)' | while read ref; do
  git lfs ls-files "$ref";
done \
  | sort -u \
  | uniq 
```

Alternatively, you can also run `git lfs prune` on your server, but this may not be exactly what you want.

technoweenie:
`git lfs ls-files` always scans the tree, so that snippet won't really dig into the history.

https://github.com/git-lfs/git-lfs/blob/eaf1395a96b7e7a125d36722a3e90d8b7bce7ad6/commands/command_ls_files.go#L63-L65

ttaylorr:
> `git lfs ls-files` always scans the tree, so that snippet won't really dig into the history.

We could teach `git lfs ls-files` to use `ScanRef` given a flag, which would make that snippet work. @technoweenie does this sound like a good idea?

technoweenie:
Possibly. Here's how `ls-files` looks right now, with `ScanTree(ref)`:

```sh
$ GIT_TRACE=1 ~p/git-lfs/bin/git-lfs ls-files master
trace git-lfs: run_command: 'git' version
trace git-lfs: run_command: 'git' config -l
trace git-lfs: run_command: git cat-file --batch
0263829989 - bin/b.bin
98ea6e4f21 - bin/hi.bin
b9f86fab47 - gif/atom-undo.gif
d1c8fab514 - gif/droidtocat.gif
d1c8fab514 - gif/dupe.gif
e1732d149d - jpg/kabuki-tattoo-left.JPG
f026c32cd7 * png/render.png
```

Here it is with `ScanRefWithDeleted()`:

```diff
diff --git a/commands/command_ls_files.go b/commands/command_ls_files.go
index 11c412f1..630286d2 100644
--- a/commands/command_ls_files.go
+++ b/commands/command_ls_files.go
@@ -60,7 +60,7 @@ func lsFilesCommand(cmd *cobra.Command, args []string) {
        })
        defer gitscanner.Close()

[...]

PR #2796, opened 2017-12-19
commands: teach '--all' to `git lfs ls-files`

ttaylorr:
This pull request adds a new flag to the `git lfs ls-files` command, `--all` to scan for _all_ LFS objects in a repository, not just those at the given reference (or HEAD, by default).

This feature was requested by @emericv in https://github.com/git-lfs/git-lfs/issues/2575:

> There is a way to have the whole LFS object sha256 list from a git repo? (like a git-lfs ls-files, but including modified/deleted objects refs of all branchs/tags).

@technoweenie suggested in https://github.com/git-lfs/git-lfs/issues/2575#issuecomment-328909675 to use `ScanRefWithDeleted`, but I think that `--all` may be sufficient (ignoring reference arguments). That said, it might be worth expanding this to cover:

1. `--all`: the entire repository, ignoring the explicit reference argument.
2. `--deleted`: the entire reference space, showing deleted objects (via `ScanTreeWithDeleted()`). 

EDIT: I implemented the above in c4d5f495.

Closes: https://github.com/git-lfs/git-lfs/issues/2575.

##

/cc @git-lfs/core https://github.com/git-lfs/git-lfs/issues/2575

### DIFF
--- a/commands/command_ls_files.go
+++ b/commands/command_ls_files.go
@@ -85,8 +85,14 @@ func lsFilesCommand(cmd *cobra.Command, args []string) {
 	if err := gitscanner.ScanIndex(ref, nil); err != nil {
 		Exit("Could not scan for Git LFS index: %s", err)
 	}
-	if err := gitscanner.ScanTree(ref); err != nil {
-		Exit("Could not scan for Git LFS tree: %s", err)
+	if lsFilesScanAll {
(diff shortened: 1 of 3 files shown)


ttaylorr:
Thanks for the review! 🙇

Issue #2798, opened 2017-12-20
Consecutive asterisks pattern not supported in git lfs migrate?


larsxschneider:
Hi,

the following, AFAIK valid, `.gitattribute` pattern does not work with `git lfs migrate`:
```
git lfs migrate import --verbose --everything -I '**/*.foo/bar/data/**'
```

[See Git docs for `**` syntax](https://github.com/git/git/blob/936d1b989416a95f593bf81ccae8ac62cd83f279/Documentation/gitignore.txt#L116-L127). I tested with Git LFS v2.3.4 and the [v2.4 pre-release](https://github.com/git-lfs/git-lfs/issues/2679#issuecomment-350896658). Is this considered a bug? If yes, then I am happy to provide more details and a test case.

Thanks,
Lars

BTW: `git lfs migrate` is awesome - insanely fast. 👍👍👍

larsxschneider:
The following workaround seems to work for the most part:
```
git lfs migrate import --verbose --everything -I 'some/path.foo/bar/data/**'
```

However, `git lfs migrate` did not migrate the following files:
```
some/path.foo/bar/data/other/dir/bla.bar/more/dirs/file
```
... my guess is that the `bla.bar` might be the cause.

larsxschneider:
```
git lfs migrate import --verbose --everything -I 'some/path.foo/bar/data/other/dir/bla.bar/more/dirs/*'
```
... doesn't migrate the file either 😢

technoweenie:
Yup, looks like a bug. Thanks for the report.

PR #2799, opened 2017-12-20
[WIP] git lfs migrate issues

larsxschneider:
Add test cases to demonstrate `git lfs migrate` problems mentioned in #2798.


test/test-migrate-import.sh
```diff
@@ -550,3 +550,43 @@ begin_test "migrate import (--everything with --exclude-ref)"
...
+begin_test "migrate import (nested sub-trees and --include with wildcard)"
+(
+  set -e
+
+  setup_single_local_branch_deep_trees
+
+  oid="$(calc_oid "$(git cat-file -p :foo/bar/baz/a.txt)")"
+  size="$(git cat-file -p :foo/bar/baz/a.txt | wc -c | awk '{ print $1 }')"
+
+  git lfs migrate import --include="**/*ar/**"
```
larsxschneider: The `*` in front of `ar` is the real problem here.

test/test-migrate-import.sh
```diff
@@ -550,3 +550,68 @@ begin_test "migrate import (--everything with --exclude-ref)"
...
+  size="$(git cat-file -p :foo/bar/baz/a.txt | wc -c | awk '{ print $1 }')"
+
+  # only import objects under "foo"
+  git lfs migrate import --include="foo/**"
+
+  assert_pointer "refs/heads/master" "foo/bar/baz/a.txt" "$oid_tree" "$size"
+  assert_local_object "$oid_tree" "$size"
+
+  # "a.txt" is not under "foo" and therefore should not be in LFS
+  refute_local_object "$oid_root"
```
larsxschneider: I admit that this is a pretty weird case. However, I just ran into it while migrating a real world repo 😄 

In this migration I move all files of the `/data` directory to Git LFS. Most of the files are actually big. However, one file exists in the exact same version in the `/src` directory. That should not be migrated, but `git lfs migrate` does migrate it. My `.gitattributes` pattern is `/data/**` and that means the migrated file in `/src` is not smudged I am left with the pointer file.

ttaylorr:
@larsxschneider 👋 here are a few updates:

1. I pushed aa1100a to stop using `refute_local_object` in tests. This is because there does rightly exist a migrated object with that OID (`foo/bar/baz/a.txt` is migrated and shares the same contents, thus an object under that OID does exist in the local object db), but `refute_local_object` does _not_ check the filepath belonging to that OID. Instead, I replaced that check to ensure that the Git OID are the same pre- and post-migration.

2. I merged `wildmatch` into this branch, causing everything to pass. Hooray!

PR #2806, opened 2018-01-04
Run 1 TransferQueue per uploaded ref

technoweenie:
This started as a rewrite of #2774 against new master (due to lots of conflicts with the progress meter in #2762). However, I decided to keep it scoped to just running multiple transfer queues, leaving the actual refspec-sending stuff to a follow up PR. I think that'll make this easier to review.

* While `commands.uploadContext` is an internal type, I'm trying to use exported functions in `commands`, outside of the actual `uploadContext` type. Also, some functions were moved to the `uploadContext` type, instead of merely accepting one as an argument. There's certainly more to be done with `*uploadContext`, but I'd rather do that in a specific refactor PR.
* `*tq.TransferQueue` now flushes the latest update, but doesn't close the Meter's internal channel. This is so there's only 1 line in the output, instead of 1 per ref. I'm open to having a single meter line per ref, but I think that should come with the upload output changes instead.
* `(*uploadContext) Await()` has been broken up into `CollectErrors(*tq.TransferQueue)` and `ReportErrors()`. This is so `CollectErrors()` can be called after each uploaded ref.

### DIFF
--- a/tq/meter.go
+++ b/tq/meter.go
@@ -169,7 +169,11 @@ func (m *Meter) Flush() {
 		return
 	}
 
-	m.update()
+	m.updates <- &tasklog.Update{
+		S:     m.str(),
+		At:    time.Now(),
+		Force: true,
+	}
 }
 
 // Finish shuts down the Meter.
--- a/tq/meter.go
+++ b/tq/meter.go
@@ -105,7 +105,7 @@ func (m *Meter) Add(size int64) {
 		return
 	}
 
-	defer m.update()
+	defer m.update(false)
 	atomic.AddInt32(&m.estimatedFiles, 1)
 	atomic.AddInt64(&m.estimatedBytes, size)
 }
@@ -117,7 +117,7 @@ func (m *Meter) Skip(size int64) {
 		return
 	}
 
-	defer m.update()
+	defer m.update(false)
 	atomic.AddInt64(&m.skippedFiles, 1)
 	atomic.AddInt64(&m.skippedBytes, size)
 	// Reduce bytes and files so progress easier to parse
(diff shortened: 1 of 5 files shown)


tq/meter.go
```diff
@@ -163,6 +163,19 @@ func (m *Meter) FinishTransfer(name string) {
 	m.fileIndexMutex.Unlock()
 }
 
+// Flush sends the latest progress update, while leaving the meter active.
+func (m *Meter) Flush() {
+	if m == nil || m.skipUpdate() {
+		return
+	}
+
+	m.updates <- &tasklog.Update{
```
ttaylorr: I think that one thing we could do to limit the number of places we do a channel write explicitly is to teach `update()` to take a `force bool` as a formal argument, and then pass that inside the body of the struct, like:

```go
func (m *Meter) update(force bool) { 
        m.updates <- &tasklog.Update{Force: true, ...}
}
```

I don't mind the duplication, but I think that channel writes in general can be fairly nuanced, so the fewer the better.

tq/transfer_queue.go
```diff
@@ -768,7 +768,7 @@ func (q *TransferQueue) Wait() {
 		close(watcher)
 	}
 
-	q.meter.Finish()
+	q.meter.Flush()
```
ttaylorr: I don't think that `Flush()` will call `close(m.updates)` which means that the `*tasklog.Logger` won't yield to future `Task` implementations.

ttaylorr: EDIT: Re-reading your PR body makes this make sense to me. Is the update channel closed elsewhere, though?

technoweenie: Yup, in `ReportErrors()`: https://github.com/git-lfs/git-lfs/pull/2806/files#diff-3cff4ed9c5b60b5b81240c88cd74e404R263

Issue #2802, opened 2017-12-23
Redesign CLI progress indicator: Upload size > Total size


bitinn:
Even for a successful `git push`, we often see this kind of reports:

```
$ git push
Git LFS: (22 of 22 files, 7 skipped) 227.26 MB / 203.26 MB, 750.98 KB skipped  
Counting objects: 63, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (63/63), done.
Writing objects: 100% (63/63), 2.01 MiB | 280.00 KiB/s, done.
Total 63 (delta 27), reused 0 (delta 0)
remote: Resolving deltas: 100% (27/27), completed with 17 local objects.
```

This line:

> Git LFS: (22 of 22 files, 7 skipped) 227.26 MB / 203.26 MB, 750.98 KB skipped

Unless you know exactly what's going on with Git LFS, you see no explanation for why upload size should exceed total size. After all, didn't we skip some files?

I am sure, there are *many* logically explanations to why upload size can be larger than total size. But I argue none of them matter to end users:

- Git LFS use *(upload size / total size)* as a progress indicator during `git push`.
- Yet upload size often exceed total file size, for an arbitrary amount.
- This is like having a progress bar going beyond its total bar length.
- The end result is users no longer trust this indicator.

----

The situation is worsen, if you are on a slow network, and run into connection problems:

```
$ git push
Git LFS: (7 of 29 files) 148.80 MB / 203.99 MB                                 Connection to github.com closed by remote host.
Git LFS: (7 of 29 files) 265.05 MB / 203.99 MB      
```

```
$ git push
Git LFS: (0 of 22 files, 7 skipped) 147.62 MB / 203.26 MB, 750.98 KB skipped   Connection to github.com closed by remote host.
Git LFS: (0 of 22 files, 7 skipped) 261.66 MB / 203.26 MB, 750.98 KB skipped
...
```

In this case users have no idea: (a) when the upload will be done, and (b) whether the push will be successful.

All they can do are wait.

And boy have I seen some disappointing failed git push. I seldom see subsequent Git LFS *skipping files* actually reduce the upload size.

----

I argue Git LFS could do better by:

A) Just don't use transfer size as a progress indicator.
B) Clarify how "skipped" files actually work.
C) Compare Git LFS to any other file syncing service (like OneDrive), and figure out what users actually want to see?

I wasn't able to convince non-technical coworkers to use Git LFS; now I am half-convinced I can't actually use it myself.

[...]

PR #2807, opened 2018-01-05
fix race conditions in upload progress meter

technoweenie:
This fixes 1 code duplication issue and 2 race conditions (plus 1 bonus fix!) causing the upload progress meter to report a higher current upload size than the estimated total size (reported in #2802).

> Git LFS: (22 of 22 files, 7 skipped) 227.26 MB / 203.26 MB, 750.98 KB skipped 

1. The `basic` and `tus` upload adapters have a lot of extremely similar logic for making and handling an HTTP request. This lead to some discrepancies with how `basic` and `tus` uploads handle failures. Introduced an internal `fileHTTPUpload()` helper so they both work consistently.
2. The LFS API client now rewinds the request body _right before_ making an initial or retried request. It used to rewind the request body after an HTTP request returned an error, meaning `ResetProgress()` was essentially doing nothing. This meant any retry during uploads would count the same transferred bytes again.
3. Found another rare race condition where the `tools.BodyWithCallback` would report transfer progress after `ResetProgress()`. So sometimes the output would report an extra 4k (a single io Read() chunk) of uploaded data. I fixed it by teaching `tq.Meter` to trust that the incoming `read` value from `tools.BodyWithCallback` is correct instead of keeping its own count. This means I had to track the value in a `map[string]int64`. Each `tools.BodyWithCallback` corresponds to an entry in the map according to the object OID. It works, but there's probably a better way to handle this.
4. BONUS: LFS now treats HTTP 503 statuses as a retriable request.

I tested this with the server at https://github.com/git-lfs/bench, on the `next` branch, with this [diff](https://gist.github.com/technoweenie/14b99ff2be13ad9a4a901b94bd683c49). Instead of randomly returning 503 errors, it now randomly closes the HTTP connection after reading a random number of bytes. That helped replicate the bug.

### DIFF
--- a/lfsapi/client.go
+++ b/lfsapi/client.go
@@ -159,13 +159,13 @@ func (c *Client) doWithRedirects(cli *http.Client, req *http.Request, via []*htt
 
 	requests := tools.MaxInt(0, retries) + 1
 	for i := 0; i < requests; i++ {
+		if seek, ok := req.Body.(io.Seeker); ok {
+			seek.Seek(0, io.SeekStart)
+		}
+
 		res, err = cli.Do(req)
 		if err == nil {
 			break
-		}
-
-		if seek, ok := req.Body.(io.Seeker); ok {
-			seek.Seek(0, io.SeekStart)
 		}
 
 		c.traceResponse(req, tracedReq, nil)
--- a/lfsapi/errors.go
+++ b/lfsapi/errors.go
@@ -58,6 +58,10 @@ func (c *Client) handleResponse(res *http.Response) error {
 		return errors.NewAuthError(err)
 	}
 
+	if res.StatusCode == 503 {
+		return errors.NewRetriableError(err)
+	}
+
 	if res.StatusCode > 499 && res.StatusCode != 501 && res.StatusCode != 507 && res.StatusCode != 509 {
 		return errors.NewFatalError(err)
 	}
@@ -95,6 +99,7 @@ var (
 		429: "Rate limit exceeded: %s",
 		500: "Server error: %s",
 		501: "Not Implemented: %s",
+		503: "Service Unavailable: %s",
 		507: "Insufficient server storage: %s",
 		509: "Bandwidth limit exceeded: %s",
 	}
--- a/tq/basic_upload.go
+++ b/tq/basic_upload.go
@@ -1,8 +1,10 @@
 package tq
 
 import (
+	"fmt"
 	"io"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -66,12 +68,21 @@ func (a *basicUploadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb Progres
 		req.Header.Set("Content-Length", strconv.FormatInt(t.Size, 10))
 	}
 
-	req.ContentLength = t.Size
+	if err := a.fileHTTPUpload(req, t, 0, cb, authOkFunc); err != nil {
+		return err
+	}
+
+	return verifyUpload(a.apiClient, a.remote, t)
+}
+
+func (a *adapterBase) fileHTTPUpload(req *http.Request, t *Transfer, offset int64, cb ProgressCallback, authOkFunc func()) error {
+	req.ContentLength = t.Size - offset
 
 	f, err := os.OpenFile(t.Path, os.O_RDONLY, 0644)
 	if err != nil {
-		return errors.Wrap(err, "basic upload")
+		return errors.Wrap(err, "lfs upload")
 	}
+
 	defer f.Close()
 
 	// Ensure progress callbacks made while uploading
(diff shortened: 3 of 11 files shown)


technoweenie:
Interesting failure. The verify retries do still work. However, since the response is a 503, it triggers a Batch API retry:

```
    10:10:48.592491 trace git-lfs: HTTP: POST http://127.0.0.1:61772/verify
    10:10:48.592714 trace git-lfs: HTTP: 503
    10:10:48.592753 trace git-lfs: HTTP: {"message":"intentionally failing verify request 3 (out of 10)"}
    10:10:48.592785 trace git-lfs: tq: verify err: LFS: intentionally failing verify request 3 (out of 10)
    10:10:48.592825 trace git-lfs: tq: retrying object 2dc9e3a3629196a5185e80d0522f9020ae9db29e21fc04b7d8016b5d44c5a107: LFS: intentionally failing verify request 3 (out of 10)
    10:10:48.592840 trace git-lfs: tq: enqueue retry #1 for "2dc9e3a3629196a5185e80d0522f9020ae9db29e21fc04b7d8016b5d44c5a107" (size: 18)
    10:10:48.592863 trace git-lfs: tq: sending batch of size 1
    10:10:48.592930 trace git-lfs: api: batch 1 files
    10:10:48.593045 trace git-lfs: creds: git credential cache ("http", "127.0.0.1:61772", "verify-fail-10-times")
    10:10:48.593054 trace git-lfs: Filled credentials for http://127.0.0.1:61772/verify-fail-10-times
    10:10:48.593067 trace git-lfs: HTTP: POST http://127.0.0.1:61772/verify-fail-10-times.git/info/lfs/objects/batch
    10:10:48.593655 trace git-lfs: HTTP: 200
    10:10:48.593713 trace git-lfs: HTTP: {"objects":[{"oid":"2dc9e3a3629196a5185e80d0522f9020ae9db29e21fc04b7d8016b5d44c5a107","size":18,"actions":{"verify":{"href":"http://127.0.0.1:61772/verify","header":{"repo":"verify-fail-10-times"},"expires_at":"0001-01-01T00:00:00Z"}}}]}
```

The fun part is the batch api response returns only a `verify` action, and not an `upload` action. Therefore, the transfer adapter doesn't think it has anything to do. I'll get the test passing in this PR, and post a fix so that the basic and tus upload adapters still process `verify` actions.

lfsapi/errors.go
```diff
@@ -58,6 +58,10 @@ func (c *Client) handleResponse(res *http.Response) error {
 		return errors.NewAuthError(err)
 	}
 
+	if res.StatusCode == 503 {
+		return errors.NewRetriableError(err)
```
ttaylorr: 🤘

tq/basic_upload.go
```diff
@@ -112,11 +135,10 @@ func (a *basicUploadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb Progres
...
 	}
 
-	// A status code of 403 likely means that an authentication token for the
-	// upload has expired. This can be safely retried.
-	if res.StatusCode == 403 {
-		err = errors.New("http: received status 403")
-		return errors.NewRetriableError(err)
+	switch res.StatusCode {
+	case 403, // likely an expired auth token
+		503: // service unavailable
```
ttaylorr: Hm, this is definitely a more pedantic review comment, but I am not sure that I'm convinced of this formatting. I like that it will make diffs easier to read should we add more HTTP status codes that warrant calling `errors.NewRetriableError` (though I'm not sure how likely that is to occur), but feel that having the `503:` case arm at the same level of indentation as the case body is confusing.

What do you think about this as an alternative?

```go
switch res.StatusCode {
case 403, 503: // expired auth token, service unavailable, respectively
        return errors.NewRetriableError(...)
}
```

lfsapi/client.go
```diff
@@ -159,15 +159,15 @@ func (c *Client) doWithRedirects(cli *http.Client, req *http.Request, via []*htt
 
 	requests := tools.MaxInt(0, retries) + 1
 	for i := 0; i < requests; i++ {
+		if seek, ok := req.Body.(io.Seeker); ok {
```
ttaylorr: I think that the performance impact is negligible, but could we avoid making this `Seek` call on the first run of the loop?

tq/meter.go
```diff
@@ -215,14 +218,21 @@ func (m *Meter) str() string {
 	// (%d of %d files, %d skipped) %f B / %f B, %f B skipped
 	// skipped counts only show when > 0
 
+	current := int64(0)
```
ttaylorr: Since `TransferBytes` ought to happen less frequently than sending an update to the terminal, do you think we could store (in addition to the map of OID -> bytes) a total `currentBytes` in the way that we used to? If we use `atomic` to compute this, I think that it will be a net performance gain by calling `atomic.Load<T>` instead of locking a mutex for the total time it takes to loop over the map.

tq/basic_upload.go
technoweenie: I formatted it this way so that the comments described what each status code means. Though a better option is to probably link to the spot in the docs instead.

tq/meter.go
technoweenie: Right, but computing it in a 2nd place is what lead to the bug in the first place. Maybe it means the way that `tools.BodyWithCallback` updates the `tq.Meter` needs to change. I thought about replacing the func callback with a channel, which _should_ ensure the updates are in order.

ttaylorr: I see; thanks for explaining. It seems like the main change that is required here are some stricter guarantees about the order in which `tools.BodyWithCallback` calls the callback, so perhaps we could guard that with a `sync/mutex`?

I don't think that we have control about the order in which updates are sent to the `tools.BodyWithCallback`, but we can definitely control when the subsequent calls are made.

That all said, I am less concerned about whether we do math once or `n` times, but I think the semantics of trust regarding the `tools.BodyWithCallback` are important for other callsites besides the `tq/meter` one.

PR #2809, opened 2018-01-05
Batch send refspec take 2

technoweenie:
This is a follow-up to #2806, finally teaching the `tq.TransferQueue` to pass on the refspec to the server.


commands/uploader.go
```diff
@@ -31,7 +31,9 @@ func uploadForRefUpdates(ctx *uploadContext, updates []*refUpdate, pushAll bool)
 
 	verifyLocksForUpdates(ctx.lockVerifier, updates)
 	for _, update := range updates {
-		q := ctx.NewQueue() // initialized here to prevent looped defer
+		q := ctx.NewQueue( // initialized here to prevent looped defer
```
ttaylorr: Nit: should this comment be moved to the line above? I think if we ever add arguments to this method, it will help keep the lines short.

test/test-pre-push.sh
```diff
@@ -2,6 +2,85 @@
...
+
+  git config "lfs.$(repo_endpoint "$GITSERVER" "$reponame").locksverify" false
+  git lfs track "*.dat"
+  echo "hi" > a.dat
+  git add .gitattributes a.dat
+  git commit -m "add a.dat"
+
+  refute_server_object "$reponame" 98ea6e4f216f2fb4b69fff9b3a44842c38686ca685f3f55dc48c5d3fb1107be4 "refs/heads/tracked"
+
+  # for some reason, using 'tee' and $PIPESTATUS does not work here
```
ttaylorr: I believe that `${PIPESTATUS[1]} would work, since there are two commands being piped here.

test/test-pre-push.sh
```diff
@@ -2,6 +2,85 @@
...
+  git lfs track "*.dat"
+  echo "hi" > a.dat
+  git add .gitattributes a.dat
+  git commit -m "add a.dat"
+
+  refute_server_object "$reponame" 98ea6e4f216f2fb4b69fff9b3a44842c38686ca685f3f55dc48c5d3fb1107be4 "refs/heads/other"
+
+  # for some reason, using 'tee' and $PIPESTATUS does not work here
+  set +e
+  echo "refs/heads/master master refs/heads/master 0000000000000000000000000000000000000000" |
```
ttaylorr: See above.

test/test-pre-push.sh
technoweenie: That must be copy pasta, I don't recall any issues. At any rate, it's just a test so it doesn't matter ;)

commands/uploader.go
technoweenie: Yeah I suppose. The comment could be moved when the first comment is added.

PR #2811, opened 2018-01-06
tq: standardized progress meter formatting

ttaylorr:
This pull request adopts the recommendation for a new style of progress meter proposed by @bitinn and @technoweenie in https://github.com/git-lfs/git-lfs/issues/2802, and https://github.com/git-lfs/git-lfs/issues/2802#issuecomment-355317842, respectively:

> Would it be possible to change the meter to `[Uploading | Downloading] LFS objects: 100% (63/63), 2.01 MiB | 280.00 KiB/s, done.` That matches the "Writing objects" message, retains the file counts, and even shows current speed.

The style that is implemented in this pull request is exactly as @technoweenie's recommendation above:

```
~/example-repository (master) $ git lfs pull
Downloading LFS objects: 100% (501/501), 1.9 KB | 210 B/s, done
```

Here's a large overview of the changes:

1. **Remove the notion of 'skipped' files.** This term has been used to indicate files that are unnecessary to re-transfer, for they are already present on either the client (in the case of a fetch) or the server (in the case of a push). This term has unfortunately also been the source of some confusion by users. To remove this source of confusion, let's instead treat "skipped" files as already having been transferred, and instead of incrementing a "skipped" count, simply increment the number of files and bytes as already having been transferred, without affecting the true transfer speed.

2. **Introduce a transfer rate indicator.** To match the style of progress meter used by Git to indicate push/pull operations, introduce to Git LFS's progress meter the average rate at which the transfer is occurring.<sup>[1]</sup> To keep track of this, each time we receive a read update, we first check if it has been sufficiently long enough since the last time we calculated an average (to avoid a false sense of precision) and, if so, recalculate the average rate of transfer by using a Cumulative Moving Average (C.M.A) function, so as to weight more heavily recent transfer speeds, so that they dominate the average.

The average is calculated as:

mathtex

Where:

- `avg` is the average rate of transfer (given as bytes/second)
- `n` is the number of samples taken
- `s` is the number of bytes transferred since the last sample was taken
- `d` is the duration of time that has passed since the last sample

I am hoping that this new formatting will clarify what is going on when LFS uploads and downloads objects. The majority of this pull request is changing outdated assertions in the integration suite to match the new formatting.

Closes: https://github.com/git-lfs/git-lfs/issues/2802.

<sup>[1]</sup>: This is actually something that I've been wanting to add to LFS for some time now, but haven't found the right time to do it. I have published [this][1] Gist around a year ago at the time of writing this pull request. That code was originally intended for something exactly like this.

##

/cc @git-lfs/core https://github.com/git-lfs/git-lfs/issues/2802

[1]: https://gist.github.com/ttaylorr/ffe27bae92ebfffe5d805b06766b2e03/

### DIFF
--- a/tq/meter.go
+++ b/tq/meter.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -18,11 +19,11 @@ import (
 // files and bytes transferred as well as the number of files and bytes that
 // get skipped because the transfer is unnecessary.
 type Meter struct {
-	DryRun bool
-	Logger *tools.SyncWriter
+	DryRun    bool
+	Logger    *tools.SyncWriter
+	Direction Direction
 
 	finishedFiles     int64 // int64s must come first for struct alignment
-	skippedFiles      int64
 	transferringFiles int64
 	estimatedBytes    int64
 	lastBytes         int64
@@ -30,13 +31,11 @@ type Meter struct {
 	sampleCount       uint64
 	avgBytes          float64
 	lastAvg           time.Time
-	skippedBytes      int64
 	estimatedFiles    int32
 	paused            uint32
 	fileIndex         map[string]int64 // Maps a file name to its transfer number
 	fileIndexMutex    *sync.Mutex
 	updates           chan *tasklog.Update
-	direction         Direction
 }
 
 type env interface {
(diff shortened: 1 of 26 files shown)


tq/meter.go
```diff
@@ -141,12 +142,29 @@ func (m *Meter) StartTransfer(name string) {
 
 // TransferBytes increments the number of bytes transferred
 func (m *Meter) TransferBytes(direction, name string, read, total int64, current int) {
+	now := time.Now()
+
```
technoweenie: I'd put this under the `if m == nil` check, since nothing happens with `now` before that return.

ttaylorr: Good call. I was wavering on this when writing the patch, and left it at the top because I had convinced myself that waiting until _after_ the `if` would introduce imprecision in the averaging. While probably true to a degree, I'm sure that the effect is negligible, and the code would be more readable if the "first thing" that is supposed to happen came after the nullity check. Changed in cf8d827.

PR #2843, opened 2018-01-26
[WIP] git lfs status issue?

larsxschneider:
We noticed that `git lfs status` reports changes in "non Git LFS tracked" files in the "Git LFS objects not staged for commit" section. Is this as it should be or is it a bug? 

If it is no bug, then I would argue it is at least confusing. 
If it is a bug, then you have a test case here.

--

PS: Are you OK with these kind of PR's to report "bugs"? Feel free to close the PRs if you don't want to tackle them. It won't hurt my feelings 😄

### DIFF
--- a/test/test-status.sh
+++ b/test/test-status.sh
@@ -2,6 +2,30 @@
 
 . "test/testlib.sh"
 
+begin_test "status with non LFS files"
+(
+  set -e
+
+  mkdir repo-0
+  cd repo-0
+  git init
+
+  echo "some text" > some.txt
+  git add some.txt
+  git commit -m "add text file"
+  echo "changed text" > some.txt
+
+  expected="On branch master
+
+Git LFS objects to be committed:
+
+
+Git LFS objects not staged for commit:"
+
+  [ "$expected" = "$(git lfs status)" ]
+)
+end_test
+
 begin_test "status"
 (
   set -e


test/test-status.sh
```diff
@@ -2,6 +2,30 @@
...
+  git add some.txt
+  git commit -m "add text file"
+  echo "changed text" > some.txt
+
+  expected="On branch master
+
+Git LFS objects to be committed:
+
+
+Git LFS objects not staged for commit:"
```
larsxschneider: The current output here is:
```
    Git LFS objects not staged for commit:

    	some.txt (Git: a23e5fd -> File: 93e2157)
```

ttaylorr:
Hi, @larsxschneider 👋! Thanks as always for opening something up. (I'm not sure if it was here or elsewhere, but I am absolutely in favor of having pull requests open showing the bug. It's a better starting point for me if there's an issue that should be addressed; that said issues are fine, too.)

> If it is no bug, then I would argue it is at least confusing.

This is not a bug; I noted this here (via: https://github.com/git-lfs/git-lfs/pull/2070):

> Since #2042, we're showing _all_ files (not just those tracked by LFS) in the status output. I think this makes sense since we want to show files that are supposed to be tracked by LFS but aren't. One thing we could do is only show files which match the patterns in `.gitattributes`, but I like the complete-ness of what we have right now.

I still think that it is beneficial to have complete results in `git lfs status` (in other words, I wouldn't want `git lfs status` to only display the status of LFS-tracked objects) but I agree that this can be confusing. Do you have any suggestions in alternative ways to display this information?

larsxschneider:
How about two more sections?
```
    Git LFS objects to be pushed to origin/master:
    Git LFS objects to be committed:
    Git LFS objects not staged for commit:
    ---
    Git objects to be committed:                         <<< new
    Git objects not staged for commit:                   <<< new
```

ttaylorr:
> How about two more sections?

That sounds good to me. I think we'll just have to do some parsing to make sure that the files aren't included in any of the `.gitattribute` sections. Is this a topic that you'd be interested in taking on?

larsxschneider:
Yeah, if time allow I will look into it.

PR #2846, opened 2018-01-30
vendor: update 'xeipuuv/gojsonpointer'

ttaylorr:
This pull request updates the vendored copy of `github.com/xeipuuv/gojsonpointer`.

Running `script/vendor` produces a diff in this tree without updating the hash of the lock file or the project itself.

##

/cc @git-lfs/core


ployees:
ะพพ

PR #2847, opened 2018-01-30
fsck: attach a filter to exclude unfetched items from fsck

larsxschneider:
If `lfs.fetchinclude` is used, then `git lfs fsck` reports errors. That might confuse users.

### DIFF
--- a/commands/command_fsck.go
+++ b/commands/command_fsck.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/git-lfs/git-lfs/filepathfilter"
 	"github.com/git-lfs/git-lfs/git"
 	"github.com/git-lfs/git-lfs/lfs"
 	"github.com/spf13/cobra"
@@ -45,6 +46,13 @@ func fsckCommand(cmd *cobra.Command, args []string) {
 			Panic(err, "Error checking Git LFS files")
 		}
 	})
+
+	// If 'lfs.fetchexclude' is set and 'git lfs fsck' is run after the
+	// initial fetch (i.e., has elected to fetch a subset of Git LFS
+	// objects), the "missing" ones will fail the fsck.
+	//
+	// Attach a filepathfilter to avoid _only_ the excluded paths.
+	gitscanner.Filter = filepathfilter.New(nil, cfg.FetchExcludePaths())
 
 	if err := gitscanner.ScanRef(ref.Sha, nil); err != nil {
 		ExitWithError(err)
--- a/test/test-fetch.sh
+++ b/test/test-fetch.sh
@@ -64,6 +64,9 @@ begin_test "fetch"
 
   git lfs fetch 2>&1 | grep "Downloading LFS objects: 100% (1/1), 1 B"
   assert_local_object "$contents_oid" 1
+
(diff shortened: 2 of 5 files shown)


test/test-fetch.sh
```diff
@@ -146,6 +164,9 @@ begin_test "fetch with exclude filters in gitconfig"
   git lfs fetch origin master newbranch
   refute_local_object "$contents_oid"
   assert_local_object "$b_oid" 1
+
+  git lfs fsck 2>&1 | tee fsck.log
+  grep "Git LFS fsck OK" fsck.log
```
larsxschneider: This check fails with:
```
Object a.dat (ca978112ca1bbdcafac231b39a23dc4da786eff8147c4e72b9807785afee48bb) could not be checked: no such file or directory
    Moving corrupt objects to /foo/test-fetch.sh-50878/clone/.git/lfs/bad
    rename /foo/test-fetch.sh-50878/clone/.git/lfs/objects/ca/97/ca978112ca1bbdcafac231b39a23dc4da786eff8147c4e72b9807785afee48bb /foo/test-fetch.sh-50878/clone/.git/lfs/bad/ca978112ca1bbdcafac231b39a23dc4da786eff8147c4e72b9807785afee48bb: no such file or directory
```

All other checks above pass!

ttaylorr:
@larsxschneider I think I pushed a set of patches which should fix it. Can you take a look and let me know if it works as expected?

commands/command_fsck.go
```diff
@@ -46,6 +47,13 @@ func fsckCommand(cmd *cobra.Command, args []string) {
 		}
 	})
 
+	// If 'lfs.fetchexclude' is set and 'git lfs fsck' is run after the
+	// initial fetch (i.e., has elected to fetch a subset of Git LFS
+	// objects), the "missing" ones will fail the fsck.
+	//
+	// Attach a filepathfilter to avoid _only_ the excluded paths.
+	gitscanner.Filter = filepathfilter.New(nil, cfg.FetchExcludePaths())
```
larsxschneider: I wonder if we should use the `includePaths` here too? Could you think of a situation where this would be necessary? E.g. is it possible to exclude `/foo/*` and include `/foo/bar` (I have never tried that).

larsxschneider:
Any idea why AppVeyor fails? I suspect it's unrelated to this PR?!

commands/command_fsck.go
ttaylorr: > I wonder if we should use the `includePaths` here too?

I don't think so, precisely since an include filter does not override an exclude filter. In the above example, the filter composed of including `/foo/bar` and excluding `/foo/*` would _not_ match `/foo/bar`.

The reason that I only rely on the exclude filter is because we want to check _all_ files, (regardless if we're only including a subset of them), but we definitely _don't_ want to check objects that we haven't downloaded.

Does this seem OK?

ttaylorr:
> Any idea why AppVeyor fails? I suspect it's unrelated to this PR?!

I think it's unrelated, and I've restarted the build manually.

commands/command_fsck.go
larsxschneider: 👍

PR #2851, opened 2018-02-01
prune: always prune excluded paths

larsxschneider:
People use 'lfs.fetchexclude' to limit the download of Git LFS content.
If a user excludes Git LFS files with 'lfs.fetchexclude' after a full 
clone, then Git LFS would not smudge the files in the working directory
but it would keep the LFS content in the LFS storage directory (usually 
.git/lfs/objects). Git LFS would keep the LFS content even after a
'git lfs prune' run. 

Change that behavior and prune all objects that match the 
'lfs.fetchexclude' pattern independent of their age.

---

⚠️  I am unfamiliar with these areas of the code. Please review carefully! All good ideas stolen from @ttaylorr 's great work in #2847 !

### DIFF
--- a/commands/command_prune.go
+++ b/commands/command_prune.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/git-lfs/git-lfs/filepathfilter"
 	"github.com/git-lfs/git-lfs/fs"
 	"github.com/git-lfs/git-lfs/git"
 	"github.com/git-lfs/git-lfs/lfs"
@@ -86,6 +87,8 @@ func prune(fetchPruneConfig lfs.FetchPruneConfig, verifyRemote, dryRun, verbose 
 	retainChan := make(chan string, 100)
 
 	gitscanner := lfs.NewGitScanner(nil)
+	gitscanner.Filter = filepathfilter.New(nil, cfg.FetchExcludePaths())
+
 	go pruneTaskGetRetainedCurrentAndRecentRefs(gitscanner, fetchPruneConfig, retainChan, errorChan, &taskwait)
 	go pruneTaskGetRetainedUnpushed(gitscanner, fetchPruneConfig, retainChan, errorChan, &taskwait)
 	go pruneTaskGetRetainedWorktree(gitscanner, retainChan, errorChan, &taskwait)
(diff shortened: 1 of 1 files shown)


larsxschneider:
`test-fetch.sh` fails right now because `git lfs fsck` does not respect the exclude pattern here. As soon as #2847 is merged (and this PR rebased) the checks should pass.

commands/command_prune.go
```diff
@@ -327,8 +330,10 @@ func pruneTaskGetRetainedAtRef(gitscanner *lfs.GitScanner, ref string, retainCha
 			return
 		}
 
-		retainChan <- p.Oid
-		tracerx.Printf("RETAIN: %v via ref %v", p.Oid, ref)
+		if gitscanner.Filter.Allows(p.Name) {
```
ttaylorr: I think this is the behavior that you're looking for, but other APIs from the `gitscanner` code do this filtering from within the `lfs` package, such that file- (or pointer-) names that don't pass the filter aren't propagated out to the callbacks.

For consistency: what do you think about moving those checks there?

commands/command_prune.go
```diff
@@ -86,6 +87,8 @@ func prune(fetchPruneConfig lfs.FetchPruneConfig, verifyRemote, dryRun, verbose
 	retainChan := make(chan string, 100)
 
 	gitscanner := lfs.NewGitScanner(nil)
+	gitscanner.Filter = filepathfilter.New(nil, cfg.FetchExcludePaths())
```
ttaylorr: Nice, I think we can avoid the `include` filters, since we only want to _not_ prune files that have been explicitly excluded from being downloaded in the first place.

commands/command_prune.go
larsxschneider: Of course! That was just necessary because I haven't used your changes from #2847 ... stupid me. Now everything is in master and rebased my branch here.

commands/command_prune.go
```diff
@@ -344,10 +347,10 @@ func pruneTaskGetPreviousVersionsOfRef(gitscanner *lfs.GitScanner, ref string, s
 		if err != nil {
 			errorChan <- err
 			return
-		} else {
-			retainChan <- p.Oid
-			tracerx.Printf("RETAIN: %v via ref %v >= %v", p.Oid, ref, since)
 		}
+
+		retainChan <- p.Oid
+		tracerx.Printf("RETAIN: %v via ref %v >= %v", p.Oid, ref, since)
```
larsxschneider: I guess aee2e19 is just my OCD 😄 
Let me know if you don't like this kind of stuff and I remove the commit from the PR!

ttaylorr: No, it's great! Thank you.

PR #2866, opened 2018-02-14
"git clone" does not smudge all files

larsxschneider:
Commit 1 and 2 are cleanup and preparation.

Commit 3 demonstrates a bug in `git clone`. Apparently files in folders with an extension that is tracked are not smudged. I haven't investigated further yet and therefore I don't know if the problem is in Git or Git LFS.

### DIFF
--- a/test/test-clone.sh
+++ b/test/test-clone.sh
@@ -4,13 +4,13 @@
 
 ensure_git_version_isnt $VERSION_LOWER "2.2.0"
 
-begin_test "clone"
+begin_test "lfs clone (deprecated)"
 (
   set -e
 
-  reponame="$(basename "$0" ".sh")"
+  reponame="lfs-clone-deprecated"
   setup_remote_repo "$reponame"
-  clone_repo "$reponame" repo
+  clone_repo "$reponame" "$reponame"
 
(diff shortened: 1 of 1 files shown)


test/test-clone.sh
```diff
@@ -4,16 +4,69 @@
...
+  grep "Cloning into" clone-plain.log
+  # should be no filter errors
+  [ ! $(grep "filter" clone-plain.log) ]
+  [ ! $(grep "error" clone-plain.log) ]
+  # should be cloned into location as per arg
+  [ -d "$newclonedir" ]
+
+  # check a few file sizes to make sure pulled
+  pushd "$newclonedir"
+    [ $(wc -c < "folder.foo/file1.txt") -eq 100 ]
```
larsxschneider: This check fails. `folder.foo/file1.txt` is not smudged.

larsxschneider: Scratch that. I just realized that `lfstest-testutils` always generates LFS files... 
I saw this bug in real life and this test case made me think that I was able to reproduce it 😞

ttaylorr:
Sounds like everything is OK? 😄

PR #2867, opened 2018-02-14
Support default TTL for authentication tokens acquired via SSH

NoEffex:
Certain LFS implementations do not return expires_in (or expires_in) from git-lfs-authenticate yet still enforce the TTL. This causes larger repositories to fail to download the LFS files if the transfer exceeds that TTL and it attempts to use the expired token. This workaround should not be necessary but I expect it's relatively common amongst implementations that support git-lfs-authenticate over SSH. The implementation that this workaround is targeting specifically is Bitbucket Server but likely affects Bitbucket Cloud as well as it is also missing the expires_in field. GitHub correctly includes the TTL which is expected as it is the reference implementation.

I understand it's not ideal to include workarounds for certain implementations but it does resolve it for Bitbucket and I cannot guarantee whether or not they will include the TTL in the near future as it is an optional field.

### DIFF
--- a/lfsapi/ssh.go
+++ b/lfsapi/ssh.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"os"
 	"os/exec"
 	"path/filepath"
 	"regexp"
@@ -71,7 +70,7 @@ func (r *sshAuthResponse) IsExpiredWithin(d time.Duration) (time.Time, bool) {
 }
 
 type sshAuthClient struct {
-	os config.Environment
+	os  config.Environment
 	git config.Environment
 }
 
--- a/lfsapi/ssh.go
+++ b/lfsapi/ssh.go
@@ -66,7 +66,7 @@ type sshAuthResponse struct {
 }
 
 func (r *sshAuthResponse) IsExpiredWithin(d time.Duration) (time.Time, bool) {
-	expiresAt := time.Now()
+	expiresAt := time.Time{}
 	if r.ExpiresAt != nil {
 		expiresAt = *r.ExpiresAt
 	}
--- a/lfsapi/ssh.go
+++ b/lfsapi/ssh.go
@@ -59,14 +59,22 @@ type sshAuthResponse struct {
 	Message   string            `json:"-"`
 	Href      string            `json:"href"`
 	Header    map[string]string `json:"header"`
-	ExpiresAt time.Time         `json:"expires_at"`
-	ExpiresIn int               `json:"expires_in"`
+	ExpiresAt *time.Time        `json:"expires_at"`
+	ExpiresIn *int              `json:"expires_in"`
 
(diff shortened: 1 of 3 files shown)


NoEffex:
I thought I accepted the CLA prior to submitting this but I evidently did not. I've since done so.

lfsapi/ssh.go
```diff
@@ -100,6 +101,12 @@ func (c *sshAuthClient) Resolve(e Endpoint, method string) (sshAuthResponse, err
 		res.Message = strings.TrimSpace(errbuf.String())
 	} else {
 		err = json.Unmarshal(outbuf.Bytes(), &res)
+		if res.ExpiresIn <= 0 {
```
ttaylorr: I think this line has the potential to cause issues when the server _does_ send the `ExpiresIn` property, but it is equal to 0, which the spec permits [[1]](https://github.com/git-lfs/git-lfs/blob/master/tq/schemas/http-batch-response-schema.json#L17-L21).

I believe what will happen here is that the token is expired, this change will instruct LFS to fallback to `lfs.defaulttokenttl`, and the request will be made in error.

Do you think that we could change this to `if res.ExpiresIn != 0`, since the server is allowed to send negative values (indicating that the token has already expired?) `res.ExpiresIn == 0` will still be broken, but that's OK since there's no way to distinguish between `0` and "I didn't send you anything".

NoEffex: I think the best approach might be to change ExpiresIn and ExpiresAt to pointers as it unmarshals pointers correctly and we need to know if they exist or not. I've updated the pull request addressing that.

Do you have any objections to adding an environment variable in addition to the configuration option?

lfsapi/ssh.go
```diff
@@ -59,18 +59,27 @@ type sshAuthResponse struct {
 	Message   string            `json:"-"`
 	Href      string            `json:"href"`
 	Header    map[string]string `json:"header"`
-	ExpiresAt time.Time         `json:"expires_at"`
-	ExpiresIn int               `json:"expires_in"`
+	ExpiresAt *time.Time        `json:"expires_at"`
```
NoEffex: I'd like to add getters for these two pointers and handle the default values in those for cleanliness but I'd like to get another opinion before I do that. Alternatively I could check them in the same way that tools.IsExpiredAtOrIn does and simply not use the value at all if its nil.

ttaylorr: > Alternatively I could check them in the same way that tools.IsExpiredAtOrIn does and simply not use the value at all if its nil.

I think that I prefer this one, since only public values are able to be seen by `encoding/json`, and it would be good to reduce the number of 'entries' into a value (either by calling a getter or accessing the field directly).

larsxschneider:
@NoEffex I wonder, would it be an option to request a new token via SSH if a once valid token stopped working (= expired)? Would that be an alternative workaround?

PR #2871, opened 2018-02-16
NTLM authentication with SSPI on windows

### LINKED ISSUE #2234: Support Kerberos (TFS 2018)
Having TFS 2018 using NTLM authentication and Git for Windows using Git Credential Manager, authentication for Git works out of the box without any user interaction using Kerberos. GCM stores the credentials with only username. As soon as Git LFS is used an error is thrown `Your user name must be of the form DOMAIN\\user. It is currently foo`. The only solution is to manually change the credential entry to a `domain\user` format, which is required to be done by every user.

It would be nice if Git LFS would support Kerberos.

See also: https://www.visualstudio.com/en-us/docs/git/manage-large-files#known-issues

stffabi:
This introduces the use of [SSPI](https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-server-2000/bb742535(v=technet.10)) on windows to authenticate via NTLM. This allows git-lfs to authenticate through NTLM by using the default credentials of the current logged in user in the case GCM returns empty credentials.

Fixes #2234


stffabi:
I'm relatively new to Go, so any help to bring this into a state that could be merged in would be really appreciated 😄.

lfsapi/ntlm.go
```diff
@@ -9,10 +9,21 @@ import (
...
 
+type ntmlCredentials struct {
+	domain   string
+	username string
+	password string
+}
+
+// Defines if the default credentials of the currently logged in user
+// should be used.
+func (c ntmlCredentials) IsDefaultCredentials() bool {
```
ttaylorr: Could you define this method to be received on the pointer type, like:

```go
func (c *ntmlCredentials) IsDefaultCredentials() bool { ... }
//      ^
```

lfsapi/ntlm.go
```diff
@@ -9,10 +9,21 @@ import (
...
+	return c == ntmlCredentials{}
```
ttaylorr: Going off of the above, I think that it might be clearer to define the body as:

```go
return c.domain == "" && c.username == "" && c.password == ""
```

lfsapi/ntlm.go
```diff
@@ -163,4 +125,21 @@ func rewoundRequestBody(req *http.Request) (io.ReadCloser, error) {
 	return body, err
 }
 
-const ntlmNegotiateMessage = "NTLM TlRMTVNTUAABAAAAB7IIogwADAAzAAAACwALACgAAAAKAAAoAAAAD1dJTExISS1NQUlOTk9SVEhBTUVSSUNB"
+func ntlmGetCredentials(creds Creds) (ntmlCredentials, error) {
```
ttaylorr: I think going along with my suggestion about this method could return a `(*ntlmCredentials, error)`, which would contain a 👍 return value indicating the absence of NTLM credentials.

stffabi: @ttaylorr thanks so much for reviewing the code. I really appreciate your feedback.

I don't quite understand what you mean by "which would contain a (thumbsup) return value". Do you mean that we could return `(nil, nil)` instead of `(&ntlmCredentials{}, nil) in case no NTLM credentials were provided? This would also mean we could get rid of `IsDefaultCredentials`, so no NTLM credentials would for windows simply mean use the default ones of the current user.

ttaylorr: > I don't quite understand what you mean by "which would contain a (thumbsup) return value". Do you mean that we could return `(nil, nil)` instead of `(&ntlmCredentials{}, nil)` in case no NTLM credentials were provided?

Your intuition is totally correct -- and my apologies for not being as clear as I could have been. My intention was to suggest that `nil` is a clearer representation of "I couldn't find credentials" than an empty `ntlmCredentials` instance.

> This would also mean we could get rid of `IsDefaultCredentials`, so no NTLM credentials would for windows simply mean use the default ones of the current user.

Indeed 👍.

stffabi:
@ttaylorr I've fixed your suggestions and it's ready for another review.

PR #2875, opened 2018-02-20
filepathfilter: implement using wildmatch

ttaylorr:
This pull request replaces the current implementation in package `filepathfilter` with a shim over `github.com/git-lfs/wildmatch`, which implements the Git/rsync-style wildmatch grammar.

The motivation for this change comes from https://github.com/git-lfs/git-lfs/issues/2798 and https://github.com/git-lfs/git-lfs/pull/2799, wherein @larsxschneider notes that certain patterns don't match filenames that they should. In particular, he shows that the following as examples [[1]](https://github.com/git-lfs/git-lfs/pull/2799):

- `**/*ar/**` (`filepathfilter` does not support `*` wildcards with characters next to them).
- `*.[mM][dD]` (`filepathfilter` does not support character classes).

I began trying to add support for these into the existing implementation of `filepathfilter`, but had difficulty when trying to compose multiple pattern matching behaviors together. For instance, it is currently hard to express "I would like to match any number of directories, then a substring, then a character class".

Instead, `github.com/git-lfs/wildmatch` was written, which composes different pattern matching behavior together in a functional style, and passes all of the tests upstream Git uses for pathspec and attribute-style matching [[2]](https://github.com/git/git/blob/v2.16.1/t/t3070-wildmatch.sh).

This pull request replaces the existing engine in filepathfilter with the implementation found in `github.com/git-lfs/wildmatch`. The aim during this replacement is to __not break existing pattern matching behavior__. On that, here is some notable information:

1. The current grammar specification in `filepathfilter` matches somewhere between the specification of pathspec, and gitattributes-style matching. (For instance, `foo` in `.gitattributes` matches only the file `foo`, but considered as a pathspec, matches a file named foo, or any direct child of a tree called foo.)
2. Existing patterns are rewritten according to a set of rules to retain behavior while being expressed as wildmatch patterns.
3. Support for subtree culling was removed, since the above translation makes almost every subtree unable to be pruned (most patterns now start with a wildcard, unless they are rooted).
4. Support for `\` directory separators is removed, since modern mingw shells translate these paths automatically by them time they reach the program [[3]](www.mingw.org/wiki/Posix_path_conversion).

Here's a breakdown of the order in which things happened:

1. 529dc7a: remove support for AllowsPattern, which is unused.
2. 44f6f7c: vendor `github.com/git-lfs/wildmatch`.
3. f54531a: remove `HasPrefix` and subtree culling
4. 3648b19...9e207ed: rewrite existing patterns in terms of the new grammar.
5. da86a43... ff04fc8: remove `filepath.Join` and `filepath.Clean` calls (to assume POSIX compliancy).
6. ff04fc8: test cleanup.
7. 451f92a: re-vendor `github.com/git-lfs/wildmatch`.

Closes: https://github.com/git-lfs/git-lfs/issues/2798.
Closes: https://github.com/git-lfs/git-lfs/issues/2799.

##


genghaibo:
This pull request replaces the current implementation in package filepathfilter with a shim over github.com/git-lfs/wildmatch, which implements the Git/rsync-style wildmatch grammar.

The motivation for this change comes from #2798 and #2799, wherein @larsxschneider notes that certain patterns don't match filenames that they should. In particular, he shows that the following as examples [1]:

**/*ar/** (filepathfilter does not support * wildcards with characters next to them).
*.[mM][dD] (filepathfilter does not support character classes).
I began trying to add support for these into the existing implementation of filepathfilter, but had difficulty when trying to compose multiple pattern matching behaviors together. For instance, it is currently hard to express "I would like to match any number of directories, then a substring, then a character class".

Instead, github.com/git-lfs/wildmatch was written, which composes different pattern matching behavior together in a functional style, and passes all of the tests upstream Git uses for pathspec and attribute-style matching [2].

This pull request replaces the existing engine in filepathfilter with the implementation found in github.com/git-lfs/wildmatch. The aim during this replacement is to not break existing pattern matching behavior. On that, here is some notable information:

The current grammar specification in filepathfilter matches somewhere between the specification of pathspec, and gitattributes-style matching. (For instance, foo in .gitattributes matches only the file foo, but considered as a pathspec, matches a file named foo, or any direct child of a tree called foo.)
Existing patterns are rewritten according to a set of rules to retain behavior while being expressed as wildmatch patterns.
Support for subtree culling was removed, since the above translation makes almost every subtree unable to be pruned (most patterns now start with a wildcard, unless they are rooted).
Support for \ directory separators is removed, since modern mingw shells translate these paths automatically by them time they reach the program [3].
Here's a breakdown of the order in which things happened:

529dc7a: remove support for AllowsPattern, which is unused.
44f6f7c: vendor github.com/git-lfs/wildmatch.
f54531a: remove HasPrefix and subtree culling
3648b19...9e207ed: rewrite existing patterns in terms of the new grammar.
da86a43... ff04fc8: remove filepath.Join and filepath.Clean calls (to assume POSIX compliancy).
ff04fc8: test cleanup.
451f92a: re-vendor github.com/git-lfs/wildmatch.
Closes: #2798.
Closes: #2799.

ttaylorr added some commits 23 days ago
 @ttaylorr
filepathfilter: remove support for 'AllowsPattern'
529dc7a
 @ttaylorr
filepath: support legacy wildcard behavior
2f37dbf
 @ttaylorr
vendor: vendor 'github.com/git-lfs/wildmatch'
44f6f7c
 @ttaylorr
filepathfilter: implement in terms of wildmatch
629a935
 @ttaylorr
filepathfilter: remove outdated benchmark
af80061
 @ttaylorr
filepathfilter: don't chomp through to basepath
c847f54
 @ttaylorr
filepathfilter: remove 'HasPrefix' from interface
f54531a
 @ttaylorr
git/githistory: remove filepath.Join call
2f8c6ee
 @ttaylorr
filepathfilter: rewrite existing wildcard patterns
3648b19
 @ttaylorr
tools/filetools: do not use filepath.Join
61fce9a
 @ttaylorr
tools/filetools: remove call to filepath.Clean
9d21cfc
 @ttaylorr
filepath: remaining test cases
9e207ed
 @ttaylorr
filepath: rewrite more legacy patterns in terms of wildmatch
b2eac6f
 @ttaylorr
filepathfilter: assume that paths are in POSIX form
da86a43
 @ttaylorr
tools/filetools: only handle non-Windows paths
297b651
 @ttaylorr
git/githistory: only handle non-Windows paths
9e93c85
 @ttaylorr
filepathfilter: use tracerx to print information about pattern matching
e8009e1
 @ttaylorr
lfs: remove filepath.Join calls
fd69029
 @ttaylorr
lfs/util: assume POSIX paths by calling filepath.ToSlash
78b04d4
 @ttaylorr
commands/migrate: OS-independent path comparison
3cad02f
 @ttaylorr
test/test-migrate-import: do not use $PATH_SEPARATOR
eef8653
 @ttaylorr
test: remove improper prefix include
ff04fc8
 @ttaylorr
vendor: ⬆️ wildmatch
451f92a
 @ttaylorr ttaylorr added the review  label 23 hours ago
 @ttaylorr ttaylorr added this to the v2.4.0 milestone 23 hours ago
 @ttaylorr
Merge branch 'master' into wildmatch

thuanrausaigon:
Day la van de nghiem trong cua ban.va ban phai chiu trach nhiem chocho nhung gi ma ban gay ra!

filepathfilter/filepathfilter.go
```diff
@@ -98,257 +90,72 @@ func (f *Filter) HasPrefix(prefix string) bool {
...
 	return true
 }
 
-// String returns a string representation of the underlying pattern for which
-// this *simpleExtPattern is matching.
-func (p *simpleExtPattern) String() string {
-	return fmt.Sprintf("*%s", p.ext)
+type wm struct {
+	w    *wildmatch.Wildmatch
+	dirs bool
```
larsxschneider: Would this be use case for [embedded fields](https://golang.org/ref/spec#Struct_types)?

filepathfilter/filepathfilter.go
```diff
@@ -143,6 +143,12 @@ func (w *wm) String() string {
 func NewPattern(p string) Pattern {
 	p = filepath.Clean(p)
 
+	// Special case: "*" and "*.*" should match everything to match existing
+	// behavior.
+	if p == "*" || p == "*.*" {
+		p = "**/*"
+	}
```
larsxschneider: I am confused here. Shouldn't `*` match everything also with the new behaviour too? I think `.gitattributes` handles it that way...

`*.*` would match all file/folder names with a dot in the new behaivior too, no?

Can you help me understand why we need this conversion?

filepathfilter/filepathfilter.go
```diff
@@ -57,311 +49,129 @@ func patternsToStrings(ps ...Pattern) []string {
...
-		return &pathlessWildcardPattern{
-			rawPattern: cleanpattern,
-			wildcardRE: regexp.MustCompile(regpattern),
+func NewPattern(p string) Pattern {
+	pp := p
+
+	// Special case: the below patterns match anything according to existing
+	// behavior.
+	switch pp {
+	case `*`, `*.*`, `.`, `./`, `.\`:
```
larsxschneider: `.` matched all files? How did this work? Git would clean/smudge all files with `.`, right? Wouldn't that only affect the Git LFS *only* include/exclude patterns?

filepathfilter/filepathfilter.go
```diff
@@ -57,311 +49,129 @@ func patternsToStrings(ps ...Pattern) []string {
...
+		} else {
+			if rooted {
+				// Special case: if there are not any directory
+				// separators, assume "pp" is a substring match.
+				pp = join(pp, "**")
+			} else {
+				// Special case: if there are not any directory
+				// separators, assume "pp" is a substring match.
+				pp = join("**", pp)
+			}
```
larsxschneider: I think this "guessing" is dangerous. We never know what the real intention of the users are. I would rather print a big warning message for a few versions that the meaning of this pattern changed.

git/githistory/rewriter.go
```diff
@@ -311,9 +310,14 @@ func (r *Rewriter) rewriteTree(commitOID []byte, treeOID []byte, path string, fn
 
 	entries := make([]*odb.TreeEntry, 0, len(tree.Entries))
 	for _, entry := range tree.Entries {
-		path := filepath.Join(path, entry.Name)
+		var epath string
```
larsxschneider: Maybe `fullpath`? No idea if this makes it clearer but I stumbled upon `epath`. Took me a second to understand the meaning.

larsxschneider:
BTW: Initially I reviewed commit by commit - but then I realized you changed some lines multiple times. At that point I started reviewing the entire diff. Sorry for the messy review.

filepathfilter/filepathfilter.go
```diff
@@ -61,30 +63,20 @@ func (f *Filter) Allows(filename string) bool {
 		return true
 	}
 
-	if len(f.include)+len(f.exclude) == 0 {
-		return true
```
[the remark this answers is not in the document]
ttaylorr: Good question -- I removed this special case because it adds no behavior that the function doesn't already accomplish. Without any includes/excludes, both of the `for` loops below don't execute, and we fall through into the base case: `return true`. I am happy to remove code when possible, so that is the rationale behind this 😄.

filepathfilter/filepathfilter.go
```diff
@@ -61,30 +63,20 @@ func (f *Filter) Allows(filename string) bool {
...
-			matched = inc.Match(cleanedName)
-			if matched {
-				break
-			}
-		}
-		if !matched {
-			return false
-		}
+	if !matched && len(f.include) > 0 {
+		return false
```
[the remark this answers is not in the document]
ttaylorr: Great suggestion, I definitely prefer `included` to `matched`, since `included` communicates inclusion, but `matched` only communicates ambiguity. "Matched what?"

I adopted this suggestion by renaming this to `included` in c48f8178.

filepathfilter/filepathfilter.go
```diff
@@ -98,257 +90,72 @@ func (f *Filter) HasPrefix(prefix string) bool {
...
 }
 
-func (p *pathlessWildcardPattern) HasPrefix(name string) bool {
-	lit, ok := p.wildcardRE.LiteralPrefix()
-	if !ok {
-		return true
+func (w *wm) chomp(filename string) string {
+	if !w.dirs {
+		base := filepath.Base(filename)
+		for _, sep := range []string{"/", "\\"} {
```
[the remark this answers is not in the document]
ttaylorr: Great catch, I updated this here: 4fb239c7.

filepathfilter/filepathfilter.go
ttaylorr: I think that my argument is in the majority motivated by personal preference, but I avoid embedded fields as is possible, since they generate a source of confusion for me. Though verbose, I tend to prefer the explicit nature of `w.w` over `w.<some-inherited-field>`.

filepathfilter/filepathfilter.go
ttaylorr: > I am confused here. Shouldn't `*` match everything also with the new behaviour too? I think `.gitattributes` handles it that way...

By default `*` is _not_ allowed to cross path separators, but with `WM_PATHNAME`, it can (behaving in-effect as `**`) [[1]](https://github.com/git/git/blob/v2.16.1/wildmatch.c#L110-L111). To the best of my knowledge, you are correct that this is the behavior that `.gitattributes` exhibits [[2]](https://github.com/git/git/blob/v2.16.1/dir.c#L1032-L1034).

I added support for this in https://github.com/git-lfs/wildmatch/pull/3, but removed it in https://github.com/git-lfs/wildmatch/commit/b6f7a2045f03a3b2f0cdfd677e809c95d2e5db6c when the complexity became too large to handle for this one particular exception.

Thus, without `WM_PATHNAME` support in `git-lfs/wildmatch`, we have to rewrite this pattern _as if_ `WM_PATHNAME` did exist in our implementation, which is `**/*` as above. In English, that's "match any number of leading directories, or none at all, and then any filename".

> `*.*` would match all file/folder names with a dot in the new behaivior too, no?

Correct, ~~but this was a special case of the prior implementation where instead of matching any filename with a dot in it, it would instead match any filename at all~~, and I misremembered that this was indeed a special case. I have reverted this behavior in 014fd601.

filepathfilter/filepathfilter.go
ttaylorr: Indeed, `.` used to match all files [[1]](https://github.com/git-lfs/git-lfs/blob/v2.3.4/filepathfilter/filepathfilter.go#L358). This is meant to preserve behavior between versions.

filepathfilter/filepathfilter.go
ttaylorr: Agreed, though I think that my language is not correct here, in particular with my usage of "assume". This translation from old pattern to new pattern is intended to be behavior-preserving, and this is how this pattern worked before. `foo` would match either the file `foo` at the tree root, or any file `foo` further down [[1]](https://github.com/git-lfs/git-lfs/blob/v2.3.4/filepathfilter/filepathfilter_test.go#L42-L49).

I have changed this comment to better convey my intentions in: 2a17ee5c.

git/githistory/rewriter.go
ttaylorr: Good idea, I renamed this to `fullpath` in: 6c782f60.

filepathfilter/filepathfilter.go
larsxschneider: I agree as I tapped into the embedded field confusion trap before. I wonder if it makes sense to rename`type wm` with `type wm_with_dirs` or something. That's long and ugly, though.
🚲-shedding 😉

ttaylorr:
@larsxschneider and I had a chance to discuss this change this morning, the summary of which I'd like to post here.

1. The workaround for `WM_PATHNAME` support in `github.com/git-lfs/wildmatch` is temporarily housed in Git LFS code. Motivated by a consistency between `github.com/git-lfs/wildmatch` and `wildmatch.c` (in git.git), we'd like to move this functionality there. This change will be accomplished in v2.4.1.

2. A bug where `git lfs clone --include='.'` would cause the working tree to appear modified is still problematic, since the Git LFS smudging process (not invoked by Git) would cause files to be converted that `.gitattributes` wouldn't ordinarily filter. Fixing this will involve changing (and defining) a specification for this grammar. This change will occur in v3.0.0.

3. It was discussed whether we should compare the old and new filepathfilter engine (`filepathfilter.go` and `github.com/git-lfs/wildmatch`) as a control/candidate respectively in order to determine which patterns will be broken in v3.0.0. I would like to avoid doing this, since the behavior of the current grammar (and its relative differences from gitattributes) will cause most patterns to be broken.

That all said, I feel comfortable merging this now and including it in the v2.4.0 release. As always, we can continue to iterate on this in subsequent releases, so any issues that should arise can be addressed quickly.

Thanks everyone!

larsxschneider:
Thanks for the great summary @ttaylorr !

One addition: If the old and the new filepathfilter engines run alongside of each other, then the results should *mostly* match even though the patterns are broken. I think we should warn the user *only* if the patterns do not match. But this is something we could implement anytime before v3.0.0

Issue #1073, opened 2016-03-14
Trying to add large file and get: Size must be less than 2147483648


chriskinsman:
I can't find anything about max file size limits.  Is Git-LFS really limited to 2GB?

Thanks,

technoweenie:
It depends on the server. 2GB is the limit on GitHub.com.

metaf:
Sorry to post in a closed issue, but where would be the appropriate place to complain about this limitation/request the 'feature' of supporting files >2GB?

technoweenie:
https://github.com/contact, since it's a github limitation and not an LFS one.

amacneil:
Cold

de-code:
Given that GitHub.com is very common and the first mentioned supported remote server, it would perhaps be good to add that to the documentation? Potentially somewhere where it mentions 'Version large files—even those as large as a couple GB in size'?

hartator:
+1 

Yes, it will be great to add this to the documentation. I've gone through the process to install everything just to be blocked because my file is 2.4 GB.

ttaylorr:
Hi everybody, this is totally something that I think we should address. In the meantime, this is also being discussed in https://github.com/git-lfs/git-lfs/issues/730, so I'm going to lock this and move discussion there. Thanks!

Issue #730, opened 2015-10-11
Reject files based on size on add


brianmingus:
If you are adding many files and just one of them is > ~2GB and it is near the end, then all of the progress in committing the files is lost as you must remove the file and then start from scratch.

[422] Size must be less than 2147483648

brianmingus:
Ok - it seems that the progress is actually cached. I'll leave this open though since it does seem that git-lfs should notify you of the github size limit when it starts tracking a file.

brianmingus:
As far as I can tell, using git rm --cached and git-lfs untrack on files that are too large does not actually remove them. The total file size to be transmitted continues to reflect all the files that I initially added. In addition to this, the messages telling you that some files are too large do not tell you which files are being referred to.

brianmingus:
I've done everything I can to remote all files from git-lfs. Git reports that my entire data directory is untracked, and git-lfs reports that no files are tracked. Nevertheless, when I try to commit it keeps trying to push all these files, and continues failing because some files are too large.

technoweenie:
It sounds like you still have commits to push that contain large files. You'll have to redo those commits.

> it does seem that git-lfs should notify you of the github size limit when it starts tracking a file.

That'd be nice, but that server side limit has nothing to do with the Git LFS client. We could add an API call every time a file is added, but that would add even more overhead to adding files. I'd be interested in PRs that solve this though.

rBatt:
Issue #1073 says it's an issue with GitHub, and not git lfs. I had not realized such a limitation had been placed on the size of individual files. Either way, that's probably the source of the problem here.

otri:
> That'd be nice, but that server side limit has nothing to do with the Git LFS client. We could add an API call every time a file is added, but that would add even more overhead to adding files. I'd be interested in PRs that solve this though.

I'd argue that it's a core Git LFS usage failure, because the repercussions are non-trivial for all users that encounter it.  A quick usability test of adding a file to any Git LFS repository with a reasonable file size limit causes a forced failure mode that a user must unwind by getting into deep internals of how git functions.  Most clients do not catch this issue, because the error is not brought to the surface.

So we Git LFS must handle this file size restriction through a formalized method.  At bare minimum there should be a way to:
- Determine the file size limits on an associated Git LFS server, and cache this locally
- Warn of adding an over-size file to Git LFS at time of the addition, allow a "--force" option
- Formalize handling of adding a file to the Git LFS server that encounters a restriction, and report the specific file in question
[...]

PR #2876, opened 2018-02-21
added note on github file size limit

de-code:
This is related to #730 and #1073. I thought while you are considering the perfect option it would be good to update the documentation. It might be good to tell users about the limitations before installing anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,11 +22,13 @@ binaries available for Mac, Windows, Linux, and FreeBSD. Check out the
 
 By default, the Git LFS client needs a Git LFS server to sync the large files
 it manages. This works out of the box when using popular git repository
-hosting providers like GitHub, Atlassian, etc. When you host your own
+hosting providers like GitHub*, Atlassian, etc. When you host your own
 vanilla git server, for example, you need to either use a separate
 [Git LFS server instance](https://github.com/git-lfs/git-lfs/wiki/Implementations),
 or use the [custom transfer adapter](docs/custom-transfers.md) with
 a transfer agent in blind mode, without having to use a Git LFS server instance.
+
+Note: The current single file size limit for GitHub is 2GB.
 
 You can install the Git LFS client in several different ways, depending on
 your setup and preferences.
--- a/README.md
+++ b/README.md
@@ -22,13 +22,11 @@ binaries available for Mac, Windows, Linux, and FreeBSD. Check out the
 
 By default, the Git LFS client needs a Git LFS server to sync the large files
 it manages. This works out of the box when using popular git repository
-hosting providers like GitHub*, Atlassian, etc. When you host your own
+hosting providers like GitHub, Atlassian, etc. When you host your own
 vanilla git server, for example, you need to either use a separate
 [Git LFS server instance](https://github.com/git-lfs/git-lfs/wiki/Implementations),
 or use the [custom transfer adapter](docs/custom-transfers.md) with
 a transfer agent in blind mode, without having to use a Git LFS server instance.
-
-Note: The current single file size limit for GitHub is 2GB.
 
 You can install the Git LFS client in several different ways, depending on
 your setup and preferences.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,12 +22,14 @@ binaries available for Mac, Windows, Linux, and FreeBSD. Check out the
...
 By default, the Git LFS client needs a Git LFS server to sync the large files
 it manages. This works out of the box when using popular git repository
-hosting providers like GitHub, Atlassian, etc. When you host your own
+hosting providers like GitHub*, Atlassian, etc. When you host your own
 vanilla git server, for example, you need to either use a separate
 [Git LFS server instance](https://github.com/git-lfs/git-lfs/wiki/Implementations),
 or use the [custom transfer adapter](docs/custom-transfers.md) with
 a transfer agent in blind mode, without having to use a Git LFS server instance.
 
+Note: The current single file size limit for GitHub is 2GB.
```
mathstuf: I think that either this should have `*` on it (which probably makes markdown parsers confused) or there should be a stronger link between the footnote mark and this note.

de-code: I thought this might come back. I don't mind either way. Which way would you suggest?

mathstuf: There isn't official support for footnotes in markdown. Something easily searchable at least.

Then again, maybe a section on limitations from common providers is warranted? There's also the monthly bandwidth limitation on GitHub at least.

de-code: Moved to a separate _Limitations_ section. I don't know about the other limitations and other people could perhaps add those. Do you think that might be look better now?

mathstuf:
Looks OK to me.

README.md
```diff
@@ -91,6 +91,10 @@ To https://github.com/git-lfs/git-lfs-test
    67fcf6a..47b2002  master -> master
 ```
 
+## Limitations
+
+The current single file size limit for GitHub is 2GB.
```
larsxschneider: I like that. Maybe extend it with `GitHub and GitHub Enterprise`? 

@bturner Do you know if there is a BitBucket/Stash limitation?
@dosire Do you know if there is a the GitLab limitation?

bturner: @larsxschneider I can't speak for Bitbucket Cloud (bitbucket.org), but Bitbucket Server/Stash, at least as of 5.8, does not impose any limits on the size of individual files; it only monitors how much free space is available and ensures users can't completely fill the disk. I can't say that that will never change (there is an open feature request for quotas, for example), but as of 5.8 (the current release) there are no limits.

I'd wager there _are_ limits for Bitbucket Cloud. Perhaps @erikvanzijst knows what they are.

mathstuf: AFAIK, there are no limitations for self-hosted GitLab (CE, probably the same for EE). I don't know about `gitlab.com`.

As for GitHub, there is also a monthly bandwidth limit for free repositories. No idea for private or Enterprise deployments.

ttaylorr:
Hi all -- I think that this is a great thing to document, but I am not sure that the README is an appropriate place for it. My concern is that this section will become a grab-bag of features that Git LFS doesn't currently support, which is something that I think we (a) should have, and (b) shouldn't be the first thing users see when they navigate to this repository.

I'd like the README to be concise and descriptive. I think that one way to do this is to instead fill this section with a link to a wiki page that documents these limitations. I have created one [here](https://github.com/git-lfs/git-lfs/wiki/Limitations), and think that this would be a good direction towards merging this.

Thoughts?

de-code:
Would it make sense to add a link to the Wiki, in particular this section?

My argument would be that Git LFS is specifically designed to work with large files. So if there is a file size limit (in particular for common provider like GitHub), then it would be fair to point potential users prominently to such limitations. I agree that it's good to keep the README concise. That is why I would probably at least scan the whole of it. Just a pointer that there are such limitations in there would be nice.

mathstuf:
I added the quota information (with links) to that wiki page.

Issue #2879, opened 2018-02-24
panic: runtime error: invalid memory address or nil pointer dereference when pulling large object


jeffreydwalter:
`go version go1.8 linux/386`
`git-lfs/2.3.4 (GitHub; linux 386; go 1.8; git 6f503b2)`

Built by cloning the git-lfs repo and running the following commands:
```
git clone https://github.com/github/git-lfs.git
cd git-lfs
./script/bootstrap
```

I'm getting the following error pulling large files via Bamboo's Git LFS feature:
```
Fetching 'refs/heads/somebranch' from 'ssh://git@git.server.com/blah/repo.git'. Will try to do a shallow fetch.
Warning: Permanently added '[127.0.0.1]:34441' (RSA) to the list of known hosts.
fetch: Fetching reference refs/heads/somebranch
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x804965c]

goroutine 1 [running]:
sync/atomic.AddUint64(0x18dfd2ec, 0x7061358, 0x0, 0x0, 0x0)
        /usr/local/go/src/sync/atomic/asm_386.s:112 +0xc
github.com/git-lfs/git-lfs/tq.(*Meter).Add(0x18dfd2d0, 0x7061358, 0x0)
        /home/build/git-lfs/src/github.com/git-lfs/git-lfs/tq/meter.go:114 +0x7a
github.com/git-lfs/git-lfs/commands.readyAndMissingPointers(0x189d9b40, 0x6, 0x8, 0x18a3a3c0, 0x189d8000, 0x1891a068, 0x0, 0x0, 0x0, 0x0, ...)
        /home/build/git-lfs/src/github.com/git-lfs/git-lfs/commands/command_fetch.go:351 +0x398
github.com/git-lfs/git-lfs/commands.fetchAndReportToChan(0x189d9b40, 0x6, 0x8, 0x18a3a3c0, 0x0, 0x8)
        /home/build/git-lfs/src/github.com/git-lfs/git-lfs/commands/command_fetch.go:271 +0x50
github.com/git-lfs/git-lfs/commands.fetchRef(0x18a08070, 0x28, 0x18a3a3c0, 0x1)
        /home/build/git-lfs/src/github.com/git-lfs/git-lfs/commands/command_fetch.go:142 +0x79
[...]

PR #2880, opened 2018-02-24
Issue #2879: Fix memory alignment issues on 32-bit systems.

jeffreydwalter:
Moved int64 declarations back to the top of the Meter struct for proper memory alignment.

See: https://github.com/git-lfs/git-lfs/issues/2879

### DIFF
--- a/tq/meter.go
+++ b/tq/meter.go
@@ -19,10 +19,6 @@ import (
 // files and bytes transferred as well as the number of files and bytes that
 // get skipped because the transfer is unnecessary.
 type Meter struct {
-	DryRun    bool
-	Logger    *tools.SyncWriter
-	Direction Direction
-
 	finishedFiles     int64 // int64s must come first for struct alignment
 	transferringFiles int64
 	estimatedBytes    int64
@@ -36,6 +32,10 @@ type Meter struct {
 	fileIndex         map[string]int64 // Maps a file name to its transfer number
 	fileIndexMutex    *sync.Mutex
 	updates           chan *tasklog.Update
+
+	DryRun    bool
+	Logger    *tools.SyncWriter
+	Direction Direction
 }
 
 type env interface {


tq/meter.go
```diff
@@ -36,6 +32,10 @@ type Meter struct {
 	fileIndex         map[string]int64 // Maps a file name to its transfer number
 	fileIndexMutex    *sync.Mutex
 	updates           chan *tasklog.Update
+
+	DryRun    bool
+	Logger    *tools.SyncWriter
+	Direction Direction
```
larsxschneider: Nice catch 👍

I wonder if these kind of problems happen often enough to setup a 32-bit build as we did for Git core: https://github.com/git/git/blob/master/ci/run-linux32-build.sh

jeffreydwalter: Thanks!

In my opinion, if you're supporting 32-bit, testing on that platform should be a given. Especially, since you guys are using libraries that depend on memory alignment.

jeffreydwalter:
FWIW, the following structs might also be problematic: https://github.com/git-lfs/git-lfs/blob/774fe1a1b5fbf77d1453243031072764d0be567a/lfsapi/stats.go#L18-L31
https://github.com/git-lfs/git-lfs/blob/42e2e061b51c2e900e8eb7ae750cc1102b2c7d7e/tasklog/percentage_task.go#L12-L23
I've never really used the atomic package so I'm not sure what the rules are for string types in structs when int64's are present. My assumption, based on their docs, is that int64's should always come first.

I guess these two are okay since there aren't any int64's in it, but maybe it's better to move the int32 to the top and put a comment in there just in case?
https://github.com/git-lfs/git-lfs/blob/1af87420505478b9fb526b6f02fe5adab4d44336/git/odb/object_writer.go#L15-L34
https://github.com/git-lfs/git-lfs/blob/f8b11e0e844edb1091cdc2274bef6b190cd6759e/git/odb/object_db.go#L18-L32

ttaylorr:
Hi, @jeffreydwalter -- thanks for opening this and fixing the issue. I have a few thoughts:

1. We should merge this change now, since it indeed fixes an issue for the Meter structure on 32-bit architectures.
2. In a future pull request, we should (a) fix this in other structures, and (b) enable 32-bit builds on supported CI services.

For now, I have queued this for release in v2.4.0.

jeffreydwalter:
@ttaylorr sounds good to me. I'll leave those other structs for you guys.

PR #2884, opened 2018-02-27
git/githistory: copy entries from cache, elsewhere

ttaylorr:
This pull request fixes an issue where the semantics of `range` would cause the resulting tree entries to be built incorrectly when the alias moved during the associated loop.

Since `range`'s second argument, when de-structured, is a pointer alias that moves through the array, appending it onto the end of a slice of other pointers is non-deterministic. If the memory layout changes, or the pointer changes into certain regions of the layout, the resulting tree will be built incorrectly.

To prevent this issue from occurring, introduce `copyEntry` which produces a new, heap-allocated `*odb.TreeEntry` not susceptible to the copying issue.

##

### DIFF
--- a/git/githistory/rewriter.go
+++ b/git/githistory/rewriter.go
@@ -314,12 +314,12 @@ func (r *Rewriter) rewriteTree(commitOID []byte, treeOID []byte, path string, fn
 		path := filepath.Join(path, entry.Name)
 
 		if !r.allows(entry.Type(), path) {
-			entries = append(entries, entry)
+			entries = append(entries, copyEntry(entry))
 			continue
 		}
 
 		if cached := r.uncacheEntry(entry); cached != nil {
-			entries = append(entries, cached)
+			entries = append(entries, copyEntry(cached))
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


git/githistory/rewriter.go
```diff
@@ -314,12 +314,12 @@ func (r *Rewriter) rewriteTree(commitOID []byte, treeOID []byte, path string, fn
...
 
 		if !r.allows(entry.Type(), path) {
-			entries = append(entries, entry)
+			entries = append(entries, copyEntry(entry))
 			continue
 		}
 
 		if cached := r.uncacheEntry(entry); cached != nil {
-			entries = append(entries, cached)
+			entries = append(entries, copyEntry(cached))
```
larsxschneider: So we are creating a deep copy here. That looks correct to me. However, TBH I don't fully see why the shallow copy doesn't work. Are the entries used in any way *after* `r.db.WriteTree(rewritten)`? Until that point `tree.Entries` shouldn't change, no?

larsxschneider: Ahh. `range` [works on a copy](https://www.ardanlabs.com/blog/2017/06/for-range-semantics.html)?! That's why the pointers are not valid anymore when used in `rewritten`?

larsxschneider: https://github.com/golang/go/wiki/Range#gotchas

ttaylorr: That's the one 😄 .

PR #2897, opened 2018-03-04
Bourne shell is enough for bootstrap

yurivict:
BSDs don't have bash installed by default, so this matters for them.


yurivict:
No, bash is needed in other places.

Issue #2900, opened 2018-03-06
Failures after upgrading to v2.4.0


goodtune:
This is not afflicting all users in our team however I have received reports from a few users since upgrading from 2.3.4 to 2.4.0 that pushes are failing.

Below is a traceback I've received.
```
panic: runtime error: index out of range 0 B/s
goroutine 34 [running]:
github.com/git-lfs/git-lfs/tools/humanize.FormatBytes(0xffffffffffffffff, 0xc4204de150, 0x6)
        /tmp/docker_run/src/github.com/git-lfs/git-lfs/tools/humanize/humanize.go:110 +0x1e5
github.com/git-lfs/git-lfs/tq.(*Meter).str(0xc420190360, 0xc420201b00, 0x4298ef)
        /tmp/docker_run/src/github.com/git-lfs/git-lfs/tq/meter.go:241 +0xed
github.com/git-lfs/git-lfs/tq.(*Meter).update(0xc420190360, 0xc42022f000)
        /tmp/docker_run/src/github.com/git-lfs/git-lfs/tq/meter.go:219 +0x4a
github.com/git-lfs/git-lfs/tq.(*Meter).Skip(0xc420190360, 0xffffffffffffffff)
        /tmp/docker_run/src/github.com/git-lfs/git-lfs/tq/meter.go:127 +0x6e
github.com/git-lfs/git-lfs/tq.(*TransferQueue).Skip(0xc420072c00, 0xffffffffffffffff)
        /tmp/docker_run/src/github.com/git-lfs/git-lfs/tq/transfer_queue.go:718 +0x39
github.com/git-lfs/git-lfs/tq.(*TransferQueue).enqueueAndCollectRetriesFor(0xc420072c00, 0xc420280000, 0x2, 0x64, 0x0, 0x0, 0x0, 0x0, 0x0)
        /tmp/docker_run/src/github.com/git-lfs/git-lfs/tq/transfer_queue.go:464 +0x68f
github.com/git-lfs/git-lfs/tq.(*TransferQueue).collectBatches.func1(0xc42027e7c0, 0xc420072c00, 0xc42027e000, 0xc4202c61e0)
        /tmp/docker_run/src/github.com/git-lfs/git-lfs/tq/transfer_queue.go:354 +0x54
created by github.com/git-lfs/git-lfs/tq.(*TransferQueue).collectBatches
[...]

PR #2902, opened 2018-03-07
tq: prevent uint64 underflow with invalid API response

ttaylorr:
This pull request prevents signed-to-unsigned conversion over- and underflows when using package `humanize`, thereby causing panics like in https://github.com/git-lfs/git-lfs/issues/2900.

As noted by @goodtune, the issue stems from a non-compliant implementation of the LFS API when it sends back a size that underflows the `uint64` type (see: https://github.com/git-lfs/git-lfs/issues/2900#issuecomment-370960857).

I think that it is worth being resilient to these failure modes, so let's add `clamp` and `clampf` to prevent overflow during these conversions.

To note, another way that we could fix this issue is by changing the internal representation of these fields such that they are unable to overflow in the first place. I pursued this implementation, but found it to be sub-par, since this approach simply pushes the problem further out through the code.

Closes: https://github.com/git-lfs/git-lfs/issues/2900.

##

/cc @git-lfs/core @goodtune https://github.com/git-lfs/git-lfs/issues/2900

### DIFF
--- a/test/test-push.sh
+++ b/test/test-push.sh
@@ -632,6 +632,7 @@ begin_test "push (with invalid object size)"
   set -e
 
   grep "invalid size (got: -1)" push.log
+  [ "0" -eq "$(grep -c "panic" push.log)" ]
   [ "0" -ne "$res" ]
 
   refute_server_object "$reponame" "$(calc_oid "$contents")"
--- a/tq/meter.go
+++ b/tq/meter.go
@@ -2,6 +2,7 @@ package tq
 
 import (
 	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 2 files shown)


goodtune:
Are we likely to see an official release which includes this fix cut in the near future?

ttaylorr:
Hi @goodtune, this commit is included in the v2.4.0 tag, which was released a few weeks ago. You can grab the release from [the releases page]. Can you give it a shot and let me know if it works for you? Thanks!

[1]: https://github.com/git-lfs/git-lfs/releases/tag/v2.4.0

goodtune:
This landed after the tag/release (see [compare](https://github.com/git-lfs/git-lfs/compare/v2.4.0...5d18f87f009b1f96d484d216868858083cdd0a13)) as #2900 was reported against the v2.4.0 release obtained from PackageCloud.

ttaylorr:
> This landed after the tag/release (see [compare](https://github.com/git-lfs/git-lfs/compare/v2.4.0...5d18f87f009b1f96d484d216868858083cdd0a13)) as #2900 was reported against the v2.4.0 release obtained from PackageCloud.

Ack, how embarrassing. I was thinking of #2883, which is similar but certainly not the same patch. I will release v2.4.1 next week.

ttaylorr:
> Hi @ttaylorr just wanting to ping and see if a 2.4.1 release build will be coming soon?

Hi @goodtune, and thanks for the ping. I'm building the release shortly in https://github.com/git-lfs/git-lfs/pull/3016, and it should be out later today.

PR #2912, opened 2018-03-10
Map local abspath to file://<abspath> endpoint URL

sprohaska:
It can be used to configure a standalonetransferagent for local files,
like:

```
git config lfs.customtransfer.file.path git-lfs-standalonetransfer-file
git config lfs.file:///.standalonetransferagent file
```

Signed-off-by: Steffen Prohaska <prohaska@zib.de>

### DIFF
--- a/lfsapi/endpoint.go
+++ b/lfsapi/endpoint.go
@@ -106,3 +106,10 @@ func endpointFromGitUrl(u *url.URL, e *endpointGitFinder) Endpoint {
 	u.Scheme = e.gitProtocol
 	return Endpoint{Url: u.String()}
 }
+
+func endpointFromLocalPath(path string) Endpoint {
+	if !strings.HasSuffix(path, ".git") {
+		path = fmt.Sprintf("%s/.git", path)
+	}
+	return Endpoint{Url: fmt.Sprintf("file://%s", path)}
+}
--- a/lfsapi/endpoint_finder.go
+++ b/lfsapi/endpoint_finder.go
@@ -170,6 +170,9 @@ func (e *endpointGitFinder) NewEndpointFromCloneURL(rawurl string) Endpoint {
 
 func (e *endpointGitFinder) NewEndpoint(rawurl string) Endpoint {
 	rawurl = e.ReplaceUrlAlias(rawurl)
+	if strings.HasPrefix(rawurl, "/") {
+		return endpointFromLocalPath(rawurl)
+	}
 	u, err := url.Parse(rawurl)
 	if err != nil {
 		return endpointFromBareSshUrl(rawurl)
--- a/lfsapi/client.go
+++ b/lfsapi/client.go
@@ -27,7 +27,18 @@ var (
 	httpRE    = regexp.MustCompile(`\Ahttps?://`)
 )
 
+var hintFileUrl = strings.TrimSpace(`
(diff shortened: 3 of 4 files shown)


lfsapi/endpoint.go
```diff
@@ -106,3 +107,10 @@ func endpointFromGitUrl(u *url.URL, e *endpointGitFinder) Endpoint {
 	u.Scheme = e.gitProtocol
 	return Endpoint{Url: u.String()}
 }
+
+func endpointFromLocalPath(path string) Endpoint {
+	if !strings.HasSuffix(path, ".git") {
+		path = slashpath.Join(path, ".git")
```
ttaylorr: Use of `path.Join` is not recommended within Git LFS. I would recommend instead:

```go
if !strings.HasSuffix(path, ".git") {
        path = fmt.Sprintf("%s/.git", path)
}

// ...
```

sprohaska:
I'm currently a bit busy. I'll reply later. Sorry.

ttaylorr:
> Sorry.

Please don't be! Looking forward to your response later. Thanks!

sprohaska:
@ttaylorr, I've amended the commit and provided an example implementation.

The `file://` scheme is required for Git config URL match to work.  An
absolute path alone cannot be used for URL matching.

An example implementation of a transfer agent is available as Gist
<https://gist.github.com/sprohaska/f3cd4287d8c41b74fc808b5a5396cad1>.

lfsapi/endpoint_finder.go
```diff
@@ -170,6 +170,9 @@ func (e *endpointGitFinder) NewEndpointFromCloneURL(rawurl string) Endpoint {
 
 func (e *endpointGitFinder) NewEndpoint(rawurl string) Endpoint {
 	rawurl = e.ReplaceUrlAlias(rawurl)
+	if strings.HasPrefix(rawurl, "/") {
+		return endpointFromLocalPath(rawurl)
```
ttaylorr: @sprohaska thanks for your clarification in https://github.com/git-lfs/git-lfs/pull/2912#issuecomment-377924331, it was very helpful for my understanding of this change. I have one remaining concern pertaining to this line. Since Git LFS does not support natively pushing or pulling without a server, I would like the user to be warned of this.

What do you think about warning when there are no custom transfer agents enabled when we're in this conditional? 

Something like:

```ShellSession
$ git lfs push
warning: Git LFS does not support pushing to local files without
warning: a custom transfer agent. Please see <url> for more.
# ...
```

sprohaska: A warning makes sense. But I'm not sure whether this is the right place to print it. Maybe the warning could be printed by the code that decides whether to use a standalone agent or contact a server. If the decision there is to contact a server for a `file://` URL, a warning would be printed instead of trying to connect to it. I'll look into it in more detail later.

sprohaska: I've added a commit that adds a hint about file:// and standalone transfer agent.

angkanasoisom:
tumzaza2018@gmail.com

ในวันที่ จ. 2 เม.ย. 2018 20:01 sprohaska <notifications@github.com>
เขียนว่า:

sprohaska:
@ttaylorr I've added a commit that adds a hint about file:// and standalone transfer agent a few days ago.

Issue #2766, opened 2017-12-02
Mapping of old sha1 to new sha1


hkrpavan:
After running git-lfs-migrate how to get mapping of old sha1 to new sha1 of all commits?

Also, what is the file git-lfs-migrate.mapdb? If it has old commit id to new commit id mapping how to dump it to text or csv file?

ttaylorr:
Hi @hkrpavan, thanks for opening this issue, and good question. I can't off the top of my head think of any intuitive ways to find an old<->new SHA-1 mapping for Git objects, but your issue has got me thinking. Beyond tags, I think that it would be useful to retain this information in some format or another.

I propose adding `--object-map=<file>` to the `git-lfs-migrate(1)` 'import' mode, which will create a CSV at the given path, containing the following:

```
previous-oid,current-oid,object-type,previous-oid-type,current-oid-type
# ...
```

The migrate command already shows _reference_ OID changes, but not all of the objects in between. I can imagine scripting with this additional information would be helpful.

/cc @git-lfs/core what do you think?

larsxschneider:
The CSV file sounds like a nice, unintrusive, approach.

Alternatively, we could extend the commit message [similar to cherry-picking](https://github.com/git/git/blob/35f6318d44379452d8d33e880d8df0267b4a0cd0/Documentation/git-cherry-pick.txt#L62-L64) (e.g. append `(Git LFS import from commit ABC123)` at the end of a each commit.

technoweenie:
I like the idea of annotating the commits, since they're being rewritten anyway. A coworker brought up [the trailer format](https://git-scm.com/docs/git-interpret-trailers), which is an existing convention of adding metadata to a commit:

```
Update PSD file.

Signed-off-by: Super Artist <super.artist@example.com>
Migrate-lfs: abc12345656
```

ttaylorr:
@larsxschneider @technoweenie I love these ideas. I think that we could split the difference and have `git-lfs-migrate(1)` accept a `--commit-msg` Go template option when running in 'import' mode.

I pushed a WIP this here: https://github.com/git-lfs/git-lfs/compare/migrate-commit-msg-tmpl.

```
~/src/repository (master) $ git lfs migrate import --everything --commit-msg="{{ original.Message }}\nMigrate-lfs: {{ original.SHA1 }}"
```

technoweenie:
A template like that doesn't quite work. Git trailers should be separated from the commit message by a line break, but can be right next to each other.

[...]

PR #2929, opened 2018-03-19
Add object-map option to get a map from the old sha1 to new sha1

jasminlapalme:
This pull request fixes partially #2766, since it does get a map file of only the commits that have been rewritten. I think that was the original request of this issue.


ttaylorr:
EDIT: on further inspection, could you fix some of these test failures (+ add documentation in `docs/main/git-lfs-migrate.1.ronn` first?)

jasminlapalme:
Sure, I will look at that.

jasminlapalme:
@ttaylorr I have fixed the tests and add documentation.

docs/man/git-lfs-migrate.1.ronn
```diff
@@ -80,6 +80,11 @@ options and these additional ones:
 * `--verbose`
     Print the commit oid and filename of migrated files to STDOUT.
 
+* `--object-map=<path>`
+    Write to 'path' a file with the mapping of each rewritten commits. The file
+    format is CSV with this pattern:
+      * OLD-SHA,NEW-SHA
```
ttaylorr: Can the `*` be removed? Since this is not part of the file format, I think this character has the potential to cause confusion for users trying to understand the file format of the `--object-map` flag.

git/githistory/rewriter.go
```diff
@@ -253,6 +267,13 @@ func (r *Rewriter) Rewrite(opt *RewriteOptions) ([]byte, error) {
 			if err != nil {
 				return nil, err
 			}
+			if len(opt.ObjectMapFilePath) > 0 {
```
ttaylorr: Could this conditional be changed to:

```go
if objectMapFile != nil {
        // ...
}
```

This way, we:

  1. Minimize referencing the options type, which has a tendency to pollute the surrounding code-base, and

  2. Have a stronger guarantee that this block will not attempt to dereference a nil-pointer, should it become the case that `len(opt.ObjectMapFilePath) > 0 =/=> objectMapFile != nil`.

git/githistory/rewriter.go
```diff
@@ -253,6 +267,13 @@ func (r *Rewriter) Rewrite(opt *RewriteOptions) ([]byte, error) {
...
+				mapStr := fmt.Sprintf("%s,%s\n", hex.EncodeToString(oid), hex.EncodeToString(newSha))
```
ttaylorr: Could this line be changed to the following:

```go
if _, err := fmt.Fprintf("%x,%x\n", oid, newSha); err != nil {
        return nil, err
}
```

for brevity?

docs/man/git-lfs-migrate.1.ronn
jasminlapalme: I have reformulated to be on the same line

git/githistory/rewriter.go
jasminlapalme: Very good points, I have done the change

git/githistory/rewriter.go
jasminlapalme: I have change it

git/githistory/rewriter.go
```diff
@@ -188,6 +193,15 @@ func (r *Rewriter) Rewrite(opt *RewriteOptions) ([]byte, error) {
 		vPerc = perc
 	}
 
+	var objectMapFile *os.File
+	if len(opt.ObjectMapFilePath) > 0 {
+		objectMapFile, err = os.Create(opt.ObjectMapFilePath)
```
ttaylorr: Should we pass the `O_EXCL` mode here? Doing so would cause the program to error when a file already exists in that location, instead of writing over it.

```go
f, err := os.OpenFile(opt.ObjectMapFilePath, os.O_RDWR|os.O_CREATE|os.O_EXCL, 0666)
if err != nil {
        // ..
}
```

git/githistory/rewriter.go
```diff
@@ -253,6 +267,12 @@ func (r *Rewriter) Rewrite(opt *RewriteOptions) ([]byte, error) {
 			if err != nil {
 				return nil, err
 			}
+			if objectMapFile != nil {
+				_, err := fmt.Fprintf(objectMapFile, "%s,%s\n", hex.EncodeToString(oid), hex.EncodeToString(newSha))
```
ttaylorr: This line can be shortened by using the `%x` formatting verb, which performs the `hex.EncodeToString` call transparently:

```go
_, err := fmt.Fprintf(objectMapFile, "%x,%x\n", oid, newSha))
```

is equivalent to:

```go
_, err := fmt.Fprintf(objectMapFile, "%s,%s\n", hex.EncodeToString(oid), hex.EncodeToString(newSha))
```

##

Since the line is shorter, do you think we can compact it even further by:

```
if _, err := fmt.Fprintf(objectMapFile, "%x,%x\n", oid, newSha)); err != nil {
        return nil, err
}
```

git/githistory/rewriter.go
jasminlapalme: Yes, I think that's safer.

git/githistory/rewriter.go
jasminlapalme: I have made the change.

Issue #2908, opened 2018-03-09
Migrate Import doesn't handle spaces in paths properly


mrackwitz:
I passed in some quoted paths via `--include`, e.g. `"External/AWS iOS SDK/AWSRuntime.framework/Versions/A/AWSRuntime"`.
This resulted them being written to the `.gitattributes` as:

```
External/AWS iOS SDK/AWSRuntime.framework/Versions/A/AWSRuntime filter=lfs diff=lfs merge=lfs -text
```

Which resulted in the following errors when checking out files:

```
SDK/AWSRuntime.framework/Versions/A/AWSRuntime is not a valid attribute name: .gitattributes:24
```

From issues in this repo, it seems like spaces should be represented as `[[:space:]]` in the `.gitattributes` file. `git lfs migrate import` should take this into account.

ttaylorr:
Hi @mrackwitz! Thanks for pointing this out; interestingly enough, `git-lfs-track(1)` and `git-lfs-untrack(1)` both support this kind of conversion, but `git-lfs-migrate-import(1)` does not. I have opened up https://github.com/git-lfs/git-lfs/pull/2933 to fix this, and have queued it for the next release.

mrackwitz:
Thanks so much for looking into this!

PR #2933, opened 2018-03-21
commands/migrate: escape paths before .gitattributes 

ttaylorr:
This pull request fixes `git-lfs-migrate-import(1)` when given `--include` and/or `--exclude` filters that contain .gitattributes-unescaped sequences, such as whitespace.

When migrating, Git LFS will do either one of the following two things: (1) adopt filters from `--include`, `--exclude`, or (2) migrate all files that aren't named `.gitattributes`. When doing (1), Git LFS previously copied these filters verbatim as they were given on the command line.

This is problematic, as @mrackwitz points out in https://github.com/git-lfs/git-lfs/issues/2908:

> I passed in some quoted paths via --include, e.g. `"External/AWS iOS SDK/*"`.
> This resulted them being written to the `.gitattributes` as:
>
> ```
> External/AWS iOS SDK/* filter=lfs diff=lfs merge=lfs -text
> ```
>
> Which resulted in the following errors when checking out files:
>
> ``` 
> * is not a valid attribute name: .gitattributes:24
> ```

We already do a similar transformation in `git-lfs-track(1)` and `git-lfs-untrack(1)` via functions previously known as `escapeTrackPattern` and `unescapeTrackPattern`, respectively. Let's repurpose those functions here to avoid creating malformed `.gitattributes` file(s).

Here's how the change occurred:

1. 2409b3e: Rename the aforementioned functions to a more generally named `convertAttrPattern` (and its counterpart). "track" patterns are a misnomer for `.gitattribute` patterns, which is what they really are.
2. 89c5fa0: Now that those functions are named appropriately, use them in `git-lfs-migrate(1)`.
3. 11199f0, 5704057: add a fixture and test case to ensure that the above functionality works as expected.

This transformation is safe to apply, since it is done so once, and has a one-to-one correspondence.

Closes: https://github.com/git-lfs/git-lfs/issues/2908

##

/cc @git-lfs/core @mrackwitz

### DIFF
--- a/commands/command_migrate_import.go
+++ b/commands/command_migrate_import.go
@@ -123,11 +123,11 @@ func trackedFromFilter(filter *filepathfilter.Filter) *tools.OrderedSet {
 	tracked := tools.NewOrderedSet()
 
 	for _, include := range filter.Include() {
-		tracked.Add(fmt.Sprintf("%s filter=lfs diff=lfs merge=lfs -text", include))
+		tracked.Add(fmt.Sprintf("%s filter=lfs diff=lfs merge=lfs -text", escapeAttrPattern(include)))
 	}
 
 	for _, exclude := range filter.Exclude() {
-		tracked.Add(fmt.Sprintf("%s text -filter -merge -diff", exclude))
+		tracked.Add(fmt.Sprintf("%s text -filter -merge -diff", escapeAttrPattern(exclude)))
 	}
 
 	return tracked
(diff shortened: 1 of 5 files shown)


ttaylorr:
@peff: would it be more appropriate to double-quote these patterns instead? I think that both have the same effect, so I am not sure which is preferred.

peff:
@ttaylor I think either is fine. Double-quoting might be a little bit more readable for a human who has to touch them later, but I'm not sure it's a big deal either way.

PR #2934, opened 2018-03-21
Retry SSH resolution 5 times

### LINKED ISSUE #2914: Add retry option if SSH auth fails
Does it make sense to add retry logic in case the SSH auth fails here? https://github.com/git-lfs/git-lfs/blob/5d18f87f009b1f96d484d216868858083cdd0a13/tq/api.go#L60-L63 I see that there is retry logic here: https://github.com/git-lfs/git-lfs/blob/5d18f87f009b1f96d484d216868858083cdd0a13/tq/api.go#L68 We've seen cases where intermittent SSH failures cause the whole repo download to stop. It'd be nice to be able to specify a retry count.

stanhu:
Closes #2914


lfsapi/client.go
```diff
@@ -23,20 +23,37 @@ import (
 const MediaType = "application/vnd.git-lfs+json; charset=utf-8"
 
 var (
-	UserAgent = "git-lfs"
-	httpRE    = regexp.MustCompile(`\Ahttps?://`)
+	UserAgent                 = "git-lfs"
+	httpRE                    = regexp.MustCompile(`\Ahttps?://`)
+	maxSshAuthenticateRetries = 5
```
ttaylorr: Do you think that this could be an instance variable belonging to `*lfsapi.Client`? I don't think that this has to be in the package-global namespace, so I'd advocate for the former.

lfsapi/client.go
```diff
@@ -23,20 +23,37 @@ import (
...
+	var err error
+
+	requests := tools.MaxInt(0, maxSshAuthenticateRetries) + 1
+	for i := 0; i < requests; i++ {
+		sshRes, err = c.SSH.Resolve(e, method)
+		if err == nil {
+			return sshRes, nil
+		}
+
 		tracerx.Printf("ssh: %s failed, error: %s, message: %s",
```
ttaylorr: Should these messages be changed to indicate when we are and aren't retrying? You can take a look at something similar, like this:

https://github.com/git-lfs/git-lfs/blob/8c8a711404391758565a946fdc36e8249344b1f1/tq/transfer_queue.go#L493

lfsapi/client.go
```diff
@@ -23,20 +23,37 @@ import (
...
-	httpRE    = regexp.MustCompile(`\Ahttps?://`)
+	UserAgent                 = "git-lfs"
+	httpRE                    = regexp.MustCompile(`\Ahttps?://`)
+	maxSshAuthenticateRetries = 5
 )
 
-func (c *Client) NewRequest(method string, e Endpoint, suffix string, body interface{}) (*http.Request, error) {
-	sshRes, err := c.SSH.Resolve(e, method)
-	if err != nil {
+func (c *Client) sshResolveWithRetries(e Endpoint, method string) (sshAuthResponse, error) {
```
ttaylorr: What do you think about changing the return type of this method to be:

```go
(*sshAuthResponse, error)
```

instead of

```go
(sshAuthResponse, error)
```

that way we can return `nil` to indicate a failure or missing `sshAuthResponse`, instead of `sshAuthResponse{}`, which could be misleading or confusing to callers of this method.

lfsapi/client.go
```diff
@@ -23,20 +23,37 @@ import (
...
+	var sshRes sshAuthResponse
+	var err error
+
+	requests := tools.MaxInt(0, maxSshAuthenticateRetries) + 1
+	for i := 0; i < requests; i++ {
+		sshRes, err = c.SSH.Resolve(e, method)
+		if err == nil {
+			return sshRes, nil
```
ttaylorr: Per my suggestion above, this line would change to:

```go
return &sshRes, nil
```

lfsapi/client.go
```diff
@@ -23,20 +23,37 @@ import (
...
 		tracerx.Printf("ssh: %s failed, error: %s, message: %s",
 			e.SshUserAndHost, err.Error(), sshRes.Message,
 		)
+	}
 
-		if len(sshRes.Message) > 0 {
-			return nil, errors.Wrap(err, sshRes.Message)
-		}
+	if len(sshRes.Message) > 0 {
+		return sshAuthResponse{}, errors.Wrap(err, sshRes.Message)
```
ttaylorr: And this, (as well as the line below it, L51) would change to:

```go
return nil, err
```

ttaylorr:
Hi @stanhu, I hope that it's OK I updated your branch myself to include feedback that I gave above and changes from master.

stanhu:
@ttaylorr Thanks!

PR #2939, opened 2018-03-23
test: ensure that git-mergetool(1) works with large files

ttaylorr:
This pull request ensures that the `git-mergetool(1)` program works correctly with large files stored using Git LFS.

This is the result of spending some time doing research into and implementing a `git-lfs-mergetool(1)`, after thinking that `git-mergetool(1)` would not convert conflicts using their appropriate clean/smudge filters before sending to the designated mergetool.

(The result of that work can be viewed in [branch `mergetool`](https://github.com/git-lfs/git-lfs/compare/mergetool)).

While discussing with @peff, we noted that `git-mergetool(1)` uses `git-checkout-index(1)` to get the contents of a file before assigning it to a temporary location and attaching an environment variable to the mergetool program. For reference, that's:

- https://github.com/git/git/blob/90bbd502d54fe920356fa9278055dc9c9bfe9a56/git-mergetool.sh#L304-L306, and
- https://github.com/git/git/blob/90bbd502d54fe920356fa9278055dc9c9bfe9a56/git-mergetool.sh#L230-L241, and
- https://github.com/git/git/blob/90bbd502d54fe920356fa9278055dc9c9bfe9a56/git-mergetool.sh#L275-L278.

Since `git-checkout-index(1)` performs filter operations on files before returning them, `git-mergetool(1)` should do the same. This test ensures that this is the case. Thus, `git-mergetool(1)` is sufficient to work with conflicting large diff-able blobs stored in Git LFS.

##

/cc @git-lfs/core, esp. @larsxschneider 
/cc @shana 
/cc @git-lfs/core

### DIFF
--- a/test/test-mergetool.sh
+++ b/test/test-mergetool.sh
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+
+. "test/testlib.sh"
+
+begin_test "mergetool works with large files"
+(
+  set -e
+
+  reponame="mergetool-works-with-large-files"
+  git init "$reponame"
+  cd "$reponame"
+
+  git lfs track "*.dat"
+  printf "base" > conflict.dat
+  git add .gitattributes conflict.dat
+  git commit -m "initial commit"
+
+  git checkout -b conflict
+  printf "b" > conflict.dat
+  git add conflict.dat
+  git commit -m "conflict.dat: b"
+
+  git checkout master
+
+  printf "a" > conflict.dat
+  git add conflict.dat
+  git commit -m "conflict.dat: a"
+
+  set +e
+  git merge conflict
+  set -e
+
+  git config mergetool.inspect.cmd '
+    for i in BASE LOCAL REMOTE; do
+      echo "\$$i=$(eval "cat \"\$$i\"")";
+    done;
+    exit 1
+  '
+  git config mergetool.inspect.trustExitCode true
+
+  yes | git mergetool \
+      --no-prompt \
+      --tool=inspect \
+      -- conflict.dat 2>&1 \
+    | tee mergetool.log
+
+  grep "\$BASE=base" mergetool.log
+  grep "\$LOCAL=a" mergetool.log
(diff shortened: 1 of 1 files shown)


peff:
Right, that all matches my experimentation. But was there some group who uses mergetool who was having trouble with LFS-tracked files? Perhaps I'm missing their use case / troubles.

shana:
@peff In game development, large files are not necessarily binary files, they can be very large text files - hundreds or gigs in size. There are tools that can perfectly merge changes in these files, but they're too large to be in git.

peff:
@shana I get the use-case. But what I don't get is what the problem is. As far as I can tell from these tests, Git would present the mergetool with the smudged (i.e., real) file content, which should be what the user wants.

But my understanding is that people are having trouble using things like UnityYAMLMerge with LFS. What are they doing differetly than what we're seeing here?

shana:
@peff Oh, I see what you're asking. When I tested this workflow last year, it didn't work, which is when we had the conversation of making sure lfs supported merge tools. However, this was around the time we open sourced GitHub for Unity, so around June 2017, and I never revisited the issue until now, so it's very possible that my assumptions around support for this are no longer valid. I'm going to run tests with all the merge tools and see if everything works as it's supposed and document it accordingly so we know if there's any edge case that we might be missing.

ttaylorr:
> I'm going to run tests with all the merge tools and see if everything works as it's supposed and document it accordingly so we know if there's any edge case that we might be missing.

Thanks. I'll wait on merging this until you confirm that `git-mergetool(1)` work as-expected with files tracked in Git LFS.

peff:
@shana Thanks. If you find any confusing behavior on the Git side, let me know. I'm happy to dig in to the root cause.

StanleyGoldman:
I'm testing this scenario and it seems to work correctly for me.

shana:
Thanks for the patience while we tested this! ❤️

PR #2963, opened 2018-04-10
Functionality to filter git lfs checkout with wildcards

### LINKED ISSUE #2952: git lfs checkout my_branch with -X "*.*"?
Hi, I'm just getting started with git lfs. It's pretty great, so far! Is there an option for git lfs checkout that avoids automatically pulling/fetching an LFS files?

My ultimate goal is to be able to clone and checkout branches without automatically downloading lfs files. So far, I can clone without the automatic download using the option `-X "*.*"`

When I ran a couple of tests, I was able to git lfs track a test file in a test branch. When I cloned the repo on another machine (with git lfs already installed), things seemed fine and dandy, no files automatically downloaded. When I used `git checkout` for my branch, the lfs test file automatically downloaded. I checked the [man page](https://github.com/git-lfs/git-lfs/blob/master/docs/man/git-lfs-checkout.1.ronn) for `git lfs checkout`, but I didn't see the option that I was hoping to find.

Is the only path forward to either use skip-smudge locally or edit `.lfsconfig`? It would be great if I could accomplish this via command line - it seems that running a filter through `.lfsconfig` isn't ideal because we want lfs files downloaded for specific cases; it's just rare that we need to download from LFS. I tried a wildcard filter in `.lfsconfig`, but there was a scenario where it blocked the fetch/pull of a file that I did want.

`git version 2.7.4`
`git-lfs/2.4.0 (GitHub; linux amd64; go 1.8.3)`
edited the output of `git lfs env`:
```
git-lfs/2.4.0 (GitHub; linux amd64; go 1.8.3)
git version 2.7.4

ConcurrentTransfers=3
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin

AccessDownload=none
AccessUpload=none
DownloadTransfers=basic
UploadTransfers=basic

### PR DESCRIPTION
**Pull request by StanleyGoldman**
Fixes #2952 

After reading the contents of `command_fetch.go` and `command_clone.go` this seems to be the way to implement this functionality.

### CODE REVIEW

File: commands/command_checkout.go [hunk 1]
```diff
@@ -43,7 +43,12 @@ func checkoutCommand(cmd *cobra.Command, args []string) {
 		pointers = append(pointers, p)
 	})
 
-	chgitscanner.Filter = filepathfilter.New(rootedPaths(args), nil)
+	include, exclude := getIncludeExcludeArgs(cmd)
+	if include != nil || exclude != nil {
```
ttaylorr: What do you think should happen in the following case?

```ShellSession
$ git lfs checkout --include=foo.bar baz.quux
```

Right now, this change will silently ignore `bar.quux`, and instead build a filter from the `--include` flag only. I am not opposed to this, but I think that there are a few options we should consider first:

1. __Ignore arguments when either `--include` or `--exclude` is given__. This seems like an OK option to me, but I think that we should `Exit()` when arguments are given in this case, something like:

    ```ShellSession
    $ git lfs checkout --include=foo.bar baz.quux
    fatal: arguments not compatible with --include, --exclude
    ```

2. __Ignore `--include` and `--exclude` when arguments are given__. This is a similarly OK option for me, but I would like to do the same sort of warning as above in this case.

3. __Combine command-line arguments with `--include`, `--exclude`__. This is my preferred option. I would like to make the following two commands synonyms of each other:

    ```ShellSession
    $ git lfs checkout --include=foo.bar baz.quux
    $ git lfs checkout --include=foo.bar;baz.quux
    ```

    This way, we don't have asymmetric clobbering behavior (i.e., in the case where one style is ignored but the other isn't), and we don't silently do something unexpected when users have `lfs.fetch{in,ex}clude` set in their `.gitconfig`'s.

I think that option (3) could be accomplished via something like:

```go
include, exclude := getIncludeExcludeArgs(cmd)
include = append(include, rootedPaths(args)...)
chgitscanner.Filter = filepathfilter.New(include, exclude)
```

(and ideally, the first two lines of that change would go into a function somewhere).

StanleyGoldman: After I can get tests going, I'll come back to this comment to try to understand it again. 🤣

StanleyGoldman:
Hey @ttaylorr I've started with the tests... I think my first two are working as expected.
I just wanted to check in with you and get your thoughts before I got too far.

test/test-checkout.sh
```diff
@@ -155,3 +155,94 @@ begin_test "checkout: outside git repository"
...
+  printf "$contents" > file2.dat
+  printf "$contents" > file3.dat
+  mkdir folder1 folder2
+  printf "$contents" > folder1/nested.dat
+  printf "$contents" > folder2/nested.dat
+  git add file1.dat file2.dat file3.dat folder1/nested.dat folder2/nested.dat
+  git add .gitattributes
+  git commit -m "add files"
+
+  [ "$contents" = "$(cat file1.dat)" ]
```
ttaylorr: I think that we can safely get rid of these `test`s. They're showing that we (1) wrote to the right files, and (2) that the files were checked out correctly to begin with.

test/test-checkout.sh
```diff
@@ -155,3 +155,94 @@ begin_test "checkout: outside git repository"
...
+  [ "$contents" = "$(cat file2.dat)" ]
+  [ "$contents" = "$(cat file3.dat)" ]
+  [ "$contents" = "$(cat folder1/nested.dat)" ]
+  [ "$contents" = "$(cat folder2/nested.dat)" ]
+
+  # Remove the working directory
+  rm -rf file1.dat file2.dat file3.dat folder1/nested.dat folder2/nested.dat
```
ttaylorr: I'm having a little bit of trouble with this line. The normal behavior for `git lfs checkout` is to smudge pointers in the working copy to turn them into large objects. While `git {,lfs} checkout` will create files if they don't exist, I don't think that this is the approach we should use for testing.

Instead, what about something like this:

```sh
begin_test "checkout: setup for --include --exclude"
(
  set -e
  reponame="checkout-include-exclude"
  setup_remote_repo "$reponame"
  clone_repo "$reponame"

  # Initialize exactly as L168-184.
)
end_test

begin_test "checkout: with include"
(
  set -e

  cd "$TRASHDIR"
  # Manually clone without downloading any objects.
  git -c "lfs.fetchexclude=*" clone "$GITSERVER/checkout-include-exclude"
  cd checkout-include-exclude

  git config credential.helper lfstest

  # Test as in line L196-197
)
end_test
```

This has the benefit of moving repeated initialization into one test, and shortening the behavior tests.

test/test-checkout.sh
```diff
@@ -155,3 +155,94 @@ begin_test "checkout: outside git repository"
...
+
+  echo "checkout should replace only file1.dat"
```
ttaylorr: This style of `echo`-ing during a test is used elsewhere in `test/test-checkout.sh`, but is not a common idiom elsewhere. Let's remove that, too. This test is testing one thing, and the test name documents it well.

test/test-checkout.sh
```diff
@@ -155,3 +155,94 @@ begin_test "checkout: outside git repository"
...
+(
+  set -e
+
+  reponame="checkout-with-include"
+  setup_remote_repo "$reponame"
+
+  clone_repo "$reponame" repo2
+
+  git lfs track "*.dat" 2>&1 | tee track.log
+  grep "Tracking \"\*.dat\"" track.log
```
ttaylorr: This pattern is also common in `test/test-checkout.sh`, but is not necessary and no longer conforms to our style. `git lfs track` alone (without `grep(1)` afterwords) is fine.

test/test-checkout.sh
```diff
@@ -155,3 +155,94 @@ begin_test "checkout: outside git repository"
...
+  [ "$contents" = "$(cat folder1/nested.dat)" ]
+  [ "$contents" = "$(cat folder2/nested.dat)" ]
+
+  # Remove the working directory
+  rm -rf file1.dat file2.dat file3.dat folder1/nested.dat folder2/nested.dat
+
+  echo "checkout should replace only file1.dat"
+  git lfs checkout --include file1.dat
+  [ "$contents" = "$(cat file1.dat)" ]
+  [ ! -f file2.dat ]
```
ttaylorr: If you adopt my suggestion of sharing the initialization sequence, you should replace `[ ! -f xxx.dat ]` with `assert_pointer "master" "xxx.dat" "$oid" "$size"`.

StanleyGoldman:
I'm so glad I asked...

test/test-checkout.sh
Zainal1: Akaun

StanleyGoldman:
Hey @ttaylorr I haven't forgotten about this. I plan on finishing soon, just been a bit crazy on my end.

ttaylorr:
Hey @StanleyGoldman, I just opened #3056, which will deprecate `git lfs checkout` in v3.0.0, since it is just as fast to use `git checkout` with #2466. I apologize for not figuring this out earlier, it had slipped my mind that we'd be deprecating this command when you started working on this.

I think that since we will eventually deprecate the command, it makes sense to close this for now.

StanleyGoldman:
No problem, sorry I was so slow on this.

Issue #2975, opened 2018-04-18
`git lfs uninstall --local` gives confusing message


hraban:
```
$ git lfs uninstall --local
Hooks for this repository have been removed.
Global Git LFS configuration has been removed.
```

First, the second line is not what I expected, given the docs:

```
* Remove the "lfs" clean and smudge filters from the global Git config.
* Uninstall the Git LFS pre-push hook if run from inside a Git repository.

Options:

* --local:
    Removes the "lfs" smudge and clean filters from the local repository's git
    config, instead of the global git config (~/.gitconfig).
```

Additionally, it doesn't seem on the face of it like anything was actually changed globally. I can't tell for sure, but a fresh git clone of a repo with lfs did start downloading lfs files, so that seems ok?

```
$ git lfs version
git-lfs/2.4.0 (GitHub; darwin amd64; go 1.10)
```

Cheers

PR #2976, opened 2018-04-18
commands/uninstall: do not log about global hooks with --local

ttaylorr:
This pull request fixes a bug where `git lfs uninstall --local` would produce output suggesting that it had uninstall global hooks when in fact it had not.

For example (as pointed out by @hraban in https://github.com/git-lfs/git-lfs/issues/2975):

```ShellSession
$ git lfs uninstall --local
Hooks for this repository have been removed.
Global Git LFS configuration has been removed.
```

To fix this, ensure that we're not running in `--local` mode before logging about the global Git LFS configuration. We do not treat `--system` as a special-case here, since it is not used by this command. (A future pull request will remove this from `git lfs uninstall`).

Closes: https://github.com/git-lfs/git-lfs/issues/2975.

##

/cc @git-lfs/core

### DIFF
--- a/commands/command_uninstall.go
+++ b/commands/command_uninstall.go
@@ -14,7 +14,9 @@ func uninstallCommand(cmd *cobra.Command, args []string) {
 		uninstallHooksCommand(cmd, args)
 	}
 
-	Print("Global Git LFS configuration has been removed.")
+	if !localInstall {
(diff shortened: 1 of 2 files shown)


751176723:
http://8828ddd.com/GR/6lz88v644v

Issue #2974, opened 2018-04-18
Random occurrences of "not a tree object" on fetch


dalemyers:
As part of our builds running on VSTS, the first thing it does is to clone our repo and checkout the commit we are building. Sometimes though, a few times in a hundred, we get a failure like this:

```
2018-04-04T09:05:06.4938770Z ##[command]git -c http.extraheader="AUTHORIZATION: bearer ********" fetch --tags --prune --progress --no-recurse-submodules origin
2018-04-04T09:05:06.9741870Z From https://[redacted].visualstudio.com/[redacted]/_git/[redacted]
2018-04-04T09:05:06.9756220Z  - [deleted]               (none)     -> origin/[redacted]
2018-04-04T09:05:07.2478720Z remote: Microsoft (R) Visual Studio (R) Team Services        
2018-04-04T09:05:07.3977900Z remote: 
2018-04-04T09:05:07.3994020Z remote: Found 14 objects to send. (150 ms)        
2018-04-04T09:05:07.6351740Z    e7989e48b8..8ec149e468  [redacted] -> origin/[redacted]
2018-04-04T09:05:07.6367720Z    9f00bb1ce0..caf9ab3dd0  develop             -> origin/develop
2018-04-04T09:05:07.7022600Z ##[command]git -c http.https://[redacted].visualstudio.com.extraheader="AUTHORIZATION: bearer ********" lfs fetch origin e0b225c0e5c2747b2ab4270238ef783113fcd673
2018-04-04T09:05:07.7452330Z fetch: Fetching reference e0b225c0e5c2747b2ab4270238ef783113fcd673
2018-04-04T09:05:07.7549930Z Could not scan for Git LFS files
2018-04-04T09:05:07.7618340Z 
2018-04-04T09:05:07.7635280Z Errors logged to /Users/[redacted]/vsts_agents/vsts_agent_2/_work/9/s/.git/lfs/logs/20180404T090507.754105495.log
2018-04-04T09:05:07.7649040Z Use `git lfs logs last` to view the log.
2018-04-04T09:05:07.7736300Z ##[command]git lfs logs last
2018-04-04T09:05:07.8064290Z git-lfs/2.4.0 (GitHub; darwin amd64; go 1.10)
2018-04-04T09:05:07.8078950Z git version 2.16.2
2018-04-04T09:05:07.8085600Z 
[...]

PR #2982, opened 2018-04-25
git: include SHA1 in ref-less revisions

ttaylorr:
This pull request _likely_ fixes an issue pointed out by @dalemyers in https://github.com/git-lfs/git-lfs/issues/2974:

> As part of our builds running on VSTS, the first thing it does is to clone our repo and checkout the commit we are building. Sometimes though, a few times in a hundred, we get a failure like this:
> 
> ```
> 2018-04-04T09:05:07.8098800Z $ git-lfs fetch origin e0b225c0e5c2747b2ab4270238ef783113fcd673
> 2018-04-04T09:05:07.8112210Z fetch: Fetching reference e0b225c0e5c2747b2ab4270238ef783113fcd673
> 2018-04-04T09:05:07.8126000Z Could not scan for Git LFS files
> 2018-04-04T09:05:07.8132650Z 
> 2018-04-04T09:05:07.8145800Z Error in git ls-tree: exit status 128 fatal: not a tree object
> ```
> There appears to be no pattern in the commits that fail. Retrying basically always works. It's just a weird intermittent issue.

The issue, I believe is related to invocations of `git-lfs-fetch(1)` that use a sha1 instead of a qualified reference name. For example:

```ShellSession
$ git lfs fetch origin 987c137
# ...
```

Code later in the command finds the attached `ref` by calling `git.ResolveRefs()`, and then uses `ref.Sha` throughout:

https://github.com/git-lfs/git-lfs/blob/1a98fa5c4065e0b692c181ddc4551d21c72945e2/commands/command_fetch.go#L87

all the way down (via `fetchRef()`, `pointerstoFetchForRef()`, `lfs.GitScanner.ScanTree()`, `lfs.runScanTree()`, `lfs.lsTreeBlobs()`) to `git.LsTree()`:

https://github.com/git-lfs/git-lfs/blob/1a98fa5c4065e0b692c181ddc4551d21c72945e2/git/git.go#L223-L232

If `ref.Sha` is empty, Git will not be able to scan it:

```ShellSession
$ git ls-tree -r -l -z --full-tree ""
fatal: Not a valid object name
```

To ensure that Git LFS faithfully reports the Sha1 to scan, assign it when we parse a non-symbolic "reference". (While looking at this code, I noticed that the terminology in this part of the project is incorrect, but that will be for a later change).

##

/cc @git-lfs/core


git/git.go
```diff
@@ -245,7 +245,8 @@ func ResolveRef(ref string) (*Ref, error) {
 
 	if len(lines) == 1 {
 		// ref is a sha1 and has no symbolic-full-name
-		fullref.Name = lines[0] // fullref.Sha
+		fullref.Name = lines[0]
+		fullref.Sha = lines[0]
```
larsxschneider: That looks sensible. I wonder, if it would make sense to check here if `lines[0]` is not empty and/or contains a 40 char string? However, in a quick check I wasn't able to make `rev-parse` return an empty result with positive error code (I guess this is not possible).

larsxschneider: The only thing that could cause confusion in the future is that `fullref.Name` is a hash. But that has been there before. So we are not worse than before I guess 😄

ttaylorr:
@larsxschneider thanks!

PR #2983, opened 2018-04-25
Do not migrate the symlinks to LFS objects.

patrickmarlier:


patrickmarlier:
Hi @ttaylorr ,

It is good to be picky. I created a new test and new fixture for the symlink. Unfortunately the Windows CI failed. I would imagine due to the symlink that is not a symlink in the Windows CI. Excerpt of the log:
"create mode 100644 link.txt"
So here I can propose to skip the test in case of Windows or find why the symlink is created as a regular file. In both case, some help is welcome because it would take time to test on Windows.

ttaylorr:
> It is good to be picky. I created a new test and new fixture for the symlink.

Thank you :smile:.

> So here I can propose to skip the test in case of Windows or find why the symlink is created as a regular file. In both case, some help is welcome because it would take time to test on Windows.

It can be a little tricky to use symlinks correctly on Windows. What about modifying the index directly? I propose something like this, in `test/testhelpers.sh`:

```sh
add_symlink() {
  local src=$1
  local dest=$2

  prefix=`git rev-parse --show-prefix`
  hash=`printf "$src" | git hash-object -w --stdin`

  git update-index --add --cacheinfo 120000 "$hash" "$prefix$dest"
  git checkout -- "$dest"
}
```

(with modifications from https://coderwall.com/p/z86txw/make-symlink-on-windows-in-a-git-repo).

We could call that in your new test and I think that it would work OK. If not, you are welcome to skip this test for now and I can come back to it later.

PR #2997, opened 2018-05-04
remove reference to CLA from contributor's guide

zackse:
The contributor's guide still linked to the now-removed CLA page. This commit removes that reference.


zackse:
Note the changelog for 2.4.0 references removal of the CLA per issue #2870 but that doesn't appear to be accurate:

```
* all: remove CLA #2870 (@MikeMcQuaid)
```

(https://github.com/git-lfs/git-lfs/blame/master/CHANGELOG.md#L103)

Issue #3019, opened 2018-05-22
Hardware accelerated hashing


m1h4:
Is Git-LFS using https://en.wikipedia.org/wiki/Intel_SHA_extensions where possible for hashing files?

Since by its nature, Git-LFS is used with possibly hundred or thousand megabyte files, having the fastes possible hashing implementation would significantly impact git operations and general developer experience.

ttaylorr:
Hi @m1h4, thanks for pointing this out and opening this issue. It's a great idea; we don't currently use any hardware acceleration, since we rely on the `crypto/sha256` to do this type of work.

I found https://github.com/minio/sha256-simd, which seems to do what we want, i.e., "be a drop-in replacement that uses hardware acceleration where possible, and falls back to crypto/sha256 where it's not." 

I pushed a branch of in: https://github.com/git-lfs/git-lfs/pull/3021.

ttaylorr:
@m1h4 interesting; I wasn't aware that Go had support for SIMD optimizations of this sort, but admit that I am only superficially aware of the `crypto/*` packages, not any of their specific details. I think that more measurement is worthwhile to see if there are speedups in a practical setting within Git LFS between `sha256-simd` and `crypto/sha256`.

ttaylorr:
Closing per #3021.

PR #3021, opened 2018-05-22
[WIP] crypto/sha256: use SIMD optimization where possible

ttaylorr:
This pull request is a take on a suggestion given in https://github.com/git-lfs/git-lfs/issues/3019, that we should use SHA256 hardware optimization techniques like SIMD where possible.

This pull request so far does three things:

1. Imports https://github.com/minio/sha256-simd, which is a drop-in replacement for `crypto/sha256` where possible.
2. Introduces `github.com/git-lfs/git-lfs/tools/sha256` as a replacement for its `crypto/*` counterpart.
3. Finally, updates all (relevant) references to `crypto/sha256` to use the new `tools` counterpart.

I wanted to get this open as a WIP to start with so that we can whether or not this compiles against all of the targets that we are interested in and to have a place that we can measure the performance impact of these changes.

Closes: https://github.com/git-lfs/git-lfs/issues/3019.

##

/cc @git-lfs/core @m1h4


ttaylorr:
In https://github.com/git-lfs/git-lfs/issues/3019#issuecomment-391301754, @m1h4 notes that Go is already using some sort of accelerated hashing instructions, and I confirmed that that is AVX2 \[[1](https://blog.minio.io/accelerate-sha256-up-to-8x-over-3-gb-s-per-core-with-avx512-a0b1d64f78f)\].

In \[[2](https://github.com/golang/go/wiki/AVX-512-support-in-Go-assembler)\], we learn that Go 1.11 will ship with SIMD support, inclining me to believe that accelerated AVX512 support for SHA256 will be coming soon as well.

Here are some performance benchmarks between the sha256-simd branch and master:

| | | | | | | | 
-- | -- | -- | -- | -- | -- | --
sha256-simd | 1.895 | 1.895 | 2.013 | 1.882 | 1.869 | 1.9108
master | 1.991 | 2.077 | 2.063 | 2.138 | 2.082 | 2.0702
  |   |   |   |   |   | **-7.70%**

As you can see, the sha256-simd branch provides a marginal speed-up as compared to master. I think that this benefit is minimal, given that other overhead from Git LFS dominates the runtime.

I think that I would rather wait for Go 1.11, so I am going to close this for now.

Issue #3025, opened 2018-05-25
Authentication not retrieved / prompted upon redirect


ashmckenzie:
Hello from GitLab :)

I've observed an issue where a HTTP 302 redirect is detected (https://github.com/git-lfs/git-lfs/blob/master/lfsapi/client.go#L372) but the new HTTP request results in a 401 because authentication is not attempted.

In my scenario, I'm attempting a `git lfs -r secondary lock|unlock` where the secondary remote is defined as:

```
secondary	http://localhost:3002/repo.git (fetch)
secondary	http://localhost:3002/repo.git (push)
```

GitLab is running at `http://localhost:3002` and is configured as a secondary (https://docs.gitlab.com/ee/administration/geo/replication/#overview) and is therefore not able to accept DB writes.  We have the necessary HTTP redirect logic implemented to redirect to the primary (`http://localhost:3001`) and is working as expected for git, but we're seeing some issues with git lfs:

`git lfs lock -r secondary` attempt:

```
$ GIT_TRACE=1 git lfs lock -r secondary 1527141413.lfs
15:35:32.637108 git.c:576               trace: exec: git-lfs lock -r secondary 1527141413.lfs
15:35:32.637938 run-command.c:640       trace: run_command: git-lfs lock -r secondary 1527141413.lfs
15:35:32.657015 trace git-lfs: exec: git 'version'
15:35:32.670150 trace git-lfs: exec: git '-c' 'filter.lfs.smudge=' '-c' 'filter.lfs.clean=' '-c' 'filter.lfs.process=' '-c' 'filter.lfs.required=false' 'rev-parse' 'HEAD' '--symbolic-full-name' 'HEAD'
15:35:32.685016 trace git-lfs: exec: git 'config' '-l'
15:35:32.694227 trace git-lfs: creds: git credential fill ("http", "localhost:3002", "")
Username for 'http://localhost:3002': root
Password for 'http://root@localhost:3002':
15:35:39.610322 trace git-lfs: Filled credentials for http://localhost:3002/repo.git
15:35:39.687054 trace git-lfs: HTTP: POST http://localhost:3002/repo.git/info/lfs/locks
15:35:43.231598 trace git-lfs: HTTP: 302
15:35:43.231624 trace git-lfs: api: redirect POST http://localhost:3002/repo.git/info/lfs/locks to http://localhost:3001/repo.git/info/lfs/locks
[...]

PR #3028, opened 2018-05-25
lfsapi: re-authenticate HTTP redirects when needed

ttaylorr:
This pull request teaches package `lfsapi` to re-authenticate HTTP requests when the response indicates a redirection.

When an HTTP request is sent through package `lfsapi`, there are a few paths that we can take down into `do()` and `doWithRedirects()`. The one that we're concerned with involves one of two outcomes:

1. We send an HTTP request that is responded to with a 301 or 302 code. If we're within our retry limit, we construct a new request sent to the new host/path, and copy over the headers from the old request \[[1](https://github.com/git-lfs/git-lfs/blob/v2.4.1/lfsapi/client.go#L360-L368)\]. Notably, we do _not_ copy the `Authorization` header iff the new host differs from the old one, as a security concern. We retry the request.

2. Same setup as (1), except we have exceeded our retry count. We complain and fail \[[2](https://github.com/git-lfs/git-lfs/blob/v2.4.1/lfsapi/client.go#L213-L216)\].

In case (1) this security measure has the opportunity to fail when _the redirected host differs from the requested host_ and _the redirected host requires authentication_. Because we recur early on `doWithRedirects()`, and _not_ `DoWithAuth()`, we don't re-authenticate the request, and instead devolve into a case like https://github.com/git-lfs/git-lfs/issues/3025.

But, recurring through `DoWithAuth()` would throw away our state `via`, allowing authenticated request to flow through different pairs of hosts infinitely. That's no good.

Instead, let's try this:

- 1a3a19c: set up `do()` to take `via []*http.Request`, the list of requests that have lead up to this one (redirects), or `nil` if there are no such requests.
- b1c2b42: keep track of the `remote` associated with a given request, or `""` if no remote matches, or we don't care which remote is associated.
- 5257e58: branch through `doWithAuth` and override the `via` list such that we can re-authenticate requests across different hosts.

One tricky thing in 5257e58 is that we don't actually know whether the original request was designated to be authenticated or not. An additional tricky thing is that we might _never_ authenticate the original request, and instead load from `http.[$url.]extraHeaders`, such that `Authorization` header is present, but Git LFS didn't put it there.

In all of those cases, we reduce the problem to checking whether the original request contained an `Authorization` header.

Closes: https://github.com/git-lfs/git-lfs/issues/3025.

##

/cc @git-lfs/core https://github.com/git-lfs/git-lfs/issues/3025

### DIFF
--- a/lfsapi/auth.go
+++ b/lfsapi/auth.go
@@ -24,6 +24,10 @@ var (
 // authentication from netrc or git's credential helpers if necessary,
 // supporting basic and ntlm authentication.
 func (c *Client) DoWithAuth(remote string, req *http.Request) (*http.Response, error) {
+	return c.doWithAuth(remote, req, nil)
+}
+
+func (c *Client) doWithAuth(remote string, req *http.Request, via []*http.Request) (*http.Response, error) {
 	req.Header = c.extraHeadersFor(req)
 
 	apiEndpoint, access, credHelper, credsURL, creds, err := c.getCreds(remote, req)
@@ -31,7 +35,7 @@ func (c *Client) DoWithAuth(remote string, req *http.Request) (*http.Response, e
 		return nil, err
 	}
 
-	res, err := c.doWithCreds(req, credHelper, creds, credsURL, access, nil)
+	res, err := c.doWithCreds(req, credHelper, creds, credsURL, access, via)
 	if err != nil {
 		if errors.IsAuthError(err) {
 			newAccess := getAuthAccess(res)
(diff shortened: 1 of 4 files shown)


lfsapi/auth.go
```diff
@@ -57,11 +67,11 @@ func (c *Client) DoWithAuth(remote string, req *http.Request) (*http.Response, e
 	return res, err
 }
 
-func (c *Client) doWithCreds(req *http.Request, credHelper CredentialHelper, creds Creds, credsURL *url.URL, access Access) (*http.Response, error) {
+func (c *Client) doWithCreds(req *http.Request, credHelper CredentialHelper, creds Creds, credsURL *url.URL, access Access, via []*http.Request) (*http.Response, error) {
```
ashmckenzie: Monster arg list!

ashmckenzie:
Looks good!  I started to prepare a fix but found myself putting `remote` in a lot of places and bailed.  

Challenging one to test!

ttaylorr:
> Challenging one to test!

Thanks for the reminder. We test package `lfsapi` from Go, mostly as a consequence of not having established ways to run "fake" HTTP servers from the integration tests. I added a test exercising this new behavior here: https://github.com/git-lfs/git-lfs/pull/3028/commits/8715e893e3f240f41cfc057234ae7d8ac16d5b28.

ttaylorr:
Windows failures are unrelated.

dosire:
Thanks for fixing this @ttaylorr

PR #3029, opened 2018-05-26
Add `--no-rewrite` flag to `migrate import` command

PastelMobileSuit:
This PR adds a new `--no-rewrite` flag to the `migrate import` command. When this flag is given, `migrate import` takes a list of files as arguments and imports these files into Git LFS. It does this by rewriting the git tree in a new commit, without modifying git history.

Some of the code in this PR was based off of code previously written by @ttaylorr. Also, a big thank you to @ttaylorr for pairing with me on parts of this PR!

### DIFF
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -40,6 +40,9 @@ var (
 	// migrateNoRewrite is the flag indicating whether or not the
 	// command should rewrite git history
 	migrateNoRewrite bool
+	// migrateCommitMessage is the message to use with the commit generated
+	// by the migrate command
+	migrateCommitMessage string
 )
 
 // migrate takes the given command and arguments, *odb.ObjectDatabase, as well
@@ -292,6 +295,7 @@ func init() {
 	importCmd.Flags().BoolVar(&migrateVerbose, "verbose", false, "Verbose logging")
 	importCmd.Flags().StringVar(&objectMapFilePath, "object-map", "", "Object map file")
 	importCmd.Flags().BoolVar(&migrateNoRewrite, "no-rewrite", false, "Add new history without rewriting previous")
+	importCmd.Flags().StringVarP(&migrateCommitMessage, "message", "m", "", "With --no-rewrite, an optional commit message")
 
 	RegisterCommand("migrate", nil, func(cmd *cobra.Command) {
 		cmd.PersistentFlags().StringVarP(&includeArg, "include", "I", "", "Include a list of paths")
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -36,6 +36,10 @@ var (
 	// objectMapFile is the path to the map of old sha1 to new sha1
 	// commits
 	objectMapFilePath string
+
+	// migrateNoRewrite is the flag indicating whether or not the
+	// command should rewrite git history
+	migrateNoRewrite bool
 )
 
 // migrate takes the given command and arguments, *odb.ObjectDatabase, as well
@@ -287,6 +291,7 @@ func init() {
 	importCmd := NewCommand("import", migrateImportCommand)
 	importCmd.Flags().BoolVar(&migrateVerbose, "verbose", false, "Verbose logging")
 	importCmd.Flags().StringVar(&objectMapFilePath, "object-map", "", "Object map file")
+	importCmd.Flags().BoolVar(&migrateNoRewrite, "no-rewrite", false, "Add new history without rewriting previous")
 
 	RegisterCommand("migrate", nil, func(cmd *cobra.Command) {
 		cmd.PersistentFlags().StringVarP(&includeArg, "include", "I", "", "Include a list of paths")
--- a/commands/command_migrate_import.go
+++ b/commands/command_migrate_import.go
@@ -49,7 +49,7 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 		root := commit.TreeID
 
 		filter := git.GetAttributeFilter(cfg.LocalWorkingDir(), cfg.LocalGitDir())
-		if len(filter.Include()) == 0 && len(filter.Exclude()) == 0 {
+		if len(filter.Include()) == 0 {
 			ExitWithError(errors.Errorf("fatal: no Git LFS filters found in .gitattributes"))
 		}
 
--- a/commands/command_migrate_import.go
+++ b/commands/command_migrate_import.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/git-lfs/git-lfs/errors"
 	"github.com/git-lfs/git-lfs/filepathfilter"
 	"github.com/git-lfs/git-lfs/git"
 	"github.com/git-lfs/git-lfs/git/githistory"
@@ -28,6 +29,59 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 		ExitWithError(err)
 	}
 	defer db.Close()
+
+	if migrateNoRewrite {
+		ref, err := git.CurrentRef()
+		if err != nil {
+			ExitWithError(errors.Wrap(err, "fatal: unable to find current reference"))
+		}
+
+		sha, _ := hex.DecodeString(ref.Sha)
+		commit, err := db.Commit(sha)
+		if err != nil {
+			ExitWithError(errors.Wrap(err, "fatal: unable to load commit"))
+		}
+
+		root := commit.TreeID
+
+		include, _ := getIncludeExcludeArgs(cmd)
+		paths, _ := determineIncludeExcludePaths(cfg, include, nil)
+		if len(paths) == 0 {
+			ExitWithError(errors.New("fatal: no matching paths found"))
+		}
+
+		for _, path := range paths {
+			root, err = rewriteTree(db, root, path)
+			if err != nil {
+				ExitWithError(errors.Wrapf(err, "fatal: could not rewrite %q", path))
+			}
+		}
+
+		name, email := cfg.CurrentCommitter()
+		author := fmt.Sprintf("%s <%s>", name, email)
+
+		oid, err := db.WriteCommit(&odb.Commit{
+			Author:    author,
+			Committer: author,
+			ParentIDs: [][]byte{sha},
+			Message:   generateMigrateCommitMessage(cmd, include),
+			TreeID:    root,
+		})
+
+		if err != nil {
+			ExitWithError(errors.Wrap(err, "fatal: unable to write commit"))
+		}
+
+		if err := git.UpdateRef(ref, oid, "git lfs migrate import --no-rewrite"); err != nil {
+			ExitWithError(errors.Wrap(err, "fatal: unable to update ref"))
+		}
+
+		if err := checkoutNonBare(l); err != nil {
+			ExitWithError(errors.Wrap(err, "fatal: could not checkout"))
+		}
+
+		return
+	}
 
 	rewriter := getHistoryRewriter(cmd, db, l)
 
@@ -104,17 +158,36 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 		UpdateRefs: true,
 	})
 
-	// Only perform `git-checkout(1) -f` if the repository is
-	// non-bare.
-	if bare, _ := git.IsBare(); !bare {
-		t := l.Waiter("migrate: checkout")
-		err := git.Checkout("", nil, true)
-		t.Complete()
-
-		if err != nil {
-			ExitWithError(err)
-		}
+	if err := checkoutNonBare(l); err != nil {
+		ExitWithError(errors.Wrap(err, "fatal: could not checkout"))
 	}
+}
+
+// generateMigrateCommitMessage generates a commit message used with
+// --no-rewrite, using --message (if given) or generating one if it isn't.
+func generateMigrateCommitMessage(cmd *cobra.Command, patterns *string) string {
+	if cmd.Flag("message").Changed {
+		return migrateCommitMessage
+	}
+	if patterns != nil {
+		return fmt.Sprintf("%s: convert to Git LFS", *patterns)
+	}
+	panic("unexpected")
+}
+
+// checkoutNonBare forces a checkout of the current reference, so long as the
+// repository is non-bare.
+//
+// It returns nil on success, and a non-nil error on failure.
+func checkoutNonBare(l *tasklog.Logger) error {
+	if bare, _ := git.IsBare(); bare {
+		return nil
+	}
+
+	t := l.Waiter("migrate: checkout")
+	defer t.Complete()
+
+	return git.Checkout("", nil, true)
 }
 
 // trackedFromFilter returns an ordered set of strings where each entry is a
@@ -201,4 +274,8 @@ func trackedToBlob(db *odb.ObjectDatabase, patterns *tools.OrderedSet) ([]byte, 
 		Contents: &attrs,
 		Size:     int64(attrs.Len()),
 	})
+}
+
+func rewriteTree(db *odb.ObjectDatabase, root []byte, path string) ([]byte, error) {
+	return root, nil
 }
--- a/commands/command_migrate_import.go
+++ b/commands/command_migrate_import.go
@@ -56,13 +56,15 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 		gf := lfs.NewGitFilter(cfg)
 
 		for _, file := range args {
-			if filter.Allows(file) {
-				root, err = rewriteTree(gf, db, root, file)
-				if err != nil {
-					ExitWithError(errors.Wrapf(err, "fatal: could not rewrite %q", file))
-				}
-			} else {
+			if !filter.Allows(file) {
 				ExitWithError(errors.Errorf("fatal: file %s did not match any entries in .gitattributes", file))
+			}
+		}
+
+		for _, file := range args {
+			root, err = rewriteTree(gf, db, root, file)
+			if err != nil {
+				ExitWithError(errors.Wrapf(err, "fatal: could not rewrite %q", file))
 			}
 		}
 
--- a/commands/command_migrate_import.go
+++ b/commands/command_migrate_import.go
@@ -31,6 +31,10 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 	defer db.Close()
 
 	if migrateNoRewrite {
+		if len(args) == 0 {
+			ExitWithError(errors.Errorf("fatal: the migrate import command requires a list of files to import when using the --no-rewrite flag"))
+		}
+
 		ref, err := git.CurrentRef()
 		if err != nil {
 			ExitWithError(errors.Wrap(err, "fatal: unable to find current reference"))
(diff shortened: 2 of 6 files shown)


commands/command_migrate.go
```diff
@@ -287,6 +294,8 @@ func init() {
 	importCmd := NewCommand("import", migrateImportCommand)
 	importCmd.Flags().BoolVar(&migrateVerbose, "verbose", false, "Verbose logging")
 	importCmd.Flags().StringVar(&objectMapFilePath, "object-map", "", "Object map file")
+	importCmd.Flags().BoolVar(&migrateNoRewrite, "no-rewrite", false, "Add new history without rewriting previous")
```
ttaylorr: I think that these are great short descriptions of what the `--no-rewrite` and `--message` flags do, which is exactly what is appropriate for these.

Could you also add a longer explanation of what these flags are in [`git-lfs-migrate(1)`](https://github.com/git-lfs/git-lfs/blob/master/docs/man/git-lfs-migrate.1.ronn)? The `docs/man` tree is used to automatically generate `commands/mancontent.go`, and this in turn is what is shown when you run `git lfs migrate --help`, or `git lfs help migrate`, for instance.

The existing prose should be a good place to start in terms of what level of detail is appropriate.

commands/command_migrate_import.go
```diff
@@ -29,6 +30,67 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 	}
 	defer db.Close()
 
+	if migrateNoRewrite {
+		if len(args) == 0 {
+			ExitWithError(errors.Errorf("fatal: the migrate import command requires a list of files to import when using the --no-rewrite flag"))
```
ttaylorr: Great; I think dying here is the right thing to do. Let's shorten this message a little bit while we're here. For instance, the phrase "the migrate import command" might be unnecessary, since this will be the first line of output after a user runs `git lfs migrate import`.

What about: "fatal: expected one or more file(s) with --no-rewrite"?

commands/command_migrate_import.go
```diff
@@ -29,6 +30,67 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
...
+		}
+
+		ref, err := git.CurrentRef()
+		if err != nil {
+			ExitWithError(errors.Wrap(err, "fatal: unable to find current reference"))
+		}
+
+		sha, _ := hex.DecodeString(ref.Sha)
```
ttaylorr: Great; throwing this error away is reasonable since we trust `git.CurrentRef()` to always give back a valid SHA-256 `ref.Sha`.

commands/command_migrate_import.go
```diff
@@ -29,6 +30,67 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
...
+		commit, err := db.Commit(sha)
+		if err != nil {
+			ExitWithError(errors.Wrap(err, "fatal: unable to load commit"))
+		}
+
+		root := commit.TreeID
+
+		filter := git.GetAttributeFilter(cfg.LocalWorkingDir(), cfg.LocalGitDir())
+		if len(filter.Include()) == 0 && len(filter.Exclude()) == 0 {
+			ExitWithError(errors.Errorf("fatal: no git lfs filters setup in .gitattributes"))
```
ttaylorr: Let's use "found" instead of "setup".

commands/command_migrate_import.go
```diff
@@ -29,6 +30,67 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
...
+		}
+
+		for _, file := range args {
```
ttaylorr: I think one interesting consequence of `rewriteTree()` processing one file at a time is that there are some penalties that we'll pay if we get part of the way through this loop. In particular, if there are `m` arguments where `m` is large, and the `m-1`th of them doesn't match a pattern in `.gitattributes`, we'll have created a lot of tree objects that are dangling in the object database.

So, I think that we could run this loop twice, once to check that all given filenames match at least one pattern, and then once more to call `rewriteTree()`, which would mean we spend more time here. I think that's an OK tradeoff for not creating more objects than we need.

Of course, we could also look into ways to make `rewriteTree()` take more than one file and therefore create less dangling tree objects, but I'll leave it up to you if you want to go down that path.

commands/command_migrate_import.go
```diff
@@ -29,6 +30,67 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
...
+
+		name, email := cfg.CurrentCommitter()
+		author := fmt.Sprintf("%s <%s>", name, email)
+
+		filelist := strings.Join(args, ",")
+		oid, err := db.WriteCommit(&odb.Commit{
+			Author:    author,
+			Committer: author,
+			ParentIDs: [][]byte{sha},
+			Message:   generateMigrateCommitMessage(cmd, &filelist),
```
ttaylorr: Let's inline `filelist`.

commands/command_migrate_import.go
```diff
@@ -202,3 +283,98 @@ func trackedToBlob(db *odb.ObjectDatabase, patterns *tools.OrderedSet) ([]byte,
...
+	if err != nil {
+		return nil, err
+	}
+
+	splits := strings.SplitN(path, "/", 2)
+
+	switch len(splits) {
+	case 1:
+		// The path points to an entry at the root of this tree, so it must be a blob.
+		// Try to replace this blob with a git lfs pointer.
```
ttaylorr: Nit; casing on "Git LFS".

commands/command_migrate_import.go
```diff
@@ -202,3 +283,98 @@ func trackedToBlob(db *odb.ObjectDatabase, patterns *tools.OrderedSet) ([]byte,
...
+		if _, err := clean(lfs.NewGitFilter(cfg), &buf, blob.Contents, blobEntry.Name, blob.Size); err != nil {
+			return nil, err
+		}
+
+		newBlob := odb.Blob{
+			Contents: &buf,
+			Size:     int64(buf.Len()),
+		}
+
+		newOid, err := db.WriteBlob(&newBlob)
```
ttaylorr: Nit, let's inline `newBlob`.

commands/command_migrate_import.go
```diff
@@ -202,3 +283,98 @@ func trackedToBlob(db *odb.ObjectDatabase, patterns *tools.OrderedSet) ([]byte,
...
+
+		blobEntry := tree.Entries[index]
+		blob, err := db.Blob(blobEntry.Oid)
+		if err != nil {
+			return nil, err
+		}
+
+		var buf bytes.Buffer
+
+		if _, err := clean(lfs.NewGitFilter(cfg), &buf, blob.Contents, blobEntry.Name, blob.Size); err != nil {
```
ttaylorr: Why do we need to pass `lfs.NewGitFilter(cfg)`? I can't recall how the current implementation of `clean()` works, and whether or not we need this. If we do need it, can it be saved elsewhere instead of being constructed every time?

commands/command_migrate_import.go
```diff
@@ -202,3 +283,98 @@ func trackedToBlob(db *odb.ObjectDatabase, patterns *tools.OrderedSet) ([]byte,
...
+		if err != nil {
+			return nil, err
+		}
+
+		newEntry := odb.TreeEntry{
+			Name:     splits[0],
+			Filemode: blobEntry.Filemode,
+			Oid:      newOid,
+		}
+		tree = tree.Merge(&newEntry)
```
ttaylorr: Nit; let's inline `newEntry` as well, to keep the current scope as narrow as possible.

commands/command_migrate_import.go
```diff
@@ -202,3 +283,98 @@ func trackedToBlob(db *odb.ObjectDatabase, patterns *tools.OrderedSet) ([]byte,
...
+		})
+
+		return db.WriteTree(tree)
+
+	default:
+		return nil, errors.Errorf("error parsing path %s", path)
+	}
+}
+
+// findEntry searches a tree for the desired entry, and returns the
```
ttaylorr: Nit; This line is wrapped at fewer than 80 characters.

git/attribs.go
```diff
@@ -97,6 +98,22 @@ func GetAttributePaths(workingDir, gitDir string) []AttributePath {
...
+// configured with the filter=lfs attribute as a file path filter which
+// file paths can be matched against
+// workingDir is the root of the working copy
+// gitDir is the root of the git repo
+func GetAttributeFilter(workingDir, gitDir string) *filepathfilter.Filter {
+	paths := GetAttributePaths(workingDir, gitDir)
+	patterns := make([]filepathfilter.Pattern, len(paths))
+
+	for i, path := range paths {
+		patterns[i] = filepathfilter.NewPattern(path.Path)
```
ttaylorr: In case we want to change this later in a way that takes away the 1-1 correspondence between paths and patterns, we can minimize that diff by calling `append()` instead of using the slicing operator `[]`. In order to make that work as-expected, you'll want to change L108 to initialize `patterns` with a capacity of `len(paths)`, instead of a `len` of `len(paths)`.

test/test-migrate-fixtures.sh
```diff
@@ -44,6 +44,46 @@ setup_local_branch_with_gitattrs() {
...
+# setup_local_branch_with_nested_gitattrs creates a repository as follows:
+#
+#   A---B
+#        \
+#         refs/heads/master
+#
+# - Commit 'A' has 120, in a.txt, and a corresponding entry in .gitattributes. There is also
+#   140 in a.md, with no corresponding entry in .gitattributes.
+#   It also has 140 in subtree/a.md, and a corresponding entry in subtree/.gitattributes
+setup_local_branch_with_nested_gitattrs() {
```
ttaylorr: ✨ Awesome!

test/test-migrate-import-no-rewrite.sh
```diff
@@ -0,0 +1,212 @@
...
+  prev_commit_oid="$(git rev-parse HEAD)"
+
+  git lfs migrate import --no-rewrite *.txt
+
+  # Ensure our desired files were imported into git-lfs
+  assert_pointer "refs/heads/master" "a.txt" "$txt_oid" "120"
+  assert_local_object "$txt_oid" "120"
+
+  # Ensure the git history remained the same
+  new_commit_oid="$(git rev-parse HEAD~1)"
```
ttaylorr: ✨ Nice.

test/test-migrate-import-no-rewrite.sh
```diff
@@ -0,0 +1,212 @@
...
+end_test
+
+begin_test "migrate import --no-rewrite (no .gitattributes)"
+(
+  set -e
+
+  setup_multiple_local_branches
+
+  # Ensure command fails if no .gitattributes files are present
+  [ !"$(git lfs migrate import --no-rewrite *.txt *.md)" ]
```
ttaylorr: Do you think we could check that the error message this invocation gives is what we expect it to be? I think one way to do this would be to write something of the ilk:

```bash
git lfs migrate import --no-rewrite *.txt *.md 2>&1 | tee migrate.log
if [ ${PIPESTATUS[0]} -eq 0 ]; then
  echo >&2 "fatal: expected git lfs migrate import --no-rewrite to fail, didn't"
fi
grep "..." migrate.log
```

commands/command_migrate_import.go
PastelMobileSuit: It looks like it's being used in a few places in the `clean()` implementation, but it shouldn't change throughout the course of the entire operation. I agree that it makes more sense to construct one in `migrateImportCommand()` and pass it down into `rewriteTree()`, so I'll go ahead and change that

commands/command_migrate_import.go
PastelMobileSuit: Ah okay, thanks! I didn't think about the possibility of dangling objects.

I agree with you on running the loop twice and think that's the best option, since we won't be able to avoid checking files against the `.gitattributes` filters.

PastelMobileSuit:
@ttaylorr Thank you for the review! I've gone ahead and pushed up some new commits addressing your feedback

test/test-migrate-import-no-rewrite.sh
```diff
@@ -0,0 +1,222 @@
...
+    exit 1
+  fi
+
+  # Ensure the provided commit message was used
+  commit_msg="$(git log -1 --pretty=format:%s)"
+  if [ "$commit_msg" != "" ]; then
+    exit 1
+  fi
+)
+end_test
```
ttaylorr: Nit: missing trailing LF.

docs/man/git-lfs-migrate.1.ronn
```diff
@@ -84,13 +84,39 @@ options and these additional ones:
...
     format is CSV with this pattern: `OLD-SHA`,`NEW-SHA`
 
-If `--include` or `--exclude` (`-I`, `-X`, respectively) are given, the
+* `--no-rewrite`
+    Migrate large objects to Git LFS in a new commit without rewriting git history.
+    Please note that when this option is used, the `migrate import` command will expect
+    a different argument list, specialized options will become available, and the core 
+    `migrate` options will be ignored. See [IMPORT (NO REWRITE)].
+
+If `--no-rewrite` is not provided and `--include` or `--exclude` (`-I`, `-X`, respectively) are given, the
```
ttaylorr: Nit: it looks like a few lines in this file are longer than 80 columns. Can you make sure that the wrapping is set properly for `.ronn` files?

docs/man/git-lfs-migrate.1.ronn
```diff
@@ -84,13 +84,39 @@ options and these additional ones:
...
+
+The `import` mode has a special sub-mode enabled by the `--no-rewrite` flag. This sub-mode
+will migrate large objects to pointers as in the base `import` mode, but will do so in a new commit
+without rewriting Git history. When using this sub-mode, the base `migrate` options, 
+such as `--include-ref`, will be ignored, as will those for the base `import` mode. 
+The `migrate` command will also take a different argument list.
+
+The `--no-rewrite` sub-mode supports the following options and arguments:
+
+* `-m` <message> `--message`=<message>
```
ttaylorr: I think the backticks should wrap from the first `-` to the last `>` in each example.

docs/man/git-lfs-migrate.1.ronn
```diff
@@ -84,13 +84,39 @@ options and these additional ones:
...
-If neither of those flags are given, the gitattributes will be incrementally
+If `--no-rewrite` is not provided and neither of those flags are given, the gitattributes will be incrementally
 modified to include new filepath extensions as they are rewritten in history.
 
+### IMPORT (NO REWRITE)
+
+The `import` mode has a special sub-mode enabled by the `--no-rewrite` flag. This sub-mode
+will migrate large objects to pointers as in the base `import` mode, but will do so in a new commit
+without rewriting Git history. When using this sub-mode, the base `migrate` options, 
+such as `--include-ref`, will be ignored, as will those for the base `import` mode.
```
ttaylorr: I think that ignoring `--include-ref` is totally OK here, but we should perhaps add a note in the documentation that all interested branches will have to merge new commits generated by `git lfs migrate import --no-rewrite` if they _also_ want to take new commits in.

docs/man/git-lfs-migrate.1.ronn
```diff
@@ -189,6 +215,18 @@ $ git lfs migrate import --everything --include="*.zip"
 
 Note: This will require a force push to any existing Git remotes.
 
+### Migrate without rewriting local history
+
+You can also migrate files without modifying the existing history of your respoitory:
+
+```
+# Without a specified commit message
```
ttaylorr: Let's move these comments out of the code block, and replace the shell with 4 spaces, and then `git lfs migrate ...` (and without the `$`).

git/attribs.go
```diff
@@ -73,7 +73,7 @@ func GetAttributePaths(workingDir, gitDir string) []AttributePath {
 				fields := strings.Fields(line)
 				pattern := fields[0]
 				if len(reldir) > 0 {
-					pattern = filepath.Join(reldir, pattern)
+					pattern = filepath.ToSlash(filepath.Join(reldir, pattern))
```
PastelMobileSuit: The necessity of this change was prompted by a consistent Windows test failure I was seeing on a VM. Every time I ran the nested `.gitattributes` test, I saw a failure.

It turns out the problem is that `filepath.Join()` uses `\` as the separator character on Windows, rather than `/`. As a result, a `*.txt` pattern in a folder `subtree` would be parsed as `subtree\*.txt`, and instead of acting as a path separator, the `\` would escape the `*` character, so the pattern would then only match a file with the name `subtree*.txt`.

PastelMobileSuit:
Okay I've pushed up some additional commits to address feedback and have finally fixed all the AppVeyor test failures, so this should be ready to go!

git/attribs.go
ttaylorr: Great fix! One nit is that `filepath.Join` calls `filepath.Clean` (see \[[1](https://github.com/golang/go/blob/master/src/path/filepath/path_unix.go#L45)\]), which can have unintended consequences with certain pattern types (i.e., escape sequences are removed, etc.)

How about we shorten this to:

```go
pattern = fmt.Sprintf("%s/%s", reldir, pattern)
```

or:

```go
pattern = strings.Join([]string{reldir, pattern}, "/")
```

PastelMobileSuit: @ttaylorr Hey! I believe this code was already being used as-is with the `track` command, so I think that was already removing escape characters from patterns. I did some investigation to confirm this, and `git lfs track "*\*.mov"` gets converted to `*/*.mov` in the `.gitattributes` file, and if you manually change it to `*\*.mov`, that's not matched by `track` on a file with the name `a*.mov`.

hoeler:
@PastelMobileSuit @ttaylorr 
Hi!
I am not sure I grasp the use-case for the new "--no-rewrite" feature of git lfs import:
I was looking for a way to migrate our huge repository to lfs withouth destroying all of our (numerous) references to commit SHAs. I came to the conclusion that this is not possible and we will need to do a clean start with a rewritten history.
Then I found the announcement of the "--no-rewrite" feature and thought it may provide a way of moving all required files to lfs without changing the history (maybe by leaving the SHA intact although the content of the commits has changed). But this does not seem to be the use-case for this feature.

Can you please elaborate on when to use this feature? Sorry if I am missing something obvious, but the manual for migrate did not help me understand it and there is no further documentation that I can find.

Thanks a lot!

ttaylorr:
Hi, @hoeler. Thanks for writing in, and I'd be more than happy to answer
your question. It seems to me that the central point is:

> Then I found the announcement of the "--no-rewrite" feature and
> thought it may provide a way of moving all required files to lfs
> without changing the history (maybe by leaving the SHA intact although
> the content of the commits has changed).

Which isn't quite the case. The '--no-rewrite' flag works by _adding_
new commits, instead of _rewriting_ existing ones.

The SHA-1 of a commit is dependent upon the contents of the files in
that commit's root tree, the parents, message, and etc., of the commit
itself. So, there is no way to "rewrite" the contents of a commit
without also changing its SHA-1.

'--no-rewrite' is used in a case where (1) you have files in Git that
should have been committed with Git LFS, but weren't, and (2) those
files are under 100 MiB (thus, don't prevent pushing to GitHub.com),
and/or (3) history rewriting is not an option.

Per (3), when you run 'git lfs migrate <import | export>', you have to
_also_ ensure that each clone of the repository has (1) fetched and (2)
reset their local references to match the remote's. When running 'git
lfs migrate import' with '--no-rewrite', the repository's cloners need
only run 'git pull' to receive the new changes.

The central difference between the two from the standpoint of "what are
my references?" is: 'git lfs migrate import' (with or without
'--no-rewrite') changes your references, the difference is whether or
not the old values are still valid locations in your repository's line
of history or not.

hoeler:
Hi @ttaylorr, thank you for the explanation.
I am still not sure if i got it right though:

1) So basically "--no-rewrite" only works with the tip of a branch, by replacing existing files passed as arguments with LFS pointers and commiting that as a new change? Isn't this the same as manually removing the files from git and simply re-adding them under LFS control?

2) As I understand it now the "--no-rewrite" option does not help with getting rid of large files referenced somewhere in the git history, because this will _always_ require a rewrite of the history, correct?

ttaylorr:
> 1) So basically "--no-rewrite" only works with the tip of a branch, by
> replacing existing files passed as arguments with LFS pointers and
> commiting that as a new change? Isn't this the same as manually
> removing the files from git and simply re-adding them under LFS
> control?

I think that the gist of that is correct, but 'git lfs migrate import
--no-rewrite' has two attributes that make it different than 'git rm
--cached', which are:

  1. Change each file in a separate commit, and

  2. Perform the change in Git LFS, so that it will work even if 'git
     lfs install' hasn't been run.

> 2) As I understand it now the "--no-rewrite" option does not help with
> getting rid of large files referenced somewhere in the git history,
> because this will _always_ require a rewrite of the history, correct?

Exactly.

PR #3039, opened 2018-05-31
add basic example in main manpage

anarcat:
Every time I go back to git LFS after a long break, I wonder how it
works and how to get started. The first manual page is currently a
reference to a fairly large number of other manual pages and, as the
core documentation of LFS, does not really help users in getting basic
things done.

In comparison, the LFS homepage has a good primer, so I have taken it
as an example for a basic example that would work in the manual page
to help users get out of the weeds quickly.

### DIFF
--- a/docs/man/git-lfs.1.ronn
+++ b/docs/man/git-lfs.1.ronn
@@ -82,3 +82,28 @@ commands and low level ("plumbing") commands.
     Git pre-push hook implementation.
 * git-lfs-smudge(1):
     Git smudge filter that converts pointer in blobs to the actual content.
+
+## EXAMPLES
+
+To get started with Git LFS, the following commands can be used.
+
+ 1. Setup Git LFS on your system. You only have to do this once per
+    repository per machine:
+
+        git lfs install
+
+ 2. Choose the type of files you want to track, for examples all `ISO`
+    images, with git-lfs-track(1):
+
+        git lfs track "*.iso"
+
+ 3. The above stores this information in gitattributes(5) files, so
+    that file need to be added to the repository:
+
+        git add .gitattributes
+
+ 3. Commit, push and work with the files normally:
+
+        git add file.iso
+        git commit -m "Add disk image"
+        git push
--- a/docs/man/git-lfs-track.1.ronn
+++ b/docs/man/git-lfs-track.1.ronn
@@ -59,6 +59,6 @@ to match paths.
 
 ## SEE ALSO
 
-git-lfs-untrack(1), git-lfs-install(1), gitattributes(5).
+git-lfs-untrack(1), git-lfs-install(1), gitattributes(5), gitignore(5).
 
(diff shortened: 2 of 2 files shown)


docs/man/git-lfs.1.ronn
```diff
@@ -82,3 +82,27 @@ commands and low level ("plumbing") commands.
     Git pre-push hook implementation.
 * git-lfs-smudge(1):
     Git smudge filter that converts pointer in blobs to the actual content.
+
+## EXAMPLES
+
+To get started with LFS, the following commands can be used.
```
larsxschneider: minor nit: Let's call it `Git LFS` here instead of just `LFS`

docs/man/git-lfs.1.ronn
```diff
@@ -82,3 +82,27 @@ commands and low level ("plumbing") commands.
...
+
+ 1. setup LFS in a git repository, this needs to be done only once:
```
larsxschneider: Maybe:
```
Setup Git LFS on your system. You only have to do this once:
```
Your version indicates that the user needs to do this for every repo.

docs/man/git-lfs.1.ronn
```diff
@@ -82,3 +82,27 @@ commands and low level ("plumbing") commands.
...
+
+        git lfs install
+
+ 2. choose the type of files you want to track, for examples all `ISO`
```
larsxschneider: Maybe start the bullet points with an upper case letter? `Choose the ...` ?

docs/man/git-lfs.1.ronn
anarcat: i thought you *did* have to run `git lfs install` in each repository. is that incorrect? how are hooks setup otherwise?

ttaylorr: That's correct, and I think we can extend @larsxschneider's idea to capture this, perhaps:

> Setup Git LFS on your system. You only have to do this once per repository per machine:

docs/man/git-lfs.1.ronn
```diff
@@ -82,3 +82,27 @@ commands and low level ("plumbing") commands.
...
+    images, see gitignore(5) for possible patterns:
+
+        git lfs track "*.iso"
+
+ 3. git-lfs-track(1) stores this information in gitattributes(5)
```
ttaylorr: This page is called `git-lfs-track(1)`, not `git-Lfs-track(1)`.

anarcat:
i believe I have addressed most of the comments here, at least those I understood. i amended the first commit to:

 1. Capitalize the first word of each bullet point
 2. This required rephrasing the bullet point starting with a manpage reference to avoid confusion, which in turn meant punting the gitignore(5) reference to the lfs-track manpage, which I have done in a separate commit
 3. Rephrased the first bullet point to clarify when install should be called

I hope this is to your liking!

Issue #3037, opened 2018-05-30
Wrong message on checkout


shuhrat:
`git lfs checkout` print 
```
Uploading LFS objects:   0% (0/85), 0 B | 0 B/s                                 
Uploading LFS objects:   0% (0/3087), 0 B | 0 B/s                               
Uploading LFS objects:   0% (0/5723), 0 B | 0 B/s                               
Uploading LFS objects:   0% (0/8546), 0 B | 0 B/s                               
Uploading LFS objects:   0% (0/11357), 0 B | 0 B/s   
....
Uploading LFS objects: 100% (47086/47132), 1.7 GB | 3.8 MB/s                    
Uploading LFS objects: 100% (47109/47132), 1.7 GB | 3.8 MB/s                    
Uploading LFS objects: 100% (47132/47132), 1.7 GB | 3.8 MB/s, done
```

Suppose that checkout does not upload anything =)

ttaylorr:
@shuhrat thanks for opening this -- that's certainly a bug, I think because we don't set the meter type here:

https://github.com/git-lfs/git-lfs/blob/37bc00aa3e5b222f0df527291337d2b6f1e83645/commands/command_checkout.go#L33

larsxschneider:
@ttaylorr In what situation do we need `git lfs checkout`? I believe it was only needed when we did not have git filter process, right? Would it make sense to deprecate it?

ttaylorr:
@shuhrat @larsxschneider 

https://github.com/git-lfs/git-lfs/pull/3055: fixes inaccurate messaging, thus closing this issue
https://github.com/git-lfs/git-lfs/pull/3056: deprecates the command entirely

PR #3055, opened 2018-06-08
commands/checkout: fix inaccurate messaging

ttaylorr:
This pull request fixes an issue where `git lfs checkout` would log a message indicating that it was "Uploading LFS objects", instead of something more appropriate, such as "Checking out LFS objects".

This issue was pointed out by @shuhrat in https://github.com/git-lfs/git-lfs/issues/3037.

The cause of this issue is interesting: since `Upload` and `Download` are defined in terms of the `iota` keyword, `Upload` (being defined closer to the top of the file) is given value `0` whereas `Download` is given value `1`. Hence, when we avoid initialization, Go steps in and gives us the zero-value of this object, which happens to be `Upload`.

To avoid this in the future, we should make the Unknown field have value zero (or avoid this altogether, and start at `iota + 1`), but this change feels out-of-scope for this pull request.

To fix this, we have to do a few things:

1. 12dc6dd: to prepare for adding a new checkout action (in package `tq`), we should teach `Verb() string`, so that we don't have to introduce a new special case when logging this message out. Specifically, this is because we used to call `strings.Title(m.direction.String())`, but this is no longer desired.

2. aa2ad12: add the new checkout direction itself, along with new cases in `Verb()` and `String()`.

3. 58815a6: assign the appropriate "direction" the `tq.Meter` when it is initialized in `commands/command_checkout.go`.

## 

/cc @git-lfs/core 
/cc @shuhurat, https://github.com/git-lfs/git-lfs/issues/3037

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -29,6 +29,7 @@ func checkoutCommand(cmd *cobra.Command, args []string) {
 	var pointers []*lfs.WrappedPointer
 	logger := tasklog.NewLogger(os.Stdout)
 	meter := tq.NewMeter()
+	meter.Direction = tq.Checkout
 	meter.Logger = meter.LoggerFromEnv(cfg.Os)
 	logger.Enqueue(meter)
 	chgitscanner := lfs.NewGitScanner(func(p *lfs.WrappedPointer, err error) {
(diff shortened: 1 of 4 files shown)


ttaylorr:
I reverted some of the extra grepping on the log messages in 58815a6 with d2e4b9b since there appears to be an unnoticed race condition. I will continue to look into this, but am removing the extra checking since the problem seems to only affect CPUs with lots of context switching (like in CI, not locally).

PR #3056, opened 2018-06-08
commands/checkout: mark 'git lfs checkout' as deprecated

ttaylorr:
This pull request marks `git lfs checkout` as deprecated, indicating that it will be removed in v3.0.0.

`git checkout` benefits from the `delay=` capability introduced by @larsxschneider upstream, thus removing the need to retain this command in the future. Since we can't get rid of it between v2.4.0 and v2.5.0, let's mark that it is deprecated (similar to #2526) and schedule it for removal in v3.0.0.

One thing to consider is a feature such as `--include` or `--exclude`. `git lfs checkout` never supported these interestingly enough, so I suppose `git lfs checkout --include` was a fairly uncommon use-case. If someone so desired it, it could be accomplished by

    $ git -c lfs.fetchinclude=foo checkout -- 

##

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/git-lfs/git-lfs/filepathfilter"
 	"github.com/git-lfs/git-lfs/git"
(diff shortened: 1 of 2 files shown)


commands/command_checkout.go
```diff
@@ -14,6 +15,15 @@ import (
 
 func checkoutCommand(cmd *cobra.Command, args []string) {
 	requireInRepo()
+
+	msg := []string{
+		"WARNING: 'git lfs checkout' is deprecated and will be removed in v3.0.0.",
+
+		"'git checkout' has been updated in upstream Git to have comparable speeds",
```
larsxschneider: Minor nit: We could say here `upstream Git 2.15+`. If you like that then I would make a PR with an according commit. If you think that's not necessary then this is great too.

The PR is 👍 👍

ttaylorr: If you feel strongly, I certainly encourage it, but I think that this is fine without it. `git lfs clone` only shows this message based on the Git version, which I think we should not replicate here.

Issue #3052, opened 2018-06-07
‘migrate’ subcommand fails with ‘Couldn't find remote ref’


fedde-s:
I'm working on a branch that seems to have several un-migrated files, as I received an unhelpful message that they should have been pointers when I switched to the branch. When I try to fix this by running `git lfs migrate info` and `git lfs migrate import`, it fails with an error message like the following:

```Console
Error running /path/to/git '-c' 'filter.lfs.smudge=' '-c' 'filter.lfs.clean=' '-c' 'filter.lfs.process=' '-c' 'filter.lfs.required=false' 'fetch' 'origin' 'upstream': 'fatal: Couldn't find remote ref upstream' 'exit status 128'
```

…as if it's trying to fetch one of my remote repositories (`upstream`) as a ref inside another remote repository (`origin`). Am I doing something wrong here?

fedde-s:
For completeness: the unhelpful message I mentioned in the first sentence is not my main issue, as that's already being addressed in #2910.

fedde-s:
To me it looks like git-lfs-migrate simply contains a bug in which it omits the `--multiple` flag to git-fetch, though.

fedde-s:
See [my comment](https://github.com/git-lfs/git-lfs/commit/36f3c951717772d2d91b0c91449e49222126e1ef#r29345999) on the code that calls git-fetch in this case.

fedde-s:
Thanks writing a fix! And sorry for bypassing your good practice of providing the parameters in which I observed the issue. It can be very annoying to have someone (who doesn't know the project or even the language it's written in) insist that they already know the cause… But I'm glad I you did have enough information to write a test in this case 🙂

ttaylorr:
> Thanks writing a fix! And sorry for bypassing your good practice of providing the parameters in which I observed the issue.

Not to worry; thanks for letting me know! Don't hesitate to ping me here or elsewhere if you run into trouble in the future.

PR #3063, opened 2018-06-13
git/git.go: pass --multiple to git-fetch(1) when appropriate

ttaylorr:
This pull request passes the `--multiple` flag to `git-fetch(1)` when appropriate, i.e., when calling `git.Fetch()` with more than one remote.

When calling `git lfs migrate` (and the `--skip-fetch` option is not given) we first fetch all remote references, to make sure that we have an up-to-date copy of what is and is not present on the remote. This is done so that we can ensure that we aren't performing a migration that will later be rejected by a remote thus requiring a `--force`.

When users have multiple remotes, we will attempt to flatten that list into a single slice, and then call `git.Fetch()` once with all of the remotes. This is a correct calling convention for 0- and 1-arity arguments (none or a single remote) but is incorrect for `>1` remotes.

Specifically, if we have remotes `origin`, and `fork`, we will call `git fetch origin fork`, which tries to fetch the reference `fork` from the remote `origin`, which is likely to fail. Instead, adopt @fedde-s's recommendation in https://github.com/git-lfs/git-lfs/commit/36f3c951717772d2d91b0c91449e49222126e1ef#r29345999:

> Does this line break when more than one remote is provided, given that the main synopsis for git-fetch looks like the line below?
> 
> ```ShellSession
> git fetch [<options>] [<repository> [<refspec>...]]
> ```
>
> If so, this is probably the intended usage here:
> 
> ```ShellSession
> git fetch --multiple [<options>] [(<repository> | <group>)...]
> ```

Closes: https://github.com/git-lfs/git-lfs/issues/3052.

##

/cc @git-lfs/core 
/cc @fedde-s https://github.com/git-lfs/git-lfs/issues/3052

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -975,7 +975,13 @@ func Fetch(remotes ...string) error {
 		return nil
 	}
 
-	_, err := gitNoLFSSimple(append([]string{"fetch"}, remotes...)...)
+	var args []string
+	if len(remotes) > 1 {
+		args = []string{"--multiple", "--"}
+	}
+	args = append(args, remotes...)
+
+	_, err := gitNoLFSSimple(append([]string{"fetch"}, args...)...)
 	return err
 }
 
--- a/test/test-migrate-fixtures.sh
+++ b/test/test-migrate-fixtures.sh
@@ -308,6 +308,31 @@ setup_single_local_branch_with_annotated_tags() {
   git tag "v1.0.0" -m "v1.0.0"
 }
 
+setup_multiple_remotes() {
+  set -e
+
+  reponame="migrate-multiple-remotes"
+  remove_and_create_remote_repo "$reponame"
+
+  forkname="$(git remote -v \
+    | head -n1 \
+    | cut -d ' ' -f 1 \
+    | sed -e 's/^.*\///g')-fork"
+  ( setup_remote_repo "$forkname" )
+
+  git remote add fork "$GITSERVER/$forkname"
+
+  base64 < /dev/urandom | head -c 16 > a.txt
+  git add a.txt
+  git commit -m "initial commit"
+  git push origin master
+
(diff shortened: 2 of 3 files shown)


fedde-s:
This may not be the right place to ask this, but I can't figure out how to run the tests based on [the docs I could find](https://github.com/git-lfs/git-lfs/blob/v2.4.2/CONTRIBUTING.md#submitting-a-pull-request). Are they missing a step or did I skip something?

```console
% docker run --rm -v "$PWD:/git-lfs:ro" -w /git-lfs golang sh -c 'script/bootstrap && script/test'
can't load package: package github.com/git-lfs/git-lfs/commands: cannot find package "github.com/git-lfs/git-lfs/commands" in any of:
        /usr/local/go/src/github.com/git-lfs/git-lfs/commands (from $GOROOT)
        /go/src/github.com/git-lfs/git-lfs/commands (from $GOPATH)
script/build.go:18:2: cannot find package "github.com/git-lfs/git-lfs/config" in any of:
        /usr/local/go/src/github.com/git-lfs/git-lfs/config (from $GOROOT)
        /go/src/github.com/git-lfs/git-lfs/config (from $GOPATH)
```

fedde-s:
Since you believe that this does fix the issue (the code suggests to me that it does, even though I never learned Go), don't let my failure to run it delay the merge. I'll just send another issue report if it turns out that I *can* still reproduce it after v2.5.0 rolls out of my package manager 😄

ttaylorr:
> Are they missing a step or did I skip something?

@fedde-s Make sure that you mount your working copy into `$GOPATH`, like:

```bash
docker run --rm \
  -v "$PWD:/go/src/github.com/git-lfs/git-lfs" \
  -w /go/src/github.com/git-lfs/git-lfs \
  golang \
  './test/test-migrate-import.sh'
```

Issue #3072, opened 2018-06-18
Undocumented behaviour change between 2.3.4 → 2.4.x


goodtune:
A recipe I was previously using to selectively fetch objects from the remote LFS server has broken under 2.4.

The documentation states that patterns provided through the `-I` switch will overload the `lfs.fetchinclude` configuration for the single execution. I am identifying the paths I wish to retrieve using our necessary logic, so for example:

```
git lfs fetch -I bin/thing/here,bin/other/thang/is/here
```

When running with `GIT_TRACE=1` set, I now observe the following output which did not previously exist.

```
10:04:30.797796 trace git-lfs: filepathfilter: rewrite "bin/thing/here" as "**/bin/thing/here/**"
10:04:30.797844 trace git-lfs: filepathfilter: rewrite "bin/other/thang/is/here" as "**/bin/other/thang/is/here/**"
```

As a result the two files I want, in my confected example, aren't being retrieved as they don't match the pattern.

It [appears that literal strings will now always receive this treatment](https://github.com/git-lfs/git-lfs/blob/v2.4.1/filepathfilter/filepathfilter.go#L116-L141), and that is entirely unexpected behaviour in my opinion. If I have gone to the trouble to provide you with an exact path, why would you ever modify it?

goodtune:
My interim solution is to be less explicit and munge the preprocessor output to result in:

```
git lfs fetch -I /bin/thing/*,/bin/other/thang/is/*
```

That is sub-optimal though because now I may end up downloading `bin/other/thang/is/reallybigthingidonotwant` as a result.

ttaylorr:
Hi @goodtune, thanks for opening this issue, and leaving such detailed notes for me to read through. 

The `filepathfilter` changes are always tricky to get right, since we replaced the underlying engine in v2.4.0, and there are lots of things to get right and wrong when attempting to match the behavior in previous versions.

I patched this up in https://github.com/git-lfs/git-lfs/pull/3081, which should do the right thing. Can you give it a whirl to make sure that it matches your expectations? Thanks!

goodtune:
Thanks @ttaylorr - are you able to provide build artifacts you want me to test?

ttaylorr:
> are you able to provide build artifacts you want me to test?

Sure, how about these:

[git-lfs-darwin-386-2.4.0.tar.gz](https://github.com/git-lfs/git-lfs/files/2125853/git-lfs-darwin-386-2.4.0.tar.gz)
[...]

PR #3081, opened 2018-06-21
filepathfilter: use literal pattern when applicable

ttaylorr:
This pull request teaches package `filepathfilter` to use the literal pattern given to `--include`, `--exclude` (or similar) when appropriate in order to retain backwards compatibility with versions earlier than v2.4.0.

In the v2.3.0 and earlier, a pattern like `a/b/c.dat` would be matched using the following code:

https://github.com/git-lfs/git-lfs/blob/4dd2bf73d7819733f7d8cc29a8e8404fc0552242/filepathfilter/filepathfilter.go#L248-L256

Which means that, in practice--on that version--files would be matched when they either are (1) in a directory called `a/b/c.dat`, (2) in a parent-directory and contain that name i.e., `d/e/a/b/c.dat`, or (3) are _equal_ (among other things, and using `filepath.Match`) to the given pattern.

In v2.4.0 and later, we rewrote patterns `P` of that form to `**/P/**`, retaining behaviors (1) and (2). However, this matches too much to be strictly equivalent to (3), so we tighten the restriction as follows:

If we have a pattern `P` that would be rewritten using the above rules, mark that we would like to fallback to the literal pattern given, and attempt to run that pattern first. This _won't_ match files with `[`-style character classes in its name, but this is consistent with the earlier behavior.

If there is a match, don't attempt to search further, i.e., don't include sibling or children files in the match.

Therefore, this change makes package `filepathfilter` behaviorally equivalent to what it previously was.

##

/cc @git-lfs/core 
/cc https://github.com/git-lfs/git-lfs/issues/3072, @goodtune

### DIFF
--- a/filepathfilter/filepathfilter.go
+++ b/filepathfilter/filepathfilter.go
@@ -151,7 +151,7 @@ func NewPattern(p string) Pattern {
 
 	msg := fmt.Sprintf("filepathfilter: rewrite %q as %q", p, pp)
 	if useLit {
-		msg = fmt.Sprintf("%s (or fallback)")
+		msg = fmt.Sprintf("%s (or fallback)", msg)
 	}
 	tracerx.Printf(msg)
 
--- a/filepathfilter/filepathfilter.go
+++ b/filepathfilter/filepathfilter.go
@@ -1,6 +1,7 @@
 package filepathfilter
 
 import (
+	"fmt"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


filepathfilter/filepathfilter.go
```diff
@@ -138,15 +148,19 @@ func NewPattern(p string) Pattern {
 			}
 		}
 	}
-	tracerx.Printf("filepathfilter: rewrite %q as %q", p, pp)
+
+	msg := fmt.Sprintf("filepathfilter: rewrite %q as %q", p, pp)
+	if useLit {
+		msg = fmt.Sprintf("%s (or fallback)", msg)
```
PastelMobileSuit: Should we prepend this with `filepathfilter:` as well like above?

test/test-pull.sh
```diff
@@ -145,6 +145,39 @@ begin_test "pull"
 )
 end_test
 
+begin_test "pull (literal --include)"
```
PastelMobileSuit: This test looks correct to me based on the description of what's going wrong with the filepath patterns, but it's passing for me on master, so I don't think it's actually testing the root issue

filepathfilter/filepathfilter.go
ttaylorr: This already happens since the argument that we're filling in for the first `%s` is the `msg` as initialized above.

ttaylorr:
@PastelMobileSuit thanks for your review; I am going to close this because of my comment in https://github.com/git-lfs/git-lfs/issues/3072#issuecomment-400009620.

Issue #3057, opened 2018-06-08
Implement `migrate export` command


PastelMobileSuit:
We should add a new `export` subcommand to the `migrate` command. `migrate export` will do the opposite of `migrate import`. When given a range of references and a set of patterns, `migrate export` will rewrite git history within the given range to replace all LFS pointers matching the given patterns with the contents of the large files being pointed to.

By default, `migrate export` should require a reference range to work, but should also provide a `--everything` flag to allow rewriting of the entire history. We may also want to consider a `--no-rewrite` flag similar to that for the `migrate import` command, which would export a given set of files in a new commit without rewriting git history.

ttaylorr:
Hi @PastelMobileSuit, this seems like a great idea to me, and I definitely am excited about seeing it land in Git LFS. Before diving into the details, I think that we should establish a good timeline for releasing this. Between the rest of the planning for v2.5.0 and the `--no-rewrite` changes you landed in #3029, I think that this is a good candidate for v2.6.0. I definitely don't want to rush you or shut you out of a release, so if you feel like v2.6.0 is a bad estimate, let me know. We can obviously re-evaluate as you work through this.

> We should add a new `export` subcommand to the `migrate` command. `migrate export` will do the opposite of `migrate import`.

OK, that sounds good. I would expect that for any patterns X, Y, and references Z, that the following would be an effective no-op:

```ShellSession
$ git lfs migrate import --include-ref=Z --include=X --exclude=Y
$ git lfs migrate export --include-ref=Z --include=X --exclude=Y
```

> When given a range of references and a set of patterns, `migrate export` will rewrite git history within the given range to replace all LFS pointers matching the given patterns with the contents of the large files being pointed to.

Sounds good. I think that there are two main conceptual challenges that I think of:

1. How do we download all of the objects that we need?
2. What is the reverse migration in `.gitattributes`?

For (1) I think that doing a scan of all history in the range of the migration (i.e., everything between `--include-ref` and `--exclude-ref`) fo all LFS pointers, and then sending those to the transfer queue (in package `tq`) for download. This has the drawback of potentially downloading a lot of data, but in practice I'm not so concerned as to pursue other approaches immediately. I think as long as we are good stewards of the cache (i.e., that we prune anything in `.git/lfs/objects` that isn't reachable any longer) that we should be OK. Definitely let me know if you feel differently.

For (2) I think we have a couple of approaches. The most obvious is to take a `.gitattributes` prior to migration, and change it as follows:

[...]

PR #3084, opened 2018-06-23
Implement `migrate export` subcommand

PastelMobileSuit:
This PR implements a `migrate export` subcommand as described in #3057. The command requires a list of patterns specified with the `--include` argument and will convert any pointers matching those patterns from a Git LFS pointer to a regular Git object. Any objects that don't already exist in the cache will be downloaded before history is rewritten using the transfer queue.

I wanted to get this PR up for review now before it gets too large. I'm planning on a few more changes in subsequent PRs after this gets merged:
* Documentation
* Pruning the cache of unreachable objects
* Smarter modification of `.gitattributes` files

### DIFF
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -297,6 +297,8 @@ func init() {
 	importCmd.Flags().BoolVar(&migrateNoRewrite, "no-rewrite", false, "Add new history without rewriting previous")
 	importCmd.Flags().StringVarP(&migrateCommitMessage, "message", "m", "", "With --no-rewrite, an optional commit message")
 
+	exportCmd := NewCommand("export", migrateExportCommand)
+
 	RegisterCommand("migrate", nil, func(cmd *cobra.Command) {
 		cmd.PersistentFlags().StringVarP(&includeArg, "include", "I", "", "Include a list of paths")
 		cmd.PersistentFlags().StringVarP(&excludeArg, "exclude", "X", "", "Exclude a list of paths")
@@ -306,6 +308,6 @@ func init() {
 		cmd.PersistentFlags().BoolVar(&migrateEverything, "everything", false, "Migrate all local references")
 		cmd.PersistentFlags().BoolVar(&migrateSkipFetch, "skip-fetch", false, "Assume up-to-date remote references.")
 
-		cmd.AddCommand(importCmd, info)
+		cmd.AddCommand(importCmd, info, exportCmd)
 	})
 }
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -298,6 +298,8 @@ func init() {
 	importCmd.Flags().StringVarP(&migrateCommitMessage, "message", "m", "", "With --no-rewrite, an optional commit message")
 
 	exportCmd := NewCommand("export", migrateExportCommand)
+	exportCmd.Flags().BoolVar(&migrateVerbose, "verbose", false, "Verbose logging")
+	exportCmd.Flags().StringVar(&objectMapFilePath, "object-map", "", "Object map file")
 
 	RegisterCommand("migrate", nil, func(cmd *cobra.Command) {
 		cmd.PersistentFlags().StringVarP(&includeArg, "include", "I", "", "Include a list of paths")
@@ -308,6 +310,6 @@ func init() {
 		cmd.PersistentFlags().BoolVar(&migrateEverything, "everything", false, "Migrate all local references")
 		cmd.PersistentFlags().BoolVar(&migrateSkipFetch, "skip-fetch", false, "Assume up-to-date remote references.")
 
-		cmd.AddCommand(importCmd, info, exportCmd)
+		cmd.AddCommand(exportCmd, importCmd, info)
 	})
 }
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -43,6 +43,10 @@ var (
 	// migrateCommitMessage is the message to use with the commit generated
 	// by the migrate command
 	migrateCommitMessage string
+
+	// exportRemote is the remote from which to download objects when
+	// performing an export
+	exportRemote string
 )
 
 // migrate takes the given command and arguments, *odb.ObjectDatabase, as well
@@ -300,6 +304,7 @@ func init() {
 	exportCmd := NewCommand("export", migrateExportCommand)
 	exportCmd.Flags().BoolVar(&migrateVerbose, "verbose", false, "Verbose logging")
 	exportCmd.Flags().StringVar(&objectMapFilePath, "object-map", "", "Object map file")
+	exportCmd.Flags().StringVar(&exportRemote, "remote", "", "Remote from which to download objects")
 
 	RegisterCommand("migrate", nil, func(cmd *cobra.Command) {
 		cmd.PersistentFlags().StringVarP(&includeArg, "include", "I", "", "Include a list of paths")
--- a/commands/command_migrate_export.go
+++ b/commands/command_migrate_export.go
@@ -1,0 +1,127 @@
+package commands
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/git-lfs/git-lfs/errors"
+	"github.com/git-lfs/git-lfs/filepathfilter"
+	"github.com/git-lfs/git-lfs/git"
+	"github.com/git-lfs/git-lfs/git/githistory"
+	"github.com/git-lfs/git-lfs/git/odb"
+	"github.com/git-lfs/git-lfs/lfs"
+	"github.com/git-lfs/git-lfs/tasklog"
+	"github.com/git-lfs/git-lfs/tools"
+	"github.com/spf13/cobra"
+)
+
+func migrateExportCommand(cmd *cobra.Command, args []string) {
+	l := tasklog.NewLogger(os.Stderr)
+	defer l.Close()
+
+	db, err := getObjectDatabase()
+	if err != nil {
+		ExitWithError(err)
+	}
+	defer db.Close()
+
+	rewriter := getHistoryRewriter(cmd, db, l)
+
+	filter := rewriter.Filter()
+	if len(filter.Include()) <= 0 {
+		ExitWithError(errors.Errorf("fatal: one or more files must be specified with --include"))
+	}
+
+	tracked := trackedFromExportFilter(filter)
+	gitfilter := lfs.NewGitFilter(cfg)
+
+	migrate(args, rewriter, l, &githistory.RewriteOptions{
+		Verbose:           migrateVerbose,
+		ObjectMapFilePath: objectMapFilePath,
+		BlobFn: func(path string, b *odb.Blob) (*odb.Blob, error) {
+			if filepath.Base(path) == ".gitattributes" {
+				return b, nil
+			}
+
+			var buf bytes.Buffer
+
+			if _, err := smudge(gitfilter, &buf, b.Contents, path, false, rewriter.Filter()); err != nil {
+				return nil, err
+			}
+
+			return &odb.Blob{
+				Contents: &buf, Size: int64(buf.Len()),
+			}, nil
+		},
+
+		TreeCallbackFn: func(path string, t *odb.Tree) (*odb.Tree, error) {
+			if path != "/" {
+				// Ignore non-root trees.
+				return t, nil
+			}
+
+			ours := tracked
+			theirs, err := trackedFromAttrs(db, t)
+			if err != nil {
+				return nil, err
+			}
+
+			// Create a blob of the attributes that are optionally
+			// present in the "t" tree's .gitattributes blob, and
+			// union in the patterns that we've tracked.
+			//
+			// Perform this Union() operation each time we visit a
+			// root tree such that if the underlying .gitattributes
+			// is present and has a diff between commits in the
+			// range of commits to migrate, those changes are
+			// preserved.
+			blob, err := trackedToBlob(db, theirs.Clone().Union(ours))
+			if err != nil {
+				return nil, err
+			}
+
+			// Finally, return a copy of the tree "t" that has the
+			// new .gitattributes file included/replaced.
+			return t.Merge(&odb.TreeEntry{
+				Name:     ".gitattributes",
+				Filemode: 0100644,
+				Oid:      blob,
+			}), nil
+		},
+
+		UpdateRefs: true,
+	})
+
+	// Only perform `git-checkout(1) -f` if the repository is
+	// non-bare.
+	if bare, _ := git.IsBare(); !bare {
+		t := l.Waiter("migrate: checkout")
+		err := git.Checkout("", nil, true)
+		t.Complete()
+
+		if err != nil {
+			ExitWithError(err)
+		}
+	}
+}
+
+// trackedFromExportFilter returns an ordered set of strings where each entry is a
+// line in the .gitattributes file. It adds/removes the fiter/diff/merge=lfs
+// attributes based on patterns included/excluded in the given filter. Since
+// `migrate export` removes files from Git LFS, it will remove attributes for included
+// files, and add attributes for excluded files
+func trackedFromExportFilter(filter *filepathfilter.Filter) *tools.OrderedSet {
+	tracked := tools.NewOrderedSet()
+
+	for _, include := range filter.Include() {
+		tracked.Add(fmt.Sprintf("%s text -filter -merge -diff", escapeAttrPattern(include)))
+	}
+
+	for _, exclude := range filter.Exclude() {
+		tracked.Add(fmt.Sprintf("%s filter=lfs diff=lfs merge=lfs -text", escapeAttrPattern(exclude)))
+	}
+
+	return tracked
+}
--- a/commands/command_migrate_export.go
+++ b/commands/command_migrate_export.go
@@ -37,7 +37,7 @@ func migrateExportCommand(cmd *cobra.Command, args []string) {
 	tracked := trackedFromExportFilter(filter)
 	gitfilter := lfs.NewGitFilter(cfg)
 
-	migrate(args, rewriter, l, &githistory.RewriteOptions{
+	opts := &githistory.RewriteOptions{
 		Verbose:           migrateVerbose,
 		ObjectMapFilePath: objectMapFilePath,
 		BlobFn: func(path string, b *odb.Blob) (*odb.Blob, error) {
@@ -92,7 +92,35 @@ func migrateExportCommand(cmd *cobra.Command, args []string) {
 		},
 
 		UpdateRefs: true,
-	})
+	}
+
+	requireInRepo()
+
+	opts, err = rewriteOptions(args, opts, l)
+	if err != nil {
+		ExitWithError(err)
+	}
+
+	// If we have a valid remote, pre-download all objects using the Transfer Queue
+	if remoteURL := getAPIClient().Endpoints.RemoteEndpoint("download", cfg.Remote()).Url; remoteURL != "" {
+		q := newDownloadQueue(getTransferManifestOperationRemote("Download", cfg.Remote()), cfg.Remote())
+		if err := rewriter.ScanForPointers(q, opts); err != nil {
+			ExitWithError(err)
+		}
+
+		q.Wait()
+
+		for _, err := range q.Errors() {
+			if err != nil {
+				ExitWithError(err)
+			}
+		}
+	}
+
+	// Perform the rewrite
+	if _, err := rewriter.Rewrite(opts); err != nil {
+		ExitWithError(err)
+	}
 
 	// Only perform `git-checkout(1) -f` if the repository is
 	// non-bare.
--- a/commands/command_migrate_export.go
+++ b/commands/command_migrate_export.go
@@ -147,7 +147,7 @@ func trackedFromExportFilter(filter *filepathfilter.Filter) *tools.OrderedSet {
 	tracked := tools.NewOrderedSet()
 
 	for _, include := range filter.Include() {
-		tracked.Add(fmt.Sprintf("%s text -filter -merge -diff", escapeAttrPattern(include)))
+		tracked.Add(fmt.Sprintf("%s text !filter !merge !diff", escapeAttrPattern(include)))
 	}
 
 	for _, exclude := range filter.Exclude() {
--- a/commands/command_migrate_export.go
+++ b/commands/command_migrate_export.go
@@ -1,7 +1,6 @@
 package commands
 
 import (
-	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -45,15 +44,20 @@ func migrateExportCommand(cmd *cobra.Command, args []string) {
 				return b, nil
 			}
 
-			var buf bytes.Buffer
-
-			if _, err := smudge(gitfilter, &buf, b.Contents, path, false, rewriter.Filter()); err != nil {
+			ptr, err := lfs.DecodePointer(b.Contents)
+			if errors.IsNotAPointerError(err) {
+				return b, nil
+			}
+			if err != nil {
 				return nil, err
 			}
 
-			return &odb.Blob{
-				Contents: &buf, Size: int64(buf.Len()),
-			}, nil
+			downloadPath, err := gitfilter.ObjectPath(ptr.Oid)
+			if err != nil {
+				return nil, err
+			}
+
+			return odb.NewBlobFromFile(downloadPath)
 		},
 
 		TreeCallbackFn: func(path string, t *odb.Tree) (*odb.Tree, error) {
--- a/commands/command_migrate_export.go
+++ b/commands/command_migrate_export.go
@@ -122,8 +122,7 @@ func migrateExportCommand(cmd *cobra.Command, args []string) {
 		ExitWithError(err)
 	}
 
-	// Only perform `git-checkout(1) -f` if the repository is
-	// non-bare.
+	// Only perform `git-checkout(1) -f` if the repository is non-bare.
 	if bare, _ := git.IsBare(); !bare {
 		t := l.Waiter("migrate: checkout")
 		err := git.Checkout("", nil, true)
@@ -135,11 +134,11 @@ func migrateExportCommand(cmd *cobra.Command, args []string) {
 	}
 }
 
-// trackedFromExportFilter returns an ordered set of strings where each entry is a
-// line in the .gitattributes file. It adds/removes the fiter/diff/merge=lfs
-// attributes based on patterns included/excluded in the given filter. Since
-// `migrate export` removes files from Git LFS, it will remove attributes for included
-// files, and add attributes for excluded files
+// trackedFromExportFilter returns an ordered set of strings where each entry
+// is a line we intend to place in the .gitattributes file. It adds/removes the
+// filter/diff/merge=lfs attributes based on patterns included/excluded in the
+// given filter. Since `migrate export` removes files from Git LFS, it will
+// remove attributes for included files, and add attributes for excluded files
 func trackedFromExportFilter(filter *filepathfilter.Filter) *tools.OrderedSet {
 	tracked := tools.NewOrderedSet()
 
--- a/commands/command_migrate_export.go
+++ b/commands/command_migrate_export.go
@@ -108,9 +108,21 @@ func migrateExportCommand(cmd *cobra.Command, args []string) {
 	// If we have a valid remote, pre-download all objects using the Transfer Queue
 	if remoteURL := getAPIClient().Endpoints.RemoteEndpoint("download", cfg.Remote()).Url; remoteURL != "" {
 		q := newDownloadQueue(getTransferManifestOperationRemote("Download", cfg.Remote()), cfg.Remote())
-		if err := rewriter.ScanForPointers(q, opts, gitfilter); err != nil {
-			ExitWithError(err)
-		}
+		gs := lfs.NewGitScanner(func(p *lfs.WrappedPointer, err error) {
+			if err != nil {
+				return
+			}
+
+			downloadPath, err := gitfilter.ObjectPath(p.Oid)
+			if err != nil {
+				return
+			}
+
+			if _, err := os.Stat(downloadPath); os.IsNotExist(err) {
+				q.Add(p.Name, downloadPath, p.Oid, p.Size)
+			}
+		})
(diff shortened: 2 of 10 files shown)


commands/command_migrate.go
```diff
@@ -297,6 +297,8 @@ func init() {
 	importCmd.Flags().BoolVar(&migrateNoRewrite, "no-rewrite", false, "Add new history without rewriting previous")
 	importCmd.Flags().StringVarP(&migrateCommitMessage, "message", "m", "", "With --no-rewrite, an optional commit message")
 
+	exportCmd := NewCommand("export", migrateExportCommand)
```
ttaylorr: I think that it might make sense to bring in a few of the flags from `import`, namely: `--verbose` and `--object-map`. I think that we can leave `--no-rewrite` off for now, but it might make sense to bring that in later.

commands/command_migrate.go
```diff
@@ -306,6 +308,6 @@ func init() {
 		cmd.PersistentFlags().BoolVar(&migrateEverything, "everything", false, "Migrate all local references")
 		cmd.PersistentFlags().BoolVar(&migrateSkipFetch, "skip-fetch", false, "Assume up-to-date remote references.")
 
-		cmd.AddCommand(importCmd, info)
+		cmd.AddCommand(importCmd, info, exportCmd)
```
ttaylorr: Nit; let's put `exportCmd` before the other two arguments, to keep this list alphabetized.

commands/command_migrate_export.go
```diff
@@ -0,0 +1,155 @@
...
+		t.Complete()
+
+		if err != nil {
+			ExitWithError(err)
+		}
+	}
+}
+
+// trackedFromExportFilter returns an ordered set of strings where each entry is a
+// line in the .gitattributes file. It adds/removes the fiter/diff/merge=lfs
```
ttaylorr: Could you clarify this line? I think the last sentence in this block makes the intended purpose of the function clear, but the term `.gitattributes` I am unsure of in this context alone.

Maybe something like:

> "... where each entry is a line we intend to place in the repository's .gitattributes file."

commands/command_migrate_export.go
```diff
@@ -0,0 +1,155 @@
...
+	filter := rewriter.Filter()
+	if len(filter.Include()) <= 0 {
+		ExitWithError(errors.Errorf("fatal: one or more files must be specified with --include"))
+	}
+
+	tracked := trackedFromExportFilter(filter)
+	gitfilter := lfs.NewGitFilter(cfg)
+
+	opts := &githistory.RewriteOptions{
+		Verbose:           migrateVerbose,
```
ttaylorr: I think this and the `ObjectMapFilePath` are good to have, but presently don't have any effect, since the flags are not added to the `export` command.

commands/command_migrate_export.go
```diff
@@ -0,0 +1,155 @@
...
+		ObjectMapFilePath: objectMapFilePath,
+		BlobFn: func(path string, b *odb.Blob) (*odb.Blob, error) {
+			if filepath.Base(path) == ".gitattributes" {
+				return b, nil
+			}
+
+			var buf bytes.Buffer
```
ttaylorr: I think we may run into a problem where the size of the contents that we're smudging are larger than what we would like to hold in memory. Since `buf` is local to the stack frame, we'll have to worry about that here.

Since we already do a pass over history below with `ScanForPointers`, I think we can avoid smudging here, since the large object that we're looking for is present in the `.git/lfs/objects` cache.

So, we can use the `MediaPath` function on `gitfilter` to find the location of this file on disk. We'll have to open that file and then pass the `*os.File` (which implements `io.Reader`) to the `*odb.Blob` constructor's `Contents` field.

We'll need to close the file later, but something like this could work:

```diff
diff --git a/git/odb/blob.go b/git/odb/blob.go
index 7f9d66ae..e9cc52f8 100644
--- a/git/odb/blob.go
+++ b/git/odb/blob.go
@@ -3,6 +3,9 @@ package odb
 import (
 	"bytes"
 	"io"
+	"os"
+
+	"github.com/git-lfs/git-lfs/errors"
 )
 
 // Blob represents a Git object of type "blob".
@@ -19,6 +22,42 @@ type Blob struct {
 	closeFn func() error
 }
 
+// NewBlobFromFile returns a new *Blob that contains the contents of the file
+// at location "path" on disk. NewBlobFromFile does not read the file ahead of
+// time, and instead defers this task until encoding the blob to the object
+// database.
+//
+// If the file cannot be opened or stat(1)-ed, an error will be returned.
+//
+// When the blob receives a function call Close(), the file will also be closed,
+// and any error encountered in doing so will be returned from Close().
+func NewBlobFromFile(path string) (*Blob, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, errors.Wrapf(err, "git/odb: could not open: %s",
+			path)
+	}
+
+	stat, err := f.Stat()
+	if err != nil {
+		return nil, errors.Wrapf(err, "git/odb: could not stat %s",
+			path)
+	}
+
+	return &Blob{
+		Contents: f,
+		Size:     stat.Size(),
+
+		closeFn: func() error {
+			if err := f.Close(); err != nil {
+				return errors.Wrapf(err,
+					"git/odb: could not close %s", path)
+			}
+			return nil
+		},
+	}, nil
+}
+
 // NewBlobFromBytes returns a new *Blob that yields the data given.
 func NewBlobFromBytes(contents []byte) *Blob {
 	return &Blob{
```

commands/command_migrate_export.go
```diff
@@ -0,0 +1,155 @@
...
+
+	requireInRepo()
+
+	opts, err = rewriteOptions(args, opts, l)
+	if err != nil {
+		ExitWithError(err)
+	}
+
+	// If we have a valid remote, pre-download all objects using the Transfer Queue
+	if remoteURL := getAPIClient().Endpoints.RemoteEndpoint("download", cfg.Remote()).Url; remoteURL != "" {
```
ttaylorr: Nice. What do you think about allowing `[--remote=<remote>]` as an optional argument in this command, to specify where clients would like to get their objects from?

We could even take it a step further and allow them to specify multiple remotes. Package `tq` today doesn't know how to interact with multiple remotes at once (and I'm inclined not to teach it how to do so), but we could work around this by looping through each given remote. Maybe something like:

```go
var wg sync.WaitGroup

for _, remote := range remotes {
        wg.Add(1)
        go downloadObjects(remotes, wg)
}

wg.Wait()
```

commands/command_migrate_export.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+		}
+	}
+
+	// Perform the rewrite
+	if _, err := rewriter.Rewrite(opts); err != nil {
+		ExitWithError(err)
+	}
+
+	// Only perform `git-checkout(1) -f` if the repository is
```
ttaylorr: Nit; I think that we can re-wrap this line here and in `git lfs migrate import` (though perhaps the later should go in a preparatory commit).

commands/command_migrate_export.go
```diff
@@ -0,0 +1,155 @@
...
+// trackedFromExportFilter returns an ordered set of strings where each entry is a
+// line in the .gitattributes file. It adds/removes the fiter/diff/merge=lfs
+// attributes based on patterns included/excluded in the given filter. Since
+// `migrate export` removes files from Git LFS, it will remove attributes for included
+// files, and add attributes for excluded files
+func trackedFromExportFilter(filter *filepathfilter.Filter) *tools.OrderedSet {
+	tracked := tools.NewOrderedSet()
+
+	for _, include := range filter.Include() {
+		tracked.Add(fmt.Sprintf("%s text -filter -merge -diff", escapeAttrPattern(include)))
```
ttaylorr: Nice. I think that we are supposed to spell the `-filter` as `!filter`, though.

git/githistory/rewriter.go
```diff
@@ -172,6 +175,33 @@ func NewRewriter(db *odb.ObjectDatabase, opts ...rewriterOption) *Rewriter {
 	return rewriter
 }
 
+// ScanForPointers scans through the range of commits given by
```
ttaylorr: Nice. Historically, we've used package `lfs`'s `GitScanner` type to do such a thing. It's backed by the now-outdated `git/git.go`, which shells out to `git(1)`, but it does roughly what you're doing here. I'd like someday to replace it with a `git/odb`-backed implementation, but for now that doesn't exist. Instead of growing two separate implementations, how about we lean on what's already provided and replace it over time in a later PR?

```diff
diff --git a/commands/command_migrate_export.go b/commands/command_migrate_export.go
index b234eac9..53e64f49 100644
--- a/commands/command_migrate_export.go
+++ b/commands/command_migrate_export.go
@@ -104,10 +104,15 @@ func migrateExportCommand(cmd *cobra.Command, args []string) {
 	// If we have a valid remote, pre-download all objects using the Transfer Queue
 	if remoteURL := getAPIClient().Endpoints.RemoteEndpoint("download", cfg.Remote()).Url; remoteURL != "" {
 		q := newDownloadQueue(getTransferManifestOperationRemote("Download", cfg.Remote()), cfg.Remote())
-		if err := rewriter.ScanForPointers(q, opts, gitfilter); err != nil {
+		scan := lfs.NewGitScanner(func(p *lfs.WrappedPointer, err error) {
+			if err != nil {
+				return
+			}
+			q.Add(p.Name, p.Name, p.Oid, p.Size)
+		})
+		if err := scan.ScanRefComplex(opts.Include, opts.Exclude, nil); err != nil {
 			ExitWithError(err)
 		}
-
 		q.Wait()
 
 		for _, err := range q.Errors() {
diff --git a/lfs/gitscanner.go b/lfs/gitscanner.go
index 15a6d58c..20d6e49d 100644
--- a/lfs/gitscanner.go
+++ b/lfs/gitscanner.go
@@ -92,6 +92,17 @@ func (s *GitScanner) ScanLeftToRemote(left string, cb GitScannerFoundPointer) er
 	return scanRefsToChan(s, callback, left, "", s.opts(ScanLeftToRemoteMode))
 }
 
+func (s *GitScanner) ScanRefComplex(include, exclude []string, cb GitScannerFoundPointer) error {
+	callback, err := firstGitScannerCallback(cb, s.FoundPointer)
+	if err != nil {
+		return err
+	}
+
+	opts := s.opts(ScanRefsMode)
+	opts.SkipDeletedBlobs = false
+	return scanRefsToChan(s, callback, include, exclude, opts)
+}
+
 // ScanRefRange scans through all commits from the given left and right refs,
 // including git objects that have been modified or deleted.
 func (s *GitScanner) ScanRefRange(left, right string, cb GitScannerFoundPointer) error {
@@ -102,7 +113,7 @@ func (s *GitScanner) ScanRefRange(left, right string, cb GitScannerFoundPointer)
 
 	opts := s.opts(ScanRefsMode)
 	opts.SkipDeletedBlobs = false
-	return scanRefsToChan(s, callback, left, right, opts)
+	return scanLeftRightToChan(s, callback, left, right, opts)
 }
 
 // ScanRefWithDeleted scans through all objects in the given ref, including
diff --git a/lfs/gitscanner_refs.go b/lfs/gitscanner_refs.go
index e66daee2..024ff117 100644
--- a/lfs/gitscanner_refs.go
+++ b/lfs/gitscanner_refs.go
@@ -33,10 +33,7 @@ func (s *lockableNameSet) Check(blobSha string) (string, bool) {
 
 func noopFoundLockable(name string) {}
 
-// scanRefsToChan takes a ref and returns a channel of WrappedPointer objects
-// for all Git LFS pointers it finds for that ref.
-// Reports unique oids once only, not multiple times if >1 file uses the same content
-func scanRefsToChan(scanner *GitScanner, pointerCb GitScannerFoundPointer, refLeft, refRight string, opt *ScanRefsOptions) error {
+func scanRefsToChan(scanner *GitScanner, pointerCb GitScannerFoundPointer, include, exclude []string, opt *ScanRefsOptions) error {
 	if opt == nil {
 		panic("no scan ref options")
 	}
@@ -134,3 +131,10 @@ func revListShas(include, exclude []string, opt *ScanRefsOptions) (*StringChanne
 
 	return NewStringChannelWrapper(revs, errs), nil
 }
+
+// scanLeftRightToChan takes a ref and returns a channel of WrappedPointer objects
+// for all Git LFS pointers it finds for that ref.
+// Reports unique oids once only, not multiple times if >1 file uses the same content
+func scanLeftRightToChan(scanner *GitScanner, pointerCb GitScannerFoundPointer, refLeft, refRight string, opt *ScanRefsOptions) error {
+	return scanRefsToChan(scanner, pointerCb, []string{refLeft}, []string{refRight}, opt)
+}
```

PastelMobileSuit:
@ttaylorr Thanks for the review! In regards to "smarter .gitattributes modification", I mean some of the things discussed in #3057, such as removing redundant patterns from the .gitattributes file.

commands/command_migrate_export.go
PastelMobileSuit: According to the [gitattributes documentation](https://git-scm.com/docs/gitattributes), it looks like `-` sets the value of an attribute to `false`, whereas `!` sets the value of an attribute to `unspecified`. Since we're doing an export here, it seems like it might be better to use `-`

ttaylorr: That's what it looks like to me, too:

https://github.com/git/git/blob/53f9a3e157dbbc901a02ac2c73346d375e24978c/attr.c#L334-L335

But, I think that `ATTR__UNSET` is more appropriate; according to the merge documentation from that same page:

> ### Unset
> Take the version from the current branch as the tentative merge result, and
> declare that the merge has conflicts. This is suitable for binary files that
> do not have a well-defined merge semantics.
>
> ### Unspecified
> By default, this uses the same built-in 3-way merge driver as is the case when
> the `merge` attribute is set. However, the `merge.default` configuration
> variable can name different merge driver to be used with paths for which the
> `merge` attribute is unspecified.

I think that given the choice between the two, we'd prefer unspecified, which defaults back to the configuration.

PastelMobileSuit: Ah, that makes sense! In that case I'll go ahead and switch this over

git/githistory/rewriter.go
PastelMobileSuit: Thanks! I agree with you, we should be using the `GitScanner` for this code - I'll go ahead and replace what I've done with the `ScanForPointers` function

commands/command_migrate_export.go
PastelMobileSuit: I think that's a good idea! We can just default to `cfg.Remote()` like we do here if the argument isn't provided.

Do we want to support multiple remotes for this? What would the use cases for that look like? I'm concerned about potential conflicts in object downloads, for example if one provided remote points to a fork of another repo, which is also provided as a remote.

PastelMobileSuit:
@ttaylorr I've pushed up an additional set of commits to address most of your feedback! Still to come:
* Cache Pruning
* A remote flag
* Docs

commands/command_migrate_export.go
```diff
@@ -45,15 +46,26 @@ func migrateExportCommand(cmd *cobra.Command, args []string) {
 				return b, nil
 			}
 
-			var buf bytes.Buffer
+			ptr, err := lfs.DecodePointer(b.Contents)
+			if errors.IsNotAPointerError(err) {
```
ttaylorr: Nit; let's push this down into the below conditional, since an error will never be `nil` and pass `errors.IsNotAPointerError(...)`.

commands/command_migrate_export.go
ttaylorr: I think that supporting just the one remote is fine for now; I can see this becoming somewhat complicated because the transfer queue really isn't designed for multi-remote operations.

I think that there are a few approaches we can take in the longer term, but for now, I think that supporting one remote via `--remote` is totally good. If a user is (1) the bleeding-edge version of Git LFS, and (2) wants to do a multi-remote export, they can always use `git lfs fetch && git lfs migrate export` instead.

Longer-term, I think we could do one of the following:

1. Teach package `tq` how to handle multiple remotes. This would provide a consistent interface for using the transfer queue, but will be only used at this singular call-site, and seems too complicated for too little benefit.

2. Keep many transfer queues alive, and round-robin the objects through them. This has the benefit of simplicity and no requirements for change inside of package `tq`, and the external call-sites should be fairly straightforward, too. If an earlier transfer queue has the same object as a later one, it doesn't really matter which one answers first. This has the drawback of having to wait for one transfer queue to cycle before the next one in line can determine which objects they need to get.

I think that we could theoretically get _more_ efficient than that, but I bet that that's an OK starting point for now.

commands/command_migrate_export.go
```diff
@@ -104,9 +116,25 @@ func migrateExportCommand(cmd *cobra.Command, args []string) {
...
+			if !filter.Allows(p.Name) {
+				return
+			}
+
+			downloadPath, err := gitfilter.ObjectPath(p.Oid)
+			if err != nil {
+				return
+			}
+
+			if _, err := os.Stat(downloadPath); os.IsNotExist(err) {
```
ttaylorr: Nice. I wonder if we could add this to the `Add()` function on the transfer queue itself. I can imagine that other callers would find this useful.

commands/command_migrate_export.go
```diff
@@ -122,8 +150,19 @@ func migrateExportCommand(cmd *cobra.Command, args []string) {
 		ExitWithError(err)
 	}
 
-	// Only perform `git-checkout(1) -f` if the repository is
-	// non-bare.
+	// Prune the cache
+	for _, object := range exported {
```
ttaylorr: Will `exported` ever contain an object evicted by the migration, but used elsewhere on disk? An example demonstrating this is a repository with two files, both tracked in LFS, sharing the same contents at different paths. If only one of those paths is migrated out, the other will still wish to retain the object.

Perhaps we could invoke the `git lfs prune` command here, instead?

lfs/gitscanner.go
```diff
@@ -89,7 +89,20 @@ func (s *GitScanner) ScanLeftToRemote(left string, cb GitScannerFoundPointer) er
 	}
 	s.mu.Unlock()
 
-	return scanRefsToChan(s, callback, left, "", s.opts(ScanLeftToRemoteMode))
+	return scanLeftRightToChan(s, callback, left, "", s.opts(ScanLeftToRemoteMode))
+}
+
+// ScanRefs through all commits reachable by refs contained in "include" and
+// not reachable by any refs included in "excluded"
+func (s *GitScanner) ScanRefs(include, exclude []string, cb GitScannerFoundPointer) error {
```
ttaylorr: ✨ Nice.

lfs/gitscanner_refs.go
```diff
@@ -91,6 +92,13 @@ func scanRefsToChan(scanner *GitScanner, pointerCb GitScannerFoundPointer, refLe
 	return nil
 }
 
+// scanLeftRightToChan takes a ref and returns a channel of WrappedPointer objects
+// for all Git LFS pointers it finds for that ref.
+// Reports unique oids once only, not multiple times if >1 file uses the same content
+func scanLeftRightToChan(scanner *GitScanner, pointerCb GitScannerFoundPointer, refLeft, refRight string, opt *ScanRefsOptions) error {
+	return scanRefsToChan(scanner, pointerCb, []string{refLeft, refRight}, nil, opt)
```
ttaylorr: Hm, should this be:

```go
return scanRefsToChan(..., []string{refLeft}, []string{refRight}, nil, opt)
```

instead? I think that they should be split, because the right is a boundary of everything we don't want in the range.

PastelMobileSuit: I think this is correct as-is. My gut reaction was the same, but 

```
[]string{refLeft, refRight}, nil
```
was being used internally by `scanRefsToChan()` before I changed its functionality here. Additionally, changing it to 

```
[]string{refLeft}, []string{refRight},
```
causes tests that leverage this function to fail. I'm still doing some investigation as to why and will come back with an update when I better understand the rev scanner.

ttaylorr: Ah, you're totally right. I just checked the code and the way that it is in your diff is 👍, my comment is incorrect.

commands/command_migrate_export.go
PastelMobileSuit: I just tried this, but unfortunately it led to a bunch of inexplicable failures about objects not being found by the transfer queue when running the `export` tests. I verified that the conditional I added to the function is indeed correct so I'm not sure what's going wrong - it's definitely trying to download the files since they don't exist on disk. This problem bled over into other tests that leverage the transfer queue as well, such as `fetch`.

I can try to do some additional debugging but I'm inclined to leave this as-is for now. This is here to prevent download of an object that exists on disk but isn't present on the remote (say, if a recently-migrated pointer hasn't been pushed up to the remote repository yet). The transfer queue is leveraged by lots of code in Git LFS, and it seems like re-downloading an object file won't necessarily be wrong in all cases

ttaylorr: > I can try to do some additional debugging but I'm inclined to leave this as-is for now. 

That's fine, let's mark this as #leftoverbits, and touch this in a future pull request.

lfs/gitscanner_refs.go
PastelMobileSuit: @ttaylorr I've gone ahead and addressed your feedback! I've also added the `--remote` flag and docs

PastelMobileSuit:
@ttaylorr I've gone ahead and addressed your feedback! I've also added the --remote flag and docs

commands/command_migrate_export.go
```diff
@@ -159,17 +159,17 @@ func migrateExportCommand(cmd *cobra.Command, args []string) {
...
 
-		if err != nil {
-			ExitWithError(err)
-		}
-	}
+	// Set our preservation time-window for objects existing on the remote to
+	// 0. Because the newly rewritten commits have not yet been pushed, some
+	// exported objects can still exist on the remote within the time window
+	// and thus will not be pruned from the cache.
+	fetchPruneCfg.PruneOffsetDays = 0
```
PastelMobileSuit: I left a comment above but this was required to ensure certain objects were properly pruned. `prune` preserves any cached objects that exist on the remote from commits within a time window of `PruneOffsetDays + FetchRecentRefsDays`. Since `migrate export` rewrites commits but does not push them, exported objects can still exist on the remote, and can be retained in the cache if this time window is not set to 0.

A negative side effect of this is that _all_ objects that exist on the remote will be pruned, regardless of whether or not the associated object was exported. It seems like this should be okay though, as if necessary Git LFS can just re-download the pruned files.

commands/command_migrate_export.go
```diff
@@ -0,0 +1,203 @@
...
+		ExitWithError(err)
+	}
+	defer db.Close()
+
+	rewriter := getHistoryRewriter(cmd, db, l)
+
+	filter := rewriter.Filter()
+	if len(filter.Include()) <= 0 {
+		ExitWithError(errors.Errorf("fatal: one or more files must be specified with --include"))
+	}
```
ttaylorr: I like this error message, but I'm wondering what it is users would expect iff they didn't provide `--include` or `--exclude`. My intuition tells me that a user invoking `git lfs migrate export --everything` would want _everything_ to be removed from Git LFS, so perhaps a default of `--include=*` (under the semantics that are implemented below) would make sense.

What do you think?

commands/command_migrate_export.go
ttaylorr: With respect to my comment below, what happens if we remove these assignments iff we perform the prune _after_ we do the checkout?

test/test-migrate-export.sh
```diff
@@ -0,0 +1,432 @@
...
+
+  setup_single_local_branch_tracked
+
+  md_oid="$(calc_oid "$(cat a.md)")"
+  txt_oid="$(calc_oid "$(cat a.txt)")"
+
+  assert_pointer "refs/heads/master" "a.txt" "$txt_oid" "120"
+  assert_pointer "refs/heads/master" "a.md" "$md_oid" "140"
+
+  git lfs migrate export --include="*" --exclude="a.md"
```
ttaylorr: I'm not sure if this would be the place to put it, but I think that the `--include="*"` flag should _only_ be the default if no `--include` or `--exclude` argument is given, since I think writing patterns against an implicit `--include="*"` would be confusing for users.

test/test-migrate-export.sh
```diff
@@ -0,0 +1,432 @@
...
+
+begin_test "migrate export (bare repository)"
+(
+  set -e
+
+  setup_single_remote_branch_tracked
+  git push origin master
+  make_bare
+
+  git lfs migrate export --everything --include="*"
```
ttaylorr: Should there be any assertions following this? I think that the corresponding `migrate import` test is blank because it was simply fixing a crash, but we should check that the export works correctly here.

test/test-migrate-export.sh
```diff
@@ -0,0 +1,432 @@
...
+
+  git lfs migrate export --include="*" --remote="zz" 2>&1 | tee migrate.log
+  if [ ${PIPESTATUS[0]} -eq 0 ]; then
+    echo >&2 "fatal: expected git lfs migrate export to fail, didn't"
+    exit 1
+  fi
+
+  grep "fatal: invalid remote zz provided" migrate.log
+)
+end_test
```
ttaylorr: Nit; I think that this is missing a newline here.

commands/command_migrate_export.go
```diff
@@ -0,0 +1,203 @@
...
+			// Create a blob of the attributes that are optionally
+			// present in the "t" tree's .gitattributes blob, and
+			// union in the patterns that we've tracked.
+			//
+			// Perform this Union() operation each time we visit a
+			// root tree such that if the underlying .gitattributes
+			// is present and has a diff between commits in the
+			// range of commits to migrate, those changes are
+			// preserved.
+			blob, err := trackedToBlob(db, theirs.Clone().Union(ours))
```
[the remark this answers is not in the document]
PastelMobileSuit: It might if there was also no pre-existing `.gitattributes` file, but right now we require a pattern being given to `migrate export` through the `--include` argument, so it shouldn't be an issue.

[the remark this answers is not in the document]
ttaylorr: Thanks for the clarification, that sounds good to me.

commands/command_migrate_export.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// Set our preservation time-window for objects existing on the remote to
+	// 0. Because the newly rewritten commits have not yet been pushed, some
+	// exported objects can still exist on the remote within the time window
+	// and thus will not be pruned from the cache.
+	fetchPruneCfg.PruneOffsetDays = 0
+	fetchPruneCfg.FetchRecentRefsDays = 0
+
+	// Prune our cache
+	prune(fetchPruneCfg, false, false, true)
```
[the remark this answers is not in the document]
PastelMobileSuit: It doesn't seem to make a difference in regards to what gets pruned (all tests still pass if it's moved below the checkout), but I agree that after the checkout is the correct place for it! I'll go ahead and move it

commands/command_migrate_export.go
PastelMobileSuit: These values only affect whether to preserve pointers referenced in the remote git history, so they're still required if the prune is moved after the checkout

PastelMobileSuit: It looks like in general though we don't actually need to set `PruneOffsetDays`, just `FetchRecentRefsDays`, so I'll remove it

test/test-migrate-export.sh
```diff
@@ -0,0 +1,432 @@
...
+  assert_pointer "refs/heads/my-feature" "a.md" "$md_feature_oid" "30"
+  assert_pointer "refs/heads/my-feature" "a.txt" "$txt_oid" "120"
+  assert_pointer "refs/heads/master" "a.md" "$md_oid" "140"
+  assert_pointer "refs/heads/master" "a.txt" "$txt_oid" "120"
+
+  git lfs migrate export --include="*.md,*.txt" my-feature
+
+  [ ! $(assert_pointer "refs/heads/my-feature" "a.md" "$md_feature_oid" "30") ]
+  [ ! $(assert_pointer "refs/heads/my-feature" "a.txt" "$txt_oid" "120") ]
+  [ ! $(assert_pointer "refs/heads/master" "a.md" "$md_oid" "140") ]
```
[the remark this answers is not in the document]
PastelMobileSuit: I'm fairly certain that migrating a branch migrates all commits reachable from that branch. Since `my-feature` was branched off of the most recent commit in `master`, all commits on `master` are thus reachable from `my-feature`, and so are migrated.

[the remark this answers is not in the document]
ttaylorr: Ah, I see. I was aware that migrations worked like that, but I thought that `master` was present on a remote. Seems all good, thanks again for the explanation!

commands/command_migrate_export.go
PastelMobileSuit: Hmm, I don't think we should use `--everything` for this, as it's already being used to specify that all local references should be migrated.

Maybe we should add a separate flag, `--export-all`?

PastelMobileSuit:
@ttaylorr I've addressed/responded to your feedback! I think there are still one or two more open questions here but hopefully this is almost there!

commands/command_migrate_export.go
ttaylorr: I agree, though I wasn't suggesting that `--everything` be used for that purpose. Rather, that `git lfs migrate export` (with `--everything`) would be a common invocation, and that because it is a common invocation, we should also assume `--include="*"` if no `--include` or `--exclude` flag was given

test/test-migrate-export.sh
ttaylorr: Regarding my comment above, I was suggesting that if we decide on a default value of `"*"` for `--include`, it should _only_ be the default when neither `--include=` nor `--exclude=` is given.

commands/command_migrate_export.go
PastelMobileSuit: Ah okay! I see what you mean.

I mentioned this below but I'm a bit wary of this. I think it's easy to mistype `git lfs migrate export` when you mean to do something more specific, thus migrating everything in your repository. If your repository has lots of large files, this could also lead to a very large number of downloads.

Saying that, this is the way that `git lfs migrate import` works so it may be fine. If we were to implement it I'd like to do it the way `import` does it, i.e., maintaining a list of file extensions that it's done an operation on and adding attributes for each one.

ttaylorr: Hmm. Good point, I agree that it could be upsetting for users to download many GiB of LFS object data.

That said, I think that we should leave it as-is, since I don't want to mirror the behavior of `migrate import` here, which I think should have generated an error as you have implemented it here.

PR #3093, opened 2018-06-28
Encoded characters in pathnames

cbuehlmann:
### Symptom
`git lfs ls-files` lists uncommited files containing escaped runes in their pathname as **not added** even if they  will be added on commit.

````sh
git add "foo/*"
git status 
     new file:   "foo/Pr\303\244sentation.txt"
     new file:   foo/Regularname
git lfs ls-files
     70f78f04ee - "foo/Pr\303\244sentation.txt"   <-- Error! File should be marked with * like 'Regularname
     ae32d82ae1 * foo/Regularname
git commit
git lfs ls-files
     70f78f04ee * foo/Präsentation.pdf
     ae32d82ae1 * foo/Regularname
````

### Cause
`os.Stat(path)` in `commands/command_ls_files.go(114)` does not find files when the path contains escaped runes. For example: "foo/Pr\303\244sentation.txt"

### Fix
Parse escaped runes and replace with the corresponding byte value.

### Remaining Work
Check if needed in other places, where `lfs.WrappedPointer.Name` is used to access files.

### DIFF
--- a/commands/command_ls_files.go
+++ b/commands/command_ls_files.go
@@ -111,7 +111,8 @@ func lsFilesCommand(cmd *cobra.Command, args []string) {
 
 // Returns true if a pointer appears to be properly smudge on checkout
 func fileExistsOfSize(p *lfs.WrappedPointer) bool {
-	info, err := os.Stat(p.Name)
+	path := cfg.Filesystem().DecodePathname(p.Name)
+	info, err := os.Stat(path)
 	return err == nil && info.Size() == p.Size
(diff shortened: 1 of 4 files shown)


cbuehlmann:
Dear @ttaylorr; thank you for the quick response.

1. 

1. bla

n https://github.com/git-lfs/git-lfs/pull/3093/commits/21c33dc757d2e4008cccfa73df1037861e66400e
and 2

Issue #3105, opened 2018-07-04
git push does not abort if local LFS objects are missing


jokram:
## Problem Description
When some local LFS objects in `.git/lfs/objects/` are missing __(see below for such situations)__ then the `git push` will report that missing objects, but will continue the push operation:
```
$ git push -u origin master
Locking support detected on remote "origin". Consider enabling it with:
  $ git config lfs.https://git-id-test0.conti.de/cds-sqe-lfs/repo_missing_oid.git/info/lfs.locksverify true
LFS upload missing objects: (0/2), 0 B | 0 B/s
  (missing) LFS_File.xxx (06b936c7439c774eca9c8ae27b39311c6b9c9508cd4413f53ff52f4320913b2e)
Uploading LFS objects:   0% (0/2), 0 B | 0 B/s, done
Counting objects: 7, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (6/6), done.
Writing objects: 100% (7/7), 786 bytes | 393.00 KiB/s, done.
Total 7 (delta 1), reused 0 (delta 0)
remote: Resolving deltas: 100% (1/1), done.
To git-id-test0.conti.de:cds-sqe-lfs/repo_missing_oid.git
 * [new branch]      master -> master
Branch 'master' set up to track remote branch 'master' from 'origin'.
```
The git repository now exists on GitHub, but the referenced LFS file has not been uploaded => data inconsitency

## Possible Situations where local LFS objects are missing
1. User has deleted that object :fearful:
2. Fetch commits containing LFS objects from remote A and push these commits to remote B  
Here is a Perl script which reproduces this use case (you need to adapt some variable at the beginning):  
[ReproduceMissingOID.pl.txt](https://github.com/git-lfs/git-lfs/files/2163554/ReproduceMissingOID.pl.txt)
And here the output of that script:  
[ReproduceMissingOID.log](https://github.com/git-lfs/git-lfs/files/2163555/ReproduceMissingOID.log)
3. Fetch commits containing LFS objects from remote A, cherry-pick some of that commits and push these new cherry-picked commits to remote B  
This use case has led to ca. 200 missing LFS objects in one case :boom:. Fortunately these objects were already on GitHub from remote A.

## Requested Change
The Git LFS pre-push hook shall abort the `git push` operation if some Git LFS objects are missing to avoid that commits can be uploaded to GitHub without the corresponding Git LFS objects.

## Workaround
[...]

PR #3109, opened 2018-07-05
change lfs.allowincompletepush default from true to false 

larsxschneider:
Incomplete pushes can corrupt a repository. Therefore we should not allow them and fail hard by default.

See #3105


commands/uploader.go
```diff
@@ -289,6 +289,8 @@ func (c *uploadContext) ReportErrors() {
 		}
 
 		if !c.allowMissing {
+			Print("hint: Your push was rejected due to missing or corrupt local objects.\n" +
```
ttaylorr: I like this message, and I think that it's good that we're only showing it once at the end of the attempted push, right before calling os.Exit.

If you don't mind, I think that we could extract this message out to a `var` and print it by referencing that instead of having a string constant in the code. We use this pattern in the `clone` and `checkout` commands, which have deprecation notices.

I was thinking something like:

```go
var (
        pushMissingHint = []string{...}
)

if !c.allowMissing {
        Print(strings.Join(pushMissingHint, "\n"))
}
os.Exit(2)
```

commands/uploader.go
```diff
@@ -289,6 +289,11 @@ func (c *uploadContext) ReportErrors() {
 		}
 
 		if !c.allowMissing {
+			pushMissingHint := []string{
+				"hint: Your push was rejected due to missing or corrupt local objects.",
+				"hint: Overwrite the reject with `git config lfs.allowincompletepush true`.",
```
terrorobe: Wording nit: Maybe something along the lines of

> hint: You can disable this check using...

larsxschneider: Fixed 👍 
Following: https://github.com/git-lfs/git-lfs/blob/f5f57312610a86eb1fdb9f1cce4c071d589ccce2/commands/lockverifier.go#L66

PR #3111, opened 2018-07-06
Update github.com/spf13/cobra to latest tagged release

ssgelm:
github.com/spf13/cobra is now providing tagged releases so debian has pulled in 0.0.3 (the latest) as a package.  Consequently this PR also updates cobra (and also pflag since it's a dependency) to the latest tagged release.  While it may not add new functionality it seems to me that it is better as a whole to use tagged releases when available instead of arbitrary git commits.  One minor change that was required upstream is that they changed the way `help` works.  I put more details in that commit (in this PR) which I will also put below:

> This was initially broken upstream in spf13/cobra@d6bf4ef.  They tried to fix it with spf13/cobra@ded646f but functionality was not actually restored.  This takes the help command code from spf13/cobra@ded646f and explicitly sets it as the help command.  On one hand this is a bug that should arguably be fixed upstream, but on the other hand I believe this is now the way they intend people to use the `help` functionality.


ssgelm:
Hmm, not sure why that PR build is failing for the old git release... AFAICT nothing I have included this PR should affect `git-lfs smudge`

ttaylorr:
@ssgelm thanks for updating this package. I tried to do this in the last release (via #2967) but building the package in isolation via `docker/run_dockers.bsh` failed, so I (roughly) reverted the change in https://github.com/git-lfs/git-lfs/pull/3017. On this branch, I was able to observe the same failure in `docker/run_dockers.bsh`.

The failure indicates to me that the installed version of Go in Docker is unable to find all of the mentioned dependencies, which I assume to mean that Glide is having trouble installing all of the necessary dependencies. I can cause Glide to fail by running `glide get github.com/spf13/cobra` and selecting version `v0.0.3`.

I'm not sure what's going on with Glide, but it seems similar to https://github.com/Masterminds/glide/issues/995. I tried debugging a running execution of `glide` and couldn't figure out what was happening, since I am unfamiliar with the code.

I tried ignoring some packages (via the top-level `ignore:` declarative), but found that this wasn't sufficient either. I think that the crux of the problem is that there is a problem pulling in spf13/viper, which is now a dependency of spf13/cobra.

So, I think we can do one of a few things here:

1. Fork spf13/cobra and remove its dependency on Viper.

2. Fix Glide by sending a patch upstream or adding the right combination of `ignore:`-d dependencies to our `glide.yaml`. We could also start using another tool in favor of Glide, like `dep` or `vgo`, but it appears that `dep` is being phased out and `vgo` is not yet ready for general consumption.

3. Remove our dependency on `spf13/cobra` entirely, and write the commands in terms of Go's standard `flag` package.

I am most in favor of option (3), but I'd be curious to hear what your thoughts are.

ttaylorr:
@ssgelm I'm going to close this for now, but please feel free to resurrect it in the future if you are interested in doing so. Once Go 1.11 ships, I think that we ought to convert to Go Modules, which might make this point moot, anyway.

Thank you again :-).

bk2204:
I've opened #3478 to track this update. The PR as it stands won't apply to the current codebase, but a new PR could be opened.

PR #3112, opened 2018-07-06
README.md: Git for Windows ships LFS by default

larsxschneider:
See https://github.com/git-for-windows/build-extra/pull/110


larsxschneider:
@ttaylorr Agreed and fixed 👍

PR #3114, opened 2018-07-06
commands/migrate: infer wildmatches with --fixup

ttaylorr:
This pull request introduces a new flag for the `git lfs migrate import` mode, `--fixup`.

This is useful in situations where the state of a repository does not mirror what is expected from the .gitattributes file(s) present in that repository.

Instead of a longer description in the body of this pull request, I have opted to include instead a thorough walk-through in each commit.

##

/cc @git-lfs/core

### DIFF
--- a/commands/command_migrate_import.go
+++ b/commands/command_migrate_import.go
@@ -12,11 +12,13 @@ import (
 	"github.com/git-lfs/git-lfs/errors"
 	"github.com/git-lfs/git-lfs/filepathfilter"
 	"github.com/git-lfs/git-lfs/git"
+	"github.com/git-lfs/git-lfs/git/gitattr"
 	"github.com/git-lfs/git-lfs/git/githistory"
 	"github.com/git-lfs/git-lfs/lfs"
 	"github.com/git-lfs/git-lfs/tasklog"
 	"github.com/git-lfs/git-lfs/tools"
 	"github.com/git-lfs/gitobj"
+	"github.com/rubyist/tracerx"
 	"github.com/spf13/cobra"
 )
 
@@ -31,6 +33,10 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 	defer db.Close()
 
 	if migrateNoRewrite {
+		if migrateFixup {
+			ExitWithError(errors.Errorf("fatal: --no-rewrite and --fixup cannot be combined"))
+		}
+
 		if len(args) == 0 {
 			ExitWithError(errors.Errorf("fatal: expected one or more files with --no-rewrite"))
 		}
@@ -94,11 +100,20 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 		return
 	}
 
+	if migrateFixup {
+		include, exclude := getIncludeExcludeArgs(cmd)
+		if include != nil || exclude != nil {
+			ExitWithError(errors.Errorf("fatal: cannot use --fixup with --include, --exclude"))
+		}
+	}
+
 	rewriter := getHistoryRewriter(cmd, db, l)
 
 	tracked := trackedFromFilter(rewriter.Filter())
 	exts := tools.NewOrderedSet()
 	gitfilter := lfs.NewGitFilter(cfg)
+
+	var fixups *gitattr.Tree
 
 	migrate(args, rewriter, l, &githistory.RewriteOptions{
 		Verbose:           migrateVerbose,
@@ -106,6 +121,21 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 		BlobFn: func(path string, b *gitobj.Blob) (*gitobj.Blob, error) {
 			if filepath.Base(path) == ".gitattributes" {
 				return b, nil
+			}
+
+			if migrateFixup {
+				var ok bool
+				attrs := fixups.Applied(path)
+				for _, attr := range attrs {
+					tracerx.Printf("%+v\n", attr)
+					if attr.K == "filter" && attr.V == "lfs" {
+						ok = !attr.Unspecified
+					}
+				}
+
+				if !ok {
+					return b, nil
+				}
 			}
 
 			var buf bytes.Buffer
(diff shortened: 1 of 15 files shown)


commands/command_migrate_import.go
```diff
@@ -108,6 +122,20 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 				return b, nil
 			}
 
+			if migrateFixup {
+				var ok bool
+				attrs := fixups.Applied(path)
+				for _, attr := range attrs {
+					if attr.K == "filter" && attr.V == "lfs" {
+						ok = !attr.Unspecified
```
ttaylorr: Note-to-self: I think that this should actually read:

```go
if attr.K == "filter" {
        ok = attr.V == "lfs" && !attr.Unspecified
}
```

What do you think?

git/gitattr/tree_test.go
```diff
@@ -0,0 +1,246 @@
...
+	tree, err := New(db, &gitobj.Tree{Entries: []*gitobj.TreeEntry{
+		{
+			Name:     "child",
+			Oid:      child,
+			Filemode: 040000,
+		},
+	}})
+	require.NoError(t, err)
+
+	attrs := tree.Applied("child/foo.dat")
```
PastelMobileSuit: Since `child/foo.dat` would be matched by a `*.dat` pattern at the top level, it might be worth explicitly asserting here that that `tree` has no attributes, but one child with 4 attributes

git/gitattr/tree_test.go
```diff
@@ -0,0 +1,246 @@
...
+	tree, err := New(db, &gitobj.Tree{Entries: []*gitobj.TreeEntry{
+		{
+			Name:     "child",
+			Oid:      child,
+			Filemode: 040000,
+		},
+	}})
+	require.NoError(t, err)
+
+	attrs := tree.Applied("child/indirect/foo.dat")
```
PastelMobileSuit: Same thing here as above - this matches on a top-level `*.dat` pattern, so it would be good to manually check the attributes and children of the tree itself

git/gitattr/tree_test.go
```diff
@@ -0,0 +1,246 @@
...
+		{
+			Name:     "child",
+			Oid:      child,
+			Filemode: 040000,
+		},
+	}})
+	require.NoError(t, err)
+
+	assert.NotContains(t, tree.Children, "child")
+}
```
PastelMobileSuit: It might also be worth adding a test that when parsing a `.gitattributes` file, the `tree` successfully picks up multiple lines of attributes

test/test-migrate-fixup.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+
+. "test/test-migrate-fixtures.sh"
+. "test/testlib.sh"
+
+begin_test "migrate import (--fixup)"
```
PastelMobileSuit: For greater test coverage, it might be good to have a test involving a nested file and a nested .gitattributes file. That's already being tested in unit tests for each individual part utilized in the fixup, but I think it would be worthwhile to have a test here involving all those parts working together

git/githistory/rewriter_test.go
```diff
@@ -292,6 +293,76 @@ func TestRewriterAllowsAdditionalTreeEntries(t *testing.T) {
...
 
+func TestHistoryRewriterCallbacks(t *testing.T) {
+	type Call struct {
+		Type string
+		Path string
+	}
+
+	var calls []*Call
+
+	db := DatabaseFromFixture(t, "linear-history.git")
```
PastelMobileSuit: It might be good to have a nested subtree in here somewhere

commands/command_migrate_import.go
ttaylorr: Hm, I think that we were mostly right, 😄. If `attr.Unspecified` holds, that means that we implicitly set `attr.V` to `""` (which happens [here][1], and is documented by Git [here][2]).

So, I think we can avoid checking `attr.Unspecified` here, since if it is true, we know that `attr.V != "lfs"`. I implemented that change in https://github.com/git-lfs/git-lfs/pull/3114/commits/853e7fce3b2b7209914466967c95fc6440ef1c2c, which holds up as I expected.

[1]: https://github.com/git-lfs/git-lfs/pull/3114/files#diff-1a27d1eb67e3578c4441409b111fc10cR102
[2]: https://git-scm.com/docs/gitattributes#gitattributes-Unspecified

commands/command_migrate_import.go
```diff
@@ -123,8 +151,21 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
...
+				}
+				return nil
+			}
+			return nil
+		},
+
 		TreeCallbackFn: func(path string, t *gitobj.Tree) (*gitobj.Tree, error) {
-			if path != "/" {
+			if path != "/" || migrateFixup {
 				// Ignore non-root trees.
```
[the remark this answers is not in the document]
ttaylorr: Great call, thank you! My 👀 missed this one. I caught it here: https://github.com/git-lfs/git-lfs/pull/3114/commits/dbe212719559a326325466de9c2ab564b9f4e01c.

git/gitattr/tree_test.go
ttaylorr: Good call! Additionally, both `Lines` and `Children` are part of the public state, so we should check that they match what we expect them to. I did that in https://github.com/git-lfs/git-lfs/pull/3114/commits/ff62456c555e0396b24585ec907fd56784469e1d.

git/gitattr/tree_test.go
ttaylorr: Thanks, I caught this as well in https://github.com/git-lfs/git-lfs/commit/ff62456c555e0396b24585ec907fd56784469e1d.

git/gitattr/tree_test.go
ttaylorr: 🤔, I think I might recommend against this unless you feel strongly about it. The `New()` function only cares about discovering trees and their .gitattributes, not parsing them. The later is the responsibility of `ParseLines()`, which does have tests ensuring that multi-line `.gitattributes` are read correctly.

I suppose that we could add a test like that that tests the composability of the two, but I'm not sure that given our knowledge of how the code is implemented it would provide worthwhile benefit.

test/test-migrate-fixup.sh
ttaylorr: Good call, I caught this in https://github.com/git-lfs/git-lfs/pull/3114/commits/853e7fce3b2b7209914466967c95fc6440ef1c2c.

git/githistory/rewriter_test.go
ttaylorr: Thanks for the suggestion, I caught that in https://github.com/git-lfs/git-lfs/pull/3114/commits/3c574d0ba1e7530866993125918694a1864036e8 and https://github.com/git-lfs/git-lfs/pull/3114/commits/22801fdaeb64c8352b9834d88dd1a65377ad1874.

ttaylorr:
@PastelMobileSuit Thanks for your careful review! I really appreciated all of your comments, and I think that I've addressed all of them. Would you mind giving this another look?

Issue #3115, opened 2018-07-09
'alternates' support doesn't conform to git's repo layout specification


leith-bartrich:
Merge request #1007 adds support for a local reference for fetch operations.  Which is good.  I love it.  I would love to see that support expanded actually.

However, upon looking into it, I found it doesn't conform to git's specification.

Specifically, the "alternates" file is supposed to take multiple alternate paths on separate lines.  The code that was merged in to git-lfs, assumes only a single line.  An "alternates" file with multiple lines shows an empty "LocalReferenceDir=" when calling "git lfs env".  Which makes sense given the way it was implemented.  But this behavior does rail against the git repository layout doc.  https://git-scm.com/docs/gitrepository-layout

The easiest semi-fix would be to just take the first line, rather than the whole file.  But really, the support should be updated to extract and cascade down a list of alternates and use all the valid ones in order, on each object.  Ideally, in the same order git's internal code implements.

Further (And maybe this is  a separate issue) I'd think this is a great place to intelligently support local clones, pulls and fetches; by checking the remote's object path and adding it to the list of LocalReferenceDir(s) if it is in fact fetching from a local path.  Then, adding a flag to disable this behavior, or alternately a flag to enable the behavior.

leith-bartrich:
I've done a local checkout of git lfs and taught myself enough go to hack up the fix and change referenced above.

Local testing on my linux successfully executes local clones via a fixed version of the 'alternates' mechanism, where it's using multiple alternates, rather than just one. 

Is this something you'd want to see a pull request for?  Again, I'm no go expert.

ttaylorr:
@leith-bartrich Thank you so much for digging into this issue! I thought that I had already commented here, but I have already opened up #3116, and #3117 which I think (taken together) should fix the issue.

If you have time, would you mind trying them out for yourself? If they don't work as expected, please feel more than free to open up a pull request, or leave a comment on mine.

I'm sorry that I forgot to leave a note here that I added support for this, but I appreciate you taking the time out to learn some Go and work on Git LFS nonetheless.

leith-bartrich:
[...]

PR #3116, opened 2018-07-09
fs: support multiple object alternates

ttaylorr:
Like in upstream Git, Git LFS supports "object database alternates"
(described in [1]) to link several object databases together.

We have supported this feature since [2] but read the entire
.git/objects/info/alternates in one single buffer via ioutil.ReadFile
and checked if a directory existed at that location.

If a user specified multiple alternates, then this will fail. Git LFS
will look for a directory that has newlines in it, when what it should
be doing is splitting that list by the '\n' character and checking each
directory if it exists.

Let's do that, and also return a []string, instead of a single string,
to indicate that there might be multiple alternates in a repository.

Closes: https://github.com/git-lfs/git-lfs/issues/3115.

[1]: https://git-scm.com/docs/gitrepository-layout#gitrepository-layout-objectsinfoalternates
[2]: https://github.com/git-lfs/git-lfs/pull/1007/files

## 

/cc @git-lfs/core


lfs/lfs.go
```diff
@@ -98,13 +100,19 @@ func LinkOrCopyFromReference(cfg *config.Configuration, oid string, size int64)
...
 		return err
 	}
-	if altMediafile != "" && tools.FileExistsOfSize(altMediafile, size) {
-		return LinkOrCopy(cfg, altMediafile, mediafile)
+	for _, altMediafile := range altMediafiles {
+		tracerx.Printf("altMediafile: %s", altMediafile)
+		if altMediafile != "" && tools.FileExistsOfSize(altMediafile, size) {
+			err = LinkOrCopy(cfg, altMediafile, mediafile)
+			if err == nil {
+				break
```
PastelMobileSuit: Maybe we could `return err` here, and `return nil` at the end of the function? I think that makes it a bit more explicit what's happening here

test/test-alternates.sh
```diff
@@ -0,0 +1,93 @@
...
+    clone_repo "$reponame" "${reponame}_alternate"
+  popd > /dev/null
+
+  rm -rf .git/lfs/objects
+
+  alternate="$TRASHDIR/${reponame}_alternate/.git/objects"
+  echo "$alternate" > .git/objects/info/alternates
+
+  GIT_TRACE=1 git lfs fetch origin master 2>&1 | tee fetch.log
+  [ "0" -eq "$(grep -c "sending batch of size 1" fetch.log)" ]
```
PastelMobileSuit: Is there something else we can check in these tests to be more confident that the files were successfully downloaded?

lfs/lfs.go
ttaylorr: Thanks for the suggestion, I think that the existing implementation might be a little clearer. The loop is trying to call `LinkOrCopy()` with each alternate to find the contents of an object and returning the last error.

When we get a nil error here, there's nothing left to do (i.e., we found an alternate that has the right objects, so we can stop going further, hence `break`-ing the loop).

I think it's prudent to keep the `return err` at the end of the function, since it indicates that we'll return the last-returned error from the loop. IIRC, it would be incorrect to write `return nil` at the end of the function, since `err` is not guaranteed to be non-`nil` by the end of the function.

test/test-alternates.sh
ttaylorr: Unfortunately not, since what we're testing is that the objects _are_ downloaded, but not via a network request (i.e., they are absent from our repository's .git/lfs/objects, but were able to be found in another alternate).

ttaylorr:
@PastelMobileSuit Thanks for all of your feedback, I think that this is ready to merge.

lfs/lfs.go
PastelMobileSuit: Ah okay! That makes sense to me, in that case I agree we should leave this as-is

PastelMobileSuit:
:shipit:

PR #3117, opened 2018-07-09
fs/fs.go: support alternates from environment variable

ttaylorr:
The Git documentation also mentions that the
"$GIT_ALTERNATE_OBJECT_DIRECTORIES" environment variable is a valid way
to configure object database remotes [1]:

  2. You could be using the `objects/info/alternates` or
     `$GIT_ALTERNATE_OBJECT_DIRECTORIES` mechanisms to borrow objects
     from other object stores.

Let's support the same by using the last commit, taking in a handle on
the operating-system configuration and using that in addition to the
list found in .git/objects/info/alternates.

[1]: https://git-scm.com/docs/gitrepository-layout#gitrepository-layout-objectsinfoalternates

## 

/cc @git-lfs/core

### DIFF
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -18,6 +18,14 @@ import (
 
 var oidRE = regexp.MustCompile(`\A[[:alnum:]]{64}`)
 
+// Environment is a copy of a subset of the interface
+// github.com/git-lfs/git-lfs/config.Environment.
+//
+// For more information, see config/environment.go.
+type Environment interface {
+	Get(key string) (val string, ok bool)
+}
+
 // Object represents a locally stored LFS object.
 type Object struct {
 	Oid  string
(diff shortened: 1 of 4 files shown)


PastelMobileSuit:
@ttaylorr It think there are some compile errors on this PR stopping the CI Builds from going through

ttaylorr:
> @ttaylorr It think there are some compile errors on this PR stopping
> the CI Builds from going through

Thanks for pointing that out. I just pushed 7bcfa6d1 (fs/fs.go: fix some
compilation issues, 2018-07-16) to fix those compilation issues.

fs/fs.go
```diff
@@ -236,6 +236,34 @@ func resolveReferenceDirs(gitStorageDir string) []string {
 	return references
 }
 
+// existsAlternate takes an object directory given in "objs" (read as a single,
+// line from .git/objects/info/alternates). If that is a satisfiable alternates
+// directory (i.e., it exists), the directory is returned along with "true". If
+// not, the empty string and false is returned instead.
+func existsAlternate(objs string) (string, bool) {
+	objs = strings.TrimSpace(objs)
+	if strings.HasPrefix(objs, "#") {
```
PastelMobileSuit: I think this should be "\"" instead

ttaylorr: Oof, got that. Thanks.

PastelMobileSuit:
:shipit:

Issue #3102, opened 2018-07-04
File Lock blocks git lfs checkout from replacing files


FattyMieo:
git lfs checkout throws error `"Could not checkout <path>"` due to the files are read-only set by the FIle Lock. How can I resolve this problem?

ttaylorr:
Hi @FattyMieo, thanks for opening this. It seems like a bug to me, and I think that we have a couple of options on how to go forward. My inclination is to suggest that you use `git checkout` instead of `git lfs checkout`, since it is deprecated and will n ot be included in future releases starting after v3.0.0.

Alternatively, we could teach `git lfs checkout` to work around these, by first making the file writeable if it is not. That would look something like:

```diff
diff --git a/lfs/gitfilter_smudge.go b/lfs/gitfilter_smudge.go
index 0d1c0d1b..21886cb2 100644
--- a/lfs/gitfilter_smudge.go
+++ b/lfs/gitfilter_smudge.go
@@ -21,6 +21,19 @@ func (f *GitFilter) SmudgeToFile(filename string, ptr *Pointer, download bool, m
 		return fmt.Errorf("Could not create working directory file: %v", err)
 	}
 	defer file.Close()
+
+	stat, err := file.Stat()
+	if err != nil {
+		return fmt.Errorf("Could not stat working directory file: %v", err)
+	}
+
+	// When we're done, return the file back to its normal permission bits.
+	defer file.Chmod(stat.Mode())
+
+	if err := file.Chmod(stat.Mode() | 0200); err != nil {
+		return fmt.Errorf("Could not make file write-able for checkout")
+	}
+
 	if _, err := f.Smudge(file, ptr, filename, download, manifest, cb); err != nil {
 		if errors.IsDownloadDeclinedError(err) {
 			// write placeholder data instead
```

But it might not be worthwhile if we're going to remove the command anyway.

@larsxschneider @PastelMobileSuit what do you think?

PastelMobileSuit:
@FattyMieo Thanks for opening the issue! In regards to @ttaylorr's comment above, I think `git checkout` is the best path forward here, since `git lfs checkout` is currently deprecated. I've also done a quick test locally and `git checkout` seems to work if a file is locked.

FattyMieo:
Thank you for the replies. 

[...]

PR #3120, opened 2018-07-09
commands/checkout: checkout over read-only files

ttaylorr:
This pull request teaches the deprecated `git lfs checkout` command to
be able to checkout read-only files in the working copy.

If a file tracked with Git LFS is locked, it will not have the writing
bit unless the user themselves holds the lock. In the former case, if a
user wants to check out a locked Git LFS file--provided that they will
not push a changed version of that file--they should be allowed to check
that file out.

To do so, let's recognize when a file is marked as read-only,
temporarily grant it writing permissions, and then restore the original
permissions once we're done.

Closes: https://github.com/git-lfs/git-lfs/issues/3102.

##

/cc @git-lfs/core

### DIFF
--- a/lfs/gitfilter_smudge.go
+++ b/lfs/gitfilter_smudge.go
@@ -16,6 +16,18 @@ import (
 
 func (f *GitFilter) SmudgeToFile(filename string, ptr *Pointer, download bool, manifest *tq.Manifest, cb tools.CopyCallback) error {
 	os.MkdirAll(filepath.Dir(filename), 0755)
+
+	if stat, _ := os.Stat(filename); stat != nil && stat.Mode()&0200 == 0 {
+		if err := os.Chmod(filename, stat.Mode()|0200); err != nil {
+			return errors.Wrap(err,
+				"Could not restore write permission")
+		}
+
+		// When we're done, return the file back to its normal
+		// permission bits.
+		defer os.Chmod(filename, stat.Mode())
+	}
+
 	file, err := os.Create(filename)
 	if err != nil {
 		return fmt.Errorf("Could not create working directory file: %v", err)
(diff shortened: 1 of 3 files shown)


lfs/gitfilter_smudge.go
```diff
@@ -16,6 +16,18 @@ import (
 
 func (f *GitFilter) SmudgeToFile(filename string, ptr *Pointer, download bool, manifest *tq.Manifest, cb tools.CopyCallback) error {
 	os.MkdirAll(filepath.Dir(filename), 0755)
+
+	if stat, _ := os.Stat(filename); stat != nil && stat.Mode()&0200 == 0 {
```
PastelMobileSuit: Since this is being done in the smudge code, is it possible this will affect read-only files in other situations that we're not expecting?

lfs/gitfilter_smudge.go
```diff
@@ -16,6 +16,18 @@ import (
...
+		if err := os.Chmod(filename, stat.Mode()|0200); err != nil {
+			return errors.Wrap(err,
+				"Could not restore write permission")
+		}
+
+		// When we're done, return the file back to its normal
+		// permission bits.
+		defer os.Chmod(filename, stat.Mode())
```
PastelMobileSuit: Nice! ✨

lfs/gitfilter_smudge.go
ttaylorr: I think we should be OK here, since SmudgeToFile is only used to smudge into the working copy, which is the only place that files could be locked or otherwise marked as read-only. The more general-purpose `Smudge()` is what goes out to the process filter and et.

PastelMobileSuit: Thanks! In that case I am 👍

Issue #3118, opened 2018-07-09
migrate import --everything is missing refs


jtatum:
I started with a bare `--mirror` clone of a repo from GitLab (important because GitLab keeps refs to their version of pull requests in the repo). I then performed a `git lfs migrate import --everything --include="..."`. The output captures all `/refs/heads/*` and `/refs/tags/*` refs. It's missing over half of the `/refs/merge-requests/*/head` refs. LFS discovers/outputs 493 of 1055. For example, `/refs/merge-requests/1/head` and `/refs/merge-requests/1000/head` are present and migrated, but `/refs/merge-requests/10/head` and `/refs/merge-requests/100/head` are not. All the missing refs do appear in `git for-each-ref`. Once the migration is complete, there are git objects that should have been migrated but were apparently part of the refs that weren't discovered.

git version 2.18.0
git-lfs/2.4.0 (GitHub; darwin amd64; go 1.10)

jtatum:
Possibly related to #2984?

ttaylorr:
@jtatum thanks for the suggestion, I think that it is related to a similar problem as is described in #2984. I fixed this up in https://github.com/git-lfs/git-lfs/pull/3121.

jtatum:
I took a look at the change... I'm not sure why the logic decided that some of these refs are local at all. That code seems to be pretty broken for bare repositories. In any case, thanks for the change! I'll spin it up today and have a look.

jtatum:
@ttaylorr verified that the PR works. `--everything` sees all refs. Thanks so much for your quick response!

ttaylorr:
> @ttaylorr verified that the PR works. `--everything` sees all refs.
> Thanks so much for your quick response!

Thanks for taking a look; I'm glad that everything works as expected.

I've queued that pull request to be generally available in the next
release, which will be v2.5.0.

ttaylorr:
> I took a look at the change... I'm not sure why the logic decided that
> some of these refs are local at all.

Me either :-).

> That code seems to be pretty broken for bare repositories. In any
> case, thanks for the change! I'll spin it up today and have a look.

Thanks for taking a look. It's possible that there is unknown breakage
for bare repositories, but we try and do as best a job as possible at
detecting when a repository is bare, and acting appropriately (i.e., not
performing worktree operations iff true).

Please don't hesitate to let me know about any breakages, and I (or
another member of @git-lfs/core would be more than happy to help fix, or
guide a contributor through fixing them).

[...]

Issue #2984, opened 2018-04-25
git lfs migrate import appears to convert files, but doesn't delete original blobs


agunnerson-ibm:
We're running into an issue where `git lfs migrate` appears to be converting files to LFS (`.git/lfs` is the expected size), but it seems like the original objects aren't getting deleted.

For example, we'll mirror clone the repo from the original server and push it to the new one.

```
[andrew.gunnerson@<REDACTED> temp]$ git clone --mirror ssh://git@<INTERNAL_GIT_1>:7999/foo/bar.git
Cloning into bare repository 'bar.git'...
remote: Counting objects: 54635, done.
remote: Compressing objects: 100% (16394/16394), done.
remote: Total 54635 (delta 30592), reused 49397 (delta 28181)
Receiving objects: 100% (54635/54635), 260.24 MiB | 34.77 MiB/s, done.
Resolving deltas: 100% (30592/30592), done.

[andrew.gunnerson@<REDACTED> temp]$ cd bar.git/

[andrew.gunnerson@<REDACTED> bar.git]$ git push --mirror git@<INTERNAL_GIT_2>:andrew-gunnerson/bar-test.git
Counting objects: 54635, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (13983/13983), done.
Writing objects: 100% (54635/54635), 260.24 MiB | 43.56 MiB/s, done.
Total 54635 (delta 30592), reused 54635 (delta 30592)
remote: Resolving deltas: 100% (30592/30592), done.
remote: error: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.
remote: error: File a/b/c.jar is 123.23 MB; this exceeds GitHub Enterprise's file size limit of 100.00 MB
To <INTERNAL_GIT_2>:andrew-gunnerson/bar-test.git
 <REDACTED OTHER REFS>
 ! [remote rejected]   master -> master (pre-receive hook declined)
 <REDACTED OTHER REFS>
error: failed to push some refs to 'git@<INTERNAL_GIT_2>:andrew-gunnerson/bar-test.git'
```

The server rejects `a/b/c.jar`, so I move that to LFS. I know the files are successfully moved because the local `lfs` folder in the git dir is non-empty and has the correct size.

```
[andrew.gunnerson@<REDACTED> bar.git]$ git lfs install
Updated git hooks.
Git LFS initialized.

[...]

PR #3121, opened 2018-07-09
commands/command_migrate.go: loosen meaning of '--everything'

ttaylorr:
In [1] and [2], callers of the 'git lfs migrate' command were surprised
when `--everything` did not migrate over everything, as its name
implies.

In the documentation, we specified that `--everything` applies only to
local references, as the default behavior of 'git lfs migrate' is to
never require a force-push unless asked to do so explicitly.

--everything sounds dangerous enough that it would imply that a user
wants _everything_ in their repository migrated. So, let's loosen the
requirement and make it mean that.

Alternatively, we could change the meaning of `--everything` in this
fashion and replace it with `--everything-local`. We could also
introduce `--force`, leave the meaning of `--evrything` unchanged, and
only exhibit this behavior with `--everything --force`. Both of these
options add too much surface area and complexity for use cases that seem
less-common, and/or could be accomplished with clever `git for-each-ref`
and `xargs`-ing.

Closes: https://github.com/git-lfs/git-lfs/issues/2984.
Closes: https://github.com/git-lfs/git-lfs/issues/3118.

##

/cc @git-lfs/core 

[1]: #2984
[2]: #3118

### DIFF
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -173,13 +173,31 @@ func includeExcludeRefs(l *tasklog.Logger, args []string) (include, exclude []st
 		include = append(include, migrateIncludeRefs...)
 		exclude = append(exclude, migrateExcludeRefs...)
 	} else if migrateEverything {
-		localRefs, err := git.LocalRefs()
+		refs, err := git.AllRefsIn("")
 		if err != nil {
 			return nil, nil, err
 		}
 
-		for _, ref := range localRefs {
-			include = append(include, ref.Refspec())
+		for _, ref := range refs {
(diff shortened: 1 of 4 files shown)


ttaylorr:
@PastelMobileSuit Thanks for the suggestion! I think I started on this and then switched branches one too many times and forgot about it. I pushed https://github.com/git-lfs/git-lfs/pull/3121/commits/e311b20d71401180cbb770b83f5421141fa32bb2, which I think should do the trick.

PastelMobileSuit:
:shipit:

Benerdchia:
#3121

Issue #3123, opened 2018-07-10
Git lfs migrate just delete a whole bunch of unstaged file changes out of my repository.


Merglasch:
What it says above. The uncommited changes weren't even in files that git lfs should track. I'm going a bit bonkers here. Does the migrate command just delete all changes that are uncommitted and unrelated to tracked filetypes without warning? Can I somehow undo this mess? 

lsf shows no tracked files.

Git LFS objects to be pushed to origin/develop:
also none

Git LFS objects to be committed:
also none

Git LFS objects not staged for commit:
also none

size of the lfs folder: 230 mb.

used commands:

git lfs migrate import --include="*psd"
migrate: Sorting commits: ..., done
migrate: Rewriting commits: NaN% (0/0), done
migrate: Updating refs: ..., done

git lfs migrate import --include="*.psd"
migrate: Sorting commits: ..., done
migrate: Rewriting commits: NaN% (0/0), done
migrate: Updating refs: ..., done

git lfs migrate import --include="*.png"
migrate: Sorting commits: ..., done
migrate: Rewriting commits: NaN% (0/0), done
migrate: Updating refs: ..., done

It seems to have thrown away every change that wasn't a newly added file

I'd be really thankful for some help

esopalumpa:
Same here.
All files that are staged (added but not commited) get eliminiated by “git lfs migrate import …”

ttaylorr:
Hi @Merglasch, thanks for opening this. Thanks also @esopalumpa for confirming the issue.

> Does the migrate command just delete all changes that are uncommitted and unrelated to tracked filetypes without warning?

Yes, by design the `git lfs migrate` command uses `git checkout --force` to refresh the working copy after a migration, which does throw away local changes.

Unfortunately, without a backup there is no way to recover thees missing files. I am hesitant to change the requirement and enforce "your working copy must be clean" before migrating, since a common problem is that a user's working copy is unable to be made clean without invasive change and hence requires a migration.

Instead, what about a warning? Perhaps, "your working copy is modified, do you wish to continue?"

Merglasch:
For the love of god please change this.

git lfs migrate should either:
...not use git checkout --force at all. If it really is a requirement, then just prevent the migration and let the user do the force checkout themselves.
... offer a force option with that. Fail on a modified working copy without the force option, inform the user and request they clean their copy or try again with the force option.

[...]

PR #3124, opened 2018-07-10
commands: warn if working copy is dirty

ttaylorr:
In 'git lfs migrate import' and 'git lfs migrate export', Git LFS makes
destructive changes to a caller's repository and therefore invokes 'git
checkout --force', which throws away local changes.

To prevent this, let's introduce a check that notifies users when they
are going to throw away local changes, and allows for the user to abort
if they do not wish to discard their local changes.

With this, a user can safely migrate over dirty repositories (i.e., in
the case that they wanted to fix a file that should have been in Git
LFS, but wasn't) without having to finagle their repository to get it
into a migrate-able state.

For users with lots of pending migrations, also teach --yes, which
allows a user to avoid the check, and instead simply prints the warning
message to STDERR.

Closes: https://github.com/git-lfs/git-lfs/issues/3123.

##

/cc @git-lfs/core 
/cc https://github.com/git-lfs/git-lfs/issues/3123, @Merglasch, @esopalumpa


ttaylorr:
> This looks great! ✨ I think this needs a test that uses the
> `setup_local_branch_with_dirty_copy()` fixture, but aside from that I
> am 👍 ! I'm really glad we have this functionality now

Thanks! I added a few tests for this in 7862593c
(test/test-migrate-import.sh: test with dirty copy, 2018-07-16). This
doesn't test every case (namely: default answer, and 'y'), but I think
what's there is sufficient to go with.

Some of the other cases are a little more brittle (like sending newlines
without causing an EOF), so I'd rather leave them out instead of tending
to them.

PastelMobileSuit:
@ttaylorr Thanks for adding the additional tests! If you feel strongly about leaving it out then I'm fine with having this merged as-is, but I would prefer if we had a single test for a `y` answer. This is going to become the standard workflow when performing a migrate, so I think it would be good to have a test so we can know if it ever breaks.

ttaylorr:
> @ttaylorr Thanks for adding the additional tests! If you feel strongly
> about leaving it out then I'm fine with having this merged as-is, but
> I would prefer if we had a single test for a `y` answer.

I don't feel strongly; how does this look? 1bea9fd5
(test/test-migrate-import.sh: test dirty copy with 'y', 2018-07-16)

> This is going to become the standard workflow when performing a
> migrate, so I think it would be good to have a test so we can know if
> it ever breaks.

I agree, and this makes good sense to me in order to add a new test to
this effect. Let's do it! :-)

PastelMobileSuit:
:shipit:

ttaylorr:
Thanks for the :shipit:, I just resolved some merge conflicts and I'm
going to let CI cook this for a while and then will merge before v2.5.0.

PR #3125, opened 2018-07-13
t: run tests using prove

ttaylorr:
This pull request converts the test output in `t` (the directory formerly known as `test`) to the [TAP specification][1], and replaces `script/integration` with `t/Makefile` that runs [`prove(1)`][2].

We would like to move away from our home-grown integration scripts and towards something that more closely approximates the Git development environment, which should make it easier for contributors who are new to the Git LFS project (but perhaps familiar with the Git project) to get started.

This change does just that, by introducing support for the TAP specification and prove to run our integration tests. The changes included in this pull request are numerous, but following is a summary of the large themes:

- Rename directory `test` to `t`. `prove` expects test files that match the shell glob `t/*.t`.

  To approximate this, we introduce the `t` directory (to replace `test`), and prefix all files in that directory with `t-`. This is similar to git, which has tests such as `t/t7810-grep.sh`. Since our tests do not have numbering, this change replaces a file such as `test/test-migrate-import.sh` with` t/t-migrate-import.sh`.

- Replace `script/integration` (and `script/integration.go`) with `make test`.

  Instead of continuing to perpetuate our home-grown `script/integration.go`, let's replace it with `prove(1)`, a widely-available test harness that understands TAP output and test parallelism instead. We can conveniently introduce a `Makefile` at the same time, which is useful for only rebuilding things that have changed since the last test invocation.

- Introduce `t/cmd/lfstest-count-tests.go`.

  Previously, `test/testhelpers.sh` was responsible for invoking `setup`, a shell function that started a single `lfstest-git-server`, which is a Git server used during integration tests. This works sufficiently well when one test is running at a time, but is prone to race conditions when two tests overlap one another in runtime. (For example, if test A starts before B (and therefore, launches a Git server) B will not launch a Git server because one has already been started by A. If A finishes before B, the Git server will go away before B has finished testing, and thus cause B to fail).

  Instead, we now maintain two files `test_count`, and `test_count.lock`, which synchronize the number of actively running tests. When the test_count goes from `0` to `1`, the `lfstest-count-tests` binary starts a Git server running in the background. When the test_count goes from `1` back to `0`, the same binary stops that server as appropriate.

  To save time (and avoid a flapping git server), we implicitly set the count to `1` when running more than one test in sequence, and clean up after ourselves later. 

For more: see [t/README.md][3].

##

/cc @git-lfs/core 

[1]: https://testanything.org/tap-specification.html
[2]: https://perldoc.perl.org/prove.html
[3]: https://github.com/git-lfs/git-lfs/blob/ac1c5f120896733fa0df9306cf155e85573d78d7/t/README.md


t/cmd/lfstest-count-tests.go
```diff
@@ -16,12 +16,33 @@ import (
 )
 
 var (
+	// countFile is the path to a file (relative to the $LFSTEST_DIR) who's
```
PastelMobileSuit: The documentation in this file is great and makes everything really easy to understand! ✨

t/cmd/lfstest-count-tests.go
```diff
@@ -0,0 +1,270 @@
...
+				fmt.Sprintf("LFSTEST_CLIENT_KEY=%s", os.Getenv("LFSTEST_CLIENT_KEY")),
+			)
+			cmd.Stdout = log
+
+			// Start performs a fork/execve, hence we can abandon
+			// this process once it has started.
+			if err := cmd.Start(); err != nil {
+				return n, err
+			}
+			return n + 1, nil
```
PastelMobileSuit: Maybe we should return a hard-coded `1` here as below?

t/cmd/lfstest-count-tests.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	contents, err := ioutil.ReadAll(f)
+	if err != nil {
+		return err
+	}
+
+	var n int = 0
+	if len(contents) != 0 {
+		n, err = strconv.Atoi(string(contents))
```
PastelMobileSuit: Would it be worth erroring out here if the `n` is less than 0? It seems like a good safety to have since the number is being read from a file.

PastelMobileSuit:
Due to the size of this PR I reviewed commit-by-commit so a few comments are listed as being outdated, but the feedback in them still applies

t/cmd/lfstest-count-tests.go
```diff
@@ -0,0 +1,201 @@
...
+
+func path(s string) (string, error) {
+	p := filepath.Join(os.Getenv("LFSTEST_DIR"), s)
+	if err := os.MkdirAll(filepath.Dir(p), 0666); err != nil {
+		return "", err
+	}
+	return p, nil
+}
+
+func alreadyExists(err error) bool {
```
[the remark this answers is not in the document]
ttaylorr: Great call; I think I caught this one in a later commit https://github.com/git-lfs/git-lfs/pull/3125/commits/55fe3b99bbf0db68d5cafde4b44803d0867bba92 (per your comment below).

t/t-ls-files.sh
```diff
@@ -334,17 +334,12 @@ begin_test "ls-files: list/stat files with escaped runes in path before commit"
 (
   set -e
 
-  if [ -n "$CIRCLECI" ]; then
```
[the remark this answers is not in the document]
ttaylorr: I'm a little bit at a loss for what what is going on with this test; my editor showed me a much different diff which did appear to change the encoding. I'll happily defer to others on what to do here (incl. dropping this commit from the series).

t/README.md
```diff
@@ -1,137 +1,116 @@
-# Git LFS Tests
+# `t`
 
-Git LFS uses two form of tests: unit tests for the internals written in Go, and
-integration tests that run `git` and `git-lfs` in a real shell environment.
-You can run them separately:
+This directory contains one of the two types of tests that the Git LFS project
```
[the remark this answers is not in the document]
ttaylorr: I think I'd rather leave this as-is, unless you feel strongly about the change. `TESTING.md` is not a standard top-level piece of documentation (like `README.md` and `CONTRIBUTING.md`, both of which do describe our testing infrastructure).

Instead, users can note that there exists a `t` folder, and if they imagine that they can run integration tests from there, this documentation should clear up any confusion.

t/README.md
```diff
@@ -1,137 +1,116 @@
...
+$ make PROVE_EXTRA_ARGS=-v ./t-checkout.sh
+$ ./t-checkout.sh
 ```
-$ script/cibuild
+
+Alternatively, one can run a selection of tests (via explicitly listing them or
+making use of the built-in shell globbing) by any of the following:
+
+```ShellSession
+$ make ./t-*.sh
```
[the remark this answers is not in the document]
ttaylorr: Sounds good, I changed this to the more broadly-applicable `t/t-*.sh` glob, and pushed the changes as 349ab994 (t/README.md: shell-glob over all files in t, 2018-07-16)
.

t/cmd/lfstest-count-tests.go
ttaylorr: Sounds like a good idea to me; this makes everything nice and consistent. I got that as: bbb8f5e4 (t/cmd/lfstest-count-tests.go: prefer consistent constant usage, 2018-07-16)/

t/cmd/lfstest-count-tests.go
ttaylorr: Great idea! I got that as: 8f6da107 (t/cmd/lfstest-count-tests.go: return errNegativeCount when negative, 2018-07-16).

t/t-ls-files.sh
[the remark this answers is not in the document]
PastelMobileSuit: @ttaylorr I'm happy to leave it, I was just confused by the diff! If your editor showed encoding changes that's fine, it's possible they're just not being rendered in this diff. I'm fine to leave this as-is

t/README.md
```diff
@@ -0,0 +1,117 @@
+# `t`
+
+This directory contains one of the two types of tests that the Git LFS project
```
PastelMobileSuit: These two paragraphs are great and I think they should be in the Git LFS repository, but I think it would be better if we made a top-level `TESTING.md` file and put them there. I think it's a bit confusing to talk about the unit tests here, since this directory strictly deals with integration testing.

If we did make a top-level file, this more general testing overview info will be more visible and easily accessible, and we can link to this more granular, integration test-specific documentation.

PR #3137, opened 2018-07-20
Set original file content type on basic upload.

calavera:
This change tries to detect the content type for a file when uploading
to the final store instead of always use `octect-stream`. It ignores
this behavior only if the response from the LFS server set an explicit
content type.

After reading the documentation, I'm not very clear about why the
content type is always octect-stream, so let me know if I missed
something obvious.

I've used this change with a repository using GitHub LFS and everything
seems to be working as expected. But I can also add tests as needed if
you point me in the right direction.

Signed-off-by: David Calavera <david.calavera@gmail.com>


tq/basic_upload.go
```diff
@@ -74,6 +72,25 @@ func (a *basicUploadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb Progres
 	}
 	defer f.Close()
 
+	if len(req.Header.Get("Content-Type")) == 0 {
```
ttaylorr: Could this be extracted into its own method? I can foresee changes like adding a configuration option to disable this, changing the default content type, and etc.

Maybe something like:

```go
func setContentTypeFor(req *http.Request, r io.Reader) error { ... }
```

tq/basic_upload.go
```diff
@@ -74,6 +72,25 @@ func (a *basicUploadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb Progres
...
+		buffer := make([]byte, 512)
+		n, err := f.Read(buffer)
+		if err != nil && err != io.EOF {
+			return errors.Wrap(err, "basic upload")
```
ttaylorr: I think this wrap description could be more informative. Perhaps something like:

```go
return errors.Wrap(err, "content type detect")
```

tq/basic_upload.go
```diff
@@ -74,6 +72,25 @@ func (a *basicUploadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb Progres
...
+		}
+
+		contentType := http.DetectContentType(buffer[:n])
+		if _, err := f.Seek(0, 0); err != nil {
+			return errors.Wrap(err, "basic upload")
```
ttaylorr: I have the same thought about this error wrap as I do the above, maybe replacing "content type detect" with "content type rewind" instead.

calavera:
Thanks for the review @ttaylorr, I made those changes.

PR #3145, opened 2018-07-25
t: fix contains_same_elements() fn

PastelMobileSuit:
Previously, the contains_same_elements() function was causing
tests to immediately exit when called, meaning that portions of
many tests were being inadvertently skipped. This fixes the
contains_same_elements() function so it no longer exits the test,
and fixes the failing tests exposed by this change.


t/testhelpers.sh
```diff
@@ -742,15 +742,7 @@ cat_end() {
...
-  printf '%s' "$1" | grep -v '^$' | sort > a.txt
-  printf '%s' "$2" | grep -v '^$' | sort > b.txt
-
-  set +e
-  diff -u a.txt b.txt 1>&2
-  res=$?
-  set -e
-  rm a.txt b.txt
-  exit $res
+  diff -u <(printf '%s' "$1" | grep -v '^$' | sort) <(printf '%s' "$2" | grep -v '^$' | sort)
```
ttaylorr: :sparkles: Nice bash-fu! I think that one thing we could do in the future (but definitely shouldn't hold up this pull request) is extract a function `normalize_lines`, like so:

```bash
normalize_lines() {
  echo "$1" | grep -v '^$' | sort
}
```

and then update this to read:

```bash
diff -u <(normalize_lines "$1") <(normalize_lines "$2")
```

PR #3150, opened 2018-07-26
release: v2.5.0

ttaylorr:
##

/cc @git-lfs/core @git-lfs/releases

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,126 @@
 # Git LFS Changelog
 
+## 2.5.0 (26 July, 2018)
+
+This release adds three new migration modes, updated developer ergonomics, and
+a handful of bug fixes to Git LFS.
+
+We would like to extend a special thanks to the following open-source
+contributors:
+
+* @calavera for fixing a broken Go test and adding support for custom
+  Content-Type headers in #3137 and #3138.
+* @cbuehlmann for adding support for encoded character names in filepaths via
+  #3093.
+* @larsxschneider for changing the default value of lfs.allowincompletepush in
+  #3109.
+* @NoEffex for supporting TTL in SSH-based authentication tokens via #2867.
+* @ssgelm for adding 'go generate' to our Debian packages via #3083.
+
+### Features
+
+* Makefile: replace many scripts with make targets #3144 (@ttaylorr)
+* {.travis,appveyor}.yml: upgrade to Go 1.10.3 #3146 (@ttaylorr)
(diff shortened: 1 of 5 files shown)


larsxschneider:
🎉 👍 😄

dscho:
Could you maybe automate the release engineering, also to guarantee a consistent file naming and release notes? For v2.5.0, the .zip file has a `v` in front of the version number, which the previous version had not. And the SHA-256s are now listed *before* the files, previously they were listed afterwards. This broke the automated upgrade script I have to maintain all kinds of 3rd-party components I bundle in Git for Windows, so I had to spend some quality time today to fix it...

ttaylorr:
> Could you maybe automate the release engineering, also to guarantee a consistent file naming and release notes?

Thanks for the note. Automating the release process is a to-do item on my list, but achieving this is not a straightforward process. Thanks for including Git LFS in GFW, and for fixing some inconsistencies between releases.

dscho:
Technically, I did not fix any inconsistency at all. I worked around them, by reacting to failures of the upgrade job.

dscho:
And now I will work around the next inconsistency, which is that the SHA-256 sums are once again listed *below*, rather than *above* the file names. Seriously, once you automate that, you will relieve me of a ton of work with every Git LFS release.

dscho:
For the record, this particular inconsistency is worked around in the Git for Windows project by this commit: https://github.com/git-for-windows/build-extra/commit/cf35e4915fb4d67810cf9fb26718c502a80a5218 (and if that diff scares you, it should: this is the consequence non-automated Git LFS releases has on Git for Windows alone; I do not want to muse about other the impact on other projects...).

ttaylorr:
> And now I will work around the next inconsistency, which is that the
> SHA-256 sums are once again listed *below*, rather than *above* the
> file names.

Sorry, and we appreciate you working around the inconsistencies between
releases. In the future, we'll do our best to make sure that the
signature is below rather than above the filename, such that the format
is:

    **git-lfs-$(GOOS)-$(GOARCH)-v$(VERSION).tar.gz**
    $shasum

and:

    **git-lfs-windows-$(GOARCH)-v$(VERSION).zip**
    $shasum

This release adopted the above styling after you [pointed out][1] that
we had been inconsistent in our usual styling with the v2.5.0 release.

You can expect the above (modulo some human error) in future releases.

[1]: https://github.com/git-lfs/git-lfs/pull/3150#issuecomment-408456045

PR #3151, opened 2018-07-26
🚧 add new options to `git lfs ls-files`

larsxschneider:
This PR adds an option to skip index scans and adds support for range scans in `git lfs ls-files`.

Currently, the output does not work as `ScanRefRange()` returns objects that it shouldn't. I am not yet sure if the bug is on the caller side or in the scanner machinery (I suspect the former).

I opened that PR for early high level feedback. @ttaylorr would these new options be ok for you?

### Todo
- [ ] investigate/fix ref range scanner bug
- [ ] add test cases
- [ ] add documentation

### DIFF
--- a/commands/command_ls_files.go
+++ b/commands/command_ls_files.go
@@ -23,12 +23,21 @@ func lsFilesCommand(cmd *cobra.Command, args []string) {
 	requireInRepo()
 
 	var ref string
+	var otherRef string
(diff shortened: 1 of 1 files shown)


larsxschneider:
I decided to take a different route and as a consequence this change is not necessary at the moment.

Issue #3158, opened 2018-07-31
Uploads fail with 422 error with Git LFS 2.5.0


zjslqshqz:
Software version
```
git-lfs/2.5.0 (GitHub; darwin 386; go 1.10.3; git 93775601)
gitlab 11.2
```
```
Locking support detected on remote "origin". Consider enabling it with:
  $ git config lfs.https://gitlab.gitlab.com/t.git/info/lfs.locksverify true
Uploading LFS objects:   0% (0/1), 0 B | 0 B/s                                                               
LFS: Client error: https://gitlab.gitlab.com/t.git/gitlab-lfs/objects/80cf36a3a93314dcae215ae8918b8b0faef1c5fb2b1af917ab3632edf5628e43/26 from HTTP 422
Uploading LFS objects:   0% (0/1), 0 B | 0 B/s, done
error: failed to push some refs to 'https://gitlab.gitlab.com/t.git'

```
BUT!

Downgrading to git-lfs=2.4.1 fixed the issue for me.

Is probably on to something with the Content-Type header theory going from:
Content-Type: application/octet-stream in 2.4.1
to the more appropiate
Content-Type: application/zip in 2.5.0.

So this problem can only be solved by downgrade?

ttaylorr:
Hi @zjslqshqz, thank you for letting us know, and we're sorry about the regression.

I think that I could have handled merging the change that introduced this behavior in a different way that would have alerted other hosting platforms, so I think that we have a few options.

1. We should (first and foremost) alert GitLab (as well as any other hosting platforms that are broken by this change) about what has happened and what they can do to go forward.

2. We should either (A) introduce an option (a-la `lfs.[url].locksverify`) to disable content-type detection in a host-specific configuration, or (B) disable the feature entirely in v2.5.1 while we wait for (1) to take effect.

I would like to get the consensus of other core team members, incl. @PastelMobileSuit and @larsxschneider. Let's figure out what to do, and then do so.

zjslqshqz:
@ttaylorr Thank you for your reply, I understand

PastelMobileSuit:
@ttaylorr I'm in favor of 2a, since we've already done the work to add the feature and it does work on some hosting platforms. My opinion is that we should default this option to `false` and have the feature be opt-in until other hosting platforms support it.

ttaylorr:
> @ttaylorr I'm in favor of 2a, since we've already done the work to add
> the feature and it does work on some hosting platforms.

[...]

PR #3163, opened 2018-07-31
tq: make Content-Type detection disable-able

ttaylorr:
This pull request makes the `Content-Type` detection introduced in #3137 disable-able, thus allowing uploads on implementations that don't support non-octet-stream headers.

Closes: https://github.com/git-lfs/git-lfs/issues/3158.

##

/cc @git-lfs/core

### DIFF
--- a/docs/man/git-lfs-config.5.ronn
+++ b/docs/man/git-lfs-config.5.ronn
@@ -266,6 +266,12 @@ be scoped inside the configuration for a remote.
   https://git-scm.com/docs/git-config#git-config-httplturlgt. To set this value
   per-host: `git config --global lfs.https://github.com/.locksverify [true|false]`.
 
+* `lfs.<url>.contenttype`
+
+  Determines whether Git LFS should attempt to detect an appropriate HTTP
+  `Content-Type` header when uploading using the 'basic' upload adapter.
+  Default: 'true'.
+
 * `lfs.skipdownloaderrors`
 
   Causes Git LFS not to abort the smudge filter when a download error is
--- a/docs/man/git-lfs-config.5.ronn
+++ b/docs/man/git-lfs-config.5.ronn
@@ -269,8 +269,9 @@ be scoped inside the configuration for a remote.
 * `lfs.<url>.contenttype`
 
   Determines whether Git LFS should attempt to detect an appropriate HTTP
-  `Content-Type` header when uploading using the 'basic' upload adapter.
-  Default: 'true'.
+  `Content-Type` header when uploading using the 'basic' upload adapter. If set
+  to false, the default header of `Content-Type: application/octet-stream` is
+  chosen instead. Default: 'true'.
 
 * `lfs.skipdownloaderrors`
 
--- a/errors/types.go
+++ b/errors/types.go
@@ -127,6 +127,20 @@ func IsDownloadDeclinedError(err error) bool {
 	return false
 }
 
+// IsDownloadDeclinedError indicates that the upload operation failed because of
+// an HTTP 422 response code.
+func IsUnprocessableEntityError(err error) bool {
+	if e, ok := err.(interface {
+		UnprocessableEntityError() bool
+	}); ok {
+		return e.UnprocessableEntityError()
+	}
+	if parent := parentOf(err); parent != nil {
+		return IsUnprocessableEntityError(parent)
+	}
+	return false
+}
+
 // IsRetriableError indicates the low level transfer had an error but the
 // caller may retry the operation.
 func IsRetriableError(err error) bool {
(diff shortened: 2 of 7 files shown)


errors/types.go
```diff
@@ -127,6 +127,20 @@ func IsDownloadDeclinedError(err error) bool {
 	return false
 }
 
+// IsDownloadDeclinedError indicates that the upload operation failed because of
+// an HTTP 422 response code.
+func IsUnprocessableEntityError(err error) bool {
+	if e, ok := err.(interface {
```
PastelMobileSuit: I've never seen this syntax before! Could you explain what it's doing here?

t/t-content-type.sh
```diff
@@ -0,0 +1,67 @@
...
+  git lfs track "*.zip"
+  printf "aaaaaaaaaa" > a.txt
+  zip -j a.zip a.txt
+  rm a.txt
+
+  git add .gitattributes a.zip
+  git commit -m "initial commit"
+  GIT_CURL_VERBOSE=1 git push origin master 2>&1 | tee push.log
+
+  [ 1 -eq "$(grep -c "Content-Type: application/zip" push.log)" ]
```
PastelMobileSuit: It looks like this is checking that `application/zip` wasn't found by the grep, but we'd be expecting the opposite

tq/transfer_queue.go
```diff
@@ -763,6 +774,17 @@ func (q *TransferQueue) toAdapterCfg(e lfsapi.Endpoint) AdapterConfig {
 	}
 }
 
+var (
+	// contentTypeWarning is the message printed when a server returns an
+	// HTTP 422 at the end of a push.
+	contentTypeWarning = []string{
+		"Uploading failed due to unsupported Content-Type header(s).",
```
PastelMobileSuit: This error message is great! I think this will help users a lot if they run into this error

errors/types.go
ttaylorr: Definitely! This is a typecast, but it looks a little odd since the type is defined inline. Normally, you'd see something like:

```go
func doSomethingIfReader(v interface{}) {
        if r, ok := v.(io.Reader); ok {
                // ok is 'true' because 'v' is an 'io.Reader' underneath.
                // So, we can use any function that io.Reader receives,
                // or use 'r' as a bona-fide io.Reader.
                io.Copy(os.Stdout, r)
        }
}

But, inside the `()` (where `io.Reader` is above), you can define any type, including doing so inline.

So, this example defines an `interface` with the function above, and then checks to see if `err` implements that (i.e., is a type that receives all of those methods). If it is, `e` receives those methods, and `ok` is true. If not, `ok` is false.

Particularly, we're checking if either `err` or any of its parents are this sentinel type by checking if it receives a method. If so, we return true, and if not, we return false otherwise. In practice, this isn't a super common technique, but it is the pattern in this file, so I'm inclined to go with that over something else.

File: t/t-content-type.sh (hunk 2 above)
Comment (ttaylorr) [reply t/t-content-type.sh]: Hm, I think that this is checking that the output contained `application/zip` exactly once. According to the manpage, `grep -c` means:

> ```
> -c, --count
>        Only a count of selected lines is written to standard output.
> ```

In which case `-eq` means a numerical comparison. So, we're checking that the number of times we found `Content-Type: application/zip` in the output is exactly one.

File: tq/transfer_queue.go (hunk 3 above)
Comment (ttaylorr) [reply tq/transfer_queue.go]: Great! Thanks

**Comment by ttaylorr** [comment]
👋 thanks for your feedback, @PastelMobileSuit. I think I addressed all of your comments. Could you please take another look?

File: t/t-content-type.sh (hunk 2 above)
Comment (PastelMobileSuit) [reply t/t-content-type.sh]: Ah okay! That makes sense to me. In that case, is the test below this one correct? If we've disabled the content type detection, I'd expect application/zip to not appear in the log.

File: docs/man/git-lfs-config.5.ronn [hunk 4]
```diff
@@ -266,6 +266,12 @@ be scoped inside the configuration for a remote.
   https://git-scm.com/docs/git-config#git-config-httplturlgt. To set this value
   per-host: `git config --global lfs.https://github.com/.locksverify [true|false]`.
 
+* `lfs.<url>.contenttype`
```
PastelMobileSuit: Would it be worth mentioning what the `Content-Type` will default to if this is false?

PastelMobileSuit:
@ttaylorr Thanks for addressing my comments! I've left two more brief pieces of feedback but once those are addressed this should be ready for merge!

t/t-content-type.sh
ttaylorr: Ack! What a great catch. I wrote a little bit about why this happened in 55ce4393 (t: fix 'git config' invocation in content type tests, 2018-07-31).

docs/man/git-lfs-config.5.ronn
ttaylorr: Definitely, that's a great idea! Here's an attempt at doing just that: 1818c594 (docs/man: indicate the default Content-Type header, 2018-07-31)
.

ttaylorr:
> I've left two more brief pieces of feedback but once those are addressed this should be ready for merge!

Thanks again! I pushed up two more patches for you to take a look at, and I agree after that this should be ready to merge.

PastelMobileSuit:
Everything looks good! :shipit:

tq/transfer_queue.go
```diff
@@ -763,6 +774,17 @@ func (q *TransferQueue) toAdapterCfg(e lfsapi.Endpoint) AdapterConfig {
...
+		"Consider disabling Content-Type detection with:",
+		"",
+		"  $ git config lfs.contenttype false",
```
tardyp: isn't it `git config lfs.$URL.contenttype false"` ?

PR #3179, opened 2018-08-09
Retry failed netrc cred requests

### LINKED ISSUE #3130: Authentication issue on MacOS
Some of my co-workers experience a weird authentication issue with git-lfs.

We have a login/password authentication over https.

I've tried to use different store methods: 
1. store
1. keychain
1. cache

None is working, while the same setup on the other machine works.
git is able to use the credentials stored while git-lfs is not.

I wonder if there is some kind of verbose mode which could give more inside on what is going on in git-lfs.

PastelMobileSuit:
When a request is made with netrc credentials and that request
fails, retry the request with non-netrc credentials

Closes #3130

### DIFF
--- a/lfsapi/auth.go
+++ b/lfsapi/auth.go
@@ -24,13 +24,13 @@ var (
 // authentication from netrc or git's credential helpers if necessary,
 // supporting basic and ntlm authentication.
 func (c *Client) DoWithAuth(remote string, req *http.Request) (*http.Response, error) {
-	return c.doWithAuth(remote, req, nil)
+	return c.doWithAuth(remote, req, nil, false)
 }
 
-func (c *Client) doWithAuth(remote string, req *http.Request, via []*http.Request) (*http.Response, error) {
+func (c *Client) doWithAuth(remote string, req *http.Request, via []*http.Request, ignoreNetRc bool) (*http.Response, error) {
 	req.Header = c.extraHeadersFor(req)
 
-	apiEndpoint, access, credHelper, credsURL, creds, err := c.getCreds(remote, req)
+	apiEndpoint, access, credHelper, credsURL, creds, fromNetRc, err := c.getCreds(remote, req, ignoreNetRc)
 	if err != nil {
 		return nil, err
 	}
@@ -57,6 +57,12 @@ func (c *Client) doWithAuth(remote string, req *http.Request, via []*http.Reques
 				// and instead call DoWithAuth.
 				return c.DoWithAuth(remote, req)
 			}
+		}
+
+		if fromNetRc && (res.StatusCode == 401 || res.StatusCode == 403) {
+			tracerx.Printf("api: netrc credentials were rejected. Trying again with new credentials...")
+			req.Header.Del("Authorization")
+			return c.doWithAuth(remote, req, nil, true)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


lfsapi/auth.go
```diff
@@ -24,13 +24,13 @@ var (
 // authentication from netrc or git's credential helpers if necessary,
 // supporting basic and ntlm authentication.
 func (c *Client) DoWithAuth(remote string, req *http.Request) (*http.Response, error) {
-	return c.doWithAuth(remote, req, nil)
+	return c.doWithAuth(remote, req, nil, false)
 }
 
-func (c *Client) doWithAuth(remote string, req *http.Request, via []*http.Request) (*http.Response, error) {
+func (c *Client) doWithAuth(remote string, req *http.Request, via []*http.Request, ignoreNetRc bool) (*http.Response, error) {
```
ttaylorr: To make sure that I'm reading this change correctly: do you intend to mark when `doWithAuth` should and should not use the credentials found in `~/.netrc` upon each request?

If this is the case, it sounds like we start with credentials found in `~/.netrc`, and (if those fail), move on to the credential helper, and then the URL credentials, and so on.

That sounds like a reasonable plan to me. I have a few stylistic remarks regarding the new parameter to this function:

1. Do you think we could name the boolean argument positively? For example, instead of `ignoreNetrc` (nit also, casing should be `Netrc` not `NetRc`), could we name it as `useNetrc`? I think that this makes the conditional on line L123 and L155 easier to grok. Instead of `if !ignoreNetrc`, it would read `if useNetrc`, which is IMHO easier to read.

1. Should this argument be added at all? I usually find that adding a boolean argument to a function is a smell indicating a split in behavior change that should instead be applied via a function extraction. Do you think that this is possible with the current code structure? I'm generally wary of changing code in this file, but I think that removing the 6th return value from `getCreds` might be a worthwhile tradeoff for the risk.

More broadly, though, I'm wondering if this is the right way to go about this change. The credential helper that this function (`c.getCreds`) returns has the following interface:

```go
type CredentialHelper interface {
        Fill(Creds) (Creds, error)
        Reject(Creds) error
        Approve(Creds) error
}
```

and has an implementation that tries multiple `CredentialHelper`s in succession until either (1) one succeeds or (2) all of them fail. That goes by the name of `*CredentialHelpers`. I'm wondering if we could replace:

https://github.com/git-lfs/git-lfs/blob/v2.5.1/lfsapi/auth.go#L174

with an implementation of the `CredentialHelper` interface that fills from a user's `~/.netrc`. In this case, `Accept()` would be a no-op, but `Reject()` could indicate that further calls to `Fill()` for a given hostname should fail immediately. A bad `~/.netrc` file could return an error anytime that it's asked to fill a request.

This way, we could return one `CredentialHelper` (backed by a `*CredentialHelpers`) that would transparently attempt the `~/.netrc` first, and then cascade down through other credential helpers without having to re-call this method and juggle the extra argument.

PastelMobileSuit: I'm with you on the credential helper approach. I think there are two really clear benefits: the code will be cleaner, and URL credentials will be prioritized above .netrc credentials, which will make us more consistent with git. 

My only concern is that this change will modify auth priorities and the order in which we perform auth operations in this file. I discovered first-hand while making this change that even a simple change in ordering can cause some weird, unexpected behaviours (for example, switching the order of the conditionals at lines [123](https://github.com/git-lfs/git-lfs/pull/3179/files#diff-6d925a0e7b322da76d57d6d1f28fc30cR123) and [127](https://github.com/git-lfs/git-lfs/pull/3179/files#diff-6d925a0e7b322da76d57d6d1f28fc30cR127) causes some requests to fall through to the credential helper when they should be using .netrc instead). I think we'd want to make extra sure the change was safe before it goes into a release.

ttaylorr:
> I'm with you on the credential helper approach. I think there are two
> really clear benefits: [...]

Sounds good, and I agree.

> My only concern is that this change will modify auth priorities and
> the order in which we perform auth operations in this file.

That's fair; this code is called from lots of different places within
Git LFS, and the authentication procedures are both complex and fragile.

Maybe it would be worth documenting the order in which different things
happen under all possible circumstances, and saving your findings in
`docs`? I think that writing tests that encode the same would be
similarly helpful.

> I think we'd want to make extra sure the change was safe before it
> goes into a release.

Agreed, as is good practice to do on all changes, small or large.

PastelMobileSuit:
I'm going to close out this PR, as @ttaylorr and I have decided that writing a new credential helper is the best way to go (more details on #3130)

PR #3187, opened 2018-08-17
🚧fetch: add option to exclude OIDs

larsxschneider:
In exceptional cases Git LFS objects might not be available anymore on
a Git LFS server. This leads to errors if those objects are still
referenced in a Git repository. The only way to overcome these errors
is to rewrite the Git history and remove the "dangling" pointers.
Unfortunately, a history rewrite is not be possible in every case.

Therefore, add a config "lfs.fetchexcludeoids" to exclude certain Git
LFS object IDs. The multiple object IDs are separate by a comma. The
config can be set via the local Git config mechanism or via the
.lfsconfig mechanism.

Example:
```
[lfs]
  fetchexcludeoids = "hash1, hash2"
```

---

This is just a proof of concept. If you agree with the idea, I will implement it properly in every Git LFS call. Furthermore, I might implement it directly in the Scanner similar to the path filter mechanism.

/cc @git-lfs/core @terrorobe @mattpollard

### DIFF
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"time"
+	"sort"
 
 	"github.com/git-lfs/git-lfs/filepathfilter"
 	"github.com/git-lfs/git-lfs/git"
@@ -65,6 +66,7 @@ func fetchCommand(cmd *cobra.Command, args []string) {
 
 	include, exclude := getIncludeExcludeArgs(cmd)
 	fetchPruneCfg := lfs.NewFetchPruneConfig(cfg.Git)
+	excludeOIDs := cfg.FetchExcludeOIDs()
 
 	if fetchAllArg {
 		if fetchRecentArg || len(args) > 1 {
@@ -76,7 +78,7 @@ func fetchCommand(cmd *cobra.Command, args []string) {
 		if len(cfg.FetchIncludePaths()) > 0 || len(cfg.FetchExcludePaths()) > 0 {
 			Print("Ignoring global include / exclude paths to fulfil --all")
 		}
-		success = fetchAll()
+		success = fetchAll(excludeOIDs)
 
 	} else { // !all
 		filter := buildFilepathFilter(cfg, include, exclude)
@@ -219,8 +221,18 @@ func fetchRecent(fetchconf lfs.FetchPruneConfig, alreadyFetchedRefs []*git.Ref, 
 	return ok
 }
 
-func fetchAll() bool {
+func fetchAll(excludedOIDs []string) bool {
 	pointers := scanAll()
+	if excludedOIDs != nil && len(excludedOIDs) > 0 {
+		var cleaned = make([]*lfs.WrappedPointer, 0, len(pointers))
+		for _, p := range pointers {
+			// TODO uppercase/lowercase?
+			if excludedOIDs[sort.SearchStrings(excludedOIDs, p.Oid)] != p.Oid {
+				cleaned = append(cleaned, p)
+			}
+		}
+		pointers = cleaned
+	}
 	Print("fetch: Fetching all references...")
 	return fetchAndReportToChan(pointers, nil, nil)
 }
--- a/config/config.go
+++ b/config/config.go
@@ -149,6 +149,23 @@ func (c *Configuration) FetchExcludePaths() []string {
 	return tools.CleanPaths(patterns, ",")
 }
 
+func (c *Configuration) FetchExcludeOIDs() []string {
+	var cleaned []string
+	oids, _ := c.Git.Get("lfs.fetchexcludeoids")
+
+	if oids = strings.TrimSpace(oids); len(oids) == 0 {
+		return cleaned
+	}
+
+	for _, part := range strings.Split(oids, ",") {
+		part = strings.TrimSpace(part)
+		cleaned = append(cleaned, part)
+	}
+
+	// TODO: normalize case?
+	return cleaned
+}
+
 func (c *Configuration) CurrentRef() *git.Ref {
 	c.loading.Lock()
 	defer c.loading.Unlock()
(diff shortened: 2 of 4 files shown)


commands/command_fetch.go
```diff
@@ -76,7 +78,7 @@ func fetchCommand(cmd *cobra.Command, args []string) {
 		if len(cfg.FetchIncludePaths()) > 0 || len(cfg.FetchExcludePaths()) > 0 {
 			Print("Ignoring global include / exclude paths to fulfil --all")
 		}
-		success = fetchAll()
+		success = fetchAll(excludeOIDs)
```
PastelMobileSuit: A similar change would need to be made below this as well for `fetchRef()` and `fetchRecent()` to ensure this applies during all usages of a fetch.

PastelMobileSuit:
If time permits, it would be worth getting @ttaylorr's thoughts on this as well

ttaylorr:
On Fri, Aug 17, 2018 at 04:46:16AM -0700, Lars Schneider wrote:
> In exceptional cases Git LFS objects might not be available anymore on
> a Git LFS server. This leads to errors if those objects are still
> referenced in a Git repository.

Right. The only case that we would do this today on GitHub.com (for
example) is if a user requests that their Git LFS objects be deleted. In
this case, support@github.com will faithfully oblige, and Git LFS
checkouts that do _not_ contain the deleted objects in .git/lfs/objects
will fail.

> The only way to overcome these errors is to rewrite the Git history
> and remove the "dangling" pointers.  Unfortunately, a history rewrite
> is not be possible in every case.

I don't think that this is quite true, though my memory could be failing
me. We have an option in git-lfs-config(5) called
"lfs.allowincompletepush", which indicates to the transfer queue that
objects are allowed to be missing, and the push can still be successful
(i.e., the pre-push hook will _not_ halt a 'git push' in the case of
missing Git LFS objects).

> Therefore, add a config "lfs.fetchexcludeoids" to exclude certain Git
> LFS object IDs. The multiple object IDs are separate by a comma. The
> config can be set via the local Git config mechanism or via the
> .lfsconfig mechanism.

I'm wary of this for a few reasons. Firstly, I'm concerned about using
'git config' as the mechanism for storing which objects are marked as
'missing'. Distributing them over `.lfsconfig` is a nice property, but
I'm not sure that it would be ideal to have potentially quite large
strings scattered throughout a user's .gitconfig file.

Further, it doesn't guarantee that all users will converge on the same
accepted set of missing OIDs. For example, if I declare the "hash1" is
'missing' in my repository-local .git/config file, but you don't declare
it missing anywhere, then you will encounter an error. I foresee this
case as potentially fracturing clusters of Git LFS clones into a game of
"who has which object(s)".

We can get around this by centralizing the list, i.e., by introducing
'git lfs mark [--missing] <oid [...]>' (or similar) and storing the list
somewhere, but that involves an API change and implementation work from
other hosting providers.

More broadly, though, I am not sure that this is the right feature that
we should be implementing. My concern is that in the case of what is
effectively a broken repository, we ought to be OK with suggesting that
a history rewrite is the only way to go. (This view is motivated by the
position that a globally missing Git LFS object is the fault of the
object store, and the repository must match the object store, not the
other way around.)

Another thing that I am thinking about is how this might interplay with
another feature that has been on our roadmap for a while, which is
object expirations. Does it make sense to have both a notion of
"missing" and "expired" objects? Could we implement this feature in
terms of an object that has expired (or expires instantly)?

I'm curious about the particular use-case that you're trying to solve.
Perhaps that might inspire us in a different direction, or convince us
that this is the right one.

ttaylorr:
@larsxschneider and I talked offline and had a few other thoughts about
ways that we could approach this problem. What this roughly boils down
to is three things:

  1. When is it OK to miss objects in pushes?
  2. When is it OK to miss objects in fetches?
  3. When is it OK to miss objects during checkouts?

One approach that addresses all three is an external filter program,
that accepts (perhaps over pktline or similar) an encoded set of objects
that are a part of a forthcoming batch API request, and filters the set
of allowed and disallowed objects. A similar protocol extension could be
implemented for filtering checkouts in a similar way.

In this world, a client can decide how they want to filter out objects
with flexibility. They could, for example, decide that they only want to
filter out objects that are missing from a remote, filter objects that
are older than some duration, or any arbitrary query.

This flexibility comes at the cost of some complexity in the
implementation on both the client end, and on the programmer in charge
of writing a filtering program. I think that using this as a means to
achieve what we're after here is too burdensome.

On the other hand, we could leverage onto the existing
lfs.allowincompletepush configuration option, and introduce
lfs.allowincompletefetch and lfs.allowincompletecheckout options to
match.

These would be easy to teach, and easy to implement, but ultimately
makes a sacrifice to the flexibility of the end user. By turning all
three knobs 'on', the user is opting-out of the Git LFS "safety net"
entirely. That's too dangerous to advocate for as a solution to this
particular problem, too.

All of these options aside, something that comes to mind when
considering how to scope checkouts (and therefore fetches, since all
"missing" objects won't be cached locally and will thusly require a
fetch), is the --exclude option (which 'git lfs pull', and 'git lfs
push' both accept, as does the filtering procedure, via the
lfs.fetchexclude configuration).

What we would like to express, roughly, is "a.txt is missing in a
particular revision, so, when you're on that revision, please do not
check it out or fetch it." The best we can do that effect today is
--exclude=a.txt, which rids ourselves of a.txt entirely over all
revisions. This doesn't cover the case where a _single_ revision of the
LFS object at path a.txt was missing, but others were not.

So, it would be therefore possible to extend the `--exclude` (and
`--include`) option format to accept a revision that they apply to, or a
range or revisions likewise.

For example, if the Git LFS object for a.txt is missing at revision
fb3675526e, we could instruct Git LFS to not check out or fetch that
object via:

    --exclude=a.txt@fb3675526e

If the object is missing over a _range_ of revisions, we could also do
so by saying:

    --exclude=a.txt@fb3675526e...975559d0f9

by essentially splitting the exclude list by `,` (our in-use separator),
and looking for the last index of `@`, flattening all of the revisions
down to a set of relevant revisions, and then extending package
`filepathfilter` to understand revisions.

Perhaps this extension could look like:

```go
func (f *Filter) AllowsAt(filename string, rev string) bool { ... }
```

where `rev` refers to the currently-relevant revision (i.e., for
checkouts it would be HEAD, for fetches it would be the range of
revisions in the fetch, for pushes it would be the range of revisions in
the push, etc.)

That might be more complicated to implement than what is here currently,
but I think that this is ultimately a more general-purpose solution to
this particular problem.

ddevyatyar:
Ddddevyatyar

larsxschneider:
We reject pushes with unavailable OIDs now. Therefore this PR is not as urgent as before. Therefore, I close it until we need it again.

Issue #3181, opened 2018-08-10
doc: git lfs install vs. install.sh


srl295:
It seems that `git lfs install` doesn't actually install the `git-lfs` binary onto the PATH. (Should it have been `git lfs configure` or `git lfs initialize`?) 

Documentation on https://git-lfs.github.com/ says:

> 1. Download and install the Git command line extension. You only have to set up Git LFS once.
>     `git lfs install`

This makes it sound like you need to:   a. fetch the tgz, b. unpack it, c. run `git lfs install` to install.  But actually, 'download and install' means that the binary needs to be put on the PATH,  and then `git lfs install` is an additional step after installing the binary.   `install.sh` is not mentioned.

The [README.md](https://github.com/git-lfs/git-lfs/blob/master/README.md) (which is also included in the .tgz) does not mention `install.sh` either. 

I note that `install.sh` _calls_ `git lfs install` as its final step.  So really, the instructions could say:

1. download and unpack the .tgz
2. execute `install.sh` which will install `git-lfs` on your PATH  _and which will run `git lfs install` for you_.

srl295:
noting the wording choices here:

> $ `git lfs install --skip-repo`
> Git LFS **initialized**.

> $ `git lfs install --help`
> …
> Perform the following actions to ensure that Git LFS is **setup properly**:
>
> * **Set up** the clean and smudge filters under the name "lfs" in the global Git
  config.
> * **Install** a pre-push hook to run git lfs pre-push for the current repository,
  if run from inside one. 
```

srl295:
https://help.github.com/articles/installing-git-large-file-storage/ has been found to be a helpful article. Perhaps the instructions there could be folded into the lfs readme/webpage.

srl295:
@ttaylorr 

> That's correct.

Also that the verb `install` is misleading? Perhaps it's really more similar to `git submodule init`. Would it be a welcome change to rename install to `git lfs init` but leave a deprecated alias `git lfs install`? 

> You're more than welcome to open a pull request

SGTM and thanks.

PR #3193, opened 2018-08-21
Update README.md

srl295:
per #3181 

- `git lfs install` does not actually install Git LFS.
- `git lfs track` is not part of the installation steps, but part of example usage.
- `install.sh` was not mentioned. Mention it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ for an overview of features.
 
 ## Getting Started
 
-### Installation
+### Downloading
 
 You can install the Git LFS client in several different ways, depending on your
 setup and preferences.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,18 +39,30 @@ built from source using the latest version of [Go](https://golang.org), and the
 available instructions in our
 [Wiki](https://github.com/git-lfs/git-lfs/wiki/Installation#source).
 
-### Usage
+#### Installing from a [binary package](https://github.com/git-lfs/git-lfs/releases)
```
ttaylorr: I have a few suggestions on this line:

1. What do you think about breaking this into an H3-level "Installing" heading and two H4-level sub-headings "From binary" and "From source"? 
2. Typically we don't have links in headings, so maybe you could write a short 1-sentence below the heading and mention this link.

README.md
```diff
@@ -39,18 +39,30 @@ built from source using the latest version of [Go](https://golang.org), and the
...
 
-Git LFS requires a global installation once per-machine. This can be done by
+The included script will install Git LFS binaries. It also runs `git lfs install` to
+perform required global configuration changes.
+
+```bash
```
ttaylorr: It would be preferred to use `ShellSession` as the language for this, which looks like:

(With `bash`):
```bash
$ ./install.sh
```

(With `ShellSession`):
```ShellSession
$ ./install.sh
# ...
```

The two look identical without a line below, but it will make the diff easier to grok in the future if someone does wish to add output below `$ ./install.sh`.

README.md
```diff
@@ -39,18 +39,30 @@ built from source using the latest version of [Go](https://golang.org), and the
...
+$ ./install.sh
+```
+
+#### Installing manually
+
+- Place the `git-lfs` binary on your system’s executable `PATH` or equivalent.
```
ttaylorr: Let's refer to `$PATH` as such (with the proceeding `
$`) to indicate that it is an environment variable.

README.md
```diff
@@ -31,26 +31,43 @@ setup and preferences.
...
-### Usage
+### Installing
+
+#### From binary
+
+The [binary packages](https://github.com/git-lfs/git-lfs/releases) include a script which will:
 
-Git LFS requires a global installation once per-machine. This can be done by
+- Install Git LFS binaries onto the system `$PATH`
+- Run `git lfs install` to
```
sffc: It's not "global"; it's per-user.  When you say "global" it makes me want to run sudo.

jefgen: Its "global" for the user?

srl295: Global as in worldwide!   Anyway, this was the original wording. It means git global as opposed to per repo. `git config --global -l | fgrep lfs`

rubelmusicwould:
Rubelmusicwould

PR #3196, opened 2018-08-24
Add build and releases for linux arm64

mmlb:
It'd be great to have linux arm64 builds available. This is about as far as I'm able to get I think. I've added the targets for other oses so that if/when go supports them all that is needed is to add to BUILD/RELEASE targets. This'll bring up the question of how to test before making said release. I don't have an easy answer for that. I think shippable has support for arm builders, but that'd bring the ci services up to 4 (if true) which seems like a pain.

The TAR_XFORM stuff maybe doesn't belong in the PR, let me know if you'd like me to drop it. I also thought about using `bsdtar` instead of `tar` which would be fine, but I don't know if that's available on the machines normally used to make releases so opted to not break your stuff.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,10 +135,15 @@ BUILD = GOOS=$(1) GOARCH=$(2) \
 
 # BUILD_TARGETS is the set of all platforms and architectures that Git LFS is
 # built for.
-BUILD_TARGETS = bin/git-lfs-darwin-amd64 bin/git-lfs-darwin-386 \
-	bin/git-lfs-linux-amd64 bin/git-lfs-linux-386 \
-	bin/git-lfs-freebsd-amd64 bin/git-lfs-freebsd-386 \
-	bin/git-lfs-windows-amd64.exe bin/git-lfs-windows-386.exe
+BUILD_TARGETS = \
+	bin/git-lfs-darwin-amd64 \
+	bin/git-lfs-darwin-386 \
+	bin/git-lfs-linux-amd64 \
+	bin/git-lfs-linux-386 \
+	bin/git-lfs-freebsd-amd64 \
+	bin/git-lfs-freebsd-386 \
+	bin/git-lfs-windows-amd64.exe \
+	bin/git-lfs-windows-386.exe
 
 # mangen is a shorthand for ensuring that commands/mancontent_gen.go is kept
 # up-to-date with the contents of docs/man/*.ronn.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -135,10 +138,16 @@ BUILD = GOOS=$(1) GOARCH=$(2) \
 
 # BUILD_TARGETS is the set of all platforms and architectures that Git LFS is
 # built for.
-BUILD_TARGETS = bin/git-lfs-darwin-amd64 bin/git-lfs-darwin-386 \
-	bin/git-lfs-linux-amd64 bin/git-lfs-linux-386 \
-	bin/git-lfs-freebsd-amd64 bin/git-lfs-freebsd-386 \
-	bin/git-lfs-windows-amd64.exe bin/git-lfs-windows-386.exe
+BUILD_TARGETS = \
```
ttaylorr: If you plan to make the ARM targets part of the default set of build targets (i.e., the ones that are usually built with `make build` or `make release`), I think that those should be added here, too.

The ones that I see that are missing are:

- `bin/git-lfs-darwin-arm64`
- `bin/git-lfs-freebsd-arm64`
- `bin/git-lfs-windows-arm64.exe`

Makefile
```diff
@@ -223,8 +240,10 @@ script/windows-installer/git-lfs-wizard-image.bmp
 # To build a specific release with a custom VERSION suffix, run the following:
 #
 # 	make VERSION=my-version bin/releases/git-lfs-darwin-amd64-my-version.tar.gz
-RELEASE_TARGETS = bin/releases/git-lfs-darwin-amd64-$(VERSION).tar.gz \
+RELEASE_TARGETS = \
```
ttaylorr: Likewise, I think that we need to add the following here:

- `bin/releases/git-lfs-freebsd-arm64-$(VERSION).tar.gz`
- `bin/releases/git-lfs-windows-arm64-$(VERSION).zip`

Makefile
mmlb: GO doesn't support arm64 on those OSes. I added them in for future go, but can just drop if you prefer.

Makefile
mmlb: same

Makefile
ttaylorr: In that case, let's drop the ones that are missing here and below from the Makefile entirely.

mmlb:
Hey @ttaylorr I'm aware of a few ci setups building on arm, most are byo type (drone, jenkins, buildroot ...). Shippable seems to be the only CIAAS doing [arm](http://docs.shippable.com/platform/tutorial/workflow/run-ci-builds-on-arm/#running-jobs-on-arm-machines) atm, though I hope to see more in this space soon. If you connect shippable I'll work on the ci.

Makefile
mmlb: dropped unused targets

Makefile
mmlb: done

mmlb:
ping @ttaylorr

ttaylorr:
Thanks for making the changes I requested and adding the arm64 targets
as a part of the default set of targets when running 'make build' and
'make release'.

I appreciate your efforts in getting the ball rolling for using
Shippable, but I don't think that we are going to pursue adding another
CI service to the set of three that we already use.

Particularly, I think that we would be most comfortable with
distributing those build artifacts as part of the standard release
process, but with the caveat that tests for those OS+Architecture
combinations are not run, so the binaries should be run with care.

In order to get this merged for v2.6.0, I think all we would need to do
is drop ceac13ce (add shippable, 2018-08-28) from your branch, in
which case it should be ready to go.

Does that sound OK?

mmlb:
SGTM. I was pretty sure you weren't going to go for a 4th CI :D but wanted to check it out just in case. Rebased on top of latest master and dropped the shippable commit.

mmlb:
:+1: :tada:

PR #3199, opened 2018-08-27
Pass GIT_SSH_COMMAND to the shell

### LINKED ISSUE #3131: Quotes in GIT_SSH_COMMAND are (still) not properly handled
#1895 reported a bug that occurred when GIT_SSH_COMMAND had quotes, e.g.:
```
GIT_SSH_COMMAND="ssh -i '/a path/with spaces'"
```

#1962 fixed this bug, but in an incomplete way. A value with two pairs of quotes still fails, e.g.:
```
GIT_SSH_COMMAND="ssh -i '/a path/with spaces' -o 'Set Some Option'"
```

The bug is in `QuotedFields` (`str_tools.go`) - the regexp is greedy so it catches anything between the outermost pair of quotes. This can be confirmed by adding a line in `TestQuotedFields` (`str_tools_test.go`):
```go
"single quotes twice": {`foo 'bar baz' 'blip blop'`, []string{"foo", "bar baz", "blip blop"}},
```
This test fails.

Unfortunately, there are many other values of GIT_SSH_COMMAND which will work in git, but won't work in git-lfs. Git uses `sh -c "command"` when running the GIT_SSH_COMMAND, which is the only way to support anything that a user may want to put in there. In fact, the documentation [explicitly states](https://git-scm.com/docs/git#git-codeGITSSHCOMMANDcode) that GIT_SSH_COMMAND is "interpreted by the shell".

(Relevant code in git: [fill_ssh_args](https://github.com/git/git/blob/c2c7d17b030646b40e6764ba34a5ebf66aee77af/connect.c#L1182), [prepare_shell_cmd](https://github.com/git/git/blob/e3331758f12da22f4103eec7efe1b5304a9be5e9/run-command.c#L254).)

If git-lfs doesn't want to use "sh -c", or needs some added argument parsing/modification magic, another solution is to use a library for shell string parsing. Some examples from a quick search: [go-shellwords](https://github.com/mattn/go-shellwords), [shlex](https://github.com/google/shlex), [go-shlex](https://github.com/anmitsu/go-shlex), [go-shellquote](https://github.com/kballard/go-shellquote), [shellwords](https://github.com/Wing924/shellwords). Git uses its own [split_cmdline](https://github.com/git/git/blob/c2c7d17b030646b40e6764ba34a5ebf66aee77af/alias.c#L54).

---
For what it's worth, I encountered this when working in TeamCity, which sets GIT_SSH_COMMAND to
```
ssh -i "path/to/key" -o "StrictHostKeyChecking=no"
```
Another user also encountered this on TeamCity, see [report](https://youtrack.jetbrains.com/issue/TW-55655). This only happens on TeamCity when `teamcity.git.use.native.ssh` is set to `true`.

### LINKED ISSUE #3136: GIT_SSH_COMMAND should be executed in a shell
There are many values of GIT_SSH_COMMAND which will work in git, but won't work in git-lfs. Git uses `sh -c "command"` when running the GIT_SSH_COMMAND, as stated in the [documentation](https://git-scm.com/docs/git#git-codeGITSSHCOMMANDcode) ("interpreted by the shell"). git-lfs, however, parses the string and executes it directly without a shell.

(Relevant code in git: [fill_ssh_args](https://github.com/git/git/blob/c2c7d17b030646b40e6764ba34a5ebf66aee77af/connect.c#L1182), [prepare_shell_cmd](https://github.com/git/git/blob/e3331758f12da22f4103eec7efe1b5304a9be5e9/run-command.c#L254).)

bk2204:
Git documents that the `GIT_SSH_COMMAND` environment variable is passed to the shell, but we're not currently doing that.  This commit fixes it so that we do pass the contents of that variable to the shell with appropriate escaping, if necessary.

This fixes #3136 and probably will fix #3131 as well.

### DIFF
--- a/lfsapi/ssh.go
+++ b/lfsapi/ssh.go
@@ -115,17 +115,29 @@ func (c *sshAuthClient) Resolve(e Endpoint, method string) (sshAuthResponse, err
 	return res, err
 }
 
+func sshFormatArgs(cmd string, args []string, needShell bool) (string, []string) {
+	if !needShell {
+		return cmd, args
+	}
+
+	joined := cmd + " " + strings.Join(tools.ShellQuote(args), " ")
+	return "sh", []string{"-c", joined}
+}
+
 func sshGetLFSExeAndArgs(osEnv config.Environment, e Endpoint, method string) (string, []string) {
-	exe, args := sshGetExeAndArgs(osEnv, e)
+	exe, args, needShell := sshGetExeAndArgs(osEnv, e)
 	operation := endpointOperation(e, method)
 	args = append(args, fmt.Sprintf("git-lfs-authenticate %s %s", e.SshPath, operation))
+	exe, args = sshFormatArgs(exe, args, needShell)
 	tracerx.Printf("run_command: %s %s", exe, strings.Join(args, " "))
 	return exe, args
 }
 
 // Return the executable name for ssh on this machine and the base args
 // Base args includes port settings, user/host, everything pre the command to execute
-func sshGetExeAndArgs(osEnv config.Environment, e Endpoint) (exe string, baseargs []string) {
+func sshGetExeAndArgs(osEnv config.Environment, e Endpoint) (exe string, baseargs []string, needShell bool) {
+	var cmd string
+
 	isPlink := false
 	isTortoise := false
 
@@ -133,12 +145,17 @@ func sshGetExeAndArgs(osEnv config.Environment, e Endpoint) (exe string, basearg
 	sshCmd, _ := osEnv.Get("GIT_SSH_COMMAND")
 	cmdArgs := tools.QuotedFields(sshCmd)
 	if len(cmdArgs) > 0 {
+		needShell = true
 		ssh = cmdArgs[0]
-		cmdArgs = cmdArgs[1:]
+		cmd = sshCmd
 	}
 
 	if ssh == "" {
 		ssh = defaultSSHCmd
+	}
+
+	if cmd == "" {
+		cmd = ssh
 	}
 
 	basessh := filepath.Base(ssh)
@@ -152,10 +169,7 @@ func sshGetExeAndArgs(osEnv config.Environment, e Endpoint) (exe string, basearg
 		isTortoise = strings.EqualFold(basessh, "tortoiseplink")
 	}
 
-	args := make([]string, 0, 5+len(cmdArgs))
-	if len(cmdArgs) > 0 {
-		args = append(args, cmdArgs...)
-	}
+	args := make([]string, 0, 7)
 
 	if isTortoise {
 		// TortoisePlink requires the -batch argument to behave like ssh/plink
@@ -185,7 +199,7 @@ func sshGetExeAndArgs(osEnv config.Environment, e Endpoint) (exe string, basearg
 		args = append(args, sshOptPrefixRE.ReplaceAllString(e.SshUserAndHost, ""))
 	}
 
-	return ssh, args
+	return cmd, args, needShell
 }
 
 const defaultSSHCmd = "ssh"
--- a/tools/str_tools.go
+++ b/tools/str_tools.go
@@ -46,6 +46,22 @@ func QuotedFields(s string) []string {
 	return out
 }
 
+// ShellQuote returns a copied string slice where each element is quoted
+// suitably for sh.
+func ShellQuote(strs []string) []string {
+	dup := make([]string, 0, len(strs))
+
+	for _, str := range strs {
+		// Quote anything that looks slightly complicated.
+		if shellWordRe.FindStringIndex(str) == nil {
+			dup = append(dup, "'"+strings.Replace(str, "'", "'\\''", -1)+"'")
+		} else {
+			dup = append(dup, str)
+		}
+	}
+	return dup
+}
+
 // Ljust returns a copied string slice where each element is left justified to
 // match the width of the longest element in the set.
 func Ljust(strs []string) []string {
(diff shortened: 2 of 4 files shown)


bk2204:
BTW, please feel free to be generous with the critiques.  This is my first submission using Go, so I fully expect to learn a lot about style and best practices.

lfsapi/ssh.go
```diff
@@ -115,32 +115,52 @@ func (c *sshAuthClient) Resolve(e Endpoint, method string) (sshAuthResponse, err
 	return res, err
 }
 
+func sshFormatArgs(cmd string, args []string, needShell bool) (string, []string) {
+	if !needShell {
+		return cmd, args
+	}
+
+	joined := cmd + " " + strings.Join(tools.ShellQuote(args), " ")
+	args = make([]string, 2)
```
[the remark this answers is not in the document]
bk2204: Yeah, I didn't realize that would affect the caller; that would definitely not be what I want to do. Your suggestion not only avoids that, it's also much more compact.

tools/str_tools_test.go
```diff
@@ -71,6 +71,35 @@ func TestQuotedFields(t *testing.T) {
...
+	for desc, c := range map[string]ShellQuoteTestCase{
+		"simple":         {[]string{"foo", "bar", "an_id"}, []string{"foo", "bar", "an_id"}},
+		"leading space":  {[]string{" foo", "bar"}, []string{"' foo'", "bar"}},
+		"trailing space": {[]string{"foo", "bar "}, []string{"foo", "'bar '"}},
+		"internal space": {[]string{"foo bar", "baz quux"}, []string{"'foo bar'", "'baz quux'"}},
+		"backslash":      {[]string{`foo\bar`, `b\az`}, []string{`'foo\bar'`, `'b\az'`}},
+		"quotes":         {[]string{`foo"bar`, "b'az"}, []string{`'foo"bar'`, "'b'\\''az'"}},
+		"mixed quotes":   {[]string{`"foo'ba\"r\"'"`}, []string{`'"foo'\''ba\"r\"'\''"'`}},
+	} {
+		t.Log(desc)
```
ttaylorr: Nit; this line can be removed, since `go test -v` will log the name of the test for you, like so:

```ShellSession
$ make PKGS=tools GO_TEST_EXTRA_ARGS='-v -run TestShellQuote' test
go test -v -run TestShellQuote ./tools
=== RUN   TestShellQuote
=== RUN   TestShellQuote/quotes
=== RUN   TestShellQuote/mixed_quotes
=== RUN   TestShellQuote/simple
=== RUN   TestShellQuote/leading_space
=== RUN   TestShellQuote/trailing_space
=== RUN   TestShellQuote/internal_space
=== RUN   TestShellQuote/backslash
--- PASS: TestShellQuote (0.00s)
    str_tools_test.go:98: quotes
    --- PASS: TestShellQuote/quotes (0.00s)
    str_tools_test.go:98: mixed quotes
    --- PASS: TestShellQuote/mixed_quotes (0.00s)
    str_tools_test.go:98: simple
    --- PASS: TestShellQuote/simple (0.00s)
    str_tools_test.go:98: leading space
    --- PASS: TestShellQuote/leading_space (0.00s)
    str_tools_test.go:98: trailing space
    --- PASS: TestShellQuote/trailing_space (0.00s)
    str_tools_test.go:98: internal space
    --- PASS: TestShellQuote/internal_space (0.00s)
    str_tools_test.go:98: backslash
    --- PASS: TestShellQuote/backslash (0.00s)
PASS
ok      github.com/git-lfs/git-lfs/tools        0.022s
```

tools/str_tools.go
```diff
@@ -46,6 +46,22 @@ func QuotedFields(s string) []string {
...
 
+// ShellQuote returns a copied string slice where each element is quoted
+// suitably for sh.
+func ShellQuote(strs []string) []string {
+	dup := make([]string, 0, len(strs))
+
+	for _, str := range strs {
+		// Quote anything that looks slightly complicated.
+		if shellWordRe.FindStringIndex(str) == nil {
+			dup = append(dup, "'" + strings.Replace(str, "'", "'\\''", -1) + "'")
```
ttaylorr: Nit; `go fmt` encourages that there not be whitespace between `+` and either of its operands.

To fix this, you can either run `make fmt`:

```ShellSession
$ make fmt
tools/str_tools.go
```

Which produces the following diff:

```diff
$ git diff
diff --git a/tools/str_tools.go b/tools/str_tools.go
index 50bba537..14f6debe 100644
--- a/tools/str_tools.go
+++ b/tools/str_tools.go
@@ -54,7 +54,7 @@ func ShellQuote(strs []string) []string {
        for _, str := range strs {
                // Quote anything that looks slightly complicated.
                if shellWordRe.FindStringIndex(str) == nil {
-                       dup = append(dup, "'" + strings.Replace(str, "'", "'\\''", -1) + "'")
+                       dup = append(dup, "'"+strings.Replace(str, "'", "'\\''", -1)+"'")
                } else {
                        dup = append(dup, str)
                }
```

Or, you can run `goimports` (which acts like `go fmt`, but also imports packages that are referenced but not imported) before saving in your editor. FWIW, I use Vim and use https://github.com/fatih/vim-go, which runs a command of your choice on buffer save, (mine is configured [here](https://github.com/ttaylorr/dotfiles/blob/7b0323b5e34bcb7ab623e98be98a1373d5d18d17/vim/.vimrc#L55-L56)).

tools/str_tools_test.go
ttaylorr: I think that I should have added some more detail here. This line:

> ```
>     str_tools_test.go:98: quotes
> ```

is caused by the call to `t.Log()`, whereas this line:

> ```
>     --- PASS: TestShellQuote/quotes (0.00s)
> ```

is caused by the call to `t.Run()`, hence removing the call to `t.Log()` (and the line printed because of it) should still leave enough information to an interested caller of `make test` to figure out what's going on.

bk2204: Sure, I'll do that.

tools/str_tools.go
bk2204: I thought I had run it through `gofmt` using Vim, but apparently I forgot.  Will fix.

PR #3201, opened 2018-08-28
Always provide a Content-Type when uploading files

### LINKED ISSUE #3197: old Content-Type not set by setting lfs.<url>.contenttype false
LFS version:
git-lfs/2.5.1 (GitHub; linux amd64; go 1.10.3)

Remote is managed by Gitlab CE 10.8.6.

Default behavior is that Content-Type is set based on file. But to support older implementation of LFS there seems to be a option introduced to set default Content-Type of  octet-stream.

Both outputs generated with GIT_CURL_VERBOSE=1.

Default:
```
23:14:00.213256 trace git-lfs: HTTP: PUT https://gitlab/group/archive.git/gitlab-lfs/objects/2e7f1651b2767ca6883f4c6ddb1b798fda71d59f1f692ba859dd6d19e871f36f/229
> PUT /group/archive.git/gitlab-lfs/objects/2e7f1651b2767ca6883f4c6ddb1b798fda71d59f1f692ba859dd6d19e871f36f/229 HTTP/1.1
> Host: gitlab
> Authorization: Basic * * * * *
> Content-Length: 229
> Content-Type: application/font-ttf
> User-Agent: git-lfs/2.5.1 (GitHub; linux amd64; go 1.10.3)
> 
```

Using **git config lfs.https://gitlab/group/archive.git.contenttype false**:
```
23:14:35.739336 trace git-lfs: HTTP: PUT https://gitlab/group/archive.git/gitlab-lfs/objects/bd7f67957567d3d7dc99e00c743d63f18a4aa4b13b96850d2969f1e45d43186e/8078
> PUT /group/archive.git/gitlab-lfs/objects/bd7f67957567d3d7dc99e00c743d63f18a4aa4b13b96850d2969f1e45d43186e/8078 HTTP/1.1
> Host: gitlab
> Authorization: Basic * * * * *
> Content-Length: 8078
> User-Agent: git-lfs/2.5.1 (GitHub; linux amd64; go 1.10.3)
> 
```
To me it seems that with commit 938df5c47f343279e8ece0aa5f86e57be4dbdf11 there will be no Content-Type set at all.

bk2204:
As of v2.5.1, we provide an option to disable automatic detection of the Content-Type and always use application/octet-stream instead.  However, when that option is set, instead of providing a Content-Type of application/octet-stream, we fail to provide one at all.

Ensure that we always set some Content-Type header when uploading files by skipping the detection but preserving the fallback behavior when the option is set.

The diff here is probably easier to understand with `git diff -w`.
This fixes #3197.


tq/basic_upload.go
```diff
@@ -150,19 +150,23 @@ func (a *basicUploadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb Progres
...
+		n, err := r.Read(buffer)
+		if err != nil && err != io.EOF {
+			return errors.Wrap(err, "content type detect")
+		}
 
-	contentType := http.DetectContentType(buffer[:n])
-	if _, err := r.Seek(0, 0); err != nil {
-		return errors.Wrap(err, "content type rewind")
+		contentType = http.DetectContentType(buffer[:n])
+		if _, err := r.Seek(0, 0); err != nil {
```
ttaylorr: Nit; we should be using [constants in the `io` package for the seek whence](https://golang.org/pkg/io/#pkg-constants) instead of the literal `0` as the second formal parameter to this function call. For example, I recommend:

```go
if _, err := r.Seek(0, io.SeekStart); err != nil {
        // ...
}
```

I suggest this only because your diff is present around this area, even though you haven't changed this line specifically. If you would like to adopt this change, perhaps you could send it as a preparatory clean-up commit?

bk2204: I've put in a preparatory commit for that. Thanks for the suggestion.

ttaylorr:
I restarted the TravisCI builds, which appear to have timed out on one
particular sub-build.

Once that builds successfully, please feel free to merge this. I think
that CircleCI did not pick up the changes, but if they build on TravisCI
then I think that is sufficient confidence-building.

Lem:
Thanks for taking care of this bug.

As the milestone was set to 2.5.2 is there any plan when 2.5.2 will be released?

ttaylorr:
> As the milestone was set to 2.5.2 is there any plan when 2.5.2 will be
> released?

Thanks for asking. We don't currently have a set date for releasing
v2.5.2, and we would likely roll this release into v2.6.0 if we felt
that that was ready sooner.

In the meantime, you can build a copy of Git LFS for yourself and
install it by running:

    $ git clone git@github.com:git-lfs/git-lfs.git
    $ cd git-lfs
    $ make
    $ cp bin/git-lfs $(which git-lfs)

PR #3202, opened 2018-08-28
Tidy misformatted files

bk2204:
There are a few misformatted files in the repository.  Run these files through gofmt to reduce diff noise for future developers.

### DIFF
--- a/lfs/gitscanner_log.go
+++ b/lfs/gitscanner_log.go
@@ -29,8 +29,8 @@ var (
 	// lfs changes and format the output suitable for parseLogOutput.. method(s)
 	logLfsSearchArgs = []string{
 		"-G", "oid sha256:", // only diffs which include an lfs file SHA change
-		"-p",   // include diff so we can read the SHA
-		"-U12", // Make sure diff context is always big enough to support 10 extension lines to get whole pointer
+		"-p",                             // include diff so we can read the SHA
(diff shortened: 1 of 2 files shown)


bk2204:
Okay, so I know what's happened here.  I have upgraded to go 1.11, and it apparently formats things differently than go 1.10.  Running the gofmt from 1.10 does indeed produce no changes.

From looking at the source for goimports, it just calls the AST serializer for the go version it's running on, which in my case is 1.11, which is why `make fmt` produced this diff.

I'm not sure how we want to move forward on this; I'm fine dropping it if you'd prefer, or I can resubmit with a fix to the documentation to indicate that we prefer the Go 1.11 gofmt.  I'm open to other suggestions as well.

bk2204:
Okay, I've rebased this onto the master branch, which is now using Go 1.11, and updated the message to indicate that it fixes the issues for the new gofmt.

ttaylorr:
> Okay, I've rebased this onto the master branch, which is now using Go
> 1.11 [ ... ]

Thanks, I was able to check this out myself and confirm that running
'make fmt' did not produce any diff from the latest (which, at the time
of writing is 79577242 (Tidy files with Go 1.11's gofmt, 2018-08-28)).

So, I think with the upgrade to Go 1.11, this is ready to go.

PR #3208, opened 2018-08-29
all: use Go modules instead of Glide

ttaylorr:
This pull request leverages #3203 and packages Git LFS as a Go 1.11 Go Module, thereby no longer using Glide.

The changes herein are described in detail in each commit, but a few notable things stick out that are worth mentioning at the top-level:

* 1819e67, & f8ffbb5: these are both preparatory commits, the first fixing a broken Makefile recipe, and the second removing a long-unused environment setting from the rules file in Debian builds.

* cd81c77: this is where the majority of the change happened. In particular:

  * `_test.go` files (and their friends) have been removed, since `go mod` does not keep `_test.go` files in the vendor tree, unlike `glide`, which did.

  * We chose to keep `$GO111MODULE` unset, to indicate that the `go mod` can automatically detect that this repository does contain a Go module, and detect it as such. This requires the repository to be checked out outside of your system `$GOPATH`, or otherwise you must set `$GO111MODULE` to `on` (the later will break in a subsequent version of Go).

  * Keeping with this theme, since we use `dh_golang` to build on Debian, and it does not yet have support for Go modules, let's explicitly enable `$GO111MODULE` to force Go on Debian builds to treat the repository as a Go module, even though the checkout takes place in a `$GOPATH`.

##

/cc @git-lfs/core


bk2204:
I did some testing on this on an Ubuntu 18.04 (bionic) Docker container.  My testing shows that this change does break building there, since Go 1.11 isn't yet available, and consequently Go module support is missing.  I don't think that's fatal, especially since Debian and Ubuntu don't use our vendored dependencies, but I thought I'd point it out anyway.

ttaylorr:
> I don't think that's fatal, especially since Debian and Ubuntu don't
> use our vendored dependencies, but I thought I'd point it out anyway.

Thanks for testing, and I agree with your assessment. I think that we
share the perspective that it's OK that the vendor tree is used instead
of go.mod, since it means that:

  1. We can compile out of the box, and

  2. If someone wants to build a package on that platform, they won't
     use either, anyway.

Thanks for your help pairing on this! :-).

Issue #3211, opened 2018-08-29
core.hooksPath isn't expanded


ktchen14:
In my `gitconfig` I have:

```ini
[core]
	hooksPath = ~/.githooks
```

This leads to `git lfs` creating a directory called `~` wherever I attempt to use it. The problem appears to be that the tilde in the hook directory isn't expanded.

bk2204:
Thanks for reporting this.  I can indeed confirm that we're not expanding paths here.

Let me see if I can come up with a fix for this.

PR #3212, opened 2018-08-30
config: expand core.hooksPath

ttaylorr:
This pull request expands the value of `core.hooksPath` as read from the Git configuration, before it is evaluate by `git lfs install`.

In practice, this makes it possible to do `git config core.hooksPath ~/path/to/hooks`, and Git LFS will properly look inside `/path/to/hooks` from within `$HOME`. It also supports incantations such as `~user/path/to/hooks`.

Closes: https://github.com/git-lfs/git-lfs/issues/3211.

##

/cc @git-lfs/core 
/cc @ktchen14 https://github.com/git-lfs/git-lfs/issues/3211

### DIFF
--- a/tools/filetools.go
+++ b/tools/filetools.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/user"
 	"path/filepath"
 	"runtime"
 	"strconv"
@@ -15,6 +16,7 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"github.com/git-lfs/git-lfs/errors"
 	"github.com/git-lfs/git-lfs/filepathfilter"
 )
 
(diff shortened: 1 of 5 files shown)


tools/filetools.go
```diff
@@ -113,6 +115,60 @@ func CleanPaths(paths, delim string) (cleaned []string) {
...
+	} else {
+		username = path[1:]
+	}
+
+	var (
+		who *user.User
+		err error
+	)
+	if len(username) == 0 {
+		who, err = currentUser()
```
bk2204: There's a tricky edge case here that we haven't considered.  If we have just a tilde (e.g. `~/hooks`), we need to use the `$HOME` environment variable in preference to what the OS (e.g. `/etc/passwd`) tells us the user's home directory is.  This is Git's behavior (and the shell's) and I suspect people will want us to match it.

ttaylorr: That makes a lot of sense. I was initially mislead because even though I thought that Git's behavior was to use `$HOME` instead of `/etc/passwd` or similar, that that _is_ what Go's `user.Current()` was doing, because of [this][1] line.

But, I was puzzled since we do set `HOME="$TESTHOME"` before running each test's subshell. So, before sending this patch I tried to look into it myself, but couldn't figure out why the two disagreed.

Because of Go's conditional compilation, the line mentioned above wasn't what we were executing. Instead, we were running [this][2], which is in a different file (and uses `getpw()` instead of `os.Getenv("HOME")`).

So, I was able to re-wrap this function so that we effectively do:

```go
func getCurrentUser() (*user.User, error) {
        u, err := user.Current()
        if err != nil {
                return nil, err
        }
        u.HomeDir = os.Getenv("HOME")
        return u, nil
}
```

and call that, instead of refactoring the code to support one case that uses package `os/user`, and one that does not.

[1]: https://github.com/golang/go/blob/go1.11/src/os/user/lookup_stubs.go#L27
[2]: https://github.com/golang/go/blob/go1.11/src/os/user/cgo_lookup_unix.go#L31-L34

Issue #3213, opened 2018-08-30
git lfs ls-files <commit> shows files from current commit


jlisee:
`git lfs ls-files <commit>` shows me more just the lfs files in a specific commit, it appears to show me the files in the current commit as well.

Lets say you have a repository and you do setup tracking and check that you see no files at your first commit:

```
$ git lfs track "*.psd"
Tracking "*.psd"
$ git add .gitattributes
$ git commit -m "attributes"
[master (root-commit) 4a725e5] attributes
 1 file changed, 1 insertion(+)
 create mode 100644 .gitattributes
$ git lfs ls-files 4a725e5
```

Then you add a file and make sure that gets listed properly:

```
$ echo foo > file.psd
$ git add file.psd
$ git commit -m "Add design file"
[master 16deebf] Add design file
 1 file changed, 3 insertions(+)
 create mode 100644 file.psd
$ git lfs ls-files
b5bb9d8014 * file.psd
```

Now if you check back at the first commit you should see nothing but...
```
$ git lfs ls-files 4a725e5
b5bb9d8014 * file.psd
```

You have to actually checkout that commit to get the right results:

```
$ git checkout -b first-commit 4a725e5
$ git lfs ls-files 4a725e5
```

While in that commit you can look forward and things are correct:

```
$ git lfs ls-files 16deebfd
b5bb9d8014 - file.psd
```

Version/Platform: Ubuntu Linux 14.04 git 2.18.0, git-lfs 2.5.1.

See [here](https://gist.github.com/jlisee/4dd921ebc32da3254b1e61313e7d0cce) for the full history.

ttaylorr:
> `git lfs ls-files <commit>` shows me more just the lfs files in a
> specific commit, it appears to show me the files in the current commit
> as well.

Thanks for the report, from reading the implementation of 'git lfs
ls-files' commands/command_ls_files.go, I can confirm that this is what
happens.

I think that this behavior was introduced by me in bcd97eef (commands:
scan all history with 'git lfs ls-files --all', 2017-12-19), and should
be changed to only do so if the given argument ref is equal to the
current ref.

For example:

  if ref == fullref {
    if err := gitscanner.ScanIndex(ref, nil); err != nil { ... }
  }

I believe that this would more closely match what Git does in
builtin/ls-files.c, but I'm not certain.

Perhaps @bk2204 or @PastelMobileSuit could chime in?

bk2204:
[...]

PR #3219, opened 2018-08-31
commands/command_ls_files.go: ignore index with argument

ttaylorr:
This pull request teaches `git lfs ls-files` to not traverse the index for Git LFS objects when given an explicit reference that is not equal to the current checkout.

Closes: https://github.com/git-lfs/git-lfs/issues/3213.

##

/cc @git-lfs/core 
/cc @jlisee, https://github.com/git-lfs/git-lfs/issues/3213

### DIFF
--- a/commands/command_ls_files.go
+++ b/commands/command_ls_files.go
@@ -21,19 +21,23 @@ var (
 func lsFilesCommand(cmd *cobra.Command, args []string) {
 	requireInRepo()
 
-	var ref string
+	var ref *git.Ref
 
 	if len(args) == 1 {
+		var err error
 		if lsFilesScanAll {
 			Exit("fatal: cannot use --all with explicit reference")
 		}
-		ref = args[0]
+		ref, err = git.ResolveRef(args[0])
+		if err != nil {
+			Exit("fatal: could not resolve ref %s", args[0])
+		}
 	} else {
 		fullref, err := git.CurrentRef()
 		if err != nil {
-			ref = git.RefBeforeFirstCommit.Sha
+			ref = git.RefBeforeFirstCommit
 		} else {
-			ref = fullref.Sha
+			ref = fullref
 		}
 	}
 
@@ -88,7 +92,7 @@ func lsFilesCommand(cmd *cobra.Command, args []string) {
 	includeArg, excludeArg := getIncludeExcludeArgs(cmd)
 	gitscanner.Filter = buildFilepathFilter(cfg, includeArg, excludeArg)
 
-	if err := gitscanner.ScanIndex(ref, nil); err != nil {
+	if err := gitscanner.ScanIndex(ref.Sha, nil); err != nil {
 		Exit("Could not scan for Git LFS index: %s", err)
 	}
 	if lsFilesScanAll {
@@ -98,9 +102,9 @@ func lsFilesCommand(cmd *cobra.Command, args []string) {
 	} else {
 		var err error
 		if lsFilesScanDeleted {
-			err = gitscanner.ScanRefWithDeleted(ref, nil)
+			err = gitscanner.ScanRefWithDeleted(ref.Sha, nil)
 		} else {
-			err = gitscanner.ScanTree(ref)
+			err = gitscanner.ScanTree(ref.Sha)
 		}
 
 		if err != nil {
--- a/commands/command_ls_files.go
+++ b/commands/command_ls_files.go
@@ -36,7 +36,16 @@ func lsFilesCommand(cmd *cobra.Command, args []string) {
 	if len(args) == 1 {
 		if lsFilesScanAll {
 			Exit("fatal: cannot use --all with explicit reference")
+		} else if args[0] == "--all" {
+			// Since --all is a valid argument to "git rev-parse",
+			// if we try to give it to git.ResolveRef below, we'll
+			// get an unexpected result.
+			//
+			// So, let's check early that the caller invoked the
+			// command correctly.
+			Exit("fatal: did you mean \"git lfs ls-files --all --\" ?")
 		}
+
 		ref, err = git.ResolveRef(args[0])
 		if err != nil {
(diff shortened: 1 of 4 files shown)


commands/command_ls_files.go
```diff
@@ -105,8 +105,16 @@ func lsFilesCommand(cmd *cobra.Command, args []string) {
 	includeArg, excludeArg := getIncludeExcludeArgs(cmd)
 	gitscanner.Filter = buildFilepathFilter(cfg, includeArg, excludeArg)
 
-	if err := gitscanner.ScanIndex(ref.Sha, nil); err != nil {
-		Exit("Could not scan for Git LFS index: %s", err)
+	if currentRef.Sha == ref.Sha {
+		// Only scan the index when "git lfs ls-files" was invoked with
+		// no arguments, or a reference pointing to the HEAD checkout
+		// (or before the first commit).
```
bk2204: Do we still want to scan the index even if the user has specified an explicit reference pointing to HEAD?  I would think that we'd only want to list the index if no treeish was given.

Consider the case that we have a branch also pointing to the same commit as this branch, or the case of a scripting tool which has provided some arbitrary resolved ref which happens to resolve to HEAD (e.g. a tag).

ttaylorr:
> Do we still want to scan the index even if the user has specified an
> explicit reference pointing to HEAD?  I would think that we'd only
> want to list the index if no treeish was given.

I think that you're right: we should scan the index iff no treeish was
given.

I'm thinking particularly about callers who want avoid an index scan,
and how they might do so if they still want to list the history
reachable via the current checkout:

  * `git lfs ls-files`: lists history from HEAD + the index
  * `git lfs ls-files HEAD`: lists history from HEAD only

I dropped the first three commits from this change, and amended the last
one, so I think that this change is ready to go now.

ttaylorr:
<details>
<summary><code>git range-diff 60bd1f35...cc7b83a9</code></summary>
<pre>
1:  cd49a517 < -:  -------- git/git.go: make RefBeforeFirstCommit a *git.Ref
2:  a3d07f66 < -:  -------- commands/command_ls_files.go: store argument as *git.Ref
3:  c9ca18f9 < -:  -------- commands/command_ls_files.go: initialize currentRef early
4:  5b136014 ! 1:  6db67f2c commands/command_ls_files.go: don't accept "git lfs ls-files -- --all"
    @@ -32,9 +32,9 @@
     +			Exit("fatal: did you mean \"git lfs ls-files --all --\" ?")
      		}
     +
    - 		ref, err = git.ResolveRef(args[0])
    - 		if err != nil {
    - 			Exit("fatal: could not resolve ref %s", args[0])
    + 		ref = args[0]
    + 	} else {
    + 		fullref, err := git.CurrentRef()
     
     diff --git a/t/t-ls-files.sh b/t/t-ls-files.sh
     --- a/t/t-ls-files.sh
5:  60bd1f35 ! 2:  cc7b83a9 commands/command_ls_files: hide index with arguments
    @@ -12,11 +12,11 @@
         However, when given an argument, 'git lfs ls-files' will traverse
         starting from a specific revision, so showing the index unconditionally
         causes a problem here. It does not make sense to show the current state
    -    of the index when listing files from HEAD~N (where N is non-zero).
    +    of the index when listing files when given a starting point explicitly.
     
    -    So, let's keep track of both the current revision, and the revision that
    -    we were asked to show, and only scan the index when the two are the
    -    same.
    +    So, let's skip calling into the gitscanner code when given an explicit
    +    reference to start from (even if that reference is HEAD, since this
    +    likely indicates that the user does not want to scan the index, too).
     
     diff --git a/commands/command_ls_files.go b/commands/command_ls_files.go
     --- a/commands/command_ls_files.go
    @@ -25,16 +25,15 @@
      	includeArg, excludeArg := getIncludeExcludeArgs(cmd)
      	gitscanner.Filter = buildFilepathFilter(cfg, includeArg, excludeArg)
      
    --	if err := gitscanner.ScanIndex(ref.Sha, nil); err != nil {
    +-	if err := gitscanner.ScanIndex(ref, nil); err != nil {
     -		Exit("Could not scan for Git LFS index: %s", err)
    -+	if currentRef.Sha == ref.Sha {
    ++	if len(args) == 0 {
     +		// Only scan the index when "git lfs ls-files" was invoked with
    -+		// no arguments, or a reference pointing to the HEAD checkout
    -+		// (or before the first commit).
    ++		// no arguments.
     +		//
     +		// Do so to avoid showing "mixed" results, e.g., ls-files output
     +		// from a specific historical revision, and the index.
    -+		if err := gitscanner.ScanIndex(ref.Sha, nil); err != nil {
    ++		if err := gitscanner.ScanIndex(ref, nil); err != nil {
     +			Exit("Could not scan for Git LFS index: %s", err)
     +		}
      	}
    @@ -76,6 +75,33 @@
     +  [ 0 -eq "$(grep -c "c.txt" ls-files.log)" ]
     +)
     +end_test
    ++
    ++begin_test "ls-files: non-HEAD reference referring to HEAD ignores index"
    ++(
    ++  set -e
    ++
    ++  reponame="ls-files-HEAD-ish-ignores-index"
    ++  git init "$reponame"
    ++  cd "$reponame"
    ++
    ++  git lfs track "*.txt"
    ++  echo "a.txt" > a.txt
    ++  echo "b.txt" > b.txt
    ++
    ++  git add .gitattributes a.txt
    ++  git commit -m "a.txt: initial commit"
    ++
    ++  tagname="v1.0.0"
    ++  git tag "$tagname"
    ++
    ++  git add b.txt
    ++
    ++  git lfs ls-files "$tagname" 2>&1 | tee ls-files.log
    ++
    ++  [ 1 -eq "$(grep -c "a.txt" ls-files.log)" ]
    ++  [ 0 -eq "$(grep -c "b.txt" ls-files.log)" ]
    ++)
    ++end_test
     +
      begin_test "ls-files: outside git repository"
      (
</pre>
</details>

PR #3228, opened 2018-09-05
t: mark test sources as .PHONY

ttaylorr:
This pull request makes `t/t-*.sh` targets `.PHONY`, so that we can repeatedly invoke them as `make t-*.sh` without having to pass `-B`.

##

/cc @git-lfs/core


bk2204:
Are you reading my mind?  This is a feature I just needed today.

ttaylorr:
🔮

PR #3235, opened 2018-09-07
Add support for core.sshCommand

### LINKED ISSUE #2215: Support core.sshCommand in gitconfig
I have issues with SSH on my corp machine + sourcetree's environment config.

By configuring the git config and setting the sshCommand I can fix this, unfortunately this setting doesn't isn't read by git-lfs

Currently having to do an ugly wrap around git-lfs to get this working

bk2204:
Git has support for specifying `core.sshCommand` in the configuration file.  Teach git-lfs support for that command, too, and ensure that it’s passed to the shell as documented.

This should fix #2215.

### DIFF
--- a/lfsapi/ssh.go
+++ b/lfsapi/ssh.go
@@ -81,7 +81,7 @@ func (c *sshAuthClient) Resolve(e Endpoint, method string) (sshAuthResponse, err
 		return res, nil
 	}
 
-	exe, args := sshGetLFSExeAndArgs(c.os, e, method)
+	exe, args := sshGetLFSExeAndArgs(c.os, c.git, e, method)
 	cmd := exec.Command(exe, args...)
 
 	// Save stdout and stderr in separate buffers
@@ -124,8 +124,8 @@ func sshFormatArgs(cmd string, args []string, needShell bool) (string, []string)
 	return "sh", []string{"-c", joined}
 }
 
-func sshGetLFSExeAndArgs(osEnv config.Environment, e Endpoint, method string) (string, []string) {
-	exe, args, needShell := sshGetExeAndArgs(osEnv, e)
+func sshGetLFSExeAndArgs(osEnv config.Environment, gitEnv config.Environment, e Endpoint, method string) (string, []string) {
+	exe, args, needShell := sshGetExeAndArgs(osEnv, gitEnv, e)
 	operation := endpointOperation(e, method)
 	args = append(args, fmt.Sprintf("git-lfs-authenticate %s %s", e.SshPath, operation))
 	exe, args = sshFormatArgs(exe, args, needShell)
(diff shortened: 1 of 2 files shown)


lfsapi/ssh.go
```diff
@@ -81,7 +81,7 @@ func (c *sshAuthClient) Resolve(e Endpoint, method string) (sshAuthResponse, err
 		return res, nil
 	}
 
-	exe, args := sshGetLFSExeAndArgs(c.os, e, method)
+	exe, args := sshGetLFSExeAndArgs(c.os, c.git, e, method)
```
ttaylorr: I wonder if we could teach this function to accept instead a `*config.Configuration` as part of a future change. The type `*config.Configuration` has both a `.OS` and `.Git` environment attached to it, so I think that it might make more sense to pass that as one argument than these two.

I took a quick look through the code, and I think that this is better suited as a future change (instead of one in this pull request) since the callers run pretty deep. From my cursory look, I think we'd need to:

1. Change the signature of `lfsapi.Config` to instead have a `Config() *config.Configuration`.
2. Change the callers of `lfsapi.NewConfig()`. Potentially we could remove this function (and type) entirely, and instead use the [options pattern](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) (a good example is [here](https://github.com/git-lfs/git-lfs/blob/v2.5.1/git/githistory/rewriter.go#L149-L173) and [here](https://github.com/git-lfs/git-lfs/blob/v2.5.1/git/githistory/rewriter.go#L196-L198)).
3. Make the rest of the changes.

It also might be worth _not_ doing this, since I think that there is a version of `NewContext` that takes two `map[string]string` and emulates an OS/Git configuration environment. It might be worth exposing something similar from the `config` package and using that, since I think that this facility is useful in places other than package `lfsapi`.

lfsapi/ssh.go
```diff
@@ -124,34 +124,44 @@ func sshFormatArgs(cmd string, args []string, needShell bool) (string, []string)
...
 	tracerx.Printf("run_command: %s %s", exe, strings.Join(args, " "))
 	return exe, args
 }
 
+// Parse command, and if it looks like a valid command, return the ssh binary
+// name, the command to run, and whether we need a shell.  If not, return
+// existing as the ssh binary name.
+func sshParseShellCommand(command string, existing string) (ssh string, cmd string, needShell bool) {
+	ssh = existing
+	cmdArgs := tools.QuotedFields(command)
```
ttaylorr: Nit: to limit the scope of `cmdArgs`, you could also write:

```go
if cmdArgs := tools.QuotedFields(command); len(cmdArgs) > 0 {
        // ...
}
```

But I don't suspect that this function will change much in the future, so I think that you'd be equally fine to ignore this suggestion.

lfsapi/ssh.go
bk2204: I think that would be a nice improvement, but I agree that would be better in a different pull request.  I thought about doing that in this change, but didn't because it would (as you mention) involve a significant refactor that seemed like more than I wanted to do here.

I also think it might conflict with some work to refactor the HTTP client that @PastelMobileSuit is doing, so deferring that change until after his code lands would probably be more courteous.

lfsapi/ssh.go
bk2204: Yeah, that would be a nice improvement.  I’ll fix and resubmit.

lfsapi/ssh.go
ttaylorr: Sounds good. Let's wait for the changes that @PastelMobileSuit is working and merge this whenever you are ready to do so.

PR #3236, opened 2018-09-07
gitobj-based Object Scanner

bk2204:
For performance and simplicity reasons, we'd like to use gitobj in our object scanner instead of invoking `git cat-file --batch`.  This series introduces code to do that, as well as a correctness fix for worktrees (discovered by the test suite).  We also update the go.mod file and the dependencies (in their own commit) to pull in the latest gitobj changes, which are required for this series to function as expected.


bk2204:
It looks like the failure for the latest Git build is due to our writing `printf "$var"` instead of `printf "%s" "$var"`; at least in some situations (although I'm unable to tell exactly what from the log), we accidentally insert format characters into the first argument to `printf`, which predictably works out poorly.  I’ll send in an additional PR to fix that in our integration test suite for improved robustness.

I think the build with the earliest supported Git is indeed broken, and I’ll need to spend some time figuring out why that is.

git/git.go
```diff
@@ -654,6 +654,22 @@ func GitDir() (string, error) {
 	return "", nil
 }
 
+func GitCommonDir() (string, error) {
+	cmd := gitNoLFS("rev-parse", "--git-common-dir")
+	out, err := cmd.Output()
+	if err != nil {
```
ttaylorr: We have a `IsVersionAtLeast()`, though I suspect that the way it is currently is better, since we'll behave more gracefully in the face of _any_ error by falling back to `GitDir()`.

git/object_scanner.go
```diff
@@ -52,55 +41,26 @@ type ObjectScanner struct {
...
-	if err := cmd.Start(); err != nil {
 		return nil, err
 	}
 
 	return &ObjectScanner{
-		from: bufio.NewReaderSize(stdout, 16384),
-		to:   stdin,
-
-		closeFn: closeFn,
+		gitobj: gitobj,
```
ttaylorr: Does it make sense here to instead say:

```go
return NewObjectScannerFrom(gitobj), nil
```

instead? I think the two are synonyms right now, but potentially the body of the later would change and we'd minimize the diff later.

git/object_scanner.go
```diff
@@ -202,36 +143,34 @@ func IsMissingObject(err error) bool {
...
-	}
-
-	l, err := s.from.ReadBytes('\n')
+	var (
+		obj      gitobj.Object
+		size     int64
+		contents *io.LimitedReader
+	)
+
+	oidbytes, _ := hex.DecodeString(oid)
```
ttaylorr: Nice, I think that it makes sense to ignore this error, since it'll turn into a `&missingErr` shortly after anyway.

Since we throw this value away, I wonder if it makes sense to introduce a `Must*` version of this function, a-la `regexp.Compile` and `regexp.MustCompile`. Maybe:

```go
func mustDecode(s string) []byte {
        x, _ := hex.DecodeString(s)
        return x
}
```

which we could then use directly as:

```go
obj, err := s.gitobj.Object(mustDecode(oid))
```

This isn't _quite_ true to the spirit of `regexp.MustCompile`'s pattern, since the standard library's implementation `panic()`s if `err != nil`, but I think that this is close enough.

Alternatively, maybe we should introduce `ObjectString`, `BlobString`, (etc.) onto `gitobj`. What do you think?

git/object_scanner.go
```diff
@@ -170,22 +127,6 @@ func (s *ObjectScanner) Err() error { return s.err }
 // necessary) and clearing both the object and error fields on the
 // `*ObjectScanner` instance.
 func (s *ObjectScanner) reset() error {
```
ttaylorr: I think that we might be able to get rid of this function entirely. This used to be used to "advance" the read-head to the end of an object in the event that a caller asked for more before reading the object that they had. I believe that the reason that we used a limited reader is that we copied the remainder of the contents to `ioutil.Discard`, but didn't want to accidentally spill over into the next object.

I think that we used an `*io.LimitedReader` as well here because we wanted to be able to call `.N` to figure out how large the object was. If we remove this function, it should be possible to remove the call to `io.LimitReader`, and instead make the file a pure `io.Reader`.

git/object_scanner.go
bk2204: I think a mustDecode function is a good idea here.

git/git.go
bk2204: It's funny you should mention that, because I think this is the cause of the failures with older Git.  I just noticed this:

```console
$ git rev-parse --random; echo $?
--random
0
```

Granted, this is on Git 2.19, but assuming that the behavior hasn't changed, that's definitely not going to provide a helpful response on older Git.

git/object_scanner.go
bk2204: I’ll take a look.  I always love deleting unused code.

bk2204:
As for performance, it seems (at least on our test repo) that there doesn't seem to be a significant performance change either way.

git/object_scanner.go
bk2204: Yup, looks like both of those things can be cleaned up.

git/object_scanner.go
bk2204: Fixed.

PR #3242, opened 2018-09-10
t: avoid using shell variables in printf's first argument

bk2204:
The printf(1) command, like it's C cousin, takes a format string as its
first argument.  If a shell variable is passed as the first argument, it
will be interpreted as a format string; this can lead to surprising
behavior and can cause the test suite to fail if we accidentally insert
a format string character into the variable.

Modify all the places in the individual tests that we use a plain quoted
variable as the format string by running the following Ruby one-liner:

  ruby -i -pe '$_.gsub!(/printf "\$/, %q(printf "%s" "$))' t/t-*.sh

Avoid modifying the test helpers, as there are places (such as calc_oid)
where we want to pass text containing escapes (such as "\n") and have
those be properly interpreted by printf(1).

### DIFF
--- a/t/t-batch-error-handling.sh
+++ b/t/t-batch-error-handling.sh
@@ -29,7 +29,7 @@ begin_test "batch error handling"
   contents="a"
   contents_oid=$(calc_oid "$contents")
 
-  printf "$contents" > a.dat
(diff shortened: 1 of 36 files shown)


bk2204:
Thanks.

Issue #3107, opened 2018-07-05
git lfs locks --local data improvement suggestions


mstrap:
We would like to implement Git LFS locks functionality for our GUI client. One basic feature would be to display "known" locks. I was hoping to find this information using `git lfs locks --local`, but unfortunately this information is inappropriate or not available at all.

**Problem**

(1) `git lfs locks --local` will report for every *own* lock, the last known state after having invoked `git lfs lock/unlock <file>`. If in the meanwhile locks in the repository have changed (because they were forcefully unlocked or the lock was released from a different clone), this information becomes outdated and it will *remain outdated* even after invoking `git lfs locks` to query updated locks information.

(2) `git lfs locks --local` does not report locks of other users.

**Suggested solution**

To follow the distributed, offline-capable approach of Git in general, `git lfs locks --local` could mirror the last known information queried by `git lfs locks`. 

**Why solve in Git LFS code?**

(1) This would make Git LFS binaries somewhat more robust in case of offline usage and allow the user to still see last known locks information, without requiring a connection to the server.

(2) It will provide an API for (GUI) clients which may want to enrich `git status` information by Git LFS locks information. As `git status` information is purely local, it will be convenient to have the additional locks information be present locally, too, to report the combined result in a single run (without requiring a connection to the server). If Git LFS itself won't provide this information, this would probably mean that interested clients will have to implement their own caches (which will be harder to implement from outside of Git LFS code and which are probably more likely to become inconsistent).

ttaylorr:
Hi @mstrap, thanks for opening this.

> **Suggested solution**
> 
> To follow the distributed, offline-capable approach of Git in general, `git lfs locks --local` could mirror the last known information queried by `git lfs locks`.

I think that this is a great idea, and we could definitely support it without too much craziness :-).

I am thinking that after every `git lfs locks` that contacts the remote server, we could update a file in `.git/lfs/cache/locks` (the `cache` sub-tree does not yet exist or serve any purpose) and hence read from that instead of the current use of `--local`.

`.git/lfs/cache/locks` could store the JSON response that we got from the remote verbatim, and would be updated when `git lfs locks` is issued. @PastelMobileSuit & @larsxschneider: what do you think of this? Would one of you be interested in implementing it?

[...]

Issue #3252, opened 2018-09-15
git-lfs locks should optionally denote own locks


mstrap:
Currently, for a Git client which is accessing Git-LFS functionality programmatically, there is no reliable way to  figure out which files are locked for the user on the server.

**Problem**

* `git lfs locks --local` may report outdated information in various cases:
  * when invoked by the same user from a different clone
  * if `.git/lfs/lockcache.db` is lost
  * if someone else has stolen locks on the server
* `git lfs locks` reports served-side user names which may be unrelated to local Git `user.name`. Only the user (who is aware of all his logins) will be able to interpret this information correctly. [1]

**Suggested solution**

Offer an option `git lfs locks --verified` which distinguishes between "ours" vs. "theirs" locks for its output. For example by just extending current output by  `ours` and `theirs` markers: 

    [theirs] huge2  Sample User     ID:25
    [theirs] huge3  Sample User     ID:29
    [ours]   huge4  MStr            ID:167

**Notes**

* With current API (version 2.5), the `/locks/verify` endpoint should provide sufficient information to implement this functionality without requiring API changes.

* The order in which locks are listed may be different compared to a plain `git lfs --locks`: the most simple approach will probably be to report locks in the same order as received by the `/locks/verify` response.

* It should be possible to combine this option with `--cached` locks information (as suggest in #3107), either by adding a forth *alternative* option here, in addition to `git lfs locks` (without option), `--local` option and proposed `--cached` option, or by adding an *orthogonal* option here.

Advantages of implementing as *alternative* option: 
  * no invalid combinations, like `--local --verified`
  * single `.git/lfs/cache/locks` file holding most recent data, which can't contradict with possible `.git/lfs/cache/locks-verified` data
  * probably a bit easier to implement

Advantages of implementing as *orthogonal* option: 
  * clear distinction between locks data with and without ours-vs-theirs information
  * programmatically processing locks information may be easier, because the consumer of the information will know from the beginning which kind of received information to expect (instead of having to "learn" on-the-fly whether locks information is with or without ours-vs-theirs information)

[1] For example, my login for our GitLab instance (version 10.8.4-ee) is "mstrap" but my display name in the GitLab UI is "MStr". GitLab will report lock information using my display name which is present nowhere on the client side:

    {"locks":{"path":"huge4","id":"80","locked_at":"2018-09-18T17:28:05Z","owner":{"name":"MStr"}}}

mstrap:
[...]

PR #3253, opened 2018-09-15
locking: cache JSON response from server

mstrap:
.git/lfs/cache/locks stores the JSON response that we got from the remote verbatim and will be updated when "git lfs locks" is issued

Partially fixes https://github.com/git-lfs/git-lfs/issues/3107

### DIFF
--- a/commands/command_locks.go
+++ b/commands/command_locks.go
@@ -32,7 +32,19 @@ func locksCommand(cmd *cobra.Command, args []string) {
 	lockClient.RemoteRef = refUpdate.Right()
 	defer lockClient.Close()
 
-	locks, err := lockClient.SearchLocks(filters, locksCmdFlags.Limit, locksCmdFlags.Local)
+	if locksCmdFlags.Cached {
+		if locksCmdFlags.Limit > 0 {
+			Exit("--cached option can't be combined with --limit")
+		}
+		if len(filters) > 0 {
+			Exit("--cached option can't be combined with filters")
+		}
+		if locksCmdFlags.Local {
+			Exit("--cached option can't be combined with --local")
+		}
+	}
+
+	locks, err := lockClient.SearchLocks(filters, locksCmdFlags.Limit, locksCmdFlags.Local, locksCmdFlags.Cached)
 	// Print any we got before exiting
 
 	if locksCmdFlags.JSON {
@@ -93,6 +105,9 @@ type locksFlags struct {
 	Local bool
 	// JSON is an optional parameter to output data in json format.
 	JSON bool
+	// for non-local queries, report cached query results from the last query
+	// instead of actually querying the server again
+	Cached bool
 }
 
 // Filters produces a filter based on locksFlags instance.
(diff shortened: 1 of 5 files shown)


mstrap:
> I think that we want to avoid introducing a general-purpose caching mechanism in package lfshttp.

I agree that this change to `lfshttp` doesn't look right. The reason why I have introduced "caching" here is that's it's the only place where I have access to the raw server response (unless changing `DecodeJSON`'s `res *http.Response` parameter to something else what wouldn't be any better). But I guess that it's not that important to cache the locks data exactly as it has been received, but it's also fine to parse it first and afterwards convert back to JSON and write to the cache file. Do you agree?

> Building on (1), I think that there could be more coordination between your implementation in lfshttp and the implementation in locking/locks.go.

Then I'll move the caching logic to `locks.go` and leave `api.go` and `lfshttp.go` untouched. This also seems to be the correct place, because only in `locks.go` we understand whether a query should be cached or not: e.g. in case of a `limit` configured, we should probably exclude this from caching, because we would overwrite a full result by a limited one.

> To expand upon (2), the flow as I see it now is to create a location on disk where the cached result might go, try and open that, and, should that fail, call into lfshttp with that caching location, expecting that lfshttp will write to it.

That would have been the next step, but this will become obsolete when moving the entire logic to `locks.go`.

> I think that what we're really after here (and what you suggest in #3252) is that we denote the difference between our own locks and the total set of known locks.

Not sure about "difference", but distinguish between our own locks and locks from other users.

> Would it be possible to do this already by comparing the lock's locker with the current user?

I don't think so, because the current user which the Git-LFS client knows in general can't be matched against `owner.name` information received from the server. What we can do is to compare lock `id`s. But here the drawback is that "own" locks will only be identified correctly in the clone where `git lfs lock` was executed. On the other hand, the server knows locks by user, not clones, and allows pushing a locked file from a different clone (for which the Git-LFS client thinks it does not have a lock). Hence, I think information from the `lockcache.db` is more confusing than helpful. Instead, it should be possible to ask the server about user-related locks (#3252).

ttaylorr:
> But I guess that it's not that important to cache the locks data exactly as it has been received, but it's also fine to parse it first and afterwards convert back to JSON and write to the cache file. Do you agree?

I agree. We assume that a serialization round-trip is meaning-preserving, so I think that we're fine to drop the code entirely and rely on our serialized cache instead.

> > Would it be possible to do this already by comparing the lock's locker with the current user?
> 
> I don't think so, because the current user which the Git-LFS client knows in general can't be matched against `owner.name` information received from the server.

That makes sense locally, though I think that I'm confused about what you're going for in general here. Could you explain a little further what you hope this change will introduce, please?

mstrap:
>> But I guess that it's not that important to cache the locks data exactly as it has been received, but it's also fine to parse it first and afterwards convert back to JSON and write to the cache file. Do you agree?
>
> I agree. We assume that a serialization round-trip is meaning-preserving, so I think that we're fine to drop the code entirely and rely on our serialized cache instead.

OK!

>>> Would it be possible to do this already by comparing the lock's locker with the current user?
>>
>> I don't think so, because the current user which the Git-LFS client knows in general can't be matched against owner.name information received from the server.
>
> That makes sense locally, though I think that I'm confused about what you're going for in general here. Could you explain a little further what you hope this change will introduce, please?

Maybe we are talking past each other :) My understanding was that you were suggesting that #3252 may be solved on the client side by comparing the username used for authentication against `owner.name` received from the server. At least for GitLab, this doesn't seem to be work -- I have added an example to #3252. This is what I meant by "in general can't be matched". Or was your suggestion "comparing the lock's locker with the current user" referring to the server side?

ttaylorr:
@mstrap Ah! I think I understand where you are going with this. Is the idea that when we ask for `--ours` (or `--verified`, etc) that we issue an API call to `/locks/verify` and cache the result thusly? 

Sorry for all of the back-and-forth, just want to make sure that we're on the same page.

ttaylorr:
> That's exactly what I have in mind.

cc @git-lfs/core does this sound OK?

mstrap:
Thanks for the feedback. Just FYI, I'll be traveling for the next two weeks, then will resume working on this task.

mstrap:
@ttaylorr seems no one else is willing to review this? I should have more time soon to continue with #3252, however I'd prefer to do small steps as I'm quite new to the code.

bk2204:
So far, this looks fine to me as well.

PR #3268, opened 2018-09-18
CI improvements

bk2204:
These are a bunch of small improvements to make troubleshooting credential-related CI failures easier.  The only user-visible change is to pass standard error through when filling credentials to make sure that we get log output from the credential helper.  This should be a net positive altogether, since it's the same behavior that Git has.

In addition, there are changes to make getting debugging output from successful tests easier and to make the lfstest credential helper log all of its operations.  Each change should be self-explanatory.

If someone has a suggestion for a better name for the environment variable in the final commit, I'm all ears.

### DIFF
--- a/lfsapi/creds.go
+++ b/lfsapi/creds.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"net/url"
+	"os"
 	"os/exec"
 	"strings"
 	"sync"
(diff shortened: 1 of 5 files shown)


bk2204:
BTW, @ttaylorr, I introduced two new commits, so if you have time, you might want to just sanity check that they look all right as well.

ttaylorr:
> you might want to just sanity check that they look all right as well.

Those look good, and thanks for letting me know 😄

PR #3271, opened 2018-09-19
Support listing only filename tracked by git lfs using --name (-n) option

skashyap7:
Some users just want to have the list of LFS files. This is particularly useful when you want to programatically scan the working tree and would like to avoid additional parsing


skashyap7:
Hi Taylor,

I think --name-only is the way to go. I will make the changes and raise a MR.

PR #3277, opened 2018-09-21
Canonicalize paths properly on Windows

### LINKED ISSUE #3276: git-lfs lock fails on Windows if the process's working directory does not match file system case
When invoking `git lfs lock` or `git lfs unlock` programatically and providing the process's working directory with wrong file system case, both commands will fail with unexpected errors. For example, when invoking Git-LFS from following Java snippet:

	public static void main(String[] args) throws Exception {
		ProcessBuilder builder = new ProcessBuilder();
		Map<String, String> env = builder.environment();
		String gitBin = "d:/temp/git64/bin";
		env.put("Path", gitBin + ";" + env.get("Path"));

		String repoRoot = "D:/temp/gitlab.lfs.clone";

		builder.directory(new File(repoRoot));
		builder.command(Arrays.asList(gitBin + "/git.exe", "lfs", "lock", "huge"));
		builder.redirectOutput(ProcessBuilder.Redirect.INHERIT);
		builder.redirectError(ProcessBuilder.Redirect.INHERIT);

		final Process process = builder.start();
		process.waitFor();
	}

this works fine and outputs:

    Locked huge

However, when using repoRoot with wrong case (small 'd'):

    String repoRoot = "d:/temp/gitlab.lfs.clone";

It will fail with:

    Lock failed: CreateFile d:\temp\gitlab.lfs.clone\D:\temp\gitlab.lfs.clone\huge: The filename, directory name or volume label syntax is incorrect.

Note that it doesn't matter whether using forward-slashes or back-slashes.

**Expected behavior**

`lock` and `unlock` should work regardless of the working directory path case.

**Notes**

`git lfs locks` works fine.

`git lfs status` works, but will report absolute paths in case of wrong path case.

Tested with git-lfs/2.5.1 (GitHub; windows amd64; go 1.10.3; git e1628484).

bk2204:
On Windows, the filesystem is typically case insensitive.  We formed the
lock path by stripping off the repo portion from the absolute pathname
using the strings package.  However, if the drive letters of the
absolute pathname and the canonicalized pathname differed in case, then
we would not strip any values off, and instead just concatenate the two
paths together.  Use the filepath.Rel function that's built into Go to
help us make the path relative in a way that is appropriate for the
operating system.

Fixes #3276

### DIFF
--- a/commands/command_lock.go
+++ b/commands/command_lock.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/git-lfs/git-lfs/errors"
 	"github.com/git-lfs/git-lfs/git"
@@ -86,8 +85,13 @@ func lockPath(file string) (string, error) {
 	}
 
 	abs := filepath.Join(wd, file)
-	path := strings.TrimPrefix(abs, repo)
-	path = strings.TrimPrefix(path, string(os.PathSeparator))
+	path, err := filepath.Rel(repo, abs)
+	if err != nil {
+		return "", err
+	}
+
+	path = filepath.ToSlash(path)
+
 	if stat, err := os.Stat(abs); err != nil {
 		return "", err
 	} else {


bk2204:
We can call `filepath.ToSlash`.  I believe the issue is that `os.Getwd` doesn't canonicalize the path (it doesn't on macOS, either), but the call to Git definitely does.  Consequently, we end up with paths differing in case where they don't share a meaningful prefix (or any non-empty prefix at all, on Windows).

I'll expand the commit message to explain this better as well.

bk2204:
I went ahead and made that change and expanded the explanation in the commit message to give more context about what's going on.  Let me know if you think I can make further improvements.

ttaylorr:
> Let me know if you think I can make further improvements.

I think that this is great, thanks! Merge away.

roozbehid-ic:
@ttaylorr I am having this problem with every file I try to lock. Dont know if something is wrong iwth ym system configuration or not. Is it in 2.5.2? 
Because with 2.5.2 I am still having this issue!

Shilpah123:
Hello,
The git lfs lock command is throwing lock failed: exit status 1 error. Can anyone please help me? Thanks.

PR #3283, opened 2018-09-25
Makefile: build source tarballs for release

### LINKED ISSUE #2911: Github generated tarballs change randomly
As the title states, github generated tarballs can change at any point. This creates issues for packaging on systems like [OpenBSD](https://marc.info/?l=openbsd-ports&m=151973450514279&w=2) ports. Would it be possible to get official tarballs created / attached to releases?

bk2204:
Due to changes between gzip versions, the source tarballs produced by GitHub may change.  To assist distributors, provide stable generated tarballs as part of the release process.

Fixes #2911.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -241,7 +241,8 @@ RELEASE_TARGETS = \
 	bin/releases/git-lfs-freebsd-amd64-$(VERSION).tar.gz \
 	bin/releases/git-lfs-freebsd-386-$(VERSION).tar.gz \
 	bin/releases/git-lfs-windows-amd64-$(VERSION).zip \
-	bin/releases/git-lfs-windows-386-$(VERSION).zip
+	bin/releases/git-lfs-windows-386-$(VERSION).zip \
+	bin/releases/git-lfs-$(VERSION).tar.gz
 
 # RELEASE_INCLUDES are the names of additional files that are added to each
 # release artifact.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -275,6 +276,9 @@ bin/releases/git-lfs-%-$(VERSION).zip : $(RELEASE_INCLUDES) bin/git-lfs-%.exe
 	@mkdir -p bin/releases
 	zip -j -l $@ $^
 
+bin/releases/git-lfs-$(VERSION).tar.gz :
```
ttaylorr: Even though this is fairly self-describing, could you add a comment explaining what this recipe makes to be consistent with the surrounding recipes?

Makefile
```diff
@@ -275,6 +276,9 @@ bin/releases/git-lfs-%-$(VERSION).zip : $(RELEASE_INCLUDES) bin/git-lfs-%.exe
...
+	git archive -o $@ --prefix=git-lfs-$(patsubst v%,%,$(VERSION))/ --format tar.gz $(VERSION)
```
ttaylorr: This looks good to me, though I'm curious why we're changing the `--prefix` to remove the leading `v`. Building this on my machine, I get:

```ShellSession
$ make VERSION=v2.5.0 bin/releases/git-lfs-v2.5.0.tar.gz
git archive -o bin/releases/git-lfs-v2.5.0.tar.gz --prefix=git-lfs-2.5.0/ --format tar.gz v2.5.0
$ cd bin/releases
$ tar -xf git-lfs-v2.5.0.tar.gz
$ ls -l
total 1848
drwxr-xr-x  40 ttaylorr  staff    1280 Jul 26 12:48 git-lfs-2.5.0
-rw-r--r--   1 ttaylorr  staff  945633 Sep 25 13:48 git-lfs-v2.5.0.tar.gz
```

Whereas without the `--prefix=...` invocation, I would expect we'd get `git-lfs-v2.5.0` as the directory name.

Makefile
bk2204: Yeah, of course.

Makefile
bk2204: I should have explained this in the commit message (and I’ll update it accordingly).  This is the behavior that the tarballs on GitHub use, and in my experience most open-source projects don’t emit the `v` in the prefix.  Git, for example, doesn't use the `v`.

I think the rationale is that the `v` is an artifact of the version control system tags and not a “real” part of the version number.  If you think it would be better, I can simply remove the `v` from the generated tarball name as well, which would be consistent in the archive, but inconsistent with our other artifacts.

ttaylorr: Sounds good, and thanks for clarifying. I think that we can do the later in a separate change, since I think that will affect some of those who bundle Git LFS in their projects (in particular, @dscho) who should be notified.

bk2204:
I've updated the commit and commit message to reflect this discussion.

Issue #3281, opened 2018-09-24
git lfs migrate import (Error in git)


zezba9000:
I have broken git lfs objects and can't fix them. Running command below is broken.

$ git lfs migrate import
migrate: override changes in your working copy? [Y/n] Y
migrate: changes in your working copy will be overridden ...
migrate: Fetching remote refs: ..., done
Error in git rev-list --stdin --reverse --topo-order --do-walk --: exit status 128 fatal: bad revision '^refs/remotes/tags/origin/v1.0'

Windows 10 x64
Git 2.19.0.windows.1
Git LFS 2.5.1

ttaylorr:
Hi, Andrew -- thanks for writing in. In truth, I'm a little puzzled with
what to do here, and I'd love some feedback from @git-lfs/core on the
best way to go forward.

What's going on here is an complicated interaction between a few things:
(1) that we don't track the remote that a *git.Ref belongs to, (2) that
we hacked *git.Ref.Name to _include_ the remote name as a prefix, and
(3) that we then call `(*git.Ref) Refspec()` in order to get out a name
that we can then give back to 'git-rev-parse(1)'.

I think that a fix would look something like the following:

  - Teach getRemoteRefs() to return a map[string][]*git.Ref (i.e., a
    mapping from remote name to a slice of references belonging to that
    remote).

  - Teach formatRefName() to special case both types of remote
    references that Git LFS knows about (e.g., git.RefTypeRemoteBranch,
    and git.RefTypeRemoteTag).

  - Call formatRefName() on each reference in the return value from
    getRemoteRefs(), falling back to r.Refspec() when we're in a local
    reference.

  - Undo our hackery to prepend the remote name to the reference name.

But all of this said, here's my confusion: I can't find anywhere in the
modern Git codebase that creates a
refs/remotes/tags/<remotename>/<tagname> remote, _other_ than
'git-svn(1)', with --prefix prior to Git 2.0.0.

My understanding is that we should instead be treating remote "tag"
references as:

  refs/remotes/<remote>/tags/<tag>

Hence, we can remove _both_ git.RefTypeRemoteBranch and
git.RefTypeRemoteTag in favor of git.RefTypeRemote and call Refspec() on
that. Having removed the prefix "refs/remotes/", we'll be left with
"<remote>/tags/<tag>" which we can append to the aforementioned prefix
to get a round-trippable reference name.

Thoughts?

bk2204:
[...]

PR #3285, opened 2018-09-25
commands: unambiguously resolve remote references

ttaylorr:
This pull request fixes a long-standing bug wherein a reference beginning with `refs/remotes/tags` would incorrectly be round-tripped back to Git, causing `git lfs migrate` and others to complain of a reference that does not exist.

Closes: https://github.com/git-lfs/git-lfs/issues/3281.

##

/cc @git-lfs/core 
/cc @zezba9000, https://github.com/git-lfs/git-lfs/issues/3281

### DIFF
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -279,15 +279,11 @@ func getRemoteRefs(l *tasklog.Logger) (map[string][]*git.Ref, error) {
 // formatRefName returns the fully-qualified name for the given Git reference
 // "ref".
 func formatRefName(ref *git.Ref, remote string) string {
-	var name []string
-
-	switch ref.Type {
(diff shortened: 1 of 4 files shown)


zezba9000:
Will this fix be in git 2.9.1 do you think?

bk2204:
This will be in Git LFS 2.6.0, which we don't yet have a release date for, although it will likely be soonish.  This is a separate project from Git, although it is bundled in Git for Windows.

Git for Windows will likely pick it up automatically once we do the release.  I can't say how soon afterwards, though.

zezba9000:
@bk2204 Oops yes I meant the next Git-LFS release sorry. Ok will keep an eye out, tnx!

Issue #1044, opened 2016-02-25
Pure SSH-based protocol


gully:
I would like to push git lfs changes via SSH, but this doesn't seem to work "out of the box".  Specifically, I have two remotes: one with https (_origin_, i.e. GitHub) and one with ssh (_alternate_, i.e. a local computer not generally accessible to the web, only ssh).  When I set up git-lfs, the https remote is fine, but I can no longer push to the ssh branch:

``` bash
$ git push alternate master
Git LFS: (0 of 2 files) 0 B / 450.17 KB
exit status 127
error: failed to push some refs to "foo@000.000.000.000:home/blah.git"
```

``` bash
$ git lfs push alternate master
Git LFS: (0 of 2 files) 0 B / 450.17 KB
exit status 127
```

I suspect this Issue is related to Issues #295, and its Pull Request #350, and maybe #798 

Will it be possible to have git-lfs work with existing ssh remotes "out of the box", like https?  Is the currently supported solution to run `git-lfs-ssh-serve` on the remote?  I have not been able to experiment with the `git-lfs-ssh-serve` yet.  Forgive me if this is a naive or duplicate question, or unique to my setup.  Thanks!

technoweenie:
Git LFS doesn't support SSH as a transport protocol. So, as a <del>hack</del> workaround, it executes a `git-lfs-authenticate` command through SSH, which should return JSON. Unfortunately, this isn't really documented that well. Here's [the JSON struct that it needs to return](https://github.com/github/git-lfs/blob/81563f2c085c9ad7b3a98f270f57f6ec19844feb/lfs/ssh.go#L15-L17). Here's a live example:

```bash
$ ssh git@github.com git-lfs-authenticate github/git-lfs download
{
  "href": "https://api.github.com/lfs/github/git-lfs",
  "header": {
    "Authorization": "RemoteAuth monkey"
  },
  "expires_at": "2016-02-25T17:42:24Z"
}
```

This gives the URL and necessary header value to access the LFS API. Not only would you have to implement this `git-lfs-authenticate` script, but you'd have to setup some kind of LFS server somewhere to communicate with the client.

gebi:
@technoweenie we also needed that feature, thus we removed all user handling from git-lfs-testserver and added the necessary infrastructure to support authentication through JWT tokens from gitolite (in our case).

https://github.com/mgit-at/lfs-test-server/tree/master/git-lfs-authenticate

MrShadowGames:
[...]

PR #3290, opened 2018-09-27
RFC: SSH protocol

bk2204:
There has been a significant amount of interest in a purely SSH-based protocol for Git LFS.  Add a proposal for an SSH-based protocol based on Git's pkt-line and protocol v2, while still mapping cleanly onto HTTP for ease of implementation on the server-side.  Implementers will likely have all of this machinery (either now or in the future) available, so this should be fairly easy to implement.

This is merely a proposal, not a commitment to implement.  Implementers who prefer to use HTTP may continue to do so.  Not specified yet is a technique to discover the SSH-based protocol, although it will probably be overloaded onto `git-lfs-authenticate`.

Comments and suggestions welcome.

/cc @git-lfs/core
/cc #1044

### DIFF
--- a/docs/proposals/ssh_adapter.md
+++ b/docs/proposals/ssh_adapter.md
@@ -1,0 +1,328 @@
+# SSH protocol proposal
+
+We'd like to implement a protocol for Git LFS that uses SSH protocol
+exclusively, avoiding the need to use HTTPS altogether.  This will make
+deployment and use easier in a variety of situations where access to certain
+ports is limited.
+
+This is a merely a proposal, not a commitment to implement for either the client
+or server side.  Implementers who prefer to use HTTP can continue to do so.
+
+## What not to do
+
+There are several possible approaches that could be adopted.  SSH provides a
+native capability for the SFTP protocol, which can be used to transfer files.
+However, in order to implement this on the server side, each access (upload or
+download) must have an access control check instead of one at the beginning of
+the operation.  This would be inefficient in some server-side implementations,
+and nearly impossible to implement securely for implementations that use the
+system OpenSSH for implementation.
+
+## A more usable approach
+
+Git already has some places we can look for inspiration.  Its SSH protocol is
+based on the Git native protocol, which is based on the pkt-line scheme.
+Recently, Git has learned about protocol version 2, which provides better
+support for expressing and negotiating capabilities.
+
+Ideally, we would allow multiple operations to occur on a single connection for
+efficiency's sake, especially on high-latency connections, where the cost of SSH
+connection setup may be high due to multiple round trips.  In addition, a
+protocol which maps well onto HTTP may be beneficial for those server-side
+implementations which would like to proxy connections to an HTTP-based backend.
+
+## Preliminary design
+
+This design assumes a reference to Git's pkt-line and protocol v2 documentation.
+pkt-line headers for this document may contain values up to 65519 decimal.
+
+To initiate a connection, Git LFS should run the following command:
+
+    $ ssh [{user}@]{server} git-lfs-transfer {path} {operation}
+
+If authentication fails, or some other connection error occurs, errors will be
+read from standard error and displayed to the user.  The operation may be
+`upload`, `download`, or `lock`.  Other operations may be implemented in the
+future.
+
+Once the connection is established, the server should send a capability
+advertisement:
+
+```
+capability-advertisement = capability-list flush-pkt
+capability-list = *capability
+capability = PKT-LINE(key[=value] LF)
+
+key = 1*(ALPHA | DIGIT | "-_")
+value = 1*(ALPHA | DIGIT | "-_.,?\/{}[]()<>!@#$%^&*+=:;")
+```
+
+Unlike the Git protocol, but like IMAP, the protocol version is specified as a
+capability.  This document defines protocol version 1, which is specified as
+`version=1`.  If the server supports other protocol versions, it may enumerate
+them here as well.
+
+If the server supports locking, the `locking` capability should be advertised,
+and the client may then use the `lock`, `unlock`, and `list-lock` commands.
+
+No capabilities other than the base functionality specified here are enabled
+without the client explicitly enabling them.  Note that the `value` production
+here, unlike in Git, does not include the space character, since it is used as a
+delimiter in parts of the protocol.
+
+The client will then issue an appropriate version command:
+
+```
+version-request = PKT-LINE("version " number LF) flush-pkt
+number = 1*DIGIT
+```
+
+The response from the server will look like the following:
+
+```
+version-response = status-command
+                   delim-pkt
+                   error-message
+                   flush-pkt
+status-command = PKT-LINE("status " http-status-code LF)
+http-status-code = 3DIGIT
+error-message = *PKT-LINE(data LF)
+```
+
+The `http-status-code` portion of the response is an HTTP status code, identical
+to those used if the request is made over HTTP.
+
+The response code should be 200 if the version is accepted or 400 if it is not.
+Other values are possible if other errors occur.
+
+### Requests to transfer objects
+
+These commands may be used if the operation was `upload` or `download`.
+
+The `batch` command is used to specify a JSON command identical those used at
+the `info/lfs/object/batch` endpoint:
+
+```
+batch-request = batch-command
+                *argument
+                delim-pkt
+                *oid-line
+                flush-pkt
+batch-command = PKT-LINE("batch" LF)
+argument = PKT-LINE(key=[data] LF)
+oid-line = PKT-LINE(oid size *(key=[value]) LF)
+oid = 1*("a-f0-9")
+size = 1*DIGIT
+```
+
+The `transfer` argument is equivalent to the corresponding value in the HTTP
+JSON API.  The `refname` argument is equivalent to the `name` argument of the
+`ref` object in the HTTP JSON API.
+
+Unknown arguments should be ignored, as should unknown key-value pairs in the
+`oid-line` production.
+
+The response from the server will look like the following:
+
+```
+batch-response = status-command
+                 *argument
+                 delim-pkt
+                 (*batch-oid-line | error-message)
+                 flush-pkt
+batch-oid-line = PKT-LINE(oid size action *(key=[value]) LF)
+error-message = PKT-LINE("message " data LF)
+```
+
+If the status command is successful (that is, the status is not 200-series
+response), the data provided matches the `*batch-oid-line` production;
+otherwise, the data provided represents a user-visible error message.
+
+The server response should contain one pkt-line per oid-size-action tuple.  That
+is, the same oid and size may be repeated if there are multiple actions.  If the
+server has no actions that are valid for an object, it should be listed once in
+the response with the `noop` action.
+
+The response for an oid may include a string, `id`, which is an opaque
+identifier relevant only to the server to help it identify the object, and
+another string, `token`, which is an opaque identifier relevant only to the
+server to help it manage authentication.  These strings must meet the syntax for
+the `value` production above; if arbitrary bytes are needed, Base64 encoding is
+recommended.
+
+The `expires-in` and `expires-at` key-value pairs have the same meaning as their
+corresponding items from the HTTP JSON API.
+
+These values, if specified, must be passed as arguments to the `get-object` and
+`put-object` commands.
+
+### Downloads
+
+These commands may be used if the operation was `download`.
+
+If the operation was `download`, the command `get-object` may be used to retrieve an object:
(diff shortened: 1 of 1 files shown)


docs/proposals/ssh_adapter.md
```diff
@@ -0,0 +1,194 @@
+# SSH protocol proposal
+
+We'd like to implement a protocol for Git LFS that uses SSH protocol
```
ttaylorr: I think that it was good that you mentioned in your commit message that this proposal is an aspiration, and not a promise that we will implement it, or a requirement of the same to anyone else. Do you think that it would be worthwhile to mention that here, too?

docs/proposals/ssh_adapter.md
```diff
@@ -0,0 +1,194 @@
...
+However, in order to implement this on the server side, each access (upload or
+download) must have an access control check instead of one at the beginning of
+the operation.  This would be inefficient in some server-side implementations,
+and nearly impossible to implement securely for implementations that use the
+system OpenSSH for implementation.
+
+## A more usable approach
+
+Git already has some places we can look for inspiration.  Its SSH protocol is
+based on the Git native protocol, which is based on the pkt-line scheme.
```
ttaylorr: I don't think that this is related to your proposal at all, but we do have a good implementation of pkt-line in package `git`, thanks to @larsxschneider's work during the Git 2.11-era push towards `filter.lfs.process`.

I guess all of that is to say: this makes implementing a pkt-line-based protocol less work for us, since we don't have to first implementation serialization over that protocol.

But, one thing we definitely _don't_ have is an implementation that will provide us convenient data structures over protocol v2 concepts, so I think that the work here lays there. Perhaps pkt-line and this proposed new implementation should reside in a separate repository?

docs/proposals/ssh_adapter.md
```diff
@@ -0,0 +1,194 @@
...
+connection setup may be high due to multiple round trips.  In addition, a
+protocol which maps well onto HTTP may be beneficial for those server-side
+implementations which would like to proxy connections to an HTTP-based backend.
+
+This design assumes a reference to Git's pkt-line and protocol v2 documentation.
+pkt-line headers for this document may contain values up to 65519 decimal.
+
+To initiate a connection, Git LFS should run the folllowing command:
+
+    $ ssh [{user}@]{server} git-lfs-transfer {path} {operation}
```
ttaylorr: Good, this matches with my expectation of what we'd want, since the program that we execute today over SSH is called `git-lfs-authenticate`. I'm glad that we're keeping the `git-lfs-` prefix, and the same order of arguments.

docs/proposals/ssh_adapter.md
```diff
@@ -0,0 +1,194 @@
...
+
+If authentication fails, or some other connection error occurs, errors will be
+read fronm standard error and displayed to the user.  The operation may be
```
ttaylorr: Typo: `fronm`.

docs/proposals/ssh_adapter.md
```diff
@@ -0,0 +1,194 @@
...
+`upload`, `download`, or `meta` (for items which are not transfers).
```
ttaylorr: Can you explain what `meta` is?

docs/proposals/ssh_adapter.md
```diff
@@ -0,0 +1,194 @@
...
+them here as well.
+
+The client will then issue an appropriate version command:
+
+```
+version-request = PKT-LINE("version " number LF) flush-pkt
+number = 1*DIGIT
+```
+
+No reply is expected.
```
ttaylorr: I think that further explanation here would be good. What if the server refuses to serve a protocol request? I assume that it would terminate the connection, but perhaps there is a convention for agreeing upon a greater-common-denominator-esque protocol(s)?

docs/proposals/ssh_adapter.md
```diff
@@ -0,0 +1,194 @@
...
+These commands may be used if the operation was `upload` or `download`.
+
+The `batch` command is used to specify a JSON command identical those used at
+the `info/lfs/object/batch` endpoint:
+
+```
+batch-request = batch-command
+                *arguments
+                delim-pkt
+                json-message
```
ttaylorr: I wonder what the benefit of `json-message` is in this context. Perhaps that this makes it easier on the client/server to transition from the HTTP(S)-based Batch API to the wire protocol, but I think that we're just deferring the cost until later.

What do you think of instead defining the protocol (as in `docs/api/batch.md`) here as a wire format? Perhaps something like:

```
batch
transfer=basic
refname=refs/heads/master
(OID size)*
```

I think that these are the important components of a batch request (where `(OID size)*` refers to 0 or more OID-size tuples pertaining to the objects that we wish to upload/download).

There are two other points that I should raise here:

1. Without `json-message`, I think we need to support `transfers` and `ref (name)` as are described in the "Requests" section of the Batch API documentation.
2. I think that we can assume that the `operation` field can safely be inferred from the context under which we invoked the program.

docs/proposals/ssh_adapter.md
```diff
@@ -0,0 +1,194 @@
...
+                flush-pkt
+batch-command = PKT-LINE("batch" LF)
+arguments = PKT-LINE(key=[value] LF)
+json-message = 1*PKT-LINE(json-data)
+```
+
+It is recommended to specify a `size` argument, which is the number of bytes in
```
ttaylorr: I think that this is good. What do you think about making this a requirement so that implementers don't have to decipher between the two?

docs/proposals/ssh_adapter.md
```diff
@@ -0,0 +1,194 @@
...
+decimal of the JSON payload without pkt-line encoding.  A server may signal an
+error if it requires a `size` argument for an operation but one is not specified.
+
+Unknown arguments should be ignored.
```
ttaylorr: I think that it would be OK to terminate the connection here, depending on your thoughts, too. Unknown arguments mean (to me) undefined behavior, which is not something that we should be subjecting the user to.

docs/proposals/ssh_adapter.md
```diff
@@ -0,0 +1,194 @@
...
+
+The response from the server will look like the following:
+
+```
+status-json-response = status-command
+                       *arguments
+                       delim-pkt
+                       json-message
```
ttaylorr: Same comment applies here about `json-message` vs spelling the protocol out in detail over the wire.

docs/proposals/ssh_adapter.md
```diff
@@ -0,0 +1,194 @@
...
+                       flush-pkt
+status-command = PKT-LINE("status " http-status-code LF)
+http-status-code = 3DIGIT
```
ttaylorr: Nice, I really like this. I think that this is an appropriate place for us to make a transition between the wire protocol and the old-style HTTP protocol. I think that this also jives with @PastelMobileSuit's current work, and makes our `lfsapi` implementation last longer.

docs/proposals/ssh_adapter.md
```diff
@@ -0,0 +1,194 @@
...
+manage authentication.  These strings must meet the syntax for the `value`
+production above; if arbitrary bytes are needed, Base64 encoding is recommended.
+
+These values, if specified, must be passed as arguments to the `get-object` and
+`put-object` commands.
+
+The `http-status-code` portion of the response is an HTTP status code, identical
+to those used if the request is made over HTTP.
+
+### Downloads
```
ttaylorr: Ah, ignore my previous comments made about unifying the batch protocol into its individual components over the wire. If my understanding is correct, that is exactly what you aim to do here. In this case, what do you think about getting rid of the batch protocol entirely for the wire protocol?

I think that there is certainly an appealing quality about introducing something (even though we know that it will be deprecated eventually) that is designed to ease the burden of implementers. (At this point, I'm requesting :eyes: from @git-lfs/implementers.)

But, I'm worried that this is a deferred cost with high interest that allows us to put off the harder change, and makes the hard transition even harder. What do you think?

docs/proposals/ssh_adapter.md
bk2204: Sure, I can do that.

docs/proposals/ssh_adapter.md
bk2204: It's possible.  I don't know how many people are going to find pkt-line work in Go super useful; I expect all of the people who are going to be doing that work are implementers of Git-related things, either client or server side.  It may be useful to start work and then split things out if we find a logical unit that works on its own.

docs/proposals/ssh_adapter.md
bk2204: Thanks.

docs/proposals/ssh_adapter.md
bk2204: Yeah, I had intended to use it for locking and other cases where we're performing operations on the repo that are not strictly related to uploads or downloads, but that's probably not a good name.  If anyone has a better idea, I'm all ears.

docs/proposals/ssh_adapter.md
bk2204: Sure.  I was thinking since we had the pkt-line encoding of the data that would be like chunked transfer encoding, and the size could be omitted in that case, but I agree that in pretty much every case, we're going to know the size (or, like with JSON, can trivially compute it), so it doesn't make sense to omit it.

docs/proposals/ssh_adapter.md
bk2204: Yeah, I didn't want to try to define what error codes might occur in this situation.  The HTTP RFC folks are much more experienced than me in defining potential failure cases.  We can always expose a human-readable error in the JSON blob or other data.

docs/proposals/ssh_adapter.md
bk2204: Yeah, I thought about that.  I assumed that the client wouldn't specify a version which wasn't implemented, but we will have to deal with people who intentionally choose to provide invalid data, so we might as well specify something.

docs/proposals/ssh_adapter.md
ttaylorr: Hm, perhaps we could get rid of `meta` entirely, and specify `locking` instead, allowing any further extensions to be negotiated during the initial connection?

bk2204: I agree we should have an operation for locking instead.  It may be useful to have a way to discover capabilities without having to commit to an operation, but I don't feel strongly about it either way.

PastelMobileSuit:
:sparkle: This looks good to me! I don't have any feedback outside of what @ttaylorr has already raised

docs/proposals/ssh_adapter.md
bk2204: I can certainly specify a pkt-line-based format for this.  I definitely think that we need a batch endpoint, since we need to determine whether the server already has objects (since we would want to avoid sending them again wastefully).

docs/proposals/ssh_adapter.md
bk2204: I think allowing additional arguments that are unknown allows us to do things like adding support for `refname` without needing to bump the protocol.  If the behavior is not backwards compatible, then we'd need a new protocol version or a new capability.

ttaylorr: Ah, that makes sense. I think this is OK as-is, then.

docs/proposals/ssh_adapter.md
bk2204: As I mentioned above, I think the batch protocol is useful to determine whether we need to upload objects at all.  Since we'd like to avoid uploading unneeded objects, we should maintain the batch protocol or specify that the client has to verify objects before uploading them.  The latter potentially means a large number of round trips, which is bad for high-latency connections.  Maybe that's not very much of a concern for Git LFS, which kind of assumes that a stable, reasonably performant connection is available, but I can imagine people traveling on airplanes being unhappy about that.

I'm fine simply specifying that one should use `verify-object` if you think that's best.  If so, then I'll simply drop the `id` and `token` arguments as well.

ttaylorr: I think leaving it as-is is fine, though I'll defer to your judgement in the final case.

Rabadash8820:
@bk2204 @ttaylorr This seems like a pretty huge deal!  Just to confirm, does Git LFS officially work over SSH now?  If not, do you have a timeline for what version of Git will include these changes?  And even more importantly, do you know when code hosting providers (particularly GitHub, BitBucket, GitLab, and Azure DevOps) will support these changes?

Thank you both again for your work on this much-needed feature!

DanVicarel:
Okay, I didn't look through the Files Changed, I see now that this is just specification stuff.   I saw "SSH" and "merged" before and got really excited 😛 Good luck implementing!

PR #3293, opened 2018-09-28
lfsapi/tq: Have DoWithAuth() caller determine URL Access Mode

PastelMobileSuit:
This PR moves the responsibility to determine the URL Access Mode out of `DoWithAuth()` and into the hands of the caller. This removes our built-in assumption that all requests will be made against the LFS API endpoint, and allows other endpoints to be used for the access mode.

The transfer adapters generally don't make requests against the LFS API endpoint, generally making requests against a storage endpoint instead. This change means that requests coming from the transfer queue will now use that storage endpoint to determine the URL Access Mode instead of the LFS API endpoint.

This PR ended up being more complicated than expected, so there's a fair number of commits here. I've tried to commit atomic changes with descriptive commit messages, so I think it makes sense to review this one on a commit-by-commit basis.

/cc @git-lfs/core

### DIFF
--- a/lfsapi/auth.go
+++ b/lfsapi/auth.go
@@ -24,10 +24,10 @@ var (
 // DoWithAuth sends an HTTP request to get an HTTP response. It attempts to add
 // authentication from netrc or git's credential helpers if necessary,
 // supporting basic and ntlm authentication.
-func (c *Client) DoWithAuth(remote string, access Access, req *http.Request, allowRetry bool) (*http.Response, error) {
+func (c *Client) DoWithAuth(remote string, access Access, req *http.Request) (*http.Response, error) {
 	res, err := c.doWithAuth(remote, access, req, nil)
 
-	if allowRetry && errors.IsAuthError(err) {
+	if errors.IsAuthError(err) {
 		if len(req.Header.Get("Authorization")) == 0 {
 			// This case represents a rejected request that
 			// should have been authenticated but wasn't. Do
@@ -35,11 +35,18 @@ func (c *Client) DoWithAuth(remote string, access Access, req *http.Request, all
 			// maximum.
 			newAccess := c.Endpoints.AccessFor(access.url)
 			tracerx.Printf("api: http response indicates %q authentication. Resubmitting...", newAccess.Mode())
-			return c.DoWithAuth(remote, newAccess, req, allowRetry)
+			return c.DoWithAuth(remote, newAccess, req)
 		}
 	}
 
 	return res, err
+}
+
+// DoWithAuthNoRetry sends an HTTP request to get an HTTP response. It works in
+// the same way as DoWithAuth, but will not retry the request if it fails with
+// an authorization error.
+func (c *Client) DoWithAuthNoRetry(remote string, access Access, req *http.Request) (*http.Response, error) {
+	return c.doWithAuth(remote, access, req, nil)
 }
 
 // DoAPIRequestWithAuth sends an HTTP request to get an HTTP response similarly
@@ -49,7 +56,7 @@ func (c *Client) DoAPIRequestWithAuth(remote string, req *http.Request) (*http.R
 	operation := getReqOperation(req)
 	apiEndpoint := c.Endpoints.Endpoint(operation, remote)
 	access := c.Endpoints.AccessFor(apiEndpoint.Url)
-	return c.DoWithAuth(remote, access, req, true)
+	return c.DoWithAuth(remote, access, req)
 }
 
 func (c *Client) doWithAuth(remote string, access Access, req *http.Request, via []*http.Request) (*http.Response, error) {
--- a/lfsapi/auth.go
+++ b/lfsapi/auth.go
@@ -34,7 +34,7 @@ func (c *Client) DoWithAuth(remote string, access Access, req *http.Request, all
 			// not count this against our redirection
 			// maximum.
 			newAccess := c.Endpoints.AccessFor(access.url)
-			tracerx.Printf("api: http response indicates %q authentication. Resubmitting...", newAccess.GetMode())
+			tracerx.Printf("api: http response indicates %q authentication. Resubmitting...", newAccess.Mode())
 			return c.DoWithAuth(remote, newAccess, req, allowRetry)
 		}
 	}
@@ -64,7 +64,7 @@ func (c *Client) doWithAuth(remote string, access Access, req *http.Request, via
 	if err != nil {
 		if errors.IsAuthError(err) {
 			newAccess := access.Upgrade(getAuthAccess(res))
-			if newAccess.GetMode() != access.GetMode() {
+			if newAccess.Mode() != access.Mode() {
 				c.Endpoints.SetAccess(newAccess)
 			}
 
@@ -83,7 +83,7 @@ func (c *Client) doWithAuth(remote string, access Access, req *http.Request, via
 }
 
 func (c *Client) doWithCreds(req *http.Request, credHelper creds.CredentialHelper, creds creds.Creds, credsURL *url.URL, access Access, via []*http.Request) (*http.Response, error) {
-	if access.GetMode() == NTLMAccess {
+	if access.Mode() == NTLMAccess {
 		return c.doWithNTLM(req, credHelper, creds, credsURL)
 	}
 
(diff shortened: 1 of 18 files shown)


lfsapi/auth.go
```diff
@@ -24,39 +24,53 @@ var (
 // DoWithAuth sends an HTTP request to get an HTTP response. It attempts to add
 // authentication from netrc or git's credential helpers if necessary,
 // supporting basic and ntlm authentication.
-func (c *Client) DoWithAuth(remote string, req *http.Request) (*http.Response, error) {
-	return c.doWithAuth(remote, req, nil)
+func (c *Client) DoWithAuth(remote string, access Access, req *http.Request, allowRetry bool) (*http.Response, error) {
+	res, err := c.doWithAuth(remote, access, req, nil)
+
+	if allowRetry && errors.IsAuthError(err) {
+		if len(req.Header.Get("Authorization")) == 0 {
```
PastelMobileSuit: This is a change to the retry conditional that previously lived in doWithAuth(), but it seemed safe to me. If a request fails with credentials then the Authorization header will be deleted before doWithAuth() returns, so this should cover that case, and so should only fire if the request had no authorization or if it used a credential helper to fill out the Authorization header.

lfsapi/endpoint_finder.go
```diff
@@ -26,14 +26,28 @@ const (
 	defaultRemote              = "origin"
 )
 
+type Access struct {
+	mode AccessMode
+	url  string
+}
+
+// Returns a copy of an AccessMode with the mode upgraded to newMode
+func (a *Access) Upgrade(newMode AccessMode) Access {
```
ttaylorr: Hmm. When we talked about this approach, I thought that there was more work to do in order to upgrade from one access mode to another.

I don't think it's bad at all thought that this function is short. In fact, I'd rather have a short function that introduces no mutable behavior than a short one that does (and I'd prefer that before a _longer_ function that introduces mutable behavior).

lfsapi/endpoint_finder.go
```diff
@@ -26,14 +26,28 @@ const (
...
+	return Access{url: a.url, mode: newMode}
+}
+
+func (a *Access) GetMode() AccessMode {
```
ttaylorr: Nit: "accessors" in Go do not ordinarily contain the prefix `Get()`. I think that you did the right thing in making `mode` non-exported, since that enforces immutability. But, let's call this `Mode()` instead, to make it appear closer to accessing a normal, publicly exported field.

lfsapi/endpoint_finder.go
```diff
@@ -202,39 +216,38 @@ func (e *endpointGitFinder) NewEndpoint(rawurl string) lfshttp.Endpoint {
...
-	tracerx.Printf("setting repository access to %s", access)
+func (e *endpointGitFinder) SetAccess(access Access) {
+	key := fmt.Sprintf("lfs.%s.access", access.url)
+	tracerx.Printf("setting repository access to %s", access.mode)
 
 	e.accessMu.Lock()
 	defer e.accessMu.Unlock()
 
-	switch access {
+	switch access.mode {
```
ttaylorr: What do you think about calling the `GetMode()`/`Mode()` function here instead? I think that that's a good way to go since it will minimize the diff if we ever need to do extra computation to determine the access mode.

This same comment applies throughout the rest of this file, and its test.

lfsapi/auth.go
```diff
@@ -24,28 +24,38 @@ var (
...
 
-func (c *Client) doWithAuth(remote string, req *http.Request, via []*http.Request) (*http.Response, error) {
+// DoAPIRequestWithAuth sends an HTTP request to get an HTTP response similarly
+// to DoWithAuth, but using the LFS API endpoint for the provided remote and
+// operation to determine the access mode.
+func (c *Client) DoAPIRequestWithAuth(remote string, req *http.Request) (*http.Response, error) {
+	operation := getReqOperation(req)
+	apiEndpoint := c.Endpoints.Endpoint(operation, remote)
+	access := c.Endpoints.AccessFor(apiEndpoint.Url)
+	return c.doWithAuth(remote, access, req, nil)
```
ttaylorr: I had a comment here that I wrote about halfway during the review that recommended that you call `DoWithAuth()` here to minimize any diff in the future should we complicate the body of `DoWithAuth()` at all later on.

But, you made that change in the subsequent commit, so I removed that comment here :-). Instead, I would recommend that in the future you make that change here (i.e., introduce this new function with a call to `DoWithAuth()` in this commit) rather than introduce something and then immediately change it in a future commit.

lfsapi/auth.go
```diff
@@ -24,8 +24,22 @@ var (
 // DoWithAuth sends an HTTP request to get an HTTP response. It attempts to add
 // authentication from netrc or git's credential helpers if necessary,
 // supporting basic and ntlm authentication.
-func (c *Client) DoWithAuth(remote string, access Access, req *http.Request) (*http.Response, error) {
-	return c.doWithAuth(remote, access, req, nil)
+func (c *Client) DoWithAuth(remote string, access Access, req *http.Request, allowRetry bool) (*http.Response, error) {
```
ttaylorr: Nice, I think that your explanation as to why we need this new behavior variant makes sense to me. What do you think about introducing another function which encodes this behavior, rather than adding a boolean argument to this one?

This has a few of benefits in my mind: (1) is that it limits the number of parameters that we have to think about when calling a function, (2) it makes clear at the call-site what the parameter is doing (i.e., rather than seeing `..., true` in a function call, we get a name that describes the behavior), and (3) it makes explicit what's going on in (2) which is really that we have two different versions of this function.

tq/basic_upload.go
```diff
@@ -215,3 +215,21 @@ func configureBasicUploadAdapter(m *Manifest) {
...
+
+func (a *basicUploadAdapter) makeRequest(t *Transfer, req *http.Request) (*http.Response, error) {
+	res, err := a.doHTTP(t, req)
+	if errors.IsAuthError(err) && len(req.Header.Get("Authorization")) == 0 {
+		// Construct a new body with just the raw file and no callbacks. Since
+		// all progress tracking happens when the net.http code copies our
+		// request body into a new request, we can safely make this request
+		// outside of the flow of the transfer adapter, and if it fails, the
+		// transfer progress will be rewound at the top level
+		f, _ := os.OpenFile(t.Path, os.O_RDONLY, 0644)
```
ttaylorr: I think that we already have the file open by the time that we call this function. Can you call `io.Seeker.Seek(0, io.SeekStart) on the _file_ instead of on the `io.Reader` in the request?

lfsapi/auth.go
PastelMobileSuit: I like that idea! I actually considered it but went with the boolean because all the names I came up with at the time felt pretty unwieldy. I actually quite like the name `DoWithAuthNoRetry()` though, so I'll add that.

tq/basic_upload.go
PastelMobileSuit: Unfortunately the issue in this case is that the underlying file has already been closed, so we can't issue a new request without re-opening it.

PastelMobileSuit:
@ttaylorr Thanks for the review! I've gone ahead and addressed your first round of feedback

bk2204:
Sure.  Let me spend a few minutes looking.

t/t-credentials.sh
```diff
@@ -350,6 +350,9 @@ begin_test "credentials from lfs.url"
   # A 401 indicates URL access mode for the /storage endpoint
   # was used instead of for the lfsapi endpoint
   grep "HTTP: 401" push.log
+  # Ensure we didn't make a second batch request, which means the request
+  # was successfully retried internally
+  [ ! $(grep "tq: retrying object" push.log) ]
```
bk2204: If for some reason this test fails, the error here will be rather unhelpful, since we'll expand the command substitution into multiple words, even potentially inserting newlines.  We could fix this if we wrote this with quotes around the command substitution, like so:

```sh
  [ ! "$(grep "tq: retrying object" push.log)" ]
```

The shell will do the right thing with the nested quote marks.  There are also a few places below that do the same thing.

PastelMobileSuit:
@bk2204 Thanks for the review! I've gone ahead and updated those last two credentials tests

PastelMobileSuit:
I'm currently in the process of getting that issue put together and have built a set of preview binaries! As such I'm going to merge this as soon as CI passes, and start work on the fix for #3130

PR #3296, opened 2018-10-01
Checkout options for conflicts

### LINKED ISSUE #3258: Proposal: teach 'git lfs checkout' --{theirs,ours,base}
From some discussion upstream in https://public-inbox.org/git/20180916075604.GB18517@gmail.com/, it was suggested that:

> Combining changes is inherently file-format specific, and I suspect
> that native authoring tools are best used in those scenarios.
> Maybe LFS can help deal with binary conflicts by having short and sweet
> ways to grab the "base", "their" and "our" versions of the conflict
> files.
> 
> Example:
> 
> 	git lfs checkout --theirs --to theirs.wav conflict.wav
> 	git lfs checkout --ours --to ours.wav conflict.wav
> 	git lfs checkout --base --to base.wav conflict.wav
> 
> Then the user can use {ours,theirs,base}.wav to produce the
> resolved result using their usual authoring tools.

##

/cc @git-lfs/core

bk2204:
When there's a conflict, it can be difficult to access the various stages of the conflict to view in a suitable diff tool.  Add the `--base`, `--ours`, and `--theirs` options to allow checking out the various stages of a conflict into different paths for easier comparison with various tools.

This fixes #3258.

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -13,8 +13,27 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	checkoutTo     string
+	checkoutBase   bool
+	checkoutOurs   bool
+	checkoutTheirs bool
+)
+
 func checkoutCommand(cmd *cobra.Command, args []string) {
 	requireInRepo()
+
+	stage, err := whichCheckout()
+	if err != nil {
+		Exit("Error parsing args: %v", err)
+	}
+
+	if checkoutTo != "" && stage != git.IndexStageDefault {
+		checkoutConflict(rootedPaths(args)[0], stage)
+		return
+	} else if checkoutTo != "" || stage != git.IndexStageDefault {
+		Exit("--to and exactly one of --theirs, --ours, and --base must be used together")
+	}
 
 	msg := []string{
 		"WARNING: 'git lfs checkout' is deprecated and will be removed in v3.0.0.",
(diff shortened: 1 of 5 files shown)


commands/command_checkout.go
```diff
@@ -75,6 +94,61 @@ func checkoutCommand(cmd *cobra.Command, args []string) {
...
+		Exit("Could not find decoder pointer for object %q: %v", ref.Sha, err)
+	}
+
+	p := &lfs.WrappedPointer{Name: file, Pointer: ptr}
+
+	singleCheckout.RunToPath(p, checkoutTo)
+	singleCheckout.Close()
+}
+
+func whichCheckout() (stage int, err error) {
```
ttaylorr: What do you think about returning a named version of `int`, perhaps like:

```go
type CheckoutStage int

const (
        CheckoutStageUnknown CheckoutStage = iota
        CheckoutStageBase
        CheckoutStageOurs
        CheckoutStageTheirs
)
```

Potentially this could type definition could also go in package `git`. It looks like there is a facility from `cobra` that provides us a way to do this sort of "assign a value to this location in memory based not on its type" via [`func Var()`](https://godoc.org/github.com/spf13/pflag#Var). I think we'd have to definite an implementation of `type Value interface`, but that seems doable.

commands/pull.go
```diff
@@ -81,22 +82,32 @@ func (c *singleCheckout) Run(p *lfs.WrappedPointer) {
...
+	// errors are only returned when the gitIndexer is starting a new cmd
+	if err := c.gitIndexer.Add(cwdfilepath); err != nil {
+		Panic(err, "Could not update the index")
+	}
+}
+
+// RunToPath checks out the pointer specified by p to the given path and returns
+// a boolean indicating whether it was successful.  It does not perform any sort
+// of sanity checking or add the path to the index.
+func (c *singleCheckout) RunToPath(p *lfs.WrappedPointer, path string) bool {
```
ttaylorr: I wonder if as a preparatory patch (or series) you might want to change the return type from `bool` to `error`. It may not be wise, since I believe that the implementation below already handles the error (i.e., sends the appropriate `pkt-line`, or similar), but I'm not sure.

It might be useful for callers to have the information, anyway.

docs/man/git-lfs-checkout.1.ronn
```diff
@@ -4,14 +4,16 @@ git-lfs-checkout(1) -- Update working copy with file content if available
 ## SYNOPSIS
 
 `git lfs checkout` <filespec>...
+`git lfs checkout` --to <path> { --ours | --theirs | --base } <file>...
 
 ## DESCRIPTION
 
-This command is deprecated, and should be replaced with `git checkout`.
+This command is deprecated when used without `--to`, and should be replaced with
```
ttaylorr: I wonder if it makes sense to continue to mark this command as deprecated in this case? I'm also not immediately sure what the behavior of `git checkout` is when given a `--{ours,theirs,base}` flag, and whether or not upstream runs the process filter.

commands/pull.go
bk2204: This commit is new, so I can make it `error` instead of `bool`.  The only reason I had to make it not return nothing is because we need to return early in `Run` in certain cases.

docs/man/git-lfs-checkout.1.ronn
bk2204: Yeah, I wasn't really sure how to go about that deprecation notice.  Removing it is fine with me.

commands/command_checkout.go
bk2204: Actually, I don't think `Var` does what we want.  I remember looking at it, and now that I'm looking again, it seems to require an argument, which we aren't passing in this case.  What we really need is a flag type that sets a variable to have a specific value, which I don't think `cobra` has.

PR #3299, opened 2018-10-03
Clean up trailing whitespace

bk2204:
Trailing whitespace is generally considered undesirable, so clean up the existing uses of it and check for it during CI to help avoid reintroducing it.  We avoid modifying vendored code and Git objects, which are binary.

### DIFF
--- a/script/cibuild
+++ b/script/cibuild
@@ -29,3 +29,7 @@ pushd t >/dev/null
   VERBOSE_LOGS=1 make X="$X" clean
   VERBOSE_LOGS=1 make X="$X" PROVE="$PROVE" PROVE_EXTRA_ARGS="$PROVE_EXTRA_ARGS"
 popd >/dev/null
+
+echo "Looking for trailing whitespace..."
+! git grep -lE '[[:space:]]+$' | \
+  grep -vE '(^vendor/|\.git/(objects/|index)|\.bat$)'
--- a/docker/README.md
+++ b/docker/README.md
@@ -4,11 +4,11 @@
 1. Run the dockers
 
         ./docker/run_dockers.bsh
-        
+
 2. **Enjoy** all your new package files in
 
         ./repos/
-        
+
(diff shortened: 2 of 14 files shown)


bk2204:
Just as a note here, I'm going to need to sanity check all the CI systems to ensure that they support `git grep -P`, and if not, fix the commit to use `git grep -E`.

script/cibuild
```diff
@@ -29,3 +29,6 @@ pushd t >/dev/null
   VERBOSE_LOGS=1 make X="$X" clean
   VERBOSE_LOGS=1 make X="$X" PROVE="$PROVE" PROVE_EXTRA_ARGS="$PROVE_EXTRA_ARGS"
 popd >/dev/null
+
+echo "Looking for trailing whitespace..."
+! git grep -lP '\s+$' | grep -vE '(^vendor/|\.git/objects/)'
```
ttaylorr: Should we use `git ls-files` here instead? I suppose that this is introducing the assumption that we're inside of a Git repository, but we make that assumption in a number of places throughout the Makefile, so I think that it is safe to do that here, too.

This carries the additional benefit that we wouldn't be grubbing through files that have been generated as a part of a previous step in script/cibuild, Makefile, or similar.

ttaylorr: Oh my, please ignore this comment. I glossed over `git` and thought that this was a recursive grep over the filesystem. `git grep` is certainly what we want here. Thanks!~

bk2204:
Okay, so I've addressed the CI failures (I think the AppVeyor one was unrelated).  I did change the invocation we use in the CI script to address false successes due to `git grep -P` missing (since PCRE was not compiled in) and to fix the issue with us flagging batch files falsely.  The commit message has been updated accordingly.

Feel free to take a look and see if you still think it's okay, and if so, I'll merge it.

PastelMobileSuit:
:shipit:

PR #3301, opened 2018-10-04
Remove `git lfs clone` command from man

gtsiolis:
The warning you get when you run `git lfs clone` is a bit misleading (see https://github.com/git-lfs/git-lfs/commit/47fa8523cda3582937077b63dd71d33db5390731#r30762174). This will remove the `git lfs clone` command from `man` (see https://github.com/git-lfs/git-lfs/commit/47fa8523cda3582937077b63dd71d33db5390731#r30753241). @PastelMobileSuit could you review this? 🍄

### DIFF
--- a/docs/man/git-lfs.1.ronn
+++ b/docs/man/git-lfs.1.ronn
@@ -33,8 +33,6 @@ commands and low level ("plumbing") commands.
     Display the Git LFS environment.
 * git-lfs-checkout(1):
     Populate working copy with real content from Git LFS files.
-* git-lfs-clone(1):
-    Efficiently clone a Git LFS-enabled repository.
 * git-lfs-fetch(1):
     Download Git LFS files from a remote.
 * git-lfs-fsck(1):


docs/man/git-lfs.1.ronn
```diff
@@ -33,8 +33,6 @@ commands and low level ("plumbing") commands.
     Display the Git LFS environment.
 * git-lfs-checkout(1):
```
gtsiolis: You get the same warning message for `git lfs checkout`. Maybe remove this one, too?

gtsiolis:
Thanks @PastelMobileSuit! I've also removed `git lfs checkout` in https://github.com/git-lfs/git-lfs/pull/3301/commits/f5a42ec24bc7ea7fa0400e7124f9259df0fb3c9d. @ttaylorr let me know if you're ok with this. Otherwise I can revert the adittional change.

gtsiolis:
@ttaylorr thanks for the meticulous response. Yes, all `git lfs clone` related commands should work as before. Also, I agree to not removing `git lfs checkout` since it's no longer deprecated. However, should we remove the `WARNING` message or the whole `msg` in the following part? https://github.com/git-lfs/git-lfs/blob/b8f9b11a2a55289fb77418fef87fa98b04d66510/commands/command_checkout.go#L38-L44

> Can you please drop this change (instead of reverting it, so as not to add additional commits to the history)? Thank you!

I shouldn't have used the word *revert* in this repository! I meant to say *remove* or even better *reset --hard*. I've dropped the change and updated the pull request title. 🍉

ttaylorr:
> However, should we remove the `WARNING` message or the whole `msg` in
> the following part?
> https://github.com/git-lfs/git-lfs/blob/b8f9b11a2a55289fb77418fef87fa98b04d66510/commands/command_checkout.go#L38-L44

@bk2204 can you take a look at this?

ttaylorr:
@gtsiolis Thanks again for your wonderful contribution!

gtsiolis:
Thanks @ttaylorr @PastelMobileSuit! 🍉

bk2204:
Sure.  I'll put in a pull request to undeprecate it.

gtsiolis:
@bk2204 I've pushed some changes in https://github.com/gtsiolis/git-lfs/commit/a411779c597686f172457a45266c80b7982c559a. Let me know if you think this is ok to submit a pull request. If this is more complex than this, feel free to take this over. 

I've also noticed the `git lfs clone` warning is also used in a test in `t/t-clone-deprecated.sh` Maybe remove this file, too? /cc @ttaylorr

bk2204:
@gtsiolis Sorry, I didn't notice your comment in time.  I just put in a pull request with identical changes at #3303.

gtsiolis:
No problem @bk2204! I'm glad we made identical changes.

gtsiolis:
> I've also noticed the `git lfs clone` warning is also used in a test in `t/t-clone-deprecated.sh` Maybe remove this file, too?

@bk2204 what do you think of this?

bk2204:
I think as long as the command itself still exists, it would be better to keep the test for the deprecated message.  Of course, when we remove the command in v3.0.0, we can remove the test for the deprecated warning.

bk2204:
Not a problem.  Glad you're thinking about ways to keep things tidy.

ttaylorr:
> is this suggestion from the [Tutorial](https://github.com/git-lfs/git-lfs/wiki/Tutorial) still valid or should it be updated as well?


Ah, this should certainly be updated.

bk2204:
I've fixed that text so it's now accurate.  Thanks for pointing it out!

mloskot:
Thank you!

PR #3304, opened 2018-10-05
Honor umask and core.sharedRepository

### LINKED ISSUE #2015: git-lfs does not respect umask values
When a file is added to git-lfs, the user's umask is not respected. Tested on RHEL6:

```
> umask
2
> cp test1.bin test2.bin
> echo haha >> test2.bin
> git add test2.bin
> git commit -a -m 'testing 12344444444'
[test_repo 0cb3423] testing 12344444444
create mode 100644 master/test2.bin
> git show -1
commit 0cb342369c2fee3abd6689904282dd153f542cf9
Author: Test User <test@gmail.com>
Date:   Thu Mar 2 13:25:52 2017 -0800

    testing 12344444444

diff --git a/master/test2.bin b/master/test2.bin
new file mode 100644
index 0000000..506b686
--- /dev/null
+++ b/master/test2.bin
@@ -0,0 +1,3 @@
+oid
sha256:d46e4b618904f939438a3f4fc13698876254fa5dc13b375631b3fb2e32c2f794
+size 801470
> find ../.git/lfs/objects/ -name
d46e4b618904f939438a3f4fc13698876254fa5dc13b375631b3fb2e32c2f794 -type f
-exec ls -l {} \;
-rw------- 1 testuser wheel 801470 Mar  2 13:25
../.git/lfs/objects/d4/6e/d46e4b618904f939438a3f4fc13698876254fa5dc13b37563
1b3fb2e32c2f794

```

This is, in the scheme of things, fairly minor. However, it is a bit odd that it happens with git-lfs and not git.

### LINKED ISSUE #2992: Does not respect umask nor core.sharedRepository
git-lfs creates internal storage in a way which prevents other users to access it.  E.g. when cloning an lfs enabled repository with

```sh
umask 0000
git -c core.sharedRepository=0777 clone https://gitlab.com/ensc/test-ci.git
```

I would expect that other users can work on it.  But filesystem objects are not group- or world writable:

```
$ ls -ld .git/lfs .git/lfs/tmp
drwxr-sr-x. 5 ensc ensc 100  4. Mai 11:47 .git/lfs
drwxr-sr-x. 2 ensc ensc  40  4. Mai 11:47 .git/lfs/tmp
```

and git operations executed by other others will fail:

```
$  su nobody -s /bin/sh -c 'git checkout -B foo HEAD'
Error cleaning LFS object: open /tmp/test-ci/.git/lfs/tmp/129768310: permission denied
fatal: lfs-file: clean filter 'lfs' failed
```

This causes problems in our CI builders which fetch the repository in a first step and build software as non-root in a second one.  See https://gitlab.com/gitlab-org/gitlab-runner/issues/3269 too.

git-lfs-2.4.0-1.fc27.x86_64
git-2.14.3-3.fc27.x86_64

bk2204:
When writing files into the repository LFS object store, Git LFS uses a temporary file and then renames it into place.  For security reasons, temporary files are created with 600 permissions; however, this causes problems in repositories accessed by multiple users, since neither the umask nor `core.sharedRepository` are honored.

This series introduces support for umask and `core.sharedRepository` into Git LFS.

This fixes #2015 and also fixes #2992.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 
@@ -81,8 +82,33 @@ func NewIn(workdir, gitdir string) *Configuration {
 }
 
 func (c *Configuration) getMask() int {
+	// This logic is necessarily complex because Git's logic is complex.
 	c.maskOnce.Do(func() {
-		c.mask = umask()
+		val, ok := c.Git.Get("core.sharedrepository")
+		if !ok {
+			val = "umask"
+		} else if Bool(val, false) {
+			val = "group"
+		}
+
+		switch strings.ToLower(val) {
+		case "group", "true", "1":
+			c.mask = 007
+		case "all", "world", "everybody", "2":
+			c.mask = 002
+		case "umask", "false", "0":
+			c.mask = umask()
+		default:
+			if mode, err := strconv.ParseInt(val, 8, 16); err != nil {
+				// If this doesn't look like an octal number, then it
+				// could be a falsy value, in which case we should use
+				// the umask, or it's just invalid, in which case the
+				// umask is a safe bet.
+				c.mask = umask()
+			} else {
+				c.mask = 0666 & ^int(mode)
+			}
+		}
 	})
 	return c.mask
 }
--- a/config/config.go
+++ b/config/config.go
@@ -48,6 +48,8 @@ type Configuration struct {
 	loadingGit sync.Mutex // guards initialization of local git and working dirs
 	remotes    []string
 	extensions map[string]Extension
+	mask       int
+	maskOnce   sync.Once
 }
 
 func New() *Configuration {
@@ -76,6 +78,13 @@ func NewIn(workdir, gitdir string) *Configuration {
 		},
 	}
 	return c
+}
+
+func (c *Configuration) getMask() int {
+	c.maskOnce.Do(func() {
+		c.mask = umask()
+	})
+	return c.mask
 }
 
 func (c *Configuration) readGitConfig(gitconfigs ...*git.ConfigurationSource) Environment {
@@ -451,4 +460,10 @@ func (c *Configuration) CurrentCommitter() (name, email string) {
 	name, _ = c.Git.Get("user.name")
 	email, _ = c.Git.Get("user.email")
 	return
+}
+
+// RepositoryPermissions returns the permissions that should be used to write
+// files in the repository.
+func (c *Configuration) RepositoryPermissions() os.FileMode {
+	return os.FileMode(0666 & ^c.getMask())
 }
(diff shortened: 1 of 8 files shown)


Noon110:
1

config/config.go
```diff
@@ -452,3 +485,13 @@ func (c *Configuration) CurrentCommitter() (name, email string) {
...
 }
+
+// RepositoryPermissions returns the permissions that should be used to write
+// files in the repository.
+func (c *Configuration) RepositoryPermissions() os.FileMode {
+	if c.mask == -1 {
+		c.setMask()
+	}
+
+	return os.FileMode(0666 & ^c.mask)
```
ttaylorr: What do you think about making this a function called `getMask()` or similar, like:

```go
func (c *Configuration) getMask() {
        c.maskOnce.Do(func() {
                c.mask = umask()
        })
        return c.mask
}
```

I think that that would allow us to forgo the `if c.mask == -1` conditional above, and rewrite this function as:

```go
return os.FileMode(0666 & ^c.getMask())
```

config/util_nix.go
```diff
@@ -0,0 +1,11 @@
+// +build !windows
+
+package config
+
+import "syscall"
+
+func umask() int {
+	umask := syscall.Umask(022)
```
ttaylorr: Can you restore the comment about how `syscall.Umask` sets the umask (so we have to unset it, which we do on L9)?

config/config.go
```diff
@@ -75,9 +77,39 @@ func NewIn(workdir, gitdir string) *Configuration {
 			return c.readGitConfig(sources...)
 		},
 	}
+	c.mask = -1
 	return c
 }
 
+func (c *Configuration) setMask() {
```
ttaylorr: Nicely done, and nicely tested below.

lfs/util.go
```diff
@@ -223,3 +223,24 @@ func LinkOrCopy(cfg *config.Configuration, src string, dst string) error {
...
+func TempFile(cfg *config.Configuration, pattern string) (*os.File, error) {
+	tmp, err := ioutil.TempFile(cfg.TempDir(), pattern)
+	if err != nil {
+		return nil, err
+	}
+
+	perms := cfg.RepositoryPermissions(false)
+	err = os.Chmod(tmp.Name(), perms)
+	if err != nil {
+		tmp.Close()
```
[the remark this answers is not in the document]
bk2204: Sure.  I think that's a good thing to document.

config/util_nix.go
bk2204: Sure.

config/config.go
bk2204: Yeah, that would be tidier, I agree, and it would avoid us working with umask more than once.  I don't think we end up calling this in a multi-threaded context, but in case a refactor happens and we do, that would avoid any problems.

bk2204:
Oh, gee, I just realized that I did a terrible job of squashing my fixes into the right commits.  Let me fix that.

bk2204:
Okay, that should address all the issues, both the ones that you noted, and the ones due to my missquash.

PR #3307, opened 2018-10-08
Add new `.netrc` credential helper

### LINKED ISSUE #3130: Authentication issue on MacOS

PastelMobileSuit:
This PR replaces the old `.netrc` logic in `doWithAuth()` with a new credential helper to handle the parsing of credentials from a `.netrc` file.

Since we're now using a credential helper for this, it also means that if `.netrc` credentials prove to be invalid, the request will automatically be retried with different credentials if they're available.

A note for reviewers: `creds/netrc.go` and `creds/netrc_test.go` were both moved over from the `lfsapi` package with some additional changes. Git unfortunately treated these as totally new files, and so the whole file appears as modified in the diff.

For `creds/netrc.go`, everything [from line 39](https://github.com/git-lfs/git-lfs/compare/netrc-retry?expand=1#diff-8614b6f60a0b40b5cdec0298392bdbc2R39) downwards is new, with everything above it just being copied over from the old file (aside from the package change).

For `creds/netrc_test.go`, I had to rewrite every test to use the new credential helper and preserve what was being tested previously. As a result, that file should be reviewed in its entirety.

/cc @git-lfs/core

Closes #3130


creds/netrc.go
```diff
@@ -0,0 +1,130 @@
...
+
+func (f *noFinder) FindMachine(host string) *netrc.Machine {
+	return nil
+}
+
+// NetrcCredentialHelper retrieves credentials from a .netrc file
+type netrcCredentialHelper struct {
+	netrcFinder NetrcFinder
+	skip        map[string]bool
+}
```
bk2204: I really like the approach of structuring this as a credential helper.  It makes the entire authentication path much easier to reason about.

PastelMobileSuit: Thanks, it was actually @ttaylorr's suggestion originally!

I like the way it turned out, I agree it makes authentication much easier to reason about. Another big plus: `.netrc` credentials were previously wrongly taking priority over credentials provided directly in a URL, and now they won't since credential helpers are queried at the end of the auth flow!

PR #3311, opened 2018-10-11
go.mod: depend on tagged gitobj

ttaylorr:
This pull requests cleans up and updates our go.mod to depend on the tagged version of gitobj, which is tagged to be identical to the current HEAD (thus no code changes are introduced in this change).

##

/cc @git-lfs/core


go.sum
```diff
@@ -6,6 +6,8 @@ github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/git-lfs/gitobj v0.0.0-20180907143119-1e97572956c1 h1:q4kxfk1Z2lnZ2b9d43GM/xTIyt4Jpx8RXWLKF8WPrtM=
 github.com/git-lfs/gitobj v0.0.0-20180907143119-1e97572956c1/go.mod h1:EdPNGHVxXe1jTuNXzZT1+CdJCuASoDSLPQuvNOo9nGM=
```
bk2204: It looks like we have some references to the old version number here.  Did we want to tidy these up?

ttaylorr: These are kind of mysterious to me. I think that Go doesn't clean this file up ever, so I resolved this by basically doing:

```
git rebase -x 'rm go.sum && make go.sum && git add go.sum && git commit --amend' HEAD~2
```

bk2204: Yeah, I got that impression as well.  I think that's why I ended up accidentally introducing the go-spew stuff, since I regenerated the file by hand. (Thanks for fixing that, by the way.)

Issue #3312, opened 2018-10-11
Honor GIT_COMMITTER_DATE and GIT_AUTHOR_DATE


bk2204:
Git uses these environment variables to set the date on commits if they're set.  Git LFS should honor them as well when writing commits.

bk2204:
As a note here, after some investigative work, the major stumbling point is the Git date format.  The ISO 8601-style formats and RFC 5322 format are easily accomplished (and I have code and tests for them).

Go's time library doesn't parse Unix timestamps.  This means we need to handle the parsing on our own.  Since we're wanting to create a `time.Time` object, we need to produce such an object based off a Unix timestamp and a time zone (a `time.Location`).  The `time.In` function may be useful.

PR #3313, opened 2018-10-11
Ensure correct syntax for commit headers in lfs migrate import

### LINKED ISSUE #3310: Import with `--no-rewrite` option adds incorrect commit timestamp
I've tried to use the import migration mode and this added in a new commit with the following date:

```shell
Date:   Thu Jan 1 00:00:00 1970 +0000 
```

The migration used is the following:

```shell
git lfs migrate import --no-rewrite -m'Commit message...' path/to/files/*.png
```
This can be replicated without the commit message, too.

bk2204:
We accidentally wrote ill-formed commit headers when writing commits for `git lfs migrate import --no-rewrite` because the author and committer fields lacked timestamps.  In addition, we could write empty email addresses if the user did not have a `user.email` field set but instead used the `EMAIL` environment variable to set their email.

Correct both of these issues and honor the distinction between author and committer.  Note that due to limitations of the Go standard library, this series does not honor `GIT_AUTHOR_DATE` and `GIT_COMMITTER_DATE` even though it honors the name- and email-related environment variables; this will be tracked in #3312.

This fixes #3310.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -479,7 +479,7 @@ func (c *Configuration) loadGitConfig() {
 	}
 }
 
-// CurrentCommitter returns the name/email that would be used to author a commit
+// CurrentCommitter returns the name/email that would be used to commit a change
 // with this configuration. In particular, the "user.name" and "user.email"
 // configuration values are used
 func (c *Configuration) CurrentCommitter() (name, email string) {
--- a/config/config.go
+++ b/config/config.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/git-lfs/git-lfs/fs"
 	"github.com/git-lfs/git-lfs/git"
@@ -51,6 +52,7 @@ type Configuration struct {
 	extensions map[string]Extension
 	mask       int
 	maskOnce   sync.Once
+	timestamp  time.Time
 }
 
 func New() *Configuration {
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -483,13 +483,48 @@ func (c *Configuration) loadGitConfig() {
...
+	email, ok = c.Os.Get("GIT_" + envType + "_EMAIL")
+	if !ok {
+		email, ok = c.Git.Get("user.email")
+	}
+	if !ok {
+		email, _ = c.Os.Get("EMAIL")
+	}
+
+	// Git filters certain characters out of the name and email fields.
+	name = strings.Map(filterString, name)
```
ttaylorr: Nice use of `strings.Map`. I think that `filterString` is too generically named to be exposed privately at the package-level scope. What do you think about either clarifying the name (perhaps, `filterSignatureString`) or introducing the function within this function scope, like:

```go
func (c *Configuration) findUserData(envType string) (name, email string) {
        var filter func(rune) rune = func(r rune) rune {
                // ...
        }

        name = strings.Map(filter, name)
}
```

t/t-migrate-import-no-rewrite.sh
```diff
@@ -33,6 +33,9 @@ begin_test "migrate import --no-rewrite (default branch)"
   # Ensure a new commit message was generated based on the list of imported files
   commit_msg="$(git log -1 --pretty=format:%s)"
   echo "$commit_msg" | grep -q "a.txt: convert to Git LFS"
+
+  # Ensure we write a valid commit object.
+  git fsck
```
ttaylorr: ✨

config/config.go
bk2204: Yeah, I was worried about that.  I think the nested function is the way to go.  I'll make that change.

PR #3314, opened 2018-10-12
config: honor GIT_AUTHOR_DATE and GIT_COMMITTER_DATE

### LINKED ISSUE #3312: Honor GIT_COMMITTER_DATE and GIT_AUTHOR_DATE

bk2204:
Git can use these environment variables to take a date for creating
commits.  Teach Git LFS these variables as well.

Note that because Go's time library doesn't parse Unix timestamps, we
need to parse Git's native format ourselves.

Fixes #3312.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -483,6 +484,30 @@ func (c *Configuration) loadGitConfig() {
 	}
 }
 
+// All the date formats that Git accepts, except for the built-in one, which is
```
ttaylorr: Nit: godoc comments must start with the name of the identifier, e.g., `// dateFormats is a list of all the ...`.

config/config.go
```diff
@@ -483,6 +484,30 @@ func (c *Configuration) loadGitConfig() {
...
+	"2006-01-02 15:04:05Z",
+	"2006.01.02T15:04:05Z",
+	"2006.01.02 15:04:05Z",
+	"01/02/2006T15:04:05Z",
+	"01/02/2006 15:04:05Z",
+	"02.01.2006T15:04:05Z",
+	"02.01.2006 15:04:05Z",
+}
+
+var defaultDatePattern = regexp.MustCompile(`\A(\d+) ([+-])(\d{2})(\d{2})\z`)
```
ttaylorr: Should these two belong to the same `var` block, as in:

```go
var (
        // ...
)
```

config/config.go
```diff
@@ -520,6 +545,40 @@ func (c *Configuration) findUserData(envType string) (name, email string) {
...
+		unixSecs, _ := strconv.ParseInt(strs[1], 10, 64)
+		hours, _ := strconv.Atoi(strs[3])
+		offset, _ := strconv.Atoi(strs[4])
+		offset = (offset + hours*60) * 60
+		if strs[2] == "-" {
+			offset = -offset
+		}
+
+		loc := time.FixedZone("", offset)
+		return time.Unix(unixSecs, int64(0)).In(loc)
```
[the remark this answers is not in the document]
bk2204: It does now.  It did not when I tested it last, but I must have been doing something wrong.  Will fix.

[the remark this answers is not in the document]
bk2204: I went back and forth but split it due to the long line.  I'll inline it.

config/config.go
bk2204: Good point.  Will fix.

PR #3317, opened 2018-10-16
Update CONTRIBUTING

bk2204:
Update the `CONTRIBUTING.md` file to reflect that we no longer use the roadmap file or roadmap issues, that we no longer use Gitter, and that there is now a howto that the maintainer should use for release steps.


bk2204:
Okay, sounds good.  If you see some things that still need to be improved, maybe you could open an issue so we can clean them up.

PR #3320, opened 2018-10-16
commands: list explicitly excluded patterns separately

### LINKED ISSUE #2442: Tracked file patterns having `-filter=lfs` are not listed correclty
### Steps to reproduce
1. In the `.gitattributes` put the following lines:
```
*.zip filter=lfs diff=lfs merge=lfs -text
*.doc filter=lfs diff=lfs merge=lfs -text
```
2. In a subdirectory `core` put the following `.gitattributes`:
```
*.doc -filter=lfs diff=lfs merge=lfs -text
```
Note that the filter is prepended with "-" to NOT track `*.doc` files in this subdirectory.

3. Call `git lfs track` to list the patterns:
```
$ git lfs track
Listing tracked patterns
    core\*.doc (core\.gitattributes)
    *.zip (.gitattributes)
    *.doc (.gitattributes)
```
=> It is not visible that `core\*.doc` is NOT tracked

4. Add a `*.doc` file in the `core` subdirectory, comit and push to origin
=> The file is not tracked by LFS as desired

### Conclusion
When `.gitattributes` contains `-filter=lfs` the `-` seems not to be respected when listing tracked patterns. But for commiting and pushing the `-` is correctly interpreted.

### Environment
```
git-lfs/2.1.1 (GitHub; windows amd64; go 1.8.1; git 3314e287)
git version 2.13.2.windows.1
```

bk2204:
Currently, if a user has disabled Git LFS for a pattern in a .gitattributes file (such as in a subdirectory) by writing "-filter=lfs", we render the pattern as a tracked pattern in "git lfs track", even though those paths are not enabled for Git LFS.

Improve this situation by listing excluded patterns as well as tracked patterns.  Compute whether a pattern is excluded by looking for an attribute entry that starts with "-filter" and if so, list the entry as excluded.  Add tests for both the "-filter" and "-filter=lfs" situations.

Fix #2442.

### DIFF
--- a/git/attribs.go
+++ b/git/attribs.go
@@ -13,7 +13,8 @@ import (
 )
 
 const (
-	LockableAttrib = "lockable"
+	LockableAttrib      = "lockable"
+	FilterDisableAttrib = "-filter"
 )
 
 // AttributePath is a path entry in a gitattributes file which has the LFS filter
@@ -24,6 +25,8 @@ type AttributePath struct {
 	Source *AttributeSource
 	// Path also has the 'lockable' attribute
 	Lockable bool
+	// Path is handled by Git LFS (i.e., filter=lfs)
+	Tracked bool
 }
 
 type AttributeSource struct {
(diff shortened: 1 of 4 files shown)


git/attribs.go
```diff
@@ -102,11 +105,14 @@ func attrPaths(path, workingDir string) []AttributePath {
 			continue
 		}
 
+		has_filter := strings.Contains(line, "filter=lfs")
```
ttaylorr: Nit: Go variables should be camel-cased.

bk2204: Ah, yes.  Will fix.

bk2204:
I was slightly worried about the parsing issue as well, but rationalized it the way that you did in that this is a porcelain, not a plumbing command. I think a `--no-excluded` option would be a good middle ground, in that it provides people an escape route if they do want to continue to parse the output of `git lfs track`.

bk2204:
Also, to respond to your suggestion about refactoring this, I think that is indeed a good idea, but I'd like to put that on the backlog for 2.7.0 instead of doing that here, and I'll open an issue to note that down.

bk2204:
There's now a `--no-excluded` flag (and documentation) and I have stopped writing Rust-style variable names in Go.  I've also opened #3328 to document our planned refactor for 2.7.0.

PR #3326, opened 2018-10-17
Uninstall improvements

### LINKED ISSUE #3322: `--skip-repo` option for uninstall
Splitting off from discussion in #3013.

`git-lfs install` has an option for `--skip-repo` to skip the local hook setup.  Can this option be added to uninstall as well to skip the local hook removal attempt?  I think that would prevent this rev-parse error in the trace output.

```
# GIT_TRACE=1 ./git-lfs uninstall --system
16:16:56.470721 trace git-lfs: exec: git 'version'
16:16:56.473940 trace git-lfs: exec: git 'config' '--system' '--remove-section' 'filter.lfs'
16:16:56.480570 trace git-lfs: Error running 'git rev-parse': Failed to call git rev-parse --git-dir --show-toplevel: "fatal: not a git repository (or any of the parent directories): .git\n"
Global Git LFS configuration has been removed.
```

### LINKED ISSUE #3323: misleading uninstall output (global vs system)
Splitting off from discussion in #3013.

When running uninstall with the `--system` flag, the output indicates that the global git config (`~/.gitconfig`) was removed, which is inaccurate.  It removes the system config (`/etc/gitconfig`).

```
# ./git-lfs uninstall --system
Global Git LFS configuration has been removed.
```

### LINKED ISSUE #3324: `--system` flag not in uninstall help
The uninstall command allows a `--system` flag that works as expected (other than the output #3323), but that flag is not described in the output of `git-lfs uninstall --help`.

```
$ ./git-lfs uninstall --help
git lfs uninstall

Perform the following actions to remove the Git LFS configuration:

* Remove the "lfs" clean and smudge filters from the global Git config.
* Uninstall the Git LFS pre-push hook if run from inside a Git repository.
  
Options:

* --local:
    Removes the "lfs" smudge and clean filters from the local repository's git
    config, instead of the global git config (~/.gitconfig).
```

bk2204:
This series introduces a set of improvements to `git lfs uninstall`.  First, the `--system` option is documented.  Second, we print a more accurate message about which configuration is modified when the `--system` option is provided.  Finally, we add a `--skip-repo` option for parallelism with `git lfs install`.

I will admit that there is some slight duplication in 5541da6, but I was unable to find an elegant way to avoid it.

This series fixes #3322, fixes #3323, and fixes #3324.

### DIFF
--- a/commands/command_uninstall.go
+++ b/commands/command_uninstall.go
@@ -10,7 +10,7 @@ func uninstallCommand(cmd *cobra.Command, args []string) {
 		Error(err.Error())
 	}
 
-	if localInstall || cfg.InRepo() {
+	if !skipRepoInstall && (localInstall || cfg.InRepo()) {
 		uninstallHooksCommand(cmd, args)
 	}
 
@@ -34,6 +34,7 @@ func init() {
 	RegisterCommand("uninstall", uninstallCommand, func(cmd *cobra.Command) {
 		cmd.Flags().BoolVarP(&localInstall, "local", "l", false, "Set the Git LFS config for the local Git repository only.")
 		cmd.Flags().BoolVarP(&systemInstall, "system", "", false, "Set the Git LFS config in system-wide scope.")
+		cmd.Flags().BoolVarP(&skipRepoInstall, "skip-repo", "", false, "Skip repo setup, just uninstall global filters.")
 		cmd.AddCommand(NewCommand("hooks", uninstallHooksCommand))
 	})
 }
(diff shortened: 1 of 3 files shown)


carlwgeorge:
I'm not familiar with your release cadence.  Will this be in a tagged release soon?  If not I'm fine with adding it to Fedora's RPMs with a patch.

ttaylorr:
> I'm not familiar with your release cadence. Will this be in a tagged release soon?

We usually release a new 2.x.0 version every ~3 or so months. There should be one out next week, which will include this.

carlwgeorge:
Perfect, I won't bother with a local patch and just wait till next week.

PR #3327, opened 2018-10-17
Clarify minimum git version

carlwgeorge:
Related change: 632f4881293d6974f169adce3a0f53b4e96393bb

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ To https://github.com/git-lfs/git-lfs-test
    67fcf6a..47b2002  master -> master
 ```
 
-Note: Git LFS requires Git v1.8.5 or higher.
+Note: Git LFS requires at least Git 1.8.2 on Linux or 1.8.5 on macOS.
 
 ## Limitations
 


bk2204:
First of all, welcome, and congratulations on your first pull request to Git LFS!

Since I'm relatively new to the project, I'm not sure what the history of our version requirements is.  It looks like at some point, we were testing 1.8.5 and have incrementally moved to 2.0.0.  The commit messages don't seem to provide a great deal of clarity, so maybe @ttaylorr can say a few words about why those changes were made.

The code does indeed seem to support 1.8.2, and I'm not opposed to updating the documentation to reflect this.  I expect you want to use RHEL 7's shipped git packages for this, and presumably that works fine, or we would have heard from the EPEL folks by now.  If possible, I'd like to see if we can get at least Travis testing this configuration so we can have confidence that we don't break it accidentally.

carlwgeorge:
For what it's worth, the [code still enforces 1.8.2](https://github.com/git-lfs/git-lfs/blob/v2.5.2/commands/commands.go#L447) as the minimum.

You are spot on, RHEL7 has git 1.8.3.1 (with multiple backported security fixes).  I am in fact one of said EPEL folks, I co-maintain the git-lfs package in Fedora and EPEL.

I don't know how easy it will be to test with that old a git version in Travis.  Isn't that an Ubuntu based environment?

bk2204:
Yeah, I took a look in the code, and we definitely are requiring 1.8.2 there.  I just want to be sure that there's not some reason things shouldn't work related to the Travis CI and CircleCI testing.  We clone Git and built it ourselves for the very old versions, and 1.8.2 does appear to build on Ubuntu 18.04, if quite noisily.

Let's see what @ttaylorr has to say about the testing, and assuming there's no hidden bombshell, I see no reason why we shouldn't take this.

ttaylorr:
> Let's see what @ttaylorr has to say about the testing, and assuming there's no hidden bombshell, I see no reason why we shouldn't take this.

I don't think that there's anything :bomb:-like going on here. As I recall, we tried to add some test, and realized that it mysteriously did not work on Git 1.8.5, but _did_ work on Git 2.0.0. I think at the time we decided to leave it (i.e., by testing on Git 2.0.0 as the "minimum version", because the change seemed unrelated to a 1.8.5-specific thing), and come back to it later. I think that we never managed to do the last part, though ;-).

All of that is to say that I think that this change looks good to me.

ttaylorr:
@bk2204 👍/ 👎 for taking this in v2.6.0? I think that it'd be OK to do so.

bk2204:
Yeah, I'm for taking this.  It's merely documenting what our current policy is, and so I can't see any reason why we shouldn't.  Ideally we'd fix the CI such that it works on 1.8.2 as well, but that can come in separately.

PR #3330, opened 2018-10-18
Fix formatting of INCLUDE AND EXCLUDE (REFS)

jsantell:
Fix formatting of the INCLUDE AND EXCLUDE (REFS) reference chart and the results to properly display newlines in Github Markdown


ChungTranThanh:
szxt14X3xrFwGybZdcQelFD6

PR #3341, opened 2018-10-24
tq: enable transfer debugging when GIT_CURL_VERBOSE is set

bk2204:
When something goes wrong, many users will happily provide output using
GIT_TRACE and GIT_CURL_VERBOSE.  This is very useful for Git, but Git
LFS doesn't honor the latter (since we don't use libcurl).  Instead, we
require users to set GIT_TRANSFER_TRACE.  This makes for a prolonged and
more difficult debugging experience.

Despite the fact that we don't use libcurl, honor the GIT_CURL_VERBOSE
flag and act as if the user had set GIT_TRANSFER_TRACE instead.  This
honors the intent of what the user wanted (transfer-related debugging
information) and makes debugging easier for end users and maintainers
alike.

### DIFF
--- a/lfs/lfs.go
+++ b/lfs/lfs.go
@@ -87,6 +87,8 @@ func init() {
 	if len(os.Getenv("GIT_TRACE")) < 1 {
 		if tt := os.Getenv("GIT_TRANSFER_TRACE"); len(tt) > 0 {
 			os.Setenv("GIT_TRACE", tt)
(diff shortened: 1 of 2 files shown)


bk2204:
I was having the exact same thought: where would a test go?  Normally I'm all for tests, but I think in this case the best we could do would be to replace an existing call, so I'm going to skip one this time.

Issue #3200, opened 2018-08-28
Too many open files when pushing lots of files


tardyp:
I am converting a big project to git lfs using git lfs migrate import

I was perhaps a bit too wide to the number of extensions I enabled, and that resulted with lots of file to upload (more than 10k)

The resulting git push fails with:

```
basic upload: open /[...]/.git/lfs/objects/a0/7d/a07d980ebff56456a0e0a1c4860a6ebf9d36f29153b3c9c794f3b1ada7a0a321: too many open files
basic upload: open /[...]/.git/lfs/objects/ec/9b/ec9b2ffebfa3260ca3f2cca021167a792b1758918e1d7d366d6496ba03de14e9: too many open files
basic upload: open /[...]/.git/lfs/objects/53/26/5326c076aac046c5093f15c4f043e9f0d41595ad3d7be3050bd6b7611a6f8ec3: too many open files
basic upload: open /[...]/.git/lfs/objects/15/43/15431f4ceec5a9b303f6723fe8fb10e881fa480a7e95cd210ed9fb30c121334f: too many open files
basic upload: open /[...]/.git/lfs/objects/cd/ba/cdba9ff4deac3fdbcf9e3684d0bfc3a7ade18b79d82df8865218b80da86b5473: too many open files
basic upload: open /[...]/.git/lfs/objects/8f/b4/8fb4a36b154da3aece5380c68437b379f052eafe68ea60d0820eab73c8acecc9: too many open files
basic upload: open /[...]/.git/lfs/objects/7b/81/7b81c8c4e50f8c6b9ca2d45e8d940cfeea4aaeb3a47234051e0b43341a039fd3: too many open files
basic upload: open /[...]/.git/lfs/objects/ba/c8/bac8add63f28740fd4fbc05a836965e49728195f14ead7090e1392f2885c7f55: too many open files
basic upload: open /[...]/.git/lfs/objects/8e/6c/8e6cc2167905185b3d23aab206ceaa5a341a401ab973a6cbda829efdbaf00650: too many open files
basic upload: open /[...]/.git/lfs/objects/88/6c/886c3eca6e1fb62449260c8c22e2acfdcbd1010d127dec8e76c2c4c623c91363: too many open files
basic upload: open /[...]/.git/lfs/objects/5d/29/5d29dd5f4e75265890715b334f99bb89a2b83d146b548fab65cb86e6d9ad439c: too many open files
[...]

PR #3347, opened 2018-10-29
Close attributes file

mbsulliv:
This change was discussed in #3200.


bk2204:
Thanks for this change.  This looks good to me, and the failure on AppVeyor appears to be a false positive.

PR #3349, opened 2018-10-30
tasklog: don't log progress status when stdout is not a tty

steffengodskesen:
To avoid filling logs with progress status when running git/git-lfs inside tools and services where the output does not go to a terminal.

The behaviour is overridable by either
 - setting `lfs.forceprogress 1`
 - setting `GIT_LFS_FORCE_PROGRESS=1`

Ideally this should be controllable by command line args (like `git push --progress`) but passing command line args down to filters and hooks is not really an option.

### DIFF
--- a/docs/man/git-lfs-config.5.ronn
+++ b/docs/man/git-lfs-config.5.ronn
@@ -306,6 +306,16 @@ be scoped inside the configuration for a remote.
   * `total` The entire size of the file, in bytes.
   * `name` The name of the file.
 
+* `GIT_LFS_FORCE_PROGRESS`
+  `lfs.forceprogress`
+
+  Controls whether Git LFS will suppress progress status when the standard
+  output stream is not attached to a terminal. The default is `false` which
+  makes Git LFS detect whether stdout is a terminal and suppress progress when
+  it's not; you can disable this behaviour and force progress status even when
+  standard output stream is not a terminal by setting either variable to 1,
+  'yes' or 'true'.
+
 * `GIT_LFS_SET_LOCKABLE_READONLY`
   `lfs.setlockablereadonly`
 
--- a/go.sum
+++ b/go.sum
@@ -1,5 +1,6 @@
 github.com/ThomsonReutersEikon/go-ntlm v0.0.0-20151030004737-b00ec39bbdd0 h1:iBnLwXNz+Mz2K6MndIM9a4sLBb+TouHB8s0tvx16s4s=
 github.com/ThomsonReutersEikon/go-ntlm v0.0.0-20151030004737-b00ec39bbdd0/go.mod h1:0eFwySJYxbNw/r8cJ01IeztpJfwrgrmiLtYig80Yvrc=
+github.com/alexbrainman/sspi v0.0.0-20180125232955-4729b3d4d858 h1:OZQyEhf4BviydsRdmK4ryeJHotDLd7vL1X8+nnxXkfk=
 github.com/alexbrainman/sspi v0.0.0-20180125232955-4729b3d4d858/go.mod h1:976q2ETgjT2snVCf2ZaBnyBbVoPERGjUz+0sofzEfro=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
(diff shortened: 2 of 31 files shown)


docs/man/git-lfs-config.5.ronn
```diff
@@ -306,6 +306,16 @@ be scoped inside the configuration for a remote.
   * `total` The entire size of the file, in bytes.
   * `name` The name of the file.
 
+* `GIT_LFS_FORCE_PROGRESS`
+  `lfs.forceprogress`
+
+  Controls whether Git LFS will suppress progress status when the
+  standard output stream is not attached to a terminal. The default is `false`
+  which make Git LFS detect whether stdout is a terminal and suppress progress
```
bk2204: Typo: ”make“ instead of ”makes”

go.sum
```diff
@@ -13,6 +14,8 @@ github.com/inconshreveable/mousetrap v1.0.0 h1:Z8tu5sraLXCXIcARxBp/8cbvlwVa7Z1NH
 github.com/inconshreveable/mousetrap v1.0.0/go.mod h1:PxqpIevigyE2G7u3NXJIT2ANytuPF1OarO4DADm73n8=
 github.com/kr/pty v0.0.0-20150511174710-5cf931ef8f76 h1:i5TIRQpbCg4aJMUtVHIhkQnSw++Z405Z5pzqHqeNkdU=
 github.com/kr/pty v0.0.0-20150511174710-5cf931ef8f76/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
+github.com/mattn/go-isatty v0.0.4 h1:bnP0vzxcAdeI1zdubAl5PjU6zsERjGZb7raWodagDYs=
+github.com/mattn/go-isatty v0.0.4/go.mod h1:M+lRXTBqGeGNdLjl/ufCoiOlB5xdOkqRJdNxMWT7Zi4=
```
bk2204: We generally want to vendor these dependencies.  Could you run `make vendor` to pull those into the tree?

tasklog/log.go
```diff
@@ -4,17 +4,25 @@ import (
...
 
 const (
 	DefaultLoggingThrottle = 200 * time.Millisecond
 )
 
+// Environment is the minimal interface that the logger needs to lookup
+// environment variables
+type Environment interface {
+	GetBool(key string, def bool) (val bool)
+}
```
bk2204: I don't see this used anywhere.  Should it be removed?

tasklog/log.go
```diff
@@ -39,8 +50,9 @@ type Logger struct {
 }
 
 // NewLogger retuns a new *Logger instance that logs to "sink" and uses the
-// current terminal width as the width of the line.
-func NewLogger(sink io.Writer) *Logger {
+// current terminal width as the width of the line. Will log progress status if
+// stdout is a terminal or if forceProgress is true
+func NewLogger(sink io.Writer, forceProgress bool) *Logger {
```
ttaylorr: I think that it may be good to adopt the same pattern of [option functions][1] here, too, even though there is only one option that we'd be supporting. I suggest this, because I think the "extra baggage" of option functions is worth the tradeoff to remove a boolean argument. Perhaps:

```go
type optionFn func(l *Logger)

var (
        forced optionFn = func(l *Logger) {
                l.forceProgress = true
        }
)

// NB: this prevents mutating the underlying variable, which _can_ be changed at runtime.
func Forced() optionFn { return forced }

func NewLogger(sink io.Writer, fns ...optionFn) {
        l := &Logger{...}
        for _, fn := range fns {
                fn(l)
        }

        // ...
}
```

Then to call this (with forcing) you would instead say:

```go
l := tasklog.NewLogger(os.Stderr, tasklog.Forced())
```

I think that you would want a helper function that calls that ☝️, since it removes the convenience of simply passing along `cfg.ForceProgress()`.

[1]: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

tasklog/log.go
```diff
@@ -214,6 +227,9 @@ func (l *Logger) logTask(task Task) {
 
 	var update *Update
 	for update = range task.Updates() {
+		if !isatty.IsTerminal(os.Stdout.Fd()) && !l.forceProgress {
```
ttaylorr: Does `isatty` cache this result? If not, and if it's expensive to compute, we should cache it ourselves, since this loop is expected to run many times.

On another note, I don't think that this is quite what we want: indeed this checks whether we can write to stdout, but stdout is not always the `io.Writer` we are writing to. Could you instead say:

```go
type Logger struct {
        ttyOnce sync.Once
        tty bool
}

func (l *Logger) tty() bool {
        l.ttyOnce.Do(func() {
                if f, ok := l.sink.(interface{ // <- could use `*os.File`, too 
                        Fd() uintptr
                }); ok {
                        l.tty = isatty.IsTerminal(f.Fd())
                        return
                }
        })
        return l.tty
}
```

And then rewrite this line to:

```go
if !(l.tty() || l.forceProgress) {
        continue
}
```

tasklog/log.go
steffengodskesen: Agree. Nice pattern; hadn't seen that before. Thanks for the pointer.

I'm not sure I understand the _"NB: this prevents mutating the underlying variable_" comment, though.

tasklog/log.go
steffengodskesen: Checking the (possibly non-existing) `Fd()` on the actual `Writer` makes good sense. Won't work in the cases where the `sink` is a `io.MultiWriter` (`command_fetch.go` and `command_prune.go` does this), but I'll try to come up with a solution to that. Maybe wrapping the `io.MultiWriter` in a type that exposes the `Fd()` of its first writer. 

The `sink` won't change during the lifetime of the logger, so could we not cache result of the call to `isatty` on instantiation of the logger?

PR #3357, opened 2018-11-02
go.mod: add two missing dependencies

avar:
These are both in vendor/modules.txt but not here. When I build with:

    GO111MODULE=on go build -v

I get errors about these being missing:

    vendor/github.com/stretchr/testify/assert/assertions.go:19:2: cannot find package "." in:
            /builddir/build/BUILD/git-lfs-2.6.0/vendor/github.com/davecgh/go-spew/spew
    vendor/github.com/stretchr/testify/assert/assertions.go:20:2: cannot find package "." in:
            /builddir/build/BUILD/git-lfs-2.6.0/vendor/github.com/pmezard/go-difflib/difflib

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/git-lfs/git-lfs
 require (
 	github.com/ThomsonReutersEikon/go-ntlm v0.0.0-20151030004737-b00ec39bbdd0
 	github.com/alexbrainman/sspi v0.0.0-20180125232955-4729b3d4d858
+	github.com/davecgh/go-spew v1.1.1
 	github.com/git-lfs/gitobj v1.0.0
 	github.com/git-lfs/go-netrc v0.0.0-20180525200031-e0e9ca483a18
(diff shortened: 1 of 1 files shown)


bk2204:
Does #3356 fix this for you?

avar:
@bk2204 : I haven't tried, and as someone who doesn't know Go that patch looks too big/scary to carry in our local build v.s. the hack I have. I was just trying to solve the x-y problem of "let's not download pre-build binaries but build from source".

Feel free to close this. I can see in later major releases if this is needed after 2.6.0,

bk2204:
Okay, I think that patch will indeed fix your issue.  We vendor dependencies into `vendor/` for people who don't want to ship their own versions of the Go packages we depend on, which is why it looks so scary.  We'll make sure that patch is in 2.6.1, which should fix your problem.

I'll close this, but feel free to reopen it #3356 doesn't meet your needs.

PR #3360, opened 2018-11-02
filepathfilter: don't say file is both accepted and rejected

bk2204:
The file path filter can mark a file as either accepted or rejected. However, one of the messages that says that a file was rejected is in the main code path, and is therefore always executed, even if the file is actually accepted.  This leads to contradictory messages and noisier output.

Ensure that we indicate that a file is rejected by the filter only if it is indeed rejected; otherwise, say only that it is accepted.


hocalvin79:
Null

PR #3377, opened 2018-11-13
Fix test server API

### LINKED ISSUE #3372: undefined lfsapi.Endpoint in test suite
Currently encountering `# github.com/git-lfs/git-lfs/t/git-lfs-test-server-api/main.go:163:4: undefined: lfsapi.Endpoint` while attempting to run the test. There is no `lfsapi.Endpoint` in the project, so i assume `lfshttp.Endpoint` is correct. However the test suite doesn't work at all with this change.

bk2204:
@Foxboron recently noticed that our server API test program didn't function in the way we expected.  It first of all didn't compile due to several refactors, and then even when fixed, it didn't pass.

This series fixes the code so that it builds, builds it as part of the integration test series to ensure we don't break it again, and fixes the reversed condition that causes it to fail.  I am fairly confident that the last patch is indeed correct, but some :eyes: on that part would be helpful.

This series fixes #3372.

### DIFF
--- a/t/Makefile
+++ b/t/Makefile
@@ -19,7 +19,12 @@ TEST_CMDS += ../bin/lfstest-gitserver$X
 TEST_CMDS += ../bin/lfstest-standalonecustomadapter$X
 TEST_CMDS += ../bin/lfstest-testutils$X
 
-TEST_SRCS = $(wildcard t-*.sh)
+# Not used for the integration tests, but build it here anyway to ensure it
+# continues to work.
+TEST_CMDS += ../bin/git-lfs-test-server-api$X
+
+TEST_SRCS     = $(wildcard t-*.sh)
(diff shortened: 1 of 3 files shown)


Foxboron:
Is there any reason why this fix was not included in `2.6.1`? Still can't build latest release on Arch Linux.

bk2204:
I think we simply didn't think of it when doing the release. It will definitely be included in 2.7.0.

Foxboron:
But `make build` simply fails at this moment. How is waiting for the next release the solution to this?

bk2204:
I'm not seeing `make build` fail on 2.6.1. The lfs-test-server-api binary shouldn't be built as part of a standard build step in that version; I think that's the reason we didn't see it fail.

Generally, if you're building a package, you just need to `make && make test integration`. `make build` builds for every platform, which is more than most people need; we have it there more for maintainer convenience to build a full binary release.

Issue #3373, opened 2018-11-11
Q: Are git lfs track and status usable for bare repositories?


mloskot:
TL;TR: Do `status` and `track` commands support bare repository, at least for read-only/inspection only?

The `status` command is documented as ["Show the status of Git LFS files in the working tree"](https://github.com/git-lfs/git-lfs/blob/master/docs/man/git-lfs-status.1.ronn) and it seems to succeed for a bare repository:

```
test.git (BARE:master) $ git lfs status
On branch master

Git LFS objects to be committed:

open .editorconfig: The system cannot find the file specified.
```

The [git lfs track manual](https://github.com/git-lfs/git-lfs/blob/master/docs/man/git-lfs-track.1.ronn) does not seem to mention any requirement of a working tree, but it apparently, does require it:

```
test.git (BARE:master) $ git lfs track
This operation must be run in a work tree.
```

Could anyone clarify usability of `git lfs status` and `git lfs track` for bare repositories?

bk2204:
Yes, both of these commands require a working tree.  `git lfs status` is meant to mirror `git status`, which requires a working tree.  Both commands show the differences between the working tree and the index, which necessarily requires that the working tree exist.  `git lfs track` modifies `.gitattributes`, which also necessarily requires a working tree.

ttaylorr:
> we should probably error out in the same way that 'git status' does on a bare repository, instead of failing to `os.Open()` a file which does not exist.

I have introduced a series to do just this in 
https://github.com/git-lfs/git-lfs/pull/3378.

mloskot:
@bk2204 Excellent explanation. Thank you!

PR #3378, opened 2018-11-13
commands/command_status.go: require a working copy

ttaylorr:
'git lfs status' is functional in a non-work tree repository only some
of the time. Particularly, when there is nothing to report (and
therefore no files to open), 'git lfs status' work as expected.

However, when there are files changed in the index, Git LFS tries to
open them to see what their hash was before/after the change, and in
particular, if they are/were/remain Git LFS pointers. This feature is
used to power the "Git <sha-1> -> LFS <sha-256>" output of 'git lfs
status'.

When in a bare repository and in the aforementioned scenario, the above
operation will fail to open a file that does not exist, for the
repository is bare and thus does not contain a working tree.

In order to bring about a more consistent experience when using 'git lfs
status', let's match the behavior upstream and exit immediately with the
message:

    $ git lfs status
    This operation must be run in a work tree.

Closes: https://github.com/git-lfs/git-lfs/issues/3373.

##

/cc @git-lfs/core 
/cc @mloskot https://github.com/git-lfs/git-lfs/issues/3373

### DIFF
--- a/commands/command_status.go
+++ b/commands/command_status.go
@@ -22,6 +22,7 @@ var (
 
 func statusCommand(cmd *cobra.Command, args []string) {
 	requireInRepo()
+	requireWorkingCopy()
 
 	// tolerate errors getting ref so this works before first commit
 	ref, _ := git.CurrentRef()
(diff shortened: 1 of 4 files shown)


mloskot:
This is very helpful. Thanks!

rosyidreza:
What you about!β£

PR #3379, opened 2018-11-13
git: close blob objects when finished

bk2204:
When we read a tree, commit, or tag from gitobj, the object is automatically closed for us, since these objects are relatively small and can be read and parsed all in memory.  However, blob objects may be large, so we return a reader which can be used to read the data from the object.

In order to properly clean up after ourselves, including closing any object files which may open, we need to call Close on the blob after reading it.  Otherwise, we'll eventually run out of file descriptors, causing us to silently hang.  Restore the reset function that was removed in e3fcde74 ("git: replace object scanner with one based on gitobj", 2018-09-05) and ensure we close objects and reset the object scanner between every call to Scan.

### DIFF
--- a/git/object_scanner.go
+++ b/git/object_scanner.go
@@ -20,6 +20,8 @@ type object struct {
 	Size int64
 	// Type is the type of the object being held.
 	Type string
+	// object is the gitobj object being handled.
+	object gitobj.Object
 }
 
 // ObjectScanner is a scanner type that scans for Git objects reference-able in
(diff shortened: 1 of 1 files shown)


git/object_scanner.go
```diff
@@ -116,6 +125,18 @@ func (s *ObjectScanner) Type() string {
 // operation.
 func (s *ObjectScanner) Err() error { return s.err }
 
+func (s *ObjectScanner) reset() error {
+	if s.object != nil && s.object.Type == "blob" {
```
ttaylorr: I might advocate that we check instead whether or not this is an object that can receive `Close()`, an if so call that. Perhaps like:

```go
func (s *ObjectScanner) reset() error {
        if c, ok := s.object.(interface{
                Close() error
        }); ok && c != nil {
                if err := c.Close(); err != nil {
                        return err
                }
        }
        return nil
}
```

bk2204: Okay, that seems like a nice improvement.  I'll go ahead and make that change.

PR #3387, opened 2018-11-20
Document default value of migrate info --top=<n>

mloskot:
It took me a couple of attempts to understand why the output does not list all I request via  `--include=` patterns :-) I hope this will help others.


bk2204:
Can you say a few words about how you found this out?  I'm looking at the code and not seeing a default; it looks like it should be showing all the items if no `--top` argument is given.  I think it would be a bug if we were limiting them by default.

mloskot:
I greped through the source tree for whole-word `top` and fished out this:

https://github.com/git-lfs/git-lfs/blob/45846c2f0520d95d7300becfed7f650363c89dee/commands/command_migrate.go#L362-L363

PR #3394, opened 2018-11-27
docs/man: suggest using Git configuration for LFS keys

bk2204:
The documentation suggests using .lfsconfig in the repository root for configuration keys.  Only a small number of keys can be set in this file for security reasons, but we don't mention this until the end of the file, which means that users often don't notice this fact, leading to confusion when it doesn't work.

Instead, suggest primarily using the Git configuration files for storing this data, since that will always work.  State that users can use .lfsconfig, but tell them up front that it is restricted to a small number of keys, and point them farther down the page where they can find that information.  Indicate to users that this restriction is for security reasons to help them better understand the rationale.

### DIFF
--- a/docs/man/git-lfs-config.5.ronn
+++ b/docs/man/git-lfs-config.5.ronn
@@ -3,12 +3,18 @@ git-lfs-config(5) -- Configuration options for git-lfs
 
 ## CONFIGURATION FILES
 
-git-lfs reads its configuration from a file called `.lfsconfig` at the root of
-the repository. The `.lfsconfig` file uses the same format as `.gitconfig`.
+git-lfs reads its configuration from any file supported by `git config -l`,
+including all per-repository, per-user, and per-system Git configuration files.
 
-Additionally, all settings can be overridden by values returned by `git config -l`.
-This allows you to override settings like `lfs.url` in your local environment
-without having to modify the `.lfsconfig` file.
+Additionally, a small number of settings can be specified in a file called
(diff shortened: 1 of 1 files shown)


docs/man/git-lfs-config.5.ronn
```diff
@@ -3,12 +3,16 @@ git-lfs-config(5) -- Configuration options for git-lfs
...
 
-git-lfs reads its configuration from a file called `.lfsconfig` at the root of
-the repository. The `.lfsconfig` file uses the same format as `.gitconfig`.
+git-lfs reads its configuration from any file supported by `git config -l`,
+including all per-repository, per-user, and per-system Git configuration files.
 
-Additionally, all settings can be overridden by values returned by `git config -l`.
-This allows you to override settings like `lfs.url` in your local environment
-without having to modify the `.lfsconfig` file.
+Additionally, a small number of settings can be specified in a file called
```
PastelMobileSuit: I think with this change it's no longer clear what `.lfsconfig` should be used for, as we're now suggesting users primarily use normal `.gitconfig` files and settings in those files will explicitly override the ones in `.lfsconfig`.

Maybe we could add a note about what the possible use cases for `.lfsconfig` are?

bk2204: Sure, that sounds good.

bk2204: Okay, I've updated this text to provide a description of why users would want to use the `.lfsconfig` file.  Let me know if there's another use case I should mention or something I've left out.

Issue #3386, opened 2018-11-20
Pushing to fork: "error: failed to push some refs to"


sandertan:
I'm experiencing an issue with pushing a branch to a fork. When I push to the source repository I can push without issues. But when pushing the same branch to a fork, this results in: 
```
...... : no such file or directory 
error: failed to push some refs to 'https.........'
```

This issue can be reproduced by running the following:
```
# Clone the source repository
git clone https://github.com/cBioPortal/datahub.git
cd datahub

# Create your own fork of this repo and add it as remote
git remote add sandertan https://github.com/sandertan/datahub.git

# Create test branch and files
git checkout -b 'master_test'
touch test_file.txt
git add test_file.txt
git commit -m 'Add test file'

# Push this branch your own fork
git push sandertan master_test
```
This results in:
```
$ git push sandertan master_test
open /Users/sander/Data/datahub/public/prad_su2c_2015/data_mutations_misc.txt: no such file or directory                                                                                                                                                                      
error: failed to push some refs to 'https://github.com/sandertan/datahub.git'
```

Is there something wrong with how these files were added to git lfs on the source repository, or is there an issue with my git-lfs configuration?

```
$ git lfs env
git-lfs/2.6.0 (GitHub; darwin amd64; go 1.11.2)
git version 2.16.1

Endpoint=https://github.com/cBioPortal/datahub.git/info/lfs (auth=none)
Endpoint (sandertan)=https://github.com/sandertan/datahub.git/info/lfs (auth=basic)
LocalWorkingDir=/Users/sander/Data/datahub
LocalGitDir=/Users/sander/Data/datahub/.git
LocalGitStorageDir=/Users/sander/Data/datahub/.git
LocalMediaDir=/Users/sander/Data/datahub/.git/lfs/objects
LocalReferenceDirs=
TempDir=/Users/sander/Data/datahub/.git/lfs/tmp
ConcurrentTransfers=3
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
FetchRecentRefsIncludeRemotes=true
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/Users/sander/Data/datahub/.git/lfs
AccessDownload=none
AccessUpload=none
DownloadTransfers=basic
UploadTransfers=basic
[...]

PR #3398, opened 2018-11-28
Improve error message on missing object

### LINKED ISSUE #1113: git lfs push --all fails with "open /FileRemovedInEarlierCommit.png: no such file or directory"
When the server is missing files (due to a bug or migrating to new server or probably other reasons), it's my understanding that `git lfs push --all` will fix it. Problem is my attempts to run that result in `open /Users/andyvn22/Desktop/lfs-testing/LFS Testing/IrisApplication.png: no such file or directory`, where `IrisApplication.png` is an old file I deleted several commits ago, and the missing files are not uploaded to the server. Shouldn't the rest of the command complete, and why is it searching for that no-longer-existent file anyway?

bk2204:
When attempting a push, we need to ensure that we have every object we want to push.  The object scanner produces pointer objects that contain the object ID of the object we want to push plus the path at which the scanner found the object.  Before pushing, we iterate over every object we want to push and if the object is not found in the LFS object store, we attempt to read the file that is at the path in question to try to create the object.

If that attempt fails since the file is no longer present (i.e., it was present in a previous revision but has since been deleted), we produce an error message that the file could not be opened without any context, causing the user to be confused as to why we would have attempted to open a file that has since been deleted.  Improve this error message by indicating what the actual problem is (that we cannot find the source file), a possible step that they can take to solve it (running git lfs fetch --all), and the error message when attempting to open the file.

This fixes #1113 and #3386.

### DIFF
--- a/commands/uploader.go
+++ b/commands/uploader.go
@@ -347,7 +347,7 @@ func (c *uploadContext) uploadTransfer(p *lfs.WrappedPointer) (*tq.Transfer, err
 	}
 
 	if len(filename) > 0 {
-		if err = c.ensureFile(filename, localMediaPath); err != nil && !errors.IsCleanPointerError(err) {
+		if err = c.ensureFile(filename, localMediaPath, oid); err != nil && !errors.IsCleanPointerError(err) {
 			return nil, err
 		}
 	}
@@ -362,7 +362,7 @@ func (c *uploadContext) uploadTransfer(p *lfs.WrappedPointer) (*tq.Transfer, err
 
 // ensureFile makes sure that the cleanPath exists before pushing it.  If it
 // does not exist, it attempts to clean it by reading the file at smudgePath.
-func (c *uploadContext) ensureFile(smudgePath, cleanPath string) error {
+func (c *uploadContext) ensureFile(smudgePath, cleanPath, oid string) error {
 	if _, err := os.Stat(cleanPath); err == nil {
 		return nil
 	}
@@ -373,7 +373,7 @@ func (c *uploadContext) ensureFile(smudgePath, cleanPath string) error {
 		if c.allowMissing {
 			return nil
 		}
-		return err
+		return errors.Wrapf(err, "Unable to find source for object %v (try running git lfs fetch --all)", oid)
 	}
 
 	defer file.Close()


commands/uploader.go
```diff
@@ -373,7 +373,7 @@ func (c *uploadContext) ensureFile(smudgePath, cleanPath string) error {
 		if c.allowMissing {
 			return nil
 		}
-		return err
+		return errors.Wrapf(err, "Unable to find source for object %v (try running git lfs fetch --all)", oid)
```
chrisd8088: Just out of interest, @bk2204, as I haven't looked at the `errors` package before, does the `%v` output the error string (from `err`) or the `oid` string?  I was looking at [Formatted printing of errors](https://godoc.org/github.com/pkg/errors#hdr-Formatted_printing_of_errors) and trying to understand what it does.

chrisd8088: Oh, wait, I see -- different `errors` package; this one is local.  Got it!  :-)

bk2204: The `%v` is for `oid`.  What `errors.Wrapf(err, fmt, args...)` does is essentially `fmt.Sprintf("%s: %s", fmt.Sprintf(fmt, args...), err)`.  It formats the string and args you give it and then tacks on the message from `err` at the end.

bk2204: The output will look something like it does here: https://github.com/git-lfs/git-lfs/issues/1113#issuecomment-442245429

chrisd8088: Yeah, indeed, thanks -- I was just looking at the wrong package, having assumed incorrectly it wasn't locally defined.  This looks great!

bk2204: Yeah, it's confusing because there's a built-in `errors` package as well, which has none of these niceties.

zachwhaley:
Hello,

Was there a new bug opened to address the underlying bug here?

We are hitting this issue fairly frequently, and doing a complete or even piece meal `git lfs fetch` is not really a fix.

Thanks!

bk2204:
I don't believe that there actually is an underlying bug to address. The local side is missing objects needed to push to the remote side, and the error message reflects that. There isn't really anything that Git LFS can do to provide those objects if it doesn't have them. Fetching would be the only way to acquire those objects for push.

zachwhaley:
I believe there is, or at least an behavior that does not seem intuitive.

Per the comment on https://github.com/git-lfs/git-lfs/issues/1113#issuecomment-366791563

> I think the core issue is that LFS is trying to read from the disk (working tree) when looking at a given hash. There's no guarantee that the observed hash is currently checked out. Git-lfs should query the index or relevant tree object for the content of the file.

There seems to be an issue where git-lfs is trying to find a large file on disk that was never checked out, mostly likely because it was removed prior to a new fetch.

The scenario works like this I think, or at least this is what we are seeing:

1. I have a clone of a repository.
2. A large file is added to the repository in one commit and then removed in another commit.
3. I then make a fetch of the repository and rightly so the deleted object is not downloaded because it is not part of my current checkout
4. I then make a commit and attempt to push
5. git-lfs makes an attempt to upload the large file that had been deleted because it sees it in my commit history (I guess?) and fails because it is not on disk.

So in my head, this seems like a weird behavior; why would git-lfs try to upload a large file that is not a part of my current checkout?

bk2204:
> So in my head, this seems like a weird behavior; why would git-lfs try to upload a large file that is not a part of my current checkout?

Git LFS downloads files on demand. However, when pushing, we need to be sure that we have all of the files involved in the entire history of the push uploaded. Otherwise, somebody could download your clone, check out an old commit where that large file exists, and then be unable to download that large file.

The server will tell us which files it has and which ones we need to upload, so if the remote you're pushing to already has those files, then they need not be uploaded again. The server may or may not consider forks of the same repository to share files, and if it does, you wouldn't need to upload the files it already has.

zachwhaley:
Okay, that makes sense.  I get the feeling we may be seeing this issue because of forked repos.

Thanks!

PR #3401, opened 2018-11-29
locking: don't modify permissions of ignored files

### LINKED ISSUE #3390: git lfs sets ignored files to read-only when file type is marked as lockable
### Setup:
```
git version
git version 2.19.2.windows.1

git lfs version
git-lfs/2.6.0 (GitHub; windows amd64; go 1.11.1; git 93702d75)
```

- Git LFS is used for locking some binary files with the extention _.vw7_, so `.gitattributes` is set to `*.vw7 lockable`
- The project contains several .vw7 files which we lock when we need to change them and unlock after push
- In the project folder there is a folder named __output_ which is generated in the build process, it also contains _.vw7_ files
- This folder is ignored in git, `.gitignore` contains `_output`

### Problem:
For example when I run a `git checkout` of any file in the repository, all the _.vw7_ files in the __output_ folder are marked as read-only.
This causes any following build process to fail, as the build can't modify the _.vw7_ files in the __output_ folder. The files have to be manually set to writable to make it work again.

### Expected behaviour:
Files marked as lockable that are in an ignored folder should not be set to read-only

bk2204:
When a lockable pattern in a gitattributes file matched an ignored file, we would set the permissions of that file on checkout, despite the fact that the file was ignored.  This could lead to problems for the user. Ensure that we only traverse objects tracked by Git when modifying the permissions of files.

This change fixes #3390.

### DIFF
--- a/locking/lockable.go
+++ b/locking/lockable.go
@@ -122,7 +122,7 @@ func (c *Client) fixFileWriteFlags(absPath, workingDir string, lockable, unlocka
 		errs = append(errs, err)
 	}
 
-	tools.FastWalkGitRepoAll(absPath, func(parentDir string, fi os.FileInfo, err error) {
+	tools.FastWalkGitRepo(absPath, func(parentDir string, fi os.FileInfo, err error) {
 		if err != nil {
 			addErr(err)
 			return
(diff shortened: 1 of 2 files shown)


bk2204:
The tests are failing because this behavior is by design, and consequently, I'm closing this PR.

Issue #3183, opened 2018-08-14
Lockable files left writable upon checkout when matching any filter in .gitignore, even with exception present


zach-r-d:
When a lockable file is checked out, it is not set to readonly if it matches any filter in .gitignore, even if there is an exception for it in .gitignore.

I've included an example below. Please let me know if there is any other information that would be helpful for me to provide. Thanks!

Example
-------

Here, we have a simple repository with LFS (at the latest version) enabled for `*.bin` files, with all three files (`.gitattributes`, `README.md`, `test.bin`) committed on master and a clean working directory with no locks held. `test.bin` is read-only as expected.
```
$ git lfs version
git-lfs/2.5.1 (GitHub; linux amd64; go 1.10.3)

$ ls -la
total 24
drwxr-xr-x 3 zach domain^users 4096 Aug 14 15:37 .
drwxr-xr-x 8 zach domain^users 4096 Aug 14 15:00 ..
drwxr-xr-x 9 zach domain^users 4096 Aug 14 15:37 .git
-rw-r--r-- 1 zach domain^users   51 Aug 14 15:00 .gitattributes
-rw-r--r-- 1 zach domain^users   27 Aug 14 15:00 README.md
-r--r--r-- 1 zach domain^users    3 Aug 14 15:37 test.bin

$ cat .gitattributes 
*.bin filter=lfs diff=lfs merge=lfs -text lockable

$ git lfs locks --local

$ git status
On branch master
Your branch is up-to-date with 'origin/master'.

nothing to commit, working directory clean
```

Removing and re-checkout-ing `test.bin` leaves it read-only, as expected.
```
$ rm test.bin && git checkout test.bin
rm: remove write-protected regular file ‘test.bin’? y

$ ls -la
total 24
drwxr-xr-x 3 zach domain^users 4096 Aug 14 15:38 .
drwxr-xr-x 8 zach domain^users 4096 Aug 14 15:00 ..
drwxr-xr-x 9 zach domain^users 4096 Aug 14 15:38 .git
-rw-r--r-- 1 zach domain^users   51 Aug 14 15:00 .gitattributes
-rw-r--r-- 1 zach domain^users   27 Aug 14 15:00 README.md
-r--r--r-- 1 zach domain^users    3 Aug 14 15:38 test.bin
```

However, adding a pattern to `.gitignore` that matches `test.bin`, then removing and re-checkout-ing `test.bin` causes it to be left writable.
```
$ echo '*t.bin' > .gitignore

[...]

Issue #434, opened 2015-06-23
Can't push to github (pre-receive hook declined)


rvi:
Hi,

This bug may be related to #333 

I just tried `git lfs`, and it doesn't worked for me...
What I did : 
- Install lfs (0.5.2)
- git lfs track myFramework.framework
- git add 
- git commit
- git push 

Here is the error : 

```logs
Counting objects: 20, done.
Delta compression using up to 4 threads.
Compressing objects: 100% (20/20), done.
Writing objects: 100% (20/20), 52.06 MiB | 7.52 MiB/s, done.
Total 20 (delta 5), reused 0 (delta 0)
remote: error: GH001: Large files detected.
remote: error: Trace: f22735321d76092fa67a1011835dfdcc
remote: error: See http://git.io/iEPt8g for more information.
remote: error: File mySDK.framework/mySDK is 185.09 MB; this exceeds GitHub's file size limit of 100.00 MB
To git@github.com:company/repo.git
 ! [remote rejected] master -> master (pre-receive hook declined)
error: failed to push some refs to 'git@github.com:company/repo.git'
```

```
$ git lfs version
git-lfs/0.5.2 (GitHub; darwin amd64; go 1.3)
```

technoweenie:
That's unrelated. Your file "mySDK.framework/mySDK" is stored in Git, not Git LFS, and exceeds GitHub's file size limit.

rvi:
So how can I store it in Git LFS ?

technoweenie:
`git lfs track "*.ext"` will add an entry to your `.gitattributes` to store "ext" files. From then on, any time an "*.exe" file is added to Git, it will be stored in Git LFS. Existing commits do not change though. If you have older commits trying to add the file to Git, you will have to squash those commits before pushing.

rvi:
Ok thanks, I understood what went wrong. I did that : `git lfs track mySDK.framework`
It worked well (at least it shows `Tracking mySDK.framework`...)

But I just tried to do `git lfs track **/mySDK`

And now on the git push I've got this : 
```
(0 of 1 files) 0 B 2s
Git LFS support is coming! Sign up for early access: https://github.com/early_access/git-lfs
Docs: https://github.com/contact
error: failed to push some refs to 'git@github.com:company/repo.git' 
```

Is it possible for you to bypass the waiting queue for my account ? :D Would be awesome !! :)

__A nice feature would be to know that tracking is not working for a directory.__

Thanks for your help !!

[...]

PR #3409, opened 2018-12-04
locking: add flag to control modification of ignored files

### LINKED ISSUE #3390: git lfs sets ignored files to read-only when file type is marked as lockable

bk2204:
In a727fea2 ("locking: remove write permission for ignored files", 2018-08-20), we learned to set and clear the read-only attribute on ignored files that match the lockable pattern, since "git lfs lock" operates on pathspecs, not actual files.

However, in doing so, we caused a regression for people who have ignored files (such as build files) which match the lockable pattern, but should not have their permissions modified.  Since there is no easy way for us to know if the user would like their files modified, add a config option to control this behavior, and leave it at the historical (i.e., pre-a727fea2) default, off.  Add documentation accordingly.

Choosing the historical behavior is compatible with the most existing use cases and preserves the behavior people expect with Git, which is that it does not touch ignored files.

Fixes #3390.
/cc @ttaylorr 
/cc @zach-r-d #3183
/cc @Kitzibua #3390

### DIFF
--- a/t/t-lock.sh
+++ b/t/t-lock.sh
@@ -250,6 +250,32 @@ begin_test "lock with .gitignore"
   git add .gitignore
   git commit -m ".gitignore: ignore 'a.txt'"
   rm -f a.txt && git checkout a.txt
+  assert_file_writeable a.txt
+)
+end_test
+
+
+begin_test "lock with .gitignore and lfs.lockignoredfiles"
+(
+  set -e
+
+  reponame="lock-with-gitignore-and-ignoredfiles"
+  setup_remote_repo_with_file "$reponame" "a.txt"
+  clone_repo "$reponame" "$reponame"
+
+  git config lfs.lockignoredfiles true
+  echo "*.txt filter=lfs diff=lfs merge=lfs -text lockable" > .gitattributes
+
+  git add .gitattributes
+  git commit -m ".gitattributes: mark 'a.txt' as lockable"
+
+  rm -f a.txt && git checkout a.txt
+  refute_file_writeable a.txt
+
+  echo "*.txt" > .gitignore
+  git add .gitignore
+  git commit -m ".gitignore: ignore 'a.txt'"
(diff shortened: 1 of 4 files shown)


t/t-lock.sh
```diff
@@ -249,7 +249,33 @@ begin_test "lock with .gitignore"
...
+  git commit -m ".gitattributes: mark 'a.txt' as lockable"
+
   rm -f a.txt && git checkout a.txt
   refute_file_writeable a.txt
+
+  echo "*.txt" > .gitignore
+  git add .gitignore
+  git commit -m ".gitignore: ignore 'a.txt'"
+  rm -f a.txt && git checkout a.txt
+  assert_file_writeable a.txt
```
PastelMobileSuit: If I'm reading this option correctly, shouldn't `a.txt` be locked due to the `lfs.lockignorefiles=true`?

bk2204: Thanks for catching this. I typoed the config option a couple lines earlier.

ali562:
#434

Issue #3329, opened 2018-10-18
URL wildcards not honored (e.g. for proxy settings)


cburger-scheidlin:
Git allows using wildcards for the host part of the URL that is used to match settings, see https://github.com/git/git/commit/a272b9e70a48a355b6dd7ff0179c11f8da7ef0f3

Git LFS does not follow these matching rules, leading to an inconsistent experience between git and git lfs.
If a proxy is configured and then unset for local domains (e.g. in an enterprise setting), the user will experience this as the git clone working but the lfs fetch failing.

If I understand the code correctly, lfshttp.go would be the place to implement this feature, specifically https://github.com/git-lfs/git-lfs/blob/6a0598985d950b9f2ce081ba91bf0ed6131a087d/lfshttp/proxy.go#L72
However, I have never programmed in go, so I dare not attempt adding it.

PR #3414, opened 2018-12-06
Allow users to specify the credential helper in .lfsconfig

keiko713:
Hello 👋 I'd like to add a credential helper setting in `.lfsconfig` so that the Git LFS will respect it and setup to local git config for me. I'm trying to see if something like this even makes sense or can be considered.

I'm trying to use Git LFS with several different servers across several repositories (the server is actually only one, but the hosts will be different). For example, the repo A will talk to the LFS api/server a.mylfs.com, the repo B will talk b.mylfs.com. These servers have the basic authentication enabled.
I basically want to have some config in the repo (.lfsconfig is good for that) so that only one person need to specify the helper setup and the other contributors will use that helper to talk to the LFS API.

I thought about several options but each has cons:

* netrc way will result lots of entries and potential dirty data (need to add for each a.mylfs.com, b.mylfs.com, ...)
* cred helper + global git config will result lots of entires and potential dirty data too
* cred helper + local git config will require the git config setup for reach repo
* all of above need to know what the LFS server host is beforehand for folks who try to clone the repo or pull from the repo that started using LFS, and clone/pull experience will not be good

I'm also happy to hear any other options or any better way to solve this problem. Thanks!

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -118,11 +118,16 @@ func (c *Configuration) getMask() int {
 }
 
 func (c *Configuration) readGitConfig(gitconfigs ...*git.ConfigurationSource) Environment {
-	gf, extensions, uniqRemotes := readGitConfig(gitconfigs...)
+	gf, extensions, uniqRemotes, credHelperKey := readGitConfig(gitconfigs...)
 	c.extensions = extensions
 	c.remotes = make([]string, 0, len(uniqRemotes))
 	for remote := range uniqRemotes {
 		c.remotes = append(c.remotes, remote)
+	}
+	if credHelperKey != "" {
+		if credHelperValue, ok := gf.Get(credHelperKey); ok {
+			c.gitConfig.SetLocal(credHelperKey, credHelperValue)
+		}
 	}
 
 	return EnvironmentOf(gf)
--- a/config/git_fetcher.go
+++ b/config/git_fetcher.go
@@ -15,7 +15,7 @@ type GitFetcher struct {
 	vals map[string][]string
 }
 
-func readGitConfig(configs ...*git.ConfigurationSource) (gf *GitFetcher, extensions map[string]Extension, uniqRemotes map[string]bool) {
+func readGitConfig(configs ...*git.ConfigurationSource) (gf *GitFetcher, extensions map[string]Extension, uniqRemotes map[string]bool, credHelperKey string) {
 	vals := make(map[string][]string)
 	ignored := make([]string, 0)
 
@@ -86,6 +86,9 @@ func readGitConfig(configs ...*git.ConfigurationSource) (gf *GitFetcher, extensi
 				uniqRemotes[remote] = remote == "origin"
 			} else if len(parts) > 2 && parts[len(parts)-1] == "access" {
 				allowed = true
+			} else if len(parts) > 2 && parts[0] == "credential" && parts[len(parts)-1] == "helper" {
+				allowed = true
+				credHelperKey = key
 			}
 
 			if !allowed && keyIsUnsafe(key) {
--- a/docs/man/git-lfs-config.5.ronn
+++ b/docs/man/git-lfs-config.5.ronn
@@ -83,6 +83,11 @@ be scoped inside the configuration for a remote.
 
   Default: `lfs` in Git repository directory (usually `.git/lfs`).
 
+* `credential.https://<host>.helper`
+
+  Sets the Git Credential helper for the host. This can be used to specify the
+  credential helper of the LFS API. This will be copied to the local gitconfig.
+
 ### Transfer (upload / download) settings
 
   These settings control how the upload and download of LFS content occurs.
@@ -345,16 +350,19 @@ The .lfsconfig file in a repository is read and interpreted in the same format
 as the file stored in .git/config. It allows a subset of keys to be used,
 including and limited to:
 
+- lfs.allowincompletepush
 - lfs.fetchexclude
 - lfs.fetchinclude
 - lfs.gitprotocol
+- lfs.locksverify
 - lfs.pushurl
 - lfs.url
 - lfs.extension.{name}.clean
 - lfs.extension.{name}.smudge
 - lfs.extension.{name}.priority
 - remote.{name}.lfsurl
-- remote.{name}.{*}.access
+- lfs.{*}.access
+- credential.{*}.helper
 
 The set of keys allowed in this file is restricted for security reasons.
 


bk2204:
First of all, thanks for sending this pull request and welcome to Git LFS!

I took a look at the code and I have some concerns about the security of providing this option in `.lfsconfig`. Since the person controlling the repo can also specify the URL from which to fetch LFS objects, if they can control the selection of credential helper, they can exploit any vulnerable credential helper on the system simply by having someone clone a URL.

I also have some concerns about whether this is a good choice for repositories that are used by people on different systems. For example, on macOS, I use the `osxkeychain` helper, but on Linux, I use `libsecret`; setting this value in the config might result in a repository that cannot have its LFS objects fetched since the credential helper can't be invoked on that system. 

As of 2.7.0, Git LFS will support wildcards for credential helpers, but unfortunately Git itself doesn't support that yet. It's possible that a patch could be sent to Git to enable that, but that itself wouldn't make it in until Git 2.21 at the earliest.  If you're interested in that, let me know, and I can look into it further.

It is currently possible to have multiple credential helpers enabled if you want.  Git will go through them in order and query them, so you could use, say, the default `osxkeychain` helper on macOS and fall back to another credential helper that understands the wildcard patterns you want to use if `osxkeychain` doesn't contain credentials itself.  If that one doesn't work, then Git will fall back to prompting, like before.  That may be the easiest way to accomplish what you're looking for with the way Git and Git LFS work now.

calavera:
@bk2204 thanks for your concerns. I'm adding some comments below:
 
>  Since the person controlling the repo can also specify the URL from which to fetch LFS objects, if they can control the selection of credential helper, they can exploit any vulnerable credential helper on the system simply by having someone clone a URL.

I understand that this is a possible attack vector. However, I'd argue that exactly the person controlling a repository should be able to specify how to download LFS objects, that includes URL and necessary credentials for that URL, including the helper to use.

It is true that in the following scenario, a malicious owner could steal credentials from a host:

1. The victim has a a helper with a known vulnerability.
2. The victim has set credentials for a given server host and the helper can read them.
3. The attacker controls a repository and can specify URL and helper.
4. The attacker also controls the server host where git is going to send the credentials.

Git won't send credentials to a host that a helper doesn't return credentials for, that's its security model. So this attack also requires some kind of social engineering and having control over the host where git sends the credentials. I think this is a risk that cannot be avoided at the end.

This patch is trying to solve a big usability problem with Git LFS. There is no way to clone a repository and download the LFS objects at the same time with `git clone`, unless you control the repository and the LFS storage. The LFS specification includes [the SSH workaround](https://github.com/git-lfs/git-lfs/blob/df881bf23a08f1b57209825e0f6b2d0b9e6dcd5c/docs/api/server-discovery.md#ssh), which only works if you control both. This is very convenient in that case, and very convenient for Git providers, but there is another case, where your Git and LFS providers are two different entities.

An even more dangerous and worrisome advice, it's what the Authentication documentation includes, which is to hardcode the user and password in the `lfs.url` attribute, that can be pushed to your git repository in plain text. Even if it says that's not recommended, that's the only other possibility if you want git clone to be able to download LFS objects. 

https://github.com/git-lfs/git-lfs/blob/df881bf23a08f1b57209825e0f6b2d0b9e6dcd5c/docs/api/authentication.md#specified-in-url

I'll really appreciate your advice to solve this use case:

1. I host my code in GitHub, ssh url "git@github.com:calavera/lfs-test.git"
2. I host my LFS objects somewhere else, https url configured in .lfsconfig: "https://lfs.calavera.io"
3. I want @keiko713 to be able to download the code and LFS objects with `git clone`.
4. @keiko713 already has a key on her computer to get access to the LFS server.
5. I want my LFS server to check @keiko713's credentials without her having to change her .gitconfig every time she needs to do the same operation with different repositories, or having to type that key in the username/password prompt.

This is exactly the same user experience that LFS provides with the SSH workaround.

bk2204:
> @bk2204 thanks for your concerns. I'm adding some comments below:
> 
> > Since the person controlling the repo can also specify the URL from which to fetch LFS objects, if they can control the selection of credential helper, they can exploit any vulnerable credential helper on the system simply by having someone clone a URL.
> 
> I understand that this is a possible attack vector. However, I'd argue that exactly the person controlling a repository should be able to specify how to download LFS objects, that includes URL and necessary credentials for that URL, including the helper to use.

I think perhaps I miscommunicated my concerns about the vulnerability issue.  My scenario looks like this:

1. I learn (either myself, or through a security advisory) that the `xyz` credential helper, which is commonly available, has a buffer overflow.
1. I create a new repo, `https://github.com/bk2204/exploit-me`.
1. I set `lfs.url` to point to a very long string containing shellcode and the `credential.helper` value to `xyz`.
1. I wait for people to clone my repository and run my shellcode by calling the credential helper.

I want to stress that `.lfsconfig` is very security sensitive. We generally don't allow anything that looks like it might execute an unexpected program on the user's system, because that opens up a potential security problem.  Upstream Git doesn't allow configuration in the repo at all, because they think it's too big of a security risk.

I'm also concerned about the fact that it copies itself into the user's Git config.  I don't think we have any precedent for that occurring, and I don't think we should allow any remote repository to override whatever the user has set.

I'm not worried about stealing credentials for the reasons you specify. I'm worried about people who are still using wildly out-of-date and insecure software and the ability to specify code to execute on those systems.

I think the situation you're looking at is a corporate environment, but we also have to consider the situation where the repository is an open source repository from an unknown party, who should probably not be able to specify which credential helper to use. 

> I'll really appreciate your advice to solve this use case:
> 
>     1. I host my code in GitHub, ssh url "[git@github.com](mailto:git@github.com):calavera/lfs-test.git"
> 
>     2. I host my LFS objects somewhere else, https url configured in .lfsconfig: "https://lfs.calavera.io"
> 
>     3. I want @keiko713 to be able to download the code and LFS objects with `git clone`.
> 
>     4. @keiko713 already has a key on her computer to get access to the LFS server.
> 
>     5. I want my LFS server to check @keiko713's credentials without her having to change her .gitconfig every time she needs to do the same operation with different repositories, or having to type that key in the username/password prompt.

I think there may be something here I'm missing. You say that the user has credentials already. I assume that they're in a credential helper. What is preventing that credential helper from being run all the time (e.g. `credential.helper=osxkeychain`)? If those credentials are in a special credential helper, why can't that be added as an additional credential helper that runs as a fallback? Is it just that the additional helper is not set up in a normal dev environment?

I would like to learn more about the scenario you're trying to solve with the additional credential helper, since I don't think I understand it very well.  I think I'd be able to discuss better how to address this use case once I understand it better.

keiko713:
Thank you so much for taking time to look into this!
The security concern was something definitely I was really skeptical of, especially when I was writing the code of "adding local git config" part. It's good to know that you have the same feeling and making a change around here is not really an option at the moment.

> I think there may be something here I'm missing. You say that the user has credentials already. I assume that they're in a credential helper. What is preventing that credential helper from being run all the time (e.g. credential.helper=osxkeychain)?

Yes, it's about having the specific credential helper (e.g. myhelper).
Interesting, I didn't think about this. I think this is possible and good enough for now. The concern (or reason I didn't think about it) would be that we know this credential helper is very specific to the certain host (`*.mylfs.com`) and the certain usage (only LFS servers, not Git host) and I'd wish that nothing else would try to use. However, we can certainly improve the helper to support running all the time (aka have proper error handling).

> As of 2.7.0, Git LFS will support wildcards for credential helpers, but unfortunately Git itself doesn't support that yet. It's possible that a patch could be sent to Git to enable that, but that itself wouldn't make it in until Git 2.21 at the earliest. If you're interested in that, let me know, and I can look into it further.

Is this about https://github.com/git-lfs/git-lfs/issues/3329 ? I was initially trying to handle this problem using `http.<hostname>.extraheader` in the git config. With this, I can do something like `http.https://*.mylfs.com.extraheader = Authorization: XXXX` in the global git config, to support specifying how to authenticate with multiple hosts. This way, I can just let the user run one setup process of adding this in the global git config, and people can pull/clone any Git LFS repos of LFS servers in *.mylfs.com as long as that auth token proves that it has an access to it.
The downside of this would be that it really does look like a "workaround" since it's not using a credential helper which Git provides for the specific purpose like this (and Git LFS supports it). Also this requires to use Git LFS 2.7.0 which is not yet released.

Given this, I'd say that adding "myhelper" as a global credential helper would be the best solution for my issue. Thanks a lot for walking this through with me.
I'm still interested in opening a PR with only doc change that I made, as I believe that the doc is a bit outdated. Is that okay?

keiko713:
Thanks! I opened a new PR #3423. I'm going to close this.

PR #3424, opened 2018-12-11
docs/man/git-lfs-migrate: make examples less confusing

### LINKED ISSUE #3421: git lfs migrate --no-rewrite couldn't see files in commit?
Say I have following branches:

- master
- update (`git checkout -b update` from master)

And if I run following command on it:

```
$ git lfs migrate import --no-rewrite LightingData.asset
```

It gives me this warning:

```
fatal: could not rewrite "LightingData.asset": unable to find entry LightingData.asset in tree
```

But if I run following command:

```
$ git lfs migrate import --include-ref=update --include="LightingData.asset"
```

Then it passes just fine:

```
migrate: Rewriting commits: 100% (4/4), done                                    
  convert	7b208bf9ddfc5c814e50b5c649fa99f41cef1b2f -> 14da0c07bd6a215b8fd71335579b492b7d7327ee
  master 	7b208bf9ddfc5c814e50b5c649fa99f41cef1b2f -> 14da0c07bd6a215b8fd71335579b492b7d7327ee

$ git lfs ls-files | grep asset

9af25f1c37 - leocities-starter-kit/Assets/Leocities/Scenes/TestSceneA/LightingData.asset
190a5dd172 - leocities-starter-kit/Assets/Leocities/Scenes/TestSceneB/LightingData.asset
108a9f53fd - leocities-starter-kit/Assets/Leocities/Scenes/TestSceneC/LightingData.asset
```

My .gitattributes contains these 2 lines:

```
LightingData.asset filter=lfs diff=lfs merge=lfs -text
...
*.asset merge=unityyamlmerge eol=lf
```

So my question, is this a bug? Is `--no-rewrite` supposed to do what I think it would do: [that is, to migrate the files into LFS without rewriting history](https://github.com/git-lfs/git-lfs/pull/3029)?

bk2204:
When invoking `git lfs migrate --no-rewrite`, we take a list of files, not a list of patterns. This differs from other situations, and the examples in the man page don't make this clear, resulting in confusion on the part of users. Improve the example by making the example commit message explain the situation a bit better.

Fixes #3421.

/cc @bitinn

### DIFF
--- a/docs/man/git-lfs-migrate.1.ronn
+++ b/docs/man/git-lfs-migrate.1.ronn
@@ -260,7 +260,8 @@ Note: This will require a force push to any existing Git remotes.
 ### Migrate without rewriting local history
 
 You can also migrate files without modifying the existing history of your
-repository:
+repository. Note that in the examples below, files in subdirectories are not
+included because they are not explicitly specified.
 
 Without a specified commit message:
 
@@ -271,8 +272,9 @@ $ git lfs migrate import --no-rewrite test.zip *.mp3 *.psd
 With a specified commit message:
 
 ```
-$ git lfs migrate import --no-rewrite -m "Import .zip, .mp3, .psd files" \
-  test.zip *.mpd *.psd
+$ git lfs migrate import --no-rewrite \
+  -m "Import test.zip, .mp3, .psd files in root of repo" \
+  test.zip *.mp3 *.psd
 ```
 
 ## SEE ALSO


bitinn:
I really think you should change the first example too, it says nothing about only working for root folder. It's not even standard practice for git to pattern match this way.

> Without a specified commit message:
>
> `$ git lfs migrate import --no-rewrite test.zip *.mp3 *.psd`
>

Also, this explanation is misleading because the `gitattributes` pattern matching is recursive by default, and `--no-rewrite` even explicitly checks for `gitattributes` entries, in order to ensure your file pattern matches an entry in `gitattributes`.

> The `--no-rewrite` sub-mode supports the following options and arguments:
>
> - `-m <message> --message=<message>` Specifies a commit message for the newly created commit.
> - [file ...] The list of files to import. These files must be tracked by patterns specified in the gitattributes.

bk2204:
I've pushed a new commit which expands on the previous one and should hopefully make this less confusing.

Note that the patterns aren't expanded by Git or Git LFS, since they're expanded by the shell instead. When invoking `git lfs track`, the patterns have to be quoted to protect them from the shell if they contain wildcards, and trying to do that here would, as you found out, not work. I agree that wasn't immediately obvious from the example, though, and it's easy to overlook.

docs/man/git-lfs-migrate.1.ronn
```diff
@@ -271,7 +272,8 @@ $ git lfs migrate import --no-rewrite test.zip *.mp3 *.psd
 With a specified commit message:
 
 ```
-$ git lfs migrate import --no-rewrite -m "Import .zip, .mp3, .psd files" \
+$ git lfs migrate import --no-rewrite \
+  -m "Import .zip, .mp3, .psd files in root of repo" \
```
frantic1048: Just getting confused with man book(`git-lfs/2.6.1 (GitHub; linux amd64; go 1.11.3)`) and found this PR.

According to the actual arguments(`test.zip *.mpd *.psd`) passed to git-lfs-migrate, to be more accurate, this commit message should be:

```diff
--  -m "Import .zip, .mp3, .psd files in root of repo" \
++  -m "Import test.zip, .mpd, .psd files in root of repo" \
```

And the [previous example](https://github.com/git-lfs/git-lfs/pull/3424/files#diff-4e71818f2c658bd6d213d1be9012df10R269) is talking about `test.zip *.mp3 *.psd`, I think this example should be:

```diff
 $ git lfs migrate import --no-rewrite \
--  -m "Import .zip, .mp3, .psd files in root of repo" \
++  -m "Import test.zip, .mp3, .psd files in root of repo" \
--   test.zip *.mpd *.psd
++   test.zip *.mp3 *.psd
```

bk2204: I've gone ahead and made that change.

docs/man/git-lfs-migrate.1.ronn
```diff
@@ -271,7 +272,8 @@ $ git lfs migrate import --no-rewrite test.zip *.mp3 *.psd
 With a specified commit message:
 
 ```
-$ git lfs migrate import --no-rewrite -m "Import .zip, .mp3, .psd files" \
+$ git lfs migrate import --no-rewrite \
+  -m "Import test.zip, .mp3, .psd files in root of repo" \
   test.zip *.mpd *.psd
```
frantic1048: These actual arguments are still different from what commit message says.

- commit message: `.mp3`
- actual: `*.mpd`

This issue had been described in the second section of https://github.com/git-lfs/git-lfs/pull/3424#discussion_r245543300

PastelMobileSuit:
I'm 👍 on these new changes! @bk2204 feel free to merge this

PR #3443, opened 2018-12-19
Trim embedded paths out of binaries

bk2204:
When building binaries, Go embeds the source filenames into the binaries, meaning that on most systems, the build directory is embedded into the binary. This means that binaries are not reproducible when built by other users and in addition that the layout of the building user's system is embedded into the binary.

Go provides an option to trim paths out of the ones stored in the binary, but unfortunately this only works for one path, and since git-lfs is now built outside the GOPATH to take advantage of modules, there are two paths that need to be stripped.

For the benefit of build reproducibility and general build hygiene, let's strip out the user's home directory when building the binary. This is better than the status quo, and we can improve this later if circumstances change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,11 +34,13 @@ EXTRA_LD_FLAGS =
 LD_FLAGS = $(BUILTIN_LD_FLAGS) $(EXTRA_LD_FLAGS)
 
 # BUILTIN_GC_FLAGS are the internal flags used to pass compiler.
-BUILTIN_GC_FLAGS =
+BUILTIN_GC_FLAGS ?= all=-trimpath="$$HOME"
 # EXTRA_GC_FLAGS are the caller-provided flags to pass to the compiler.
 EXTRA_GC_FLAGS =
 # GC_FLAGS are the union of the above two BUILTIN_GC_FLAGS and EXTRA_GC_FLAGS.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -34,12 +34,14 @@ EXTRA_LD_FLAGS =
 LD_FLAGS = $(BUILTIN_LD_FLAGS) $(EXTRA_LD_FLAGS)
 
 # BUILTIN_GC_FLAGS are the internal flags used to pass compiler.
-BUILTIN_GC_FLAGS =
+BUILTIN_GC_FLAGS = all=-trimpath="$$HOME"
```
PastelMobileSuit: Should there be two `$` here?

bk2204: Yeah, that's intentional, since in a Makefile, `$` is used to specify make variables. If you want a literal `$` so it can be interpreted by the shell, you need to escape it, so `$$`. So what the shell sees is `all=-trimpath="$HOME"` and it does the right substitution.

We want this to be interpreted by the shell, since on Windows it's possible that we may end up with backslashes, and it's better to let the shell handle all of that, since writing `$(HOME)` and letting the Makefile do the substitution could result in odd behavior with backslashes and such.

I probably should have put all of this in the commit message, come to think of it.

bk2204: I've updated the commit message and made these values overridable for cases when users would rather user `PWD` instead of `HOME`.

PR #3444, opened 2018-12-21
Makefile additions for reproducible builds and asmflags

Foxboron:
This patch set include setting `$LDFLAGS` as a default `EXTRA_LD_FLAGS` parameter for supporting system linked flags. `trimpath` is set as a default for `BUILTIN_GC_FLAGS` so we are able to support [reproducible builds](https://reproducible-builds.org/) in the Makefile without any extra additions. Also added support for passing `-asmflags` to `go build` as we have support for `-ldflags` and `-gcflags`. This has the same variable for trimpath set as a default.

I also think enabling PIE by default with `-buildmode=pie` should be considered. However that has caused problems for other projects so it should be up to the developers if they want to support this as a default setting or not.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,8 +28,8 @@ BUILTIN_LD_FLAGS += -s
 BUILTIN_LD_FLAGS += -w
 endif
 # EXTRA_LD_FLAGS are given by the caller, and are passed to the Go linker after
-# BUILTIN_LD_FLAGS are processed.
-EXTRA_LD_FLAGS =
+# BUILTIN_LD_FLAGS are processed. By default the system LDFLAGS are passed.
+EXTRA_LD_FLAGS ?= -extldflags ${LDFLAGS}
 # LD_FLAGS is the union of the above two BUILTIN_LD_FLAGS and EXTRA_LD_FLAGS.
 LD_FLAGS = $(BUILTIN_LD_FLAGS) $(EXTRA_LD_FLAGS)
 


Foxboron:
Oof, I saw now that `trimpath` has a PR with #3443. If there is any specifics in this patch set you want i can modify as you see fit.

bk2204:
Yeah, as mentioned, in #3443, we have a bit of a problem in this scenario. Go only provides one path to trim, not two, and we really need two in this scenario; you've chosen `$PWD`, and I've chosen `$HOME`.

I'm going to modify that pull request to allow the variable to be overridden on the command line so that packagers can specify the right values for their environment if `$HOME` isn't it.

bk2204:
I agree that `$PWD` is better overall. However, since we rely on building outside of `$GOPATH` in order to get Go 1.11 module support to work, we end up with two different types of paths: those in `$PWD`, and those in `$GOPATH`. If we limit ourselves to `$PWD`, we'll continue to embed the user's `$GOPATH` into the binaries.

If Go 1.12 comes out and we can move back to building inside of the `$GOPATH` and still get module support working, then we can switch to `$GOPATH` instead.

Foxboron:
I see the other PR has been merged. Are you interested in the `$LDFLAGS` addition or should we close this PR?

bk2204:
I'm definitely fine with honoring the system LDFLAGS, so updating the PR so that it did that would be great.

Foxboron:
Done :)

Foxboron:
Ack and done

PR #3449, opened 2018-12-26
Add support for retries with delays (ex. rate limiting)

### LINKED ISSUE #3122: Obey rate-limits, instead of crashing
Some Git LFS servers might be rate-limited. When that is the case and the HTTP response 429 (Too many requests) is send, Git LFS will just crash.

Instead it should wait as long as specified in the [Retry-After response HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After), if this header was send by the server.

Example of a crash in Git LFS, when rate limit is exceeded. 

```
git clone https://git:MYTOKEN@code.somegit.com/example/example/example.git example
Cloning into 'example'...
remote: Counting objects: 1027, done.
remote: Compressing objects: 100% (885/885), done.
remote: Total 1027 (delta 91), reused 1017 (delta 86)
Receiving objects: 100% (1027/1027), 151.15 KiB | 2.19 MiB/s, done.
Resolving deltas: 100% (91/91), done.
Downloading example/data/example/class01/00000185.png (208 KB)
Error downloading object: example/data/example/class01/00000185.png (922316c): Smudge error: Error downloading example/data/example/class01/00000185.png
(942616ce675326ca5e8g31392aaa19eg3cfe3d17ae0p4e5991a930f716ca6eaa): batch response: Rate limit exceeded: https://git:MYTOKEN@code.somegit.com/example/example/example.git example/info/lfs/objects/batch

Errors logged to /home/example/code/tmp/example/.git/lfs/objects/logs/20180628T152316.531135072.log
fatal: example/data/example/class01/00000185.png: smudge filter lfs failed

```

zkry:
Fixes #3122 
This pull request introduces a new type of error which contains information regarding "Retry-After" header when a rate limit is hit. The error then behaves similar to the is-retriable error, except that this doesn't get re-added to the queue immediately, but rather waits until the time is up.

To test this I modified the git-lfs/lfs-test-server server to rate limit requests (maybe I can clean this up and make a PR for this as well), though I guess some automatic test would be needed. What would you suggest for this?

### DIFF
--- a/t/t-batch-retries-ratelimit.sh
+++ b/t/t-batch-retries-ratelimit.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+
+. "$(dirname "$0")/testlib.sh"
+
+begin_test "batch storage upload causes retries"
+(
+  set -e
+
+  reponame="batch-storage-upload-retry-later"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" batch-storage-repo-upload
+
+  contents="storage-upload-retry-later"
+  oid="$(calc_oid "$contents")"
+  printf "%s" "$contents" > a.dat
+
+  git lfs track "*.dat"
+  git add .gitattributes a.dat
+  git commit -m "initial commit"
+
+  GIT_TRACE=1 git push origin master 2>&1 | tee push.log
+  if [ "0" -ne "${PIPESTATUS[0]}" ]; then
+    echo >&2 "fatal: expected \`git push origin master\` to succeed ..."
+    exit 1
+  fi
+
+  assert_server_object "$reponame" "$oid"
+)
+end_test
+
+begin_test "batch storage download causes retries"
+(
+  set -e
+
+  reponame="batch-storage-download-retry-later"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" batch-storage-repo-download
+
+  contents="storage-download-retry-later"
+  oid="$(calc_oid "$contents")"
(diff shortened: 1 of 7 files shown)


bk2204:
Hey, this looks like a great improvement. Welcome to Git LFS!

We have a set of integration tests which use the test server in `t/cmd` and can be run using `make integration`. The best way to handle this would be to wire up a special name in the test server and then write a test case in shell that used that special repo name.

t/t-batch-retries-ratelimit.sh
```diff
@@ -0,0 +1,70 @@
...
+
+  git push origin master
+  assert_server_object "$reponame" "$oid"
+
+  pushd ..
+    git \
+      -c "filter.lfs.process=" \
+      -c "filter.lfs.smudge=cat" \
+      -c "filter.lfs.required=false" \
+      clone "$GITSERVER/$reponame" "$reponame-assert"
```
bk2204: Should we not have a test for the clone case as well, since that's a situation in which we're going to have a large number of requests?

zkry: Yeah, that makes sense. I updated the PR with the added test.

sharidz313: Tq

max-wittig:
Awesome! Thanks for implementing this :+1:

zkry:
No problem.  👍 Thank you guys for the support.

PR #3458, opened 2019-01-07
Modify logic of 'migrate info' to process extensionless files

### LINKED ISSUE #3453: git lfs migrate info doesn't list all files
I have a large file (>100MB) checked into my repo, but when I run the following command it doesn't show up:
` git lfs migrate info --everything --include="GoogleMobileAds"`

The file is located at 
`~/[redacted]/⁨Pods⁩/⁨Google-Mobile-Ads-SDK⁩/Frameworks⁩/⁨frameworks⁩/GoogleMobileAds.framework/GoogleMobileAds`

When I actually do the conversion to LFS using git lfs migrate import... it seems to correctly convert this file as I'm able to upload my cloned repo to GitHub (whereas before the import I get an error message).

But I'm wondering why this file won't show up when I do "... info..."

zkry:
Fixes #3453

It seems like to fix this, the the name to group the collection of file sizes should fall back to the file name if no extension exists. This leads to output that looks like this:

```
large5noext	315 MB	  1/1 files(s)	100%
*.c        	315 MB	  1/1 files(s)	100%
*.go       	98 KB 	11/11 files(s)	100%
LICENSE    	35 KB 	  1/1 files(s)	100%
*.sed      	25 KB 	45/45 files(s)	100%

### DIFF
--- a/t/fixtures/migrate.sh
+++ b/t/fixtures/migrate.sh
@@ -203,6 +203,32 @@ setup_multiple_local_branches() {
   git checkout master
 }
 
+# setup_multiple_local_branches_with_alternate_names performs the same task
+# as setup_multiple_local_branches, but creates a file with no extension.
+setup_multiple_local_branches_with_alternate_names() {
+  set -e
+
+  reponame="migrate-info-multiple-local-branches"
+
+  remove_and_create_local_repo "$reponame"
+
+  base64 < /dev/urandom | head -c 120 > no_extension
+  base64 < /dev/urandom | head -c 140 > a.txt
+
+  git add no_extension a.txt
+  git commit -m "initial commit"
+
+  git checkout -b my-feature
+
+  base64 < /dev/urandom | head -c 30 > a.txt
+  base64 < /dev/urandom | head -c 100 > no_extension
+
+  git add no_extension a.txt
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by bk2204** [comment]
This looks like a nice improvement. Would you mind adding a test for this somewhere under `t/`, maybe in `t-migrate-info.t`?

File: t/fixtures/migrate.sh [hunk 1]
```diff
@@ -203,6 +203,32 @@ setup_multiple_local_branches() {
...
+# as setup_multiple_local_branches, but creates a file with no extension.
+setup_multiple_local_branches_with_alternate_names() {
+  set -e
+
+  reponame="migrate-info-multiple-local-branches"
+
+  remove_and_create_local_repo "$reponame"
+
+  base64 < /dev/urandom | head -c 120 > no_extension
+  base64 < /dev/urandom | head -c 140 > a.txt
```
bk2204: I'll just make a note here that `head -c` isn't portable across systems. OpenBSD, for example, lacks it, and POSIX doesn't require it.

I don't think that's critical in this case, since we're basically doing the same thing slightly above, so anyone for whom this is a problem won't be significantly worse off than before.

PR #3477, opened 2019-01-15
Fix parameter name on List Locks API Documentation

Erwyn:
On locks listing, the git-lfs client is not querying the `ref` parameter but the `refspec` one. Fix the documentation accordingly.

For reference: https://github.com/git-lfs/git-lfs/blame/84f47a430beb9695ceb8427443f304e091128edd/locking/api.go#156-L157


bk2204:
Thanks for this, and welcome to Git LFS!

This looks good to me, and again, I appreciate you sending this in.

Erwyn:
You're welcome!

PR #3480, opened 2019-01-15
Run go generate only on Windows

### LINKED ISSUE #3473: Can't run "go generate" on non-Windows
As part of building the git-lfs Debian package, `go generate` must be run to generate the help information.  Dh_golang, the debian build tool for go, has a built in flag `DH_GOLANG_GO_GENERATE` which, if set to `1`, will automatically run `go generate` on the package.

The problem with using this flag is that running `go generate` on git-lfs also attempts to run https://github.com/git-lfs/git-lfs/blob/master/git-lfs.go#L1, which fails because `goversioninfo` is not installed.

One solution could be to package it for Debian but since it's windows-specific that seems to be of questionable value.  I currently have worked around it by patching out that line [1] but that obviously would not be an acceptable solution because it would break Windows builds.  One thing that was suggested was to move the `go generate` to a file with a windows-specific but since there are none of those in the root of the project that may not work based on my reading of the `goversioninfo` docs.

While the git-lfs Makefile works around this by only calling `go generate` on Windows builds, I do think that this is at least a valid bug - one might reasonably expect to be able to run `go generate` successfully regardless of platform.  I will do some testing and see if I can come up with a good solution but I wanted to at least file the issue so you all could be aware of it since you know the codebase better than I do and might have a better idea for how to fix it.

[1] https://salsa.debian.org/go-team/packages/git-lfs/commit/8a4bad783a43ee05c144b96f44671eec62324a77

bk2204:
On Windows, we want to run `goversioninfo`, which embeds certain version information into Windows binaries. However, despite compiling and running on non-Windows systems, it isn't very useful there. Move the `go:generate` comment into a separate file that's only built on Windows.

This improves the lives of packagers on Debian, where the packaging tools want to run `go generate` by default.

I've confirmed that running `go generate` produces a `resource.syso` file on Windows and does not on macOS.

Fixes #3473.

/cc @ssgelm

### DIFF
--- a/git-lfs.go
+++ b/git-lfs.go
@@ -1,5 +1,3 @@
-//go:generate goversioninfo
-
 package main
 
 import (
--- a/git-lfs_windows.go
+++ b/git-lfs_windows.go
@@ -1,0 +1,4 @@
+//+build windows
+//go:generate goversioninfo
+
+package main


ssgelm:
This worked correctly on Linux.

PR #3483, opened 2019-01-17
Update Cobra

### LINKED ISSUE #3478: Consider updating github.com/spf13/cobra to latest tagged release
Back in #3111, there was a proposal to update Cobra to the latest release, in order to make the lives of distro packagers easier. However, doing so caused some failures with Glide. Now that we're using Go modules, this should be less of an issue, so we can reconsider updating this dependency to a newer version.

bk2204:
It's generally helpful for packagers if we pick released versions of dependencies when possible. With that in mind, update Cobra and Pflag to their latest released versions. Cherry-pick a patch from @ssgelm to preserve the way that invocations such as `git lfs help status` work.

I've tested that it's still possible to run the Docker containers successfully with this patch, which was the problem for us when we were using Glide.

Fixes #3478.


ssgelm:
Yeah this would be great and a big help for Debian.  We currently apply my patch when packaging (https://salsa.debian.org/go-team/packages/git-lfs/blob/master/debian/patches/fix-help.patch) but it'd be awesome to not have to!

PR #3486, opened 2019-01-21
smudge-filter: try all remotes when getting lfs objects

wsprent:
Hi,

I have encountered this behaviour when working with a repo that contains lfs pointers and has two remotes:

1. Say we have `origin` and some other remote, `other-remote`.
2. `git fetch other-remote some-branch`.
3. `git checkout other-remote/some-branch`.
4. git lfs fails as it tries fetching all lfs objects from `origin`.

I feel like the above is a pattern that*should *just work*, and that falling back on trying another remote is a better alternative than bailing. This patch solves that exact workflow, but I recognise that it is a bit of a hack (in fact, I don't fully understand why it works with `filter-process`).
My first impression is that teaching the `TransferQueue` about multiple remotes would probably be the cleaner change, but would also require much more work and be more risky.

I'm submitting this PR mainly to hear if you think this kind of behaviour even makes sense for the lfs client, and, if so, advice on where it would be appropriate to implement it. :)

### DIFF
--- a/lfs/gitfilter_smudge.go
+++ b/lfs/gitfilter_smudge.go
@@ -86,22 +86,50 @@ func (f *GitFilter) Smudge(writer io.Writer, ptr *Pointer, workingfile string, d
 func (f *GitFilter) downloadFile(writer io.Writer, ptr *Pointer, workingfile, mediafile string, manifest *tq.Manifest, cb tools.CopyCallback) (int64, error) {
 	fmt.Fprintf(os.Stderr, "Downloading %s (%s)\n", workingfile, humanize.FormatBytes(uint64(ptr.Size)))
 
-	// NOTE: if given, "cb" is a tools.CopyCallback which writes updates
-	// to the logpath specified by GIT_LFS_PROGRESS.
-	//
-	// Either way, forward it into the *tq.TransferQueue so that updates are
-	// sent over correctly.
+	mainRemote := f.cfg.Remote()
+	remotes := append([]string{mainRemote}, f.cfg.Remotes()...)
+	for i, remote := range remotes {
+		if remote == mainRemote {
+			remotes[0] = mainRemote
+			remotes[i] = remote
+			break
+		}
+	}
 
-	q := tq.NewTransferQueue(tq.Download, manifest, f.cfg.Remote(),
-		tq.WithProgressCallback(cb),
-		tq.RemoteRef(f.RemoteRef()),
-	)
-	q.Add(filepath.Base(workingfile), mediafile, ptr.Oid, ptr.Size)
-	q.Wait()
+	allErrs := make([]error, 0)
+	completed := false
+	for i, remote := range remotes {
 
-	if errs := q.Errors(); len(errs) > 0 {
+		// NOTE: if given, "cb" is a tools.CopyCallback which writes updates
+		// to the logpath specified by GIT_LFS_PROGRESS.
+		//
+		// Either way, forward it into the *tq.TransferQueue so that updates are
+		// sent over correctly.
+
+		q := tq.NewTransferQueue(tq.Download, manifest, remote,
+			tq.WithProgressCallback(cb),
+			tq.RemoteRef(f.RemoteRef()),
+		)
+		q.Add(filepath.Base(workingfile), mediafile, ptr.Oid, ptr.Size)
+		q.Wait()
+
+		errs := q.Errors()
+
+		if len(errs) == 0 {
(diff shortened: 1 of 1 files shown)


lfs/gitfilter_smudge.go
```diff
@@ -86,22 +86,50 @@ func (f *GitFilter) Smudge(writer io.Writer, ptr *Pointer, workingfile string, d
...
-		tq.RemoteRef(f.RemoteRef()),
-	)
-	q.Add(filepath.Base(workingfile), mediafile, ptr.Oid, ptr.Size)
-	q.Wait()
-
-	if errs := q.Errors(); len(errs) > 0 {
+	mainRemote := f.cfg.Remote()
+	remotes := append([]string{mainRemote}, f.cfg.Remotes()...)
+	for i, remote := range remotes {
+		if remote == mainRemote {
```
ttaylorr: Since you initialize `remotes` as:

```go
removes := append([]string{mainRemote}, f.cfg.Remotes()...)
```

I think that it would always be the case that `remotes[0] == mainRemote`, and hence this loop would terminate in the same fashion after only a single iteration. Can you clarify its original intent, and see whether or not it is necessary?

lfs/gitfilter_smudge.go
```diff
@@ -86,22 +86,50 @@ func (f *GitFilter) Smudge(writer io.Writer, ptr *Pointer, workingfile string, d
...
+
+		errs := q.Errors()
+
+		if len(errs) == 0 {
+			completed = true
+			break
+		}
+
+		if i < len(remotes)-1 {
+			fmt.Fprintf(os.Stderr, "%s not found on remote \"%s\" - trying \"%s\" next\n", workingfile, remote, remotes[i+1])
```
ttaylorr: I'm not sure that we should be printing these messages to the user; I'd expect that if this were the behavior that smudging exhibited, it would proceed on through the remotes list transparently and not tell me about it. I am glad, however, that we are printing to stderr instead of stdout.

Perhaps we could add a knob here, but I'm not sure that it's worthwhile.

lfs/gitfilter_smudge.go
```diff
@@ -86,22 +86,50 @@ func (f *GitFilter) Smudge(writer io.Writer, ptr *Pointer, workingfile string, d
...
+		}
+
+		allErrs = append(allErrs, errs...)
```
ttaylorr: I am not sure that this works for all cases that you are trying to implement it over. For example, if we try remote `A` which does _not_ object `X`, but then continue to remote `B` which does, won't `len(allErrs) != 0`? In other words, won't we retain the missing object error from `A`, even though it has been obsoleted by `B`?

lfs/gitfilter_smudge.go
wsprent: Ah nice catch. My intent was to have a list of all remotes but keep the remote suggested by the original heuristics in front, so behaviour would only change if an object is missing.

Basically, `remotes` should be initialised as `f.cfg.Remotes()`.

lfs/gitfilter_smudge.go
wsprent: We retain the errors, but don't act on them as `completed` is `true`, and the length of `allErrs` is never checked. But I guess it might be more clear to set `allErrs` to an empty slice and scrap the boolean.

bk2204:
> > The price for having it _just work_ here is obviously having to spend time trying multiple remotes, but the alternative is an error and a working directory in a bad state.
> 
> I think that this is the crux of it for me: trying multiple remotes can be quite costly in lots of situations, so I think that the tradeoff for an extra `git lfs fetch` command is a fair one. @bk2204 what do you think?

Yeah, I think that's a fair tradeoff here.

As I said before, if we can automate knowing which remote to pull from, I'm all for that as an improvement.

wsprent:
While I think it's obviously fair enough to have the extra `lfs fetch` in the workflow, I would like to point out two things:
1. If we order the remotes git lfs look at by first looking at the git lfs would've tried now, then getting objects will only behave differently in the cases where both more than one remote is configured _and_ the first remote tried did not have the object. In all of these cases, current behaviour would be to fail outright.
I.e: any use of `git lfs` which would be previously succesful, would take the same amount of time -- only commands that would now fail would take longer time, and then possibly succeed.
2. If a user has explicitly configured a second remotes (or more), is it not fair to assume that these remotes are interesting to the user?

steffengodskesen:
@larsxschneider ^^

Issue #3490, opened 2019-01-23
git commit take unexpected long time for Cygwin


nhatkhai:
It seem to me every time I commit a normal files (text file) with small change took more than 40seconds to be finished. 
When I look at the TRACE, it seem to be delay for a while at ``11:42:05.182320 trace git-lfs: exec: git 'version'``. 
Also, run ``git lfs install`` is really slow too. Wounder they are relate. 
I Also noted that ``git lfs env`` took a quite a long time to complete too. 
See all the log bellow:

```
git commit -a -m ".................."
11:42:04.475541 git.c:344               trace: built-in: git commit -a -m '...........'
11:42:04.842330 run-command.c:640       trace: run_command: git gc --auto
11:42:04.920327 git.c:344               trace: built-in: git gc --auto
11:42:04.935927 run-command.c:640       trace: run_command: GIT_EDITOR=: GIT_INDEX_FILE=/cygdrive/c/projects/git/XYZ/.git/modules/XX/YY/index /cygdrive/c/projects/git/XYZ/.git/modules/XX/YY/hooks/post-commit
11:42:05.091922 git.c:576               trace: exec: git-lfs post-commit
11:42:05.091922 run-command.c:640       trace: run_command: git-lfs post-commit
11:42:05.182320 trace git-lfs: exec: git 'version'
11:43:23.773230 trace git-lfs: exec: git 'config' '-l'
11:43:34.398492 trace git-lfs: exec: git '-c' 'filter.lfs.smudge=' '-c' 'filter.lfs.clean=' '-c' 'filter.lfs.process=' '-c' 'filter.lfs.required=false' 'rev-parse' 'HEAD' '--symbolic-full-name' 'HEAD'
11:43:47.358082 trace git-lfs: filepathfilter: rewrite ".git" as "**/.git/**"
11:43:47.358082 trace git-lfs: filepathfilter: rewrite "**/.git" as "**/.git"
11:43:47.358082 trace git-lfs: filepathfilter: rejecting "YY" via []
11:43:47.358082 trace git-lfs: filepathfilter: accepting "YY"
11:43:47.358082 trace git-lfs: filepathfilter: rewrite "~*.*" as "**/~*.*"
....
```

```
(519)$ git-lfs version
git-lfs/2.6.1 (GitHub; windows amd64; go 1.11.1; git dc072c3e)
(519)$ git version
11:52:17.642972 git.c:344               trace: built-in: git version
git version 2.17.0
```

```
[...]

PR #3492, opened 2019-01-23
git/version.go: replace sync.Mutex usage with sync.Once

ttaylorr:
We use a sync.Mutex to synchronize access to the string pointer
`gitVersion`, which indicates the version of Git used by the system
running Git LFS.

Our basic usage of the sync.Mutex is not incorrect, but we can improve
the readability by instead using a sync.Once. A sync.Once determines
very quickly (and in an atomic, goroutine-safe fashion) whether or not
_any_ function has been run, and if it hasn't, run it.

By doing this, we can--at the first request--produce a value for the
result of running 'git version', and then return it later to the caller.
This has the following benefits:

  - If 'git version' has already been run, we do not need to hold the
    lock for the entire duration of the function.

  - If 'git version' has not already been run, we only run it once,
    retaining the existing behavior.

Only one change, which is the introduction of the `gitVersionErr`
variable. This is a consequence of executing the 'git version' call in a
closure: since we're in a new stack frame, we can't return from our
parent.

Instead, we retain the value for all time, and return _it_, along with
whatever value we got from running 'git version' in the first place.

##

/cc @git-lfs/core 
/cc https://github.com/git-lfs/git-lfs/issues/3490

### DIFF
--- a/git/version.go
+++ b/git/version.go
@@ -10,21 +10,17 @@ import (
 )
 
 var (
-	gitVersion   *string
-	gitVersionMu sync.Mutex
+	gitVersionOnce sync.Once
+	gitVersion     string
+	gitVersionErr  error
 )
 
 func Version() (string, error) {
-	gitVersionMu.Lock()
-	defer gitVersionMu.Unlock()
-
-	if gitVersion == nil {
-		v, err := subprocess.SimpleExec("git", "version")
-		gitVersion = &v
-		return v, err
-	}
-
-	return *gitVersion, nil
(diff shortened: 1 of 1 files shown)


ttaylorr:
> This looks great. Thanks for writing this up, and my apologies for taking so long to get to review this.

Thanks for reviewing it, and no worries. I had forgotten about it myself until I was browsing through open pull requests (as I tend to do every ~week or so just to make sure things aren't getting stale).

Issue #3500, opened 2019-01-29
[Suggestion] Provide a way to check if a blob is an LFS pointer


markonius:
I can get a regular old file with `git cat-file blob <blob_hash>`, but I need a way to check if that blob is an LFS pointer so that I know whether I need to pipe it into `git lfs smudge`.  
Something like `git lfs pointer --check <pointer_contents>`

I'm sorry if something like this is already possible, I couldn't find that functionality.

vbjay:
```git cat-file -p $TREE_SHA1:.gitattributes | git check-attr --stdin -- $FILENAME | grep "filter: lfs"```

This is a great idea.  Could even accept a blob sha.  

```bash
git lfs pointer --check <blob sha>
git lfs pointer --check <file contents>
```

ttaylorr:
> `git cat-file -p $TREE_SHA1:.gitattributes | git check-attr --stdin -- $FILENAME | grep "filter: lfs"`

I'm not quite sure this is doing exactly what @markonius was asking for. Your command seems to look for all entries in the `.gitattributes` file that are affected with the Git LFS filter. I think that this can be slightly simplified to:

```ShellSession
$ git check-attr filter -- $filename
```

In any case, this doesn't dictate whether or not `$filename` is a Git LFS pointer in the history at all points in time, since a user can bypass the `.gitattributes` mechanism in a number of ways, e.g., by not having Git LFS installed, disabling their clean filter, and so on.

> I need a way to check if that blob is an LFS pointer so that I know whether I need to pipe it into `git lfs smudge`.

I think what `git lfs pointer --check` might do is something like this: read in a file, and report whether or not it is parse-able as a valid pointer. I can recommend some hack-y ways to do this sort of thing, but really this sort of behavior should live within Git LFS, since it can act as an authoritative source on whether or not something is a pointer.

I implemented such a functionality in https://github.com/git-lfs/git-lfs/pull/3501.

vbjay:
No.  I was using it as a reverse point of what can be used to find out if pathx is lfs tracked.  We would LOVE for the feature of I have blob sha(sha from git-ls-tree that is a blob)

ttaylorr:
@vbjay I think that this feature is not Git LFS specific (in other words, whereas @markonius's request is asking what _Git LFS_ thinks about something, this is asking only about what _Git_ thinks of the world).

One thing you could do is:

```ShellSession
[...]

PR #3501, opened 2019-01-30
commands/command_pointer.go: introduce `--check` option

ttaylorr:
We have used the 'git-lfs-pointer(1)' command historically to compare
varying implementation of the Git LFS protocol. This has been done by
passing both `--file` and `--stdin` to the command, which will in turn
compare the contents of both handles and return appropriately whether
they are the same or different.

One use case that Git LFS pointer does not yet support directly is
_checking_ to see whether a given file is a valid Git LFS pointer. This
is a different use-case than before, since we aren't doing any
comparison, we're simply checking whether the official implementation of
Git LFS can parse a given stream as a pointer.

As an aside, one way to do this today is the following:

  $ git lfs clean < a.txt | git lfs pointer --stdin --file my.ptr

Where we first generate a pointer of the file 'a.txt' (via
`git-lfs-clean(1)`), and then redirect that to `git-lfs-pointer(1)`
against our own file.

Let's make this above incantation easier to execute by providing a
functionally equivalent `--check` option. Running `git lfs pointer
--check` (and passing either `--file`, or `--stdin`) will return either
1 or 0 depending on whether or not the given pointer file was invalid or
not.

Closes: https://github.com/git-lfs/git-lfs/issues/3500.

##

/cc @git-lfs/core 
/cc https://github.com/git-lfs/git-lfs/issues/3500 @markonius

### DIFF
--- a/commands/command_pointer.go
+++ b/commands/command_pointer.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 
 	"github.com/git-lfs/git-lfs/git"
@@ -18,6 +19,7 @@ var (
 	pointerFile    string
 	pointerCompare string
 	pointerStdin   bool
+	pointerCheck   bool
 )
 
 func pointerCommand(cmd *cobra.Command, args []string) {
@@ -25,6 +27,36 @@ func pointerCommand(cmd *cobra.Command, args []string) {
 	something := false
 	buildOid := ""
 	compareOid := ""
+
+	if pointerCheck {
+		var r io.ReadCloser
+		var err error
+
+		if len(pointerCompare) > 0 {
+			ExitWithError(fmt.Errorf("fatal: cannot combine --check with --compare"))
+		}
+
+		if len(pointerFile) > 0 {
+			if pointerStdin {
+				ExitWithError(fmt.Errorf("fatal: with --check, --file cannot be combined with --stdin"))
+			}
+			r, err = os.Open(pointerFile)
+			if err != nil {
+				ExitWithError(err)
+			}
+		} else if pointerStdin {
+			r = ioutil.NopCloser(os.Stdin)
+		} else {
+			ExitWithError(fmt.Errorf("fatal: must specify either --file or --stdin with --compare"))
+		}
+
+		_, err = lfs.DecodePointer(r)
+		if err != nil {
+			os.Exit(1)
+		}
+		r.Close()
+		return
+	}
 
 	if len(pointerCompare) > 0 || pointerStdin {
 		comparing = true
@@ -129,5 +161,6 @@ func init() {
 		cmd.Flags().StringVarP(&pointerFile, "file", "f", "", "Path to a local file to generate the pointer from.")
 		cmd.Flags().StringVarP(&pointerCompare, "pointer", "p", "", "Path to a local file containing a pointer built by another Git LFS implementation.")
 		cmd.Flags().BoolVarP(&pointerStdin, "stdin", "", false, "Read a pointer built by another Git LFS implementation through STDIN.")
+		cmd.Flags().BoolVarP(&pointerCheck, "check", "", false, "Check whether the given file is a Git LFS pointer.")
 	})
 }
--- a/docs/man/git-lfs-pointer.1.ronn
+++ b/docs/man/git-lfs-pointer.1.ronn
@@ -1,11 +1,12 @@
-git-lfs-pointer(1) -- Build and compare pointers
-================================================
+git-lfs-pointer(1) -- Build, compare, and check pointers
+========================================================
 
 ## SYNOPSIS
 
 `git lfs pointer --file=path/to/file`<br>
 `git lfs pointer --file=path/to/file --pointer=path/to/pointer`<br>
 `git lfs pointer --file=path/to/file --stdin`
+`git lfs pointer --check --file=path/to/file`
 
 ## Description
 
@@ -24,6 +25,12 @@ between different Git LFS implementations.
 * `--stdin`:
     Reads the pointer from STDIN to compare with the pointer generated from
     `--file`.
+
+* `--check`:
+    Reads the pointer from STDIN (if `--stdin` is given) or the filepath (if
+    `--file`) is given. If neither or both of `--stdin` and `--file` are given,
+    the invocation is invalid. Exits 0 if the data read is a valid Git LFS
+    pointer. Exits 1 otherwise.
 
 ## SEE ALSO
 
--- a/t/t-pointer.sh
+++ b/t/t-pointer.sh
@@ -2,7 +2,6 @@
 
 . "$(dirname "$0")/testlib.sh"
 
-
 begin_test "pointer --file --stdin"
 (
   set -e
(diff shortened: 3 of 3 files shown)


vbjay:
I want to pass a blob sha into ```git lfs pointer --check ``` somehow and it do the checking.  Yes we can use git cat-file to get the contents of file but adding sha option would mean we could call one command and get a result from it.

``bash
`git lfs pointer --check <blob sha>```
OR
`git lfs pointer --check --sha <blob sha>```
```

**Comment by ttaylorr** [comment]
@vbjay I am opposed to adding `--sha`. The reason is this: `git lfs pointer` abuses the `--file` and `--stdin` flags as sides `A` and `B` in a comparison. Adding a _new_ flag makes these semantics more confusing: what happens if I pass `--file` and `--sha`, or all three?

Of course, we could defend around this, but I think that `git cat-file -p <sha1> | git lfs pointer --stdin` is much easier.

**Comment by vbjay** [comment]
I did give an option without --sha.  Sure.  We could run the multipart command.  


The context is https://github.com/gitextensions/gitextensions/pull/6189  We have a blob sha.  We want to know is it LFS?  We can provide the contents of cat-file but I was thinking it would be a nice to have.

**Comment by ttaylorr** [comment]
> We could run the multipart command.

Yeah, I think that `git cat-file <blobsha> | git lfs pointer --check --stdin` would be the best here.

**Comment by vbjay** [comment]
Here is the thinking behind this.  If we cat-file, we get the full contents of the file.  lfs spec says you check x number of bytes for a header to see if it is a pointer.  Instead of us sending full file to tour stdin, we could pass in sha and your command could get head...contents from file and verify pointer and give us a yes or no answer.

File: docs/man/git-lfs-pointer.1.ronn [hunk 1]
```diff
@@ -25,6 +26,11 @@ between different Git LFS implementations.
     Reads the pointer from STDIN to compare with the pointer generated from
     `--file`.
 
+* `--check`:
```
PastelMobileSuit: Should we mention here that this flag has to be paired with either `--file` or `--stdin`? And that it's incompatible with `--pointer`?

commands/command_pointer.go
```diff
@@ -26,6 +28,33 @@ func pointerCommand(cmd *cobra.Command, args []string) {
...
 
+	if pointerCheck {
+		var r io.ReadCloser
+		var err error
+
+		if len(pointerCompare) > 0 {
+			ExitWithError(fmt.Errorf("fatal: cannot combine --check with --compare"))
+		}
+
+		if len(pointerFile) > 0 {
```
PastelMobileSuit: It looks like if a user passes in both `--pointer` and `--stdin` the check will automatically default to the pointer file.

I'm wondering if instead we should error if both flags are provided? Given that absent an error this command just returns an exit status, I could see somebody getting confused about what exactly is being checked in this case.

ttaylorr:
> If we cat-file, we get the full contents of the file. lfs spec says you check x number of bytes for a header to see if it is a pointer.

I don't think that that's necessarily the case. Git LFS opens the _stream_ for reading, but does not buffer the entire contents of your file in memory, whether it comes from `--stdin` or `--file`. Even if your file is 10 GiB large, Git LFS will [only read the first 1024 bytes of it](https://github.com/git-lfs/git-lfs/blob/v2.6.1/lfs/pointer.go#L105-L127) in order to decide whether or not what you gave was a pointer.

Even so, it's not clear how Git LFS would solve this problem any better than `git cat-file ... |` could. If the object you were sending was loose, Git LFS could open the file and only read the first `N` bytes (where `N` is large enough to encode a pointer). But in the case where your file is packed, Git would have to first _unpack_ the entire contents of your file, and then send it across. This is a quick operation in practice, but has to happen nonetheless.

The overwhelmingly common case, I argue, occurs from the working copy, where the file is already laid out, and the only cost we pay (speaking from a 10,000-foot perspective) is opening the file, reading the first 1 KiB of it, and then closing the file.

> Since headers can change over versions, it would be a bad idea for us to keep up to date with your header size. Instead, we just use your official command.

This is a fair point, but I don't think that it is a problem because of the above. Even if the header size changes (say, to be more than 1024 bytes long), Git LFS would simply update to read more of the payload. Because you're not paying an extra cost on sending data that we don't read, you can simply send the file as normal.

docs/man/git-lfs-pointer.1.ronn
ttaylorr: That's a great idea, and I appreciate your 👀 which are certainly much more careful than mine ;-).

commands/command_pointer.go
ttaylorr: This idea is great, too. These flags both exist for archaic reasons, but don't make sense when combined together with `--check`. I think that we should exit with a non-zero code and print a helpful message (as well as document this in `git-lfs-pointer.1.ronn`, and test it in `t/`, too)

ttaylorr:
@PastelMobileSuit thanks for your excellent review. I just pushed a new round; could you review that, too?

ttaylorr:
@vbjay I think that I am unclear on why you're asking `git lfs pointer --check` take in a blob SHA. If you can pipe it through `git cat-file -p :<blobsha> |`, what is the benefit that Git LFS would provide you specifically?

vbjay:
Why are you so resistant to adding it?  What is your problem with adding it?  shas are used all over the place.  What is so wrong with us passing a sha of to you and you can read x bytes in from cat-file on your end.  I understand how to use the cat-file | ...   I am just curious why you feel it is wrong to provide this functionality.

ttaylorr:
> I am just curious why you feel it is wrong to provide this functionality.

I don't feel that it's wrong per-se, but I feel that it's important I have a foundational understanding of why a user might need such a functionality, when it is equivalently provided through piping.

vbjay:
Did you follow the link I sent of the lines of code?  We are working on a system we built that lets us run git and other commands.  If we have to call cat-file... and then take that stream and copy it to a new process that is your git lfs pointer --check, I think that is foolish when we could pass blobSha to a run of git lfs pointer --check.  Then we could adjust our cat-file  call or run any other command based on teh result.  We could also show icons saying it is an LFS file or one of many other things we could do with the answer of Is blob X an LFS pointer.

Think of it as a  benefit in scripting and other uses of LFS.  If we can pass in a blob sha, we can have confidence that your cat-file call will use the right args and we don't have to figure out the right args to send correct input to your stdin.  Instead, you know what you want.  You proces the output of cat-file and give the user of the command the needed info.

ttaylorr:
> If we have to call cat-file... and then take that stream and copy it to a new process that is your git lfs pointer --check, I think that is foolish when we could pass blobSha to a run of git lfs pointer --check.

Ah, thanks for clarifying. It seems the crux of the issue is that it may not be possible (?) to simulate this sort of pipelining from your programming language. I do not use C# and thus am not sure one way or another. In Go, one would write something like [this](https://play.golang.org/p/gOqJXwjB9oz).

Is the above possible in C#? If so, we would prefer that you do that, instead of relying on us implementing `--sha`. Here is the reason that we would prefer not to implement `--sha`:

- Since `git lfs pointer`'s inception, it was designed to compare two pointers for equality, where one pointer comes from `--file`, and the other comes from `--stdin`. (This is perhaps not how we'd design the command today, but it is the interface that we are bound to uphold.)

- When we introduce `--check`, it makes sense that this could be combined with `--file`, OR `--stdin` OR even `--sha`. In other words, it makes sense for `git lfs pointer --check` to take a single file as the source for the pointer that it would like to check.

- So, on its own, `git lfs pointer --check --sha` makes sense, I agree. _But_, in conjunction with `--stdin`, and `--file`, we now have `2^3` combinations that we have to retrofit onto the old-style pointer comparison operation, which is undesirable. Further, we can imagine this number growing exponentially, as Git adds more hashing algorithms beyond `--sha1`. (E.g., do we need to support `--sha3`, `--sha256`, `--sha1` and so on? As we add more of these there are exponentially larger numbers of combinations that we have to support.

I hope that this reasoning is OK for you. It's not that we're opposed to adding `--check --sha` on its own, it's the ramifications that it would have on the existing usage of the command. That's why I'm encouraging the use of `git cat-file -p | git lfs pointer --check`, to accomplish what you are looking for without having to add the feature to LFS.

vbjay:
C# does support stream redirection.  

I would think --sha would be enough.  git would handle the sha parsing.  You pass the sha on to cat-file and it gets the contents.  You shouldn't care if it is a sha256 or sha1.

vbjay:
Example:

C= client
L=LFS

C-> git cat-file ...->L pointer --check --stdin  I now have to send you full stream and I can't say send first 1024 bytes because C doesn't know header size of LFS.

C-> L pointer --check --sha 11223344 ->(git cat-file ... 11223344)  L knows official header size and can stop reading result from cat-file early.

ttaylorr:
> I now have to send you full stream and I can't say send first 1024 bytes because C doesn't know header size of LFS.

I don't think that that's the case: Git LFS will close the stream after either (1) reading a pointer successfully or (2) reading 1024 bytes and determining that it wasn't a pointer. There is no buffering or similar going on underneath, Git LFS will only read what it needs to, and discard the rest.

> I would think --sha would be enough.

As I have mentioned, our concern is not that `--sha` would be difficult to implement for `git lfs pointer --check`, it's that there would be unintended complications with all of the _existing_ flags that `git lfs pointer` takes.

Pending @PastelMobileSuit's review, I will merge support for this `git lfs pointer --check` without `--sha`.

vbjay:
> Git LFS will close the stream after either...

That is what I did not know.  So passing in stdin of cat-file will not make it send all 4GB of a file that is in the repo?

Let's say we have a 4GB file in a git repo(not lfs tracked).  Ignoring the fun of that and just using it as a crazy example.  We run git cat-file on that blob.  If we send that stream to your stdin, you don't read all bytes, you close the stream when you parse the possible header and return yes or no to the caller.

PR #3510, opened 2019-02-05
Release automation

bk2204:
Our current release process is highly manual and requires several individual steps to go through. This series is a first step at automating parts of this process.

Included are changes to automate most of the Windows build process, which is an area that has caused us problems in the past, a script to create a GitHub release and upload assets, and a script to bump version numbers.

The entire process is not yet complete, but since this series provides valuable improvements that we can take advantage of now, I thought I'd submit it and let us use parts of it. I've already used the Windows build process to great success in the last release.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -288,6 +288,12 @@ bin/releases/git-lfs-%-$(VERSION).zip : $(RELEASE_INCLUDES) bin/git-lfs-%.exe
 bin/releases/git-lfs-$(VERSION).tar.gz :
 	git archive -o $@ --prefix=git-lfs-$(patsubst v%,%,$(VERSION))/ --format tar.gz $(VERSION)
 
+# release-linux is a target that builds Linux packages. It must be run on a
+# system with Docker that can run Linux containers.
+.PHONY : release-linux
+release-linux:
+	./docker/run_dockers.bsh
+
 # release-windows is a target that builds and signs Windows binaries.  It must
 # be run on a Windows machine under Git Bash.
 #
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,10 @@ GOIMPORTS_EXTRA_OPTS ?= -w -l
 TAR_XFORM_ARG ?= $(shell tar --version | grep -q 'GNU tar' && echo '--xform' || echo '-s')
 TAR_XFORM_CMD ?= $(shell tar --version | grep -q 'GNU tar' && echo 's')
 
+# CERT_SHA1 is the SHA-1 hash of the Windows code-signing cert to use.  The
+# actual signature is made with SHA-256.
+CERT_SHA1 ?= 516f21950afecd3779b8b77da92f738fec501f03
+
 # SOURCES is a listing of all .go files in this and child directories, excluding
 # that in vendor.
 SOURCES = $(shell find . -type f -name '*.go' | grep -v vendor)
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -284,6 +288,51 @@ bin/releases/git-lfs-%-$(VERSION).zip : $(RELEASE_INCLUDES) bin/git-lfs-%.exe
...
+	mv git-lfs-x86.exe git-lfs-windows-386.exe
+	@# We use tar because Git Bash doesn't include zip.
+	tar -cf $@ git-lfs-windows-amd64.exe git-lfs-windows-386.exe git-lfs-windows.exe
+	$(RM) git-lfs-windows-amd64.exe git-lfs-windows-386.exe git-lfs-windows.exe
+
+# release-windows-rebuild takes the archive produced by release-windows and
+# incorporates the signed binaries into the existing zip archives.
+.PHONY : release-windows-rebuild
+release-windows-rebuild:
+	[ -f "bin/releases/git-windows-assets-$(VERSION).tar.gz" ] || { \
```
ttaylorr: Can we depend on these directly, as in:

```Makefile
.PHONY : release-windows-rebuild
release-windows-rebuild : bin/releases/git-windows-assets-$(VERSION).tar.gz
        # ...
```

Makefile
```diff
@@ -284,6 +288,51 @@ bin/releases/git-lfs-%-$(VERSION).zip : $(RELEASE_INCLUDES) bin/git-lfs-%.exe
...
+		echo "Windows assets are missing." >&2; exit 1; }
+	temp=$$(mktemp -d); \
+	file="$$PWD/bin/releases/git-windows-assets-$(VERSION).tar.gz"; \
+		( \
+			tar -C "$$temp" -xzf "$$file" && \
+			for i in amd64 386; \
```
ttaylorr: I think that flattening these is fine. It might be preferable to have a flat structure like the following, instead of the increased level in indentation. These lines look similarly enough that I imagine a reader could quickly spot the difference in `amd64` vs `386`.

```Makefile
( \
        zip -j -l bin/releases/git-lfs-windows-$(VERSION).zip "$$temp/git-lfs-windows-amd64.exe"; \
        zip -j -l bin/releases/git-lfs-windows-$(VERSION).zip "$$temp/git-lfs-windows-386.exe"; \
        cp "$$temp/git-lfs-windows.exe" bin/releases/git-lfs-windows-$(VERSION).exe; \
\ )
```

script/changelog
```diff
@@ -20,18 +20,50 @@ commit_summary() {
   echo "* $title #$id (@$author)"
 }
 
+revisions () {
```
ttaylorr: I'm wondering if there is another name we could use for `revisions()`. The reason I suggest this is that I was myself confused at what this function does. Perhaps instead we could call it:

- `revisions_in` (i.e., so that it would read `revisions_in $range`), or
- `get_revisions_in` (so that it is more clear that this method "looks something up").

What do you think?

Makefile
```diff
@@ -288,6 +288,10 @@ bin/releases/git-lfs-%-$(VERSION).zip : $(RELEASE_INCLUDES) bin/git-lfs-%.exe
 bin/releases/git-lfs-$(VERSION).tar.gz :
 	git archive -o $@ --prefix=git-lfs-$(patsubst v%,%,$(VERSION))/ --format tar.gz $(VERSION)
 
+.PHONY : release-linux
```
ttaylorr: I think that it's unnecessary, but it may be good to document this target in a similar fashion as `release-windows` below.

script/upload
```diff
@@ -0,0 +1,262 @@
...
+    echo FreeBSD
+  else
+    ruby -e 'puts ARGV[0].capitalize' "$os"
+  fi
+}
+
+categorize_arch () {
+  local arch="$1"
+
+  ruby -e 'puts ARGV[0].upcase' "$arch"
```
ttaylorr: Hmm... I'm hesitant to introduce new Ruby dependencies in the Git LFS code base, since I think that it would be nice to remove it (in particular people who build Git LFS from source would like this, since we can build the documentation without Ronn).

I'm not a Ruby expert by any means (I'm barely familiar with it), but I think that this line could instead be released with:

```bash
echo "$arch" | tr -s a-z A-Z
```

script/upload
```diff
@@ -0,0 +1,262 @@
...
+}
+
+categorize_os () {
+  local os="$1"
+
+  if [ "$os" = "freebsd" ]
+  then
+    echo FreeBSD
+  else
+    ruby -e 'puts ARGV[0].capitalize' "$os"
```
ttaylorr: Likewise here, though I don't think that it's as straightforward for `#capitalize` than it is for `#upcase`.

script/upload
```diff
@@ -0,0 +1,262 @@
...
+}
+
+# Categorize a release asset and print its human readable name to standard
+# output.
+categorize_asset () {
+  local file="$1"
+  local os=$(echo "$file" | sed -e 's/^git-lfs-//' -e 's/[-.].*$//')
+  local arch=$(echo "$file" | ruby -pe '$_.gsub!(/\Agit-lfs-[^-]+-([^-]+)[-.].*/, "\\1")')
+
+  if ruby -e 'exit 1 unless /\Agit-lfs-v\d+\.\d+\.\d+\.tar\.gz\z/ =~ ARGV[0]' "$file"
```
ttaylorr: And the same here, but perhaps with a `case ... esac` statement?

script/upload
```diff
@@ -0,0 +1,262 @@
...
+    echo "text/plain"
+  fi
+}
+
+# Format the JSON for creating the release and print it to standard output.
+format_release_json () {
+  local version="$1"
+  local bodyfile="$2"
+
+  ruby -rjson -e 'puts JSON.generate({
```
ttaylorr: Hmmm. This one is tricky, too: again I think that it would be beneficial to not have a dependency on Ruby here, but we might be able to get out easily since we _do_ assume that `jq` is installed. I'm not sure of how to do this either, but it seems like it ought to be possible.

script/upload
```diff
@@ -0,0 +1,262 @@
...
+  curl https://api.github.com/repos/$REPO/releases | \
+    jq -r '.[] | select(.name == "'"$version"'") | .assets | .[] | .name' \
+    > "$WORKDIR/existing-assets"
+
+  for file in $(release_files "$version" | filter_files "$WORKDIR/existing-assets")
+  do
+    base=$(basename "$file")
+    desc=$(categorize_asset "$base")
+    ct=$(content_type "$base")
+    encbase=$(ruby -ruri -e 'print URI.escape(ARGV[0])' "$base")
```
ttaylorr: Can we use `curl(1)`'s `--data-urlencode` here instead?

Makefile
bk2204: Sure. The risk that we run doing that is that the user on the Unix system will attempt to run the Windows build process on their Unix system and it will fail if they don't already have the file copied from the Windows, but I guess that's an okay thing to do.

Makefile
bk2204: Okay, will revise.

script/changelog
bk2204: `revisions_in` sounds good.

script/upload
bk2204: No, I don't believe we can. `--data-urlencode` requires that we be passing those values as a body, not as a URL, as far as I understand. The body we're passing here is actually the file itself, so we need to have some way to encode the text.

script/upload
bk2204: This portion could probably be replaced. However, I should point out that this script is only designed to be run by a maintainer. End users will never want to run this script and any attempt to do so will fail due to lack of permissions. I think it's safe to assume that a reasonably modern Ruby is available on macOS and Linux systems, which are the systems on which Git LFS maintainers are likely to be doing releases.

As you've pointed out below, there are some tricky places where we really would like to produce something that shell scripting can't provide us natively. However, since we require `prove`, we can replace the Ruby with Perl instead, if you like. I felt Ruby would be more familiar to current and future maintainers and it wouldn't be a problem to use it since we already use `ronn`, but if you'd prefer Perl, I can go along with that.

script/upload
bk2204: As a further note, this is one place where core Perl no longer provides a built-in mechanism for escaping since `CGI.pm` has been removed, so if we switched to Perl here, we'd need to write something like `encbase=$(perl -e 'print $ARGV[0] =~ s/([^A-Za-z0-9])/sprintf "%%%02x", $1/gre;')`.

script/upload
ttaylorr: Hm, I think that you make an excellent point about this script being run by a maintainer, and hence its contents do not add to our dependency list. Given this, I think that the Ruby is fine to stick around without issue.

Makefile
bk2204: You actually pointed out a bug here, which is that the zip file should have the architecture in its name.

bk2204: I actually found several issues with this on a second glance. Because the code now uses the variables multiple times in a multi-command loop, I think it's better to keep it as a loop.

script/upload
bk2204: This does indeed read better as a case statement. I've update another function to use that approach as well.

bk2204:
Okay, this should address all of the outstanding issues and fix the issue where we serialized the wrong filename into the Windows archives.

viladoit015:
True

PR #3518, opened 2019-02-12
Retry on oversize file

### LINKED ISSUE #3494: LFS Client Error (InvalidArgumentError)
Hello,
trying to perform a git lfs fetch but it just hangs (0 B/s for a while) and then quit with 

`
LFS: Client error: https://api.media.atlassian.com/file/992a6467-031c-499d-841f-19711e45640d/binary
`

The command
 `
GIT_TRACE=1 GIT_CURL_VERBOSE=1 git lfs fetch
`
results in a long series of errors like this one below:

`
< HTTP/1.1 400 Bad Request                                                                              < Connection: close                                                                                     < Content-Length: 167                                                                                   < Access-Control-Allow-Origin: *                                                                        < Access-Control-Expose-Headers: Accept-Ranges, Content-Encoding, Content-Length, Content-Range         < Content-Type: application/json                                                                        < Date: Thu, 24 Jan 2019 15:14:19 GMT                                                                   < Strict-Transport-Security: max-age=15552000; includeSubDomains                                        < X-B3-Spanid: 13f3b59a8f95a973                                                                         < X-B3-Traceid: 13f3b59a8f95a973                                                                        < X-Content-Type-Options: nosniff                                                                       < X-Dns-Prefetch-Control: off                                                                           < X-Download-Options: noopen                                                                            < X-Frame-Options: SAMEORIGIN                                                                           < X-Xss-Protection: 1; mode=block                                                                       <                                                                                                       15:14:22.479535 trace git-lfs: HTTP: {"error":{"code":"InvalidArgumentError","title":"range first index  must not be greater than last index","href":"https://api.media.atlassian.com#InvalidArgumentError"}}   {"error":{"code":"InvalidArgumentError","title":"range first index must not be greater than last index" ,"href":"https://api.media.atlassian.com#InvalidArgumentError"}}15:14:22.479535 trace git-lfs: tq: retr ying object 77dfe7e27990ba66880d64d5423e862b37cb7461405cdc03789e3bf055a050e1: LFS: Client error: https: //api.media.atlassian.com/file/7392df22-bf24-4717-aa21-7efb92efa827/binary                              15:14:22.479535 trace git-lfs: tq: enqueue retry #1 for "77dfe7e27990ba66880d64d5423e862b37cb7461405cdc 03789e3bf055a050e1" (size: 122880044)                                                                   15:14:22.588244 trace git-lfs: filepathfilter: rewrite ".git" as "**/.git/**"                           15:14:22.588244 trace git-lfs: filepathfilter: rewrite "**/.git" as "**/.git"                           15:14:22.588244 trace git-lfs: filepathfilter: rejecting "tmp" via []                                   15:14:22.588244 trace git-lfs: filepathfilter: accepting "tmp"
`

I'd put this in evidence as should be relevant:
`
trace git-lfs: HTTP: {"error":{"code":"InvalidArgumentError","title":"range first index  must not be greater than last index","href":"https://api.media.atlassian.com#InvalidArgumentError"}
`

I suspected the service was down but all operational on the bitbucket status page.
What am I doing wrong?

bk2204:
If we're attempting to resume a download but the file we have on disk is the expected size or too large, we'll send an invalid Range header where the lower bound is greater than the upper bound. If this situation occurs, simply retry the download from the beginning, since we clearly don't have the expected data (or we wouldn't be resuming) and it isn't clear how else to recover.

This fixes #3494.

### DIFF
--- a/t/cmd/lfstest-gitserver.go
+++ b/t/cmd/lfstest-gitserver.go
@@ -749,6 +749,37 @@ func storageHandler(w http.ResponseWriter, r *http.Request) {
 					byteLimit = 8
 					batchResumeFailFallbackStorageAttempts++
 				}
+			} else if string(by) == "status-batch-retry" {
+				if rangeHdr := r.Header.Get("Range"); rangeHdr != "" {
+					regex := regexp.MustCompile(`bytes=(\d+)\-(.*)`)
+					match := regex.FindStringSubmatch(rangeHdr)
+					// We have a Range header with two
+					// non-empty values.
+					if match != nil && len(match) > 2 && len(match[2]) != 0 {
+						first, _ := strconv.ParseInt(match[1], 10, 32)
+						second, _ := strconv.ParseInt(match[2], 10, 32)
+						// The second part of the range
+						// is smaller than the first
+						// part (or the latter part of
+						// the range is non-integral).
+						// This is invalid; reject it.
+						if second < first {
+							w.WriteHeader(400)
+							return
+						}
+						// The range is valid; we'll
+						// take the branch below.
+					}
+					// We got a valid range header, so
+					// provide a 206 Partial Content. We
+					// ignore the upper bound if one was
+					// provided.
+					if match != nil && len(match) > 1 {
+						statusCode = 206
+						resumeAt, _ = strconv.ParseInt(match[1], 10, 32)
+						w.Header().Set("Content-Range", fmt.Sprintf("bytes %d-%d/%d", resumeAt, len(by), resumeAt-int64(len(by))))
+					}
+				}
 			}
 			w.WriteHeader(statusCode)
 			if byteLimit > 0 {
--- a/t/t-resume-http-range.sh
+++ b/t/t-resume-http-range.sh
@@ -85,3 +85,44 @@ begin_test "resume-http-range-fallback"
 )
 end_test
 
+begin_test "resume-http-range-retry"
+(
+  set -e
+
+  reponame="resume-http-range-retry"
+  setup_remote_repo "$reponame"
+
+  clone_repo "$reponame" $reponame
+
+  git lfs track "*.dat" 2>&1 | tee track.log
+  grep "Tracking \"\*.dat\"" track.log
+
+  # This string announces to server that we want a test that strictly handles
+  # Range headers, rejecting any where the latter part of the range is smaller
+  # than the former part.
+  contents="status-batch-retry"
+  contents_oid=$(calc_oid "$contents")
+
+  printf "%s" "$contents" > a.dat
+  git add a.dat
+  git add .gitattributes
+  git commit -m "add a.dat" 2>&1 | tee commit.log
+  git push origin master
+
+  assert_server_object "$reponame" "$contents_oid"
+
+  # Delete local copy then fetch it back.
+  rm -rf .git/lfs/objects
+  refute_local_object "$contents_oid"
+
+  # Create a partial corrupt object.
+  mkdir .git/lfs/incomplete
+  printf "%s" "${contents/st/aa}" >".git/lfs/incomplete/$contents_oid.tmp"
+
+  # The first download may fail with an error; run a second time to make sure
+  # that we detect the corrupt file and retry.
+  GIT_TRACE=1 git lfs fetch 2>&1 | tee fetchresume.log
+  GIT_TRACE=1 git lfs fetch 2>&1 | tee fetchresume.log
+  assert_local_object "$contents_oid" "${#contents}"
+)
+end_test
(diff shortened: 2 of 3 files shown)


t/t-resume-http-range.sh
```diff
@@ -85,3 +85,42 @@ begin_test "resume-http-range-fallback"
...
+  # server will abort the transfer mid way (so will error) when not resuming
+  # then we can restart it
+  rm -rf .git/lfs/objects
+  refute_local_object "$contents_oid"
+
+  mkdir .git/lfs/incomplete
+  printf "%s" "${contents/st/aa}" >".git/lfs/incomplete/$contents_oid.tmp"
+
+  GIT_TRACE=1 git lfs fetch 2>&1 | tee fetchresume.log
+  GIT_TRACE=1 git lfs fetch 2>&1 | tee fetchresume.log
```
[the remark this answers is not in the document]
bk2204: Yeah, that's the idea. I probably need to add a comment to that effect here.

I'll add some assertions that we've deleted the problematic file and ensure that the new file has the correct hash.

t/cmd/lfstest-gitserver.go
```diff
@@ -749,6 +749,24 @@ func storageHandler(w http.ResponseWriter, r *http.Request) {
 					byteLimit = 8
 					batchResumeFailFallbackStorageAttempts++
 				}
+			} else if len(by) == len("status-batch-retry") && string(by) == "status-batch-retry" {
```
ttaylorr: We may be copying what's already here, but is there a need to check the string both for string and length equality? Shouldn't the later follow from the former?

t/cmd/lfstest-gitserver.go
```diff
@@ -749,6 +749,24 @@ func storageHandler(w http.ResponseWriter, r *http.Request) {
...
+				if rangeHdr := r.Header.Get("Range"); rangeHdr != "" {
+					regex := regexp.MustCompile(`bytes=(\d+)\-(.*)`)
+					match := regex.FindStringSubmatch(rangeHdr)
+					if match != nil && len(match) > 2 && len(match[2]) != 0 {
+						first, _ := strconv.ParseInt(match[1], 10, 32)
+						second, _ := strconv.ParseInt(match[2], 10, 32)
+						if second < first {
```
ttaylorr: Do you think it would be worth adding a comment indicating what it means to be in this conditional?

t/cmd/lfstest-gitserver.go
```diff
@@ -749,6 +749,24 @@ func storageHandler(w http.ResponseWriter, r *http.Request) {
...
+							w.WriteHeader(400)
+							return
+						}
+					}
+					if match != nil && len(match) > 1 {
```
ttaylorr: Likewise in this conditional?

t/t-resume-http-range.sh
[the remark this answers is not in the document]
ttaylorr: I'm not sure that that behavior is worth testing explicitly. I think that what we care about at the integration level is that the fetch was executed successfully (by however that fetch actually works).

Testing an implementation like deleting the file in question makes such a test brittle in my mind.

t/t-resume-http-range.sh
```diff
@@ -85,3 +85,42 @@ begin_test "resume-http-range-fallback"
...
+  setup_remote_repo "$reponame"
+
+  clone_repo "$reponame" $reponame
+
+  git lfs track "*.dat" 2>&1 | tee track.log
+  grep "Tracking \"\*.dat\"" track.log
+
+  # this string announces to server that we want a test that
+  # interrupts the transfer when started from 0 to cause resume
+  contents="status-batch-retry"
```
ttaylorr: Thanks for documenting this. When I first explored these tests, I was often confused by why these magic strings caused certain things to happen; this is really helpful for similar newcomers.

bk2204: I actually need to revise this, since this comment is pulled in from another test assertion. But I will replace it with a hopefully equally helpful comment.

t/cmd/lfstest-gitserver.go
bk2204: I think this was an optimization (perform the cheap check first), but I'm happy to just remove it and switch to a string comparison. I expect Go does the same comparison internally anyway.

t/cmd/lfstest-gitserver.go
bk2204: Sure, I can add comments for these conditionals.

t/t-resume-http-range.sh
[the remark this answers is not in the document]
bk2204: I've decided to leave it as it is and add some comments to help make the situation a bit easier to understand.

bk2204:
I think I've addressed all of the concerns here. There are now copious correct comments, and I've removed the “optimization” which Go likely already performs.

PR #3534, opened 2019-02-19
Avoid nil pointer dereference on unexpected failure

### LINKED ISSUE #3533: [2.7.0] panic: runtime error: invalid memory address or nil pointer dereference
```
$ git version
git version 2.20.1.windows.1
$ git lfs version
git-lfs/2.7.0 (GitHub; windows amd64; go 1.11.5; git 3b3fa90b)
```

For around a month now, I have been regularly pushing to GitHub and Azure Repos around 25 repos migrated to Git LFS.  I used to use 2.6.1 and I had never noticed such crash.
Yesterday, I updated Git LFS from 2.6.1 to 2.7.0.

While pushing around 25 repos freshly migrated to Git LFS, using 2.7.0, the following error occurred 3-4 times:

```
[signal 0xc0000005 code=0x0 addr=0x10 pc=0x74f0cf]

goroutine 555 [running]:
github.com/git-lfs/git-lfs/tq.(*basicUploadAdapter).DoTransfer(0xc000826088, 0x0, 0x0, 0xc00084fe00, 0xc0002e7470, 0x0, 0x0, 0x0)
        C:/Users/brian m. carlson/checkouts/git-lfs/tq/basic_upload.go:126 +0x70f
github.com/git-lfs/git-lfs/tq.(*adapterBase).worker(0xc0003e61b0, 0x3, 0x0, 0x0)
        C:/Users/brian m. carlson/checkouts/git-lfs/tq/adapterbase.go:177 +0x54b
created by github.com/git-lfs/git-lfs/tq.(*adapterBase).Begin
        C:/Users/brian m. carlson/checkouts/git-lfs/tq/adapterbase.go:90 +0x2b4
error: failed to push some refs to 'https://XXX'
```

Restarted push completed successfully. So, I don't think this is related to any particular repo or repo size.

I haven't managed to reproduce the crash while running with `GIT_TRACE=1` yet. I will update this report as soon as I get any more details.

bk2204:
When we get an error making an API request, we check to see if the response is a 429, and if so, we attempt to back off and retry later. However, we failed to check if we actually had a response; if the response was missing (say, because we had a network issue), then we would dereference a nil pointer.

Instead, let's explicitly check for a non-nil response before querying it for a status code.

This series doesn't include a test because I'm not aware of a way to synthesize this failure. It clearly is occurring, though, and it seems fairly obvious what's going on.

Fixes #3533.

/cc @mloskot

### DIFF
--- a/tq/basic_upload.go
+++ b/tq/basic_upload.go
@@ -123,6 +123,12 @@ func (a *basicUploadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb Progres
 			err = errors.Wrap(err, perr.Error())
 		}
 
+		if res == nil {
+			// We encountered a network or similar error which caused us
+			// to not receive a response at all.
+			return errors.NewRetriableError(err)
+		}
+
 		if res.StatusCode == 429 {
 			retLaterErr := errors.NewRetriableLaterError(err, res.Header["Retry-After"][0])
 			if retLaterErr != nil {


mloskot:
Thanks for fixing it so quickly!

tq/basic_upload.go
```diff
@@ -123,7 +123,7 @@ func (a *basicUploadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb Progres
 			err = errors.Wrap(err, perr.Error())
 		}
 
-		if res.StatusCode == 429 {
+		if res != nil && res.StatusCode == 429 {
```
ttaylorr: I think that this change is safe, since it's the only place in this block that accesses `res`, and the block returns regardless on either L129 or L132. To harden this change for the future, it might be good to make sure that we can add other checks that involve dereferencing `res`, without having to repeat the `if res != nil` bits.

What do you think about the following?

```suggestion
		if res != nil {
			// We encountered a network or similar error which caused us
			// to not receive a response at all.
			return errors.NewRetriableError(err)
		}
		
		if res.StatusCode == 429 {
```

bk2204: Yeah, that sounds like a good idea, except I think we want to check for `res == nil`. I'll push up a change with that.

dscho:
@bk2204 do you intend to release Git LFS v2.7.1 before Git v2.21.0 comes out? I ask because I already integrated v2.7.0 into Git for Windows, and one user suggested to downgrade Git LFS in Git for Windows rather than shipping v2.21.0 with v2.7.0...

bk2204:
My tentative plans are to do a 2.7.1 release this Friday. I'll need to fix the other side of this (which I just realized we haven't fixed) in #3537.

mloskot:
@dscho The user who suggested the downgrade was me 😄

dscho:
Sounds good, @bk2204!

@mloskot hehe, I totally missed that you were in this PR thread, too ;-)

bk2204:
@dscho Just FYI, I will likely do the release come Monday or Tuesday due to the fact that I just got some build-related patches done, which means the earliest I can do the PR for the release itself is tomorrow.

dscho:
@bk2204 thanks for letting me know!

PR #3537, opened 2019-02-20
Avoid nil pointer dereference on download failure

bk2204:
In 4775f9c1 ("tq: avoid nil pointer dereference on unexpected failure", 2019-02-19), we fixed an issue on uploads where we could dereference a nil http.Response pointer when uploading. We should also check for this when downloading. Since the check a few lines below is now redundant, remove it.


Desires62:
boostrap checked array and hookjed over hacth map

PR #3539, opened 2019-02-21
Ensure trimpath works properly on Windows

bk2204:
We recently added code to trim paths out of the binary to aid in reproducible builds. However, our code had a bug: since there are multiple sets of double quotes in the build invocation, if the home directory contained spaces, as it often does on Windows, the build would fail because the extra unquoted spaces confused the compiler.

To address this issue, remove the extra double quotes.


bk2204:
Actually, this doesn't work properly in all cases on Windows. I need to do some additional work, so I'll reopen a new PR once I fix all of the cases.

PR #3545, opened 2019-02-25
Don't set -extldflags unless LDFLAGS has a value

hartzell:
Unconditionally setting `-extldflags` to the contents of `LDFLAGS`
causes problems when `LDFLAGS` isn't defined (`-extldflags` doens't
have an corresponding value).

In out case this led to `git-lfs` aborting.

```
$ ./bin/git-lfs
Aborted
```

This change only sets `-extldflags` if `LDFLAGS` is defined.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,9 @@ BUILTIN_LD_FLAGS += -w
 endif
 # EXTRA_LD_FLAGS are given by the caller, and are passed to the Go linker after
 # BUILTIN_LD_FLAGS are processed. By default the system LDFLAGS are passed.
+ifdef LDFLAGS
 EXTRA_LD_FLAGS ?= -extldflags ${LDFLAGS}
+endif
 # LD_FLAGS is the union of the above two BUILTIN_LD_FLAGS and EXTRA_LD_FLAGS.
 LD_FLAGS = $(BUILTIN_LD_FLAGS) $(EXTRA_LD_FLAGS)
 


bk2204:
Hey, thanks for the patch and welcome to Git LFS!

I think this looks obviously correct. I'm not sure why we abort in this case; it's likely due to something in Go itself on whatever version you're building on. I will admit I'm curious as to why this fails for you and what version of Go you're building against, but that shouldn't prevent us from merging this as it stands.

It looks like from the repo you've linked, that you might be building against Go 1.5. Is that correct? If so, it might be worth using a newer version, since we generally only support the latest version of Go and don't test against older versions.

hartzell:
We use the [Spack package manager](https://spack.io) to build a variety of things.  We recently noticed that we had builds of `git-lfs` that didn't work.  Sometimes.  Dig, dig, dig and I ended up with this.

We're building on minimal-ish CentOS 7 systems (not very many packages installed).

I'm running a fairly recent version of `go`, I'm not sure what you see that makes you think `go@1.5`:

```consolesession
$ which go
~/spack/opt/spack/linux-centos7-x86_64/gcc-8.2.0/go-1.11.5-pei4g2iupo6y2yvogg4vde2poix2ho7d/bin/go
$ go version
go version go1.11.5 linux/amd64
[ghartzell@bifx1n03 git-lfs]$
```

I'm tracking this discussion over here: https://github.com/spack/spack/issues/10702

bk2204:
> I'm running a fairly recent version of `go`, I'm not sure what you see that makes you think `go@1.5`:

Ah, I saw the build-dependency listed [in this file](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/git-lfs/package.py) and thought it might have been an old version of Go.

I'm going to merge this, and if we do a 2.7.2 release, I'll make sure we include it.

hartzell:
At, [that build dependency](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/git-lfs/package.py#L22) is `go@1.5` **or later**.  The spack "concretizer" (SAT solver) will put together a graph that satisfies the various requirements/constraints, generally preferring the newest version that it knows about.

PR #3549, opened 2019-03-04
Switch from manually running go generate to using dh-golang to run it

### LINKED ISSUE #2013: No usage text found after bash script .deb installation
I just installed a deb package with the bash script on Ubuntu 16.04 and:
`git lfs help`
gives:
`Sorry, no usage text found for "git-lfs"`

However,  `git  lfs --help` opens man pages as expected.

ssgelm:
This fixes #2013.  For some reason I could only reproduce the issue from within the docker builder, not when running outside of docker.  Regardless, this appears to fix the problem and is a generally cleaner
solution anyway that will continue to work if any other generate commands are added to git-lfs.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -18,6 +18,7 @@ BUILD_DIR := obj-$(DEB_HOST_GNU_TYPE)
 export DH_GOPKG := github.com/git-lfs/git-lfs
 # DH_GOLANG_EXCLUDES typically incorporates vendor exclusions
 export DH_GOLANG_EXCLUDES := test github.com/olekukonko/ts/* github.com/xeipuuv/* github.com/spf13/cobra/* github.com/kr/* github.com/pkg/errors github.com/alexbrainman/sspi/*
+export DH_GOLANG_GO_GENERATE := 1
 export PATH := $(CURDIR)/$(BUILD_DIR)/bin:$(PATH)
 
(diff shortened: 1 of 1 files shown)


ssgelm:
Also, should there be a test to confirm that there is help text?  Seems like this issue has crept up a few times so I'd imagine there should be tests.  I wasn't sure where to add them though so I did not write them.

bk2204:
Thanks for fixing this. If you wanted to add a test, you could add it to `t/t-happy-path.sh`, or simply write a new test file. If not, I'm happy to take this as it is, and I can write up a test file for the future.

ssgelm:
Yeah, I'd like to but I don't think I have time right now.  Can you take this for now and I can try and add one in the future?

bk2204:
Absolutely.

ssgelm:
Thanks for that!  You might also want to consider rebuilding 2.7.1 and uploading fixed debs.

bk2204:
I don't think I can upload over an existing object. I hit that every time I build ronn for CentOS 6, when I have to delete the Ruby RPMs we build so they don't fail the upload.

However, we have one or two things which could potentially go into a 2.7.2, so we may do such a release, and if so, I'll include this.

ssgelm:
Yeah, no worries, just a thought.  Deleting and re-uploading would likely manage to cause more problems anyway. 😄

Issue #3554, opened 2019-03-07
git lfs lock works incorrectly if the CD contains a junction, and the lock can't be removed


JFLarvoire:
Using Git for Windows v2.21.0.
In my root directory, I have a C:\JFL junction that points to my home directory C:\Users\Larvoire\Documents. To save typing, I usually use that junction to quickly change the current directory to one of my sources directories.

```
C:\JFL\Temp>dir C:\ | findstr JFL
2018-02-13  11:01    <JUNCTION>     JFL [C:\Users\Larvoire\Documents]

C:\JFL\Temp>cd C:\JFL\SRC\HPSD\SAFcollector4\Collectors\UserGuide

C:\JFL\SRC\HPSD\SAFcollector4\Collectors\UserGuide>
```

I ran the commands below, and now I can't get rid of the lock:

```
C:\JFL\SRC\HPSD\SAFcollector4\Collectors\UserGuide>git lfs lock SAFcollectorUserGuide.docx
Locked ../../../../Users/Larvoire/Documents/SRC/HPSD/SAFcollector4/Collectors/UserGuide/SAFcollectorUserGuide.docx

C:\JFL\SRC\HPSD\SAFcollector4\Collectors\UserGuide>git lfs locks
../../../../Users/Larvoire/Documents/SRC/HPSD/SAFcollector4/Collectors/UserGuide/SAFcollectorUserGuide.docx     jf-larvoire     ID:11

C:\JFL\SRC\HPSD\SAFcollector4\Collectors\UserGuide>git lfs unlock SAFcollectorUserGuide.docx
Git status failed: exit status 128

C:\JFL\SRC\HPSD\SAFcollector4\Collectors\UserGuide>git lfs unlock ../../../../Users/Larvoire/Documents/SRC/HPSD/SAFcollector4/Collectors/UserGuide/SAFcollectorUserGuide.docx
Unable to determine path: FindFirstFile C:\Users\Larvoire\Documents\SRC\Users\Larvoire\Documents\SRC\HPSD\SAFcollector4\Collectors\UserGuide\SAFcollectorUserGuide.docx: The system cannot find the path specified.

C:\JFL\SRC\HPSD\SAFcollector4\Collectors\UserGuide>git lfs unlock -i 11
Git status failed: exit status 128

C:\JFL\SRC\HPSD\SAFcollector4\Collectors\UserGuide>git lfs locks
[...]

PR #3560, opened 2019-03-14
Handle Windows junctions properly

bk2204:
On Windows, there is a type of file system object called a junction point, which is a by-name reference from the junction point to a directory. Go's filepath.EvalSymlinks, which we use in the locking code, resolves these and provides a canonical path. However, Git for Windows, which provides the repository root, does not resolve them, resulting in the relative paths we generate against the repository root containing ".." components. As a consequence, the locks generated in this way are broken and cannot be unlocked.

To avoid this problem, we canonicalize the paths we get back from Git to ensure that all lock paths are relative. In addition, we check for locks containing ".." and reject them, and we allow users to force unlocking by ID, even if we think that will fail. If it does indeed fail, we are no worse off than before, and if it does not, then the user is able to clean up their broken locks.

Note that we have no tests for these cases because the only way to generate this issue is with junctions, which are not able to be created except with the mklink CMD built-in, which is not available in the bash-based environment we use.

/cc #3554 and git-for-windows/git#2114
/cc @JFLarvoire as reporter

### DIFF
--- a/commands/command_lock.go
+++ b/commands/command_lock.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/git-lfs/git-lfs/errors"
 	"github.com/git-lfs/git-lfs/git"
@@ -91,6 +92,9 @@ func lockPath(file string) (string, error) {
 	}
 
 	path = filepath.ToSlash(path)
+	if strings.HasPrefix(path, "../") {
+		return "", fmt.Errorf("lfs: unable to canonicalize path %q", path)
+	}
 
 	if stat, err := os.Stat(abs); err != nil {
 		return "", err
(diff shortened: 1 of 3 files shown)


commands/command_lock.go
```diff
@@ -91,6 +92,9 @@ func lockPath(file string) (string, error) {
 	}
 
 	path = filepath.ToSlash(path)
+	if strings.HasPrefix(path, "../") {
+		return "", fmt.Errorf("lfs: unable to canonicalize path %q", path)
```
PastelMobileSuit: Can the `../` component occur elsewhere in the path, or can it only be the first element?

bk2204: I believe it can only be the first element. We produce a path relative to the repository root. If the path is under the repository root (by whatever metric), it won't contain a `..` component, since we'll have canonicalized both paths already. If it isn't, the first component will contain `..`, since we'll need to leave the repository root.

I considered checking in the middle of the path as well, but it's not strictly necessary and there isn't a simple way to check the entire path without also catching weird paths like `/home/foo/path-ending-in-two-dots../`, which I feel confident someone, somewhere, would complain about.

PR #3455, opened 2019-01-02
Issue 3252: git-lfs locks should optionally denote own locks

mstrap:

PR #3569, opened 2019-03-19
Issue 3252: git-lfs locks should optionally denote own locks

mstrap:
This is the same pull request as #3455 which I had closed unfortunately (by temporarily deleting my feature branch).

### DIFF
--- a/docs/man/git-lfs-locks.1.ronn
+++ b/docs/man/git-lfs-locks.1.ronn
@@ -29,6 +29,16 @@ Lists current locks from the Git LFS server.
   last known locks in case you are offline. There is no guarantee that locks
   on the server have not changed in the meanwhile.
 
+* `--verify`:
+  Verifies the lock owner on the server and marks our own locks by 'O'.
+  Own locks are actually held by us and corresponding files can be updated for
+  the next push. All other locks are held by someone else.
+  Contrary to --local, this option will also detect locks which are held by us
+  despite no local lock information being available (e.g. because the file had
+  been locked from a different clone);
+  it will also detect 'broken' locks (e.g. if someone else has forcefully
+  unlocked our files).
+
 * `-l <num>` `--limit=<num>`:
   Specifies number of results to return.
 
--- a/locking/locks.go
+++ b/locking/locks.go
@@ -3,6 +3,7 @@ package locking
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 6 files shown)


mstrap:
@git-lfs/core can you please take a look? It's the same pull request as #3455. The Travis CI build failure seems to be unrelated to my changes.

locking/locks.go
```diff
@@ -218,100 +219,108 @@ func (c *Client) SearchLocks(filter map[string]string, limit int, localOnly bool
...
 	theirLocks = make([]Lock, 0, limit)
-	body := &lockVerifiableRequest{
-		Ref:   &lockRef{Name: ref.Refspec()},
-		Limit: limit,
-	}
 
-	c.cache.Clear()
+	if cached {
+		if limit != 0 {
+			return []Lock{}, []Lock{}, errors.New("can't search cached locks when limit is set")
```
bk2204: Is there a reason we use `[]Lock{}` instead of `nil`? I think it's customary to use `nil` as the zero value for slices when returning errors.

locking/locks.go
```diff
@@ -218,100 +219,108 @@ func (c *Client) SearchLocks(filter map[string]string, limit int, localOnly bool
...
-		if res != nil {
-			switch res.StatusCode {
-			case http.StatusNotFound, http.StatusNotImplemented:
-				return ourLocks, theirLocks, errors.NewNotImplementedError(err)
-			case http.StatusForbidden:
-				return ourLocks, theirLocks, errors.NewAuthError(err)
-			}
+		locks := &lockVerifiableList{}
+		err := c.readLocksFromCacheFile("verifiable", func(decoder *json.Decoder) error {
+			locks = &lockVerifiableList{}
```
bk2204: Why do we need to set this value to an empty value before decoding?

docs/man/git-lfs-locks.1.ronn
```diff
@@ -29,6 +29,16 @@ Lists current locks from the Git LFS server.
   last known locks in case you are offline. There is no guarantee that locks
   on the server have not changed in the meanwhile.
 
+* `--verify`:
+  Verifies the lock owner on the server and marks our own locks by 'O'.
+  Own locks are actually hold by us and corresponding files can be updated for
```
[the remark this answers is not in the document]
mstrap: Thanks, fixed.

locking/locks.go
mstrap: It seems that the lock-related functions work differently here: they will always return results collected up to the point where the error has occurred. For instance, `VerifiableLocks` (before any of my changes from last year) return `ourLocks, theirLocks, err` with `ourLocks` and `theirLocks` being non-`nil` always:

https://github.com/git-lfs/git-lfs/blob/8b41ba5cb9bbff6564f2c75c8c5050cee17707ba/locking/locks.go#L238-L240

This somehow seems reasonable because this will give at least all results which could be collected until a (possible connection) error has occurred.

For the particular piece of code, we could change it to `nil, nil, errors.New(...)` but this requires to introduce some more checks in calling methods. Hence, I would prefer to leave as is.

locking/locks.go
mstrap: That actually redundant. Fixed. Thanks!

mstrap:
@bk2204, thanks for reviewing. Do you prefer to have individual "fixup!" commits added or should I rewrite the feature branch by amending/squashing existing commits?

bk2204:
@mstrap, you're welcome to do either. If you write individual `fixup!` commits, you'd have to squash before merging, but I'm okay with either. What I tend to do depends on how complex my branch is and how many fixups I plan to have; in this case, I'd probably just squash and amend.

locking/locks.go
bk2204: Okay, that seems fine.

mstrap:
@ttaylorr may I ask you to give this PR a review, too? It complements the (remote) lock handling improvements which I had started with issue #3253.

PR #3584, opened 2019-03-28
Properly handle config options for URLs with upper case letters

### LINKED ISSUE #3571: LFS 2.7.1 doesn't include extra header from git config
Steps to reproduce:
```
GIT_LFS_SKIP_SMUDGE=1 git clone https://account@dev.azure.com/account/Test/_git/mitesch-lfs mitesch-lfs
cd mitesch-lfs
git config http.https://account@dev.azure.com/account/Test/_git/mitesch-lfs.extraheader "AUTHORIZATION: bearer notvalid"
export GIT_CURL_VERBOSE=1
git lfs fetch
```

In 2.6.1, it includes the authorization header in the requests made by `git lfs fetch`.
In 2.7.1, the header is not included.

We (Azure DevOps) optionally set the bearer token in the config for users during builds so that later steps can perform authenticated git commands. Our upgrade to 2.7.1 on the build machines breaks this functionality for LFS. 

OS: Windows 10
Git: 2.20.1

bk2204:
Git configuration options have a complicated situation with regard to case. For the most part, they are case-insensitive: you may write any case into the file, but Git interprets it as lower case. However, there are exceptions.

Because the middle component of a three-component option can be a URL, branch name, or remote name, this component (and only this component) is treated as case sensitive. Since this component can be a URL, which may (and, due to the ".git" portion, usually does) contain a dot, the first component of the config option is read up until the first dot, and the last component is read from the end to the last dot.

When git config writes a value into the file, it preserves the case the user has provided, and when it prints the config values, it canonicalizes the keys by folding the case-insensitive portions to lowercase. Git LFS then reads this canonicalized form in as our source of config options, relieving us from having to parse the files ourselves.

However, when we read this data in, we forced the key to lowercase, and when we looked up a key, we also forced the key we were looking up to lowercase. While this behavior was wrong (since URLs, at least, are case-sensitive), it did happen to mostly work if the configuration didn't contain settings for two URLs differing in case.

In the 2.7.0 cycle, we changed the way we did URL config lookups to match the way Git does them. Previously, we performed configuration lookups on several possible keys (forcing them to lower case, URL portion included) and took the first that matched. Now, we directly compare the URL we're connecting to (which may be in mixed case) to the values we got in the configuration (which we've forced to lowercase). Consequently, we will fail to find configuration options for a mixed-case URL, resulting in things like "http.extraHeader" not being used.

Fix this issue by letting Git do the canonicalization of configuration keys for us instead of lowercasing them ourselves and then canonicalizing the key when looking it up in the table. Add tests for this behavior with "http.extraHeader" in the integration suite and several canonicalization assertions in the unit tests. Update several tests to use the canonical version of the data in their test data stores.

Fixes #3571
/cc @mitesch and @arche89 as reporters

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -13,7 +13,7 @@ func TestRemoteDefault(t *testing.T) {
 	cfg := NewFrom(Values{
 		Git: map[string][]string{
 			"branch.unused.remote":     []string{"a"},
-			"branch.unused.pushRemote": []string{"b"},
+			"branch.unused.pushremote": []string{"b"},
 		},
 	})
 	assert.Equal(t, "origin", cfg.Remote())
@@ -24,7 +24,7 @@ func TestRemoteBranchConfig(t *testing.T) {
 	cfg := NewFrom(Values{
 		Git: map[string][]string{
 			"branch.master.remote":    []string{"a"},
-			"branch.other.pushRemote": []string{"b"},
+			"branch.other.pushremote": []string{"b"},
 		},
 	})
 	cfg.ref = &git.Ref{Name: "master"}
@@ -37,8 +37,8 @@ func TestRemotePushDefault(t *testing.T) {
 	cfg := NewFrom(Values{
 		Git: map[string][]string{
 			"branch.master.remote":    []string{"a"},
-			"remote.pushDefault":      []string{"b"},
-			"branch.other.pushRemote": []string{"c"},
+			"remote.pushdefault":      []string{"b"},
+			"branch.other.pushremote": []string{"c"},
 		},
 	})
 	cfg.ref = &git.Ref{Name: "master"}
@@ -51,8 +51,8 @@ func TestRemoteBranchPushDefault(t *testing.T) {
 	cfg := NewFrom(Values{
 		Git: map[string][]string{
 			"branch.master.remote":     []string{"a"},
-			"remote.pushDefault":       []string{"b"},
-			"branch.master.pushRemote": []string{"c"},
+			"remote.pushdefault":       []string{"b"},
+			"branch.master.pushremote": []string{"c"},
 		},
 	})
 	cfg.ref = &git.Ref{Name: "master"}
--- a/config/git_fetcher.go
+++ b/config/git_fetcher.go
@@ -32,7 +32,10 @@ func readGitConfig(configs ...*git.ConfigurationSource) (gf *GitFetcher, extensi
 			}
 
 			allowed := !gc.OnlySafeKeys
-			key, val := strings.ToLower(pieces[0]), pieces[1]
+
+			// We don't need to change the case of the key here,
+			// since Git will already have canonicalized it for us.
+			key, val := pieces[0], pieces[1]
 
 			if origKey, ok := uniqKeys[key]; ok {
 				if ShowConfigWarnings && len(vals[key]) > 0 && vals[key][len(vals[key])-1] != val && strings.HasPrefix(key, gitConfigWarningPrefix) {
@@ -114,7 +117,8 @@ func readGitConfig(configs ...*git.ConfigurationSource) (gf *GitFetcher, extensi
 // empty string and false will be returned, signaling that the value was
 // absent.
 //
-// Map lookup by key is case-insensitive, as per the .gitconfig specification.
+// Map lookup by key is case-insensitive, except for the middle part of a
+// three-part key, as per the .gitconfig specification.
 //
 // Get is safe to call across multiple goroutines.
 func (g *GitFetcher) Get(key string) (val string, ok bool) {
@@ -130,7 +134,7 @@ func (g *GitFetcher) GetAll(key string) []string {
 	g.vmu.RLock()
 	defer g.vmu.RUnlock()
 
-	return g.vals[strings.ToLower(key)]
+	return g.vals[g.caseFoldKey(key)]
 }
 
 func (g *GitFetcher) All() map[string][]string {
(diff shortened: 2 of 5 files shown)


config/git_fetcher.go
```diff
@@ -148,6 +149,23 @@ func (g *GitFetcher) All() map[string][]string {
...
+	// URL, it may have dots in it. We'll downcase the part before the first
+	// dot and after the last dot, but preserve the piece in the middle,
+	// which may be a branch name, remote, or URL, all of which are
+	// case-sensitive. This is the algorithm Git uses to canonicalize its
+	// keys.
+	if len(parts) < 3 {
+		return strings.ToLower(key)
+	}
+
+	return fmt.Sprintf("%s.%s.%s", strings.ToLower(parts[0]), strings.Join(parts[1:last], "."), strings.ToLower(parts[last]))
```
ttaylorr: One alternative option here (which might--or might not be--be clearer than `%s.%s.%s`) is:

```go
return strings.Join([]string{
        strings.ToLower(parts[0]),
        strings.Join(parts[1:last], "."),
        strings.ToLower(parts[last]),
}, ".")
```

But, I'm not convinced that one is better than the other overall. Totally your call on which you'd like to adopt.

config/git_fetcher.go
```diff
@@ -32,7 +32,7 @@ func readGitConfig(configs ...*git.ConfigurationSource) (gf *GitFetcher, extensi
 			}
 
 			allowed := !gc.OnlySafeKeys
-			key, val := strings.ToLower(pieces[0]), pieces[1]
+			key, val := pieces[0], pieces[1]
```
ttaylorr: I think that a comment here is worthwhile. Something to the effect of: "don't re-case the incoming configuration key/value pairs from Git, since it is already done for us correctly".

config/config_test.go
```diff
@@ -13,7 +13,7 @@ func TestRemoteDefault(t *testing.T) {
 	cfg := NewFrom(Values{
 		Git: map[string][]string{
 			"branch.unused.remote":     []string{"a"},
-			"branch.unused.pushRemote": []string{"b"},
```
ttaylorr: One thing that we could do to harden the use of `config.Values` throughout these tests against adding _new_ non-URL keys containing uppercase characters is to warn about it during construction, something like:

```diff
diff --git a/config/config.go b/config/config.go
index 6e38c595..112bf972 100644
--- a/config/config.go
+++ b/config/config.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode"
 
 	"github.com/git-lfs/git-lfs/fs"
 	"github.com/git-lfs/git-lfs/git"
@@ -155,6 +156,13 @@ func NewFrom(v Values) *Configuration {
 			}
 
 			for key, values := range v.Git {
+				parts := strings.Split(key, ".")
+				isUrl := len(parts) >= 3
+				hasUpper := strings.IndexFunc(key, unicode.IsUpper) > -1
+
+				if !isUrl && hasUpper {
+					fmt.Fprintf(os.Stderr, "key %q has uppercase, shouldn't", key)
+				}
 				for _, value := range values {
 					fmt.Printf("Config: %s=%s\n", key, value)
 					source.Lines = append(source.Lines, fmt.Sprintf("%s=%s", key, value))
```

I'm not totally sold on the above, though, since it doesn't _really_ prevent us from breaking things in the future, it just prints some extra bits out to stderr.

Something that we could do to go even further, would be to teach the Values constructor to take a `*testing.T` instance, and call `Fail()` on it, instead of this. What do you think?

config/git_fetcher_test.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetCanonicalization(t *testing.T) {
+	vals := map[string][]string{
+		"user.name":                   []string{"Pat Doe"},
+		"branch.MixedCase.pushremote": []string{"Somewhere"},
```
ttaylorr: Here's a compelling case where we _wouldn't_ want the warning about uppercase keys... maybe we don't want it at all?

config/git_fetcher.go
bk2204: Yeah, that would in fact be tidier, I think. I'll make that change.

config/git_fetcher.go
bk2204: Sure, I can do that.

config/config_test.go
bk2204: I don't think the print to stderr will be effective, since I think Go suppresses most output if the test is successful. Of course, if we use mixed case, the test probably won't be successful.

I'm a bit negative on passing test contexts into non-test code, personally. It will likely bring dependencies into the binary that we don't need, and it seems a bit unhygienic. However, if we know that this is always an error (and it certainly looks like that's the case), we could just call `panic(fmt.Sprintf("key %q has uppercase, shouldn't", key))`, which will certainly alert the developer making this mistake. What would you think about that?

config/git_fetcher_test.go
bk2204: Ah, but in your code, we never take the branch that prints the warning, because we have at least three components. I don't think this case makes a compelling argument for or against your proposed solution.

config/config_test.go
ttaylorr: Ah, your approach is a good way forward. This suggestion came from a (now outdated) understanding that the `struct Values` was test-only code (which it used to be, but it's now part of the publicly exported surface area from this package). Of course, we could introduce a test-only variant, but I think that it's probably not necessary...

bk2204:
I think all of the issues here have been addressed, so it should be ready for a round 2 review.

PR #3588, opened 2019-04-01
Use git-lfs version of go-ntlm

### LINKED ISSUE #3475: Take over go-ntlm?
I'm just pointing out the message in the readme on `go-ntlm`:

> This repository will be removed on December 14th and will no longer be maintained and ownership will also not be transferred. Please fork this repositiory if you wish to continue to use it.

As perhaps the largest user of it (maybe? dunno how to measure), maybe `go-ntlm` can be folded into the `git-lfs` organization like was done with `go-netrc`?

Of course, it didn't actually get removed, only archived...

bk2204:
The upstream of go-ntlm has archived its repository and is no longer doing releases. Because this dependency is required for Git LFS, we've created our own fork to ensure that the upstream repo doesn't disappear on us. Use our own copy of go-ntlm within Git LFS.

Fixes #3475


bk2204:
I can explain that. Our fork has a few additional commits (to update the README and change the package declarations), so the object ID and a few of the files (e.g. the ones that refer to the MD4 package) are slightly different, leading to the different hash.

ttaylorr:
> I can explain that. Our fork has a few additional commits (to update
> the README and change the package declarations), so the object ID and
> a few of the files (e.g. the ones that refer to the MD4 package) are
> slightly different, leading to the different hash.

Ah, that makes good sense. It seems that there are no implementation
changes, so I think that this is still good to go.

PR #3590, opened 2019-04-02
tq/adapterbase: support rewriting href

h-hirokawa:
Use insteadOf/pushInsteadOf aliases to rewrite href to upload/download LFS objects.
This is useful in situations such like you need to access the LFS server via a reverse proxy.

### DIFF
--- a/tq/adapterbase.go
+++ b/tq/adapterbase.go
@@ -51,6 +51,11 @@ type transferImplementation interface {
 	DoTransfer(ctx interface{}, t *Transfer, cb ProgressCallback, authOkFunc func()) error
 }
 
+const (
+	enableHrefReriteKey     = "lfs.transfer.enablehrefrewrite"
+	defaultEnableHrefRerite = false
+)
+
 func newAdapterBase(f *fs.Filesystem, name string, dir Direction, ti transferImplementation) *adapterBase {
 	return &adapterBase{
 		fs:           f,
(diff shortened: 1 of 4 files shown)


tq/adapterbase.go
```diff
@@ -194,12 +194,14 @@ func (a *adapterBase) worker(workerNum int, ctx interface{}) {
 var httpRE = regexp.MustCompile(`\Ahttps?://`)
 
 func (a *adapterBase) newHTTPRequest(method string, rel *Action) (*http.Request, error) {
-	if !httpRE.MatchString(rel.Href) {
-		urlfragment := strings.SplitN(rel.Href, "?", 2)[0]
+	href := a.apiClient.Endpoints.NewEndpoint(a.direction.String(), rel.Href).Url
```
bk2204: I have some questions about how this is going to work with existing implementations.

For example, if I have the following, is this going to still work, or will things break?

```
[url "ssh://git@github.com/"]
    insteadOf = "https://github.com/"
```

Similar questions apply to how this will work with Bitbucket and GitLab's implementations. My gut tells me this will work fine for GitHub, but it will probably not for Bitbucket. If that's the case, we'll probably need to leave this until a 3.0 version where we have an SSH-based protocol, since otherwise this will break working configurations.

h-hirokawa:
@bk2204 
Certainly, it will break in the situations that use insteadof config to force SSH protocol because this try to download LFS objects through SSH.
And I have checked this works as expected when using HTTPS with GitHub and GitLab, but not checked with Bitbucket.

Therefore, for backward compatibility, I added a new config `lfs.transfer.enablehrefrewrite`, so there are no breaking changes any more.
Please re-review it.

Issue #3615, opened 2019-04-17
SSPI should be checked prior to prompting for credentials


AJH16:
Currently ntlmReAuth in ntlm.go calls ntlmGetCredentials prior to testing if SSPI already provides access.  It would be preferable to attempt authentication using SSPI first and only prompt the user for credentials if SSPI fails.  Without creating a Windows Credential with the Git Credential Manager, every access to LFS prompts for a username and password and then succeeds when you enter nothing and simply hit enter.  Pre-checking the SSPI credentials would prevent this from occurring by allowing the SSPI authentication to occur without a user prompt.

I think this should be a pretty simple change, but I am not aware enough of the overall design around the code to be certain of making the change myself.

AJH16:
The creds could also be checked to make sure it's the proper return for no credentials being provided by Git Credential Manager.  This would prevent excessive login attempts as environments not using SSPI would only have the SSPI auth failure when the git credential manager does not have a record for the location.  After git credential manager stores the credentials, they would be used prior to attempting the SSPI authentication in the future.

bk2204:
Hey, thanks for writing in.

I will admit to not being very familiar with NTLM, but it sounds like you're describing a situation analogous to Kerberos: you have some sort of cached token that can be accessed without needing a username and password, and as long as that authentication succeeds, no prompt for credentials is necessary.

I definitely think that this would be great to fix, but since I don't have access to a Windows domain controller, I can't fix it myself. The NTLM code was contributed by someone else and I don't have a way to test it. If you're interested in looking at the code, I'm happy to provide helpful feedback and answer any questions you might have. The NTLM code is generally pretty easy to find with a `git grep -i ntlm`, and hopefully it's not too complex.

My only request here is that this function with any credential manager, not just the one with Git for Windows. It may be the case that users are using other credential managers or non-Windows machines with a Windows domain and we should make sure that continues to work well.

AJH16:
Ok, so the part I was unsure about is how exactly you are making the request and how to make the attempt in such a way that it can still make the request with the provided credentials if it fails.  I'm also not sure exactly what the client c is going to provide from the abstract credHelper when there is no record.  It might also be possible to sidestep it somewhat by using SSPI rather than NTLM in the lfs.{env}.access setting.

[...]

PR #3617, opened 2019-04-18
Wrap credential data to allow late credential prompting and update NTLM/SSPI  to attempt SSPI login prior to prompting for credentials.

AJH16:
This is very rough at the moment as I just picked up Go and have very little idea about it's structures, so I may be mishandling objects or over-exposing things, but this provides what should be a general framework for encapsulating the information related to credentials and making it so that the Fill can be called later in the process to support things like NTLM SSPI or anything else where an automatic credential should be tried prior to prompting the user, but where we can't take the initial action until after the credential helpers have been setup. 

The code builds and the general concept should be sound as far as I can tell, but I can't guarantee anything towards proper function at the moment.

This is an attempt at resolving https://github.com/git-lfs/git-lfs/issues/3615.


bk2204:
Sorry for not getting back to you sooner on this. I've taken a first pass over it, and I hope to give it a more comprehensive look on Tuesday.

lfsapi/ntlm.go
```diff
@@ -29,32 +28,42 @@ func (c *Client) doWithNTLM(req *http.Request, credHelper creds.CredentialHelper
...
 		return res, err
 	}
 
-	switch res.StatusCode {
-	case 401:
-		credHelper.Reject(creds)
-		if retry {
-			return c.ntlmReAuth(req, credHelper, creds, false)
+	// If SSPI failed, then we need to try the normal.
+	if res.StatusCode >= 300 || res.StatusCode < 199 {
```
bk2204: It might be tidier to reverse this condition and `return res, nil` on success. Then you don't need to indent this block.

AJH16:
It appears that the test for lfsapi is timing out after 10 minutes, but I'm not sure how anything I changed could have caused that.  I'm not sure how I would even go about debugging the test.

bk2204:
I recently hit a test timeout as well and it was a hang in the code. Does the test hang when you run `make && make test integration` on your system?

AJH16:
when I do make test integration I get a bunch of \ and then make: *** [lint] Error 1.  When I do "go test ./lfsapi" it runs for 10 minutes and then times out.

bk2204:
If you're seeing that failure in `go test`, then something in your code is causing a hang. That would probably be why CI is failing.

The `make lint` failure means that you may have introduced a new dependency, in which case you may need to do `make vendor` to include it.

AJH16:
"make vendor" also fails with an error 1.

bk2204:
Can you provide the actual error output? Are you using a version of Go that's at least 1.11?

AJH16:
Apparently there is a temporary environment variable that I had to set to get vendor to work.

AJH16:
But yeah, I know that it's the lfsapi test that's failing.  I'm pretty sure it is because the timing of the credential prompt changed and the test execution doesn't provide anything for the input, but I'm not really sure how to make that happen as I don't really understand what the current test is actually doing.

bk2204:
I think the problem is that something is getting called in a loop without making progress. For example, if I apply the following patch and run `go test -v ./lfsapi`, it prints “called credential helper” over and over:

```
diff --git a/lfsapi/auth_test.go b/lfsapi/auth_test.go
index 358e984e..0dcabdda 100644
--- a/lfsapi/auth_test.go
+++ b/lfsapi/auth_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"os"
 	"strings"
 	"sync/atomic"
 	"testing"
@@ -292,6 +293,7 @@ func newMockCredentialHelper() *mockCredentialHelper {
 }
 
 func (m *mockCredentialHelper) Fill(input creds.Creds) (creds.Creds, error) {
+	fmt.Fprintf(os.Stderr, "called credential helper\n")
 	if found, ok := m.Approved[credsToKey(input)]; ok {
 		return found, nil
 	}
```

I'm not sure why this is, but it's definitely the cause of the hang.

AJH16:
So I tracked it down but I don't know go well enough to know what the fix is.  It appears the problem is that FillCreds does not end up updating the Creds property correctly.  When I call in to FillCreds on auth.go:152, within the method I have the correct creds, but after the function returns, the creds on the object are empty.

bk2204:
> So I tracked it down but I don't know go well enough to know what the fix is. It appears the problem is that FillCreds does not end up updating the Creds property correctly. When I call in to FillCreds on auth.go:152, within the method I have the correct creds, but after the function returns, the creds on the object are empty.
> 
> It appears perhaps it needed a *. Changing that seems to have fixed it to be pass by reference.

Oh, yeah, I missed that. Typically you want to make methods use the pointer version in Go (the version with the `*`). Glad you got it figured out.

AJH16:
Looks like something is still failing with the NTLM test, I hadn't been able to get that one working locally yet either and wasn't 100% clear on what it's doing.  I'll do some more digging.  It's the actual ntlm_test - TestNtlmAuth that seems to be the problem.

AJH16:
Ok, so I have tracked down the problem.  There apparently is not currently any SSPI test suite.  Switching the order so that SSPI goes first resulted in unexpected requests to the test server which result in it crashing.  I do not have the knowledge to implement an SSPI server side test emulator, but what I can probably do is alter the NTLM test to make it go straight to using the non SSPI NTLM credentials which should allow the basic NTLM test to run.  Any objections?

bk2204:
That's probably fine for now. If you're interested, there's a [good doc](http://davenport.sourceforge.net/ntlm.html) that goes into depth about NTLM, should you want to synthesize a test.

AJH16:
I'm not even sure if we could synthesis a good cross platform one since with SSPI the credential is going to be pulled from the active user.  We'd be faking it hard to the point of probably being pretty close to no value anyway.

AJH16:
It appears there is now some kind of fault in the ntlm_auth_test_nix, but I'm not able to debug that test as I don't have a *nix environment handy.  Any idea what that test is supposed to be doing?

bk2204:
I think that test is supposed to be checking that you can authenticate on Unix. In Go, you can have build commands at the top of the file to control the platform, and the `_nix.go` files are only built on non-Windows.

Part of the problem here is that SSPI doesn't work on Unix, because the credentials that are available on Windows won't exist, so the test is failing because it's getting nil credentials, which it doesn't expect. It may be a good idea to wrap the nil credentials code in something like `if runtime.GOOS == "windows"`, since we know that code will never work.

AJH16:
Yeah, sorry I didn't update earlier, but I figured that out and added a value to the ntlm_auth implementation that is platform specific to indicate if SSPI is available or not.  Right now it's just true and false hardcoded, but it's possible it could have more complicated logic to determine in the future.

AJH16:
Well, after updating the NTLM test to respond correctly when skipping the SSPI attempt on *nix, it's now made it past the unit tests.  Guess we'll see how the rest of the test suite does.  That is quite the set of tests.

AJH16:
Ok, so I think things should be good now.  I'm not sure how to make it so that the requested changes are addressed.  I updated them in the code, but it seems that I did something wrong process wise for having them flag as clear.  I was also able to test this in my local TFS environment and it successfully did a clone of an LFS enabled repo without prompting for a password and succeeded via SSPI.

AJH16:
And doing a clone of an LFS enabled repository just worked on a machine that is not on a trusted domain using NTLM credentials without going through SSPI as well.

AJH16:
Updated title of pull request now that this is an actual fix for the issue and has been more rigorously developed and tested.

PR #3623, opened 2019-04-23
Various release updates

### LINKED ISSUE #3621: Enable disco as a release target
Currently we have Ubuntu "disco" commented out as a release target. We should enable it before the 2.8.0 release.

bk2204:
There are several release-related issues that I discovered while releasing 2.7.2. The first of these, of course, is the nwe signing certificate, which requires an update to the Makefile. In addition, I discovered several issues with the PackageCloud upload script, including some syntax errors and a release that PackageCloud refused to accept. Finally, I hit an additional issue with attempting to upload existing package that I've seen repeatedly before but have just hacked around; this time I finally fixed it.

I've also added Ubuntu disco as a release target, which fixes #3621. Note that I didn't use this change in this release; I simply applied it so as to not forget about it for 2.8.0.

### DIFF
--- a/script/packagecloud.rb
+++ b/script/packagecloud.rb
@@ -51,7 +51,6 @@ $distro_name_map = {
     "sles/12.1", # LTSS ends 31 May 2020
     "sles/12.2", # LTSS ends 31 Mar 2021
     "sles/12.3", # LTSS ends 30 Jun 2022
-    "sles/12.4", # Current
     "sles/15.0"  # Current
   ],
   # Debian EOL https://wiki.debian.org/LTS/
--- a/script/packagecloud.rb
+++ b/script/packagecloud.rb
@@ -60,7 +60,7 @@ $distro_name_map = {
   "debian/7" => [
     "debian/wheezy", # EOL 31st May 2018
     "ubuntu/precise" # ESM April 2019
-  ),
+  ],
   "debian/8" => [
     "debian/jessie",     # EOL June 30, 2020
     "linuxmint/qiana",   # EOL April 2019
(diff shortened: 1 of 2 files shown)


script/packagecloud.rb
```diff
@@ -114,7 +114,9 @@ def distro_names_for(filename)
     puts "pushing #{full_path} to #{$distro_id_map.key(distro_id).inspect}"
     result = $client.put_package("git-lfs", pkg, distro_id)
     result.succeeded || begin
-      raise "packagecloud put_package failed, error: #{result.response}"
+      if result.response != '{"filename":["has already been taken"]}'
```
PastelMobileSuit: Could you give some additional context on why this if-statement is necessary? I don't know much about packagecloud, so I'm still a bit confused after reading the commit message.

bk2204: Sure. During the release process, the upload aborted halfway through due to `sles/12.4` not existing. Re-running the script causes us to attempt to re-upload the same files. If the file already exists and this conditional isn't present, PackageCloud returns this error and we throw an exception and abort the upload process, since PackageCloud doesn't allow us to upload over a file with the same name.

Instead, the thing we want to do is just ignore the fact that we've already uploaded those files and continue the process. Similarly, when we build the CentOS 6 packages, we always build the same Ruby and ronn packages, and without this change, they have to be manually deleted from the directory or the script will abort when the upload gets to them.

Hopefully this makes things a bit clearer.

PastelMobileSuit: Ah okay! So we can still upload the package if we receive this "package already exists" error, we just weren't doing it before? Could you add a comment here to that effect?

bk2204: Well, not exactly. This error means we've already uploaded the package, and we can't upload it a second time, so we're ignoring the failure when we attempt to do so.

I'll add an appropriate comment.

PR #3624, opened 2019-04-23
Improve error handling in git lfs install

### LINKED ISSUE #3622: git lfs install --system returns zero exit code on failure
Trying to install LFS system-wide on windows without sufficient access rights fails with reasonable error message, however returns zero exit code. That's a bit unpleasant when one wants to do it in a script.

Git-LFS version: 2.7.1 (included in Git for Windows 2.21.0)

```
C:\>git lfs install --system
WARNING: current user is not root/admin, system install is likely to fail.
WARNING: Error running C:\Program Files\Git\mingw64\libexec\git-core\git.exe 'config' '--system' '--replace-all' 'filter.lfs.required' 'true': 'error: could not lock config file C:/Program Files/Git/mingw64/etc/gitconfig: Permission denied' 'exit status 255'
Run `git lfs install --force` to reset git config.

C:\>echo %ERRORLEVEL%
0
```

A somewhat related cosmetic issue is, that when the access rights are sufficient, it still wrongly prints the warning that they are not, but then succeeds:
```
C:\>git lfs install --system
WARNING: current user is not root/admin, system install is likely to fail.
Git LFS initialized.
```
Observed on Windows 10 1809 x64 & Windows 7 x64, also with Git-LFS 2.7.1.

bk2204:
This series introduces two improvements to git lfs install. The first is to return unsuccessfully (exit code 2) when the operation fails, so that scripting users can easily detect the failure. The second is to avoid printing the warning about the lack of root privileges on Windows, where we will always be unable to detect these privileges properly.

/cc r0nw as reporter
Fixes #3622

### DIFF
--- a/t/t-install.sh
+++ b/t/t-install.sh
@@ -35,7 +35,7 @@ begin_test "install with old (non-upgradeable) settings"
   git config --global filter.lfs.clean "git-lfs clean --something %f"
 
   git lfs install | tee install.log
-  [ "${PIPESTATUS[0]}" = 0 ]
+  [ "${PIPESTATUS[0]}" = 2 ]
 
   grep -E "(clean|smudge)\" attribute should be" install.log
   [ `grep -c "(MISSING)" install.log` = "0" ]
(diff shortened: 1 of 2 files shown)


t/t-install.sh
```diff
@@ -224,6 +225,34 @@ begin_test "install --local"
 )
 end_test
 
+begin_test "install --local with failed permissions"
```
PastelMobileSuit: What causes the install to fail in this case?

bk2204: The `chmod 500 .git` operation. The directory isn't writable, so `.git/config` can't be updated.

Let me add a comment to that effect.

berrylaw:
Call me if any Enquiry 63313350 

Berry Law

berrylaw:
I don’t know how to play ur game 
This is not funny sorry 

Berry Law

PR #3634, opened 2019-04-30
Don't fail if we lack objects the server has

### LINKED ISSUE #3468: Avoid failing when pushing objects the server has
We have had a variety of reports related to #1113 where a user wants to push a branch with a deleted LFS file. Almost all of the time, the server already has the object, although the local side does not, and we fail because we check for the object locally before determining if the server has it.

It would be better if we could defer checking for missing objects until we know if they'll actually be needed for upload. This approach would solve most of the problems that people have in these kinds of cases.

bk2204:
A Git LFS client may not have the entire history of the objects for the repository. However, in some situations, we traverse the entire history of a branch when pushing it, meaning that we need to process every LFS object in the history of that branch. If the objects for the entire history are not present, we currently fail to push.

Instead, let's mark objects we don't have on disk as missing and only fail when we would need to upload those objects. We'll know the server has the objects if the batch response provides no actions to take for them when we request an upload. Pass the missing flag down through the code, and always set it to false for non-uploads.

If for some reason we fail to properly flag a missing object, we will still fail later on when we cannot open the file, just in a messier and more poorly controlled way. The technique used here will attempt to abort the batch as soon as we notice a problem, which means that in the common case (less than 100 objects) we won't have transferred any objects, so the user can notice the failure as soon as possible.

Update the tests to look for a string which will occur in the error message, since we no longer produce the system error message for ENOENT.

Fixes #3468

### DIFF
--- a/commands/command_migrate_export.go
+++ b/commands/command_migrate_export.go
@@ -136,7 +136,7 @@ func migrateExportCommand(cmd *cobra.Command, args []string) {
 			}
 
 			if _, err := os.Stat(downloadPath); os.IsNotExist(err) {
-				q.Add(p.Name, downloadPath, p.Oid, p.Size)
+				q.Add(p.Name, downloadPath, p.Oid, p.Size, false)
 			}
 		})
 		gs.ScanRefs(opts.Include, opts.Exclude, nil)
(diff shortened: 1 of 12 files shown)


bk2204:
Gentle ping, @git-lfs/core. It would be nice to get this into the next release, since this issue results in significant slowness with Eclipse and in some other cases as well.

Aa88s:
#3734 lfs pre-push implementation

PR #3636, opened 2019-05-08
Allow vendors to customize the version info

bk2204:
It's common for vendors (such as Linux distros) to want to customize the User-Agent string and version information for software they provide. This is valuable for vendors, but it also allows troubleshooting problems with custom patches that may have been applied by the vendor. Allow vendors to set a VENDOR makefile flag to set the vendor field for the User-Agent and version information.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,9 @@ GO_TEST_EXTRA_ARGS =
 # the config.GitCommit variable is always set via this variable, and
 # DWARF-stripping is enabled unless DWARF=YesPlease.
 BUILTIN_LD_FLAGS =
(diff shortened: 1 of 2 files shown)


QuLogic:
Do you have any expectations of what to put in this field?

bk2204:
The default is "GitHub", and I had anticipated an alternative such as "Debian" or "Fedora". If, for example, SourceTree were distributing their own version, they could use "Atlassian", "SourceTree", or "Atlassian SourceTree".

I just know that Debian often wants to customize these kinds of fields, so I provided a technique for them to do so should they want to do so.

PR #3639, opened 2019-05-10
Document a new batch error code

bk2204:
In some cases, it may be useful to return an error code if the number of objects in the batch response is too larger for the server to handle. Document the 413 error code for this case, since it is in use by at least one implementation for this purpose.

### DIFF
--- a/docs/api/batch.md
+++ b/docs/api/batch.md
@@ -262,6 +262,8 @@ The following status codes can optionally be returned from the API, depending on
 the server implementation.
 
 * 406 - The Accept header needs to be `application/vnd.git-lfs+json`.
+* 413 - The batch API request contained too many objects or the request was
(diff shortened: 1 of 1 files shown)


docs/api/batch.md
```diff
@@ -262,6 +262,8 @@ The following status codes can optionally be returned from the API, depending on
 the server implementation.
 
 * 406 - The Accept header needs to be `application/vnd.git-lfs+json`.
+* 413 - The batch API request contained too many objects or the request was
```
PastelMobileSuit: Would it be worth updating our implementation to return this status code?

bk2204: I assume you're referring to our test server. I've opened git-lfs/lfs-test-server#84 to track implementing that.

PR #3652, opened 2019-05-15
Fix 'locked_at' lock field not documented as optional

slonopotamus:
According to `locking/schemas/http-lock-create-response-schema.json`, `locked_at` field is not required,
so fix locking API documentation to match schema.

### DIFF
--- a/docs/api/locking.md
+++ b/docs/api/locking.md
@@ -61,7 +61,7 @@ Successful responses return the created lock:
 as long as it's returned as a string.
 * `path` - String path name of the locked file. This should be relative to the
 root of the repository working directory.
-* `locked_at` - The timestamp the lock was created, as an ISO 8601 formatted string.
+* `locked_at` - Optional timestamp the lock was created, as an ISO 8601 formatted string.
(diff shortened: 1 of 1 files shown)


bk2204:
Hey, I would say that the documentation is canonical, not the schemas, so would you mind updating the schemas instead of the documentation?

slonopotamus:
But... but schemas were given higher priority in #3650 and #3651 :( I fail to follow this logic. It isn't safe to fix schemas because that way we will make `locked_at` mandatory while there can be implementations in the wild that already rely on the fact that this field was optional for `git-lfs`.

bk2204:
The issue is that when we have a conflict, we have to look at what real-world implementations do. In this case, I think most implementations do in fact return this value, and so the proper behavior is that we return the value. Moreover, the Git LFS client deserializes this into an object, in which case we're going to want to provide a value that's useful for that case.

slonopotamus:
Okay, will update this PR tomorrow.

slonopotamus:
#3655 fixes schema, closing this PR.

PR #3663, opened 2019-05-22
Ensure we always use correct Content-Type and Accept headers

### LINKED ISSUE #3662: git-lfs doesn't send 'Accept: application/vnd.git-lfs+json' HTTP header for /verify URL
LFS spec [says](https://github.com/git-lfs/git-lfs/blob/master/docs/api/basic-transfers.md#verification) that LFS client needs to send both `Accept: application/vnd.git-lfs+json` and `Content-Type: application/vnd.git-lfs+json`. But [verifyUpload](https://github.com/git-lfs/git-lfs/blob/84f47a430beb9695ceb8427443f304e091128edd/tq/verify.go#L16) function only sends `Content-Type`, without sending `Accept` header.

Also see git-lfs/lfs-test-server#85 to fully feel how messed up situation with `/verify` is.

bk2204:
There is one place in the code and two places in the testsuite where we fail to provide the proper Content-Type and Accept headers when making a request to the server. Correct these places and adjust the test server to always check for the Accept header and check for the Content-Type header whenever we have a JSON body (i.e., when we have a POST request).

Fixes #3662
/cc @slonopotamus as reporter

### DIFF
--- a/tq/verify.go
+++ b/tq/verify.go
@@ -35,10 +35,11 @@ func verifyUpload(c *lfsapi.Client, remote string, t *Transfer) error {
 		return err
 	}
 
+	req.Header.Set("Content-Type", "application/vnd.git-lfs+json")
+	req.Header.Set("Accept", "application/vnd.git-lfs+json")
 	for key, value := range action.Header {
 		req.Header.Set(key, value)
 	}
-	req.Header.Set("Content-Type", "application/vnd.git-lfs+json")
 
 	mv := c.GitEnv().Int(maxVerifiesConfigKey, defaultMaxVerifyAttempts)
 	mv = tools.MaxInt(defaultMaxVerifyAttempts, mv)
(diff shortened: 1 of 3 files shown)


tq/verify.go
```diff
@@ -39,6 +39,7 @@ func verifyUpload(c *lfsapi.Client, remote string, t *Transfer) error {
 		req.Header.Set(key, value)
 	}
 	req.Header.Set("Content-Type", "application/vnd.git-lfs+json")
+	req.Header.Set("Accept", "application/vnd.git-lfs+json")
```
slonopotamus: No, this will break compatibility with lfs-test-server and current Gitea. See git-lfs/lfs-test-server#85. Change the order: first set your headers and **then** overwrite them with headers from `action.Header`.

bk2204: Okay, fixed and squashed.

PR #3666, opened 2019-05-26
Fix typo in t-locks.sh

TBK:


bk2204:
And I'm pretty sure this CI failure is one of our periodic failures; it appears unrelated to this test.

PR #3676, opened 2019-06-04
Update README to draw attention to lfs track pattern

richardwu:
I naively ran into this unintended consequence when first using git-lfs. I think this subtle detail should be pointed out more boldly in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,11 @@ $ git lfs track "*.psd"
 more about this pattern syntax
 [here](https://git-scm.com/docs/gitattributes)).
 
+**Note**: the quotations surrounding the pattern are important. Otherwise
+the glob pattern will be interpreted first by your shell and expanded into matching
+existing files. Multiple individual file entries will be created in 
+`.gitattributes` voiding the glob pattern in the future.
+
 After any invocation of `git-lfs-track(1)` or `git-lfs-untrack(1)`, you _must
 commit changes to your `.gitattributes` file_. This can be done by running:
 
--- a/README.md
+++ b/README.md
@@ -77,10 +77,8 @@ $ git lfs track "*.psd"
 more about this pattern syntax
 [here](https://git-scm.com/docs/gitattributes)).
 
-**Note**: the quotations surrounding the pattern are important. Otherwise
-the glob pattern will be interpreted first by your shell and expanded into matching
-existing files. Multiple individual file entries will be created in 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -77,6 +77,11 @@ $ git lfs track "*.psd"
 more about this pattern syntax
 [here](https://git-scm.com/docs/gitattributes)).
 
+**Note**: the quotations surrounding the pattern are important. Otherwise
+the glob pattern will be interpreted first by your shell and expanded into matching
+existing files. Multiple individual file entries will be created in
```
bk2204: I think we have some trailing whitespace on this line, which is causing the CI system to be upset. Could you remove it?

README.md
```diff
@@ -77,6 +77,11 @@ $ git lfs track "*.psd"
 more about this pattern syntax
 [here](https://git-scm.com/docs/gitattributes)).
 
+**Note**: the quotations surrounding the pattern are important. Otherwise
+the glob pattern will be interpreted first by your shell and expanded into matching
+existing files. Multiple individual file entries will be created in 
+`.gitattributes` voiding the glob pattern in the future.
```
bk2204: I definitely think this is an important thing we should note. It's a tricky thing for users and I've seen other people get it wrong in the past with surprising results, as you mentioned.

Having said that, we'd like to keep the README file as short and concise as possible, since people are less likely to read longer documents. What would you think about condensing this into a single sentence, like, "Note that the quotation marks surrounding the pattern are important to prevent the glob pattern from being expanded by the shell"?

README.md
```diff
@@ -77,6 +77,11 @@ $ git lfs track "*.psd"
 more about this pattern syntax
 [here](https://git-scm.com/docs/gitattributes)).
 
+**Note**: the quotations surrounding the pattern are important. Otherwise
```
ttaylorr: It may be good to match the existing style of the "you-should-probably-know-about-this" as we have elsewhere in the README.md, such as writing it instead with a blockquote and italics, as:

```md
> *Note*: the quotations surrounding the pattern are important [...]
```

README.md
richardwu: Sounds good: updated.

README.md
richardwu: 👍

richardwu:
Updated PR re: comments. Feel free to squash and merge when green.

seba1997:
N

PR #3677, opened 2019-06-06
Add support for SOCKS proxies

### LINKED ISSUE #1424: Error with socks5 proxy
Hey there.

I got some problems when using git-lfs behind a socks proxy.  
The proxy settings in the git config are as follows:  
`http.proxy=socks5://127.0.0.1:3334`

When  `git push`ing an error is thrown:  

```
$ git push
Git LFS: (0 of 1 files) 0 B / 154.28 MB
Post https://_username_@_ip_/binaries/lfs.git/info/lfs/objects/batch: http: error connecting to proxy http://socks5://127.0.0.1:3334: dial tcp: lookup socks5: getaddrinfow: No such host is known.
```

So the config is read, but git-lfs tries to connect to the proxy while appending another protocol (`http://`). When there's already one in place it should be used, imo.

KR,
bcersows

bk2204:
Currently, we prepend "http://" to the proxy URL if it's not an HTTP or HTTPS URL. Unfortunately, that breaks support for SOCKS 5 proxies, which use URLs that start with "socks5://". Fix this by allowing URL schemes starting with "socks" in addition to those starting with "http".

Note that this does not introduce any support for socks5h proxies, since Go does not support them, but we will support them automatically once Go does.

Fixes #1424.

### DIFF
--- a/lfshttp/proxy.go
+++ b/lfshttp/proxy.go
@@ -32,7 +32,7 @@ func proxyFromClient(c *Client) func(req *http.Request) (*url.URL, error) {
 		}
 
 		proxyURL, err := url.Parse(proxy)
-		if err != nil || !strings.HasPrefix(proxyURL.Scheme, "http") {
+		if err != nil || !(strings.HasPrefix(proxyURL.Scheme, "http") || strings.HasPrefix(proxyURL.Scheme, "socks")) {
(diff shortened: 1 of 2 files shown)


xkszltl:
@bk2204 Should we list socks5h there as well?

bk2204:
socks5h support was added in 8f5cd667d.

PR #3680, opened 2019-06-11
Be more precise about what timestamps we accept

### LINKED ISSUE #3660: git-lfs doesn't accept valid ISO-8601 date
This was discovered against a custom LFS server, so for now there is no easy way to reproduce this.

```
$ git lfs --version
git-lfs/2.6.1 (GitHub; linux amd64; go 1.11.5)
$ git lfs lock .gitattributes
Lock failed: parsing time ""2019-05-20T15:51:24.303+0000"" as ""2006-01-02T15:04:05Z07:00"": cannot parse "+0000"" as "Z07:00"
```

According to [Wikipedia](https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC), this is a perfectly valid ISO-8601 date string.

Bug at least applies to `locked_at` field of LFS locks. Possibly other places where LFS spec talks about "ISO 8601 formatted timestamp" are also affected.

bk2204:
In a couple of places in our documentation, we specified that a timestamp should be in ISO 8601 format. However, ISO 8601 specifies a large number of variants, and Go only parses a subset of those. For example, using a timestamp without the colon in the time zone offset will cause Git LFS to fail.

Be explicit about what we want as a timestamp, which is an uppercase RFC 3339-formatted timestamp with second precision. That's what Go parses, and that's what we expect. It is, in addition, a profile of ISO 8601, so other tools that can parse the entirety of ISO 8601 will still accept these date formats.

Fixes #3660
/cc @slonopotamus as reporter

### DIFF
--- a/docs/api/batch.md
+++ b/docs/api/batch.md
@@ -145,8 +145,9 @@ omitted.
     * `expires_in` - Whole number of seconds after local client time when
       transfer will expire. Preferred over `expires_at` if both are provided.
       Maximum of 2147483647, minimum of -2147483647.
-    * `expires_at` - String ISO 8601 formatted timestamp for when the given
-    action expires (usually due to a temporary token).
+    * `expires_at` - String uppercase RFC 3339-formatted timestamp with second
+      precision for when the given action expires (usually due to a temporary
+      token).
 
 Download operations MUST specify a `download` action, or an object error if the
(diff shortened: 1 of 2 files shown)


slonopotamus:
lgtm

PR #3686, opened 2019-06-14
Don't recurse into submodules when walking repository

### LINKED ISSUE #3638: Post commit hook traverses submodules, causing git commit to take an extremely long time to complete
Maybe similar to #3490, but I opened a new issue since the root cause might be different.

Here is the command I run:

```
GIT_TRACE=1 git commit -a --amend --no-edit
```

The trace file is attached here: [commit-trace.txt](https://github.com/git-lfs/git-lfs/files/3163294/commit-trace.txt)

My repository structure has 2 submodules in it. Here is my `.gitmodules`:

```
[submodule "zPayService/ZPayUtilities"]
    path = zPayService/ZPayUtilities
    url = ../zpay-utilities.git
[submodule "Core"]
    path = Core
    url = ../core.git
    branch = .
```

The `Core` submodule is also an LFS repository. The `post-commit` started going into the Core subdirectory, which it shouldn't do because that's a submodule. Evidence is shown in the attached trace, here's a small snippet:

```
12:29:47.164317 trace git-lfs: filepathfilter: accepting "Core/External/PowerVR/source/external/concurrent_queue"
12:29:47.164317 trace git-lfs: filepathfilter: accepting "Core/External/PowerVR/source/LICENSE_POWERVR_SDK.txt"
12:29:47.164317 trace git-lfs: filepathfilter: accepting "Core/External/PowerVR/source/include/CL"
```

Because the post commit hook is traversing the submodule files, it makes commits in the parent repository take a very long time. Is this a bug or a configuration issue?

bk2204:
When we're walking the tree to select objects which need permissions adjusted, don't recurse into any submodules that exist. This is not only slow, but the submodules may have their own gitattributes files which we wouldn't have honored.

Fixes #3638.

### DIFF
--- a/t/t-post-commit.sh
+++ b/t/t-post-commit.sh
@@ -74,3 +74,33 @@ begin_test "post-commit (locked file outside of LFS)"
   refute_file_writeable a.dat
 )
 end_test
+
+begin_test "post-commit does not enter submodules"
+(
+  set -e
+
+  reponame="post-commit-submodules"
+  setup_remote_repo "$reponame"
+
+  clone_repo "$reponame" "$reponame"
+
+  git lfs track --lockable "*.dat"
+  git lfs track "*.big" # not lockable
+  git add .gitattributes
+  git commit -m "add git attributes"
+
+  mkdir submodule
+  (cd submodule && git init && echo abc >foo && git add foo && git commit -m 'foo')
+  git submodule add ./submodule submodule
+  git commit -m 'Add submodule'
+
+  echo "Come with me" > pcfile1.dat
+  echo "and you'll be" > pcfile2.dat
+  echo "in a world" > pcfile3.big
+  echo "of pure imagination" > pcfile4.big
+
+  git add *.dat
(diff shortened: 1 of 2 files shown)


t/t-post-commit.sh
```diff
@@ -74,3 +74,33 @@ begin_test "post-commit (locked file outside of LFS)"
...
+
+  mkdir submodule
+  (cd submodule && git init && echo abc >foo && git add foo && git commit -m 'foo')
+  git submodule add ./submodule submodule
+  git commit -m 'Add submodule'
+
+  echo "Come with me" > pcfile1.dat
+  echo "and you'll be" > pcfile2.dat
+  echo "in a world" > pcfile3.big
+  echo "of pure imagination" > pcfile4.big
```
PastelMobileSuit: 👍, this is the kind of good test data we need

bk2204: It's copied from slightly above, but I agree this is good test data.

PR #3710, opened 2019-07-11
Fix CircleCI

bk2204:
Our CircleCI configuration seems to be being flagged as being 1.0 instead of 2.0, preventing jobs from running. This PR is an attempt to address this issue and get things working again.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,4 @@
 version: 2
-
 jobs:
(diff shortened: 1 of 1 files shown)


bk2204:
I've just reported it to CircleCI, and will follow up with them on Monday or whenever they get back to me. I agree that this shouldn't be an issue (and especially not suddenly when it worked before).

PR #3715, opened 2019-07-17
Use a download token when searching locks

### LINKED ISSUE #3653: git-lfs requires write access to list locks
[List locks](https://github.com/git-lfs/git-lfs/blob/master/docs/api/locking.md#list-locks) spec says:

> LFS Servers should ensure that users have at least pull access to the repository.

But in fact:

```
$ GIT_TRACE=1 git lfs locks
15:00:24.147599 git.c:670               trace: exec: git-lfs locks
15:00:24.147621 run-command.c:643       trace: run_command: git-lfs locks
15:00:24.150194 trace git-lfs: exec: git 'version'
15:00:24.151202 trace git-lfs: exec: git '-c' 'filter.lfs.smudge=' '-c' 'filter.lfs.clean=' '-c' 'filter.lfs.process=' '-c' 'filter.lfs.required=false' 'rev-parse' 'HEAD' '--symbolic-full-name' 'HEAD'
15:00:24.152904 trace git-lfs: exec: git 'config' '-l'
15:00:24.153889 trace git-lfs: run_command: ssh -- git@github.com git-lfs-authenticate slonopotamus/git-lfs-java.git upload
15:00:24.608683 trace git-lfs: HTTP: GET https://lfs.github.com/slonopotamus/git-lfs-java/locks
15:00:25.270450 trace git-lfs: HTTP: 200
```

Notice that `git-lfs-authenticate` is invoked with `upload` mode. Either spec is wrong and write access is **required** to list locks or git-lfs should fallback to `download` mode instead. In current shape, if user doesn't have write access, he doesn't pass `git-lfs-authenticate` step.

Way to reproduce:

1. Clone any GitHub repo where you only have read access
2. Invoke `GIT_TRACE=1 git lfs locks`
3. Observe that you do not get any locks because `git-lfs-authenticate` fails

bk2204:
The locking API documentation specifies that a user need only have pull access to the repository when searching locks. However, we requested an upload token, which requires push access. Instead, ask for a download token, which requires only pull access, as documented.

Fixes #3653
/cc @slonopotamus as reporter

### DIFF
--- a/locking/api.go
+++ b/locking/api.go
@@ -180,7 +180,7 @@ type lockList struct {
 }
 
 func (c *lockClient) Search(remote string, searchReq *lockSearchRequest) (*lockList, *http.Response, error) {
-	e := c.Endpoints.Endpoint("upload", remote)
+	e := c.Endpoints.Endpoint("download", remote)
 	req, err := c.NewRequest("GET", e, "locks", nil)
 	if err != nil {
 		return nil, nil, err
--- a/t/t-locks.sh
+++ b/t/t-locks.sh
@@ -50,6 +50,33 @@ begin_test "list a single lock"
 )
 end_test
 
+begin_test "list a single lock (SSH)"
+(
+  set -e
+
+  reponame="locks-list-ssh"
+  setup_remote_repo_with_file "$reponame" "f.dat"
+  clone_repo "$reponame" "$reponame"
+
+  sshurl="${GITSERVER/http:\/\//ssh://git@}/$reponame"
+  echo $sshurl
+  git config lfs.url "$sshurl"
+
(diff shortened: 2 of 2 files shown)


slonopotamus:
LGTM

bk2204:
I should mention for the benefit of reviewers that I did ping the implementors team on the issue and nobody screamed loudly. That doesn't mean that everything is guaranteed to work wonderfully, but it does mean that folks were informed.

bk2204:
I'm not sure we have any tests for SSH URL handling, but I'll check and add one if we do.

bk2204:
We do have handling for SSH things and there's now a test for this.

PR #3736, opened 2019-07-26
Do not log skipped checkouts to file

### LINKED ISSUE #3732: Reduce volume of logs when using include/exclude
### setup
- Git LFS versions 2.7.2 and 2.8.0
- Nexus 3 and GitLab LFS servers.

### what
I have a repo with ~8000 files, using `--include` / `--exclude` to only fetch a small subset of these.  The checkout operation results in a significant amount of logging written to disk.

The logs appear to be generated by this

https://github.com/git-lfs/git-lfs/blob/v2.8.0/commands/pull.go#L87-L88

which says

> // acceptable error, data not local (fetch not run or include/exclude) 

The operations I perform are

```
git clone -b the/branch https://the.url
git lfs fetch -I'xxx.*'
git lfs checkout
```

The problem is that the `.git/lfs/logs` loads up with significant logging, ~2.7GB for the 8000 file example.  Is there a way to configure a reduction or redirection of this, or some other way to quiet it down?

Thanks

jw3:
Fixes #3732 

Stderr logging is reduced and logging to `.git/lfs/logs` is eliminated

### DIFF
--- a/commands/pull.go
+++ b/commands/pull.go
@@ -85,7 +85,7 @@ func (c *singleCheckout) Run(p *lfs.WrappedPointer) {
 	if err := c.RunToPath(p, cwdfilepath); err != nil {
 		if errors.IsDownloadDeclinedError(err) {
 			// acceptable error, data not local (fetch not run or include/exclude)
-			LoggedError(err, "Skipped checkout for %q, content not local. Use fetch to download.", p.Name)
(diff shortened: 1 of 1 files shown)


bk2204:
I'm almost certain this is a spurious CI failure. We've seen this test fail before.

PR #3754, opened 2019-08-04
Add CheckCloneFileSupported

kazuki-ma:


bk2204:
As mentioned, I think this would be better included in #3753, since it's logically a part of that change and it would be easier to review as part of the same PR.

kazuki-ma:
@bk2204 Sorry for late reply.
Understood. Let me close this PR. 

Thank you.

PR #3755, opened 2019-08-05
Optimize traversal of Git objects with URL remotes

### LINKED ISSUE #3661: Improve push optimization to work with URL remotes
Currently, when we push to a remote specified by a URL, we traverse the entire history in order to discover which LFS objects need to be pushed. However, with a named remote, we use the heads of the remote to exclude objects we know the remote already has.

Since the pre-push hook knows which Git objects are being pushed, we should optimize it by traversing only those Git objects we're pushing, since we can assume that the remote side will also have any LFS objects that are referenced by the Git objects it already has.

This will improve performance in a variety of cases, notably with some JGit and EGit usages where the remote for pre-push hooks will never provide a named remote.

bk2204:
We have an optimization that means we avoid traversing objects that exist in the remote tracking branches of named remotes. We assume that the server already has the LFS objects referred to by those Git objects.

However, this doesn't work for URL remotes, since there are no remote tracking branches for them. Instead, let's make sure that we don't traverse Git objects that are included in the old value of the ref, which we know the server must have.

Note that this will not help the case where someone is pushing a new branch to a URL remote; in that case, the old value is the all-zeros value and we'll still have to traverse the entire history since we haven't learned anything about what the server has.

Fixes #3661

### DIFF
--- a/git/rev_list_scanner.go
+++ b/git/rev_list_scanner.go
@@ -23,9 +23,9 @@ const (
 	ScanRefsMode ScanningMode = iota
 	// ScanAllMode will scan all history.
 	ScanAllMode
-	// ScanLeftToRemoteMode will scan the difference between any included
+	// ScanRangeToRemoteMode will scan the difference between any included
 	// SHA1s and a remote tracking ref.
-	ScanLeftToRemoteMode
+	ScanRangeToRemoteMode
 )
 
 // RevListOrder is a constant type that allows for variation in the ordering of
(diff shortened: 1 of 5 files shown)


git/rev_list_scanner.go
```diff
@@ -25,7 +25,7 @@ const (
 	ScanAllMode
 	// ScanLeftToRemoteMode will scan the difference between any included
```
PastelMobileSuit: This comment will need to be updated to reflect the change

lfs/gitscanner.go
```diff
@@ -74,9 +74,9 @@ func (s *GitScanner) RemoteForPush(r string) error {
 	return nil
 }
 
-// ScanLeftToRemote scans through all commits starting at the given ref that the
+// ScanRangeToRemote scans through all commits starting at the given ref that the
```
PastelMobileSuit: Since there's a new parameter here (`right`) and we're now providing two refs, could you update this comment accordingly?

bk2204:
Thanks for pointing these out. I've fixed both of these, plus addressed the test failures by handling some edge cases that I hadn't previously considered.

PR #3756, opened 2019-08-05
Provide an option to track to handle paths literally

### LINKED ISSUE #3506: Add a way to automatically escape/quote filenames given to git lfs track
We use the following simple command to track all files larger than 100 KB in our repo:

```bash
$ find * -type f -size +100k -exec git lfs track '{}' +
```

The issue is that `git lfs track` in fact expects a git pattern and not filename as an argument. So if a filename contains special characters it can happen that this does not configure git lfs to really track a file. For example, if file has a filename `foo[bar].jpg`, calling `git lfs track foo[bar].jpg` will in fact not configure it to track that file.

I would suggest an argument is made to `git lfs track` to allow passing filenames and `git lfs track` would then make sure to properly escape/quote the filename before storing it into `.gitattributes` file. For example, something like `git lfs track --filename` or something like that.

bk2204:
Normally, the arguments to `git lfs track` are taken as glob patterns. This is generally useful, but sometimes there are filenames that contain special characters. Add a `--filename` option that forces arguments to be taken as literal paths instead of glob patterns to make handling these files easier.

In the test helpers, pass -F to grep so that our unusual file names don't trigger undesired pattern matching. This has no effect on other tests, as they all pass a literal filename here.

Fixes #3506.

### DIFF
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -27,6 +27,7 @@ var (
 	trackDryRunFlag         bool
 	trackNoModifyAttrsFlag  bool
 	trackNoExcludedFlag     bool
+	trackFilenameFlag       bool
 )
 
 func trackCommand(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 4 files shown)


gdlxn-ibm:
I've found that the `--filename` option doesn't work properly if the file name contains spaces in that `[` and `]` characters in `[[:space:]]` are incorrectly escaped.  Here's a example `.getattribues` file:

```
git-issue-103-component/200MiB filter=lfs diff=lfs merge=lfs -text
git-issue-103-component/200MiB+1 filter=lfs diff=lfs merge=lfs -text
git-issue-103-component/50MiB+1 filter=lfs diff=lfs merge=lfs -text
git-issue-103-component/55MiB\[1\].test filter=lfs diff=lfs merge=lfs -text
git-issue-103-component/test\[\[:space:\]\]folder/55MiB\[1\].test filter=lfs diff=lfs merge=lfs -text
```

The last entry was added by `git lfs track --filename "git-issue-103-component/test folder/55MiB[1].test"`

bk2204:
Thanks for the information.  I've opened #3785 to fix that.

gdlxn-ibm:
Thanks for the quick response.  I'll test with the fix for #3785 and let you know how it goes.

gdlxn-ibm:
The fix for #3785 works for me.  Thanks for the quick turn around on getting this issue fixed.

Issue #1755, opened 2016-12-10
`git bundle` support or an "offline" workflow?


christian-korneck:
`git bundle` is a a great git functionality to bundle individual commits into a container file that can then be used to update a git repository that is located on a computer which cannot directly reach the originating repository or the repository manager (i.e. `github.com`). I use it regularly to "pull" github repositories from isolated computers that don't have Internet access.

However, these bundle files don't include files tracked in `git-lfs`.

Are there any workarounds to achieve the same with `git-lfs`?
Are there any plans to add `git bundle` support to `git-lfs` (i.e. something like `git lfs bundle`)?

ttaylorr:
Adding a `git lfs bundle` command would certainly be possible, but isn't something that's immediately on our roadmap. As always, the core team would welcome and assist in reviewing a pull-request to add support.

ttaylorr:
Hi, circling back on the above. It is still the case that we do not have immediate plans to work on something like this, and I would still be more than happy to help anyone who is interested and does have the time. For now, since this isn't actively on my radar (or anyone else on `@git-lfs/core`, from my knowledge), I am going to close this.

christian-korneck:
:-(

zhengyongtao:
I have the same needs, haven't I planned to achieve them after so many years?

dvdveer:
I use a shell script to workaround this issue. It basically creates a tar archive with all the changed LFS files between the commits you want to bundle. When you want to unbundle you should extract the tar file before running git pull.

I've created a gist of it here: https://gist.github.com/dvdveer/a47966db83af115e6200124f92e22761

christian-korneck:
@dvdveer thanks for sharing

zhengyongtao:
I used a similar way to solve,but this way can only package the resources of the current branch. What should I do with the resources of other branches? @dvdveer

dvdveer:
The script that I shared also works with other branches. The only condition is that the lfs objects are already downloaded in your local repository. I solve this by running `git lfs fetch --all` before running the script, but this may not be feasible if you have a large total size of LFS objects. Unfortunately, I do not know of a git-lfs command that downloads an lfs object when you have it's hash. As far as I know the `git lfs fetch` command only takes git commits as argument.

ModischFabrications:
[...]

PR #3764, opened 2019-08-09
ls-files: add support for reference range

dvdveer:
Add an option to the `ls-files` command where the user can specify a commit range. This will return all the lfs files which changed between those two commits.

This change is based on the aborted pull request #3151 by @larsxschneider. I finished implementation and added unit tests and documentation.

As background information, the reason for this option is that I have a script that creates a git bundle. To workaround the missing bundle support in lfs (see #1755), I want to know which lfs objects were changed/created in the commit range of the bundle. Currently, in the script, I walk over all commits in the range and do a compare of the `ls-files` output. This is very slow however, so if `ls-files` has support for a commit range that would help me a lot.

### DIFF
--- a/docs/man/git-lfs-ls-files.1.ronn
+++ b/docs/man/git-lfs-ls-files.1.ronn
@@ -3,12 +3,16 @@ git-lfs-ls-files(1) -- Show information about Git LFS files in the index and wor
 
 ## SYNOPSIS
 
-`git lfs ls-files` [<ref>]
+`git lfs ls-files` [<ref>]<br>
+`git lfs ls-files` <ref> <ref>
 
 ## DESCRIPTION
 
 Display paths of Git LFS files that are found in the tree at the given
 reference.  If no reference is given, scan the currently checked-out branch.
+If two references are given, the LFS files that are modified between the two
+references are shown; deletions are not listed.
+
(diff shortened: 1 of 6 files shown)


docs/man/git-lfs-ls-files.1.ronn
```diff
@@ -3,12 +3,15 @@ git-lfs-ls-files(1) -- Show information about Git LFS files in the index and wor
...
-`git lfs ls-files` [<ref>]
+`git lfs ls-files` [<ref>]<br>
+`git lfs ls-files` <ref> <ref>
 
 ## DESCRIPTION
 
 Display paths of Git LFS files that are found in the tree at the given
 reference.  If no reference is given, scan the currently checked-out branch.
+If two references are given, the LFS files that are modified between the two
+references are shown.
```
bk2204: It looks like your test checks that we don't print deletions. That's fine, but we probably want to mention it here, since Git considers a deletion to be a modification. Maybe add a semicolon and “deletions are not listed”?

dvdveer: Sounds good! What do you prefer, that I create a new commit or amend the original commit?

bk2204: If you could amend the original commit, that would be great.

PR #3765, opened 2019-08-09
Support GIT_ALTERNATE_OBJECT_DIRECTORIES

bk2204:
There are two ways to specify an alternate in Git: via the repository file or via the environment. We've long supported the former, but not the latter. Add support for the environment variable form to ensure that all the standard use cases work.

During the writing of this series, I discovered an existing integration test for alternates that wasn't being run because it was named differently than the other tests. It's been renamed and adjusted so it now runs automatically and it's the basis for the test for this series.

Fixes git-lfs/gitobj#14.
/cc @rsrchboy as reporter

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -54,7 +54,7 @@ func checkoutCommand(cmd *cobra.Command, args []string) {
 	meter.Direction = tq.Checkout
 	meter.Logger = meter.LoggerFromEnv(cfg.Os)
 	logger.Enqueue(meter)
-	chgitscanner := lfs.NewGitScanner(func(p *lfs.WrappedPointer, err error) {
+	chgitscanner := lfs.NewGitScanner(cfg, func(p *lfs.WrappedPointer, err error) {
 		if err != nil {
 			LoggedError(err, "Scanner error: %s", err)
 			return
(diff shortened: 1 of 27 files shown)


nhatkhai:
Hi, about this GIT_ALTERNATE_OBJECT_DIRECTORIES. I think Git LFS interpreted it incorrectly. It mean to be one or multiple objects folders separated by `:`. As my current version `git-lfs/2.13.3 (GitHub; windows amd64; go 1.16.2; git a5e65851)` - It only work for one folder. When specified two folders in warning during `git lfs fetch` as not understand, and ignore using it.
According this this link https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables

> GIT_ALTERNATE_OBJECT_DIRECTORIES is a colon-separated list (formatted like /dir/one:/dir/two:…) which tells Git where to check for objects if they aren’t in GIT_OBJECT_DIRECTORY. If you happen to have a lot of projects with large files that have the exact same contents, this can be used to avoid storing too many copies of them.

bk2204:
I believe we have tests for this case.  Note that it's only colon-separated on Unix; on Windows, it's [documented to be semicolon-separated](https://git-scm.com/docs/git#Documentation/git.txt-codeGITALTERNATEOBJECTDIRECTORIEScode), which we honour.

PR #3780, opened 2019-08-16
Use an absolute path for smudging

### LINKED ISSUE #3772: `git lfs pull` does not work on long path (windows)
This is a problem observed on windows 10 machines.
Some of our repos have some long file names.
While running locally we have no problems since we use the default smudge filter and just do `git clone ...` / `git checkout ...` / ...
But on the CI we have the machines provisioned with `git lfs install --skip-smudge --skip-repo --force` and let jenkins do his stuff =).
Most builds doesn't have problems but since we add the branch name in the workspace folder name used once in a while some path hit the 260 windows limit.

When hitting long paths `git lfs pull origin` will output:

```
Could not check out "alfasim_data/initial_aaaaaaaaa_ssss/dddddddddddddd/initial-fffffffffffffffffffffffffffffffffffffffffffffffffffffffff.ext"
Could not check out "alfasim_data/initial_aaaaaaaaa_ssss/dddddddddddddd/initial-ggggggggggggggggggggggggggggggggggggggggggggggggggggggggg.ext"
...
Downloading LFS objects: 100% (146/146), 137 MB | 33 MB/s
```

Return `0` as error code and left the mentioned files unchanged.

Doing the same procedure using a short workspace folder name does not show problems.
Note that the error messages have been collected locally using makeshift names since jenkins itself show no outputs for this error and the jobs will just fail when some of those files are used.

With `git lfs install --force` clone and checkout will apply correctly the filters.

bk2204:
On Windows, paths are limited to 260 characters with two exceptions: manifested applications on Windows 10 may use long path, and all systems may use UNC paths, which must be absolute. Since we are not currently using a manifest, use an absolute path when smudging to make so that we can take advantage of UNC paths on Windows.

Fixes #3772

### DIFF
--- a/lfs/gitfilter_smudge.go
+++ b/lfs/gitfilter_smudge.go
@@ -28,7 +28,12 @@ func (f *GitFilter) SmudgeToFile(filename string, ptr *Pointer, download bool, m
 		defer os.Chmod(filename, stat.Mode())
 	}
 
-	file, err := os.Create(filename)
+	abs, err := filepath.Abs(filename)
+	if err != nil {
+		return fmt.Errorf("Could not produce absolute path for %q", filename)
+	}
+
+	file, err := os.Create(abs)
 	if err != nil {
 		return fmt.Errorf("Could not create working directory file: %v", err)
 	}


markuskramerIgitt:
Hi @bk2204 , 

I have to `skip smudge` to workaround the [4GB file limit issue](https://github.com/git-lfs/git-lfs/issues/2434).

When I clone my repo I reproducible get only errors related to manifests and dll:

could not check out "Operating Systems/Windows 10 Enterprise SAC 1809/SW_DVD9_Win_Pro_Ent_Edu_N_10_1809/sources/replacementmanifests/microsoft-certificateservices-webenrollment-serverupgrade-replacement.man"
could not check out "Operating Systems/Windows 10 Enterprise SAC 1809/SW_DVD9_Win_Pro_Ent_Edu_N_10_1809/sources/replacementmanifests/microsoft-client-license-platform-service-migration/clipmigplugin.dll"

I first thought that this is related to locking, so I executed
```
$ git lfs locks
Error while retrieving locks: Repository or object not found: https://artifactory.fantasy.com/artifactory/api/lfs/fantasy-local/locks?refspec=refs%2Fheads%2Fmaster
Check that it exists and that you have proper access to it
```
So I assume there are no locks.

The above path are only 185 characters,  but could the errors  related to `skip smudge` and absolute path?

PR #3786, opened 2019-08-21
Fix couple of 'the the' typos

pmeerw:


pmeerw:
thank you for considering the patch; I've dropped changes under vendor/

PR #3790, opened 2019-08-25
Add clonefile on Windows over ReFS support.

kazuki-ma:
ReFS ([wikipedia](https://en.wikipedia.org/wiki/ReFS)) is filesystem on Windows OS and it's support copy-on-write / file cloning since v3.2.
ReFS v3.2 is available since Windows Server 2016, Windows Server 2019 and Windows 10 also supports format & read-write.

I think many asset heavy product CI system and developer start using ReFS if git-lfs support dedup on it.

Reference:

* https://docs.microsoft.com/en-us/windows/win32/fileio/block-cloning

This closes https://github.com/git-lfs/git-lfs/issues/3791

### DIFF
--- a/tools/util_windows.go
+++ b/tools/util_windows.go
@@ -1,0 +1,158 @@
+// +build windows
+
+package tools
+
+import (
+	"io"
+	"io/ioutil"
+	"os"
+	"unsafe"
+
+	"golang.org/x/sys/windows"
+)
+
+var (
+	availableClusterSize = []int64{64 * 1024, 4 * 1024} // ReFS only supports 64KiB and 4KiB cluster.
+	GiB                  = int64(1024 * 1024 * 1024)
+)
+
+// fsctlDuplicateExtentsToFile = FSCTL_DUPLICATE_EXTENTS_TO_FILE IOCTL
+// Instructs the file system to copy a range of file bytes on behalf of an application.
+//
+// https://docs.microsoft.com/windows/win32/api/winioctl/ni-winioctl-fsctl_duplicate_extents_to_file
+const fsctlDuplicateExtentsToFile = 623428
+
+// duplicateExtentsData = DUPLICATE_EXTENTS_DATA structure
+// Contains parameters for the FSCTL_DUPLICATE_EXTENTS control code that performs the Block Cloning operation.
+//
+// https://docs.microsoft.com/windows/win32/api/winioctl/ns-winioctl-duplicate_extents_data
+type duplicateExtentsData struct {
+	FileHandle       windows.Handle
+	SourceFileOffset int64
+	TargetFileOffset int64
+	ByteCount        int64
+}
+
+// CheckCloneFileSupported runs explicit test of clone file on supplied directory.
+// This function creates some (src and dst) file in the directory and remove after test finished.
+//
+// If check failed (e.g. directory is read-only), returns err.
+func CheckCloneFileSupported(dir string) (supported bool, err error) {
+	src, err := ioutil.TempFile(dir, "src")
+	if err != nil {
+		return false, err
+	}
+	defer os.Remove(src.Name())
+
+	// Make src file not empty.
+	// Because `FSCTL_DUPLICATE_EXTENTS_TO_FILE` on empty file is always success even filesystem don't support it.
+	_, err = src.WriteString("TESTING")
+	if err != nil {
+		return false, err
+	}
+
+	dst, err := ioutil.TempFile(dir, "dst")
+	if err != nil {
+		return false, err
+	}
+	defer os.Remove(dst.Name())
+
+	return CloneFile(dst, src)
+}
+
+func CloneFileByPath(dst, src string) (success bool, err error) {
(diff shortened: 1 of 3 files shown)


tools/util_windows.go
```diff
@@ -0,0 +1,163 @@
...
+var (
+	availableClusterSize = []int64{64 * 1024, 4 * 1024} // ReFS only supports 64KiB and 4KiB cluster.
+	GiB                  = int64(1024 * 1024 * 1024)
+)
+
+// _FSCTL_DUPLICATE_EXTENTS_TO_FILE :
+// Instructs the file system to copy a range of file bytes on behalf of an application.
+//
+// https://docs.microsoft.com/windows/win32/api/winioctl/ni-winioctl-fsctl_duplicate_extents_to_file
+const _FSCTL_DUPLICATE_EXTENTS_TO_FILE = 623428
```
bk2204: Even if Windows names this with a leading underscore, we probably don't want to.

tools/util_windows.go
```diff
@@ -0,0 +1,163 @@
...
+
+// _DUPLICATE_EXTENTS_DATA :
+// Contains parameters for the FSCTL_DUPLICATE_EXTENTS control code that performs the Block Cloning operation.
+//
+// https://docs.microsoft.com/windows/win32/api/winioctl/ns-winioctl-duplicate_extents_data
+type _DUPLICATE_EXTENTS_DATA struct {
```
bk2204: Same issue about the leading underscore.  Since this is not supposed to be accessible outside of this package, we probably want to call it `duplicateExtentsData`.

Is there a way for us to get the definition of this struct and constant from a library so that we don't have to implement it ourselves?

tools/util_windows.go
```diff
@@ -0,0 +1,163 @@
...
+package tools
+
+import (
+	"io"
+	"io/ioutil"
+	"os"
+	"unsafe"
+
+	"github.com/rubyist/tracerx"
+	"golang.org/x/sys/windows"
```
bk2204: Do we need to add this to `go.mod`, or is it already brought in by the `golang.org/x/sys` package?

tools/util_windows.go
```diff
@@ -0,0 +1,163 @@
...
+//
+// If check failed (e.g. directory is read-only), returns err.
+func CheckCloneFileSupported(dir string) (supported bool, err error) {
+	src, err := ioutil.TempFile(dir, "src")
+	if err != nil {
+		return false, err
+	}
+	defer os.Remove(src.Name())
+
+	_, err = src.WriteString("TESTING") // Non-empty file is required for correct result.
```
bk2204: Is a non-empty file required just for this case or for the function itself to work correctly?  If the latter, we probably want to check for a non-zero size below.

tools/util_windows.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	fileSize := srcStat.Size()
+
+	err = dst.Truncate(fileSize) // set file size. Thre is a requirements "The destination region must not extend past the end of file."
+	if err != nil {
+		return
+	}
+
+	tracerx.Printf("dst:%s src:%s\n", dst, src)
```
bk2204: This looks like it may be debugging information that should be removed.

tools/util_windows.go
```diff
@@ -0,0 +1,163 @@
...
+		windows.Handle(dst.Fd()),
+		_FSCTL_DUPLICATE_EXTENTS_TO_FILE,
+		(*byte)(unsafe.Pointer(&request)),
+		uint32(unsafe.Sizeof(request)),
+		(*byte)(unsafe.Pointer(nil)), // = nullptr
+		0,
+		&bytesReturned,
+		&overlapped)
+
+	tracerx.Printf("DeviceIoControl:%+v result:%+v\n", request, err)
```
bk2204: This also looks like perhaps it's debugging information.

tools/util_windows_test.go
```diff
@@ -0,0 +1,111 @@
+// +build windows
+
+package tools
+
+import (
+	"crypto/sha1"
```
bk2204: If we're going to use a hash for comparing two files, let's use SHA-256.  SHA-1 is considered weak and we'd like to use less of it in our codebase, and we already use SHA-256 in a lot of places.

tools/util_windows.go
kazuki-ma: We don't add entry to `go.mod`. It's part of [golang.org/x/sys](https://godoc.org/golang.org/x/sys).

tools/util_windows.go
kazuki-ma: Calling `FSCTL_DUPLICATE_EXTENTS_TO_FILE` to empty file is always success even filesystem don't support it.
To check filesystem support `FSCTL_DUPLICATE_EXTENTS_TO_FILE` or not, we have to use non-empty file.

I'll write those as comment.

tools/util_windows_test.go
kazuki-ma: Got it. Thank you.

kazuki-ma:
Updated. Please review again.

kazuki-ma:
Thank you very much for reviewing.

PR #3796, opened 2019-08-27
NON-ISSUE Use FICLONE instead of BTRFS_IOC_CLONE.

kazuki-ma:
Sorry it's my miss-understanding that we should use `FICLONE` first and fallback to `BTRFS_IOC_CLONE`. Actually, `FICILONE` is equal to `BTRFS_IOC_CLONE`.

So we already support copy-on-write on XFS not only Btrfs.

In this commit, let me use standarized name `FICILONE` instead of `BTRFS_IOC_CLONE`.

FICILONE definition:
----

`#define FICLONE                _IOW(0x94, 9, int)`  in https://github.com/torvalds/linux/blob/v5.2/include/uapi/linux/fs.h#L195

BTRFS_IOC_CLONE definition:
----

`#define BTRFS_IOC_CLONE        _IOW(BTRFS_IOCTL_MAGIC, 9, int)` in https://github.com/torvalds/linux/blob/v5.2/include/uapi/linux/btrfs.h#L850

and

`#define BTRFS_IOCTL_MAGIC 0x94` in https://github.com/torvalds/linux/blob/v5.2/include/uapi/linux/btrfs.h#L25

### DIFF
--- a/tools/util_linux.go
+++ b/tools/util_linux.go
@@ -5,10 +5,10 @@ package tools
 /*
 #include <sys/ioctl.h>
 
-#undef BTRFS_IOCTL_MAGIC
-#define BTRFS_IOCTL_MAGIC 0x94
-#undef BTRFS_IOC_CLONE
-#define BTRFS_IOC_CLONE _IOW (BTRFS_IOCTL_MAGIC, 9, int)
+#undef FICLONE
+#define FICLONE		_IOW(0x94, 9, int)
+// copy from https://github.com/torvalds/linux/blob/v5.2/include/uapi/linux/fs.h#L195 for older header files.
+// This is equal to the older BTRFS_IOC_CLONE value.
 */
 import "C"
 
@@ -20,7 +20,7 @@ import (
 )
 
 const (
-	BtrfsIocClone = C.BTRFS_IOC_CLONE
+	ioctlFiClone = C.FICLONE
 )
 
 // CheckCloneFileSupported runs explicit test of clone file on supplied directory.
@@ -51,7 +51,7 @@ func CloneFile(writer io.Writer, reader io.Reader) (bool, error) {
 	fdst, fdstFound := writer.(*os.File)
 	fsrc, fsrcFound := reader.(*os.File)
 	if fdstFound && fsrcFound {
-		if _, _, err := syscall.Syscall(syscall.SYS_IOCTL, fdst.Fd(), BtrfsIocClone, fsrc.Fd()); err != 0 {
+		if _, _, err := syscall.Syscall(syscall.SYS_IOCTL, fdst.Fd(), ioctlFiClone, fsrc.Fd()); err != 0 {
 			return false, err
 		}
 		return true, nil


tools/util_linux.go
```diff
@@ -48,15 +50,32 @@ func CheckCloneFileSupported(dir string) (supported bool, err error) {
...
+		return false, nil
+	}
+
+	// Try FICLONE, this supported since kernel 4.5.
+	// This can work with XFS, Btrfs and other format which support copy on write.
+	err := syscall.Syscall(syscall.SYS_IOCTL, dstFD.Fd(), C.FICLONE, srcFD.Fd())
+	if err == nil {
+		return true, nil // success
+	} else { // error case
+		if err != C.EINVAL {
```
bk2204: Can we move these into one else-if statement?  That would keep things tidy.

We could even bring the `BTRFS_IOC_CLONE` case up as an additional else-if case, although that might require that we switch from `err == 0` back to `err != 0`.

kazuki-ma: Sorry I miss-understand about `FICLONE` and `BTRFS_IOC_CLONE`.
I updated my PR, could you review again. Thank you.

tools/util_linux.go
```diff
@@ -5,10 +5,11 @@ package tools
...
 
-#undef BTRFS_IOCTL_MAGIC
-#define BTRFS_IOCTL_MAGIC 0x94
-#undef BTRFS_IOC_CLONE
-#define BTRFS_IOC_CLONE _IOW (BTRFS_IOCTL_MAGIC, 9, int)
+#undef FICLONE
+#define FICLONE		_IOW(0x94, 9, int)
+// copy from https://github.com/torvalds/linux/blob/v5.2/include/uapi/linux/fs.h#L195
+// for older header files.
+// This is eaual to BTRFS_IOC_CLONE. https://github.com/torvalds/linux/blob/v5.2/include/uapi/linux/btrfs.h#L850
```
bk2204: This should probably say "equal".  I'm also not sure we need the references; it's probably fine to just say something like, "This is equal to the older BTRFS_IOC_CLONE value."

tools/util_linux.go
```diff
@@ -48,15 +49,19 @@ func CheckCloneFileSupported(dir string) (supported bool, err error) {
 }
 
 func CloneFile(writer io.Writer, reader io.Reader) (bool, error) {
-	fdst, fdstFound := writer.(*os.File)
-	fsrc, fsrcFound := reader.(*os.File)
-	if fdstFound && fsrcFound {
-		if _, _, err := syscall.Syscall(syscall.SYS_IOCTL, fdst.Fd(), BtrfsIocClone, fsrc.Fd()); err != 0 {
```
bk2204: It looks like the old and new code are intended to be equivalent, but that's hard to see from the diff.  To make it easier for future readers trying to reason about this code, could we limit the refactoring here and simply change the constant `BtrfsIocClone` to `ioctlFiClone`?

kazuki-ma: Good suggesting. Sorry for refactoring contamination. Let' me do that.

kazuki-ma:
updated, please review again.

PR #3800, opened 2019-08-28
Avoid deadlock when transfer queue fails

### LINKED ISSUE #3798: t-push-bad-dns gets stuck if no network is available
I'm trying to update the Fedora package for git-lfs to 2.8.0, but the tests are getting stuck. With `prove` running in parallel, everything runs except for `t-push-bad-dns.sh`, which seemingly gets stuck somewhere, and never finishes after an hour.

If I run the build in a chroot with networking enabled, then the test works. However, the official builders have no network access, so this is not an acceptable workaround.

So far, I have bisected the problem to 1412d6e47af351d678f50cf9a5d6e0bc9c92c919.

bk2204:
In 1412d6e4 ("Don't fail if we lack objects the server has", 2019-04-30), we changed the code to abort later if a missing object occurs.  In doing so, we had to consider the case where the transfer queue aborts early for some reason and ensure that the sync.WaitGroup does not unnecessarily block due to outstanding objects never getting processed.

However, the approach we used, which was to explicitly add the number of items we skipped processing, was error prone and didn't cover all cases. Notably, a DNS failure could randomly cause a hang during a push.  Solve this by creating a class for a wait group which is abortable and simply abort it if we encounter an error, preventing any deadlocks caused by miscounting the number of items.

There's no test for this case because triggering it requires network to be disabled, which isn't something which we can do in our testsuite.  As reported, existing tests do sporadically fail in this case, so when run in a suitable environment, our testsuite already covers this case. 

/cc @QuLogic as reporter
Fixes #3798.

### DIFF
--- a/tq/transfer_queue.go
+++ b/tq/transfer_queue.go
@@ -123,6 +123,43 @@ func (b batch) Len() int           { return len(b) }
 func (b batch) Less(i, j int) bool { return b[i].Size < b[j].Size }
 func (b batch) Swap(i, j int)      { b[i], b[j] = b[j], b[i] }
 
+type abortableWaitGroup struct {
+	wq      sync.WaitGroup
+	counter int
+	mu      sync.Mutex
+}
+
+func newAbortableWaitGroup() *abortableWaitGroup {
+	return &abortableWaitGroup{}
+}
+
+func (q *abortableWaitGroup) Add(delta int) {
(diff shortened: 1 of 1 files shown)


QuLogic:
This [works](https://koji.fedoraproject.org/koji/buildinfo?buildID=1368184) for me.

PR #3812, opened 2019-09-09
Fix Remote Name Parsing Bug

nataliechen1:
git config allows "." in subsection name but not in key name.
while splitting key with "." into parts, config key name should be the last part and subsection name should be parts from the second to the second-to-last joint by ".".

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -330,3 +330,13 @@ func TestCurrentTimestamp(t *testing.T) {
 		}
 	}
 }
+
+func TestRemoteNameWithDotDefault(t *testing.T) {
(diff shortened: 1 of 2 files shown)


nataliechen1:
I ran the failed test, 'credentials from lfs.url', on my machine (ubuntu 18.04) with the same go version (1.12.6) and git version (2.23.0.162.gf1d4a28250) but the test was successful.

bk2204:
Yeah, that's one of our flaky tests.  I tried to kick off a re-run, but apparently that doesn't work with PRs.  I think this is fine, and I'm happy to merge it.

Thanks again for the patch and for including a test for the fix.

PR #3813, opened 2019-09-09
Do not fail when multiple processes download the same lfs file

### LINKED ISSUE #2825: shared lfs.storage is not threadsafe
When multiple same modules share one lfs storage (by git config lfs.storage=STORAGE). Access from multiple instances with command 'git lfs fetch' fails. But it's expected to shared the storage.

slonopotamus:
When several git-lfs processes use the same LFS storage simultaneously,
there's a race condition between checking whether LFS file exists and renaming temporary
file to final name in LFS storage. When rename fails, check that target file exists.
If it does, pretend that download has finished successfully.

Fixes #2825

----

I'm not sure how to write a test for this.

### DIFF
--- a/tq/basic_download.go
+++ b/tq/basic_download.go
@@ -242,7 +242,12 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
 		return fmt.Errorf("Expected OID %s, got %s after %d bytes written", t.Oid, actual, written)
 	}
 
-	return tools.RenameFileCopyPermissions(dlfilename, t.Path)
+	err = tools.RenameFileCopyPermissions(dlfilename, t.Path)
+	if _, err2 := os.Stat(t.Path); err2 == nil {
+		// Target file already exists, possibly was downloaded by other git-lfs process
+		return nil
+	}
+	return err
 }
 
 func configureBasicDownloadAdapter(m *Manifest) {


slonopotamus:
cicleci failure is not related to my changes.

tq/basic_download.go
```diff
@@ -242,7 +242,12 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
 		return fmt.Errorf("Expected OID %s, got %s after %d bytes written", t.Oid, actual, written)
 	}
 
-	return tools.RenameFileCopyPermissions(dlfilename, t.Path)
+	err = tools.RenameFileCopyPermissions(dlfilename, t.Path)
+	if _, err2 := os.Stat(t.Path); err2 == nil {
```
bk2204: Could we change this check to look like the following instead?  That would check for the error which is problematic on Windows and ignore it explicitly instead of assuming that it's the cause if the file already exists.

I assume that this is only a problem on Windows, since Unix systems allow replacing files in use.

```
if os.IsExist(err) {
		// Target file already exists, possibly was downloaded by other git-lfs process
		return nil
}
```

slonopotamus: 1. Sure, will update PR soon.
2. I don't think you can rename a file on top of already existing one even on Linux.

slonopotamus: OK, to be more exact:

On Windows, os.IsExist is:
```
func isExist(err error) bool {
	err = underlyingError(err)
	return err == syscall.ERROR_ALREADY_EXISTS ||
		err == syscall.ERROR_DIR_NOT_EMPTY ||
		err == syscall.ERROR_FILE_EXISTS || err == ErrExist
}
```

However, "The process cannot access the file because it is being used by another process." is `ERROR_SHARING_VIOLATION == 32` (Go doesn't have such constant at all).

slonopotamus:
Rebased onto fresh master to fix circleci breakage.

Issue #3797, opened 2019-08-27
"git commit" of single file takes 5 minutes, mounted fileystem/diskimage with 50G GIT repo + 900G of builds articles


sarvi:
[gitcommitexp.log](https://github.com/git-lfs/git-lfs/files/3549053/gitcommitexp.log)
Some context
  We have a 50G GIT/GIT-LFS based repository. We then build about 900G of build articles.
To accelerate development, we copy this workspace into a EXT4 formatted disk image and snapshot it
Developers clone this disk image quickly, mount it and they have a fully built workspace that they can do incremental builds on.

Problem 1: resolved
after mounting the diskimage, "git status" and "git diff" used to take 40 minutes.
After some discussions on thhe GIT alias, I was asked do a "git update-index --refresh" after copying over the GIT workspace to the diskimage. It was suggested that the git-index that tracks source tree information included node information that would not be valid after it GIT wokspace was copied into a new filesystem.
So we now do a "git update-index --refresh"  in the diskimage after git workspace copy and before we snapshot. After this the cloned/mounted diskimage performed better. First time "git status/diff" is now down to 1m30 seconds, subsequence commands were always in <10 seconds.

GIT Thread: https://marc.info/?t=156649700700001&r=1&w=2

Problem 2:
Now in the above workspace, if I now try to do a local commit of a simple 1 line code change, It takes about 5 minutes.
 "git commit -m "dummy commit" of a 1 line change in 1 file takes about 5-6 \
minutes, everytime in this workspace. Tracing shows a whole bunch. 

The entire 5-6 \
minutes worth of the following sort of trace logs. 3:13:50.320930 trace git-lfs: \
filepathfilter: rejecting "x/y/z.o.command" via [] 13:13:50.320940 trace git-lfs: \
filepathfilter: accepting " x/y/z.o.command " 13:13:50.320862 trace git-lfs: \
filepathfilter: rejecting "a/b/c/d.o.command" via [] 13:13:50.320972 trace git-lfs: \
filepathfilter: accepting " a/b/c/d..o.command"

The git folks suggest this is not a GIT problem but might be git-lfs

> These are messages from the Git LFS tool, and this list does not provide
> support for Git LFS, since it is a separate project not affiliated with
> Git.
> 
> I would recommend that you create an issue on Git LFS's tracker at:
> 
>   https://github.com/git-lfs/git-lfs/issues/new
> 
> and mention that you are seeing lots of 'filepathfilter' logs. My
[...]

PR #3823, opened 2019-09-17
Reworked to use git ls-files in some circumstances instead of FastWalkGitRepo

SeamusConnor:
This is a rough cut of addressing #3797 by replacing FastWalkGitRepo with `git ls-files`. For my use case, it cuts post-hook execution time down from ~40 seconds, to a few hundred ms.

At this point, I am looking for a high-level review of the changes, and direction on wether I should proceed w/ completing this change.

There are a few things that need to be addressed:

1. The new code needs tests.
2. Some existing tests are failing.
3. The new behavior needs to be verified against the old behavior on more repos.

I am also requesting any feedback that anyone familiar w/ the code might have. As it stands, git-lfs is unusable for our project without this (or a similar) fix, and I would like to verify that this fix is sound before we use it internally.

### DIFF
--- a/git/ls_files.go
+++ b/git/ls_files.go
@@ -1,0 +1,90 @@
+package git
+
+import (
+	"bufio"
+	"strings"
+	"path"
+	"io/ioutil"
+
+	"github.com/git-lfs/git-lfs/errors"
+	"github.com/git-lfs/git-lfs/tools"
+	"github.com/rubyist/tracerx"
+)
+
+type lsFileInfo struct {
+	BaseName string
+	FullPath string
+}
+type LsFiles struct {
+	Files map[string]*lsFileInfo
+	FilesByName map[string][]*lsFileInfo
+}
+
+func NewLsFiles(workingDir string, standardExclude bool) (*LsFiles, error) {
+
+	args := []string{
+		"ls-files",
+		"-z", // Use a NUL separator. This also disables the escaping of special characters.
+		"--others",
+		"--cached",
+	}
+
+	if standardExclude {
+		args = append(args, "--exclude-standard")
+	}
+	cmd := gitNoLFS(args...)
+	cmd.Dir = workingDir
+
+    tracerx.Printf("NewLsFiles: running in %s git %s",
+			workingDir, strings.Join(args, " "))
+
+	// Capture stdout and stderr
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, err
+	}
+	stderr, err := cmd.StderrPipe()
+	if err != nil {
+		return nil, err
+	}
+
+	scanner := bufio.NewScanner(stdout)
+	scanner.Split(tools.SplitOnNul)
+
+	if err := cmd.Start(); err != nil {
+		return nil, err
+	}
+
+	rv := &LsFiles {
+		Files: make(map[string]*lsFileInfo),
+		FilesByName: make(map[string][]*lsFileInfo),
+	}
+
+	// Setup a goroutine to drain stderr as large amounts of error output may cause
+	// the subprocess to block.
+	errorMessages := make(chan []byte)
+	go func() {
+		msg, _ := ioutil.ReadAll(stderr)
+		errorMessages <- msg
+	}()
+
+	// Read all files
+	for scanner.Scan() {
+		base := path.Base(scanner.Text())
+		finfo := &lsFileInfo {
+			BaseName: base,
+			FullPath: scanner.Text(),
+		}
+		rv.Files[scanner.Text()] = finfo
+		rv.FilesByName[base] = append(rv.FilesByName[base], finfo)
+	}
+
+	// Check the output of the subprocess, output stderr if the command failed.
+	msg := <-errorMessages
+	if err := cmd.Wait(); err != nil {
+		return nil, errors.Errorf("Error in git %s: %v %s",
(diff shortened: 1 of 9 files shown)


bk2204:
The approach I'm taking, which I'm still waiting on feedback on, is the `ls-files-filter` branch on bk2204/git-lfs.  It does pass the testsuite, although it has the limitations that it doesn't handle the case where users lock ignored files and it doesn't handle files with newlines (which we don't handle anyway at the moment).

The general approach of using `git ls-files` is definitely sound, but I'm not sure how it performs on large repos.  I'm not opposed to taking a patch that you write up if you like, although if you're happy with my patch, it may be easier for me to rework and submit.

SeamusConnor:
RE large repos: I think the limiting factor is going to be the size of the working tree, and the complexity of the gitignores. To that end, our repo is the result of an SVN conversion, and it is fairly large at 278,000 files, with a gitignore over 3k lines.

I will take a look at your changes, and see where we differ. If you are open to accepting a PR, I will keep working on this.

SeamusConnor:
I spent a minute and reviewed your change. It looks pretty similar to mine, except you are hitting the filesystem a little more.

I think I need to understand a little more:

1. When searching for .gitattributes: What are the requirements here? AFAIU, we should be looking for all git attributes files that are either staged, or committed to the repository. Is that correct? What about .gitattributes files that have been created but not yet staged?
2. Fixing up file locks: Does a file have to be staged/commited to be "locked"? If so, could the calls to stat be eliminated? For reference, it is about 10x faster if `ls-files` doesn't have to traverse the work tree.
3. In your change, you are now stepping through the output of ls-files one at at time and calling the callbacks sequentially. What do you supposed the performance impact of this will be on querying the server for file locks?

bk2204:
For 1, we need to honor `.gitattributes` files that are in the tree but not yet staged.  We have tests that check for this case and require it to work.

For 2, it isn't practically useful to lock a file that isn't yet committed, since nobody else will be contending over it.  I don't think we have to support that case.

For 3, I don't believe we're querying the server if you're talking about the code in `locking/lockable.go`.  There, we're just looking at the `.gitattributes` file for the `lockable` attribute and setting those files read only.

I don't think we intrinsically need to call `stat` in this case.  I can imagine an additional commit on top of my change that instead of passing an `os.FileInfo`, changes the code to never run the callback on directories and instead passes just the name of the file in place of that argument.  I don't believe we actually care about the directories in any case where we traverse and it would be cheaper, as you point out.

SeamusConnor:
I ran some trivial bench marking on the two change sets:

```
# With your change set, including all of the calls to stat.
$ time git checkout master
Already on 'master'
Your branch is up to date with 'origin/master'.

real    0m6.902s
user    0m2.746s
sys     0m2.729s

# With my change set
$ time git checkout master
Already on 'master'
Your branch is up to date with 'origin/master'.

real    0m0.673s
user    0m0.716s
sys     0m0.553s
```

My guess is that the majority of the difference is the calls to stat. I will rework my change so the test cases pass. My guess is it will end up looking very similar to your change, except for the calls to stat, and the NUL separated output from `git ls-files`.

SeamusConnor:
Updated benchmark:
```
$ time git checkout master
Already on 'master'
Your branch is up to date with 'origin/master'.

real    0m1.741s
user    0m1.568s
sys     0m0.692s
```
As expected, it is faster than running stat on all of the files, but not as fast as the initial version because it has to scan the working directory (`git ls-files --other`).

SeamusConnor:
Not sure why the windows ci build is failing. Is that expected?

slonopotamus:
I believe there's something broken with Windows CI. Reported as #3828.

slonopotamus:
@SeamusConnor Please, rebase onto latest `master` to make sure tests pass on Windows.

SeamusConnor:
Alright, looks like whatever was done to the windows CI build on master fixed my issue.

git/ls_files.go
```diff
@@ -0,0 +1,90 @@
...
+
+	// Read all files
+	for scanner.Scan() {
+		base := path.Base(scanner.Text())
+		finfo := &lsFileInfo {
+			BaseName: base,
+			FullPath: scanner.Text(),
+		}
+		rv.Files[scanner.Text()] = finfo
+		rv.FilesByName[base] = append(rv.FilesByName[base], finfo)
```
bk2204: I think this approach has the real possibility of using a lot of memory on large repositories, since we're storing data twice for each file instead of operating incrementally.  I think we'd want to continue to do things incrementally like we do for the current walker.

SeamusConnor: Yeah, I considered that when I did it. For my repo (which IMO is quite large w/ a working tree of ~300k files), `ls-files -z -o --cached` produced around 20 MiB of data. I made a completely untested tradeoff for speed here, with the assumption that a callback or channel based approach would be slower. Perhaps the most prudent thing to do would be for me to try it with an iterative interface, and see what happens to the execution time/memory consumption.

slonopotamus: Largest git installation I worked on was 1M files with 1.5TB `.git/objects` (before git-lfs existed) and this was much beyond what any sane person would do. So if what you did works reasonable well for 300K files, I believe that's Good Enough to leave as is.

bk2204: Okay, let's go with it.  It's going to scale better than the current status quo, and it's not too difficult to fix if it doesn't.

PR #3825, opened 2019-09-19
Let git-lfs HTTPS transport send cookies

### LINKED ISSUE #3824: Git-lfs cannot access repositories behind Google Cloud Identity-Aware Proxy
To access service protected by Gcloud IAP a client need to send HTTP cookies (like `GCP_IAAP_AUTH_TOKEN_*`).

Git works fine with it after setting up the `http.cookieFile` configuration option, but git-lfs don't send HTTP cookies at all.

kupson:
This allows Git LFS to use the same cookies as configured for Git
(http.cookieFile). Those cookies may be needed for e.g. Gcloud Identity-Aware
Proxy.

Cookies are sent only over HTTPS.

Fixes #3824

### DIFF
--- a/lfshttp/cookies.go
+++ b/lfshttp/cookies.go
@@ -14,7 +14,7 @@ func isCookieJarEnabledForHost(c *Client, host string) bool {
 }
 
 func getCookieJarForHost(c *Client, host string) (http.CookieJar, error) {
-cookieFile, _ := c.uc.Get("http", fmt.Sprintf("https://%v", host), "cookieFile")
+	cookieFile, _ := c.uc.Get("http", fmt.Sprintf("https://%v", host), "cookieFile")
 
 	return cookie.NewJar(cookieFile)
 }
--- a/lfshttp/cookies.go
+++ b/lfshttp/cookies.go
@@ -1,0 +1,20 @@
+package lfshttp
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/google/slothfs/cookie"
+)
+
+func isCookieJarEnabledForHost(c *Client, host string) bool {
+	_, cookieFileOk := c.uc.Get("http", fmt.Sprintf("https://%v", host), "cookieFile")
+
+	return cookieFileOk
+}
+
+func getCookieJarForHost(c *Client, host string) (http.CookieJar, error) {
+cookieFile, _ := c.uc.Get("http", fmt.Sprintf("https://%v", host), "cookieFile")
+
+	return cookie.NewJar(cookieFile)
+}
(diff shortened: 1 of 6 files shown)


lfshttp/cookies.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+func isCookieJarEnabledForHost(c *Client, host string) bool {
+	_, cookieFileOk := c.uc.Get("http", fmt.Sprintf("https://%v", host), "cookieFile")
+
+	return cookieFileOk
+}
+
+func getCookieJarForHost(c *Client, host string) (http.CookieJar, error) {
+cookieFile, _ := c.uc.Get("http", fmt.Sprintf("https://%v", host), "cookieFile")
```
bk2204: This line doesn't seem to be indented properly.  Could you run `goimports` or `go fmt` to make sure your code is tidy?

lfshttp/cookies.go
```diff
@@ -0,0 +1,20 @@
+package lfshttp
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/google/slothfs/cookie"
```
bk2204: I assume we need this dependency because the builtin Go cookie jar can't read from a file?

lfshttp/cookies.go
kupson: Should be fixed now.

lfshttp/cookies.go
kupson: Yes. Git-lfs should read the same file as git and libcurl - a parser for Netscape cookie jar file format is needed (https://curl.haxx.se/docs/http-cookies.html).

The format is quite simple and the parser can be added directly to `lfshttp/cookies.go` if you would like to avoid additional dependencies.

t/t-clone.sh
```diff
@@ -665,3 +665,94 @@ begin_test "clone bare empty repository"
...
 )
 end_test
+
+begin_test "clone (HTTP server/proxy require cookies)"
+(
+  set -e
+
+  # golang net.http.Cookie ignores cookies with IP instead of domain/hostname
+  GITSERVER_SAVED="$GITSERVER"
+  GITSERVER=$(echo "$GITSERVER" | sed 's/127\.0\.0\.1/localhost/')
```
bk2204: This is in a subshell, so you need not save the old version and restore it.  Any variables you modify here are not persisted in other tests.

slonopotamus:
@kupson Please, rebase onto latest `master` to make sure tests pass on Windows

t/t-clone.sh
kupson: This should be fixed now.

PR #3833, opened 2019-09-23
ci: force Windows Git version to 2.22.0

### LINKED ISSUE #3828: Tests do not pass on Windows CI
See #3827

bk2204:
In Git for Windows 2.23, the /dev and /dev/fd directories did not get created.  As a result, it's not possible to use a process substitution in bash, which breaks a significant portion of the test suite.  In the mean time, let's downgrade Git for Windows to 2.22.0, which is what Actions used to use and worked for us.  We can remove this hack when the next Git for Windows release occurs.

Fixes #3828.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,8 @@ jobs:
     - run: set PATH=%GOPATH%\bin;%PATH%
     - run: cinst InnoSetup -y
     - run: cinst strawberryperl -y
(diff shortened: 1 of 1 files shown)


slonopotamus:
Interesting. Is there a bugreport/fix for this in Git for Windows? I just want to make sure that subsequent Git for Windows releases will actually fix things back.

slonopotamus:
Found it: git-for-windows/git#2291

PR #3849, opened 2019-10-08
Support fetching entire history of specific refs

pluehne:
### Motivation

I’m currently working on an automation that synchronizes Git repositories between remotes, some of which use Git LFS. Compared to `git lfs fetch --all`, I’d like to fetch all objects referenced in the entire commit history but, and here comes the difference, not for all refs but *only a specific set of refs.*

For this reason, I’d like extend `git lfs fetch` to support downloading all objects referenced by any commit reachable from a user-defined set of refs.

### Design

I suggest allowing `git lfs fetch` to accept a combination of the `--all` option and a list of refs specified as positional command-line arguments. This would then instruct Git LFS to download all objects referenced by any commit reachable from any of the specified refs (but no other objects).

For example, the call
```sh
$ git lfs fetch --all origin master develop v1.0.0
```
would download all objects referenced by *any commit* in the `master` and `develop` branches as well as the `v1.0.0` tag—not just the tips of those branches and tags. However, objects stemming from a `sandbox` branch or a `v0.1.0` tag would not be downloaded if they aren’t contained in `master`, `develop`, or `v1.0.0`.

In contrast,
```sh
$ git lfs fetch origin master develop v1.0.0
```
only downloads objects referenced by the tips of `master`, `develop`, and `v1.0.0` but none of the objects from earlier commits.

### Implementation

I based the implementation off the preexisting functions used when providing the `--all` option and when providing just a list of refs without `--all`. I’ve added tests covering the new usage of the `--all` option to show that indeed only objects referenced by commits reachable from the requested refs (but none aside from those) will be fetched.

### DIFF
--- a/docs/man/git-lfs-fetch.1.ronn
+++ b/docs/man/git-lfs-fetch.1.ronn
@@ -25,10 +25,11 @@ This does not update the working copy.
   that would otherwise be downloaded. See [RECENT CHANGES]
 
 * `--all`:
-  Download all objects referenced by any commit that is reachable; this is
-  primarily for backup / migration purposes. Cannot be combined with --recent or
-  --include/--exclude. Ignores any globally configured include and exclude paths
-  to ensure that all objects are downloaded.
+  Download all objects that are referenced by any commit reachable from the refs
+  provided as arguments. If no refs are provided, then all refs are pushed. This
+  is primarily for backup and migration purposes. Cannot be combined with
+  --recent or --include/--exclude. Ignores any globally configured include and
+  exclude paths to ensure that all objects are downloaded.
 
 * `--prune` `-p`:
   Prune old and unreferenced objects after fetching, equivalent to running
@@ -125,6 +126,11 @@ What changes are considered 'recent' is based on a number of gitconfig options:
 * Fetch the LFS objects for the current ref from a secondary remote 'upstream'
 
   `git lfs fetch upstream`
+
+* Fetch all the LFS objects from the default remote that are referenced by any
+  commit in the `master` and `develop` branches
+
+  `git lfs fetch --all master develop`
 
 * Fetch the LFS objects for a branch from origin
 
--- a/docs/man/git-lfs-fetch.1.ronn
+++ b/docs/man/git-lfs-fetch.1.ronn
@@ -26,8 +26,8 @@ This does not update the working copy.
 
 * `--all`:
   Download all objects that are referenced by any commit reachable from the refs
-  provided as arguments. If no refs are provided, then all refs are pushed. This
-  is primarily for backup and migration purposes. Cannot be combined with
+  provided as arguments. If no refs are provided, then all refs are fetched.
+  This is primarily for backup and migration purposes. Cannot be combined with
   --recent or --include/--exclude. Ignores any globally configured include and
   exclude paths to ensure that all objects are downloaded.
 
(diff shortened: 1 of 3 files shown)


pluehne:
As I now realized, I believe that my suggested change to `git lfs fetch` is in line with the current behavior of `git lfs push`. According to the documentation of `git lfs push`:
>`--all`: This pushes all objects to the remote that are referenced by any commit reachable **from the refs provided as arguments.** If no refs are provided, then all refs are pushed.
>—https://github.com/git-lfs/git-lfs/blob/f92bd3b/docs/man/git-lfs-push.1.ronn

bk2204:
Hey,

This seems to be a nice improvement, and I agree that it mirrors `git lfs push`.  I'd definitely like to see some integration tests for this and documentation, as you've pointed out, but I think the basic idea is sound and the code looks reasonably sane.

pluehne:
@bk2204: Thanks a lot, this is exactly the kind of feedback I was looking for before proceeding with integration tests and documentation. I’ll make this pull request ready for review, or I’ll let you know if I have questions.

pluehne:
@bk2204: I’ve updated the documentation to match the new behavior of `git lfs fetch`, and I have added a few tests covering the combination of `--all` with a list of refs.

I wouldn’t mind if someone had a look at my changes now :slightly_smiling_face:. Thanks!

docs/man/git-lfs-fetch.1.ronn
```diff
@@ -25,10 +25,11 @@ This does not update the working copy.
   that would otherwise be downloaded. See [RECENT CHANGES]
 
 * `--all`:
-  Download all objects referenced by any commit that is reachable; this is
-  primarily for backup / migration purposes. Cannot be combined with --recent or
-  --include/--exclude. Ignores any globally configured include and exclude paths
-  to ensure that all objects are downloaded.
+  Download all objects that are referenced by any commit reachable from the refs
+  provided as arguments. If no refs are provided, then all refs are pushed. This
```
bk2204: I think we probably want to say "all refs are downloaded", right?

docs/man/git-lfs-fetch.1.ronn
```diff
@@ -126,6 +127,11 @@ What changes are considered 'recent' is based on a number of gitconfig options:
 
   `git lfs fetch upstream`
 
+* Fetch all the LFS objects from the default remote that are referenced by any
+  commit in the `master` and `develop` branches
+
+  `git lfs fetch --all master develop`
```
bk2204: I think we'd want this to say `git lfs fetch --all origin master develop`.

docs/man/git-lfs-fetch.1.ronn
pluehne: Right. I don’t know why I didn’t spot this one :sweat_smile:. I’ll fix this right away.

docs/man/git-lfs-fetch.1.ronn
pluehne: Also correct, thanks for finding this one as well!

docs/man/git-lfs-fetch.1.ronn
pluehne: I used the wording “all refs are fetched” if that’s okay.

pluehne:
@bk2204: Thanks a lot for your feedback, I’ve fixed the two issues you spotted.

PR #3851, opened 2019-10-08
Fix JSON comma problems in docs

dyrone:
Signed-off-by: Teng Long <dyroneteng@gmail.com>

### DIFF
--- a/docs/api/batch.md
+++ b/docs/api/batch.md
@@ -54,7 +54,7 @@ transfer adapters.
   "objects": [
     {
       "oid": "12345678",
-      "size": 123,
+      "size": 123
     }
   ]
 }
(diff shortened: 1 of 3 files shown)


dyrone:
After PR [#3841 ](https://github.com/git-lfs/git-lfs/pull/3841), I still found some documentation JSON comma issures when i was testing git-lfs HTTP API.

I checked all API documentations and fixed them.

docs/api/batch.md
```diff
@@ -194,7 +194,7 @@ return a 200 status code, and provide per-object errors. Here is an example:
   "objects": [
     {
       "oid": "1111111",
-      "size": 123,
+      "size": 123
```
bk2204: I don't think this is correct.  There's another entry after this, so we need the comma.

PR #3859, opened 2019-10-15
ls-files --all man patch

yamiacat:
I've fixed some minor formatting on the man page of the `git lfs ls-files --all`.

I believe it may be worth adding a sentence along the lines of "This will include previous versions of LFS objects that are no longer found in the current tree." to make its use clearer. Should I add that to this PR?

Context - I've been providing support for LFS for a while but have only just found out about the --all flag introduced in https://github.com/git-lfs/git-lfs/pull/2796 ."How do I find the object IDs of old versions of LFS files?" is a common support request.

I also plan to add that flag to the tutorial - https://github.com/git-lfs/git-lfs/wiki/Tutorial

### DIFF
--- a/docs/man/git-lfs-ls-files.1.ronn
+++ b/docs/man/git-lfs-ls-files.1.ronn
@@ -28,7 +28,7 @@ LFS pointer.
   Show as much information as possible about a LFS file. This is intended
   for manual inspection; the exact format may change at any time.
 
-* `-a --all`:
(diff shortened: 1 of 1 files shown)


bk2204:
Hey,

Thanks for the patch, and welcome to Git LFS!  I think that adding that sentence would be fine, and adding that flag to the wiki somewhere appropriate would also be fine.  The current patch also looks good.

Issue #3863, opened 2019-10-16
Git LFS doesn't properly handle zero-sized file


slonopotamus:
We accidentally got a zero-sized LFS file commited to our repository:

```
version https://git-lfs.github.com/spec/v1
oid sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
size 0
```

Please, don't ask how we achieved it. File was properly uploaded to LFS server (also, with zero size). According to LFS spec, file size is required to be "at least zero" so I believe that our custom [LFS server](https://github.com/bozaro/git-as-svn) is behaving properly by accepting such file.

But now, attempts to download this file into working copy go crazy:

```
$ git status skyway_panel_single_yellow_s.dds 
On branch dev
Your branch is behind 'origin/dev' by 1 commit, and can be fast-forwarded.
  (use "git pull" to update your local branch)

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        deleted:    skyway_panel_single_yellow_s.dds

no changes added to commit (use "git add" and/or "git commit -a")

$ GIT_TRACE=1 git reset --hard e9af3e7ab7055fbd2bc5d09d80356d92034747dc                               
17:20:06.091252 git.c:440               trace: built-in: git reset --hard e9af3e7ab7055fbd2bc5d09d80356d92034747dc
17:20:06.328220 run-command.c:663       trace: run_command: 'git-lfs filter-process'
17:20:06.332127 trace git-lfs: exec: git 'version'
17:20:06.333926 trace git-lfs: exec: git 'config' '-l'
17:20:06.334832 trace git-lfs: Install hook: pre-push, force=false, path=/home/marat/wfgo/.git/hooks/pre-push, upgrading...
17:20:06.334961 trace git-lfs: Install hook: post-checkout, force=false, path=/home/marat/wfgo/.git/hooks/post-checkout, upgrading...
17:20:06.335022 trace git-lfs: Install hook: post-commit, force=false, path=/home/marat/wfgo/.git/hooks/post-commit, upgrading...
17:20:06.335091 trace git-lfs: Install hook: post-merge, force=false, path=/home/marat/wfgo/.git/hooks/post-merge, upgrading...
17:20:06.335161 trace git-lfs: Initialize filter-process
[...]

PR #3864, opened 2019-10-16
Do not delete zero-sized file from LFS storage

slonopotamus:
Because it:
1. Doesn't add any value
2. Introduces a source of race conditions to download logic when multiple git-lfs instances share LFS storage
3. Forces git-lfs to oscillate between two strange states when empty file is downloaded from LFS server. See #3863

### DIFF
--- a/lfs/gitfilter_smudge.go
+++ b/lfs/gitfilter_smudge.go
@@ -62,7 +62,7 @@ func (f *GitFilter) Smudge(writer io.Writer, ptr *Pointer, workingfile string, d
 	stat, statErr := os.Stat(mediafile)
 	if statErr == nil && stat != nil {
 		fileSize := stat.Size()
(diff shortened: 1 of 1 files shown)


slonopotamus:
Rejecting files will force us to rewrite history. This is much worse than marking files as modified. There are numerous scenarios that trigger marking files as modified, most trivial one is changing `.gitattributes` without renormalizing existing files in repository.

mathstuf:
Why would history rewrite be required for anything other than checking out older commits with the `size 0` pointer? Why not replace it with empty content and work with all future LFS implementations?

slonopotamus:
Exactly for checking out old commits. I suggest continuing discussion in #3865. It aims at rejecting much more stuff than just zero-sized LFS pointer.

PR #3868, opened 2019-10-18
Support schannel ssl backend

nikola-sh:
This PR adds schannel [http.sslBackend](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httpsslBackend) support for Git LFS.
Main goal is to ignore [http.sslCaInfo](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httpsslCAInfo) if schannel is active and [http.schannelUseSSLCAInfo](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httpschannelUseSSLCAInfo) is false.
Similar logic in git [here](https://github.com/git-for-windows/git/blob/master/http.c).


bk2204:
Hey,

Since this is a Git for Windows-specific option, and I don't use WIndows, could you explain why this option is valuable?  I'm aware of the difference between the `schannel` (Windows TLS) backend and the `openssl` one.  I know that people use the `schannel` backend so they can use credentials from the normal Windows credential store or Windows smartcard support.

What I'm not familiar with is the reason for ignoring `http.sslCaInfo` in this case.  It seems like it should be simple enough to just not set it if you don't want it used, which is the way other Git options work.

nikola-sh:
Hi!

`http.schannelUseSSLCAInfo` it is probably an attempt to support new cURL feature (`CURLOPT_CAINFO` for `schannel`) but keep old behaviour by default.
You can see [this commit](https://github.com/git-for-windows/git/commit/b67d40adbbaf4f5c4898001bf062a9fd67e43368) and [CURLOPT_CAINFO documentation](https://curl.haxx.se/libcurl/c/CURLOPT_CAINFO.html) for details.

Anyway I think Git LFS should interpret git config settings in same way as git does.

bk2204:
I've seen the commit, and I'm still not clear on why the original code in Git or this code is necessary if it's possible to just not set `http.sslCaInfo`.  Technically, we don't have multiple SSL backends, and changing the configuration of the SSL backend has no effect on us.

I'm not opposed to accepting this patch for compatibility with Git, but before I do so, I need to understand why this option exists in the first place, so I can determine if it's a sensible thing for us to support or not.  It may be that there's a reason it doesn't make sense for us.

nikola-sh:
I agree that user can just unset `sslCaInfo` (it is set in system gitconfig for Windows by default).
But I don't know preconditions of git http config schema.
As I said, some motivation is described in [commit](https://github.com/git-for-windows/git/commit/b67d40adbbaf4f5c4898001bf062a9fd67e43368) message.

Problem I met is that `git lfs` didn't work for corporate git server but `git clone` worked.
Corporate git server uses custom SSL sertificate installed in Windows Certificate Store by admin policies.
I had default system config:
```
[http]
	sslBackend = openssl
	sslCAInfo = C:/Program Files/Git/mingw64/ssl/certs/ca-bundle.crt
```
and edited global config:
```
[http]
	sslBackend = schannel
```

bk2204:
@dscho, can you tell us why this feature originally came into Git for Windows?  It would be helpful for us to understand a little bit more about it before we merge a change here.

dscho:
> can you tell us why this feature originally came into Git for Windows? It would be helpful for us to understand a little bit more about it before we merge a change here.

@bk2204 certainly.

The common way to add custom certificates on Windows, especially in large corporate settings, is by using the Windows Certificate Store. There are convenient administrator tools to do this company-wide, so that e.g. internal servers can have a self-signed certificate and all internal users' machines automatically accept that certificate.

If you use OpenSSL, you have to add that self-signed certificate to the `ca-bundle.crt` file, though. And if there are multiple OpenSSL installations (e.g. one from Git for Windows, and another one from, say, in your local Python installation), _all_ of their respective `ca-bundle.crt` files need to be patched.

Obviously, this is super cumbersome for administrators!

Enter Secure Channel. It is the Windows-native library for TLS/SSL and it integrates nicely with the Windows Certificate Store.

In fact, even if you are not an administrator, but a user who wants to access a given HTTPS site that has a self-signed certificate, it is super convenient to use Secure Channel (AKA schannel): simply direct your web-browser to that site, accept that certificate (which adds it to the Windows Certificate Store), then use schannel-enabled Git to clone from there (which picks up that stored certificate from the Windows Certificate Store).

Convenient, right?

nikola-sh:
@dscho Thank you for explanation.

But we need more details about config entry `http.schannelUseSSLCAInfo`.
Why it was introduced? 

It seems simpler to act in this way.

For config below use schannel with certificates from `/path/to/ca-bundle.crt`:
```
http.sslBackend = schannel
http.sslCaInfo = /path/to/ca-bundle.crt
```

For config below use schannel certificates from Windows Certificate Store.
```
http.sslBackend = schannel
http.sslCaInfo = 
```

dscho:
> But we need more details about config entry `http.schannelUseSSLCAInfo`.
> Why it was introduced?

It was introduced because some cURL version (I forgot which one) introduced not only support for reading the `ca-bundle.crt` file with Secure Channel, but since we used to set `http.sslCAInfo` _all the time_, Git for Windows' Secure Channel support would then pick up that file _instead of_ the Windows Certificate Store.

Let me dig out the commit for you (the commit message should be informative): git/git@b67d40adbbaf4f5c4898001bf062a9fd67e43368

It is a relatively new thing that Git for Windows' installer unsets `http.sslCAInfo` when enabling Secure Channel, we only do this because of Git LFS, so that it does not override the Windows Certificate Store. There were user reports triggering that change of behavior.

nikola-sh:
@bk2204 Take a look to the recent comments pls.

PR #3869, opened 2019-10-21
Update git-lfs-migrate.1.ronn

gertcuykens:
add export mode


docs/man/git-lfs-migrate.1.ronn
```diff
@@ -13,6 +13,9 @@ git-lfs-migrate(1) - Migrate history to or from git-lfs
 * `import`
     Convert large Git objects to LFS pointers.
 
+* `export`
+    Convert LFS pointer files out of Git into corresponding object files.
```
bk2204: Hmm, I'm finding this a little confusing.  I wonder if we could make it mirror the entry for `import` and say something more along the lines of "Convert LFS pointers to large Git objects".

gertcuykens: agree, done

PR #3871, opened 2019-10-22
Fix error strings to follow Go guidelines

slonopotamus:
Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation:
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

### DIFF
--- a/errors/errors_test.go
+++ b/errors/errors_test.go
@@ -6,7 +6,7 @@ import (
 )
 
 func TestChecksHandleGoErrors(t *testing.T) {
-	err := errors.New("Go Error")
+	err := errors.New("go error")
 
 	if IsFatalError(err) {
 		t.Error("go error should not be a fatal error")
@@ -14,7 +14,7 @@ func TestChecksHandleGoErrors(t *testing.T) {
 }
 
 func TestCheckHandlesWrappedErrors(t *testing.T) {
-	err := errors.New("Go error")
+	err := errors.New("go error")
 
 	fatal := NewFatalError(err)
 
@@ -24,7 +24,7 @@ func TestCheckHandlesWrappedErrors(t *testing.T) {
 }
 
 func TestBehaviorWraps(t *testing.T) {
-	err := errors.New("Go error")
+	err := errors.New("go error")
 
 	fatal := NewFatalError(err)
 	ni := NewNotImplementedError(fatal)
@@ -43,7 +43,7 @@ func TestBehaviorWraps(t *testing.T) {
 }
 
 func TestContextOnGoErrors(t *testing.T) {
-	err := errors.New("Go error")
+	err := errors.New("go error")
 
 	SetContext(err, "foo", "bar")
 
@@ -54,7 +54,7 @@ func TestContextOnGoErrors(t *testing.T) {
 }
 
 func TestContextOnWrappedErrors(t *testing.T) {
-	err := NewFatalError(errors.New("Go error"))
+	err := NewFatalError(errors.New("go error"))
 
 	SetContext(err, "foo", "bar")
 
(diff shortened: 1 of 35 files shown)


git/git.go
```diff
@@ -237,10 +237,10 @@ func LsTree(ref string) (*subprocess.BufferedCmd, error) {
 func ResolveRef(ref string) (*Ref, error) {
 	outp, err := gitNoLFSSimple("rev-parse", ref, "--symbolic-full-name", ref)
 	if err != nil {
-		return nil, fmt.Errorf("Git can't resolve ref: %q", ref)
+		return nil, fmt.Errorf("git can't resolve ref: %q", ref)
 	}
 	if outp == "" {
-		return nil, fmt.Errorf("Git can't resolve ref: %q", ref)
+		return nil, fmt.Errorf("git can't resolve ref: %q", ref)
```
bk2204: In these two cases, I think we're referring to the project Git, not the binary `git`, so keeping these upper case is probably better.

errors/errors_test.go
```diff
@@ -54,7 +54,7 @@ func TestContextOnGoErrors(t *testing.T) {
 }
 
 func TestContextOnWrappedErrors(t *testing.T) {
-	err := NewFatalError(errors.New("Go error"))
+	err := NewFatalError(errors.New("go error"))
```
bk2204: I think it was intended in this file for these to refer to the Go language, in which case an upper case first letter would be appropriate here.

git/rev_list_scanner_test.go
```diff
@@ -160,7 +160,7 @@ func TestRevListArgs(t *testing.T) {
 
 func TestRevListScannerCallsClose(t *testing.T) {
 	var called uint32
-	err := errors.New("Hello world")
+	err := errors.New("hello world")
```
bk2204: I don't have a strong opinion on this one, but it is customary to write "Hello world" with an initial capital.

lfsapi/ntlm.go
```diff
@@ -144,7 +144,7 @@ func ntlmGetCredentials(creds creds.Creds) (*ntmlCredentials, error) {
 
 	splits := strings.Split(username, "\\")
 	if len(splits) != 2 {
-		return nil, fmt.Errorf("Your user name must be of the form DOMAIN\\user. It is currently '%s'", username)
+		return nil, fmt.Errorf("your user name must be of the form DOMAIN\\user. It is currently '%s'", username)
```
bk2204: I think this one and the one above are intended to be sentences, and so should begin with a capital.

lfshttp/verbose.go
```diff
@@ -23,7 +23,7 @@ func (c *Client) traceRequest(req *http.Request) (*tracedRequest, error) {
 
 	body, ok := req.Body.(ReadSeekCloser)
 	if body != nil && !ok {
-		return nil, fmt.Errorf("Request body must implement io.ReadCloser and io.Seeker. Got: %T", body)
+		return nil, fmt.Errorf("request body must implement io.ReadCloser and io.Seeker. Got: %T", body)
```
bk2204: Same comment here about the sentences.

errors/errors_test.go
slonopotamus: But at the same time this file says `t.Error("go error should not be a fatal error")` even before my changes.

git/git.go
slonopotamus: OK, reverted.

git/rev_list_scanner_test.go
slonopotamus: Completely changed error text here

lfsapi/ntlm.go
slonopotamus: Ok, reverted

lfshttp/verbose.go
slonopotamus: Ok, reverted

PR #3873, opened 2019-10-23
Stop replacing files in LFS storage when downloading them concurrently

slonopotamus:
On Windows, there is no way to replace file atomically. Instead, MoveFileExA:

1. Calls CreateFileA(access=Delete, shareMode=Delete)
2. Calls SetRenameInformationFile to rename file
3. Calls CloseFile

The problem is that if parallel process attempts to read destination file between
steps 2 and 3, it will try to do that without shareMode=Delete and will hit a SHARING_VIOLATION error.

There are two solutions here:
1. Do not overwrite file if it already exists
2. Retry reading from file if got a SHARING_VIOLATION
3. Do not use Windows

This PR implements option 1.


slonopotamus:
@bk2204: I'm not sure how to fix formatting in `util_generic.go` and `util_linux.go`. I have go 1.12.9 and 1.13.3 but running `go fmt` on both of them doesn't produce any changes in those files.

slonopotamus:
There's is no actual need to avoid overwriting on Linux/MacOS because they don't have problematic race, I just did it for consistency and more adequate API.

slonopotamus:
There's something weird going on with Mac CI, it is definitely unrelated to my changes: https://github.com/git-lfs/git-lfs/pull/3873/checks?check_run_id=272010376#step:4:58

tools/util_generic.go
```diff
@@ -21,3 +21,8 @@ func CloneFile(writer io.Writer, reader io.Reader) (bool, error) {
 func CloneFileByPath(_, _ string) (bool, error) {
 	return false, nil
 }
+
+// This is almost identical to os.rename but doesn't replace newname if it already exists
+func RenameNoReplace(oldname, newname string) error {
+	return false, errors.New("unsupported platform")
```
bk2204: We ship binaries for FreeBSD and just added support for Dragonfly BSD, so I don't think this is going to be a suitable solution.

slonopotamus:
Hmm... Okay.

I see two options:
a) We introduce a "TryRename" function (or whatever you want to name it) that does plain `os.Rename` on all platforms except Windows. On Windows, we still need all stuff I did that ends up in `MoveFileEx` without overwriting.
b) We guard opening files from storage with check-sleep-retry when it hits SHARING_VIOLATION error.

Which one you would prefer?

slonopotamus:
In ideal world, non-replacing rename would be provided by underlying platform. Unfortunately, Go doesn't do that.

bk2204:
I'd definitely prefer #1.  I think we can probably trim `fixLongPath` by simply converting to an absolute path and prepending `\\?\` as appropriate, since Go will have cleaned the path and done most of that for us when converting to an absolute path.  That means our implementation can be a lot simpler.

slonopotamus:
Superceded by #3880.

PR #3878, opened 2019-10-24
Fix several causes of CI problems

bk2204:
For an unknown reason, sometimes libintl.h is missing on macOS when it should not be.  Try harder to fix this by reinstalling the packages which are already allegedly installed and forcing an unlink and re-link.

In addition, address a problem in Windows CI where `%PATH%` is for some reason no longer recognized.  Since it's not strictly necessary to set the path in such a case, stop setting it.

### DIFF
--- a/script/build-git
+++ b/script/build-git
@@ -5,7 +5,12 @@ DIR="$1"
 case $(uname -s) in
   Darwin)
     brew install curl zlib pcre2 gettext openssl
-    brew link --force zlib pcre2 gettext openssl;;
+    brew reinstall curl zlib pcre2 gettext openssl
+    for i in curl zlib pcre2 gettext openssl
+    do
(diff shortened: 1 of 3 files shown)


script/build-git
```diff
@@ -5,7 +5,12 @@ DIR="$1"
 case $(uname -s) in
   Darwin)
     brew install curl zlib pcre2 gettext openssl
-    brew link --force zlib pcre2 gettext openssl;;
+    brew reinstall curl zlib pcre2 gettext openssl
+    for i in zlib pcre2 gettext openssl
```
chrisd8088: Out of interest, why is `curl` not in the `for` list?

bk2204: Because I apparently did not copy and paste correctly.  Thanks for pointing it out.

PR #3880, opened 2019-10-25
Stop replacing files in LFS storage when downloading them concurrently on Windows

### LINKED ISSUE #2825: shared lfs.storage is not threadsafe

slonopotamus:
On Windows, there is no way to replace file atomically. Instead, MoveFileExA:

1. Calls CreateFileA(access=Delete, shareMode=Delete)
2. Calls SetRenameInformationFile to rename file
3. Calls CloseFile

The problem is that if parallel process attempts to open destination file for reading between
steps 2 and 3, it will try to do that without shareMode=Delete and will hit a SHARING_VIOLATION error.

In practice, this race condition results in:

  Smudge error: Error opening media file.: open .git\lfs\objects\<sha>: The process cannot access the file because it is being used by another process.

There are two solutions here:

 1. Do not overwrite file if it already exists
 2. Retry reading from file if got a SHARING_VIOLATION

This commit implements option 1.

Fixes #2825 (for Windows, other OSes are race-free already). Also see #3813 and #3826.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,6 @@ jobs:
     - run: mkdir -p "$HOME/go/bin"
       shell: bash
     - run: set GOPATH=%HOME%\go
-    - run: set PATH=%GOPATH%\bin;%PATH%
     - run: cinst InnoSetup -y
     - run: cinst strawberryperl -y
     - run: choco uninstall git.install -y --force
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,7 +13,6 @@ jobs:
     - run: mkdir -p "$HOME/go/bin"
       shell: bash
     - run: set GOPATH=%HOME%\go
-    - run: set PATH=%GOPATH%\bin;%PATH%
     - run: cinst InnoSetup -y
     - run: cinst strawberryperl -y
     - run: cinst zip -y
--- a/script/build-git
+++ b/script/build-git
@@ -5,7 +5,12 @@ DIR="$1"
 case $(uname -s) in
   Darwin)
     brew install curl zlib pcre2 gettext openssl
-    brew link --force zlib pcre2 gettext openssl;;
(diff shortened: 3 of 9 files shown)


slonopotamus:
This is a v2 attempt for #3873.

slonopotamus:
FTR, here's an example program (Java) that reproduces described SHARING_VIOLATION race:
```java
import java.io.File;
import java.io.FileInputStream;
import java.nio.file.Files;
import java.nio.file.Path;
import java.nio.file.StandardCopyOption;
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.TimeUnit;

public class Main {

  public static void main(String[] args) {
    final int processors = Runtime.getRuntime().availableProcessors();
    final ExecutorService executorService = Executors.newFixedThreadPool(processors);

    final File tmpDir = new File("tmp");

    if (!tmpDir.isDirectory() && !tmpDir.mkdirs())
      System.exit(1);

    final Path dest = new File(tmpDir, "dest").toPath();

    for (int i = 0; i < processors; ++i)
      executorService.submit(() -> {
        while (true) {
          final Path file = File.createTempFile("bla", "", tmpDir).toPath();
          try {
            Files.move(file, dest, StandardCopyOption.ATOMIC_MOVE, StandardCopyOption.REPLACE_EXISTING);
          } catch (Exception e) {
            Files.delete(file);
          }

          TimeUnit.MILLISECONDS.sleep(10);

          try (FileInputStream in = new FileInputStream(dest.toFile())) { // Fails here
            // noop, we just want to open file
          } catch (Exception e) {
            e.printStackTrace();
            System.exit(3);
          }
        }
      });
  }
}
```

slonopotamus:
Rebased on top of #3878 to pass CI.

Issue #3875, opened 2019-10-23
GIT lfs 2.9.0 gives stream error: stream ID 39; HTTP_1_1_REQUIRED on Azure Devops servers


YawJatah:
Hi guys,
We have been having a core issue with git lfs and Azure Devops Servers (and also it's precedent Team Foundation Server 2018). The issue steams from when a binary file is introduced in to the repo before being tracked by git lfs. before 2.9.0 git lfs will just hang on the problematic binary without any feedback. The issue is related to fix #3806, so we updated to 2.9.0 to see if our issue will be resolved. Unfortunately with 2.9.0 we can't even do git lfs on an Azure DevOps Server because right away you run in to stream error: stream ID 39; HTTP_1_1_REQUIRED. This is a possible regression introducted because of fix #3793 which was suppose to add HTTP/2 support.
From my understanding IIS 10 has a deep support for HTTP/2 (and apparent can't be turned of in recent version of Windows Server), and since hosted Azure DevOps Servers run off off IIS they support HTTP/2 by default for ssl connections and will fall back to HTTP/1.1 if the client doesn't support it by usse the error  HTTP_1_1_REQUIRED.

From the logs (GIT_TRACE=1), git lfs starts the connection as HTTP/1.1, is it possible that the implementation of support of HTTP/2 isn't complete (hence there might be a mis match in between the git lfs client and server when it comes to agreeing on what protocol to use).

From the trace log git lfs does a: 
POST /url_to_git_repo/project.git/info/lfs/objects/batch HTTP/1.1.

Any assistance will be greatly appreciated.

We can't really use git 2.8.0 and prior, because of the hang, and can't use 2.9.0 because of the HTTP/2.0 issue.

As a side note: it feels like git lfs needs a lot of testing on a self hosted TFS/Azure Devop servers as they seem to be lots of little blocking issues that you don't  encounter on a GitHub server :S. You guys are doing a great job 👍

YawJatah:
Update:
We were able to disable HTTP/2 on our Windows Server (not ideal because now all the IIS hosted webservices are running with HTTP/1.1) by using the registery change described here: [https://forums.iis.net/t/1235335.aspx?HTTP+2+connections+with+iOS+or+Android+devices+not+working+](https://forums.iis.net/t/1235335.aspx?HTTP+2+connections+with+iOS+or+Android+devices+not+working+). And this allows us properly connect. But we are running into an issue where one binary file keeps giving:
**No credentials are available in the security package**
[...]

PR #3881, opened 2019-10-25
Fall back if there are no system Windows credentials when using NTLM

bk2204:
On Windows, we attempt to use the system credentials when authenticating using NTLM if they exist.  However, in some cases, the system is not appropriately configured and we get an error, which causes the push attempt to abort.

If we get such an error, fall back to prompting for credentials like we did before 2.8.0.

This is a shot in the dark.  It may or may not work, but this is the best attempt I can give things without having a Windows system with AD set up.

### DIFF
--- a/lfsapi/ntlm.go
+++ b/lfsapi/ntlm.go
@@ -36,12 +36,12 @@ func (c *Client) ntlmReAuth(req *http.Request, credWrapper creds.CredentialHelpe
 	// Try SSPI first.
 	if c.ntlmSupportsSSPI() == true {
 		res, err := c.ntlmAuthenticateRequest(req, nil)
-		if err != nil && !errors.IsAuthError(err) {
+		if err != nil && !errors.IsAuthError(err) && !IsSystemCredentialError(err) {
 			return res, err
 		}
 
 		// If SSPI succeeded, then we can move on.
-		if res.StatusCode < 300 && res.StatusCode > 199 {
+		if res != nil && res.StatusCode < 300 && res.StatusCode > 199 {
 			return res, nil
 		}
 	}
--- a/lfsapi/ntlm_auth_nix.go
+++ b/lfsapi/ntlm_auth_nix.go
@@ -10,6 +10,10 @@ import (
 	"github.com/git-lfs/go-ntlm/ntlm"
 )
 
+func IsSystemCredentialError(err error) bool {
+	return false
+}
+
 func (c *Client) ntlmSupportsSSPI() bool {
 	return false
 }
--- a/lfsapi/ntlm_auth_windows.go
+++ b/lfsapi/ntlm_auth_windows.go
@@ -4,10 +4,20 @@ package lfsapi
 
 import (
 	"net/http"
+	"syscall"
 
 	"github.com/alexbrainman/sspi"
 	"github.com/alexbrainman/sspi/ntlm"
 )
+
+const systemCredentialError = 0x8009030E
+
+func IsSystemCredentialError(err error) bool {
+	if e, ok := err.(syscall.Errno); ok {
+		return uintptr(e) == systemCredentialError
+	}
+	return false
+}
 
 func (c *Client) ntlmSupportsSSPI() bool {
 	return true


YawJatah:
Is there anyway for git lfs to access the same windows credentials store as git. Usually git (via GCM) stores the right credentials in the windows credentials store  (usually under: git:https://username@domain.com).
Prompting will be great for dev (on a side note prior to 2.8.0 git lfs had a hard time getting GCM to show the prompt when using NTLM), but might be an issue for un-attendant environment like build machines.

Unless prompt also means getting the credentials from git?

AJH16:
If I'm understanding this issue correctly, it might also be able to proactively prevent this by altering the ntlmSupportsSSPI value in the ntlm_auth_windows file if we can determine that the account isn't valid for SSPI before making the request.  I'm not sure what exactly the case is that this is occurring for though, so I'm not sure if we can tell before we try or not.

bk2204:
"Prompt" in this case means get the credentials using the credential helper.

YawJatah:
Hi Guys, while going through the git config file (the config file inside the .git folder), to get a better understanding of the git lfs settings that was set, I noticed that there were two git entries for our Azure Devops endpoint.

```
[lfs "https://domain-name/tfs/collection/ProjectName/_git/ProjectName.git/info/lfs"]
	access = ntlm
[lfs "https://domain-name/tfs/collection/ProjectName/_git/ProjectName/info/lfs/objects/"]
	access = basic
```
The first entry was the main one before 2.9.0 (we were using 2.8.0). After I started using 2.9.0 the second entry was added. That seems to be the one that git lfs in 2.9.0 is using, because the access of the first one doesn't seem to affect the authentication type, but the second one does.

As described in this issues  #3875 , the work around was to turn off HTTP/2 and enable Basic Authentication (but this work around only works with 2.9.0, doesn't seem to work with 2.8.0).

Don't know if the above information will help shed some light on the underlining issue.

bk2204:
> ```
> [lfs "https://domain-name/tfs/collection/ProjectName/_git/ProjectName.git/info/lfs"]
> 	access = ntlm
> [lfs "https://domain-name/tfs/collection/ProjectName/_git/ProjectName/info/lfs/objects/"]
> 	access = basic
> ```

These entries are used to record the authentication type we used last so we can pick the right one the next time.  Since you switched from NTLM to Basic, we added a new entry.  I'm not sure why the remote URL is slightly different, but it should be fine either way.

bk2204:
Okay, I've rebased this on the latest master.

@YawJatah, could you try this version with `http.version` set to `HTTP/1.1` and using NTLM?  If I'm correct, this version should fix both problems.  If you need binaries, you can go to the [CI check run page](https://github.com/git-lfs/git-lfs/pull/3881/checks?check_run_id=282267937) (when it completes) and download the `windows-latest` artifact, which should have binaries for you.

YawJatah:
> Okay, I've rebased this on the latest master.
> 
> @YawJatah, could you try this version with `http.version` set to `HTTP/1.1` and using NTLM? If I'm correct, this version should fix both problems. If you need binaries, you can go to the [CI check run page](https://github.com/git-lfs/git-lfs/pull/3881/checks?check_run_id=282267937) (when it completes) and download the `windows-latest` artifact, which should have binaries for you.

@bk2204 , will try it out tonight. Just to confirm, http.version, is already set, or we have to manually set that in the git config (the config file under the .git folder)? Also for the server should we turn HTTP/2 back on and disable Basic authentication (So only leaving Negotiate enabled)?

bk2204:
> @bk2204 , will try it out tonight. Just to confirm, http.version, is already set, or we have to manually set that in the git config (the config file under the .git folder)? Also for the server should we turn HTTP/2 back on and disable Basic authentication (So only leaving Negotiate enabled)?

You'll need to set `http.version` manually, either in `.git/config` or in the per-user configuration (`~/.gitconfig`).  That should make it work when you turn HTTP/2 on.  And yes, please disable Basic, leaving only NTLM and Negotiate enabled.  If things are fixed, it should continue to work with those settings enabled.

YawJatah:
@bk2204 I was able to finally test out the latest build. There is still an issue with ntlm. I can only get git lfs to work with basic  authentication (I now have HTTP/2 turned on on the server side and atleast that works).
The only way I am able to update my repo is by doing 

`GIT_LFS_SKIP_SMUDGE=1 GIT_TRACE=1 GIT_TRANSFER_TRACE=1 GIT_CURL_VERBOSE=1 GCM_TRACE=1 git pull --verbose`

then

updating the config file entry for git fls

[lfs "https://repo-url.com/tfs/collection/project/_git/project.git/info/lfs"]
	access = basic

[lfs "https://repo-url.com/tfs/collection/project/_git/project.git/info/lfs/objects/"]
	access = basic

then

running

`GIT_TRACE=1 GIT_TRANSFER_TRACE=1 GIT_CURL_VERBOSE=1 GCM_TRACE=1 git lfs fetch origin --all`

`GIT_TRACE=1 GIT_TRANSFER_TRACE=1 GIT_CURL_VERBOSE=1 GCM_TRACE=1 git lfs checkout`


it seems that for Azure Devops git lfs doesn't work properly  on the first url. The second url is added automatically after I set the first url to basic.

bk2204:
Okay, then I'm not sure how to fix NTLM.  I'm happy to accept a patch to do so, but without test infrastructure, I'm unable to reproduce this problem.

I'm going to close this PR, since it doesn't work.

YawJatah:
@bk2204 , if we setup a test Azure DevOps server for you, will that be helpfull in testing  and validating this issue (there is no config to be don on the client side, plus you will be connecting directly to the actually enviroment that seems problematic)?

bk2204:
Yes, if I can have a test account with NTLM, that would be helpful in tracking this down.  I have a Windows 10 Pro VM I can use for testing.

YawJatah:
@bk2204 Sounds good, will setup a test server and pm you when it is ready. Thanks

PR #3887, opened 2019-10-28
Allow specifying HTTP version to use

bk2204:
There are some servers that cannot speak HTTP/2 in all cases and demand to fall back to HTTP/1.1 with a `HTTP_1_1_REQUIRED`.  Notably, this happens with IIS 10 when using NTLM.  Go's HTTP library doesn't seem to like this response and aborts the transfer, leading to a failure.

Fortunately, Git has an option (`http.version`) to control the protocol used when speaking HTTP to a remote server.  Implement this option to allow users to set the protocol to use when speaking HTTP and work around these servers.

/cc @YawJatah as reporter

### DIFF
--- a/lfshttp/client.go
+++ b/lfshttp/client.go
@@ -346,6 +346,26 @@ func (c *Client) doWithRedirects(cli *http.Client, req *http.Request, remote str
 	return c.doWithRedirects(cli, redirectedReq, remote, via)
 }
 
+func (c *Client) configureProtocols(u *url.URL, tr *http.Transport) error {
+	version, _ := c.uc.Get("http", u.String(), "version")
+	switch version {
+	case "HTTP/1.1":
+		// This disables HTTP/2, according to the documentation.
+		tr.TLSNextProto = make(map[string]func(authority string, c *tls.Conn) http.RoundTripper)
+	case "HTTP/2":
+		if u.Scheme != "https" {
+			return fmt.Errorf("HTTP/2 cannot be used except with TLS")
+		}
+		http2.ConfigureTransport(tr)
+		delete(tr.TLSNextProto, "http/1.1")
+	case "":
+		http2.ConfigureTransport(tr)
+	default:
+		return fmt.Errorf("Unknown HTTP version %q", version)
+	}
+	return nil
+}
+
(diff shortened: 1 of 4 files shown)


lfshttp/client.go
```diff
@@ -346,6 +346,24 @@ func (c *Client) doWithRedirects(cli *http.Client, req *http.Request, remote str
 	return c.doWithRedirects(cli, redirectedReq, remote, via)
 }
 
+func (c *Client) configureProtocols(u *url.URL, tr *http.Transport) error {
+	version, _ := c.uc.Get("http", u.String(), "version")
+	switch version {
+	case "HTTP/1.1":
+		// This disables HTTP?2, according to the documentation.
```
chrisd8088: Possible typo: `HTTP/2` not `HTTP?2`?

lfshttp/client.go
```diff
@@ -346,6 +346,24 @@ func (c *Client) doWithRedirects(cli *http.Client, req *http.Request, remote str
...
+		tr.TLSNextProto = make(map[string]func(authority string, c *tls.Conn) http.RoundTripper)
+	case "HTTP/2":
+		if u.Scheme != "https" {
+			return fmt.Errorf("HTTP/2 cannot be used except with TLS")
+		}
+		http2.ConfigureTransport(tr)
+		delete(tr.TLSNextProto, "http/1.1")
+	case "":
```
chrisd8088: I'm not too familiar with the `URLConfig.Get()` method (well, or any of the LFS client code! :-) but is it possible it might return some other non-empty value here, perhaps due to a misconfiguration?  I just wondered about a `default:` here instead ...

lfshttp/client.go
bk2204: Good catch.

lfshttp/client.go
bk2204: That method returns either `value, true`, where `value` is the value that's from the config file (which can either be `http.version` or `http.MATCHING-URL.version`, as appropriate), or `"", false` if no such value exists.  In other words, this is the same return value as a map would produce.

I agree that a default is appropriate here, since we should probably return an error in that case; thanks for the suggestion.  Will fix.

PR #3890, opened 2019-10-29
Implement retry logic to fix LFS storage race conditions on Windows

slonopotamus:
Testing showed that while race condition analysis in #3880 was correct, the way it tries to fix that
does not work for the *first* git-lfs process that will actually perform file move.

So, I revert #3880. Instead, this PR performs multiple attempts when working with files in LFS storage.

Similar logic is already implemented in "cmd/go/internal/robustio" and "cmd/go/internal/renameio" packages. However, they are not public, so we cannot use them.

Marking this PR as draft because we need 2-3 days of internal testing before will be sure that it actually fixes the problem.

P.S. If you have better names for RobustXXX functions - I'm fully open for ideas.


slonopotamus:
I believe that `t-fetch.sh` failure on Ubuntu is not related to this PR. See #3891.

bk2204:
I've kicked off another run.  We'll see how that works; trying to do that in the past has made everything fail, so hopefully things are working better now.

bk2204:
This seems like a sane approach, BTW.  The `Robust` prefix is fine with me.

slonopotamus:
Well, this approach was described as an alternative in #3880 but I hoped to the last moment that more reliable stuff will work. However, cruel reality showed that it won't.

tools/robustio_windows.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+const (
+	// This is private in [go]/src/internal/syscall/windows/syscall_windows.go :(
+	ERROR_SHARING_VIOLATION syscall.Errno = 32
+)
+
+// isEphemeralError returns true if err may be resolved by waiting.
+func isEphemeralError(err error) bool {
+	return err == ERROR_SHARING_VIOLATION
```
slonopotamus: Bug: I need to use `errors.Is` instead of `==` because `os.foo()` wraps underlying error.

slonopotamus:
Force-pushed updated changed, days-without-incidents timer reset to zero, need several days of testing again.

slonopotamus:
Ouch. Turns out `errors.Is` is a Go-1.13 feature. Looking how to achieve the same on older versions...

slonopotamus:
Okay, I believe we've tested this hard enough

slonopotamus:
FYI: we're running for almost a month with this fix and haven't seen a single failure caused by `SHARING_VIOLATION`.

bk2204:
That's great to hear.

PR #3903, opened 2019-11-06
Revert "ci: force Windows Git version to 2.22.0"

bk2204:
Now that Git for Windows 2.24.0 is available with a fixed version of bash, we can revert the changes to CI that pinned it to 2.22.0 and have things work properly.

This reverts commit 52257344da89a7ba0891d49ac4d60a8e04563aad.


bk2204:
Apparently Actions doesn't have 2.24.0 out yet, so this will fail.  I'll let this hang out until 2.24.0 is available and then rebase onto `master` and things should pass.

PR #3906, opened 2019-11-07
Add issue templates

bk2204:
To help folks provide the information we need to be successful in
helping them, let's provide issue templates that prompt them for
information about their systems.


bk2204:
Suggestions on any way to improve this are welcome; most of what's in here is in the `CONTRIBUTING.md`, but folks don't always read that before submitting an issue.

Issue #3898, opened 2019-11-01
hooks and lfs folders are created on project root instead of inside .git folder when using git-lfs


StarWars999123:
The issue occured after Installation of the latest Git LFS 2.9.0 on a PC with Visual Studio 2019 16.3.6. git-lfs --version from cmd says, it runs with GitLFS version 2.8.0. 

Finally the project root looks like this: 
```
.git/
    hooks/
        pre-push
        post-checkout
        post-merge
        post-commit
hooks/
    pre-push
    post-checkout
    post-merge
    post-commit
ourFilesAndStuff
```
The issue occured already several times on different developer PCs. We are not sure, how we should handle this. A simple delete does not help, as it gets allways recreated immedeatly.

- Does this happen due to a version conflict of the git LFS delivered by VisualStudio or why can it occur?
- Shall we check-in this Hooks-Folder or put it into the .gitignore?
- Whats the reason that it gets created?

bk2204:
Hey,

Can you provide the output of `git config -l` on an affected system?  I think there might be a `core.hooksPath` setting that's causing this.

StarWars999123:
Luckily I have currently a working repository without this anomaly. On this system does git config -l does not list any core.hooksPath. I'll try to install the mentioned lfs version 2.9.0 again in a few days.

bk2204:
It would also be useful to have the output of `git lfs env`.  There were some changes in 2.8.0 that had to do with hooks, and knowing what Git LFS thinks some of those paths are would help troubleshooting.

StarWars999123:
```
git-lfs/2.8.0 (GitHub; windows amd64; go 1.12.2; git 30af66bb)
git version 2.23.0.windows.1

LocalWorkingDir=
LocalGitDir=
LocalGitStorageDir=
LocalMediaDir=lfs\objects
LocalReferenceDirs=
TempDir=lfs\tmp
ConcurrentTransfers=3
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
FetchRecentRefsIncludeRemotes=true
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=lfs
AccessDownload=none
AccessUpload=none
DownloadTransfers=basic
UploadTransfers=basic
[...]

PR #3910, opened 2019-11-08
Handle missing cygpath gracefully

bk2204:
It looks like there are situations in which some Git for Windows installations don't have the cygpath binary.  In that case, we end up picking an empty string, which leads us to install hooks and LFS objects in the current directory.  Fix this by looking for the cygpath binary and if it's missing, disabling Cygwin path canonicalization on Windows.

/cc #3898 
/cc @StarWars999123

### DIFF
--- a/tools/os_tools.go
+++ b/tools/os_tools.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"os/exec"
 	"strings"
 
 	"github.com/git-lfs/git-lfs/subprocess"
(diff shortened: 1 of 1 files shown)


bk2204:
@StarWars999123, can you try one of the artifacts from the successful Windows CI job and see if it solves your problem?

mastercoms:
This fixed the issue for me.

mastercoms:
This fixed the issue for me.

bk2204:
Okay, great, thanks for confirming.  This won't make it into 2.9.1, since I've already done the release PR for that, but it's possible to do a 2.9.2 in the future.

PR #3911, opened 2019-11-08
Allow migrate export to handle non-pointer files gracefully

### LINKED ISSUE #3908: lfs migrate export: error while parsing a file to be rewritten
So I'm rewriting a repo to migrate out of LFS, using `git lfs migrate export --include="*" --everything`.
migrate chokes on a file during the `Rewriting commits` step (I think), with this error:
```
10:06:15.841053 trace git-lfs: filepathfilter: accepting "problematic_file"
Error reading line 0: #!/bin/sh
```

Here's a very simple repro (btw I'm on Git for Windows 2.24.0):
```
#!/bin/bash

# init repo
rm -rf test_repo/
mkdir test_repo
cd test_repo
git init

echo "git-lfs" > problematic_file
git add .
git commit -m "create repo"

# git lfs migrate export --include="*" --everything --yes
GIT_TRACE=1 git lfs migrate export --include="*" --everything --yes 2>&1 | tee ../lfs_migrate_export.log
```

This is the smallest repro I could make. The error only happens if problematic_file contains "git-lfs".

I pinpointed the error to [this line](https://github.com/git-lfs/git-lfs/blob/master/lfs/pointer.go#L256)... unfortunately I have zero experience with go, so trying to fix it seems a bit daunting.

I'm guessing it's trying to parse the file to figure out if it's an lfs pointer file like this:
```
oid sha256:d3a4dad979bc00ce2c0e08553f4a8822c3649efbfd88c9ebf0d4ed3509183928
size 11
```
... probably seeing "git-lfs" in the file triggers it, but then not all lines are key-value pairs, so it fails?

In my real-world example, the actual file is a shell script meant to toggle git hooks on or off, which I think is more or less a copy of a vanilla git lfs hook:
```
#!/bin/sh
command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting .git/hooks/post-checkout.\n"; exit 2; }
git lfs post-checkout "$@"
```

bk2204:
If a file that was being handled by git lfs migrate export contained the text "git-lfs", but was not a pointer file, we would abort.  This occurred because the error was not of a type that indicated that the file was not a pointer, so we treated it as an unrecoverable error instead of a recoverable one.

Wrap the error in an appropriate type so that we can recover from this and treat it as not being a pointer.  Add tests for both error branches in the function to ensure we recover from them.

Fixes #3908.
/cc @benblo as reporter

### DIFF
--- a/lfs/pointer.go
+++ b/lfs/pointer.go
@@ -253,7 +253,7 @@ func decodeKVData(data []byte) (kvps map[string]string, exts map[string]string, 
 
 		parts := strings.SplitN(text, " ", 2)
 		if len(parts) < 2 {
-			err = fmt.Errorf("error reading line %d: %s", line, text)
+			err = errors.NewNotAPointerError(fmt.Errorf("error reading line %d: %s", line, text))
 			return
 		}
 
@@ -261,7 +261,7 @@ func decodeKVData(data []byte) (kvps map[string]string, exts map[string]string, 
 		value := parts[1]
 
 		if numKeys <= line {
-			err = fmt.Errorf("extra line: %s", text)
+			err = errors.NewNotAPointerError(fmt.Errorf("extra line: %s", text))
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


benblo:
Thanks a lot!

PR #3918, opened 2019-11-13
Add support for local paths

### LINKED ISSUE #3914: git local clone failed
```sh
mkdir test
git init
git lfs track *.pdf
git add ngspice-manual.pdf
git commit -m "test"

cd ..
git clone test test2
```

The failure occurred in the last command:
```
(488)$ git clone test test2
Cloning into 'test2'...
done.
Downloading ngspice-manual.pdf (3.2 MB)
Error downloading object: ngspice-manual.pdf (eb4de49): Smudge error: Error downloading ngspice-manual.pdf (eb4de49c0ab6870cafe4cb412179420d89d97872aafbd2fd0fb16554d5451925): batch request: ssh: Could not resolve hostname c: Name or service not known: exit status 255

Errors logged to C:\projects\git\test2\.git\lfs\logs\20191112T144004.7574523.log
fatal: ngspice-manual.pdf: smudge filter lfs failed

```

bk2204:
Currently, we only support local remotes using file URLs, but not local paths.  This is a highly requested feature, however, so implement support for local paths as remotes.

First, fix the handling of local paths we already have: the handling of absolute Unix-style paths.  Instead of checking for the remote to contain a file URL to determine whether to append "info/lfs" to the URL, look to see if the URL we're using is a file URL, which will catch local paths which will have been rewritten as such by this point.

If we get something that looks like it's not a URL, check if it's a file on the file system, and if so, don't attempt to interpret it as an SSH URL.  This fixes the confusion with Windows paths, which resemble SSH-style locations with a single-letter alias as the host name.

Finally, when turning a local path into a file URL, turn the path into an absolute one if possible and rewrite it using slashes, which is required for file URLs.  Add several tests for the various cases: one for Unix-style paths, one for native paths, and one for relative paths.

Fixes #3914.

### DIFF
--- a/lfsapi/endpoint_finder.go
+++ b/lfsapi/endpoint_finder.go
@@ -175,7 +175,7 @@ func (e *endpointGitFinder) NewEndpointFromCloneURL(operation, rawurl string) lf
 		ep.Url = rawurl[0 : len(rawurl)-1]
 	}
 
-	if strings.HasPrefix(rawurl, "file://") {
+	if strings.HasPrefix(ep.Url, "file://") {
 		return ep
 	}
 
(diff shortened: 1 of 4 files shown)


bk2204:
Yeah, I'd hope that we could have a more generic approach in the future, but for now, it's kind of a bear.  The code is necessarily complicated by the fact that some Windows paths looks like SSH remotes; that's been the inconvenient issue preventing us from making progress on this before.

Issue #3929, opened 2019-11-22
Retries failed with "tls: use of closed connection" error


bluekeyes:
We're occasionally seeing failures downloading LFS files from GitHub Enterprise (2.18.5 and 2.19.2) where a failed request is retried, but then the retry fails due to a `tls: use of closed connection` error. Here's a `GIT_TRACE=1` log from one occurrence:

```
04:34:43.194170 trace git-lfs: exec: git '-c' 'filter.lfs.smudge=cat' '-c' 'filter.lfs.clean=cat' '-c' 'filter.lfs.process=' '-c' 'filter.lfs.required=false' 'rev-parse' 'HEAD' '--symbolic-full-name' 'HEAD'
Downloading path/to/file.png (136 KB)
04:34:43.200267 trace git-lfs: tq: running as batched queue, batch size of 100
04:34:43.200331 trace git-lfs: tq: sending batch of size 1
04:34:43.200446 trace git-lfs: run_command: ssh -- git@github.company.domain git-lfs-authenticate org/repository.git download
04:35:01.175099 trace git-lfs: api: batch 1 files
04:35:01.175492 trace git-lfs: HTTP: POST https://github.company.domain/_lfs/org/repository/objects/batch
04:35:02.002612 trace git-lfs: HTTP: 200
04:35:02.002782 trace git-lfs: HTTP: {"objects":[{"oid":"fdb49757db5ee89cd90af00c1671c23912e11c60237d743a6c1fa1487006edfa","size":136253,"actions":{"download":{"href":"https://media.github.company.domain/lfs/58863/objects/fdb49757db5ee89cd90af00c1671c23912e11c60237d743a6c1fa1487006edfa","header":{"Authorization":"RemoteAuth REDACTED"}}}}]}
04:35:02.002881 trace git-lfs: tq: starting transfer adapter "basic"
04:35:02.003394 trace git-lfs: HTTP: GET https://media.github.company.domain/lfs/58863/objects/fdb49757db5ee89cd90af00c1671c23912e11c60237d743a6c1fa1487006edfa
04:35:17.093724 trace git-lfs: tq: retrying object fdb49757db5ee89cd90af00c1671c23912e11c60237d743a6c1fa1487006edfa: LFS: Get https://media.github.company.domain/lfs/58863/objects/fdb49757db5ee89cd90af00c1671c23912e11c60237d743a6c1fa1487006edfa: EOF
[...]

PR #3930, opened 2019-11-22
Retry batch failures

bk2204:
If the batch operation fails due to an error instead of a bad HTTP status code, we'll abort the batch operation and retry.  This appears to be a regression from 1412d6e4 ("Don't fail if we lack objects the server has", 2019-04-30), which caused us to handle errors differently.

Since there are two error returns from enqueueAndCollectRetriesFor, let's wrap the batch error case as a retriable error and not abort if we find a retriable error later on.  This lets us continue to abort if we get a missing object, which should be fatal, but retry in the more common network failure case.

/cc #3929
/cc @bluekeyes as reporter

### DIFF
--- a/tq/transfer_queue.go
+++ b/tq/transfer_queue.go
@@ -442,7 +442,7 @@ func (q *TransferQueue) collectBatches() {
 		// don't process further batches.  Abort the wait queue so that
 		// we don't deadlock waiting for objects to complete when they
 		// never will.
-		if err != nil {
+		if err != nil && !errors.IsRetriableError(err) {
 			q.wait.Abort()
 			break
 		}
@@ -538,7 +538,7 @@ func (q *TransferQueue) enqueueAndCollectRetriesFor(batch batch) (batch, error) 
 				}
 			}
 
-			return next, err
+			return next, errors.NewRetriableError(err)
 		}
 	}
 


chrisd8088 (approved):
This seems reasonable to me, after a brief detour through (some of) the relevant history here -- thank you for the detailed notes!

As far as I can tell, the [comments](https://github.com/git-lfs/git-lfs/blob/4098474f682ac683fc94d0a9f5141f79ffc4b00b/tq/transfer_queue.go#L494-L496) above `enqueueAndCollectRetriesFor()` are still accurate as to the intended functionality, and the evolution of the serious error check in `collectBatches()` through #3634 and #3800 makes sense.

My only question would be one of those tedious irritating ones as to whether we should implement any additional tests -- and I realize that might be a tall order, as on a first quick pass I don't see anything in the current suite which be easily adapted (but I would be happy to be wrong about that).  On the assumption it's not straightforward, though, I think it's fine to merge this as it stands.

Thanks as always!!  👍

bk2204:
The question about the tests is a good one.  Unfortunately, the error case we're handling here is when we get unexpected errors; that is, where we don't get an HTTP response.  Go's HTTP library doesn't really let me close the connection without a response (and this is not the first time I've wanted to do this in a test).

Let me see if I can come up with a test that can emulate something with an unexpected HTTP status code, which if possible would be better than nothing.

bk2204:
I've just checked, and I don't seem able to synthesize this with a 500.  That error is flagged as non-retriable, which means the process aborts.

chrisd8088:
Fair enough -- thank you for checking!  This did seem to fall into the category of tests which would require more of an integration test framework, and possibly some low-level plumbing.  As I said, this looks fine to merge as-is:
> On the assumption it's not straightforward, though, I think it's fine to merge this as it stands.

PR #3936, opened 2019-12-02
Add nil-check on defer block of DoTransfer()

exceed-alae:
If os.OpenFile() returns error then variable "f" is nil, so support this case

### DIFF
--- a/tq/basic_download.go
+++ b/tq/basic_download.go
@@ -47,15 +47,17 @@ func (a *basicDownloadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb Progr
 	if err != nil {
 		return err
 	}
+	tmpName := f.Name()
 	defer func() {
+		// Fail-safe: Most implementation of os.File.Close() does nil check
 		if f != nil {
 			f.Close()
-			// This will delete temp file if:
-			// - we failed to fully download file and move it to final location including the case when final location already
-			//   exists because other parallel git-lfs processes downloaded file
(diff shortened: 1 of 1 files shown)


slonopotamus:
Ugh... Seems like things are not that simple.

The relevant code is:

    f, err = os.OpenFile(f.Name(), os.O_RDWR, 0644)

So, there is a filename that should be deleted (`f.Name()` *before* call to `OpenFile`), but it gets lost if `OpenFile` fails.

exceed-alae:
> So, there is a filename that should be deleted (f.Name() before call to OpenFile), but it gets lost if OpenFile fails and overwrites f with nil.

I see ... that's right... I'm going to push new patch.

slonopotamus:
@exceed-ale: just curious: did you find this problem by reading through the code or actually hit it in real-world scenario? Even after your fix, hitting this codepath means that download has failed in a kinda unexpected way. The bug you're fixing was introduced during #3826 that has severely changed this code to deal with concurrent git-lfs processes downloading to the same LFS storage. So I'm interested in conditions that trigger this code to fail because that might mean that measures taken to fix concurrent problems were not enough.

exceed-alae:
@slonopotamus That problem occured when I was testing lfs.storage sharing by multi user. (2.9.0 for WIndows. I've set lfs.storage to //MACHINE/SHARE/)
The exact situation at the time of that problem occured have not been confirmed. But I think so that same object was accessed on same timing by different user.

PR #3941, opened 2019-12-04
Kerberos (SPNEGO) support for HTTP

### LINKED ISSUE #2576: git-lfs does not support negotiate authentication
Any chance lfs could use negotiate authentication in addition to basic auth? We use kerberos/sso in front of our GIT server and we currently can't utilize lfs for this reason. When the lfs authentication happens, it just fails as it won't retry authentication with a different method. This would be similar to the fix in the GIT bug:
https://github.com/git/git/commit/4dbe66464b4fd695c5989cc272fa0edd6475037c

bk2204:
Currently, Git LFS doesn't support Kerberos; it only supports NTLM using the Negotiate authentication method.  Kerberos is commonly used on Windows as part of Active Directory, as well as on Unix systems.  It is well understood, and provided the standard cryptographic assumptions are met, is far more secure than NTLM.

This series introduces support for Kerberos when the remote server advertises support for the Negotiate method. Negotiate supports both Kerberos and NTLM, but the choice can only be distinguished by the messages sent, which necessarily requires an authentication attempt.  Consequently, we attempt to use Kerberos, and if it fails for any reason, we fall back to NTLM.  This ensures we don't break existing users.

This series works correctly on a Debian sid system against a Kerberos domain running a MIT Kerberos KDC on Debian buster.  Although the series should work fine on Windows, it has not been tested there, and consequently Windows support is experimental.  The number of algorithms supported by the Go Kerberos library is limited, but should support the most common and secure types (AES with HMAC-SHA-1 and AES with HMAC-SHA-2), which are also supported on Windows.

Because Kerberos requires a trusted third party (the KDC, key distribution center, or Kerberos server), there are few tests for this case, as it wouldn't be reasonable to set up an entire Kerberos infrastructure in the testsuite.

Fixes #2576


itsho:
This will be extremely useful. when is the next release planned for?
thanks in advance! 😃

bk2204:
The next release is planned for this month.  Please feel free to try it out using one of the CI artifact binaries (or one you've built yourself), especially if you're using Windows, so that if there are any problems, we can fix them before the release.

itsho:
I've compiled the latest version and tested it with our local build machine.
1. I'm still getting HTTP_1_1_REQUIRED
2. the Windows user got locked. 😕

log summary:

> git-lfs/2.9.3 (GitHub; windows amd64; go 1.13.5; git 770719d4)
> Receiving objects: 100% .............. done.
> Resolving deltas: 100% (..............), done.
> git -c http.https://-------------.extraheader="AUTHORIZATION: bearer ********" lfs fetch origin ..............................bd
> fetch: Fetching reference ........................bd
> LFS: Get https://................/info/lfs/objects/...........................6f: stream error: stream ID 3205; HTTP_1_1_REQUIRED
> error: failed to fetch some objects from 'https://................/_git/ufed.git/info/lfs'
> ##[command]git lfs logs last
> No logs to show

bk2204:
It's known that you have to set `http.version` to `HTTP/1.1` if you use Kerberos or NTLM with IIS; that isn't changing.

It's interesting that it got locked, because we do fall back to NTLM if Kerberos fails for some reason.  Unfortunately, we don't have a good way of knowing what kind of Negotiate authentication the remote side supports.  Can you open a new issue with the output of running with `GIT_TRACE=1 GIT_TRANSFER_TRACE=1 GIT_CURL_VERBOSE=1` with at least the first 12 characters of the Base64-encoded Negotiate responses so I can look into this?

go.mod
```diff
@@ -21,7 +25,10 @@ require (
 	github.com/xeipuuv/gojsonschema v0.0.0-20170210233622-6b67b3fab74d
 	golang.org/x/net v0.0.0-20191027093000-83d349e8ac1a
 	golang.org/x/sync v0.0.0-20181221193216-37e7f081c4d4
-	golang.org/x/sys v0.0.0-20190215142949-d0b11bdaac8a
+	golang.org/x/sys v0.0.0-20190412213103-97732733099d
+	gopkg.in/jcmturner/goidentity.v3 v3.0.0 // indirect
+	gopkg.in/jcmturner/gokrb5.v7 v7.3.0 // indirect
```
QuLogic: Is it right that you have v7 when [go-spnego is using v5.3.0](https://github.com/dpotapov/go-spnego/blob/master/go.mod#L11)?

bk2204: It appears that's a mistake.  I think #4035 should fix that.

It perhaps snuck in before I used go-spnego and was doing things by hand.  Thanks for pointing it out.

PR #3973, opened 2020-01-08
Improve license files

bk2204:
Add a note to the LICENSE file that components included in the binary contain modules licensed under other terms, and update the Debian copyright file to reflect the components used in the binary.


Chest1:
Thanks

PR #3992, opened 2020-01-20
Allow locking and unlocking non-existent files

mversluys:
The lockPath sanitization was slightly amended to allow os.Stat() to fail. All other conditions, such as testing to see if it's a directory were retained.

If a lock is obtained, attempting to set it writable only occurs if the file exists.


mversluys:
Will try running the tests locally

mversluys:
Added and updated tests

PR #4016, opened 2020-02-10
Check error when creating local storage directory

### LINKED ISSUE #4015: cannot replace "" with "<$TMPDIR/oid-nonce>": rename <$TMPDIR/oid-nonce> : no such file or directory
**Describe the issue**
Failure to create object path in the lfs database is silently discarded, resulting in a misleading error message at the end of the fetch operation

**System environment**
macOS
Darwin Kernel Version 19.2.0: root:xnu-6153.61.1~20/RELEASE_X86_64 x86_64

**Output of `git lfs env`**
```
git-lfs/2.10.0 (GitHub; darwin amd64; go 1.13.6)
git version 2.21.1 (Apple Git-122.3)

Endpoint=https://server.com/git/lfs (auth=basic)
LocalWorkingDir=/Users/user/project
LocalGitDir=/Users/user/project/.git
LocalGitStorageDir=/Users/user/project/.git
LocalMediaDir=/Users/user/project/.git/lfs/objects
LocalReferenceDirs=
TempDir=/Users/user/project/.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/Users/user/project/.git/lfs
AccessDownload=basic
AccessUpload=basic
GIT_EXEC_PATH=/Library/Developer/CommandLineTools/usr/libexec/git-core
```

**Additional context**
The symptoms are such:
git lfs fetch would properly connect to the servers and start file transfer. It would download all files but the progress indicator would stay at 0. At the end of all downloads the process fails with a list of errors per every object of the same form:

    cannot rename "" with "<path>": rename <path> : no such file or directory

where `<path>` is the local $TMPDIR concatenated with the OID and some nonce. 
After the process $TMPDIR would actually contain the files with names OID.part, meaning that they were indeed downloaded successfully 

After some rather long investigation which involved checking out master of git-lfs and debug printing it was eventually discovered that the actual issue was that the ./git/lfs folder had been owned by root. The reason why it wasn't obvious lies in [commands.go:135](https://github.com/git-lfs/git-lfs/blob/b37a75010414946fb565bae80b2371f9f6a52b53/commands/commands.go#L135):

```
func downloadTransfer(p *lfs.WrappedPointer) (name, path, oid string, size int64, missing bool) {
	path, _ = cfg.Filesystem().ObjectPath(p.Oid)
	return p.Name, path, p.Oid, p.Size, false
}
```

here the `cfg.Filesystem().ObjectPath(p.Oid)` returned error being unable to create the db dir for the object, but that error is discarded.
```
error trying to create local storage directory in "/Users/user/project/.git/lfs/objects/5c/63": mkdir /Users/user/project/.git/lfs/objects: permission denied
```

bk2204:
When we have an error creating the local storage directory, such as if the permissions are wrong, we fail to produce a helpful error message, since we fail to check the error we produce.

Let's take the error we get in such a case and pass it through to the transfer queue, where we can handle it as any other error.  This means that if only one directory has a problem, we can transfer the rest of the objects and fail for only the one problematic object.

Fixes #4015.
/cc @leonid-s-usov as reporter

### DIFF
--- a/commands/command_migrate_export.go
+++ b/commands/command_migrate_export.go
@@ -136,7 +136,7 @@ func migrateExportCommand(cmd *cobra.Command, args []string) {
 			}
 
 			if _, err := os.Stat(downloadPath); os.IsNotExist(err) {
-				q.Add(p.Name, downloadPath, p.Oid, p.Size, false)
+				q.Add(p.Name, downloadPath, p.Oid, p.Size, false, nil)
 			}
 		})
 		gs.ScanRefs(opts.Include, opts.Exclude, nil)
--- a/commands/command_smudge.go
+++ b/commands/command_smudge.go
@@ -59,7 +59,7 @@ func delayedSmudge(gf *lfs.GitFilter, s *git.FilterProcessScanner, to io.Writer,
 
 	if !skip && filter.Allows(filename) {
 		if _, statErr := os.Stat(path); statErr != nil {
-			q.Add(filename, path, ptr.Oid, ptr.Size, false)
+			q.Add(filename, path, ptr.Oid, ptr.Size, false, err)
 			return 0, true, ptr, nil
 		}
 
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -131,9 +131,9 @@ func buildFilepathFilter(config *config.Configuration, includeArg, excludeArg *s
 	return filepathfilter.New(inc, exc)
 }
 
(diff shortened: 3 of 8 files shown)


leonid-s-usov:
Thanks for being this fast!

PR #4017, opened 2020-02-10
commands/command_filter_process: cherry-pick of several commits cause panic error

shalashik:
**Describe the bug**
You can receive the `panic: send on closed channel` error while cherry-pick several commits and lfs files are not presented in the cache.

The current `filter-process` realisation assumes that when the `list_available_blobs` command is received from GIT `smudge` command won't be  issued:  
https://github.com/git-lfs/git-lfs/blob/c3b776db6ee273a6c231f9d5922e3aa3421b7b3e/commands/command_filter_process.go#L124

But cherry-picking of several commits with LFS files demonstrate different behavior. It seems that GIT creates single long running process filter which is used for the entire life of a single Git command:
https://git-scm.com/docs/gitattributes#_code_filter_code

As a result GIT LFS can receive several `smudge` commands within single `filter-process`. The panic error appears when two `smudge` commands trigger `status=delayed` reply from LFS. The first `smudge` causes GIT LFS filter process to close transfer queue, the second `smudge` attempts to write in closed queue which causes panic.

Log: [log.txt](https://github.com/git-lfs/git-lfs/files/4182891/log.txt)
 
**To Reproduce**
I've created test to reproduce this issue:
t/t-cherry-pick-commits.sh
[t-cherry-pick-commits.sh.txt](https://github.com/git-lfs/git-lfs/files/4182890/t-cherry-pick-commits.sh.txt)

**Expected behavior**
Cherry-picked commits without panic

**System environment**
I was able to reproduce this issue in different environments including Windows, OSX and Ubuntu 18.08

**Output of `git lfs env`**
```
$ git lfs env
git-lfs/2.10.0 (GitHub; darwin amd64; go 1.13.6; git 7237dd0f)
git version 2.25.0

Endpoint=https://github.com/shalashik/gitlfstest.git/info/lfs (auth=none)
  SSH=git@github.com:shalashik/gitlfstest.git
LocalWorkingDir=/Users/shalashik/repos/pets/gitlfstest
LocalGitDir=/Users/shalashik/repos/pets/gitlfstest/.git
LocalGitStorageDir=/Users/shalashik/repos/pets/gitlfstest/.git
LocalMediaDir=/Users/shalashik/repos/pets/gitlfstest/.git/lfs/objects
LocalReferenceDirs=
TempDir=/Users/shalashik/repos/pets/gitlfstest/.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/Users/shalashik/repos/pets/gitlfstest/.git/lfs
AccessDownload=none
AccessUpload=none
GIT_EXEC_PATH=/usr/local/Cellar/git/2.25.0/libexec/git-core
```

### DIFF
--- a/commands/command_filter_process.go
+++ b/commands/command_filter_process.go
@@ -63,21 +63,10 @@ func filterCommand(cmd *cobra.Command, args []string) {
 	ptrs := make(map[string]*lfs.Pointer)
 
 	var q *tq.TransferQueue
-	closeOnce := new(sync.Once)
-	available := make(chan *tq.Transfer)
-
-	if supportsDelay {
-		q = tq.NewTransferQueue(
-			tq.Download,
-			getTransferManifestOperationRemote("download", cfg.Remote()),
-			cfg.Remote(),
-			tq.RemoteRef(currentRemoteRef()),
-		)
-		go infiniteTransferBuffer(q, available)
-	}
-
 	var malformed []string
 	var malformedOnWindows []string
+	var closeOnce *sync.Once
+	var available chan *tq.Transfer
 	gitfilter := lfs.NewGitFilter(cfg)
 	for s.Scan() {
 		var n int64
@@ -99,6 +88,19 @@ func filterCommand(cmd *cobra.Command, args []string) {
 				n = ptr.Size
 			}
 		case "smudge":
+			closeOnce = new(sync.Once)
+			available = make(chan *tq.Transfer)
+
+			if supportsDelay {
+				q = tq.NewTransferQueue(
+					tq.Download,
+					getTransferManifestOperationRemote("download", cfg.Remote()),
+					cfg.Remote(),
+					tq.RemoteRef(currentRemoteRef()),
+				)
+				go infiniteTransferBuffer(q, available)
+			}
+
 			w = git.NewPktlineWriter(os.Stdout, smudgeFilterBufferCapacity)
 			if req.Header["can-delay"] == "1" {
 				var ptr *lfs.Pointer
@@ -124,8 +126,8 @@ func filterCommand(cmd *cobra.Command, args []string) {
 			closeOnce.Do(func() {
 				// The first time that Git sends us the
 				// 'list_available_blobs' command, it is given
-				// that no more smudge commands will be issued
-				// with _new_ checkout entries.
+				// that now it waiting until all delayed blobs
+				// are available within this smudge filter call
 				//
 				// This means that, by the time that we're here,
 				// we have seen all entries in the checkout, and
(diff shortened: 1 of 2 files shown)


t/t-cherry-pick-commits.sh
```diff
@@ -0,0 +1,36 @@
...
+  git add b.dat
+  git commit -m "add a.dat"
+  commit2=$(git log -n1 --format="%H")
+
+  git push origin master
+
+  git checkout secondbranch
+  rm -rf .git/lfs/objects
+
+  GIT_TRACE_PACK_ACCESS=2 git cherry-pick $commit1 $commit2
```
bk2204: Hmm, do we need this `GIT_TRACE_PACK_ACCESS`?

commands/command_filter_process.go
```diff
@@ -99,6 +88,19 @@ func filterCommand(cmd *cobra.Command, args []string) {
...
+			closeOnce = new(sync.Once)
+			available = make(chan *tq.Transfer)
+
+			if supportsDelay {
+				q = tq.NewTransferQueue(
+					tq.Download,
+					getTransferManifestOperationRemote("download", cfg.Remote()),
+					cfg.Remote(),
+					tq.RemoteRef(currentRemoteRef()),
+				)
```
bk2204: If we use a new transfer queue on each smudge command, then we'll issue a whole new batch command for each object, which would be very inefficient, since we're essentially downloading each object one by one.  That's the issue that the filter-process is trying to avoid.

I think what we might want to do instead is track if the queue has been closed (say, by setting `q` to `nil` when we do), and then create a new one only if one doesn't exist.  Does that sound reasonable to you?

t/t-cherry-pick-commits.sh
shalashik: Thank you. The `GIT_TRACE_PACK_ACCESS` is not necessary there. Removed it in https://github.com/shalashik/git-lfs/commit/26c58f

commands/command_filter_process.go
shalashik: Thank you for pointing me to right direction to solve the issue. I'll need a few days to create the appropriate solution and test it, will get back to you asap.

shalashik: I've pushed new commit: https://github.com/shalashik/git-lfs/commit/ddbe7eacab32d4ba83254c6d180f26afe3cfec69

PR #4022, opened 2020-02-11
README: explain how to verify releases

### LINKED ISSUE #4018: Verifying Binaries
**Describe the issue**
I am trying to verify the hashes of the binaries posted on the GitHub release page. I am trying to verify them in a systematic way, so I can put the commands in a Dockerfile. 

Usually, I use `sha256sum` to verify the downloads:

```
wget --no-check-certificate --no-cookies http://archive.apache.org/dist/ant/binaries/apache-ant-${ANT_VERSION}-bin.tar.gz \
    && wget --no-check-certificate --no-cookies http://archive.apache.org/dist/ant/binaries/apache-ant-${ANT_VERSION}-bin.tar.gz.sha512 \
    && echo "$(cat apache-ant-${ANT_VERSION}-bin.tar.gz.sha512) apache-ant-${ANT_VERSION}-bin.tar.gz" | sha512sum -c
```

Or if the file is signed with a GPG I do the following:

```
wget --no-check-certificate --no-cookies https://piccolo.link/sbt-${SBT_VERSION}.tgz \
    && wget --no-check-certificate --no-cookies https://piccolo.link/sbt-${SBT_VERSION}.tgz.asc \
    && curl "https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x2EE0EA64E40A89B84B2DF73499E82A75642AC823" | gpg --import \
    && gpg --verify sbt-${SBT_VERSION}.tgz.asc sbt-${SBT_VERSION}.tgz \
```

But since the file you provide is a combination of the both I am confused on how to verify this systematically. 

Also what key are you using to sign these files? I do not see that mentioned anywhere. When I look around for a key I only see the one mentioned on packagecloud.

**Additional context**
What I want to do is either just download the `sha256` file and verify the hashes. Or import your key and verify the signature. Like how https://github.com/krallin/tini/releases does it.

bk2204:
Some folks aren't familiar with the signed SHA-256 sums file we use, or aren't sure how to get the keys to verify releases.  To help them, point them to the tag we've created that contains the keys (similar to the way Git does this) and add an explanation to the README to help them verify the releases.

Currently, I'm the only core team member with keys, so I'm the only one with keys in the tree.  If other folks add keys to their GitHub profile, ping me and I'll rewrite the tag.  I've opted to use a tree instead of the blob that Git uses (git/git's `junio-gpg-pub`) because I can't find a way to let people download a tagged blob from GitHub, while tagged trees do work.

Fixes #4018

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,23 @@ running:
 $ git lfs install
 ```
 
+#### Verifying releases
+
+Releases are signed with the OpenPGP key of one of the core team members.  To
+get these keys, you can run the following command, which will print them to
+standard output:
+
+```ShellSession
+$ curl -L https://api.github.com/repos/git-lfs/git-lfs/tarball/core-gpg-keys | tar -Ozxf -
+```
+
+Once you have the keys, you can download the `sha256sums.asc` file and verify
+the file you want like so:
+
+```ShellSession
+$ gpg -d sha256sums.asc | grep git-lfs-linux-amd64-v2.10.0.tar.gz | shasum -a 256 -c
+```
+
 ## Example Usage
 
 To begin using Git LFS within a Git repository that is not already configured


bk2204:
I've updated this to use an API URL instead of a codeload URL for future-proofing.

PR #4026, opened 2020-02-12
Stop using cgo on amd64 Linux

### LINKED ISSUE #4025: Issues with binary install on Alpine
**Describe the bug**
When performing a binary install on Alpine, I get the following error:

```
bash-5.0# ./install.sh
fatal: 'lfs' appears to be a git command, but we were not
able to execute it. Maybe git-lfs is broken?
```

I tried two methods to install `git-lfs`. First I tried to follow the official method of using the `install.sh` file. After that failed I tried to do it manually by moving `git-lfs` into `/usr/bin/` and then updating the `PATH`. Both seemed to produce the above error.

Manually, steps followed:

```
curl -fsSLO https://github.com/git-lfs/git-lfs/releases/download/v2.10.0/git-lfs-linux-amd64-v2.10.0.tar.gz
tar -zvxf git-lfs-linux-amd64-v2.10.0.tar.gz
mv git-lfs /usr/bin/ 
export PATH=$PATH:/usr/bin
```   

**To Reproduce**
See above ^^^

**Expected behavior**
I expect either install method to work. This install method works on `debian` based system just fine. 

**System environment**
Testing out in a Docker image: `alpine:latest`

**Output of `git lfs env`**
N/A

**Additional context**
A work around is just to use `apk add git-lfs`. This works but deviates how we install `git-lfs` in our other Dockefiles for [Jenkins](https://github.com/jenkinsci/docker).

bk2204:
This series allows us to stop using cgo on amd64 Linux, which makes our binaries static and therefore portable to musl-based systems.  The commit messages explain more in detail.

The constant was generated using a throwaway C program.

Fixes #4025.

### DIFF
--- a/tools/util_linux.go
+++ b/tools/util_linux.go
@@ -1,16 +1,6 @@
-// +build linux,cgo
+// +build linux
 
 package tools
-
-/*
-#include <sys/ioctl.h>
-
-#undef FICLONE
-#define FICLONE		_IOW(0x94, 9, int)
-// copy from https://github.com/torvalds/linux/blob/v5.2/include/uapi/linux/fs.h#L195 for older header files.
-// This is equal to the older BTRFS_IOC_CLONE value.
-*/
-import "C"
 
 import (
 	"io"
@@ -20,7 +10,10 @@ import (
 )
 
 const (
-	ioctlFiClone = C.FICLONE
+	// This is the FICLONE ioctl value found in linux/fs.h on a typical
+	// system.  It's equivalent to the older BTRFS_IOC_CLONE.  We hard-code
+	// the value here to avoid a dependency on cgo.
+	ioctlFiClone = 0x40049409
 )
 
 // CheckCloneFileSupported runs explicit test of clone file on supplied directory.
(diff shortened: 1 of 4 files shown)


tools/util_linux.go
```diff
@@ -20,7 +10,7 @@ import (
 )
 
 const (
-	ioctlFiClone = C.FICLONE
+	ioctlFiClone = 0x40049409
```
chrisd8088: It might be valuable for future maintainers if this came with a comment akin to the one removed above from the `import "C"` section, just explaining where it comes from ... I realize the name of the variable provides the context on its own, but I figure the link to the Linux source would ease comprehension too.

bk2204: Sure, I can do that.

bk2204: Done.

james-crowley:
@bk2204 Thanks for the PR. Will these changes be backported to `v2.10.0` or be on the next release of `git-lfs`?

bk2204:
They'll be in the next release, which will likely be v2.11.

PR #4034, opened 2020-02-21
Clearer output when expected file isn't a pointer

ThomasNigro:
Hello,

Suggesting this minor output change to make the error message clearer.

### DIFF
--- a/commands/command_filter_process.go
+++ b/commands/command_filter_process.go
@@ -203,7 +203,7 @@ func filterCommand(cmd *cobra.Command, args []string) {
 	}
 
 	if len(malformed) > 0 {
-		fmt.Fprintf(os.Stderr, "Encountered %d file(s) that should have been pointers, but weren't:\n", len(malformed))
+		fmt.Fprintf(os.Stderr, "Encountered %d file(s) that should have been Git LFS pointers, but they're Git files:\n", len(malformed))
 		for _, m := range malformed {
 			fmt.Fprintf(os.Stderr, "\t%s\n", m)
 		}


commands/command_filter_process.go
```diff
@@ -203,7 +203,7 @@ func filterCommand(cmd *cobra.Command, args []string) {
 	}
 
 	if len(malformed) > 0 {
-		fmt.Fprintf(os.Stderr, "Encountered %d file(s) that should have been pointers, but weren't:\n", len(malformed))
+		fmt.Fprintf(os.Stderr, "Encountered %d file(s) that should have been Git LFS pointers, but they're Git files:\n", len(malformed))
```
bk2204: This is definitely more detailed, but it's also more than 80 characters long, which means it will wrap on some screens.  Can we shorten it?  Maybe `Encountered %d file(s) that should have been Git LFS pointers, but weren't`?  Or something else?

ThomasNigro: Your suggestion is definitely better as it will fit in one line. However I'm afraid it might not be clear enough. I'm now getting inspiration from the next `if` condition, which displays an initial error message, then loops on a collection, then display a help message with a `git lfs` command to get more details. I think we could do the same: an initial (short) but a bit clearer output message, and then instructions on what to do next.

commands/command_filter_process.go
```diff
@@ -203,10 +203,14 @@ func filterCommand(cmd *cobra.Command, args []string) {
...
 	if len(malformed) > 0 {
-		fmt.Fprintf(os.Stderr, "Encountered %d file(s) that should have been Git LFS pointers, but they're Git files:\n", len(malformed))
+		fmt.Fprintf(os.Stderr, "Encountered %d file(s) that should have been Git LFS pointers, but weren't:\n", len(malformed))
+
 		for _, m := range malformed {
 			fmt.Fprintf(os.Stderr, "\t%s\n", m)
 		}
+
+		fmt.Fprintf(os.Stderr, "\Type: `git lfs ls-files` to see which files are in the Git LFS index.\n");
+		fmt.Fprintf(os.Stderr, "\See: `git lfs help migrate` for migrating files to the Git LFS index.\n");
```
bk2204: I think these two lines have syntax errors.

Also, since this is a warning message, we should try to be fairly brief.  I think we could just say, “See "git lfs migrate --help" for how to migrate these files to Git LFS.” I'm not sure we need the previous line, and there isn't really a Git LFS index separate from the Git index.

Filipova96 (approved):

PR #4066, opened 2020-03-10
lfshttp: allow renegotiation

### LINKED ISSUE #4062: "local error: tls: no renegotiation" when using CA certificate
GIT LFS throws following error when CA certificate is used with Stash/Bitbucket server.

```
15:52:19.245451 trace git-lfs: api error: Post https://mycompany.com/scm/myproj/git_lfs.git/info/lfs/objects/batch: local error: tls: no renegotiation
batch response: Post https://mycompany.com/scm/myproj/git_lfs.git/info/lfs/objects/batch: local error: tls: no renegotiation
Uploading LFS objects: 0% (0/1), 0 B | 0 B/s, done
error: failed to push some refs to 'ssh://git@mycomany.com:7999/myproj/git_lfs.git'
```

On server side, /info/lfs/locks/verify was received and responded with 200 but /info/lfs/objects/batch was never received.

GIT LFS VERSION : 2.9.2
GIT version : 2.15.1

bk2204:
There are some TLS servers that want to perform a renegotiation to request users use a client certificate.  Allow us to use renegotiation as a client when making a TLS connection.  The DoS potential for this is low, since the server must not only perform another key exchange, but must re-sign the request, leading to a larger performance impact on the server than the client.

Fixes #4062 

@manishgo, can you please test this commit and see if it works for you?  If you need binaries, they should be available for download from the CI builds.

### DIFF
--- a/lfshttp/client.go
+++ b/lfshttp/client.go
@@ -449,7 +449,9 @@ func (c *Client) Transport(u *url.URL, access creds.AccessMode) (http.RoundTripp
 		tr.DialContext = dialer.DialContext
 	}
 
-	tr.TLSClientConfig = &tls.Config{}
+	tr.TLSClientConfig = &tls.Config{
+		Renegotiation: tls.RenegotiateFreelyAsClient,
+	}
 
 	if isClientCertEnabledForHost(c, host) {
 		tracerx.Printf("http: client cert for %s", host)


bk2204:
Okay, I'm going to assume that this does indeed fix the problem and open it up for review.

bk2204:
Yeah, it should be fine here.  I'm pretty sure Go doesn't support insecure renegotiation, and as mentioned, the DoS potential is low.

PR #4084, opened 2020-03-28
make Go tests run consistently using local binary

### LINKED ISSUE #4046: some Go tests fail due to missing local temp directories
**Describe the bug**

When running the Go test suite on some development systems (macOS, at least) with `make test`, some tests fail with the errors of the form:
```
    testutils.go:228: Error running git command 'git push origin ...': exit status 1 EOF
        Uploading LFS objects: ...
        error: failed to push some refs to '/var/folders/.../T/lfsRepo...'
```

Running `make test` again with the `GIT_TRACE=1` environment variable yields the information that the given path plus `/.git` does not exist as a local directory.  As an example:
```
12:21:40.442232 trace git-lfs: xfer[git-lfs]: error creating handler: chdir /var/folders/f4/3tj6q1w11xd4mnnx7fm3jk8m0000gn/T/lfsRepo758791364/.git: no such file or directory
        EOF
        Uploading LFS objects: 100% (2/2), 45 B | 0 B/s, done.
        error: failed to push some refs to '/var/folders/f4/3tj6q1w11xd4mnnx7fm3jk8m0000gn/T/lfsRepo758791364'
```

**To Reproduce**
Steps to reproduce the behaviour:
1. Build from source per the Wiki [instructions](https://github.com/git-lfs/git-lfs/wiki/Installation#source).
2. Run the Go test suite with `make test` (not the shell tests in the `t/` directory).
3. Observe the `TestCurrentRefAndCurrentRemoteRef`, `TestRecentBranches`, and `TestScanUnpushed` tests fail with errors as described above.

**Expected behaviour**
All Go tests should pass.

**System environment**
macOS Catalina with Go version 1.13.3.

**Output of `git lfs env`**
Not applicable.

**Additional context**
Not applicable.

chrisd8088:
The PR attempts to refine a number of related issues with the Go test suite and the `make test` Makefile target.

The `make test` target is adjusted so that it depends on the local `bin/git-lfs[.exe]` binary being built first, and then the Go tests are run with `-count=1` so no caching of results is done between invocations.  (A developer wishing for Go's default caching of package list results can use `go test ./...` instead of `make test`.)

The common test utility package in `t/cmd/util/testutils.go` is given an `init()` function which attempts to ensure that the first path in the `PATH` environment variable's list is an absolute path to the working tree's `bin` directory.  This prevents the tests, when they run `git-lfs` as a system command, from accidentally using whatever `git-lfs` binary happens to be installed on the system instead, which obviously can produce inconsistent or unexpected test results.

Finally, the standalone transfer adapter's logic for running `git rev-parse --git-dir` when setting up a new handler based on a file path is refined so that it first removes any trailing `/.git` path segment.  This ensures the logic succeeds when presented with the path to a bare repository (which lacks a `.git` directory) even when the path has had `/.git` appended by [`EndpointFromLocalPath()`](https://github.com/git-lfs/git-lfs/blob/e8be7992b1c42956a351abf876e7c137b2254d83/lfshttp/endpoint.go#L112).

As [discussed](https://github.com/git-lfs/git-lfs/issues/4046#issuecomment-604649900) in #4046, we might be able to change `EndpointFromLocalPath()` as well, or instead, of changing `gitDirAtPath()`.  I chose this approach primarily because I was concerned that I didn't fully understand the potential use cases for custom transfer adapters, such as were described in #2912, and whether users might have configurations which depended on the existing behaviour of the `EndpointFromLocalPath()` function.

I believe the CI test failures are due to the issue addressed by #4082.

Fixes #4046.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -406,7 +406,7 @@ test-race : GO_TEST_EXTRA_ARGS=-race
 # And so on.
 test : fmt
 	(unset GIT_DIR; unset GIT_WORK_TREE; \
-	$(GO) test $(GO_TEST_EXTRA_ARGS) $(addprefix ./,$(PKGS)))
+	$(GO) test -count=1 $(GO_TEST_EXTRA_ARGS) $(addprefix ./,$(PKGS)))
 
 # integration is a shorthand for running 'make' in the 't' directory.
 .PHONY : integration
--- a/Makefile
+++ b/Makefile
@@ -404,7 +404,7 @@ test-race : GO_TEST_EXTRA_ARGS=-race
 # 		make PKGS="config lfsapi tools/kv" test-race
 #
 # And so on.
-test : fmt
+test : fmt $(.DEFAULT_GOAL)
 	(unset GIT_DIR; unset GIT_WORK_TREE; \
 	$(GO) test -count=1 $(GO_TEST_EXTRA_ARGS) $(addprefix ./,$(PKGS)))
 
--- a/lfshttp/endpoint.go
+++ b/lfshttp/endpoint.go
@@ -3,6 +3,7 @@ package lfshttp
 import (
 	"fmt"
 	"net/url"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -108,9 +109,6 @@ func EndpointFromHttpUrl(u *url.URL) Endpoint {
 }
 
 func EndpointFromLocalPath(path string) Endpoint {
-	if !strings.HasSuffix(path, ".git") {
-		path = fmt.Sprintf("%s/.git", path)
-	}
 	var slash string
 	if abs, err := filepath.Abs(path); err == nil {
 		// Required for Windows paths to work.
@@ -118,6 +116,17 @@ func EndpointFromLocalPath(path string) Endpoint {
 			slash = "/"
 		}
 		path = abs
+	}
+
+	var gitpath string
+	if filepath.Base(path) == ".git" {
+		gitpath = path
+		path = filepath.Dir(path)
+	} else {
+		gitpath = filepath.Join(path, ".git")
+	}
+	if file, err := os.Lstat(gitpath); err == nil && file.IsDir() {
+		path = gitpath
 	}
 	return Endpoint{Url: fmt.Sprintf("file://%s%s", slash, filepath.ToSlash(path))}
 }
--- a/lfshttp/endpoint.go
+++ b/lfshttp/endpoint.go
@@ -125,7 +125,7 @@ func EndpointFromLocalPath(path string) Endpoint {
 	} else {
 		gitpath = filepath.Join(path, ".git")
 	}
-	if file, err := os.Lstat(gitpath); err == nil && file.IsDir() {
+	if _, err := os.Stat(gitpath); err == nil {
 		path = gitpath
 	}
 	return Endpoint{Url: fmt.Sprintf("file://%s%s", slash, filepath.ToSlash(path))}
--- a/lfshttp/standalone/standalone.go
+++ b/lfshttp/standalone/standalone.go
@@ -101,6 +101,11 @@ func gitDirAtPath(path string) (string, error) {
 	}
 	env = env[:n]
 
+	// Trim any trailing .git path segment.
+	if filepath.Base(path) == ".git" {
+		path = filepath.Dir(path)
+	}
+
 	curdir, err := os.Getwd()
 	if err != nil {
 		return "", err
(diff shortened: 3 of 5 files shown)


chrisd8088:
I think the work you did in your `standalone-failure` branch, @bk2204 (per https://github.com/git-lfs/git-lfs/issues/4046#issuecomment-597116097), may be useful but orthogonal ... you mentioned, I think, that it resolved some other issues on the CI servers?  In any case, if it's valuable, I didn't want it to be forgotten.

lfshttp/standalone/standalone.go
```diff
@@ -101,6 +101,11 @@ func gitDirAtPath(path string) (string, error) {
 	}
 	env = env[:n]
 
+	// Trim any trailing .git path segment.
+	if filepath.Base(path) == ".git" {
+		path = filepath.Dir(path)
+	}
+
```
bk2204: So I think this solves our present problem.  However, in the general case, I think our testsuite is telling us something: that we're passing a remote value that isn't valid.  If we have a bare repository, then that's the equivalent of the `.git` directory; there isn't a `.git` directory inside it, and so we're right to reject it.

I understand that it's nice to maintain backward compatibility here, but I don't think it makes sense for us to pass a URL that isn't valid over to the adapter and then fix it up.  I wonder if we could try a different approach: calling `os.Stat` on the path with the `.git` component in `EndpointFromLocalPath` and using it if it's valid, and omitting it if it's not.  That would preserve backwards compatibility while still making sure we have valid data.

Makefile
```diff
@@ -406,7 +406,7 @@ test-race : GO_TEST_EXTRA_ARGS=-race
 # And so on.
 test : fmt
 	(unset GIT_DIR; unset GIT_WORK_TREE; \
-	$(GO) test $(GO_TEST_EXTRA_ARGS) $(addprefix ./,$(PKGS)))
+	$(GO) test -count=1 $(GO_TEST_EXTRA_ARGS) $(addprefix ./,$(PKGS)))
```
bk2204: Just for my enlightenment, is there a place where the caching is not doing the right thing here?

chrisd8088: No, the caching is doing the right thing -- albeit, if you're invoking the wrong `git-lfs`, it caches those results, hence my confusion in thinking there might be some transient behaviour in the tests.  With the other changes in this PR, that _should_ stop being an issue, so one could leave this as-is.

But, speaking just for myself, I expect a test suite to run fully every time, at least if I'm using a `make test` kind of invocation.  I realize Go likes to cache its test results, although only when using "package list mode".  For my part, if I'm using a wrapper like `make test` I may not realize what's going on under the hood, and if it's being invoked in a CI context, I'd prefer to know it is forcing a full redo on all the tests.

Anyway, that's why I figured this was a reasonable change; someone can always use `go test ./...` if they want the default Go behaviour.  But I can live without it too.  :-)

lfshttp/standalone/standalone.go
chrisd8088: I'm happy to do that instead, or as well, since you advise it!  I figure this change, in itself, is effective and might as well stay, but having a belt-and-suspenders approach and fixing the output from `EndpointFromLocalPath()` is good with me.  I'll take a look at that next.

chrisd8088: OK, I've tried to implement your suggestion, @bk2204 -- see how that sits with you now.

lfshttp/endpoint.go
```diff
@@ -119,6 +117,17 @@ func EndpointFromLocalPath(path string) Endpoint {
...
 	}
+
+	var gitpath string
+	if filepath.Base(path) == ".git" {
+		gitpath = path
+		path = filepath.Dir(path)
+	} else {
+		gitpath = filepath.Join(path, ".git")
+	}
+	if file, err := os.Lstat(gitpath); err == nil && file.IsDir() {
```
bk2204, line 128: It's possible to have a file as the `.git` path (referred to as a _gitlink_), such as in a submodule or worktree, and I think we'd definitely want to let people push into submodules.  This is a tricky corner case, but I think we'd probably be fine with just checking if it exists.

It also occurred to me that this behavior is broken in our standalone helper, so I'll fix that in an independent PR.

bk2204: Actually, this is not broken in our standalone helper (:tada:), but we lack tests for it, so I'm going to add those tests in a separate PR.

chrisd8088: OK, I've adjusted this to just use `os.Stat()` and not try to check that the `.git` segment refers to a directory, just that it exists.

bk2204:
It looks like we had a temporary network problem building the Docker images, so I kicked off a restart, but it looks like that broke things.  You may just want to rebase on `master` and force-push to kick off a new CI job.

PR #4096, opened 2020-04-14
git: avoid trying to rewrite remote tags as remote branches

### LINKED ISSUE #3818: git lfs migrate import --fixup not working; fatal error with "bad revisions"
```
herbst@PFC MINGW32 /e/git/project (master)
$ git lfs migrate import --fixup
migrate: Fetching remote refs: ..., done
Error in git rev-list --stdin --reverse --topo-order --do-walk --: exit status 1
28 fatal: bad revision '^refs/remotes/origin/Alpha-0.5'

$ git lfs version
git-lfs/2.6.1 (GitHub; windows 386; go 1.11.1; git dc072c3e)
```

See #3285 and #3281 which are closed and state that this is supposed to be fixed in lfs 2.6.0.

**Edit:** just updated to git-lfs 2.7.1 but the behaviour stays the same. Git version is 2.21.0.windows.1.

bk2204:
When invoking "git lfs migrate import --fixup", we call Git to fetch the remote branches and tags.  However, we mark both as having the type remote branch, so when we look up the local copies for them, we try to use the tag name as the name of a remote tracking branch, which fails.

Instead, let's mark them as remote tags, which means we won't try to rewrite them into branches at all.  This ensures the operation can succeed.

Fixes #3818

### DIFF
--- a/t/t-migrate-fixup.sh
+++ b/t/t-migrate-fixup.sh
@@ -100,3 +100,32 @@ begin_test "migrate import (--fixup, --no-rewrite)"
   grep -q "fatal: --no-rewrite and --fixup cannot be combined" migrate.log
 )
 end_test
+
+begin_test "migrate import (--fixup with remote tags)"
+(
+  set -e
+
+  setup_single_local_branch_tracked_corrupt
+
+  git lfs uninstall
+
+  base64 < /dev/urandom | head -c 120 > b.txt
+  git add b.txt
+  git commit -m "b.txt"
+
+  git tag -m tag1 -a tag1
+  git reset --hard HEAD^
+
+  git lfs install
+
+  cwd=$(pwd)
+  cd "$TRASHDIR"
+
+  git clone "$cwd" "$reponame-2"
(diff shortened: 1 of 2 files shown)


t/t-migrate-fixup.sh
```diff
@@ -100,3 +100,34 @@ begin_test "migrate import (--fixup, --no-rewrite)"
...
+
+  git lfs install
+
+  cwd=$(pwd)
+  cd "$TRASHDIR"
+
+  git clone "$cwd" "$reponame-2"
+  cd "$reponame-2"
+
+  txt_oid="$(calc_oid "$(git cat-file -p :a.txt)")"
```
chrisd8088, line 127: Is this line useful here?

bk2204: No, it is not, and I'll remove it.

bk2204:
It does still return the branches and tags, but we just don't use the tag values it returns.  I will fix the typo, though, while I'm here.

PR #4097, opened 2020-04-14
tq: add exponential backoff for retries

bluekeyes:
### Background

Our GitHub Enterprise server occasionally returns 500 errors for certain objects for a few seconds at a time (we've seen up to 10) and then recovers. We're tracking this separately with GitHub support, but the problem is made worse by LFS's retry behavior. Because retries happen as fast as possible, even a higher-than-default number of retries is not enough to get past these temporary errors.

### Proposed Changes

This PR adds exponential backoff for retries with a configurable maximum delay and enables them by default. The delays starts at 250ms and double with each retry up to a maximum of 10 seconds. The new `lfs.transfer.maxretrydelay` property controls the maximum value and can also disable delays if set to 0.

Ten seconds is an arbitrary maximum, but seemed like a good compromise between waiting long enough for a server to recover and waiting so long that LFS appears to stall. As before, if a server provides the `Retry-After` header, that value is used instead and the retry is no counted against the limit.

With the default configuration (8 retries, 10s max delay), a failed operation will now take 35.75 seconds before failing. I'm not sure if this is too long or not, so it may make sense to adjust either the number of retries, the base delay, or the maximum delay.

### Other Considerations

- The maximum delay value seemed the like the minimal useful configuration, but maybe the base delay or other options should be configurable as well.
- Delays are enabled by default because this should improve reliability without hurting performance in most cases.
- I applied the delay to all transfer-related retries, but perhaps it should only apply to the API request at the start of the batch.


tq/transfer_queue.go
```diff
@@ -530,10 +559,10 @@ func (q *TransferQueue) enqueueAndCollectRetriesFor(batch batch) (batch, error)
 			// retried, they will be marked as failed.
 			for _, t := range batch {
 				if q.canRetryObject(t.Oid, err) {
-					q.rc.Increment(t.Oid)
-
-					next = append(next, t)
+					enqueueRetry(t, err)
 				} else if readyTime, canRetry := q.canRetryObjectLater(t.Oid, err); canRetry {
+					// Do not use enqueueRetry() here because retries requested
+					// by a server do not count against the normal retry limit
```
bluekeyes, line 565: I don't know if this is actually the desired behavior: `rc.Increment` is not called, but it's unclear if that was intentional or an oversight.

bk2204: I definitely think we should count them against the limit.

docs/man/git-lfs-config.5.ronn
```diff
@@ -160,6 +160,14 @@ be scoped inside the configuration for a remote.
   not an integer, is less than one, or is not given, a value of eight will be
   used instead.
 
+* `lfs.transfer.maxretrydelay`
+
+  Specifies the maximum time in seconds LFS will wait between each retry
+  attempt. LFS uses exponential backoff for retries, doubling the time between
```
bk2204, line 166: Is this still true if the server gives us a larger value in the `Retry-After` header?

tq/transfer_queue.go
```diff
@@ -506,6 +522,19 @@ func (q *TransferQueue) enqueueAndCollectRetriesFor(batch batch) (batch, error)
...
+		count := q.rc.Increment(t.Oid)
+		t.ReadyTime = q.rc.ReadyTime(t.Oid)
+		delay := time.Until(t.ReadyTime).Seconds()
+
+		var errMsg string
+		if err != nil {
+			errMsg = ": " + err.Error()
+		}
+		tracerx.Printf("tq: enqueue retry #%d after %.2fs for %q (size: %d)%s", count, delay, t.Oid, t.Size, errMsg)
+		next = append(next, t)
```
bk2204, line 535: I take it this is written inline and not as an independent method on the queue so that we can close over `next`.

tq/transfer_queue.go
```diff
@@ -506,6 +522,19 @@ func (q *TransferQueue) enqueueAndCollectRetriesFor(batch batch) (batch, error)
...
 	tracerx.Printf("tq: sending batch of size %d", len(batch))
 
+	enqueueRetry := func(t *objectTuple, err error) {
+		count := q.rc.Increment(t.Oid)
+		t.ReadyTime = q.rc.ReadyTime(t.Oid)
+		delay := time.Until(t.ReadyTime).Seconds()
+
+		var errMsg string
+		if err != nil {
+			errMsg = ": " + err.Error()
```
bk2204, line 532: We usually would prefer to write this as `fmt.Sprintf(": %s", err)`.  That's slightly more efficient as well as more common in our codebase.

tq/transfer_queue_test.go
```diff
@@ -31,6 +44,36 @@ func TestRetryCounterCanNotRetryAfterExceedingRetryCount(t *testing.T) {
...
+	rc := newRetryCounter()
+	start := time.Now()
+
+	rc.Increment("oid")
+	ready1 := rc.ReadyTime("oid")
+	assert.True(t, ready1.Sub(start) >= baseRetryDelayMs*time.Millisecond)
+
+	rc.Increment("oid")
+	ready2 := rc.ReadyTime("oid")
+	assert.True(t, ready2.Sub(start) >= 2*baseRetryDelayMs*time.Millisecond)
```
bk2204, line 62: Can we use `assert.GreaterOrEqual` here and above so that we get a better error message on failure?

docs/man/git-lfs-config.5.ronn
bluekeyes: Not quite. If I change it so that `Retry-After` counts against the retry limit, then the simplest behavior to implement is to wait for the server-provided time and then if that fails again _and_ the server does not provide another time, wait for the delay implied by the retry count. The `Retry-After` value always wins for a particular retry, but does not influence the the length of future delays.

Maybe noting that the `Retry-After` value is always used if provided but otherwise doesn't change the delay behavior in the docs is sufficient?

tq/transfer_queue.go
bluekeyes: Yes, but I can also make it an independent method on the queue that takes a `*batch` if you prefer that style.

tq/transfer_queue_test.go
bluekeyes: It's not in the version of `stretchr/testify` that's currently vendored, but I can updated that.

docs/man/git-lfs-config.5.ronn
bluekeyes: I updated this paragraph to clarify the behavior when a server provides the `Retry-After` header.

tq/transfer_queue.go
bk2204: It's fine as it is.  I just wanted to be clear on what's going on.

tq/transfer_queue_test.go
```diff
@@ -31,6 +44,36 @@ func TestRetryCounterCanNotRetryAfterExceedingRetryCount(t *testing.T) {
...
+	assert.Equal(t, time.Time{}, rc.ReadyTime("oid"))
+}
+
+func TestRetryCounterDelaysExponentially(t *testing.T) {
+	rc := newRetryCounter()
+	start := time.Now()
+
+	rc.Increment("oid")
+	ready1 := rc.ReadyTime("oid")
+	assert.GreaterOrEqual(t, ready1.Sub(start).Milliseconds(), int64(baseRetryDelayMs))
```
bk2204, line 58: It looks like this item and the below one are failing since they don't work on Go 1.12.  Could we maybe use `Nanoseconds` and multiply?

docs/man/git-lfs-config.5.ronn
bk2204: Great.  That's exactly what I was looking for.

bluekeyes:
Thanks for the reviews. I fixed my new test and rebased on `master`, but it looks like there is now an unrelated failure in a different test and I'm not sure how to fix it.

bk2204:
I think this just happens to be a test that randomly failed, since it looks like it worked on the other systems.  I've tried to kick off a rebuild, but it turns out the Actions workflow doesn't like that.  If you can just `git rebase -m -f master` and force-push again , I think that will fix it.

PR #4109, opened 2020-04-27
Add lock creation date to 'git lfs locks' output

slonopotamus:


slonopotamus:
[CI failure](https://github.com/git-lfs/git-lfs/pull/4109/checks?check_run_id=621615779#step:4:224) isn't caused by changes in this PR.

slonopotamus:
Before doing that, I have a question: is there any way to list locks in machine-friendly format? If no, people might parse output of `git lfs locks` and we would break them, so it would be safer to move lock creation timestamp to the _last_ column.

bk2204:
There is a `--json` option.  I did check that before responding.  Although, since you mention it, putting it at the end would be a good idea for folks who don't want to use JSON, so if we can, let's do that.

Plataformaalgoritmica (commented):
Hola mundo soy Max,sugerencia de refuerzo de seguridad.

Plataformaalgoritmica:
Hola mundo soy Max Ramírez.

PR #4114, opened 2020-04-30
Add fetch/prune options to safe keys for lfsconfig

### LINKED ISSUE #4111: Allow to set fetch options in .lsfconfig
At the end of [config/git_fetcher.go](https://github.com/git-lfs/git-lfs/blob/master/config/git_fetcher.go) there is a list of safe keys which limit what can be read from `.lfsconfig`. Some of the fetch options like `lfs.fetchinclude/exclude` are marked as safe, while others are not.
I agree that some settings are not safe, however, there are others that are but are disabled regardless.
Namely, the remaining fetch settings:
```
lfs.fetchrecentrefsdays
lfs.fetchrecentremoterefs
lfs.fetchrecentcommitsdays
lfs.fetchrecentalways
```

In my case, we have a team of testers which switch regularly between branches. Without specifying `--recent` or setting `lfs.fetchrecentalways` you quickly end up on `branchA` wanting to switch to `branchB` but lfs files are missing while recent commits have been received via`git pull` on `branchA`. Now a working network connection is required to switch branches and access required lfs files on `branchB`
We want to specify sensible defaults for our testers without setting these properties locally.

But whatever the use-case, I think these settings can be considered safe and there shouldn't be a reason to exclude them even if the use-case seems limited.

JungStar:
Some users want to specify other defaults for fetch/prune options, e.g.,
enable lfs.fetchrecentalways.

Add fetch and prune options to list of safe keys for .lfsconfig.
Update the documentation to include the new list safe keys.

This adds the following keys to the list of safe keys:
- lfs.fetchrecentrefsdays
- lfs.fetchrecentremoterefs
- lfs.fetchrecentcommitsdays
- lfs.fetchrecentalways
- lfs.pruneoffsetdays

Fixes #4111


Bosch-Eli-Black:
I realize this PR is pretty old, but I'd still be interested in having this merged 🙂  Anyone around to review? 🙂

bk2204:
I think my position as outlined in #4111 still stands.  I'm not comfortable with adding more keys to the `.lfsconfig` in this way, but if the PR is fixed to avoid the conflict, perhaps one of the other core team members would be willing to do it.

Bosch-Eli-Black:
Thanks, @bk2204 🙂  After thinking about this some more, I think our team probably wouldn't use this feature, anyway, because we wouldn't want these settings to apply to both our dev machines and the build server.

PR #4116, opened 2020-05-01
commands/command_migrate.go: fix bug

Electric26:
Default hint ([Y/n]) wasn't being honored - 'y' is now default.


chrisd8088:
Hey, thanks for the patch!  This seems quite reasonable to me ... I'll let @bk2204 weigh in next week too, but I suspect this is OK as-is.  Thanks again!

Issue #2464, opened 2017-07-31
[Question] ARM Support


zafields:
It appears that you don't release a package for ARM. Is this for technical reasons, or is ARM not on your radar? Do you intend to support ARM sometime in the future? Do you already support ARM and I'm just missing it entirely?

ttaylorr:
@zafields thanks for opening this! Great question. LFS can be built to target ARM and ARM64 architecture (see: `GOOS=linux GOARCH=arm64 script/bootstrap`, and https://github.com/git-lfs/git-lfs/pull/2088), but we don't offer official support or distribution of it.

If there's enough demand in the future, I'd be more than happy to distribute those packages in an official capacity.

jjehl:
+1 for demand on arm architecture

Surmeh:
+1 for demand on arm architecture

ttaylorr:
@andyneff what do you think about this?

andyneff:
@ttaylorr I don't actually know a whole lot about arm or have much opinion on the matter.

If its as simple as just calling the distribution scripts with the arm `TARGET_PLATFORM` now, I say go for it!

ttaylorr:
@andyneff it looks a little more complicated than that; I started work in these two places:

- https://github.com/git-lfs/git-lfs/compare/arm-arm64-distributions
- https://github.com/git-lfs/build-dockers/compare/arm-arm64-distributions

I don't have time to finish this right now, so someone else is more than welcome to take a look. Otherwise, I will try and look at this myself before the next release, which would be v2.5.0.

MarekPetr:
Hi @ttaylorr, have you tried running git-lfs on android's arm? I am developing an app using Termux's git-lfs but I can not make ot track files.

bk2204:
Hey,

Is your question about whether the binary can be compiled or, having compiled it, how to get it to work?  If it's the latter, I think it would probably be better to open a new issue using the template so we can track down what's going wrong.

I don't believe any of the core team use Git LFS on Android ourselves, but we're happy to try to help you figure things out.

MarekPetr:
Thanks for a quick answer. That would be awesome, I can totally open an issue :)

PR #4126, opened 2020-05-08
Makefile: add linux arm build and release targets

saracen:
Can we have support for ARM please?

It would close https://github.com/git-lfs/git-lfs/issues/2464

Thank you!


saracen:
@bk2204 I see, thank you. Is doing it like I have in the latest okay?

Issue #3238, opened 2018-09-09
Not all files are migrated with git lfs migrate import


tcamin:
We’re trying to migrate a largish (22k commits) repo to Git LFS, but are facing some issues with some specific files that seems not to migrate:

```
❯ git lfs migrate import --include=Subito/Pods/Optimizely-iOS-SDK/Optimizely.framework/Versions/A/Optimizely --everything --verbose
migrate: Sorting commits: ..., done
migrate: commit 92d98a4fad452573a6773effb781b863f59a4622: Subito/Pods/Optimizely-iOS-SDK/Optimizely.framework/Versions/A/Optimizely
migrate: Rewriting commits: 100% (21195/21195), done
  develop                                                	bf01409b82a35a3bf1c00b13bd49d39b0aaf0e8e -> ccdc1dbfca07daf403fb3c6bcd9281dc1b4b3ad1
....
.... other branches and tags
....
migrate: Updating refs: ..., done
migrate: checkout: ..., done

❯ GIT_LFS_SKIP_SMUDGE=1 git checkout 92d98a4fad452573a6773effb781b863f59a4622
Checking out files: 100% (1795/1795), done.
Note: checking out '92d98a4fad452573a6773effb781b863f59a4622'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at 92d98a4fa adding Optimizely SDK

subito git/92d98a4...
❯ cat Subito/Pods/Optimizely-iOS-SDK/Optimizely.framework/Versions/A/Optimizely | head -n 5
����
    	X�K��L���ޑ���
                     3���!<arch>
#1/20           1474927163  503   20    100644  62236     `
__.SYMDEF�
```

As shown here the files seems to be matched during migration and the history gets indeed rewritten at commit 92d98a4fad452573a6773effb781b863f59a4622 however when I checkout that commit I don't see the LFS pointer, but instead I still see the entire binary data.

If I checkout without the `GIT_LFS_SKIP_SMUDGE=1` I'm getting:

```
Encountered 1 file(s) that should have been pointers, but weren't:
	Subito/Pods/Optimizely-iOS-SDK/Optimizely.framework/Versions/A/Optimizely
```

[...]

PR #4133, opened 2020-05-13
git: consider full refspec when determining seen refs

### LINKED ISSUE #4120: git lfs migrate import --everything doesn't migrate tags whose name matches that of a branch
**Describe the bug**
When migrating a repository to LFS, any tags that have the same name as a branch are not migrated. Presumably this is because somewhere inside LFS tags are referred to in am ambiguous way as `<name>` rather than `refs/tags/<name>`.

**To Reproduce**
```
git init .
git lfs install --local
echo "hello" > hello.txt
git add hello.txt
git commit -m "Add hello.txt"
git branch release_42
git tag release_42
git tag release_other
git lfs migrate import --everything --include="*.txt"
```

**Expected behavior**
Both branches (master and release_42) and both tags (release_42 and release_other) should all point to the same new, migrated commit.

**Observed behaviour**
The two branches and the release_other tag point to the same new, migrated commit. But the release_42 tag still points to the old, unmigrated commit.

**System environment**
Windows 10, Git Bash environment.

**Output of `git lfs env`**
```
git-lfs/2.10.0 (GitHub; windows amd64; go 1.12.7; git a526ba6b)
git version 2.26.2.windows.1

LocalWorkingDir=C:\Data\Code\repro
LocalGitDir=C:\Data\Code\repro\.git
LocalGitStorageDir=C:\Data\Code\repro\.git
LocalMediaDir=C:\Data\Code\repro\.git\lfs\objects
LocalReferenceDirs=
TempDir=C:\Data\Code\repro\.git\lfs\tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=C:\Data\Code\repro\.git\lfs
AccessDownload=none
AccessUpload=none
```

bk2204:
We avoid rewriting references that we've already seen, since in some cases we will need to order references to properly handle tags pointing to tags.  However, we use ref.Name for this, which is the short name, not the full refspec.  Consequently, if we've already seen a ref with the same short name (e.g., a branch when we're processing a tag), we end up skipping the second ref.

This is not what we want, so let's use the full refspec when populating the set of items we've already seen.

Fixes #4120
/cc @karlhendikse as reporter
/cc #3238, which may be fixed by this

### DIFF
--- a/git/githistory/ref_updater.go
+++ b/git/githistory/ref_updater.go
@@ -79,10 +79,11 @@ func (r *refUpdater) updateOneRef(list *tasklog.ListTask, maxNameLen int, seen m
 		return errors.Wrapf(err, "could not decode: %q", ref.Sha)
 	}
 
-	if _, ok := seen[ref.Name]; ok {
+	refspec := ref.Refspec()
+	if _, ok := seen[refspec]; ok {
 		return nil
 	}
-	seen[ref.Name] = struct{}{}
+	seen[refspec] = struct{}{}
 
 	to, ok := r.CacheFn(sha1)
 
--- a/t/t-migrate-import.sh
+++ b/t/t-migrate-import.sh
@@ -499,6 +499,32 @@ begin_test "migrate import (identical contents, different permissions)"
 )
 end_test
 
+begin_test "migrate import (tags with same name as branches)"
+(
+  set -e
+
+  setup_multiple_local_branches
+  git checkout master
+
+  contents="hello"
+  oid=$(calc_oid "$contents")
+  printf "$contents" >hello.dat
+  git add .
+  git commit -m "add file"
+
+  git branch foo
+  git tag foo
+  git tag bar
+
+  git lfs migrate import --everything --include="*.dat"
+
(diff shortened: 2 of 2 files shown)


git/githistory/rewriter.go
```diff
@@ -554,6 +554,7 @@ func (r *Rewriter) refsToMigrate() ([]*git.Ref, error) {
 		}
 
 		local = append(local, ref)
+		fmt.Fprintf(os.Stderr, "dx: b: to migrate=%q\n", ref.Refspec())
```
chrisd8088, line 557: Is this deliberate, or leftover from debugging, by any chance?

bk2204: That is indeed debugging output.  I'll strip it out.

PR #4143, opened 2020-05-27
Sign and notarize binaries on macOS

### LINKED ISSUE #3714: developer signature problem in macos (catalina)
I'd like to use git-lfs in my macos catalina machine. I'm aware that catalina is in beta release, but still I wonder what will happen to git-lfs in a few months (?) when catalina is released.

My OS can find the existing git-lfs, i.e.
```
$ which git-lfs
/usr/local/bin/git-lfs
```
but when I try to run it (from the terminal, or in other ways), I get an error message stating that 
```
"git-lfs" can't be opened because Apple cannot check it for malicious software
```

I know that I can get around that with an `xattr -d com.apple.quarantine` command, but I sort of like how apple is being cautious.

I wonder ... is there a chance the binaries could be signed in the future, to avoid this problem?

Again, to emphasize, I'm using Catalina, so it's not what most mac people will have, for a few months at least.

Thanks!!

bk2204:
Apple complains if a binary is downloaded from the Internet but has not been signed and notarized as of macOS Catalina.  Because this is an impediment for users, let's fix that.

This series contains several different patches:
1. We now build with Go 1.14 in CI in addition to other versions.
2. We no longer build i386 Darwin binaries since we can't notarize those.
3. We build the Darwin release assets as a zip file, since we can't notarize a tarball.
4. We build Darwin binaries on Darwin instead of Linux for releases.
5. We sign and notarize binaries.

Each patch contains a description of what it does and why it's valuable.

Currently we use my (GitHub) Apple ID as the actor for notarization plus an app password.  This value will be stored in the secret store for Git LFS so it can be changed out without further code changes if I should be hit by a bus (or just move on to other projects).

This has been tested in a private repo in the git-lfs organization and appears to work there.

Fixes #3714.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -168,7 +168,6 @@ BUILD = GOOS=$(1) GOARCH=$(2) \
 # built for.
 BUILD_TARGETS = \
 	bin/git-lfs-darwin-amd64 \
-	bin/git-lfs-darwin-386 \
 	bin/git-lfs-linux-arm \
 	bin/git-lfs-linux-arm64 \
 	bin/git-lfs-linux-amd64 \
@@ -206,8 +205,6 @@ all build : $(BUILD_TARGETS)
 # embeds the versioninfo into the binary.
 bin/git-lfs-darwin-amd64 : $(SOURCES) mangen
 	$(call BUILD,darwin,amd64,-darwin-amd64)
-bin/git-lfs-darwin-386 : $(SOURCES) mangen
-	$(call BUILD,darwin,386,-darwin-386)
 bin/git-lfs-linux-arm : $(SOURCES) mangen
 	GOARM=5 $(call BUILD,linux,arm,-linux-arm)
 bin/git-lfs-linux-arm64 : $(SOURCES) mangen
@@ -272,7 +269,6 @@ script/windows-installer/git-lfs-wizard-image.bmp
 # 	make VERSION=my-version bin/releases/git-lfs-darwin-amd64-my-version.tar.gz
 RELEASE_TARGETS = \
 	bin/releases/git-lfs-darwin-amd64-$(VERSION).tar.gz \
-	bin/releases/git-lfs-darwin-386-$(VERSION).tar.gz \
 	bin/releases/git-lfs-linux-arm-$(VERSION).tar.gz \
 	bin/releases/git-lfs-linux-arm64-$(VERSION).tar.gz \
 	bin/releases/git-lfs-linux-amd64-$(VERSION).tar.gz \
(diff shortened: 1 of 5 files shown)


bk2204:
I also snuck a general CI fix in here (the `docs/man/mangen` patch) since this PR fails without it and I'm already touching the CI system all over the place in this PR.

Makefile
```diff
@@ -384,12 +391,53 @@ release-windows-rebuild: bin/releases/git-lfs-windows-assets-$(VERSION).tar.gz
...
+
 .PHONY : release-write-certificate
 release-write-certificate:
 	@echo "Writing certificate to $(CERT_FILE)"
 	@echo "$$CERT_CONTENTS" | base64 --decode >"$$CERT_FILE"
 	@printf 'Wrote %d bytes (SHA256 %s) to certificate file\n' $$(wc -c <"$$CERT_FILE") $$(shasum -ba 256 "$$CERT_FILE" | cut -d' ' -f1)
 
+# release-import-certificate imports the given certificate into the macOS
+# keychain "CI".  It is not generally recommended to run this on a user system,
+# since it creates a new keychain and modifies the keychain search path.
```
chrisd8088, line 426: I don't suppose there's any way to have this target's build commands start by checking that they are running as an Action CI-specific user, and pausing for user confirmation if not?

bk2204: I don't think we can do an Actions-specific user since there's no guarantee that won't change, but we can certainly check for an Actions-specific environment variable and fail otherwise.  Thanks for the idea.

bk2204: I've checked for the `CI` environment variable and gently chided the user if it's not set.

MarkCallow:
Please make a release containing this fix. I just installed 2.11 thinking it had this fix and what a right royal pain it was.

Even if there are no other features ready, this by itself is enough to justify a release.

bk2204:
I plan to do a release this month.  I appreciate it's inconvenient to have to deal with the unnotarized binaries.  In the mean time, you may prefer using Homebrew.

cliffhall:
@bk2204 How is that release coming along? I notice there is a 2.12 release out, but the download page still only shows 2.11.

bk2204:
The release is out.  I've pushed a change to update the website; it should take effect in a few minutes.  Thanks for the reminder.

cliffhall:
@bk2204 Looks like the download link is broken:
Screen Shot 2020-09-09 at 12 02 24 PM

bk2204:
Yeah, we now distribute zip files instead of tarballs because tarballs can't be notarized.  I've pushed a change to update that.

PR #4157, opened 2020-06-12
Makefile: ensure temp Go modules can be deleted

chrisd8088:
As a followup to commit 1d4e46b311a3 in PR #4132, we ensure any files created during the `make test` build can be fully deleted by setting the user write permission bit.  Otherwise, at least on macOS, some Go modules' files are installed in directories with read-only permissions, and the final `rm -fr` cleanup is then unable to remove those files, leading to lots of spurious error messages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -473,6 +473,7 @@ test : fmt $(.DEFAULT_GOAL)
 		export GIT_CONFIG_NOSYSTEM=1; \
 		$(GO) test -count=1 $(GO_TEST_EXTRA_ARGS) $(addprefix ./,$(PKGS)); \
 		RET=$$?; \
+		chmod -R u+w "$$tempdir"; \
 		rm -fr "$$tempdir"; \
 		exit $$RET; \
 	)


Makefile
```diff
@@ -473,6 +473,7 @@ test : fmt $(.DEFAULT_GOAL)
 		export GIT_CONFIG_NOSYSTEM=1; \
 		$(GO) test -count=1 $(GO_TEST_EXTRA_ARGS) $(addprefix ./,$(PKGS)); \
 		RET=$$?; \
+		chmod -R u+w "$$tempdir"; \
```
bk2204, line 476: This seems like most limited solution we can apply.  I've seen security problems from loosening permissions before deleting, but I don't think we have that here because it only applies to the user and not anyone else.

Do we know how this works on Windows?

chrisd8088: Hmm, Windows ... good question.  The Windows CI jobs seem to have passed, and I see these commands in the log output ahead of successful Go tests (e.g., in the ` Run GOPATH="$HOME/go" PATH="$HOME/go/bin:$PATH" script/cibuild` section of the `Build on Windows` job [output](https://github.com/git-lfs/git-lfs/pull/4157/checks?check_run_id=766167307)).  That's running under Mingw-w64, I think?

bk2204: If the answer is, “It doesn't break the testsuite and is no worse off than before,” which it looks like it is, then I'm fine with that.

chrisd8088: Cool.  And at least on some CI jobs, I see this is failing as I experienced it locally; see this recent job [output](https://github.com/git-lfs/git-lfs/runs/766491205), for instance.  So I think it'll be helpful to merge this, which I'll do in a moment.

bk2204: I'm very curious as to why it's failing now and wasn't before.  I guess that's the question I always ask about our CI jobs, though, so :man_shrugging:.

chrisd8088: Yeah, dunno.  I suppose the change could be upstream somewhere in the Go ecosystem, since we install a bunch of Go modules into this temporary directory, and the troublesome permissions are on certain directories in that hierarchy, e.g., `.../go/pkg/mod/gopkg.in/yaml.v2@v2.2.8` and so forth.

PR #4171, opened 2020-06-25
Don't mark unlocked files that aren't lockable as read-only

### LINKED ISSUE #4169: Unlocking a file that is not `lockable` should not make the file read only
**Describe the bug**

If a file is locked (doesn't need to be tracked by LFS) and is not `lockable`, when unlocking the file it will be marked read only until the user manually corrects the permissions.

**To Reproduce**

```
# Setup a clean repo
mkdir lfs-repro
cd lfs-repro
git init
echo "# LFS test" > README.md
git add README.md
git commit -m "Init"

# Add a remote that supports LFS locking
git remote add origin https://github.com/jamesramsay/lfs-repro.git

# Lock and unlock
git lfs install
git lfs lock README.md
git lfs unlock README.md

# Observe read-only permissions
ls -la README.md
```

As far as I can tell, Git LFS never changes the permissions again, so they are stuck as read only until manually resolved.

**Expected behavior**

Since the file is not marked `lockable`, the file should not be set to read only when unlocked.

**System environment**

macOS 10.15.5

<details><summary>**Output of `git lfs env`**</summary>

```
git-lfs/2.11.0 (GitHub; darwin amd64; go 1.14.2)
git version 2.27.0

Endpoint=https://github.com/jamesramsay/lfs-repro.git/info/lfs (auth=none)
  SSH=git@github.com:jamesramsay/lfs-repro.git
LocalWorkingDir=/Users/jramsay/GitLab/temp/lfs-repro
LocalGitDir=/Users/jramsay/GitLab/temp/lfs-repro/.git
LocalGitStorageDir=/Users/jramsay/GitLab/temp/lfs-repro/.git
LocalMediaDir=/Users/jramsay/GitLab/temp/lfs-repro/.git/lfs/objects
LocalReferenceDirs=
TempDir=/Users/jramsay/GitLab/temp/lfs-repro/.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/Users/jramsay/GitLab/temp/lfs-repro/.git/lfs
AccessDownload=none
AccessUpload=none
GIT_EXEC_PATH=/usr/local/Cellar/git/2.27.0/libexec/git-core
GIT_TRACE2_PARENT_SID=20200625T025724.504881Z-H0bad8183-P00002d11

```

</details>

bk2204:
Whenever we use a file path filter and there are no included patterns, we default to assuming the files match.  For many cases, this is fine, but this is not useful for checking whether a file is lockable: if no files are listed as lockable, then no files are lockable.

Let's add an option for our filters to specify the default value if no patterns match.  If that default is false, like it should be for locking, then skip traversing the exclude patterns altogether, since anything that's not included cannot change the result.  We test both the case that lockable patterns are present and absent in our test.

Fixes #4169

### DIFF
--- a/filepathfilter/filepathfilter.go
+++ b/filepathfilter/filepathfilter.go
@@ -16,18 +16,37 @@ type Pattern interface {
 }
 
 type Filter struct {
-	include []Pattern
-	exclude []Pattern
+	include      []Pattern
+	exclude      []Pattern
+	defaultValue bool
 }
 
-func NewFromPatterns(include, exclude []Pattern) *Filter {
-	return &Filter{include: include, exclude: exclude}
+type options struct {
+	defaultValue bool
 }
 
-func New(include, exclude []string) *Filter {
+type option func(*options)
+
+// DefaultValue is an option representing the default value of a filepathfilter
+// if no patterns match.  If this option is not provided, the default is true.
+func DefaultValue(val bool) option {
+	return func(args *options) {
+		args.defaultValue = val
+	}
+}
+
+func NewFromPatterns(include, exclude []Pattern, setters ...option) *Filter {
+	args := &options{defaultValue: true}
+	for _, setter := range setters {
+		setter(args)
+	}
+	return &Filter{include: include, exclude: exclude, defaultValue: args.defaultValue}
+}
+
+func New(include, exclude []string, setters ...option) *Filter {
 	return NewFromPatterns(
 		convertToWildmatch(include),
-		convertToWildmatch(exclude))
+		convertToWildmatch(exclude), setters...)
 }
 
 // Include returns the result of calling String() on each Pattern in the
@@ -66,6 +85,15 @@ func (f *Filter) Allows(filename string) bool {
 		return false
 	}
 
+	// Beyond this point, the only values we can logically return are false
+	// or the default value.  If the default is false, then there's no point
+	// traversing the exclude patterns because the return value will always
+	// be false; we'd do extra work for no functional benefit.
+	if !included && !f.defaultValue {
+		tracerx.Printf("filepathfilter: rejecting %q", filename)
+		return false
+	}
+
 	for _, ex := range f.exclude {
 		if ex.Match(filename) {
 			tracerx.Printf("filepathfilter: rejecting %q via %q", filename, ex.String())
@@ -73,6 +101,7 @@ func (f *Filter) Allows(filename string) bool {
 		}
 	}
 
+	// No patterns matched and our default value is true.
 	tracerx.Printf("filepathfilter: accepting %q", filename)
 	return true
 }
--- a/locking/lockable.go
+++ b/locking/lockable.go
@@ -47,7 +47,7 @@ func (c *Client) refreshLockablePatterns() {
 			c.lockablePatterns = append(c.lockablePatterns, filepath.ToSlash(p.Path))
 		}
 	}
-	c.lockableFilter = filepathfilter.New(c.lockablePatterns, nil)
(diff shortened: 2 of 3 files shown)


filepathfilter/filepathfilter.go
```diff
@@ -16,18 +16,35 @@ type Pattern interface {
...
-func New(include, exclude []string) *Filter {
+type option func(*options)
+
+func DefaultValue(val bool) option {
+	return func(args *options) {
+		args.defaultValue = val
+	}
+}
+
+func NewFromPatterns(include, exclude []Pattern, setters ...option) *Filter {
```
chrisd8088, line 36: I like that you've made this more generic than just adding a simple `bool` flag parameter.

filepathfilter/filepathfilter.go
```diff
@@ -61,7 +78,7 @@ func (f *Filter) Allows(filename string) bool {
 		}
 	}
 
-	if !included && len(f.include) > 0 {
+	if !included && (len(f.include) > 0 || !f.defaultValue) {
 		tracerx.Printf("filepathfilter: rejecting %q via %v", filename, f.include)
```
chrisd8088, line 82: If `len(f.include) == 0` but `f.defaultValue` is `false`, then we'll return `false` here, which is intended, obviously, but the trace message may be confusing as `f.include` will be empty.  I might recommend adding some additional logic here to output a more meaningful trace message when we're rejecting the match because we fell through to the filter's default value.

filepathfilter/filepathfilter.go
```diff
@@ -16,18 +16,35 @@ type Pattern interface {
...
-func NewFromPatterns(include, exclude []Pattern) *Filter {
-	return &Filter{include: include, exclude: exclude}
+type options struct {
+	defaultValue bool
 }
 
-func New(include, exclude []string) *Filter {
+type option func(*options)
+
+func DefaultValue(val bool) option {
```
chrisd8088, line 30: My only hesitation here is that `DefaultValue` is something of a generic phrase and, while I have no good alternatives to suggest, I wondered if we could add a comment describing it's function, at least.

I think my hesitation stems from the fact that, IIUC, this isn't so much a default in the sense of "if no patterns match, then do X" but rather "if no **_include_** patterns are **_defined_**, then do X".  So it feels more specific than a generic default ... maybe there's a way to express that in a different name, like `NoIncludeDefaultValue`?  Sheesh, naming things is hard.  :-(

filepathfilter/filepathfilter.go
bk2204: Besides the design benefits, this also has the benefit of not requiring us to retrofit every other caller in the codebase.

filepathfilter/filepathfilter.go
bk2204: True, we only consider include patterns, but this is an optimization: if we know that the filter has only excludes left, the only thing we can return is `false` or the default value.  If the default value is `false`, then we need not consider the excludes at all, since we know what the return value is.  The optimization isn't required for correctness, but it's clear to me that we need multiple additional comments here, since this is tricky to reason about.

bk2204: I've added a comment here explaining what `DefaultValue` does and explained the logic here much better.

chrisd8088: Thanks, and I think I noticed the logic changed too to make it more clearer as well!  ❤️

PR #4175, opened 2020-06-30
subprocess: trace all command execution

bluekeyes:
Trace all command execution at the time when the command starts. Previously, only "simple" execution created trace entries, leading to mysterious gaps in trace logs where time passed but it looked like nothing was happening.

An alternate approach might be to log the trace line when the `subprocess.Cmd` is created. in `newCmd`. If commands are always executed right after creation, that might be enough and it avoids overriding the additional methods from `os/exec.Cmd`.

### DIFF
--- a/git/config.go
+++ b/git/config.go
@@ -175,7 +175,6 @@ func (c *Configuration) Source() (*ConfigurationSource, error) {
 
 func (c *Configuration) gitConfig(args ...string) (string, error) {
 	args = append([]string{"config"}, args...)
-	subprocess.Trace("git", args...)
 	cmd := subprocess.ExecCommand("git", args...)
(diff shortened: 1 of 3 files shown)


bk2204:
This seems fine to me; I'll wait to formally approve it until #4177 lands, at which point I'll ask you to rebase it on top of the latest HEAD so we can be sure that CI passes.

bluekeyes:
Thanks for fixing the worktree issue, I wasn't sure why that was failing. I rebased on top of #4177, but it looks like there's a new failure that seems unrelated. Should I rebase again to rerun?

bk2204:
I think that's just a flaky test that we have an issue for somewhere.  I've kicked off a re-run of the job, and it seems to be fixed.

PR #4176, opened 2020-06-30
git: improve performance of remote ref listing

bluekeyes:
When repositories have a large number (hundreds of thousands) of refs, Git LFS adds noticeable overhead on the time taken by `git show-ref` and `git ls-remote`. To reduce this overhead:

* Avoid regular expressions when parsing lines of Git command output. Regexp evaluation was the most expensive part of this process in profiling.
* Minimize the number of loops when computing skipped refs by combining the missing and skipped checks and converting the list of remote refs into a pre-filtered set.

In a repository with ~300k remote references, these changes reduce the time spent in the `pre-push` command from 8.26s to 6.31s. The individual optimizations were identified and verified with pprof.

I'm happy to break these up into two separate PRs or commits if that's preferred.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -1086,20 +1086,30 @@ func CachedRemoteRefs(remoteName string) ([]*Ref, error) {
 	cmd.Start()
 	scanner := bufio.NewScanner(outp)
 
-	r := regexp.MustCompile(fmt.Sprintf(`([0-9a-fA-F]{40})\s+refs/remotes/%v/(.*)`, remoteName))
+	refPrefix := fmt.Sprintf("refs/remotes/%v/", remoteName)
 	for scanner.Scan() {
-		if match := r.FindStringSubmatch(scanner.Text()); match != nil {
-			name := strings.TrimSpace(match[2])
+		if sha, name, ok := parseShowRefLine(refPrefix, scanner.Text()); ok {
 			// Don't match head
 			if name == "HEAD" {
 				continue
 			}
-
-			sha := match[1]
 			ret = append(ret, &Ref{name, RefTypeRemoteBranch, sha})
 		}
 	}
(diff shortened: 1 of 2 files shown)


git/git.go
```diff
@@ -1086,22 +1086,29 @@ func CachedRemoteRefs(remoteName string) ([]*Ref, error) {
...
 			ret = append(ret, &Ref{name, RefTypeRemoteBranch, sha})
 		}
 	}
 	return ret, cmd.Wait()
 }
 
+func parseShowRefLine(refPrefix, line string) (sha, name string, ok bool) {
+	const shaLen = 40
+
+	refIdx := strings.Index(line, refPrefix)
```
bluekeyes, line 1105: I used `strings.Index` here because the original regexp allowed one or more spaces after the SHA. However, the output from `show-ref` and `ls-remote` seems well-defined, with a single space or a single tab respectively.

If supporting variations on the format isn't necessary, these two new parse functions might be simplified by looking for the ref at a specific offset in the string.

bk2204: I think the way I'd like to go here is to search for the first space or tab, as appropriate, and not use a hard-coded offset.  Git will learn to support SHA-256, which has a longer length, very soon (within the next two releases) and I have a series for Git LFS which removes all of the hard-coded `40` constants here, as well as updating regexes to support it here as well.  I'd like to avoid regressing that series if possible, so adding more assumptions about the length of the hash should be avoided.

It's probably safe to rely on the output of Git not changing to add additional spaces or tabs that aren't in the current version.

bluekeyes: That makes sense, I updated both functions to look for the space/tab separator.

bk2204:
Thanks for the patch and the prompt turnaround on revising it!

PR #4182, opened 2020-07-05
update go.mod to reflect Go 1.12 minimum

### LINKED ISSUE #4179: go.mod requires go1.11, but os.ProcessState.ExitCode requires go1.12
**Describe the bug**
When building git-lfs from source with go1.11, the build fails because os.ProcessState.ExitCode is undefined (see https://github.com/git-lfs/git-lfs/issues/3945). Go added ExitCode to os.ProcessState in go1.12 ([release notes](https://golang.org/doc/go1.12#os)), however the git-lfs go.mod states a minimum requirement of go1.11.

Listing go1.11 is misleading and should be changed to 1.12. Additionally, it would be nice to note a minimum requirement on go1.12 in the README notes for installing [from source](https://github.com/git-lfs/git-lfs#from-source).

**To Reproduce**
1. Install go1.11
2. Build git-lfs
3. Observe build failures as described in https://github.com/git-lfs/git-lfs/issues/3945

**Expected behavior**
- go.mod should depend on `go 1.12` instead of `go 1.11`.
- README should note go1.12 minimum requirement in section on installing from source.

**System environment**
- FreeBSD 8, amd64
- `go version go1.11.4 freebsd/amd64`

**Output of `git lfs env`**
git-lfs did not build, thus this information is not available.

**Additional context**
We bootstrapped a patched go1.11.4 from source as Go no longer supports FreeBSD 8. Had it been known that git-lfs requires go1.12, we would have bootstrapped go1.12.17 instead.

Alternatively, if you want to continue supporting go1.11, [git/git.go:688](https://github.com/git-lfs/git-lfs/blob/43a83115f3d5b13e8432ccb9d7ac0d1880ff6d46/git/git.go#L688) could be changed to remove the call to `ExitCode`:
```diff
-if e, ok := err.(*exec.ExitError); ok && e.ProcessState.ExitCode() == 128 {
+if e, ok := err.(*exec.ExitError); ok && e.ProcessState.Sys().(syscall.WaitStatus).ExitStatus() == 128 {
```

chrisd8088:
Bump the minimum Go version specified in our `go.mod` to match our CI tooling and our dependency on some Go 1.12 features, specifically our use of `os.ProcessState.ExitCode()`.

Note that an alternative is proposed in #4183 and we should adopt either that one or this, but not both.

Fixes #4179.
/cc @andrewarchi as reporter.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,4 +29,4 @@ require (
 	gopkg.in/yaml.v2 v2.2.8 // indirect
 )
 
-go 1.11
+go 1.12


chrisd8088:
Closing in favour of #4183.

PR #4200, opened 2020-07-28
config: optionally read .lfsconfig from the repository

### LINKED ISSUE #4077: Read .lfsconfig from repository if absent from working tree
**Describe the bug**

`git clone` doesn't honor the `lfs.url` setting from `.lfsconfig` for some files. By the looks of it the problem is with files that are sorted alphabetically prior to `.lfsconfig` itself.

**To Reproduce**
Steps to reproduce the behavior, let's say with a GitHub repo but requiring also a different LFS server:
1. Create a new repo in GitHub with a `.lfsconfig` pointing to a different Git LFS server
2. Create and push an LFS-tracked file called `.abigfile`: `git lfs track .abigfile && echo "hello world" > .abigfile && git add -A && git commit -m test && git push`
3. `git clone git@github.com:your-repo` and observe the error:

```
[..]
Receiving objects: 100% (5/5), done.
Downloading .abigfile (12 B)
Error downloading object: .abigfile (a948904): Smudge error: Error downloading .abigfile (a948904f2f0f479b8f8197694b30184b0d2ed1c1cd2a1ec0fb85d299a192a447): [a948904f2f0f479b8f8197694b30184b0d2ed1c1cd2a1ec0fb85d299a192a447] Object does not exist on the server: [404] Object does not exist on the server
[..]
```

4. Verify that the file is actually available: `git lfs pull && git checkout -f HEAD && cat .abigfile` produces "hello world" as expected
5. Move it to a different path (remove leading `.`): `git mv .abigfile abigfile && git lfs untrack .abigfile && git lfs track abigfile && git commit -am test2 && git push`
6. `git clone git@github.com:your-repo` now reports no errors and `cat abigfile` produces "hello world" as expected

Using `GIT_TRACE=1` shows that in step 3 it's trying to get it from `lfs.github.com` instead of the URL specified in `.lfsconfig`, whereas in step 6 it uses the correct URL.

The `GIT_TRACE=1` outputs diverge here:

```
[..]
Receiving objects: 100% (8/8), done.
09:39:22.258562 run-command.c:663       trace: run_command: git rev-list --objects --stdin --not --all --quiet --alternate-refs '--progress=Checking connectivity'
09:39:22.263653 git.c:439               trace: built-in: git rev-list --objects --stdin --not --all --quiet --alternate-refs '--progress=Checking connectivity'
09:39:22.298787 run-command.c:663       trace: run_command: 'git-lfs filter-process'
07:39:22.336627 trace git-lfs: exec: git 'version'
07:39:22.347550 trace git-lfs: exec: git 'config' '-l'
07:39:22.353481 trace git-lfs: exec: git 'config' '-l' '-f' '/tmp/test-lfs-2/.lfsconfig'
[..]
```

The last line is not present when the LFS file is called `.abigfile`.

**Expected behavior**
The URL specified in `.lfsconfig` should always be used when fetching LFS files.

**System environment**
Arch Linux but I don't think it's relevant. I don't think I have any configurations that could affect this.

**Output of `git lfs env`**
From within one of the test repos:
```
git-lfs/2.10.0 (GitHub; linux amd64; go 1.14)
git version 2.25.2

Endpoint=<the expected one>
LocalWorkingDir=/tmp/test-lfs
LocalGitDir=/tmp/test-lfs/.git
LocalGitStorageDir=/tmp/test-lfs/.git
LocalMediaDir=/tmp/test-lfs/.git/lfs/objects
LocalReferenceDirs=
TempDir=/tmp/test-lfs/.git/lfs/tmp
ConcurrentTransfers=32
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/tmp/test-lfs/.git/lfs
AccessDownload=basic
AccessUpload=basic
```

bk2204:
Currently we only read .lfsconfig from the working tree.  This is better than nothing, but it means that if there's an LFS file that starts with a name earlier than .lfsconfig, we won't read the proper config file, and we won't use any LFS remote URLs that are located in that file.

If the file is missing, let's additionally read from the index, if there's a working tree, and then from HEAD.  If the repository is bare, let's just read from HEAD.  This is very similar to what Git does for .gitmodules.

It does, however, differ in one significant way: Git will read from the index first if it's doing a checkout, since that might have newer information.  We don't do that here, since it's not totally clear that we can articulate all the cases where that occurs in Git LFS, but we explicitly allow for the possibility of changing the behavior in the future and document the feature accordingly.

We no longer read the file .lfsconfig from a bare repository, but that we did this in the first place was very questionable and pretty clearly a bug, so it's intentional that we no longer do.

It's intentional that we ignore errors from this code path, since Git doesn't distinguish well between older versions which don't support the `--blob` flag to `git config` (those before 1.8.4), cases where the revision is missing, and actual errors.  We don't want to do a `git rev-parse` on each revision, since that would add potentially two more Git calls to every invocation of Git LFS.

Fixes #4077

### DIFF
--- a/git/config.go
+++ b/git/config.go
@@ -2,6 +2,7 @@ package git
 
 import (
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -134,15 +135,31 @@ func (c *Configuration) UnsetLocalKey(key string) (string, error) {
 	return c.gitConfigWrite("--unset", key)
 }
 
-func (c *Configuration) Sources(optionalFilename string) ([]*ConfigurationSource, error) {
+func (c *Configuration) Sources(dir string, optionalFilename string) ([]*ConfigurationSource, error) {
 	gitconfig, err := c.Source()
 	if err != nil {
 		return nil, err
 	}
 
-	fileconfig, err := c.FileSource(optionalFilename)
-	if err != nil && !os.IsNotExist(err) {
+	bare, err := IsBare()
+	if err != nil {
 		return nil, err
+	}
+
+	// First try to read from the working directory and then the index if
+	// the file is missing from the working directory.
+	var fileconfig *ConfigurationSource
+	if !bare {
+		fileconfig, err = c.FileSource(filepath.Join(dir, optionalFilename))
+		if err != nil {
+			if !os.IsNotExist(err) {
+				return nil, err
+			}
+			fileconfig, _ = c.RevisionSource(fmt.Sprintf(":%s", optionalFilename))
+		}
+	}
+	if fileconfig == nil {
+		fileconfig, _ = c.RevisionSource(fmt.Sprintf("HEAD:%s", optionalFilename))
 	}
 
 	configs := make([]*ConfigurationSource, 0, 2)
@@ -159,6 +176,14 @@ func (c *Configuration) FileSource(filename string) (*ConfigurationSource, error
 	}
 
 	out, err := c.gitConfig("-l", "-f", filename)
+	if err != nil {
+		return nil, err
+	}
+	return ParseConfigLines(out, true), nil
+}
+
+func (c *Configuration) RevisionSource(revision string) (*ConfigurationSource, error) {
+	out, err := c.gitConfig("-l", "--blob", revision)
 	if err != nil {
 		return nil, err
 	}
--- a/t/t-config.sh
+++ b/t/t-config.sh
@@ -42,6 +42,80 @@ begin_test "default config"
 )
 end_test
 
+begin_test "config reads from repository"
+(
+  set -e
+  reponame="repository-config"
+  setup_remote_repo "$reponame"
+  mkdir $reponame
+  cd $reponame
+  git init
+  git remote add origin "$GITSERVER/$reponame"
+  git lfs env | tee env.log
+  grep "Endpoint=$GITSERVER/$reponame.git/info/lfs (auth=none)" env.log
+
+  git config --file=.lfsconfig lfs.url http://lfsconfig-file
+  git config --file=.lfsconfig lfs.http://lfsconfig-file.access lfsconfig
+  git add .lfsconfig
+  git commit -m 'Add file'
+  git push origin HEAD
+
+  git checkout -b side
+  git config --file=.lfsconfig lfs.url http://lfsconfig-file-side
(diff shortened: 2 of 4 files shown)


git/config.go
```diff
@@ -134,17 +135,33 @@ func (c *Configuration) UnsetLocalKey(key string) (string, error) {
...
+	// First try to read from the working directory and then the index if
+	// the file is missing from the working directory.
+	var fileconfig *ConfigurationSource
+	if !bare {
+		fileconfig, err = c.FileSource(filepath.Join(dir, optionalFilename))
+		if err != nil {
+			if !os.IsNotExist(err) {
+				return nil, err
+			}
+			fileconfig, _ = c.RevisionSource(fmt.Sprintf(":%s", optionalFilename))
```
chrisd8088, line 158: Just as a question, is it the case that there are not interesting errors to report from trying to run `git config ...` here?  In the `c.FileSource()` call above it looks like `c.gitConfig()` is called and any errors from it get returned as long as they aren't `os.IsNotExist()`, but similar returns from `c.RevisionSource()` look to be ignored here and below in the `HEAD` call.

chrisd8088: Ah, I see you anticipated my question in the PR description!
> It's intentional that we ignore errors from this code path, since Git doesn't distinguish well between older versions which don't support the `--blob` flag to `git config` (those before 1.8.4), cases where the revision is missing, and actual errors.

t/t-config.sh
```diff
@@ -42,6 +42,81 @@ begin_test "default config"
...
+  git remote add origin "$GITSERVER/$reponame"
+
+  git lfs track "*.bin"
+  git config --file=.lfsconfig lfs.url "$GITSERVER/$reponame.git/info/lfs"
+
+  echo "abc" > .bin
+  echo "def" > a.bin
+
+  git add .
+  git ls-files
```
chrisd8088, line 107: Just for my own understanding, how does the `git ls-files` work in this test?  Would it fail with an exit code under some circumstances if the new functionality wasn't working as expected?

bk2204: I believe that's a leftover debugging statement, so I'll remove it.

PR #4209, opened 2020-08-07
prune: fix deleting objects referred to by stashes

### LINKED ISSUE #4206: git lfs prune should not delete objects referenced by stashes
Currently running "git lfs prune" on objects which are only referenced in stashed changes will result in those objects being deleted, and therefore the stashed changes can never be re-applied. 

**To Reproduce**
* Make a change to an existing, tracked LFS file
* `git stash push`
* `git lfs prune`
* `git stash pop`

The pop will fail because the LFS content is missing (it will try to download it, and of course fail). If you ran `git stash pop` before

**Expected behavior**
`git lfs prune` should not delete the object(s) because they are still referenced by a stash. Stashes should be included in the list of referenced local commits alongside branches.

**System environment**
Windows 10

**Output of `git lfs env`**
```
git-lfs/2.11.0 (GitHub; windows amd64; go 1.14.2; git 48b28d97)
git version 2.27.0.windows.1

Endpoint=https://Foo.git/info/lfs (auth=none)
  SSH=git@Foo.git
LocalWorkingDir=C:\Users\steve\...
LocalGitDir=C:\Users\steve\...\.git
LocalGitStorageDir=C:\Users\steve\...\.git
LocalMediaDir=C:\Users\steve\...\.git\lfs\objects
LocalReferenceDirs=
TempDir=C:\Users\steve\...\.git\lfs\tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=C:\Users\steve\...\.git\lfs
AccessDownload=none
AccessUpload=none
DownloadTransfers=basic,bitbucket-media-api,lfs-standalone-file
UploadTransfers=basic,bitbucket-media-api,lfs-standalone-file
GIT_SSH=C:\Program Files\PuTTY\plink.exe
git config filter.lfs.clean = "git-lfs clean -- %f"```
```

sinbad:
It was impossible to pop a stash with LFS data successfully
after running any lfs prune command before this fix, because
prune would consider stashed data unreferenced.

This fixes #4206


sinbad:
FYI: I wrote the `git lfs prune` command originally so this is my bad, hence why I'm fixing it 😃

chrisd8088:
Hey, thanks for both the bug report and the proposed patch!  I was intending to look at the bug report today, and am delighted to have a potential solution in hand before I even do so!  😄 

On a first glance this seems reasonable, but to get the CI jobs to pass I think we'll need to fix the trailing whitespace issue in `t-prune.sh` (I believe there's a missing final newline?)

Thanks again for the patch; that's always appreciated!

sinbad:
Ah yes, sorry I ran the tests individually locally rather than the full CI build, I forgot to check the trailing new line. Due to time zone I won’t be able to fix this until tomorrow but feel free to patch it up yourself if you want / have time. :)

sinbad:
Might need some help with this, I've tried to please the cibuild god but it keeps complaining about whitespace; I've added a couple of newlines but no luck. Its been a few years since I've had to bash my head against it so maybe there's a trick I've forgotten :)

lfs/gitscanner_log.go
```diff
@@ -63,6 +64,32 @@ func scanUnpushed(cb GitScannerFoundPointer, remote string) error {
...
+	scanner := bufio.NewScanner(cmd.Stdout)
+
+	for scanner.Scan() {
+		err = s.ScanRef(strings.TrimSpace(scanner.Text()), cb)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
```
chrisd8088, line 89: After a bit of a deeper look, I had a few notes on this `scanStashed()` function.

The first thing which puzzled me for a while was that it seemed to succeed in a freshly initialized, empty repo with no stashes, where `.git/refs/stash` doesn't exist yet.  I couldn't get `git reflog show stash` to work on the command line, but it seemed to be OK when run this way.  But the immediate issue is that the possible error exit code returned by `cmd.Wait()` is not being checked.  If you add to the end of this function something like:
```go
        stderr, _ := ioutil.ReadAll(cmd.Stderr)
        err = cmd.Wait()
        if err != nil {
                return fmt.Errorf("error in git reflog: %v %v", err, string(stderr))
        }
```

then it does fail in a repo where there's no `refs/stash` ref, or if you change `"stash"` to `"foobar"` or something.  So that's the first issue, that the Git command's exit code is not checked.

Looking at how the `git stash list` command works, I believe it starts by [checking](https://github.com/git/git/blob/dc04167d378fb29d30e1647ff6ff51dd182bc9a3/builtin/stash.c#L691-L692) for an extant `refs/stash` ref, and exiting early if it doesn't exist.

So we probably need something similar here once we're checking the Git command's exit code; otherwise we'll bail with `Prune error: error in git reflog: exit status 128 fatal: ambiguous argument 'stash': unknown revision or path not in the working tree.`  (That's what I get once I added the check above on `cmd.Wait()`'s return value, plus failures in the `t/t-prune.sh` and other tests.)

Beyond that, I looking again at how `git stash list` [uses](https://github.com/git/git/blob/dc04167d378fb29d30e1647ff6ff51dd182bc9a3/builtin/stash.c#L695-L696) the `git log -g refs/stash --` command, I wondered if we could avoid the need for a new `git.RefLog()` function in `git/git.go` entirely and just use `git.Log()`?

git/git.go
```diff
@@ -269,6 +269,11 @@ func Log(args ...string) (*subprocess.BufferedCmd, error) {
 	return gitNoLFSBuffered(logArgs...)
 }
 
+func RefLog(args ...string) (*subprocess.BufferedCmd, error) {
+	reflogArgs := append([]string{"reflog"}, args...)
+	return gitNoLFSBuffered(reflogArgs...)
+}
+
```
chrisd8088, line 276: See my other inline comment on `scanStashed()` for some musings on whether we could dodge the need for this new function by using `git.Log()` to read the stashed refs instead.

lfs/gitscanner_log.go
chrisd8088: I worked up a tentative first half of an alternate version of this function, which checks for `refs/stash` first, and then uses `git.Log()`.  The hard part, which I skipped for now, is reading the output; I thought that could borrow from the design of `parseScannerLogOutput()` but without the need for a `direction` parameter.

```go
func scanStashed(cb GitScannerFoundPointer, s *GitScanner) error {
        if _, err := git.ResolveRef("refs/stash"); err != nil {
                return nil
        }

        logArgs := []string{"-g"}

        // Add standard search args to find lfs references
        logArgs = append(logArgs, logLfsSearchArgs...)
        // Search reflog (with -g) for refs/stash
        logArgs = append(logArgs, "refs/stash", "--")

        cmd, err := git.Log(logArgs...)
        if err != nil {
                return err
        }
```

sinbad: Hmm, personally, I think the original version is more elegant - I know it's not checking the result code, but this is a pattern that already exists in the codebase (see RemoteList, RemoteURLs, RecentBranches). The only reason for this command to fail (and others to not) is that there are no stashes to list.

I didn't use `git log` because there's no need to "walk" the history - all stashes are singular commits. reflog is more efficient to get the data needed and I thought the RefLog command might be useful later on. But I don't mind if you'd prefer to write it another way.

chrisd8088: Ah, I see the similarity to `RemoteList` and friends now; I had not thought to look at those, only at the other functions nearby.  Well, given that, maybe there's a tidy way to combine the two approaches and get a minimal set of changes?

Here's what seems to be working for me tonight, replacing just the `git reflog show` with `git log -g`, i.e., something like:
```go
	// git log -g --format="%H" refs/stash --
	logArgs := []string{"-g", "--format=%H", "refs/stash", "--"}

	cmd, err := git.Log(logArgs...)
```

I can't claim to be an expert on the exact functional differences between those two commands (but I'll ask an expert :-) ... however, if they're effectively the same for this purpose, that would allow for reuse of the `git.Log()` function and avoid the need to add `git.RefLog()`.  Although I confess this is a very small thing, really; I could go either way (particularly if it turns out there's a specific advantage to the `git reflog` command)!

Thanks again so much for working on this; it's very much appreciated.

sinbad: OK you're right I think `-g` is basically an alias for reflog and makes log behave very differently. Unfortunately for the same reason `-g` makes some of the other options in `logLfsSearchArgs` to `git log` stop working, such as the `-G "oid sha256:" search, so we can't combine everything into one `git log` and send that  to `parseScannerLogOutput`. So you're still down to executing 2 commands, albeit that both of them are now `git log` so the RefLog command isn't needed.

chrisd8088: Yeah, I tossed out my attempt to use `logLfsSearchArgs`; that was a non-starter.  All I was thinking of was changing the first couple of lines of this function to use `git.Log()` and the `git log -g ...` command format.  (And according to the folks I consulted, `git log -g` is equivalent to `git reflog show`.

That said, based on the feedback I got, I realized there may be a different, more "interesting" issue yet to solve here, which is what happens when someone uses `git stash pop --index` and had LFS file changes in both their index *and* their working tree.  I'll write that up in the main comments, though.

chrisd8088:
What was pointed out to me today is that the stash uses multiple commits to store the state of the working tree and the index separately, and our various experiments with `git reflog show` and `git log -p` both only give the SHAs of the working tree stash commits.  That got me thinking that if someone used `git stash pop --index` and then tried to restore the working tree to the index, it might still cause a problem, even with this PR's changes.  And indeed that seems to be the case.

I wrote up a test which demonstrates the outstanding issue here, I believe; at least for me, using this PR as it stands, it failed on the second `assert_local_object "$oid_indexstashed" "${#content_indexstashed}"` (and in a simpler manual test, without the test helpers, I get a failure on `git checkout .` when the smudge filter can't download the missing LFS data):

```bash
begin_test "prune keep stashed changes in index"
(
  set -e

  reponame="prune_keep_stashed_index"
  setup_remote_repo "remote_$reponame"

  clone_repo "remote_$reponame" "clone_$reponame"

  git lfs track "*.dat" 2>&1 | tee track.log
  grep "Tracking \"\*.dat\"" track.log

  # generate content we'll use
  content_inrepo="This is the original committed data"
  oid_inrepo=$(calc_oid "$content_inrepo")
  content_indexstashed="This data will be stashed from the index and should not be deleted"
  oid_indexstashed=$(calc_oid "$content_indexstashed")
  content_stashed="This data will be stashed and should not be deleted"
  oid_stashed=$(calc_oid "$content_stashed")

  # We just need one commit of base data, makes it easier to test stash
  echo "[
  {
    \"CommitDate\":\"$(get_date -1d)\",
    \"Files\":[
      {\"Filename\":\"stashedfile.dat\",\"Size\":${#content_inrepo}, \"Data\":\"$content_inrepo\"}]
  }
  ]" | lfstest-testutils addcommits

  # now modify the file, and add it to the index
  echo -n "$content_indexstashed" > stashedfile.dat
  git add stashedfile.dat

  # now modify the file again, and stash it
  echo -n "$content_stashed" > stashedfile.dat

  git stash

  # Prove that the stashed data was stored in LFS (should call clean filter)
  assert_local_object "$oid_stashed" "${#content_stashed}"
  assert_local_object "$oid_indexstashed" "${#content_indexstashed}"

  # Prune data, should NOT delete stashed file or stashed changes to index
  git lfs prune

  assert_local_object "$oid_stashed" "${#content_stashed}"
  assert_local_object "$oid_indexstashed" "${#content_indexstashed}"

  # Restore working tree from stash
  git stash pop --index

  # Reset working tree to index from stash
  git checkout .
)
end_test
```

So I think we should add a test like this, first of all, and also consider how best to walk back through all the SHAs associated with the stash.

sinbad:
Nice catch, I never realised stash could create >1 commit but thinking about it I guess it makes sense. I guess it can only ever create 2. I'll have another crack at it.

sinbad:
Here's on failed attempt that I thought would work:

```go
func scanUnpushed(cb GitScannerFoundPointer, remote string) error {
	logArgs := []string{
		"--branches", "--tags", // include all locally referenced commits
		"--not"} // but exclude everything that comes after

	if len(remote) == 0 {
		logArgs = append(logArgs, "--remotes")
	} else {
		logArgs = append(logArgs, fmt.Sprintf("--remotes=%v", remote))
	}

	// Add standard search args to find lfs references
	logArgs = append(logArgs, logLfsSearchArgs...)

	cmd, err := git.Log(logArgs...)
	if err != nil {
		return err
	}

	parseScannerLogOutput(cb, LogDiffAdditions, cmd)
	return nil
}
```

While this does list the correct commits, the diff generated by the `-p` option *only* includes the OID for the combined change for the entire stash, i.e. the working copy version and not the one that was in the index. I can't seem to get the diff to output both and I'm not sure why, it's not how it behaves for other commits. 

Looking at the result of `git show -p <stash_sha>` it seems to be that it knows the changes are different to a normal log entry:

```diff
- oid sha256:67ae8ef97d615140fcb42b3fa7e300a5ce5960cd8ff9807eef3dcb85aec3cecd
- size 25
 -oid sha256:32609ffbe33e263e807f7e3b8ee6ae7d990da65de5c15a51f1b62782a8cc7276
 -size 43
++oid sha256:c9d22606d40fec4e99cc7d09fbef109f5301a30e35c177e0d14bafb36894cdc5
++size 64
```

This looks more like a `git status --porcelain` style output where positioning of the -/+ tells you whether it's the index or working copy version. In the example above OID 67ae8ef9 (- in col 0) was the previous commit, OID 32609ffbe (- in col 1) was the stashed index, and OID c9d22606d (++) was the working copy version. `git log` only shows the first and last OIDs even if you're definitely getting the commit inbetween in the output, the diff for it just isn't there.

FYI for stash commits that only had the working copy, `git show -p <stash_sha>` displays:

```diff
--oid sha256:67ae8ef97d615140fcb42b3fa7e300a5ce5960cd8ff9807eef3dcb85aec3cecd
--size 25
++oid sha256:c7fe2c1adc4c9fbe8df7d1bd56b85d98e52288ffc6547fcb7a15fab39a4942ed
++size 15
```

So I think the best approach is going to be to use `git show -p <stash_sha>...` and parse out any OIDs which have either "+" or "-" in **column 1** of the diff output. That should pick up the index and working copy state of stashes. Thoughts? I'd love to figure out a way to get `git log` to output that full detail like `git show` but so far I've had no luck.

sinbad:
Sorry I pasted the wrong "failed attempt" function in the previous comment at first, fixed now.

chrisd8088:
One other thing we should consider (again, based on the tips I got about how `git stash push` operates) is that `git stash push -u` makes yet a third commit, to capture any untracked files.

Doing something like:
```bash
echo 123 > foo.dat; git add foo.dat
git commit -m initial
echo 456 > foo.dat
git add foo.dat
echo 789 > foo.dat
echo abc > bar.dat
git stash push -u
git log -g refs/stash
```

shows something like the following, where each of the three SHAs in the `Merge:` line represents the stashed state of the index, the stashed changed files from the working tree, and the stashed untracked files from the working tree:
```
Reflog message: WIP on master: 2678f27 initial
Merge: 2678f27 2762e90 a71ca8c
```

sinbad:
OK that suggests that we need both commands. To summarise:

1. `git log -g --format=%h refs/stash --` to get the 'leaf' stash SHAs
1. `git log -p <stashSHA>... --not --branches --tags` (+logSearchArgs) to pick up changes from untracked files and the final diff (but it will miss the diff between working copy and index)
1. `git show -p <stashSHA>...` to pick up the diff between working copy & index on same file, if it exists

The set of OIDs from all of those should be the complete set.

sinbad:
Alternatively if we used the "Merge" line:

```
Reflog message: WIP on master: a70e14d First
Merge: a70e14d 4cf30f4 1bad043
```

In my case a70e14d is the parent commit of the stash (which should be ignored), 4cf30f4 is the index contents and 1bad043 is the untracked files. The leaf SHA is then the working copy. If that's a consistent we could ignore the first SHA on the merge line and use the rest plus the leaf as arguments to `git show`. It's a little more parsing though (becomes 2 stages); in the previous suggestion you only need the leaf SHAs and everything else is discoverable via OID lines (albeit with a slightly different regex to handle index/working copy).

The downside of this as well as assuming the order is it's a bit more parsing

chrisd8088:
Yes, exactly -- I realized I needed to edit my description of which commits were which in https://github.com/git-lfs/git-lfs/pull/4209#issuecomment-672981922 and then noticed you'd already explained the relevant commits more correctly.  :-)

sinbad:
Bad news, I'm not sure exactly what version of git started reporting merge parents for stashes, but it was somewhere later than git 2.7 because during my testing on an older linux setup with older git, I couldn't get `git log -g refs/stash` to report the merge line, or the parent data when using `--format=%p`. It's just blank. So I don't think we want to rely on that behaviour.

sinbad:
For reference in this git 2.7 the parentage of the commits are still the same, it's just that `git log -g --format=%p` doesn't report them (and neither does `git reflog`). So at some point the reflog must have been enhanced to support this.

`git show --format=%p` does show the merge parents of a stash leaf commit, however. So that's probably the way to go.

sinbad:
OK this took more a lot more work than I expected, but it's finally working. I had to add a "Show" command and I had to also change the OID pointer diff parsing because it couldn't handle merge commits. I added a test which includes all 3 kinds of stash change at once because there was a bug that only happened when you had all 3 together.

Hoping this is all OK now!

chrisd8088:
OK, wow, that's a lot of work indeed, thank you!  I plan to take a look tomorrow but from the tests it looks like this covers a lot of territory and all the use cases.  ❤️

chrisd8088:
Thanks so much for sticking with this!  I just wanted to say I'm still mid-review and have not forgotten this, and will get you more feedback as soon as I can.  Thanks so much again!

lfs/gitscanner_log.go
```diff
@@ -63,6 +64,106 @@ func scanUnpushed(cb GitScannerFoundPointer, remote string) error {
...
+	if err != nil {
+		// Ignore this error, it really only happens when there's no refs/stash
+		return nil
+	}
+
+	// Now we need to specifically use "git show" to parse results
+	// We can't use "git log" because weirdly that omits the index changes
+	// in the diff display, it collapses both into one diff and only shows the
+	// final change (not a 3-way like show). Only "show" on all the shas
+	// above displays them separately
```
chrisd8088, line 137: Please forgive me if I've got the wrong end of the stick here!  If I'm following along, we use `git log -g` to get the reflog list of stashes; for each stash, this is the final `WIP` merge commit, which may have additional parents in the form of the index and untracked-files commits.

So we use `git show` to retrieve the SHAs of each of those, for each stash, and then iterate again over that last running `git show` on the whole concatenated list of SHAs, in order to get any LFS pointer data out of them.

I can't help but feel that there might be a more compact way to achieve the same results.  I haven't tried this in code yet, but what about using `git log -m` plus the usual `logLfsSearchArgs` (i.e., `-p` in particular) just each of the SHAs from the first list from the reflog, of the WIP merge commits?  Doing that seemed like it might get all the necessary LFS data from all of the parents in one go, without needing run `git show` at all.

I realize older Git versions may throw a wrinkle in here, but testing with Git v2.0.0 it seems like it might work (again, on the command line only, not in code yet).

I can try to write up another version if you don't get to it first, just to test this out ... but I may not get it done tomorrow, so I won't promise that!  :-)

sinbad: If you can find a better way to do it I'm quite happy! I kind of ran out of time with this, I spent way longer than I expected last week tinkering with it and this way was the only one I came up with in the time that worked consistently. I did try a few log variations but there were 2 problems:

1. `log -g` doesn't show merges in older git versions
1. In my experiments `log` often collapsed the diff of the WIP and index, making it impossible to get all 3 changes. I don't know why it did this, but only `show` reliably gave me all 3 OIDs. If you can find a way to make it work all the time, great!

sinbad: Ah sounds like from your comments the `-m` option might fix this and the more awkward diff format. That will be much better!

I don't know when I'm going to have time to revisit this though, it'll probably be a few days.

chrisd8088: No worries at all; you've gotten this a huge distance down the road and much faster than I could have.  Let me see if I can find some time in the next day or two to tinker a bit more with this.

Would it be OK if I pushed commits onto your branch in your repo?  I haven't wanted to do that without asking for permission first (even if the "allow edits from maintainers" checkbox is checked on your side).

sinbad: Oh yeah that's fine, thanks for asking but I'm very happy to collaborate on this. :)

lfs/gitscanner_log.go
```diff
@@ -155,7 +256,8 @@ func newLogScanner(dir LogDiffDirection, r io.Reader) *logScanner {
 		commitHeaderRegex:    regexp.MustCompile(fmt.Sprintf(`^lfs-commit-sha: (%s)(?: (%s))*`, git.ObjectIDRegex, git.ObjectIDRegex)),
 		fileHeaderRegex:      regexp.MustCompile(`diff --git a\/(.+?)\s+b\/(.+)`),
 		fileMergeHeaderRegex: regexp.MustCompile(`diff --cc (.+)`),
-		pointerDataRegex:     regexp.MustCompile(`^([\+\- ])(version https://git-lfs|oid sha256|size|ext-).*$`),
+		// stash diff can have up to 3 +/- characters
+		pointerDataRegex: regexp.MustCompile(`^([\+\- ]{1,3})(version https://git-lfs|oid sha256|size|ext-).*$`),
```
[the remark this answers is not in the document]
chrisd8088: I've implemented this in 34b323d4f76f6efd4a9c2fec7d1f9410af0b762d, and then 7fb982f946cdd0db9ec0e3a2d96285be41a2b129 lets us make things even tidier by avoiding the need for `git show` entirely, I believe.

lfs/gitscanner_log.go
chrisd8088: I circled back around to this (thanks for your patience!) and put together 7fb982f946cdd0db9ec0e3a2d96285be41a2b129, which seems to pass the new test suite and lets us use `git log` exclusively again, I think.

sinbad: Thanks for doing that, it's a much better solution. We win! 👍 😄

sinbad:
Looks good to me! 👍

commands/command_prune.go
```diff
@@ -71,7 +71,7 @@ func prune(fetchPruneConfig lfs.FetchPruneConfig, verifyRemote, dryRun, verbose
 	// Add all the base funcs to the waitgroup before starting them, in case
 	// one completes really fast & hits 0 unexpectedly
 	// each main process can Add() to the wg itself if it subdivides the task
-	taskwait.Add(4) // 1..4: localObjects, current & recent refs, unpushed, worktree
+	taskwait.Add(5) // 1..5: localObjects, current & recent refs, unpushed, worktree, stashes
 	if verifyRemote {
 		taskwait.Add(1) // 5
```
bk2204, line 76: I think this comment should now be `6`.

t/t-prune.sh
```diff
@@ -594,3 +594,161 @@ begin_test "prune verify large numbers of refs"
...
+  echo "[
+  {
+    \"CommitDate\":\"$(get_date -1d)\",
+    \"Files\":[
+      {\"Filename\":\"stashedfile.dat\",\"Size\":${#content_inrepo}, \"Data\":\"$content_inrepo\"}]
+  }
+  ]" | lfstest-testutils addcommits
+
+  # now modify the file, and stash it
+  echo -n "$content_stashed" > stashedfile.dat
```
bk2204, line 626: We generally prefer to use `printf "%s"` instead of `echo -n` in this codebase because `printf` is both more flexible and more portable.  `echo` is fine if you want its normal behavior of a newline.

There are a couple of other instances of this below.

bk2204 (commented):
Overall, looks great.  I found two minor things that we might want to fix, but I'm pretty happy with the way this turned out.

commands/command_prune.go
chrisd8088: Good catch, thank you!

t/t-prune.sh
chrisd8088: Indeed, thanks for spotting that -- fixed!

PR #4217, opened 2020-08-17
Add filename in verify request header.

xzy57369:
When using Alibaba OSS(Object Storage Service) as binary assets server, the content-type field will be included in checking pre-signed url due to the security reason. But as the LFS api specifications, the api will not know the content-type information when generating pre-signed url, causing the url unmatched. One possible solution is to using default content-type "application/octet-stream" when signing url and set content-type to "application/octet-stream" in batch reponse's header field.
This causes another problem, since both LFS api and oss does not the file name and see the content-type as octet-stream. The file type information is losing and it is important, because we want to provide a dashboard to our users for managing those files tracked by LFS. It will be a convience if they can filt files by file types.
By checking the source code we find that it is easy to provide the filename in verify requests. We decided to provide the filename in request header instead of in request body, because it will do less effect to the original api.

Signed-off-by: yunhuai.xzy <yunhuai.xzy@alibaba-inc.com>

### DIFF
--- a/tq/verify.go
+++ b/tq/verify.go
@@ -1,6 +1,7 @@
 package tq
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/git-lfs/git-lfs/lfsapi"
@@ -37,6 +38,8 @@ func verifyUpload(c *lfsapi.Client, remote string, t *Transfer) error {
 
 	req.Header.Set("Content-Type", "application/vnd.git-lfs+json")
 	req.Header.Set("Accept", "application/vnd.git-lfs+json")
+	req.Header.Set("Content-Disposition", fmt.Sprintf("attachment; filename=%q", t.Name))
+
 	for key, value := range action.Header {
 		req.Header.Set(key, value)
 	}


xzy57369:
Thanks for your reply @bk2204 
It is true that the same object may have many different file names or types, but Bitbucket has supported the file type detection. I guess the content-type may be used, which is not available in our cases. How do you think they achieved it, is there any another methods?
image


We have come up with three ideas. 
First of all it to adding filename in verify request header, although it may not be so accurate, it still provide useful information for most cases. 
The second is to read some bytes from object storage and detect the file header, which is a bit slower and require extra work on the server side. 
The third is to put a service before object storage, receiving the object first and then upload it to object storage. This wastes server's bandwidth and performance, and may become a bottleneck when facing high concurrency situation.

Which method do you prefer?

bk2204:
Usually the approach one uses here is to use libmagic or its command-line tool, `file`, to inspect the object and detect its type.  I think that's the thing you want to do if you can't use the Content-Type header that we provide.  After all, only you know what types of objects you can render and display, and inspecting the object is the best way to know whether you can, in fact, render it.  I suspect that's the thing that Atlassian's doing, although I don't know for certain.

xzy57369:
libmagic or file would be useful if the file exists on our server, but in fact they don't. Actually, what we want is not a 100% right file type, maybe a confidence of 95% is enough by checking the filename. Thus we wish that verify request could have the ability to providing filenames.

PR #4230, opened 2020-09-03
Add manual pages to release assets

### LINKED ISSUE #4228: Include man pages in release binaries
**Describe the issue**
Would you mind to include built manpages in the binary tarballs published on Github releases?

This'd make manpages available to folks which install git lfs locally into their $HOME.

**System environment**
n.a.

**Output of `git lfs env`**
n.a.

**Additional context**
n.a.

bk2204:
In general, it would be nice to include pre-built manual pages in the release archives, both in the normal troff format and as HTML.  They're already baked into the binary as help output, but for some folks, `man` is easier to use.

This PR includes the manual pages in the archives and sets up the CI system to make sure Ruby and `ronn` are installed so we can build the release assets on every platform.  During the Windows CI run, we don't explicitly run `make release`, so call `make man` so we test that they can be built there.

Fixes #4228
/cc @lunaryorn as submitter

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v1
+    - uses: actions/setup-ruby@v1
+    - run: gem install ronn
     - run: script/cibuild
     - run: CGO_ENABLED=0 make release
     - run: mkdir -p bin/assets
@@ -35,6 +37,7 @@ jobs:
     runs-on: windows-latest
     steps:
     - uses: actions/checkout@v1
+    - uses: actions/setup-ruby@v1
     - run: |
         echo "::add-path::C:\Program Files\Git\mingw64\bin"
         echo "::add-path::C:\Program Files\Git\usr\bin"
@@ -47,6 +50,10 @@ jobs:
     - run: choco uninstall git.install -y --force
     - run: cinst git --version 2.26.2 -y --force
     - run: refreshenv
+    - run: gem install ronn
+      shell: bash
+    - run: make man
+      shell: bash
     - run: GOPATH="$HOME/go" PATH="$HOME/go/bin:$PATH" go get github.com/josephspurrier/goversioninfo/cmd/goversioninfo
       shell: bash
     - run: GOPATH="$HOME/go" PATH="$HOME/go/bin:$PATH" script/cibuild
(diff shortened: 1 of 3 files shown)


chrisd8088 (approved):
Thanks for doing this, and for breaking it up into two tidy commits! 👍

Alhadis:
@bk2204 Just curious, have you considered using [`mandoc`](https://mandoc.bsd.lv/man/) for generating HTML versions of *actual* man pages? There's a modern, semantic alternative to the traditional [`man(7)`](https://man.openbsd.org/man.7) macros called [`mdoc(7)`](https://man.openbsd.org/mdoc.7) which also supports markdown output (should you need it).

I'd be happy to submit a pull-request if you're interested.

bk2204:
Hey,

We use Markdown here because we embed the text output into the binary for the `--help` output.  I'm familiar with both `man(7)` and `mdoc(7)`, but because they both require troff or a similar program to render, they aren't a good choice for our needs.

Also, while I am personally fine with writing manual pages in troff-based formats, a lot of newer folks are more familiar with Markdown and AsciiDoc and other similar formats, and continuing to use them benefits new contributors.

Alhadis:
> but because they both require troff or a similar program to render, they aren't a good choice for our needs.

Hold up. Troff "or a similar program" (I assume you mean `mandoc`?) is required to view man pages anyway, so that argument makes no sense.

bk2204:
> Hold up. Troff "or a similar program" (I assume you mean `mandoc`?) is required to view man pages anyway, so that argument makes no sense.

Sorry, I was unclear.  We embed the literal Markdown from the source (modified slightly to remove HTML) into the `git-lfs` binary as part of the build process.  The program is in `docs/man/mangen.go`, and it produces output in `commands/mancontent_gen.go`.  So when you run e.g., `git lfs help track`, what you're seeing is the Markdown from the manual page.

We _do_ want some way to indicate things like headings in such a case, but we _don't_ want terminal escape codes, because some people may be using dumb terminals (e.g., emacs's `M-x shell`), so Markdown and similar formats are ideal for this situation.

We _also_ have to consider the tooling available on various Linux systems because we build packages there.  CentOS 7 does ship Ruby, but it doesn't ship `mandoc`, so we're forced to either build it (as we do for the Ruby gems, which are much easier) or use something else.  Additionally, we can usually depend on some version of Ruby being available for Windows, but we cannot depend on `mandoc` or `groff`, since it isn't guaranteed that a C compiler is available there, and even if it were, the tool probably wouldn't compile.  Chocolatey, where we get our extra tooling from, doesn't ship `mandoc`, and it looks like it's pretty Unix specific.

PR #4231, opened 2020-09-03
tools: always force a UTF-8 locale for cygpath

### LINKED ISSUE #4227: git lfs does not work on windows bash.exe when used with unicode paths
**Describe the bug**
This is a follow up from https://stackoverflow.com/questions/63691374/git-lfs-checkout-fails-on-windows-with-unicode-path-error-createfile
I am running Windows 10 Pro in a virtual machine with nothing but Git for Windows installed. I can reproduce the issue on a physical machine as well, however.
Running any LFS command that modifies the working directory fails on bash.exe when the path to the repository contains Unicode characters. This can also be reproduced when executing a git lfs pull through GitKraken (which also seems to invoke bash.exe).

**To Reproduce**
Run the following shell script on bash.exe and compare it to git-bash.exe
```
createRepo()
{
  directory=$1
  mkdir $directory
  cd $directory
  git init
  git lfs install --local
  git config --local user.email "test@mail.com"
  git config --local user.name "test"
  git lfs track bin.dat
  git add .gitattributes
  git commit -m "attribute file"

  file="bin.dat"
  echo "qwertyuiop" > "bin.dat"

  git add .
  git commit -m "added binary file"

  cd ..
}

mkdir lfsissue
cd lfsissue

echo "create repo in path with ascii characters only"
createRepo test

echo "create repo in path with unicode characters"
createRepo äöü
```

Compare the output from bash.exe
```
$ sh issue.sh
create repo in path with ascii characters only
Initialized empty Git repository in C:/Users/VM/Desktop/lfsissue/test/.git/
Updated git hooks.
Git LFS initialized.
Tracking "bin.dat"
[master (root-commit) 88018d1] attribute file
 1 file changed, 1 insertion(+)
[master 9d69353] added binary file
 create mode 100644 bin.dat
create repo in path with unicode characters
Initialized empty Git repository in C:/Users/VM/Desktop/lfsissue/äöü/.git/
Error: CreateFile C:\Users\VM\Desktop\lfsissue\���: The system cannot find the file specified.
Updated git hooks.
Git LFS initialized.
Error: CreateFile C:\Users\VM\Desktop\lfsissue\���: The system cannot find the file specified.
This operation must be run in a work tree.
fatal: pathspec '.gitattributes' did not match any files
On branch master

Initial commit

nothing to commit (create/copy files and use "git add" to track)
warning: LF will be replaced by CRLF in bin.dat.
The file will have its original line endings in your working directory
Error: CreateFile C:\Users\VM\Desktop\lfsissue\���: The system cannot find the file specified.
[master (root-commit) d316732] added binary file
 1 file changed, 1 insertion(+)
 create mode 100644 bin.dat
```
to the correct and expected output of git-bash.exe
```
$ sh issue.sh
create repo in path with ascii characters only
Initialized empty Git repository in C:/Users/VM/Desktop/lfsissue/test/.git/
Updated git hooks.
Git LFS initialized.
Tracking "bin.dat"
[master (root-commit) 5816b6c] attribute file
 1 file changed, 1 insertion(+)
[master b5276fa] added binary file
 create mode 100644 bin.dat
create repo in path with unicode characters
Initialized empty Git repository in C:/Users/VM/Desktop/lfsissue/├ñ├Â├╝/.git/
Updated git hooks.
Git LFS initialized.
Tracking "bin.dat"
[master (root-commit) d4c2fba] attribute file
 1 file changed, 1 insertion(+)
[master 7bf822e] added binary file
 create mode 100644 bin.dat

```

**Output of `git lfs env`**
git-lfs/2.11.0 (GitHub; windows amd64; go 1.14.2; git 48b28d97)
git version 2.28.0.windows.1

LocalWorkingDir=
LocalGitDir=
LocalGitStorageDir=
LocalMediaDir=lfs\objects
LocalReferenceDirs=
TempDir=lfs\tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=lfs
AccessDownload=none
AccessUpload=none

bk2204:
When we look up the repository path from Git, we pass it through cygpath -w to canonicalize it into a Windows path, since Cygwin's Git will give us a Unix-style path.  We perform path canonicalization not only on Cygwin, but also on MINGW as well, which include Git Bash, since we want to accept and canonicalize Unix-style paths there as well.

Normally, this works great.  However, if invoked not from Git Bash, but via the Git for Windows bash.exe command, no locale is set in the environment, despite the locale binary indicating UTF-8 locales.  As a result, if non-ASCII character exist in the path name, it tries to encode them in ISO-8850-1.

On a standard Unix, where paths are always bytes, defaulting to ISO-8859-1 might be fine, because regardless of the encoding, paths are always bytes and no encoding needs to be performed.  On macOS, where the file system and all locales use UTF-8, this is also not a problem, because again, no encoding needs to be done.

However, on Windows, where paths are natively stored as UTF-16, this is remarkably unhelpful, since the majority of Unicode code points cannot be represented in ISO-8859-1.  Thus, the vast majority of paths are broken by cygpath when the locale is not set.

Since we know we always want UTF-8 from cygpath, let's just force that in the environment we pass it.  We need to copy the environment since the value we have is shared among all executed subcommands and we don't want to modify other commands' locales, since that would cause error messages to be printed in English instead of the user's locale.

Note that before this change, the test would fail because the local working directory was not read, and therefore it would be empty on Windows.

Fixes #4227

### DIFF
--- a/tools/os_tools.go
+++ b/tools/os_tools.go
@@ -29,6 +29,15 @@ func Getwd() (dir string, err error) {
 
 func translateCygwinPath(path string) (string, error) {
 	cmd := subprocess.ExecCommand("cygpath", "-w", path)
+	// cygpath uses ISO-8850-1 as the default encoding if the locale is not
+	// set, resulting in breakage, since we want a UTF-8 path.
+	env := make([]string, 0, len(cmd.Env)+1)
+	for _, val := range cmd.Env {
+		if !strings.HasPrefix(val, "LC_ALL=") {
+			env = append(env, val)
+		}
+	}
+	cmd.Env = append(env, "LC_ALL=C.UTF-8")
 	buf := &bytes.Buffer{}
 	cmd.Stderr = buf
 	out, err := cmd.Output()
--- a/t/t-env.sh
+++ b/t/t-env.sh
@@ -975,3 +975,68 @@ UploadTransfers=basic,lfs-standalone-file
   contains_same_elements "$expected" "$actual"
 )
 end_test
+
+
+begin_test "env with unicode"
+(
+  set -e
+  # This contains a Unicode apostrophe, an E with grave accent, and a Euro sign.
+  # Only the middle one is representable in ISO-8859-1.
+  reponame="env-d’autre-nom-très-bizarr€"
+  unset_vars
+  mkdir $reponame
+  cd $reponame
+  git init
+  git remote add origin "$GITSERVER/env-origin-remote"
+  git remote add other "$GITSERVER/env-other-remote"
+
+  touch a.txt
+  git add a.txt
+  git commit -m "initial commit"
+
+  # Set by the testsuite.
+  unset LC_ALL
+
+  endpoint="$GITSERVER/env-origin-remote.git/info/lfs (auth=none)"
+  endpoint2="$GITSERVER/env-other-remote.git/info/lfs (auth=none)"
+  localwd=$(canonical_path "$TRASHDIR/$reponame")
+  localgit=$(canonical_path "$TRASHDIR/$reponame/.git")
+  localgitstore=$(canonical_path "$TRASHDIR/$reponame/.git")
+  lfsstorage=$(canonical_path "$TRASHDIR/$reponame/.git/lfs")
+  localmedia=$(canonical_path "$TRASHDIR/$reponame/.git/lfs/objects")
+  tempdir=$(canonical_path "$TRASHDIR/$reponame/.git/lfs/tmp")
+  envVars=$(printf "%s" "$(env | grep "^GIT")")
+  expected=$(printf '%s
+%s
+
+Endpoint=%s
+Endpoint (other)=%s
+LocalWorkingDir=%s
+LocalGitDir=%s
+LocalGitStorageDir=%s
+LocalMediaDir=%s
+LocalReferenceDirs=
+TempDir=%s
+ConcurrentTransfers=8
+TusTransfers=false
+BasicTransfersOnly=false
+SkipDownloadErrors=false
+FetchRecentAlways=false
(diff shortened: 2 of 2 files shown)


tools/os_tools.go
```diff
@@ -29,6 +29,13 @@ func Getwd() (dir string, err error) {
 
 func translateCygwinPath(path string) (string, error) {
 	cmd := subprocess.ExecCommand("cygpath", "-w", path)
+	// cygpath uses ISO-8850-1 as the default encoding if the locale is not
+	// set, resulting in breakage, since we want a UTF-8 path.
+	env := make([]string, 0, len(cmd.Env)+1)
+	for _, val := range cmd.Env {
+		env = append(env, val)
+	}
+	cmd.Env = append(env, "LC_ALL=C.UTF-8")
```
chrisd8088, line 38: I would imagine that by appending this at the end of the `env` array, if `LC_ALL` appears earlier as well, it will be supplanted by this value ... I'm virtually certain that would be the effect on Unix, but a little less so about Windows.  Might it be useful to adjust the test to explicitly try setting `LC_ALL` to something else before running `git lfs env`?

chrisd8088 (approved):
This looks great.  I had one question but it's more for my own curiosity than anything else.

tools/os_tools.go
bk2204: Good point.  I'll just strip out the `LC_ALL` entry if it exists.

PR #4249, opened 2020-09-14
Add FAQ to README and CONTRIBUTING

bk2204:
We have some frequently asked questions that we get, so I've written up a FAQ on the wiki.  Let's add a link to the FAQ in the README and CONTRIBUTING documentation to help users find it more easily.

In addition, rephrase a part of the contributing documentation which uses language similar to a well-known political campaign.  We don't want people to associate our project with any political campaign, since such associations are irrelevant and distracting.


chrisd8088 (approved):
👍 Thanks!

PR #4251, opened 2020-09-16
tools/util_darwin.go: Remove use of direct syscalls

stanhu:
Previously, trying to compile `util_darwin.go` against a recent version
of `golang/sys` library failed with:

```
util_darwin.go:127:3: undefined: unix.SYS_CLONEFILEAT
```

Go v1.12 removed the ability to use direct syscalls. The clonefile
system calls was added to the latest `golang/sys` in
https://github.com/golang/go/issues/41366, so we can use that now.

### DIFF
--- a/tools/util_darwin.go
+++ b/tools/util_darwin.go
@@ -9,7 +9,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
-	"unsafe"
+	"syscall"
 
 	"github.com/git-lfs/git-lfs/errors"
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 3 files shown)


tools/util_darwin.go
```diff
@@ -112,30 +112,11 @@ func CloneFileByPath(dst, src string) (bool, error) {
...
-		uintptr(atFDCwd),
-		uintptr(unsafe.Pointer(dstCString)),
-		uintptr(C.CLONE_NOFOLLOW),
-		0,
-	)
-	if errNo != 0 {
 		return &CloneFileError{
-			Unsupported: errNo == C.ENOTSUP,
-			errorString: fmt.Sprintf("%s. from %v to %v", unix.ErrnoName(errNo), src, dst),
+			Unsupported: err == syscall.ENOTSUP,
```
stanhu, line 118: This is not quite right.

tools/util_darwin.go
```diff
@@ -112,30 +112,11 @@ func CloneFileByPath(dst, src string) (bool, error) {
...
+			errorString: fmt.Sprintf("%v from %v to %v: %v", err, src, dst),
```
stanhu, line 119: We might be able to a type assertion here with `unix.ErrnoName` here because the syscall takes the `syscall.Errno` value and uses this wrapper:

```go
// errnoErr returns common boxed Errno values, to prevent
// allocations at runtime.
func errnoErr(e syscall.Errno) error {
	switch e {
	case 0:
		return nil
	case EAGAIN:
		return errEAGAIN
	case EINVAL:
		return errEINVAL
	case ENOENT:
		return errENOENT
	}
	return e
}
```

But I'm not quite sure how to test this, so I'm just putting the raw string.

stanhu: It looks like `syscall.Errno` already returns a human-readable string, so we're good here.

bk2204 (approved):
This looks great!  Thanks for the patch, and I agree this is the best way forward since it works on all the Go versions we want to use.

Issue #4193, opened 2020-07-18
Feature request: Allow defining pointers to existing HTTP files


mitar:
This is maybe more of a git-lfs design proposal/feature proposal issue. If there is a different venue to do those, please tell me.

So we are using git LFS to track and version scientific datasets. And it works great. We can track both metadata (as JSON files) and binary data and users can use familiar tool (git) to do so. It allows great reproducibility as well as you can store particular git hash with your computations and know that you can retrieve this later. While working more on this I realized there could be small optimization in terms of storage requirements, and also something which would help when storing large datasets on GitHub and GitLab. Namely, both of those services limit the size of files and/or repositories you can have to store large files. Which has posed some issues to us in the past but we have found some workarounds (compressing data, etc.) but I anticipate workarounds might not always be possible. So as we are currently using GitHub and GitLab to those those datasets, I would not like to hit a point when this is not possible anymore.

On the other hand, many datasets are already published publicly online in different data repositories, accessible through public HTTP. So could we extend the pointer spec to allow also:

````
version https://git-lfs.github.com/spec/v1
oid sha256:4d7a214614ab2935c943f9e0ff69d22eadbb8f32b1258daaa5e2ca24d17e2393
size 12345
available https://example.com/file.csv
(ending \n)
```

So that when one would be cloning the repository, `https://example.com/file.csv` would be fetched instead, checked for hash, and then checked out? Pointers with `available` key (name just a suggestion) would not be pushed at all during committing/pushing, so this could operate without any git LFS server side component. (We could have some option that during first time this is committed, the remote URL is validated.)

This would allow us to transparently decide where to store git LFS files. For most we could store them at services like GitHub and GitLab, but in the case that any individual file hits the limit, we could easily use alternatives: or existing files available on the web, or store it in a public S3 bucket or something. While tooling and user experience would stay exactly the same.

There could be a `git lfs remote-file` CLI tool which you would call like:

```
git lfs remote-file https://example.com/file.csv local/path.csv
```

And would create `local/path.csv` pointer file, pointing to `https://example.com/file.csv`, storing hash and size in it, and prepare everything to be simply committed and pushed.

## Related work

git LFS can already handle `file://` [files](https://github.com/git-lfs/git-lfs/blob/master/docs/man/git-lfs-standalone-file.1.ronn). Could this be expanded to handle also `https` files?

bk2204:
Hey,

[...]

PR #4256, opened 2020-09-18
Adding specificatiojn for static transfers adapter

mitar:
A RFC with a formal spec to address #4193.

Feedback welcome.

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -18,6 +18,7 @@ API Specification:
 
 Current transfer adapters include:
   * [Basic](./basic-transfers.md)
+  * [Static](./static-transfers.md)
 
 Experimental transfer adapters include:
   * Tus.io (upload only)
--- a/docs/api/static-transfers.md
+++ b/docs/api/static-transfers.md
@@ -1,0 +1,79 @@
+# Static Transfer API
+
+The Static transfer API is a non-interactive API for downloading LFS objects.
+Its main intent is to provide simple interoperability with existing data
+storage solutions by offloading serving LFS objects to them while using
+git and git LFS for version management.
+
+Support for it by Git LFS clients and servers is optional.
+
+Git LFS clients MAY implement this transfer API in a fallback manner:
+first attempting to download the LFS object using the Static transfer
+API and if this fails (e.g., the manifest file does not exist,
+oid cannot be found in the manifest file) use the Basic transfer API.
+
+## Downloads
+
+Downloading an object first requires fetching a manifest file from using
+HTTP from the remote server. The proposed location of the manifest file
+is `https://<remote domain>/.well-known/git-lfs-manifest.json`. The
+location is configured using the `lfs.staticurl` configuration option,
+or remote-specific `remote.{name}.lfsstaticurl` configuration option.
+Both of them can be provided in `.lfsconfig` file as well.
+
+The manifest file MUST contain a JSON payload that looks like this:
+
+```json
+{
+  "version": "1",
+  "transfer": "static",
+  "objects": [
+    {
+      "oid": "1111111",
+      "size": 123,
+      "authenticated": true,
+      "actions": {
+        "download": {
+          "href": "https://some-download.com/1111111",
+          "header": {
+            "Authorization": "Basic ..."
+          },
+          "expires_at": "2016-11-10T15:29:07Z"
+        }
+      }
+    }
+  ]
+}
+```
+
+The remote server SHOULD return together with the contents of the
+manifest file also suitable caching response headers (e.g.,
+`Cache-Control`, `Pragma`, `Last-Modified`, `ETag`). Moreover, it SHOULD
+respond to GET and HEAD HTTP requests and support caching request
+headers (e.g., `If-Modified-Since`, `If-None-Match`). Git LFS client
+SHOULD attempt to fetch the manifest file again if any of the
+authorization tokens expires.
+
+The Static transfer adapter will make a GET request on the `href`, expecting the
+raw bytes returned in the HTTP response.
+
+```
+> GET https://some-download.com/1111111
+> Authorization: Basic ...
+<
+< HTTP/1.1 200 OK
+< Content-Type: application/octet-stream
+< Content-Length: 123
+<
+< {contents}
+```
+
+Git LFS clients MAY support non-HTTP `href` URIs (e.g., IPFS, dat). Those URIs
+should use dedicated URI schemas to differentiate them from HTTP URIs.
+
+## Uploads
+
+Uploading is not supported with this transfer adapter. Managing uploads and
+updating the manifest file accordingly is out of the scope. This enables
+one to integrate it with existing data storage solutions in the way which
+makes the most sense for a particular data storage solution.


bk2204:
Hey,

Sorry for the delay.  I'm looking at this and thinking about it, and I hope to have a review for you in the next few days.

docs/api/static-transfers.md
```diff
@@ -0,0 +1,79 @@
...
+first attempting to download the LFS object using the Static transfer
+API and if this fails (e.g., the manifest file does not exist,
+oid cannot be found in the manifest file) use the Basic transfer API.
+
+## Downloads
+
+Downloading an object first requires fetching a manifest file from using
+HTTP from the remote server. The proposed location of the manifest file
+is `https://<remote domain>/.well-known/git-lfs-manifest.json`. The
+location is configured using the `lfs.staticurl` configuration option,
```
bk2204, line 20: I'd rather we didn't force things into a fixed path, because that prevents serving multiple repositories from the same server.  What if we said that it's under the `/static-v1` path under the normal `/info/lfs` URL (so `/info/lfs/static-v1`)?  That would let us autodiscover static support based on the same URL and wouldn't require a separate option.  It would also let this work magically for non-HTTP URLs.

docs/api/static-transfers.md
```diff
@@ -0,0 +1,79 @@
...
+      "oid": "1111111",
+      "size": 123,
+      "authenticated": true,
+      "actions": {
+        "download": {
+          "href": "https://some-download.com/1111111",
+          "header": {
+            "Authorization": "Basic ..."
+          },
+          "expires_at": "2016-11-10T15:29:07Z"
```
bk2204, line 41: I don't think we really want an `expires_at` component here.  If this is supposed to be static, let's make people provide persistent, cacheable URLs.

docs/api/static-transfers.md
```diff
@@ -0,0 +1,79 @@
...
+}
+```
+
+The remote server SHOULD return together with the contents of the
+manifest file also suitable caching response headers (e.g.,
+`Cache-Control`, `Pragma`, `Last-Modified`, `ETag`). Moreover, it SHOULD
+respond to GET and HEAD HTTP requests and support caching request
+headers (e.g., `If-Modified-Since`, `If-None-Match`). Git LFS client
+SHOULD attempt to fetch the manifest file again if any of the
+authorization tokens expires.
```
bk2204, line 55: My strong preference is that we don't have expiring tokens.  If we need expiring tokens, then we need a server to handle that properly, especially since caching can prevent re-fetching from working.

Let's also recommend that it support proper Range headers.

docs/api/static-transfers.md
```diff
@@ -0,0 +1,79 @@
...
+
+```json
+{
+  "version": "1",
+  "transfer": "static",
+  "objects": [
+    {
+      "oid": "1111111",
+      "size": 123,
+      "authenticated": true,
```
bk2204, line 34: Let's note that `authenticated` needs to have its normal value.  That is, this may not be set, and a user may need to authenticate using the normal mechanisms.

docs/api/README.md
```diff
@@ -18,6 +18,7 @@ API Specification:
 
 Current transfer adapters include:
   * [Basic](./basic-transfers.md)
+  * [Static](./static-transfers.md)
```
bk2204, line 21: I'd like to leave this document in the `proposals` directory until it's implemented, since otherwise folks might think we support this when we don't yet.

bk2204 (commented):
I think overall I like the direction of this, but I've left some comments on the way I'd like to see it go, and of course I'm interested in hearing thoughts from other folks on @git-lfs/core.

mitar:
Thank you for your comments and review!

docs/api/static-transfers.md
mitar: I was thinking here about allowing to cache discrepancies between manifest caching headers and individual entry token expiration. So I could have a long lasting (a year) token here and I can specify precise token expiration here. Then my caching of the manifest file might be some other value, just because a simple implementation using a regular HTTP server would look at a manifest file like static file. So I wanted to allow the client to detect that maybe file caching is not in sync with token expiration. And retry the manifest again. Because ideally that file got replaced in that year.

But sure, I can remove those tokens as well. I would leave `header` in though. But we can remove expiration.

docs/api/static-transfers.md
mitar: Sounds good. I can remove expiration from the spec and mention range headers.

docs/api/static-transfers.md
mitar: If we do not want expiration of tokens here, then we should probably just remove `authenticated` as well, no?

docs/api/README.md
mitar: Sure.

docs/api/static-transfers.md
bk2204: I don't think we need to, because if the repository is public and doesn't require authentication, then `authenticated` can be `true`.

docs/api/static-transfers.md
mitar: Oh, path is just proposed location, not forcing anything. I really think that here some simple HTTP server would be serving this static manifest file. So this is why I went with more standard `.well-known` here, which we could also standardize.

I do think that this should definitely be configurable in `lfs.staticurl` because generally the use case here is that you are not running a git LFS server-side component here. But I am open to what a default or proposed path is. In current version of my proposal we do not have a default and `lfs.staticurl` has to be set explicitly to enable this transfer adapter, but we do suggest where one should put the manifest.

Are you suggesting we should have auto-discovery and that git LFS client should always try to resolve the manifest file at some standard location before trying to use the basic transfer adapter? I thought this extra request every time at the beginning of a transfer is a no-go? But sure, if you prefer, I can add that.

>  It would also let this work magically for non-HTTP URLs.

Hm, not sure how you plan to support non-HTTP URLs for manifest file? Is this in-scope of this proposal? But sure, I can make it that:
* Unless `lfs.staticurl` is set, `/info/lfs/static-v1/manifest.json` is attempted to be fetched, based on the current origin, whichever protocol it is.
* If it does not find the manifest or the entry does not exist in the manifest, the basic transfer adapter is used instead.

Is this what you are proposing?

One open question to with this approach is: what happens if fetching the manifest fails for some other reason? Do we still fallback on basic transfer or abort? Because fallback might incur unexpected transfer costs for example. With opt-in use of static transfer adapter (by requiring `lfs.staticurl`) the answer is simpler: if it is set and we cannot fetch it, we have an error and we abort.

docs/api/static-transfers.md
mitar: Ah, I see. I misunderstood this flag. So `authenticated=true` means that entry is already authenticated and nothing extra has to be done to obtain the file. File `authenticated=false` means they have to authenticate with other means. Sounds good.

docs/api/static-transfers.md
bk2204: Yes, I'm suggesting we should have autodiscovery based on the remote or LFS URL.  While we don't have support for non-HTTP remotes now, other folks may offer them via extensions, and we should gracefully support them if we can, especially since this proposal is well suited to a variety of remote types.

If the manifest fails for some reason, we fall back.  What we could have is a `lfs.statictransfersonly` option like `lfs.basictransfersonly` that performs only static transfers.

mitar: OK, so we can have autodiscovery (`/info/lfs/static/manifest-v1.json`, is that OK?), if fetching fails, we fall back to regular transfer. I would also want `lfs.staticurl`. If it is set, that location is used. If it fails, it is an error.

I think `lfs.statictransfersonly` could be added, but I do not think it addresses the problem I described. That would be more of `lfs.statictransfermanifestshouldexist` flag, because I still want fallback to basic transfer for entries which do not exist there. I think by making failure explicit if `lfs.staticurl` is set addresses the problem. So should I still add `lfs.statictransfersonly`?

Does this sound good to you?

bk2204: I don't want to treat static transfers differently from basic or TUS transfers.  For example, we don't have a lfs.tusurl.  What we could do is provide an option with a list of transfer types (`lfs.transfertypes`) that are allowed in order and then use only those.

mitar: So you are saying that it should fall back always to the other transport, with both auto-discovered URL or with `lfs.staticurl` URL? And that if you explicitly list transfer adapter in `lfs.transfertypes` then it would fail if manifest cannot be fetched?

mitar: If you are saying that there should be no `lfs.staticurl` option, than this is breaking my main use case and something I thought we agreed on to have [in our initial discussion](https://github.com/git-lfs/git-lfs/issues/4193#issuecomment-663270154). My main use case for this is that I want to reuse existing stuff stored on the Internet which serve files on a server which does not run git LFS backend. As such I might not be able to put manifest anywhere, but at a location I can, next to those files, for example. So being able to exactly configure where manifest is, not necessary on origin, and not necessary under default path, is a requirement for me to be able to do that.

bk2204: > If you are saying that there should be no `lfs.staticurl` option, than this is breaking my main use case and something I thought we agreed on to have [in our initial discussion](https://github.com/git-lfs/git-lfs/issues/4193#issuecomment-663270154). My main use case for this is that I want to reuse existing stuff stored on the Internet which serve files on a server which does not run git LFS backend. As such I might not be able to put manifest anywhere, but at a location I can, next to those files, for example. So being able to exactly configure where manifest is, not necessary on origin, and not necessary under default path, is a requirement for me to be able to do that.

I'm saying there should be no `lfs.staticurl` option.  Either we can have remote autodiscovery using the normal remote lookup, or we can specify the manifest directly in the `lfs.url` option.

My reasoning is that I don't want us to learn separate URLs for every type of transfer adapter.  That leads to a lot of complexity and results in behavior that's hard to reason about.  Also, as it stands, Git LFS cannot handle multiple different remote URLs, so any attempt to write a proposal which requires it do so will necessarily require a major restructuring of the code before it can ever be implemented.  There's no approach where we can fall back to a different remote or different URL for fetching; that simply isn't viable at the moment.

mitar: > we can specify the manifest directly in the `lfs.url` option.

But how does then fallback work? How do we then know if `lfs.url` then points to a manifest file or to git LFS API endpoint? How we get git LFS API endpoint from URI which points to the manifest file?

I find this approach adding more complexity, not less.

> My reasoning is that I don't want us to learn separate URLs for every type of transfer adapter. 

I think this is a logical fallacy. I am not advocating for every type of transfer adapter, but only for this particular one, because to me it makes sense: it is a static file which could be hosted anywhere. That is the whole point of this transfer adapter. To be able to use legacy/existing infrastructure to serve files.

> Git LFS cannot handle multiple different remote URLs

We are not introducing a new full origin, just one URL the transfer adapter downloads. I do not understand why fetching that file from inside the static adapter code would require any changes to the rest of the git LFS codebase? I mean, manifest file can also then have entries which each can point to a completely different remote URL. Are you saying git LFS is unable to fetch them and download them when needed? That they can be just under one origin and this is it? Then whole this static adapter idea is not feasible anyway.

I really do not understand where is this complexity. How I understand is, when git LFS wants to download an OID, it will:
* Call the static adapter to download it.
* Static adapter will check if it has manifest cached.
  * If not, it will download it from `lfs.staticurl` or from automatically inferred URL.
    * If download fails, it will abort download if `lfs.staticurl` was explicitly set, otherwise pass it through to the next transfer adapter.
* Static adapter will check if OID is in the manifest.
* If yes, it will download the file.
* If not, it will pass it through to the next transfer adapter.

bk2204: > > we can specify the manifest directly in the `lfs.url` option.
> 
> But how does then fallback work? How do we then know if `lfs.url` then points to a manifest file or to git LFS API endpoint? How we get git LFS API endpoint from URI which points to the manifest file?

Git LFS cannot do fallback with separate URLs at this point.  As mentioned, there is no way to use multiple endpoints (including `lfs.url` and the like), regardless of where those come from.  For example, any remote URL we provide needs to have support for URL rewriting, config inheritance, URL-specific options (e.g., certificates), per-remote fetch and push URLs, preferred authentication types, and so forth.  All of that code is designed to work only on endpoints, so the URL needs to be a remote URL using the existing framework.
 
Moving this out to be a totally separate URL from the rest of the endpoint code means we have a giant exception to how things work that's special-cased for the static adapter.  That's not a desirable way forward.  For example, some people need to force the use of Basic authentication even though the server supports NTLM, and that is handled by the endpoint code.  We can't just tell people, “Too bad, this is a special case, none of the normal functionality works here.”  Users will be justifiably angry about that.

Fallback can work just fine if the fallback is to the same endpoint URL.

To distinguish between the direct URL and autodiscovery, we could either fetch the URL directly and see if it contains a manifest, and then fall back to autodiscovery, or we could specify two similar transfer adapters (`static-direct` and `static`).

> I think this is a logical fallacy. I am not advocating for every type of transfer adapter, but only for this particular one, because to me it makes sense: it is a static file which could be hosted anywhere. That is the whole point of this transfer adapter. To be able to use legacy/existing infrastructure to serve files.

Yes, I don't think there's a problem with specifying that using `lfs.url` or a remote endpoint as normal.
 
> We are not introducing a new full origin, just one URL the transfer adapter downloads. I do not understand why fetching that file from inside the static adapter code would require any changes to the rest of the git LFS codebase? I mean, manifest file can also then have entries which each can point to a completely different remote URL. Are you saying git LFS is unable to fetch them and download them when needed? That they can be just under one origin and this is it? Then whole this static adapter idea is not feasible anyway.

What I'm saying is that currently there is a big assumption in our codebase that there is a single place to fetch and push data to, which is either the remote URL or one of several specific settings (e.g., `lfs.url`).  That URL, the endpoint, is the source of where data should be uploaded to and downloaded from, which may point to one or more URLs elsewhere.

If your goal is to have a widely disparate location from which you fetch static data, and you want to fall back to non-static data at a totally different URL, then that's not going to work with the codebase as it stands without a lot of ugly special cases.  That isn't to say that it couldn't in the future, but we'd need to add support for that first.

Note that this can be reasonably easily solved if you have _some_ HTTP server that can host both the static and basic endpoints that are just redirects to the actual locations you care about, because you have one base URL and we'll just resolve the redirects as normal.  But if you want to have just two totally separate URLs, then it just won't work as things stand.  I'm sorry if that isn't what you were hoping for, but it's the limitation of the codebase as it stands.

mitar: > Moving this out to be a totally separate URL from the rest of the endpoint code means we have a giant exception to how things work that's special-cased for the static adapter.

Yea, this is why I would make it opt-in. Also, many things you listed just make no sense for static transfer adapter (it is download-only, it does not support authentication).

> For example, some people need to force the use of Basic authentication even though the server supports NTLM

OK, but this is all because of pushing this simple idea further than I am advocating. For me this is really about linking public HTTP files into a manifest so that I can version them. Whole authentication support for static transfers is something I am not sure about at all.

If we want feature parity with basic transfers then everything you are writing is true. But do we really want that? I personally do not have a need nor use case for that. But I do have an use case for some files being served from existing known public location and some files being served from local git LFS instance.

> we could specify two similar transfer adapters

Could we then have also `static-simple` with just simple HTTP manifest without any of fancy features, but allowing dedicated manifest URL?

> and you want to fall back to non-static data at a totally different URL

So if I understand you correctly, the main issue is with "automatic fallback". That fallback cannot work currently against two different endpoints, but if we do not have fallback (because it cannot yet work) then why having two URLs at all to begin with and not just use `lfs.url`? So it is not the issue of adapter, but of fallback between adapters?

mitar: I still do not get though how would you support that `lfs.url` can point directly to a manifest file or that it can point to git LFS API endpoint? And support fallback in this way? I understand that if it is the latter, we can do auto-discovery to find if manifest is there. But if we point `lfs.url` just to the manifest, how do we know where is the git LFS API endpoint?

bk2204: I think we would specify that by the `lfs.transfertypes` setting.  We could have `static-simple` where `lfs.url` points directly to the manifest and `static` where autodiscovery happens as part of a repository.

mitar: Hm, I think that given that we already have a transfer called `basic`, I feel having `static-simple` and `static` be a bit confusing. Like "simple" and "basic" seems very similar in meaning.

Maybe we should have "basic" and "static" (one you call `static-simple`) and then also `static-auto` for one with auto-discovery?

bk2204: Sure, that sounds fine.

mitar: OK, let me try to do a pass then with all the feedback you provided and let's see where we land.

ux-8 (approved):

mitar:
Sorry. I will not be continuing with this. So my main motivation was to support storing files larger than 2 GB which cannot be stored on GitHub somewhere else (on a HTTP server) and then point only for that file to that server. While for other files use existing GitHub LFS capabilities. The review pointed out that it is hard to do so and requested simplifications, which I agree would be an useful feature, but not really directly applicable to my needs. So I didn't get time to work on this further, nor I anticipate I will have time in the future. So I am closing this and somebody else can take it from here (and integrate comments above).

PR #4259, opened 2020-09-22
lfshttp: add support for socks5h proxies

### LINKED ISSUE #4258: Add support for socks5h proxies
I have started using git-lfs recently, and found, that accessing repositories through a socks5h server has become quite a headache. Git itself supports both socks5 and socks5h without problem, trying to resolve host names locally for socks5 and forwarding them to the socks server for socks5h. For what I'm doing, I need the latter, so I used to specify a socks5h proxy.
For some reason, golang chose not to support proxy URL starting with socks5h, only with socks5, but if you give it a socks5 URL, it actually does things the socks5h way and forwards all host names to the socks server. I found bug reports and change proposals to remedy this situation in golang, that were written two years ago. So I don't expect the situation to improve any time soon.
In the meantime, would it be possible to implement a workaround in git-lfs, that replaces socks5h URLs with socks5 URLs before handing them off to golang? Then I could set HTTPS_PROXY to socks5h://server:port and both git and git-lfs would do the same thing.

bk2204:
Go currently doesn't support the socks5h protocol, but it does know how to handle it if given a URL starting with socks5:// instead.  Let's rewrite the URLs internally to make things just magically work.

Fixes #4258.

### DIFF
--- a/lfshttp/proxy.go
+++ b/lfshttp/proxy.go
@@ -27,6 +27,10 @@ func proxyFromClient(c *Client) func(req *http.Request) (*url.URL, error) {
 			return nil, nil
 		}
 
+		if strings.HasPrefix(proxy, "socks5h://") {
+			proxy = strings.Replace(proxy, "socks5h://", "socks5://", 1)
+		}
+
 		cfg := &httpproxy.Config{
 			HTTPProxy:  proxy,
 			HTTPSProxy: proxy,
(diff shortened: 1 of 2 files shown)


chrisd8088 (approved):
Nice!  (I think the test is new from yesterday, too. 👍 )

koxu1996:
@bk2204 Latest _2.12.1_ does not include support for _socks5h_ - do you know when this PR will be part of official release?

bk2204:
We're planning to do a release in early December.

koxu1996:
Ok, thanks :smile:

PR #4260, opened 2020-09-22
ci: don't reinstall openssl with Homebrew

bk2204:
GitHub Actions ships its own version of the openssl Homebrew package that fails to reinstall with an EPERM error.  Since the reason we did these contortions was to relink gettext and we finally gave up and disabled gettext, let's just go back to a brew link --force and stop trying to reinstall.


chrisd8088 (approved):
👍

PR #4262, opened 2020-09-23
workflows/ci: build latest Git with SHA-256

bk2204:
We'd like to make sure that our code works with SHA-256 Git repositories, so when we build with the latest Git, let's test with SHA-256 repositories.


chrisd8088 (approved):
👍 << 8.

PR #4266, opened 2020-09-30
script/upload: avoid using Ruby's URI.escape

### LINKED ISSUE #4226: script/upload should not use URI.escape
**Describe the issue**
We're using `URI.escape` in `script/upload`, but that is deprecated and will cause warnings:

```
-e:1: warning: URI.escape is obsolete
-e:1: warning: URI.escape is obsolete
```
**System environment**
Debian amd64/sid and our ubuntu-latest CI images

**Additional context**
This will probably break in a future release, so we should fix it before our CI jobs mysteriously fail during a release.

bk2204:
Ruby's `URI.escape` is now deprecated because different components of a path require different escaping rules.  However, we still want to perform a similar escaping style, so let's do it ourselves with a `String#gsub` call.  We could use `CGI.escape`, but that encodes spaces as plus signs and it's unclear whether that's a desirable change to make here.  Note that there are no negative consequences for encoding more data than we need to, since the server will decode it for us.

Fixes #4226

### DIFF
--- a/script/upload
+++ b/script/upload
@@ -21,6 +21,10 @@ abort () {
   exit 2
 }
 
+uri_encode () {
+  ruby -e 'print ARGV[0].gsub(/[^A-Za-z0-9_.-]/) { |x| "%%%02x" % x.ord }' "$1"
+}
+
 curl () {
   if [ -n "$GITHUB_TOKEN" ]
   then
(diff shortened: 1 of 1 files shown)


chrisd8088 (approved):
LGTM.

PR #4269, opened 2020-10-06
Honor GIT_WORK_TREE

### LINKED ISSUE #4213: git lfs track does not honor GIT_WORK_TREE and GIT_DIR
**Describe the bug**
`git lfs track` does not honor `GIT_WORK_TREE` nor `GIT_DIR` environment variables as well as `--git-dir` and `--work-tree` options.

**To Reproduce**
```
mkdir foo bar
export GIT_WORK_TREE=foo GIT_DIR=bar
git init
git lfs track '*.xz'
```
`.gitattributes` file is produced in current directory and error message is printed:
> Tracking "*.xz"
> Error changing lockable file permissions: Error in git ls-files -z --others --cached --exclude-standard: exit status 128 fatal: not a git repository: 'bar'
>
> Errors logged to <…>/testtwo/bar/lfs/logs/20200814T181926.04705318.log
> Use `git lfs logs last` to view the log.

**Expected behavior**
`.gitattributes` file in `foo` directory.

**System environment**
Linux, git version 2.28.0, git-lfs/2.11.0 (GitHub; linux amd64; go 1.14.4)

**Output of `git lfs env`**
```
git-lfs/2.11.0 (GitHub; linux amd64; go 1.14.4)
git version 2.28.0

LocalWorkingDir=/home/hideo/testtwo/foo
LocalGitDir=/home/hideo/testtwo/bar
LocalGitStorageDir=/home/hideo/testtwo/bar
LocalMediaDir=/home/hideo/testtwo/bar/lfs/objects
LocalReferenceDirs=
TempDir=/home/hideo/testtwo/bar/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/home/hideo/testtwo/bar/lfs
AccessDownload=none
AccessUpload=none
GIT_DIR=bar
GIT_NOTES_REF=refs/notes/build
GIT_WORK_TREE=foo
```

bk2204:
Currently, we don't honor GIT_WORK_TREE like Git does.  Git implicitly changes into the working tree of the repository if this variable is set, but we currently don't do that.

Let's fix that by mimicking Git's behavior so this works correctly, which necessitates transforming the various environment variables we pass to Git so that they are valid when we change directories.

These commits are logical, bisectable, and independent and have sane commit messages.

Fixes #4213
/cc @Mr-Tao as reporter

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -316,15 +316,49 @@ func requireInRepo() {
 // that it not be bare. If it is bare (or the state of the repository could not
 // be determined), this function will terminate the program.
 func requireWorkingCopy() {
+	if cfg.LocalWorkingDir() == "" {
+		Print("This operation must be run in a work tree.")
+		os.Exit(128)
+	}
+}
+
+func setupRepository() {
+	requireInRepo()
 	bare, err := git.IsBare()
 	if err != nil {
 		ExitWithError(errors.Wrap(
 			err, "fatal: could not determine bareness"))
 	}
 
-	if bare {
-		Print("This operation must be run in a work tree.")
-		os.Exit(128)
+	if !bare {
+		changeToWorkingCopy()
+	}
+}
+
+func setupWorkingCopy() {
+	requireInRepo()
+	requireWorkingCopy()
+	changeToWorkingCopy()
+}
+
+func changeToWorkingCopy() {
+	workingDir := cfg.LocalWorkingDir()
+	cwd, err := tools.Getwd()
+	if err != nil {
+		ExitWithError(errors.Wrap(
+			err, "fatal: could not determine current working directory"))
+	}
+	cwd, err = filepath.EvalSymlinks(cwd)
+	if err != nil {
+		ExitWithError(errors.Wrap(
+			err, "fatal: could not canonicalize current working directory"))
+	}
+
+	// If the current working directory is not within the repository's
+	// working directory, then let's change directories accordingly.  This
+	// should only occur if GIT_WORK_TREE is set.
+	if !(strings.HasPrefix(cwd, workingDir) && (cwd == workingDir || (len(cwd) > len(workingDir) && cwd[len(workingDir)] == os.PathSeparator))) {
+		os.Chdir(workingDir)
 	}
 }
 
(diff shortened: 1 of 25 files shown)


lfs/lfs.go
```diff
@@ -72,10 +76,15 @@ func Environ(cfg *config.Configuration, manifest *tq.Manifest) []string {
 	}
 
 	for _, e := range osEnviron {
-		if !strings.Contains(strings.SplitN(e, "=", 2)[0], "GIT_") {
+		key := strings.SplitN(e, "=", 2)[0]
+		if !strings.Contains(key, "GIT_") {
```
chrisd8088, line 80: Just looking at this with fresh eyes, I wonder if `strings.HasPrefix()` would be more appropriate ... are there environment variables where `GIT_` would not be at the start of the name?

commands/commands.go
```diff
@@ -315,15 +315,49 @@ func requireInRepo() {
...
+
+func setupWorkingCopy() {
+	requireInRepo()
+	requireWorkingCopy()
+	changeToWorkingCopy()
+}
+
+func changeToWorkingCopy() {
+	workingDir := cfg.LocalWorkingDir()
+	cwd, err := os.Getwd()
```
chrisd8088, line 345: IIUC, the path which we get for the current working directory will be compared down below (after running `filepath.EvalSymlinks()` on it) to the output from `cfg.LocalWorkingDir()` which I believe [ultimately](https://github.com/git-lfs/git-lfs/blob/68a8f8a49d834c91c9e9d39bc53c6fc065122dbd/config/config.go#L368) comes from `git.GitAndRootDirs()`which as of 23fd6c5973b69694776e630e11eda48bfe1f838c is now always going to call the new `tools.CanonicalizePath()` which will call `TranslateCygwinPath()`.

So in order to also run `TranslateCygwinPath()` on the output from `os.Getwd()`, perhaps this should instead call `tools.Getwd()`?

chrisd8088 (approved):
This is great, thank you!  I had a couple of minor comments, but otherwise LGTM!

lfs/lfs.go
bk2204: Yeah, this is a behavior change, but I believe it's a prudent change to make.

commands/commands.go
bk2204: Good point.  Updated.

Issue #4275, opened 2020-10-08
Add --above parameter to git-lfs migrate import


EliRibble:
**Describe the issue**
Currently `git-lfs migrate info` has a parameter, `--above=<size>` which only considers files whose individual filesize is above a given size. `git-lfs migrate import` does not have this parameter but relies instead on `--include` and `--exclude` file paths. For very large repositories this can be too onerous to specify. Instead it would be useful to simply migrate all files over a specific size, say 100K or 1M.

**System environment**
N/A

**Output of `git lfs env`**
N/A

chrisd8088:
Resolved in #4276 ... thanks again for all the work on the patch submission, it's very much appreciated!

EliRibble:
Thank you, I'm extremely impressed at how you managed the process with care, respect and kindness.

PR #4276, opened 2020-10-09
Add --above parameter to 'migrate import'.

EliRibble:
The purpose is to allow users to tell git-lfs to migrate
all files over a particular threshold without having to determine
the various paths and patterns that would. This is very useful
for large code bases with inconsistent history of adding large
files in different places. This option makes it possible to
supply `git-lfs migrate import --everything --above=1Mb` and
have git-lfs do the convenient thing and rewrite history. 

Issue #4275

### DIFF
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -31,6 +31,11 @@ var (
 	// of updated references.
 	migrateSkipFetch bool
 
+	// migrateImportAboveFmt indicates the presence of the --above=<size>
+	// flag and instructs 'git lfs migrate import' to import all files
+	// above the provided size.
+	migrateImportAboveFmt string
+
 	// migrateEverything indicates the presence of the --everything flag,
 	// and instructs 'git lfs migrate' to migrate all local references.
 	migrateEverything bool
@@ -365,6 +370,7 @@ func init() {
 	info.Flags().StringVar(&migrateInfoUnitFmt, "unit", "", "--unit=<unit>")
 
 	importCmd := NewCommand("import", migrateImportCommand)
+	importCmd.Flags().StringVar(&migrateImportAboveFmt, "above", "", "--above=<n>")
 	importCmd.Flags().BoolVar(&migrateVerbose, "verbose", false, "Verbose logging")
 	importCmd.Flags().StringVar(&objectMapFilePath, "object-map", "", "Object map file")
 	importCmd.Flags().BoolVar(&migrateNoRewrite, "no-rewrite", false, "Add new history without rewriting previous")
--- a/commands/command_migrate_import.go
+++ b/commands/command_migrate_import.go
@@ -17,6 +17,7 @@ import (
 	"github.com/git-lfs/git-lfs/lfs"
 	"github.com/git-lfs/git-lfs/tasklog"
 	"github.com/git-lfs/git-lfs/tools"
+	"github.com/git-lfs/git-lfs/tools/humanize"
 	"github.com/git-lfs/gitobj/v2"
 	"github.com/spf13/cobra"
 )
@@ -133,12 +134,19 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 	gitfilter := lfs.NewGitFilter(cfg)
 
 	var fixups *gitattr.Tree
+	above, err := humanize.ParseBytes(migrateImportAboveFmt)
+	if err != nil {
+		ExitWithError(errors.Wrap(err, "fatal: cannot parse --above=<n>"))
+	}
 
 	migrate(args, rewriter, l, &githistory.RewriteOptions{
 		Verbose:           migrateVerbose,
 		ObjectMapFilePath: objectMapFilePath,
 		BlobFn: func(path string, b *gitobj.Blob) (*gitobj.Blob, error) {
 			if filepath.Base(path) == ".gitattributes" {
+				return b, nil
+			}
+			if (above > 0) && (uint64(b.Size) < above) {
 				return b, nil
 			}
 
--- a/t/t-migrate-import.sh
+++ b/t/t-migrate-import.sh
@@ -409,6 +409,31 @@ begin_test "migrate import (include/exclude ref with filter)"
 )
 end_test
 
+begin_test "migrate import (above)"
+(
+  set -e
+  setup_single_local_branch_untracked
+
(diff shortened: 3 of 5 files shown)


commands/command_migrate_import.go
```diff
@@ -133,6 +134,10 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 	gitfilter := lfs.NewGitFilter(cfg)
 
 	var fixups *gitattr.Tree
+	above, err := humanize.ParseBytes(migrateImportAboveFmt)
+	if err != nil {
+		ExitWithError(errors.Wrap(err, "cannot parse --above=<n>"))
```
chrisd8088, line 139: All the other error messages above this include the prefix `"fatal: "`, so I think it would be nice to make this do the same.

chrisd8088 (changes requested):
Hey, thanks very much for this PR, it looks good to me, with one minor caveat regarding the wording of the new error message.  If you don't mind tweaking that one thing, I think we're good to merge this.

Thanks again very much for the contribution!

chrisd8088:
Oh, one other things which just occurred to me -- we should also add a test in `t/t-migrate-import.sh` for this new option, just something simple which checks for files above and below and at the threshold given by the `--above` option.

EliRibble:
Sorry, I'm not sure how to run the tests locally. The contributing documentation seems to say that "make test" is the way to run tests, but for me it just downloads a bunch of things and prints that all the tests are passing. That doesn't seem right considering how many CI builds are failing - how do I go about running the tests locally so I can fix this PR?

chrisd8088:
Thanks for the test code, that's great!  I usually run a specific test script by doing something like this in my working tree:
```
$ cd t
$ ./t-migrate-import.sh
```

You can also use `cd t && make test` to run the whole suite of test scripts.  Sometimes I find I need to `rm -r t/remote` first to clean out stale test data from a previous run.

chrisd8088 (commented):
I had two last comments, both on the test -- one regarding whether there are some extra debugging `echo` statements which could be removed, and one re the possible addition of a `refute`-type statement to confirm that the smaller file was not imported in the test.

Thoughts?  And thank you again so much for pursuing this!

t/t-migrate-import.sh
```diff
@@ -409,6 +409,30 @@ begin_test "migrate import (include/exclude ref with filter)"
...
+  txt_main_oid="$(calc_oid "$(git cat-file -p "refs/heads/main:a.txt")")"
+
+  echo "md_main_oid" $md_main_oid
+  echo "txt_main_oid" $txt_main_oid
+  git lfs migrate import --above 121B
+  # Ensure that 'a.md', whose size is above our 121 byte threshold
+  # was converted into a git-lfs pointer by the migration.
+  assert_local_object "$md_main_oid" "140"
+  assert_pointer "refs/heads/main" "a.md" "$md_main_oid" "140"
+
```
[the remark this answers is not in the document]
EliRibble: thanks, great ideas, added.

t/t-migrate-import.sh
```diff
@@ -409,6 +409,30 @@ begin_test "migrate import (include/exclude ref with filter)"
...
+begin_test "migrate import (above)"
+(
+  set -e
+  setup_single_local_branch_untracked
+
+  md_main_oid="$(calc_oid "$(git cat-file -p "refs/heads/main:a.md")")"
+  txt_main_oid="$(calc_oid "$(git cat-file -p "refs/heads/main:a.txt")")"
+
+  echo "md_main_oid" $md_main_oid
+  echo "txt_main_oid" $txt_main_oid
```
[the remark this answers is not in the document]
EliRibble: Absolutely, they can be removed. Thanks for catching that.

commands/command_migrate_import.go
EliRibble: Added.

commands/command_migrate.go
```diff
@@ -31,6 +31,11 @@ var (
 	// of updated references.
 	migrateSkipFetch bool
 
+	// migrateImportAboveFmt indicates the presence of the --amount=<size>
```
chrisd8088, line 34: One very minor thing I'm just noticing now is that this comment should probably say `--above=<size>` instead of `--amount=<size>`, but I won't bother to make approval conditional on this change as it's not too important, just a nice-to-have fixup.

t/t-migrate-import.sh
```diff
@@ -409,6 +409,30 @@ begin_test "migrate import (include/exclude ref with filter)"
...
+  assert_local_object "$md_main_oid" "140"
+  assert_pointer "refs/heads/main" "a.md" "$md_main_oid" "140"
+  refute_pointer "refs/heads/main" "a.txt" "$md_main_oid" "120"
+  refute_local_object "$txt_main_oid" "120"
+
+  # The migration should have identified that *.md files are now
+  # tracked because it migrated a.md
+  main_attrs="$(git cat-file -p "$main:.gitattributes")"
+
+  echo "$main_attrs" | grep -q "*.md filter=lfs diff=lfs merge=lfs"
```
chrisd8088, line 432: Thanks so much for making the other changes above in this test!

The only other nice-to-have fixup I can see here would be to add a line after this one which checks that there is no `*.txt` entry in `.gitattributes`.  I believe some of the other tests do this [already](https://github.com/git-lfs/git-lfs/blob/1f4e688e32bedfa067bb34b88ca52995387b3d02/t/t-migrate-import.sh#L109) with a `-vq` line like:
```suggestion
  echo "$main_attrs" | grep -q "*.md filter=lfs diff=lfs merge=lfs"
  echo "$main_attrs" | grep -vq "*.txt filter=lfs diff=lfs merge=lfs"
```

chrisd8088 (approved):
I'm going to approve this as it looks pretty darn good to me now.  There are a couple of notes I made, one in particular to add a further negative check to the new test, but neither is critical.  If you get a chance to make those two changes, let me know!

Thanks again very much for pursuing this, it's looking great!

commands/command_migrate.go
EliRibble: done

t/t-migrate-import.sh
EliRibble: done

chrisd8088 (approved):
This looks perfect -- I'm going to merge in the latest `master` changes, and then merge this PR.  👍

PR #4282, opened 2020-10-12
config: map missing port to default for HTTP key lookups

### LINKED ISSUE #4281: Proxy URL does not use default ports if omitted
**Describe the bug**
In our setup the LFS server returns URLs containing the port (HTTPS, Port 443). If they are matched against URLs from the git config they don't match because the git config omits the port while the URL from the server does not. This behavior does not match the one described in rule 3 in the [git documentation](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httplturlgt).

The relevant comparison is done here: https://github.com/git-lfs/git-lfs/blob/effc2b1b64c9594f0159f14f5214fa7470e62596/config/url_config.go#L116 
And I think the problem is with the `net.url.Port()` which returns an empty string for a missing port instead of the default port for said scheme (see https://golang.org/pkg/net/url/#URL.Port)

Introduced in #3392.

**To Reproduce**
1. Use `GIT_TRACE=1 git lfs fetch` to examine the advertised URLs from your test server.
2. Define a global proxy server that DOES NOT WORK (something like localhost) using `http.proxy=localhost`
3. Test that a `GIT_TRACE=1 git lfs fetch` fails with the config
4. Add an exception for the proxy. If your server includes the proxy, omit it here, otherwise include the default proxy for the scheme:
   `git config http.<url>.proxy ''`
5. `GIT_TRACE=1 git lfs fetch` will fail
6. `git config --get-urlmatch http.proxy <url from step 1>` returns the correct proxy (None) and not the proxy from step 2

Example with my setup:
1. Server response: `https://my.company.int:443/artifactory/repo/objects/...`
2. `git config http.proxy localhost`
3. It doesn't work
4. `git config http.https://my.company.int/.proxy ''`
5. Still fails
6. `git config --get-urlmatch http.proxy https://my.company.int:443` -> empty (expected)

**Expected behavior**
Both LFS and git should have the same behavior.

**System environment**
Not relevant

**Output of `git lfs env`**

```
git-lfs/2.12.0 (GitHub; windows amd64; go 1.14.7; git dad7b9e6)
git version 2.28.0.windows.1

Endpoint=https://my.company.int/artifactory/repo (auth=none)
  SSH=my.company.int:artifactory/repo
Endpoint (origin_https)=https://my.company.int/artifactory/api/lfs/repo (auth=basic)
LocalWorkingDir=C:\repo
LocalGitDir=C:\repo\.git
LocalGitStorageDir=C:\repo\.git
LocalMediaDir=C:\repo\.git\lfs\objects
LocalReferenceDirs=
TempDir=C:\repo\.git\lfs\tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=C:\repo\.git\lfs
AccessDownload=none
AccessUpload=none
GIT_CURL_VERBOSE=1
GIT_TRACE=0
GIT_TRANSFER_TRACE=1
```

bk2204:
When Git performs a lookup for a key with urlmatch semantics, it matches the default port if none is specified in the URL.  However, our code currently doesn't do that, instead having a missing port match only another missing port entry.

Let's fix the behavior to match Git's by mapping a missing port to the default port.  We add tests for checking both URLs (the config one and the one we're sending data to) to make sure that our code is robust.

Fixes #4281
/cc @slietzau as reporter

### DIFF
--- a/config/url_config_test.go
+++ b/config/url_config_test.go
@@ -8,17 +8,19 @@ import (
 
 func TestURLConfig(t *testing.T) {
 	u := NewURLConfig(EnvironmentOf(MapFetcher(map[string][]string{
-		"http.key":                           []string{"root", "root-2"},
-		"http.https://host.com.key":          []string{"host", "host-2"},
-		"http.https://user@host.com/a.key":   []string{"user-a", "user-b"},
-		"http.https://user@host.com.key":     []string{"user", "user-2"},
-		"http.https://host.com/a.key":        []string{"host-a", "host-b"},
-		"http.https://host.com:8080.key":     []string{"port", "port-2"},
-		"http.https://host.com/repo.git.key": []string{".git"},
-		"http.https://host.com/repo.key":     []string{"no .git"},
-		"http.https://host.com/repo2.key":    []string{"no .git"},
-		"http.http://host.com/repo.key":      []string{"http"},
-		"http.https://host.*/a.key":          []string{"wild"},
+		"http.key":                                []string{"root", "root-2"},
+		"http.https://host.com.key":               []string{"host", "host-2"},
+		"http.https://user@host.com/a.key":        []string{"user-a", "user-b"},
+		"http.https://user@host.com.key":          []string{"user", "user-2"},
+		"http.https://host.com/a.key":             []string{"host-a", "host-b"},
+		"http.https://host.com:8080.key":          []string{"port", "port-2"},
+		"http.https://host.com/repo.git.key":      []string{".git"},
+		"http.https://host.com/repo.key":          []string{"no .git"},
+		"http.https://host.com/repo2.key":         []string{"no .git"},
+		"http.http://host.com/repo.key":           []string{"http"},
+		"http.https://host.com:443/repo3.git.key": []string{"port"},
+		"http.ssh://host.com:22/repo3.git.key":    []string{"ssh-port"},
+		"http.https://host.*/a.key":               []string{"wild"},
 	})))
 
 	getOne := map[string]string{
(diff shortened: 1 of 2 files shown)


config/url_config_test.go
```diff
@@ -8,17 +8,18 @@ import (
...
+		"http.https://user@host.com/a.key":        []string{"user-a", "user-b"},
+		"http.https://user@host.com.key":          []string{"user", "user-2"},
+		"http.https://host.com/a.key":             []string{"host-a", "host-b"},
+		"http.https://host.com:8080.key":          []string{"port", "port-2"},
+		"http.https://host.com/repo.git.key":      []string{".git"},
+		"http.https://host.com/repo.key":          []string{"no .git"},
+		"http.https://host.com/repo2.key":         []string{"no .git"},
+		"http.http://host.com/repo.key":           []string{"http"},
+		"http.https://host.com:443/repo3.git.key": []string{"port"},
+		"http.https://host.*/a.key":               []string{"wild"},
```
chrisd8088, line 22: Since the new `portForURL()` function checks for the URLs with the SSH protocol, might it be advisable to add test or two here which use `ssh://...` format URLs?

chrisd8088 (approved):
LGTM; I only wondered about adding some tests for URLs of the `ssh://` form, since there's explicit logic now depending on that protocol's default port.

config/url_config_test.go
bk2204: I included it only for completeness, since I'm pretty sure we only use this code path for `http.*` keys, which don't work for SSH requests.  But I'll add some tests here nevertheless, in case we use it in a situation like the credential codepath.

PR #4291, opened 2020-10-25
tools: use ERROR_SHARING_VIOLATION const from golang.org/x/sys/windows

tklauser:
Use ERROR_SHARING_VIOLATION defined in the golang.org/x/sys/windows
package instead of duplicating it in robustio_windows.go.


chrisd8088:
Thanks very much for the patch; it looks like it tidies this detail up nicely and passes the CI suite, including with Go 1.11, so I'm inclined to merge it as-is unless another @git-lfs/core member has concerns.  Thanks again!

slonopotamus:
I wonder why I didn't find it when was writing this code.

bk2204:
No objections here.  It's possible that we didn't have that package available when this was written which may have been why we didn't notice it.

chrisd8088 (approved):
Very nice cleanup, thank you!

PR #4307, opened 2020-11-16
fs: don't panic when using a too-short object ID to push

### LINKED ISSUE #4303: Git LFS crash when push with empty object id
**Describe the bug**
When run `git lfs push --oid ''` git lfs would crash with a stack trace

**To Reproduce**
1. Try passing an empty object ID into git LFS push

**Expected behavior**
Git LFS gracefully exit, prompting user that an empty object IDs was not allowed.

Better yet if we can just filter out all of the empty object IDs among the provided objects.

**System environment**
Centos7

**Output of `git lfs env`**
The output of running `git lfs env` as a code block.
```
# Omit as its not very relevant
git-lfs/2.12.0 (GitHub; linux amd64; go 1.15.2)
git version 2.27.0.284.g2915947361
```

**Additional context**
Any other relevant context about the problem here.

```
> git lfs push --object-id origin ''
panic: runtime error: slice bounds out of range [:2] with length 0

goroutine 1 [running]:
github.com/git-lfs/git-lfs/fs.(*Filesystem).localObjectDir(0xc00024e000, 0x7ffc63dc01ca, 0x0, 0xc000254030, 0x7fb94c5a9048)
        /builddir/build/BUILD/git-lfs-2.12.0/fs/fs.go:123 +0xef
github.com/git-lfs/git-lfs/fs.(*Filesystem).ObjectPath(0xc00024e000, 0x7ffc63dc01ca, 0x0, 0xc000254030, 0x6, 0xc00021ab70, 0xc000220f10)
        /builddir/build/BUILD/git-lfs-2.12.0/fs/fs.go:68 +0x5a
github.com/git-lfs/git-lfs/lfs.(*GitFilter).ObjectPath(...)
        /builddir/build/BUILD/git-lfs-2.12.0/lfs/gitfilter.go:21
github.com/git-lfs/git-lfs/commands.uploadsWithObjectIDs(0xc00026c0b0, 0xc000098fa0, 0x1, 0x2)
        /builddir/build/BUILD/git-lfs-2.12.0/commands/command_push.go:80 +0xce
github.com/git-lfs/git-lfs/commands.pushCommand(0xc0000db180, 0xc000098f90, 0x2, 0x3)
        /builddir/build/BUILD/git-lfs-2.12.0/commands/command_push.go:52 +0xf2
github.com/spf13/cobra.(*Command).execute(0xc0000db180, 0xc000098f60, 0x3, 0x3, 0xc0000db180, 0xc000098f60)
        /builddir/go/pkg/mod/github.com/spf13/cobra@v0.0.3/command.go:766 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0xc000092000, 0xc000139f08, 0x1, 0x1)
        /builddir/go/pkg/mod/github.com/spf13/cobra@v0.0.3/command.go:852 +0x2fe
github.com/spf13/cobra.(*Command).Execute(...)
        /builddir/go/pkg/mod/github.com/spf13/cobra@v0.0.3/command.go:800
github.com/git-lfs/git-lfs/commands.Run(0x10)
        /builddir/build/BUILD/git-lfs-2.12.0/commands/run.go:103 +0x318
main.main()
        /builddir/build/BUILD/git-lfs-2.12.0/git-lfs.go:33 +0xec
```

bk2204:
Previously, we'd panic when trying to slice into the object ID if the user provided a too-short object ID on the command line.  Instead, let's fix this so we reject too-short object IDs with an error.

Fixes #4303 
/cc @sluongng as reporter

### DIFF
--- a/t/t-push.sh
+++ b/t/t-push.sh
@@ -112,10 +112,10 @@ begin_test "push"
 )
 end_test
 
-# sets up the tests for the next few push --all tests
+# helper used by the next few push --all tests to set up their repos
 push_all_setup() {
   suffix="$1"
-  reponame="$(basename "$0" ".sh")-all"
+  reponame="$(basename "$0" ".sh")-all-$suffix"
   content1="initial"
   content2="update"
   content3="branch"
@@ -129,10 +129,7 @@ push_all_setup() {
   oid5=$(calc_oid "$content5")
   extraoid=$(calc_oid "$extracontent")
 
-  # if the local repo exists, it has already been bootstrapped
-  [ -d "push-all" ] && exit 0
-
-  clone_repo "$reponame" "push-all"
+  clone_repo "$reponame" "push-all-$suffix"
   git config "lfs.$(repo_endpoint "$GITSERVER" "$reponame").locksverify" true
   git lfs track "*.dat"
 
@@ -254,7 +251,7 @@ begin_test "push --all (1 ref arg)"
   [ $(grep -c "^push " < push.log) -eq 3 ]
 
   git lfs push --all origin branch 2>&1 | tee push.log
-  grep "3 files" push.log
+  [ $(grep -c "Uploading LFS objects: 100% (3/3)" push.log) -eq 1 ]
   assert_server_object "$reponame-$suffix" "$oid1"
   assert_server_object "$reponame-$suffix" "$oid2"
   assert_server_object "$reponame-$suffix" "$oid3"
@@ -281,8 +278,8 @@ begin_test "push --all (1 ref arg)"
   grep "push $oid3 => file1.dat" push.log
   [ $(grep -c "^push " push.log) -eq 3 ]
 
-  git push --all origin branch 2>&1 | tee push.log
-  grep "5 files, 1 skipped" push.log # should be 5?
+  git push origin branch 2>&1 | tee push.log
+  [ $(grep -c "Uploading LFS objects: 100% (3/3)" push.log) -eq 1 ]
   assert_server_object "$reponame-$suffix-2" "$oid2"
   assert_server_object "$reponame-$suffix-2" "$oid3"
   refute_server_object "$reponame-$suffix-2" "$oid4"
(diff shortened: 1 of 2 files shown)


t/t-push.sh
```diff
@@ -820,3 +820,16 @@ begin_test "push custom reference"
...
 end_test
+
+begin_test "push --object-id (invalid value)"
+(
+  set -e
+
+  push_all_setup "everything"
+
+  git lfs push --object-id origin '' 2>&1 | tee push.log
+  git lfs push --object-id origin "${oid1:1:4}" 2>&1 | tee -a push.log
```
chrisd8088, line 831: I was intrigued by how this works, because the `${oid1:1:4}` would seem to me to be generating a 4-character string starting with the 2nd character in `$oid1`, and that didn't jive with the `len(oid) < 4` check added in `ObjectPath()`.

Well, therein lies a tale.  I've put up two extra commits in my `chrisd8088/empty-oid-push` [branch](https://github.com/chrisd8088/git-lfs/tree/empty-oid-push) which explain it in detail, but the gist is that `push_all_setup()` appears to have been accidentally short-circuiting all the tests which call it (except the first one), and when that bug is fixed, this new test starts failing because it generates a 4-character string which doesn't trigger a second `too short object ID` message from `git lfs push`, as I'd idly puzzled over in the first place.

If you're willing to pull those two commits (f9381fa37cb25af17c6d90f4eea054770cb5692d and 63a0b7a388a35fd507468d2474ef63d0c1cc30b2) into this PR, I think that would resolve all my concerns here and get the test fully working.

chrisd8088 (changes requested):
This change seems great and will resolve the reported issue nicely, thank you!

I hope I've explained my rather unexpected findings and proposed additions to the PR to make the test run completely ... not at all the speedy 👍 I'd anticipated when I started looking at this!  :-)

t/t-push.sh
bk2204: Good catch.  Done.

chrisd8088 (approved):
Thanks for pulling those commits in!  I should probably have used `push_all_setup "invalid-oid"` instead of having the redundant extra `push` in there, but it's not material to the fixes or tests, so I'll leave it for now.

PR #4308, opened 2020-11-16
.github/workflows: remove special ::add-path commands

bk2204:
We had these commands to work around a problem with an earlier GitHub Actions image not finding the expected Git binaries, but they are no longer necessary.  In addition, GitHub Action now fails our commands because honoring these options causes a security problem, so our CI jobs are failing.  Let's remove them since they're no longer needed nor useful.


chrisd8088 (approved):
👍  Thanks!

PR #4318, opened 2020-11-23
Make git lfs migrate import handle missing extensions

### LINKED ISSUE #4312: git lfs migrate import with --above does not always track file
**Describe the bug**
I needed to migrate a project where a file and a directory had the same name and I only wanted to migrate the file.

This does not appear to be supported by `git-lfs migrate` today so I built it from source to use the `--above` flag. Initially it appeared to work just fine as the file in question was migrated but looking in `.gitattributes` it is not being tracked correctly.

The file in question has no file extension.

**To Reproduce**
I'm not sure how to reproduce it but it's using OpenCV with both iOS (https://github.com/opencv/opencv/releases/download/4.5.0/opencv-4.5.0-ios-framework.zip) and Android (https://github.com/opencv/opencv/releases/download/4.5.0/opencv-4.5.0-android-sdk.zip). I would like to migrate the `opencv2` binary in the iOS framework but not the `opencv2` directory in the Android SDK.

**Expected behavior**
I would expect the binary file to be tracked.

**System environment**
`macOS 11.0.1`
```
credential.helper=osxkeychain
user.name=Nicklas Ansman Giertz
user.email=nicklas@ansman.se
user.signingkey=94C20A317A743C13
core.excludesfile=/Users/nicklas/.gitignore
core.editor=vim
core.autocrlf=input
color.ui=auto
alias.feature-start=!zsh -c 'git branch feature/$0 master && git checkout feature/$0'
alias.feature-finish=!zsh -c 'export FEATURE=`git rev-parse --symbolic-full-name --abbrev-ref HEAD` && git checkout master && git merge $FEATURE'
alias.rev=describe --tags --dirty --always --long
push.default=current
filter.media.clean=git-media-clean %f
filter.media.smudge=git-media-smudge %f
difftool.sourcetree.cmd=opendiff "$LOCAL" "$REMOTE"
difftool.sourcetree.path=
mergetool.sourcetree.cmd=/Applications/Sourcetree.app/Contents/Resources/opendiff-w.sh "$LOCAL" "$REMOTE" -ancestor "$BASE" -merge "$MERGED"
mergetool.sourcetree.trustexitcode=true
mergetool.keepbackup=false
credential.helper=osxkeychain
rerere.enabled=true
http.cookiefile=/Users/nicklas/.gitcookies
filter.lfs.required=true
init.templatedir=~/.git_templates
pager.branch=false
commit.template=/Users/nicklas/.stCommitMsg
pull.rebase=false
core.filemode=true
core.bare=false
core.logallrefupdates=true
core.ignorecase=true
core.precomposeunicode=true
```

**Output of `git lfs env`**
```
git-lfs/2.12.0 (GitHub; darwin amd64; go 1.15.5; git 19e217f6)
git version 2.29.2

Endpoint=<redacted> (auth=none)
  SSH=<redacted>
LocalWorkingDir=<redacted>
LocalGitDir=<redacted>
LocalGitStorageDir=<redacted>
LocalMediaDir=<redacted>
LocalReferenceDirs=
TempDir=<redacted>
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=<redacted>
AccessDownload=none
AccessUpload=none
```

bk2204:
Currently, if we have a file and it contains an extension, we'll add that extension to the list of paths that should be placed in .gitattributes.  However, if the file does not contain an extension, it won't be listed at all.  As a result, we may convert a file flagged by the --above option but never cause Git to filter it properly.

Let's fix this by writing an attribute into the repository with the path relative to the root if we don't contain an extension.  This means that people can store typical Unix binaries, which lack extensions, as LFS files more easily.

In addition, there is a commit to clean up a thinko in the existing test.

/cc @ansman as reporter
Fixes #4312

### DIFF
--- a/commands/command_migrate_import.go
+++ b/commands/command_migrate_import.go
@@ -172,6 +172,8 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 
 			if ext := filepath.Ext(path); len(ext) > 0 {
 				exts.Add(fmt.Sprintf("*%s filter=lfs diff=lfs merge=lfs -text", ext))
+			} else {
+				exts.Add(fmt.Sprintf("/%s filter=lfs diff=lfs merge=lfs -text", path))
 			}
 
 			return &gitobj.Blob{
--- a/t/fixtures/migrate.sh
+++ b/t/fixtures/migrate.sh
@@ -95,6 +95,8 @@ setup_local_branch_with_nested_gitattrs() {
 setup_single_local_branch_untracked() {
   set -e
 
+  local name="${1:-a.md}"
+
   reponame="single-local-branch-untracked"
 
   remove_and_create_local_repo "$reponame"
@@ -102,10 +104,10 @@ setup_single_local_branch_untracked() {
   git commit --allow-empty -m "initial commit"
 
   base64 < /dev/urandom | head -c 120 > a.txt
-  base64 < /dev/urandom | head -c 140 > a.md
+  base64 < /dev/urandom | head -c 140 > "$name"
 
-  git add a.txt a.md
-  git commit -m "add a.{txt,md}"
+  git add a.txt "$name"
+  git commit -m "add a.txt and $name"
 }
 
 # setup_single_local_branch_tracked creates a repository as follows:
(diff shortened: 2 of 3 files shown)


ansman:
Seems like a great solution to me 👍

chrisd8088 (approved):
Looks great, thanks for the extra tuneup on this!

PR #4328, opened 2020-12-01
Drop obsolete OS support

bk2204:
This PR drops support for CentOS 6 and Debian 8, which are no longer publicly supported upstream.  While various versions do have extended security support available, we can't access it, and it doesn't seem like a good idea (or responsible to the rest of the Internet) for us to continue to build packages on versions that are known to be insecure.

This also makes our CI runs significantly faster, since we currently build many packages for CentOS 6 that we can now avoid building.  We've shaved 15 minutes off of our build time for Linux packages.

We also add some additional distros that are not in the file.  Note that we don't support Fedora 33 because packagecloud.io does not list it as a supported version at this time.

### DIFF
--- a/docker/run_dockers.bsh
+++ b/docker/run_dockers.bsh
@@ -58,7 +58,7 @@ done
 
 if [[ ${#IMAGES[@]} == 0 ]]; then
   # If you change this list, change script/upload as well.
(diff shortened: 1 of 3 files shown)


chrisd8088 (approved):
LGTM, thanks.

Issue #4330, opened 2020-12-02
git lfs migrate import fails if two refs have the same name but are case insensitive


IAmHughes:
**Describe the bug**
When working in a repository that has large files, performining `git lfs migrate import --everything --include <extensions>` will go through every `ref` and rewrite history to create lfs pointers. However, due to git on Windows (and I believe Mac) not recognizing case sensitivity, two refs named: `feature1` and `Feature1` are treated as a single ref by the `git lfs migrate import` command.

The problem leads to having a ref that has been rewritten, but then another that has not and needs to be cleaned up manually by editing the `packed-refs` file.

**To Reproduce**
1. Star with a repository that has large files in two branches that are not in LFS, and the two branches with the same name but different case sensitivity (i.e. `feature1` and `Feature1`
2. Run `git lfs migrate import --everything --include <extension>`
3. Run `git log --numstat --all -- <ref/large-file>`
4. Note two shas with the same timestamp and message, however one has been rewritten to a lfs pointer (3 lines) and the other hasn't

**Expected behavior**
When running a `git lfs migrate --everything`, I would expect all refs to be migrated appropriately.

**System environment**
- Windows 10
- Git Bash for Windows (git v2.27, may have been v2.29)

**Output of `git lfs env`**
This was ran on a machine I don't have access too, I can't do this command unfortunately.

**Additional context**
Case sensitivity seems like the root of the issue, though I'm unsure of a path to fix it.

/cc @larsxschneider as you found this for me while debugging.

bk2204:
Hey,

This is unfortunately a limitation of Git on Windows.  Git doesn't support multiple refs differing in case if your file system doesn't preserve them.  While they can happen to exist in the `packed-refs` file, as soon as the ref needs to be rewritten, it will be rewritten as a loose ref in the file system, which takes precedence over the old value in the `packed-refs` file.  Since it's not possible to represent both refs in the file system, it's not possible to rewrite both refs, and when Git LFS tries to do so, it rewrites only the one.  This would likely also happen if you used `git filter-branch`.

If you need to support refs differing only in case, you need to use a case-sensitive file system or (on Windows) directory.  On Windows, you can do this by using WSL as well.

larsxschneider:
@bk2204 Could it be a solution to delete the `packed-refs` file on Windows if `--everything` is passed?

bk2204:
[...]

PR #4332, opened 2020-12-02
migrate import: warn about refs on case insensitive file systems

larsxschneider:
addresses #4330

/cc @iamhughes

### DIFF
--- a/docs/man/git-lfs-migrate.1.ronn
+++ b/docs/man/git-lfs-migrate.1.ronn
@@ -38,6 +38,11 @@ git-lfs-migrate(1) - Migrate history to or from git-lfs
 * `--everything`:
     See [INCLUDE AND EXCLUDE (REFS)].
 
+    Note: Git refs are "case-sensitive" on all platforms in "packed from" (see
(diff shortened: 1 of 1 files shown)


docs/man/git-lfs-migrate.1.ronn
```diff
@@ -38,6 +38,11 @@ git-lfs-migrate(1) - Migrate history to or from git-lfs
 * `--everything`:
     See [INCLUDE AND EXCLUDE (REFS)].
 
+    Note: Git refs are case sensitive on all platforms in packed from (see
+    `git-pack-refs(1)). On case insensitive file systems, e.g. NTFS on Windows
```
bk2204, line 42: I think this should be “packed form” and “case-insensitive” (with a hyphen).  Also, it looks like the backtick is missing after `git-pack-refs(1)`.

larsxschneider: Fixed. Thank you 👍

bk2204 (approved):

PR #4341, opened 2020-12-10
Makefile: don't fail the second time macOS builds are built

bk2204:
zip has an interesting feature: if there is nothing to do, it exits 12. Since in our macOS release CI target, we build the same asset multiple times due to using two separate invocations of make, we end up with a case where zip exits 12 unexpectedly.

Ideally make would realize that our assets don't need to be rebuilt, but as long as we're going to rebuild them, let's remove the existing file so zip doesn't get bent out of shape about the fact that our operation isn't doing anything useful.

This caused the v2.13.0 release job to fail, so I'll need to do a v2.13.1 with this change.


chrisd8088 (approved):
❤️

Rithy77:
Hello word

PR #4343, opened 2020-12-11
release: v2.13.1

bk2204:
This is an in-progress look at the next release of Git LFS, v2.12.0, which is scheduled for Friday, December 11, 2020.

I'm not providing binaries here because this is a very simple bug fix for a CI problem that prevented us from releasing v2.13.0.  Therefore, it should not function any differently from v2.13.0 other than containing a different version number.

The CI system will produce artifacts for Windows, Linux, and macOS if you need some test binaries.

/cc @git-lfs/core
/cc @git-lfs/implementers
/cc @git-lfs/releases

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Git LFS Changelog
 
+## 2.13.1 (11 Dec 2020)
+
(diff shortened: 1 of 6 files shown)


chrisd8088 (approved):
👍 Now with macOS!

PR #4344, opened 2020-12-11
Makefile: allow make release to be run twice in a row

bk2204:
We recently fixed a problem with the macOS target to make it possible to build a second time after running `make release`, but we forgot a Windows target as well.  Let's make it possible to run that command twice in a row.


chrisd8088 (approved):
👍 Now with Windows!

PR #4345, opened 2020-12-14
Add test for download gzip transport compression

bk2204:
Since we're uploading and downloading a large amount of data, we may prefer to use transport compression.  Currently, Go will do this for us automatically unless we ask it not to.  Let's detect this case and add a trace output so people can find it and we can add a test that it continues to work.

Note that it's not possible to detect this using the headers since Go adds the headers on the request and strips the headers on the response as part of making the request, so it's not readily apparent to users without this trace output.


chrisd8088 (approved):
This seems all good to me, thank you!

PR #4347, opened 2020-12-17
Fix numbered list in git lfs examples

gison93:
When the file is rendered in the browser the numbers in the numbered list are automatically corrected, while when the file is shown in the command line the numbers are wrong.


bk2204 (approved):
Thanks for the patch, and welcome to Git LFS!  This looks like a pretty straightforward fix, and pretty obviously correct.

PR #4368, opened 2021-01-13
prune: add options to be more aggressive about pruning

### LINKED ISSUE #4367: Allow unrestricted pruning
The default git lfs prune is really restricted, as a result one often has to:

1. Find out what combinations of options are needed among:
    lfs.fetchrecentrefsdays
    lfs.fetchrecentremoterefs
    lfs.fetchrecentcommitsdays
    lfs.fetchrecentalways
    lfs.pruneoffsetdays

2. Deal with the fact that these options have no effect when used in .lfsconfig

3. Notice that the current checkout is never prune, so that the only solution is to create a dummy empty branch, prune, then come back to the original branch

Can a simple option be added to allow unrestricted pruning ?
By unrestricted, I mean without any heuristic (time, current checkout, ...), but keeping the current pushed / remote verification logic.

bk2204:
Some users really want to minimize the amount of data they store on the local system, especially for large checkouts.  Let's add two options, --recent and --force, the former which ignores the recency settings and the latter which additionally prunes objects which are currently checked out.  We will never prune objects that are unpushed, even with these options, including stashed objects.

Add tests and documentation for both of these options.

Fixes #4367
/cc @Pierre-Bartet as reporter

### DIFF
--- a/commands/command_prune.go
+++ b/commands/command_prune.go
@@ -26,6 +26,8 @@ var (
 	pruneDryRunArg      bool
 	pruneVerboseArg     bool
 	pruneVerifyArg      bool
+	pruneRecentArg      bool
+	pruneForceArg       bool
 	pruneDoNotVerifyArg bool
 )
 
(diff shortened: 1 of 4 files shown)


Pierre-Bartet:
Thanks! It looks perfect to me.

chrisd8088 (approved):
Super, thank you!

PR #4384, opened 2021-01-27
Remove NTLM

### LINKED ISSUE #4247: Remove NTLM support
NTLM support has been broken in a variety of ways for some time.  It's most commonly used by Azure DevOps, although it's possible that other implementations support it.  None of the core team members use Windows, which is where it fails most of the time, nor do we use NTLM.  In addition, NTLM uses weak cryptography, like MD4, which has been known to be insecure for at least 25 years.

Absent someone stepping up to maintain NTLM across platforms, I propose that we drop support for it after the 2.13 release (which will probably take place in early December).  We have support for Kerberos that is known to work on at least Linux (and should work on Windows as well), which can be used for secure SSO instead.

/cc #4234, #3829, #3351
/cc @git-lfs/core

bk2204:
Our NTLM support has been known to be broken in various situations for a while, specifically on Windows.  The core team is unable to troubleshoot these problems, and nobody has stepped up to maintain the NTLM support. In addition, NTLM uses cryptography and security techniques that are known to be insecure, such as the algorithms DES, MD4, and MD5, as well as simple, unsalted hashes of passwords.

Since we now support Kerberos, most users should be able to replace their use of NTLM with Kerberos instead.  Users have reported this working on Windows and it is known to work well on at least Debian as well.  Drop support for NTLM and remove it from the codebase.

Fixes #4247


chrisd8088 (approved):
I did a quick grep through the codebase and it looks like you caught everything related to NTLM (at least, that I could see).  Sweet!

PR #4385, opened 2021-01-27
.github/workflows: don't try to install older Git

bk2204:
Git for Windows 2.27.0 had a bug where symlinks would appear perpetually modified when core.fscache was enabled, causing our CI job to fail because it looks for modified files after running the formatter. However, a newer version of Git is now available and our old code is failing, so let's remove it and use the newer version of Git that's shipped by default.


chrisd8088 (approved):
Thanks!

bk2204:
I actually have to close this PR, because in order for this workflow change to be effective, it has to build out of the main repo, and this is on my fork.

PR #4387, opened 2021-01-29
tools: implement cloneFileSyscall on darwin without cgo

tklauser:
Use `AT_FDCWD` and `CLONE_NOFOLLOW` consts from `golang.org/x/sys/unix`
instead of the respective C constants via cgo.


chrisd8088:
Thanks for the patch!  This seems entirely functional (i.e., it cleanly passed all the builds, including on macOS) and definitely tidies things up a bit, so I'll merge it tomorrow unless another @git-lfs/core person has an objection.

Thanks again!

chrisd8088 (approved):

bk2204 (approved):

Issue #4153, opened 2020-06-08
Rename master branch


bk2204:
There's been discussion recently about the [problematic use of `master` and `slave`](https://blog.carbonfive.com/2018/10/08/problematic-terminology-in-open-source/) in technology.  Similar to what GitHub Desktop (desktop/desktop#6478) and other projects have done, it would be a good idea to rename our default branch to something different, and do this across all of our repositories.

For consistency with what [Git for WIndows is doing](https://github.com/git-for-windows/git/issues/2674), I propose `main`, although we can discuss other names as well.

This issue doesn't track the replacement of this terminology in the codebase, which I think should be split out into another issue.

There are some technical issues to discuss on this that would need doing:

- [ ] Create the new branch in each repository.
- [ ] Set it as the default branch.
- [ ] Update the release howto to reflect the new name.
- [ ] Update the docs link on the website to reflect the new name.
- [ ] Remove the old CI job files that refer to the old default branch, and make sure Actions is set up for the subsidiary repositories if need be.
- [ ] Remove the old `master` branches to avoid confusion.

/cc @git-lfs/core

fcharlie:
The economic foundation determines the superstructure. To fight for equal rights, it is necessary for the disadvantaged groups to get rich and have more political, economic and cultural rights. Can changing the vocabulary alone solve the problem?

bk2204:
I have a `to-main` branch on my remote which updates all of the links and functional documentation (e.g., howtos, contributing, readme) to use `main` instead.  It also will remove the old CI files that refer to `master`.  I know there are some outstanding PRs which we'd want to merge first to let the branch settle, and then we can rename and change the default branch.

raaaaay-c:
**Stop renaming** 
In git concept master has to be understood as the 'original', the 'master copy', the 'first-ever copy'. 
There is no association with master and slaves. 
There is no git slave branch concept, and it's not a slavery-origin terminology. 

In some projects the initial branch is not the main branch. But it will always be the first ever branch, the original branch.

To show black people our respect and solidarity we do not have to rename everything that is called master or black as long as there is no association with master & slave or black as something bad.

If we rename all the things this tends to have the opposite effect, if we always associate master with slavery and when the word "black" in our minds is something bad that we have to rename directly. 

We would be the ones who created racism, not terms.

bNobo:
and what do you think about "root" ? It could also be a good alternative to "master", especially when we are talking about branches.

bk2204:
[...]

PR #4390, opened 2021-02-01
Rename default branch

bk2204:
As outlined in #4153, we'd like to change the default branch name to `main`.  There are a few things which refer to the old name, including some unused config files and scripts, some of our internal documentation, and some user-facing documentation as well.  This doesn't remove the entire use of the old name from the codebase because various repositories need to be updated, but it is sufficient for us to perform the rename in this repository without breaking things.

There is a similar change for doc links on the website (git-lfs/git-lfs.github.com#45) which I'll push immediately after merging this and right before the rename happens.  This is the last of our active repositories to undergo the rename (excluding the website, which I must verify will continue to work); the others should already be updated.  This PR does not close the issue because there are manual steps to be done.

The commit messages explain things in more detail.

Note that this PR is not for discussion about the desirability of this approach; the proper forum for that is #4153, provided the discussion can remain within the bounds of our Code of Conduct.  Discussion outside those bounds should remain outside our community.


znull (approved):

docs/man/git-lfs-migrate.1.ronn
```diff
@@ -192,25 +192,25 @@ include by `--include-ref` and not reachable by `--exclude-ref`.
...
      \
-      refs/remotes/origin/master
+      refs/remotes/origin/main
 ```
 
 In the above configuration, the following commits are reachable by each ref:
 
 ```
-refs/heads/master:         C, B, A
+refs/heads/main  :         C, B, A
```
chrisd8088, line 203: Here the whitespace is:
```
main  : A
```
while below it's
```
main:   A
```

Was that intentional, or just an inadvertent difference?

chrisd8088 (approved):
Other than one small question, this looks like a good cleanup and rename, thanks!

docs/man/git-lfs-migrate.1.ronn
bk2204: It was an inadvertent difference.  Fixed.

Issue #3531, opened 2019-02-19
Checkout Error: Pointer Parsing: Expected key oid (NO file specified)


dylanh724:
When I `git pull`, all is well.
However, when I specifically `git lfs pull`, I get this error:

![image](https://user-images.githubusercontent.com/8840024/52994328-25f0a000-3452-11e9-840a-87a6d9efbf37.png)

From my Google searches, this is supposed to link to a specific file. I'm told that I can normally fix merge errors, **but there is NO file specified!**

Someone noted it was a Git LFS error that just "randomly happens" after many pulls.

I can't find any way to fix this. What can I do?

# Logs:

$ git lfs logs last
git-lfs/2.3.4 (GitHub; windows amd64; go 1.8.3; git d2f6752f)
git version 2.15.0.windows.1

$ git-lfs pull
Checkout error: pointer parsing: Expected key oid, got <<<<<<<
Checkout error: pointer parsing: Expected key oid, got <<<<<<<

Expected key oid, got <<<<<<<
github.com/git-lfs/git-lfs/errors.Errorf
        C:/Users/techn/go/src/github.com/git-lfs/git-lfs/errors/errors.go:69
github.com/git-lfs/git-lfs/errors.NewBadPointerKeyError
        C:/Users/techn/go/src/github.com/git-lfs/git-lfs/errors/types.go:306
github.com/git-lfs/git-lfs/lfs.decodeKVData
        C:/Users/techn/go/src/github.com/git-lfs/git-lfs/lfs/pointer.go:276
github.com/git-lfs/git-lfs/lfs.decodeKV
        C:/Users/techn/go/src/github.com/git-lfs/git-lfs/lfs/pointer.go:150
github.com/git-lfs/git-lfs/lfs.DecodeFrom
        C:/Users/techn/go/src/github.com/git-lfs/git-lfs/lfs/pointer.go:131
github.com/git-lfs/git-lfs/lfs.DecodePointer
        C:/Users/techn/go/src/github.com/git-lfs/git-lfs/lfs/pointer.go:107
github.com/git-lfs/git-lfs/lfs.DecodePointerFromFile
        C:/Users/techn/go/src/github.com/git-lfs/git-lfs/lfs/pointer.go:104
github.com/git-lfs/git-lfs/commands.(*singleCheckout).Run
        C:/Users/techn/go/src/github.com/git-lfs/git-lfs/commands/pull.go:44
github.com/git-lfs/git-lfs/commands.pull.func1
[...]

PR #4394, opened 2021-02-03
Log file name when failing to decode pointer from file

MartinLCoatsink:
This will make it easier to find 'bad' git-lfs files that probably got mangled in a git merge at some point.

The issue is related to this discussion: https://github.com/git-lfs/git-lfs/issues/3531


lfs/pointer.go
```diff
@@ -95,8 +95,16 @@ func DecodePointerFromFile(file string) (*Pointer, error) {
 		return nil, err
 	}
 	defer f.Close()
-	return DecodePointer(f)
+	
+	// Since the lower level functions don't have any awareness of which file caused the error
+	// log the filename here when the error eventually bubbles up
+	p, err := DecodePointer(f)
+	if err != nil {
+		errors.NewNotAPointerError(errors.New("expected file to contain a pointer, but it didn't. File: " + file))
```
bk2204, line 103: I think this fails to pass on the message from lower down.  It can be useful for folks to know what exactly made it invalid instead of having to guess.  Also, this doesn't actually change anything, since it just instantiates an object here and doesn't return anything. 

Maybe something like this:

```go
if err != nil {
        err = errors.NewNotAPointerError(fmt.Errorf("expected %q to contain a pointer, but it didn't: %s", file, err))
}
```

And of course, some tests, so we can have confidence that this does work as expected.  There are tests in `t` that would be a good place for this.

MartinLCoatsink: Passing on the error is a good idea, I'll fix that. 

I'll try taking a look at test as well. Found a very similar test in t-smudge.sh called "smudge with invalid pointer". However I'm not sure if that runs the code path I'm testing, since it reads the data directly from stdin, rather than from a file.

commands/pull.go
```diff
@@ -67,7 +67,7 @@ func (c *singleCheckout) Run(p *lfs.WrappedPointer) {
 	// Check the content - either missing or still this pointer (not exist is ok)
 	filepointer, err := lfs.DecodePointerFromFile(cwdfilepath)
 	if err != nil && !os.IsNotExist(err) {
-		if errors.IsNotAPointerError(err) || errors.IsBadPointerKeyError(err) {
+		if errors.IsNotAPointerError(err) {
```
MartinLCoatsink, line 70: Was scratching my head for a while, until I found this line of code, which was silencing the error "expected key oid, got <<<<<".

This line seems to have been added fairly recently, in v2.13.2 ([commit](https://github.com/git-lfs/git-lfs/commit/9e3bc45b934758e422c3e800161dabd0711775bf))

My git installation was still runnign git-lfs 2.11.0, which explains why I could repro easily using my regular git, but not with the latest version.  

Is silencing the error still the expected behaviour? I find that this sometimes leads to accidently commiting files with conflics (e.g. _BACKUP files). While it's not a terrible problem, it seems undesirable.

bk2204: If you want to revert it, that's fine, but please remove the test so it doesn't fail.

MartinLCoatsink: It's probably good 'as is'.

PR #4396, opened 2021-02-03
Consider scheme of request URL, not proxy URL, when choosing proxy

### LINKED ISSUE #4392: gitconfig http proxy values should be honored for https traffic even if proxy url does not start with https
Currently, gitconfig http proxy values are only honored for https if the proxy url starts with https.

Code snippet from proxy.go:

```
if strings.HasPrefix(gitProxy, "https://") {
  httpsProxy = gitProxy
}
```

However, some corporate IT environments configure https proxies to use http:// URLs. Ideally, if an http proxy rule is defined for a given LFS url, it would be honored regardless of the protocol of that proxy URL.

For example, in .gitconfig:

[http "https://lfs.github.com"]
  proxy = http://hostname.domain.com:PORT

In this case, we are being explicit that the given proxy should be used for HTTPS traffic going to lfs.github.com. It shouldn't matter that the proxy URL starts with http://.

In lfshttp/proxy.go, I would suggest examining the target URL rather than the proxy URL to determine if httpsProxy should be set.

bk2204:
If a user has a proxy using HTTP but a URL using HTTPS, we fail to honor the http.proxy configuration setting because we check the scheme of the proxy URL, not the scheme of the request URL.  Let's fix this by deciding which proxy variable to set based on the request URL, as was probably intended, rather than the proxy URL.

Add some tests for this case, as well as the case that we correctly ignore the HTTPS_PROXY environment variable if the scheme of the request URL is HTTP, which we were previously lacking.

/cc @bhilton (as submitter)
Fixes #4392

### DIFF
--- a/lfshttp/proxy.go
+++ b/lfshttp/proxy.go
@@ -75,7 +75,7 @@ func getProxyServers(u *url.URL, urlCfg *config.URLConfig, osEnv config.Environm
 	if urlCfg != nil {
 		gitProxy, ok := urlCfg.Get("http", u.String(), "proxy")
 		if len(gitProxy) > 0 && ok {
-			if strings.HasPrefix(gitProxy, "https://") {
+			if u.Scheme == "https" {
 				httpsProxy = gitProxy
 			}
 			httpProxy = gitProxy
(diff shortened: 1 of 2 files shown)


chrisd8088 (approved):
Looks good!  (Took me a while to understand the sense of the tests, but that's entirely my own fault.)

bk2204:
This is a tricky edge case, which is why we haven't seen it, so I'm not surprised that it took you a second.

jerelev-intel:
Thank you for fixing this!

PR #4398, opened 2021-02-08
vendor,go.{mod,sum}: update x/net and dependencies

chrisd8088:
Update the vendored `golang.org/x/net` module to the latest version, which in turn updates the vendored copy of the `golang.org/x/text` module to 0.3.3.  That version of `x/text` includes a mitigation of CVE-2020-14040, which pertains to processing UTF-16 data with a Byte Order Mark.

The Git LFS client should not be affected by that security issue, since it does not open streams in UTF-16 mode, but we update our modules to stay current with the upstream Go sources and also to avoid being flagged by security scanners.

The specific commands run to perform this update were:
`go get golang.org/x/net@latest && go mod tidy && go mod vendor`


bk2204 (approved):

Myoazwan95 (commented):
Duplicate of #
#4633

PR #4407, opened 2021-02-19
lfs: don't invoke diff drivers when pruning repositories

### LINKED ISSUE #4401: git lfs prune calls external diff tools ang hangs
**Describe the bug**

We found out that `git lfs prune -d -v` started to hang on one of our Windows boxes. Analysis showed that:

1. There's main `git lfs prune -d -v` process
2. There's a child process: `git -c filter.lfs.smudge= -c filter.lfs.clean= -c filter.lfs.process= -c filter.lfs.required=false log --branches --tags --not --remotes=origin -G "oid sha256:" -p -U12 "--format=lfs-commit-sha: %H %P"`
3. There's a subchild process: `"C:\Program Files\Git\mingw64\bin\antiword.exe" -m UTF-8 C:\Users\CUCUMB~1.ONL\AppData\Local\Temp\2/LM9YyF_classes_No_Name.dot`

It turns out that `git log -G` command runs external diff tools. [AntiWord is set up to diff `*.dot` files in Git for Windows](https://github.com/git-for-windows/build-extra/blob/git-sdk-1.0.8/git-extra/astextplain#L15). But if you pass it something that is not a Word file, AntiWord logs to stderr: `C:\Users\MEA71~1.RAD\AppData\Local\Temp/EJNcO8_1.dot is not a Word Document.`

Now, the **final and most important piece** of this problem is [parseScannerLogOutput](https://github.com/git-lfs/git-lfs/blob/v2.13.2/lfs/gitscanner_log.go#L122-L144) function that doesn't properly read subprocess stderr and is able to deadlock if `git log` process writes enough data to stderr and blocks on I/O buffer availability.

**Almost a way to reproduce**

I'm not yet sure how to fully reproduce this within a `git lfs prune` context, but can clearly see that `git log` invoked in the same way DOES write to stderr. Within a Git Bash:

```
$ mkdir foo
$ cd foo/
$ git init
$ echo "I cho?" > 1.dot
$ git add 1.dot 
$ git commit -am "initial"
$ git log --branches --tags --not -G "cho" -p -U12 "--format=lfs-commit-sha: %H %P"
C:\Users\MEA71~1.RAD\AppData\Local\Temp/EJNcO8_1.dot is not a Word Document.
```

**Expected behavior**

At least, `git lfs prune` is not supposed to hang. So, [parseScannerLogOutput](https://github.com/git-lfs/git-lfs/blob/v2.13.2/lfs/gitscanner_log.go#L122-L144) needs to be fixed to properly consume both stdout/stderr. Second, I'm not sure whether it is correct that `git lfs prune` invokes `git log` in such a way that it involves external diff tools, possibly this is a bug too. Third, I'm not sure whether it is correct that Git for Windows writes to stderr when one does `git log -G` and its output happens to include `*.dot` files.

**System environment**

```
$ git --version
git version 2.26.0.windows.1

$ git lfs version
git-lfs/2.10.0 (GitHub; windows amd64; go 1.12.7; git a526ba6b)
```

**Output of `git lfs env`**

```
$ git lfs env
git-lfs/2.10.0 (GitHub; windows amd64; go 1.12.7; git a526ba6b)
git version 2.26.0.windows.1
LocalWorkingDir=
LocalGitDir=
LocalGitStorageDir=
LocalMediaDir=C:\lfs-cache\objects
LocalReferenceDirs=
TempDir=C:\lfs-cache\tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=C:\lfs-cache
AccessDownload=none
AccessUpload=none
```

bk2204:
Certain file paths are associated with diff textconv patterns with Git for Windows, and these are invoked by default when using git log. However, when we're using git log to scan data to find LFS objects, we don't want to invoke those programs.

Let's fix this by passing appropriate options to disabled these invocations.  In the test, we set the driver to false, since Git will fail if the process it calls also fails.

Fixes #4401

### DIFF
--- a/t/t-prune.sh
+++ b/t/t-prune.sh
@@ -858,3 +858,22 @@ begin_test "prune --force"
   refute_local_object "$oid_inrepo" "${#content_inrepo}"
 )
 end_test
+
+begin_test "prune does not invoke external diff programs"
+(
+  set -e
+
+  reponame="prune-external-diff"
+  setup_remote_repo "remote-$reponame"
+
+  clone_repo "remote-$reponame" "clone-$reponame"
+
+  git config diff.word.textconv 'false'
+  echo "*.dot diff=word" >.git/info/attributes
+
+  for n in $(seq 1000); do (echo "$n" > "$n.dot") done
+  git add .
+  git commit -am "initial"
+  git lfs prune
+)
+end_test
--- a/lfs/gitscanner_log.go
+++ b/lfs/gitscanner_log.go
@@ -29,6 +29,8 @@ var (
 	// Arguments to append to a git log call which will limit the output to
 	// lfs changes and format the output suitable for parseLogOutput.. method(s)
 	logLfsSearchArgs = []string{
+		"--no-ext-diff",
+		"--no-textconv",
 		"-G", "oid sha256:", // only diffs which include an lfs file SHA change
 		"-p",                             // include diff so we can read the SHA
 		"-U12",                           // Make sure diff context is always big enough to support 10 extension lines to get whole pointer


slonopotamus:
Are we going to ignore for now the possibility of hanging if any other reasons cause `git log` to write something to stderr?

bk2204:
Yes, we are, since Git should not normally print a large amount of data to standard error.

t/t-prune.sh
```diff
@@ -858,3 +858,22 @@ begin_test "prune --force"
   refute_local_object "$oid_inrepo" "${#content_inrepo}"
 )
 end_test
+
+begin_test "prune does not invoke exterrnal diff programs"
```
chrisd8088, line 862: Minor typo in `exterrnal`.

chrisd8088 (approved):
Seems reasonable, thank you!

t/t-prune.sh
bk2204: Fixed, thanks.

slonopotamus:
> since Git should not normally print a large amount of data to standard error.

Normally yes, but. It is common to use things like `GIT_TRACE=1` and other similar things when debugging git/git-lfs issues. Even you often ask users to provide such logs. And these things DO print a lot to stderr.

Today, we got another deadlock between git/git-lfs, now caused by too much data on stderr from git called from [RevListScanner](https://github.com/git-lfs/git-lfs/blob/a32af909e770dfac946d3690e48968834e821add/git/rev_list_scanner.go#L165-L214).

bk2204:
Okay, well, please open an issue for it with reproduction steps and we'll take a look.  I agree that in hindsight we probably should have fixed it differently originally.

slonopotamus:
We're currently trying to make up a repro for RevListScanner deadlock and will create an issue then.

PR #4418, opened 2021-03-01
Canonicalize Windows paths like Git does

### LINKED ISSUE #4012: Paths are not canonicalized in the same way as Git on Windows
I have several repositories on a network path. I am able to use git-lfs on those existing repositories just fine, but I am running into issues when trying to track new files. I'm not sure what triggered the change in behavior. I moved from a Windows 7 to a Windows 10 machine, and am probably running a different version of git and git-lfs than before. I'm not sure of the relevance of this.

Let's say the network drive with UNC path `\\network` is mapped to the drive letter N. 
When I run the following commands from the Windows command shell (`cmd`), 
```
$ N:
$ mkdir repo
$ cd repo
$ git init
$ git lfs install
$ git lfs track *.bin
```
I get an error,
```
Current directory "N:\\repo" is outside of git working directory "\\\\network\\repo"
```
When I run `git lfs env`, I get the following entries for directories directories:
```
git-lfs/2.10.0 (Github; windows amd64; go 1.12.7; git a526ba6b
git version 2.25.0.windows.1

LocalWorkingDir=\\network\repo
LocalGitDir=N:\repo\.git
LocalGitStorageDir=N:\repo\.git
LocalMediaDir=N:\repo\.git\lfs\objects
LocalReferenceDirs=
TempDir=N:\repo\.git\lfs\tmp
(...)
LfsStorageDir=N:\repo\.git\lfs
```
It seems to me the issue arises from the fact that the `LocalWorkingDir` is set to the UNC path, whereas all other `*Dir` entries are set to the mapped drive letter. 

I have no issues in adding/committing changes to files in existing repos as lfs objects (they are already tracked), however, I can no longer track new files in those existing repositires, which I was able to do before just fine. The `git lfs env` shows similar pattern (UNC path on `LocalWorkingDir`, network drive letters on other `*Dir` settings).

edit: small additions

### LINKED ISSUE #4238: LFS does not use a SUBSTed Path in every place
I'm using git lfs as Part of the git for Windows Installation v 2.27.0. It's git lfs 2.11 - see below the output of git lfs env for details.

Im using the dos SUBST command to redirect a subdirectory to the driveletter D:. Today I cloned an empty repository and wanted to track "*.docx" files and got the following error:

> /d/repos/PROD/TEST/test.push_test_wg_git_lfs (master)
> $ git lfs track "*.docx"
> Current directory "D:\\repos\\PROD\\TEST\\test.push_test_wg_git_lfs" outside of git working directory "C:\\Users\\Public\\Documents\\subst_D\\repos\\PROD\\TEST\\test.push_test_wg_git_lfs".

I cd'ed to the "real" directory and repeated the command successfully. After that cd'ed back to the D:-version of the folder an could work with lfs flawlessly. (git add, git commit, git push - all worked like expected).

When I do a git lfs env I see that all directorys are located on the D:-Drive. Only the **LocalWorkingDir** is located on C:

> $ git lfs env
> git-lfs/2.11.0 (GitHub; windows amd64; go 1.14.2; git 48b28d97)
> git version 2.27.0.windows.1
> 
> Endpoint=https://xxxxxxx/test/test.push_test_wg_git_lfs.git/info/lfs (auth=none)
>   SSH=git@xxxxxxx:test/test.push_test_wg_git_lfs.git
> LocalWorkingDir=C:\Users\Public\Documents\subst_D\repos\PROD\TEST\test.push_test_wg_git_lfs
> LocalGitDir=D:\repos\PROD\TEST\test.push_test_wg_git_lfs\.git
> LocalGitStorageDir=D:\repos\PROD\TEST\test.push_test_wg_git_lfs\.git
> LocalMediaDir=D:\repos\PROD\TEST\test.push_test_wg_git_lfs\.git\lfs\objects
> LocalReferenceDirs=
> TempDir=D:\repos\PROD\TEST\test.push_test_wg_git_lfs\.git\lfs\tmp
> ConcurrentTransfers=8
> TusTransfers=false
> BasicTransfersOnly=false
> SkipDownloadErrors=false
> FetchRecentAlways=false
> FetchRecentRefsDays=7
> FetchRecentCommitsDays=0
> FetchRecentRefsIncludeRemotes=true
> PruneOffsetDays=3
> PruneVerifyRemoteAlways=false
> PruneRemoteName=origin
> LfsStorageDir=D:\repos\PROD\TEST\test.push_test_wg_git_lfs\.git\lfs
> AccessDownload=none
> AccessUpload=none
> DownloadTransfers=basic,lfs-standalone-file
> UploadTransfers=basic,lfs-standalone-file
> GIT_EXEC_PATH=C:/Program Files/Git/mingw64/libexec/git-core
> GIT_SSH=C:\Program Files (x86)\WinSSHTerm\tools\plink.exe
> git config filter.lfs.process = "git-lfs filter-process"
> git config filter.lfs.smudge = "git-lfs smudge -- %f"
> git config filter.lfs.clean = "git-lfs clean -- %f"

My OS is Windows 10 1709 (Build 16299.2045). I tested it also on Win 10 1909 (Build 18363.1016) - same error.

I don't think this is an intended behavior. At least it is not the behavior I did expect.

bk2204:
Git consistently uses canonicalized paths internally.  This is for many reasons, but mostly to verify that a single path is within a repository. In order to interoperate properly with Git, we need to canonicalize paths and do it in the same way as Git.

On Unix systems, to canonicalize a path, it is sufficient to make the path absolute and then resolve any symlinks.  Go provides two functions to do these two steps, filepath.Abs and filepath.EvalSymlinks, and they work as advertised.

Windows, however, has much more complex path handling and these functions do not handle all cases.  The typical way to canonicalize paths on Windows is using GetFinalPathNameByHandle, and this is the technique Git uses.

Go, however, does not provide a general API to canonicalize paths, unlike Rust's std::fs::canonicalize and similar functionality in myriad other languages.  Therefore, in order to get this working on Windows, let's add a function to canonicalize paths in the appropriate system way, one for Unix systems and one for Windows.  The code comes from Go's standard library, so update the copyright and license accordingly.

Update the CanonicalizePath function to use the new function.  We duplicate the Abs call because we an absolute path in CanonicalizePath in some cases even if the path is missing, whereas the new function needs to do it in all cases because we will use it other situations in the future.  This should be a simple string check, so it should not involve an extra system call or other overhead.

Use the new function in some other parts of the code where we were previously using filepath.EvalSymlinks.

There are no tests for this case because the situations which are problematic require either SMB shares or SUBST, and neither is acceptable in a testsuite like ours.

Fixes #4238
Fixes #4012

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -20,8 +20,8 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 
-Portions of the subprocess directory are copied from Go and are under the
-following license:
+Portions of the subprocess and tools directories are copied from Go and are
+under the following license:
 
 Copyright (c) 2010 The Go Authors. All rights reserved.
 
--- a/commands/command_lock.go
+++ b/commands/command_lock.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/git-lfs/git-lfs/errors"
 	"github.com/git-lfs/git-lfs/git"
+	"github.com/git-lfs/git-lfs/tools"
 	"github.com/spf13/cobra"
 )
 
@@ -79,7 +80,7 @@ func lockPath(file string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	wd, err = filepath.EvalSymlinks(wd)
+	wd, err = tools.CanonicalizeSystemPath(wd)
 	if err != nil {
 		return "", errors.Wrapf(err,
 			"could not follow symlinks for %s", wd)
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -347,7 +347,7 @@ func changeToWorkingCopy() {
 		ExitWithError(errors.Wrap(
 			err, "fatal: could not determine current working directory"))
 	}
-	cwd, err = filepath.EvalSymlinks(cwd)
+	cwd, err = tools.CanonicalizeSystemPath(cwd)
 	if err != nil {
 		ExitWithError(errors.Wrap(
 			err, "fatal: could not canonicalize current working directory"))
--- a/debian/copyright
+++ b/debian/copyright
@@ -31,6 +31,7 @@ Files: vendor/github.com/alexbrainman/sspi/*
  subprocess/path.go
  subprocess/path_nix.go
  subprocess/path_windows.go
+ tools/filetools_windows.go
 Copyright: 2009, 2010, 2012, 2017 The Go Authors
 License: Go
  Redistribution and use in source and binary forms, with or without
--- a/lfshttp/standalone/standalone.go
+++ b/lfshttp/standalone/standalone.go
@@ -137,7 +137,7 @@ func gitDirAtPath(path string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return filepath.EvalSymlinks(gitdir)
+	return tools.CanonicalizeSystemPath(gitdir)
 }
 
 func fixUrlPath(path string) string {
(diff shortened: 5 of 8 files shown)


chrisd8088 (approved):
Yay!

PR #4420, opened 2021-03-03
invoke git-config with --includes to ensure it always evaluates include.* directives

codykrieger:
Per `git-config`'s man page, `include.*` directives are not respected by default when asked to evaluate a specific file (whether via `--global`, `--file /path/to/.gitconfig`, or `GIT_CONFIG=/path/to/.gitconfig`).

This patch ensures that, if `GIT_CONFIG` is set in the environment when e.g. `git-lfs-push` is invoked, config directives from included files are evaluated.

This is important if, for example, an included file contains a `url.<foo>.insteadOf` directive or `lfs.<foo>.locksverify` directive that's shared between many repositories.

Before this change, with the following setup:

```
% git config --file /tmp/a.config -l
...
remote.origin.url=https://my-server-alias/foo/bar.git
include.path=/tmp/b.config

% git config --file /tmp/b.config -l
...
url.https://my-actual-server.com/.insteadOf=https://my-server-alias/
```

I'd get something like:

```
% GIT_CONFIG=/tmp/a.config git push
...
batch response: Post "https://my-server-alias/foo/bar.git/info/lfs/objects/batch": dial tcp: lookup my-server-alias: no such host
batch response: Post "https://my-server-alias/foo/bar.git/info/lfs/objects/batch": dial tcp: lookup my-server-alias: no such host
Uploading LFS objects:   0% (0/2), 0 B | 0 B/s, done.
error: failed to push some refs to 'https://my-actual-server.com/foo/bar.git'
```

But now it works as expected. :)

I'm open to adding tests if someone can point me at the best place to add them.

### DIFF
--- a/git/config.go
+++ b/git/config.go
@@ -199,7 +199,7 @@ func (c *Configuration) Source() (*ConfigurationSource, error) {
 }
 
 func (c *Configuration) gitConfig(args ...string) (string, error) {
-	args = append([]string{"config"}, args...)
+	args = append([]string{"config", "--includes"}, args...)
 	cmd := subprocess.ExecCommand("git", args...)
 	if len(c.GitDir) > 0 {
 		cmd.Dir = c.GitDir
--- a/t/t-config.sh
+++ b/t/t-config.sh
@@ -256,3 +256,23 @@ begin_test "config: ignoring unsafe lfsconfig keys"
   grep "  core.askpass" status.log
 )
 end_test
+
+begin_test "config respects include.* directives when GIT_CONFIG is set"
+(
+  set -e
+
+  mkdir include-directives
+  cd include-directives
+
+  git init
+
+  git config lfs.url "http://some-url/rest"
+  GIT_CONFIG="$(pwd)/.git/config" git lfs env | tee env.log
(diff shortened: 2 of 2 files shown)


bk2204 (commented):

This looks like a great improvement.  I was surprised to find that `--includes` is actually supported by Git 1.8.3.1 (and I contribute to Git), so this doesn't fortunately doesn't need a feature check and the code looks like it should be fine.

I would like to see some tests for this, as you alluded to.  I think the typical place for those here would be `t/t-config.sh`.  I think the easiest way to test this might be trying to use a config option that would influence `git lfs env` and then invoking that, but other options are possible, too.  If you have questions or run into trouble, please let us know, and we'll try to help out.

codykrieger:
@bk2204 Brilliant, thanks for the pointer—I'll give that a shot. :)

codykrieger:
@bk2204 OK! I think that'll do it. I ran the new test with and without my changes—fails without, passes with. Fingers crossed it passes all of the CI checks. :) 🤞

bk2204 (approved):
This looks great.  Thanks again for the PR.

codykrieger:
@bk2204 Cheers! 🙌

PR #4421, opened 2021-03-04
Tighten LFS pointer regexp

stanhu:
This prevents a potential parameter injection that might result from
decoding an LFS pointer file with trailing characters in the OID.

### DIFF
--- a/lfs/pointer.go
+++ b/lfs/pointer.go
@@ -22,7 +22,7 @@ var (
 	}
 	latest      = "https://git-lfs.github.com/spec/v1"
 	oidType     = "sha256"
-	oidRE       = regexp.MustCompile(`\A[[:alnum:]]{64}`)
+	oidRE       = regexp.MustCompile(`\A[0-9a-f]{64}\z`)
 	matcherRE   = regexp.MustCompile("git-media|hawser|git-lfs")
 	extRE       = regexp.MustCompile(`\Aext-\d{1}-\w+`)
 	pointerKeys = []string{"version", "oid", "size"}
--- a/lfs/pointer_test.go
+++ b/lfs/pointer_test.go
@@ -274,6 +274,11 @@ size 12345`,
 ext-0-foo boom:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
 oid sha256:4d7a214614ab2935c943f9e0ff69d22eadbb8f32b1258daaa5e2ca24d17e2393
 size 12345`,
+
+		// bad OID
+		`version https://git-lfs.github.com/spec/v1
+oid sha256:4d7a214614ab2935c943f9e0ff69d22eadbb8f32b1258daaa5e2ca24d17e2393&
+size 177735`,
 	}
 
 	for _, ex := range examples {


lfs/pointer.go
```diff
@@ -22,7 +22,7 @@ var (
 	}
 	latest      = "https://git-lfs.github.com/spec/v1"
 	oidType     = "sha256"
-	oidRE       = regexp.MustCompile(`\A[[:alnum:]]{64}`)
+	oidRE       = regexp.MustCompile(`\A[0-9A-Fa-f]{64}\z`)
```
stanhu, line 25: https://github.com/git-lfs/git-lfs/blob/master/docs/spec.md#the-pointer says this is lowercase:

`oid tracks the unique object id for the file, prefixed by its hashing method: {hash-method}:{hash}. Currently, only sha256 is supported. The hash is lower case hexadecimal.`

Do we care about supporting uppercase?

lfs/pointer.go
```diff
@@ -195,7 +195,7 @@ func parseOid(value string) (string, error) {
 	if parts[0] != oidType {
 		return "", errors.New("Invalid Oid type: " + parts[0])
 	}
-	oid := parts[1]
+	oid := strings.TrimSpace(parts[1])
```
bk2204, line 198: It's unclear to me why we need this.  Can you say a few words about this change?

bk2204 (commented):
Hey,

I think overall this is a nice improvement.  I agree with your decision not to allow uppercase, since pointer files are supposed to be unique.

I did have a question about one aspect of the change, though.

lfs/pointer.go
stanhu: @bk2204 I was concerned some LFS pointer files might have trailing space, newlines, or line feed characters (e.g. with Windows) now that we have a `\z` in the regexp. I think `decodeKVData` calls `strings.SplitN`, which might keep `\r`?

bk2204: Ah, that makes sense.  I believe `bufio.NewScanner` strips off all of the possible line ending characters, so that shouldn't be a problem and we can just remove this line.  Pointer files should not have CRLF line feeds, since they're supposed to be bit-for-bit unique, but I believe we'll handle it gracefully even if they do.

bk2204 (approved):
Looks great.  Thanks for the patch.

lfs/pointer.go
stanhu: @bk2204 Now I remember why I added this. Do we care about tolerating trailing spaces? This now fails:

```diff
diff --git a/lfs/pointer_test.go b/lfs/pointer_test.go
index 7960bd38..88ab2e58 100644
--- a/lfs/pointer_test.go
+++ b/lfs/pointer_test.go
@@ -88,8 +88,7 @@ func TestDecodeTinyFile(t *testing.T) {
 
 func TestDecode(t *testing.T) {
 	ex := `version https://git-lfs.github.com/spec/v1
-oid sha256:4d7a214614ab2935c943f9e0ff69d22eadbb8f32b1258daaa5e2ca24d17e2393
-size 12345`
+oid sha256:4d7a214614ab2935c943f9e0ff69d22eadbb8f32b1258daaa5e2ca24d17e2393` + "  \nsize 12345"
 
 	p, err := DecodePointer(bytes.NewBufferString(ex))
 	assertEqualWithExample(t, ex, nil, err)
```

bk2204: No, we definitely do not want people to use trailing spaces.  We want people to use the single, canonical pointer file without any modifications.  The fact that we happen to tolerate CRs in the file is an oversight, not a feature.

PR #4427, opened 2021-03-08
Update README and Contributing docs

### LINKED ISSUE #4426: Documentation lacks instructions on how to remove git lfs
The markdown doc and "git lfs help" explain how I can start using lfs and hook it into my git repo. Great.

But they don't tell me what I should do if I decide git lfs is more trouble than it's worth and I want to stop using it. Let's say I converted my repo to lfs using import --everything, then added 200 commits, some files tracked by lfs, some not. 

Now I want to stop using lfs and want my repo to be the way it would've been if I had been using plain git from the start: so for every commit in history, to have lfs-tracked files replaced by normal "managed by plain git" files.

I'd like to use this tool but I'm apprehensive about it because I don't have the technical knowledge / git expertise to stop using it. So please fix this glaring omission from the doc. It's a best practice anyway to show how to uninstall your app.

bk2204:
This introduces two new pieces to our common Markdown files.

The first one covers uninstalling Git LFS, since that's a question that a lot of users have.  We now have an entry for it in the FAQ as well, so hopefully that documentation should be easier to find.

The second piece encourages users to use the discussions functionality for things that aren't bug reports or feature requests.  As outlined in the commit message, that allows askers to easily mark a correct answer, which both helps future users and also prevents those users from popping up in an issue asking for the canonical answer when one has been given.

Fixes #4426

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,6 +74,17 @@ them as separate pull requests.
 * Explain the rationale for your change in the pull request. You can often use
   part of a good commit message as a starting point.
 
+## Discussions
+
+[Our discussions](https://github.com/git-lfs/git-lfs/discussions) are the
+perfect place to ask a question if you're not sure on something, provide
+feedback that isn't a bug report or feature request, or learn about use cases or
+best practices with Git LFS.  There's even a search box to help you see if
+someone has already answered your question!
+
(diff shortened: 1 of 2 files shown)


chrisd8088 (approved):
Great!

Issue #4419, opened 2021-03-03
--fixup flag to support git lfs migrate info


kid1412621:
**Describe the issue**
`--fixup` flag now only support `git lfs migrate import`, but the tracked file type could be numerous. Using `--include` flag could be cumbersome. 

**System environment**
```
NAME=Fedora
VERSION="33 (Workstation Edition)"
ID=fedora
VERSION_ID=33
VERSION_CODENAME=""
PLATFORM_ID="platform:f33"
PRETTY_NAME="Fedora 33 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:33"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f33/system-administrators-guide/"
SUPPORT_URL="https://fedoraproject.org/wiki/Communicating_and_getting_help"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=33
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=33
PRIVACY_POLICY_URL="https://fedoraproject.org/wiki/Legal:PrivacyPolicy"
VARIANT="Workstation Edition"
VARIANT_ID=workstation
```

**Output of `git lfs env`**
```
git-lfs/2.13.2 (GitHub; linux amd64; go 1.14.13; git fc664697)
git version 2.29.2

Endpoint=https://gitlab.xxx.com/rdcdevcd/CD_FY_PRD_ZNBQ_DOC.git/info/lfs (auth=none)
  SSH=git@gitlab.xxx.com:rdcdevcd/CD_FY_PRD_ZNBQ_DOC.git
LocalWorkingDir=/home/kid/CD_FY_PRD_ZNBQ_DOC
LocalGitDir=/home/kid/CD_FY_PRD_ZNBQ_DOC/.git
LocalGitStorageDir=/home/kid/CD_FY_PRD_ZNBQ_DOC/.git
LocalMediaDir=/home/kid/CD_FY_PRD_ZNBQ_DOC/.git/lfs/objects
LocalReferenceDirs=
[...]

PR #4436, opened 2021-03-12
Allow reporting of LFS pointers separately in migrate info command

chrisd8088:
The output of the `git lfs migrate info` command, when the repository history contains existing Git LFS pointers, treats those pointers as regular files and counts only the size of the pointer, not the referenced object.  This can result in misleading output for repositories where there is a potential use case for the `git lfs migrate import --fixup` command and option, among others.

This PR adds a `--pointers` option to the `git lfs migrate info` command, which accepts one of three values, `ignore`, `follow`, or `no-follow` (the default).  The latter is the default only because it ensures we maintain backwards compatibility with the output of the command in previous Git LFS versions.

The `--pointers=ignore` option simply skips any existing Git LFS pointers when calculating the totals for the summary, while the `follow` alternative reports (on a separate line) the total sizes and numbers of Git LFS objects in the traversed Git history.  For example, on a small test repo:
```
$ git lfs migrate info --pointers=follow --everything
...                              
*.gitattributes	69 B	2/2 files(s)	100%
*.txt          	1 B 	1/1 files(s)	100%

LFS Objects    	1 B 	1/1 files(s)	100%
```

We also adjust the sorting logic for the summary output to make it fully deterministic; previously, if two entries for different file extensions had the same totals, they might appear in any order, but now we sort them using Go's default string comparison ordering.

In the `t/t-migrate-info.sh` test suite, we remove a spurious early `exit 0` which has been in place for some time, meaning the last test was always skipped.  We then add additional tests of the new `--pointers` option and its various settings, with and without Git LFS objects in the test repositories.

We also add, for good measure, a pair of tests which check the behaviour of the `git lfs migrate info` command when there are files in the history which should be Git LFS objects according to the corresponding `.gitattributes` files but are not.  This is another type of situation where `git lfs migrate import --fixup` might be used, so we want to ensure the `info` subcommand produces correct output for it as well.

### DIFF
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -368,6 +368,7 @@ func init() {
 	info.Flags().IntVar(&migrateInfoTopN, "top", 5, "--top=<n>")
 	info.Flags().StringVar(&migrateInfoAboveFmt, "above", "", "--above=<n>")
 	info.Flags().StringVar(&migrateInfoUnitFmt, "unit", "", "--unit=<unit>")
+	info.Flags().StringVar(&migrateInfoPointers, "pointers", "", "Ignore, dereference, or include LFS pointer files")
 
 	importCmd := NewCommand("import", migrateImportCommand)
 	importCmd.Flags().StringVar(&migrateImportAboveFmt, "above", "", "--above=<n>")
--- a/commands/command_migrate_info.go
+++ b/commands/command_migrate_info.go
@@ -21,8 +21,8 @@ import (
 type migrateInfoPointersType int
 
 const (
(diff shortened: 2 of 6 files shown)


sami555555 (approved):

bk2204 (approved):
This looks good.

One note: since we're doing a 3.0 release, you can break backwards compatibility if you want and think that would be more intuitive.

chrisd8088:
Alright, thanks!  I've flipped the sense of the default around, then.  (Hopefully this work will get me closer to the `--fixup` option proposed in #4419.)

Issue #4403, opened 2021-02-18
Please update to Go 1.16 and also release for macOS ARM64


dbast:
Go 1.16 was just released with macOS ARM64 support, see https://blog.golang.org/go1.16

bk2204:
Hey,

Thanks for writing in.  We can certainly do this for the next major release.  However, it will require some manual work because we have to make sure binaries are signed in order to distribute them, and that means we need to run multiple steps during our build.  So while this certainly isn't impossible, it isn't immediately trivial, either.

Note that if Homebrew provides suitable binaries, you may want to use theirs in the mean time.

dbast:
Hey @bk2204,

thanks for looking into the topic.

The code signing stuff actually got much simpler as go 1.16 is adding an adhoc signature to the darwin_arm64 binary without any extra effort, which is good enough, see the long discussion inside https://github.com/golang/go/issues/42684

That is the reason why e.g. for gocloc building+releasing for darwin_arm64 was an one-liner https://github.com/hhatto/gocloc/pull/43 (they use also the GitHub action goreleaser/goreleaser-action@v1 which automatically also released the binary for that new arch... there are no extra signing calls there)

Downloading the darwin_arm64 binary and checking the signature shows:
* `wget https://github.com/hhatto/gocloc/releases/download/v0.4.1/gocloc_darwin_arm64.tar.gz`
* `tar -xf gocloc_darwin_arm64.tar.gz`
* `codesign -d -v gocloc`:
```
Executable=/Users/.../gocloc
Identifier=a.out
Format=Mach-O thin (arm64)
CodeDirectory v=20400 size=90110 flags=0x20002(adhoc,linker-signed) hashes=2813+0 location=embedded
Signature=adhoc
Info.plist=not bound
TeamIdentifier=not set
Sealed Resources=none
Internal requirements=none
```

Thanks

bk2204:
Unfortunately, unless macOS has stopped suggesting that binaries downloaded from the Internet that aren't signed by an Apple Developer account and notarized be moved to the trash, we actually need to get real signing set up.  So this isn't as easy as it seems.  I'm also not very excited about doing one thing for amd64 and another for arm64.

bk2204:
[...]

PR #4437, opened 2021-03-12
Darwin ARM64 support

bk2204:
macOS has recently started supporting ARM64 (Apple Silicon) in Big Sur.  Go 1.16 now supports this target, so let's build a suitable binary as part of our release process.

There are several changes here:
- Add support for module versions suitable for Go 1.16.
- Switch to the Ubuntu 20.04 version, since it isn't clear if the Ubuntu 18.04 versions will natively support Go 1.16.
- Switch to Go 1.16 for building.
- Build and upload the ARM64 targets for Darwin.

Note that now that we're using Go 1.16, we must be careful about using cgo.  Go 1.16 has dropped support for using the 387 on 32-bit x86 targets (golang/go#40255), but outside of Darwin (which no longer supports 32-bit targets), the 387 is part of the x86 platform ABI when calling functions with floating-point arguments, and not using it will break things.  We must therefore be careful not to use cgo with floating-point numbers, which we do not presently do.

The plan is to do a point release with this change and some bug fixes while we finalize things for 3.0.0.

/cc #4403
/cc @dbast as reporter


script/upload
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-REPO="git-lfs/git-lfs"
+REPO="git-lfs/git-lfs-ci-test"
```
chrisd8088, line 5: Is this change from commit a46bc9d9770587f974c5ee5a6e4582380287a8a4 (with the commit message "WIP" and no description) still desired?

chrisd8088 (approved):
This is great news, thank you!  ❤️   I had one question about a "WIP" commit, and one other thing I'm a little unclear on is the change from `ubuntu-latest` to `ubuntu-20.04`; the commit description for that change reads:

> Ubuntu 20.04 is the latest LTS version and it provides the newest
software available for us to build.  More importantly, it appears that
the latest Go, 1.16, will be available for Ubuntu 20.04 but possibly not
for Ubuntu 18.04, so because we want at least that version in order to
successfully build for ARM64 macOS, let's prefer the latest Ubuntu.

But we're not preferring the latest Ubuntu; we're fixing it at 20.04 -- which is fine, since it supports Go 1.16, but it's not quite the same thing.  I notice the GitHub Actions docs are a [little](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) [conflicted](https://github.com/actions/virtual-environments#available-environments) on the subject of whether `-latest` is now 20.04 in all cases -- was the change here to avoid ending up on a 18.04 system?

script/upload
bk2204: No, that should be dropped.  I've done so.

bk2204:
> This is great news, thank you! heart I had one question about a "WIP" commit, and one other thing I'm a little unclear on is the change from `ubuntu-latest` to `ubuntu-20.04`; the commit description for that change reads:
> 
> > Ubuntu 20.04 is the latest LTS version and it provides the newest
> > software available for us to build.  More importantly, it appears that
> > the latest Go, 1.16, will be available for Ubuntu 20.04 but possibly not
> > for Ubuntu 18.04, so because we want at least that version in order to
> > successfully build for ARM64 macOS, let's prefer the latest Ubuntu.
> 
> But we're not preferring the latest Ubuntu; we're fixing it at 20.04 -- which is fine, since it supports Go 1.16, but it's not quite the same thing. I notice the GitHub Actions docs are a [little](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) [conflicted](https://github.com/actions/virtual-environments#available-environments) on the subject of whether `-latest` is now 20.04 in all cases -- was the change here to avoid ending up on a 18.04 system?

Yes, unfortunately, for Actions, `ubuntu-latest` isn't really the latest; it's the default.  That name, in retrospect, was perhaps suboptimal.  `ubuntu-latest` is in the process of transitioning to 20.04, but it may still be 18.04 in some cases.  It isn't clear to me whether Go 1.16 will be available on 18.04; the issue mentioning adding it said it would be available on 20.04, macOS, and Windows.

Regardless, we want to use the actual latest version, not the "latest" version, and it's better for us to make this transition now while we're in control of it rather than having an unpleasant surprise down the line.  In this case, no changes are needed, and I didn't expect any.  But I also have not enjoyed putting out emergency fixes to our CI in the past when things have suddenly broken and someone opens a PR, so I'd rather let us make the decision at a convenient time for us.

Hopefully that makes things a little clearer.

PR #4438, opened 2021-03-12
Multipart transfer mode proposal

shevron:


docs/proposals/multipart_trasnfer_mode.md
```diff
@@ -0,0 +1,296 @@
...
+{
+  "transfer": "multipart",
+  "objects": [
+    {
+      "oid": "20492a4d0d84f8beb1767f6616229f85d44c2827b64bdbfb260ee12fa1109e0e",
+      "size": 10000000,
+      "actions": {
+        "parts": [
+          {
+            "href": "https://foo.cloud.com/storage/upload/20492a4d0d84?part=0",
```
bk2204, line 116: Can we use the RFC 6761 TLD `.example` here and below (e.g., `foo.cloud.example` and `lfs.mycompany.example`)?  That way we won't accidentally conflict with a real domain.

docs/proposals/multipart_trasnfer_mode.md
```diff
@@ -0,0 +1,296 @@
...
+parts may take several hours from `batch` to `verify`.
+
+### Falling back to `basic` transfer for small files
+Using multipart upload APIs has some complexity and speed overhead, and for this reason it is recommended that servers
+implement a "fallback" to `basic` transfers if the uploaded object is small enough to handle in a single part.
+
+Clients *should* support such fallback natively, as it rides on existing transfer method negotiation capabilities.
+
+The server can simply respond with `{"transfer": "basic", ...}`, even if `mutipart` was request by the client
+and *is supported* by the server in order to achieve this.
```
bk2204, line 275: Typo: “mutipart” for “multipart”.  Also, let's not make the transfers magical, so let's permit servers to fall back if both `basic` and `multipart` were specified.  That makes it a lot easier for the client to have a little more control over things, and theoretically have a simple, multipart-only implementation if they know the server they use supports that.

docs/proposals/multipart_trasnfer_mode.md
```diff
@@ -0,0 +1,296 @@
...
+of the request. This is a useful capability even if not required, as it allows backends to validate each part
+separately as it is uploaded.
+
+To support this, `parts` request objects may include a `want_digest` value, which may be any value specified by
+[RFC-3230](https://tools.ietf.org/html/rfc3230) or [RFC-5843](https://tools.ietf.org/html/rfc5843) (the design for this
+feature is highly inspired by these RFCs).
+
+Possible values include a comma-separated list of q-factor flagged algorithms, such as SHA-256 and SHA-512.
+Of one or more of these are specified, the digest of the payload is to be specified by the client as part of
+the Digest header, using the format specified by [RFC-3230 section 4.3.2](https://tools.ietf.org/html/rfc3230#section-4.3.1).
```
bk2204, line 189: We say “section 4.3.2” but seem to link to section 4.3.1.

docs/proposals/multipart_trasnfer_mode.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+## Uploaded Part Digest
+Some storage backends will support, or even require, uploading clients to send a digest of the uploaded part as part
+of the request. This is a useful capability even if not required, as it allows backends to validate each part
+separately as it is uploaded.
+
+To support this, `parts` request objects may include a `want_digest` value, which may be any value specified by
+[RFC-3230](https://tools.ietf.org/html/rfc3230) or [RFC-5843](https://tools.ietf.org/html/rfc5843) (the design for this
+feature is highly inspired by these RFCs).
```
bk2204, line 185: Let's specify that it can be any digest registered with IANA via the process outlined in RFC 3230.  Maybe someone wants to specify SHA-3 in a future RFC, and we'd want to allow that.

docs/proposals/multipart_trasnfer_mode.md
```diff
@@ -0,0 +1,296 @@
...
+
+Possible values include a comma-separated list of q-factor flagged algorithms, such as SHA-256 and SHA-512.
+Of one or more of these are specified, the digest of the payload is to be specified by the client as part of
+the Digest header, using the format specified by [RFC-3230 section 4.3.2](https://tools.ietf.org/html/rfc3230#section-4.3.1).
+
+It should be noted that digests based on `MD5` or `SHA-1` algorithms are not supported by this transfer mode,
+as they are no longer to be considered secure. Therefore, the `contentMD5`, `MD5` and `SHA` values are not to be used as
+part of `want_digest` headers.
```
bk2204, line 193: Let's also specify that we only allow cryptographically secure hash algorithms in this field.  For example, one can specify `adler32` or `unixcksum` in a Digest header, but we're not going to implement that, either, because as a practical matter nobody uses those.

docs/proposals/multipart_trasnfer_mode.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+#### Example of part upload request send to the storage server
+Following on the `want_digest` value specified in the last example, the client should now send the following headers
+to the server when uploading the part:
+
+```
+HTTP/1.1 PUT /storage/upload/20492a4d0d84?part=3
+Authorization: Bearer someauthorizationtokenwillbesethere
+Digest: SHA-256=thvDyvhfIqlvFe+A9MYgxAfm1q5=,SHA-512=thvDyvhfIqlvFe+A9MYgxAfm1q5thvDyvhfIqlvFe+A9MYgxAfm1q5=
```
bk2204, line 227: Can we allow the client to specify just one of these?  I always viewed the negotiation as allowing the client to specify just the most preferred option negotiated.

docs/proposals/multipart_trasnfer_mode.md
```diff
@@ -0,0 +1,296 @@
...
+
+### `parts` actions
+Each `parts` action should include the `pos` and `size` attributes, in addition to the attributes specified
+above:
+* `pos` indicate the position in bytes within the file in which the part should begin. If not specified, `0` (that is the
+beginning of the file) is assumed.
+* `size` is the size of the part in bytes. If `size` is omitted, default to read until the end of file.
+* If both `pos` and `size` are omitted, the action is expected to be a single-part upload of the entire file
+
+In adition, `parts` actions may include the following parameters:
```
bk2204, line 73: Typo: “adition” for “addition”.

docs/proposals/multipart_trasnfer_mode.md
```diff
@@ -0,0 +1,296 @@
...
+* `method`, with `PUT` as the default method if none is specified. This allows customizing the HTTP method used when uploading
+object parts.
+* `want_digest` to specify an expected HTTP `Digest` header, as described below.
+
+### `verify` action
+The `verify` action is similar to `basic` transfer mode `verify`, with the following additional parameters:
+* `params` - an object with additional parameters to send to the server as part of the `POST` JSON body.
+  These parameters are to be sent as provided back to the server.
```
bk2204, line 81: Can we specify that this `params` structure is passed as a `params` structure?  That is, that our response looks like this:

```
{
  "params": {
    "a": 256,
    "b": "string",
    "c": null
  }
}
```

Doing so means we can add data in the future easier and it also means that implementing parsing of this field in strongly typed languages like Go and Rust is a _lot_ easier than if we insert them randomly into the top-level body.

docs/proposals/multipart_trasnfer_mode.md
```diff
@@ -0,0 +1,296 @@
...
+        },
+        "abort": {
+          "href": "https://foo.cloud.com/storage/upload/20492a4d0d84",
+          "authenticated": true,
+          "header": {
+            "Authorization": "Basic 123abc123abc123abc123abc123="
+          },
+          "method": "DELETE",
+          "expires_in": 86400
+        }
```
bk2204, line 171: Can we provide example responses of what the `verify` and `abort` requests look like below?

bk2204 (commented):
This looks great.  I left some comments about a few changes I'd like to see, but overall I'm very happy with the direction this is going.

docs/proposals/multipart_trasnfer_mode.md
shevron: Yes, that was generally my intention. The assumption was that clients should continue to always support and specify `basic`, but I'm ok with dropping that assumption and saying something like "if the client specified both `multipart` and `basic` as requested transfer modes, the server *may* decide to fall back to `basic` transfer if ..."

docs/proposals/multipart_trasnfer_mode.md
shevron: Yes, the client is allowed to do so, as per the `Digest` RFC. Perhaps the example should emphasize that approach instead of sending both.

docs/proposals/multipart_trasnfer_mode.md
shevron: Fixed, BTW `.example` TLD is RFC-2606, RFC-6761 AFAIK is `example.com` and friends :wink:

docs/proposals/multipart_trasnfer_mode.md
shevron: Typo fixed

docs/proposals/multipart_trasnfer_mode.md
shevron: Added below.

shevron:
@bk2204 I updated the proposal based on your feedback, resolved the trivial stuff and left the more complex stuff open for you to review and resolve. 

I've also decided to drop the `params` attribute from `abort` requests as I feel it complicates things and won't be needed. I see two options here - either aborting is a simple request with no body that could be sent directly to the storage backend, or it is more complex and may require some parameters. In the latter case, the LFS server could take care of handling `abort` and get the parameters it needs to do so via the URL.

bk2204 (approved):
This looks great.  Thanks for working with me on the changes.

bk2204:
I'll just note for folks who might come later that there is the slightest possibility that, since this is a proposal, things might change as we implement them in the client and test server and start testing, but I feel confident such changes will be minor.

PR #4442, opened 2021-03-16
docs: note that -I and -X override configuration settings

### LINKED ISSUE #4425: git lfs pull --exclude="" ambiguous behavior
Some of my lfs files are ignored through the `.lsconfig`
```
[lfs]
    fetchexclude = "my/path/to/large/files/*"
```

I know that `--exclude=""` overrides the config and downloads those files. This is what I want for CI and production builds.

Initially I thought that `--exclude` would recursively join with the file config, not override it. It was kind of unclear and took a while to figure out. 

I'm also afraid that eventually someone might ask them to be applied together and then production builds will be missing important binaries.

bk2204:
Note in the documentation that -I and -X override their respective configuration settings so it's clear what behavior people can expect. Mention that setting either value to an empty string clears the value.

Fixes #4425

### DIFF
--- a/docs/man/git-lfs-clone.1.ronn
+++ b/docs/man/git-lfs-clone.1.ronn
@@ -44,6 +44,9 @@ of paths to include/exclude in the fetch (wildcard matching as per gitignore).
 Only paths which are matched by fetchinclude and not matched by fetchexclude
 will have objects fetched for them.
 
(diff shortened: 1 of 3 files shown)


chrisd8088 (approved):
👍 Thanks!

Issue #4288, opened 2020-10-22
Git lfs prune hangs on windows


jflefebvre06:
**Describe the bug**
When running git lfs prune with lfs version > 2.10.0, the process freeze

![image](https://user-images.githubusercontent.com/2685607/96893101-a3bc4880-148a-11eb-8c06-535bbe1a6fea.png)

![image](https://user-images.githubusercontent.com/2685607/96893155-b171ce00-148a-11eb-85c9-d2e6167a1108.png)

The only solution is to kill the GIT-LFS process. `CTRL+C` does not works.

**To Reproduce**
`git lfs prune --no-verify-remote --verbose`

Works fine with 2.10.0, breaks with 2.11 and 2.12

**System environment**
Windows 10

**Output of `git lfs env`**

```
D:\@WAC>git lfs env
git-lfs/2.12.0 (GitHub; windows amd64; go 1.14.7; git dad7b9e6)
git version 2.29.0.windows.1

Endpoint=https://XXX@YYY/wac.git/info/lfs (auth=none)
LocalWorkingDir=C:\@WAC
LocalGitDir=C:\@WAC\00.git
LocalGitStorageDir=C:\@WAC\00.git
LocalMediaDir=C:\@WAC\00.git\lfs\objects
LocalReferenceDirs=
TempDir=C:\@WAC\00.git\lfs\tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
FetchRecentRefsIncludeRemotes=true
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=C:\@WAC\00.git\lfs
AccessDownload=none
AccessUpload=none
DownloadTransfers=basic,lfs-standalone-file
UploadTransfers=basic,lfs-standalone-file
GIT_EXEC_PATH=C:/Program Files/Git/mingw64/libexec/git-core
git config filter.lfs.process = "git-lfs filter-process"
git config filter.lfs.smudge = "git-lfs smudge -- %f"
git config filter.lfs.clean = "git-lfs clean -- %f"
```

**Additional context**
Any other relevant context about the problem here.

bk2204:
Hey,

[...]

PR #4449, opened 2021-03-23
Parse stash log entries parsimonously in prune command

### LINKED ISSUE #4439: git lfs prune never completes after the most recent git update
**Describe the bug**

This has been verified by multiple colleagues of mine.

Running `git lfs prune` before git update finished normally. After the most recent git update (`2.30.2`) `git lfs prune` never finished. In fact, I am getting more objects retained, in the previous runs, before the update, I had around `108k` objects retained

image

Downgrading to `2.30.1` did not help

**To Reproduce**

1. Upgrade to the latest version of git for windows (with the security fix) `v2.30.2`
2. Clone an lfs repository
3. `git lfs prune`
4. prune never completes

**Expected behavior**
`git lfs prune` should complete in a timely manner

**System environment**

```
λ git config -l
http.sslcainfo=C:/Program Files/Git/mingw64/ssl/certs/ca-bundle.crt
http.sslbackend=openssl
diff.astextplain.textconv=astextplain
filter.lfs.required=true
credential.helper=manager-core
core.autocrlf=true
core.fscache=true
core.symlinks=false
pull.rebase=false
init.defaultbranch=master
user.name=samaursa
user.email=samaursa@tuquegames.com
user.admin=true
diff.renamelimit=10000
core.autocrlf=false
core.preloadindex=true
core.fscache=true
alias.dry=!f() { git merge-tree `git merge-base $2 $1` $2 $1; }; f
gc.auto=256
filter.lfs.required=true
lfs.pruneoffsetdays=0
core.filemode=false
core.bare=false
core.logallrefupdates=true
core.symlinks=false
core.ignorecase=true
branch.master.remote=origin
```

**Output of `git lfs env`**
```
λ git lfs env
git-lfs/2.13.2 (GitHub; windows amd64; go 1.14.13; git fc664697)
git version 2.30.2.windows.1

LocalReferenceDirs=
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=0
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
AccessDownload=basic
AccessUpload=basic
GIT_INSTALL_ROOT=C:\Program Files\Git
```

**Additional context**
Our repository is very big, with the `.git/lfs` size greater than `300GB`. Usually, `git lfs prune`, with the default settings, takes a few minutes to complete. Immediately after the upgrade, the prune would never finish.

Edit: removing sensitive info

chrisd8088:
When looking for Git LFS pointers in the Git stash in order to avoid pruning the corresponding local objects during a `git lfs prune` command (functionality which was added in #4209), we want to avoid looking further back in the Git log file than necessary; otherwise, we may potentially find many LFS object references which we should prune but incorrectly consider to be part of the stashes.

We therefore proceed in two steps; first looking solely at each stash entry's WIP merge commit using the command `git log -m --first-parent <stash-sha>^..<stash-sha>`, and then looking at each entry's index merge commit and possible additional untracked files merge commit using the command `git log <stash-sha>^..<stash-sha>`.  By limiting these commands' commit ranges to just the stash merge commits, we skip improperly traversing the full log.

We update the relevant tests to include older commits, one of which should be pruned in each case; note that
with these changes the tests fail with the previous version of the `git lfs prune` logic for stashes.  We also add stashes to two other tests and confirm that in both cases the LFS objects from the stashed commits are retained, but no additional objects are also unexpectedly retained as a result of the stash.

Further, we make sure the three dedicated stash tests work with more than a single stash entry each, stashing work on a branch which is then deleted.  We expect the stashed LFS content to be retained while the unreferenced LFS objects committed on the deleted branch are pruned.  In the case where we test this with `git stash -u`, we need to ensure we first commit the `.gitattributes` file created by `git lfs track`, because otherwise our first `git stash -u` removes it and then we are unable to create LFS objects after checking out the to-be-deleted-later branch.

(It may be very marginally easier to review this PR by ignoring whitespace changes.)

Fixes #4439 and (part of) #4288.
/cc @samaursa as reporter.

### DIFF
--- a/lfs/gitscanner_log.go
+++ b/lfs/gitscanner_log.go
@@ -83,7 +83,7 @@ func scanStashed(cb GitScannerFoundPointer, s *GitScanner) error {
 	// older Git versions (at least <=2.7) don't report merge parents in
 	// the reflog, we can't extract the parent SHAs from "Merge:" lines
 	// in the log; we can, however, use the "git log -m" option to force
-	// individual diffs of all the merge parents in a second step.
+	// an individual diff with the first merge parent in a second step.
 	logArgs := []string{"-g", "--format=%h", "refs/stash", "--"}
 
 	cmd, err := git.Log(logArgs...)
@@ -95,7 +95,8 @@ func scanStashed(cb GitScannerFoundPointer, s *GitScanner) error {
 
 	var stashMergeShas []string
 	for scanner.Scan() {
-		stashMergeShas = append(stashMergeShas, strings.TrimSpace(scanner.Text()))
+		stashMergeSha := strings.TrimSpace(scanner.Text())
+		stashMergeShas = append(stashMergeShas, fmt.Sprintf("%v^..%v", stashMergeSha, stashMergeSha))
 	}
 	err = cmd.Wait()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


bk2204 (approved):

chrisd8088:
@bk2204 -- I've added a few commits to this PR while expand the tests, as the updated description now mentions.

bk2204 (approved):

PR #4454, opened 2021-03-26
Make migrate commands default to preserving uncommitted changes

### LINKED ISSUE #3741: Improve warning that git lfs migrate will destroy uncommitted changes
As suggested by @bk2204's [comment](https://github.com/git-lfs/git-lfs/issues/3123#issuecomment-515999144), this issue is a follow-up to #3123.

There, [I asked](https://github.com/git-lfs/git-lfs/issues/3123#issuecomment-515707490) that the warning for the command `git lfs migrate import --include="*.csv"` (or other extensions) about deleting uncommitted changes should be a lot more flashy! Perhaps a red color and/or all-caps message asking for confirmation twice that the user understands proceeding with the migration can result in loss of data.

chrisd8088:
We update the prompt to be more explicit when the import (or export) migration commands notice a "dirty" working tree with uncommitted changes, and we also revise the default action in this case to be to exit early and preserve the work in progress.

Thanks to @bk2204 for the suggested revisions (see https://github.com/git-lfs/git-lfs/issues/3741#issuecomment-516045440).

Fixes #3741.
/cc @janosh as reporter.


janosh:
Much appreciated!￼

bk2204 (approved):

PR #4458, opened 2021-03-30
Update git-lfs-migrate man page and add description section

### LINKED ISSUE #4408: feature/add: lfs migrate import --remote
It seems that there is no `--remote` flag for `lfs migrate import`, only `export`. 

It's easy enough to workaround by renaming the original origin: 
```
git remote rename origin oldorigin
git remote add git@gitlab.com:group/project.git
git lfs migrate import --everything --include="*.whatever"
```

However, would still be nice to specify the remote in the import command. 
Does this a) already exist and simply isn't documented, or b) not make sense as a feature due to my not fully understanding how import works?

chrisd8088:
We add a lengthy description section to the `it-lfs-migrate` man page to order to provide some context and general guidance on how and when the command should be used, as well as trying to explain several potentially non-intuitive concepts and encourage users to check their repositories thoroughly before and after attempting an import or export operation.

We also adjust some formatting details in the page, fix some typos, and make other minor revisions to suggest using the `--all` option when force-pushing after importing with `--everything`, stop implying that only large files are ever imported or exported, and avoid stating that a single use case is definitely more prevalent than all others.

And we allow man page headers to include parentheses so that the specialized headers in the `git-lfs-migrate` man page (e.g., `"IMPORT (NO REWRITE)"`) are supported and properly linked within the rendered page.

Fixes #4408.
/cc @glass-ships as reporter.

### DIFF
--- a/docs/man/git-lfs-migrate.1.ronn
+++ b/docs/man/git-lfs-migrate.1.ronn
@@ -5,6 +5,100 @@ git-lfs-migrate(1) - Migrate history to or from Git LFS
 
 `git lfs migrate` <mode> [options] [--] [branch ...]
 
+## DESCRIPTION
+
+Convert files in a Git repository to or from Git LFS pointers, or
+summarize Git file sizes by file type.  The `import` mode converts Git
+files (i.e., blobs) to Git LFS, while the `export` mode does the reverse,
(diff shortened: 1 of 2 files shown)


bk2204 (approved):
This is a great improvement to the documentation.

PR #4468, opened 2021-04-08
Mark skipdownloaderrors as safe

### LINKED ISSUE #4467: lfs.skipdownloaderrors is not considered a safe key
**Describe the bug**
If you put `lfs.skipdownloaderrors` in `.gitattributes` then Git LFS ignores it with the error

```
WARNING: These unsafe lfsconfig keys were ignored:

  lfs.skipdownloaderrors
```

**To Reproduce**
1. Create a repo using LFS.
2. Add `.gitattributes` with `lfs.skipdownloaderrors` set to `true`.
3. `git push`

**Expected behavior**
It shouldn't print a warning, and when you clone the repo it should respect the setting and ignore download errors.

**System environment**
    git-lfs/2.13.2 (GitHub; darwin amd64; go 1.15.6)

**Output of `git lfs env`**
```
git-lfs/2.13.2 (GitHub; darwin amd64; go 1.15.6)
git version 2.30.0

WARNING: These unsafe lfsconfig keys were ignored:

  lfs.skipdownloaderrors
Endpoint=redacted (auth=none)
  SSH=redacted
Local*=redacted
LocalReferenceDirs=
TempDir=redacted
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=redacted
AccessDownload=none
AccessUpload=none
GIT_ASKPASS=/Applications/Visual Studio Code.app/Contents/Resources/app/extensions/git/dist/askpass.sh
GIT_EXEC_PATH=/usr/local/Cellar/git/2.30.0/libexec/git-core
```

**Additional context**

This issue is basically the same as #2490. It's hard to see how `skipdownloaderrors` could possibly be unsafe so I think it should be whitelisted.

Timmmm:
This option can't be used to do anything dangerous.

Fixes #4467

### DIFF
--- a/config/git_fetcher.go
+++ b/config/git_fetcher.go
@@ -190,4 +190,5 @@ var safeKeys = []string{
 	"lfs.locksverify",
 	"lfs.pushurl",
 	"lfs.url",
+	"lfs.skipdownloaderrors",
 }
--- a/config/git_fetcher.go
+++ b/config/git_fetcher.go
@@ -189,6 +189,6 @@ var safeKeys = []string{
 	"lfs.gitprotocol",
 	"lfs.locksverify",
 	"lfs.pushurl",
-	"lfs.url",
 	"lfs.skipdownloaderrors",
+	"lfs.url",
 }
--- a/docs/man/git-lfs-config.5.ronn
+++ b/docs/man/git-lfs-config.5.ronn
@@ -405,6 +405,7 @@ including and limited to:
 - lfs.url
 - remote.{name}.lfsurl
 - lfs.{*}.access
+- lfs.skipdownloaderrors
 
 The set of keys allowed in this file is restricted for security reasons.
 
--- a/docs/man/git-lfs-config.5.ronn
+++ b/docs/man/git-lfs-config.5.ronn
@@ -402,10 +402,10 @@ including and limited to:
 - lfs.gitprotocol
 - lfs.locksverify
 - lfs.pushurl
-- lfs.url
-- remote.{name}.lfsurl
-- lfs.{*}.access
 - lfs.skipdownloaderrors
+- lfs.url
+- lfs.{*}.access
+- remote.{name}.lfsurl
 
 The set of keys allowed in this file is restricted for security reasons.
 


config/git_fetcher.go
```diff
@@ -190,4 +190,5 @@ var safeKeys = []string{
 	"lfs.locksverify",
 	"lfs.pushurl",
 	"lfs.url",
+	"lfs.skipdownloaderrors",
```
chrisd8088, line 193: One suggestion to keep these in alphabetical order:

```suggestion
	"lfs.skipdownloaderrors",
	"lfs.url",
```

docs/man/git-lfs-config.5.ronn
```diff
@@ -405,6 +405,7 @@ including and limited to:
 - lfs.url
 - remote.{name}.lfsurl
 - lfs.{*}.access
+- lfs.skipdownloaderrors
```
chrisd8088, line 408: In this case, we might also fix the ordering of `lfs.{*}.access` and `remote.{*}.access` while we're at it.

```suggestion
- lfs.skipdownloaderrors
- lfs.url
- lfs.{*}.access
- remote.{name}.lfsurl
```

chrisd8088 (changes requested):
Hey, thanks for the bug report and patch!

I think this should be OK, but I'm wondering if you would mind reordering the new lines so they fall in alphabetical order in the respective lists, something like the suggestions I added (and in one case, it could help fix another ordering typo in the docs).

Thanks again very much for this PR, and welcome to the project!

Timmmm:
Ooo I can just click buttons - fancy. Thanks for the quick reply!

chrisd8088 (approved):
Thanks very much, this looks good to me!  Thanks again for the bug report and patch; it's much appreciated.

PR #4476, opened 2021-04-12
Make lfshttp package builds more portable

### LINKED ISSUE #4475: Import NetBSD pkgsrc patches to support Solaris
NetBSD pkgsrc, which, despite the name, can be used on a variety of OSes, not just NetBSD, contains [a patch to support Solaris](https://github.com/NetBSD/pkgsrc/blob/trunk/devel/git-lfs/patches/patch-lfshttp_certs__solaris.go) which we should import.  Presumably this is sufficient to cause Git LFS to otherwise work on Solaris, but lacking a Solaris system, I can't be sure.

I suspect the case is that without this change, the code fails to build altogether on Solaris, which we should be able to test with a cross-compile.  If so, we may want to refactor out the code for non-Darwin Unix systems into a single file with a build directive and cover FreeBSD, NetBSD, DragonflyBSD, OpenBSD, Solaris, and Linux.

bk2204:
Currently, we have a large number of identical cert files for various platforms, all of which are non-macOS Unix systems.  These files all simply provide the default system configuration and nothing else.

We recently noticed by looking at the NetBSD pkgsrc repository, which, despite its name, can be used on many systems, that there is a patch to support Solaris.  Unsurprisingly, it duplicates this same file.

Since using the default system behavior is the prudent thing to do if we don't have anything else to do, let's make our project more portable by simply using this same file for all systems other than macOS or Windows. That way, we'll work on Solaris and NetBSD, among other systems, and any other Unix systems which support Go in the future will also magically work.

Fixes #4475

### DIFF
--- a/lfshttp/certs_freebsd.go
+++ b/lfshttp/certs_freebsd.go
@@ -1,8 +1,0 @@
-package lfshttp
-
-import "crypto/x509"
-
-func appendRootCAsForHostFromPlatform(pool *x509.CertPool, host string) *x509.CertPool {
-	// Do nothing, use golang default
-	return pool
-}
--- a/lfshttp/certs_linux.go
+++ b/lfshttp/certs_linux.go
@@ -1,8 +1,0 @@
-package lfshttp
-
-import "crypto/x509"
-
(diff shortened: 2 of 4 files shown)


chrisd8088 (approved):
Another 👍

PR #4478, opened 2021-04-14
Fix downloadFile in gitfilter_smudge.go to actually propagate all errors

slonopotamus:
Before this commit, loop over errors didn't actually loop
and instead returned only the first error

----

I have to idea how to test this change. I found it because GoLand was complaining that condition `multiErr != nil` inside loop over errors was always false.


bk2204 (approved):
Yeah, this looks like an obviously correct fix.  Thanks for the patch.

PR #4482, opened 2021-04-14
Bump go-spnego to the latest version

bk2204:
This version uses a newer version of the Kerberos library, which should work better in some environments.  Tidy the go.mod and go.sum files and vendor the dependencies.

/cc #4479 
/cc @BenAigen as reporter


chrisd8088 (approved):
👍

PR #4490, opened 2021-04-26
fs: be a little less aggressive with cleanup

bk2204:
When we invoke the file helper to download files from a remote system, we use the repository's temporary directory in order to make sure we can quickly and easily rename files into the object store without the need for a copy, which might be necessary if we used the system temporary directory.  These files don't look like a normal object ID; instead, they have a name generated by Go's temporary file code.

However, this causes a problem if, during the pull, a file is checked out and Git causes it to be uselessly cleaned, which happens in some cases.  That's because our temporary file cleanup code will remove all files in the temporary directory that don't look like normal object ID components, and as a result, the clean operation can remove files that are in use by the fetch.

The solution here is simple: instead of immediately purging anything that looks like it's not a valid object ID, let's wait an hour before pruning any temporary file unless it's an object that we've already downloaded and verified is in our object store.  This won't leave significantly more cruft around, but it will avoid this problem.


lfshttp/standalone/standalone.go
```diff
@@ -244,6 +245,15 @@ func (h *fileHandler) download(oid string, size int64) (string, string, error) {
 		return oid, "", err
 	}
 
+	// Set the timestamp so if we hard link this file into a temporary
+	// location, our temporary files don't get deleted because they're too
+	// old.
+	now := time.Now()
+	err = os.Chtimes(src, now, now)
```
slonopotamus, line 252: I'm a bit worried by performance effect of such change. Unfortunately, if `lfs.LinkOrCopy` ends up doing `os.Link`, we don't get a change to change mtime.

On the other side, if we're able to do `os.Link(src, dst)` we don't need temp file at all!

So, I believe a better fix is:

1. Try to do `os.Link(src, dst)`. If it succeeded (either directly or via a race condition where a parallel process managed to create `dst` earlier than us), end.
2. Create temp, copy src -> temp
3. `os.Rename(temp, dst)`.

But it looks like current level of abstraction doesn't provide us enough context in order to perform `os.Link(src, dst)`?

bk2204: We do need the temporary file regardless, because the standalone transfer protocol doesn't let us write directly into the LFS storage.  We have to provide a download location for the file so that the main process can copy or link it into the repository.

slonopotamus: That's unfortunate. Direct `os.Link` from src to LFS storage would kill two birds with one stone - no need to write to src and no issues with temp file mtime.

bk2204: I just realized this won't work if the remote repository is read only or otherwise doesn't have permissions.  I'll need to revise this.

bk2204: Fixed.

chrisd8088 (approved):
This looks great, thank you for all the work on it!

PR #4492, opened 2021-04-28
Use host architecture and OS when running `go generate`

sergiou87:
I was trying to [cross-compile git-lfs to macOS for arm64](https://github.com/desktop/dugite-native/pull/354) from an x86_64 machine and I noticed that just [passing a `GOARCH=arm64` variable to `make`](https://github.com/desktop/dugite-native/pull/354/files#diff-b079a6aebd2c75ac12727b1cfcf5ffb9ddf9055a823e3015d7c729fac1628f93R81) would cause a failure like this:
```
go generate github.com/git-lfs/git-lfs/commands
fork/exec /var/folders/4d/hy7gsq2n22jccp96wdbpt3100000gn/T/go-build614314336/b001/exe/mangen: bad CPU type in executable
commands/commands.go:28: running "go": exit status 1
make: *** [commands/mancontent_gen.go] Error 1
```

I don't understand _why_ exactly, but [overriding `GOARCH` with _nothing_](https://github.com/desktop/dugite-native/pull/354/files#diff-b079a6aebd2c75ac12727b1cfcf5ffb9ddf9055a823e3015d7c729fac1628f93R80) (to effectively use the host CPU arch) for the specific `go generate` command fixed the issue. However, I don't know if this could cause issues somewhere else or if there is a better way to fix this problem.

Please, let me know what you think 😄

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,7 +196,7 @@ mangen : commands/mancontent_gen.go
 # 'commands' of Git LFS. It depends upon the contents of the 'docs' directory
 # and converts those manpages into code.
 commands/mancontent_gen.go : $(wildcard docs/man/*.ronn)
-	GOARCH= $(GO) generate github.com/git-lfs/git-lfs/commands
+	GOOS= GOARCH= $(GO) generate github.com/git-lfs/git-lfs/commands
 
 # Targets 'all' and 'build' build binaries of Git LFS for the above release
 # matrix.
--- a/Makefile
+++ b/Makefile
@@ -196,7 +196,7 @@ mangen : commands/mancontent_gen.go
 # 'commands' of Git LFS. It depends upon the contents of the 'docs' directory
 # and converts those manpages into code.
 commands/mancontent_gen.go : $(wildcard docs/man/*.ronn)
-	$(GO) generate github.com/git-lfs/git-lfs/commands
+	GOARCH= $(GO) generate github.com/git-lfs/git-lfs/commands
 
 # Targets 'all' and 'build' build binaries of Git LFS for the above release
 # matrix.


bk2204 (commented):
Hey,

Thanks for the patch, and welcome to Git LFS!

There are two ways to accomplish what you're looking for here.  One of them is to do this and _also_ add `GOOS` so folks can cross-compile across operating systems.

The other approach is to use the specific Makefile target for those OS/architecture combinations, such as `bin/git-lfs-darwin-arm64`.  These are what we use for building our own releases, and it avoids this problem altogether.

I don't have a strong preference about which way to go; we can take this patch if you're willing to add `GOOS` here, or, if you don't want to do that, you can just use the existing targets.

sergiou87:
Thanks for the quick reply!

I saw the specific Makefile targets, but dugite was already doing "manual" builds instead of using those targets: https://github.com/desktop/dugite-native/blob/f221234dcd40d00702e43494e4ce5a7a3e271833/script/build-macos.sh#L63

Not sure if there is a problem with them, but I wanted to make the minimum possible changes :flushed:

I still don’t understand 100% why would GOARCH/GOOS affect go generate like that, specially given it’s only generating manpages content 🤔 

So I'd rather go with the other option you suggested, which if I understand you correctly is just adding `GOOS= GOARCH=` before the `go generate` command instead of just `GOARCH=`, is that right @bk2204?

bk2204:
> So I'd rather go with the other option you suggested, which if I understand you correctly is just adding `GOOS= GOARCH=` before the `go generate` command instead of just `GOARCH=`, is that right @bk2204?

Yup, that's definitely what I was suggesting, and I think it should be fine if we do that.

Just as a note, our main branch is in a holding pattern right now.  While we can merge this, we're not planning on doing a release until June, which will be our 3.0.0 release (because of the removal of NTLM) and so it may be a while before any change here makes it into a release.

sergiou87:
Yeah don't worry, we also like to be a bit behind the latest version, so I'll just use `sed` to change the Makefile before building git-lfs, and drop a comment with our conversation here 😄 

Thank you!! ❤️

bk2204 (approved):

bk2204:
I'll wait for CI to finish running and then merge this.

PR #4502, opened 2021-05-17
Bump transitive dependencies

bk2204:
When using Go 1.16 in CI, go get is upgrading some of our dependencies automatically, and since these items are not in the go.sum file, Go complains and refuses to build our binary.  Let's bump the relevant dependencies and vendor the resulting modules to make sure that Go can build our binary successfully on a fresh Go 1.16.


chrisd8088:
Thanks for tackling this -- I was also working on it, and it seems like it's due to the recent Go 1.16.4 release.

bk2204:
Yeah, I'm not sure what the relevant difference is from before, but apparently something is making things unhappy here, and this seems like the easiest way to fix it.  I tried a more aggressive upgrade, but that broke support for Go 1.12 and older (due to a dependency on the `errors` package).

chrisd8088 (approved):
LGTM, thank you!

PR #4509, opened 2021-05-24
locking: avoid nil pointer dereference with invalid response

### LINKED ISSUE #4508: A malformed server response to lock causes nil pointer deference
**Describe the issue**
When response from a server to the lock creation doesn't include "lock" or "message", `lockRes.lock` is nil when it is derefrenced [locking/locks.go:118](https://github.com/git-lfs/git-lfs/blob/c7c6901893c3712f0910421fab454a4f87d3f126/locking/locks.go#L118)

**System environment**
Debian, buster

bk2204:
If the server responds with a JSON response with neither a valid lock nor a valid message, we can end up performing a NULL pointer dereference.  Let's detect this case and return an error that indicates that the server has sent an invalid response.

I haven't added tests here because I think this is extremely unlikely to happen in practice and the consequence here is very minor, but if folks feel it's better to add one, I can do so.

Fixes #4508

### DIFF
--- a/locking/api.go
+++ b/locking/api.go
@@ -64,7 +64,14 @@ func (c *lockClient) Lock(remote string, lockReq *lockRequest) (*lockResponse, *
 	}
 
 	lockRes := &lockResponse{}
-	return lockRes, res, lfshttp.DecodeJSON(res, lockRes)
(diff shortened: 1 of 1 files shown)


chrisd8088 (approved):
Great, thanks!

PR #4527, opened 2021-06-11
README: improve steps for building from source

### LINKED ISSUE #4522: Expand requirements in README for installing from source
The install section for installing from source mentions a binary that has to be placed in PATH, which obviously cannot be right!

bk2204:
The `README` currently makes a lot of assumptions about the requirements to build from source.  Let's explain to the user what those requirements are and to invoke `make`, since we tend to build things differently from most other Go projects.

Fixes #4522
/cc @matthiasbeyer as reporter

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,10 @@ $ ./install.sh
 
 #### From source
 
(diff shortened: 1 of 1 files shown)


chrisd8088 (approved):
LGTM!

PR #4531, opened 2021-06-16
Let packagecloud.io recognize almalinux

saper:
This seems to be the change necessary to let almalinux-based system to be treated the same way as Red Hat Enterprise Linux or CentOS.

Workaround: set `os=el` in the environment

### DIFF
--- a/rpm/build_rpms.bsh
+++ b/rpm/build_rpms.bsh
@@ -14,7 +14,7 @@ else #Basically Centos 5/6
 fi
 
 case "${OS_NAME}" in
-  centos*|red*)
+  centos*|red*|almalinux)
     RPM_DIST=".el${VERSION_ID}"
     ;;
   fedora)
--- a/script/packagecloud.rb
+++ b/script/packagecloud.rb
@@ -43,6 +43,7 @@ $distro_name_map = {
     "sles/15.2",  # Current
   ],
   "centos/8" => [
+    "almalinux/8",
     "el/8",
     "fedora/32",
     "fedora/33",


saper:
This is what happens when trying to install git-lfs via packagecloud.io:

```
[root@0d1a459d4c11 tmp]# curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.rpm.sh > script.rpm.sh
[root@0d1a459d4c11 tmp]# ./script.rpm.sh
bash: ./script.rpm.sh: Permission denied
[root@0d1a459d4c11 tmp]# bash script.rpm.sh 
Detected operating system as almalinux/8.
Checking for curl...
Detected curl...
Downloading repository file: https://packagecloud.io/install/repositories/github/git-lfs/config_file.repo?os=almalinux&dist=8&source=script
curl: (22) The requested URL returned error: 404 


Unable to download repo config from: https://packagecloud.io/install/repositories/github/git-lfs/config_file.repo?os=almalinux&dist=8&source=script

This usually happens if your operating system is not supported by 
packagecloud.io, or this script's OS detection failed.

You can override the OS detection by setting os= and dist= prior to running this script.
You can find a list of supported OSes and distributions on our website: https://packagecloud.io/docs#os_distro_version

For example, to force CentOS 6: os=el dist=6 ./script.sh

If you are running a supported OS, please email support@packagecloud.io and report this.
```

rpm/build_rpms.bsh
```diff
@@ -14,7 +14,7 @@ else #Basically Centos 5/6
 fi
 
 case "${OS_NAME}" in
-  centos*|red*)
+  centos*|red*|almalinux)
```
bk2204, line 17: I don't think this is required.  We don't actually call this for any builds except for `centos`.  I don't think it hurts anything to leave it, though.

script/packagecloud.rb
```diff
@@ -43,6 +43,7 @@
     "sles/15.2",  # Current
   ],
   "centos/8" => [
+    "almalinux/8",
```
bk2204, line 46: This looks fine.  This just means that when we build the CentOS 8 package, we'll upload the exact same artifact for AlmaLinux 8, which is probably what we wanted here.

bk2204 (approved):
Hey, thanks for the patch, and welcome to Git LFS!  The patch looks good.

Assuming CI passes (which I expect it will), I'll merge this, and the next release we do, which should be 3.0.0, should contain this change and suitable packages.  That will likely happen mid-July, at the current rate.

saper:
Thank you for such a nice welcome! I am not sure how a whole packagecloud.io machinery works and how to test it; I stopped short of sending email to their support about the message I got - when I realized OS packaging is done here in this repo instead.

bk2204:
Honestly, `script/packagecloud.rb` is hard to test because it uploads data directly to the production account, so we usually just try to be very careful not to break it.  If you're interested in running the build of one or more containers, the CI job will show you how to do that:

https://github.com/git-lfs/git-lfs/blob/e0b315cb37462b2f6f5af8e273f22ab93c8fcb59/.github/workflows/ci.yml#L100-L108

And I believe the CI test failure here is not your fault, but a flaky test I haven't had time to look into.  I've kicked off a fresh run to verify.

bk2204:
Just to make a note about this, I'm having to revert this PR before the release because AlmaLinux is not supported by packagecloud.io and therefore we won't be able to upload an RPM for this as part of the release.  If we attempt to do so nevertheless, the upload will fail, as will the release.

We can revisit this if packagecloud.io supports it in the future.

saper:
What does it mean? I was going to file the issue with them but I realized it is only that script that does not recognize alma linux as the "el" flavour. It should be the same package as for the Red Hat Enterprise Linux.

bk2204:
It means that we cannot upload a package for AlmaLinux, unlike for CentOS.  If packagecloud.io decides to support AlmaLinux in the future, then we can add a package.  It's also possible that the script could be updated to use the RHEL packages as well.

saper:
Fair enough, I am not going to follow up on this. Thank you for your patience!

PR #4536, opened 2021-06-25
Allow users to explicitly limit scan depth

kellyma2:
In some scenarios git-lfs is unable to correctly deduce how to bound
the range of commits to look and when determining what to push. For
example, with Gerrit you push to refs/for/master in lieu of pushing to
refs/heads/master.  When this happens we will attempt to push
substantially more LFS objects than necessary resulting in poor
performance.

In lieu of using this fallback, if the user has a reasonable idea of
the underlying push depth we should allow them to limit the search
range.  Because we're unable to directly pass parameters via the
underlying git hooks we need to specify this in the command
environment.

### DIFF
--- a/git/rev_list_scanner.go
+++ b/git/rev_list_scanner.go
@@ -6,7 +6,9 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 
@@ -247,6 +249,10 @@ func revListArgs(include, exclude []string, opt *ScanRefsOptions) (io.Reader, []
 	case ScanAllMode:
 		args = append(args, "--all")
 	case ScanRangeToRemoteMode:
+		depth, _ := strconv.Atoi(os.Getenv("GIT_LFS_MAX_SCAN_DEPTH"))
+		if depth > 0 {
+			args = append(args, fmt.Sprintf("-%d", depth))
+		}
 		args = append(args, "--ignore-missing")
 		if len(opt.SkippedRefs) == 0 {
 			args = append(args, "--not", "--remotes="+opt.Remote)


chrisd8088:
Hey, thanks for the suggestion and the patch, and welcome to the project!

If we adopted this as-is, one thing we'd definitely want to also see in the PR is some additional documentation in the `docs/man/git-lfs-config.5.ronn` page, at a minimum; an example in another man page would be good too, if it was appropriate (e.g., maybe the `git-lfs-push` page?)

But before we get that far, would you be able to give more of an example of how this would be used?  (I'm not familiar with Gerrit, myself.)  If you run with `GIT_TRACE=1`, what are the Git commands we're executing now in your situation, and how does the change from `refs/heads/master` to something else cause problems we can only solve with an environment variable?

kellyma2:
Gerrit creates reviews by having you push to special refs.  For example, in order to push a review to master (ie: refs/heads/master) you push your commits to a ref named refs/for/master.  The line handed off to the git-lfs pre-push hook is of the form:
```
refs/heads/master b19ae1135aed6d1d7f5b34db556a79566a57f783 refs/for/master 0000000000000000000000000000000000000
```
Internally, what is happening is that we call `git rev-list --objects --ignore-missing --stdin --` and passing only `b19ae1135aed6d1d7f5b34db556a79566a57f783` to it, which returns the entire world and induces us to try and push everything we find (in our case it causes us to try and push >1000 LFS objects).

I had used an environment variable as a control for this because there doesn't appear to be any other knobs through which I can produce varied behaviour as the only inputs we get are what git passes to us via the pre-push hook.

chrisd8088:
Thanks for the additional context, that's very helpful.  I'll need a bit of time to absorb the details, I admit, and we should consider whether this env var would be useful with other Git LFS commands, or would have adverse or unintentional effects on them. 

The `RevListScanner` [class](https://github.com/git-lfs/git-lfs/blob/502801373173e28268b4f8417707117d4ee5fecd/git/rev_list_scanner.go#L158-L165) is obviously used in a number of places, like the `git lfs migrate` [command](https://github.com/git-lfs/git-lfs/blob/502801373173e28268b4f8417707117d4ee5fecd/git/githistory/rewriter.go#L514-L515) and I believe `git lfs fetch` [as well](https://github.com/git-lfs/git-lfs/blob/502801373173e28268b4f8417707117d4ee5fecd/commands/command_fetch.go#L183) as `git lfs push` and `git lfs pre-push`, and maybe others.

bk2204:
I should point out that we have an optimization that should prevent scanning objects that we know a specific named remote has, and that if you're using a named remote, that should kick in.  If it's not, fixing that such that it does kick in here would be preferable because it doesn't let the user break things accidentally.  Is there some reason that optimization wouldn't work for you?

kellyma2:
@bk2204 I'm familiar with this optimization and it has worked for me in other contexts, however, I believe it would require that you're pushing a branch.  My example didn't illustrate this properly, but when pushing HEAD or tags, this presumably won't kick in?

kellyma2:
@chrisd8088 Sure.  I can poke around a bit more and try to grasp the surrounding implications a bit as well.  I'm still familiarizing myself with the code base so some of these things are less evident to me.

bk2204:
> @bk2204 I'm familiar with this optimization and it has worked for me in other contexts, however, I believe it would require that you're pushing a branch. My example didn't illustrate this properly, but when pushing HEAD or tags, this presumably won't kick in?

I haven't looked, but if that's the case, I'd much rather we fix it so that it does work in this case because (a) that's a much better user experience because it works automatically and (b) assuming the user sets your environment variable to something reasonably conservative, fixing the optimization will almost certainly result in us traversing far less data and therefore better performance.

If we can't fix it, then I'm open to your solution, but I'd prefer to make things as automatic as possible if we can.

kellyma2:
@bk2204 When I say that it presumably won't kick in, I mean that HEAD and tags don't have a remote associated with them, presumably.   

I'm happy to work something up, but in this situation what would the preferred approach be?  eg: should we pick a remote and use that as a baseline? If so, how to decide which remote?

bk2204:
The `pre-push` hook, which invokes `git lfs pre-push`, is always provided with the name and location of the destination remote as arguments.  If a named remote isn't being used, they'll be the same, but otherwise you should have the remote name as the first argument.  In fact, if a URL is used, we try to compare it to a remote and use that one if possible because EGit always passes a URL.

So theoretically, even if you're pushing HEAD or tags, you're pushing it _somewhere_ and that remote should exist.

kellyma2:
Yes, we know the name and location of the remote *repository* but that doesn't tell us the remote *ref*. The refs/for/* remote refs in gerrit are not visible to something like `git ls-remote`.  How should we choose which remote ref to use?

bk2204:
The remote optimization doesn't use just the single remote ref.  It's designed to add `--not --remotes=origin` (where the remote is called `origin`) to exclude data reachable from all items under `refs/remotes/origin/`.  Every known branch on the remote origin is supposed to be excluded, so if you're working mostly on feature branches off a main branch and you fetch occasionally, it should exclude everything reachable from any of the remote-tracking branches.

That means that you should still be able to exclude most of the objects unless the history of your branch is very long or divergent.

kellyma2:
I'll take another look and see if there's something useful in there to do.  Thanks for the pointers

PR #4552, opened 2021-07-19
LFS repository format version

### LINKED ISSUE #4533: Implement an LFS repository format version
Currently, we assume that all LFS repositories look the same.  However, there have been proposals to store Git LFS data differently, such as by compressing objects on disk to save space.  Such a change would necessarily impede interoperability with older Git LFS versions.

The way that Git handles this is to set `core.repositoryFormatVersion` to a version number, and if it is not set, default to 0.  If the version number is not known, then the repository is rejected.  Presently, Git understands version 0, which is the original, and version 1, which is the same, but for a configuration options starting with `extension.`.  If the repository is a v1 repository and any unknown extension is found, the repository is rejected as being in an unknown format.

I propose we adopt a config option, `lfs.repositoryFormatVersion`, that by default is 0 and that we write into the repository from v3.0.0 on.  If we see a version number we don't support, we reject it.  We can then adopt a v1 in a minor version in the future with appropriate rejection for extensions (e.g., `lfs.repositoryExtensions.*`).

bk2204:
We've had several proposals to change how data is stored on disk, such as by compressing files when stored inside the .git directory. Unfortunately, those proposals are incompatible with how data is stored now and we would likely break older clients if we implemented them.

To make it possible for us to implement them in the future, let's do what Git does and add support for a repository format version.  We read the configuration option `lfs.repositoryFormatVersion` from the local configuration, and if it is nonzero, we exit with an error.  We also assume that if the value is not set, then it is implicitly 0 and set it if possible.  We ignore any errors when writing to allow working with read-only repositories in the limited circumstances where that can be valuable.

The plan in the future is to support repository format v1, which, like Git's, will be the same but for a set of `lfs.repositoryextension.*` keys which denote extensions.  Such extensions will be ignored in v0, but will be mandatory to understand in v1; unknown keys will cause Git LFS to abort.  The implementation of this functionality is left to the future, however.

In addition, make sure we set up the repositories where appropriate so that all major subsystems set the proper value.

Fixes #4533

### DIFF
--- a/t/t-repo-format.sh
+++ b/t/t-repo-format.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+
+. "$(dirname "$0")/testlib.sh"
+
+begin_test "repository format version"
+(
+  set -e
+
+  reponame="lfs-repo-version"
+  git init $reponame
+  cd $reponame
+
+  [ -z "$(git config --local lfs.repositoryFormatVersion)" ]
+
+  git lfs track '*.dat'
+
+  [ "$(git config --local lfs.repositoryFormatVersion)" = "0" ]
+
+  git config --local lfs.repositoryFormatVersion 1
+
+  git lfs track '*.bin' >output 2>&1 && exit 1
+  cat output
+  grep "Unknown repository format version: 1" output
+
+  git config --local --unset lfs.repositoryFormatVersion
+  # Verify that global settings are ignored.
+  git config --global lfs.repositoryFormatVersion 1
+
+  git lfs track '*.bin'
+)
+end_test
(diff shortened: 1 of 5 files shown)


t/t-repo-format.sh
```diff
@@ -0,0 +1,31 @@
...
+  [ -z "$(git config --local lfs.repositoryFormatVersion)" ]
+
+  git lfs track '*.dat'
+
+  [ "$(git config --local lfs.repositoryFormatVersion)" = "0" ]
+
+  git config --local lfs.repositoryFormatVersion 1
+
+  git lfs track '*.bin' >output 2>&1 && exit 1
+  cat output
```
chrisd8088, line 22: I'm guessing the `cat` is here in case of test failure, so it's a bit easier to debug what went wrong -- similar to the one in `t/t-locks.sh` discussed in https://github.com/git-lfs/git-lfs/pull/4446#discussion_r668771782.

chrisd8088 (approved):
LGTM!

t/t-repo-format.sh
bk2204: Yes, that's exactly the case.

PR #4554, opened 2021-07-21
Drop unused ClearTempStorage() transfer adapter method and tune stale comments

chrisd8088:
This PR follows up on some notes from a [conversation](https://github.com/git-lfs/git-lfs/pull/4446#discussion_r665860126) in PR #4446, specifically the fact that the `ClearTempStorage()` method of the transfer `Adapter` interface is never utilized, and the current implementations in the basic and SSH transfer adapters would be dangerous if they ever were called.  We therefore remove the method from the interface and drop its implementations from the various transfer adapaters.

(The reason the current implementations are potentially dangerous is that they perform `os.RemoveAll(a.tempDir())` where `tempDir()` is an internal method of each adapter that tries to create a temporary directory within the local Git repository, but if it fails, just returns `os.TempDir()`, and also does not cache the path that it initially returns.  So if `tempDir()` managed to create the local temporary directory, but then it was removed by some external process and file permissions changed such that a final call to `tempDir()` by `ClearTempStorage()` failed to re-create the directory, `ClearTempStorage()` would just try to remove whatever had been returned by `os.TempDir()`, which [according](https://pkg.go.dev/os#TempDir) to the docs may be the system temporary directory, e.g., `/tmp` on Unix.  Fortunately `ClearTempStorage()` has never actually been called, however, since it's introduction in PR #1265.)

We also tidy a few items discovered during the recent reviews of PRs #4446 and #4525, namely:

- We update the comment describing the `scanRefsToChan()` internal function in `lfs/gitscanner_refs.go`, which as noted in this [conversation](https://github.com/git-lfs/git-lfs/pull/4525#discussion_r669249563) has been out of date since PR #1743.
- We update the comment describing the `Begin()` method of the transfer adapter interface in `tq/transfer.go`, which as noted in this [conversation](https://github.com/git-lfs/git-lfs/pull/4446#discussion_r665871222) has been out of date since PR #1774.
- We drop an extra comment about temporary directory creation in `tq/ssh.go` added in PR #4446.
- In `t/t-batch-transfer.sh` we double-quote one instance of the `$sshurl` shell variable (added in PR #4446) and drop several unused lines, as discussed in this [pair of](https://github.com/git-lfs/git-lfs/pull/4446#discussion_r664861460) [conversations](https://github.com/git-lfs/git-lfs/pull/4446#discussion_r664861034).

### DIFF
--- a/lfs/gitscanner_refs.go
+++ b/lfs/gitscanner_refs.go
@@ -33,8 +33,8 @@ func (s *lockableNameSet) Check(blobSha string) (string, bool) {
 func noopFoundLockable(name string) {}
 
 // scanRefsToChan scans through all commits reachable by refs contained in
-// "include" and not reachable by any refs included in "excluded" and returns
-// a channel of WrappedPointer objects for all Git LFS pointers it finds.
+// "include" and not reachable by any refs included in "exclude" and invokes
+// the provided callback for each pointer file, valid or invalid, that it finds.
 // Reports unique oids once only, not multiple times if >1 file uses the same content
 func scanRefsToChan(scanner *GitScanner, pointerCb GitScannerFoundPointer, include, exclude []string, gitEnv, osEnv config.Environment, opt *ScanRefsOptions) error {
 	if opt == nil {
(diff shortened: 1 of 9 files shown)


bk2204 (approved):
Looks good.

Thanks for this cleanup and for fixing up some of the issues I created. :sparkles:

PR #4555, opened 2021-07-21
go.mod: bump gitobj to v2.0.2

bk2204:
Bump gitobj to v2.0.2 to fix support for zOS, which has different stat constants than most other Unices.


chrisd8088 (approved):
Great, thanks!

PR #4575, opened 2021-08-10
Update go.mod module path with explicit v2

### LINKED ISSUE #4574: "go get" command fails to checkout master
**Describe the bug**
The [installation wiki](https://github.com/git-lfs/git-lfs/wiki/Installation) has provides instructions for obtaining the git-lfs source code.  These instructions are:

> Building requires a standard Unix environment with GNU make and bash. On Windows, > you should install GNU make (e.g., via Chocolatey) and build under Git Bash.
>
> 1. Ensure that you have a reasonably modern version of Go.
> 2. If you are using Windows, build goversioninfo (e.g., with go get github.com/josephspurrier/goversioninfo/cmd/goversioninfo) and place the binary in your PATH.
> 3. Obtain a copy of the repository, either with git clone (into the appropriate location within your $GOPATH), or with go get, i.e., go get github.com/git-lfs/git-lfs.
> 4. In your copy of the source, execute make to ensure that you are able to build a fresh copy of Git LFS. If successful, the binary will appear in bin/git-lfs.

When I use the go get line
```
go get github.com/git-lfs/git-lfs
```

It fails with the following message:
```
-bash-4.2# go get github.com/git-lfs/git-lfs
# cd /tmp/gitlfs/src/github.com/git-lfs/git-lfs; git checkout master
error: pathspec 'master' did not match any file(s) known to git.
package github.com/git-lfs/git-lfs: exit status 1
```

**System environment**
* Fedora 20
* Latest go
   ```
   -bash-4.2# go version
   go version go1.4.2 linux/amd64
   ```

chrisd8088:
When our `go.mod` file was introduced in commit 114e85c2002091eb415040923d872f8e4a4bc636 in PR #3208, the module path chosen did not include a trailing `/v2` component.  However, the Go modules specification now advises that module paths must have a "major version suffix" which matches the release version.

We therefore add a `/v2` suffix to our module path and all its instances in import paths.  We'll need to bump this again before we release v3.0.0, obviously (along with things like version [number](https://github.com/git-lfs/git-lfs/blob/48ffab044a182a24e515c4449b74b713cab1e2d5/config/version.go#L16) in `config/version.go` and the [JSON](https://github.com/git-lfs/git-lfs/blob/48ffab044a182a24e515c4449b74b713cab1e2d5/versioninfo.json#L5-L7) in `versioninfo.json`).

See also the Go module system's major version suffix [rule](https://golang.org/ref/mod#major-version-suffixes).

As noted in https://github.com/git-lfs/git-lfs/issues/4574#issuecomment-895752677, the current output from Go's proxy service only [lists](https://proxy.golang.org/github.com/git-lfs/git-lfs/@v/list) Git LFS version tags up to v2.5.2, and it marks all the v2 ones with `+incompatible`.

<details><summary>Click for current version listing (sorted):</summary>

```
v0.1.0
v0.2.0
v0.2.1
v0.2.1-p1
v0.2.2
v0.2.3
v0.3.0
v0.3.1
v0.3.2
v0.3.3
v0.3.4
v0.3.5
v0.3.6
v0.4.0
v0.4.1
v0.4.2
v0.5.0
v0.5.1
v0.5.1-tracing
v0.5.2
v0.5.3
v0.5.4
v0.5.4-homebrew
v0.6.0
v1.0.0
v1.0.1
v1.0.2
v1.1.0
v1.1.1
v1.1.1-pre-push-tracing
v1.1.2
v1.2.0
v1.2.1
v1.3.0
v1.3.1
v1.4.0
v1.4.1
v1.4.2
v1.4.3
v1.4.4
v1.5.0
v1.5.1
v1.5.2
v1.5.2-go1.8beta1
v1.5.3
v1.5.4
v1.5.5
v1.5.6
v2.0.0+incompatible
v2.0.1+incompatible
v2.0.2+incompatible
v2.0.2-rc1+incompatible
v2.1.0+incompatible
v2.1.1+incompatible
v2.2.0+incompatible
v2.2.1+incompatible
v2.2.2-example+incompatible
v2.3.0+incompatible
v2.3.1+incompatible
v2.3.2+incompatible
v2.3.3+incompatible
v2.3.4+incompatible
v2.4.0+incompatible
v2.4.1+incompatible
v2.4.2+incompatible
v2.5.0+incompatible
v2.5.1+incompatible
v2.5.2+incompatible
```
</details>

Making an explicit [request](https://sum.golang.org/lookup/github.com/git-lfs/git-lfs@v2.13.3) for the latest v2.13.3 version from the checksum service currently returns:
```
invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

This presumably also explains why the Git LFS [entry](https://pkg.go.dev/github.com/git-lfs/git-lfs) in the pkg.go.dev search index returns a stale page:

Screen Shot 2021-08-09 at 11 52 19 PM

Partially resolves #4574.
/cc @RochaStratovan as reporter.


bk2204 (approved):
I'm not intrinsically opposed to this, but as I mentioned in https://github.com/git-lfs/git-lfs/issues/4574#issuecomment-896168736, it might give folks the impression that our internals are semantically versioned when they are not: only the behavior of the external binary is semantically versioned.

So I'm approving this in case you want to go ahead anyway, even though I feel (not strongly) that a documentation fix might be appropriate instead.

chrisd8088:
Thanks -- and just to be clear, a documentation update is absolutely required too.  This was just something I stumbled on while trying understand what was going on.

From what I can tell, there's no way to use Go modules and specify that your package is, itself, not a module for re-use, but I would be very pleased to hear otherwise.  Some other major projects like [InfluxDB](https://github.com/influxdata/influxdb) and [Consul](https://github.com/hashicorp/consul) have a top-level `go.mod` even though they provide binary executables, and the former, at least, has added `/v2` for their next major release.

Assuming we continue to follow that same model, I would be inclined to think that we should stick to the Go modules' rules, but also put prominent notices in our docs that we don't provide a stable ABI, rather than trying to hint that that's the case by relying on an artifact of the way we number releases as compared to our `go.mod` module path.

However, I also don't feel strongly about this; it was just the outcome of some investigation into what was going on with proxy.golang.org.

PR #4584, opened 2021-08-17
Bump Go to 1.17

dennisameling:
Go 1.17 [has been released](https://golang.org/doc/go1.17). One of the new features is that it now supports Windows ARM64 :tada: I'd like to create a PR for that, but before I do, here's a separate PR for bumping Go to 1.17 first.

There's new [`//go:build` lines](https://golang.org/doc/go1.17#build-lines) - Go now understands `//go:build` lines and prefers them over `// +build` lines:

> gofmt (and go fmt) now synchronizes //go:build lines with // +build lines. If a file only has // +build lines, they will be moved to the appropriate location in the file, and matching //go:build lines will be added. Otherwise, // +build lines will be overwritten based on any existing //go:build lines. For more information, see https://golang.org/design/draft-gobuild.

This PR adds the lines that `make fmt` has created. For backward compatibility reasons, the new `//go:build` lines were added and kept in sync with `// +build` as mentioned above.

PR #4586, opened 2021-08-18
Add support for Windows ARM64

dennisameling:
Follow-up PR on #4584

Go 1.17 added support for the Windows ARM64 build target. This PR adds that support to Git LFS, including the required CI + installer logic.

The installer correctly recognizes ARM64 and writes to the `Program Files` folder (not `Program Files (x86)`):

image

The installed executable is native ARM64 indeed:

image

```
PS C:\Program Files\Git LFS> .\git-lfs
git-lfs/2.13.0 (GitHub; windows arm64; go 1.17)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -186,7 +186,8 @@ BUILD_TARGETS = \
 	bin/git-lfs-freebsd-amd64 \
 	bin/git-lfs-freebsd-386 \
 	bin/git-lfs-windows-amd64.exe \
-	bin/git-lfs-windows-386.exe
+	bin/git-lfs-windows-386.exe \
+	bin/git-lfs-windows-arm64.exe
 
 # mangen is a shorthand for ensuring that commands/mancontent_gen.go is kept
 # up-to-date with the contents of docs/man/*.ronn.
@@ -236,6 +237,8 @@ bin/git-lfs-windows-amd64.exe : resource.syso $(SOURCES) mangen
 	$(call BUILD,windows,amd64,-windows-amd64.exe)
 bin/git-lfs-windows-386.exe : resource.syso $(SOURCES) mangen
 	$(call BUILD,windows,386,-windows-386.exe)
+bin/git-lfs-windows-arm64.exe : resource.syso $(SOURCES) mangen
+	$(call BUILD,windows,arm64,-windows-arm64.exe)
 
 # .DEFAULT_GOAL sets the operating system-appropriate Git LFS binary as the
 # default output of 'make'.
@@ -291,6 +294,7 @@ RELEASE_TARGETS = \
 	bin/releases/git-lfs-freebsd-386-$(VERSION).tar.gz \
 	bin/releases/git-lfs-windows-amd64-$(VERSION).zip \
 	bin/releases/git-lfs-windows-386-$(VERSION).zip \
+	bin/releases/git-lfs-windows-arm64-$(VERSION).zip \
 	bin/releases/git-lfs-$(VERSION).tar.gz
 
 # RELEASE_INCLUDES are the names of additional files that are added to each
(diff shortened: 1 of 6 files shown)


git-lfs_windows_arm64.go
```diff
@@ -0,0 +1,6 @@
+//go:build windows && arm64
+// +build windows,arm64
+
+//go:generate goversioninfo -arm=true
```
dennisameling, line 4: The `-arm` flag needs to be defined explicitly when building for Windows ARM64. Support for this was added a few days ago: https://github.com/josephspurrier/goversioninfo/pull/51

bk2204: I don't have a better suggestion.  We practically never modify those files, so the burden of adding one more should be negligible.

Makefile
```diff
@@ -367,10 +371,13 @@ bin/releases/git-lfs-windows-assets-$(VERSION).tar.gz :
 	@# work properly.
 	$(MAKE) -B GOARCH=amd64 && cp ./bin/git-lfs.exe ./git-lfs-x64.exe
 	$(MAKE) -B GOARCH=386 && cp ./bin/git-lfs.exe ./git-lfs-x86.exe
+	$(MAKE) -B GOARCH=arm64 && cp ./bin/git-lfs.exe ./git-lfs-arm64.exe
 	@echo Signing git-lfs-x64.exe
 	@signtool.exe sign -debug -fd sha256 -tr http://timestamp.digicert.com -td sha256 $(CERT_ARGS) -v git-lfs-x64.exe
 	@echo Signing git-lfs-x86.exe
 	@signtool.exe sign -debug -fd sha256 -tr http://timestamp.digicert.com -td sha256 $(CERT_ARGS) -v git-lfs-x86.exe
+	@echo Signing git-lfs-arm64.exe
+	@signtool.exe sign -debug -fd sha256 -tr http://timestamp.digicert.com -td sha256 $(CERT_ARGS) -v git-lfs-arm64.exe
```
bk2204, line 380: I assume this can be done successfully on an amd64 machine?

script/windows-installer/inno-setup-git-lfs-installer.iss
```diff
@@ -99,8 +105,10 @@ begin
       if not (Pos('Git\cmd', ExtractFilePath(ExecStdOut)) = 0) then begin
         // Proxy Git path detected
         Result := ExpandConstant('{pf}');
-      if Is64BitInstallMode then
+      if IsX64 then
         Result := Result + '\Git\mingw64\bin'
+      else if IsARM64 then
+        Result := Result + '\Git\arm64\bin'
```
bk2204, line 111: I believe this is supposed to be the path that Git for Windows uses for its binaries.  Just to confirm, Git for WIndows on arm64 uses `arm64\bin`, not `mingw64\bin`?

bk2204 (commented):
Hey,

Thanks for the patch.  I'm especially delighted you put fixes into the installer script, which is an area I'm not very familiar with.  I had a few questions here, but overall I feel good about this change and I think once those are answered we should be good to go.

script/windows-installer/inno-setup-git-lfs-installer.iss
dennisameling: That's correct. [We're working on ARM64 support for Git for Windows](https://github.com/git-for-windows/git/issues/3107) and [it was decided to use `arm64` as the folder name](https://github.com/git-for-windows/MINGW-packages/pull/44/files#diff-df00ed51839a39d696b0022219f61624cc1003df1244dc646f88ae7499f941a0R693) indeed 😊

Makefile
dennisameling: That's correct. The amd64 `signtool` binary is used and it can indeed sign arm64 binaries 👍🏼 I also built my test build/installer on an amd64 machine so can confirm it works

script/windows-installer/inno-setup-git-lfs-installer.iss
dennisameling: But to be honest, I'm not sure whether this logic is ideal in the first place. E.g. a user can install x86 Git for Windows on a x64 Windows machine, so this logic would fail in that case. When that happens, a user will only have `\Git\mingw32\bin` on their x64 machine then. On top of this function a comment says `// Currently only used to support running the uninstaller for the old Git LFS version.` - [the commit was added 6 years ago](https://github.com/git-lfs/git-lfs/commit/1651e55ce5cc37c69bf5cb3bf92e626b83ad5e0a). I'm not familiar with Git LFS' codebase, but my gut says this code might no longer be needed in the first place. I'll leave that up to you to decide!

bk2204: I also agree that this might no longer be used, but I'm happy to leave it as it is for now.  The InnoSetup stuff is unfamiliar to me as a non-Windows developer, so I'm going to defer to you, since you seem to be more familiar with it.

bk2204 (approved):

PR #4591, opened 2021-08-23
Switch from which to command -v

bk2204:
Debian has recently deprecated the which utility and will remove it in a future version, since POSIX specifies `command -v` for finding the location of a path and has for some time.  In addition, which has started complaining loudly to standard error, so it causes test and makefile output to be hard to read.

Let's switch to the preferred invocation, `command -v`, in the Makefile, build tooling, and test suite.


chrisd8088 (approved):
LGTM, thanks.

PR #4603, opened 2021-08-31
subprocess: don't treat empty PATH component as . on Windows

bk2204:
It's common for Windows machines to have trailing empty PATH components and users there don't expect that to mean "." like it does on Unix.  To avoid unexpected behavior and somewhat improve security, treat the empty PATH component differently on Windows.

### DIFF
--- a/subprocess/path.go
+++ b/subprocess/path.go
@@ -32,6 +32,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


subprocess/path.go
```diff
@@ -52,6 +53,11 @@ func LookPath(file string) (string, error) {
 	path := os.Getenv("PATH")
 	for _, dir := range filepath.SplitList(path) {
 		if dir == "" {
+			// Windows often has empty components in the PATH and
+			// treating them as "." not expected.
```
chrisd8088, line 57: Very minor nit: maybe revise to `treating them as "." is not expected` (i.e., add an "is")?

chrisd8088 (approved):
Thank you!

subprocess/path.go
bk2204: Fixed.

PR #4604, opened 2021-08-31
Support for locking and unlocking multiple files

bk2204:
We've seen a request to lock and unlock multiples files on the command line.  This has been difficult due to the need to change the JSON output format, which was a backwards incompatible change.  Since we're introducing a new major version, we can now make that change, and it's done here.


chrisd8088 (approved):
Looks really good to me, thank you!

Pakon2543 (commented):
gh pr checkout 4604

Pakon2543 (commented):
gh pr checkout  #4604[
Duplicate of #](url)

PR #4611, opened 2021-09-01
Bump package version to v3

### LINKED ISSUE #4580: Update module path to /v3 in go.mod file
Before we release v3.0, we should remember to update the module path in the `go.mod` file, akin to the change in #4575.

bk2204:
Since we're about to do a v3.0.0 release, let's bump the version to v3.

Make this change automatically with the following command to avoid any missed items:

```
  git grep -l github.com/git-lfs/git-lfs/v2 | \
  xargs sed -i -e 's!github.com/git-lfs/git-lfs/v2!github.com/git-lfs/git-lfs/v3!g'
```

Update one place that's currently missing the version.

Fixes #4580


chrisd8088 (approved):
LGTM.  I suppose it might be worthwhile changing the other version strings and settings at the same time, like the ones in `config/version.go`?

bk2204:
I'm going to the command that does that as part of the release PR, as usual, so it will be in a separate PR.

PR #4622, opened 2021-09-09
corrected debian 11 & 12  derived variants

jvimr:
corrected debian 11 & 12  derived variants - bullseye (debian 11) and bookworm (debian 12) - for the packagecloud deployment

debian versions synced with the list on the [packagecloud](https://packagecloud.io/docs#anchor-debian)

### DIFF
--- a/script/packagecloud.rb
+++ b/script/packagecloud.rb
@@ -71,6 +71,7 @@ $distro_name_map = {
   ],
   "debian/11" => [
     "debian/bullseye",  # Current stable
+    "debian/bookworm", # Current testing
   ]
 }
 
--- a/script/packagecloud.rb
+++ b/script/packagecloud.rb
@@ -70,10 +70,7 @@ $distro_name_map = {
     "ubuntu/hirsute",   # EOL January 2022
   ],
   "debian/11" => [
-    "debian/bullseye",  # Current
-  ],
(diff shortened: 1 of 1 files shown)


bk2204 (commented):
Hey,

Thanks for the patch.

In general, we don't build packages based on their own containers specifically for pre-release versions, like Debian testing.  That's why you won't have seen a package for Debian 11 before now, and we'll have shipped the Debian 10 packages there in the past.  That's because Debian testing can change very quickly and we don't really want to follow a moving target for things that are in our CI system.

The change you've introduced would require a change to build a Debian 12 container specifically in order for this to work, and as such, we probably don't want to do that for the reasons I've mentioned.

I do appreciate you sending the patch nevertheless, though.

jvimr:
Hi,
OK, I've changed the pull req just to correct the debian 11 codename. Is this OK?

bk2204:
I think we'd want to have both `debian/bullseye` and `debian/bookworm` in the `debian/11` section in order to have things work as expected.

jvimr:
OK, both `debian/bullseye` and `debian/bookworm` are now  in the debian/11 section

bk2204 (approved):
Thanks for the patch.  This looks good.

PR #4624, opened 2021-09-09
Advertise hash algorithm supported in batch request

bk2204:
To future-proof ourselves against the need to change hash algorithms in the future, let's advertise the hash algorithm to the remote side.  If the value is not `sha256` or absent (which is treated as the same thing), we fail.

### DIFF
--- a/tq/api.go
+++ b/tq/api.go
@@ -1,6 +1,7 @@
 package tq
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/git-lfs/git-lfs/v3/errors"
@@ -24,11 +25,13 @@ type batchRequest struct {
 	Objects              []*Transfer `json:"objects"`
 	TransferAdapterNames []string    `json:"transfers,omitempty"`
 	Ref                  *batchRef   `json:"ref"`
+	HashAlgorithm        string      `json:"hash_algo"`
 }
 
 type BatchResponse struct {
 	Objects             []*Transfer `json:"objects"`
 	TransferAdapterName string      `json:"transfer"`
+	HashAlgorithm       string      `json:"hash_algo"`
 	endpoint            lfshttp.Endpoint
 }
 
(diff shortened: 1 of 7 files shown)


tq/api.go
```diff
@@ -38,6 +41,7 @@ func Batch(m *Manifest, dir Direction, remote string, remoteRef *git.Ref, object
 	}
 
 	return m.batchClient().Batch(remote, &batchRequest{
+		HashAlgorithm:        "sha256",
```
chrisd8088, line 44: In the `type batchRequest struct` definition just above, `HashAlgorithm` is the last element, while here it's given as the first.  It makes no practical difference but as the others are in the same order, perhaps this should also go last?

tq/ssh.go
```diff
@@ -73,6 +73,19 @@ func (a *SSHBatchClient) Batch(remote string, bReq *batchRequest) (*BatchRespons
...
+			continue
+		}
+		if entries[0] == "hash-algo" {
+			bRes.HashAlgorithm = entries[1]
+			if bRes.HashAlgorithm != "sha256" {
+				return nil, fmt.Errorf("batch response: unsupported hash algorithm: %q", entries[1])
+			}
+		}
+	}
+
```
chrisd8088, line 88: I may be missing something here, but I'm not clear on what this achieves, as I don't see how the `args` array would change from where it is set just above with the value `"hash-algo=sha256"`.  Is this code intended to future-proof against a time when the `args` array is set dynamically?  Or am I just missing how its contents are dynamic now?

chrisd8088 (approved):
I had a question about one section, but I'm going to approve this as it overall makes good sense to me, and I may be misunderstanding how that one section is intended to operate.

tq/ssh.go
bk2204: You're not missing anything.  I intended to make `batchInternal` return an `args` value, but didn't.  I've fixed that in the latest version.

tq/api.go
bk2204: Fixed.

tq/ssh.go
chrisd8088: Okay, that makes more sense; it's parsing the arguments returned in the batch response.

PR #4641, opened 2021-09-24
Stop supporting Go older than 1.13

bk2204:
When using Go 1.12, we need to download some of the HTTP/2 code as part of the build process.  However, that code uses the Go 1.13 errors package and expects errors.Is to work, which it does not.  Since this isn't really practically avoidable and Go upstream is not going to care about getting those older Go versions to work, let's drop support for Go before 1.13.

Note that we only guarantee compatibility with the latest Go version, so this doesn't violate our support policy.

I'm not including newer versions of Go because they are extremely picky about the build process and things are currently broken on at least 1.16, and my goal here is to unbreak the build system as quickly as possible.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
     name: Build with specific Go
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


chrisd8088 (approved):
LGTM, thank you.

PR #4647, opened 2021-09-25
Fix Windows arm64 release

### LINKED ISSUE #4642: Release process fails on Windows ARM64
Our Windows release CI job is [failing on ARM64](
https://github.com/git-lfs/git-lfs/runs/3701266790) due to a bad relocation in `resource.syso`.  This is preventing us from doing a release.

I've tried removing the `resource.syso` file before rebuilding the ARM64 binaries, but that doesn't fix the problem.  If we can't figure out how to solve this before Monday, I'll need to revert the Windows ARM64 support until such time as it builds properly in CI.

Note that this problem does not occur when building the binaries on Linux.

/cc @dennisameling as the original submitter

dennisameling:
Fixes #4642

This PR does the following:
- Fix Windows arm64 build as reported in https://github.com/git-lfs/git-lfs/issues/4642. The problem was that `go generate` didn't have a GOARCH set, so it used the host architecture (amd64) to build `resource.syso`. That in turn caused the `unknown ARM64 relocation type 7` error. Also, once the `resource.syso` file was generated, GNU Make wouldn't update it for other architectures as it can't recognize if GOARCH has changed.
- Build arm64 `resource.syso` instead of arm (32-bit). While 32-bit arm does work on arm64, Git LFS only supports arm64, so we can safely go with that one.
- Fix small typo in InnoSetup configuration from my previous PR (https://github.com/git-lfs/git-lfs/pull/4586)
- Don't install `windows-sdk-10.0` in the CI release process as [it unnecessarily adds 3 mins to the job](https://github.com/dennisameling/git-lfs/runs/3709739120?check_suite_focus=true). GitHub Actions [already includes several Windows 10 SDK versions out of the box](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md#workloads-components-and-extensions).

Example release pipeline which succeeded on Windows: https://github.com/dennisameling/git-lfs/runs/3709768442?check_suite_focus=true - ~I don't have my macOS code signing cert at hand currently, so that job obviously failed 😊~

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,11 +34,11 @@ jobs:
       shell: bash
       env:
         CERT_CONTENTS: ${{secrets.WINDOWS_CERT_BASE64}}
-    - run: PATH="$HOME/go/bin:$PATH" make bin/releases/git-lfs-windows-amd64-$(git describe).zip
+    - run: PATH="$HOME/go/bin:$PATH" GOARCH=amd64 go generate && make bin/releases/git-lfs-windows-amd64-$(git describe).zip
       shell: bash
-    - run: PATH="$HOME/go/bin:$PATH" make bin/releases/git-lfs-windows-386-$(git describe).zip
+    - run: PATH="$HOME/go/bin:$PATH" GOARCH=386 go generate && make bin/releases/git-lfs-windows-386-$(git describe).zip
       shell: bash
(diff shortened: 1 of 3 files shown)


bk2204 (approved):

bk2204:
Thanks for the quick patch.  This looks good.

PR #4654, opened 2021-09-27
fs: specify a file as existing if it's empty

### LINKED ISSUE #4645: git-lfs 3.0.0 - zero byte files not tracked by git-lfs are resulting in Object does not exist on the server
**Describe the bug**
In two separate repositories today, after upgrading to git lfs 3.0.0 (these happened to be on circle-ci pipelines with unpinned versions of git lfs), the following error occurred:

```
[e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855] Object does not exist on the server: [404] Object does not exist on the server
error: failed to fetch some objects from 'https://github.com/<repo name redacted>/info/lfs'
```

Oddly enough - the exact same object ID was reported in two completely different repos.

Running
```
git log --all -p -S e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
```
Did not return any objects.

However, running
```
git-lfs ls-files --all | grep e3b0c44
```
returned a series of files, none of which are being tracked by git-lfs, all of which are zero bytes in size.

Pinning the version of git-lfs used to 2.13.3 resolved the issue.

**To Reproduce**
1. Create a repository with an empty file in it and some other non-empty file tracked with git-lfs
2. Pull

**Expected behavior**
No error, as these files are not tracked by git lfs.

**System environment**
Ubuntu 16.04 (ubuntu-1604:202007-01) running on circle-ci
Git config:
```
filter.lfs.required=true
url.ssh://git@github.com.insteadof=https://github.com
gc.auto=0
filter.lfs.required=true
core.filemode=true
core.bare=false
core.logallrefupdates=true
remote.origin.url=git@github.com:carta/ds-reporting-service.git
branch.master.remote=origin
lfs.repositoryformatversion=0
```

**Output of `git lfs env`**
```
git-lfs/3.0.0 (GitHub; linux amd64; go 1.17)
git version 2.27.0

Endpoint=https://github.com/<redacted repo name>.git/info/lfs (auth=none)
  SSH=git@github.com:<redacted>/<redacted repo name>.git
LocalWorkingDir=/home/circleci/<redacted repo name>
LocalGitDir=/home/circleci/<redacted repo name>/.git
LocalGitStorageDir=/home/circleci/<redacted repo name>/.git
LocalMediaDir=/home/circleci/<redacted repo name>/.git/lfs/objects
LocalReferenceDirs=
TempDir=/home/circleci/<redacted repo name>/.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/home/circleci/<redacted repo name>/.git/lfs
AccessDownload=none
AccessUpload=none
GIT_ASKPASS=echo
```

bk2204:
If the object has a size of zero, then we already have its object: it's the empty file, and we shouldn't need to download it from anywhere.  If we inquire if it exists, then say it does, and say its location is the system's equivalent of `/dev/null`.

The only tricky case is if we're de-duplicating, linking, or writing to a file, in which case we should not try to use a link or replace on the file, since we neither want to link nor replace the system's `/dev/null`.

Fixes #4645

### DIFF
--- a/commands/command_dedup.go
+++ b/commands/command_dedup.go
@@ -112,6 +112,10 @@ func dedup(p *lfs.WrappedPointer) (success bool, err error) {
 
 	// Do clone
 	srcFile := cfg.Filesystem().ObjectPathname(p.Oid)
+	if srcFile == os.DevNull {
+		return true, nil
+	}
+
 	dstFile := filepath.Join(cfg.LocalWorkingDir(), p.Name)
 
 	// Clone the file. This overwrites the destination if it exists.
--- a/commands/command_fsck.go
+++ b/commands/command_fsck.go
@@ -107,7 +107,11 @@ func fsckCommand(cmd *cobra.Command, args []string) {
 
 	for _, oid := range corruptOids {
 		badFile := filepath.Join(badDir, oid)
-		if err := os.Rename(cfg.Filesystem().ObjectPathname(oid), badFile); err != nil {
+		srcFile := cfg.Filesystem().ObjectPathname(oid)
+		if srcFile == os.DevNull {
+			continue
+		}
+		if err := os.Rename(srcFile, badFile); err != nil {
 			ExitWithError(err)
 		}
 	}
--- a/commands/command_prune.go
+++ b/commands/command_prune.go
@@ -307,6 +307,9 @@ func pruneDeleteFiles(prunableObjects []string, logger *tasklog.Logger) {
 			problems.WriteString(fmt.Sprintf("Unable to find media path for %v: %v\n", oid, err))
 			continue
 		}
+		if mediaFile == os.DevNull {
+			continue
+		}
 		err = os.Remove(mediaFile)
 		if err != nil {
 			problems.WriteString(fmt.Sprintf("Failed to remove file %v: %v\n", mediaFile, err))
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -3,6 +3,8 @@ package fs
 import (
 	"bufio"
 	"bytes"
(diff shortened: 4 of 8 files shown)


commands/command_dedup.go
```diff
@@ -114,6 +114,10 @@ func dedup(p *lfs.WrappedPointer) (success bool, err error) {
 	srcFile := cfg.Filesystem().ObjectPathname(p.Oid)
 	dstFile := filepath.Join(cfg.LocalWorkingDir(), p.Name)
 
+	if srcFile == os.DevNull {
+		return true, nil
+	}
+
```
chrisd8088, line 120: I suppose this could move up to just below where `srcFile` is initialized, since we don't need `dstFile` yet.  But it's a minor thing.

chrisd8088 (approved):
Thanks for fixing this!  The new tests definitely fail without these changes, which is excellent.  Would there be any value in adding additional tests for the `prune` and `dedup` use cases as well?

commands/command_dedup.go
bk2204: I've moved it.

bk2204:
I've added tests for the prune case; unfortunately, dedup doesn't run on my system, so I can't test it here.

Issue #4628, opened 2021-09-13
`git-lfs migrate import` does not add attributes for copied files (same name and content)


ycongal-smile:
**Describe the bug**
When trying to `git-lfs migrate import` a repo in which there are copied files (eg: `dirA/file` = `dirB/file`), the `.gitattributes` file does not list all instances of the file but they are all changed to a pointer blob. This results in a incoherent repo.

**To Reproduce**
1. Create a test repo with this script : 
```
git init
echo test > a
mkdir dir
cp a dir/
git add a dir/a
git commit -m "add a dir/a"
# In this repo a & dir/a have the same name and content
```
2. Import the repo in git-lfs : `git lfs migrate import --above=1b`
3. Look at .gitattributes and notice that `dir/a` is missing
```
$ cat .gitattributes 
/a filter=lfs diff=lfs merge=lfs -text
```
4. Look at `dir/a` and notice that it is a "LFS pointer blob" :
```
$ cat dir/a 
version https://git-lfs.github.com/spec/v1
oid sha256:f2ca1bb6c7e907d06dafe4687e579fce76b37e4e93b7605022da52e6ccc26fd2
size 5
```
5. To go further, doing a `git-lfs checkout` result in a dirty repo (`dir/a`pointer is replaced by its content but without a filter, so git notices the change)
```
$ git-lfs checkout
Checking out LFS objects: 100% (2/2), 10 B | 0 B/s, done.                                                                                                                                     
$ git status
On branch main
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	modified:   dir/a
```
6. As a workaround, I can `git-lfs track dir/a` and commit the resulting modifications.

NB: I can't reproduce this behavior if the name are different (eg `a` = `dir/b`)

**Expected behavior**
After the `git lfs migrate import ...`, `a` and `dir/a`are listed in `.gitattributes`.

**System environment**
Tested on a master (commit 16da80350b9ec7abf131e878b532d472ca149f77) build on a Debian 11 (bullseye).

**Output of `git lfs env`**
```
git-lfs/2.13.0 (GitHub; linux amd64; go 1.15.9)
git version 2.30.2

LocalWorkingDir=$BASE/git-lfs/debug_dir/tmp.1631551057.6tar
LocalGitDir=$BASE/git-lfs/debug_dir/tmp.1631551057.6tar/.git
LocalGitStorageDir=$BASE/git-lfs/debug_dir/tmp.1631551057.6tar/.git
[...]

PR #4671, opened 2021-10-07
Call migrate() BlobFn on every blob

ycongal-smile:
Caching is now done inside the BlobFn function instead of in the caller. To do this, the OID of the original blob is given as an argument to BlobFn.

Adapt migrate() calls to this :
* For "migrate info", a set of "seen" OID is cached to avoid counting them twice.
* For "migrate import", the cleaned blob is cached and reused.

Fixes git-lfs/git-lfs#4628

Also add a test case for git-lfs/git-lfs#4628 and adapt  unittest to follow the BlobFn change


ycongal-smile:
Thanks!

PR #4673, opened 2021-10-07
Correct t-lock regular expression to be musl compatible

fh1ch:
This MR corrects a regular expression within `t/t-lock.sh` to make it compatible with the musl version of libc (see https://wiki.musl-libc.org/functional-differences-from-glibc.html#regular-expressions), as it's used in e.g. Alpine Linux. There is no functional change, just a format one.

I'm currently in the process of bumping *git-lfs* to version *3.0.1* in Alpine Linux (see https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/26213), but running the integration tests (specifically this one regular expression) as part of the packaging process fail. This can also be reproduced locally:

```sh
% docker run -ti alpine /bin/sh
$ grep -E '\[{"id":"[^"]+","path":"a\.dat","owner":{"name":"Git LFS Tests"},"locked_at":"[^"]+"},{"id":"[^"]+","path":"b\.dat","owner":{"name":"Git LFS Tests"},"locked_at":"[^"]+"}\]' lock.json 
grep: bad regex '\[{"id":"[^"]+","path":"a\.dat","owner":{"name":"Git LFS Tests"},"locked_at":"[^"]+"},{"id":"[^"]+","path":"b\.dat","owner":{"name":"Git LFS Tests"},"locked_at":"[^"]+"}\]': Invalid contents of {}
```

/cc @bk2204

### DIFF
--- a/t/t-lock.sh
+++ b/t/t-lock.sh
@@ -113,7 +113,7 @@ begin_test "lock multiple files (JSON)"
   git push origin main:other
 
   git lfs lock --json *.dat | tee lock.json
-  grep -E '\[{"id":"[^"]+","path":"a\.dat","owner":{"name":"Git LFS Tests"},"locked_at":"[^"]+"},{"id":"[^"]+","path":"b\.dat","owner":{"name":"Git LFS Tests"},"locked_at":"[^"]+"}\]' lock.json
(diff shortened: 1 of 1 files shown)


bk2204:
> Regarding CI, there seems to be an unrelated failure (probably just a flaky test) as well as a problem with installing Git (due to the Let's Encrypt intermediate CA expiry, see https://github.com/git-lfs/git-lfs/pull/4673/checks?check_run_id=3828861031). I guess we need to re-trigger those and hope that the second time is a charm, otherwise I'll dig into the issue a bit deeper.

Yes, I think that's a flaky test.  I've opened git-lfs/build-dockers#37 to fix the Let's Encrypt issue and will re-run CI once that merges.

PR #4713, opened 2021-10-29
Avoid errors in git lfs env

### LINKED ISSUE #4712: Error reading git config: error running /usr/libexec/git-core/git 'rev-parse' '--is-bare-repository': 'fatal: not a git repository (or any of the parent directories): .git' 'exit status 128'
**Describe the issue**
when I used "git lfs env", it will report error：
Error reading git config: error running /usr/libexec/git-core/git 'rev-parse' '--is-bare-repository': 'fatal: not a git repository (or any of the parent directories): .git' 'exit status 128'
LocalWorkingDir=
LocalGitDir=
LocalGitStorageDir=
LocalMediaDir=lfs/objects
LocalReferenceDirs=
TempDir=lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=lfs
AccessDownload=none
AccessUpload=none
GIT_EXEC_PATH=/usr/libexec/git-core
git config filter.lfs.process = ""
git config filter.lfs.smudge = ""
git config filter.lfs.clean = ""

**System environment**
git-lfs version
git-lfs/2.13.3 (GitHub; linux arm64; go 1.16.2; git a5e65851)

**Output of `git lfs env`**
git-lfs/2.13.3 (GitHub; linux arm64; go 1.16.2; git a5e65851)
git version 2.19.1

Error reading git config: error running /usr/libexec/git-core/git 'rev-parse' '--is-bare-repository': 'fatal: not a git repository (or any of the parent directories): .git' 'exit status 128'
LocalWorkingDir=
LocalGitDir=
LocalGitStorageDir=
LocalMediaDir=lfs/objects
LocalReferenceDirs=
TempDir=lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=lfs
AccessDownload=none
AccessUpload=none
GIT_EXEC_PATH=/usr/libexec/git-core
git config filter.lfs.process = ""
git config filter.lfs.smudge = ""
git config filter.lfs.clean = ""

**Additional context**
Any other relevant context about the problem here.

bk2204:
There are a couple different cases in which we can get errors when running `git lfs env` outside of a repository.  One of these can also cause us to fail to load Git configuration, which makes the output confusing, and another only shows up when the user (like the author) is using a non-English language.  Let's fix these issues and add a test to be sure we don't regress the behavior.

Each commit contains an explanatory commit message.  Note that this PR is probably best reviewed with whitespace differences turned off.

Fixes #4712

### DIFF
--- a/git/config.go
+++ b/git/config.go
@@ -140,31 +140,29 @@ func (c *Configuration) Sources(dir string, optionalFilename string) ([]*Configu
 	if err != nil {
 		return nil, err
 	}
+	configs := make([]*ConfigurationSource, 0, 2)
 
 	bare, err := IsBare()
-	if err != nil {
-		return nil, err
-	}
-
-	// First try to read from the working directory and then the index if
-	// the file is missing from the working directory.
-	var fileconfig *ConfigurationSource
-	if !bare {
-		fileconfig, err = c.FileSource(filepath.Join(dir, optionalFilename))
-		if err != nil {
-			if !os.IsNotExist(err) {
-				return nil, err
+	if err == nil {
+		// First try to read from the working directory and then the index if
+		// the file is missing from the working directory.
+		var fileconfig *ConfigurationSource
+		if !bare {
+			fileconfig, err = c.FileSource(filepath.Join(dir, optionalFilename))
+			if err != nil {
+				if !os.IsNotExist(err) {
+					return nil, err
+				}
+				fileconfig, _ = c.RevisionSource(fmt.Sprintf(":%s", optionalFilename))
 			}
-			fileconfig, _ = c.RevisionSource(fmt.Sprintf(":%s", optionalFilename))
 		}
-	}
-	if fileconfig == nil {
-		fileconfig, _ = c.RevisionSource(fmt.Sprintf("HEAD:%s", optionalFilename))
-	}
+		if fileconfig == nil {
+			fileconfig, _ = c.RevisionSource(fmt.Sprintf("HEAD:%s", optionalFilename))
+		}
 
-	configs := make([]*ConfigurationSource, 0, 2)
-	if fileconfig != nil {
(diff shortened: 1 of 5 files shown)


git/config.go
```diff
@@ -140,31 +140,30 @@ func (c *Configuration) Sources(dir string, optionalFilename string) ([]*Configu
...
-	// First try to read from the working directory and then the index if
-	// the file is missing from the working directory.
-	var fileconfig *ConfigurationSource
-	if !bare {
-		fileconfig, err = c.FileSource(filepath.Join(dir, optionalFilename))
-		if err != nil {
-			if !os.IsNotExist(err) {
-				return nil, err
+	if err == nil {
+
```
chrisd8088, line 147: Minor note: I think there's an extra line of whitespace here.

I suppose one could avoid indenting this block by instead doing something like:
```go
	if err != nil {
		return append(configs, gitconfig), nil
	}
```
But this approach has the advantage of not repeating the `return` line, so I think it's actually better as written.

chrisd8088 (approved):
LGTM!

git/config.go
bk2204: Whitespace removed.  And I agree this is slightly nicer to avoid duplication.

PR #4717, opened 2021-11-01
Pakon2543 patch 2

Pakon2543:


bk2204:
Hey,

I don't think this is helpful since it doesn't seem to add anything on top of our existing workflows and the description doesn't tell us why we'd want this.  As a result, I'm going to close it.

Issue #4324, opened 2020-11-30
git clone with specified user in url doesn't work as expected


howardlyliao:
**Describe the bug**
For git command we can specified user in url.  Ex: git clone "http://mtk10145@xxxxxx:8080/Git-Test"
But for git-lfs, if netrc file has multiple lines with same machine but different login.
git-lfs would always use the first line with the same machine it finds in .netrc.
If the account in the first line doesn't have the permission. 
It which would cause authentication fail.

In our environment we have different login with different permissions for the same machine(gerrit).
Thus this issue is critical for us.

**To Reproduce**
1.create a netrc file ~/.netrc with a fake_acc as the account without permission at the first line:
machine myHost login fake_acc password FakePassword
machine myHost login howard password RealPassword

2.clone a git-lfs repo:
git clone http://howard@myHost:8080/test_lfs1

3.It will ask to type password due to authentication failed:
Cloning into 'test_lfs1'...
remote: Counting objects: 3, done
remote: Finding sources: 100% (3/3)
remote: Total 178 (delta 45), reused 177 (delta 45)
Receiving objects: 100% (178/178), 16.64 KiB | 8.32 MiB/s, done.
Resolving deltas: 100% (45/45), done.
Password for 'http://howard@myHost:8080':

4.However if we marked the first line of ~/.netrc, it would clone properly:
#machine myHost login fake_acc password FakePassword
machine myHost login howard password RealPassword

$git clone http://howard@myHost:8080/test_lfs1
Cloning into 'test_lfs1'...
remote: Counting objects: 3, done
remote: Finding sources: 100% (3/3)
remote: Total 178 (delta 45), reused 177 (delta 45)
Receiving objects: 100% (178/178), 16.64 KiB | 16.64 MiB/s, done.
Resolving deltas: 100% (45/45), done.

**Output of `git lfs env`**
git-lfs/2.11.0 (GitHub; linux amd64; go 1.14.1)
git version 2.18.0

LocalWorkingDir=
LocalGitDir=
LocalGitStorageDir=
LocalMediaDir=lfs/objects
LocalReferenceDirs=
TempDir=lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
FetchRecentRefsIncludeRemotes=true
PruneOffsetDays=3
[...]

PR #4726, opened 2021-11-04
netrc: consider same machine may have different login names

howardlyliao:
This is related change for solving #4324


bk2204 (approved):
Hey,

Thanks for the patch.  I agree this looks like the right way forward.

PR #4728, opened 2021-11-08
Add support for ARM64 Debian packages

### LINKED ISSUE #4546: Publish arm64 linux packages to packagecloud
**Describe the issue**
I can't install recent git-lfs via package on arm64

(I assume same for s390x and ppc would be good to get those sorted too)

**System environment**
The version of your operating system, plus any relevant information about platform or configuration (e.g., container or CI usage, Cygwin, WSL, or non-Basic authentication).  If relevant, include the output of `git config -l` as a code block.

macOs big sur 11.4
M1 mac
arm64

**Output of `git lfs env`**
The output of running `git lfs env` as a code block.
not relevant

**Additional context**
I've tested on the debian:buster-slim and centos:8.3.2011 images and cannot find the corresponding package.

```
docker run -it debian:buster-slim bash
apt install -y curl
curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash
apt-cache showpkg git-lfs
```

If I run on an amd64 machine all works fine

Work around for docker:

```dockerfile
ARG TARGETARCH
ARG GIT_LFS_VERSION=2.13.3

# required for multi-arch support, revert to package cloud after:
# https://github.com/git-lfs/git-lfs/issues/4546
RUN GIT_LFS_ARCHIVE="git-lfs-linux-${TARGETARCH}-v${GIT_LFS_VERSION}.tar.gz" \
    GIT_LFS_RELEASE_URL="https://github.com/git-lfs/git-lfs/releases/download/v${GIT_LFS_VERSION}/${GIT_LFS_ARCHIVE}"\
    set -x; curl --quiet --fail --location --show-error --output "/tmp/${GIT_LFS_ARCHIVE}" "${GIT_LFS_RELEASE_URL}" && \
    mkdir -p /tmp/git-lfs && \
    tar xzvf "/tmp/${GIT_LFS_ARCHIVE}" -C /tmp/git-lfs && \
    bash -x /tmp/git-lfs/install.sh && \
    rm -rf /tmp/git-lfs*
```

bk2204:
Users have recently requested packages for arm64, which is an increasingly popular architecture.  In addition, it is the architecture used on newer Macs and some Windows systems, which may commonly run Linux VMs or WSL.

However, GitHub Actions does not support arm64 natively, so we must build in emulation.  This is extraordinarily slow, so we'll build only for the latest Debian release.  All Debian-based images use our Debian images, and Debian is by far the most portable mainstream Linux distribution in terms of architecture support, so the cost to add additional architectures should be extremely minimal.  Users who would like versions for older releases may build their own packages with our tooling.

Build these packages in both CI and for releases.

Fixes #4546

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -122,3 +122,30 @@ jobs:
     - run: '[ -z "${GITHUB_REF%%refs/tags/*-pre*}" ] || ./script/packagecloud.rb'
       env:
         PACKAGECLOUD_TOKEN: ${{secrets.PACKAGECLOUD_TOKEN}}
+  build-docker-cross:
+    name: Build Cross Linux packages
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        arch: [arm64]
+        container: [debian_11]
+    - uses: actions/checkout@v1
+    - uses: actions/setup-ruby@v1
+    - run: |
+        echo '{"experimental": true}' | sudo tee /etc/docker/daemon.json
+        sudo systemctl restart docker.service
+        docker version -f '{{.Server.Experimental}}'
+    - uses: docker/setup-qemu-action@v1
+    - run: git clone https://github.com/git-lfs/build-dockers.git "$HOME/build-dockers"
+    - run: (cd "$HOME/build-dockers" && ./build_dockers.bsh --arch=$ARCH $CONTAINER)
+      env:
+        ARCH: ${{matrix.arch}}
+        CONTAINER: ${{matrix.container}}
+    - run: DOCKER_AUTOPULL=0 ./docker/run_dockers.bsh --prune --arch=$ARCH $CONTAINER
+      env:
+        ARCH: ${{matrix.arch}}
+        CONTAINER: ${{matrix.container}}
+    # If this is a pre-release tag, don't upload anything to packagecloud.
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -122,3 +122,28 @@ jobs:
...
       env:
         PACKAGECLOUD_TOKEN: ${{secrets.PACKAGECLOUD_TOKEN}}
+  build-docker-cross:
+    name: Build Cross Linux packages
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        arch: [arm64]
+        container: [debian_11]
+    - run: |
```
chrisd8088, line 132: I notice there two lines in the `build-docker` workflows from both this file and  `ci.yaml`, and also in the new `build-docker-cross` workflow in `ci.yaml`, which aren't present here:
```yaml
    - uses: actions/checkout@v1
    - uses: actions/setup-ruby@v1
```
Are these needed everywhere, or might they dropped everywhere, or is this workflow unique in some way in not needing them?

chrisd8088 (approved):
One question but otherwise LGTM!

.github/workflows/release.yml
bk2204: Oh, no we definitely need those.  Good catch.

PR #4741, opened 2021-11-16
ssh: avoid using -- where possible

### LINKED ISSUE #4733: SSH invalid argument error with TeamCity since 3.x
**Describe the issue**
I imagine you want to class this as TeamCity's fault instead of a regression in Git LFS, but it's more discoverable here, and possibly you want to restore compatibility, so I'm posting this here anyway.

When using TeamCity for CI, if it's configured to do agent-side checkout (a requirement for Git LFS support) and a key's been set in the VCS root settings, at least in the 2017 version (I don't have access to anything up-to-date, so I can't verify whether anything's been changed), it applies the key by directing Git to use a temporary batch file as its SSH command, which presumably adds the necessary arguments and forwards the command to the real SSH.

Git LFS 2.x would call SHH with commands of the form `ssh arg arg arg`, whereas 3.x calls it as `ssh -- arg arg arg` (i.e. with `--`). It appears TeamCity's shim doesn't support this, as the following error is emitted:

```
Error downloading object: path (short hash): Smudge error: Error downloading path (long hash): batch request: Invalid arguments [--, git@github.com, git-lfs-authenticate org/repo.git download]: exit status 1
```

Or, in the `GIT_TRACE` file:
```
12:08:50.384288 trace git-lfs: run_command: C:\buildAgentData\temp\git-ssh-3502009113789195714.bat -- git@github.com git-lfs-authenticate org/repo.git download
12:08:50.384860 trace git-lfs: exec: C:\buildAgentData\temp\git-ssh-3502009113789195714.bat '--' 'git@github.com' 'git-lfs-authenticate org/repo.git download'
12:08:50.515545 trace git-lfs: ssh: git@github.com failed, error: exit status 1, message: Invalid arguments [--, git@github.com, git-lfs-authenticate org/repo.git download] (try: 0/6)
```

Comparing `GIT_TRACE` output between this and the same command from the last working version shows the command given to the SSH shim batch file is identical except for the absence of the `--`.

This isn't another case of #4672 as there's no leading slash in the org name.

It's definitely the shim batch file that's triggering the problem as running the same commands as TeamCity does, but manually, allows the checkout to work and the normal SSH command is used instead of the batch file

**System environment**
Windows Server 2019 Datacenter version 1809 build 17763.2237
Git called via TeamCity Enterprise 2017.2.2 (build 50909) with the agent process running as a service.
Git for Windows 2.33.1.windows.1
Git LFS 3.0.1

**Output of `git lfs env`**
This is likely inaccurate as I can't run this in the same context as Git runs when things break and I expect that some environment variables are different.
```
git-lfs/3.0.1 (GitHub; windows amd64; go 1.17.1)
git version 2.33.1.windows.1

Endpoint=https://github.com/org/repo.git/info/lfs (auth=none)
  SSH=git@github.com:org/repo.git
LocalWorkingDir=C:\buildAgentData\work\c43e8748fe05225d\repo
LocalGitDir=C:\buildAgentData\work\c43e8748fe05225d\repo\.git
LocalGitStorageDir=C:\buildAgentData\work\c43e8748fe05225d\repo\.git
LocalMediaDir=C:\buildAgentData\work\c43e8748fe05225d\repo\.git\lfs\objects
LocalReferenceDirs=
TempDir=C:\buildAgentData\work\c43e8748fe05225d\repo\.git\lfs\tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=C:\buildAgentData\work\c43e8748fe05225d\repo\.git\lfs
AccessDownload=none
AccessUpload=none
GIT_TRACE=C:\git-trace.log
```

**Additional context**
As mentioned right at the top, this is really TeamCity's fault for making an SSH shim that can't handle all the arguments that might be passed to SSH, not Git LFS' fault for passing SSH arguments SSH is compatible with.

bk2204:
In Git LFS version 3.0, we added support for understanding the GIT_SSH_VARIANT environment variable and its corresponding configuration option.  When we did so, we dropped the special-casing for program names and instead always used the -- setting for the default variant, "ssh". This variant represents OpenSSH, which as a program with normal POSIX option handling, supports -- just fine.

Unfortunately, not all clients do.  TeamCity's SSH client doesn't set the variant and also doesn't understand --.  We need to handle this case to prevent option injection in case someone tries to use a URL like "ssh://-oProxyCommand=exploit/", which could otherwise lead to arbitrary code execution.

However, we don't have to add this all the time, but only when the user-and-host portion starts with a dash.  The reason is that the only other place an attacker could inject an option is in the path, and the path is not a separate option by itself.  The path is always preceded by a command (either "git-lfs-authenticate" or "git-lfs-transfer" in a single option that contains the name of the command, the path, and the operation, separated by spaces.  As a result, option injection is not possible in the path since those commands don't take options.

Note that Git just dies in this case with a message like so:

  fatal: strange hostname '-oProxyCommand=exploit' blocked

Let's adopt this approach of using -- less often so that we can more gracefully deal with this case and fix some SSH clients, even if they really should learn to understand this option.

Fixes #4733

### DIFF
--- a/ssh/ssh.go
+++ b/ssh/ssh.go
@@ -173,18 +173,22 @@ func GetExeAndArgs(osEnv config.Environment, gitEnv config.Environment, meta *SS
 		args = append(args, meta.Port)
 	}
 
-	if variant == variantSSH {
-		// inserts a separator between cli -options and host/cmd commands
-		// example: $ ssh -p 12345 -- user@host.com git-lfs-authenticate ...
-		args = append(args, "--", meta.UserAndHost)
+	if sshOptPrefixRE.MatchString(meta.UserAndHost) {
+		if variant == variantSSH {
+			// inserts a separator between cli -options and host/cmd commands
+			// example: $ ssh -p 12345 -- user@host.com git-lfs-authenticate ...
+			args = append(args, "--", meta.UserAndHost)
+		} else {
+			// no prefix supported, strip leading - off host to prevent cmd like:
+			// $ git config lfs.url ssh://-proxycmd=whatever
+			// $ plink -P 12345 -proxycmd=foo git-lfs-authenticate ...
+			//
+			// Instead, it'll attempt this, and eventually return an error
+			// $ plink -P 12345 proxycmd=foo git-lfs-authenticate ...
+			args = append(args, sshOptPrefixRE.ReplaceAllString(meta.UserAndHost, ""))
+		}
 	} else {
-		// no prefix supported, strip leading - off host to prevent cmd like:
-		// $ git config lfs.url ssh://-proxycmd=whatever
-		// $ plink -P 12345 -proxycmd=foo git-lfs-authenticate ...
-		//
-		// Instead, it'll attempt this, and eventually return an error
-		// $ plink -P 12345 proxycmd=foo git-lfs-authenticate ...
-		args = append(args, sshOptPrefixRE.ReplaceAllString(meta.UserAndHost, ""))
+		args = append(args, meta.UserAndHost)
 	}
 
 	return cmd, args, needShell
--- a/ssh/ssh_test.go
+++ b/ssh/ssh_test.go
@@ -22,7 +22,6 @@ func TestSSHGetLFSExeAndArgs(t *testing.T) {
 	exe, args := ssh.GetLFSExeAndArgs(cli.OSEnv(), cli.GitEnv(), &meta, "git-lfs-authenticate", "download", false)
 	assert.Equal(t, "ssh", exe)
 	assert.Equal(t, []string{
-		"--",
 		"user@foo.com",
 		"git-lfs-authenticate user/repo download",
 	}, args)
@@ -30,7 +29,6 @@ func TestSSHGetLFSExeAndArgs(t *testing.T) {
 	exe, args = ssh.GetLFSExeAndArgs(cli.OSEnv(), cli.GitEnv(), &meta, "git-lfs-authenticate", "upload", false)
 	assert.Equal(t, "ssh", exe)
 	assert.Equal(t, []string{
-		"--",
 		"user@foo.com",
 		"git-lfs-authenticate user/repo upload",
 	}, args)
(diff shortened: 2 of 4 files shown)


chrisd8088 (changes requested):
The code changes here all look good, thank you!  It took me a while to wrap my head around what was changing, even though it's not particularly complex, mostly because I got myself mixed up about the test suite and test helper.

However, in the process of untangling myself, I did find some oddities about how the SSH arguments are processed by the test helper (I believe it can still be passed `--` in the default test case, for instance) and also that I could bork a test by setting `GIT_SSH_VARIANT`, and that when doing so one `grep` passed in an unexpected fashion.

Anyway, as a result I ended up putting together a [branch](https://github.com/chrisd8088/git-lfs/tree/dash-dash-ssh-tests) with a couple of additional commits that I think might be valuable.

If that doesn't look OK, though, the main thing I would suggest changing is just to in some way restore the handling of `--` in `t/cmd/lfs-ssh-echo.go` because it does, I believe, still get passed that end-of-options delimiter, so it should be looking for it in the default case.

bk2204:
I've looked through your comments here and they all seem sensible and reasonable.  What I've done is squashed your first commit into mine, changing the commit message slightly and adding a co-author credit, so that we don't leave my commit with nonsensical behavior uncorrected, and then cherry-picked your second commit, since that seems like it's fine as it is.

Thanks for the thorough review, and hopefully what I've done is okay with you.

chrisd8088 (approved):
All looks good to me -- thank you for making this so tidy!

chrisdjali-wrld3d:
It's been several weeks since this was merged. What kind of timeframe can we expect for it to appear in a tagged release?

chrisd8088:
We expect to do a release sometime this month (January), but of course that's always dependent on external factors (jobs, life, etc.) which may cause a delay.  But the expectation is for a release in January.

PR #4760, opened 2021-12-03
post-checkout: don't modify permissions of untracked files

### LINKED ISSUE #4757: Why does git lfs track not ignore untracked files in git
The description of the LFS post-checkout hook is "Responds to Git post-checkout events. It makes sure that any files which are marked as lockable by git lfs track are read-only in the working copy, if not currently locked by the local user."

This makes sense for git tracked files.  Apparently though, it applies to git untracked files in the working copy.

echo tracked > tracked.txt
git add tracked.txt
git commit tracked.txt -m "tracked"
echo untracked > untracked.txt
ls -l  
total 2
-r--r--r-- 1 My Name 1049089 8 Dec 2 09:17 tracked.txt
-rw-r--r-- 1 My Name 1049089 10 Dec 2 09:18 untracked.txt

git checkout master -- tracked.txt
ls -l
total 2
-r--r--r-- 1 My Name 1049089 8 Dec 2 09:17 tracked.txt
-r--r--r-- 1 My Name 1049089 10 Dec 2 09:18 untracked.txt

.gitattributes
    * filter=lfs diff=lfs merge=lfs -text lockable
    .gitattributes -filter -diff -merge -text
    .gitattributes export-ignore
    README.md export-ignore

bk2204:
Git doesn't know about untracked files, so we should avoid modifying them when we adjust permissions in the post-checkout hook.  Add an option to our invocation of git ls-files that controls the use of the `--others` flag (which controls listing untracked files), and disable it when we're looking for files to process with the post-checkout hook.

Fixes #4757
/cc @rainwf as reporter

### DIFF
--- a/git/attribs.go
+++ b/git/attribs.go
@@ -177,7 +177,7 @@ func findAttributeFiles(workingDir, gitDir string) []attrFile {
 		paths = append(paths, attrFile{path: repoAttributes, readMacros: true})
 	}
 
-	lsFiles, err := NewLsFiles(workingDir, true)
+	lsFiles, err := NewLsFiles(workingDir, true, true)
 	if err != nil {
 		tracerx.Printf("Error finding .gitattributes: %v", err)
 		return paths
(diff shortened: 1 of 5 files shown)


chrisd8088 (approved):
Looks good, thanks

PR #4768, opened 2021-12-07
filter-process: don't print large file warning on fixed versions

### LINKED ISSUE #4744: Encountered (N) file(s) that may not have been copied correctly on Windows
**Describe the bug**
with Git-for-windows 2.34.0 the checksum now supports 4GB+ and therefor fixed the smudge with the recent release.

**To Reproduce**
1. Clone a repository with configured lfs
2. Add a +4gb file and commit

4. See warning: 
```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        test.bin

nothing added to commit but untracked files present (use "git add" to track)

d:\Github\nexuslfs>git add test.bin
Encountered 1 file(s) that may not have been copied correctly on Windows:
        test.bin

See: `git lfs help smudge` for more details.
```

**Expected behavior**
don't show warning when git.exe is of greater version than 2.34.0

**System environment**
Windows 10 Pro,
git-for-windows 2.34.0.windows.1
.gitconfig = user info only.

**Output of `git lfs env`**

```
```

bk2204:
In Git for Windows 2.34.0 and the upcoming Git 2.35.0, the issue that's caused corruption on Windows has been fixed.  Let's avoid warning users that there's a problem if they're using 2.34.0 or newer unless they've specifically set the option to true.  Note that we only consider 2.34.0 and not 2.35 since Unix has never had this issue.  Update the documentation as well to reflect the new default.

Fixes #4744

### DIFF
--- a/docs/man/git-lfs-config.5.ronn
+++ b/docs/man/git-lfs-config.5.ronn
@@ -111,8 +111,9 @@ be scoped inside the configuration for a remote.
 * `lfs.largefilewarning`
 
   Warn when a file is 4 GiB or larger. Such files will be corrupted when using
-  Windows (unless smudging is disabled) due to a limitation in Git.  Default:
-  true.
+  Windows (unless smudging is disabled) with a Git for Windows version less than
+  2.34.0 due to a limitation in Git.  Default: true if the version is less than
+  2.34.0, false otherwise.
 
 ### Transfer (upload / download) settings
 
--- a/commands/command_filter_process.go
+++ b/commands/command_filter_process.go
@@ -212,7 +212,7 @@ func filterCommand(cmd *cobra.Command, args []string) {
 		}
 	}
 
-	if len(malformedOnWindows) > 0 && cfg.Git.Bool("lfs.largefilewarning", true) {
+	if len(malformedOnWindows) > 0 && cfg.Git.Bool("lfs.largefilewarning", !git.IsGitVersionAtLeast("2.34.0")) {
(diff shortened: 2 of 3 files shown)


docs/man/git-lfs-config.5.ronn
```diff
@@ -111,8 +111,9 @@ be scoped inside the configuration for a remote.
 * `lfs.largefilewarning`
 
   Warn when a file is 4 GiB or larger. Such files will be corrupted when using
-  Windows (unless smudging is disabled) due to a limitation in Git.  Default:
-  true.
+  Windows (unless smudging is disabled) with a Git for Windows version less than
+  2.34.0 due to a limitation in Git.  Default: true if the version is less than
+  2.34.0, false otherwise.
```
chrisd8088, line 117: I would also suggest changing the line in "Known Bugs" [section](https://github.com/git-lfs/git-lfs/blob/b197e981b5da134bdc5a94db3b23d5179f9ad6ed/docs/man/git-lfs-smudge.1.ronn#known-bugs) of `docs/man/git-lfs-smudge.1.ronn` which reads:
> `On Windows, Git does not handle files in the working tree larger than 4 gigabytes.`
>
> `For more information, see: https://github.com/git-lfs/git-lfs/issues/2434.`

chrisd8088 (approved):
One suggestion only ... it'll be great to have this merged!

docs/man/git-lfs-config.5.ronn
bk2204: Updated.  I've removed the link to the bug and stated that newer versions are unaffected.

PR #4771, opened 2021-12-08
migrate import: don't allow path filters with --above

bk2204:
The `--include`, `--exclude`, and `--fixup` options don't work properly with `--above`, so let's print an error instead of silently doing the wrong thing.  Document this fact as well to make sure that people aren't surprised.


chrisd8088 (approved):
LGTM, thanks.

PR #4772, opened 2021-12-08
Add CODEOWNERS to automatically ping on review

bk2204:
Currently, when someone opens a PR, they either have to ping the @git-lfs/core team as reviewers or hope that someone sees the PR.  Let's help the users out by automatically pinging the team for review by adding a CODEOWNERs, which we can then work into our branch protection rules.


chrisd8088 (approved):
👍

PR #4776, opened 2021-12-10
Move `err` checking to before the value was used

donno2048:
Or else `apiClient` might not have an `Endpoints` attribute and just be `nil` when accessed


bk2204:
Hey,

Thanks for the patch.  Are you actually seeing a problem with this code, or is this purely precautionary?

donno2048:
Purely precautionary

bk2204 (approved):

PR #4795, opened 2021-12-21
.github/workflows: switch to actions/checkout@v2

### LINKED ISSUE #4789: [Security] Workflow release.yml is using vulnerable action actions/checkout
The workflow release.yml is referencing action actions/checkout using references v1. However this reference is missing the commit [a6747255bd19d7a757dbdda8c654a9f84db19839](https://github.com/actions/checkout/commits/a6747255bd19d7a757dbdda8c654a9f84db19839) which may contain fix to the some vulnerability. 
The vulnerability fix that is missing by actions version could be related to: 
(1) CVE fix 
(2) upgrade of vulnerable dependency 
(3) fix to secret leak and others. 
Please consider to update the reference to the action.

bk2204:
The actions/checkout workflow looks like it may have had a security-sensitive change to it in the latest v2.  It doesn't look like it affects us, but let's update to be safe.

Switch to actions/checkout@v2 everywhere and use the fetch-depth: 0 argument, which makes us fetch the entire history.  This is the main reason we stuck with v1 for so long, and it's required for certain parts of our build and release process to function correctly.

Fixes #4789

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,9 @@ jobs:
         go: ['1.17.x']
     runs-on: ${{ matrix.os }}
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
     - uses: actions/setup-ruby@v1
     - uses: actions/setup-go@v2
       with:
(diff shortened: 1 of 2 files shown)


chrisd8088 (approved):
👍

Issue #4167, opened 2020-06-20
git lfs migrate import on path doesn't handle files that have been moved


amaidment:
During the course of using `git lfs migrate import [path]/**` (on a `--mirror clone` of our repo), I found it gave me some odd behaviour where files had been moved between directories in the repository - for example, if the file was in com/xyz/abc/big_file.csv, but at some point in time was moved to com/xyz/def/big_file.csv.

**Migrate Current Path**
If I do `git lfs migrate import com/xyz/def/** --verbose`, it does _not_ output that the file (com/xyz/def/big_file.csv) has been moved to LFS. While it adds com/xyz/def/** to .gitattributes, it doesn't actually move the file to LFS storage. As such, if I then do a (non-mirror) clone from my mirror repo, I get the following error:
> message:Encountered 1 file(s) that should have been pointers, but weren't:
>         com/xyz/def/big_file.csv

**Migrate Original Path**
Instead, if I do the `git lfs migrate import com/xyz/abc/** --verbose`, it outputs that the file (com/xyz/abc/big_file.csv) has been moved to LFS. 

> migrate: Sorting commits: ..., done.
> migrate: commit [blah]: com/xyz/abc/big_file.csv
> migrate: Rewriting commits: 100% (x/x), done.

So, it adds com/xyz/abc/** to .gitattributres, and moves the file to LFS storage. However, it doesn't seem to handle that the file has been moved correctly. As such, if I then do a (non-mirror) clone from my mirror repo, there are no error messages, but the repo's state seems inconsistent.That is that .gitattributes contains the original path:

> com/xyz/abc/** filter=lfs diff=lfs merge=lfs -text

However, if I look at what files are stored under LFS, it shows the file at its new path.

> $ git lfs ls-filesblah - com/xyz/def/big_file.csv

I'm not quite sure what the desired behaviour should be, in either of these cases, but what it's actually doing in both these cases seems confusingly inconsistent.

**Workarounds**
I found two workarounds for this problem:

1. Rather than doing the LFS migration for files in specific locations, I broadened the migration (e.g. across test resources) to a higher directory - i.e. equivalent to doing git lfs migrate import com/xyz/**

2. Where 1) was neither possible or appropriate (i.e. the path change was at high level directory, e.g. between modules), I found that I could do both the lfs migration on both directories (in the appropriate order) - i.e.

> $ git lfs migrate import com/xyz/abc/** 
> ...
> $ git lfs migrate import com/xyz/def/**  
> ...
[...]

PR #4813, opened 2022-01-12
Fix migrate import speed regression

### LINKED ISSUE #4750: git lfs migrate import performance issues with version 3.0.2
**Describe the bug**
Running **git lfs migrate import** using version 3.0.2 is around 60 times slower compared to 3.0.1. Git repository with 5941 commits takes around 9 minutes to complete using version 3.0.1 and using 3.0.2 only managed to reach 60% of the commits in 6 hours of running.

**To Reproduce**
1. Download and install Git version 2.34.1 or 2.34.0 from https://github.com/git-for-windows/git/releases
2. Run **git lfs migrate import --yes --include="..."** on a Git repository
3. Download 3.0.1 version of Git LFS from https://github.com/git-lfs/git-lfs/releases/download/v3.0.1/git-lfs-windows-amd64-v3.0.1.zip
4. Apply the git-lfs.exe file on top of mingw64\bin\git-lfs.exe in the install location of Git
5. Repeat step 2 on a fresh Git repository

**Expected behavior**
Time taken for git lfs migrate import command to run should be identical with v3.0.1 and v3.0.2.

**System environment**
Windows 10 v1909

**Output of `git lfs env`**
git-lfs/3.0.2 (GitHub; windows amd64; go 1.17.2)
git version 2.34.0.windows.1

LocalWorkingDir=D:\Work\SubTest\Subsurface_Git-2.34.0
LocalGitDir=D:\Work\SubTest\Subsurface_Git-2.34.0\.git
LocalGitStorageDir=D:\Work\SubTest\Subsurface_Git-2.34.0\.git
LocalMediaDir=D:\Work\SubTest\Subsurface_Git-2.34.0\.git\lfs\objects
LocalReferenceDirs=
TempDir=D:\Work\SubTest\Subsurface_Git-2.34.0\.git\lfs\tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=D:\Work\SubTest\Subsurface_Git-2.34.0\.git\lfs
AccessDownload=none
AccessUpload=none
GIT_EXE="C:\Git\Git-2.34.0\cmd\git.exe"
GIT_EXEC_PATH=C:/Git/Git-2.34.0/mingw64/libexec/git-core

**Additional context**
This issue seems to be independent from Git version itself as I managed to reproduce the same issue with replacing the Git LFS delivered in Git versions 2.33.0, 2.33.1, 2.34.0 and 2.34.1 with versions 2.13.3, 3.0.1 or 3.0.2 of Git LFS and all Git versions are affected when using 3.0.2.

bk2204:
When we cache files, do so on the full path instead of just the directory entry.  This means that when we have an identical file with the same name in two different direectories, we distinguish between the two paths and ensure both are added to .gitattributes.

This is an alternate solution to #4671 which should perform better.  For compmarison, with a clone of Git's main repository with the following command, we get:

`git lfs migrate import --everything --include="*.h"`:

* v3.0.1      (broken):  608s user,   53s system,    5:34 total
* v3.0.2      (fixed): 13435s user, 1255s system, 1:43:17 total
* this commit (fixed):   716s user,   67s system,    6:59 total

This is a much better performance characteristic for equivalent results.

Preserve the integration from the earlier attempt at fixing this plus add an additional one.  Avoid using `assert_pointer` in the new test because that helper doesn't always work correctly when there are two files with the same file name.

Fixes #4750

### DIFF
--- a/t/t-migrate-import.sh
+++ b/t/t-migrate-import.sh
@@ -1053,6 +1053,25 @@ begin_test "migrate import (copied file)"
 )
 end_test
 
+begin_test "migrate import (copied file with only a single path)"
+(
+  set -e
+
+  setup_local_branch_with_copied_file
+
+  oid="$(calc_oid "$(git cat-file -p :a.txt)")"
+
+  # Prevent MSYS from rewriting /a.txt into a Windows path.
+  MSYS_NO_PATHCONV=1 git lfs migrate import --include="/a.txt" --everything
+
+  # Expect attribute for only "/a.txt".
+  if grep -q "^/dir/a.txt" ./.gitattributes || ! grep -q "^/a.txt" ./.gitattributes; then
+    exit 1
+  fi
+  refute_pointer "refs/heads/main" "dir/a.txt" "$oid" 5
+)
+end_test
+
 begin_test "migrate import (filename special characters)"
 (
   set -e
--- a/commands/command_migrate_export.go
+++ b/commands/command_migrate_export.go
@@ -43,7 +43,7 @@ func migrateExportCommand(cmd *cobra.Command, args []string) {
 	opts := &githistory.RewriteOptions{
 		Verbose:           migrateVerbose,
 		ObjectMapFilePath: objectMapFilePath,
-		BlobFn: func(path string, oid []byte, b *gitobj.Blob) (*gitobj.Blob, error) {
+		BlobFn: func(path string, b *gitobj.Blob) (*gitobj.Blob, error) {
 			if filepath.Base(path) == ".gitattributes" {
 				return b, nil
 			}
--- a/commands/command_migrate_import.go
+++ b/commands/command_migrate_import.go
@@ -146,12 +146,10 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 		}
 	}
 
-	blobCache := make(map[string]bytes.Buffer)
-
 	migrate(args, rewriter, l, &githistory.RewriteOptions{
 		Verbose:           migrateVerbose,
 		ObjectMapFilePath: objectMapFilePath,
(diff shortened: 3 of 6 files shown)


t/t-migrate-import.sh
```diff
@@ -1053,6 +1053,25 @@ begin_test "migrate import (copied file)"
...
+  set -e
+
+  setup_local_branch_with_copied_file
+
+  oid="$(calc_oid "$(git cat-file -p :a.txt)")"
+
+  # Prevent MSYS from rewriting /a.txt into a Windows path.
+  MSYS_NO_PATHCONV=1 git lfs migrate import --include="/a.txt" --everything
+
+  # Expect attributes for "/dir/a" and "/a"
```
chrisd8088, line 1067: Minor nit: I think this comment should say something like `Expect attribute only for "/a"`.

chrisd8088 (approved):
This looks to be an incredibly tidy and elegant solution to a long-standing problem, along with reversing the regression.  Wow, thank you!  🙇 

I believe this may also resolve #4167, from my own testing.  I don't know if it's worth adding another test case, though; I think the one you've added here might be sufficient, although it deals with copied files, not moved ones.

One note: I notice that the commit description and this PR's description mention PR #4176, but I think the PR in question is actually #4671.  It might be good to update at least the commit description, for future clarity.  (Note that that's different again from #4167 I mentioned above, which is a still-open bug report.)

Again, thank you for this.  ❤️

t/t-migrate-import.sh
bk2204: Good point.

PR #4815, opened 2022-01-13
Fall back from Negotiate to Basic

bk2204:
The Negotiate authentication scheme can support multiple different types of authentication.  The two most popular are NTLM and Kerberos.  When we supported NTLM, we'd first try Kerberos, and if it failed, fall back to NTLM.

However, we no longer support NTLM, but some people still have server configuration that uses NTLM via Negotiate.  For these people, authentication may be broken.  Let's fall back to Basic in such a case by keeping track of which authentication mechanisms we've tried, keeping only the supported mechanisms if we got a response, and stripping out failing mechanisms, falling back to Basic.

To help with servers that support both Negotiate and Basic, we specifically consider SPNEGO (Negotiate) errors as authentication errors.  This is because if the server supports Kerberos but the client does not have a ticket, then we'll get an error trying to read the client's tickets, which will manifest in this way.

### DIFF
--- a/lfsapi/auth.go
+++ b/lfsapi/auth.go
@@ -68,9 +68,11 @@ func (c *Client) doWithAuth(remote string, access creds.Access, req *http.Reques
 	res, err := c.doWithCreds(req, credWrapper, access, via)
 	if err != nil {
 		if errors.IsAuthError(err) {
-			newAccess := access.Upgrade(getAuthAccess(res))
+			newMode, newModes := getAuthAccess(res, access.Mode(), c.access)
+			newAccess := access.Upgrade(newMode)
 			if newAccess.Mode() != access.Mode() {
 				c.Endpoints.SetAccess(newAccess)
+				c.access = newModes
 			}
 
 			if credWrapper.Creds != nil {
@@ -312,20 +314,34 @@ var (
 	authenticateHeaders = []string{"Lfs-Authenticate", "Www-Authenticate"}
 )
 
-func getAuthAccess(res *http.Response) creds.AccessMode {
-	for _, headerName := range authenticateHeaders {
-		for _, auth := range res.Header[headerName] {
-			pieces := strings.SplitN(strings.ToLower(auth), " ", 2)
-			if len(pieces) == 0 {
-				continue
-			}
+func getAuthAccess(res *http.Response, access creds.AccessMode, modes []creds.AccessMode) (creds.AccessMode, []creds.AccessMode) {
+	newModes := make([]creds.AccessMode, 0, len(modes))
+	for _, mode := range modes {
+		if access != mode {
+			newModes = append(newModes, mode)
+		}
+	}
+	if res != nil {
+		supportedModes := make(map[creds.AccessMode]struct{})
+		for _, headerName := range authenticateHeaders {
+			for _, auth := range res.Header[headerName] {
+				pieces := strings.SplitN(strings.ToLower(auth), " ", 2)
+				if len(pieces) == 0 {
+					continue
+				}
 
-			switch creds.AccessMode(pieces[0]) {
-			case creds.NegotiateAccess:
-				return creds.AccessMode(pieces[0])
+				switch creds.AccessMode(pieces[0]) {
+				case creds.BasicAccess, creds.NegotiateAccess:
+					supportedModes[creds.AccessMode(pieces[0])] = struct{}{}
+				}
+			}
+		}
+		for _, mode := range newModes {
+			if _, ok := supportedModes[mode]; ok {
+				return mode, newModes
 			}
 		}
 	}
 
-	return creds.BasicAccess
+	return creds.BasicAccess, newModes
 }
--- a/lfsapi/auth_test.go
+++ b/lfsapi/auth_test.go
@@ -39,9 +39,21 @@ func TestAuthenticateHeaderAccess(t *testing.T) {
 			res := &http.Response{Header: make(http.Header)}
 			res.Header.Set(key, value)
 			t.Logf("%s: %s", key, value)
-			assert.Equal(t, expected, getAuthAccess(res))
+			result, _ := getAuthAccess(res, creds.NoneAccess, creds.AllAccessModes())
+			assert.Equal(t, expected, result)
 		}
 	}
+}
+
+func TestDualAccessModes(t *testing.T) {
+	res := &http.Response{Header: make(http.Header)}
+	res.Header["Www-Authenticate"] = []string{"Negotiate 123", "Basic 456"}
+	access, next := getAuthAccess(res, creds.NoneAccess, creds.AllAccessModes())
+	assert.Equal(t, creds.NegotiateAccess, access)
+	access, next = getAuthAccess(res, access, next)
+	assert.Equal(t, creds.BasicAccess, access)
+	access, _ = getAuthAccess(res, access, next)
+	assert.Equal(t, creds.BasicAccess, access)
 }
 
 func TestDoWithAuthApprove(t *testing.T) {
--- a/creds/access.go
+++ b/creds/access.go
@@ -31,3 +31,12 @@ func (a *Access) Mode() AccessMode {
 func (a *Access) URL() string {
 	return a.url
 }
+
+// AllAccessModes returns all access modes in the order they should be tried.
+func AllAccessModes() []AccessMode {
+	return []AccessMode{
+		NoneAccess,
+		NegotiateAccess,
+		BasicAccess,
+	}
+}
--- a/lfsapi/lfsapi.go
+++ b/lfsapi/lfsapi.go
@@ -17,6 +17,7 @@ type Client struct {
 
 	client  *lfshttp.Client
 	context lfshttp.Context
+	access  []creds.AccessMode
 }
 
 func NewClient(ctx lfshttp.Context) (*Client, error) {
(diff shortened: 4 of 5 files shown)


lfsapi/auth.go
```diff
@@ -311,20 +314,41 @@ var (
...
+			newModes = append(newModes, mode)
+		}
+	}
+	if len(newModes) == 0 {
+		for _, mode := range creds.AllAccessModes() {
+			if mode != creds.NoneAccess {
+				newModes = append(newModes, mode)
 			}
+		}
+	}
```
chrisd8088, line 330: I confess I'm a little unclear on why this block is here ... I see, I think, that it causes the set of potential modes to be "reset", as it were, if none are left; that's why, in the accompanying test, the returned modes cycle through `negotiate`, `basic`, then `negotiate` again, `basic` again, and so on if one keeps going.  (I tried tinkering with this and it does cycle without end.)

What I'm unclear on is how this will interoperate with the [recursive call](https://github.com/git-lfs/git-lfs/blob/ee3088d4804bd2b699eb6a76aa1318d1d2283ed6/lfsapi/auth.go#L35) in `DoWithAuth()` above; see my comments in that function.

lfsapi/auth.go
```diff
@@ -66,10 +66,13 @@ func (c *Client) doWithAuth(remote string, access creds.Access, req *http.Reques
 
 	res, err := c.doWithCreds(req, credWrapper, access, via)
 	if err != nil {
+		tracerx.Printf("dx: a: %s %v\n", err, errors.IsAuthError(err))
 		if errors.IsAuthError(err) {
-			newAccess := access.Upgrade(getAuthAccess(res))
+			newMode, newModes := getAuthAccess(res, access.Mode(), c.access)
+			newAccess := access.Upgrade(newMode)
```
chrisd8088, line 72: Looking through the code in this section I was struck by the fact that this function's caller, `DoWithAuth()`, appears to call itself recursively when the request header `Authorization` has been deleted but the server sent a 401; this is sort of a signal from the client to itself to try again, as it deletes the `Authorization` header just a bit below in this block.  That logic looks to have been introduced in #3293, along with similar blocks in the basic [download](https://github.com/git-lfs/git-lfs/blob/ee3088d4804bd2b699eb6a76aa1318d1d2283ed6/tq/basic_download.go#L280-L282) and [upload](https://github.com/git-lfs/git-lfs/blob/ee3088d4804bd2b699eb6a76aa1318d1d2283ed6/tq/basic_upload.go#L228-L239) adapters.

But I don't see a way to break out of this loop should the server keep sending 401s, and so the `getAuthAccess()` changes here which add another endless loop of access modes (negotiate -> basic -> negotiate -> basic -> ...) leaves me uncomfortable.

I tried, as an experiment, taking this PR's branch and replacing a [conditional](https://github.com/git-lfs/git-lfs/blob/ee3088d4804bd2b699eb6a76aa1318d1d2283ed6/t/cmd/lfstest-gitserver.go#L1416-L1420) in the `t/cmd/lfstest-gitserver.go` test server with an unconditional return of a 401, e.g.:
```go
	writeLFSError(w, 401, "Error: Authorization Required")
	return true
```
which simulates a broken server that keeps returning 401s.  Sure enough, the `t/t-credential.sh` test suite hangs indefinitely on a [push](https://github.com/git-lfs/git-lfs/blob/ee3088d4804bd2b699eb6a76aa1318d1d2283ed6/t/t-credentials.sh#L409) in `"credentials from lfs.url"` test.  Adding some tracing in `lfsapi/auth.go` shows it definitely just keeps calling `doWithAuth()` again, and while doing so, it keeps getting `basic` back from `getAuthAccess()`.

Personally I would love to see the logic here made somehow a bit more clear and with a definite end.  Naively, I would expect this code to start with a list of supported access modes (like this PR defines in `AllAccessModes()`) and then try those in sequence, stopping when they run out.  If the server replies with a way to negotiate, great, but I feel like there should be a simple `for` here stepping through the available modes the client supports, skipping any not supported by the server, and stopping when the list is exhausted.

I realize that adds to the scope, but I'm honestly unsure how to assess the changes here at the moment, given the recursive situation that appears to exist.

chrisd8088 (commented):
Thank you for jumping on this; I don't know how I would simulate the problem, as my Windows skills are limited.

I'm uncertain how some of the negotiation logic here interacts with what I think might be the existing potential in the Git LFS code for a broken server to cause the client to hang indefinitely.  My apologies for not getting further with an analysis.

lfsapi/auth.go
```diff
@@ -66,10 +66,13 @@ func (c *Client) doWithAuth(remote string, access creds.Access, req *http.Reques
 
 	res, err := c.doWithCreds(req, credWrapper, access, via)
 	if err != nil {
+		tracerx.Printf("dx: a: %s %v\n", err, errors.IsAuthError(err))
```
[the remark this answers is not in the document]
bk2204: That's a debugging comment which I thought I deleted but apparently did not.

lfsapi/auth.go
bk2204: Yeah, that was supposed to be intentional, but if you'd prefer we don't, I can remove it.

lfsapi/auth.go
bk2204: It looks like we already have that behavior in the main branch.  What that means is that you'll get prompted for credentials again if your password is bad, since we'll erase the credentials and try again.  In our test code, we don't ever prune the credentials like a real credential helper, so we don't see that behavior.  If you enter real credentials but don't have access, you're supposed to get a 403 or 404, not a 401.

I'm going to remove the cycling behavior and preserve this behavior, because it's intended to work that way from what I can tell.

lfsapi/auth_test.go
```diff
@@ -39,11 +39,25 @@ func TestAuthenticateHeaderAccess(t *testing.T) {
...
+func TestDualAccessModes(t *testing.T) {
+	res := &http.Response{Header: make(http.Header)}
+	res.Header["Www-Authenticate"] = []string{"Negotiate 123", "Basic 456"}
+	access, next := getAuthAccess(res, creds.NoneAccess, creds.AllAccessModes())
+	assert.Equal(t, creds.NegotiateAccess, access)
+	access, next = getAuthAccess(res, access, next)
+	assert.Equal(t, creds.BasicAccess, access)
+	access, next = getAuthAccess(res, access, next)
+	assert.Equal(t, creds.NegotiateAccess, access)
+	access, next = getAuthAccess(res, access, next)
```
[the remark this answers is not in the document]
bk2204: Fixed.

bk2204:
I've made the changes I mentioned above.  I do want to point out that this was not tested on Windows; instead, I use Kerberos for my network of Debian machines, and I tested this on a subdomain of my regular web server with a proxy to the lfs-test-server.  However, it does show that we fall back if Negotiate fails, whereas we did not before.

lfsapi/auth.go
chrisd8088: It's definitely behaviour that's already in the main branch, so it's not immediately germane to this PR, but I couldn't untangle the two cycling interactions myself.

What I was struck by with the existing code is that if the server was broken somehow and always returned 401 instead of the expected 403, I'm not sure we wouldn't just enter an infinite loop.  I would certainly expect that if the server returned 403, we'd prompt with another chance to re-enter credentials.  But the logic in these three places (one here and one each in the upload and download adapters), where there's a recursive call without a maximum number of calls, just makes me suspicious that a bad server or a bad credential helper or something could put the client into a loop.  I'd love to be proven wrong, though.

bk2204: We have the same behaviour as Git does here.  If we get a 403, we give up, since the user has valid credentials, but they're not authorized to access the data.  That's always fatal.

We will technically enter an infinite loop if we keep getting 401s, but we'll keep prompting them for credentials using `git credential fill`, since we'll eventually always fall back to Basic.  Technically the user can just keep filling the credentials, or they can use a credential helper like ours in the test that never returns anything different, in which case they'll loop forever.  But that's also, as I understand it from working on it, Git's behaviour here as well, and in the normal case, they'll give up because the credential prompt is annoying or get a 403 or 404 and stop.  Notably, in many cases, the server will actually return a 403 or 404 if there are credentials provided and they're not valid (since that represents anonymous access, which isn't valid), which is often why this typically does abort in the general case.

Technically, we could hit that case, but, as I said, that's also true of Git, and I'm not worried about it.  I've removed my new cycling behavior, and we just devolve into an instance of Basic, so hopefully that's at least easier to reason about here.  If you want to add a check on the number of retries we do, I'm happy to do so, but I'd prefer to do so in a new PR, if that's all right with you.  If it's really too hard to reason about and it's important that it's in this PR, I can do that, though.

chrisd8088: No, I trust your experience here; and I don't think this should change in this PR, at a minimum.

lfsapi/auth.go
```diff
@@ -311,20 +313,34 @@ var (
...
-				continue
-			}
+func getAuthAccess(res *http.Response, access creds.AccessMode, modes []creds.AccessMode) (creds.AccessMode, []creds.AccessMode) {
+	newModes := make([]creds.AccessMode, 0, len(modes))
+	for _, mode := range modes {
+		if access != mode {
+			newModes = append(newModes, mode)
+		}
+	}
+	supportedModes := make(map[creds.AccessMode]struct{})
```
chrisd8088, line 323: I suppose this line could move inside the `if res != nil { ... }` block.

chrisd8088 (approved):
I'm going to go ahead and give this an approval as it fixes the reported problem and is certainly comprehensible on its own.

I do wish I found the authentication and authorization code in this area simpler to understand, but that's on me.

PR #4846, opened 2022-01-30
Update and mark message strings for translation

chrisd8088:
Following from the work in PR #4781 and #4729, we mark a number of additional message strings for translation, and revise various message strings with the aim of improving their consistency and clarity.

This PR may be most easily reviewed in commit-by-commit fashion, although it should also be readable as a single diff.

Aside from adding messages to the set of translatable string, which we undertake in the final commit, the preceding commits attempt to address a number of issues, including:

- Using consistent capitalization for acronyms, abbreviations, and project names (e.g., "Git" and "Git LFS").
    - Note that this PR does not attempt to ensure all messages begin with a capitalized (or all-lowercase) first word.
- Improving our grammar and punctuation in a few messages.
- Removing short message prefixes where they are used inconsistently within a package and do not add particular value.
    - We leave these types of prefixes in place in [trace](https://github.com/git-lfs/git-lfs/blob/02448f5fde353bd34236f7e2679678d46198195d/tq/transfer_queue.go#L523) messages intended for debugging, however, as they have utility there.
- Moving other short message prefixes specific to a Git LFS command, such as `migrate:`, out of the translatable strings, as they are intended to reference they command's name.
- Using backticks to delimit `git ...` and `git lfs ...` commands where they appear in messages.
- Using single quotes to delimit file names and paths, e.g., `'.gitattributes'`, where they appear in messages.
    - This PR does not achieve perfect consistency in this regard, though, as many message string formats use `%q` to quote user-supplied file paths, Git remotes, Git ref names, etc.
- Making some strings untranslatable, either because they appear in Go test files, or because they are fixed strings such as Git LFS command examples, Git configuration examples, etc.
- Moving most newlines outside of the translatable strings, as this should aid the task of ensuring translations have equivalent formatting, and because we can often make use of the formatting capabilities of the function enclosing the call to `tr.Tr.Get()` to do so.
- Dropping or changing the functions enclosing calls to `tr.Tr.Get()` to ensure we never pass a translated and interpolated string as another format string.

Re the last item, because the `tr.Tr.Get()` family of methods insert arguments into `printf(3)`-style format strings after translating the format string, we can in a few cases drop a surrounding call to `fmt.Sprintf()` or a similar method, as those now take no arguments and so are redundant.  This will help us avoid situations where either the translated string or the argument values interpolated by `tr.Tr.Get()` produce an output string which itself happens to contain character sequences that resemble Go format specifiers (e.g., `%s`, `%d`, etc.)  In such cases passing the string at runtime to a method such as `fmt.Fprintf()` will result in the output containing a warning such as `%!s(MISSING)`, which is not ideal.

As an [example](https://github.com/git-lfs/git-lfs/blob/02448f5fde353bd34236f7e2679678d46198195d/commands/command_pointer.go#L93), if `pointerFile` contained `%d` then the output from `fmt.Fprintf()` would contain `%!d(MISSING)`:
```go
	fmt.Fprintf(os.Stderr, tr.Tr.Get("Git LFS pointer for %s\n\n", pointerFile))
```

In one [case](https://github.com/git-lfs/git-lfs/blob/02448f5fde353bd34236f7e2679678d46198195d/lfs/attribute.go#L162-L163), in `lfs/attribute.go`, we can now also simplify the format string to use standard format specifiers instead of double-escaped ones (e.g., `%%q`) since we can just allow `tr.Tr.Get()` to do the interpolation.

We also take the opportunity to remove explicit leading or trailing newlines from translation messages wherever it is possible to convert the enclosing function to `fmt.Print()`, `fmt.Println()`, etc., noting that the former can be used to concatenate multiple strings without intermediate spaces, while the latter will insert spaces between all its arguments.

Within the `commands` package we can also make use of the formatting provided by the `Print()`, `Error()`, etc. functions, as they ultimately [utilize](https://github.com/git-lfs/git-lfs/blob/02448f5fde353bd34236f7e2679678d46198195d/commands/commands.go#L200-L203) `fmt.Fprintln()` when no arguments are provided, so a newline is automatically appended, and when arguments are provided, the `fmt.Fprintf()` is used but these utility functions [append](https://github.com/git-lfs/git-lfs/blob/02448f5fde353bd34236f7e2679678d46198195d/commands/commands.go#L204) a newline for us to the format string we pass.  But in that instance we then need to be careful not to pass our translated string as the format string, but as an argument, as noted above.

Lastly, we include panic messages amongst our translated strings as there is [already](https://github.com/git-lfs/git-lfs/blob/02448f5fde353bd34236f7e2679678d46198195d/lfs/diff_index_scanner.go#L73) [precedent](https://github.com/git-lfs/git-lfs/blob/02448f5fde353bd34236f7e2679678d46198195d/commands/command_fetch.go#L52) for translating these.

### DIFF
--- a/commands/command_filter_process.go
+++ b/commands/command_filter_process.go
@@ -206,9 +206,9 @@ func filterCommand(cmd *cobra.Command, args []string) {
 	}
 
 	if len(malformed) > 0 {
-		fmt.Fprintf(os.Stderr, tr.Tr.GetN(
-			"Encountered %d file that should have been pointers, but wasn't:\n",
-			"Encountered %d files that should have been pointers, but weren't:\n",
+		fmt.Fprintln(os.Stderr, tr.Tr.GetN(
+			"Encountered %d file that should have been pointers, but wasn't:",
+			"Encountered %d files that should have been pointers, but weren't:",
 			len(malformed),
 			len(malformed),
 		))
@@ -218,9 +218,9 @@ func filterCommand(cmd *cobra.Command, args []string) {
 	}
 
 	if len(malformedOnWindows) > 0 && cfg.Git.Bool("lfs.largefilewarning", !git.IsGitVersionAtLeast("2.34.0")) {
-		fmt.Fprintf(os.Stderr, tr.Tr.GetN(
-			"Encountered %d file that may not have been copied correctly on Windows:\n",
-			"Encountered %d files that may not have been copied correctly on Windows:\n",
+		fmt.Fprintln(os.Stderr, tr.Tr.GetN(
+			"Encountered %d file that may not have been copied correctly on Windows:",
+			"Encountered %d files that may not have been copied correctly on Windows:",
 			len(malformedOnWindows),
 			len(malformedOnWindows),
 		))
@@ -229,7 +229,7 @@ func filterCommand(cmd *cobra.Command, args []string) {
 			fmt.Fprintf(os.Stderr, "\t%s\n", m)
 		}
 
-		fmt.Fprintf(os.Stderr, tr.Tr.Get("\nSee: `git lfs help smudge` for more details.\n"))
+		fmt.Fprint(os.Stderr, "\n", tr.Tr.Get("See: `git lfs help smudge` for more details."), "\n")
 	}
 
 	if err := s.Err(); err != nil && err != io.EOF {
--- a/commands/command_filter_process.go
+++ b/commands/command_filter_process.go
@@ -207,7 +207,7 @@ func filterCommand(cmd *cobra.Command, args []string) {
 
 	if len(malformed) > 0 {
 		fmt.Fprintln(os.Stderr, tr.Tr.GetN(
-			"Encountered %d file that should have been pointers, but wasn't:",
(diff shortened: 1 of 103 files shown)


commands/command_filter_process.go
```diff
@@ -206,9 +206,9 @@ func filterCommand(cmd *cobra.Command, args []string) {
 	}
 
 	if len(malformed) > 0 {
-		fmt.Fprintf(os.Stderr, tr.Tr.GetN(
-			"Encountered %d file that should have been pointers, but wasn't:\n",
-			"Encountered %d files that should have been pointers, but weren't:\n",
+		fmt.Fprintln(os.Stderr, tr.Tr.GetN(
+			"Encountered %d file that should have been pointers, but wasn't:",
```
bk2204, line 210: Should we also have changed this to be "should have been a pointer"?

bk2204 (approved):
This looks great!  Thanks so much.

I had one minor thing I noticed, but I think this would be fine with or without that change.

commands/command_filter_process.go
chrisd8088: Yup, good eye, thank you!  Fixed.

PR #4835, opened 2022-01-25
fix merging of `.gitattributes` with execute file mode during migration

chrisd8088:
We currently generate incorrect duplicate `.gitattributes` entries when performing an import or export migration if an existing `.gitattributes` file has execute file permissions and we intend to merge new attribute filter lines into that file.

We first add a pair of tests for this condition, one each for the migration import and export operations, which each check for a single `.gitattributes` file after migration and therefore will fail with the current behaviour.

We then bump our `git-lfs/gitobj` module to v2.1.0, as it fixes merges of tree entries where an existing tree entry has a different file permissions mode than the tree entry being merged with it.  This should resolve the problem and allow our new tests to pass.

We also revise the names used for several test repositories created by the `git lfs migrate` tests, aligning them with their actual contents and with the majority of the other names used by the test suite.  And we slightly simplify some execute permission mode checks on a file in an existing test to use the shell `-x` operator, since that's what we're using in our new tests as well.

Note that handling of `.gitattributes` tree entries which are symbolic links is left for a subsequent PR.

Fixes #4796.
/cc @yhe13 as reporter.

Issue #4796, opened 2021-12-22
git lfs migrate import command corrupted object tree


yhe13:
**Describe the bug**
git lfs migrate import command corrupted object tree.
>Note: This happends 100% on one of our repo (see below log), but does not happen on other repos. it is repo specifc.

**To Reproduce**
1. clone the repo in bare mode
2. run **git fsck** on the cloned repo, no issue reported.
3. run **git lfs migrate import --include="*.zip"** (commits rewritten)
4. run **git fsck** again, see lots of tree error: 
**error in tree <SHA>: duplicateEntries: contains duplicate file entries**

```
ad_yhe13@JFS1SAPP0042 MINGW64 /c/Hoffer/temp
$ git clone ssh://xxxxxxxxxxxxxxxx/hadas_cmt_testprog.git --bare
Cloning into bare repository 'hadas_cmt_testprog.git'...
Welcome to Intel Hardened OS
remote: Enumerating objects: 90, done.
remote: Counting objects: 100% (90/90), done.
remote: Compressing objects: 100% (75/75), done.
remote: Total 6685 (delta 36), reused 36 (delta 13), pack-reused 6595
Receiving objects: 100% (6685/6685), 191.77 MiB | 11.97 MiB/s, done.
Resolving deltas: 100% (1791/1791), done.

ad_yhe13@JFS1SAPP0042 MINGW64 /c/Hoffer/temp
$ cd hadas_cmt_testprog.git

ad_yhe13@JFS1SAPP0042 MINGW64 /c/Hoffer/temp/hadas_cmt_testprog.git (BARE:dev)
$ git --version
git version 2.34.1.windows.1

ad_yhe13@JFS1SAPP0042 MINGW64 /c/Hoffer/temp/hadas_cmt_testprog.git (BARE:dev)
$  git lfs --version
git-lfs/3.0.2 (GitHub; windows amd64; go 1.17.2)

ad_yhe13@JFS1SAPP0042 MINGW64 /c/Hoffer/temp/hadas_cmt_testprog.git (BARE:dev)
$ git fsck
Checking object directories: 100% (256/256), done.
Checking objects: 100% (6685/6685), done.

ad_yhe13@JFS1SAPP0042 MINGW64 /c/Hoffer/temp/hadas_cmt_testprog.git (BARE:dev)
$ git lfs migrate import --include="*.zip"
migrate: Sorting commits: ..., done.
migrate: Rewriting commits: 100% (154/154), done.
[...]

PR #4849, opened 2022-01-31
halt migration when `.gitattributes` symbolic link encountered

chrisd8088:
As [discussed](https://github.com/git-lfs/git-lfs/issues/4796#issuecomment-1015782841) in issue #4796 and PR #4835, we add checks to the `git lfs migrate` command so that in all of its modes, if a symbolic link named `.gitattributes` is encountered, the migration halts with an error.  This will occur regardless of whether the migration would normally modify a `.gitattributes` file in the same location.

We add tests for the `import`, `export`, and `info` sub-commands which exercise this new behaviour, including with and without the `--fixup` option for the `import` and `info` sub-commands.

Note that this is an unexpected condition because `.gitattributes` tree entries should never be symbolic links; Git itself will complain about them with `"Too many levels of symbolic links"` warning messages.

In the manual page for the `git lfs migrate` command, we add notes to clarify this new behaviour, and also explain that any existing `.gitattributes` files with execute permissions will have those removed, as per PR #4835.

### DIFF
--- a/commands/command_migrate_info.go
+++ b/commands/command_migrate_info.go
@@ -164,15 +164,24 @@ func migrateInfoCommand(cmd *cobra.Command, args []string) {
 		},
 
 		TreePreCallbackFn: func(path string, t *gitobj.Tree) error {
-			if migrateFixup && path == "/" {
-				var err error
+			if migrateFixup {
+				if path == "/" {
+					var err error
 
-				fixups, err = gitattr.New(db, t)
-				if err != nil {
-					return err
+					fixups, err = gitattr.New(db, t)
+					if err != nil {
+						return err
+					}
 				}
 				return nil
 			}
+
+			for _, e := range t.Entries {
+				if e.Name == ".gitattributes" && e.IsLink() {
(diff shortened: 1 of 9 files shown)


commands/command_migrate_info.go
```diff
@@ -164,15 +164,24 @@ func migrateInfoCommand(cmd *cobra.Command, args []string) {
...
+						return err
+					}
 				}
 				return nil
 			}
+
+			for _, e := range t.Entries {
+				if e.Name == ".gitattributes" && e.IsLink() {
+					return errors.Errorf("migrate: %s", tr.Tr.Get("expected '.gitattributes' to be a file, got a symbolic link"))
+				}
```
bk2204, line 182: It's likely that `.gitattributes` will be early on in the tree.  If we've seen it, can we break out of the loop even if it's not a link?  That would help us perform better on repos with large trees.

bk2204 (commented):
This looks really good.  I had just one thought about a possible performance improvement we could make here.

commands/command_migrate_info.go
chrisd8088: Oh, good point; I should have pulled over that same logic from the `import` command's version.

bk2204 (approved):

PR #4856, opened 2022-02-04
Enable localization whenever we build binaries

bk2204:
When we build binaries, we'd like to make sure they're localized.  However, our Makefile, to be kind to external users, doesn't always force localization to be enabled.

Let's add a Makefile knob to do that and use it everywhere we build binaries.  There is an additional cleanup at the beginning of the series, and each commit has an explanatory commit message.


chrisd8088 (approved):
This looks fine to me, but I did want to note that there's a `fixup!` commit which I think was intended to be squashed into the first commit in the PR.

PR #4859, opened 2022-02-07
fix `git lfs fsck --objects A..B` handling and drop all left/right ref terminology

### LINKED ISSUE #4855: gif lfs fetch error even though fsck has been ok
**Describe the issue**

My repository has 2 branches, both starts with an empty commit X. The branches have a remote counterpart, which currently points to the same latest commits.

I have been doing some operations with the 2 branches, with them checked with `git lfs fsck X..<branch1>` and `git lfs fsck X..<branch2>` and reported ok.

However, when I recently ran `git lfs fetch --all`, it reports an error like:

```
fetch: 2618 object(s) found, done.
fetch: Fetching all references...
error transferring "d494e93a608893d986af6e133a71fc97c39e6fa8b50b9a7c080183f26395c29e": [0] remote missing object d494e93a608893d986af6e133a71fc97c39e6fa8b50b9a7c080183f26395c29e
error: failed to fetch some objects from '...'
```

I run `git fsck` in my local repository and no error is reported.

Does that mean my local or remote repository is corrupted? Is there a quick way to get the corresponding git blob object from the LFS object hash `d494e93a608893d986af6e133a71fc97c39e6fa8b50b9a7c080183f26395c29e` for me to further investigate?

**System environment**

Both the local and remote repos are in the same device. The OS is Windows 10 and using Git for Windows (`git version 2.35.1.windows.1` / `git-lfs/3.0.2 (GitHub; windows amd64; go 1.17.2)`).

**Output of `git lfs env`**

```
Endpoint=file:///D:/Users/Danny/Projects/test/myrepo/.git (auth=none)
LocalWorkingDir=D:\Users\Danny\Projects\test\myrepo2
LocalGitDir=D:\Users\Danny\Projects\test\myrepo2\.git
LocalGitStorageDir=D:\Users\Danny\Projects\test\myrepo2\.git
LocalMediaDir=D:\Users\Danny\Projects\test\myrepo2\.git\lfs\objects
LocalReferenceDirs=
TempDir=D:\Users\Danny\Projects\test\myrepo2\.git\lfs\tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=D:\Users\Danny\Projects\test\myrepo2\.git\lfs
AccessDownload=none
AccessUpload=none
git config filter.lfs.process = "git-lfs filter-process --skip"
git config filter.lfs.smudge = "git-lfs smudge --skip -- %f"
```

chrisd8088:
As described in https://github.com/git-lfs/git-lfs/issues/4855#issuecomment-1030769273, the `git lfs fsck --objects` command currently handles refs expressed on the command line in two-dot range notation backwards, so we resolve that problem, add tests to confirm it is working as intended, and also revise some internal variable and function names to remove the ambiguity of the terms "left" and "right" as they apply to Git ref names.

This PR is probably most easily reviewed in a commit-by-commit manner, although it should be feasible to review it as a single diff as well.

#### Reversed Ref Arguments

The unintended reversal of the ref names in a `git lfs fsck --objects A..B` command is due to the fact that we [pass](https://github.com/git-lfs/git-lfs/blob/12a957202906667343ef9fc62b312ee53353f266/commands/command_fsck.go#L151) the first ref (i.e., `A`, identifying everything to be excluded) as the `left` argument to the gitscanner.ScanRefRange(), and the second ref (i.e., `B`) as the `right` argument.  That would seem to be correct, except that in the `GitScanner` methods, `left` is meant to define all the commits to include, while `right` is meant to define the ones to exclude.

When these arguments are [passed](https://github.com/git-lfs/git-lfs/blob/12a957202906667343ef9fc62b312ee53353f266/lfs/gitscanner.go#L140) to `scanLeftRightToChan()` and then [passed](https://github.com/git-lfs/git-lfs/blob/12a957202906667343ef9fc62b312ee53353f266/lfs/gitscanner_refs.go#L99) to `scanRefsToChan()`, the `left` argument [becomes](https://github.com/git-lfs/git-lfs/blob/12a957202906667343ef9fc62b312ee53353f266/lfs/gitscanner_refs.go#L40) the `include` argument and the `right` one becomes the `exclude` one.

This terminology likely stems from the way refs are [delivered](https://github.com/git-lfs/git-lfs/blob/12a957202906667343ef9fc62b312ee53353f266/commands/command_pre_push.go#L25-L26) to the `git lfs pre-push` hook command, with the local one coming first (on the left), and the remote one coming second:

```
  <local ref> <local sha> <remote ref> <remote sha>
```

Since in that context we want to exclude all the objects already on the remote, `right` (remote) becomes the `exclude` ref, while `left` (local) becomes the `include` one.  However, outside of this context, `left` and `right` are ambiguous
terms.

For the `git lfs fsck --objects` command we can simply reverse the two arguments to `gitscanner.ScanRefRange()`.

For the `--pointers` command option, we similarly reverse the two arguments to `gitscanner.ScanRefRangeByTree()`, but we also need to reverse the way those arguments are then [passed](https://github.com/git-lfs/git-lfs/blob/12a957202906667343ef9fc62b312ee53353f266/lfs/gitscanner.go#L154) to `scanRefsByTree()` because at the moment they are flipped, which is why the `git lfs fsck --pointers` command performs as expected and hence why the existing test suite does not fail.

#### Expanded Tests

Another reason the test suite does not fail now although there are various tests of the default `git lfs fsck` option, which implicitly performs the actions of both the `--objects` and `--pointers` options, the ones which happen to test the `A..B` two-dot range notation do not set up any corrupt Git LFS object files, and therefore pass.  There are also no explicit tests of the `--objects` option.

So we first add a pair of tests dedicated to the `--objects` option, one checking for expected output when Git LFS object files are corrupt or missing, and one checking for no output when no Git LFS objects exist in the repository.

Then we expand the `"fsck operates on specified refs"` [test](https://github.com/git-lfs/git-lfs/blob/12a957202906667343ef9fc62b312ee53353f266/t/t-fsck.sh#L267-L2900 so that it explicitly checks both the `--pointers` and `--objects options`, checking the former against a commit with bad Git LFS pointer objects, and checking the latter against a different commit with an object file with incorrect contents, using `A..B` notation in each case.  This new second new check would fail without the changes made in this PR to resolve the ordering of the refs passed to the `gitscanner.ScanRefRange()` method by the `git lfs fsck` command.

#### Revised Terminology

We then reorder and rename various arguments, functions, and structure members, and rewrite various comments, in order to replace the terms `left` and `right` throughout the codebase wherever they pertain to Git ref names.

In PR #2308 many internal functions which take arguments that determine the set of Git objects to scan for Git LFS pointers already had those arguments renamed from `left` and `right` to `include` and `exclude`.  However, some of the `GitScanner` methods still use the older terminology, although they are used in many contexts other than processing pre-push Git hook input.  Therefore we try to complete this change in terms, using `include` and `exclude` wherever possible.

We also update some comments to better describe the functionality of various object scanning methods.

Since at least PR #2706, and indeed earlier, the local and remote refs which are part of the `RefUpdate` [structure ](https://github.com/git-lfs/git-lfs/blob/12a957202906667343ef9fc62b312ee53353f266/git/refs.go#L12-L13) have been referred to as the `left` and `right` refs, which demonstrates the origin of these terms in the development of the pre-push hook command.  In this context, `include` and `exclude` make less sense than `local` and `remote`, so we replace the older terms with `local` and `remote` wherever they appear.

#### Proceeding Despite Missing Files

Lastly, when the `git lfs fsck` command detects one or more missing or corrupt Git LFS object files, it attempts to move them into the `.git/lfs/bad` directory.  If a file is missing, though, the command immediately [exits](https://github.com/git-lfs/git-lfs/blob/12a957202906667343ef9fc62b312ee53353f266/commands/command_fsck.go#L116) with exit code 2, rather than the exit code 1 which it would otherwise produce.

Aside from the exit code difference, this behaviour also has the effect that a single missing object file may prevent all other corrupt ones from being moved into the `bad` directory.

We therefore ignore any "file not found" errors in this loop and attempt to move/rename all invalid object files, which means we then always exit with exit code 1 when any object files are corrupt or missing (unless an exceptional circumstance occurs, such as being unable to create the `.git/lfs/bad` directory or failing to move an object file due to permissions issues or something like that), and we update our tests to match this new behaviour.

Fixes #4855.
/cc @danny0838 as reporter.

### DIFF
--- a/t/t-fsck.sh
+++ b/t/t-fsck.sh
@@ -264,6 +264,72 @@ EOF
 )
 end_test
 
+setup_invalid_objects () {
+  git init $reponame
+  cd $reponame
+
+  # Create a commit with some files tracked by git-lfs
+  git lfs track *.dat
+  echo "test data" > a.dat
+  echo "test data 2" > b.dat
+  git add .gitattributes *.dat
+  git commit -m "first commit"
+
+  oid1=$(calc_oid_file a.dat)
+  oid2=$(calc_oid_file b.dat)
+  echo "CORRUPTION" >>".git/lfs/objects/${oid1:0:2}/${oid1:2:2}/$oid1"
+  rm ".git/lfs/objects/${oid2:0:2}/${oid2:2:2}/$oid2"
+}
+begin_test "fsck detects invalid objects"
+(
+  set -e
+
+  reponame="fsck-objects"
+  setup_invalid_objects
+
+  set +e
+  git lfs fsck >test.log 2>&1
+  RET=$?
+  set -e
+
+  [ "$RET" -eq 2 ]
+  [ $(grep -c 'objects: corruptObject: a.dat (.*) is corrupt' test.log) -eq 1 ]
+  [ $(grep -c 'objects: openError: b.dat (.*) could not be checked: .*' test.log) -eq 1 ]
+  [ $(grep -c 'objects: repair: moving corrupt objects to .*' test.log) -eq 1 ]
+
+  cd ..
+  rm -rf $reponame
+  setup_invalid_objects
+
+  set +e
+  git lfs fsck --objects >test.log 2>&1
+  RET=$?
+  set -e
+
+  [ "$RET" -eq 2 ]
+  [ $(grep -c 'objects: corruptObject: a.dat (.*) is corrupt' test.log) -eq 1 ]
+  [ $(grep -c 'objects: openError: b.dat (.*) could not be checked: .*' test.log) -eq 1 ]
+  [ $(grep -c 'objects: repair: moving corrupt objects to .*' test.log) -eq 1 ]
+)
+end_test
+
+begin_test "fsck does not detect invalid objects with no LFS objects"
+(
+  set -e
+
+  reponame="fsck-objects-none"
+  git init "$reponame"
+  cd "$reponame"
+
+  echo "# README" > README.md
+  git add README.md
+  git commit -m "Add README"
+
+  git lfs fsck
+  git lfs fsck --objects
+)
+end_test
+
 begin_test "fsck operates on specified refs"
 (
   set -e
--- a/t/t-fsck.sh
+++ b/t/t-fsck.sh
@@ -127,8 +127,8 @@ begin_test "fsck: outside git repository"
   set +e
   git lfs fsck 2>&1 > fsck.log
   res=$?
-
   set -e
+
   if [ "$res" = "0" ]; then
     echo "Passes because $GIT_LFS_TEST_DIR is unset."
     exit 0
@@ -193,6 +193,7 @@ begin_test "fsck detects invalid pointers with GIT_OBJECT_DIRECTORY"
   git init "$reponame-2"
   gitdir="$(lfstest-realpath "$reponame-2/.git")"
   GIT_WORK_TREE="$reponame-2" GIT_DIR="$gitdir" GIT_OBJECT_DIRECTORY="$objdir" git update-ref refs/heads/main "$head"
+
   set +e
   GIT_WORK_TREE="$reponame-2" GIT_DIR="$gitdir" GIT_OBJECT_DIRECTORY="$objdir" git lfs fsck --pointers >test.log 2>&1
   RET=$?
@@ -280,6 +281,7 @@ setup_invalid_objects () {
   echo "CORRUPTION" >>".git/lfs/objects/${oid1:0:2}/${oid1:2:2}/$oid1"
   rm ".git/lfs/objects/${oid2:0:2}/${oid2:2:2}/$oid2"
 }
+
 begin_test "fsck detects invalid objects"
 (
   set -e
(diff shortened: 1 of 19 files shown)


t/t-fsck.sh
```diff
@@ -264,6 +264,72 @@ EOF
...
+  echo "test data 2" > b.dat
+  git add .gitattributes *.dat
+  git commit -m "first commit"
+
+  oid1=$(calc_oid_file a.dat)
+  oid2=$(calc_oid_file b.dat)
+  echo "CORRUPTION" >>".git/lfs/objects/${oid1:0:2}/${oid1:2:2}/$oid1"
+  rm ".git/lfs/objects/${oid2:0:2}/${oid2:2:2}/$oid2"
+}
+begin_test "fsck detects invalid objects"
```
bk2204, line 283: Minor nit: maybe a blank line would be nice here.

bk2204 (approved):
Overall, looks good.  Thanks for fixing this.

PR #4863, opened 2022-02-12
Improve formatting

bbodenmiller:


bk2204 (approved):

This looks good to me, and it seems to be mostly consistent with what we do in other manual pages, so I think this should be a nice improvement.

I'll wait for CI to pass (I don't expect any problems there) and then merge this.  Thanks again.

PR #4878, opened 2022-02-15
lfs: add old hook content to the list of old hooks

### LINKED ISSUE #4876: Bitbucket and git lfs v3.1.1 Issue
Until yesterday's v3.1.1 release we installed and ran git lfs v3.0.2 in our Bitbucket Pipeline without issue. Since the v3.1.1 update I've seen the following error after the Bitbucket Pipeline installs git lfs:

```
curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash (successful)
apt-get --assume-yes install git-lfs (successful and installs v3.1.1)
git lfs install
+ git lfs install
Hook already exists: pre-push
	#!/bin/sh
	command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting .git/hooks/pre-push.\n"; exit 2; }
	git lfs pre-push "$@"
To resolve this, either:
  1: run `git lfs update --manual` for instructions on how to merge hooks.
  2: run `git lfs update --force` to overwrite your hook.
```

I didn't see anything obvious in the release notes that would lead me to the cause of this issue. Could you offer a suggestion on how to fix this and what changed between releases?

Thank you!

bk2204:
Our hook code is designed to automatically replace old hooks that it knows about with new ones.  However, in ed3decf7 ("use backticks around commands in messages", 2022-01-26), we made a change to the hook and accidentally forgot to include the old hook content.

Let's fix this by providing our old hook content, which contains a template, and processing each of the corresponding old hooks to replace the template.  This will cause unmodified hooks to be automatically updated, just like we expect.

Note that I opted not to provide any tests since we know the hook upgrade code already works and this is doing little more than just adding a single new entry.

Fixes #4876
/cc @chuckgregoryACES as reporter

### DIFF
--- a/lfs/hook.go
+++ b/lfs/hook.go
@@ -18,6 +18,7 @@ import (
 var (
 	// The basic hook which just calls 'git lfs TYPE'
 	hookBaseContent = "#!/bin/sh\ncommand -v git-lfs >/dev/null 2>&1 || { echo >&2 \"\\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting '.git/hooks/{{Command}}'.\\n\"; exit 2; }\ngit lfs {{Command}} \"$@\""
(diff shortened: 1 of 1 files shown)


asaneh:
#4193 #4890 #4895

PR #4880, opened 2022-02-16
Call out destructive command in README

btoll:
While it is the user's responsiblity to read the documentation and know
what impact a command will have that they choose to run on their
machine, I believe it would be helpful to draw more attention to the
destructive commands that are featured in the README that cause Git to
rewrite its commit objects.

Many users will only skim the README, and the migration command that
addresses a common use case should be called out more than it is for its
side effects, in my opinion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,9 @@ $ git commit -m "add psd"
 > $ git lfs migrate import --include="*.psd" --everything
 > ```
 >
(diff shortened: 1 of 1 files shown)


chrisd8088:
Hey, thanks for the patch and welcome to the Git LFS project!  This looks good to me, so I'm going to merge it in.  Thanks again!

chrisd8088 (approved):

PR #4896, opened 2022-03-02
Revert "lfs: add old hook content to the list of old hooks"

asaneh:
Reverts git-lfs/git-lfs#4878


chrisd8088:
This also doesn't look like a useful request, so I'm going to close it as well.

Please don't submit any more spurious requests or we'll have to block you from the project.

PR #4903, opened 2022-03-04
Drop vendoring

### LINKED ISSUE #4827: Drop vendoring
We get a lot of false positive reports about data that we vendor in our repositories, and it would be nice to avoid many of those.  Vendoring also makes our diffs needlessly large and there's no technical reason we need it in this project.

I'd like to propose that we stop vendoring and simply allow build processes to download as necessary as is standard with Go.  Distros can still package dependencies to avoid needing network access during builds, and we can retain a vendoring target for people to prepare repositories for building if they'd like to do so.

bk2204:
We currently vendor all dependencies in the repository.  This causes a bunch of problems, notably that sometimes these dependencies are out of date, we get security reports about them, and they needlessly bloat the repository.  There are some legitimate reasons to vendor dependencies, but none of them really apply to this project.

Since there are few benefits here and several downsides, let's remove the vendor directory altogether.  It has already been ignored, so if users would like to use our `make vendor` target, which remains, they can do so without problems.

There are two preparatory commits here which set us up for proper behavior at the end.  I should point out that this pull request removes over 358,000 more lines than it adds, which I believe is a compelling argument in favor of it.

Fixes #4827


chrisd8088 (approved):
Lovely, thank you!  This will save some headaches dealing with security reports that don't actually affect Git LFS, for sure.

Issue #4922, opened 2022-03-30
Pointer parsing doesn't match pointer specification


olsen232:
The spec at https://github.com/git-lfs/git-lfs/blob/main/docs/spec.md provides a list of things that tools should allow and not allow in pointer files. Git LFS itself doesn't quite follow the same rules.

Specifically, regarding the following rules:
 - Keys MUST only use the characters [a-z] [0-9] . -.
 - Lines of key/value pairs MUST be sorted alphabetically in ascending order (with the exception of version, which is always first).
And:
 - Tools that parse and regenerate pointer files MUST preserve keys that they don't know or care about.

Git LFS has instead the following behavior:
 - it doesn't allow any keys it doesn't recognize
 - it allows extensions (which is fine) but it allows them in any order (unlike the required keys, where it enforces the correct order)

There are no accidental logic errors in the pointer parsing code, all behavior is verified by the tests. For this reason, I haven't included any information to reproduce the issue, since the tests already cover it. However, perhaps Git LFS should follow the same spec?

olsen232:
I've taken the liberty of including a PR that will fix this - but of course the reviewer will first have to decide if they agree with this issue and think that Git LFS should more closely follow the spec described in https://github.com/git-lfs/git-lfs/blob/main/docs/spec.md

PR #4923, opened 2022-03-30
Match pointer parsing to pointer specification

olsen232:
Fixes https://github.com/git-lfs/git-lfs/issues/4922

From https://github.com/git-lfs/git-lfs/blob/main/docs/spec.md
- Keys MUST only use the characters [a-z] [0-9] . -.
- Lines of key/value pairs MUST be sorted alphabetically in ascending order
  (with the exception of version, which is always first).

And:
- Tools that parse and regenerate pointer files MUST preserve keys that
  they don't know or care about.

Git LFS itself doesn't exactly follow these rules:
- it doesn't allow any keys it doesn't recognize
- it allows extensions (which is fine) but it allows them to in the
  wrong order

This change makes Git LFS pointer parsing code match the specification:
- extra keys that are not recognized are allowed, and have no effect
  (except for making the pointer non-canonical)
- unless the extra keys use disallowed characters, in which case they
  cause pointer parsing to fail
- extensions are no longer allowed to be out of order since that
  would violate the alphabetically sorted rule

### DIFF
--- a/errors/types.go
+++ b/errors/types.go
@@ -135,17 +135,6 @@ func IsProtocolError(err error) bool {
 	return false
 }
 
-// If an error is abad pointer error of any type, returns NotAPointerError
-func StandardizeBadPointerError(err error) error {
-	if IsBadPointerKeyError(err) {
-		badErr := err.(badPointerKeyError)
-		if badErr.Expected == "version" {
-			return NewNotAPointerError(err)
-		}
-	}
-	return err
-}
-
 // IsDownloadDeclinedError indicates that the smudge operation should not download.
 // TODO: I don't really like using errors to control that flow, it should be refactored.
 func IsDownloadDeclinedError(err error) bool {
@@ -375,9 +364,6 @@ func NewPointerScanError(err error, treeishOid, path string) error {
 }
 
 type badPointerKeyError struct {
-	Expected string
-	Actual   string
-
 	*wrappedError
 }
 
@@ -385,9 +371,8 @@ func (e badPointerKeyError) BadPointerKeyError() bool {
 	return true
 }
 
-func NewBadPointerKeyError(expected, actual string) error {
-	err := Errorf(tr.Tr.Get("Expected key %s, got %s", expected, actual))
-	return badPointerKeyError{expected, actual, newWrappedError(err, tr.Tr.Get("pointer parsing"))}
+func NewBadPointerKeyError(err error) error {
+	return badPointerKeyError{newWrappedError(err, tr.Tr.Get("Pointer parsing error"))}
 }
 
 // Definitions for IsDownloadDeclinedError()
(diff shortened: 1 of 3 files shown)


bk2204:
Hey,

First of all, thanks for the patch.  I have a couple thoughts.

If we're regenerating a pointer file to serialize it, then we'll want to include the extra unknown fields.  As you point out, we're required to preserve them.  However, outside of that, we're not required to accept them, and I'd argue that we probably should not.  If some sort of critical new extension comes in, we'd need to adequately understand it in order to avoid corrupting the file or omitting it when someone runs `git add`.

Additionally, I'm concerned that making parsing stricter here may be considered a breaking change.  If people have broken pointer files, then we'll end up requiring them to rewrite their entire repository.  So I think as a practical matter, we'll need to simply make these pointers non-canonical (and therefore subject to fail under `git lfs fsck`), and then we can end up refusing them in version 4.  Otherwise, we'd be making a breaking change in a minor release, which wouldn't be allowed under semantic versioning.

Finally, it appears CI thinks there's a whitespace error somewhere here.

olsen232:
Hi Brian,

I think perhaps I misunderstood the spec slightly - it's possible that what will come out of this will be a slight clarification to the spec, which is okay too.

Here is the spec:
> Tools that parse and regenerate pointer files MUST preserve keys that they don't know or care about.

Here is one possible understanding of that spec, which was my first understanding:
> Tools that parse pointer files MUST ignore keys that they don't know or care about.
> Tools that regenerate pointer files MUST preserve keys that they don't know or care about.

And here is another, which I think is more in line with what you have said in this PR:
> Tools that parse pointer files (may? must?) fail if they encounter keys that they don't know about.
> Tools that regenerate pointer files MUST preserve keys that they don't know or care about.

Can you confirm which is your understanding of the spec?

bk2204:
I don't think either is particularly the case.  Tools that parse _and_ regenerate pointer files MUST _preserve_ keys that they don't know about.  That is, if we parse a pointer file and then write it somewhere, we must preserve the keys.  That is, `git lfs pointer` parses files, and then regenerates them, and it is obligated to preserve unknown keys.

The spec doesn't require us to _ignore_ or _smudge pointer files with_ keys that we don't know about.  If we parse a pointer file and never write that pointer file back out into the tree, then whether we preserve keys is irrelevant because we're not preserving the pointer.  In this case, we have to consider what possible data could be in the pointer entries and whether we might need to change our behaviour based on the (to us) extraneous data.  Thus, tools that _parse_ (but not regenerate) pointer files MAY accept or reject unknown keys as they see fit.  In our case, it is probably safer to reject them, since, due to the way that Git works, as soon as the file is modified on disk, Git will re-read and re-clean it, and the extraneous fields will be destroyed as Git LFS creates a fresh new pointer from the input and Git replaces the old one.

PR #4929, opened 2022-04-05
commands/update: add --append option

gpanders:
This option can be used to append the LFS hooks to existing git hooks.

I have pre-existing `post-checkout`, `post-merge`, and `post-commit` hooks, which means each time I want to initialize git-lfs in a repo I have to run `git lfs install --manual` and manually append the LFS hooks to each of my pre-existing hooks. This is a bit tedious, so I thought it might be handy if git-lfs had a way to do this automatically.

N.B. I was unable to run the tests locally. I get the following error:

```
No subtests run

Test Summary Report
-------------------
t-update.sh (Wstat: 0 Tests: 0 Failed: 0)
  Parse errors: No plan found in TAP output
```

I searched through the project documentation and online, but wasn't able to figure out the problem. As far as I can tell it's a Perl issue. I'm hoping the CI can run the tests and I can fix any issues found there until I can figure this out locally.

### DIFF
--- a/commands/command_clean.go
+++ b/commands/command_clean.go
@@ -92,7 +92,7 @@ func clean(gf *lfs.GitFilter, to io.Writer, from io.Reader, fileName string, fil
 func cleanCommand(cmd *cobra.Command, args []string) {
 	requireStdin(tr.Tr.Get("This command should be run by the Git 'clean' filter"))
 	setupRepository()
-	installHooks(false)
+	installHooks(false, false)
 
 	var fileName string
 	if len(args) > 0 {
--- a/commands/command_clone.go
+++ b/commands/command_clone.go
@@ -97,7 +97,7 @@ func cloneCommand(cmd *cobra.Command, args []string) {
 		// If --skip-repo wasn't given, install repo-level hooks while
 		// we're still in the checkout directory.
 
-		if err := installHooks(false); err != nil {
+		if err := installHooks(false, false); err != nil {
 			ExitWithError(err)
 		}
 	}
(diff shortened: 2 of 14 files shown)


bk2204:
Hey,

First of all, welcome to Git LFS.  While I think you've probably noticed that CI has failed, and I'll let you look into that at your leisure, let me share some thoughts about this approach.

The hook you didn't mention using is the `pre-push` hook.  Unlike the other three, that hook requires standard input, which means that just appending the Git LFS hook isn't useful, since the earlier part of the hook script will already have read all of the input, and then the `pre-push` hook will get no data and not push anything.

I think, therefore, that we need to have a decision about what to do in that case with this option, since it can't functionally work there.  There are three possible approaches:

1. Make `--append` not work on the `pre-push` hook and clearly document that, and pretend as if `--append` weren't passed for `pre-push`.  (We need an update to the manual page here anyway.)  I think this is probably the most reasonable approach, but it does require some explanation.
2. Make `--append` not work on the `pre-push` hook and document that, and do nothing.  I think this is unsuitable because it runs the risk of people not having their hook installed and therefore not pushing their data, which could well lead to data loss.
3. Decide not to implement `--append` at all.

There may be other approaches here, and I'm happy to hear your thoughts, as well as those of other folks on @git-lfs/core on the best approach.  I think @chrisd8088 has recently run the tests on Windows, so if you're on that platform and having trouble, hopefully he can give you some pointers about how to make the testsuite run (I remember it being very finicky.)

I will just point out that there's a proposal on the Git list to allow using multiple hooks via the config.  That will probably come down the line and then we won't need this because you can simply spawn your custom hook _and_ the Git LFS hook and everything will magically work.  However, that may be farther away than you'd like, so I think this PR is still worth considering.

You're welcome to try to track down the failure here and see what it is, or if you'd prefer to discuss the design more before proceeding, that's fine too.  We're not in any immediate rush to do a feature release.  I do appreciate you taking time to send in a patch, though.

PR #4935, opened 2022-04-11
lfs: don't write hooks when they haven't changed

### LINKED ISSUE #4865: Avoid overwriting unchanged hooks
**Describe the issue**
We came across a timestamp issue regarding git hooks. We have a custom set-up for our git repository, where we changed `.git/hooks` to be a symlink pointing to a tracked folder inside the repository. This is because we want to track the differences to the hooks in the repo. When `filter-process` is triggered, git-lfs is trying to overwrite the hooks. The files are identical, so git-lfs is overwriting them with the same contents, resulting in a timestamp change. This is picked up as a difference by commands like `git diff-files`. Normally, this is not a problem because we update the index, however, I've seen cases of the `filter-process` being triggered right after `git update-index --refresh`, or after `git status`. So we noticed that we were updating the index, then immediately saw differences that were not expected, because the timestamp of the hook files was changed.

I tried to reproduce this [here](https://github.com/RuxandraM/lfs_hooks) 

Any thoughts? Is it possible to skip writing the hook files if they are identical? I understand git-lfs checks for the contents anyway (for not overwriting custom hooks).

**System environment**
Running on MacOS, git-lfs/2.13.3, git version 2.23.0

bk2204:
Some people store their hooks inside their repository.  This isn't a secure approach, but in some cases it can be acceptable if all users are trusted.

In such a case, if Git LFS always rewrites a hook even if it hasn't changed, it can result in the file being marked as modified by `git status`, since `git status` can invoke Git LFS, which in turn invokes Git LFS, which updates the hooks.  Thus, the hooks will always appear modified.

Let's solve this by not writing the hooks if they haven't changed.  We don't include tests here because the only way to determine if changes have occurred is the mtime, and that's difficult to effectively test in the testsuite in a portable way without introducing sleeps, which would make our tests slow.

Fixes #4865

### DIFF
--- a/lfs/hook.go
+++ b/lfs/hook.go
@@ -107,12 +107,12 @@ func (h *Hook) write() error {
 // the member variable `Upgradeables`. It halts and returns any errors as they
 // arise.
 func (h *Hook) Upgrade() error {
-	match, err := h.matchesCurrent()
+	upgradable, match, err := h.matchesCurrent()
 	if err != nil {
 		return err
 	}
 
-	if !match {
+	if !upgradable || match {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


chrisd8088 (approved):
Very nice, thanks!

PR #4936, opened 2022-04-11
script/upload: correct RHEL 8 package repo

### LINKED ISSUE #4934: Download link for RHEL8/CentOS version not functioning for 3.1.1 and 3.1.2
**Describe the issue**
When trying to download the git-lfs rpm's for RHEL8/CentOS8 through the link (https://packagecloud.io/github/git-lfs/packages/el/8/git-lfs-3.1.2-1.el8.x86_64.rpm/download) an error is displayed. Package is not available. 

**System environment**
Any web browser.

**Additional context**
The link in versions older than 3.1.1 do work.

bk2204:
Because CentOS 8 is now EOL, we've switched to Rocky Linux for building RHEL 8-compatible packages.  However, those packages don't end up containing the ".el8" segment, so the link is broken.  Fix this by removing that segment from the URL.

Fixes #4934


chrisd8088 (approved):
👍 Thanks!

PR #4938, opened 2022-04-11
Makefile: remove legacy trimpath code

bk2204:
Before Go 1.13, it was possible to trim paths with GC flags and ASM flags.  In Go 1.13, it became possible to simply specify -trimpath, at which point all the old behaviour became unnecessary.

Since we require Go 1.13 in our code due to the use of the errors package, let's simplify our code by always using the -trimpath flag and omitting the old forms.  The old forms additionally were broken when using a path with a space, which could happen on Windows if the current user's username contained a space (which is not uncommon).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,17 +39,11 @@ endif
 LD_FLAGS = $(BUILTIN_LD_FLAGS) $(EXTRA_LD_FLAGS)
 
 # BUILTIN_GC_FLAGS are the internal flags used to pass compiler.
-BUILTIN_GC_FLAGS ?= all=-trimpath="$$HOME"
+BUILTIN_GC_FLAGS =
 # EXTRA_GC_FLAGS are the caller-provided flags to pass to the compiler.
 EXTRA_GC_FLAGS =
 # GC_FLAGS are the union of the above two BUILTIN_GC_FLAGS and EXTRA_GC_FLAGS.
 GC_FLAGS = $(BUILTIN_GC_FLAGS) $(EXTRA_GC_FLAGS)
-
-ASM_FLAGS ?= all=-trimpath="$$HOME"
-
-# TRIMPATH contains arguments to be passed to go to strip paths on Go 1.13 and
-# newer.
-TRIMPATH ?= $(shell [ "$$($(GO) version | awk '{print $$3}' | sed -e 's/^[^.]*\.//;s/\..*$$//;')" -ge 13 ] && echo -trimpath)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -39,17 +39,13 @@ endif
 LD_FLAGS = $(BUILTIN_LD_FLAGS) $(EXTRA_LD_FLAGS)
 
 # BUILTIN_GC_FLAGS are the internal flags used to pass compiler.
-BUILTIN_GC_FLAGS ?= all=-trimpath="$$HOME"
+BUILTIN_GC_FLAGS ?=
```
chrisd8088, line 42: Perhaps this could go back to the `BUILTIN_GC_FLAGS =` form that it was in prior to #3443?  (Or even be removed entirely, but it seems like it was added intentionally in that form in 2235198650d8dfadded525cc308ae8dfa786da52 of PR #3144.)

Makefile
```diff
@@ -39,17 +39,13 @@ endif
...
 EXTRA_GC_FLAGS =
 # GC_FLAGS are the union of the above two BUILTIN_GC_FLAGS and EXTRA_GC_FLAGS.
 GC_FLAGS = $(BUILTIN_GC_FLAGS) $(EXTRA_GC_FLAGS)
 
-ASM_FLAGS ?= all=-trimpath="$$HOME"
-
-# TRIMPATH contains arguments to be passed to go to strip paths on Go 1.13 and
-# newer.
-TRIMPATH ?= $(shell [ "$$($(GO) version | awk '{print $$3}' | sed -e 's/^[^.]*\.//;s/\..*$$//;')" -ge 13 ] && echo -trimpath)
+ASM_FLAGS ?=
```
chrisd8088, line 48: Could this be removed completely, since it seems like there's no need for it anymore?  It looks like it was introduced in 65b65d074de753f0166ce13cd6dba0695dcdd154 of #3443 just to support this use case, so maybe it could be dropped.

Makefile
```diff
@@ -186,7 +182,7 @@ BUILD = GOOS=$(1) GOARCH=$(2) \
 	-ldflags="$(LD_FLAGS)" \
 	-gcflags="$(GC_FLAGS)" \
 	-asmflags="$(ASM_FLAGS)" \
```
chrisd8088, line 184: As above, maybe this `-asmflags=...` like could dropped now, since it seems like it was added along with `ASM_FLAGS` for this one purpose.

chrisd8088 (approved):
I wondered about some additional simplifications, but otherwise this looks all good!

PR #4946, opened 2022-04-15
Update Windows signing certificate SHA hash in `Makefile`

chrisd8088:
The Git LFS signing certificate for Windows binaries has changed, so we update the default signing certificate SHA-1 hash in the `Makefile`.

(The last update was in 2019 in PR #3623.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ TAR_XFORM_CMD ?= $(shell tar --version | grep -q 'GNU tar' && echo 's')
 
 # CERT_SHA1 is the SHA-1 hash of the Windows code-signing cert to use.  The
 # actual signature is made with SHA-256.
-CERT_SHA1 ?= 824455beeb23fe270e756ca04ec8e902d19c62aa
+CERT_SHA1 ?= 27ea8f81ce920bccd2174e3b272f2ba247605be6
 
 # CERT_FILE is the PKCS#12 file holding the certificate.
 CERT_FILE ?=


znull:
Judging from 3b7c60f782c61a6568956aa93cb9ff756a72e4c3 and 5cb868bb418e2ae89fb39df6cfc8b13594ce5e97, we could just drop support for this instead? Referencing a cert by sha1 seems to be a holdover from when a custom VM was used to build release packages, and I'm getting the feeling that's not something needed anymore.

chrisd8088:
That's entirely possible; there's definitely a chance we don't need this.

bk2204 (approved):
It is possible we don't need this, but since the alternative is that our release fails and I'd like for at least our next release to be rather boring considering the problems we had last time, let's go ahead with this.  We can investigate further with our test repo and see if it's not needed in the future, and if so, remove it.

Also, I'm sure you already thought of this, but since I didn't see the cert myself, this is specifically the certificate fingerprint, and not a hash of the PKCS#12 file, so you'd need to query OpenSSL for this directly if you didn't.

bk2204:
Actually, OpenSSL says the certificate SHA-1 should be `27:EA:8F:81:CE:92:0B:CC:D2:17:4E:3B:27:2F:2B:A2:47:60:5B:E6`.

The thing I did here is this:

```
$ openssl pkcs12 -info -in codesign.pfx -out $TMPDIR/foo.pem -provider default -provider legacy -nokeys
$ openssl x509 -text -in $TMPDIR/foo.pem -fingerprint
```

I think the hash you provided is that of the PKCS#12 file, which I don't think is what Windows wants here.

chrisd8088:
Thanks for the correction, @bk2204 -- I think I've updated this again now with the expected hash of the fingerprint.

Issue #4751, opened 2021-11-25
git lfs migrate import include filter issue starting with version 3.0.1


rolandas-rimkus:
**Describe the bug**
Starting with version 3.0.1 of Git LFS running **git lfs migrate import** with a include pattern like *.extension started affecting directory names that also end with the same pattern. For example if *.model is used all files under a directory VS.Project.Parse.Model would be added under Git LFS. This was not the case in version 2.13.3.

The documentation for [gitattributes ](https://git-scm.com/docs/gitattributes) seems to indicate that using *.extension should not affect directories so it seems like a bug was introduced in the version 3.0.1 of Git LFS.

**To Reproduce**
1. Install Git version 2.33.1 or newer from: https://github.com/git-for-windows/git/releases
2. Run the following commands under a Windows Command Prompt
```
git init D:\GitLfsBug
mkdir D:\GitLfsBug\Data
mkdir D:\GitLfsBug\VS.Project.Parse.Model
echo Test > D:\GitLfsBug\Data\aaa.model
echo Test > D:\GitLfsBug\Data\aaa.txt
echo Test > D:\GitLfsBug\VS.Project.Parse.Model\bbb.txt
echo Test > D:\GitLfsBug\VS.Project.Parse.Model\ccc.txt
git -C D:\GitLfsBug add --all .
git -C D:\GitLfsBug commit -m "Test" --all
git -C D:\GitLfsBug lfs install
git -C D:\GitLfsBug lfs migrate import --yes --include="*.model"
git -C D:\GitLfsBug lfs ls-files --size --long
```
3. Note down that the files tracked under LFS are: Data\aaa.model, VS.Project.Parse.Model\bbb.txt and VS.Project.Parse.Model\ccc.txt
4. Download Git LFS version 2.13.3 from https://github.com/git-lfs/git-lfs/releases/download/v2.13.3/git-lfs-windows-amd64-v2.13.3.zip
5. Apply the git-lfs.exe file on top of mingw64\bin\git-lfs.exe in the install location of Git
6. Repeat step 2
7. Note down that the only file tracked under LFS is: Data\aaa.model

**Expected behavior**
Directories that match the file filter passed into git lfs migrate import should not be tracked under Git LFS as was the case in versions up to 3.0.1.

**System environment**
Windows 10 v1909

**Output of `git lfs env`**
git-lfs/3.0.1 (GitHub; windows amd64; go 1.17.1)
git version 2.33.1.windows.1

[...]

PR #4951, opened 2022-04-19
Use `gitignore`-style path matching for additional commands

### LINKED ISSUE #4945: lfs.fetchinclude no longer works with directories
**Describe the bug**

Using `lfs.fetchinclude` with a directory no longer works with the 3.x releases of git-lfs. (It worked fine with the 2.x releases)

**To Reproduce**
1. `git clone https://github.com/nuxi/git-lfs-bug.git`
2. Check the contents of `.lfsconfig` and see that `lfs.fetchinclude = foo` is set.
3. Check to see if `foo/should_be_a_binary.bin` is the pointer or a binary.
4. Check to see if `bar/should_be_a_pointer.bin` is the pointer or a binary.

**Expected behavior**

I think the filenames are obvious enough, but with git-lfs 3.x, the `foo/should_be_a_binary.bin` file is still a pointer. Under git-lfs 2.x, the file is the binary.

**System environment**

Debian Sid

**Output of `git lfs env`**

```
git-lfs/3.0.2 (GitHub; linux amd64; go 1.17.2)
git version 2.35.1

Endpoint=https://github.com/nuxi/git-lfs-bug.git/info/lfs (auth=none)
LocalWorkingDir=/home/nuxi/src/git-lfs-bug/lfs3-main
LocalGitDir=/home/nuxi/src/git-lfs-bug/lfs3-main/.git
LocalGitStorageDir=/home/nuxi/src/git-lfs-bug/lfs3-main/.git
LocalMediaDir=/home/nuxi/src/git-lfs-bug/lfs3-main/.git/lfs/objects
LocalReferenceDirs=
TempDir=/home/nuxi/src/git-lfs-bug/lfs3-main/.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/home/nuxi/src/git-lfs-bug/lfs3-main/.git/lfs
AccessDownload=none
AccessUpload=none
FetchInclude=foo
```

**Additional context**

The same repo, but with the config option pointing directly at the exact binary file is fine. You can try this with `git clone -b exact https://github.com/nuxi/git-lfs-bug.git`

Support for specifying a directory is documented in the manual as the very first example - https://github.com/git-lfs/git-lfs/blob/v3.1.2/docs/man/git-lfs-fetch.1.ronn#examples

chrisd8088:
The `lfs.fetchinclude` and `lfs.fetchexclude` Git configuration options, if set, are used to control the action of a number of Git LFS commands.  Since PR #4556, the `git lfs clone`, `git lfs fetch`, and `git lfs pull` commands have strictly applied `gitignore(5)`-style matching rules to these configuration options.

However, other commands including `git lfs filter-process` and `git lfs smudge` now apply `gitattributes(5)`-style matching rules to these same configuration options, leading to confusion.

We therefore revise all remaining uses of these configuration options to also use `gitignore`-style matching rules.

We also add new tests for the `git lfs filter-process` and `git lfs fsck` commands and adjust or expand existing tests for the `git lfs prune` and `git lfs smudge` commands in order to confirm that `gitignore`-style matching is used for all of them.  These new and updated tests fail if `gitattributes`-style matching is used instead.

As well, we update the existing manual page documentation on how these `lfs.fetch*` filter configuration options function, and add additional notes on their operation with regard to several other commands, and we fix a few related formatting issues and missing references.
    
(Note that the `git lfs migrate` command does not require any changes because it does not read the `lfs.fetch*` configuration options.  Instead, it [supplies](https://github.com/git-lfs/git-lfs/blob/0df26c176a2f345d547b77f46ec2e18f40da6cf7/commands/command_migrate.go#L312) a `false` value for the `useFetchOptions` flag to the `determineIncludeExcludePaths()` function, so any `lfs.fetch*` configuration values are ignored.  This is significant because `git lfs migrate` deliberately uses `gitattributes`-style matching for any path patterns supplied via its `-I`/`-X` command-line arguments, unlike all other commands that accept `-I`/`-X` arguments as overrides for the `lfs.fetch*` configuration options.  See #4758 and #4751 for more details.)

Fixes #4945.
/cc @nuxi as reporter.

### DIFF
--- a/commands/command_filter_process.go
+++ b/commands/command_filter_process.go
@@ -61,7 +61,7 @@ func filterCommand(cmd *cobra.Command, args []string) {
 	}
 
 	skip := filterSmudgeSkip || cfg.Os.Bool("GIT_LFS_SKIP_SMUDGE", false)
-	filter := filepathfilter.New(cfg.FetchIncludePaths(), cfg.FetchExcludePaths(), filepathfilter.GitAttributes)
+	filter := filepathfilter.New(cfg.FetchIncludePaths(), cfg.FetchExcludePaths(), filepathfilter.GitIgnore)
 
 	ptrs := make(map[string]*lfs.Pointer)
 
--- a/commands/command_fsck.go
+++ b/commands/command_fsck.go
@@ -144,7 +144,7 @@ func doFsckObjects(include, exclude string, useIndex bool) []string {
 	// objects), the "missing" ones will fail the fsck.
 	//
 	// Attach a filepathfilter to avoid _only_ the excluded paths.
-	gitscanner.Filter = filepathfilter.New(nil, cfg.FetchExcludePaths(), filepathfilter.GitAttributes)
+	gitscanner.Filter = filepathfilter.New(nil, cfg.FetchExcludePaths(), filepathfilter.GitIgnore)
 
 	if exclude == "" {
 		if err := gitscanner.ScanRef(include, nil); err != nil {
--- a/commands/command_prune.go
+++ b/commands/command_prune.go
@@ -98,7 +98,7 @@ func prune(fetchPruneConfig lfs.FetchPruneConfig, verifyRemote, dryRun, verbose 
 	retainChan := make(chan string, 100)
 
 	gitscanner := lfs.NewGitScanner(cfg, nil)
-	gitscanner.Filter = filepathfilter.New(nil, cfg.FetchExcludePaths(), filepathfilter.GitAttributes)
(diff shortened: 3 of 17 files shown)


bk2204 (approved):
Yeah, having a consistent interpretation here seems like the right thing to do.

PR #4954, opened 2022-04-19
Build test helper commands with `.exe` file extension on Windows

chrisd8088:
On Windows, it is convenient to have the `.exe` file extension added to our compiled test helpers when running tests manually, so we use the `X` variable to do so in the same manner as in the primary `Makefile`.

Also, we add a `test-commands` target in `t/Makefile` which is convenient when rebuilding just the test helpers.


bk2204 (approved):
I've also found myself wanting a target to rebuild just the test commands, so thanks for adding this.

PR #4971, opened 2022-04-27
Multiple hash support

### LINKED ISSUE #4937: Include a larger set of hashes with release assets
Right now, we provide only SHA-256 hashes for our releases.  That's convenient for most people and fine from a security perspective, but some distributors, such as NetBSD, use different hashes for release assets (in their case, SHA-512 and BLAKE2s).  For the convenience of distributors, it would be nice to provide other hashes as well.

We can use the BSD format, which in the tooling is usually called `--tag`, which includes both the hash name and the file name, so it's possible to distinguish between hashes, like so:

```
SHA256 (/dev/null) = e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
SHA3-256 (/dev/null) = a7ffc6f8bf1ed76651c14756a061d662f580ff4de43b49fa82d80a4b80f8434a
```

Using this format would let us provide all of these hashes in one additional file to avoid needing many different files for different hashes.  This format can be verified by either the Perl tools or the GNU tools, and we can generate them with a Ruby script, since OpenSSL, which is used by Ruby, contains all the hash algorithms we care about (SHA-2, SHA-3, and BLAKE2).  It's also generally verifiable on BSD machines, which makes it substantially more portable and convenient for users.  We can add and remove hashes there as we see fit.

We should, for backward compatibility, keep the `sha256sums.asc` file as well.

bk2204:
Right now, we provide signed SHA-256 hashes for our releases.  This is fine and sufficient, and also cryptographically secure.  However, many distributors use other algorithms, and it would be convenient if we could provide easy access to those hashes as well.  For example, NetBSD uses SHA-512 and BLAKE2s.

Let's add an additional file, hashes.asc, which contains a general set of hashes in the BSD format. The advantage of the BSD format over the traditional GNU format is that it includes the hash algorithm, which allows us to distinguish between hashes of the same length, such as SHA-256, SHA-512/256, and SHA3-256.  It is generated by shasum, sha*sum, sha3sum, and b2sum with the --tag format, and all of these programs accept it for verification with no problems.

Using the BSD format means that we need only provide one additional file with all the additional algorithms.  There is therefore no need to add multiple new files, and if we desire to add additional algorithms in the future, that's easily done without modification.

I don't love the name `hashes` and would prefer something different, but I am, as usual, bad at naming things, so suggestions are welcome.

Fixes #4937

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,11 @@ the file you want like so:
 $ gpg -d sha256sums.asc | grep git-lfs-linux-amd64-v2.10.0.tar.gz | shasum -a 256 -c
 ```
 
+For the convenience of distributors, we also provide a wider variety of signed
+hashes in the `hashes.asc` file.  Those hashes are in the tagged BSD format, but
+can be verified with Perl's `shasum` or the GNU hash utilities, just like the
+ones in `sha256sums.asc`.
+
 ## Example Usage
 
 To begin using Git LFS within a Git repository that is not already configured
--- a/script/hash-files
+++ b/script/hash-files
@@ -1,0 +1,59 @@
+#!/usr/bin/env ruby
+
+require "openssl"
+
+# This maps the OpenSSL name to the name used in the output file.
+# The order used is the order they should appear in the output file.
+DIGESTS = {
+  'BLAKE2b512' => 'BLAKE2b',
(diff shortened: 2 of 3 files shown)


chrisd8088 (approved):
LGTM!

PR #4980, opened 2022-04-29
Tarballs, not tarbombs

### LINKED ISSUE #4977: Make tarballs not tarbombs
Our tarballs and zip files should not extract to the current directory.  I seem to recall that we had this working at some point, but it may have gotten broken.  Regardless, we should fix it.

bk2204:
Right now, extracting one of our archives (except for the source archives) causes all files to be extracted in the current directory. This is commonly referred to as a tarbomb, and is generally unwelcome.

Instead, let's create a top-level directory in each of our archives named the same way as we currently name our source tarballs: the string "git-lfs-" and then a version number without the initial "v".  This is the same algorithm GitHub would use to create its autogenerated archives.

Because we can no longer use zip's -j option to strip off path components (since we now want to preserve at least some of them), using zip becomes tricky.  Instead, let's use the BSD (libarchive) tar, which is natively available on Windows, macOS, and FreeBSD as the system tar implementation, and which can be easily installed on all Linux distros under the name "bsdtar".  This program supports tar files, zip files, and a wide variety of other formats that we don't need here and is ideal for this usage.

In addition, to make testing these operations easier, let's move several commands into variables so they can be overridden.  Finally, we change our tar archives to use the pax format, which is already used by Git and is substantially more portable than GNU tar's default format.

This series is best reviewed commit by commit.

Fixes #4977

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -104,7 +104,7 @@ jobs:
     - uses: actions/setup-go@v2
       with:
         go-version: ${{ matrix.go }}
-    - run: sudo apt-get update && sudo apt-get -y install gettext
+    - run: sudo apt-get update && sudo apt-get -y install gettext libarchive-tools
       env:
           DEBIAN_FRONTEND: noninteractive
     - uses: actions/download-artifact@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
         go-version: ${{ matrix.go }}
     - run: brew install gettext
       if: ${{ startsWith(matrix.os, 'macos-') }}
-    - run: sudo apt-get update && sudo apt-get -y install gettext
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -97,14 +97,14 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        go: ['1.18.x']
+        go: ['1.17.x']
```
chrisd8088, line 100: Should we keep this on `1.18.x`?

chrisd8088 (approved):
I just had one question about the release workflow going back to Go 1.17; otherwise this looked great, thank you!

.github/workflows/release.yml
bk2204: Yeah, I rebased and there was a conflict which I apparently misresolved.  Will fix.

PR #4987, opened 2022-05-04
note deprecation of NTLM

vtbassmatt:
note deprecation of NTLM and point to Azure DevOps Server recommendation


bk2204 (approved):
This looks good.  Thanks for the patch.

bk2204:
Yup.  Our CI takes a long time, so I went and did other, productive things while waiting for it, but it's green now.

PR #4992, opened 2022-05-05
workflows: use ronn-ng

bk2204:
Recently, we changed to use the windows-2019 target because ronn fails to build on newer versions.  Unfortunately, it also fails to build on that version due to updates in GitHub Actions.

Since ronn itself is dead upstream, let's switch to ronn-ng, just like Debian has done.  This provides an identically named binary but is a supported codebase that is more likely to work now and in the future.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       if: ${{ startsWith(matrix.os, 'ubuntu-') }}
       env:
           DEBIAN_FRONTEND: noninteractive
-    - run: gem install ronn
+    - run: gem install ronn-ng
     - run: script/cibuild
       env:
           FORCE_LOCALIZE: true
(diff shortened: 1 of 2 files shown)


chrisd8088:
Sweet!  🚀  I'm delighted this exists and very happy you spotted it, as it should save us a lot of effort!!

chrisd8088 (approved):
LGTM.  I wonder if it will require `ruby/setup-ruby@v1` on the Windows 2022 runners or if it will work with the older, deprecated `actions/setup-ruby@v1` we're using now.  The latter would mean we could probably switch to the `windows-latest` runners again, per #4883.  If we need `ruby/setup-ruby@v1`, though, we may still run into problems, IIRC.

Regardless, though, this is definitely a strict improvement over the status quo.

bk2204:
I'll open an issue to track that and investigate the best way forward there.  Since things aren't working now, I'd like to go with the smallest fix that will unbreak things, and then we can iteratively improve on things in the future.

PR #4996, opened 2022-05-06
lfshttp: fix invalid Accept header syntax

rcoup:
git-lfs uses the same constant value for the `Content-Type` and `Accept` request HTTP headers:

```
POST /objects/batch/
Accept: application/vnd.git-lfs+json; charset=utf-8
Content-Type: application/vnd.git-lfs+json; charset=utf-8
...
```

But `charset` [isn't valid syntax for `Accept`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept). This change splits out the constant to fix this:

```
POST /objects/batch/
Accept: application/vnd.git-lfs+json
Content-Type: application/vnd.git-lfs+json; charset=utf-8
...
```

### DIFF
--- a/lfshttp/client.go
+++ b/lfshttp/client.go
@@ -27,7 +27,8 @@ import (
 	"golang.org/x/net/http2"
 )
 
-const MediaType = "application/vnd.git-lfs+json; charset=utf-8"
+const MediaType = "application/vnd.git-lfs+json"
+const RequestContentType = MediaType + "; charset=utf-8"
 
 var (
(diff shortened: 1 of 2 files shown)


chrisd8088:
Hey, thanks for the patch and welcome to Git LFS!  This seems correct and reasonable to me, so I'll merge it once the CI jobs turn green.  Thanks again for the PR!

chrisd8088 (approved):

PR #5006, opened 2022-05-11
Add --json output for git lfs track

### LINKED ISSUE #4921: [Feature Request] Separate tracked listing versus excluded
Two projects I use have been developed to parse `git-lfs` output, such as `git-lfs tracked` and they're unable to differentiate the output of tracked items versus excluded due to the uncertainty of localization.

Perhaps the tracked/excluded could be listed separately, or a guaranteed output delimiter could be used? A double-newline, perhaps?

bk2204:
We want a machine readable output for git lfs track, so add a `--json` flag to add this.

Fixes #4921

### DIFF
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"bufio"
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -29,6 +30,7 @@ var (
 	trackNoModifyAttrsFlag  bool
 	trackNoExcludedFlag     bool
 	trackFilenameFlag       bool
+	trackJSONFlag           bool
 )
 
 func trackCommand(cmd *cobra.Command, args []string) {
@@ -42,6 +44,10 @@ func trackCommand(cmd *cobra.Command, args []string) {
 	if len(args) == 0 {
 		listPatterns()
 		return
+	}
+
+	if trackJSONFlag {
+		Exit(tr.Tr.Get("--json option can't be combined with arguments"))
 	}
 
 	mp := gitattr.NewMacroProcessor()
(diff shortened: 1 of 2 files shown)


t/t-track.sh
```diff
@@ -711,3 +711,26 @@ begin_test "track: escaped glob pattern with spaces in .gitattributes"
...
+  set -e
+
+  reponame="track-json"
+  git init "$reponame"
+  cd "$reponame"
+
+  git lfs track '*.dat'
+  git lfs track --lockable '*.bin'
+  echo 'a.dat !filter' >>.gitattributes
+  git add .gitattributes
```
chrisd8088, line 726: Out of interest, is this necessary?  I tried a `git lfs track` on an untracked `.gitattributes` and it worked, so it may not need to be added-but-not-committed here.

commands/command_track.go
```diff
@@ -337,5 +369,6 @@ func init() {
 		cmd.Flags().BoolVarP(&trackNoModifyAttrsFlag, "no-modify-attrs", "", false, "skip modifying .gitattributes file")
 		cmd.Flags().BoolVarP(&trackNoExcludedFlag, "no-excluded", "", false, "skip listing excluded paths")
 		cmd.Flags().BoolVarP(&trackFilenameFlag, "filename", "", false, "treat this pattern as a literal filename")
+		cmd.Flags().BoolVarP(&trackJSONFlag, "json", "", false, "print output in json")
```
chrisd8088, line 372: Very minor nit since these strings are unused ... maybe "print output in JSON" here?

commands/command_track.go
```diff
@@ -221,8 +227,34 @@ ArgsLoop:
...
+		}{Patterns: make([]PatternData, 0, len(knownPatterns))}
+		for _, p := range knownPatterns {
+			patterns.Patterns = append(patterns.Patterns, PatternData{
+				Pattern:  p.Path,
+				Source:   p.Source.String(),
+				Tracked:  p.Tracked,
+				Lockable: p.Lockable,
+			})
+		}
+		err := json.NewEncoder(os.Stdout).Encode(patterns)
```
chrisd8088, line 251: Would it be possible to use the `SetIndent("", "  ")` [function](https://pkg.go.dev/encoding/json#Encoder.SetIndent) here to make the output format have a bit of indenting?

chrisd8088 (approved):
Some minor questions only; this is great, thank you!

t/t-track.sh
bk2204: Technically, no.  I can remove it.

commands/command_track.go
bk2204: Yup.  That was a copy-paste problem.

commands/command_track.go
bk2204: Sure.  I doubt this is going to be very interesting for human-readable output, but it does at least make the tests nicer, so I figure that's an improvement.

PR #5014, opened 2022-05-19
Update INSTALLING.md to fix ubuntu derivative command and allow easy install for some distros

ry167:
- Fix for duplicate `sudo` which causes it to not correctly pass environment variables
- Comment for Pop OS and automatic dist selection for Ubuntu based distributions

It would be cool if packagecloud could support more distros considering Mint and Pop OS are fairly popular, but that isn't a problem for git-lfs.


INSTALLING.md
```diff
@@ -24,10 +24,16 @@ If you are running a distribution which does not match exactly a repository uplo
 
 If you are running LinuxMint 17.1 Rebecca, which is downstream of Ubuntu Trusty and Debian Jessie, you can run:
 
-`curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | os=debian dist=jessie sudo -E sudo bash`
+`curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | os=debian dist=jessie sudo -E bash`
```
bk2204, line 27: This looks obviously correct. This is now obsolete, though, but we don't have to change that now.

INSTALLING.md
```diff
@@ -24,10 +24,16 @@ If you are running a distribution which does not match exactly a repository uplo
...
 
 The `os` and `dist` variables passed-in will override what would be detected for your system and force the selection of the upstream distribution's repository.
 
+You may also be able to run the following to automatically detect the dist for Ubuntu based distributions such as Pop OS:
+```
+. /etc/lsb-release
```
bk2204, line 33: This line is going to import everything in that file into the current shell, which may not be what people want.  Could we use parentheses around these two commands to import that in a subshell?  I'd also recommend in general avoiding `sudo -E` because sometimes that functionality is disabled, so maybe we'd end up with something like this:

```
(. /etc/lsb-release &&
curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | 
sudo env os=ubuntu dist="${DISTRIB_CODENAME}" bash)
```

Would that work for you?

bk2204 (approved):

INSTALLING.md
ry167: Yep, that looks much better. Will update the PR

INSTALLING.md
ry167: Agreed, didn't want to go changing the entire page though

bk2204 (approved):
This looks great.  Thanks so much for the patch and for the quick turnaround time in getting this in shape.  Since I'll be cutting a release today, we can include this as part of it.

bk2204:
It looks like we have some trailing whitespace here that's causing CI to fail.  Could you fix that?

ry167:
@bk2204 Should be fixed now! Single space after the pipe character.

bk2204 (approved):
Thanks, this looks good.  I'll merge it after we've done our release on Wednesday.

MoneyDaMotive:
>  #``

chrisd8088:
This still looks good to me, so I'm going to merge it once the CI jobs are complete.

Issue #3484, opened 2019-01-18
Remove LFS from repository


thomase1993:
I tried to remove git lfs with this function:

$ git lfs migrate export --verbose --include "*.png"
migrate: Fetching remote refs: ..., done                                                                                                                                                      
migrate: Sorting commits: ..., done                                                                                                                                                           
migrate: Rewriting commits: 100% (0/0), done                                                                                                                                                  
migrate: Updating refs: ..., done                                                                                                                                                             
migrate: checkout: ..., done                                                                                                                                                                  
prune: 1365 local object(s), 1365 retained, done                                                                                                                                              

The history is not rewritten. And gitlab still shows LFS files. What do I do wrong?

I want to remove LFS because I do not want to have a gitlab instance just for some of my private projects. I only want ssh access to the server and LFS doesn't support ssh only (as I read somewhere).

I found this ticket: #3463 with the answer from  @bk2204.

bk2204:
Hey, sorry to hear you're having trouble.

I think you'll want to specify one or more references that you want to modify, or use the `--everything` option. For example, you could say, `git lfs migrate --export --verbose --include="*.png" --include-ref="refs/heads/master"` or `git lfs migrate --export --verbose --include="*.png" --everything`.

jklippel:
As I ran into a similar issue just now, I decided to try to enhance the man-page a little. Even though it is a rather old issue I think it is a good idea to extend the documentation.

thomase1993:
Hi, 

Thanks for your effort. :-) 

Part of the misunderstanding comes from the fact that the output seems to do something. Perhaps one would need to use a ref or the --everything option. Maybe thius some sort of user interface validation. :-)

On the other hand, maybe you could give a message that n commits have been adjusted.

PR #5018, opened 2022-05-21
enhance git-lfs-migrate(1) man page

jklippel:
Add a hint that a reference must be provided at the command line or export will not perform any changes.

Fixes: #3484

### DIFF
--- a/docs/man/git-lfs-migrate.1.ronn
+++ b/docs/man/git-lfs-migrate.1.ronn
@@ -311,6 +311,9 @@ argument to specify which files to export. Files matching the `--include`
 patterns will be removed from Git LFS, while files matching the `--exclude`
 patterns will retain their Git LFS status. The export command will modify the
 `.gitattributes` to set/unset any filepath patterns as given by those flags.
+The `export` mode also requires at least one or more references (specified by
+``--include-ref` or `--everything`) to be given. Otherwise, the export won't do
+anything. The given references will be rewritten during the export.
 
 ## INCLUDE AND EXCLUDE
 
--- a/docs/man/git-lfs-migrate.1.ronn
+++ b/docs/man/git-lfs-migrate.1.ronn
@@ -312,8 +312,8 @@ patterns will be removed from Git LFS, while files matching the `--exclude`
 patterns will retain their Git LFS status. The export command will modify the
 `.gitattributes` to set/unset any filepath patterns as given by those flags.
 The `export` mode also requires at least one or more references (specified by
-``--include-ref` or `--everything`) to be given. Otherwise, the export won't do
-anything. The given references will be rewritten during the export.
+`--include-ref` or `--everything`) to be given. The given references will be
+rewritten during the export.
 
 ## INCLUDE AND EXCLUDE
 


chrisd8088:
Hey, thanks for the PR and for contributing to Git LFS!

I do have a concern here which is that I'm not sure it's accurate to say the `git lfs migrate export` command does nothing if one doesn't supply an explicit reference or use the `--everything` option.  For instance, here's an example set of commands:
```
$ git init test
$ cd test
$ echo abc >abc.bin
$ git add abc.bin
$ git commit -m initial

$ git lfs migrate import --include="*.bin"
$ git lfs ls-files
edeaaff3f1 - abc.bin
$ cat .gitattributes 
*.bin filter=lfs diff=lfs merge=lfs -text

$ git lfs migrate export --include="*.bin"
$ git lfs ls-files
$ cat .gitattributes 
*.bin filter=lfs diff=lfs merge=lfs -text
*.bin !text !filter !merge !diff
```

So that `git lfs migrate export` clearly did something (i.e., reversed the import) on the `HEAD` ref without any need to specify that ref.

It is true that the command must have a file path pattern specified, and we note that in [manual page](https://github.com/git-lfs/git-lfs/blob/315a79216cbaec35b38abdca853ed4cd82389b78/docs/man/git-lfs-migrate.1.ronn?plain=1) already with the comment:
```
(At least one include pathspec is required for the `export` mode.)
```

Out of interest, what is the exact problem you encountered?  If it's reproducible with a small example, that would help to clarify the situation and then we can look at whether it's a bug or something that needs to be addressed in the documentation.

chrisd8088:
The current changes in this PR add the statement: "The `export` mode also requires at least one or more references (specified by`--include-ref` or `--everything`) to be given".  But that's not accurate, as the example I gave above demonstrates; it uses `HEAD` by default and neither `--include-ref` nor `--everything` are supplied.

What is the problem you encountered, exactly?  I think knowing that will help clarify things.

jklippel:
> I do have a concern here which is that I'm not sure it's accurate to say the `git lfs migrate export` command does nothing if one doesn't supply an explicit reference or use the `--everything` option. For instance, here's an example set of commands:

You are of course correct. I am sorry I missed that. It was not directly visible to me, I should have checked.

> 
> It is true that the command must have a file path pattern specified, and we note that in [manual page](https://github.com/git-lfs/git-lfs/blob/315a79216cbaec35b38abdca853ed4cd82389b78/docs/man/git-lfs-migrate.1.ronn?plain=1) already with the comment:

I propose to add a statement on this to the export-section, too.

If I read the whole man page, I should/would have read this. But the first time I read the man page, I tried the import. When I returned to the man page and read on the export, I skipped directly to the (rather short) export-section. Then I wondered why it did not work. I searched, found this open bug, tested it with `--everything` and it worked as expected. I wanted other people to not make the same mistake and as there was this old and open issue, I opened this pull-request.

> Out of interest, what is the exact problem you encountered? If it's reproducible with a small example, that would help to clarify the situation and then we can look at whether it's a bug or something that needs to be addressed in the documentation.

I don't think there is any bug here. I did not provide any reference and assumed that this export command did nothing.

Anyway, to resolve this:

- Add a line to the documentation (as proposed in the adapted PR or any further adapted version).
- Discard the PR

Either way I'd propose to close this old issue afterwards.

jklippel:
> The current changes in this PR add the statement: "The `export` mode also requires at least one or more references (specified by`--include-ref` or `--everything`) to be given". But that's not accurate, as the example I gave above demonstrates; it uses `HEAD` by default and neither `--include-ref` nor `--everything` are supplied.
> 
> What is the problem you encountered, exactly? I think knowing that will help clarify things.

Sorry I pushed and then took my time replying. I hope I answered your question already.

chrisd8088:
Thanks for the comments!  I certainly don't mind adding a note in the manual page's section on the `export` subcommand about needing a path pattern.

What I still want to clarify, though, is whether that's sufficient for your case or not, because there's an important difference between requiring at least one path pattern (with `include` or `exclude`) and requiring a Git _reference_, which is what `--include-ref` and `--everything` do.

The old issue #3484 reports a problem where a path pattern is supplied (as is required) with `--include` but the `export` subcommand _still_ doesn't do anything.

Is that the situation you ran into?  If so, it would be great to have a proper reproduction case we can test with to see why that's happening, because that may not be the expected behaviour (but see below).  If you don't supply an explicit reference, `HEAD` should be assumed.

But if what you meant is that you didn't supply a path pattern, that's a different story -- in that case, yes, the expected behaviour is not to do anything.

Now maybe you did give a path pattern with `--include` and then found there was no change because there were no files matching that pattern in `HEAD`, and when you added `--everything`, all refs were checked and so the files you were inspecting were migrated.  Again, though, that's expected behaviour and the manual page describes it:
```
To examine or modify files in branches other than the currently checked-out
one, branch refs may be specified directly, or provided in one or more
`--include-ref` options.  They may also be excluded by prefixing them
with `^` or providing them in `--exclude-ref` options.  Use the `--everything`
option to specify that all refs should be examined, including all remote refs.
```

chrisd8088:
I'm going to close this as there's been no followup recently, but if it is still relevant, please do reopen it.  Thanks!

Issue #4810, opened 2022-01-06
git lfs migrate import --everything ignores malformed references silently


wuhaochen:
**Describe the bug**
We were trying to migrate a repo to git lfs which had a malformed local reference:
`refs/head/some-description` instead of `refs/heads/some-description`
After running `git lfs migrate import everything` that reference was ignored silently. Then when we tried to push it to the remote, it failed because we still have files larger than the size limit in that reference.

**Root cause**
In this part of the code https://github.com/git-lfs/git-lfs/blob/ee3088d4804bd2b699eb6a76aa1318d1d2283ed6/commands/command_migrate.go#L193-L213
The reference was recognized as `RefTypeOther` and got skipped because it doesn't fit the `case "merge-requests", "pull", "pull-requests"`

**Questions**
I understand this is more of an user issue that caused by crazy input but when we are trying to automate things and migrate code bases in batch, errors like this is very hard to debug.
Would be better to give some warnings on the refs that was ignored?
Also is there any concern to just try to migrate all the refs with `RefTypeOther`?

bk2204:
Hey,

I think my preference here would be to just migrate everything that isn't a special ref to Git (that is, everything that has at least three levels that's not `refs/notes`, `refs/bisect`, or `refs/replace`), since that's what we've documented our behavior to be for `--everything`.  I'm a little concerned that we may be missing some sort of special ref to Git and then we'd end up with undesirable behavior, though.

Thoughts on the best approach here, @git-lfs/core?

wuhaochen:
I would agree with migrating everything.
I'm not familiar with all the special refs in git but I'm also not sure if migrating special ref to git is actually going to cause any issue.   Even if so, what if these special refs include large files? Is that possible to also support them?
I would even suggest not to have the three levels constraint, the refs being filtered are returned from `git for-each-ref` from https://github.com/git-lfs/git-lfs/blob/ee3088d4804bd2b699eb6a76aa1318d1d2283ed6/git/git.go#L1217, if git decides to return a ref that are not three levels, it probably should also be processed.

The other option could be to introduce more parameters to differentiate actually migrating everything vs migrating currently what `--everything` is doing.

chrisd8088:
I wondered at first whether we could output interactive prompts ("do you want to migrate ...?") but that might break the workflow of folks who are using the migration command in scripts.

I would also agree that `--everything` should migrate, well, everything, if it possibly can.

[...]

PR #5045, opened 2022-06-11
Change git-lfs migrate import --everything to migrate everything except for special git refs

wuhaochen:
This PR implements @bk2204 's suggestion in #4810 to only exclude special git refs.

### DIFF
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -128,6 +128,13 @@ func rewriteOptions(args []string, opts *githistory.RewriteOptions, l *tasklog.L
 // isSpecialGitRef checks if a ref spec is a special git ref to exclude from
 // --everything
 func isSpecialGitRef(refspec string) bool {
+	// Special refspecs.
+	switch refspec {
+	case "refs/stash":
+		return true
+	}
+
+	// Special refspecs from namespaces.
 	parts := strings.SplitN(refspec, "/", 3)
 	if len(parts) < 3 {
 		return false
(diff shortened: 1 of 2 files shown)


commands/command_migrate.go
```diff
@@ -125,6 +125,21 @@ func rewriteOptions(args []string, opts *githistory.RewriteOptions, l *tasklog.L
...
+// isSpecialGitRef checks if a ref spec is a special git ref to exclude from
+// --everything
+func isSpecialGitRef(refspec string) bool {
+	parts := strings.SplitN(refspec, "/", 3)
+	if len(parts) < 3 {
+		return false
+	}
+	prefix := strings.Join(parts[:2], "/")
+	switch prefix {
+	case "refs/notes", "refs/bisect", "refs/replace":
```
bk2204, line 137: I think we'll also want to include `refs/stash` here, since we won't want to rewrite stashes.

bk2204 (commented):

This looks mostly correct.  I think there's one change we should make to the code, and some tests, probably in `t/t-migrate-import.sh`, would also be a good idea.

commands/command_migrate.go
wuhaochen: > 

Added a small unit test, I'll probably need a bit more time to figure out the how to get the test right in `.sh` unless people want to give some help on generating examples of excluded refs.

bk2204: I think this is probably fine as it stands.

bk2204 (approved):

PR #5048, opened 2022-06-16
installer: handle `BashOnly` Git for Windows gracefully

dscho:
This addresses the issue raised in https://github.com/git-lfs/git-lfs/discussions/5031.

/cc @chrisd8088

### DIFF
--- a/script/windows-installer/inno-setup-git-lfs-installer.iss
+++ b/script/windows-installer/inno-setup-git-lfs-installer.iss
@@ -101,6 +101,56 @@ begin
     Result := Pos(';' + UpperCase(ParamExpanded) + '\;', ';' + UpperCase(OrigPath) + ';') = 0;
 end;
 
+function SetEnvironmentVariable(lpName,lpValue:String):Boolean;
+external 'SetEnvironmentVariableW@Kernel32.dll stdcall delayload';
+
+// When Git for Windows is installed with the PATH option "Bash only", i.e.
+// _without_ adding anything to the global `PATH`, we will not find `git.exe`
+// there.
+//
+// Detect that situation and add `<Git>\cmd` to the `PATH` so that we find it
+// when registering Git LFS later.
+function AddGitForWindowsCMDToPATHIfNeeded: boolean;
+var
+  Domain: Integer;
+  Key, PathOption, AppPath, Path: string;
+begin
+  Result := False;
+
+  Key := 'Microsoft\Windows\CurrentVersion\Uninstall\Git_is1';
+  if RegKeyExists(HKEY_LOCAL_MACHINE, 'Software\Wow6432Node\' + Key) then begin
+    Domain := HKEY_LOCAL_MACHINE;
+    Key := 'Software\Wow6432Node\' + Key;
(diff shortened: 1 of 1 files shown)


script/windows-installer/inno-setup-git-lfs-installer.iss
```diff
@@ -101,6 +101,56 @@ begin
...
+    (not FileExists(AppPath + '\cmd\git.exe'))
+  then
+    Exit;
+
+  // Extend PATH so that it finds `git.exe`
+  Path := GetEnv('PATH');
+  if Path<>''
+  then begin
+    Path := AppPath + '\cmd;' + Path;
+  end;
```
chrisd8088, line 148: I realize it's very unlikely that `Path` would be empty, but since we're checking the non-empty case, should we include an else block like:
```suggestion
  if Path = '' then
    Path := AppPath + '\cmd'
  else
    Path := AppPath + '\cmd;' + Path;
```

chrisd8088 (commented):
Thanks for this!  I had one comment so far, and I'll give it a try later on a fresh Windows VM (once I remember how to tinker with the installer again!)

script/windows-installer/inno-setup-git-lfs-installer.iss
```diff
@@ -101,6 +101,56 @@ begin
...
+    Domain := HKEY_LOCAL_MACHINE;
+    Key := 'Software\' + Key;
+  end else if RegKeyExists(HKEY_CURRENT_USER, 'Software\' + Key) then begin
+    Domain := HKEY_CURRENT_USER;
+    Key := 'Software\' + Key;
+  end else
+    Exit;
+
+  if (not RegQueryStringValue(Domain, Key, 'Inno Setup CodeFile: Path Option', PathOption)) or
+    (PathOption <> 'Cmd') or // (PathOption <> 'BashOnly') or
```
chrisd8088, line 137: OK, I've run a test on a fresh Windows VM where I first installed Git for Windows and chose not to install Git LFS and also chose the "Use Git from Git Bash only" option, and then built a new installer using this PR's changes and ran it.

That failed to prevent the "Could not find Git; can not register Git LFS" message because the `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Git_is1` registry subkey's `Inno Setup CodeFile: Path Option` value has the data string `BashOnly`.

Changing this line as follows made the installer able to proceed, however:
```suggestion
    (PathOption <> 'BashOnly') or
```

So I wonder if this was a code change and comment left over from debugging?

chrisd8088 (commented):
I've run a first test and found one thing I think needs to be resolved -- let me know if you agree.

script/windows-installer/inno-setup-git-lfs-installer.iss
dscho: Darn it. That's _totally_ from debugging.

chrisd8088 (approved):
This looks great and works as expected on my test VM -- thank you!  🙇

PR #5053, opened 2022-06-23
- fix `unhandled error` warning

ovaldi:
image


ovaldi:
image

@bk2204

bk2204:
Hey,

I think we need to wait on #5054 to merge to fix that, but I wouldn't stress out about the failure for now.  Let me take a look at this, and if it looks good, we'll merge it as soon as that PR merges.

bk2204 (approved):
Hey,

Thanks for the patch, and welcome to Git LFS!  We typically don't do this kind of checking in Git LFS, but since you've sent the patch, I think we can take it here.

I'll merge this once #5054 merges and we should be in good shape.

Issue #3835, opened 2019-09-25
Smudge error: Error downloading object for "git checkout" from alternative remote


jamie-pate:
When I add a 2nd remote, and try to checkout a remote branch from that remote git-lfs will try to obtain objects from the `origin` remote instead

```
git-lfs --version
git-lfs/2.8.0 (GitHub; linux amd64; go 1.12.8)

# relevant variables from git config?

file:/home/jpate/.gitconfig     filter.lfs.clean=git-lfs clean -- %f
file:/home/jpate/.gitconfig     filter.lfs.smudge=git-lfs smudge -- %f
file:/home/jpate/.gitconfig     filter.lfs.process=git-lfs filter-process
file:/home/jpate/.gitconfig     filter.lfs.required=true
file:.git/config        remote.origin.url=ssh://user@origin.example.com:29418/project
file:.git/config        remote.origin.fetch=+refs/heads/*:refs/remotes/origin/*
file:.git/config        remote.gitea.url=ssh://git@172.254.254.254:222/user/escaperoom_project.git
file:.git/config        remote.gitea.fetch=+refs/heads/*:refs/remotes/gitea/*

```

`git checkout not-origin/feat_somebranch -b feat_somebranch`

See attached log with `GIT_TRACE=1`:

It seems like it's trying to download the files from `origin` even though i'm checking out from `gittea/feat_somebranch`

**excerpt**
```
Updating files: 100% (1678/1678), done.
17:57:48.938063 trace git-lfs: tq: sending batch of size 77
17:57:48.938248 trace git-lfs: run_command: ssh -p 29418 -- user@origin.example.com git-lfs-authenticate securityfabricinaction/escaperoom download
17:57:49.320258 trace git-lfs: api: batch 77 files
17:57:49.320661 trace git-lfs: HTTP: POST https://origin.example.com/projectinfo/lfs/objects/batch
17:57:49.414343 trace git-lfs: HTTP: 200
```
[git-lfs-checkout-bug.log](https://github.com/git-lfs/git-lfs/files/3650100/git-lfs-checkout-bug.log)

<hr>

I was able to work around the problem using 
```bash
GIT_LFS_SKIP_SMUDGE=1 git checkout gitea/feat_somebranch -b feat_somebranch
git pull gitea feat_sdbranch_cutscene
```

bk2204:
Hey,

[...]

Issue #3803, opened 2019-09-01
git lfs pull don't respect sparsecheckout.


kazuki-ma:
Some files don't exist in worktree if core.sparsecheckout is true.

But `git lfs pull` will download and restore those files.

bk2204:
Hey,

Seeing as we don't have any code for sparse checkout, I'm not surprised that this doesn't work.  I do think we should fix Git LFS so that it does, however.

anb0s:
We found similar issue today too: 
- worktree configured with sparse checkouts
- update of worktree with new binary files
- right Git credentials, but wrong LFS credetials (password changed at artifactory server)
- git checkout succesfull
- content filtering showed error message, because of wrong credetials, but no error / abort in script (have to investigate if script does not check errors or if git lfs returns no error in this case)
- started build --> weird errors that new exe cannot be executed (pointer file, no real exe binary)
- tried to repair with "git checkout -f" etc. -> does not update the pointer files
- tried to repair with
  - "git lfs fetch origin <branch-or-tag>" --> tooks much longer 
  - "git lfs checkout <branch-or-tag>" --> tooks much longer
  - now the checkout has all folders and files and increased from 12GB to 48GB, because the whole content was checked out, because LFS fetch is not sparse aware
  - also the LFS cache is very big now, because LFS fetch is not sparse aware

For us that means we cannot easily repair the worktree with simple "git lfs" commands. Maybe some script can help to iterate all pointer files and fetch + checkout them file by file?

woopla:
Let me add my vote to this one. Right now it's an all-or-nothing affair, which makes me sad :) Especially with Git's per-worktree config not working that well with Git 2.25, if I have 10 sparse worktrees, I'm gonna have 10 copies :crying_cat_face:

zhuda:
I will vote this one as well. In our pipeline it takes more than 10 minutes to run `git lfs pull origin` but actually none of them are needed basing on sparse-checkout settings.

Mike-Moreton-Lifesize:
+1 for this.  As more and more people learn about sparse-checkout, you're going to get more requests!

srohmen:
is there at least any workaround  to fetch only lfs objects from a folder? git lfs fetch seems to download all the data at once

xiaolongli827:
I have this problem, too. Did you solve it

xiaolongli827:
@kazuki-ma

kazuki-ma:
I've been away from work that requires git-lfs.
[...]

PR #5066, opened 2022-07-04
Allow alternative remotes to be handled by LFS

srohmen:
This fixes issue #3835. LFS could not derive the right remote endpoint under certain circumstances. For example, in a detached state, a simple heuristic was used to derive the remote (i.e. branch.*.remote, remote.lfsdefault, any other single remote, origin). This fails if there are multiple remotes in play. This patch derives the corresponding remote from the commit hash / treeish.

Thanks to @stanhu for providing the first version which this PR is based on.

### DIFF
--- a/commands/command_filter_process.go
+++ b/commands/command_filter_process.go
@@ -95,6 +95,14 @@ func filterCommand(cmd *cobra.Command, args []string) {
 				closeOnce = new(sync.Once)
 				available = make(chan *tq.Transfer)
 
+				if cfg.AutoDetectRemoteEnabled() {
+					// update current remote with information gained by treeish
+					newRemote := git.FirstRemoteForTreeish(req.Header["treeish"])
+					if newRemote != "" {
+						cfg.SetRemote(newRemote)
+					}
+				}
+
 				q = tq.NewTransferQueue(
 					tq.Download,
 					getTransferManifestOperationRemote("download", cfg.Remote()),
--- a/git/git.go
+++ b/git/git.go
@@ -1481,3 +1481,71 @@ func ObjectDatabase(osEnv, gitEnv Environment, gitdir, tempdir string) (*gitobj.
 	}
 	return odb, nil
 }
+
+func remotesForTreeish(treeish string) []string {
+	var outp string
+	var err error
+	if treeish == "" {
+		//Treeish is empty for sparse checkout
+		tracerx.Printf("git: treeish: not provided")
+		outp, err = gitNoLFSSimple("branch", "-r", "--contains", "HEAD")
+	} else {
+		tracerx.Printf("git: treeish: %q", treeish)
+		outp, err = gitNoLFSSimple("branch", "-r", "--contains", treeish)
+	}
+	if err != nil || outp == "" {
+		tracerx.Printf("git: symbolic name: can't resolve symbolic name for ref: %q", treeish)
+		return []string{}
+	}
+	return strings.Split(outp, "\n")
+}
+
+// remoteForRef will try to determine the remote from the ref name.
+// This will return an empty string if any of the remote names have a slash
+// because slashes introduce ambiguity. Consider two refs:
+//
+// 1. upstream/main
+// 2. upstream/test/main
+//
+// Is the remote "upstream" or "upstream/test"? It could be either, or both.
+// We could use git for-each-ref with %(upstream:remotename) if there were a tracking branch,
+// but this is not guaranteed to exist either.
+func remoteForRef(refname string) string {
+	tracerx.Printf("git: working ref: %s", refname)
+	remotes, err := RemoteList()
+	if err != nil {
+		return ""
+	}
+	parts := strings.Split(refname, "/")
+	if len(parts) < 2 {
+		return ""
+	}
+	for _, name := range remotes {
+		if strings.Contains(name, "/") {
+			tracerx.Printf("git: ref remote: cannot determine remote for ref %s since remote %s contains a slash", refname, name)
+			return ""
+		}
+	}
+	remote := parts[0]
+	tracerx.Printf("git: working remote %s", remote)
+	return remote
+}
+
+func getValidRemote(refs []string) string {
+	for _, ref := range refs {
+		if ref != "" {
+			return ref
+		}
+	}
+	return ""
+}
+
+// FirstRemoteForTreeish returns the first remote found which contains the treeish.
+func FirstRemoteForTreeish(treeish string) string {
+	name := getValidRemote(remotesForTreeish(treeish))
+	if name == "" {
+		tracerx.Printf("git: remote treeish: no valid remote refs parsed for %q", treeish)
+		return ""
+	}
+	return remoteForRef(name)
+}
--- a/lfs/gitfilter_smudge.go
+++ b/lfs/gitfilter_smudge.go
@@ -77,6 +77,14 @@ func (f *GitFilter) Smudge(writer io.Writer, ptr *Pointer, workingfile string, d
 	} else if statErr != nil || stat == nil {
 		if download {
 			n, err = f.downloadFile(writer, ptr, workingfile, mediafile, manifest, cb)
+
+			// In case of a cherry-pick the newly created commit is likely not yet
+			// be found in the history of a remote branch. Thus, the first attempt might fail.
+			if err != nil && f.cfg.SearchAllRemotesEnabled() {
+				tracerx.Printf("git: smudge: default remote failed. searching alternate remotes")
+				n, err = f.downloadFileFallBack(writer, ptr, workingfile, mediafile, manifest, cb)
+			}
+
 		} else {
 			return 0, errors.NewDownloadDeclinedError(statErr, tr.Tr.Get("smudge filter"))
 		}
@@ -121,6 +129,46 @@ func (f *GitFilter) downloadFile(writer io.Writer, ptr *Pointer, workingfile, me
 	}
 
 	return f.readLocalFile(writer, ptr, mediafile, workingfile, nil)
+}
+
+func (f *GitFilter) downloadFileFallBack(writer io.Writer, ptr *Pointer, workingfile, mediafile string, manifest *tq.Manifest, cb tools.CopyCallback) (int64, error) {
+	// Attempt to find the LFS objects in all currently registered remotes.
+	// When a valid remote is found, this remote is taken persistent for
+	// future attempts within downloadFile(). In best case, the ordinary
+	// call to downloadFile will then succeed for the rest of files,
+	// otherwise this function will again search for a valid remote as fallback.
+	remotes := f.cfg.Remotes()
+	for index, remote := range remotes {
+		q := tq.NewTransferQueue(tq.Download, manifest, remote,
+			tq.WithProgressCallback(cb),
+			tq.RemoteRef(f.RemoteRef()),
+		)
+		q.Add(filepath.Base(workingfile), mediafile, ptr.Oid, ptr.Size, false, nil)
+		q.Wait()
+
+		if errs := q.Errors(); len(errs) > 0 {
+			var multiErr error
+			for _, e := range errs {
+				if multiErr != nil {
+					multiErr = fmt.Errorf("%v\n%v", multiErr, e)
+				} else {
+					multiErr = e
+				}
+			}
+			wrappedError := errors.Wrapf(multiErr, tr.Tr.Get("Error downloading %s (%s)", workingfile, ptr.Oid))
+			if index >= len(remotes)-1 {
+				return 0, wrappedError
+			} else {
+				tracerx.Printf("git: download: remote failed %s %s", remote, wrappedError)
+			}
+		} else {
+			// Set the remote persistent through all the operation as we found a valid one.
+			// This prevents multiple trial and error searches.
+			f.cfg.SetRemote(remote)
+			return f.readLocalFile(writer, ptr, mediafile, workingfile, nil)
+		}
+	}
+	return 0, errors.Wrapf(errors.New("No known remotes"), tr.Tr.Get("Error downloading %s (%s)", workingfile, ptr.Oid))
 }
 
 func (f *GitFilter) readLocalFile(writer io.Writer, ptr *Pointer, mediafile string, workingfile string, cb tools.CopyCallback) (int64, error) {
--- a/t/t-multiple-remotes.sh
+++ b/t/t-multiple-remotes.sh
@@ -1,0 +1,192 @@
+#!/usr/bin/env bash
+# Test lfs capability to download data when blobs are stored in different
+# endpoints
+
+. "$(dirname "$0")/testlib.sh"
+
+# This feature depends on the treeish parameter that is provided as metadata
+# in git versions higher or equal than 2.27
+ensure_git_version_isnt $VERSION_LOWER "2.27.0"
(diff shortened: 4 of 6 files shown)


srohmen:
what can I do about the ruby gem issue within the CI job? it does not look like my patch is responsible for this

bk2204:
It's not.  I'm working on finishing the patch which should fix that and I'll re-run the job once it's merged.

bk2204 (commented):
Hey,

Thanks for the patch, and welcome to Git LFS!

Overall, I think this could be a valuable feature.  I think we'll need to make these options since overall they change behaviour in an incompatible way.  We'll also need some integration tests here (those would be in the `t/` directory).  Finally, CI has been fixed, but you'll need to rebase on the main branch to pick up the fixes.

If you need any help on getting this into shape, let me know, and I'll try to help you out.  My apologies for taking so long to get to this.

commands/command_filter_process.go
```diff
@@ -95,6 +95,12 @@ func filterCommand(cmd *cobra.Command, args []string) {
 				closeOnce = new(sync.Once)
 				available = make(chan *tq.Transfer)
 
+				//Update current remote with treeish
+				newRemote := git.FirstRemoteForTreeish(req.Header["treeish"])
+				if newRemote != "" {
+					cfg.SetRemote(newRemote)
+				}
```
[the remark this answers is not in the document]
srohmen: done in my latest push

lfs/gitfilter_smudge.go
```diff
@@ -77,6 +77,14 @@ func (f *GitFilter) Smudge(writer io.Writer, ptr *Pointer, workingfile string, d
 	} else if statErr != nil || stat == nil {
 		if download {
 			n, err = f.downloadFile(writer, ptr, workingfile, mediafile, manifest, cb)
+
+			// In case of a cherry-pick the newly created commit is likely not yet
+			// be found in the history of a remote branch. Thus, the first attempt might fail.
+			if err != nil {
+				tracerx.Printf("Default remote failed. Searching alternate remotes")
+				n, err = f.downloadFileFallBack(writer, ptr, workingfile, mediafile, manifest, cb)
```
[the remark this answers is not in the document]
srohmen: done in my latest push

lfs/gitfilter_smudge.go
```diff
@@ -123,6 +131,39 @@ func (f *GitFilter) downloadFile(writer io.Writer, ptr *Pointer, workingfile, me
 	return f.readLocalFile(writer, ptr, mediafile, workingfile, nil)
 }
 
+func (f *GitFilter) downloadFileFallBack(writer io.Writer, ptr *Pointer, workingfile, mediafile string, manifest *tq.Manifest, cb tools.CopyCallback) (int64, error) {
+	// attempt to find the LFS objects in all currently registered remotes
+	remotes := f.cfg.Remotes()
+	for index, remote := range remotes {
+		q := tq.NewTransferQueue(tq.Download, manifest, remote,
```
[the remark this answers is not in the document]
srohmen: I introduced a treatment for that, and also extended the code documentation how that is done. I suggest we start a new discussion based on that, if that is still not sufficient

srohmen:
Thanks @bk2204 for you feedback. We (a co-worker @Leo1690 and me) have treated most of your comments. I rebased my branch, squashed the new changes in the old commits and pushed it to my feature branch. However, we are still working on the test environment. We have also shell scripts which reproduce the issue, but to port that to the environment within t/ of git-lfs it takes a bit more effort. I hope we can at least review and discuss the updated patch in the meantime.

Leo1690:
Thanks @bk2204 for your feedback! 
I have a couple of issues creating the tests.
1. By default all remotes uses the same lfs server. To reproduce the issue I need each remote to be linked to a different lfs server and I cannot find out how to do it
2. I tried to use lfs directly with the file system but I get this error batch {request: missing protocol: "file://}. I tried to set up a standalone transfer mode that supports the file system but It did not work.
3. I can try to mount the remotes as docker containers and and use git with ssh to transfer lfs data, but I do not know if you would accept a merge request that adds docker containers to your testing environment.
I think that if we use docker containers as remotes or repos in the test suite it would help to isolate the git configuration and Env variables between tests. I believe it is worth considering.

srohmen:
I pushed a commit authored by @Leo1690 which implements the tests. I hope it is fine that this MR contains now two different authors in the line of commits.

bk2204:
> Thanks @bk2204 for your feedback! I have a couple of issues creating the tests.
> 
>     1. By default all remotes uses the same lfs server. To reproduce the issue I need each remote to be linked to a different lfs server and I cannot find out how to do it

I believe, provided you use different URLs, the test LFS server should expose different repositories.  It should be sufficient to expose multiple repositories on the same server instead of different servers, I would think.

>     2. I can use lfs directly with the file system but I get this error batch {request: missing protocol: "file://} when I use sparse-checkout and cherry-pick. I tried to set up the standalone transfer mode but It did not work.

The standalone transfer agent should be built-in automatically for file URLs.  However, it is known that it requires an absolute URL and it doesn't work for Windows SMB paths.  Sparse checkout almost certainly doesn't work and is tracked in #3803.

Can you provide an example of a shell script that doesn't work for you (without sparse checkout)?

Leo1690:
Hi @bk2204.
- If all remotes uses the same lfs server, the tests succeed even without our patch. The problem shows up only if different remotes point to different lfs servers. The issue can be reproduced with 2 gitlab repos.

- I solved the problem with the file system remotes in the /t folder, and our tests succeed with our patch and fails without it (as expected). So I was finally able to implement all necessary tests including the sparse-checkout test.

- I saw that the CI failed. If I am not wrong, it looks like the problem is when I attempt to access a folder outside of $TRASHDIR. Anyway, I changed the tests to work only inside a subdirectory in $TRASHDIR/home
`make: Leaving directory /root/src/git-lfs/rpm/BUILD/git-lfs-3.2.0/t`
`error: Bad exit status from /var/tmp/rpm-tmp.cqNFbK (%check)`

t/t-multiple-remotes.sh
```diff
@@ -0,0 +1,137 @@
...
+reponame="$(basename "$0" ".sh")"
+
+prepare_mrtest(){
+  MRHOME=$TRASHDIR/home/mrhome
+  rm -rf $MRHOME
+  mkdir $MRHOME
+  HOME=$MRHOME
+  git config --global user.name "Git LFS Tests"
+  git config --global user.email "git-lfs@example.com"
+}
```
bk2204, line 14: I don't think we should be setting `HOME` here, since the test library does that, and we also don't need to set the Git config values, since those should already be set.

Also, please here and everywhere else, double-quote all variables.  On Windows, it's extremely common for a user's home directory (and thus all paths to their repositories) to have spaces, and we need to handle that gracefully.

t/t-multiple-remotes.sh
```diff
@@ -0,0 +1,137 @@
...
+
+prepare_remote() {
+  local repodir="$1"
+  mkdir -p "$repodir"
+  cd "$repodir"
+  git init --bare
+  git config http.receivepack false
+  git lfs install
+}
```
bk2204, line 35: Is there a reason our existing function `setup_remote_repo` doesn't work here?  Why is setting `http.receivepack` necessary (a comment would be appropriate)?

t/t-multiple-remotes.sh
```diff
@@ -0,0 +1,137 @@
...
+
+prepare_forks () {
+  smain=$HOME/$reponame"-main-remote.git"
+  sfork=$HOME/$reponame"-fork-remote.git"
+  cmain=$HOME/$reponame"-main-repo"
+  cfork=$HOME/$reponame"-fork-repo"
+  rm -rf $smain
+  rm -rf $sfork
+  rm -rf $cmain
+  rm -rf $cfork
```
bk2204, line 45: Instead of deleting and recreating repos like this, let's use a parameter which is specific to each call and set that differently in each individual test.  Then we can avoid different tests affecting each other and let the cleanup code at the end of the test clean up all of the test data for us.

You can see an example in `t/t-pull.sh` how we do this with `setup_remote_repo`.

t/t-multiple-remotes.sh
```diff
@@ -0,0 +1,137 @@
...
+  prepare_consumer $cmain
+  git checkout -b master
+  git lfs track '*.bin'
+  git add --all
+  git commit -m "Initial commit"
+  git push -u mr master
+  git push -u fr master
+  #Add a .bin in main repo
+  touch a.bin
+  echo 1234 >> a.bin
```
bk2204, line 57: We typically prefer `printf` over `echo` in our codebase.  It's more consistent and flexible, and in some Windows environments, `echo` may insert CRLF, which makes our hashes change.

t/t-multiple-remotes.sh
```diff
@@ -0,0 +1,137 @@
...
+  git add --all
+  git commit -m "Add Bin file"
+  git push mr master
+  prepare_consumer $cfork
+}
+begin_test "pull from different remote"
```
bk2204, line 63: Let's put a space here and in between each set of tests.

git/git.go
```diff
@@ -1481,3 +1481,72 @@ func ObjectDatabase(osEnv, gitEnv Environment, gitdir, tempdir string) (*gitobj.
...
+// because slashes introduce ambiguity. Consider two refs:
+//
+// 1. upstream/main
+// 2. upstream/test/main
+//
+// Is the remote "upstream" or "upstream/test"? It could be either, or both.
+// We could use git for-each-ref with %(upstream:remotename) if there were a tracking branch,
+// but this is not guaranteed to exist either.
+func remoteForRef(refname string) string {
+	tracerx.Printf("Git Working Ref %", refname)
```
bk2204, line 1514: I don't think this works with just a single percent sign.  Also, typically, we write the trace messages downcase, often with a prefix (so, here, “git: working ref: %s”), which would be great to fix here and elsewhere.  We typically write mostly fragments or phrases which are punctuated with colons as necessary to avoid ambiguity.

bk2204 (commented):
This is looking pretty good.  I left some more comments, but I think we should be good to go soon.

t/t-multiple-remotes.sh
Leo1690: You are right. I added a comment to explain my motivation for that. 
`The global configuration sets a common http endpoint (ip:port) for lfs`
 `We deactivate the git global configuration related with the http protocol`
 `by setting a new .gitconfig file`
 Additionally, I am not setting a new $HOME. Instead I am removing .gitconfig file and setting the settings I need. At the begging of the next test the original .gitconfig file is restored by the function begin_test()

t/t-multiple-remotes.sh
Leo1690: I added 2 comments to explain the new function.
`Set up remote repo that does not use the http API`
`Uses the basic stand alone transfer instead of the http API`
Additionally, I removed the `http.receivepack=false`. It was not necessary after all.

t/t-multiple-remotes.sh
Leo1690: I agree. That looks cleaner. I changed the new function prepare_forks to create remote and local repos with different paths
`prepare_forks test_case`

t/t-multiple-remotes.sh
Leo1690: I fixed it.
`printf "1234" > a.bin`

git/git.go
Leo1690: I fixed it.
`tracerx.Printf("git: working ref: %s", refname)`
`tracerx.Printf("git: ref remote: cannot determine remote for ref %s since remote %s contains a slash", refname, name)`
`tracerx.Printf("git: working remote %s", remote)`
`tracerx.Printf("git: remote treeish: no valid remote refs parsed for %q", treeish)`
`tracerx.Printf("git: smudge: default remote failed. searching alternate remotes")`

t/t-multiple-remotes.sh
bk2204: I'm sorry, I still think there's something I'm missing.  What specific setting is causing the problem?  In `t/t-standalone-file.sh`, we use `file:///` URLs just fine, so it can't be `lfs.url`.

Can we keep the regular `.gitconfig` and simply unset the offending option in each test with `git config --global --unset`?  I'd rather we do that because it means that if we add an important new setting into the testsuite in the future, the entire testsuite picks it up.

t/t-multiple-remotes.sh
```diff
@@ -0,0 +1,136 @@
...
+  git lfs install
+}
+
+prepare_forks () {
+  local testcase="$1"
+  smain="$HOME"/"$reponame"-"$testcase"-main-remote.git
+  sfork="$HOME"/"$reponame"-"$testcase"-fork-remote.git
+  cmain="$HOME"/"$reponame"-"$testcase"-main-repo
+  cfork="$HOME"/"$reponame"-"$testcase"-fork-repo
+  prepare_remote "$smain"
```
bk2204, line 46: Can we change this to `setup_remote_repo "$smain"` and then use `"$REMOTEDIR/$smain.git"`?  We shouldn't need the `git lfs install` since that's already been done once by our setup code, and if we don't delete the `.gitconfig` file as I suggested, then we'll have the configuration persisted.

Basically, my request here is to use as much of the same setup code and other helper functions as possible.  The style of this test is currently very different from our other tests, and if we can make it as similar as possible, then it will be easier to maintain.  You can see what `t/t-verify.sh` looks like, which I think is a good example of a typical test.

bk2204 (commented):
I left a few more comments.  We probably want to also add some documentation in AsciiDoc in the `git-lfs-config` manual page so folks will know how this works.

t/t-multiple-remotes.sh
Leo1690: I am sorry. I seems that it was not necessary to chage the global configuration after all. Thank you for pointing that out. The new patch uses the global seetings provided by `setup()`

t/t-multiple-remotes.sh
Leo1690: You are right, I adpted the tests to use the function `setup_remote_repo()`

srohmen:
I added a commit which extends the documentation. I hope this is to your liking what I have phrased @bk2204

bk2204 (approved):
This looks good to me.  Thanks so much for the patch and working with me on this.

I'm going to announce this as an experimental feature in v3.3.0 so that we can see if there are any bugs and shake them out before we commit to supporting it long term.

I've kicked off CI, and I'll check back on Monday to verify it's passed and then merge.

srohmen:
Oh, I just fixed the phrasing of a comment of a function. I didn't knew that the approval would reset completely.

bk2204:
Hmmm, it looks like CI is broken.  Could you look into that?

srohmen:
@bk2204 2 CI jobs failed because the feature can only be supported for git version 2.27 or newer. We added an exclusion rule for the tests. The other job, which uses the git development version of git, is not explainable. We never touched any of the other tests and the new feature is disabled in that setup. My guess is right now that git upstream behavior changed such that the tests are failing now. Can we run the same CI job without our patch to confirm this theory?

bk2204:
I'm testing a PR in #5082 to see if it makes any difference.

srohmen:
@bk2204 I just rebased the branch on your latest fix. I just hope this fixes the CI issues we had before.

PR #5079, opened 2022-07-15
Adding locking_notes.md

pratap043:
This is a PR with the proposal for enhancing the Git LFS REST API to capture additional notes on the locks.

### DIFF
--- a/docs/proposals/locking_notes.md
+++ b/docs/proposals/locking_notes.md
@@ -9,55 +9,7 @@ The git-lfs REST API for Locks creation should be able to accept additonal attri
 
 ## API extensions
 
-The current Git LFS File Locking API [https://github.com/git-lfs/git-lfs/blob/v3.2.0/docs/api/locking.md] doesn't have a field to capture some information related to Locks creation which would be useful to understand why and from where the lock was acquired. To get more understanding, below is the use case:
-
-Consider an application where source control management is one of the feature and GitHub is the repository being used. Here there are 3 workspaces with 2 users working on a single Repository containing 6 files.
-
-Here the User1 is working on FileA, FileB under Workspace W1, User2 is working on FileC, FileD under Workspace W2 and also User1 is working on FileE, FileF under Workspace W3.
-
-So we acquire locks before working on these files respectively. Now we can see that W1 has locks on FileA, FileB. W2 has locks on FileC, FileD. W3 has locks on FileE, FileF.
-
-Repository
-Wl
--- User1
------- FileA
------- FileB
-W2
--- User2
------- FileC
------- FileD
-W3
--- User1
------- FileE
------- FileF
-
-Now when we run List Locks request from W1, we will get all locks on the Repo as given below. Here we doesn't know from which workspaces, the locks were acquired.
-
-User Workspace File
-User1 -- FileA
-User1 -- FileB
-User1 -- FileE
-User1 -- FileF
-User2 -- FileC
-User2 -- FileD
-
-Our requirement is to get the workspace information along with locks, to know from where the locks were acquired as given below (say When fetchLocks is run from W1):
-
-User Workspace File
-User1 W1 FileA
-User1 W1 FileB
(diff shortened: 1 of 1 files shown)


docs/proposals/locking_notes.md
```diff
@@ -0,0 +1,62 @@
...
+  - Allow to have additional attribute to store the lock message/notes during lock creation
+  - Include lock message/notes in the git-lfs List Locks REST API response
+
+## API extensions
+
+The current Git LFS File Locking API [https://github.com/git-lfs/git-lfs/blob/v3.2.0/docs/api/locking.md] doesn't have a field to capture some information related to Locks creation which would be useful to understand why and from where the lock was acquired. To get more understanding, below is the use case:
+
+In our application team development feature, there are 3 workspaces with 2 users working on a single Repository containing 6 files.
+
+Here the User1 is working on FileA, FileB under Workspace W1, User2 is working on FileC, FileD under Workspace W2 and also User1 is working on FileE, FileF under Workspace W3.
```
bk2204, line 16: I think I'd like to phrase this as a generic comment field rather than being used for any particular workflow.  It may be that you want to use it for this purpose, but others may simply be interested in a free-form comment field, much like a commit message.

bk2204 (commented):
Hey,

I think this is a good start.  In addition to a more generic description, I'd like to see a proposed API body and response, much like in the actual API documentation.

pratap043:
Thanks Brian (@bk2204). I have updated the document with the changes as suggested. Kindly check once and let me know for further modifications if any.

docs/proposals/locking_notes.md
bk2204: To be clearer, I think we should avoid the discussion of your particular use case.  It is one possible use case, but it isn't a particularly common one, and I think that now we have the Concept section and the “With this enhancement” section below, it should be clear to folks why a general-purpose comment field is useful and would be valuable.

pratap043:
Hi Brian (@bk2204 ). I have updated the document by removing the specific use case to make it generic, thank you.

bk2204 (approved):
This looks good, thanks.

bk2204:
I'm going to accept this as a proposal.  That means that we'll consider it for possible implementation in a future version, but due to implementation constraints, it's possible that it will change before it's shipped as we discover additional requirements.

It _doesn't_ mean that we'll be implementing it in a particular timeframe or that we'll immediately accept a PR implementing it, since we typically will want to see at least one major implementer commit to implement it at some point.

I'm also going to ping @git-lfs/implementers since this is a thing they'll be interested in.

pratap043:
Thanks so much Brian @bk2204.

docs/proposals/locking_notes.md
```diff
@@ -0,0 +1,193 @@
...
+<      },
+<      "notes": "Lock applied from Workspace A"
+<    }
+<  ],
+<  "theirs": [],
+<  "next_cursor": "optional next ID"
+< }
+```
+
+### Request (with out notes)
```
slonopotamus, line 156: Do we actually want this? Such feature (answering without notes after you added notes support) will actually require noticeable amounts of code. Both git-lfs and https://github.com/git-as-svn/git-lfs-java simply call JSON serializer on a Lock structure.

git-lfs: https://github.com/git-lfs/git-lfs/blob/v3.2.0/locking/locks.go#L200-L211
git-lfs-java: https://github.com/git-as-svn/git-lfs-java/blob/v0.19.0/gitlfs-common/src/main/kotlin/ru/bozaro/gitlfs/common/data/Lock.kt#L6-L30

And it is nontrivial to conditionally exclude a field from serialization.

The only reason to omit notes that I can come up with is network traffic.

bk2204: I think it's fine for the server to simply return any notes that are present, whether or not any were requested.  If no notes are requested on the client side, we'll just not emit the field, and then the server can either return them if appropriate or not.

PR #5082, opened 2022-07-25
t: set init.defaultBranch

bk2204:
Previously, we used a template to set the default HEAD ref to "main" so that our branch would be correct.  It appears that Git has changed things so that that no longer works, so set init.defaultBranch as well to make sure we initialize new repositories properly.


chrisd8088 (approved):

jessicakink:
Need to suck big dick!

PR #5086, opened 2022-08-12
push: add ability to read refs/oids from stdin

rcoup:
Add the ability to provide refs or object ids to `git lfs push` via standard input. This can make life simpler when plumbing together commands or working with large numbers of OIDs/refs. Particularly on Windows where the maximum command line length is much shorter.

Read refs from stdin:
```sh
echo "mybranch" | git lfs push origin --stdin
```

Read object IDs from stdin:
```sh
printf "4c48d2a6991c9895bcddcf027e1e4907280bcf21975492b1afbade396d6a3340\n82be50ad35070a4ef3467a0a650c52d5b637035e7ad02c36652e59d01ba282b7\n\n" \
  | git lfs push origin --object-id --stdin
```

Values read from stdin are newline-delimited, and blank lines are ignored. If `--stdin` is used, any values passed via the command line arguments ~are ignored~ raise an error.

### DIFF
--- a/commands/command_push.go
+++ b/commands/command_push.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"bufio"
 	"os"
 
 	"github.com/git-lfs/git-lfs/v3/errors"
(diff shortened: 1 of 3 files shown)


bk2204:
Can you help me understand why `xargs` isn't sufficient here?  It seems like that should work just fine in most cases, and I know it's included with Git for Windows, so it's available even on Windows.

rcoup:
Use case: safely setting up piping via xargs is annoying when git-lfs is being called from Python.

commands/command_push.go
```diff
@@ -44,15 +45,31 @@ func pushCommand(cmd *cobra.Command, args []string) {
...
+		scanner := bufio.NewScanner(os.Stdin) // line-delimited
+		for scanner.Scan() {
+			line := scanner.Text()
+			if line != "" {
+				argList = append(argList, line)
+			}
+		}
+		if err := scanner.Err(); err != nil {
+			ExitWithError(errors.Wrap(err, tr.Tr.Get("Error reading from stdin:")))
+		}
```
bk2204, line 60: Can we produce an error here if there are other command-line arguments, since we'll be ignoring them?

bk2204 (commented):
This looks like a reasonable feature to add.  I had one request for an improvement.

commands/command_push.go
rcoup: Done 😃

bk2204 (approved):
This looks good to me.  Thanks again for the patch.

PR #5136, opened 2022-10-05
ssh: disable concurrent transfers if no multiplexing

### LINKED ISSUE #5064: Avoid issuing multiple SSH connections with pure SSH protocol if multiplexing is not enabled
As mentioned in #5062, we can end up with some cases where multiplexing is not enabled.  In such a situation, we'd want to avoid issuing multiple connections, since it can be very expensive to create and tear down such connections and might prompt the user for the SSH key passphrase multiple times.

bk2204:
Right now, we try to spawn multiple SSH connections using ControlMaster when making multiple requests.  However, if multiplexing is not enabled, we can spawn multiple actual connections, which can be expensive and require lots of authentication requests.

Instead, let's make sure we don't enable multiple connections if multiplexing is not enabled, since this may cause the user to be prompted for multiple SSH key connection attempts (say, if they're using a security key) and is substantially more heavyweight than simply spawning a new process over the same connection.

Fixes #5064

### DIFF
--- a/ssh/ssh.go
+++ b/ssh/ssh.go
@@ -30,7 +30,7 @@ type SSHMetadata struct {
 	Path        string
 }
 
-func FormatArgs(cmd string, args []string, needShell bool) (string, []string) {
+func FormatArgs(cmd string, args []string, needShell bool, multiplex bool) (string, []string) {
 	if !needShell {
 		return cmd, args
 	}
@@ -38,12 +38,12 @@ func FormatArgs(cmd string, args []string, needShell bool) (string, []string) {
 	return subprocess.FormatForShellQuotedArgs(cmd, args)
 }
 
-func GetLFSExeAndArgs(osEnv config.Environment, gitEnv config.Environment, meta *SSHMetadata, command, operation string, multiplexDesired bool) (string, []string) {
-	exe, args, needShell := GetExeAndArgs(osEnv, gitEnv, meta, multiplexDesired)
+func GetLFSExeAndArgs(osEnv config.Environment, gitEnv config.Environment, meta *SSHMetadata, command, operation string, multiplexDesired bool) (string, []string, bool) {
+	exe, args, needShell, multiplexing := GetExeAndArgs(osEnv, gitEnv, meta, multiplexDesired)
 	args = append(args, fmt.Sprintf("%s %s %s", command, meta.Path, operation))
-	exe, args = FormatArgs(exe, args, needShell)
+	exe, args = FormatArgs(exe, args, needShell, multiplexing)
 	tracerx.Printf("run_command: %s %s", exe, strings.Join(args, " "))
-	return exe, args
+	return exe, args, multiplexing
 }
 
 // Parse command, and if it looks like a valid command, return the ssh binary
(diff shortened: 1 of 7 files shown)


ssh/ssh.go
```diff
@@ -30,20 +30,20 @@ type SSHMetadata struct {
 	Path        string
 }
 
-func FormatArgs(cmd string, args []string, needShell bool) (string, []string) {
+func FormatArgs(cmd string, args []string, needShell bool, multiplex bool) (string, []string) {
```
chrisd8088, line 33: I see the `multiplex` argument added here but I don't see it being used ... am I missing something here?

ssh/ssh.go
```diff
@@ -155,10 +155,12 @@ func GetExeAndArgs(osEnv config.Environment, gitEnv config.Environment, meta *SS
 		args = append(args, "-batch")
 	}
 
+	isMultiplexing = false
```
chrisd8088, line 158: Since `GetLFSExeAndArgs()` uses the argument name `multiplexing`, I might be inclined to do the same here as well rather than `isMultiplexing`, but I recognize that's a bit of a bikeshed concern.

ssh/ssh_test.go
```diff
@@ -515,7 +515,7 @@ func TestSSHGetExeAndArgsInvalidOptionsAsPath(t *testing.T) {
 	assert.Equal(t, "git@git-host.com", e.SSHMetadata.UserAndHost)
 	assert.Equal(t, "/-oProxyCommand=gnome-calculator", e.SSHMetadata.Path)
 
-	exe, args, needShell := ssh.GetExeAndArgs(cli.OSEnv(), cli.GitEnv(), &e.SSHMetadata, false)
+	exe, args, needShell, _ := ssh.GetExeAndArgs(cli.OSEnv(), cli.GitEnv(), &e.SSHMetadata, false)
```
chrisd8088, line 518: All of the existing tests just ignore the new return value ... should we add some tests which confirm its true/false state in particular conditions?

chrisd8088 (commented):
I only had a few notes, and I notice the Windows CI jobs seems to be failing, so maybe that's something worth review as well (but perhaps it's not related to this PR)?

ssh/ssh.go
bk2204: No, other than in the tests and some of the code we pass all of the arguments directly from the return value of another function into this function, and it makes it a lot easier to not update all the tests to just pass the extra argument and not strip it off.

ssh/ssh.go
bk2204: Okay, sounds good.

ssh/ssh_test.go
bk2204: I can look into that.  Part of this depends on whether we can create a specific temporary directory, which is potentially tricky and I was concerned might lead to flaky tests, but I'll see what I can do to try to make that less of a problem.

bk2204:
Okay, the test failure has magically disappeared, I've added tests (and some additional commits since I discovered a bug), and this should be ready for another review when that's convenient.

ssh/ssh.go
chrisd8088: Got it, thanks!

ssh/ssh_test.go
chrisd8088: Thanks for the extra test; looks good!

chrisd8088 (approved):
This looks great, thanks!

PR #5153, opened 2022-10-27
Fixed broken link for git-lfs-migrate

dhiwakarK:
Fixed broken link for git-lfs-migrate

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ $ git commit -m "add psd"
 > **Note that this will rewrite history and change all of the Git object IDs in your
 > repository, just like the export version of this command.**
 >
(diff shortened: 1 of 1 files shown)


dhiwakarK:
I was also wondering why the link title should say `git-lfs-migrate(1)` instead of just `git-lfs-migrate` 🤔

chrisd8088 (approved):
Thank you for the PR; this looks good!

And as for your question about the `(1)`, that's a Unix convention to indicate the section of the Unix manual pages to which the command belongs.  See, for reference, the Wikipedia [entry](https://en.wikipedia.org/wiki/Man_page#Manual_sections) and sites like man7.org, whose [entry](https://man7.org/linux/man-pages/man1/man.1.html) on the `man(1)` command provides good background.

Issue #3048, opened 2018-06-06
git lfs clone aborts with "Skipping object checkout, Git LFS is not installed"


DavidGriffith:
Tried to do ```git lfs clone https://github.com/gewtnet/tops10-disks```.  This proceeded to download about 487 megabytes of stuff into ```.git/lfs``` whereupon it aborted with ```Skipping object checkout, Git LFS is not installed.```.  But git LFS **IS** in fact installed.  ```git lfs pull``` also responds with the complaint that Git LFS isn't installed and does nothing else.

DavidGriffith:
Yes, it does.  The instructions I was following didn't mention that step.

ttaylorr:
Hi @DavidGriffith, I'm glad that you were able to resolve this issue. @robobenklein: thanks for stepping in! My apologies for the delayed response, I was out of the office last week and have been catching up on email since.
 
Please don't hesitate to let me know if there is anything else that I can help you out with in the future by sending me a ping here or elsewhere. Thanks!

bomgar:
It would be great to mention `git lfs install` in the error message

rkost:
Another way would be modifying the message to be less ambiguous: `Git LFS is not installed in this repository.` I honestly thought git would not find my LFS extension installation on my system (to be fair, I am completely new to LFS and just came across a repo that uses it).

suraj-mathew:
Another observation i found is that, if we run **git lfs install** using a root user, and if we try to use it from another user (say X) on the same machine, it wont work. User X will end up in the getting the error message "Git LFS is not installed"

To overcome this issue, we need to 

- either run **git lfs install** again from the respective user's terminal (here X's)
- or run **git lfs install --system** from the root user initially

Talgin:
> Does running `git lfs install` again after cloning fix the issue? It's probably talking about the repo-specific installation.

For newbies who met this problem for the first time the order of operation would look like:
- Download and install Git LFS (git-lfs.com)
- Go to your cloned git repo and inside it run `git lfs install`
- Then run `git pull`
- Now you can check the sizes of your files which should be the same as in your git repo (top left corner before the words "Stored with Git LFS").

thecadams:
> It would be great to mention git lfs install in the error message

Hello folks. 5 years later and this is still true.

PR #5172, opened 2022-11-09
Add information to ambiguous error message.

WhatTheFuzz:
Users could confuse the previous error message with a shell path issue, as though it could not resolve the installation of git-lfs. The proposed message in this commit offers additional insight by:

- Detailing that LFS has not been installed for this repository.
- Offering a fix (git lfs install) for the error.

Relate issue #3048. Judging by the continued addition of upvotes, even after the issue has been closed, users are still running into this problem.

Passing local test through the use of `make test`.

### DIFF
--- a/commands/command_pull.go
+++ b/commands/command_pull.go
@@ -113,7 +113,7 @@ func pull(filter *filepathfilter.Filter) {
 	}
 
 	if singleCheckout.Skip() {
-		fmt.Println(tr.Tr.Get("Skipping object checkout, Git LFS is not installed."))
+		fmt.Println(tr.Tr.Get("Skipping object checkout, Git LFS is not installed for this repository. Consider installing it with 'git lfs install'."))
 	}
 }
 
--- a/commands/command_pull.go
+++ b/commands/command_pull.go
@@ -113,7 +113,8 @@ func pull(filter *filepathfilter.Filter) {
 	}
 
 	if singleCheckout.Skip() {
-		fmt.Println(tr.Tr.Get("Skipping object checkout, Git LFS is not installed for this repository. Consider installing it with 'git lfs install'."))
+		fmt.Println(tr.Tr.Get("Skipping object checkout, Git LFS is not installed for this repository.\n" +
+		                      "Consider installing it with 'git lfs install'."))
 	}
 }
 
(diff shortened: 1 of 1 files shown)


commands/command_pull.go
```diff
@@ -113,7 +113,7 @@ func pull(filter *filepathfilter.Filter) {
 	}
 
 	if singleCheckout.Skip() {
-		fmt.Println(tr.Tr.Get("Skipping object checkout, Git LFS is not installed."))
+		fmt.Println(tr.Tr.Get("Skipping object checkout, Git LFS is not installed for this repository. Consider installing it with 'git lfs install'."))
```
bk2204, line 116: I think this text is clearly an improvement, but I think it's longer than 80 characters, which might make it a bit difficult to read in some cases.  Could we insert a newline here after `repository.` to make it easier to read on systems with small screens?

bk2204 (commented):
Hey, thanks for the patch, and welcome to Git LFS!  I think this is clearly a helpful improvement, and I had just one suggestion to make it a little easier to read.

commands/command_pull.go
WhatTheFuzz: I totally agree. I amended the PR with a multiline message. This makes it easier to read for both developers and end-users. However, I couldn't find an example of a multiline string in the project. If you prefer backticks, I can adjust it to that as well. Thank you for the warm welcome.

commands/command_pull.go
```diff
@@ -113,7 +113,8 @@ func pull(filter *filepathfilter.Filter) {
 	}
 
 	if singleCheckout.Skip() {
-		fmt.Println(tr.Tr.Get("Skipping object checkout, Git LFS is not installed."))
+		fmt.Println(tr.Tr.Get("Skipping object checkout, Git LFS is not installed for this repository.\n" +
+		                      "Consider installing it with 'git lfs install'."))
```
bk2204, line 117: I think one of these lines has trailing whitespace, which is causing CI to fail.

While you're making that change, I think it will be easier for the code which extracts strings for translation if we leave it as one long string on one line, just with a newline in the middle.

WhatTheFuzz: Easy. I amended the snippet to just one line with the newline character. Sorry it took three reviews for a one-line string. :)

bk2204 (approved):
This looks good.  Thanks so much for the patch.

PR #5187, opened 2022-11-15
Use --sparse parameter for ls-files for performance optimization

Leo1690:
When lfs is used with a huge mono repository with sparse-checkout, the lfs post-commit command took a long time. Per default git ls-files, which excute during the post-commit, retrieves all file paths, irrespectevly of the sparse set. When sparse-index is used ls-files respects the --sparse parameter, which omits the irrelevant files that are not in the sparse set, resulting in a performance improvement.

### DIFF
--- a/git/ls_files.go
+++ b/git/ls_files.go
@@ -29,6 +29,10 @@ func NewLsFiles(workingDir string, standardExclude bool, untracked bool) (*LsFil
 		"--cached",
 	}
 
+	if IsGitVersionAtLeast("2.35.0") {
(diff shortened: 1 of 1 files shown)


bk2204 (approved):

This seems fine to me.  I noted the `--sparse` option lists directories for sparse directories instead of files, but since the only two places we use this function are in locking (where directories cannot be locked) and for finding `.gitattributes` files, which will never match, I think this is okay here and we shouldn't have any problems.

PR #5223, opened 2022-12-15
fix(ssh): use /tmp to place control dir on darwin

aymanbagabas:
OpenSSH complains if the control path is too long (>=104) On Darwin, os.MkdirTemp is too long, use /tmp instead.

ControlPath too long ('/var/folders/v3/5pnzpycs0pzdpytmgfzc7tj80000gn/T/sock-96855798/sock-b677e145501e220cfb5583ba1638a77ffe779df9' >= 104 bytes)

### DIFF
--- a/ssh/ssh.go
+++ b/ssh/ssh.go
@@ -2,10 +2,10 @@ package ssh
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strings"
 	"syscall"
 
@@ -110,9 +110,14 @@ func findRuntimeDir(osEnv config.Environment) string {
 }
 
 func getControlDir(osEnv config.Environment) (string, error) {
+	tmpdir, pattern := "", "sock-*"
+	if runtime.GOOS == "darwin" {
+		// On Darwin, the default temporary directory results in a socket path that's too long.
+		tmpdir = "/tmp"
+	}
 	dir := findRuntimeDir(osEnv)
 	if dir == "" {
-		return ioutil.TempDir("", "sock-*")
+		return os.MkdirTemp(tmpdir, pattern)
 	}
 	dir = filepath.Join(dir, "git-lfs")
 	err := os.Mkdir(dir, 0700)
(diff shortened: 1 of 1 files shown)


ssh/ssh.go
```diff
@@ -103,6 +103,10 @@ func getVariant(osEnv config.Environment, gitEnv config.Environment, basessh str
 // findRuntimeDir returns a path to the runtime directory if one exists and is
 // guaranteed to be private.
 func findRuntimeDir(osEnv config.Environment) string {
+	switch runtime.GOOS {
+	case "darwin":
+		return "/tmp"
```
bk2204, line 108: Can we add a comment here for future developers?  Maybe something like:

```
// On Darwin, the default temporary directory results in a socket path that's too long.

File: ssh/ssh.go [hunk 2]
```diff
@@ -121,7 +125,7 @@ func getControlDir(osEnv config.Environment) (string, error) {
 		// os.ErrExist, but that's not available on Go 1.11.
 		perr, ok := err.(*os.PathError)
 		if !ok || perr.Err != syscall.EEXIST {
-			return ioutil.TempDir("", "sock-*")
+			return os.MkdirTemp("", "sock-*")
```
bk2204, line 128: I think we still have a problem here.  If we set `dir` to `/tmp`, we then try to create `/tmp/git-lfs`, and if it fails, we use `$TMPDIR` to create a random path, which is still going to be too long.  That means the second time the same user runs Git LFS, or the first time a different user runs Git LFS, `/tmp/git-lfs` will fail to be created and we'll be back to fix this all over again.

bk2204 (commented):
Hey, thanks for the patch.  I think there are still a few things we need to address here before going forward.

ssh/ssh.go
aymanbagabas: In that case, I think we should bubble up the error instead of suppressing it.

On another note, what do you think of using the user's cache directory for storing the control socket? We could use `os.UserCacheDir()`

aymanbagabas: I've updated the patch to create a random `sock-*` dir under `/tmp` on darwin using `os.MkdirTemp`

chrisd8088:
Just FYI, most of the Git LFS core team is away on holidays, so it might take a few weeks for this to be reviewed again.

chrisd8088 (approved):
This looks good to me now, as it seems tightly constrained to just the `getControlDir()` function.  But I know brian was looking at it too, so I'll leave it for him to review as well after the holidays.

Thanks again for the patch!

bk2204 (approved):

PR #5228, opened 2022-12-19
Check for github action updates monthly

jlosito:
This will use dependabot to check any github action which this project uses for updates in order to keep it up-to-date.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot


chrisd8088:
Hey, thanks for the PR and welcome to Git LFS!

This seems like a good idea, in general, but I am concerned we may see a number of issues reported about our use of the `actions/setup-ruby@v1` action rather than the `ruby/setup-ruby@v1` one.  However, as I reported in ruby/setup-ruby#293, the latter changes `PATH` in such a way as to cause the system default Git LFS to run instead of our newly-built `git-lfs.exe` binary which we want to test.  Fixing that did not seem trivial, so for the moment we are sticking with the older action, although it's deprecated.

chrisd8088:
Just as a note for the future, if we adopt this change, we should also make a parallel changes to the Actions workflows in `git-lfs/gitobj` and `git-lfs/wildmatch`.

jlosito:
@chrisd8088 should I add an ignore section for the action which you are referring to so that the bot leaves that action alone?

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore

chrisd8088:
I've opened #5236 which upgrades our use of `actions/setup-ruby@v1` to `ruby/setup-ruby@v1`.  The core team is mostly going to be on vacation for a while, so it might take a while to be approved to merge.  However, once that's done (and maybe a few other stale Actions are updated too, in other PRs), I think this should be good to go.

Thanks again for the PR and for your patience!

chrisd8088:
I've also added #5243 which should upgrade all the other Actions steps to the latest available versions.  It's currently built on top of #5236 so I'll wait to merge that one first before taking the new one out of draft mode.

chrisd8088 (approved):
Looks good and I think we've now updated all our workflow steps in advance, so I'll merge this once CI is green.  Thanks again!

PR #5230, opened 2022-12-19
fix(tq): stop adding transfers to queue after abort

aymanbagabas:
When the transfer gets aborted, dangling goroutines might add more transfers to the queue resulting in a wait deadlock.
Conditionally add transfers to queue if abort flag is not set.

Using a pure-SSH approach makes the client hang unable to abort.

Signed-off-by: Ayman Bagabas <ayman.bagabas@gmail.com>

### DIFF
--- a/tq/transfer_queue.go
+++ b/tq/transfer_queue.go
@@ -142,6 +142,7 @@ type abortableWaitGroup struct {
 	wq      sync.WaitGroup
 	counter int
 	mu      sync.Mutex
+	abort   bool
 }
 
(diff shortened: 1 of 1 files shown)


chrisd8088:
Thanks for the patch!  This looks reasonable to me, so I'm going to merge it.

While looking at this I realize it would be nice to have some tests for the `abortableWaitGroup` functions in the `tq/transfer_queue_test.go` Go tests, at least, but I don't think we need to hold this up on that point.

chrisd8088 (approved):
Looks good, thanks!

PR #5236, opened 2022-12-22
Upgrade GitHub Actions workflows to use `ruby/setup-ruby@v1`

chrisd8088:
As the `actions/setup-ruby` GitHub Action is deprecated in favour of the `ruby/setup-ruby` Action, we upgrade to use the latter in our CI and release workflows.

As testing the `release.yml` workflow is not convenient without generating a release, it has instead been run successfully on a separate, private repository cloned from this one.

#### MSYS2 Build Environment

In the Windows builds, the `ruby/setup-ruby@v1` Action [installs](https://github.com/ruby/setup-ruby/blob/4b2d1d631efa087f8896c15a0c6023dc2f483198/windows.js#L76-L78) the MSYS2 build tools and [sets](https://github.com/ruby/setup-ruby/blob/4b2d1d631efa087f8896c15a0c6023dc2f483198/index.js#L175-L182) some environment variables which persist for subsequent steps in the workflow, such as the `TMPDIR` variable.

These differences from the `actions/setup-ruby` Action introduce failures into our test suite, principally by [causing](https://github.com/ruby/setup-ruby/blob/4b2d1d631efa087f8896c15a0c6023dc2f483198/common.js#L229-L231) Actions to run the Bash shell in `C:\msys64\usr\bin` instead of the one in `C:\Program Files\Git\bin`, which in turn results in the `git` command running the `git.exe` binary found in `C:\Program Files\Git\bin` instead of `C:\Program Files\Git\mingw64\bin`.

The Git binary in `C:\Program Files\Git\bin`, when asked to run a `git lfs` command, then finds the `git-lfs.exe` binary in `C:\Program Files\Git\cmd` instead of the one we have built, despite our [changing](https://github.com/git-lfs/git-lfs/blob/4140f638e9e305605f2e8cd941355340bb71fbf0/t/testenv.sh#L76-L77) the `PATH` environment variable so that our built `git-lfs.exe` binary should be found first.

We resolve this problem by renaming the `C:\msys64` directory with a temporary name after installing the `asciidoctor` Ruby gem.

Also note that we do not need to specify the Bash shell to run the `gem install acsiidoctor` step, and that we move this step below the `ruby/setup-ruby@v1` step in all cases so as to make the dependency more clear.  (We also move the `gem install packagecloud-ruby` step into the same relative location, for the same reason.)

#### Temporary Files

The `ruby/setup-ruby@v1` Action also sets the `TMPDIR` environment variable for all subsequent steps, and this results in other test suite failures on Windows because we [create](https://github.com/git-lfs/git-lfs/blob/4140f638e9e305605f2e8cd941355340bb71fbf0/t/testenv.sh#L87-L91) paths for our temporary test [repositories](https://github.com/git-lfs/git-lfs/blob/4140f638e9e305605f2e8cd941355340bb71fbf0/t/testhelpers.sh#L415) in our `GIT_LFS_TEST_DIR` path, which we [set](https://github.com/git-lfs/git-lfs/blob/4140f638e9e305605f2e8cd941355340bb71fbf0/t/testenv.sh#L82) using `mktemp(1)`, and that derives its default path from `TMPDIR`.  However, some of our test helper programs such as `lfstest-customadapter` [use](https://github.com/git-lfs/git-lfs/blob/4140f638e9e305605f2e8cd941355340bb71fbf0/t/cmd/lfstest-customadapter.go#L130) the Go language `ioutil.TempFile()` function to construct paths, which [calls](https://pkg.go.dev/io/ioutil@go1.19.4#TempFile) `os.TempDir()`, and on Windows that does not [refer](https://pkg.go.dev/os@go1.19.4#TempDir) to the `TMPDIR` environment variable but the `TMP` and `TEMP` ones instead.

The consequence is that tests using those helper programs try to move files between directories creating using both techniques of temporary path creation, and these return paths on different volumes, so the test fails when trying to move files between `D:\a\_temp` and `C:\Users\RUNNER~1\AppData\Local\Temp`.

Therefore we take care to unset the `TMPDIR` variable before running commands which might make use of `mktemp(1)`.

Thanks to @MSP-Greg for the suggestion to rename `C:\msys64`!

/cc ruby/setup-ruby#293
/cc #5228


.github/workflows/ci.yml
```diff
@@ -51,7 +51,9 @@ jobs:
     runs-on: windows-latest
     steps:
     - uses: actions/checkout@v1
-    - uses: actions/setup-ruby@v1
+    - uses: ruby/setup-ruby@v1
+    - run: gem install asciidoctor
+    - run: Rename-Item -Path C:\msys64 -NewName msys64-tmp -Force
```
larsxschneider, line 56: Minor nit: I wonder if it would be useful to add a short comment here to explain the rename. Reading this line without the context of this PR would make me 🤔 in a few months 🤣

larsxschneider (approved):

.github/workflows/ci.yml
chrisd8088: Good idea, thanks!  Done!

Issue #5237, opened 2022-12-22
git lfs migrate export way too long without userfeedback


axkibe:
**Describe the issue**
I'm running ```git lfs migrate export --include="*" --everything```
on a very large repository.

First the process does a lot of CPU (without any userfeedback what it is doing)

then it seemd to hang a long way on following syscall (acording to strace)
futex(0xe488a8, FUTEX_WAIT_PRIVATE, 0, NULL

Now it resumed after many minutes.. (i guess, because it's multithreaded).

Could you maybe give the user some head ups that something is happening and what is happening?

BTW: Is this the correct way to leave LFS for a repository (don't worry, I'm staying with LFS but I want this repository LFS-less for testing/developing purposes).

I tried just master instead --everything .. but that does nothing, could you add an export everything example to Readme (for maybe only one ref)? I'm pretty confused how export is working.

chrisd8088:
Hey, I'm sorry you're having trouble.  I would definitely recommend reading through the `git-lfs-migrate(1)` [manual page](https://github.com/git-lfs/git-lfs/blob/4140f638e9e305605f2e8cd941355340bb71fbf0/docs/man/git-lfs-migrate.adoc), if you haven't already; there's a section there specifically about the `git lfs migrate export` sub-command.

At a high level, `git lfs migrate export`, like `git lfs migrate import`, will rewrite your Git history (changing commit SHAs in the process) to remove or negate Git LFS entries in `.gitattributes` files and change any Git LFS objects matching those entries' patterns into regular Git objects.

If you're using Git LFS in a repository and want to "undo" some or all of the rules in your `.gitattributes` files, and convert the corresponding Git LFS objects into regular Git blobs, then yes, `git lfs migrate export` is an effective tool, so in that regard it's a recommended way to stop using Git LFS.  That said, note that just like the `import` mode, the `export` mode will rewrite your Git history, so you'll want to force-push all the branches and tags of your repository to your remote (assuming things look good once the command is finished).

The `--everything` option means "rewrite history on all refs", as opposed to just the current ref.  The `--include="*"` option is going to cause a final `* !text !filter !merge !diff` rule to be added to any `.gitattributes` files, which should override any other patterns which assign specific file extensions or paths to Git LFS.  That should work, but you might also prefer a more targeted set of rules, in which case providing one or more specific patterns like `--include="*.jpg"` will work.

[...]

PR #5239, opened 2022-12-23
Add `git lfs migrate export` command examples to manual page

chrisd8088:
We add several examples of the use of the `git lfs migrate export` subcommand to the `git-lfs-migrate(1)` manual page, showing use of both the `--include` option and the `--remote` option.

/cc @axkibe
/cc https://github.com/git-lfs/git-lfs/issues/5237#issuecomment-1363811700


bk2204 (approved):
Looks good.  I'm wholly in favour of making our documentation better to answer common user questions.

PR #5245, opened 2023-01-04
Make hooks refer to `core.hookspath`

bk2204:
In some cases, users may use `core.hooksPath` to locate their hooks in a different location.  However, right now, our hook and installation instructions mention only `.git/hooks`.

We can update the instructions without a problem, but we don't want to hard-code the hook path into the hooks themselves because of things like symlinks and moved repository, plus the fact that we can't update a hook automatically unless it's identical minus some whitespace changes.  To avoid spuriously failing to update a hook, let's print the right location with the instructions, and just mention `core.hookspath` and `.git/hooks` in the message, leaving it to the user to discover.

While we're at it, fix a small portion of the test which wasn't testing what we thought it was in a preparatory commit.  The gory details are mentioned in the commit message.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -150,10 +150,12 @@ func getHookInstallSteps() string {
 		ExitWithError(err)
 	}
 	hooks := lfs.LoadHooks(hookDir, cfg)
+	hookDir = filepath.ToSlash(hookDir)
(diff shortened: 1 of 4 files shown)


chrisd8088 (approved):
Great, thank you!

PR #5282, opened 2023-02-10
t: avoid incorrect negated commands

### LINKED ISSUE #5183: Bare negated `grep` commands in tests do not catch errors
**Describe the bug**

In a number of our tests, such [several](https://github.com/git-lfs/git-lfs/blob/4e767fd9540426ecd31a297ac6ece539dbbff8a3/t/t-standalone-file.sh#L458) [places](https://github.com/git-lfs/git-lfs/blob/4e767fd9540426ecd31a297ac6ece539dbbff8a3/t/t-standalone-file.sh#L468) in the `t/t-standalone-file.sh` and [others](https://github.com/git-lfs/git-lfs/blob/4e767fd9540426ecd31a297ac6ece539dbbff8a3/t/t-fetch.sh#L98), we use a construction such as the following:
```
  # We should not use the custom adapter process here.
  ! grep "xfer: started custom adapter process" push.log
```

However, unless such a line happens to appear as the last line of a test (as it does in [some](https://github.com/git-lfs/git-lfs/blob/4e767fd9540426ecd31a297ac6ece539dbbff8a3/t/t-attributes.sh#L37) [places](https://github.com/git-lfs/git-lfs/blob/4e767fd9540426ecd31a297ac6ece539dbbff8a3/t/t-track.sh#L77)) the test will not actually be sensitive to whether the string appears in a given log file or not.

This is because the `-e` shell attribute set at the beginning of most of our tests is documented in the `set` subsection of the `SHELL BUILTIN COMMANDS` section of `bash(1)` manual [page](https://www.man7.org/linux/man-pages/man1/bash.1.html#SHELL_BUILTIN_COMMANDS) as making the shell or subshell exit immediately if a command pipeline exists with a non-zero status, but only under certain conditions, including that the command's return value is not inverted with `!`:

> The shell does not exit if the command that fails is part of the command list immediately following a `while` or `until` keyword, part of the test following the `if` or `elif` reserved words, part of any command executed in a `&&` or `||` list except the command following the final `&&` or `||`, any command in a pipeline but the last, or if the command's return value is being inverted with `!`.

We should ensure that all occurrences of this usage of a negated `grep` command in our tests are replaced with a suitable alternative construction, such as either `grep "string" file.log && false` or `[ ! grep "string" file.log ]`.

We may also want to introduce a linter check of some type to help us avoid use of this construction in our tests in the future.

**To Reproduce**

The following example explicitly inserts the string `xfer: started custom adapter process` into the `push.log` file which the subsequent `grep` [command](https://github.com/git-lfs/git-lfs/blob/4e767fd9540426ecd31a297ac6ece539dbbff8a3/t/t-standalone-file.sh#L458) in `t/t-standalone-file.sh` then searches for that string, with the result from that `grep` then inverted with `!`.  The test still succeeds, though, as shown.

```
$ git clone https://github.com/git-lfs/git-lfs.git
$ cd git-lfs
$ make
$ cd t
$ ./t-standalone-file.sh
ok 1 - standalone-file-upload-download-bare ...                    
ok 2 - standalone-file-upload-download-non-bare ...                
ok 3 - standalone-file-download-missing-file ...                   
ok 4 - standalone-file-clone ...                                   
ok 5 - standalone-file-local-path ...                              
ok 6 - standalone-file-lfs.url file URL ...                        
ok 7 - standalone-file-lfs.url http URL ...                        
1..7

$ sed -ie 's/adapter process here\./adapter process here.\n  echo "xfer: started custom adapter process" >>push.log/' \
  t-standalone-file.sh
$ git diff -U1 t-standalone-file.sh
--- a/t/t-standalone-file.sh
+++ b/t/t-standalone-file.sh
@@ -457,2 +457,3 @@ begin_test "standalone-file-lfs.url http URL"
   # We should not use the custom adapter process here.
+  echo "xfer: started custom adapter process" >>push.log
   ! grep "xfer: started custom adapter process" push.log

$ rm -rf remote test_count
$ ./t-standalone-file.sh
ok 1 - standalone-file-upload-download-bare ...                    
ok 2 - standalone-file-upload-download-non-bare ...                
ok 3 - standalone-file-download-missing-file ...                   
ok 4 - standalone-file-clone ...                                   
ok 5 - standalone-file-local-path ...                              
ok 6 - standalone-file-lfs.url file URL ...                        
ok 7 - standalone-file-lfs.url http URL ...                        
1..7
```

**Expected behaviour**

The tests should fail if the strings for which they are searching appear in given log files.

**System environment**

The example above was performed on macOS, but should behave the same way on any platform.

**Output of `git lfs env`**

```
git-lfs/3.2.0 (GitHub; darwin amd64; go 1.19.1; git b64865c0)
git version 2.33.0
```

**Additional context**

We resolved this problem in several places in the `t/t-path.sh` test suite in commit 11092ef2b17eaa67f3363edc68b59331b979f7ee.  The description for that commit includes a relevant explanation of the issue:
```
In the final checks in both tests we search for the text string
"exploit" in the output log file captured after running a Git LFS
command, using the shell command "! grep -q 'exploit' output.log".
The "!" reverses the exit code from "grep", so if the word "exploit"
is found, the test should fail.  This works in the first test because
the command is the last one in the test, so the inverted exit code
from "grep" is returned as the exit code from whole test's subshell.

However, in the second test several other commands follow this command,
and because "set -e" (which is standard at the top of every test)
ignores commands' exit codes when they are inverted with "!", the
test proceeds even if the word "exploit" is seen in the output log.

To resolve this problem we instead use a command pipeline and ensure
that when the "grep" succeeds, the exit code from the final command
in the pipeline is generated by "false".  This successfully causes
the test to fail immediately when the word "exploit" is seen in the
output log file.
```

bk2204:
When `set -e` is enabled, not all commands trigger an error exit if they return false.  For example, it's clear that commands in an `if` or `while` statement don't cause an error if they are false.

What is less obvious, however, is that negated commands and negated pipelines also have no effect on `set -e`.  From POSIX 1003.1-2017 (on `sh -e`):

    When this option is on, if a simple command fails for any of the
    reasons listed in Consequences of Shell Errors or returns an exit
    status value >0, and is not part of the compound list following a
    while, until, or if keyword, and is not a part of an AND or OR list,
    and is not a pipeline preceded by the ! reserved word, then the
    shell shall immediately exit.

As such, writing something like `! grep` will never fail.  Fortunately, we can append `&& exit 1` instead of the `!` and that will work correctly.

To make this work, run the following command to make the code properly check the exit status of our commands:

```
git grep -l '! [a-z]' t | \
  xargs ruby -pi -e '$_.gsub!(/^(\s+)! ([a-z].*)$/, "\\1\\2 && exit 1")'
```

Fixes #5183

### DIFF
--- a/t/t-attributes.sh
+++ b/t/t-attributes.sh
@@ -34,7 +34,8 @@ begin_test "macros"
 
   cd dir
   git lfs track '*.bin' 2>&1 | tee track.log
-  ! grep '"*.bin" already supported' track.log
+  grep '"*.bin" already supported' track.log && exit 1
+  true
 
   # NOTE: At present we do not test that "git lfs track" reports
   #       "already supported" when it finds a pattern in a subdirectory's
--- a/t/t-checkout.sh
+++ b/t/t-checkout.sh
@@ -49,7 +49,7 @@ begin_test "checkout"
   [ "$contents" = "$(cat folder2/nested.dat)" ]
   grep "Checking out LFS objects: 100% (5/5), 95 B" checkout.log
   grep 'accepting "file1.dat"' checkout.log
-  ! grep 'rejecting "file1.dat"' checkout.log
+  grep 'rejecting "file1.dat"' checkout.log && exit 1
 
   # Remove the working directory
   rm -rf file1.dat file2.dat file3.dat folder1/nested.dat folder2/nested.dat
@@ -231,7 +231,7 @@ begin_test "checkout: conflicts"
     git commit -m "second"
 
     # This will cause a conflict.
-    ! git merge first
+    git merge first && exit 1
 
     git lfs checkout --to base.txt --base file1.dat
     git lfs checkout --to ours.txt --ours file1.dat
--- a/t/t-fetch.sh
+++ b/t/t-fetch.sh
@@ -95,7 +95,7 @@ begin_test "fetch (shared repository)"
   rm -rf .git/lfs/objects
 
   git lfs fetch 2>&1 | tee fetch.log
-  ! grep "Could not scan" fetch.log
+  grep "Could not scan" fetch.log && exit 1
   assert_local_object "$contents_oid" 1
 
   git lfs fsck 2>&1 | tee fsck.log
--- a/t/t-migrate-import.sh
+++ b/t/t-migrate-import.sh
@@ -846,7 +846,7 @@ begin_test "migrate import (--everything with tag pointing to tag)"
   assert_pointer "refs/heads/my-feature" "a.md" "$md_feature_oid" "30"
 
   git tag --points-at refs/tags/abc | grep -q def
-  ! git tag --points-at refs/tags/def | grep -q abc
+  git tag --points-at refs/tags/def | grep -q abc && exit 1
 
   assert_local_object "$md_main_oid" "140"
   assert_local_object "$md_feature_oid" "30"
--- a/t/t-post-commit.sh
+++ b/t/t-post-commit.sh
@@ -101,6 +101,7 @@ begin_test "post-commit does not enter submodules"
 
   git add *.dat
   GIT_TRACE=1 git commit -m "Committed large files" 2>&1 | tee output
-  ! grep -E 'filepathfilter:.*submodule/foo' output
+  grep -E 'filepathfilter:.*submodule/foo' output && exit 1
+  true
 )
 end_test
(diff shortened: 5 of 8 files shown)


chrisd8088 (approved):
Looks great, and catches all the spots I'd noticed, plus some `git` ones too.  Thank you!

PR #5284, opened 2023-02-13
locks: ensure local locks are synced on error

### LINKED ISSUE #5055: Local locks aren't updated correctly when lock/unlock fails on some files
The local locks seems to be not updated when lock/unlock command fails on some files. I have tried the following steps. Lets assume w have a repo that GIT-LFS is applied to and it has 2 files out1.txt and out2.txt

1. Run 'git lfs lock out1.txt'
2. Run 'git lfs locks --local' -> Now I can see that a lock is reported on out1.txt and that is correct
3. Run 'git lfs lock out1.txt out2.txt' -> The lock will fail on out1.txt because it is already locked. But will succeed on out2.txt
4. Run 'git lfs locks --local' -> I can see that out1.txt lock only is reported 

**Expected behavior**
I expect the lock of out2.txt to be reported too in step 4. Note when I run 'git lfs locks' in step 4 I can see that out2.txt lock is reported successfully.

Note : I am using GIT-LFS v3.0.2

bk2204:
On an error, we call `os.Exit(2)`.  However, the defer for the locking client that saves the local locks is not called in such a case, which can leave the locks out of sync even if some of the operation fails. If we're going to exit early, be sure we close the lock client to avoid this problem.

Fixes #5055

### DIFF
--- a/commands/command_lock.go
+++ b/commands/command_lock.go
@@ -62,6 +62,7 @@ func lockCommand(cmd *cobra.Command, args []string) {
 	}
 
 	if !success {
+		lockClient.Close()
 		os.Exit(2)
 	}
 }
--- a/commands/command_unlock.go
+++ b/commands/command_unlock.go
@@ -120,6 +120,7 @@ func unlockCommand(cmd *cobra.Command, args []string) {
 		}
 	}
 	if !success {
+		lockClient.Close()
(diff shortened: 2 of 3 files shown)


chrisd8088 (approved):
Great, thanks for the test, that explains a lot!

PR #5295, opened 2023-02-26
Error out for invalid ref name

philip-peterson:
I noticed that `git lfs push origin jibberishnamehere` returns `0` regardless of whether `jibberishnamehere` is a valid ref. This PR adds a check to see if all the refs you provided are valid, and if they aren't, errors out early.

### DIFF
--- a/commands/command_push.go
+++ b/commands/command_push.go
@@ -145,13 +145,26 @@ func lfsPushRefs(refnames []string, pushAll bool) ([]*git.RefUpdate, error) {
 		reflookup[ref.Name] = ref
 	}
 
+	suspiciousRefs := make([]string, 0)
+
 	refs := make([]*git.RefUpdate, len(refnames))
 	for i, name := range refnames {
 		if ref, ok := reflookup[name]; ok {
 			refs[i] = git.NewRefUpdate(cfg.Git, cfg.PushRemote(), ref, nil)
 		} else {
 			ref := &git.Ref{Name: name, Type: git.RefTypeOther, Sha: name}
+			suspiciousRefs = append(suspiciousRefs, name)
 			refs[i] = git.NewRefUpdate(cfg.Git, cfg.PushRemote(), ref, nil)
+		}
+	}
+
+	for _, suspiciousRef := range suspiciousRefs {
+		isValid, err := git.IsValidRef(suspiciousRef)
+		if err != nil {
+			return nil, errors.New(tr.Tr.Get("Unable to determine ref validity of %v. Error: %v", suspiciousRef, err))
+		}
+		if !isValid {
+			return nil, errors.New(tr.Tr.Get("Invalid ref: %v", suspiciousRef))
 		}
 	}
 
--- a/commands/command_push.go
+++ b/commands/command_push.go
@@ -159,11 +159,12 @@ func lfsPushRefs(refnames []string, pushAll bool) ([]*git.RefUpdate, error) {
 	}
 
 	for _, suspiciousRef := range suspiciousRefs {
-		isValid, err := git.IsValidRef(suspiciousRef)
+		resolvedRef, err := git.ResolveRef(suspiciousRef)
+
 		if err != nil {
 			return nil, errors.New(tr.Tr.Get("Unable to determine ref validity of %v. Error: %v", suspiciousRef, err))
 		}
-		if !isValid {
+		if resolvedRef == nil {
 			return nil, errors.New(tr.Tr.Get("Invalid ref: %v", suspiciousRef))
 		}
 	}
--- a/git/git.go
+++ b/git/git.go
@@ -14,6 +14,7 @@ import (
 	"io/ioutil"
 	"net/url"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"regexp"
 	"strconv"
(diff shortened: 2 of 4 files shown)


chrisd8088:
Hey, thanks for the patch, and welcome to the project!  This looks like a good thing to fix!

I notice that the `git lfs fetch` command, which also takes a ref, does validate it and returns a non-zero exit code if it's not valid.  It appears to do so by [calling](https://github.com/git-lfs/git-lfs/blob/91bac118ed88bc858612f4df69c35c342332b427/commands/command_fetch.go#L50-L53), which uses `git rev-parse`.  I might be inclined to suggest following the same approach here, just to keep things in line with each other.

That said, I don't know why `Panic()` is [used](https://github.com/git-lfs/git-lfs/blob/91bac118ed88bc858612f4df69c35c342332b427/commands/command_fetch.go#L52) instead of just `Exit()` or `ExitWithError()`.  This appears to stem from the earliest commits which created the command, and was then overlooked in commit 6ec4b57ed3dbe56d450d39bad177383691be9237 of PR #713 when some `Panic()` calls were converted to just `Exit()`.  (Our use of these various functions is inconsistent, I know.)

It looks like we have [two](https://github.com/git-lfs/git-lfs/blob/91bac118ed88bc858612f4df69c35c342332b427/t/t-fetch.sh#L631-L638) [tests](https://github.com/git-lfs/git-lfs/blob/91bac118ed88bc858612f4df69c35c342332b427/t/t-push.sh#L550-L557) for the use of invalid remotes with `git lfs fetch` and `git lfs push`, but we don't have comparable tests for those commands with invalid refs.  
so we should add those tests in this PR too.

Thanks again so much for working on this, and let us know we can help further!

chrisd8088 (commented):
We might be able to avoid the need to add the `git.IsValidRef()` function by using `git.ResolveRef()` in the same manner the `git lfs fetch` command does, for the same purpose of checking for valid refs.

It would also be great to add tests for invalid refs for both commands in the `t/t-push.sh` and `t/t-fetch.sh` scripts.

philip-peterson:
@chrisd8088 done!

chrisd8088:
Hey, thanks for the PR update!  We're heading into the holidays now, but I'll make sure to take a look and get back to you as soon as I can.  Thanks again very much!

commands/command_push.go
```diff
@@ -145,16 +145,30 @@ func lfsPushRefs(refnames []string, pushAll bool) ([]*git.RefUpdate, error) {
...
+		resolvedRef, err := git.ResolveRef(suspiciousRef)
+
+		if err != nil {
+			return nil, errors.New(tr.Tr.Get("Unable to determine ref validity of %v. Error: %v", suspiciousRef, err))
+		}
+		if resolvedRef == nil {
+			return nil, errors.New(tr.Tr.Get("Invalid ref: %v", suspiciousRef))
+		}
+	}
+
```
chrisd8088, line 171: This definitely resolves the problem and makes the new test in `t/t-push.sh` succeed, which is great (and the test fails without this change, so it's a good proof the fix works!)

Looking at the logic here more closely, I wonder if we could simply do this:
```patch
--- a/commands/command_push.go
+++ b/commands/command_push.go
@@ -157,6 +157,9 @@ func lfsPushRefs(refnames []string, pushAll bool) ([]*git.RefUpdate, error) {
 			refs[i] = git.NewRefUpdate(cfg.Git, cfg.PushRemote(), ref, nil)
 		} else {
 			ref := &git.Ref{Name: name, Type: git.RefTypeOther, Sha: name}
+			if _, err := git.ResolveRef(name); err != nil {
+				return nil, errors.New(tr.Tr.Get("Invalid ref argument: %v", name))
+			}
 			refs[i] = git.NewRefUpdate(cfg.Git, cfg.PushRemote(), ref, nil)
 		}
 	}
```

At first I thought we might replace the loop over the `suspiciousRefs` slice with a call to the `git.ResolveRefs()` [function](https://github.com/git-lfs/git-lfs/blob/baf40ac99850a62fe98515175d52df5c513463ec/git/git.go#L324-L335), which would do the loop for us.

Then I realized that since we're just going to report an error for the first ref which doesn't resolve, we don't really need to collect a list of them at all; we can just do the check right away while iterating over the ref names from the command line arguments.  So we can, I think, avoid the need to create `suspiciousRefs` in the first place.

Next I looked at [some](https://github.com/git-lfs/git-lfs/blob/baf40ac99850a62fe98515175d52df5c513463ec/commands/command_checkout.go#L101-L104) [other](https://github.com/git-lfs/git-lfs/blob/baf40ac99850a62fe98515175d52df5c513463ec/commands/command_migrate.go#L188-L191) [callers](https://github.com/git-lfs/git-lfs/blob/baf40ac99850a62fe98515175d52df5c513463ec/git/git.go#L327-L330) of `git.ResolveRef()` and concluded we don't need to check for a `nil` ref, just for a non-nil `err` return value.  That's because the function [checks](https://github.com/git-lfs/git-lfs/blob/baf40ac99850a62fe98515175d52df5c513463ec/git/git.go#L304-L306) for an empty response from `git rev-parse` and turns it into an error, and otherwise we can assume Git's response is valid enough to be parsed.

Finally, I think that if we use the same error message as the `git lfs fetch` command [generates](https://github.com/git-lfs/git-lfs/blob/baf40ac99850a62fe98515175d52df5c513463ec/commands/command_fetch.go#L52), we can ever-so-slightly simplify the job of a future contributor who's translating error messages for us, so they only have to translate one version of the same message rather than two.  This is, admittedly, a very minor concern.

Anyway, that's how I wound up at the suggestion above.  Would you be amenable to adopting that change?

chrisd8088 (commented):
Thanks so much for adopting the `git.ResolveRef()` function here and adding the tests; those are hugely appreciated.

I have a few more suggestions which I think will streamline the changes here.  Please do take a look and let me know if you're interested in adding them to the PR.

And thank you again very much for pursuing this patch!  The project will benefit from the improvement here, for certain.

chrisd8088:
I've merged PR #5639, which should address this and includes your work, @philip-peterson (with credit, of course!), so I'm going to close this PR out, if that's OK.  Thanks again for the bug report and contribution to the project!

PR #5300, opened 2023-03-01
Update Windows signing certificate hash

chrisd8088:
The Git LFS signing certificate for Windows binaries has changed, so we update the default signing certificate SHA-1 hash in the Makefile.

(The last update was in commit 16eec1724ef916ef1e9eb4be585c2dc27b7a7270 of PR #4946 in 2022.)

The commands used to generate this hash were:

```
$ openssl pkcs12 -info -in codesign.pfx -out codesign.pem
$ openssl x509 -text -in codesign.pem -fingerprint | \
  grep Fingerprint | sed 's/^SHA1 Fingerprint=//' | \
  sed 's/://g' | tr [:upper:] [:lower:]
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ TAR_XFORM_CMD ?= $(shell $(TAR) --version | grep -q 'GNU tar' && echo 's')
 
 # CERT_SHA1 is the SHA-1 hash of the Windows code-signing cert to use.  The
 # actual signature is made with SHA-256.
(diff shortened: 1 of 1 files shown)


chrisd8088:
Just FYI to @git-lfs/core reviewers -- the failing Windows CI check in `t/t-track.sh` appears to be a relatively recent phenomenon, which exists in `main` as well and is unrelated to this PR's change.

bk2204 (approved):

PR #5311, opened 2023-03-14
[WIP] Add shell completion generation

### LINKED ISSUE #5104: [feature] Shell completion
**Describe the issue**

How about add bash/zsh/fish/posh/etc completion? Just like other git subcommands?

```
❯ git add --<TAB>
option
--all                  add, modify, and remove index entries to match the working tree
--chmod                override the executable bit of the listed files
--dry-run              do not actually add files; only show which ones would be added
--edit                 open diff against index in editor
--force                allow adding otherwise ignored files
--ignore-errors        continue adding if an error occurs
--ignore-removal       like "--all" but ignore removals
--intent-to-add        record only that path will be added later
...
```

A test example:

```
❯ compdef '_arguments -S -s "*: :((checkout\:'\''Populate working copy with real content from Git LFS files.'\'' dedup\:'\''De-duplicate Git LFS files.'\''))"' git-lfs
_arguments -S -s "*: :((checkout\:'Populate working copy with real content from Git LFS files.' dedup\:'De-duplicate Git LFS files.'))" git-lfs
❯ git-lfs <TAB>
checkout   Populate working copy with real content from Git LFS files.
dedup      De-duplicate Git LFS files.
```

I guess <https://jmtirado.net/shell-completion-with-cobra-and-go/> and <https://github.com/cli/cli/blob/cf4b73ff958b272cf3c9c0cf9351459f76b793a0/pkg/cmd/completion/completion.go> may be helpful?

Thanks.

anihm136:
Fixes #5104. Currently still WIP - completions work correctly for a command named 'git-lfs', but not for 'git lfs'. Need to figure out how to modify the generated script to complete the subcommand correctly.

### DIFF
--- a/commands/run.go
+++ b/commands/run.go
@@ -111,12 +111,15 @@ PowerShell:
 		Run: func(cmd *cobra.Command, args []string) {
 			switch args[0] {
 			case "bash":
-				buf := new(bytes.Buffer)
-				cmd.Root().GenBashCompletion(buf)
-				buf.WriteString("_git_lfs() { __start_git-lfs; }\n") // this is needed for git bash completion to pick up the completion for the subcommand
-				buf.WriteTo(os.Stdout)
+				completion := new(bytes.Buffer)
+				cmd.Root().GenBashCompletion(completion)
+				completion.WriteString("_git_lfs() { __start_git-lfs; }\n") // this is needed for git bash completion to pick up the completion for the subcommand
+				completion.WriteTo(os.Stdout)
 			case "zsh":
-				cmd.Root().GenZshCompletion(os.Stdout)
+				completion := new(bytes.Buffer)
+				cmd.Root().GenZshCompletion(completion)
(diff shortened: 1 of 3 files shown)


bk2204:
Hey,

Is this still a WIP, or are you ready for review?

anihm136:
Still WIP, apologies if it notified you for a review. Should I convert it to a draft PR until it's ready?

bk2204:
Sure, converting to draft would be a good idea.  Since we've already been notified, please just mention in the PR body when this is ready for review and we'll try to get to it reasonably quickly.

Thanks for taking the time to send a pull request for this; I know a bunch of people want this feature.

anihm136:
@bk2204 this is ready for review now. I do not have access to PowerShell, so I am unable to manually verify if completions are working. Do you expect tests for this?

bk2204 (commented):
Thanks for the patch!

I noticed one thing we should fix for localization purposes, but otherwise I think this looks good.  If you think you can write a few tests for bash (we don't use zsh, fish, or PowerShell in the testsuite, so we can't test them), that would be great, but if it's not possible, then we can skip it.

anihm136:
Getting the list of completions for a command in bash seems to be tricky, so I'm leaving out the test for now. Please let me know if anything else is needed. Thanks!

bk2204 (approved):
This looks good.  Thanks again for the patch!

PR #5317, opened 2023-03-20
Fix Windows CI with Go 1.20

bk2204:
As some have recently noticed, our Windows CI has been broken due to the `t/t-track.sh` test failing, which has been caused by a change to our CI in Go 1.20 wherein it always “helpfully” includes a dot component for relative paths with no other directory.  This is not very helpful in our case, and patch 2 here fixes that problem.

Patch 3 also fixes a problem that had gone unnoticed by our testsuite but that is related.

As usual, these commits all have their own reasonable commit messages and explain independently why they're valuable.


chrisd8088 (approved):
Great detective work, thank you!

Issue #5316, opened 2023-03-20
git lfs migrate import --everything ignores remote refs


GLaN1K:
**Describe the bug**

`--everything` in `git lfs migrate import --everything` ignored all remote refs

**To Reproduce**

1. Clone repo which was prepared for the test:
```
git clone https://gitlab.com/nikita.savitsky1/lfs-migration-test
```

In it,  there are two branches `main` and `test_branch`:

```
glanik@Luna ~/lfs-migration-test:main 0> git show-ref
127eefa52cdb16bf4ad726aa1a83600f11ccf753 refs/heads/main
27db8eb4158f34493f0e238695274e6ea014cc51 refs/remotes/origin/HEAD
27db8eb4158f34493f0e238695274e6ea014cc51 refs/remotes/origin/main
27db8eb4158f34493f0e238695274e6ea014cc51 refs/remotes/origin/test_branch
```

2. Execute migration with `--everything`

```
 git lfs migrate import --everything --include="*.dat"
```

3. In output, only `main` was migrated, but `test_branch` was ignored:

```
migrate: Sorting commits: ..., done.
migrate: Rewriting commits: 100% (3/3), done.
  main  27db8eb4158f34493f0e238695274e6ea014cc51 -> 127eefa52cdb16bf4ad726aa1a83600f11ccf753
migrate: Updating refs: ..., done.
migrate: checkout: ..., done.
```

**Expected behavior**

All remote refs, including `test_branch` should have been migrated

**System environment**

M1 ARM CPU, MacOS "Ventura" Version 13.2.1 (22D68)

**Output of `git lfs env`**
```
glanik@Luna ~/lfs-migration-test:main 0> git lfs env
git-lfs/3.1.2 (GitHub; darwin arm64; go 1.17.6)
git version 2.30.1

Endpoint=https://gitlab.com/nikita.savitsky1/lfs-migration-test.git/info/lfs (auth=none)
  SSH=git@gitlab.com:nikita.savitsky1/lfs-migration-test.git
LocalWorkingDir=/Users/glanik/lfs-migration-test
LocalGitDir=/Users/glanik/lfs-migration-test/.git
LocalGitStorageDir=/Users/glanik/lfs-migration-test/.git
LocalMediaDir=/Users/glanik/lfs-migration-test/.git/lfs/objects
LocalReferenceDirs=
TempDir=/Users/glanik/lfs-migration-test/.git/lfs/tmp
ConcurrentTransfers=8
[...]

PR #5327, opened 2023-03-31
Clarify `git lfs migrate` handling of local and remote references

chrisd8088:
When using the `git lfs migrate` command with its `--everything` option, all commits reachable from any refs, local or remote, are considered for migration; however, per PR #2559, only local refs are updated after the migration to avoid desynchronizing remote refs with the origin.

We therefore clarify this distinction in the command's manual page, as well as clarifying that the `git push` command's `--all` option refers only to local branches.

Resolves the documentation issue noted in #5316.

### DIFF
--- a/docs/man/git-lfs-migrate.adoc
+++ b/docs/man/git-lfs-migrate.adoc
@@ -98,14 +98,21 @@ pointers as files by using `--pointers=no-follow`. (The latter option is
 akin to how existing Git LFS pointers were handled by the `info` mode in
 prior versions of Git LFS).
 
-When using the `--everything` option, take note that it means all refs
-(local and remote) will be considered, but not necessarily all file
-types. The `import` and `info` modes consider all file types by default,
-although the `--include` and `--exclude` options constrain this
-behavior. Also note that after importing across all branches with the
-`--everything` option (and then checking to ensure the results are
-satisfactory!) it may be convenient to update multiple branches on your
-remotes by using the `--all` option to `git push`.
+When using the `--everything` option, take note that it means all commits
+reachable from all refs (local and remote) will be considered, but not
+necessarily all file types. The `import` and `info` modes consider all file
+types by default, although the `--include` and `--exclude` options constrain
+this behavior.
+
+While the `--everything` option means all commits reachable from any
+ref will be considered for migration, after migration only local refs will
+be updated even when `--everything` is specified. This ensures remote
+refs stay synchronized with their remote. In other words, `refs/heads/foo`
+will be updated with the `--everything` option, but `refs/remotes/origin/foo`
+will not, so it stays in sync with the remote until `git push origin foo`
+is performed.  After checking that the results of a migration with
+`--everything` are satisfactory, it may be convenient to push all local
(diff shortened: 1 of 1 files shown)


bk2204 (approved):
Thanks for clarifying this.  I think this is going to be help people understand the way things work much better.

chrisd8088:
I thought it might be useful, for future reference, to detail an example which shows how a commit reachable only from a remote reference will be migrated when `--everything` is used, but is then left unreachable because it is not reachable from any local references.

Note that all three matching files matching the `--include="*.bin"` pattern are converted into blobs, and the commits where they were introduced are rewritten, but only the local refs are updated and so the one commit which is reachable only from `refs/remotes/origin/other` is left unreachable.

```
$ git init source
$ cd source
$ echo init >init.bin
$ git add init.bin
$ git commit -m init
$ git checkout -b other
$ echo foo >foo.bin
$ git add foo.bin
$ git commit -m foo
$ echo bar >bar.bin
$ git add bar.bin
$ git commit -m bar

$ cd ..
$ git clone source target
$ cd target
$ git reset --hard HEAD^
HEAD is now at f4303e8 foo
$ git checkout main
branch 'main' set up to track 'origin/main'.

$ git lfs migrate import --everything --include="*.bin"
migrate: Rewriting commits: 100% (3/3), done.                                   
  main	efddaae30c09e756686832d34a7dd69427b7fcd4 -> bd56ad43eb62db78fadbf51d0d5eb1c91b71d0c3
  other 	41d4dd8a12ddb445843adbf8f52f6aae877007e7 -> 4daa4eff655003d395e20d16960e186c4342e14c
migrate: Updating refs: ..., done.                                              
migrate: checkout: ..., done.        

$ find .git/lfs/objects -type f
.git/lfs/objects/b5/bb/b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c
.git/lfs/objects/7d/86/7d865e959b2466918c9863afca942d0fb89d7c9ac0c99bafc3749504ded97730
.git/lfs/objects/06/11/06111c6797d5e5932b792d0c3afdf13bc912ba44af806058bbca4770d8915305

$ cat .git/lfs/objects/7d/86/7d865e959b2466918c9863afca942d0fb89d7c9ac0c99bafc3749504ded97730
bar

$ git fsck --unreachable
Checking object directories: 100% (256/256), done.
unreachable blob 01613474789de934ab15077e96fb168c6a14f694
unreachable commit ccb4958973a439675849756ce055866970223612
unreachable tree 24125f03bace9df92575e9c10f981bdebb546917

$ git show ccb4958973a439675849756ce055866970223612:bar.bin
oid sha256:7d865e959b2466918c9863afca942d0fb89d7c9ac0c99bafc3749504ded97730
size 4

$ git show other:foo.bin
oid sha256:b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c
size 4
$ git show other:bar.bin
fatal: path 'bar.bin' does not exist in 'other'

$ git show refs/remotes/origin/other:foo.bin
foo
$ git show refs/remotes/origin/other:bar.bin
bar
```

PR #5355, opened 2023-05-01
Add a `--file` option to install and uninstall

### LINKED ISSUE #4947: Proposal: add a `--file` flag to `git lfs install` in order to write the filter configuration to an arbitrary file
**Describe the issue**
Currently, `git lfs install` only supports setting the smudge and clean filters globally by writing to the `~/.gitconfig` file.
Git itself also supports reading its global configuration from the `~/.config/git/config` file, if the user wishes to keep their home directory clean.
Adding a `--file` flag would allow such user to set the filters in their config file of choice, thus avoiding the presence of multiple global git configs.
**System environment**
git version 2.35.2.windows.1, Microsoft Windows [Version 10.0.22000.556]

**Output of `git lfs env`**
```
git-lfs/3.0.2 (GitHub; windows amd64; go 1.17.2)
git version 2.35.2.windows.1
```

bk2204:
There are actually two possible global configuration files, `$HOME/.gitconfig` and `$XDG_CONFIG_HOME/git/config`.  However, the `--global` file by default modifies only the former if it's present. Add a `--file` option to `git lfs install` and `git lfs uninstall` so that users can choose which of those files they'd like to use.

Fixes #4947

### DIFF
--- a/commands/command_install.go
+++ b/commands/command_install.go
@@ -10,6 +10,7 @@ import (
 )
 
 var (
+	fileInstall       = ""
 	forceInstall      = false
 	localInstall      = false
 	worktreeInstall   = false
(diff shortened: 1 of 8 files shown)


chrisd8088 (approved):
Nice, thanks!

Issue #5333, opened 2023-04-07
Smudge fails when there is not remote


b-camacho:
**Describe the bug**
Any command that triggers the smudge filter in a repo that:
- uses git-lfs
- has to remote specified
Will cause the smudge filter to fail with
```
Downloading big.chung (10 MB)
Error downloading object: big.chung (f5e02aa): Smudge error: Error downloading big.chung (f5e02aa71e67f41d79023a128ca35bad86cf7b6656967bfe0884b3a3c4325eaf): batch request: missing protocol: ""
```

**To Reproduce**
1. Install and configure `git-lfs` globally (`git-lfs` has to be on your `PATH`, and you need a `[filter "lfs"]` entry in your `.gitconfig`
2. `git init test-lfs`
3. `cd test-lfs`
4. `git fetch https://gitlab.com/b-camacho/test-lfs.git refs/heads/main:refs/heads/main # this fetches the objects from the remote, without actually storing it as a remote known to local git`
5. `git archive 76e14b4d94fed6af4cb1243adf6f3fdcf5e73077 # this command fails`

**Expected behavior**
The smudge filter should work the same way when no remote is set.

**System environment**
`Linux vrd 5.15.0-69-generic #76~20.04.1-Ubuntu SMP Mon Mar 20 15:54:19 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`
`git-lfs` build from source (rev `78ecb1fd75e84891cd23a1e266142f8ac8d3641e`) with `go version go1.19 linux/amd64`

**Output of `git lfs env`**
```
git-lfs/3.3.0 (GitHub; linux amd64; go 1.19; git 78ecb1fd)
git version 2.37.3

Endpoint=https://github.com/git-lfs/git-lfs.git/info/lfs (auth=none)
LocalWorkingDir=/home/bmc/sources/git-lfs
LocalGitDir=/home/bmc/sources/git-lfs/.git
LocalGitStorageDir=/home/bmc/sources/git-lfs/.git
LocalMediaDir=/home/bmc/sources/git-lfs/.git/lfs/objects
LocalReferenceDirs=
TempDir=/home/bmc/sources/git-lfs/.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
FetchRecentRefsIncludeRemotes=true
[...]

PR #5357, opened 2023-05-02
Add `FETCH_HEAD` as fallback remote source

b-camacho:
Addresses https://github.com/git-lfs/git-lfs/issues/5333

# What changed?
Some git commands (`git archive` is an example) read the special `FETCH_HEAD` file to retrieve information about the most recently fetched remote, even when no remote is stored in the local git repository.

With this change, if reading remotes in other ways fails, `git-lfs` will try to remote information directly from the `.git/FETCH_HEAD` file.

# Implementation
I couldn't find a `git` command for reading the `FETCH_HEAD` file so I read in and parsed it directly. The format for the remote URL in `FETCH_HEAD` seems non standard, there may be variations of it that I missed, or a better way to normalize the URL.

# Testing
I tried this on a couple repos of mine and it does work without a remote! What would be a good starting point for regression testing? Should I add unit tests for this as well?


b-camacho:
One outstanding problem:
git's internal files are not always under the `.git` path which I currently hardcoded, looking into ways to retrieve that

bk2204:
Hey, sorry for the delay.  I just wanted to say thanks for the pull request, and welcome to Git LFS.  In general, I think this feature is useful, but the core team is currently discussing when reading from `FETCH_HEAD` is the right thing to do, and we'll try to get back with you to a review as soon as possible.

b-camacho:
No worries! Happy to hear you are discussing it.

b-camacho:
Hey @bk2204, I was wondering if the maintainers had a chance to talk about this PR, or is there anything you'd like to see changed/added? I'd really want this change to land!

bk2204 (commented):
Sorry about the delay.  I wrote up a review, promptly forgot to hit the comment button, and then went on vacation for 8 days.  I think this is moving in a good direction, but there are a few changes we need to make, and we also need some tests.

b-camacho:
Sorry for taking a while to get back to you! Thanks for all the review comments, I totally missed that there's already a function for doing all the URL conversions I needed. I retested the code, and fetching still works while using the existing `NewEndpointFromCloneURL` which is great!

lfsapi/endpoint_finder.go
```diff
@@ -117,14 +119,53 @@ func (e *endpointGitFinder) RemoteEndpoint(operation, remote string) lfshttp.End
...
+	if scanner.Scan() {
+		line := scanner.Text()
+		return ExtractRemoteUrl(line)
+	}
+
+	return "", fmt.Errorf("Failed to read content from %s", filePath)
+}
+
+func ExtractRemoteUrl(line string) (string, error) {
+        re := regexp.MustCompile(`\s(https://|ssh://)?([\/\.\-\:\_a-zA-Z0-9]+)$`)
```
bk2204, line 158: How does this work for local fetches with local paths and `file:///` URLs?

bk2204 (commented):
Sorry for the delay in getting back to you.  I think this is overall moving in the right direction, but I have a question and we'd still want to see some tests for this in `t`.

lfsapi/endpoint_finder.go
b-camacho: interesting, I tested this with a local checkout:
```
# make a local repo with an lfs file
mkdir lfs-local-src && cd lfs-local-src
git init .
fallocate -l 10000000 big.chung
git lfs track --filename big.chung
git add .
git commit -m "add big file"

# make another local repo, fetch from the other repo via `file://`
mkdir ~/lfs-local && cd ~/lfs-local
git init .
git fetch file:///home/bmc/sources/lfs-local-src refs/heads/master
GIT_TRACE=1 GIT_TRANSFER_TRACE=1 GIT_CURL_VERBOSE=1 git archive 76e14b4d94fed6af4cb1243adf6f3fdcf5e73077
# git archive output:
# pax_global_header00006660000000000000000000000064144602460010014507gustar00rootroot0000000000000052 comment=7d4efb35e1db865d69edf7aa254ab8c6e93fedac
# big.chung000066400000000000000000000002051446024600100126370ustar00rootroot00000000000000version https://git-lfs.github.com/spec/v1
# oid sha256:f5e02aa71e67f41d79023a128ca35bad86cf7b6656967bfe0884b3a3c4325eaf
# size 10000000
```
and it worked fine! but I agree with your implication that `ExtractRemoteUrl` should not work here, trying to figure out what other codepath must have been triggered

b-camacho:
Got some more time to look into the `FETCH_HEAD` formats. I couldn't find the format specified anywhere, so I ran a [sourcegraph search](https://sourcegraph.com/search?q=context%3Aglobal+file%3AFETCH_HEAD%24&patternType=standard&sm=1&groupBy=repo) then scraped the `FETCH_HEAD` file contents from github  ([here](https://gist.github.com/b-camacho/8d633bc44bb96312f74177baaa7b8812) is the result if you're curious)

This gave me a pretty good idea of what does and doesn't occur in a `FETCH_HEAD` file and I came up with [this slightly monstrous regex](https://regex101.com/r/lYla7c/1)

b-camacho:
Anyway, added unit tests and retested the integration with nix, happy to report everything still works while using the big, much more specific regex.

bk2204 (commented):
I think this looks mostly fine to me, but I'd like to see some tests in the `t` directory (that is, integration tests) to be sure that we do fetch data from the proper remote when we clone or fetch repositories.  They're written in shell, so hopefully they aren't too hard to write.

bitcoinfinancier (approved):

b-camacho:
I really like the way the tests in `t/` are set up!, I added two. 
First one to check if we use FETCH_HEAD as fallback. Verified it fails on main, as expected, with
```
#     Downloading test_file.dat (14 B)
#     Error downloading object: test_file.dat (bd06b87): Smudge error: Error downloading test_file.dat (bd06b879000490b5b90ea20d30e5d1068ab3d848f36b8c8d8d9622fa45aa0c7c): batch request: missing protocol: ""
#
#     Errors logged to '/tmp/git-lfs_TEMP.I0icgr/t-no-remote.sh-418417/t-no-remote-dst/lfs/logs/20230822T005155.539528639.log'.
#     Use `git lfs logs last` to view the log.
#     error: external filter 'git-lfs filter-process' failed
#     fatal: test_file.dat: smudge filter lfs failed
#     + test_status=128
```
but passes with this change.

Also added a test to verify FETCH_HEAD is not used when a remote is set.

Lmk how this sounds

bk2204 (approved):
Thanks, this looks good.

I think we'll mark this as an experimental feature for the next release just to make sure there aren't any unexpected bugs before we mark it ready for prime time.  I do think this is a valuable feature, though, so thanks for working on it.

bk2204:
Ah, drat, I spoke too soon.  It looks like there are some broken tests here that will need fixing.

b-camacho:
Makes sense to add as an experimental feature.
The test was a genuine failure. For some commands, if you requested a specific remote but that remote was missing, we'd still use the fallback (definitely a bug). I fixed it by only using the fallback if we're using the default remote, which seems like a good resolution?

bitcoinfinancier (approved):

bk2204:
Yes, I think that seems like a good solution here.

bk2204:
It looks like we still have a few formatting errors here.

b-camacho:
fixed the formatting, ran `script/cibuild` locally and it seemed happy 🤞

PR #5358, opened 2023-05-03
Improve HTML formatting of git-lfs-push.adoc

shoelzer:
Avoid a line break between backticks because that produces a hardcoded line break in HTML.

### DIFF
--- a/docs/man/git-lfs-push.adoc
+++ b/docs/man/git-lfs-push.adoc
@@ -28,8 +28,8 @@ by the local clone of the remote.
   all local refs are pushed. Note that this behavior differs from that of
   git-lfs-fetch(1) when its `--all` option is used; in that case, all refs are
   fetched, including refs other than those under `refs/heads` and `refs/tags`.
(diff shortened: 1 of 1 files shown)


bk2204:
Hey,

Where are you seeing this?  I've run `make man/html/git-lfs-push.1.html` and this doesn't happen in the output.  If this is on GitHub's rendering, it would probably be better to get the rendering pipeline fixed rather than try to work around it here.

shoelzer:
I see it on GitHub here: https://github.com/git-lfs/git-lfs/blob/main/docs/man/git-lfs-push.adoc

bk2204:
Okay, I believe that's a bug in the GitHub rendering pipeline rather than a bug in this repository.  Could you please report that at https://github.com/github/markup/issues, since it works correctly when built from the repository using Asciidoctor?

shoelzer:
OK, moved to https://github.com/github/markup/issues/1689 and closing this PR.

PR #5364, opened 2023-05-11
Drop pre-1.13 Go compatibility code

slonopotamus:
I failed to find which Go version git-lfs declares as oldest supported. CI runs against 1.19 and 1.20. But anyway, 1.12 became EOL back in 2020.


bk2204 (approved):
This looks good.  Thanks for the patch.

PR #5376, opened 2023-05-28
Add Refspec value to lockIdFromPath request

bogomolets-owl:
This PR fixes an inconsistent behaviour in git lfs client requests behaviour.
Currently all locking requests send refspec value to server except lockIdFromPath. It actually looks like the refspec value should be passed thought it dosent at the moment.

Motivation:
This inconsistency prevents custom server side locking logic that takes current user branch into account. Specificaly the unlocking request that is called with filespec param instead of lock id.
git lfs unlock "path/to/file" 

The usecase for this custom logic is a "custom lockspace" feature for ie hotfix branches, that should not interfere with main development branches locks and wont be ever merged as is back into main branch.

I work in game development where locking is used extensively due to high amount of binary assets and requires some "in place" logic tweaks for default locking logic.
Usecases include:
- hotfix branches that are made for a current product version and wont be merged into the main development branch as is;
- feature branches for "dirty" experiments that wont be merged as is and require broad project modifications with several team members;
- release branches that wont follow git flow development process, where some in place changes are applied to the current release version and backported in modified state into final product;

I dont see any directly related tests for this specific request. If they should actually be created please advice.

### DIFF
--- a/locking/locks.go
+++ b/locking/locks.go
@@ -402,6 +402,7 @@ func (c *Client) lockIdFromPath(path string) (string, error) {
 		Filters: []lockFilter{
 			{Property: "path", Value: path},
 		},
+		Refspec: c.RemoteRef.Refspec(),
 	})
 
 	if err != nil {
--- a/t/t-unlock.sh
+++ b/t/t-unlock.sh
@@ -451,3 +451,42 @@ begin_test "unlock with git-lfs-transfer"
   refute_server_lock_ssh "$reponame" "$id" "refs/heads/main"
 )
 end_test
+
+begin_test "client is sending a remote ref to the server"
+(
+  set -e
+
+  reponame="client-is-sending-a-remote-ref-to-the-server"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" "$reponame"
+
+# setup repo gitattributes and commit a test file
+  git lfs track "*.dat"
+  echo "a" > a.dat
+  git add .gitattributes a.dat
+  git commit -m "add a.dat"
+  git push origin main
+
+# lock the file with curl verbose output and record the output to lock.log
+  GIT_CURL_VERBOSE=1 git lfs lock a.dat | tee lock.log
+
+# retrieve the lock id from the lock.log
+  id=$(assert_lock lock.log a.dat)
+
+# check file is locked on server
+  assert_server_lock "$reponame" "$id"
+
+# check that the refspec is sent via the request body in the lock request
+# not sure in this pattern cause the output is multiline
+  grep -E "POST.|\n*\"ref\":{\"name\":\"refs\/heads\/main\"}" lock.log
+
+# unlock the file with curl verbose output and record the output to unlock.log
+  GIT_CURL_VERBOSE=1 git lfs unlock a.dat | tee unlock.log
+
+# check that the refspec is sent via the locks request url
+  grep -E "GET.*&refspec=refs%2Fheads%2Fmain" unlock.log
+
+# check that the refspec is sent via the request body in the unlock request
+# not sure in this pattern cause the output is multiline
+  grep -E "POST.|\n*\"ref\":{\"name\":\"refs\/heads\/main\"}" unlock.log
+)


chrisd8088:
Thank you for the patch, and welcome to the Git LFS project!  Your explanation of the problem is very helpful and the proposed resolution seems correct.

It would be ideal to have a test included in this PR which demonstrates the fix; i.e., a test which fails without the patch but succeeds with it.  That would also give us a reproduction example we could test against major Git LFS services like GitHub to make sure this change to the client doesn't introduce any problems when working with their service.

If you have time to try to add such a test to the `t/t-unlock.sh` [file](https://github.com/git-lfs/git-lfs/blob/1efc16b424ea1f88bfbcfe81a667bcfb2cdba3d5/t/t-unlock.sh), that would be great, and if you need help with that, please just let us know.  It might be necessary to adjust the `t/cmd/lfstest-gitserver.go` file's [handler](https://github.com/git-lfs/git-lfs/blob/1efc16b424ea1f88bfbcfe81a667bcfb2cdba3d5/t/cmd/lfstest-gitserver.go#L1365-L1409) for the `/locks` endpoint, as this is the test Git LFS server used in the test suite.  We would want that test server's implementation of the Git LFS locking API to reflect what is expected when a `/locks` query is made in the specific conditions you are concerned about.

I built a Git LFS client with this PR's change and did a little testing against GitHub's locking API.  It looks as though, when working on a branch, the ref name is not passed to the `/locks` API endpoint when trying to look up the lock ID by path, which I think is what you are aiming to fix.

Without this change:

```
$ git checkout foo
$ git lfs lock bar.txt
$ GIT_CURL_VERBOSE=1 git lfs unlock bar.txt
... 
> GET /chrisd8088/....git/info/lfs/locks?path=bar.txt HTTP/1.1
...
> POST /chrisd8088/....git/info/lfs/locks/3106505/unlock HTTP/1.1
...
{"force":false,"ref":{"name":"refs/heads/foo"}}
...
```

With this change:
```
$ GIT_CURL_VERBOSE=1 git lfs unlock bar.txt
...
> GET /chrisd8088/....git/info/lfs/locks?path=bar.txt&refspec=refs%2Fheads%2Ffoo HTTP/1.1
...
```

So it seems as though GitHub's locking API implementation may ignore the ref name and only track locks by path, which would be consistent with Git LFS client behaviour pre-v2.4.0.  Other Git LFS services (GitLab, Atlassian, etc.) may be different and have implemented ref-aware behaviour.

It appears that ref support in the client's locking behaviour was added in a series of PRs tracked in #2712, such as PR #2773.

At any rate, I'd suggest adding a test (which may involve changing the test server to be ref-aware) which illustrates the problem by failing without this change and succeeding with it.

Thanks again for taking the time to investigate and develop the patch!

bogomolets-owl:
Hi chrisd8088!

Thank you!

I dont know much about go or linux development so it took me a while to understand the clues and make anything plausable. 
I am not entirely sure i did the job, but this is what i came up with: for now i added a client test, that checks that every request to server contains actual refspec value in request url or request body for GET and POST requests respectively.
It checks that the refspec is actually sent from client side and this test should fail without the change cause search request is missing the refspec value without the fix. This is a straitforward fix check.

I dont have environment to check how this test works with grep regex patterns that i used so i am not entierly sure that test does what it should. I left comments to explain it.

But as you mentioned the server modifiation - i believe this test is not actually what you wanted cause it dont interfiere with server logic in any way. Did you mean to build a client side test that demonstrates "separate locks in different branches" logic? 

I don't seem to be very helpful with testing, so help is appreciated!

chrisd8088:
Thanks for the additional work!  I'll take a look as soon as I can, and see if maybe I can help out with some changes to our test server as well.  Thanks again, and also for your patience!

phuchau1989 (approved):

chrisd8088 (commented):
Hey, thanks again for your contribution and your patience!  I finally had a few brain cells I could align in the right direction and got some time to focus on this PR again.

Thank you very much for adding a test!  With a few tweaks I think we could get that working and reported as such.  However, as it turns out, I think we can do something even more substantive and won't need to parse any debug output or anything like that!

Before I describe the details, is it OK with you if I add commits to this PR?  I would like to add the three patches from my `lock-id-path-refspec` [branch](https://github.com/chrisd8088/git-lfs/tree/lock-id-path-refspec).

---

I tried to put a fair bit of detail in the commit descriptions, but to summarize, I'll start by referring to what I wrote some time ago in https://github.com/git-lfs/git-lfs/pull/5376#issuecomment-1566245177:

> It might be necessary to adjust the `t/cmd/lfstest-gitserver.go` file's [handler](https://github.com/git-lfs/git-lfs/blob/1efc16b424ea1f88bfbcfe81a667bcfb2cdba3d5/t/cmd/lfstest-gitserver.go#L1365-L1409) for the `/locks` endpoint, as this is the test Git LFS server used in the test suite.  We would want that test server's implementation of the Git LFS locking API to reflect what is expected when a `/locks` query is made in the specific conditions you are concerned about.

In other words, I wanted to make sure that the test server we use in our test suite, which emulates a full production Git LFS API service, would look for and use the `refspec` query parameter.  Without that, our tests would be somewhat incomplete -- we could check that the client was sending the parameter, but we wouldn't know if it was effective or not.

Also, I wanted our test server to be able to emulate a Git LFS API service which did _not_ check for the `refspec` query parameter or filter search results using it, the way GitHub's API appears to still behave.  Since our Git LFS client has to work with their service and others like it, we would like to make sure that adding the `refspec` query parameter won't have an effect if the server just ignores it.

Fortunately, it turns out that these needs were all anticipated back when the support for the `refspec` parameter was added in PR #2773!  The `lfstest-gitserver.go` utility, which our tests use to simulate a production API service, already has a [mechanism](https://github.com/git-lfs/git-lfs/blob/ab3f2288bdb3cface5a4a0af2daa9c302214ab6f/t/cmd/lfstest-gitserver.go#L1231-L1241) by which it can be switched between "old" and "new" modes when handling requests to the `/locks` endpoint.  If the test repository name has a special suffix, the server expects the `refspec` query parameter and ensures it matches what the test wants; if the suffix is not part of the test repository name, though, the server acts like an older service which just ignores the `refspec` parameter.

The problem, it seems, is that while there are [tests](https://github.com/git-lfs/git-lfs/blob/ab3f2288bdb3cface5a4a0af2daa9c302214ab6f/t/t-unlock.sh#L13-L120) which use the special suffix to enable the "new" behaviour in the test API server, they were inadvertently changed to all use the `git lfs unlock --id` option, which means they don't ever make lock search requests, and so the `refspec` behaviour is not tested.  (In fact, two of the tests ended up being identical except for their names.)  This gap in the tests is why the behaviour you've fixed in this PR wasn't detected by the test suite.

Arguably, the most important thing we want to test is that when `git lfs unlock` is used with a file path and not the `--id` argument, and the `refspec` is sends does _not_ match the ref used when creating the only lock for that file path, the request fails to remove the lock if the test server is acting in the "new" mode and filtering search results using the `refspec`, but succeeds and removes the lock if the test server is acting in the "old" mode and ignoring the `refspec` parameter.

This latter version of the test in particular would confirm that things work as they do today with a service like GitHub, which doesn't pay attention to the `refspec` argument even if one is sent in a request to the `.../locks` API endpoint.

So, I've attempted to fix up the problems in the test suite and make better use of the facilities already available in the test server, and add tests which confirm that the `refspec` parameter now sent by the `git lfs unlock` command works (and does not work) as expected under multiple conditions and with both modes of the test server.

I hope that is all acceptable to you!  If so, I'll push these commits into your branch, if I can, or feel free to pull them into this PR, and then we can look for approval from another member of the Git LFS core team.

Thanks again for raising this issue and contributing a fix, and for your patience!

chrisd8088:
I included the original commit from this PR in PR #5538, which is now merged, so I think we can close this.  Thanks again for the bug report and the patch!

Issue #5222, opened 2022-12-14
Improve credential manager handling


bk2204:
There's [currently a series on the Git list](https://lore.kernel.org/git/pull.1352.v4.git.1670880984.gitgitgadget@gmail.com/) to expand the capabilities of credential helpers somewhat.  We should extend Git LFS to support it, which should be backwards compatible, and verify that our credential handling is correct, including that we handle an empty string correctly to reset helpers.

PR #5381, opened 2023-05-31
Add support for `wwwauth[]` to credential helpers

bk2204:
Git recently added a new field to the credential helper, `wwwauth[]`, which may be repeated and includes all of the `WWW-Authenticate` headers so that the credential helper may choose an appropriate set of credentials and extract any sort of necessary data from the field (such as challenge).

In Git LFS, we also want to do this with the `LFS-Authenticate` headers as well, since those are often used for the same purpose, so include both these headers in that field when passing them to `git credential fill`.

Note that `git credential fill` only honours this value and passes it to the credential helper in Git 2.41 and newer (including the latest `HEAD`).  However, just to be safe, let's add an undocumented and experimental option (`credential.*.skipwwwauth`) that users can use to control this, which we can remove in a few releases if it turns out it's not needed.  Similarly, skip our new tests if we have an older version of Git where this doesn't work, since they'll otherwise fail.

In addition, perform some preparatory work to allow multiple values for a key, which we'll use in the second commit.

Fixes: #5222

### DIFF
--- a/creds/netrc.go
+++ b/creds/netrc.go
@@ -63,7 +63,7 @@ func newNetrcCredentialHelper(osEnv config.Environment) *netrcCredentialHelper {
 }
 
 func (c *netrcCredentialHelper) Fill(what Creds) (Creds, error) {
-	host, err := getNetrcHostname(what["host"])
+	host, err := getNetrcHostname(firstEntryForKey(what, "host"))
 	if err != nil {
 		return nil, credHelperNoOp
 	}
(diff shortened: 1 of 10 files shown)


creds/netrc.go
```diff
@@ -63,7 +63,7 @@ func newNetrcCredentialHelper(osEnv config.Environment) *netrcCredentialHelper {
 }
 
 func (c *netrcCredentialHelper) Fill(what Creds) (Creds, error) {
-	host, err := getNetrcHostname(what["host"])
+	host, err := getNetrcHostname(what["host"][0])
```
chrisd8088, line 66: I noticed that in `creds/creds.go` the changes in this commit use `firstEntryForKey(what, "host")`, e.g., in `(*AskPassCredentialHelper).Fill()`.  Should the methods for `*netrcCredentialHelper` do the same (here, and in `Approve()` and `Reject()` below)?

chrisd8088 (approved):

creds/netrc.go
bk2204: Ah, yes, you're right.  I'll make that change.

PR #5385, opened 2023-06-06
Unlock by ID with JSON Flag returns empty array

salvorizza:
When unlock is called with the JSON flag and ID, the response is empty.

The current behavior:
git lfs lock --json <SOME_PATH>
[{"id":"103","path":"<SOME_PATH>","owner":{"name":"<Author>"},"locked_at":"2023-06-06T10:44:37Z"}]

git lfs unlock --json --id=103 
[]

The expected behavior:
git lfs lock --json <SOME_PATH>
[{"id":"103","path":"<SOME_PATH>","owner":{"name":"<Author>"},"locked_at":"2023-06-06T10:44:37Z"}]

git lfs unlock --json --id=103 
[{"id":"103","unlocked":true}]

### DIFF
--- a/commands/command_unlock.go
+++ b/commands/command_unlock.go
@@ -109,6 +109,11 @@ func unlockCommand(cmd *cobra.Command, args []string) {
 			success = false
 		} else if !locksCmdFlags.JSON {
 			Print(tr.Tr.Get("Unlocked Lock %s", unlockCmdFlags.Id))
+		} else {
+			locks = append(locks, unlockResponse{
+				Id:     unlockCmdFlags.Id,
+				Unlocked: true,
+			})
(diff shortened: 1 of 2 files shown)


bk2204 (commented):
Hey, thanks for the patch, and welcome to Git LFS!  This looks fine, but I noticed that the unlock code has the same problem.  Would you be willing to fix that as well and add some tests for this case in the `t` directory?

salvorizza:
@bk2204 I've created a test for this scenario, currently the only thing that is failing is a test on askpass on Mac, i think that's not related to this task.

bk2204 (approved):
Looks good, thanks.

PR #5392, opened 2023-06-13
Update distributions

### LINKED ISSUE #5227: Update distros for Linux packages
Dupe of #4988, but for v3.4.0 I guess?

bk2204:
This PR contains four commits.  The first does some tidying up of the various locations we have for base distros (that is, those for which we build packages).  The next two update the code for Debian 12.  Finally, we update the remaining distributions for v3.4.0.

Fixes #5227


chrisd8088 (approved):
Lovely, thanks!

PR #5402, opened 2023-06-21
Handle local paths with trailing slashes

### LINKED ISSUE #5400: cloning local repo with trailing slash fails
**Describe the bug**

For a given local repo with LFS, `git clone /path/to/repo` succeeds, but `git clone /path/to/repo/` fails.

**To Reproduce**

```
~/git$ git init --bare repo.git
Initialized empty Git repository in /home/ed/git/repo.git/
~/git$ git init test
Initialized empty Git repository in /home/ed/git/test/.git/
~/git$ cd test
~/git/test$ cp /usr/bin/bash .
~/git/test$ git lfs track bash
Tracking "bash"
~/git/test$ git commit -a -m "initial commit"
~/git/test$ git remote add origin $PWD/../repo.git
~/git/test$ git push origin main
Uploading LFS objects: 100% (1/1), 0 B | 0 B/s, done.
Enumerating objects: 4, done.
Counting objects: 100% (4/4), done.
Delta compression using up to 4 threads
Compressing objects: 100% (3/3), done.
Writing objects: 100% (4/4), 407 bytes | 407.00 KiB/s, done.
Total 4 (delta 0), reused 0 (delta 0), pack-reused 0
To /home/ed/git/test/../repo.git
 * [new branch]      main -> main

# This works.
~/git$ git clone $PWD/repo.git new2
Cloning into 'new2'...
done.

# This should work, but fails to fetch LFS objects.
~/git$ git clone $PWD/repo.git/ new3
Cloning into 'new3'...
done.
Downloading bash (1.4 MB)
Error downloading object: bash (2c336c6): Smudge error: Error downloading bash (2c336c63e26881d2f02f34379024e7c314bce572c08cbaa319bacbbec29f93ed): batch request: missing protocol: "/home/ed/git/repo.git/info/lfs"

Errors logged to '/home/ed/git/new3/.git/lfs/logs/20230621T004858.750834098.log'.
fatal: bash: smudge filter lfs failed
and retry with 'git restore --source=HEAD :/'

~/git$
```

**Expected behavior**

This should mimic git's behavior for local paths and treat them both identically. 

**System environment**

```
Description:	Ubuntu 22.04.2 LTS
Release:	22.04
Codename:	jammy
```

Output of git lfs env

```
git-lfs/3.3.0 (GitHub; linux amd64; go 1.19.3)
git version 2.41.0

Endpoint=file:///home/ed/git/repo.git (auth=none)
LocalWorkingDir=/home/ed/git/test3
LocalGitDir=/home/ed/git/test3/.git
LocalGitStorageDir=/home/ed/git/test3/.git
LocalMediaDir=/home/ed/git/test3/.git/lfs/objects
LocalReferenceDirs=
TempDir=/home/ed/git/test3/.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/home/ed/git/test3/.git/lfs
AccessDownload=none
AccessUpload=none
GIT_TEMPLATE_DIR=/home/ed/.git_template
```

bk2204:
When we want to handle a local path, we must rewrite it internally into a `file:///` URL, because we dispatch our standalone transfer agent based on the URL scheme.  However, once we've looked up an appropriate URL and formatted it as the scheme prefers, if it contains a trailing slash, we ignore the entire lookup and replace the URL with the raw one given, but without the trailing slash.

This behaviour seems a bit bizarre, and it causes us to take a local path with a trailing slash and treat it not as the `file:///` URL it needs to be, but as a raw local path, which gets refused.  Let's avoid this problem by looking at the rewritten URL, and modifying that instead if we find its trailing slash to be offensive.

Fixes #5400
/cc @ebardsley as reporter

### DIFF
--- a/lfsapi/endpoint_finder.go
+++ b/lfsapi/endpoint_finder.go
@@ -151,8 +151,8 @@ func (e *endpointGitFinder) NewEndpointFromCloneURL(operation, rawurl string) lf
 		return ep
 	}
 
-	if strings.HasSuffix(rawurl, "/") {
-		ep.Url = rawurl[0 : len(rawurl)-1]
+	if strings.HasSuffix(ep.Url, "/") {
+		ep.Url = ep.Url[0 : len(ep.Url)-1]
 	}
 
 	if strings.HasPrefix(ep.Url, "file://") {
--- a/t/t-standalone-file.sh
+++ b/t/t-standalone-file.sh
@@ -66,6 +66,7 @@ do_upload_download_test () {
 
 do_local_path_test () {
   local reponame="$1"
+  local suffix="$2"
 
   git lfs track "*.dat" 2>&1 | tee track.log
   grep "Tracking \"\*.dat\"" track.log
@@ -111,17 +112,17 @@ do_local_path_test () {
   cd "$TRASHDIR"
 
   # Check a clone using an absolute Unix-style path.
-  GIT_TRACE=1 GIT_TRANSFER_TRACE=1 git clone "$testdir" "$reponame-repo1"  2>&1 | tee clonecustom.log
+  GIT_TRACE=1 GIT_TRANSFER_TRACE=1 git clone "$testdir$suffix" "$reponame-repo1"  2>&1 | tee clonecustom.log
   (cd "$reponame-repo1" && git lfs fsck)
   grep "xfer: started custom adapter process" clonecustom.log
 
   # Check a clone using a relative path.
-  GIT_TRACE=1 GIT_TRANSFER_TRACE=1 git clone "$reponame-repo1" "$reponame-repo2"  2>&1 | tee clonecustom.log
+  GIT_TRACE=1 GIT_TRANSFER_TRACE=1 git clone "$reponame-repo1$suffix" "$reponame-repo2"  2>&1 | tee clonecustom.log
   (cd "$reponame-repo2" && git lfs fsck)
   grep "xfer: started custom adapter process" clonecustom.log
 
   # Check a clone using an absolute native-style path.
-  GIT_TRACE=1 GIT_TRANSFER_TRACE=1 git clone "$(native_path "$testdir")" "$reponame-repo3"  2>&1 | tee clonecustom.log
+  GIT_TRACE=1 GIT_TRANSFER_TRACE=1 git clone "$(native_path "$testdir")$suffix" "$reponame-repo3"  2>&1 | tee clonecustom.log
   (cd "$reponame-repo3" && git lfs fsck)
   grep "xfer: started custom adapter process" clonecustom.log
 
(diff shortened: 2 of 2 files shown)


t/t-standalone-file.sh
```diff
@@ -64,6 +64,67 @@ do_upload_download_test () {
   [ "$(echo "$objectlist" | wc -l)" -eq 12 ]
 }
 
+do_local_path_test () {
```
chrisd8088, line 67: I might suggest having this function take `$reponame` as an argument.  I suspect this still works as-is because variables are global in the shell by default, so the caller can set `$reponame` and then this function can just read it below.  But it would be a bit more future-proof to maybe have it take the variable as an argument and set a local variable from it.

t/t-standalone-file.sh
```diff
@@ -109,17 +109,17 @@ do_local_path_test () {
   cd "$TRASHDIR"
 
   # Check a clone using an absolute Unix-style path.
-  GIT_TRACE=1 GIT_TRANSFER_TRACE=1 git clone "$testdir" "$reponame-repo1"  2>&1 | tee clonecustom.log
+  GIT_TRACE=1 GIT_TRANSFER_TRACE=1 git clone "$testdir$suffix" "$reponame-repo1"  2>&1 | tee clonecustom.log
```
chrisd8088, line 112: Again, my suggestion here would be to pass `$suffix` as a second argument to the function, rather than rely on a global (although of course that does work).

chrisd8088 (approved):
Looks good!  I had a suggestion about the tests only.  Thanks!

t/t-standalone-file.sh
bk2204: That sounds like a good suggestion.  I've squashed in changes to that effect here and below.

lfsapi/endpoint_finder.go
```diff
@@ -151,8 +151,8 @@ func (e *endpointGitFinder) NewEndpointFromCloneURL(operation, rawurl string) lf
 		return ep
 	}
 
-	if strings.HasSuffix(rawurl, "/") {
-		ep.Url = rawurl[0 : len(rawurl)-1]
+	if strings.HasSuffix(ep.Url, "/") {
```
ebardsley, line 154: Thanks for the quick fix! `strings.TrimSuffix` would be appropriate here too.

bk2204: That is a good idea, but I'm going to leave this as it stands for now in the interests of making the minimum possible fix.

PR #5412, opened 2023-06-29
git: use `git var` to find system gitattributes file

bk2204:
At the moment, Git doesn't provide us a way to know where the system gitattributes file is.  To help find it, we use the `PREFIX` environment variable and tack on `etc/gitattributes`.  However, as a practical matter, nobody sets `PREFIX`, so typically we always use `/etc/gitattributes`.

That being said, Git has just learned a way to provide this information using `git var GIT_ATTR_SYSTEM` to print this information, and this will be included in Git 2.42.  If we have such a version of Git, use it to find this information.  If it fails, return an error, but if it simply prints no values (such as with `GIT_ATTR_NOSYSTEM`), then that means Git does not consider us to have a system gitattributes file, and in that case, ignore trying to load that data.


git/attribs.go
```diff
@@ -64,19 +67,36 @@ func GetRootAttributePaths(mp *gitattr.MacroProcessor, cfg Env) []AttributePath
...
-	prefix, _ := env.Get("PREFIX")
-	if len(prefix) == 0 {
-		prefix = string(filepath.Separator)
-	}
+func GetSystemAttributePaths(mp *gitattr.MacroProcessor, env Env) ([]AttributePath, error) {
+	var path string
+	if IsGitVersionAtLeast("2.42.0") {
+		cmd, err := gitNoLFS("var", "GIT_ATTR_SYSTEM")
+		if err != nil {
+			return nil, errors.New(tr.Tr.Get("failed to set up `git var GIT_ATTR_SYSTEM`: %v", err))
```
chrisd8088, line 75: As a minor note, it looks like [many](https://github.com/git-lfs/git-lfs/blob/1820b646e05c24497652b3f9eaf91500dc1ee988/git/git.go#L392) of our [other](https://github.com/git-lfs/git-lfs/blob/1820b646e05c24497652b3f9eaf91500dc1ee988/git/git.go#L261) [invocations](https://github.com/git-lfs/git-lfs/blob/1820b646e05c24497652b3f9eaf91500dc1ee988/git/git.go#L614) of `gitNoLFS()` return any error with the message `failed to find` rather than `failed to set up`, so I might suggest doing the same here, just for consistency.

git/attribs.go
```diff
@@ -64,19 +67,36 @@ func GetRootAttributePaths(mp *gitattr.MacroProcessor, cfg Env) []AttributePath
...
+		}
+		out, err := cmd.Output()
+		if err != nil {
+			return nil, errors.New(tr.Tr.Get("failed to fetch output from `git var GIT_ATTR_SYSTEM`: %v", err))
```
chrisd8088, line 79: In a similar vein to the comment above, at least some of our [similar](https://github.com/git-lfs/git-lfs/blob/1820b646e05c24497652b3f9eaf91500dc1ee988/git/git.go#L766) [calls](https://github.com/git-lfs/git-lfs/blob/1820b646e05c24497652b3f9eaf91500dc1ee988/git/git.go#L843) report this error as `failed to call`.

chrisd8088 (approved):

git/attribs.go
bk2204: Good point.  I've updated both.

PR #5423, opened 2023-07-13
git: match patterns, not file names, for tracked files

### LINKED ISSUE #5409: git lfs track --filename "/[2023-06-23]/1.tset" "/[2023-06-23]/2.aaaa" fail to execute in win 10
**Describe the bug**

As git lfs track doc said，we can use `--filename` to set path with special character such as `[ ] * ?`. But when I do this in my win 10 computer, it went wrong.

**To Reproduce**
1. In win 10, New two files named: "/[2023-06-23]/1.tset" "/[2023-06-23]/2.aaaa" in dir `[2023-06-23]` in any installed git lfs repo's root folder.
2. Open git bash command execute `git lfs track --filename "/[2023-06-23]/1.tset" "/[2023-06-23]/2.aaaa"`
3. See error
image

And the `.gitattributes` file will add one line config which is wired, because it has already ended with error. Why should have some effect on really disk file?

**Expected behavior**

using `--filename` for paths with special character  and worked as [doc](https://github.com/git-lfs/git-lfs/blob/300f4ed3e5f9f4f0d42d326d95e7ca418ce61891/docs/man/git-lfs-track.adoc#L4) said 

**System environment**

git-lfs/3.3.0 (GitHub; windows amd64; go 1.19; git e815aff5)
git version 2.36.1.windows.1

image

**Additional context**

Alse, I tried execute `GIT_TRACE=1 git lfs track --filename "/[2023-06-23]/1.tset" "/[2023-06-23]/2.aaaa"`, the execution details follwoing:

image

From the details, I guess it may related to the `git 'ls-files' '--cached' '--' '\[2023-06-23\]/1.tset'`，because when i execute the comand longly in my git bash it show that:
image

bk2204:
Right now, when we run `git lfs track`, we warn the user if the pattern they've provided matches an existing file name.  However, when we invoke that command with `--filename` and a pattern with brackets on Windows, we get an error because the escaped pattern contains backslashes, which results in broken behaviour on Windows.

The real cause of this problem, however, is that we're ultimately asking for a file name and not a pattern.  Our invocation happens to work on Unix, but we need it to also work on Windows.  To solve this, let's instead ask for ignored cached files and specify a single pattern. Since we've asked for a pattern, this results in a different behaviour from before and avoids the error.

We would have caught this on Windows if our weird file name code had worked there, but since we included asterisks and question marks, which are not valid on Windows, the test couldn't run there.  Instead, let's leave the brackets and run the test on all platforms.  We still have the test below which does the same thing with spaces on Unix, so we're not really regressing our tests there.

In addition, since we don't have any tests for this verbose logging that prints the messages that trigger the problem, let's add some, both with more common patterns and some unusual ones as well.

Fixes #5409
/cc @zhoushaokun as reporter

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -1315,8 +1315,9 @@ func GetTrackedFiles(pattern string) ([]string, error) {
 	cmd, err := gitNoLFS(
 		"-c", "core.quotepath=false", // handle special chars in filenames
 		"ls-files",
+		"--ignored",
 		"--cached", // include things which are staged but not committed right now
-		"--",       // no ambiguous patterns
+		"-x",
 		safePattern)
 	if err != nil {
 		return nil, errors.New(tr.Tr.Get("failed to find `git ls-files`: %v", err))
--- a/t/t-track.sh
+++ b/t/t-track.sh
@@ -663,14 +663,11 @@ begin_test "track: escaped glob pattern in .gitattributes"
 (
   set -e
 
-  # None of these characters are valid in the Win32 subsystem.
-  [ "$IS_WINDOWS" -eq 1 ] && exit 0
-
   reponame="track-escaped-glob"
   git init "$reponame"
   cd "$reponame"
 
-  filename='*[foo]bar?.txt'
+  filename='[foo]bar.txt'
   contents='I need escaping'
   contents_oid=$(calc_oid "$contents")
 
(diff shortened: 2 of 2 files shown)


chrisd8088 (approved):
Thanks for the diagnosis and the fix!  I think this is much pretty much a strict improvement over the current state.

I did notice that the Git [documentation](https://git-scm.com/docs/git-ls-files#_exclude_patterns) on `git-ls-files(1)` notes:
> `gitignore(5)` specifies the format of exclude patterns

But, of course, the rules created with `git lfs track` go into `.gitattributes`, where certain `gitignore(5)` patterns don't apply; in particular, `foo/` doesn't match anything in the directory the way it does for a `.gitignore` rule.

This means we get a technically incorrect match and consequent output from `git lfs track`, and it will unnecessarily touch files when given a `foo/` pattern.
```
$ git init test
$ cd test
$ mkdir foo
$ echo foo >foo/foo.bin
$ git lfs track '*.bin'
$ git add .gitattributes foo
$ git commit -m foo

$ git-lfs track -v foo/
Tracking "foo/"
Searching for files matching pattern: foo/
Found 1 files previously added to Git matching pattern: foo/
Touching "foo/foo.bin"
```

This behaviour exists already, though, because passing `foo/` to `git ls-files` as a filename rather than an exclude pattern also matches the existing file in the directory.  Since that's the case, this PR doesn't introduce any regression in this regard, as far as I can tell.

bk2204:
Yeah, I agree we could do better here, but I think, as you said, that this is a strict improvement, so I'm going to merge it.

PR #5425, opened 2023-07-14
proposals: fix filename typo

dyrone:


chrisd8088 (approved):
Thanks for the correction!

PR #5438, opened 2023-07-24
Add RISC-V support

AaronDewes:
This adds support for the RISC-V architecture.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,6 +215,7 @@ BUILD_TARGETS = \
 	bin/git-lfs-linux-arm64 \
 	bin/git-lfs-linux-amd64 \
 	bin/git-lfs-linux-ppc64le \
+	bin/git-lfs-linux-riscv64 \
 	bin/git-lfs-linux-s390x \
 	bin/git-lfs-linux-386 \
 	bin/git-lfs-freebsd-amd64 \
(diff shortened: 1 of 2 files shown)


bk2204:
Thanks for the patch, and welcome to Git LFS!  I think we'll also need a patch to `script/upload` in `categorize_arch`, which should call it `64-bit RISC-V`.

If you want to test that script, you can do so against your remote by adding a personal access token into your `~/.netrc`, updating the `REPO` field in the top of the script to your personal fork, creating an annotated tag of your local version (call it _TAG_), and running `script/upload --skip-verify TAG`.  That should create a draft release in your version for the tag that you can use to verify that the name appears as it should.

AaronDewes:
Thanks! I tested this and it works.

bk2204 (approved):
Looks good, thanks.

Just to set expectations, this won't be included in v3.4.0, since we've already cut the release, but this should be included in v3.5.0.  We'll merge it once v3.4.0 is out.

AaronDewes:
Thanks for letting me know! I saw the PR and noticed the release was scheduled for tomorrow, so I didn't expect this to make it into the release, which is unfortunate for me, but I guess I'll have to be patient and I can already build from source for now.

PR #5439, opened 2023-07-25
Add a single source of truth for distro info

bk2204:
Right now, we have multiple different locations for our distro information.  This is suboptimal in many ways because it requires developers to keep multiple sources of information in sync.

To solve this, let's write a Ruby script that contains our single source of truth.  It contains all of the existing data we have now in an easily processable format.  It uses a style which is called a _modulino_ in the Perl community, which is essentially a script which can also be required as a library for testing purposes.

Since we can require any file which ends in `.rb`, create the script as a library with this trailing suffix and require it in the RSpec tests we're adding so that we can test it produces the expected results. These tests can be run with a simple `rspec script/spec` using a standard distro Ruby and RSpec.

For our script itself, symlink it into the `script` directory without a trailing suffix.  Various authorities, including Debian, recommend not using a trailing suffix on programs so that the implementation language can change easily if need be.  We follow this advice here.

The symlink itself should not pose a problem because this is only run on Unix by CI or core team members, and our CI system should have developer mode enabled on Windows, allowing symlinks without elevated permissions. Hapless end users cloning this repository on Windows without developer mode enabled will receive a plain text file in place of the symlink, which should not pose a practical problem since they won't be using this script.

### DIFF
--- a/docker/run_dockers.bsh
+++ b/docker/run_dockers.bsh
@@ -39,7 +39,7 @@ function split_image_name()
 }
 
 # Parse Arguments
-IMAGES=()
+IMAGES=
 PRUNE=
 ARCH=amd64
 while [[ $# -gt 0 ]]; do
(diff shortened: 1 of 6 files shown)


chrisd8088 (approved):
This looks really nice, thank you for the cleanup and centralization!

PR #5443, opened 2023-07-31
Remove vendoring instructions from contributing docs

QuLogic:
The `vendor` directory is `.gitignore`d, and vendoring as a process was fully removed in #4903.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -152,9 +152,10 @@ $ script/cibuild     # runs everything, with verbose debug output
 
 ## Updating 3rd party packages
 
-1. Update `go.mod`.
-1. Run `make vendor` to update the code in the `vendor` directory.
-1. Commit the change.  Git LFS vendors the full source code in the repository.
+1. Update `go.mod` and `go.sum`. You can ensure the latter is up-to-date by
+   using Go tools instead of manually editing `go.mod` (e.g., `go get
+   gopath@version`) or by running `go mod tidy` or `make go.sum`.
+1. Commit the change.
 1. Submit a pull request.
 
 ## Releasing


chrisd8088 (approved):
Thanks for the correction, much appreciated!

CONTRIBUTING.md
```diff
@@ -153,8 +153,7 @@ $ script/cibuild     # runs everything, with verbose debug output
 ## Updating 3rd party packages
 
 1. Update `go.mod`.
```
chrisd8088, line 155: It occurs to me on a second look that we actually should mention something here about updating `go.sum` as well, since it is tracked in Git.  Perhaps something about running `go mod tidy` to update `go.sum` would suffice, or possibly `make go.sum` (which [relies](https://github.com/git-lfs/git-lfs/blob/2137b4976f4c6e4af3c17bcedb87a1f23270caf2/Makefile#L585-L587) on `go mod verify` at the moment)?

(I also notice we have [some](https://github.com/git-lfs/git-lfs/blob/2137b4976f4c6e4af3c17bcedb87a1f23270caf2/Makefile#L105-L107) [stale](https://github.com/git-lfs/git-lfs/blob/2137b4976f4c6e4af3c17bcedb87a1f23270caf2/Makefile#L268) [references](https://github.com/git-lfs/git-lfs/blob/2137b4976f4c6e4af3c17bcedb87a1f23270caf2/Makefile#L589-L621) to the `vendor` directory in the `Makefile`; those are likely best dealt with in another PR, though.)

Would you be willing to add some text here regarding `go.sum`, before we merge this PR?

Thanks again for the help!

QuLogic: Done.

chrisd8088 (approved):
Lovely, thank you!

CONTRIBUTING.md
chrisd8088: Fantastic, thanks very much!

PR #5444, opened 2023-07-31
Update cobra to 1.7.0

QuLogic:
This release has several updates to generated completions, requiring the fixture to be updated. AFAICT, the patching that is intended to be checked by this test is still working, so I just updated the files from the just-built output.


chrisd8088 (commented):
Thanks for the patch and the update!

If I read the changes correctly, it looks like, for Zsh, the `compdef` command is now included in the script, and if that's true, we should amend the manual page's [example](https://github.com/git-lfs/git-lfs/blob/2137b4976f4c6e4af3c17bcedb87a1f23270caf2/docs/man/git-lfs-completion.adoc#L214) as part of this PR to take that example step out if it's no longer necessary.

It's probably also worth double-checking in all the shells that the updated scripts perform as expected.  This is a new feature and so we don't have an automated way of doing that.  I'll try to find some time to check over the next while, and by all means report any test results you have.

Thanks again very much!

QuLogic:
I did test with bash, but not the others. It was not extensive of course, but I did confirm that at least _something reasonable_ appeared when tab-completing `git lfs ...`.

And yes, the `compdef` was explicitly added to allow `source` of zsh completions: https://github.com/spf13/cobra/pull/1917

chrisd8088 (approved):
This looks great, thank you!  I tested all three shells and things seem to be working nicely, as anticipated.

PR #5487, opened 2023-08-26
Update to Go 1.21

chrisd8088:
As our [official policy](https://github.com/git-lfs/git-lfs/blob/a3829f486597702297737bb6677534cb5cd0bbc6/CONTRIBUTING.md#prerequisites) is to support the latest version of Go, we upgrade our CI and release workflows to version 1.21, which was recently released.

We also drop CI job runs (and therefore official support) for versions older than 1.20, as they are not supported by upstream Go.

Note that several older required CI jobs will not run for this PR, and the newer jobs should be marked as required after this PR is merged.


chrisd8088:
OK, that makes sense.  I'll look at adding the necessary env vars in some additional commits in this PR.

chrisd8088:
This PR now also sets the `GOTOOLCHAIN=local` environment variable in our `Makefile`s and Actions workflows, and when combined with PR git-lfs/build-dockers#58, I think that covers all the build contexts where we want to ensure the Go toolchain is fixed to the expected version.

For example, if we run our Actions CI job using Go 1.21 using this PR's changes to our workflows (but without any changes to our `Makefile`s), and after committing a change to the `go.mod` file to specify `go 1.22`, the result is the failure we expect the first time Go is invoked:
```
GOOS= GOARCH= go generate github.com/git-lfs/git-lfs/v3/commands
go: go.mod requires go >= 1.22 (running go 1.21.1; GOTOOLCHAIN=local)
make: *** [Makefile:236: commands/mancontent_gen.go] Error 1
```

And if we use the same modified `go.mod` but with this PR's changes to our primary `Makefile`, and run `make` locally with a Go 1.21 installation, we also get a failure:
```
GOOS= GOARCH= go build -ldflags=" -X github.com/git-lfs/git-lfs/v3/config.GitCommit=0865224b -s -w " -gcflags=" " -trimpath -o ./bin/git-lfs ./git-lfs.go
go: go.mod requires go >= 1.22 (running go 1.21.0; GOTOOLCHAIN=local)
```

So both sets of changes, to the workflows and to the `Makefile`s, appear to have the desired effect.

bk2204 (approved):
This looks great.  Thanks so much for getting this done.

Issue #5464, opened 2023-08-11
panic: runtime error: index out of range [0] with length 0  goroutine 27


wa1one:
**Describe the bug**
`git lfs fetch` shows panic runtime error (goroutine 27).

**To Reproduce**
1. Go to 'some cloned repo folder'
2. Run `git lfs fetch`
3. Enter HTTP auth/passwd (using security tokens)

**Expected behavior**
`git lfs fetch` runs successfully

**System environment**
Ubuntu 22.04

**Output of `git lfs env`**
```
panic: runtime error: index out of range [0] with length 0

goroutine 27 [running]:
github.com/git-lfs/git-lfs/v3/lfsapi.(*Client).getCreds(0xc00045bb60, {0xc000380200, 0x6}, {{0xc000380285, 0x5}, {0xc00039fe30, 0x2e}}, 0xc0000f6100)
	github.com/git-lfs/git-lfs/v3/lfsapi/auth.go:165 +0x798
github.com/git-lfs/git-lfs/v3/lfsapi.(*Client).doWithAuth(0xc00045bb60, {0xc000380200, 0x6}, {{0xc000380285, 0x5}, {0xc00039fe30, 0x2e}}, 0xc0000f6100, {0x0, 0x0, ...})
	github.com/git-lfs/git-lfs/v3/lfsapi/auth.go:63 +0xd8
github.com/git-lfs/git-lfs/v3/lfsapi.(*Client).DoWithAuth(0xc00045bb60, {0xc000380200, 0x6}, {{0xc000380285, 0x5}, {0xc00039fe30, 0x2e}}, 0xc0000f6100)
	github.com/git-lfs/git-lfs/v3/lfsapi/auth.go:26 +0x79
github.com/git-lfs/git-lfs/v3/lfsapi.(*Client).DoAPIRequestWithAuth(0xc00045bb60, {0xc000380200, 0x6}, 0xc0000f6100)
	github.com/git-lfs/git-lfs/v3/lfsapi/auth.go:57 +0x147
github.com/git-lfs/git-lfs/v3/tq.(*tqClient).Batch(0xc000384b60, {0xc000380200, 0x6}, 0xc0004b85a0)
	github.com/git-lfs/git-lfs/v3/tq/api.go:94 +0x3fe
[...]

PR #5490, opened 2023-08-31
Fix a panic in the credential code

bk2204:
In some cases, we might end up with an entry that's missing either a username or password, and in such a case, we wouldn't want to panic, like we do now.  Solve this by using the function `FirstEntryForKey` to look up the value and return an empty string if none is present.

Do some refactoring in preparatory commits to make this possible.

### DIFF
--- a/lfsapi/auth.go
+++ b/lfsapi/auth.go
@@ -162,7 +162,7 @@ func (c *Client) getCreds(remote string, access creds.Access, req *http.Request)
 		err = credWrapper.FillCreds()
 		if err == nil {
 			tracerx.Printf("Filled credentials for %s", credsURL)
-			setRequestAuth(req, credWrapper.Creds["username"][0], credWrapper.Creds["password"][0])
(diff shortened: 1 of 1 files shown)


yosi-codefresh:
@bk2204 - what is the plan about releasing it with a formal release?

bk2204:
This will be in v3.5.0.  If people would like us to do a v3.4.1 for this, we can, but we need someone who's affected to report back to us that it's fixed, either here or in #5464.

Users can build from the main branch or use the CI artifacts (which are unsigned) in the mean time or to test.

yosi-codefresh:
Is there any workaround? or either what are the cases that it happens? because i see that sometimes it haapens and sometime no. i will test it soon and will udpate

chrisd8088:
I was able to reproduce the crash, as described in https://github.com/git-lfs/git-lfs/issues/5464#issuecomment-1836787828, and demonstrate that this PR does resolve the problem, as noted in https://github.com/git-lfs/git-lfs/issues/5464#issuecomment-1836830898.

PR #5504, opened 2023-09-12
Limit CopyCallbackFile to print every 200 ms

cmaves:
Prior to this commit, the callback would print every single iteration of io.Copy. This results in a print every 32kb and
can slow clean operations by several orders of magnitude when GIT_LFS_PROGRESS is set.
By limiting the print to every quarter second in this PR, most of negative impact is mitigated.

### DIFF
--- a/lfs/util.go
+++ b/lfs/util.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/git-lfs/git-lfs/v3/config"
 	"github.com/git-lfs/git-lfs/v3/errors"
@@ -51,12 +52,14 @@ func (f *GitFilter) CopyCallbackFile(event, filename string, index, totalFiles i
 	}
 
 	var prevWritten int64
-
+	deadline := time.Now().Add(time.Millisecond * 250)
 	cb := tools.CopyCallback(func(total int64, written int64, current int) error {
-		if written != prevWritten {
-			_, err := file.Write([]byte(fmt.Sprintf("%s %d/%d %d/%d %s\n", event, index, totalFiles, written, total, filename)))
+		now := time.Now()
+		if written != prevWritten && (!now.Before(deadline) || written == total) {
+			_, err := fmt.Fprintf(file, "%s %d/%d %d/%d %s\n", event, index, totalFiles, written, total, filename)
 			file.Sync()
 			prevWritten = written
+			deadline = now.Add(time.Millisecond * 250)
 			return wrapProgressError(err, event, logPath)
 		}
 
--- a/lfs/util.go
+++ b/lfs/util.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/git-lfs/git-lfs/v3/config"
 	"github.com/git-lfs/git-lfs/v3/errors"
+	"github.com/git-lfs/git-lfs/v3/tasklog"
 	"github.com/git-lfs/git-lfs/v3/tools"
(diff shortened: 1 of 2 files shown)


lfs/util.go
```diff
@@ -51,12 +52,14 @@ func (f *GitFilter) CopyCallbackFile(event, filename string, index, totalFiles i
 	}
 
 	var prevWritten int64
-
+	deadline := time.Now().Add(time.Millisecond * 250)
```
chrisd8088, line 55: It seems to me this is similar in purpose to the `DefaultLoggingThrottle` value that's [defined](https://github.com/git-lfs/git-lfs/blob/2d8549c0348b97d04fa0ac96092a5f7167886e1a/tasklog/log.go#L18) in the `tasklog` package, which [limits](https://github.com/git-lfs/git-lfs/blob/2d8549c0348b97d04fa0ac96092a5f7167886e1a/tasklog/log.go#L211-L212) logging to once per every 200 milliseconds.

Rather than using a slightly different value here, I would suggest at least using that `tasklog.DefaultLoggingThrottle` constant both here and below.

(If we wanted to get fancy we could make this a configurable value, which is then used in both places, but for now I think it's OK to just use the same constant, as I believe we never set the `tasklog.Logger` with any other value than the [default](https://github.com/git-lfs/git-lfs/blob/2d8549c0348b97d04fa0ac96092a5f7167886e1a/tasklog/log.go#L71).)

One other thing we might consider is adding a test or two, akin to the [throttling tests](https://github.com/git-lfs/git-lfs/blob/2d8549c0348b97d04fa0ac96092a5f7167886e1a/tasklog/log_test.go#L120-L172) in the `tasklog` package.  That's a bit of a heavier lift, though.  I'm not sure we have any Go tests of the `lfs.CopyCallbackFile()` function at all, so let us know if we can help introduce some.

lfs/util.go
```diff
@@ -51,12 +52,14 @@ func (f *GitFilter) CopyCallbackFile(event, filename string, index, totalFiles i
...
 	cb := tools.CopyCallback(func(total int64, written int64, current int) error {
-		if written != prevWritten {
-			_, err := file.Write([]byte(fmt.Sprintf("%s %d/%d %d/%d %s\n", event, index, totalFiles, written, total, filename)))
+		now := time.Now()
+		if written != prevWritten && (!now.Before(deadline) || written == total) {
```
chrisd8088, line 58: I don't know if it's necessary, but I might be inclined to write `written >= total` just to cover the exceptional case where more bytes are written than expected.  That would, of course, mean the throttling would also stop and we'd write the progress message on every copy callback, but I think that's tolerable as this would be an exceptional condition which may never actually happen.

chrisd8088 (commented):
Thanks for the patch, and welcome to the Git LFS project!

This looks reasonable to me, on the whole!  I had a couple of suggestions to make, but nothing too dramatic, I hope.

That said, adding some tests may be a bit involved, so let us know if we can help with that.

Thanks again very much for the contribution to the project!

lfs/util.go
cmaves: Done

lfs/util.go
cmaves: I've updated the value to use the `DefaultLoggingThrottle` as the sleep duration. I've added a test for the throttle portion of this. It is not the cleanest test with a fair bit of setup and a sleep in it so if there is a better way to test it I'm all ears.

chrisd8088 (approved):
This looks great, thanks!  I'll kick off CI, and see if we can get this merged.  Thanks again!

Issue #5498, opened 2023-09-07
Installation instructions for Windows could use a note


zpostfacto:
**Describe the issue**
The Windows installation instructions should include a note that you will need to restart your command prompt because the installation modifies your PATH.

It is not obvious to a non-expert that the way "git lfs" works is by locating an executable named "git-lfs" in your PATH, or that installing git lfs would need to modify your path.  I just lost an hour because I followed the installation instructions, tabbed back to my command prompt and typed "git lfs", and git complains that "lfs" is not a known git command.

chrisd8088:
Hey, thanks for the suggestion and I'm sorry you're having trouble.  May I ask, which particular set of instructions were you looking at, so we can adjust them?  For instance, are you using the Git for Windows project, or a binary download of Git LFS, or building from source?  We have a few [different](https://github.com/git-lfs/git-lfs/tree/42bf682677b3f79724601750617e435e37db7b23#on-windows) [sections](https://github.com/git-lfs/git-lfs/tree/42bf682677b3f79724601750617e435e37db7b23#from-binary) of [instructions](https://github.com/git-lfs/git-lfs/tree/42bf682677b3f79724601750617e435e37db7b23#from-source) for each case.

zpostfacto:
Hm, actually none of the above!  So...my apologies if this is not be the best place to report this feedback.

I looked at these two sets of instructions:

https://git-lfs.com/
https://docs.github.com/en/repositories/working-with-files/managing-large-files/installing-git-large-file-storage

chrisd8088:
Thanks for getting back to us!  So out of interest, to make sure we understand your use case, I take it you're not using Git LFS as bundled with Git for Windows, but installed it separately, perhaps after downloading one of our Windows release binaries?  Is that correct?

zpostfacto:
I'm sorry, I cannot remember how I installed git, to be honest.  Can you tell from this by any chance?

```
f:\dev\Steam\main>where git
C:\Program Files\Git\cmd\git.exe

f:\dev\Steam\main>git --version
git version 2.33.0.windows.2
```

But yes, however I installed git, it did not have git-lfs installed and that is what caused me to try to install git-lfs separately.

chrisd8088:
[...]

PR #5507, opened 2023-09-13
Add installation note about restarting shells on Windows for PATH changes

chrisd8088:
As discussed in #5498, Windows users in particular may need to restart their command shell after installing a Git LFS binary in order for any changes to `$PATH` to take effect, so we add a note to our installation instructions on this point.

/cc @zpostfacto as reporter.


zpostfacto:
Nice, maybe it will save somebody a small amount of time / frustration in the future.
Cheers.

bk2204 (approved):

Issue #5511, opened 2023-09-16
LFS can track and clean `.gitattributes`, breaking itself (again)


RuRo:
**Describe the bug**

It seems that issue #1099 and/or PR #1346 either broke again or were never fully fixed in the first place.

**To Reproduce**

```sh
git lfs --version
# git-lfs/3.0.2 (GitHub; linux amd64; go 1.18.1)

git init .
# Initialized empty Git repository in /tmp/tmp.86UIbTNFDJ/.git/

git lfs track '*'
# Tracking "*"

git add .
git commit -m "initial commit"
# [master (root-commit) e3fafb4] initial commit
#  1 file changed, 3 insertions(+)
#  create mode 100644 .gitattributes

cat .gitattributes
# * filter=lfs diff=lfs merge=lfs -text

git rebase --root
# https://git-lfs.github.com/spec/v1 is not a valid attribute name: .gitattributes:1
# sha256:8f1f58fd32af1d9afe3c8d06ebe19deef3656ab96408bbdfb46cf45e8eeb0bc3 is not a valid attribute name: .gitattributes:2
# Successfully rebased and updated refs/heads/master.

cat .gitattributes
```

**Expected behavior**

`.gitattributes` doesn't get affected by LFS even when it is "supposed" to be tracked based on the specified patterns.

**System environment**

I was able to reproduce this issue inside a stock `ubuntu:22.04` container (after a simple `apt-get install git-lfs`).

bk2204:
Hey,

Thanks for the report.  It looks like we do print the correct message, but we still modify the file.  I'll open a PR that should fix this shortly.

Note also that 3.0.2 has different behaviour and doesn't print the correct message.

bk2204:
Unfortunately, whether a file is filtered is determined by Git, and not Git LFS.  There's not really anything we can do here to address this, and this would be broken even without Git LFS if you asked for some other transformation (such as changing the working-tree-encoding to UTF-16).

If you'd like Git to fail in that case, then that's a conversation you need to have with Git upstream.

RuRo:
@bk2204 does the filter not know the path to the file that is being filtered? My idea was to skip processing `.gitattributes` inside the filter itself.

bk2204:
[...]

PR #5515, opened 2023-09-18
track: reject attempts to modify `.gitattributes`

bk2204:
If the user requests a pattern that matches `.gitattributes` and that's the only pattern, we currently return a zero exit status, even though we've failed.  Similarly, if an error occurs, we fail to exit nonzero. In both of these cases, switch to exiting nonzero to indicate that the operation was unsuccessful.

However, continue to return successfully if the reason we didn't process any patterns was because they're already matched, since what the user wanted is already done.

/cc #5511
/cc @RuRo as reporter

### DIFF
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -163,6 +163,8 @@ ArgsLoop:
 		}
 	}
 
+	modified := false
+	sawError := false
 	// Any items left in the map, write new lines at the end of the file
 	// Note this is only new patterns, not ones which changed locking flags
 	for pattern, newline := range changedAttribLines {
@@ -204,6 +206,7 @@ ArgsLoop:
 			// Newline already embedded
 			attributesFile.WriteString(newline)
 		}
+		modified = true
 
 		for _, f := range gittracked {
 			if trackVerboseLoggingFlag || trackDryRunFlag {
@@ -215,6 +218,7 @@ ArgsLoop:
 				err := os.Chtimes(f, now, now)
 				if err != nil {
 					LoggedError(err, tr.Tr.Get("Error marking %q modified: %s", f, err))
+					sawError = true
 					continue
 				}
 			}
@@ -226,6 +230,19 @@ ArgsLoop:
 	err = lockClient.FixFileWriteFlagsInDir(relpath, readOnlyPatterns, writeablePatterns)
 	if err != nil {
 		LoggedError(err, tr.Tr.Get("Error changing lockable file permissions: %s", err))
+		sawError = true
+	}
+
+	if sawError {
+		os.Exit(2)
+	}
+	// If we didn't modify things, but that's because the patterns
+	// were already supported, don't return an error, since what the
+	// user wanted has already been done.
+	// Otherwise, if we didn't modify things but only because the
+	// patterns were disallowed, return an error.
+	if !modified && len(changedAttribLines) > 0 {
+		os.Exit(1)
 	}
 }
 
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -166,11 +166,6 @@ ArgsLoop:
 	// Any items left in the map, write new lines at the end of the file
 	// Note this is only new patterns, not ones which changed locking flags
 	for pattern, newline := range changedAttribLines {
-		if !trackNoModifyAttrsFlag {
-			// Newline already embedded
-			attributesFile.WriteString(newline)
-		}
-
 		// Also, for any new patterns we've added, make sure any existing git
 		// tracked files have their timestamp updated so they will now show as
 		// modified note this is relative to current dir which is how we write
@@ -203,6 +198,11 @@ ArgsLoop:
 		}
 		if matchedBlocklist {
 			continue
+		}
+
+		if !trackNoModifyAttrsFlag {
+			// Newline already embedded
+			attributesFile.WriteString(newline)
 		}
 
 		for _, f := range gittracked {
--- a/t/t-track.sh
+++ b/t/t-track.sh
@@ -377,7 +377,7 @@ begin_test "track blocklisted files by name"
   git add .gitattributes
   git commit -m 'Initial commit'
 
-  git lfs track .gitattributes 2>&1 > track.log
+  git lfs track .gitattributes 2>&1 > track.log && exit 1
   grep "Pattern '.gitattributes' matches forbidden file '.gitattributes'" track.log
   [ -z "$(git status --porcelain | grep -v '^??')" ]
 )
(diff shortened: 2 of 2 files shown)


t/t-track.sh
```diff
@@ -375,9 +375,11 @@ begin_test "track blocklisted files by name"
 
   touch .gitattributes
   git add .gitattributes
+  git commit -m 'Initial commit'
 
   git lfs track .gitattributes 2>&1 > track.log
   grep "Pattern '.gitattributes' matches forbidden file '.gitattributes'" track.log
+  [ -z "$(git st --porcelain | grep -v '^??')" ]
```
chrisd8088, line 382: Both here and in the test below, is this an alias for `git status`?  It may not be something that's installed locally.

I notice this succeeds as a test in CI and on my system because the pipe into `grep` means the non-zero exit code from `git st`, reporting `'st' is not a git command`, doesn't trigger an immediate failure (thanks, `set -e` in Bash!), and since there's nothing on stdout, the `grep -v` then returns nothing and the `-z` passes.

commands/command_track.go
```diff
@@ -226,6 +230,16 @@ ArgsLoop:
...
+		sawError = true
+	}
+
+	if sawError {
+		os.Exit(2)
+	} else if !modified && len(changedAttribLines) > 0 {
+		// If we didn't modify things, but that's because the patterns
+		// were already supported, don't return an error, since what the
+		// user wanted has already been done.
+		os.Exit(1)
```
chrisd8088, line 242: Is it the case that we would get here if the requested patterns are already supported?  IIUC, the loop which [checks](https://github.com/git-lfs/git-lfs/blob/ab3f2288bdb3cface5a4a0af2daa9c302214ab6f/commands/command_track.go#L96-L97) for that case returns before [adding](https://github.com/git-lfs/git-lfs/blob/ab3f2288bdb3cface5a4a0af2daa9c302214ab6f/commands/command_track.go#L107) anything to `changedAttribLines`.  We would get here, though, if a pattern is [reported](https://github.com/git-lfs/git-lfs/blob/ab3f2288bdb3cface5a4a0af2daa9c302214ab6f/commands/command_track.go#L200) as matching a forbidden file, I think.

Should the comment reflect this, and that we are returning an error (i.e., a non-zero exit code)?  Or am I misunderstanding the context?

bk2204: The comment refers to the condition above it.  Would you prefer that I place it above the condition instead?  Unfortunately, Go effectively requires cuddling the brace and the else if, so I can't split them here.

bk2204: I've moved this out into its own separate `if` block to make the comment cleaner.

And yes, without the `len(changedAttribLines) > 0` condition, one of our other tests fails.

chrisd8088: Thanks!  I understood that the comment applied to the condition of the enclosing `if` statement block, but it's definitely even clearer now.

What I'm perhaps fuzzy about is how exiting with a `1` counts as "not returning an error".  Is the intent of this comment to state "we don't return a Go error to the caller" and not "we don't exit with a non-zero exit code, which indicates a general error, from the command itself", which is how I've been interpreting it?

And, separately, I still wonder if it is actually possible to get into this conditional if, as the comment says, "we didn't modify things, but that's because the patterns were already supported".

I see that, for instance, that without this [conditional](https://github.com/git-lfs/git-lfs/blob/8e14a7e89a6ba03f6139e88b3834892fc944e184/commands/command_track.go#L239-L244), the `blocklisted` tests fail when `git lfs track .gitattributes` is [run](https://github.com/git-lfs/git-lfs/blob/8e14a7e89a6ba03f6139e88b3834892fc944e184/t/t-track.sh#L380) or `git lfs track .git*` is [run](https://github.com/git-lfs/git-lfs/blob/8e14a7e89a6ba03f6139e88b3834892fc944e184/t/t-track.sh#L399) and those commands don't return a non-zero (i.e., error) exit code.  That makes sense to me.  When the conditional exists, the tests pass.

But IIUC the situation caught by this conditional block is not one where nothing is being changed because "the patterns were already supported", as the comment states currently.  Rather, nothing is being changed because the patterns are disallowed -- the code would change something, but it's noticing that we're not supposed to change `.git*` or `.gitattributes`, and causing the command to exit with a non-zero error code (which is good!)

It seems to me that the case where we don't change any patterns because the patterns were already supported, like in the example below, is not passing through this conditional.  Rather, the condition is [caught](https://github.com/git-lfs/git-lfs/blob/8e14a7e89a6ba03f6139e88b3834892fc944e184/commands/command_track.go#L96-L97) in a prior loop over the input command-line arguments, and the `continue` in that conditional block means nothing is [added](https://github.com/git-lfs/git-lfs/blob/8e14a7e89a6ba03f6139e88b3834892fc944e184/commands/command_track.go#L107) to `changedAttribLines` since that occurs later in the loop.  Thus when we get to the conditional under discussion here (and assuming there aren't any valid or blocklisted patterns to consider among the input arguments),  `modified` will be false, but `changedAttribLines` will also just be empty, and this conditional doesn't apply.
```
$ git lfs track "*.bin"
Tracking "*.bin"
$ git lfs track "*.bin"
"*.bin" already supported
```

Does that all make sense?

chrisd8088 (approved):

commands/command_track.go
bk2204: If we had just `if !modified` for this block, then we'd return 1 if we had no modified lines because they were all disallowed.  However, that condition would also be triggered if we had no modified lines because all of our patterns were already supported.  The way we know that has occurred is if `changedAttribLines` is empty, since, as you mention, the previous loop will have skipped adding them.

So, with this comment, we're describing part of the behaviour of the implicit and empty else branch.  It's true that exiting 1 is, indeed, returning an error.  However, if we don't take the if-branch, then we return 0 implicitly, which is the non-error case, and what the comment describes.  I agree the comment is confusing if we read it as applying to the if branch.

Hopefully that makes things clearer here.  If you think that the comment should be reworded to refer to the if-branch instead in a clearer way, I can do that.  Maybe something like this would be better:

> If we didn't modify anything successfully, only return an error if we have at least one new or changed pattern that isn't alredy supported.  If all of the patterns were already supported and we didn't modify anything for that reason, we don't consider that an error case and skip the conditional below, implicitly returning 0, because what the user wanted has already been done.

Of course I'm happy to hear other suggestions about how to make this less confusing if you have any.

chrisd8088: Ah, I see -- the comment is intended to apply to the case where we _don't_ enter the conditional block that exits with `os.Exit(1)`!  That makes it make much more sense to me.

I suppose I might just suggest adding a sentence following the comment, because (and maybe this is just me) I read a comment above or within a conditional block as applying to that block.  Let me draft something in a suggestion real quick.

commands/command_track.go
```diff
@@ -226,6 +230,17 @@ ArgsLoop:
...
 		LoggedError(err, tr.Tr.Get("Error changing lockable file permissions: %s", err))
+		sawError = true
+	}
+
+	if sawError {
+		os.Exit(2)
+	}
+	// If we didn't modify things, but that's because the patterns
+	// were already supported, don't return an error, since what the
+	// user wanted has already been done.
```
chrisd8088, line 241:
```suggestion
	// If we didn't modify things, but that's because the patterns
	// were already supported, don't return an error, since what the
	// user wanted has already been done.
	// Otherwise, if we didn't modify things but only because the
	// patterns were disallowed, return an error.
```

chrisd8088 (approved):
I see now what the intent of the comment is; thank you!  I added one suggested extra sentence to help me in the future remember what this is all about, if that's OK.

commands/command_track.go
bk2204: I've squashed that in with a co-author credit.

PR #5522, opened 2023-09-26
Filter --above by path and by type

sleekweasel:
This augments --above, so it can be given:
* `--above-exclude-by-path=a regex of paths to ignore`
* `--above-exclude-by-type=a regex of types to ignore` where type is determined by running them through the unix 'file' utility
Since the latter can be an expensive operation, it can also be given:
* `--above-include-by-path=a regex of paths to include` which skips the 'file' check.

The rationale for using the 'file' utility is that under Unix, executables are not given an extension like '.exe' and so cannot be usefully matched by trivial globs/regexes.

Note that these 'regex' expressions are actually a ';' separated list of regexes, where the first regex is used to match, the second to countermand the result if the first matches, etc, so - for example - '.*ASCII.*;.*postscript.*' would match strings that contained `ASCII` but not `postscript`.

### DIFF
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -38,6 +38,8 @@ var (
 	// above the provided size.
 	migrateImportAboveFmt string
 
+	migrateImportAboveOnlyBinary bool
+
 	// migrateEverything indicates the presence of the --everything flag,
 	// and instructs 'git lfs migrate' to migrate all local references.
 	migrateEverything bool
@@ -396,6 +398,7 @@ func init() {
 
 	importCmd := NewCommand("import", migrateImportCommand)
 	importCmd.Flags().StringVar(&migrateImportAboveFmt, "above", "", "--above=<n>")
+	importCmd.Flags().BoolVar(&migrateImportAboveOnlyBinary, "above-only-binary", false, "--above only applies to binary files")
 	importCmd.Flags().BoolVar(&migrateVerbose, "verbose", false, "Verbose logging")
 	importCmd.Flags().StringVar(&objectMapFilePath, "object-map", "", "Object map file")
 	importCmd.Flags().BoolVar(&migrateNoRewrite, "no-rewrite", false, "Add new history without rewriting previous")
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -38,7 +38,24 @@ var (
 	// above the provided size.
 	migrateImportAboveFmt string
 
-	migrateImportAboveOnlyBinary bool
+	// migrateImportAboveNotByTypeFmt indicates the presence of the
+	// --above-not-by-type=<regexes> flag and modifies --above=<size> to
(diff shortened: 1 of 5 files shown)


bk2204:
Hey,

First of all, thanks for sending a patch.

I have some concerns about the usage of `file` here.  In general, there are no portable options to that command; for example, `-k` doesn't exist for OpenBSD and neither option exists on Solaris.  (Git LFS is available on both.)  I also believe there are some environments where people are using Portable Git on Windows where `file` will be missing.  In addition, the text is definitely not stable across systems and isn't guaranteed to be stable across versions, so I'm not sure that we can use that as a pattern.  (The `file(1)` manual page famously claims, “Don't do as Berkeley did and change ‘shell commands text’ to ‘shell script’”, which I think illustrates the problem very well.)

I think if we're adding inclusions and exclusions, we'd probably want to make them work like the existing ones do using glob patterns (and use the existing options) rather than regexes.  That's more like what Git does, and while they're a little less powerful, we'd like to preserve as much of the behavioural similarity to Git as possible.

I'm also worried that this behaviour is perhaps a little too niche, especially since it relates specifically to `--above` and isn't more general.  I don't have a great solution here, though; `git filter-branch` is a possible approach that allows people to use a shell to customize things, but it's also quite slow and has some sharp edges.

I'm interested to hear a little more about your potential use cases for this PR to understand if there's maybe some places we could improve.

sleekweasel:
Hi

I did find myself thinking about Windows users who probably wouldn't have 'file' but I didn't consider that different versions of file might not support -k. I wasn't so worried about different text output between different versions, given that it seemed users will only run migrate occasionally and on a single system, and will have context about what 'file' actually returns for them. I take your point about sticking with globs, but again - import --above didn't seem likely to be a frequent thing for an individual to run.

Our aim was probably like most folks': to LFS as much as practical with limited prior experience doing it, particularly for archaological files everyone had forgotten about, to take our 30G repo down to something in the 5G range, to avoid LFSing things that seemed likely to compress well, while keeping as much history as possible to keep the anti-squashers happy, and working round the problem that GitHub Enterprise doesn't support Nexus-based LFS storage, so image diffs don't work (without a chrome plugin that's likely to stop working when they restrict functionality still further in Manifest v4, or other solutions). And that LFSing all our pngs resulted in unacceptably slow checkouts (to the point it was much faster to archive the lfs cache and distribute it with curl - another reason we didn't eventually LFS them).

Ultimately, I think we're probably better off with what we've settled on, which is a monster --import and --export for migration, and then trimming out the archeological directives from the resulting gitattributes. Notably the executables turned out mostly to be in **/*.(xc)?framework/** directories, so but since I'd already coded this during the investigation, it seemed reasonable to offer it for your consideration, or at least for someone else to perhaps find useful. We hope to find time to write up our experience, including my colleague's methods for digging around in ancient history, and include some newbie nitty-gritty details we didn't understand until afterwards.

Issue #5535, opened 2023-10-03
Pure SSH protocol deadlocks when SSH multiplexing is enabled


KyleFromKitware:
**Describe the bug**
When using the pure SSH `git-lfs-transfer` protocol, upon ending the session, `git-lfs` deadlocks unless `lfs.ssh.automultiplex` is set to `false`. This is because `git-lfs` tries to terminate the master ssh instance first, which never ends because it's waiting for all the other multiplexed connections to end.

**To Reproduce**
1. `git lfs push` to a remote that supports the pure SSH protocol
2. See that `git lfs` deadlocks while waiting for its ssh instances to end

**Expected behavior**
`git lfs push` should not deadlock. The deadlock goes away when `lfs.ssh.multiplex` is set to `false`.

**System environment**
Ubuntu 20.04 with an in-development GitLab Shell SSH server

**Output of `git lfs env`**
```
git-lfs/3.4.0 (GitHub; linux amd64; go 1.21.1)
git version 2.25.1

Endpoint (gitlab/ssh)=https://localhost/root/lfs-test.git/info/lfs (auth=none)
  SSH=git@localhost:/root/lfs-test.git
LocalWorkingDir=
LocalGitDir=<REDACTED>/lfs-test.git
LocalGitStorageDir=<REDACTED>/lfs-test.git
LocalMediaDir=<REDACTED>/lfs-test.git/lfs/objects
LocalReferenceDirs=
TempDir=<REDACTED>/lfs-test.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
FetchRecentRefsIncludeRemotes=true
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=<REDACTED>/lfs-test.git/lfs
AccessDownload=none
AccessUpload=none
DownloadTransfers=basic,lfs-standalone-file,ssh
UploadTransfers=basic,lfs-standalone-file,ssh
git config filter.lfs.process = ""
git config filter.lfs.smudge = ""
git config filter.lfs.clean = ""
```

**Additional context**
Note that I encountered this while working on adding `git-lfs-transfer` support to GitLab. I don't know if this is reproducible with other SSH servers.

bk2204:
Hey,

[...]

PR #5537, opened 2023-10-04
ssh: Specifically designate a master multiplex connection

KyleFromKitware:
SSHTransfer.Shutdown() was attempting to first shut down the first connection it created (which happened to be the master connection), but this deadlocked because the master connection was waiting for the extra connections to shut down. Designate the first connection as the master connection, make sure that it truly is the master connection, and shut it down after shutting down all extra connections.

Issue: #5535


ssh/ssh.go
```diff
@@ -163,11 +140,18 @@ func GetExeAndArgs(osEnv config.Environment, gitEnv config.Environment, meta *SS
...
-		controlPath, err := getControlDir(osEnv)
-		if err == nil {
-			multiplexing = true
-			controlPath = filepath.Join(controlPath, "sock-%C")
-			args = append(args, "-oControlMaster=auto", fmt.Sprintf("-oControlPath=%s", controlPath))
+		multiplexing = true
+		controlMasterArg := "-oControlMaster=no"
+		controlPath = multiplexControlPath
+		if multiplexControlPath == "" {
+			controlMasterArg = "-oControlMaster=yes"
```
bk2204, line 147: How does this work on the Windows OpenSSH, which [doesn't support `ControlMaster`](https://github.com/PowerShell/Win32-OpenSSH/issues/1328)?  Is it at least no less functional?

ssh/ssh.go
```diff
@@ -100,41 +100,18 @@ func getVariant(osEnv config.Environment, gitEnv config.Environment, basessh str
...
-	if err != nil {
-		// Ideally we would use errors.Is here to check against
-		// os.ErrExist, but that's not available on Go 1.11.
-		perr, ok := err.(*os.PathError)
-		if !ok || perr.Err != syscall.EEXIST {
-			return os.MkdirTemp(tmpdir, pattern)
-		}
-	}
-	return dir, nil
+	return os.MkdirTemp(tmpdir, pattern)
```
bk2204, line 109: There's no explanation here on why we've removed the functionality to store sockets in `XDG_RUNTIME_DIR`.  Unlike `/tmp`, that's guaranteed to live as long as the user's session and won't get cleaned up by a temporary file cleaner until the system reboots.  It's also guaranteed to be private and on a file system that is fully functional; that is, it supports sockets.

If we need to pick a random socket or create a temporary directory in that directory, I'd rather we do that than remove the `XDG_RUNTIME_DIR` functionality.

bk2204 (commented):
Thanks for the patch.  This looks like it's going in the right direction, but I have a few questions.

ssh/ssh.go
KyleFromKitware: What happens right now if you pass `-oControlMaster=auto` to Windows OpenSSH? Does it just not do anything with it? If so, does `-oControlMaster=yes` or `-oControlMaster=no` also not do anything?

ssh/ssh.go
KyleFromKitware: > If we need to pick a random socket or create a temporary directory in that directory

I basically need a unique socket for every instance of `SSHTransfer`. Making a temporary directory in `XDG_RUNTIME_DIR` would probably do the trick.

> I'd rather we do that than remove the XDG_RUNTIME_DIR functionality.

Sure, I could do that.

ssh/ssh.go
KyleFromKitware: Looking at that issue, it seems that the functionality is so broken that they haven't even implemented a "This doesn't work on Windows" error based on [this comment](https://github.com/PowerShell/Win32-OpenSSH/issues/1328#issuecomment-922003814):

> Can we at least have the client report that the feature is not supported instead of printing opaque errors about sockets, as though the feature were implemented but happened to have suffered some incidental mishap?

Would this test be sufficient?

```go
if variant == variantSSH && runtime.GOOS != "windows" && multiplexDesired && multiplexEnabled {
```

bk2204: I don't think that's the right behaviour, since it apparently does work for Git for Windows's OpenSSH.  Perhaps the default behaviour of the setting above should be `runtime.GOOS != "windows"` instead of `true`.  That would at least allow users to override it if they're using the OpenSSH from Git for Windows.

KyleFromKitware: OK, I've implemented your suggestion.

KyleFromKitware: Actually, now I'm wondering if we should simply have Windows OpenSSH users opt out instead of requiring Git for Windows OpenSSH users to opt in. Seems to me like someone who's using Git is more likely to use Git for Windows OpenSSH. WDYT?

bk2204: Sure, that seems fine.  If the default behaviour doesn't cause any problems (as the message you quoted implies) and is just less functional, then it should be fine to leave it enabled by default.

KyleFromKitware: > If the default behaviour doesn't cause any problems (as the message you quoted implies)

Well, to me the message implies that `-oControlMaster=yes` on Windows produces cryptic errors, which means this behavior doesn't work. My thinking was that if someone's using Git on Windows, they're more likely to be using Git's OpenSSH, so we should cater to them by default and just hope they're not using Windows OpenSSH. However, if we want to make things always work by default, even for the Windows OpenSSH minority, then this should be off by default on Windows.

KyleFromKitware: I'm keeping the multiplexing off by default on Windows, and (I think) I've fixed the CI issues.

KyleFromKitware:
One of the macOS tests failed. It looks unrelated to my changes. My guess is it's a spurious failure. Could you please run it again?

bk2204 (approved):

PR #5554, opened 2023-10-25
Update notarization to use `notarytool`

### LINKED ISSUE #5394: Fix notarization on macOS
As of November 1, 2023, Apple will no longer accept notarized apps using `altool`, which is used by https://github.com/mitchellh/gon, which we're using.  For v3.5.0, we'll need to either make sure that gon is updated (which it may or may not be) or use a different blocking notarization tool.

Note that Apple's notarization tool `altool` isn't blocking, which is why we don't use it, since it can take some time to complete the notarization but we don't want to complete the operation until it succeeds.

bk2204:
As of November 1, 2023, Apple will no longer accept notarized apps using altool, which is used by https://github.com/mitchellh/gon, which we've been using.  Port this code to notarytool, which is Apple's preferred command-line tool.

The secrets have been placed in the regular and CI test secret stores already and this has worked successfully on a test build of the release workflow.

Fixes #5394

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -100,6 +100,7 @@ jobs:
       env:
         DARWIN_DEV_USER: ${{secrets.MACOS_DEV_USER}}
         DARWIN_DEV_PASS: ${{secrets.MACOS_DEV_PASS}}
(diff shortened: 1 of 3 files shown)


chrisd8088 (approved):
Awesome, thanks!

PR #5555, opened 2023-10-25
Allow configuring the SSH protocol

bk2204:
Some people have very slow SSH round trips and the extra time to attempt a failed git-lfs-transfer operation is bothersome.  Let's add an option to allow users to use either the pure SSH protocol or the older hybrid protocol based on the URL in question to let them choose the right value if they know what it should be.

### DIFF
--- a/docs/man/git-lfs-config.adoc
+++ b/docs/man/git-lfs-config.adoc
@@ -318,6 +318,13 @@ Supports URL config lookup as described in:
 https://git-scm.com/docs/git-config#git-config-httplturlgt. To set this
 value per-host:
 `git config --global lfs.https://github.com/.locksverify [true|false]`.
+* `lfs.sshtransfer` / `lfs.<url>.sshtransfer`
++
+Configures whether SSH transfers (the pure SSH protocol) are used.
+By default (or if the value is set to `negotiate`), the pure SSH protocol is
+tried first, and then the older hybrid protocol.  If `always` is used, then
+only the pure SSH protocol is tried.  Similarly, if `never` is used, then only
+the hybrid protocol is attempted.
 * `lfs.<url>.contenttype`
 +
 Determines whether Git LFS should attempt to detect an appropriate HTTP
--- a/lfshttp/client.go
+++ b/lfshttp/client.go
@@ -209,6 +209,12 @@ func (c *Client) sshResolveWithRetries(e Endpoint, method string) (*sshAuthRespo
 	var sshRes sshAuthResponse
 	var err error
 
+	uc := config.NewURLConfig(c.gitEnv)
+	if val, ok := uc.Get("lfs", e.OriginalUrl, "sshtransfer"); ok && val != "negotiate" && val != "never" {
+		tracerx.Printf("skipping SSH-HTTPS hybrid protocol connection by request")
+		return nil, errors.New("git-lfs-authenticate has been disabled by request")
+	}
+
 	requests := tools.MaxInt(0, c.sshTries) + 1
 	for i := 0; i < requests; i++ {
 		sshRes, err = c.SSH.Resolve(e, method)
--- a/t/t-locks.sh
+++ b/t/t-locks.sh
@@ -72,6 +72,16 @@ begin_test "list a single lock (SSH; git-lfs-authenticate)"
   grep "f.dat" locks.log
   grep "Git LFS Tests" locks.log
   grep "lfs-ssh-echo.*git-lfs-authenticate /$reponame download" trace.log
+
+  GIT_TRACE=1 git -c lfs."$sshurl".sshtransfer=never lfs locks --path "f.dat" 2>trace.log | tee locks.log
+  [ $(wc -l < locks.log) -eq 1 ]
+  grep "f.dat" locks.log
+  grep "Git LFS Tests" locks.log
+  grep "lfs-ssh-echo.*git-lfs-authenticate /$reponame download" trace.log
+  grep "skipping pure SSH protocol" trace.log
+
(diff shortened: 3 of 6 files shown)


docs/man/git-lfs-config.adoc
```diff
@@ -318,6 +318,13 @@ Supports URL config lookup as described in:
 https://git-scm.com/docs/git-config#git-config-httplturlgt. To set this
 value per-host:
 `git config --global lfs.https://github.com/.locksverify [true|false]`.
+* `lfs.<url>.sshprotocol`
```
chrisd8088, line 321: As it happens, I've been slowly working on some revised and expanded documentation for many of our config settings, and in the process have written some for the currently undocumented `lfs.gitProtocol` setting, which takes values like `http` or `https`; the intent is that whatever is specified (with the default being `https`) is then substituted if the original URL begins with the `git://` scheme.  So it's a full protocol substitution.

This new setting has a different intent, isn't to change from the SSH protocol to something else entirely, but rather to adjust "how much" SSH to use, as it were.  So I wonder if we could use a different term here other than "protocol", although I appreciate the challenge in coming up with something clear and succinct.

One alternative which occurred to me is something like `lfs.sshTransferProtocol` or `lfs.sshAdapterProtocol`; another, which I maybe like better, might be `lfs.sshTransfer`, which aligns with the `lfs.customTransfer.*` settings.

Following from that last idea, I then wondered if it might be reasonable to allow as the possible values some simpler terms such as `always`, `never`, and `negotiate` (the default), with the goal of making the configuration settings more readable to the lay user.

I also toyed with the idea of having two Boolean settings instead of a single ternary one, maybe something like:
- `lfs.sshTransfer` - if false, `git-lfs-authenticate` is always used; otherwise, `git-lfs-transfer` may be used (default: true)
- `lfs.sshTransfer.negotiate` - only meaningful if `lfs.sshTransfer` is true, in which case it indicates whether or not to try `git-lfs-transfer` before falling back to `git-lfs-authenticate` (default: true)

That would avoid the need to come up with any terms at all for the settings, which I kind of appreciate, because naming things is hard.

t/t-locks.sh
```diff
@@ -72,6 +72,17 @@ begin_test "list a single lock (SSH; git-lfs-authenticate)"
   grep "f.dat" locks.log
   grep "Git LFS Tests" locks.log
   grep "lfs-ssh-echo.*git-lfs-authenticate /$reponame download" trace.log
+
+  GIT_TRACE=1 git -c lfs."$sshurl".sshprotocol=git-lfs-authenticate lfs locks --path "f.dat" 2>trace.log | tee locks.log
+  cat trace.log
```
chrisd8088, line 77: This `cat` and the ones in the subsequent test could probably be removed; I wondered if they were left over from the test construction and debugging process.

chrisd8088 (commented):
This is a great enhancement for people who want to simplify their setups and avoid the SSH "probe" for full transfer protocol support on the server side.

My one main comment is whether we might profit by trying to align the configuration option's name and possible settings a bit more with some of our existing configuration terminology, and maybe splitting it into two Boolean options (which avoids a whole set of naming problems!)  I kind of like that latter approach, I think, but I'm open to a range of alternatives.  Mostly I just thought we should avoid an accidental misalignment with `lfs.gitProtocol` and could perhaps also find some simpler terms to describe the valid options of the new setting.

docs/man/git-lfs-config.adoc
chrisd8088: On reflection, `lfs.sshTransfer.negotiate` is a bit ambiguous and maybe suggests some kind of negotiation of the underlying SSH protocol or the LFS transfer protocol over SSH, and neither is the case (particularly not the former).

What about something like `lfs.sshTransfer.httpFallback` or `lfs.sshTransfer.basicTransferFallback`, which might make it really clear what's being enabled/disabled by a Boolean value?

lfshttp/client.go
```diff
@@ -209,6 +209,12 @@ func (c *Client) sshResolveWithRetries(e Endpoint, method string) (*sshAuthRespo
 	var sshRes sshAuthResponse
 	var err error
 
+	uc := config.NewURLConfig(c.gitEnv)
+	if val, ok := uc.Get("lfs", e.OriginalUrl, "sshprotocol"); ok && val != "auto" && val != "git-lfs-authenticate" {
+		tracerx.Printf("skipping SSH-HTTPS hybrid protocol connection by request")
```
KyleFromKitware, line 214: +1 for adding this. While adding pure-SSH support to GitLab, I've had to take care to make sure git-lfs isn't falling back to `git-lfs-authenticate` in the event that `git-lfs-transfer` fails, resulting in me not noticing that my GitLab `git-lfs-transfer` code doesn't actually work. It would be nice to have a way to ensure that only `git-lfs-transfer` is ever attempted.

docs/man/git-lfs-config.adoc
bk2204: I think having a single tri-state option is better here.  I'll take `lfs.sshTransfer` with `always`, `never`, and `negotiate`.

t/t-locks.sh
bk2204: I've removed them.

docs/man/git-lfs-config.adoc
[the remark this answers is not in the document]
bk2204: Done.

bk2204:
This looks green and is ready for another review when it's convenient.

chrisd8088 (approved):

docs/man/git-lfs-config.adoc
chrisd8088: Sounds good, thanks!

PR #5561, opened 2023-10-27
Improve locking performance

bk2204:
When performing locking for multiple files, we can spend a lot of time repeating the same operations once for each file.  In most cases, we don't need to do that, and we can get better performance by doing the operation only once.  This is going to be especially noticeable on Windows, where spawning a process is more expensive.

### DIFF
--- a/commands/command_lock.go
+++ b/commands/command_lock.go
@@ -23,6 +23,11 @@ func lockCommand(cmd *cobra.Command, args []string) {
 		cfg.SetRemote(lockRemote)
 	}
 
+	lockData, err := computeLockData()
+	if err != nil {
+		ExitWithError(err)
+	}
+
 	refUpdate := git.NewRefUpdate(cfg.Git, cfg.PushRemote(), cfg.CurrentRef(), nil)
 	lockClient := newLockClient()
 	lockClient.RemoteRef = refUpdate.RemoteRef()
@@ -31,7 +36,7 @@ func lockCommand(cmd *cobra.Command, args []string) {
 	success := true
 	locks := make([]locking.Lock, 0, len(args))
 	for _, path := range args {
-		path, err := lockPath(path)
+		path, err := lockPath(lockData, path)
 		if err != nil {
 			Error(err.Error())
 			success = false
@@ -67,6 +72,28 @@ func lockCommand(cmd *cobra.Command, args []string) {
 	}
 }
 
+type lockData struct {
+	rootDir    string
+	workingDir string
+}
+
+// computeLockData computes data about the given repository and working
+// directory to use in lockPath.
+func computeLockData() (*lockData, error) {
+	wd, err := tools.Getwd()
+	if err != nil {
+		return nil, err
+	}
+	wd, err = tools.CanonicalizeSystemPath(wd)
+	if err != nil {
+		return nil, err
+	}
+	return &lockData{
+		rootDir:    cfg.LocalWorkingDir(),
+		workingDir: wd,
+	}, nil
+}
+
 // lockPaths relativizes the given filepath such that it is relative to the root
 // path of the repository it is contained within, taking into account the
 // working directory of the caller.
@@ -74,42 +101,28 @@ func lockCommand(cmd *cobra.Command, args []string) {
 // lockPaths also respects different filesystem directory separators, so that a
 // Windows path of "\foo\bar" will be normalized to "foo/bar".
 //
-// If the root directory, working directory, or file cannot be
-// determined, an error will be returned. If the file in question is
-// actually a directory, an error will be returned. Otherwise, the cleaned path
-// will be returned.
+// If the file path cannot be determined, an error will be returned. If the file
+// in question is actually a directory, an error will be returned. Otherwise,
+// the cleaned path will be returned.
 //
 // For example:
 //   - Working directory: /code/foo/bar/
 //   - Repository root: /code/foo/
 //   - File to lock: ./baz
 //   - Resolved path bar/baz
-func lockPath(file string) (string, error) {
-	repo, err := git.RootDir()
-	if err != nil {
-		return "", err
-	}
-
-	wd, err := os.Getwd()
-	if err != nil {
-		return "", err
-	}
-	wd, err = tools.CanonicalizeSystemPath(wd)
-	if err != nil {
-		return "", errors.Wrapf(err,
-			tr.Tr.Get("could not follow symlinks for %s", wd))
-	}
-
+func lockPath(data *lockData, file string) (string, error) {
 	var abs string
+	var err error
+
 	if filepath.IsAbs(file) {
 		abs, err = tools.CanonicalizeSystemPath(file)
 		if err != nil {
 			return "", errors.New(tr.Tr.Get("unable to canonicalize path %q: %v", file, err))
 		}
 	} else {
-		abs = filepath.Join(wd, file)
+		abs = filepath.Join(data.workingDir, file)
 	}
-	path, err := filepath.Rel(repo, abs)
+	path, err := filepath.Rel(data.rootDir, abs)
 	if err != nil {
 		return "", err
 	}
--- a/commands/command_locks.go
+++ b/commands/command_locks.go
@@ -19,7 +19,11 @@ var (
 )
 
 func locksCommand(cmd *cobra.Command, args []string) {
-	filters, err := locksCmdFlags.Filters()
+	lockData, err := computeLockData()
+	if err != nil {
+		ExitWithError(err)
+	}
+	filters, err := locksCmdFlags.Filters(lockData)
 	if err != nil {
 		Exit(tr.Tr.Get("Error building filters: %v", err))
 	}
@@ -154,11 +158,11 @@ type locksFlags struct {
 }
 
 // Filters produces a filter based on locksFlags instance.
-func (l *locksFlags) Filters() (map[string]string, error) {
+func (l *locksFlags) Filters(data *lockData) (map[string]string, error) {
 	filters := make(map[string]string)
 
 	if l.Path != "" {
-		path, err := lockPath(l.Path)
+		path, err := lockPath(data, l.Path)
 		if err != nil {
 			return nil, err
 		}
--- a/commands/command_unlock.go
+++ b/commands/command_unlock.go
@@ -58,6 +58,11 @@ func unlockCommand(cmd *cobra.Command, args []string) {
 		cfg.SetRemote(lockRemote)
 	}
 
+	lockData, err := computeLockData()
+	if err != nil {
+		ExitWithError(err)
+	}
+
 	refUpdate := git.NewRefUpdate(cfg.Git, cfg.PushRemote(), cfg.CurrentRef(), nil)
 	lockClient := newLockClient()
 	lockClient.RemoteRef = refUpdate.RemoteRef()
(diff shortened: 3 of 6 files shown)


commands/command_lock.go
```diff
@@ -67,6 +73,33 @@ func lockCommand(cmd *cobra.Command, args []string) {
 	}
 }
 
+type lockData struct {
+	rootDir    string
+	workingDir string
+}
+
+func computeLockData() (*lockData, error) {
+	repo, err := git.RootDir()
```
chrisd8088, line 82: Looking at this in the context of this PR's refactoring, I notice that this [calls](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/git/git.go#L790) `git rev-parse --show-toplevel`, then [performs](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/git/git.go#L800) a Cygwin translation, and then [calls](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/git/git.go#L804) `tools.CanonicalizePath()`, which then starts by [running](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/tools/filetools.go#L486) another Cygwin translation before [calling](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/tools/filetools.go#L495) `tools.CanonicalizeSystemPath()`.

So, first of all, I think we should drop that first redundant [call](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/git/git.go#L800-L803) to `tools.TranslateCygwinPath()` in `git.RootDir()`, since it'll just be run again when that calls `tools.CanonicalizePath()`.

Next, I think this function might profit by just using `cfg.LocalWorkingDir()` instead of calling `git.RootDir()`, which will always run `git rev-parse`, whereas `cfg.LocalWorkingDir()` will first [run](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/config/config.go#L367-L368) its `loadGitDirs()` internal function, which [skips](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/config/config.go#L380-L382) taking action if `git rev-parse` has already been used to load the various directory paths.  If the paths aren't known yet, then it [runs](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/config/config.go#L384) `tools.GitAndRootDirs()` which [invokes](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/git/git.go#L749) the `git rev-parse ... --show-toplevel` command and [passes](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/git/git.go#L785) the result through `tools.CanonicalizePath()`.

(The `loadGitDirs()` function then also [uses](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/config/config.go#L396) `tools.ResolveSymlinks()` on that path, which `git.RootDir()` and `git.GitAndRootDirs()` don't do themselves.  That may not be necessary, though, because that just [calls](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/tools/filetools.go#L64) `tools.CanonicalizeSystemPath()` which `git.RootDir()` and `git.GitAndRootDirs()` have already done by [calling](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/git/git.go#L785) `tools.CanonicalizePath()` which also [calls](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/tools/filetools.go#L495) `tools.CanonicalizeSystemPath()`.  So that's yet another possible simplification, removing the `tools.ResolveSymlinks()` [calls](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/config/config.go#L394-L396) in `loadGitDirs()`.)

Whew ... anyway, my point in all of that is that if we use `cfg.LocalWorkingDir()` here I think we'll avoid any excess calls to `git rev-parse`, and that ties in with my comment above regarding moving the `computeLockData()` call to the top of `lockCommand()`, before `newLockClient()`, since that also [uses](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/commands/commands.go#L107) `cfg.LocalWorkingDir()` and therefore will get the benefit of this function having already run and cached the output from `git rev-parse`.  I hope that all makes sense!

commands/command_lock.go
```diff
@@ -30,8 +30,14 @@ func lockCommand(cmd *cobra.Command, args []string) {
 
 	success := true
 	locks := make([]locking.Lock, 0, len(args))
+	lockData, err := computeLockData()
+	if err != nil {
+		lockClient.Close()
```
chrisd8088, line 35: I think we could avoid the need to call `lockClient.Close()` here by simply making this `computeLockData()` call further up, before the `newLockClient()` [call](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/commands/command_lock.go#L27).  That's what is already being done in this PR in the `commands/command_locks.go` source file, so if we do that here and in `commands/command_unlock.go` I think we can dodge the problem of deferred functions not running during `os.Exit()`.

commands/command_locks.go
```diff
@@ -19,7 +19,12 @@ var (
 )
 
 func locksCommand(cmd *cobra.Command, args []string) {
-	filters, err := locksCmdFlags.Filters()
+	lockData, err := computeLockData()
+	if err != nil {
+		Error(err.Error())
+		os.Exit(3)
```
chrisd8088, line 25: My comments above about `ExitWithError()` apply here too, obviously enough.

I like that we're running `computeLockData()` here before `newLockClient()` so we don't have to close the client on error.

commands/command_unlock.go
```diff
@@ -65,9 +65,15 @@ func unlockCommand(cmd *cobra.Command, args []string) {
 
 	locks := make([]unlockResponse, 0, len(args))
 	success := true
+	lockData, err := computeLockData()
+	if err != nil {
+		lockClient.Close()
+		Error(err.Error())
+		os.Exit(3)
+	}
```
chrisd8088, line 73: All the thoughts I had about this construction in the `commands/command_lock.go` file apply here too, of course.

locking/locks.go
```diff
@@ -67,6 +67,8 @@ func NewClient(remote string, lfsClient *lfsapi.Client, cfg *config.Configuratio
 		cache:              &nilLockCacher{},
 		cfg:                cfg,
 		ModifyIgnoredFiles: lfsClient.GitEnv().Bool("lfs.lockignoredfiles", false),
+		LocalWorkingDir:    cfg.LocalWorkingDir(),
+		LocalGitDir:        cfg.LocalGitDir(),
```
chrisd8088, line 71: Do we make use of `LocalGitDir` at all?  If not, perhaps we can drop it here, and add it later should it become needed in a future change.

Big :+1: to the use of `cfg.LocalWorkingDir()` here because it validates all the comments I made about the preceding commit, I think!

chrisd8088 (approved):
This has a whole bunch of efficiency improvements, thank you!

I think I might have spotted a few more possible ones (but might be mistaken!) and had some minor notes on how to structure the calls to the new `computeLockData()` helper function; that's about it, though.

commands/command_lock.go
```diff
@@ -30,8 +30,14 @@ func lockCommand(cmd *cobra.Command, args []string) {
...
+		Error(err.Error())
+		os.Exit(3)
```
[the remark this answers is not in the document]
bk2204: I always like to use new, different error codes so that we can distinguish various cases, which I think is valuable to the user.  I hadn't intended to add any new tests for this case, though.  If you prefer, I can change it back.

locking/locks.go
bk2204: It's initialized elsewhere, and I think it's better to initialize them both at the same time, since it makes it easier to use them both together.

commands/command_lock.go
```diff
@@ -84,32 +117,19 @@ func lockCommand(cmd *cobra.Command, args []string) {
 //   - Repository root: /code/foo/
 //   - File to lock: ./baz
 //   - Resolved path bar/baz
```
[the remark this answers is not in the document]
bk2204: I've updated the comments to reflect what the function does.

commands/command_lock.go
```diff
@@ -67,6 +73,33 @@ func lockCommand(cmd *cobra.Command, args []string) {
...
+	workingDir string
+}
+
+func computeLockData() (*lockData, error) {
+	repo, err := git.RootDir()
+	if err != nil {
+		return nil, err
+	}
+
+	wd, err := os.Getwd()
```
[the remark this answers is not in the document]
bk2204: Done.

bk2204:
Other than the case I mentioned where I think it makes sense to initialize both values at once, I've gone ahead and made all the suggested changes.

commands/command_lock.go
chrisd8088: Thanks for the fixup!  I'm just going to highlight two notes from my previous comments so I don't forget that they things we could tune in another PR, but they're OK for now.
> So, first of all, I think we should drop that first redundant [call](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/git/git.go#L800-L803) to `tools.TranslateCygwinPath()` in `git.RootDir()`, since it'll just be run again when that calls `tools.CanonicalizePath()`.

> (The `loadGitDirs()` function then also [uses](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/config/config.go#L396) `tools.ResolveSymlinks()` on that path, which `git.RootDir()` and `git.GitAndRootDirs()` don't do themselves.  That may not be necessary, though, because that just [calls](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/tools/filetools.go#L64) `tools.CanonicalizeSystemPath()` which `git.RootDir()` and `git.GitAndRootDirs()` have already done by [calling](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/git/git.go#L785) `tools.CanonicalizePath()` which also [calls](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/tools/filetools.go#L495) `tools.CanonicalizeSystemPath()`.  So that's yet another possible simplification, removing the `tools.ResolveSymlinks()` [calls](https://github.com/git-lfs/git-lfs/blob/df9b79dc4b96547127c91e03688a7d9cb1619ae7/config/config.go#L394-L396) in `loadGitDirs()`.)

chrisd8088:
This looks great and I've approved it already, so please merge as and when you have time!

PR #5589, opened 2023-12-13
Fix git-scm.com links

tigsikram:
The anchors in [git-config](https://git-scm.com/docs/git-config) have changed.


bk2204 (approved):
Thanks for the patch, and welcome to Git LFS!  This looks good.

PR #5590, opened 2023-12-14
Update release documentation and changelog summary script for patch releases

chrisd8088:
When we prepare a patch release of Git LFS, we cherry-pick specific merged changes from the `main` branch onto a `release-next` branch using the `git cherry-pick` command with the `-m1` option, which flattens all the commits unique to the merge on the `main` branch into a single non-merge commit on the release branch.

As a result, when we run our `script/changelog` script to generate a summary of the changes for the patch release, the `git rev-list` command it [executes](https://github.com/git-lfs/git-lfs/blob/7832469836703657c55b85e9cba1b2f7809dfc12/script/changelog#L24) returns nothing because it uses the `--merges` option, and the flattened commits on the release branch are not actual merge commits.

We therefore add a `--patch` option (as suggested by @bk2204) to our script which runs `git rev-list` without the `--merges` option, and note its use in the appropriate step in the how-to documentation of our Git LFS release process.

We also add the `-s` option to the `git show` command the script [runs](https://github.com/git-lfs/git-lfs/blob/7832469836703657c55b85e9cba1b2f7809dfc12/script/changelog#L67) so as to avoid cluttering the output with the patch diff for each commit it presents for categorization by the user.  When run for a non-patch Git LFS release, this is not important because the commits found by `git rev-list` are all merges and generally have no blob changes directly associated with them, so the `git show` output is relatively succinct.  But with the commits on the release branch for a Git LFS patch release, which are not merges, the `git show` output without the `-s` option is potentially quite verbose.

We also expand our documentation's description of several of our release steps with additional notes, and fix one typo.

### DIFF
--- a/docs/howto/release-git-lfs.md
+++ b/docs/howto/release-git-lfs.md
@@ -101,8 +101,8 @@ to zero, we are releasing a PATCH version.
        MINOR release number in that series.  For a MINOR release, use
        `script/changelog vM.(N-1).0...HEAD`, where `(N-1)` is the previous
        MINOR release number, and for a PATCH release, use
-       `script/changelog vM.N.(P-1)...HEAD`, where `(P-1)` is the previous
-       PATCH release number.
+       `script/changelog --patch vM.N.(P-1)...HEAD`, where `(P-1)` is the
+       previous PATCH release number.
 
        * Optionally write 1-2 paragraphs summarizing the release and calling
          out community contributions.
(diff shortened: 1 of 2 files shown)


docs/howto/release-git-lfs.md
```diff
@@ -117,11 +117,17 @@ to zero, we are releasing a PATCH version.
          `release-M.(N-1)`.)
 
      * Run `script/update-version vM.N.P` to update the version number in all
-       of the relevant files.
+       of the relevant files.  Note that this script requires a version of
+       `sed(1)` compatible with the GNU implementation.
```
bk2204, line 121: Just FYI, you could add a patch to make this use Ruby instead of documenting this behaviour.  I'm sure this is my fault, and we already use Ruby elsewhere in the developer tooling.

bk2204 (approved):

docs/howto/release-git-lfs.md
chrisd8088: I thought about that, but I confess I took the easy route here, because there are a number of other places where our release tooling more or less assumes the user is on a Linux system, and I figured that if I ever tackled those more comprehensively, I'd do so in a separate PR.

Issue #4674, opened 2021-10-07
Several flaky tests


bk2204:
We've had a bunch of flaky test runs recently, which can be seen in the Actions tab.  We should look into identifying what makes them flaky and fix them, so we can have a nice, reliable testsuite.

bk2204:
The most notable test run that we now have that's flaky is `t/t-path.sh` on Windows.  Fixing that would really improve the development experience, since Windows takes almost 60 minutes to run the testsuite.

chrisd8088:
Our other flaky test(s) at the moment appear to the ones which involve credentials, such as `t/t-credentials.sh`, which occasionally fails on several systems including Windows and macOS.

PR #5595, opened 2023-12-21
Replace deprecated `io/ioutil` functions

Juneezee:
The `io/ioutil` package has been deprecated as of Go 1.16 [^1]. This commit replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.

[^1]: https://golang.org/doc/go1.16#ioutil

### DIFF
--- a/tools/util_windows.go
+++ b/tools/util_windows.go
@@ -42,7 +42,10 @@ func CheckCloneFileSupported(dir string) (supported bool, err error) {
 	if err != nil {
 		return false, err
 	}
-	defer os.Remove(src.Name())
+	defer func() {
+		_ = src.Close()
+		_ = os.Remove(src.Name())
+	}()
 
 	// Make src file not empty.
 	// Because `FSCTL_DUPLICATE_EXTENTS_TO_FILE` on empty file is always success even filesystem don't support it.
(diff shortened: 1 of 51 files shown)


chrisd8088:
Hey, thanks for the patch!  While this looks really good, it also seems like these changes may be consistently triggering one of the CI test flakes (per #4674) we're tracking in our Windows builds.  So I think we'll have to pin that down and fix it before we can merge this change.

tools/util_windows.go
Juneezee, line 1:
```
Looking for files that are not formatted correctly...
?? tools/dst3911297958
?? tools/src3438863009
```

https://github.com/git-lfs/git-lfs/actions/runs/7477718701/job/20351044777?pr=5595#step:14:361

Okay this is weird because [`ioutil.TempFile` simply calls `os.CreateTemp` starting from Go 1.17+](https://github.com/golang/go/blob/cc85462b3d23193e4861813ea85e254cfe372403/src/io/ioutil/tempfile.go#L24-L26). I didn't expect the CI to fail here

For some reasons they are not being removed by the `defer os.Remove` calls below :thinking:

Juneezee: Some related links:

1. https://stackoverflow.com/a/49163789/7902371
2. https://stackoverflow.com/a/64569543/7902371
3. https://github.com/golang/go/issues/58860

chrisd8088 (approved):
Thanks for all these updates and for digging into the unexpected test failures!

PR #5615, opened 2024-01-11
Remove test references to old CI

bk2204:
We no longer use Appveyor or TravisCI, so let's remove references to them in the testsuite.

### DIFF
--- a/t/t-askpass.sh
+++ b/t/t-askpass.sh
@@ -34,14 +34,6 @@ begin_test "askpass: push with core.askPass"
 (
   set -e
 
-  if [ ! -z "$TRAVIS" ] ; then
-    # This test is known to be broken on Travis, so we skip it if the $TRAVIS
-    # environment variable is set.
-    #
-    # See: https://github.com/git-lfs/git-lfs/pull/2500 for more.
-    exit 0
-  fi
-
   reponame="askpass-with-config"
   setup_remote_repo "$reponame"
   clone_repo "$reponame" "$reponame"
@@ -70,14 +62,6 @@ end_test
 begin_test "askpass: push with SSH_ASKPASS"
 (
   set -e
-
-  if [ ! -z "$TRAVIS" ] ; then
-    # This test is known to be broken on Travis, so we skip it if the $TRAVIS
-    # environment variable is set.
-    #
-    # See: https://github.com/git-lfs/git-lfs/pull/2500 for more.
-    exit 0
-  fi
 
   reponame="askpass-with-ssh-environ"
   setup_remote_repo "$reponame"
--- a/t/t-clone.sh
+++ b/t/t-clone.sh
@@ -89,10 +89,6 @@ end_test
 begin_test "cloneSSL"
 (
   set -e
-  if $TRAVIS; then
-    echo "Skipping SSL tests, Travis has weird behaviour in validating custom certs, test locally only"
-    exit 0
-  fi
 
   reponame="test-cloneSSL"
   setup_remote_repo "$reponame"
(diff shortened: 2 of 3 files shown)


chrisd8088:
I think PR #5882 has now addressed the Travis CI portions of this draft PR, and it turns out the AppVeyor bits are actually still important (although we don't use that service anymore), so I'd suggest we might be able to close this PR.

---

For reference, I've included below the patch I expect to use to deal with the seemingly AppVeyor-related environment variable.  I'll put this into a future PR, but the crux of it is that the `APPVEYOR_REPO_COMMIT_MESSAGE` environment variable is really just an example designed to check that the `Environ()` function in our `lfs` package [ignores](https://github.com/git-lfs/git-lfs/blob/f7bd4fd417052b86a960e1b9e569eb6748cfb548/lfs/lfs.go#L81-L83) all environment variables that don't start with `GIT_`, including those which happen to have `GIT_` in their values.

This was originally a problem exposed by commit 0951c698ad22a53c04d8d5d98a100f18f6e3993a in PR #1825, and then quickly addressed in commit 32244d92a44aaf24b144312bc1327ac7b4664d2e of the same PR, at which time the fake AppVeyor variable added to the test suite's environment to confirm that the problem was fixed.

Specifically, the earlier commit had the string `GIT_SSH` in its message, and the AppVeyor CI service automatically provided the message from the most recent commit in a Git branch under test in the `APPVEYOR_REPO_COMMIT_MESSAGE` environment variable.  At the time, the `Environ()` function would [populate](https://github.com/git-lfs/git-lfs/blob/0951c698ad22a53c04d8d5d98a100f18f6e3993a/lfs/lfs.go#L112-L114) the map it returns with the names and values of any environment variables whose names or values contained the string `GIT_`.  The result was that the `APPVEYOR_REPO_COMMIT_MESSAGE` variable was included in the returned environment, which would cause the tests in what is now our `t/t-env.sh` test script to fail.

Anyway, as I said, I have another round or two of test suite fixups underway, and I'll deal with this environment variable curiosity in those with a change like the one below.

```patch
diff --git a/t/t-env.sh b/t/t-env.sh
index 34fe0575..cfdadea9 100755
--- a/t/t-env.sh
+++ b/t/t-env.sh
@@ -10,6 +10,10 @@ if [ "$IS_WINDOWS" -eq 1 ]; then
   export MSYS2_ENV_CONV_EXCL="GIT_LFS_TEST_DIR"
 fi
 
+# The "git lfs env" command should ignore this environment variable
+# despite the "GIT_" strings in its name and value.
+export TEST_GIT_EXAMPLE="GIT_EXAMPLE"
+
 begin_test "env with no remote"
 (
   set -e
diff --git a/t/t-worktree.sh b/t/t-worktree.sh
index ef237238..775067f8 100755
--- a/t/t-worktree.sh
+++ b/t/t-worktree.sh
@@ -11,6 +11,10 @@ if [ "$IS_WINDOWS" -eq 1 ]; then
   export MSYS2_ENV_CONV_EXCL="GIT_LFS_TEST_DIR"
 fi
 
+# The "git lfs env" command should ignore this environment variable
+# despite the "GIT_" strings in its name and value.
+export TEST_GIT_EXAMPLE="GIT_EXAMPLE"
+
 begin_test "git worktree"
 (
     set -e
diff --git a/t/testenv.sh b/t/testenv.sh
index a63122ee..111d2024 100644
--- a/t/testenv.sh
+++ b/t/testenv.sh
@@ -152,7 +152,6 @@ GIT_CONFIG_NOSYSTEM=1
 GIT_TERMINAL_PROMPT=0
 GIT_SSH=lfs-ssh-echo
 GIT_TEMPLATE_DIR="$(native_path "$ROOTDIR/t/fixtures/templates")"
-APPVEYOR_REPO_COMMIT_MESSAGE="test: env test should look for GIT_SSH too"
 LC_ALL=C
 
 export CREDSDIR
@@ -160,7 +159,6 @@ export GIT_LFS_FORCE_PROGRESS
 export GIT_CONFIG_NOSYSTEM
 export GIT_SSH
 export GIT_TEMPLATE_DIR
-export APPVEYOR_REPO_COMMIT_MESSAGE
 export LC_ALL
 
 # Don't fail if run under git rebase -x.
```

bk2204:
Great, I'll close this, then.

PR #5616, opened 2024-01-11
Fix flaky test `t-credentials.sh`

### LINKED ISSUE #5609: t-credential.sh fails when building with lower level of parallelism
**Describe the bug**
When building the test suite with `make -C t PROVE_EXTRA_ARGS="-j40 -v"`, the tests pass. However, when turning the parallelism down to -j4, the t-credentials.sh fails consistently on test [13 - credentials with bad netrc creds will retry](https://github.com/git-lfs/git-lfs/blob/badde878a38fa2e37898ada0cbb007e7d6bb118b/t/t-credentials.sh#L457)

**To Reproduce**
1. make tests with `make -C t PROVE_EXTRA_ARGS="-j4 -v"`
2. see how they fail

**Expected behavior**
Passing test suit, same as when running with `-j40`

This issue occurred, when building git-lfs for CentOS Stream. We use `make -C t PROVE_EXTRA_ARGS="-j$(getconf _NPROCESSORS_ONLN) -v"` to get the number of physical cores present on the build system. When building in Koji, we can't see the issue, because by default we use -j40, however Zuul CI running on MR in our GitLab uses -j4, which then fails.
I was able to reproduce this locally in mockbuild environment. 

As I don't really know the test suite of git-lfs, I don't know what is the root cause of the problem.

bk2204:
We've noticed that this test has become flaky sometimes and fails in CI. More recently, a distributor noticed that this test fails consistently when run with 4 jobs at a time with `prove`.  Not all of the tests fail every time, but at least one of them does.

Part of the relevant difference is that our netrc tests use `localhost` as the machine name instead of the default 127.0.0.1.  This is helpful, because it means we don't call the credential helper, when we really want to fall back to the askpass code.  That's because we use different credentials for the `netrctest` remote, so if there were credentials available for `localhost`, we'll fail since we'll provide the wrong ones.

The reason, therefore, that this test is flaky is that the clone code creates an entry for the credential helper for `localhost`, and the credential directory is shared among tests.  Thus, `t-credentials.sh` fails whenever these tests run concurrently.

The obvious solution is to simply create a per-test credentials directory so that we aren't affected by what other tests are doing. Let's do that, which makes this test pass consistently.

The initial commit here simply sets up a function so that we can call it in our test.

Fixes #5609

### DIFF
--- a/t/t-credentials.sh
+++ b/t/t-credentials.sh
@@ -4,6 +4,9 @@
 
 ensure_git_version_isnt $VERSION_LOWER "2.3.0"
 
+export CREDSDIR="$REMOTEDIR/creds-credentials"
+setup_creds
+
 begin_test "credentials with url-specific helper skips askpass"
 (
   set -e
--- a/t/testhelpers.sh
+++ b/t/testhelpers.sh
@@ -557,6 +557,13 @@ write_creds_file() {
   fi
 }
 
+setup_creds() {
+  mkdir -p "$CREDSDIR"
+  write_creds_file "user:pass" "$CREDSDIR/127.0.0.1"
+  write_creds_file ":pass" "$CREDSDIR/--$certpath"
+  write_creds_file ":pass" "$CREDSDIR/--$keypath"
+}
+
 # setup initializes the clean, isolated environment for integration tests.
 setup() {
   cd "$ROOTDIR"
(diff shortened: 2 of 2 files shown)


opohorel:
I included both of these patches, added them to v3.4.1 release and tested with `-j4` inside the CentOS Stream 9 mock build.
Tests now pass as expected, thank you for taking a look and fixing this!

t/t-credentials.sh
```diff
@@ -4,6 +4,9 @@
 
 ensure_git_version_isnt $VERSION_LOWER "2.3.0"
 
+export CREDSDIR="$REMOTEDIR/creds-credentials"
+setup_creds
```
chrisd8088, line 8: I confess to being a bit puzzled here because the commit description indicates the goal is to create a per-test credentials directory, but wouldn't this create a directory shared among all the tests in `t/t-credentials.sh`, and if I understand the problem there are several tests in that test suite which end up conflicting with each other, so this might not fully resolve the issue?  Quite possibly I'm missing some important detail here.

chrisd8088 (approved):
Thanks for digging into this!  I had a question about how this change would work if the problem stems from multiple tests in the same suite conflicting with each other, but if it entirely fixes the problem, I'm good with it.

t/t-credentials.sh
bk2204: Sorry, by per-test I mean test file, not `begin_test`/`end_test`.  Thus, the goal is to prevent the `t/t-credential.sh` and `t/t-clone.sh` tests from conflicting with each other.  The sets of tests in each file are fine on their own, just not together.

PR #5618, opened 2024-01-17
Add a `--local` argument to install.sh

murez:
add option `--local` to install at `$HOME/.local` and check permission of the prefix before installing

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -9,14 +9,37 @@ elif [ "${BOXEN_HOME:-}" != "" ] ; then
   prefix=${BOXEN_HOME:-}
 fi
 
-mkdir -p $prefix/bin
-rm -rf $prefix/bin/git-lfs*
+local_install=false
+
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --local)
+      local_install=true
+      prefix="$HOME/.local"
+      shift
+      ;;
+    *)
+      echo "Unknown option: $1"
+      exit 1
+      ;;
+  esac
+done
+
+# Check if the user has permission to install in the specified prefix
+if [ ! -w "$prefix" ]; then
+  echo "Error: Insufficient permissions to install in $prefix. Try running with sudo or choose a different prefix."
+  exit 1
+fi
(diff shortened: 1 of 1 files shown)


script/install.sh
```diff
@@ -9,14 +9,37 @@ elif [ "${BOXEN_HOME:-}" != "" ] ; then
...
 fi
 
-mkdir -p $prefix/bin
-rm -rf $prefix/bin/git-lfs*
+local_install=false
+
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --local)
+      local_install=true
```
bk2204, line 17: It doesn't look like this variable is being used.  Perhaps we can remove it here and on line 12?

script/install.sh
```diff
@@ -9,14 +9,37 @@ elif [ "${BOXEN_HOME:-}" != "" ] ; then
...
+    *)
+      echo "Unknown option: $1"
+      exit 1
+      ;;
+  esac
+done
+
+# Check if the user has permission to install in the specified prefix
+if [ ! -w "$prefix" ]; then
+  echo "Error: Insufficient permissions to install in $prefix. Try running with sudo or choose a different prefix."
```
bk2204, line 30: Let's redirect this to standard error with `>&2` so that errors go the right place.

bk2204 (commented):
Hey, thanks for the patch and welcome to Git LFS!  This looks like a valuable feature.  I have just a few things that we could clean up here and then this should be ready to go.

murez:
Thanks for suggestions, I just found out that on multi tenant like HPC env, it is impossible to install `git-lfs` to `/usr/local`, and install to home local may be the best practice.

bk2204:
It's not actually impossible to do.  You need to specify the `PREFIX` environment variable, and if you do `PREFIX="$HOME/.local" ./install.sh`, then it should work.  That's a pretty standard way to specify installation targets for build systems, but it would be nice to have support for this out of the box.

bk2204 (approved):

PR #5629, opened 2024-01-23
pull: improve error message on failing checkout

bk2204:
If the user is having a problem checking out a file, it would be helpful to know why.  Let's include the reason that the error is occurring so they can learn what the cause is and report it helpfully if there's a problem.


bk2204:
That would be fantastic if you want to do that.

larsxschneider (approved):
I love that!

PR #5630, opened 2024-01-23
Use Azure Code Signing for Windows release binaries

bk2204:
It will soon be the case that all private keys used for signing Windows binaries must be stored in an HSM.  To make that possible, let's use Azure Code Signing to generate a temporary certificate which is backed by an HSM and use that to sign the data.

I've successfully tested this on our internal CI repo, where I've verified that the intermediate assets are all correctly signed (using `osslsigncode`).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,10 +40,6 @@ jobs:
       shell: bash
     - run: mkdir -p bin/releases
       shell: bash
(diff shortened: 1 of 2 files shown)


chrisd8088 (approved):
This looks great, thank you!  I see that you've already added the `SPN_GIT_LFS_SIGNING*` keys to the project's secrets, and presumably we can remove the `WINDOWS_CERT_*` ones once this PR is merged.

Issue #5622, opened 2024-01-18
git-lfs-transfer opens a lot of unused sessions


axkibe:
**Describe the issue**
This isn't a bug that breaks anything, but something that wastes resources.

I'm currently implementing git-lfs-transfer for the SSH protocol for my node.js git server.

I tried to make the log most sensible to understand.

My syntax is to assign events to streams is "connection number:session number".

The git client simply tries to make a push with 1 LFS object.

So far its all good, first git makes one ssh connection with git-receive-pack command, then comes a second ssh connection with the LFS stuff and a third one for git-upload-pack.*

Within the LFS connection client and server version-hello each other, the client makes a batch request, which I reply them, with a go ahead (otherwise as one server solution I don't care about announcing these things much anyway, but I do the batch things to make the client happy) .. but then the client then makes 7 more sessions within that connection, makes 7 times version hellos with the server and then never to use them and only to uses the primary one for the put-object request (which I didn't implement yet) 

```
Jan 18 16:22:28 ssh-lfs[2:1]: client wants to transfer: /axel.git upload
Jan 18 16:22:28 ssh-lfs[2:1]: user axel lfs transfers axel.git (upload)
2:1 OUT:000eversion=1\n0000
2:1  IN:000eversion 1\n
2:1  IN:0000
2:1 OUT:000fstatus 200\n0000
Jan 18 16:22:28 ssh[3]: client connected {
  ip: '::ffff:127.0.0.1',
  family: 'IPv6',
  port: 40928,
  header: {
    greeting: '',
    identRaw: 'SSH-2.0-OpenSSH_9.3p1 Ubuntu-1ubuntu3.2',
    versions: { protocol: '2.0', software: 'OpenSSH_9.3p1' },
    comments: 'Ubuntu-1ubuntu3.2'
  }
}
Jan 18 16:22:28 ssh-git[3:1]: client wants to git-upload-pack: '/axel.git'
Jan 18 16:22:28 ssh-git[3:1]: user axel accesses axel.git (git-upload-pack)
Jan 18 16:22:28 ssh-git[3:1]: spawning axel git-upload-pack
Jan 18 16:22:28 ssh-git[3:1]: path /vcs/git/axel.git
Jan 18 16:22:28 ssh-git[3:1]: git spawn close 0
Jan 18 16:22:28 ssh[3]: client disconnected
2:1  IN:000abatch\n
2:1  IN:0011transfer=ssh\n
[...]

PR #5634, opened 2024-01-29
Initialize sessions lazily

bk2204:
Right now, we spawn potentially several connections without necessarily needing to.  For example, if we're transferring two objects, we can practically use at most three connections: one for the batch and one for each object.

Instead of spawning all the workers up front, let's change how we do things and simply pass in the worker number at first.  Then, acquire the connection when we're uploading or downloading so that the connection is only spawned on demand.  This means that we spawn fewer workers in case there are few objects.

/cc #5622

### DIFF
--- a/ssh/connection.go
+++ b/ssh/connection.go
@@ -86,14 +86,28 @@ func (tr *SSHTransfer) IsMultiplexingEnabled() bool {
 }
 
 // Connection returns the nth connection (starting from 0) in this transfer
-// instance or nil if there is no such item.
-func (tr *SSHTransfer) Connection(n int) *PktlineConnection {
+// instance if it is initialized and otherwise initializes a new connection and
+// saves it in the nth position.  In all cases, nil is returned if n is greater
+// than the maximum number of connections.
+func (tr *SSHTransfer) Connection(n int) (*PktlineConnection, error) {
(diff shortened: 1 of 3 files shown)


ssh/connection.go
```diff
@@ -87,13 +87,25 @@ func (tr *SSHTransfer) IsMultiplexingEnabled() bool {
 
 // Connection returns the nth connection (starting from 0) in this transfer
 // instance or nil if there is no such item.
```
chrisd8088, line 89: It might be worth updating this comment to reflect the revised logic in the function, maybe something like:
```suggestion
// Connection returns the nth connection (starting from 0) in this transfer
// instance if it is initialized and otherwise initializes a new connection and
// saves it in the nth position.  In all cases, nil is returned if n is greater than
// the maximum number of connections.
```

chrisd8088 (approved):
Looks good, thank you!  I had one suggestion only for an update to a comment.

ssh/connection.go
bk2204: Squashed in with a co-author credit.

MosTLive: > Squashed in with a co-author credit.

MosTLive: ]()

Issue #5636, opened 2024-01-30
Git LFS prune deletes staged objects


jochenhz:
**Describe the bug**
Staged LFS files are deleted by `git lfs prune`. If I create a commit after that, the commit is broken and cannot be pushed. Even worse, if I have multiple local commits, I have permanently lost data. Note, `--verify-remote` does not prevent this

**To Reproduce**
git init
git lfs install
git lfs track myfile.data
git add .
git lfs prune

Verify that the LFS objects have been deleted from .git/lfs/objects

**Expected behavior**
Staged files should be retained in the same way as stashed or reachable files. 
`--verify-remote` should check every file that is deleted, I guess?

**System environment**
macOS 13.4

**Output of `git lfs env`**
```
git-lfs/3.4.0 (GitHub; darwin arm64; go 1.20.6)
git version 2.42.0

Endpoint=https://github.com/jochenhz/lfs-bug.git/info/lfs (auth=basic)
LocalWorkingDir=/Users/jochenhunz/Downloads/gitlfs-bug/lfs-bug
LocalGitDir=/Users/jochenhunz/Downloads/gitlfs-bug/lfs-bug/.git
LocalGitStorageDir=/Users/jochenhunz/Downloads/gitlfs-bug/lfs-bug/.git
LocalMediaDir=/Users/jochenhunz/Downloads/gitlfs-bug/lfs-bug/.git/lfs/objects
LocalReferenceDirs=
TempDir=/Users/jochenhunz/Downloads/gitlfs-bug/lfs-bug/.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
FetchRecentRefsIncludeRemotes=true
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/Users/jochenhunz/Downloads/gitlfs-bug/lfs-bug/.git/lfs
AccessDownload=basic
AccessUpload=basic
DownloadTransfers=basic,lfs-standalone-file,ssh
UploadTransfers=basic,lfs-standalone-file,ssh
GIT_EXEC_PATH=/opt/homebrew/opt/git/libexec/git-core
git config filter.lfs.process = "git-lfs filter-process"
git config filter.lfs.smudge = "git-lfs smudge -- %f"
[...]

PR #5637, opened 2024-01-31
Fix git lfs prune is deleting staged files in the index

jochenhz:
This PR should fix https://github.com/git-lfs/git-lfs/issues/5636

Running git lfs prune right after git add will prune the indexed files from the LFS cache so that any commit will result in a rightfully rejected push. Even worse, doing multiple commits on changes of the same file will permanently delete LFS data that is not recoverable.

This change retains all files that are in the index by diffing the index against the HEAD.

### DIFF
--- a/commands/command_prune.go
+++ b/commands/command_prune.go
@@ -481,7 +481,7 @@ func pruneTaskGetRetainedWorktree(gitscanner *lfs.GitScanner, fetchconf lfs.Fetc
 	// Don't repeat any commits, worktrees are always on their own branches but
 	// may point to the same commit
 	commits := tools.NewStringSet()
-	
+
 	if !fetchconf.PruneForce {
 		// current HEAD is done elsewhere
 		headref, err := git.CurrentRef()
@@ -544,7 +544,6 @@ func pruneTaskGetRetainedIndex(gitscanner *lfs.GitScanner, ref string, workingDi
 		return
 	}
 }
-
 
 // Background task, must call waitg.Done() once at end
 func pruneTaskGetReachableObjects(gitscanner *lfs.GitScanner, outObjectSet *tools.StringSet, errorChan chan error, waitg *sync.WaitGroup, sem *semaphore.Weighted) {
--- a/commands/command_prune.go
+++ b/commands/command_prune.go
@@ -471,13 +471,9 @@ func pruneTaskGetRetainedUnpushed(gitscanner *lfs.GitScanner, fetchconf lfs.Fetc
 func pruneTaskGetRetainedWorktree(gitscanner *lfs.GitScanner, fetchconf lfs.FetchPruneConfig, retainChan chan string, errorChan chan error, waitg *sync.WaitGroup, sem *semaphore.Weighted) {
 	defer waitg.Done()
 
-	if fetchconf.PruneForce {
-		return
-	}
-
 	// Retain other worktree HEADs too
 	// Working copy, branch & maybe commit is different but repo is shared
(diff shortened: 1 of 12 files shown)


commands/command_prune.go
```diff
@@ -521,6 +522,26 @@ func pruneTaskGetRetainedStashed(gitscanner *lfs.GitScanner, retainChan chan str
 	}
 }
 
+// Background task, must call waitg.Done() once at end
+func pruneTaskGetRetainedIndex(gitscanner *lfs.GitScanner, retainChan chan string, errorChan chan error, waitg *sync.WaitGroup, sem *semaphore.Weighted) {
+	defer waitg.Done()
+
+	err := gitscanner.ScanIndex("HEAD", func(p *lfs.WrappedPointer, err error) {
```
bk2204, line 529: We probably want to handle multiple indices, one for each worktree.  It might therefore be helpful to call this from `pruneTaskGetRetainedWorktree`.  If we don't do that, we might produce wrong results if I do `git add` in worktree A and then `git lfs prune` in another worktree.

bk2204 (commented):
I think there's a situation with multiple worktrees we'll probably want to handle here (and some tests for that case would be helpful as well).  In addition, it looks like we're failing some tests, too.

jochenhz:
Good point, I'll take a look at this

bk2204:
I do also want to just say thanks for sending a patch for this before we even got a chance to look at the issue.  I'm overall really happy with the direction this is going in.

jochenhz:
> I do also want to just say thanks for sending a patch for this before we even got a chance to look at the issue. I'm overall really happy with the direction this is going in.

Sure thing, I appreciate that you guys are very busy and would like to increase the likelihood of this being fixed. 

I am actually a little bit stuck how to implement this for worktrees. The problem is that to use git diff-index I'd need to `cd` into the worktree and don't really know how to do that as I cannot change the config of the GitScanner being used

bk2204:
> I am actually a little bit stuck how to implement this for worktrees. The problem is that to use git diff-index I'd need to `cd` into the worktree and don't really know how to do that as I cannot change the config of the GitScanner being used

The way I would do this is to pass an additional argument to `ScanIndex` that would eventually get passed down to `DiffIndex` that, if it's not the empty string, is the working directory.  That can be passed used as `git -C WORKING-DIRECTORY diff-index` to invoke Git in the proper location.  If it is the empty string, then it's ignored and we don't change the working directory.  Does that make sense to you?

jochenhz:
> > I am actually a little bit stuck how to implement this for worktrees. The problem is that to use git diff-index I'd need to `cd` into the worktree and don't really know how to do that as I cannot change the config of the GitScanner being used
> 
> The way I would do this is to pass an additional argument to `ScanIndex` that would eventually get passed down to `DiffIndex` that, if it's not the empty string, is the working directory. That can be passed used as `git -C WORKING-DIRECTORY diff-index` to invoke Git in the proper location. If it is the empty string, then it's ignored and we don't change the working directory. Does that make sense to you?

Makes perfect sense, thanks for your input. I'll update the PR once I'd like some feedback!

jochenhz:
@bk2204 I have force pushed a new solution taking worktrees into account

bk2204:
It looks like we still have one or two failing tests here.  Once that's fixed, I'll do a final review.

jochenhz:
@bk2204 yeah, it was failing with git 2.0.0 as no worktrees were supported back then and the git.go RootDir() function was not handling missing working directories correctly. Should be all green now 🤞

bk2204 (approved):

Issue #5642, opened 2024-02-02
Proposal: Add a --safe option to LFS prune


jochenhz:
**Describe the issue**
My motivation:
While working on https://github.com/git-lfs/git-lfs/pull/5637 I figured that prune is not secure enough for our use cases. Our git client is used by game developers, so the LFS cache is (force) pruned very regularly to significantly reduce the size of the project folder. This is necessary for git (LFS) to be a valid competitor to e.g. perforce.
The one thing that version control must prevent is data loss, and git LFS prune may be very destructive as it does not take the reflog or indexed files into account, for example. Beside https://github.com/git-lfs/git-lfs/pull/5637, there is also this open issue https://github.com/git-lfs/git-lfs/issues/4593.

My proposal:
In this commit https://github.com/Anchorpoint-Software/git-lfs/commit/749af8b50063ad0642bcf8ccdfda0fbb52c461b4 I've added a `--safe` option to LFS prune that only prunes objects that are 100% on the remote server. So in contrast to `--verify-remote`, `--safe` does not exit if a file is not available on the remote and orphan commits and indexed files are not pruned.

I am curious to hear what you think about this?

bk2204:
Hey,

I think such a feature is useful and welcome.  I would say that I don't love the name `--safe` and might call this feature `--verify-continue` or some such, but I'm interested to hear what @chrisd8088 thinks, since he's often better at naming things than I am.

chrisd8088:
I'd agree such a feature addition would be valuable.  We should add a matching configuration option to the `lfs.pruneVerifyRemoteAlways` setting as well, to be consistent with how `--verify-remote` is implemented.

My own instinct, FWIW, might be to add an option (and config setting) which requires `--verify-remote` or `lfs.pruneVerifyRemoteAlways` to be set, and then controls whether or not to [exit](https://github.com/git-lfs/git-lfs/blob/195fff823fecdcbd950942b8cc838bcd0c48015e/commands/command_prune.go#L245-L247) when a problem is detected.

I'm not great with naming things either, and I've struggled to come up with anything that's both succinct and self-explanatory in this case, where I think we're trying to express, "When there are locally prunable objects, and some can be verified to exist on the remote, but others can not, either prune nothing, or prune only the objects that were verified to exist on the remote."  Which is quite a mouthful.

[...]

PR #5648, opened 2024-02-08
Add --verify-unreachable option to LFS prune

jochenhz:
This PR should close https://github.com/git-lfs/git-lfs/issues/5642

LFS prune offers `--verify-remote`, a command that exits whenever a reachable object is not on the remote. 
Hence, it still prunes non-reachable objects, e.g. referenced by orphan commits. 

This PR adds a new option `--verify-unreachable` that additionally verifies those unreachable objects, so there is a guarantee that data is never lost. 

In addition to that, a new option `--when-unverified=<halt,continue>` can change the default behaviour to not exit when a file is unverified but continue to prune the verified objects.


jochenhz:
@bk2204 the t-credentials.sh fails on macOS and I don't know why that might happen. If I run cibuild on my mac, everything is green. Any idea?

bk2204:
That test can be a little flaky; I've marked it for rebuild.

bk2204 (approved):

PR #5650, opened 2024-02-08
workflows: update to Go 1.22

bk2204:
We want to support and test the latest versions of Go.  Now that Go 1.22 is out, let's test against 1.22 and 1.21 instead of 1.21 and 1.20.


bk2204:
Once this is approved, we can adjust the required builds here and switch to Go 1.22 by default.

chrisd8088 (approved):

PR #5655, opened 2024-02-22
Fix logScanner fails to parse pointer file containing unicode chars

jochenhz:
Git will quote paths using Unicode characters in git log, but the fileHeaderRegex does not account for optional quotes in the log.

This change fixes the regex to account for quotes due to Unicode characters and improves the tests. 
Due to this bug, git lfs prune might accidentally prune recent files that contain unicode characters.

One might see a git log looking like this, for example (note the quotation marks):

```
lfs-commit-sha: cde89bbb531b285170c0119c34e656141dca269a fba0486e7940971a64f87d8ef8e65fb1be44b2ad

diff --git a/Source Assets/Audio/01 - Am Flughafen - Ankunftsbereich.mp3 b/Source Assets/Audio/01 - Am Flughafen - Ankunftsbereich.mp3
new file mode 100644
index 0000000..8bf5d61
--- /dev/null
+++ b/Source Assets/Audio/01 - Am Flughafen - Ankunftsbereich.mp3       
@@ -0,0 +1,3 @@
+oid sha256:0b21b27eb6f510ef4da2ef3b35c3e394e42e735913fdd2c9bd3dbfde3be19b7b
+size 1535686
diff --git "a/Source Assets/Audio/Alltagsger\303\244usche - Kamera - Blitzlicht.mp3" "b/Source Assets/Audio/Alltagsger\303\244usche - Kamera - Blitzlicht.mp3"
new file mode 100755
index 0000000..63a83b8
--- /dev/null
+++ "b/Source Assets/Audio/Alltagsger\303\244usche - Kamera - Blitzlicht.mp3"   
@@ -0,0 +1,3 @@
+oid sha256:9b265c0416fe1332300bdb08e3d59326b92e948f9a3cab4d4401b11e22100d5b
+size 251836

```

### DIFF
--- a/lfs/gitscanner_log.go
+++ b/lfs/gitscanner_log.go
@@ -216,7 +216,7 @@ func newLogScanner(dir LogDiffDirection, r io.Reader) *logScanner {
 		// no need to compile these regexes on every `git-lfs` call, just ones that
 		// use the scanner.
 		commitHeaderRegex:    regexp.MustCompile(fmt.Sprintf(`^lfs-commit-sha: (%s)(?: (%s))*`, git.ObjectIDRegex, git.ObjectIDRegex)),
-		fileHeaderRegex:      regexp.MustCompile(`^diff --git a\/(.+?)\s+b\/(.+)`),
+		fileHeaderRegex:      regexp.MustCompile(`^diff --git "?a\/(.+?)\s+"?b\/(.+)`),
 		fileMergeHeaderRegex: regexp.MustCompile(`^diff --cc (.+)`),
 		pointerDataRegex:     regexp.MustCompile(`^([\+\- ])(version https://git-lfs|oid sha256|size|ext-).*$`),
 	}
@@ -341,6 +341,11 @@ func (s *logScanner) scan() (*WrappedPointer, bool) {
 }
 
 func (s *logScanner) setFilename(name string) {
+	// Trim last character if it's a quote
+	if len(name) > 0 && name[len(name)-1] == '"' {
+		name = name[:len(name)-1]
+	}
+
 	s.currentFilename = name
 	s.currentFileIncluded = s.Filter.Allows(name)
 }
--- a/lfs/scanner_test.go
+++ b/lfs/scanner_test.go
@@ -82,11 +82,11 @@ index 0000000..8519883
 +size 125873
 lfs-commit-sha: 64b3372e108daaa593412d5e1d9df8169a9547ea e99c9cac7ff3f3cf1b2e670a64a5a381c44ffceb
 
-diff --git a/hobbit_5armies_2.mov b/hobbit_5armies_2.mov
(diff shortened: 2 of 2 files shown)


bk2204 (commented):
This seems fine, but don't we need to handle the octal escapes as well?  The example in the PR message shows a couple of octal escapes that we'd probably want to handle in `setFilename`.

jochenhz:
> This seems fine, but don't we need to handle the octal escapes as well? The example in the PR message shows a couple of octal escapes that we'd probably want to handle in `setFilename`.

They are already matched by the current regex, [see](https://regex101.com/r/nwEuAS/1)

bk2204:
> They are already matched by the current regex, [see](https://regex101.com/r/nwEuAS/1)

Right, but `setFilename` checks to see if the file is matched by the filter, and the escaped filename wouldn't necessarily match the filter.  If the filter is `*ç*.bin` and the filename is `républiquefrançaise.bin`, then the encoded name would be `r\303\251publiquefran\303\247aise.bin`, and that wouldn't match.

jochenhz:
> > They are already matched by the current regex, [see](https://regex101.com/r/nwEuAS/1)
> 
> Right, but `setFilename` checks to see if the file is matched by the filter, and the escaped filename wouldn't necessarily match the filter. If the filter is `*ç*.bin` and the filename is `républiquefrançaise.bin`, then the encoded name would be `r\303\251publiquefran\303\247aise.bin`, and that wouldn't match.

Ah, good catch!
I have pushed a change that converts the incoming name to utf-8 and added some tests for it. Not sure if this is the most elegant solution in golang though.

lfs/gitscanner_log.go
```diff
@@ -341,6 +342,17 @@ func (s *logScanner) scan() (*WrappedPointer, bool) {
...
+	// Trim last character if it's a quote
+	if len(name) > 0 && name[len(name)-1] == '"' {
+		name = name[:len(name)-1]
+	}
+
+	// Convert octals to proper UTF-8 code
+	unquotedName, err := strconv.Unquote(`"` + name + `"`)
+	if err == nil {
+		name = unquotedName
+	}
```
bk2204, line 354: I think we still have one problem here, which is that this should only apply to strings which were quoted in the first place.  If they're not quoted, then we don't want to unquote them, since that could result in incorrect results.

So this block should go within the previous `if` statement.

jochenhz: Not sure about this. The if statement before is just to catch the ending `"` of the second matching group because I could not find a regex that excludes the last `"` as lookaheads are not supported in go.

But anyway, I think it's a no-op for non quoted strings, no? [See this example](https://go.dev/play/p/CZz4nYz1A3v)

bk2204: Yeah, I think you're right.  We wouldn't end up with a backslash escape unless we were quoted, and if we had a backslash at all, we'd have an escape and quoting.

bk2204 (approved):

PR #5671, opened 2024-03-06
Supports downloading each file using multi-threading for basicDownloa…

SURA907:
git-lfs downloads multiple lfs files at the same time, but only uses one thread to download each file

Multi-threaded downloading is very useful, I hope to add this feature to basicDownloadAdapter

Control the number of threads through the newly added configuration `lfs.downloadthreads`, default 8, maximum 64

There are some additional tests. If it does not pass, it will return to the original single-threaded download:
1. File smaller than 100MB
2. An error occurred in the HEAD request, or the server did not respond to the ‘Accept-Ranges’ header

### DIFF
--- a/tools/iotools.go
+++ b/tools/iotools.go
@@ -19,6 +19,20 @@ const (
 	memoryBufferLimit = 1024
 )
 
+func CopyNWithCallBack(w io.Writer, r io.Reader, totalSize, startByte, endByte int64, cb CopyCallback) (int64, error) {
+	if cb == nil {
+		return io.CopyN(w, r, endByte-startByte+1)
+	}
+
+	cbReader := &CallbackReader{
+		C:         cb,
+		TotalSize: totalSize,
+		Reader:    r,
+	}
+
+	return io.CopyN(w, cbReader, endByte-startByte+1)
+}
+
 // CopyWithCallback copies reader to writer while performing a progress callback
 func CopyWithCallback(writer io.Writer, reader io.Reader, totalSize int64, cb CopyCallback) (int64, error) {
 	if success, _ := CloneFile(writer, reader); success {
--- a/tq/basic_download.go
+++ b/tq/basic_download.go
@@ -1,19 +1,28 @@
 package tq
 
 import (
+	"encoding/hex"
 	"fmt"
-	"hash"
 	"io"
 	"net/http"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strconv"
+	"sync"
 
+	"github.com/git-lfs/git-lfs/v3/config"
 	"github.com/git-lfs/git-lfs/v3/errors"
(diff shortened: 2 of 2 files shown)


SURA907:
My object storage provider limits the bandwidth of a single download request even though I am paying based on traffic. 囧

bk2204:
Hey,

My concern with this is that it's going to compete with the number of download threads, so we may get combinatorial explosion (so 8 threads for 8 concurrent downloads is 64 threads), which can easily exceed the CPU and bandwidth of the machine.  Also, in many cases, servers, despite the RFC, don't properly handle `HEAD` requests, so this will simply fail to work in many cases.  Using an additional `HEAD` request will also make at least twice as many requests, which doubles the latency and will make small- to medium-sized files slower.

However, the main reason I think we can't accept this is that it changes the [basic transfer API](https://github.com/git-lfs/git-lfs/blob/main/docs/api/basic-transfers.md) in an incompatible way, since the API doesn't mention the usage of `HEAD` requests.  We could in theory include a new transfer type that supported this, but I think for the reasons above, it may not see much use.

It does sound unfortunate that your object storage provider limits your download bandwidth needlessly, but I'd argue that should be handled by switching storage provider instead of having all clients work around that fact.

For future reference, we generally prefer changes to the API or proposals for new API to be opened as a proposal with a doc first, so that we can discuss them first.  That avoids situations where folks put in a lot of extra work to write code that we don't end up accepting.

SURA907:
> Hey,
> 
> My concern with this is that it's going to compete with the number of download threads, so we may get combinatorial explosion (so 8 threads for 8 concurrent downloads is 64 threads), which can easily exceed the CPU and bandwidth of the machine. Also, in many cases, servers, despite the RFC, don't properly handle `HEAD` requests, so this will simply fail to work in many cases. Using an additional `HEAD` request will also make at least twice as many requests, which doubles the latency and will make small- to medium-sized files slower.
> 
> However, the main reason I think we can't accept this is that it changes the [basic transfer API](https://github.com/git-lfs/git-lfs/blob/main/docs/api/basic-transfers.md) in an incompatible way, since the API doesn't mention the usage of `HEAD` requests. We could in theory include a new transfer type that supported this, but I think for the reasons above, it may not see much use.
> 
> It does sound unfortunate that your object storage provider limits your download bandwidth needlessly, but I'd argue that should be handled by switching storage provider instead of having all clients work around that fact.
> 
> For future reference, we generally prefer changes to the API or proposals for new API to be opened as a proposal with a doc first, so that we can discuss them first. That avoids situations where folks put in a lot of extra work to write code that we don't end up accepting.

OK, I get it, changing the default behavior is hardly a good idea.

Perhaps the server should inform the batch whether multi-threaded downloading can be used

For small and medium files, it may be feasible to automatically fall back to single-threaded downloads.

I will submit a proposal later

PR #5674, opened 2024-03-07
Include remote error when pure SSH protocol fails

bk2204:
Right now, if spawning the remote command fails for some reason, we get a message like the following in the trace output:

  pure SSH protocol connection failed: Unable to negotiate version with remote side (unable to read capabilities): EOF

However, all we know is that we failed to connect to the remote `git-lfs-transfer` process.  That could be for any number of reasons: the remote program doesn't exist or is not in `PATH`, the user has misconfigured something locally or remotely, or any number of other problems.  We won't know unless we log the output, so let's do that so we can get more information.

### DIFF
--- a/ssh/connection.go
+++ b/ssh/connection.go
@@ -1,10 +1,14 @@
 package ssh
 
 import (
+	"bytes"
+	"fmt"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


chrisd8088 (approved):
Looks good, thanks!

PR #5680, opened 2024-03-13
Match `PATH` case insensitively in Windows installer

### LINKED ISSUE #5679: Install on Windows fails due to "unexpected" git executable
Starting with 3.20, the chocolatey package install is failing and I'm seeing the same behavior when I attempt a silent install from the command line using the exe file. 

This just fails without any indication of what happened unless I add the switch to create a log file
`.\git-lfs-windows-v3.2.0.exe /VERYSILENT /SUPPRESSMSGBOXES /NORESTART /LOG`

The log shows that a Warning message was suppressed and defaulted to No which cancelled the install process.

```
Warning: Found Git in unexpected location: "c:\program files\git\cmd\git.EXE"
Defaulting to No for suppressed message box (Yes/No):

An executable Git program was found in an unexpected location outside of Program Files:
     "c:\program files\git\cmd\git.EXE"
If this looks dubious, Git LFS should not be registered using it.

Do you want to register Git LFS using this Git program?

Refusing to use Git found at: "c:\program files\git\cmd\git.EXE"
```
Depending on the exact configuration of the PC, the error message might say c:\program files\git\bin\git.exe. I can't understand why git-lfs thinks that location is "outside of Program Files." If I run the installer manually without the /SILENT switch and choose Yes then it installs as expected but this breaks any scripted install.

bk2204:
The Windows installer expands the path to the `C:\Program Files` directory using the standard casing for this.  However, our `PATH` value does not need to contain the same case, since on Windows this directory is case insensitive, and as a result, if the two differ in case, we don't correctly detect that the Git executable is in `PATH`.

Let's fix this by lowercasing both versions of the text.  Note that instead of using the `Lowercase` function, which deals only with ASCII characters, we use `AnsiLowercase`, which, despite its name, uses the current Windows locale.

This is not guaranteed to produce correct results in all cases because the file system's case folding is done in the kernel in a locale-insensitive way based on the tables created when the file system was created, but there's no way to do better than this in the general case since we don't have functionality to implement the kernel's case mapping and it is impossible to correctly fold arbitrary Unicode text in a locale-insensitive way.  We therefore adopt the least bad option available while continuing to support Windows.

Fixes #5679

### DIFF
--- a/script/windows-installer/inno-setup-git-lfs-installer.iss
+++ b/script/windows-installer/inno-setup-git-lfs-installer.iss
@@ -167,9 +167,9 @@ begin
   else
     RegisterOrDeregister := 'register';
 
-  PFiles32 := ExpandConstant('{commonpf32}\')
+  PFiles32 := AnsiLowercase(ExpandConstant('{commonpf32}\'))
   if IsWin64 then
-    PFiles64 := ExpandConstant('{commonpf64}\')
+    PFiles64 := AnsiLowercase(ExpandConstant('{commonpf64}\'))
   else
     PFiles64 := PFiles32; // `commonpf64` is not available on 32-bit Windows
 
@@ -186,7 +186,7 @@ begin
       PathExt := Copy(PathExt, j+1, Length(PathExt)-j);
 
       if FileExists(Path + Ext) then begin
-        if (Pos(PFiles32, Path) = 1) or (Pos(PFiles64, Path) = 1) then begin
+        if (Pos(PFiles32, AnsiLowercase(Path)) = 1) or (Pos(PFiles64, AnsiLowercase(Path)) = 1) then begin
           Result := True;
           Exit;
         end;


chrisd8088:
My apologies for being a bit slow to take a look at this; life has intervened.  I hope to give it a whirl on a Windows VM shortly.

chrisd8088 (approved):
This looks good; thanks for your patience.  From my own testing, I agree this fixes the reported problem.

I also found that attempting to make `C:\` case-sensitive with `fsutil.exe file setCaseSensitiveInfo 'C:\' enable` or similar commands, even in the `Administrator` role, was prevented by Windows.  So, I believe that rules out the possibility there could be both a `C:\Program Files` and a `C:\program files` (which, if it were possible, would argue for case-sensitive matching against `C:\Program Files` since that's the OS-defined location we get from the `{commonpf*}` variables).

I also note that the comparison we perform here is strictly against that value (the one from the appropriate `{commonpf*}` variable), so although we may find a `git` executable somewhere else, including in a case-sensitive folder location, the case-sensitivity of that location doesn't matter for the sake of our check since the comparison value we use is always the (now lower-cased) `C:\program files` string.

PR #5698, opened 2024-04-02
checkout: gracefully handle files deleted from the index

### LINKED ISSUE #5669: `checkout` does not respect files which have been staged for removal with `git rm`
**Describe the bug**
Using `git lfs checkout` to apply LFS files in a working copy where some LFS files have been removed with `git rm` incorrectly attempts to check out the removed files, failing because the pointer file is no longer present on disk.

**To Reproduce**
1. Clone a repository that uses LFS entries with `--skip-smudge` so that LFS files are not checked out.
2. Run `git lfs fetch` to acquire all LFS files into the local cache.
3. Stage one of the LFS pointer files for deletion using `git rm`.
4. Run `git lfs checkout`. Observe that the checkout reports an error about failing to check out the removed file.

**Expected behavior**
Git LFS should check out all LFS files corresponding to the pointer files present in the working copy, and should ignore files which have been staged for deletion with `git rm`.

**System environment**
Windows 11

**Output of `git lfs env`**
```
git-lfs/3.4.0 (GitHub; windows amd64; go 1.20.6; git d06d6e9e)
git version 2.43.0.windows.1

Endpoint=https://github.cds.internal.unity3d.com//unity/neutron.git/info/lfs (auth=none)
  SSH=git@github.cds.internal.unity3d.com:/unity/neutron.git
LocalWorkingDir=C:\UnitySrc\neutron
LocalGitDir=C:\UnitySrc\neutron\.git
LocalGitStorageDir=C:\UnitySrc\neutron\.git
LocalMediaDir=C:\UnitySrc\lfs\objects
LocalReferenceDirs=
TempDir=C:\UnitySrc\lfs\tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=C:\UnitySrc\lfs
AccessDownload=none
AccessUpload=none
```

**Additional context**
My team is working on collectively resolving conflicts in a very large merge, which is why we have files that are staged for deletion but that cannot yet be committed. We'd like to be able to use `git lfs checkout` to check out the files that we have so far resolved, in order to use them in building/verifying parts of the codebase that we've resolved so far.

bk2204:
Right now, when someone deletes a pointer from the index with `git rm` and then runs `git lfs checkout`, the operation fails with a message of "Could not update the index" because our invocation of `git update-index` is missing the `--add` flag.

Obviously, the user does not expect an error in this case, and `git checkout` simply ignores files staged for deletation, so let's do the same thing.  If a file on disk is deleted, check the index with `git diff-index` to see if it's deleted from `HEAD`.  If so, ignore the file, just like Git does. Note that we use `git diff-index` specifically because it doesn't refresh the index and is therefore much cheaper than alternatives, such as `git status`, which might do that.

Fixes #5669

### DIFF
--- a/t/t-checkout.sh
+++ b/t/t-checkout.sh
@@ -51,6 +51,14 @@ begin_test "checkout"
   grep 'accepting "file1.dat"' checkout.log
   grep 'rejecting "file1.dat"' checkout.log && exit 1
 
+  git rm file1.dat
+
+  echo "checkout should skip replacing files deleted in index"
+  git lfs checkout
+  [ ! -f file1.dat ]
+
+  git reset --hard
+
   # Remove the working directory
   rm -rf file1.dat file2.dat file3.dat folder1/nested.dat folder2/nested.dat
 
--- a/commands/pull.go
+++ b/commands/pull.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"io"
 	"os"
+	"strings"
 	"sync"
 
 	"github.com/git-lfs/git-lfs/v3/config"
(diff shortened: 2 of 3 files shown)


t/t-checkout.sh
```diff
@@ -51,6 +51,14 @@ begin_test "checkout"
   grep 'accepting "file1.dat"' checkout.log
   grep 'rejecting "file1.dat"' checkout.log && exit 1
 
+  git rm file1.dat
+
+  echo "checkout should skip replacing files deleted in index"
+  GIT_TRACE=1 git lfs checkout 2>&1 | tee checkout.log
```
chrisd8088, line 57: I think this might be simplified to just `git lfs checkout`, since the output of the command is not subsequently examined (and so neither is any tracing out).  That would be similar to what's done immediately below in the `checkout with filters` check, where `git lfs checkout file2.dat` is run, also without examining its output.

chrisd8088 (approved):

t/t-checkout.sh
bk2204: Good point.  I'll squash that change in.

Issue #4110, opened 2020-04-27
Improve support for case-insensitive file systems / OSes


mojca:
**Describe the bug**

Suppose that a git repository contains `file1.png` and `file2.PNG`, with the latter only existing in the earlier commits (it is no longer present in the latest commit in the `master` branch).

Say that the developer uses Linux to run a straightforward command to convert `png` files to large files as documented in the [Tutorial](https://github.com/git-lfs/git-lfs/wiki/Tutorial#migrating-existing-repository-data-to-lfs) and elsewhere:
```
git lfs migrate import --include="*.png"
```
The new repository looks reasonable, with no evident problems, however only `file1.png` gets converted into a large file, while `file2.PNG` remains in the repository as a regular file.

As soon as someone tries to check out an older commit / branch / tag, run a bisection etc. on Windows, an unpleasant surprise is waiting:
```
Encountered 1 file(s) that should have been pointers, but weren't:
    file2.PNG
```
and the only way out is to force checkout of another commit.

According to my understand of #2858 this is considered to be a broken repository on Windows.

Some suggestions to fix the problem include creating a commit to fix the behaviour, but the latest commit is already fine. The problem happens when someone wants to do something with older branches / tags, and fixing the latest commit doesn't help at all.

The "best" alternative is to fix the full history, but that comes with tons of headaches related to everyone needing to do a clean checkout, loosing links in tickets, pull requests, some commits like `revert commit abcde123456` no longer pointing to a valid sha reference etc.

A proper solution would have been to use
```
git lfs migrate import --include="*.[pP][nN][gG]"
```
from the start, but that is hardly documented everywhere, and potential problems are nearly impossible to spot until it's too late.

I didn't really test it, but I assume the opposite problem would happen when a user on Windows adds `file3.PNG` to the repository which gets stored as large file, and another user on Linux checks this out.

It would be great if:
* there was a way to specify a case-insensitive pattern in a slightly less clumsy way than `*.[pP][nN][gG]`
* the conversion `git lfs migrate ...` would at least warn the developer in case there are files in repository that match the pattern in case-insensitive way, but not in case-sensitive way, so that future users on Windows or macOS don't end up suffering from the issue
* a similar warning is thrown when users try to add such files to repository later

**To Reproduce**

See above.

**Expected behavior**

Either of the following options:
* `git` should never be complaining about `Encountered 1 file(s) that should have been pointers, but weren't` on case-insensitive systems
[...]

PR #5699, opened 2024-04-03
Optimize performance for scanning trees in partial clones

bk2204:
Right now, if the user is using partial clone, our call to `git ls-tree` against HEAD is expensive because `git ls-tree` needs to download each blob, which it does incrementally instead of all at once.

If we're scanning the tree from HEAD, then we can avoid the expense of doing this by running `git ls-files` with a pattern that matches only LFS files, which makes the operation much cheaper, since we avoid needing to download blobs for many of those objects.  We can format the data such that it matches the pattern we expect for `git ls-tree` so that we can avoid modifying most of the calls and continue to let things function in the same way.  Do so, but limit our changes to Git 2.13.0 and newer, since the `attr` pathspec is new in that version.

In addition, let's fix some tests which rely on an invalid assumption about how we discover and process LFS files.


chrisd8088 (approved):
Very nice, thank you!

stanhu:
@bk2204 Note that this change appears to have broken some CI jobs that relied on the previous behavior of scanning all files in the tree instead of only the ones that match `.gitattributes`. More specifically, previously LFS pointers that were not listed in `.gitattributes` would still have a chance of matching. I think this might be worth documenting because I think this is a matter of updating `.gitattributes`.

Suppose you had a repository with these files:

```
test1.bin
TEST2.BIN
```

Let's say `.gitattributes` were set to:

```
*.bin filter=lfs diff=lfs merge=lfs -text
```

Let's also assume that the repository was cloned on a case-sensitive filesystem, like ext4. As a result, the Git config `core.ignorecase` is `false`.

In `git-lfs` v3.3.0, `git-lfs` runs `git ls-tree -r <SHA>` to find all files in the repository. In the example above, this includes both `test1.bin` and `TEST2.BIN`. It appears that `filepathfilter` [defaults to true](https://github.com/git-lfs/git-lfs/blob/04abbd843622600279baeb31c92862c4d73e419e/filepathfilter/filepathfilter.go#L118-L119) if the file doesn't match any attribute, and `runScanLFSFiles` appears to [scan any blob that might be an LFS pointer](https://github.com/git-lfs/git-lfs/blob/beae1146bda7e9fa712c3c73dacbcf9ed5067fd9/lfs/gitscanner_tree.go#L40-L41).

With this change, in git-lfs v3.6.0, git-lfs now runs something along the lines of `git ls-files -r :(top,attr:filter=lfs) <SHA>`, which filters out files that only are matched by `.gitattributes`. Since `TEST2.BIN` does not match the `*.bin` rule, it is ignored.

The solution, of course, is to update the `.gitattributes` [as described in this issue](https://github.com/git-lfs/git-lfs/issues/4110#issue-607559373):

```
*.[bB][iI][nN] filter=lfs diff=lfs merge=lfs -text
```

Of course if `core.ignorecase` is `true`, then the previous `.gitattributes` matches, and there are no issues.

I'd agree that the current method of being explicit in `.gitattributes` is the correct behavior. However, previously `git-lfs` was more forgiving and thus users may have come to depend on the behavior of searching for all possible LFS pointers, even if it was never intended.

PR #5709, opened 2024-04-15
chore: fix function names in comment

rustfix:


bk2204 (approved):

I think our CI is a little flaky right now, and while I can re-run these jobs to see if they'll pass, there's a larger issue that we need to address.  Once that's fixed, we'll go ahead and merge this, since the patch looks very reasonable.

PR #5779, opened 2024-05-28
Support arbitrary HTTP credential schemes for authentication

bk2204:
Git recently learned to announce capabilities for credential helpers and this functionality will be in Git 2.46.  One of the capabilities is the ability for credential helpers to specify arbitrary HTTP credential schemes and the credentials for these schemes.  This allows users with a suitable credential helper to implement support for, say, Bearer authentication where the credentials come from a credential helper.

With this capability, Git LFS will advertise the `capability[]=authtype` line, and a suitable version of Git will pass this along to the helper, which can then return something like `authtype=Bearer` and `credential=my-bearer-token` to provide a `Authorization: Bearer my-bearer-token` header.  It can also return other schemes as well, and the `credential` field can include any necessary parameters that the scheme may require.

In order to implement this in Git LFS, we first need to adjust our test credential helper and the server so that they can handle this new authentication type and return appropriate credentials.  Then, we add support in the main codebase and verify that it works as expected.

Note that Git versions which do not support this capability will exit unsuccessfully for `git credential capability` (causing our tests to be skipped) and will do nothing with the `capability[]=authtype` declaration, which means that it will not be passed to the credential helper.  At that point, the credential helper can either return the previous username/password pair, or decide to return nothing at all.

There is an additional capability to allow multi-stage authentication, such as NTLM or Kerberos, in the credential helper.  These both require two round trips for authentication to succeed, which needs additional support from the helper and Git or Git LFS.  Supporting NTLM in Git LFS via a suitable credential helper was in fact one of the goals of the upstream series, in addition to discouraging the use of the config file for storing authentication credentials (such as used to be required for Bearer auth).  However, the additional capability required for that is not implemented here, and can be implemented in an additional series in due course.

### DIFF
--- a/t/cmd/git-credential-lfstest.go
+++ b/t/cmd/git-credential-lfstest.go
@@ -70,19 +70,33 @@ func fill() {
 	}
 
 	hostPieces := strings.SplitN(firstEntryForKey(creds, "host"), ":", 2)
-	user, pass, err := credsForHostAndPath(hostPieces[0], firstEntryForKey(creds, "path"))
+	authtype, user, cred, err := credsForHostAndPath(hostPieces[0], firstEntryForKey(creds, "path"))
 	if err != nil {
 		fmt.Fprintln(os.Stderr, err.Error())
 		os.Exit(1)
 	}
 
-	if user != "skip" {
+	capas := discoverCapabilities(creds)
+
+	switch authtype {
+	case "skip":
+	case "":
 		if _, ok := creds["username"]; !ok {
 			creds["username"] = []string{user}
 		}
 
 		if _, ok := creds["password"]; !ok {
-			creds["password"] = []string{pass}
+			creds["password"] = []string{cred}
+		}
+	default:
+		if _, ok := capas["authtype"]; ok {
+			if _, ok := creds["authtype"]; !ok {
+				creds["authtype"] = []string{authtype}
+			}
+
+			if _, ok := creds["credential"]; !ok {
+				creds["credential"] = []string{cred}
+			}
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


t/t-credentials.sh
```diff
@@ -290,6 +290,40 @@ begin_test "credentials skips wwwauth[] with option"
...
+  printf "Bearer::token" > "$CREDSDIR/127.0.0.1--$reponame"
+
+  clone_repo "$reponame" "$reponame"
+  git checkout -b new-branch
+
+  git lfs track "*.dat" 2>&1 | tee track.log
+  grep "Tracking \"\*.dat\"" track.log
+
+  contents="b"
+  contents_oid=$(calc_oid "$contents")
```
chrisd8088, line 310: I'm not sure this `contents_oid` variable is used, so perhaps it could be removed?

t/t-credentials.sh
```diff
@@ -290,6 +290,40 @@ begin_test "credentials skips wwwauth[] with option"
...
+
+  printf "%s" "$contents" > b.dat
+  git add b.dat
+  git add .gitattributes
+  git commit -m "add b.dat"
+
+  GIT_TERMINAL_PROMPT=0 GIT_TRACE=1 GIT_TRANSFER_TRACE=1 GIT_CURL_VERBOSE=1 git push origin new-branch 2>&1 | tee push.log
+  grep "Uploading LFS objects: 100% (1/1), 1 B" push.log
+  echo "approvals:"
```
chrisd8088, line 319: Minor thought: these `echo` statements could possibly just be inline comments, as I don't think anything output to `stdout` is usually visible when running the tests.

chrisd8088 (approved):
Very nifty, thank you!

t/cmd/git-credential-lfstest.go
```diff
@@ -120,25 +160,25 @@ func credsForHostAndPath(host, path string) (string, string, error) {
...
+			return authtype, u, cred, err
 		}
 	}
 
 	return credsFromFilename(hostFilename)
 }
 
-func credsFromFilename(file string) (string, string, error) {
+func credsFromFilename(file string) (string, string, string, error) {
 	userPass, err := os.ReadFile(file)
```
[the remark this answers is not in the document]
bk2204: Good point, I'll fix that.

PR #5803, opened 2024-06-17
Support multistage authentication with a Git credential helper

bk2204:
In 2.46, Git will support multistage authentication from credential helpers.  This will allow Git and Git LFS to implement functionality like NTLM and Kerberos in the credential helper, which lets this functionality to work even if Git LFS doesn't support it natively.

This requires two separate pieces of data.  First, it involves a `state[]` field, which each credential helper can add to keep track of state.  Second, it allows the usage of a boolean `continue` field, which indicates that the response is multistage and this is not the final stage.

In order to make this work, we adjust a few things.  First, we advertise the `state` capability.  Additionally, we save and pass back the `state[]` fields that the credential helper may send to us.  We also don't change the authentication scheme if the helper told us that this was a multistage response.  Finally, we add a check to avoid a credential helper getting stuck in an infinite loop if it keeps handing back the same credentials.

There are also a variety of preparatory steps which are necessary and are present in their own commits for easy review.

### DIFF
--- a/lfsapi/auth.go
+++ b/lfsapi/auth.go
@@ -16,14 +16,16 @@ import (
 )
 
 var (
-	defaultEndpointFinder = NewEndpointFinder(nil)
+	defaultEndpointFinder  = NewEndpointFinder(nil)
+	defaultMaxAuthAttempts = 3
 )
 
 // DoWithAuth sends an HTTP request to get an HTTP response. It attempts to add
 // authentication from netrc or git's credential helpers if necessary,
 // supporting basic authentication.
 func (c *Client) DoWithAuth(remote string, access creds.Access, req *http.Request) (*http.Response, error) {
-	res, err := c.doWithAuth(remote, access, req, nil)
+	count := 0
+	res, err := c.doWithAuth(remote, &count, access, req, nil)
 
 	if errors.IsAuthError(err) {
 		if len(req.Header.Get("Authorization")) == 0 {
@@ -44,7 +46,8 @@ func (c *Client) DoWithAuth(remote string, access creds.Access, req *http.Reques
 // the same way as DoWithAuth, but will not retry the request if it fails with
 // an authorization error.
 func (c *Client) DoWithAuthNoRetry(remote string, access creds.Access, req *http.Request) (*http.Response, error) {
-	return c.doWithAuth(remote, access, req, nil)
+	count := 0
+	return c.doWithAuth(remote, &count, access, req, nil)
 }
 
 // DoAPIRequestWithAuth sends an HTTP request to get an HTTP response similarly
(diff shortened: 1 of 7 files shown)


lfsapi/auth.go
```diff
@@ -57,18 +59,24 @@ func (c *Client) DoAPIRequestWithAuth(remote string, req *http.Request) (*http.R
 	return c.DoWithAuth(remote, access, req)
 }
 
-func (c *Client) doWithAuth(remote string, access creds.Access, req *http.Request, via []*http.Request) (*http.Response, error) {
+func (c *Client) doWithAuth(remote string, count *int, access creds.Access, req *http.Request, via []*http.Request) (*http.Response, error) {
+	if *count == 3 {
```
chrisd8088, line 63: I might suggest putting this value of `3` into a `const` with a name like `defaultMaxAuthAttempts` or something along those lines.

We might eventually want to make this configuration with a setting like `lfs.auth.retries` or `lfs.auth.maxRetries` (akin to the existing `lfs.ssh.retries` setting, for instance).

lfsapi/auth.go
```diff
@@ -77,7 +85,11 @@ func (c *Client) doWithAuth(remote string, access creds.Access, req *http.Reques
 
 			if credWrapper.Creds != nil {
 				req.Header.Del("Authorization")
-				credWrapper.CredentialHelper.Reject(credWrapper.Creds)
+				if multistage && *count < 3 && res != nil && res.StatusCode == 401 {
```
chrisd8088, line 88: I'd suggest using the `const` here too in place of `3`.

chrisd8088 (approved):

t/cmd/git-credential-lfstest.go
```diff
@@ -27,16 +28,28 @@ type credential struct {
...
 	skip       bool
 }
 
-func (c *credential) Serialize(capabilities map[string]struct{}) map[string][]string {
+func (c *credential) Serialize(capabilities map[string]struct{}, state []string) map[string][]string {
+	formattedState := fmt.Sprintf("lfstest:%s", c.state)
+	formattedMatchState := fmt.Sprintf("lfstest:%s", c.matchState)
 	creds := make(map[string][]string)
 	if c.skip {
-		// Do nothing.
```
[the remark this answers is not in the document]
bk2204: Uh, no, that was a bad squash.  Will restore.

t/cmd/git-credential-lfstest.go
```diff
@@ -111,8 +111,14 @@ func fill() {
 		os.Exit(1)
 	}
 
+	result := map[string][]string{}
 	capas := discoverCapabilities(creds)
-	result := cred.Serialize(capas, creds["state[]"])
+	for _, thisCred := range cred {
```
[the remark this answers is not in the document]
bk2204: Will do.

lfsapi/auth.go
bk2204: Good idea.

PR #4573, opened 2021-08-08
Fix 429 retry-after handling for LFS batch API endpoint

fh1ch:
This MR corrects the handling of `429` retry-after HTTP errors for the *LFS batch API* endpoint and adds the matching integration tests. Said functionality was already working for the *storage API* endpoint and was also covered by integration tests.

Receiving `429` retry-after HTTP errors also for the *LFS batch API* endpoint is a likely scenario when using a Git-server with shared rate-limit bucket like GitLab or reverse HA-Proxy and project with multiple LFS files.

**Scenario before 5b6a929**

``` txt
#     21:45:55.445971 trace git-lfs: tq: running as batched queue, batch size of 100
#     21:45:55.446155 trace git-lfs: run_command: git rev-list --objects --ignore-missing --not --remotes=origin --stdin --
#     21:45:55.447408 trace git-lfs: exec: git '-c' 'filter.lfs.smudge=' '-c' 'filter.lfs.clean=' '-c' 'filter.lfs.process=' '-c' 'filter.lfs.required=false' 'cat-file' '--batch-check'
#     21:45:55.448571 trace git-lfs: exec: git '-c' 'filter.lfs.smudge=' '-c' 'filter.lfs.clean=' '-c' 'filter.lfs.process=' '-c' 'filter.lfs.required=false' 'rev-parse' '--git-common-dir'
21:45:55.453201 trace git-lfs: tq: sending batch of size 1                            
#     21:45:55.453405 trace git-lfs: api: batch 1 files
#     21:45:55.453493 trace git-lfs: creds: git credential cache ("http", "127.0.0.1:50427", "batch-upload-retry-later")
#     21:45:55.453497 trace git-lfs: Filled credentials for http://127.0.0.1:50427/batch-upload-retry-later
#     21:45:55.453506 trace git-lfs: HTTP: POST http://127.0.0.1:50427/batch-upload-retry-later.git/info/lfs/objects/batch
#     21:45:55.453994 trace git-lfs: HTTP: 429
#     21:45:55.454012 trace git-lfs: HTTP: rate limit reached
#     21:45:55.454042 trace git-lfs: api error: LFS: Error
#     21:45:55.454057 trace git-lfs: tq: enqueue retry #1 after 11.00s for "5404ee773986b403c227c48c61913b32109d8b2aa512b6fba044acc14cfe30a5" (size: 17)
#     21:46:06.458175 trace git-lfs: tq: sending batch of size 1
[...]

PR #5804, opened 2024-06-18
Fix panic caused by accessing non-existent header

blanet:
We encountered a panic while checking out a repository with many LFS files, even with a fresh build from main:

```
panic: runtime error: index out of range [0] with length 0

goroutine 49 [running]:
github.com/git-lfs/git-lfs/v3/tq.(*basicDownloadAdapter).download(0xc0000148d0, 0xc0009391a0, 0xc0005e7970, 0x0, 0xc000014620, 0x0, {0xb4d7b0, 0xc000552780})
	/data00/go/pkg/mod/github.com/git-lfs/git-lfs/v3@v3.4.0/tq/basic_download.go:156 +0x14a8
github.com/git-lfs/git-lfs/v3/tq.(*basicDownloadAdapter).DoTransfer(0xc0000148d0, {0x40?, 0x0?}, 0xc0009391a0, 0x0?, 0x0?)
	/data00/go/pkg/mod/github.com/git-lfs/git-lfs/v3@v3.4.0/tq/basic_download.go:96 +0x3a8
github.com/git-lfs/git-lfs/v3/tq.(*adapterBase).worker(0xc0000c46c0, 0x5, {0x0, 0x0})
	/data00/go/pkg/mod/github.com/git-lfs/git-lfs/v3@v3.4.0/tq/adapterbase.go:183 +0x58b
created by github.com/git-lfs/git-lfs/v3/tq.(*adapterBase).Begin
	/data00/go/pkg/mod/github.com/git-lfs/git-lfs/v3@v3.4.0/tq/adapterbase.go:96 +0x270
```

After some digging, we found the panic was caused by accessing a non-existent "Retry-After" response header when receiving a 429 status code. We should use the `Get()` method to safely access response headers.


chrisd8088:
Hey, thanks for the bug report and fix!  This seems correct and valuable to me.

I think it does expose some other problems which we should address at the same time, however.

First, it would be ideal if we could also adds some tests for this condition.  For example, I was able to trigger the panic condition by commenting out the lines which set the `Retry-After` header for [uploads](https://github.com/git-lfs/git-lfs/blob/d065d1f17c293fbbb0544932af7668adbfc1ffe3/t/cmd/lfstest-gitserver.go#L705) or [downloads](https://github.com/git-lfs/git-lfs/blob/d065d1f17c293fbbb0544932af7668adbfc1ffe3/t/cmd/lfstest-gitserver.go#L757) in our test server, building the `lfstest-gitserver` with those changes, and then running the `t/t-batch-storage-retries-ratelimit.sh` test script.

Second, a similar kind of error occurs if one comments out the `Retry-After` header that's set for [batch](https://github.com/git-lfs/git-lfs/blob/d065d1f17c293fbbb0544932af7668adbfc1ffe3/t/cmd/lfstest-gitserver.go#L416) responses, builds the test server, and runs the `t/t-batch-retries-ratelimit.sh` script.  The proposed change to `enqueueAndCollectRetriesFor()` in this PR doesn't prevent that on its own, unfortunately.  This condition doesn't cause a Go panic, but the client exits with an error code and doesn't retry the batch requests at all.

There's a history of efforts to better support 429 response handling, starting with PR #3449, with fixes and improvements in PRs #3930, #4097, and #4573 (and possibly others).  I've spent a fair bit of time today reading through these PRs, and trying to put together some additional commits for this PR.

My `fix-batch-retry-without-header` [branch](https://github.com/chrisd8088/git-lfs/tree/fix-batch-retry-without-header) contains four additional commits to the one in this PR.  Would you consider pulling those into this PR?  Otherwise  I can open a separate PR which contains your changes too.

The purpose of these four extra commits is to add the tests I outlined above, and also fix the related problem with how we handle batch API responses when a 429 is received without a `Retry-After` header.  My changes should ensure that in that situation, as for the ones you addressed with individual object uploads and downloads, we follow the exponential backoff strategy from PR #4097 and make a series of batch API requests until either one succeeds or we reach the maximum number of requests permitted by the `lfs.transfer.maxRetries` configuration setting.

And thank you again for your contribution; we greatly appreciate the help!

errors/types.go
```diff
@@ -412,6 +412,10 @@ type retriableLaterError struct {
 }
 
 func NewRetriableLaterError(err error, header string) error {
+	if header == "" {
+		return nil
+	}
```
chrisd8088, line 417: This is a good change, thank you!

As I mentioned in https://github.com/git-lfs/git-lfs/pull/5804#issuecomment-2178035354, this change on its own isn't enough to help us handle batch API responses when we receive a 429 without a `Retry-After` header, because we don't convert those into any kind of retriable error; we [fall through](https://github.com/git-lfs/git-lfs/blob/d065d1f17c293fbbb0544932af7668adbfc1ffe3/lfshttp/errors.go#L66-L73) and just return a non-retriable error, which means the client just exits in such a case.

To help us handle 429s without `Retry-After` headers from the batch API as well as from individual object uploads and downloads, I've proposed some additional commits in my `fix-batch-retry-without-header` [branch](https://github.com/chrisd8088/git-lfs/tree/fix-batch-retry-without-header).

Would you be willing to pull those into this PR as well, if they seem reasonable to you?

chrisd8088 (commented):
Thanks again for finding this bug and fixing it!

It would be great to add some tests for this condition, like the ones I've put into my `fix-batch-retry-without-header` [branch](https://github.com/chrisd8088/git-lfs/tree/fix-batch-retry-without-header), if you're willing to consider adding the commits from there to this PR.

Thanks again very much for the contribution!

errors/types.go
blanet: Sure! It's difficult for me to construct those tests because I am not very familiar with the implementation details of git-lfs, thanks for your help. :-)

I really appreciate your efforts in tracking the long history of the retry mechanism.

chrisd8088: No worries at all!  It was an interesting exercise and think we're making real improvements.  Thank you again for the key analysis and fix.

chrisd8088:
I'm going to request a review from @bk2204 here, because I don't think I should review my own commits.  :-)

bk2204 (approved):

PR #5863, opened 2024-09-12
git-lfs: omit tags in ls-remote; optimize pre-push

### LINKED ISSUE #5848: Fetching all refs and tags on every push can be really slow for large repositories
**Describe the bug**
At the moment, doing a `git push` on a repository with `git lfs` enabled executes `git lfs pre-push`. One of the commands executed as a part of git pre-push seems to be — 

```
trace git-lfs: exec: git '-c' 'filter.lfs.smudge=' '-c' 'filter.lfs.clean=' '-c' 'filter.lfs.process=' '-c' 'filter.lfs.required=false' 'ls-remote' '--heads' '--tags' '-q' 'origin'
```

In a very large repository, which has CI systems tagging releases every hour and a large number of refs, this currently takes ~10s to execute slowing down `git push` for everyone. 

**Expected behavior**
`--tags` are omitted or there is some sort of caching that avoids making a call to fetch all remotes on all pushes. Tags can be an ever-growing list and most git commands do not fetch tags by default (or it is possible to set fetching of tags off by default easily).

**Output of `git lfs env`**
```
git-lfs/3.5.1 (GitHub; darwin amd64; go 1.22.1)
git version 2.46.0

LocalReferenceDirs=
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
AccessDownload=basic
AccessUpload=basic
GIT_EXEC_PATH=/usr/local/opt/git/libexec/git-core
GIT_TRACE2_PARENT_SID=20240827T034806.415503Z-H0456b86c-P0000772d
git config filter.lfs.process = "git-lfs filter-process --skip"
git config filter.lfs.smudge = "git-lfs smudge --skip -- %f"
```

pastelsky:
Fixes #5848.

Improves the performance of git lfs pre-push particularly for large repositories with numerous tags.

Previously, the pre-push hook would execute 'git ls-remote' with both --heads and --tags flags, which could be slow in repositories with many tags.

Tags can be omitted for pre-push however.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -1277,12 +1277,12 @@ func Fetch(remotes ...string) error {
 	return err
 }
 
-// RemoteRefs returns a list of branches & tags for a remote by actually
-// accessing the remote via git ls-remote.
-func RemoteRefs(remoteName string, ignoreTags bool) ([]*Ref, error) {
+// RemoteRefs returns a list of branches and, optionally, tags for a remote
+// by actually accessing the remote via git ls-remote.
+func RemoteRefs(remoteName string, withTags bool) ([]*Ref, error) {
 	var ret []*Ref
 	args := []string{"ls-remote", "--heads", "-q"}
-	if !ignoreTags {
+	if withTags {
 		args = append(args, "--tags")
 	}
 	args = append(args, remoteName)
@@ -1308,7 +1308,7 @@ func RemoteRefs(remoteName string, ignoreTags bool) ([]*Ref, error) {
 
 			typ := RefTypeRemoteBranch
 			if ns == "tags" {
-				if ignoreTags {
+				if !withTags {
 					return nil, errors.New(tr.Tr.Get("unexpected tag returned by `git ls-remote --heads`: %s %s", name, sha))
 				}
 				typ = RefTypeRemoteTag
--- a/git/git.go
+++ b/git/git.go
@@ -1279,9 +1279,15 @@ func Fetch(remotes ...string) error {
 
 // RemoteRefs returns a list of branches & tags for a remote by actually
 // accessing the remote via git ls-remote.
-func RemoteRefs(remoteName string) ([]*Ref, error) {
+func RemoteRefs(remoteName string, ignoreTags bool) ([]*Ref, error) {
 	var ret []*Ref
-	cmd, err := gitNoLFS("ls-remote", "--heads", "--tags", "-q", remoteName)
+	args := []string{"ls-remote", "--heads", "-q"}
+	if !ignoreTags {
+		args = append(args, "--tags")
+	}
+	args = append(args, remoteName)
+	
+	cmd, err := gitNoLFS(args...)
 	if err != nil {
 		return nil, errors.New(tr.Tr.Get("failed to find `git ls-remote`: %v", err))
 	}
--- a/git/git.go
+++ b/git/git.go
@@ -1286,7 +1286,7 @@ func RemoteRefs(remoteName string, ignoreTags bool) ([]*Ref, error) {
 		args = append(args, "--tags")
 	}
 	args = append(args, remoteName)
-	
+
 	cmd, err := gitNoLFS(args...)
 	if err != nil {
 		return nil, errors.New(tr.Tr.Get("failed to find `git ls-remote`: %v", err))
@@ -1308,6 +1308,9 @@ func RemoteRefs(remoteName string, ignoreTags bool) ([]*Ref, error) {
 
 			typ := RefTypeRemoteBranch
 			if ns == "tags" {
+				if ignoreTags {
+					return nil, errors.New(tr.Tr.Get("unexpected tag returned by `git ls-remote --heads`: %s %s", name, sha))
+				}
 				typ = RefTypeRemoteTag
 			}
 			ret = append(ret, &Ref{name, typ, sha})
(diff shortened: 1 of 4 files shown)


chrisd8088:
Hey, thanks very much for the patch!  I'll take a look in the next few days, and consider whether we should add a test.

Our CI jobs are also not working 100% right now, so I won't kick off CI jobs for this PR yet, but I have some fixes in the pipeline and will get those up into a PR soon, so we can return our CI suite to full functionality.  Once that's done, we have a small queue of PRs like yours to merge.  So, stay tuned, and thanks very much for the contribution!

pastelsky:
Thanks! My understanding was that this was just removing an unused flag from a code path, which is why I didn't add a test initially. What are the assertions we could add here in your opinion?

chrisd8088:
> What are the assertions we could add here in your opinion?

Honestly, I haven't had time to even think about this yet -- it's entirely possible there's nothing we need to do.  I only meant to say that I try to always consider, for any PR, whether tests are something we can add or update.  It was just a generic comment.

Thanks for you patience, and once `$DAYJOB` is slightly less intensive, I should be able to get our CI stack working again.

chrisd8088:
Just a quick update on the state of our CI suite: PR #5866 fixed the macOS jobs, PR #5868 should fix the Windows jobs, and PR #5872 will upgrade us to Go 1.23, which should fix the `Build with specific Go` jobs.  With any luck, we'll be back to 100% success after those are merged.

Thanks very much for your patience!

pastelsky:
Thanks for the update, Chris.

Regards,
Shubham Kanodia

git/git.go
```diff
@@ -1279,9 +1279,15 @@ func Fetch(remotes ...string) error {
...
-func RemoteRefs(remoteName string) ([]*Ref, error) {
+func RemoteRefs(remoteName string, ignoreTags bool) ([]*Ref, error) {
 	var ret []*Ref
-	cmd, err := gitNoLFS("ls-remote", "--heads", "--tags", "-q", remoteName)
+	args := []string{"ls-remote", "--heads", "-q"}
+	if !ignoreTags {
+		args = append(args, "--tags")
+	}
+	args = append(args, remoteName)
+
```
chrisd8088, line 1289: There's some extra trailing whitespace on this line which `go fmt` is detecting -- would you mind removing that?  That should allow our CI suite to pass.

lfs/gitscanner_remotes.go
```diff
@@ -13,7 +13,7 @@ import (
 // remote, use a more explicit diff command.
 func calcSkippedRefs(remote string) []string {
 	cachedRemoteRefs, _ := git.CachedRemoteRefs(remote)
-	actualRemoteRefs, _ := git.RemoteRefs(remote)
+	actualRemoteRefs, _ := git.RemoteRefs(remote, true)
 
 	// The set of remote refs can be very large. Since CachedRemoteRefs only
 	// returns branches, filter the remote refs and convert them to a set for
```
chrisd8088, line 19: I notice that we can also simplify the following loop because we no longer need to filter for branch refs only.  I've written up that simplification and can push a commit to this branch, if you're willing to accept it.  Otherwise I'll put it into a separate branch in my fork, and you can take a look at it there.

chrisd8088 (commented):
Thanks again for this PR!

I realized we could go a bit further with these changes, removing some unneeded logic and adding an error check, and that we should perhaps also expand our test suite of the `git lfs push` command slightly to cover the case where there is a tag ref on the Git remote.

I've written up three commits in my `ls-remote-skip-tags` [branch](https://github.com/chrisd8088/git-lfs/tree/ls-remote-skip-tags) and I'd like to push these into this PR's branch, if that's OK with you.

We'd then solicit a second review from another member of the core team, and that should be everything we need to get this merged!

Let me know if that's all acceptable.  I can also open a second PR, but I think it's easier to combine efforts here in this one, if possible.  If I don't hear back in a while, I may go ahead with my plan as I outlined it; I hope that's OK.

Thanks again for the contribution!

lfs/gitscanner_remotes.go
pastelsky: Hey Chris, please feel free to add any commits to this.

pastelsky:
@chrisd8088 You can add your commits to this, no issues.

chrisd8088:
> You can add your commits to this, no issues — added you as a collaborator to my fork.

Thank you very much!  I've done so, and I'll solicit a review from another @git-lfs/core person (just so I'm not reviewing my own commits!), and we'll go from there.  Thanks again!

git/git.go
```diff
@@ -1279,9 +1279,15 @@ func Fetch(remotes ...string) error {
 
 // RemoteRefs returns a list of branches & tags for a remote by actually
 // accessing the remote via git ls-remote.
-func RemoteRefs(remoteName string) ([]*Ref, error) {
+func RemoteRefs(remoteName string, ignoreTags bool) ([]*Ref, error) {
 	var ret []*Ref
-	cmd, err := gitNoLFS("ls-remote", "--heads", "--tags", "-q", remoteName)
+	args := []string{"ls-remote", "--heads", "-q"}
+	if !ignoreTags {
```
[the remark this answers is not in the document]
chrisd8088: Great suggestion, thank you!  I've adopted it and also took the opportunity to update the comments describing the function a bit too.

PR #5874, opened 2024-10-02
fix: fix slice init length

rusttech:
The intention here should be to initialize a slice with a capacity of  len(refs)  rather than initializing the length of this slice.

### DIFF
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -76,7 +76,7 @@ func fetchCommand(cmd *cobra.Command, args []string) {
 		}
 
 		if len(args) > 1 {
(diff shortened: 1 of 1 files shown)


chrisd8088 (approved):
Thanks for the contribution!

This looks obviously correct to me, so I'll merge it in once our CI tests pass.

Fortunately, there appears to be an existing test which [exercises](https://github.com/git-lfs/git-lfs/blob/65e3ceaeccbe52e8ed69cc62dc8b686b26d1cee8/t/t-fetch.sh#L428-L450) this code, so we don't need to add any extra tests for this PR.

Thanks again for the patch!

PR #5876, opened 2024-10-02
Add batch size config value and use it everywhere

bogomolets-owl:
Hi team! Hope you are doing great!

Me again with some minor tweaks for git lfs.

What this request is about:
git lfs has batch api. Current batch api implemenataion has fixed batch request size (100 oids). 
The purpose of this request is to make it possible to override the batch request size through the configuration.
The default value remains unchanged (100).

Motivation.
100 oids per batch is pretty good for general case of basic transfer agent. 
But custom transfer agents implementations may want to override this value for some reasons.
I am working on a particular example of such transfer agent. The purpose of my agent is to amplify the transfer of small binary files using git lfs.
Generally using git lfs for small files is a bad practice, but the project i am currently working on requires me to use git lfs that way cause
- i have exactly one file extension for files with different sizes (from 5 KB to 5 GB all have the same extension)
- i can't rely on any sort of naming conventions to solve the issue via gitattributes (this will quickly go out of control due to team size)

And to solve my particular problem i need to override the size of a batch request.

Best regards, Alexander Bogomoletz.

### DIFF
--- a/docs/man/git-lfs-config.adoc
+++ b/docs/man/git-lfs-config.adoc
@@ -145,6 +145,16 @@ to simple HTTP.
 If set to true, this enables resumable uploads of LFS objects through
 the tus.io API. Once this feature is finalized, this setting will be
 removed, and tus.io uploads will be available for all clients.
+* `lfs.transfer.batchSize`
++
+The number of objects to download/upload sent in a single batch request
+to the LFS server. Default is 100.
++
+This value should be changed with caution, as it can have a significant
+impact on the performance of the LFS server and the server is free to
+return an HTTP 413 status code if this value is too high as the Batch
+API specification states.
+
 * `lfs.standalonetransferagent`
 +
 Allows the specified custom transfer agent to be used directly for
--- a/docs/man/git-lfs-config.adoc
+++ b/docs/man/git-lfs-config.adoc
@@ -145,16 +145,6 @@ to simple HTTP.
 If set to true, this enables resumable uploads of LFS objects through
 the tus.io API. Once this feature is finalized, this setting will be
 removed, and tus.io uploads will be available for all clients.
-* `lfs.transfer.batchSize`
-+
-The number of objects to download/upload sent in a single batch request
-to the LFS server. Default is 100.
-+
-This value should be changed with caution, as it can have a significant
-impact on the performance of the LFS server and the server is free to
-return an HTTP 413 status code if this value is too high as the Batch
-API specification states.
-
 * `lfs.standalonetransferagent`
 +
 Allows the specified custom transfer agent to be used directly for
@@ -221,6 +211,15 @@ If set to true, this enables rewriting href of LFS objects using
 `url.*.insteadof/pushinsteadof` config. `pushinsteadof` is used only for
 uploading, and `insteadof` is used for downloading and for uploading
 when `pushinsteadof` is not set.
+* `lfs.transfer.batchSize`
++
+The number of objects to download/upload sent in a single batch request
+to the LFS server. Default is 100.
++
+This value should be changed with caution, as it can have a significant
+impact on the performance of the LFS server and the server is free to
+return an HTTP 413 status code if this value is too high as the Batch
+API specification states.
 
 === Push settings
 
--- a/t/t-batch-transfer-size.sh
+++ b/t/t-batch-transfer-size.sh
@@ -68,30 +68,18 @@ begin_test "batch storage download with small batch size"
   assert_server_object "$reponame" "$oid2"
   assert_server_object "$reponame" "$oid3"
 
-  pushd ..
-    git \
-      -c "filter.lfs.process=" \
-      -c "filter.lfs.smudge=cat" \
-      -c "filter.lfs.required=false" \
-      clone "$GITSERVER/$reponame" "$reponame-assert"
-
-    cd "$reponame-assert"
-
-    git config credential.helper lfstest
-    git config --local lfs.transfer.batchSize 1
-
-    GIT_TRACE=1 git lfs pull origin main 2>&1 | tee pull.log
-    if [ "0" -ne "${PIPESTATUS[0]}" ]; then
-      echo >&2 "fatal: expected \`git lfs pull origin main\` to succeed ..."
-      exit 1
-    fi
-
-    actual_count="$(grep -c "tq: sending batch of size 1" pull.log)"
-    [ "3" = "$actual_count" ]
-
-    assert_local_object "$oid1" "${#contents1}"
-    assert_local_object "$oid2" "${#contents2}"
-    assert_local_object "$oid3" "${#contents3}"
-  popd
+  cd ..
+  git config --global lfs.transfer.batchSize 1
+  GIT_TRACE=1 git clone "$GITSERVER/$reponame" "${reponame}-assert" 2>&1 | tee clone.log
+  if [ "0" -ne "${PIPESTATUS[0]}" ]; then
+    echo >&2 "fatal: expected \`git clone\` to succeed ..."
+    exit 1
+  fi
+  actual_count="$(grep -c "tq: sending batch of size 1" clone.log)"
+  [ "3" = "$actual_count" ]
+  cd "${reponame}-assert"
+  assert_local_object "$oid1" "${#contents1}"
+  assert_local_object "$oid2" "${#contents2}"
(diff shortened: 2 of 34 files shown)


chrisd8088 (commented):
Hey, thanks for the idea and the implementation!  I think this is a useful additional feature and option to have, for sure.

Aside from a few implementation notes, the main three things I'd like to suggest for this PR are to rename the new configuration option to `lfs.transfer.batchSize`, add some documentation of the option to the `docs/man/git-lfs-config.adoc` file in the `Upload and download transfer settings` [section](https://github.com/git-lfs/git-lfs/blob/8b2882b7b791635d5deefebcc4edaa9e8eef638a/docs/man/git-lfs-config.adoc#L124-L213), and probably also add a `t/t-batch-size.sh` test file with some small tests of the new option.

As far as documentation is concerned, we should probably note that a server is free to return an HTTP 413 status code 
in its response if it deems the requested batch size too large, as our Batch API specification [states](https://github.com/git-lfs/git-lfs/blob/8b2882b7b791635d5deefebcc4edaa9e8eef638a/docs/api/batch.md?plain=1#L277-L278).

Regarding a test file, I think a single pair of upload and download tests is probably sufficient, very similar to what we have now for the `lfs.transfer.maxRetries` configuration option in the `t/t-batch-retries.sh` [file](https://github.com/git-lfs/git-lfs/blob/8b2882b7b791635d5deefebcc4edaa9e8eef638a/t/t-batch-retries.sh).  In these tests we can set `lfs.transfer.batchSize` to a small number like 1 or 2, and try a push or pull of more than that number of objects while setting `GIT_TRACE=1`.  Checking the trace logs should then find the expected number of API requests (e.g., if we push 3 objects with a batch size of 1, we should find 3 API requests in the logs).

bogomolets-owl:
Hi @chrisd8088!
Thank you for your review!
I added the chages you mentioned to the merge request.

Though i have troubles with running my tests. I am currently working on windows and i managed to test the feature itself, but not the tests so i am not sure the tests are actully behave as expected.
Maybe there are some docs i missed to clarify the development pipeline for lfs? 
I have two specific troubles:
- i can build only windows platform currently by making some changes in makefile. This one can be skipped cause i dont have any other platforms to check the tests against.
- i dont understand how to use my git lfs with my git not ruinig my daily workspace setup. This is what i need help with.

Best regards, Alexander Bogomoletz.

chrisd8088:
Thanks very much for the additional work!  This is looking really good.  I ran your new test on my system and it passes, so that's great.

I may not have time for a full second review for a few days, but I'll make sure to get to it shortly.

As for developing on Windows, the core team typically has not done that, so I'm not sure how much advice I can give.  When I need to check some behaviour on Windows, I use a Git Bash terminal, as provided by the Git for Windows project, and install Go there so I can build the client.  That usually works OK for me to build and run tests, but it's also been a while since I did that, so maybe some details have changed.  You might also try an Azure VM or something like that, where you can develop without affecting your own regular workspace.

chrisd8088:
One small detail which I did notice just now, while running your new test script, is that it has non-executable permissions (100644).  Would you mind giving it executable permissions (100755) in Git?

t/t-batch-transfer-size.sh
```diff
@@ -0,0 +1,97 @@
...
+
+    actual_count="$(grep -c "tq: sending batch of size 1" pull.log)"
+    [ "3" = "$actual_count" ]
+
+    assert_local_object "$oid1" "${#contents1}"
+    assert_local_object "$oid2" "${#contents2}"
+    assert_local_object "$oid3" "${#contents3}"
+  popd
+)
+end_test
```
chrisd8088, line 97: This is nicely done!  I happen to know two tricks we can use to simplify this a bit.

First, we could use `GIT_LFS_SKIP_SMUDGE=1` to avoid fetching Git LFS objects during the `git clone` step, so that would mean you don't have to explicitly override the various `filter.lfs.*` Git configuration options.

But, we don't even need to do that here, because we can actually set the `lfs.transfer.batchSize` option right in the global Git configuration file, and then just do a `git clone` and check the results.

We don't have to worry we might be contaminating the Git configuration of the "real" global Git configuration file (e.g., the `$HOME/.gitconfig` of the person running the tests in their local development environment) because the `begin_test()` shell function [resets](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/testlib.sh#L80-L84) the `HOME` environment variable at the start of every test, and copies in a `.gitconfig` file which has been explicitly [created](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/testhelpers.sh#L601-L618) with a common set of configurations so all our tests run in the same context and are never affected by any settings which might be unique to a particular development environment.  So we're free to run `git config --global` in any test, as the change will be scoped to just that test.

So I'd make the following suggestion:

```suggestion
  cd ..
  git config --global lfs.transfer.batchSize 1

  GIT_TRACE=1 git clone "$GITSERVER/$reponame" "${reponame}-assert" 2>&1 | tee clone.log
  if [ "0" -ne "${PIPESTATUS[0]}" ]; then
    echo >&2 "fatal: expected \`git clone\` to succeed ..."
    exit 1
  fi

  actual_count="$(grep -c "tq: sending batch of size 1" clone.log)"
  [ "3" = "$actual_count" ]

  cd "${reponame}-assert"
  assert_local_object "$oid1" "${#contents1}"
  assert_local_object "$oid2" "${#contents2}"
  assert_local_object "$oid3" "${#contents3}"
)
end_test
```

t/t-batch-transfer-size.sh
```diff
@@ -0,0 +1,97 @@
+﻿#!/usr/bin/env bash
```
chrisd8088, line 1: Thanks for adding these tests, this is great!

Other than one simplification, the only other thing I'd ask is if we can run:
```
$ git update-index --chmod=+x t/t-batch-transfer-size.sh 
```

and commit the result, so that the new file has the same executable permissions as the other `t/t-*.sh` files.  It's a minor detail, really, and makes no difference if you're using the `make t-foo.sh` approach to running the tests.  However, for those of us (like me!) who sometimes cheat a bit and just run `./t-foo.sh`, it helps to have the files all be executable.

chrisd8088 (commented):
Hey, thanks so much for the additions; this is looking really good.  It will be a nice new feature to include in the next release!

I had only a couple of minor suggestions for the new test file.  I'd be happy to put those into a commit and push them to your branch, if you're willing to grant me permission to do so in GitHub.

Otherwise, if you don't mind pulling in the one suggested code change and running the `git update-index --chmod=+x`, that would be all we need before we can merge this.

Thanks again so much for the idea and the implementation!  We sincerely appreciate these kinds of contributions from the community.

docs/man/git-lfs-config.adoc
```diff
@@ -145,6 +145,16 @@ to simple HTTP.
...
+* `lfs.transfer.batchSize`
++
+The number of objects to download/upload sent in a single batch request
+to the LFS server. Default is 100.
++
+This value should be changed with caution, as it can have a significant
+impact on the performance of the LFS server and the server is free to
+return an HTTP 413 status code if this value is too high as the Batch
+API specification states.
+
```
chrisd8088, line 157: This is great documentation, thank you!

I realized on a second read-through that I should perhaps ask if you wouldn't mind shifting this down a bit so that it's organized with the other `lfs.transfer.*` settings, like `lfs.transfer.maxRetries` and the ones following that.  Our listing of configuration options is, sadly, not sorted alphabetically within each section, but if we at least keep the most closely-related entries together, I think that will help people a bit.

bogomolets-owl:
Hi @chrisd8088 !

Sorry for the delay, i've had some busy time recently.

Thank you for your comments!
I agree with your suggestions so i updated the pull request with a new commit!
Also i granted you acces to the fork repository so you can modify it directly just in case it will be needed, so please feel free to make any changes to it you see appropriate!

chrisd8088 (approved):
This looks great!  Thank you again for the contribution!

chrisd8088:
> Sorry for the delay, i've had some busy time recently.

No worries at all, and I really appreciate your time and effort on this PR.  The CI failures are just due to an unrelated change in upstream Git for which I'm working on a fix, and once that's in place we can merge this PR.

Thanks again very much!

PR #5879, opened 2024-10-05
Run tests in parallel on Windows and always cleanup test directories

chrisd8088:
Since we revised our test suite to use the `prove` command in PR #3125, an undiagnosed bug in our `lfstest-count-tests` test helper [program](https://github.com/git-lfs/git-lfs/blob/8b2882b7b791635d5deefebcc4edaa9e8eef638a/t/cmd/lfstest-count-tests.go) has prevented us from running our test scripts in parallel on Windows.  Instead, several workarounds were introduced in that PR to avoid counting parallel test script executions and run all our test scripts sequentially on Windows, which slows down our CI jobs.

(The `lfstest-count-tests` program is used to track the number of active test scripts, so our `lfstest-gitserver` test server [program](https://github.com/git-lfs/git-lfs/blob/8b2882b7b791635d5deefebcc4edaa9e8eef638a/t/cmd/lfstest-gitserver.go) can be started once, shared by all our test scripts, and then stopped after they are all complete.  This works as expected on Linux and macOS systems.)

In this PR, we fix the bug in the `lfstest-count-tests` program, which is the result of the fact that we never call `Close()` on the `test_count.lock` file we [open](https://github.com/git-lfs/git-lfs/blob/8b2882b7b791635d5deefebcc4edaa9e8eef638a/t/cmd/lfstest-count-tests.go#L186) to ensure exclusive access to the `test_count` file.  On Unix systems, this causes no problems, because when we [call](https://github.com/git-lfs/git-lfs/blob/8b2882b7b791635d5deefebcc4edaa9e8eef638a/t/cmd/lfstest-count-tests.go#L209) `Remove()` to delete the lock file, the operating system allows us to unlink the file while a file handle remains open.  On Windows, however, this is not permitted, so the `Remove()` call does not succeed, leaving the lock file in place.  Thus when we run the `lfstest-count-tests` program a second time, it is unable to acquire the lock and exits with an error.

With the bug fixed, we then remove the workarounds from PR #3125, which set the `GIT_LFS_NO_TEST_COUNT` and `GIT_LFS_LOCK_ACQUIRE_DISABLED` environment variables on Windows and use their presence to avoid running the `lfstest-count-tests` program and, when it is run, to avoid creating or removing the lock file.

---

In addition, we also make sure our test suite once again removes all the temporary directories it creates to hold the Git repositories created by our tests.  Although we [set](https://github.com/git-lfs/git-lfs/blob/8b2882b7b791635d5deefebcc4edaa9e8eef638a/t/testenv.sh#L89-L90) the `RM_GIT_LFS_TEST_DIR` environment variable to `yes` when we use `mktemp(1)` to [create](https://github.com/git-lfs/git-lfs/blob/8b2882b7b791635d5deefebcc4edaa9e8eef638a/t/testenv.sh#L87) a temporary directory, this is a remnant of PR #1107, and has not been honoured since the relevant cleanup code was removed from our `shutdown()` shell [function](https://github.com/git-lfs/git-lfs/blob/8b2882b7b791635d5deefebcc4edaa9e8eef638a/t/testhelpers.sh#L644-L654) in PR #3125.  In this PR, we restore the cleanup steps from the original implementation.

Moreover, we ensure that we export the `GIT_LFS_TEST_DIR` and `RM_GIT_LFS_TEST_DIR` environment variables so that our test scripts can reuse the same temporary directory, instead of each script creating a unique directory, as happens now.

At present, running the `make test` build target on any platform results in a large number of stale temporary directories being left behind, which no longer occurs with the changes in this PR.

---

Finally, we remove some unused test variables, improve the `lfstest-count-tests` program's handling of file permissions and errors, revise our Linux RPM package builds to run more test scripts in parallel, and skip installing Strawberry Perl on Windows as our CI workflow now [installs](https://github.com/git-lfs/git-lfs/blob/8b2882b7b791635d5deefebcc4edaa9e8eef638a/.github/workflows/ci.yml#L89-L93) the Git for Windows SDK, which includes Perl and the `prove` command.

This PR will be most easily reviewed on a commit-by-commit basis, with whitespace differences ignored.  Each commit has a detailed description of its changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,7 +81,6 @@ jobs:
       shell: bash
     - run: set GOPATH=%HOME%\go
     - run: choco install -y InnoSetup
-    - run: choco install -y strawberryperl
     - run: make man
       shell: bash
     - run: GOPATH="$HOME/go" PATH="$HOME/go/bin:$PATH" go install github.com/josephspurrier/goversioninfo/cmd/goversioninfo@latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,7 +33,6 @@ jobs:
       shell: bash
     - run: set GOPATH=%HOME%\go
     - run: choco install -y InnoSetup
-    - run: choco install -y strawberryperl
     - run: choco install -y zip
     - run: choco install -y jq
     - run: GOPATH="$HOME/go" PATH="$HOME/go/bin:$PATH" go install github.com/josephspurrier/goversioninfo/cmd/goversioninfo@latest
--- a/rpm/SPECS/git-lfs.spec
+++ b/rpm/SPECS/git-lfs.spec
@@ -71,7 +71,7 @@ export SKIPAPITESTCOMPILE=1
 
 pushd src/github.com/git-lfs/%{name}
   make test
-  make -C t PROVE_EXTRA_ARGS=-j4 test
+  make -C t PROVE_EXTRA_ARGS=-j9 test
 popd
 
 rm -rf ${GIT_LFS_TEST_DIR}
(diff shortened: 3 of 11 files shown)


larsxschneider (approved):
Great find! Thank you 🙇

PR #5882, opened 2024-10-09
Repair and restore all tests of cloning over TLS

### LINKED ISSUE #5658: Some tests are functionally disabled due to `if $TRAVIS`
We have some tests in our testsuite that are functionally disabled because they run `if $TRAVIS`, which is always true if the `TRAVIS` environment variable is not set (fun shell fact!).  Many of these deal with our TLS certificate handling, and they are broken in CI for various reasons depending on the OS.

We should probably investigate what's wrong and how to fix it, since these tests are never run unless `TRAVIS` is set to a shell command returning false.  I should note that at least some of them pass on my Ubuntu 23.10 laptop, so simply reproducing on Linux is not sufficient to trigger the problem.

chrisd8088:
Our `t/t-clone.sh` test script includes one [test](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/t-clone.sh#L225-L269) which is skipped on Windows and two [tests](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/t-clone.sh#L91-L223) that have been accidentally completely disabled for many years.

This PR restores all three tests to full functionality on all our CI and build platforms, with the exception of the legacy CentOS 7 Linux distribution.

It should be easiest to review this PR on a commit-by-commit basis.  Each commit contains a detailed description of its changes.

(Note that we expect to further revise the tests in the `t/t-clone.sh` test script in subsequent PRs, once they are functional again.)

Resolves #5658.

#### Background

As described in #5658, the `cloneSSL` and `clone ClientCert` tests contain [invalid](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/t-clone.sh#L94-L97) [checks](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/t-clone.sh#L155-L158) for the Travis CI environment, which in practice cause the tests to trivially pass in all contexts.  The `if $TRAVIS` shell statement, which lacks any test command or shell builtin like `test` or `[`, evaluates as true when the `TRAVIS` variable is not defined.  Since we migrated our CI jobs to GitHub Actions in PR #3808, this environment variable is never set, so the tests are always skipped.

Further, the `clone ClientCert with homedir certs` test is [skipped](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/t-clone.sh#L229-L232) because the default TLS validation library on Windows does not accept the TLS certificate of our `lfstest-gitserver` test [utility](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/cmd/lfstest-gitserver.go).

We also [set](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/testhelpers.sh#L616) the `http.*.sslVerify` Git configuration option to `false` for our entire test suite, which is a legacy of an attempt get the `clone ClientCert` test to succeed in the Travis CI service in PR #1893.

#### Change Summary

Before we can remove the broken checks for the Travis CI environment, we first resolve a number of problems which have accumulated in the tests over the years.  Some of these are relatively trivial, but others are fairly involved, and sometimes only affect a single platform.

After making these fixes, we remove all the remaining checks of the `TRAVIS` variable, including [two](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/t-askpass.sh#L37-L43) [checks](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/t-askpass.sh#L74-L80) in the `t/t-askpass.sh` test script that do not have the same shell scripting bug as the ones in the `t/t-clone.sh` script.

Finally, we remove the command which sets the `http.*.sslVerify` Git configuration to `false`, thus allowing it to default to `true` again for all our tests.

#### Change Details

We make several revisions to our `git-credential-lfstest` test [utility](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/cmd/git-credential-lfstest.go) in order to permit the `clone ClientCert` test to pass, and also take the opportunity to refactor and simplify its handling of input and output fields, as well as to align its behaviour more closely with that of the `git-credential-store(1)` command and other actual Git credential stores.  One key change we make is to ensure the `git-credential-lfstest` utility  accepts an empty `username` field, as it did prior to PR #5803.  We also make sure that it rewrites paths with Windows volume identifiers into Unix-style paths so that it can find its credential record files when running in the Git Bash environment on Windows.

Next, to help the `clone ClientCert` test pass on the Debian systems we currently run in CI (Debian 10, 11, and 12), we update our `lfstest-gitserver` utility so that it rewrites the `DEK-Info` header of the encrypted key it [generates](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/cmd/lfstest-gitserver.go#L1766) with uppercase `A` through `F` letters in its hexadecimal value string.  Without this workaround, the version of the GnuTLS library used by Git and libcurl on our Debian CI systems rejects the encrypted key file, due to a bug which has subsequently been fixed as of version 3.8.0 of GnuTLS by commit gnutls/gnutls@4604bbde14d2c6adb2af5315f9063ad65ab50aa6 in [merge request 1655](https://gitlab.com/gnutls/gnutls/-/merge_requests/1655).

We then revise the three tests in `t/t-clone.sh` so they are skipped when Git and libcurl have a dependency on the legacy libnss3 library, as is only the case for our builds on the legacy CentOS 7 platform.  Support for this TLS library has been removed entirely from curl as of version 8.3.0, in commit curl/curl@7c8bae0d9c9b2dfeeb008b9a316117d7b9675175.  On CentOS 7, though, Git and libcurl are built with libnss3, and it rejects the TLS [certificate](https://github.com/golang/go/blob/7529d09a11496a77ccbffe245607fbd256200991/src/net/http/internal/testcert/testcert.go#L10-L33) [presented](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/cmd/lfstest-gitserver.go#L73-L74) by our `lfstest-gitserver` program, due to an apparently overly-strict [interpretation](https://security.stackexchange.com/questions/176177/why-does-curl-nss-encryption-library-not-allow-a-ca-with-the-extended-key-usage) of RFC 5280.  Note that we expect to drop support for CentOS 7 builds soon, as the only [equivalent](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/script/lib/distro.rb#L38) platform which has not reached the end of its support lifecycle is SUSE Linux Enterprise Server 12 SP5, and its general support [ends](https://www.suse.com/lifecycle/#suse-linux-enterprise-server-12) this month.  Once we no longer build on CentOS 7, we can remove this PR's checks for the libnss3 library from the tests in `t/t-clone.sh`.

The last non-trivial change we make is to set the `http.sslBackend` Git configuration option to `openssl` when running our TLS cloning tests on Windows.  The default TLS verification backend used by Git and libcurl on Windows is the Secure Channel (Schannel) library, which rejects the certificate presented by our `lfstest-gitserver` program, but not for the same reason as libnss3 does.  Instead, Schannel does not read the `IPAddress` fields of the certificate's Subject Alternative Name (SAN) attributes, one of which has the IPv4 address 127.0.0.1, and that is the address we use to connect to our test Git server.  Schannel only looks for a matching `DNSName` field, and when it does not find one, it rejects the certificate.  As we only care that our tests validate the behaviour of Git LFS, not Git itself, changing Git and libcurl's TLS backend to OpenSSL does not compromise the integrity of our tests.

### DIFF
--- a/commands/command_prune.go
+++ b/commands/command_prune.go
@@ -508,7 +508,7 @@ func pruneTaskGetRetainedWorktree(gitscanner *lfs.GitScanner, fetchconf lfs.Fetc
 
 	// Retain other worktree HEADs too
 	// Working copy, branch & maybe commit is different but repo is shared
-	allWorktrees, err := git.GetAllWorkTrees(cfg.LocalGitStorageDir())
+	allWorktrees, err := git.GetAllWorktrees(cfg.LocalGitStorageDir())
 	if err != nil {
 		errorChan <- err
 		return
@@ -536,9 +536,11 @@ func pruneTaskGetRetainedWorktree(gitscanner *lfs.GitScanner, fetchconf lfs.Fetc
 			go pruneTaskGetRetainedAtRef(gitscanner, worktree.Ref.Sha, retainChan, errorChan, waitg, sem)
 		}
 
-		// Always scan the index of the worktree
-		waitg.Add(1)
-		go pruneTaskGetRetainedIndex(gitscanner, worktree.Ref.Sha, worktree.Dir, retainChan, errorChan, waitg, sem)
+		if !worktree.Prunable {
+			// Always scan the index of the worktree if it exists
+			waitg.Add(1)
+			go pruneTaskGetRetainedIndex(gitscanner, worktree.Ref.Sha, worktree.Dir, retainChan, errorChan, waitg, sem)
+		}
 	}
 }
 
--- a/git/git.go
+++ b/git/git.go
@@ -902,17 +902,114 @@ func GitCommonDir() (string, error) {
 	return tools.CanonicalizePath(path, false)
 }
 
-// A git worktree (ref + path)
+// A git worktree (ref + path + flags)
 type Worktree struct {
-	Ref Ref
-	Dir string
+	Ref      Ref
+	Dir      string
+	Prunable bool
 }
 
-// GetAllWorkTrees returns the refs that all worktrees are using as HEADs plus the worktree's path.
-// This returns all worktrees plus the master working copy, and works even if
+// GetAllWorktrees returns the refs that all worktrees are using as HEADs plus the worktree's path.
+// This returns all worktrees plus the main working copy, and works even if
 // working dir is actually in a worktree right now
-// Pass in the git storage dir (parent of 'objects') to work from
-func GetAllWorkTrees(storageDir string) ([]*Worktree, error) {
+//
+// Pass in the git storage dir (parent of 'objects') to work from, in case
+// we need to fall back to reading the worktree files directly.
+func GetAllWorktrees(storageDir string) ([]*Worktree, error) {
+	// Versions before 2.7.0 don't support "git-worktree list", and
+	// those before 2.36.0 don't support the "-z" option, so in these
+	// cases we fall back to reading the .git/worktrees directory entries
+	// and then reading the current worktree's HEAD ref.  This requires
+	// the contents of .git/worktrees/*/gitdir files to be absolute paths,
+	// which is only true for Git versions prior to 2.48.0.
+	if !IsGitVersionAtLeast("2.36.0") {
+		return getAllWorktreesFromGitDir(storageDir)
+	}
+
+	cmd, err := gitNoLFS(
+		"worktree",
+		"list",
+		"--porcelain",
+		"-z", // handle special chars in filenames
+	)
+	if err != nil {
+		return nil, errors.New(tr.Tr.Get("failed to find `git worktree`: %v", err))
+	}
+
+	stdout, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, errors.New(tr.Tr.Get("failed to open output pipe to `git worktree`: %v", err))
+	}
+	stderr, err := cmd.StderrPipe()
+	if err != nil {
+		return nil, errors.New(tr.Tr.Get("failed to open error pipe to `git worktree`: %v", err))
+	}
+
+	if err := cmd.Start(); err != nil {
+		return nil, errors.New(tr.Tr.Get("failed to start `git worktree`: %v", err))
+	}
+
+	scanner := bufio.NewScanner(stdout)
+	scanner.Split(tools.SplitOnNul)
+	var dir string
+	var ref *Ref
+	var prunable bool
+	var worktrees []*Worktree
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if len(line) == 0 {
+			if len(dir) > 0 && ref != nil && len(ref.Sha) > 0 {
+				worktrees = append(worktrees, &Worktree{
+					Ref:      *ref,
+					Dir:      dir,
+					Prunable: prunable,
+				})
+			}
+			dir = ""
+			ref = nil
+			continue
+		}
+
+		parts := strings.SplitN(scanner.Text(), " ", 2)
+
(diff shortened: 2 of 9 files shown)


t/cmd/git-credential-lfstest.go
```diff
@@ -51,8 +51,10 @@ func (c *credential) Serialize(capabilities map[string]struct{}, state []string)
 				}
 			}
 		}
-	} else if len(c.username) != 0 && len(c.password) != 0 {
-		creds["username"] = []string{c.username}
+	} else if len(c.authtype) == 0 && (len(username) == 0 || username[0] == c.username) {
+		if len(username) == 0 {
+			creds["username"] = []string{c.username}
+		}
```
larsxschneider, line 58: TBH the only thing that I am still puzzled about is why we check for `len(c.authtype) == 0` here.

t/cmd/git-credential-lfstest.go
```diff
@@ -175,6 +175,19 @@ func credsForHostAndPath(host, path string) ([]credential, error) {
 		if err == nil {
 			return cred, err
 		}
+
+		// Ideally we might run cygpath to convert paths like D:/...
+		// to /d/... paths, but we only need to do this to support
+		// one test of the deprecated git-lfs-clone command in our
+		// CI suite, so for simplicity we just do basic rewriting.
+		if len(path) > 2 && path[0] >= 'A' && path[0] <= 'Z' && path[1] == ':' {
+			path = "/" + string(path[0]+32) + path[2:]
```
larsxschneider, line 184: tiny nit: Maybe `strings.ToLower(string(path[0]))` would easier to grasp?

t/cmd/lfstest-gitserver.go
```diff
@@ -1769,6 +1770,20 @@ func generateClientCertificates(rootCert *x509.Certificate, rootKey interface{})
...
+	// https://github.com/gnutls/gnutls/commit/4604bbde14d2c6adb2af5315f9063ad65ab50aa6
+	// https://github.com/gnutls/gnutls/blob/a0aa4780892dcc3c14cc10d823f8766ac75bcd85/lib/x509/privkey_openssl.c#L205-L206
+	dekInfoIndexes := dekInfoRE.FindSubmatchIndex(clientKeyEncPEM)
+	if dekInfoIndexes == nil || len(dekInfoIndexes) != 4 {
+		log.Fatalf("DEK-Info header not found in encrypted private key: %s", string(clientKeyEncPEM))
+	}
+	for i := dekInfoIndexes[2]; i < dekInfoIndexes[3]; i++ {
+		c := clientKeyEncPEM[i]
+		if c >= 'a' && c <= 'f' {
+			clientKeyEncPEM[i] = c - 32
```
larsxschneider, line 1783: Maybe `unicode.ToUpper(rune(c))`

larsxschneider (approved):
Thank you for figuring all these things out! 🙇

PR #5884, opened 2024-10-10
git-lfs: omit tags in show-ref; optimize pre-push

### LINKED ISSUE #5883: git lfs push can be slow in a repository with a very large number of tags due to show-ref
**Describe the bug**
This issue is congruent with https://github.com/git-lfs/git-lfs/issues/5848. After https://github.com/git-lfs/git-lfs/pull/5863,  lfs pre push longer queries tags when running `ls-remote` to fetch remote refs.

However a pre-push still queries local tags here — 

https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/lfs/gitscanner_remotes.go#L15

via 
https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/git/git.go#L1224-L1229

since by default `git show-ref` lists all branches and tags. 
However, one can supply `--branches` to `git show-ref` to restrict this to only local branches.

It seems like from the comment here — 
https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/lfs/gitscanner_remotes.go#L17

that one may not need to query & iterate over tags here as well. 

## Additionally

On a related note, the push command queries tags too via — 

https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/commands/command_push.go#L135-L139

I don't think it's possible to remove the querying of tags here, but let me know if you think otherwise.

pastelsky:
Fixes #5883

Previously, the pre-push hook would execute 'git show-ref' without `--branches`, which could be slow in repositories with many tags.

Improves the performance of git lfs pre-push marginally for repositories that have a large number of local tags.

### DIFF
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -274,7 +274,7 @@ func getRemoteRefs(l *tasklog.Logger) (map[string][]*git.Ref, error) {
 	for _, remote := range remotes {
 		var refsForRemote []*git.Ref
 		if migrateSkipFetch {
-			refsForRemote, err = git.CachedRemoteRefs(remote)
+			refsForRemote, err = git.CachedRemoteRefs(remote, true)
 		} else {
 			refsForRemote, err = git.RemoteRefs(remote, true)
 		}
--- a/git/git.go
+++ b/git/git.go
@@ -1219,11 +1219,17 @@ func Checkout(treeish string, paths []string, force bool) error {
 	return err
 }
 
-// CachedRemoteRefs returns the list of branches & tags for a remote which are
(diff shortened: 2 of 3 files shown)


jarlungoodoo73 (commented):
A client asked us to help MAC OS downloading large files that fail , please help this person with there problem Github Community

jarlungoodoo73:
All yours Github Please MAC OS user.

PR #5886, opened 2024-10-14
hook: fix newlines in command missing message

xdavidwu:
Under most shells `echo "\n"` produces \n, not a newline. This fixes it
via moving to printf, which should be more portable and is encouraged by
POSIX.

### DIFF
--- a/lfs/hook.go
+++ b/lfs/hook.go
@@ -16,9 +16,10 @@ import (
 
 var (
 	// The basic hook which just calls 'git lfs TYPE'
-	hookBaseContent = "#!/bin/sh\ncommand -v git-lfs >/dev/null 2>&1 || { echo >&2 \"\\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, remove this hook by deleting the '{{Command}}' file in the hooks directory (set by 'core.hookspath'; usually '.git/hooks').\\n\"; exit 2; }\ngit lfs {{Command}} \"$@\""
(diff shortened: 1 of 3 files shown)


chrisd8088:
Hey, thanks for the contribution and welcome to the Git LFS project!

On a first look this appears to be a valuable and comprehensive set of changes, which is really great.  I'd like to take a few days to reflect on whether there are any other changes we should include the hooks at the same time, or anything I didn't see on a first review, but either way, I expect we'll be able to merge this shortly.  In the meantime I'll get the CI suite started on this PR.

Thanks again very much!

chrisd8088:
Just a note re the CI builds—it appears GitHub Actions have upgraded their default runners to Ubuntu 24.04, and we'll have to adjust one of our build scripts for that change, so for now, just ignore those CI failures.  I'll get a patch up to deal with the CI breakage shortly.

chrisd8088:
Thanks for your patience!

I haven't had a chance to circle back past this PR yet, and while I've resolved our CI problems with Ubuntu 24.04, unfortunately our CI suite has broken once again, this time due to a small upstream change in Git's `master` development branch.  Once I've cleared that up I'll make sure to give this PR a full review.

chrisd8088 (commented):
Thank you for this improvement!  The hook script changes look good to me.

I did notice that we have an extra [check](https://github.com/git-lfs/git-lfs/blob/dfa3b2d83b99a67b1ae399efbe4bfe3e897f1d58/t/t-update.sh#L90) in our existing `t/t-update.sh` test script at the end of the `replace old hook 6` section in the `update` test, and that we're missing the same check at the end the `replace old hook 5` [section](https://github.com/git-lfs/git-lfs/blob/dfa3b2d83b99a67b1ae399efbe4bfe3e897f1d58/t/t-update.sh#L78-L82) in the test.  This PR then copies the duplicate check another time into its new `replace old hook 7` section.

Would you mind either adding a commit to this PR to address those issues, with a bit of a description in the commit message to explain the history, or granting me permission to push a commit to your branch to do the same?

Aside from that, this all seems ready to merge.  Thank you again for the contribution!

chrisd8088 (approved):
Thank you for the extra commit; this looks great!

PR #5887, opened 2024-10-14
t: increase portability

xdavidwu:
Fix when /bin/sh is not bash, or XDG_CONFIG_HOME is set

### DIFF
--- a/t/Makefile
+++ b/t/Makefile
@@ -1,3 +1,4 @@
+SHELL = bash
 RM ?= rm -f
 PROVE ?= prove
 PROVE_EXTRA_ARGS =
--- a/t/Makefile
+++ b/t/Makefile
@@ -46,9 +46,9 @@ test-commands : $(TEST_CMDS)
 
 test : test-commands
 	$(RM) -r remote test_count{,.lock}
-	@bash -c ". ./testenv.sh && setup && cd t && \
+	@. ./testenv.sh && setup && cd t && \
 		RM_GIT_LFS_TEST_DIR=no $(PROVE) $(PROVE_EXTRA_ARGS) t-*.sh && \
-		shutdown"
+		shutdown
 
 .PHONY : $(TEST_SRCS)
 $(TEST_SRCS) : $(TEST_CMDS)
--- a/t/testhelpers.sh
+++ b/t/testhelpers.sh
@@ -604,6 +604,10 @@ setup() {
     mkdir "$HOME"
   fi
 
+  # do not let Git use a different configuration file
+  unset GIT_CONFIG
+  unset XDG_CONFIG_HOME
+
   if [ ! -f $HOME/.gitconfig ]; then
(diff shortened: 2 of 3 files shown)


chrisd8088:
Thanks for these changes!  I think they seem reasonable, on a first look.

I was initially going to suggest that we could avoid setting the `Makefile`'s `SHELL` [variable](https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html) if we just expanded the [two](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/Makefile#L47) [places](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/Makefile#L54) where we use braces to list the two relevant files, `test_count` and `test_count.lock`.

Then it occurred to me that if we do set `SHELL = bash` as you've done, we might instead be able to simplify our primary target's [recipe](https://github.com/git-lfs/git-lfs/blob/df0afdd19eaf422651b6d131e4ae8a83f6399981/t/Makefile#L48-L50), which has needed an explicit `bash -c "..."` shell command since commit 4f2372215f41a4d8386966df4f37df987cca66f8 in PR #3125.

Would you be amenable to trying that in this PR?  I can push a small commit to your branch, if you're willing to grant me permission to do that; otherwise, would you be able to try a patch like this?

```patch
--- a/t/Makefile
+++ b/t/Makefile
@@ -46,9 +46,9 @@ test-commands : $(TEST_CMDS)
 
 test : test-commands
        $(RM) -r remote test_count{,.lock}
-       @bash -c ". ./testenv.sh && setup && cd t && \
+       @ . ./testenv.sh && setup && cd t && \
                RM_GIT_LFS_TEST_DIR=no $(PROVE) $(PROVE_EXTRA_ARGS) t-*.sh && \
-               shutdown"
+               shutdown
```

---

Also. just a note re the CI builds—it appears GitHub Actions have upgraded their default runners to Ubuntu 24.04, and we'll have to adjust one of our build scripts for that change, so for now, just ignore those CI failures. I'll get a patch up to deal with the CI breakage shortly.

t/Makefile
```diff
@@ -45,9 +46,9 @@ test-commands : $(TEST_CMDS)
 
 test : test-commands
 	$(RM) -r remote test_count{,.lock}
-	@bash -c ". ./testenv.sh && setup && cd t && \
+	@. ./testenv.sh && setup && cd t && \
 		RM_GIT_LFS_TEST_DIR=no $(PROVE) $(PROVE_EXTRA_ARGS) t-*.sh && \
-		shutdown"
+		shutdown
```
chrisd8088, line 51: Thanks so much for adding this change!  It's great to be able to take advantage of the fixed shell environment here.

chrisd8088 (approved):
This looks good!

Thanks again for the contributions.

PR #5900, opened 2024-10-29
doc: mention the pointer size constraint

bk2204:
We check in a variety of places that our pointer files are less than 1 KiB in size.  Because this is a limitation, let's document it in the spec so that users can easily understand the constraint.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -19,6 +19,8 @@ only a pointer file is written.
 * Values MUST NOT contain return or newline characters.
 * Pointer files MUST be stored in Git with their executable bit matching that
 of the replaced file.
+* Pointer files must be less than 1024 bytes in size, including any pointer
+  extension lines.
 * Pointer files are unique: that is, there is exactly one valid encoding for a
   pointer file.
 


docs/spec.md
```diff
@@ -19,6 +19,7 @@ only a pointer file is written.
 * Values MUST NOT contain return or newline characters.
 * Pointer files MUST be stored in Git with their executable bit matching that
 of the replaced file.
+* Pointer files must be less than 1024 bytes in size.
```
chrisd8088, line 22: Thank you very much for adding this key bit of documentation!

The only thing I can think we might expand on here is to clarify that this maximum includes any Git LFS pointer extension lines as well, maybe just by saying something like:
```suggestion
* Pointer files must be less than 1024 bytes in size, including any pointer extension lines.
```

chrisd8088 (approved):

docs/spec.md
bk2204: I've squashed that in.

chrisd8088 (approved):
Thanks very much for this!

PR #5914, opened 2024-11-18
Fix improper negated test expressions and refine TLS client certificate tests

chrisd8088:
Following from the work in PR #5882 to fully enable all the tests in our `t/t-clone.sh` test script, this PR further refines the `clone ClientCert` and `clone ClientCert with homedir certs` [tests](https://github.com/git-lfs/git-lfs/blob/3990c7ab4fa6c99540af3f73f33da47a72df08af/t/t-clone.sh#L166-L289) so they both perform the same sets of checks, including a reworked version of the loop in the `clone ClientCert` test which clarifies when the encrypted version of the TLS/SSL certificate's private key is in use.

In particular, the `clone ClientCert with homedir certs` test does not, at the moment, actually test the `git lfs clone` command, unlike all the other tests in `t/t-clone.sh` test script, and it also does not use the encrypted private key file at all, although we [create](https://github.com/git-lfs/git-lfs/blob/3990c7ab4fa6c99540af3f73f33da47a72df08af/t/testhelpers.sh#L582-L583) some record files for our `git-credential-lfstest` helper program as if this were the case.

We also make minor formatting adjustments and simplifications to our `t/t-clone.sh` test script, and update an incorrect [comment](https://github.com/git-lfs/git-lfs/blob/3990c7ab4fa6c99540af3f73f33da47a72df08af/t/testenv.sh#L140) in our `t/testenv.sh` shell library.

In addition to these changes, we fix a number of instances throughout our test suite where we use incorrectly negated test expressions of the form `[ ! $(grep "foo" bar.log) ]` to try to check that a given log file does not contain a specific string.  These checks succeed at present, but for the wrong reason.  When the given string is not found in the log file, the command substitution returns an empty value, so the test devolves to just `[ ! ]`, and the shell treats the lack of a test expression as a false value, which it then inverts due to the `!` operator, and so the test always passes.  We replace these incorrectly formed checks using either a `-z` test operator (when appropriate) or a simple `grep(1)` command in an `AND` list with an `exit 1` statement, as we already did in PR #5282 to deal with incorrect uses of the `!` shell word.

Prior to these corrections, we adopt a [suggestion](https://github.com/git-lfs/git-lfs/pull/5905#discussion_r1829271742) made during review of PR #5905 to improve the readability of the commands [used](https://github.com/git-lfs/git-lfs/blob/3990c7ab4fa6c99540af3f73f33da47a72df08af/t/testhelpers.sh#L65) to parse the `LocalMediaDir` line from `git lfs env` command's output in various functions in our `t/testhelpers.sh` shell library.

Lastly, we make sure our `lfstest-gitserver` program closes all of its instances of the `Server` structure from the `net/http/httptest` Go package before it [exits](https://github.com/git-lfs/git-lfs/blob/3990c7ab4fa6c99540af3f73f33da47a72df08af/t/cmd/lfstest-gitserver.go#L159).  This has no effect on our tests, but is idiomatic and recommended by the Go documentation.

This PR will be most easily reviewed on a commit-by-commit basis, with whitespace changes ignored.

### DIFF
--- a/t/t-askpass.sh
+++ b/t/t-askpass.sh
@@ -112,7 +112,7 @@ begin_test "askpass: defaults to provided credentials"
 
   GIT_ASKPASS="lfs-askpass" GIT_TRACE=1 GIT_CURL_VERBOSE=1 git push origin main 2>&1 | tee push.log
 
-  [ ! $(grep "filling with GIT_ASKPASS" push.log) ]
+  grep "filling with GIT_ASKPASS" push.log && exit 1
   grep "main -> main" push.log
 )
 end_test
--- a/t/t-clone.sh
+++ b/t/t-clone.sh
@@ -57,8 +57,8 @@ begin_test "clone"
   grep "Cloning into" lfsclone.log
   grep "Downloading LFS objects:" lfsclone.log
   # should be no filter errors
-  [ ! $(grep "filter" lfsclone.log) ]
-  [ ! $(grep "error" lfsclone.log) ]
+  grep "filter" lfsclone.log && exit 1
+  grep "error" lfsclone.log && exit 1
   # should be cloned into location as per arg
   [ -d "$newclonedir" ]
 
@@ -78,8 +78,8 @@ begin_test "clone"
   grep "Cloning into" lfsclone.log
   grep "Downloading LFS objects:" lfsclone.log
   # should be no filter errors
-  [ ! $(grep "filter" lfsclone.log) ]
-  [ ! $(grep "error" lfsclone.log) ]
+  grep "filter" lfsclone.log && exit 1
+  grep "error" lfsclone.log && exit 1
   # clone location should be implied
   [ -d "$reponame" ]
 
@@ -142,8 +142,8 @@ begin_test "cloneSSL"
   grep "Cloning into" lfsclone.log
   grep "Downloading LFS objects:" lfsclone.log
   # should be no filter errors
-  [ ! $(grep "filter" lfsclone.log) ]
-  [ ! $(grep "error" lfsclone.log) ]
+  grep "filter" lfsclone.log && exit 1
+  grep "error" lfsclone.log && exit 1
   # should be cloned into location as per arg
   [ -d "$newclonedir" ]
 
@@ -216,8 +216,8 @@ begin_test "clone ClientCert"
     grep "Cloning into" lfsclone.log
     grep "Downloading LFS objects:" lfsclone.log
     # should be no filter errors
-    [ ! $(grep "filter" lfsclone.log) ]
-    [ ! $(grep "error" lfsclone.log) ]
+    grep "filter" lfsclone.log && exit 1
+    grep "error" lfsclone.log && exit 1
     # should be cloned into location as per arg
     [ -d "$newclonedir" ]
 
@@ -804,8 +804,8 @@ begin_test "clone (HTTP server/proxy require cookies)"
   grep "Cloning into" lfsclone.log
   grep "Downloading LFS objects:" lfsclone.log
   # should be no filter errors
-  [ ! $(grep "filter" lfsclone.log) ]
-  [ ! $(grep "error" lfsclone.log) ]
+  grep "filter" lfsclone.log && exit 1
+  grep "error" lfsclone.log && exit 1
   # should be cloned into location as per arg
   [ -d "$newclonedir" ]
 
@@ -825,8 +825,8 @@ begin_test "clone (HTTP server/proxy require cookies)"
   grep "Cloning into" lfsclone.log
   grep "Downloading LFS objects:" lfsclone.log
   # should be no filter errors
-  [ ! $(grep "filter" lfsclone.log) ]
-  [ ! $(grep "error" lfsclone.log) ]
+  grep "filter" lfsclone.log && exit 1
+  grep "error" lfsclone.log && exit 1
   # clone location should be implied
   [ -d "$reponame" ]
 
--- a/t/t-clone.sh
+++ b/t/t-clone.sh
@@ -94,6 +94,26 @@ begin_test "clone"
     [ ! -e "lfs" ]
     assert_clean_status
   popd
+
+  # Now check clone with standard 'git clone' and smudge download
+  rm -rf "$reponame"
+  git clone "$GITSERVER/$reponame" 2>&1 | tee clone.log
+  if [ "0" -ne "${PIPESTATUS[0]}" ]; then
+    echo >&2 "fatal: expected clone to succeed ..."
+    exit 1
+  fi
+  grep "Cloning into" clone.log
+  [ -d "$reponame" ]
+
+  pushd "$reponame"
+    [ $(wc -c < "file1.dat") -eq 110 ]
+    [ $(wc -c < "file2.dat") -eq 75 ]
+    [ $(wc -c < "file3.dat") -eq 66 ]
+    assert_hooks "$(dot_git_dir)"
+    [ ! -e "lfs" ]
+    [ "6" -eq "$(find "$(dot_git_dir)/lfs/objects" -type f | wc -l)" ]
+    assert_clean_status
+  popd
 )
 end_test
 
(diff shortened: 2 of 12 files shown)


t/t-askpass.sh
```diff
@@ -112,7 +112,7 @@ begin_test "askpass: defaults to provided credentials"
 
   GIT_ASKPASS="lfs-askpass" GIT_TRACE=1 GIT_CURL_VERBOSE=1 git push origin main 2>&1 | tee push.log
 
-  [ ! $(grep "filling with GIT_ASKPASS" push.log) ]
```
larsxschneider, line 115: Nice catch!

t/t-migrate-export.sh
```diff
@@ -52,8 +52,9 @@ begin_test "migrate export (default branch)"
   echo "$main_attrs" | grep -q "*.md !text !filter !merge !diff"
   echo "$main_attrs" | grep -q "*.txt !text !filter !merge !diff"
 
-  [ ! $(echo "$feature_attrs" | grep -q "*.md !text !filter !merge !diff") ]
-  [ ! $(echo "$feature_attrs" | grep -q "*.txt !text !filter !merge !diff") ]
+  echo "$feature_attrs" | grep -q "*.md !text !filter !merge !diff" && exit 1
```
larsxschneider, line 55: Should we single quote the grep argument to avoid possible wildcard expansion (`*`) or history expansion (`!`)?

t/t-clone.sh
```diff
@@ -178,6 +178,33 @@ begin_test "clone ClientCert"
...
+  #
+  # When these options are set, however, Git and Git LFS will independently
+  # invoke "git credential fill" to retrieve the passphrase for the
+  # encrypted private key.  Because the "http.sslCertPasswordProtected"
+  # option is set, Git will query the credential helper, passing a
+  # "protocol=cert" line and a "path=<certfile>" line with the path
+  # from the "http.<url>.sslCert" option.  Note that this path refers
+  # to our unencrypted certificate file; Git does not use the path to
+  # the encrypted private key file from the "http.<url>.sslKey" option
+  # in its query to the credential helper.
```
larsxschneider, line 195: Is that a bug in Git? Why does Git query for the cert file and not the key file?

t/t-clone.sh
```diff
@@ -285,16 +359,48 @@ begin_test "clone ClientCert with homedir certs"
...
+      [ $(wc -c < "file2.dat") -eq 75 ]
+      [ $(wc -c < "file3.dat") -eq 30 ]
+      assert_hooks "$(dot_git_dir)"
+      [ ! -e "lfs" ]
+      assert_clean_status
+    popd
+
+    # Now check clone with standard 'git clone' and smudge download
+    rm -rf "$reponame"
+    git clone "$CLIENTCERTGITSERVER/$reponame"
```
larsxschneider, line 395: Shouldn't we check that `git clone` creates files in `$reponame` that have the expected names and sizes as we did above for `git lfs clone`?

t/t-clone.sh
```diff
@@ -227,16 +263,14 @@ begin_test "clone ClientCert"
       [ $(wc -c < "file2.dat") -eq 75 ]
       [ $(wc -c < "file3.dat") -eq 30 ]
       assert_hooks "$(dot_git_dir)"
+      [ ! -e "lfs" ]
       assert_clean_status
```
larsxschneider, line 267: Why don't we clone here with `git clone` as a test, too?

larsxschneider (approved):
Great PR. I commented with a few minor questions/suggestions but the PR would still be ready to go.

t/t-clone.sh
chrisd8088: Well, it's behaviour that stems from this [line](https://github.com/git/git/blob/090d24e9af6e9f59c3f7bee97c42bb1ae3c7f559/http.c#L698) where the certificate file's path, not the key file's path, is sent to the credential helper, and it's clearly [documented](https://github.com/git/git/blob/090d24e9af6e9f59c3f7bee97c42bb1ae3c7f559/Documentation/config/http.txt#L198-L202), because the certificate file *may* be encrypted, since those PEM files often contain both the certificate and the private key.  From git/git@30dd916348001e4313708473d91d633d3b14d1b5 there's this explanation, as well:

> The user is always prompted, even if the certificate is not encrypted.  This should be fine; unencrypted certificates are rare and a security risk anyway.

In our `lfstest-gitserver` test utility, we only [encrypt](https://github.com/git-lfs/git-lfs/blob/3990c7ab4fa6c99540af3f73f33da47a72df08af/t/cmd/lfstest-gitserver.go#L1768) the private key, and the client certificate file we create does not include it, so there's no particular need to encrypt that file.

To your point, the Git LFS client should arguably behave more like Git, and pass the file to the certificate file, not the key file.  At the moment, we simply expect to be able to [read](https://github.com/git-lfs/git-lfs/blob/3990c7ab4fa6c99540af3f73f33da47a72df08af/lfshttp/certs.go#L87) the client certificate file, but [check](https://github.com/git-lfs/git-lfs/blob/3990c7ab4fa6c99540af3f73f33da47a72df08af/lfshttp/certs.go#L102) if the private key file was encrypted, and if it was, [call](https://github.com/git-lfs/git-lfs/blob/3990c7ab4fa6c99540af3f73f33da47a72df08af/lfshttp/certs.go#L103) a function and pass the key file's path to it, so it [uses](https://github.com/git-lfs/git-lfs/blob/3990c7ab4fa6c99540af3f73f33da47a72df08af/lfshttp/certs.go#L43) that path in the call to the credential helper.  Changing that is outside the scope of this PR, but it's worth noting in case we want to adjust it in the future.

I think having the extra comments in these tests is a good step in that direction, since it makes it clear what's going on and the differences between Git and Git LFS in this regard.

t/t-migrate-export.sh
chrisd8088: I've opened #5915 to track the problems in our tests' use of regular expression meta-characters.

t/t-clone.sh
chrisd8088: If I understand the question correctly, I think this PR's changes implement that suggestion.

The old version of this test (prior to this PR's changes) would only [run](https://github.com/git-lfs/git-lfs/blob/f1249933f1e33b758743c6cfdf02e473f2935b07/t/t-clone.sh#L236-L238) a single `git clone` after exiting its loop, which is why it [reset](https://github.com/git-lfs/git-lfs/blob/f1249933f1e33b758743c6cfdf02e473f2935b07/t/t-clone.sh#L234) the `http.<url>.sslKey` configuration option, since the second pass through the loop would have [changed](https://github.com/git-lfs/git-lfs/blob/f1249933f1e33b758743c6cfdf02e473f2935b07/t/t-clone.sh#L211) that to refer to the encrypted private key file instead of the unencrypted one.

But this PR's commit 3079fd783c92a2368dd4bd400e6e3606e7ed2f72 moves the `git clone` into the loop, so it now runs twice, once with each version of the private key file.  Quoting from that commit's description:

> As we expect to make several other adjustments to this test in subsequent commits in this PR, we first bring the `git clone` command into the same loop with the `git lfs clone` command, which means we no longer need to reset the `http.<url>.sslKey` configuration option, as we want to use the same value for both cloning commands.  This change also ensures we perform all the same checks with both types of TLS/SSL certificate keys.

t/t-clone.sh
chrisd8088: That's a good question.  This particular test design, with a `git clone` [command](https://github.com/git-lfs/git-lfs/blob/f1249933f1e33b758743c6cfdf02e473f2935b07/t/t-clone.sh#L160-L162) after the `git lfs clone` command and subsequent checks, dates back to commits  4c64e8212454f248478f2d431f3e6931e3b67fe6 and 8f91a1bc7fe0c74cde254b432302e6ac871d1a86 of PR #1067, when the `cloneSSL` test was introduced.  It was then replicated in the `clone ClientCert` test when that was added in commit daba49ae43340ee30b2e7c9a6b8d5927dbd71f51 of PR #1893, and now this PR replicates it again into this `clone ClientCert with homedir certs` test.

In all of these cases, we're just relying on the `set -e` shell option to cause the tests to fail if the `git clone` doesn't succeed.  We definitely have some other tests, like those in the `t/t-filter-process.sh` script, which [confirm](https://github.com/git-lfs/git-lfs/blob/f1249933f1e33b758743c6cfdf02e473f2935b07/t/t-filter-process.sh#L44-L63) Git LFS objects have been fetched by `git clone`.

Still, we might as well flesh out our tests in `t/t-clone.sh` with some additional checks, just to be on the safe side.  I'll push a commit which does that to this PR, and then merge.  Thank you for the speedy review!

t/t-clone.sh
larsxschneider: Sorry about the comment. That didn't made sense.

PR #5920, opened 2024-11-24
Update release process documentation and scripts

chrisd8088:
This PR updates our "Releasing Git LFS" documentation [page](https://github.com/git-lfs/git-lfs/blob/6340befc60876f4f039f215479d9d5a945f817e1/docs/howto/release-git-lfs.md) to reflect the current state of our release process, and also adjusts two of the scripts used in that process so they function successfully on macOS systems (or provide more useful error messages if the necessary dependencies are not available).

In particular, our `script/upload` script expects that the `sha3sum(1)` utility is available and that it [accepts](https://github.com/git-lfs/git-lfs/blob/6340befc60876f4f039f215479d9d5a945f817e1/script/upload#L297) input in the format [generated](https://github.com/git-lfs/git-lfs/blob/6340befc60876f4f039f215479d9d5a945f817e1/script/hash-files#L33) by our `script/hash-files` script, i.e.:
```
<hash-name> (<file-name>) = <hash-value>
```

However, the version of the `sha1sum(1)` program packaged and installed by Homebrew is different from the one [packaged](https://sources.debian.org/src/libdigest-sha3-perl/) by Debian, which is a Perl script that [accepts](https://sources.debian.org/src/libdigest-sha3-perl/1.04-1/sha3sum/#L252-L259) a wider range of input formats, including the format output by our `script/hash-files` script.  The program provided by Homebrew's `sha3sum` [formula](https://formulae.brew.sh/formula/sha3sum) requires input in the following format, and is not able to determine the appropriate algorithm to use without it being specified by an `-a` command-line option.
```
<hash-value>  <file-name>
```
Therefore we adjust our `script/upload` script so it is able to function with either implementation of the `sha1sum(1)` utility, since the Perl script also [accepts](https://sources.debian.org/src/libdigest-sha3-perl/1.04-1/sha3sum/#L260-L264) the same format as the Homebrew version.

This PR will be most easily reviewed on a commit-by-commit basis, with whitespace changes ignored.

### DIFF
--- a/docs/howto/release-git-lfs.md
+++ b/docs/howto/release-git-lfs.md
@@ -282,15 +282,15 @@ and then use that branch as the base for the PATCH release.
      $ git log --merges --first-parent vM.N.(P-1)...main
      ```
 
-   3. For each merge that you want to backport, run:
+  3. For each merge that you want to backport, run:
 
-      ```ShellSession
-      $ git cherry-pick -m1 <SHA-1>
-      ```
+     ```ShellSession
+     $ git cherry-pick -m1 <SHA-1>
+     ```
 
-      This will cherry-pick the merge onto your release branch, using
-      the `-m1` option to specify that the first parent of the merge
-      corresponds to the mainline.
+     This will cherry-pick the merge onto your release branch, using
+     the `-m1` option to specify that the first parent of the merge
+     corresponds to the mainline.
 
-   4. Then follow the [guidelines](#building-a-release) above, using the
-      `release-M.N` branch as the base for the new PATCH release.
+  4. Then follow the [guidelines](#building-a-release) above, using the
+     `release-M.N` branch as the base for the new PATCH release.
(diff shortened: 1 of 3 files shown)


docs/howto/release-git-lfs.md
```diff
@@ -270,7 +275,12 @@ When building a PATCH release, we cherry-pick merges from `main` to the
 `vM.N` release branch, creating the branch first if it does not exist,
 and then use that branch as the base for the PATCH release.
 
-  1. If the `release-M.N` branch does not already exist, create it from
+  1. Upgrade or downgrade the Go version used in the `git-lfs/build-dockers`
```
larsxschneider, line 278: When would we downgrade?

larsxschneider (approved):

docs/howto/release-git-lfs.md
chrisd8088: The most obvious situation occurs when Go releases a new minor (or major) version which drops support for older platforms, and we've moved our development branches ahead to keep pace with that.

However, we normally want our patch releases to support the same set of platform versions as all the other Git LFS releases we've made in that series, so we generally don't want to upgrade to a new minor (or major) version of Go for a patch release, and might have to downgrade Go temporarily as a result.

PR #5930, opened 2024-12-09
Tighten security of Actions workflows

bk2204:
I was recently reading an article about a repository whose release workflow was compromised due to code injection and how it could have been detected using [zizmor](https://github.com/woodruffw/zizmor).  I did some auditing here of the Actions workflows, both manually and with zizmor, and I didn't find any obvious security problems, but I did some hardening.

The first patch was from my manual auditing and covers some improved shell quoting.  This isn't currently exploitable, but we might as well avoid the problem altogether and using proper shell quoting is a best practice anyway.  The second patch comes from the recommendation of zizmor.

We _are_ following the best practice of storing expanded Actions patterns (those with `${{}}`) in an environment variable before using them, because this prevents injections (such as from refs containing shell characters).  Actions doesn't have any sort of quoting for these cases, but by passing them in via the environment, we can use the normal shell quoting rules for these purposes.

So overall, I think we're in a good spot, but this PR just makes it a little less inviting as a target.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,11 +205,11 @@ jobs:
         docker version -f '{{.Server.Experimental}}'
     - uses: docker/setup-qemu-action@v3
     - run: git clone https://github.com/git-lfs/build-dockers.git "$HOME/build-dockers"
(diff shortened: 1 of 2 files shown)


chrisd8088 (approved):
Looks great, thanks!

PR #5931, opened 2024-12-10
Simplify macOS CI jobs and specify Go toolchain version

chrisd8088:
The macOS runners provided by GitHub Actions are now provisioned with pre-installed versions of several Homebrew formulae that we currently re-install, which leads to annotation warnings in our CI job summaries.  We can therefore simplify our CI jobs on macOS slightly by eliminating the unnecessary Homebrew invocations.

As well, we now also have a warning from CodeQL that our `go.mod` file does not specify a Go toolchain, which we can resolve most easily by providing a full Go version in the `go` line of that file.


script/build-git
```diff
@@ -4,8 +4,8 @@ DIR="$1"
 
 case $(uname -s) in
   Darwin)
-    brew install curl zlib pcre2 openssl
-    brew link --force curl zlib pcre2 openssl
```
larsxschneider, line 8: Would it make sense to add something like the following just in case the base image ever changes?
```
which curl >/dev/null || echo "curl not installed"
which openssl >/dev/null || echo "openssl not installed"
which pcre2grep >/dev/null || echo "pcre2 not installed"
```

Or maybe even print the version numbers to the logs. Since the script run with `-e` we would stop if the dependencies are not available.
```
curl --version
openssl --version
pcre2grep --version
```

larsxschneider (approved):
LGTM

script/build-git
chrisd8088: That's a good question, and it prompted me to take a much deeper look into which libraries we actually require to build Git for our test suite on macOS.

As it turns out, we don't even need to install the `zlib` Homebrew formula, as the `libz` library provided in contemporary macOS versions is perfectly sufficient for our needs, so we can drop `brew install` entirely.

Moreover, while for now we still need to link against the Homebrew version of `libcurl` in order to deal with the regression in the version currently provided by macOS (per PR #5866), we can avoid running `brew link --force` by just using `brew --prefix curl` to set the `CURLDIR` variable.

As for reporting the versions of our build dependencies, I concluded that since we don't actually care about the exact versions of the command-line tools, only whether we link against the `libcurl` from Homebrew or not, it was most useful to run `otool -L` at the end of our script to see the dynamic library dependencies of the `git-http-fetch` binary we've built.  (That Git binary in particular, unlike most of the others, will be linked against `libcurl` in order to be able to perform HTTP requests, so it's a good one to check.)

For completeness we'll also run `ldd` against the `git-http-fetch` binary on Linux, although at present we don't have the same kind of concern about which installation of a specific library we link against on Linux as we do on macOS.

As well, I added the `--build-options` argument to the `git --version` command we already [run](https://github.com/git-lfs/git-lfs/blob/a32a02b44bf8a511aa14f047627c49e1a7fd5021/script/build-git#L41), but of course that will only output extra information in our `build-latest` CI job, as the Git version our `build-earliest` job builds lacks support for that option (but conveniently just ignores it).

PR #5942, opened 2024-12-20
Rename git-lfs-migrate.adoc to migrate.ads

backforwards:


chrisd8088:
Can you explain why this is a useful change, given that the PR doesn't alter any other AsciiDoc filenames, and that `.adoc` is the [recommended](https://asciidoctor.org/docs/asciidoc-recommended-practices/#document-extension) file extension for this file type?

If there's no valid rationale for this change, we'll close this PR as it appears to be a spurious contribution.

PR #5944, opened 2024-12-22
docs/man/git-lfs-migrate.adoc: remove duplicate flag

fedirz:
Removed the duplicate flag `--include-ref` flag mention.
Before:
```
--include-ref=<refname>:
  See "Include and exclude references".
--include-ref=<refname>:
  See "Include and exclude references".
```
After:
```
--include-ref=<refname>:
  See "Include and exclude references".
```


chrisd8088 (approved):
Hey, thanks for the correction!

It looks like we inadvertently introduced these duplicate lines back in commit a9c8c0eadfb1e3b1b93470385afa9e8b78de90a4 of PR #5054 when we converted the manual pages to AsciiDoc.

Issue #5947, opened 2024-12-27
Add option to migrate import to add old SHA to the new commit message text


dparrini:
The import mode has the `--object-map=<path>` flag which helps after the migration. But having the old SHA metadata directly in the commit messages' body will help using git-based tools built-in search to look for specific commits.

It would be nice to have a flag that allows `migrate import` to update the messages.

Thanks.

chrisd8088:
Hey, thanks for the suggestion and the accompanying PR!

We always appreciate contributions like this, and I think it's a good idea to provide an option like this with the `git lfs migrate` command.  I'll make some suggestions on the PR itself, but the basic implementation looks reasonable.  Thank you again for the idea and the work on the PR!

PR #5948, opened 2024-12-27
Add --append-old-sha option to migrate import command, appends old-SH…

dparrini:
Hi, this PR is a candidate solution to the issue that I opened today: 
https://github.com/git-lfs/git-lfs/issues/5947

The original tests passes (tested on WSL) but I'm not sure how to add tests for this feature. It also lacks mention on documentation. Any help is appreciated if this PR is reviewed as useful.

### DIFF
--- a/commands/command_migrate.go
+++ b/commands/command_migrate.go
@@ -63,6 +63,10 @@ var (
 	// migrateFixup is the flag indicating whether or not to infer the
 	// included and excluded filepath patterns.
 	migrateFixup bool
+
+	// appendOldShaToMessages is the flag indicating whether or not to append
+	// to rewritten commit messages the old commit SHA.
+	appendOldShaToMessages bool
 )
 
 // migrate takes the given command and arguments, *gitobj.ObjectDatabase, as well
@@ -115,9 +119,10 @@ func rewriteOptions(args []string, opts *githistory.RewriteOptions, l *tasklog.L
 		Include: include,
 		Exclude: exclude,
 
-		UpdateRefs:        opts.UpdateRefs,
-		Verbose:           opts.Verbose,
-		ObjectMapFilePath: opts.ObjectMapFilePath,
+		UpdateRefs:             opts.UpdateRefs,
+		Verbose:                opts.Verbose,
+		ObjectMapFilePath:      opts.ObjectMapFilePath,
+		AppendOldShaToMessages: opts.AppendOldShaToMessages,
 
 		BlobFn:            opts.BlobFn,
 		TreePreCallbackFn: opts.TreePreCallbackFn,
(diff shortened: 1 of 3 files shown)


chrisd8088 (commented):
Thank you so much for this PR!  I think it's a good idea to be able to offer this additional functionality when using the `git lfs migrate` command, and our users will appreciate it.

There are a couple of things we'll need for this PR to be complete, in addition to the things I've mentioned in review comments regarding terminology and handling commits with pre-existing trailers.

First, I think we should offer this functionality when using both the `export` and `import` sub-commands of the `git lfs migrate` command, not just with the `import` sub-command.

An entry describing the behaviour of this feature and its new command-line option will need to be added to both the `IMPORT` [section](https://github.com/git-lfs/git-lfs/blob/e4df48288edb0fae2f985fd88ad847dd98e532ec/docs/man/git-lfs-migrate.adoc?plain=1#L235-L262) and the `EXPORT` [section](https://github.com/git-lfs/git-lfs/blob/e4df48288edb0fae2f985fd88ad847dd98e532ec/docs/man/git-lfs-migrate.adoc?plain=1#L298-L313) of the `git-lfs-migrate(1)` manual page.

Last but not least, we'll want to add a number of tests to the `t/t-migrate-import.sh` [test script](https://github.com/git-lfs/git-lfs/blob/e4df48288edb0fae2f985fd88ad847dd98e532ec/t/t-migrate-import.sh) and the `t/t-migrate-export.sh` [test script](https://github.com/git-lfs/git-lfs/blob/e4df48288edb0fae2f985fd88ad847dd98e532ec/t/t-migrate-export.sh).  (I'd probably add them near the tests of the `--object-map` flag, myself.)

These don't need to be especially complicated tests, but should exercise the handling of commits with pre-existing trailers (i.e., no extra whitespace line between the existing trailers and the `LFS-*` one we generate) and commits without such trailers (i.e., one extra whitespace line only).  Other things to check include:
- Handling commits with and without trailing line feed characters.  Git normally ensures there is one trailing `\n`, per the `git-whitespace(1)` manual [page](https://git-scm.com/docs/git-stripspace), but we should be prepared to handle unusual cases created by, for instance, using the `git commit --cleanup=verbatim` [option](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---cleanupltmodegt).  If there is no trailing LF character, we'll need to add two before our trailer; if there's just one, we'll need to add one (unless the final line is itself a trailer); and if there's more than one we can just append our trailer line.
- Handling commits where no changes are made and the SHA stays the same, so we want to verify our trailer is not appended in this case.
- Using the new flag with the `--no-rewrite` option should result in an error message and no changes.

Thank you again very much for getting this work underway, we really appreciate these kinds of contributions from the community!

PR #5959, opened 2025-01-12
docs/man/git-lfs-prune.adoc: fix --(no-)verify-unreachable description

m3ka24:
I could not find an option `--verify-reachable` / `--no-verify-reachable` on prune sub command.
```
$ git lfs prune --verify-reachable
Error: unknown flag: --verify-reachable

$ git lfs prune --no-verify-reachable
Error: unknown flag: --no-verify-reachable
```

On the other hand, the option `--verify-unreachable` / `--no-verify-unreachable` seems to exist.
```
$ git lfs prune --verify-unreachable
prune: 0 local objects, 1 retained, done.

$ git lfs prune --no-verify-unreachable
prune: 0 local objects, 1 retained, done.
```

So I changed the content of adoc to `--verify-unreachable` / `--no-verify-unreachable`.

### DIFF
--- a/docs/man/git-lfs-prune.adoc
+++ b/docs/man/git-lfs-prune.adoc
@@ -60,10 +60,10 @@ gitignore(5).
 `--no-verify-remote`::
   Disables remote verification if lfs.pruneverifyremotealways was enabled in
   settings. See <<_verify_remote>>.
-`--verify-reachable`::
+`--verify-unreachable`::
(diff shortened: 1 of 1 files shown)


chrisd8088 (approved):
Thank you for the correction!  This looks like a minor oversight we missed when PR #5648 was merged last year.

We always appreciate improvements like this to our documentation, so thank you very much!

Issue #5888, opened 2024-10-16
Provide a command to get temporary download URLs


redsun82:
**Describe the issue**
I have a use case where I would like some way to get the temporary download URLs for files stored in LFS.

In our project, we use git LFS for some artifacts and [bazel](https://github.com/bazelbuild/bazel) for building. As not all LFS files are required on all platforms and for all builds, I wanted bazel to drive downloading LFS files by need, with normal git LFS checkout disabled via `.lfsconfig` files and `lfs.fetchinclude`/`lfs.fetchexclude`.

As bazel has its own way of managing downloads (avoiding too many downloads at the same time and caching them), I needed a way to provide it download URLs for LFS files. By reading up on the LFS protocol I came up with [this script](https://github.com/github/codeql/blob/main/misc/bazel/internal/git_lfs_probe.py) which works pretty well in the general case. However, it turned out to be pretty fragile on the authentication side (there are so many ways `git` and `git lfs` can authenticate!).

Would it be possible to have a plumbing command exposed that would provide those URLs directly out of `git lfs`, so that I would not have to go through the headache of reimplementing the whole authentication routine?

What I'm thinking is something like
```
git lfs batch-request path/to/file path/to/another/file
```
that would return the response to a [batch request](https://github.com/git-lfs/git-lfs/blob/main/docs/api/batch.md) for the files listed as inputs. Even better if a `--skip-local` or similar flag would make this skip files that are already smudged on disk, and only process LFS pointers (but I could also do that kind of filtering on my side).

cc @criemen

chrisd8088:
Hey, thanks for the suggestion!  I'll mark this as a possible enhancement and put it in the backlog.

And if anyone has time to offer a PR with an implementation and some tests and documentation, that would be fantastic, as our core team is quite small right now and just responding to bug reports and general questions absorbs a good percentage of our available time, leaving not much left over for feature development.

redsun82:
I've jotted down https://github.com/git-lfs/git-lfs/pull/5913 as a draft implementation that does what we need. Obviously, it's far from a mergeable state, as it has no tests, no documentation and nothing done for localization, but it can be a good starting point (and we can actually start using that from another fork). I will not be able to contribute much more in the future, but maybe some colleagues will 🙂

chrisd8088:
[...]

PR #5973, opened 2025-01-21
Add `--dry-run` option to `fetch`

redsun82:
This was modelled after the `git-lfs push --dry-run` option. I've removed some apparently unused code:
* computing and reporting of ready nodes
* deduplication of multiple objects with the same OID (as far as I can tell, that is already carried out by the transfer queue class, so no need to redo it in the fetch command)
* the output channel of the main fetch function which was always `nil`.

For testing, I added `--dry-run` testing to a couple of cases, but I wonder if we should add it for _all_ cases.

This is a part of what was discussed in https://github.com/git-lfs/git-lfs/issues/5888.

### DIFF
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"time"
+	"sync"
 
 	"github.com/git-lfs/git-lfs/v3/filepathfilter"
 	"github.com/git-lfs/git-lfs/v3/git"
@@ -324,10 +325,11 @@ func scanAll() []*lfs.WrappedPointer {
 	return pointers
 }
 
-func printTransfers(out <-chan *tq.Transfer) {
+func printTransfers(wg sync.WaitGroup, out <-chan *tq.Transfer) {
 	for p := range out {
 		Print("%s %s => %s", tr.Tr.Get("fetch"), p.Oid, p.Name)
 	}
+	wg.Done()
 }
 
 // Fetch
@@ -338,9 +340,11 @@ func fetch(allpointers []*lfs.WrappedPointer) bool {
 		getTransferManifestOperationRemote("download", cfg.Remote()),
 		cfg.Remote(), tq.WithProgress(meter), tq.DryRun(fetchDryRunArg),
 	)
+	var wg sync.WaitGroup
 
 	if fetchDryRunArg {
-		go printTransfers(q.Watch())
+		wg.Add(1)
+		go printTransfers(wg, q.Watch())
 	}
 
 	for _, p := range pointers {
@@ -358,6 +362,7 @@ func fetch(allpointers []*lfs.WrappedPointer) bool {
 		ok = false
 		FullError(err)
 	}
+	wg.Wait()
 	return ok
 }
 
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -361,7 +361,7 @@ func fetch(allpointers []*lfs.WrappedPointer) bool {
 		ok = false
 		FullError(err)
 	}
-	if (fetchDryRunArg) {
+	if fetchDryRunArg {
 		wg.Wait()
 	}
 	return ok
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -107,7 +107,7 @@ func fetchCommand(cmd *cobra.Command, args []string) {
 		verifyUnreachable := fetchPruneCfg.PruneVerifyUnreachableAlways
 
 		// assume false for non available options in fetch
-		prune(fetchPruneCfg, verify, verifyUnreachable, false, false, false)
+		prune(fetchPruneCfg, verify, verifyUnreachable, false, fetchDryRunArg, fetchDryRunArg)
 	}
 
 	if !success {
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -324,34 +324,7 @@ func scanAll() []*lfs.WrappedPointer {
 	return pointers
 }
 
-func WatchFetchedPointers(pointers []*lfs.WrappedPointer, queue *tq.TransferQueue) <-chan *lfs.WrappedPointer {
-	out := make(chan *lfs.WrappedPointer)
-	dlwatch := queue.Watch()
-
-	go func() {
-		// fetch only reports single OID, but OID *might* be referenced by multiple
-		// WrappedPointers if same content is at multiple paths, so map oid->slice
-		oidToPointers := make(map[string][]*lfs.WrappedPointer, len(pointers))
-		for _, pointer := range pointers {
-			plist := oidToPointers[pointer.Oid]
-			oidToPointers[pointer.Oid] = append(plist, pointer)
-		}
-
-		for t := range dlwatch {
-			plist, ok := oidToPointers[t.Oid]
-			if !ok {
-				continue
-			}
-			for _, p := range plist {
-				out <- p
-			}
-		}
-		close(out)
-	}()
-	return out
-}
-
-func PrintFetchedPointers(out <-chan *lfs.WrappedPointer) {
+func PrintTransfers(out <-chan *tq.Transfer) {
 	for p := range out {
 		Print("%s %s => %s", tr.Tr.Get("fetch"), p.Oid, p.Name)
 	}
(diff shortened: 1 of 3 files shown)


chrisd8088:
Hi @redsun82!  Thanks again for all these PRs!  I am starting to work my way through them and thought I'd begin here.  I just wanted to let you know I haven't forgotten about these PRs and that we really appreciate all your hard work (and patience!)

t/t-fetch.sh
```diff
@@ -63,11 +63,18 @@ begin_test "fetch"
...
+  refute_local_object "$contents_oid"
+
   git lfs fetch
   assert_local_object "$contents_oid" 1
 
   git lfs fsck 2>&1 | tee fsck.log
   grep "Git LFS fsck OK" fsck.log
+
+  git lfs fetch --dry-run 2>&1 | tee fetch.log
+  ! grep "fetch .* => a\.dat" fetch.log
```
chrisd8088, line 77: The `!` negation here does work as expected, because this happens to be the last command in the test.  However, it's a tricky situation because any command that we add in the future which follows this one will cause the check to become one which passes in all cases, even if the `grep` finds the pattern in the file and succeeds, meaning the `!` should result in a failure exit code from the pipeline.

That would happen because we have the `set -e` option set, and as Bash's [documentation](https://www.gnu.org/software/bash/manual/bash.html#index-set) states, that option has the following behaviour:

> Exit immediately if a pipeline ... returns a non-zero status.
> The shell does not exit if the command that fails is part of the command list immediately following a `while` or `until` keyword, part of the test in an `if` statement, part of any command executed in a `&&` or `||` list except the command following the final `&&` or `||`, any command in a pipeline but the last, or if the command’s return status is being inverted with `!`.

Over time we inadvertently accumulated a number of these silently non-functional checks due to the use of negated pipelines, which we resolved in commit a5d20de4a8de9660720c79858ae57e40cffa1def of PR #5282.

In order to try to avoid that problem in the future, I'd suggest we change this check and the one in the test below to use a different idiom:
```suggestion
  grep "fetch .* => a\.dat" fetch.log && exit 1
  true
```

t/t-fetch.sh
```diff
@@ -128,12 +135,21 @@ begin_test "fetch with remote and branches"
...
+
   git lfs fetch origin main newbranch
   assert_local_object "$contents_oid" 1
   assert_local_object "$b_oid" 1
 
   git lfs fsck 2>&1 | tee fsck.log
   grep "Git LFS fsck OK" fsck.log
+
+  git lfs fetch origin main newbranch --dry-run | tee fetch.log
+  ! grep "fetch .* => [ab]\.dat" fetch.log
```
chrisd8088, line 152: As above, I would suggest we change this to future-proof the test a bit.
```suggestion
  grep "fetch .* => [ab]\.dat" fetch.log && exit 1
  true
```

chrisd8088 (approved):
This PR looks great, thank you so much!  We also really appreciate the removal of unnecessary, stale code; that's a very pleasant bonus.

I had a few small suggestions before we merge this PR, but other than those minor details, I think this is all set.  Thank you again!

chrisd8088:
Intriguingly, we're getting a flaky test result from the modified `fetch with remote and branches` test in the `t/t-fetch.sh` script, where in one of the Linux package builds the `fetch.log` appears to be:
```
#     fetch: Fetching reference refs/heads/main
#     fetch ca978112ca1bbdcafac231b39a23dc4da786eff8147c4e72b9807785afee48bb => a.dat
#     fetch: Fetching reference refs/heads/newbranch
#     fetch ca978112ca1bbdcafac231b39a23dc4da786eff8147c4e72b9807785afee48bb => a.dat
```

The lack of the expected output line for `b.dat` means the test fails.  I'm not sure what's causing that difference from all the other CI runs, but we'll have to figure that out and try to get the output and the test to have consistent results.

redsun82:
> Intriguingly, we're getting a flaky test result from the modified `fetch with remote and branches` test in the `t/t-fetch.sh` script, where in one of the Linux package builds the `fetch.log` appears to be:
> 
> ```
> #     fetch: Fetching reference refs/heads/main
> #     fetch ca978112ca1bbdcafac231b39a23dc4da786eff8147c4e72b9807785afee48bb => a.dat
> #     fetch: Fetching reference refs/heads/newbranch
> #     fetch ca978112ca1bbdcafac231b39a23dc4da786eff8147c4e72b9807785afee48bb => a.dat
> ```
> 
> The lack of the expected output line for `b.dat` means the test fails. I'm not sure what's causing that difference from all the other CI runs, but we'll have to figure that out and try to get the output and the test to have consistent results.

Hmm, that's indeed odd. I couldn't reproduce locally... Yeah, needs some more looking into.

redsun82:
@chrisd8088 oh silly me, I had added the printing coroutine without waiting for it to finish. Should be fixed now!

t/t-fetch.sh
chrisd8088: > I actually though that a `!` command would count as a single command inverting the exit code as far as !-e! was concerned

I think pretty much everyone who's worked on this project, including myself, has had the same expectation.  And it's sneaky because a command with `!`, if it's the last one in the test, it all appears to work as you'd expect: the test fails if the command succeeds and `!` inverts the exit code.  But that's only because the inverted exit code from that last command is what's returned from the test; `-e` doesn't get involved in this special case.

I like the `command && exit 1 || true` technique you've used.  It's a little tidier than the two-line [idiom](https://github.com/git-lfs/git-lfs/blob/19a879b885a762c9d00895a49edd84af5b7b345b/t/t-pre-push.sh#L1219-L1220) we used in a number of other places, and I don't think it will cause any problems if someone later adds subsequent commands to the test and forgets to remove the `|| true` from the preceding command list.

chrisd8088:
> I had added the printing coroutine without waiting for it to finish.

Indeed, I just discovered that same issue.  Adding a `time.Sleep(5 * 100 * time.Second)` into the original version of `printTransfers()` after each `Print()` call makes the problem reproducible on a faster CPU, and it explains why we only saw the flake on some of our slowest CI jobs where we're running an old Linux version in a Docker container.

chrisd8088 (commented):
Thanks for tackling the flaky test problem!

I think we'll have to make a few more changes, as I've suggested, to avoid the reverse situation where the command never finishes in the non-dry-run case.  But with those additional changes, I _think_ we'll be all set.

I'm happy to push a commit to your branch, if you grant me permission; otherwise I'll leave this in your hands.  And thank you again very much for working on these PRs!

commands/command_fetch.go
```diff
@@ -323,44 +325,26 @@ func scanAll() []*lfs.WrappedPointer {
...
-				}
-				for _, p := range plist {
-					out <- p
-				}
-			}
-			close(out)
-		}()
+	if fetchDryRunArg {
+		wg.Add(1)
+		go printTransfers(wg, q.Watch())
```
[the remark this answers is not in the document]
redsun82: Ah, I was kind of expecting `wg.Wait()` to exit immediately in case `wg.Add` was not called. I've gone ahead with your suggestions, thanks!

redsun82:
@chrisd8088 I've implemented your suggestions, but also went ahead and gave you access to my fork, so feel free to add any commits! Thanks for the feedback, it's good to brush up on some Go knowledge :slightly_smiling_face:

commands/command_fetch.go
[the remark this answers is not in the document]
chrisd8088: Yeah, I've been caught short by exactly that same assumption myself.  And the need to pass `sync.WaitGroup` only as a pointer has also tripped me up in the past, including while I was hacking on things last night.  :-)

chrisd8088 (approved):
This is looking really good!

Thank you for the invitation to collaborate; I've pushed up what I hope is one final change to deal with an edge case where the queue's `Wait()` method is called before the anonymous goroutine has a chance to register a new `watcher` channel.  Otherwise, I found that inserting a `time.Sleep(1 * time.Second)` at the start of the goroutine was enough to make the command stall indefinitely.

chrisd8088:
OK, I think we're all good here ... except that, sigh, our Windows CI job is failing for an unrelated reason.  I'll work on getting that fixed ASAP, and then we can re-run the CI jobs here and merge this.

PR #5974, opened 2025-01-21
Add `--json` option to `fetch`

redsun82:
This adds a `--json` option to fetch, which is done on top of https://github.com/git-lfs/git-lfs/pull/5973.

It turned out there was a minor bug in `--dry-run`, where the same object was being reported fetched multiple times from different references, which diverged from non-dry-run behaviour, where the object being fetched for one ref prevented refetching for the other. This has been fixed by keeping a set of virtually fetched OIDs. This and gathering the transfers across references required sharing a `fetchWatcher` object across `fetch*` calls.

Additionally, this PR also harmonizes the `--json` flag occurring in different commands:
* all now accept `-j` as a shorcut (previously only `status` did)
* all now have the same documentation

### DIFF
--- a/commands/command_fetch.go
+++ b/commands/command_fetch.go
@@ -25,16 +25,13 @@ var (
 	fetchJsonArg   bool
 )
 
-type FetchWatcher struct {
+type fetchWatcher struct {
 	transfers        []*tq.Transfer
 	virtuallyFetched map[string]bool
 }
 
-func newFetchWatcher() *FetchWatcher {
-	ret := &FetchWatcher{
-		transfers:        nil,
-		virtuallyFetched: nil,
-	}
+func newfetchWatcher() *fetchWatcher {
+	ret := &fetchWatcher{}
 	if fetchJsonArg {
 		ret.transfers = make([]*tq.Transfer, 0)
 	}
@@ -44,7 +41,7 @@ func newFetchWatcher() *FetchWatcher {
 	return ret
 }
 
-func (d *FetchWatcher) registerTransfer(t *tq.Transfer) {
+func (d *fetchWatcher) registerTransfer(t *tq.Transfer) {
 	if d.transfers != nil {
 		d.transfers = append(d.transfers, t)
 	}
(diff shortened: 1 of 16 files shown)


commands/command_fetch.go
```diff
@@ -21,8 +22,65 @@ var (
...
+}
+
+func (d *FetchWatcher) registerTransfer(t *tq.Transfer) {
+	if d.transfers != nil {
+		d.transfers = append(d.transfers, t)
+	}
+	if d.virtuallyFetched != nil {
+		d.virtuallyFetched[t.Oid] = true
+	}
+	if fetchDryRunArg {
```
chrisd8088, line 54: I hemmed and hawed a bit over this, because `append()` will accept a `nil` slice and just allocate a backing array, so we could drop the `make([]*tq.Transfer, 0)` in the `newFetchWatcher()` function above, and then do something like:
```go
	if fetchJsonArg {
		d.transfers = append(d.transfers, t)
	}
	if fetchDryRunArg {
		d.virtuallyFetched[t.Oid] = true
		printProgress("%s %s => %s", tr.Tr.Get("fetch"), t.Oid, t.Name)
	}
```

However, since in your next PR #5975 you're adding another flag whose state affects both functions, I think it's probably more elegant to leave it as you've written it.

commands/command_fetch.go
```diff
@@ -21,8 +22,65 @@ var (
 	fetchAllArg    bool
 	fetchPruneArg  bool
 	fetchDryRunArg bool
+	fetchJsonArg   bool
 )
 
+type FetchWatcher struct {
```
chrisd8088, line 28: As we don't need to export this structure outside of the `commands` package, we can just call it `fetchWatcher`.

chrisd8088 (approved):
This looks awesome.  Thank you so much!

I've left a few comments and suggestions, but nothing major, I think.  If it's easier for you, I can push a commit or two with those minor revisions and then merge the PR; just let me know what you prefer.

I should also say, thanks very much for straightening up all the `--json` and `-j` options and their documentation!

commands/command_fetch.go
```diff
@@ -398,5 +470,6 @@ func init() {
 		cmd.Flags().BoolVarP(&fetchAllArg, "all", "a", false, "Fetch all LFS files ever referenced")
 		cmd.Flags().BoolVarP(&fetchPruneArg, "prune", "p", false, "After fetching, prune old data")
 		cmd.Flags().BoolVarP(&fetchDryRunArg, "dry-run", "d", false, "Do not fetch, only show what would be fetched")
+		cmd.Flags().BoolVar(&fetchJsonArg, "json", false, "Give the output in a stable JSON format for scripts")
```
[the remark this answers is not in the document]
redsun82: ha, I was so caught up adding that in the existing commands I totally forgot it here, good catch, thanks!

commands/command_fetch.go
redsun82: Fair point, I'll mull it over when I get back to that PR, maybe I can indeed move choosing the behaviour from the constructor to this function. For now, I left it like it was.

chrisd8088 (approved):
Fantastic, thank you!

PR #5975, opened 2025-01-21
Add `--force` option to `fetch`

redsun82:
This adds a `--force` option that makes `fetch` refetch objects that are already present locally. This can be useful especially in combination with the `--dry-run` option to see what is fetchable regardless of the current local status, or with `--json` from https://github.com/git-lfs/git-lfs/pull/5974 to get download information.


chrisd8088:
I usually avoid commenting on draft PRs, since I know folks don't always appreciate comments on work-in-progress, so I'll keep this brief!

I just happened to notice that in the final commit here, which builds on the work from PR #5973, there's a mention in one comment of an option named `--refetch` rather than `--force`, and that prompted me to go re-read the Git documentation and realize that you're really onto something there.

The `git fetch` command's `--refetch` [option](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt---refetch) is actually the one whose functionality is closest to what this PR implements, so let's call the new option `--refetch` rather than `--force`.  That's a much better idea, since it keeps the two commands in closer alignment than my proposal.

Apologies for steering you in the wrong direction there, and thanks for spotting the inconsistency!

redsun82:
> I usually avoid commenting on draft PRs, since I know folks don't always appreciate comments on work-in-progress, so I'll keep this brief!
> 
> I just happened to notice that in the final commit here, which builds on the work from PR #5973, there's a mention in one comment of an option named `--refetch` rather than `--force`, and that prompted me to go re-read the Git documentation and realize that you're really onto something there.
> 
> The `git fetch` command's `--refetch` [option](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt---refetch) is actually the one whose functionality is closest to what this PR implements, so let's call the new option `--refetch` rather than `--force`. That's a much better idea, since it keeps the two commands in closer alignment than my proposal.
> 
> Apologies for steering you in the wrong direction there, and thanks for spotting the inconsistency!

ah, thanks a lot for this comment! Yeah, I think I hesitated between `--force` and `--refetch` and that comment captured that. Once I'll have some more time to get back to this I'll go for `--refetch` then, I agree with you that it's better!

t/t-fetch.sh
```diff
@@ -118,6 +118,27 @@ EOF
...
+  git lfs fetch --refetch
+  assert_local_object "$contents_oid" 1
+
+  git lfs fsck 2>&1 | tee fsck.log
+  grep "Git LFS fsck OK" fsck.log
+)
+end_test
+
+
 begin_test "fetch (empty file)"
```
chrisd8088, line 142: While experimenting with the new tests in this PR, I discovered (as you may have already!) that the new `fetch --refetch with remote and branches` test you've added validates the behaviour of the `--refetch` option with empty files, namely that it correctly ignores them.

I was a bit surprised to realize, though, that the reason this occurs is solely due to the ordering of these tests.  If the new `p.Size == 0` conditional is removed from the `pointers()`/`pointersToFetch()` function in the Go code, the test fails, as we would expect ... but only if the test is located after the existing `fetch (empty file)` test.  That's because the `fetch (empty file)` test [adds](https://github.com/git-lfs/git-lfs/blob/aa0ca91e2287d4b7b9e060e7141a61531290113e/t/t-fetch.sh#L127-L129) an empty file to the Git repository which is shared by many of the tests in the `t/t-fetch.sh` script.

(Reusing a test repository between tests is not really a practice we follow any more, but the origins of the current `t/t-fetch.sh` script go back a long way.)

If we were ever to reorder the tests, we might silently lose the validation of the `--refetch` option's behaviour with regard to empty files; the test would continue to pass, even if we subsequently changed the implementation of the command such that we accidentally disabled the code which skips empty files when re-fetching.

That's a bit unlikely, I admit, but to be on the safe side, would you be willing to add a comment here along the lines of the following?
```suggestion
# NOTE: Do not reorder this test below the subsequent tests, as they depend on
#       the empty file it creates in the shared test repository.
begin_test "fetch (empty file)"
```

chrisd8088 (approved):
Wow, thank you again for all the work on these PRs!  This one also looks great.

I had a few suggestions here and there, but nothing at all significant.

Thank you again so much for writing up a proposal in #5888 and seeing it through to completion.  We sincerely appreciate the help in developing new features for Git LFS!

t/t-fetch.sh
redsun82: I went ahead and did that in its own commit, I can revert that if you don't think it's a good idea :slightly_smiling_face:

chrisd8088: Cool, that works too and means we test the empty file case in all the interdependent tests, which is great.

PR #5977, opened 2025-01-25
Update workflows to use ARM runners and new Apple signing certificate

chrisd8088:
This PR updates our GitHub Actions workflows so they avoid several permissions issues, make use of the newly available ARM runners, and sign our macOS release binaries with a new Apple Developer ID certificate that replaces the previous one, which has now expired.

First, we revise our CI and release workflows to avoid a problem introduced when we changed the `actions/checkout` actions in PR #5930 to no longer leave Git credentials in place for subsequent steps to use.  While this had no effect on CI jobs run for PRs, it does affect CI and release jobs run when we push a new Git LFS version tag, because the `git fetch` commands we added in PR #5243 to work around the problem described in actions/checkout#290 and actions/checkout#882 now fail due to the lack of cached Git credentials.

Next, because we would like to run our Actions workflows with the minimal required permissions, we alter the `build-main` job in our release workflow so that is does not try to create a draft release announcement or attach binary release assets to that announcement.  At present, this step is the only one in our workflows which requires write permissions on our whole repository's contents, and GitHub does not offer a way to make those permissions more fine-grained (e.g., by limiting them to just the creation of release announcements, for instance).  Instead, in the future we will just run our `script/upload` script manually, which allows us to restrict all our workflows to having just read permissions on our repository contents.

We then change our release workflow to replace our expired Apple Developer ID certificate with a new one that is provided by GitHub.  Because Apple [limits](https://developer.apple.com/help/account/create-certificates/create-developer-id-certificates) the number of certificates which can be created for a Developer ID to just five, GitHub is no longer able to assign a dedicated certificate to this project.  In order to share a certificate with other projects, we adopt a GitHub management tool which allows us to approve our release workflow and confirm the use of the new certificate's secrets to sign our release binaries.

Finally, we revise our Actions workflows to make use of the ARM64 runners GitHub now [offers](https://github.com/orgs/community/discussions/148648) for public repositories like ours.  This allows us to bypass some recent issues seen in our Linux package builds that use QEMU-based ARM64 emulation, and because these jobs now run significantly faster than before, we can add a build of a Debian 11 release package for ARM64 systems.  (We should also be able to offer a Rocky 9 RPM package for ARM64 systems, once we make some adjustments to our RPM package metadata and Docker containers.)

The changes to our release workflow in this PR have been tested using a private repository to confirm they work as expected when a new version tag is pushed.  One exception is the use of ARM64 runners in the release workflow, as GitHub does not yet support ARM64 runners in private repositories without an Enterprise account, but we have tested these build jobs in our CI workflow in public repositories to confirm they succeed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,11 +16,11 @@ jobs:
       with:
         fetch-depth: 0
         persist-credentials: false
-    - run: git fetch origin "+${GITHUB_REF}:${GITHUB_REF}"
-      if: ${{ github.ref_type == 'tag' }}
-      # We update the current tag as the checkout step turns annotated tags
-      # into lightweight ones by accident, breaking "git describe".
-      # See https://github.com/actions/checkout/issues/882 for details.
+        ref: ${{ github.ref }}
+      # We specify the current ref because otherwise the checkout turns
+      # annotated tags into lightweight ones, breaking "git describe".
+      # See https://github.com/actions/checkout/issues/290 and
+      # https://github.com/actions/checkout/issues/882 for details.
     - uses: ruby/setup-ruby@v1
     - run: gem install asciidoctor
     - uses: actions/setup-go@v5
@@ -53,8 +53,7 @@ jobs:
       with:
         fetch-depth: 0
         persist-credentials: false
-    - run: git fetch origin "+${GITHUB_REF}:${GITHUB_REF}"
-      if: ${{ github.ref_type == 'tag' }}
+        ref: ${{ github.ref }}
     - uses: actions/setup-go@v5
       with:
         go-version: ${{ matrix.go }}
@@ -67,9 +66,7 @@ jobs:
       with:
         fetch-depth: 0
         persist-credentials: false
-    - run: git fetch origin "+${GITHUB_REF}:${GITHUB_REF}"
-      if: ${{ github.ref_type == 'tag' }}
-      shell: bash
+        ref: ${{ github.ref }}
     - uses: ruby/setup-ruby@v1
     - run: gem install asciidoctor
     - run: Rename-Item -Path C:\msys64 -NewName msys64-tmp -Force
@@ -136,11 +133,10 @@ jobs:
       with:
         fetch-depth: 0
         persist-credentials: false
+        ref: ${{ github.ref }}
     - uses: actions/setup-go@v5
       with:
         go-version: '1.23.x'
-    - run: git fetch origin "+${GITHUB_REF}:${GITHUB_REF}"
-      if: ${{ github.ref_type == 'tag' }}
     - run: git clone -b master https://github.com/git/git.git "$HOME/git"
     - run: |
         echo "GIT_INSTALL_DIR=$HOME/git" >> "$GITHUB_ENV"
@@ -161,11 +157,10 @@ jobs:
       with:
         fetch-depth: 0
         persist-credentials: false
+        ref: ${{ github.ref }}
     - uses: actions/setup-go@v5
       with:
         go-version: '1.23.x'
-    - run: git fetch origin "+${GITHUB_REF}:${GITHUB_REF}"
-      if: ${{ github.ref_type == 'tag' }}
     - run: git clone -b v2.0.0 https://github.com/git/git.git "$HOME/git"
     - run: |
         echo "GIT_INSTALL_DIR=$HOME/git" >> "$GITHUB_ENV"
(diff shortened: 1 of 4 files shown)


docs/howto/release-git-lfs.md
```diff
@@ -213,13 +213,26 @@ to zero, we are releasing a PATCH version.
...
 
-  6. From the command line, finalize the release process by signing the release:
+     The release assets should now be present in the `bin/releases` directory.
+     The `script/upload` utility will create a new draft release announcement
+     and then upload the release assets and attach them to the announcement:
 
-     ```ShellSession
-     $ script/upload --finalize vM.N.P
+     ```
+     $ script/upload --skip-verify vM.N.P
```
larsxschneider, line 238: I think you skip verify here because this step would also try to download the asset which we already have downloaded manually. However, wouldn't it make sense to verify the signatures of the files that we have locally before we upload them or is that already checked elsewhere?

larsxschneider (approved):
LGTM! Thank you 🙇

docs/howto/release-git-lfs.md
chrisd8088: The "verify" step in this script downloads the assets attached to the release announcement page and then verifies that the GPG-signed `hashes.asc` and `sha256sums.asc` are valid and that the hashes in those files match the hashes of the binary assets.  So that's a different operation from downloading the assets from the summary page of the release Actions job, which we'll now do manually, at least for the time being.

When we run `script/upload --skip-verify` without the `--finalize` option, that will [create](https://github.com/git-lfs/git-lfs/blob/a8bf08074988651b19fbb1f003e3fb6691f78aa0/script/upload#L460) the draft release announcement page, and then [upload](https://github.com/git-lfs/git-lfs/blob/a8bf08074988651b19fbb1f003e3fb6691f78aa0/script/upload#L463) the binary assets as attachments to that page.  The `hashes` and `sha256sums` files have not been created yet.

When we run `script/upload --finalize`, that will first [download](https://github.com/git-lfs/git-lfs/blob/a8bf08074988651b19fbb1f003e3fb6691f78aa0/script/upload#L337) the binary assets attached the draft release announcement (as created by `script/upload --skip-veify`), [create](https://github.com/git-lfs/git-lfs/blob/a8bf08074988651b19fbb1f003e3fb6691f78aa0/script/upload#L347-L355) the `hashes.asc` and `sha256sum.asc` files for the first time, and then [upload](https://github.com/git-lfs/git-lfs/blob/a8bf08074988651b19fbb1f003e3fb6691f78aa0/script/upload#L367-L368) those and attach them to the announcement page.  Finally, because we don't use `--skip-verify` in this version of the command, it will end by [verifying](https://github.com/git-lfs/git-lfs/blob/a8bf08074988651b19fbb1f003e3fb6691f78aa0/script/upload#L469), meaning it will download the binary assets and the hash files from the draft release announcement and check that they match.

Lastly, we manually click the "Publish" button on the announcement page, which makes the release public.

.github/workflows/ci.yml
```diff
@@ -16,11 +16,11 @@ jobs:
...
-      if: ${{ github.ref_type == 'tag' }}
-      # We update the current tag as the checkout step turns annotated tags
-      # into lightweight ones by accident, breaking "git describe".
-      # See https://github.com/actions/checkout/issues/882 for details.
+        ref: ${{ github.ref }}
+      # We specify the current ref because otherwise the checkout turns
+      # annotated tags into lightweight ones, breaking "git describe".
+      # See https://github.com/actions/checkout/issues/290 and
+      # https://github.com/actions/checkout/issues/882 for details.
     - uses: ruby/setup-ruby@v1
```
bondtomasznowak-stack, line 24: d----            9/7/2026 10:55 PM                compat_dbs

.github/workflows/ci.yml
```diff
@@ -16,11 +16,11 @@ jobs:
...
-    - run: git fetch origin "+${GITHUB_REF}:${GITHUB_REF}"
-      if: ${{ github.ref_type == 'tag' }}
-      # We update the current tag as the checkout step turns annotated tags
-      # into lightweight ones by accident, breaking "git describe".
-      # See https://github.com/actions/checkout/issues/882 for details.
+        ref: ${{ github.ref }}
+      # We specify the current ref because otherwise the checkout turns
+      # annotated tags into lightweight ones, breaking "git describe".
+      # See https://github.com/actions/checkout/issues/290 and
+      # https://github.com/actions/checkout/issues/882 for details.
```
bondtomasznowak-stack, line 23: 23


A

PR #5983, opened 2025-02-06
Update Windows CI jobs to find Perl `prove` command

chrisd8088:
As described in commit 8d456a76727bbadf690b0093e1343331bff9ac23 of PR #5879, we currently depend on the Git for Windows SDK to install a version of the `prove(1)` command from the Perl `Test::Harness` distribution, because we use that command to run our suite of shell tests.

However, in our CI jobs on Windows, the SDK does not at present install the `prove` command into a location listed in the `PATH` environment variable, so these CI jobs having begun failing, as seen in this job [run](https://github.com/git-lfs/git-lfs/actions/runs/13166188381/job/36746979899) for PR #5973.

This change may be related to the recent migration of the SDK's release artifacts to GitHub from Azure Blobs, as outlined in PR git-for-windows/git-for-windows-automation#109, and is possibly a consequence of PR git-for-windows/build-extra#588.

Regardless, the `prove` command is fortunately still included in the `minimal` flavour of the SDK, since it is [listed](https://github.com/git-for-windows/git-sdk-64/blob/f845bb725e56e058a0fbf93aba18c70906a1068e/.sparse/minimal-sdk#L188) in the manifest for that flavour. We simply need to ensure that the directory in which the command is now located, namely `/usr/bin/core_perl`, is added to the set of paths in the `PATH` variable for the Git Bash environment.

Note that we only need to add this extra path to `PATH` when we are running our `script/cibuild` script, as that [runs](https://github.com/git-lfs/git-lfs/blob/19a879b885a762c9d00895a49edd84af5b7b345b/script/cibuild#L32) the default recipe from our `t/Makefile` file, which in turn [executes](https://github.com/git-lfs/git-lfs/blob/19a879b885a762c9d00895a49edd84af5b7b345b/t/Makefile#L43) the `test` recipe, which [uses](https://github.com/git-lfs/git-lfs/blob/19a879b885a762c9d00895a49edd84af5b7b345b/t/Makefile#L50) the `prove` command to run and summarize all our shell test scripts.

/cc @dscho as a Git for Windows maintainer who may have additional recommendations

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,10 +88,12 @@ jobs:
         flavor: minimal
       # We install the SDK so as to have access to the msgfmt.exe binary
       # from the GNU gettext package.
-    - run: GOPATH="$HOME/go" PATH="$HOME/go/bin:$PATH" env -u TMPDIR script/cibuild
+    - run: GOPATH="$HOME/go" PATH="$HOME/go/bin:/usr/bin/core_perl:$PATH" env -u TMPDIR script/cibuild
       shell: bash
       # We clear the TMPDIR set for Ruby so mktemp and Go use the same
       # volume for temporary files.
+      # We include the /usr/bin/core_perl path in PATH so the Perl prove
+      # command provided by the SDK can be located.
     - run: rm -f commands/mancontent_gen.go
       shell: bash
     - run: GOPATH="$HOME/go" PATH="$HOME/go/bin:$PATH" make GOARCH=386 -B


dscho:
> in our CI jobs on Windows, the SDK does not at present install the `prove` command into a location listed in the `PATH` environment variable

That is strange, as the `setup-git-for-windows-sdk` Action [specifically adds `core_perl` so that `prove` is found](https://github.com/git-for-windows/setup-git-for-windows-sdk/blob/599f62ed8ee1dfe40055aae062183f28cd57563c/main.ts#L110-L119).

And indeed, [in my test `prove` is found on the `PATH`](https://github.com/dscho/setup-git-for-windows-sdk/actions/runs/13174434621/job/36770547281#step:3:11). The workflow definition reads like this:

```yaml
name: 'prove-test'

on:
  push:

jobs:
  build:
    runs-on: windows-latest
    steps:
      - uses: git-for-windows/setup-git-for-windows-sdk@v1
        with:
          flavor: minimal
      - shell: bash
        run: |
          set -x
          type prove
          echo $PATH
          prove --version
```

> This change may be related to the recent migration of the SDK's release artifacts to GitHub from Azure Blobs, as outlined in PR [git-for-windows/git-for-windows-automation#109](https://github.com/git-for-windows/git-for-windows-automation/pull/109), and is possibly a consequence of PR [git-for-windows/build-extra#588](https://github.com/git-for-windows/build-extra/pull/588).

Neither of these changes is concerned with `setup-git-for-windows-sdk` nor `ci-artifacts` nor do they change the SDKs, therefore I would be really surprised if they had any effect here.

I had a quick look whether `/etc/profile` might potentially play a role here, but I do not think so, its contents in the minimal Git for Windows SDK are:

```sh
export MSYSTEM=MINGW64
export PATH=/mingw64/bin:/usr/bin/:/usr/bin/core_perl:/c/WINDOWS/system32:/c/WINDOWS:/c/WINDOWS/System32/Wbem
```

Crucially, it includes `core_perl`!

What _is_ strange, though, is that there is a `Path` (not `PATH`!) listed [in the `env` section of your failed run](https://github.com/git-lfs/git-lfs/actions/runs/13166188381/job/36746979899#step:13:10), and it does _not_ include `D:\minimal-sdk\usr\bin`. It _does_ include `C:\mingw64\bin` though, so you may not even run the intended Bash at all in your workflow.

How about inserting `set -x; env; type bash; cygpath -aw /;cat /etc/profile;` in the `run:` clause of the failing step to diagnose?

chrisd8088:
Thanks for taking a look, @dscho, and all the tips!

Intriguingly, whatever the problem is, it's not consistent right now, so I was able to get a few successful CI jobs through the queue.

The issue doesn't seem to be related to the specific Actions Windows runner image version, as I've seen failing and successful runs which both had the same image version (`20250127.1` or now `20250203.1.0`, both `windows-2022` images), and even ran the exact same workflows.  In one case, just re-running the workflow a few times led to both successful and failing outcomes.

As a sidebar, the weird `Path` environment variable you noticed is set by the `ruby/setup-ruby` action, which we used to require in order to install some Ruby gems and run several Ruby scripts.  (I left a lot of documentation about this in commit b7fa3a54e4d65e5f000b30e9df41cddf1daeed41 and PR #5236 back in 2022.)  Fortunately, your question about that led me to experiment a bit with removing that action, in case we could now use the Ruby that's already installed on the latest versions of the various GitHub Actions runners, and it turns out we can!  So I've written up PR #5984, which yanks all that stuff out.

Alas ... this apparently doesn't resolve the flakiness we're seeing.  As it turns out, it's not just the `prove` executable but also the `msgfmt` one which are sometimes in the `PATH` and sometimes not.

I tried running versions of our CI workflow in a private repository, with the same changes from PR #5984 (i.e., no `ruby/setup-ruby` action), and saw it both succeed and fail on Windows in different runs.  I added a step right after the `git-for-windows/setup-git-for-windows-sdk` action which output the contents of `PATH` and the contents of `/etc/profile`, and also ran `command -v` to look for `bash`, `msgfmt`, and `prove` executables.

In the case of a successful run, the `/etc/profile` contents are just:
```
export MSYSTEM=MINGW64
export PATH=/mingw64/bin:/usr/bin/:/usr/bin/core_perl:/c/WINDOWS/system32:/c/WINDOWS:/c/WINDOWS/System32/Wbem
```

and the three binaries are found as:
```
/usr/bin/bash
/mingw64/bin/msgfmt
/usr/bin/core_perl/prove
```

The `PATH` variable contains a long list of paths, but it starts with all the stuff one might expect on a Windows runner:
```
/mingw64/bin:/usr/bin:/usr/bin/core_perl:/c/Users/runneradmin/go/bin:/c/hostedtoolcache/windows/go/1.23.5/x64/bin:/c/Program Files/MongoDB/Server/5.0/bin:...
```

By contrast, in a failing run, the `/etc/profile` contents are substantial and look to be the default MSYS2 version of this file.  The only binary found is:
```
/usr/bin/bash
```

And the `PATH` contents start with:
```
/mingw64/bin:/usr/bin:/c/Users/runneradmin/bin:/d/git-sdk-64-minimal/mingw64/bin:/d/git-sdk-64-minimal/usr/bin:/d/git-sdk-64-minimal/usr/bin/core_perl:/c/Users/runneradmin/go/bin:/c/hostedtoolcache/windows/go/1.23.5/x64/bin:/c/Program Files/MongoDB/Server/5.0/bin:...
```

For the time being, I'll just keep this PR in reverse, and if we see a flake like I've described in a public Actions job run I'll add a link to it here.

dscho:
> By contrast, in a failing run, the `/etc/profile` contents are substantial and look to be the default MSYS2 version of this file.

Since you run these from a Bash step, it might actually hide what is actually going on: the hosted `windows-latest` runners have a full MSYS2 at `C:\msys2` and if, for whatever reason, `C:\msys2\usr\bin` is in the `PATH`, you will not run the minimal SDK but MSYS2 instead _and_ `echo $PATH` would still show `/mingw64/bin:/usr/bin/:...`.

Read: the symptom suggests to me that in the failing case, you are somehow running the Bash step using the regular MSYS2 Bash instead of the minimal SDK's Bash.

This is also consistent with the `PATH` you showed for the failing case:

> ```
> /mingw64/bin:/usr/bin:/c/Users/runneradmin/bin:/d/git-sdk-64-minimal/mingw64/bin:/d/git-sdk-64-minimal/usr/bin:/d/git-sdk-64-minimal/usr/bin/core_perl:[...]
> ```

Notice the `/d/git-sdk-64-minimal/mingw64/bin:` part? That's supposed to come at the beginning, but obviously isn't (the MSYS2 path translation would reduce this path to `/mingw64/bin/` if running in the minimal SDK, therefore we _know_ we're not running in the minimal SDK).

Now the big question is: What _on earth_ is changing the `PATH` between the `setup-git-for-windows-sdk` step and the `shell: bash`/`run` step?

dscho:
> you are somehow running the Bash step using the regular MSYS2 Bash instead of the minimal SDK's Bash.

I was wrong, it's not the regular MSYS2 Bash, [it's _regular Git for Windows'_ Bash](https://github.com/chrisd8088/git-lfs/actions/runs/13166461822/job/36747702223#step:13:3):

> ```
>   shell: C:\Program Files\Git\bin\bash.EXE --noprofile --norc -e -o pipefail {0}
> ```

When everything is working as it should, we see [this](https://github.com/chrisd8088/git-lfs/actions/runs/13169935566/job/36758281838#step:13:3) instead:

> ```
>   shell: D:/git-sdk-64-minimal/usr/bin\bash.EXE --noprofile --norc -e -o pipefail {0}
> ```

Puzzling. I have no idea why [`setup-git-for-windows-sdk`'s `PATH` munging](https://github.com/git-for-windows/setup-git-for-windows-sdk/blob/599f62ed8ee1dfe40055aae062183f28cd57563c/main.ts#L110-L119) seems to work in some cases, and in other cases it does not.

May be worth investigating by trimming the workflow down so that it not only skips non-Windows jobs, but also replacing the `script/cibuild` step (and all subsequent steps) by a `shell: bash` step that succeeds only if `cygpath -am /` outputs `/d/git-sdk-64-minimal/`, then run that workflow as often as you need to see whether it is flaky. If it is, continue by ripping out random steps that precede the `setup-git-for-windows-sdk` step, to see whether they perhaps interfere.

If this all takes too much time to investigate and you need something that simply works, you could add an `id: minimal-sdk` to the `setup-git-for-windows-sdk` step and then replace [the `shell: bash` line](https://github.com/git-lfs/git-lfs/blob/main/.github/workflows/ci.yml#L92) of the `GOPATH="$HOME/go" PATH="$HOME/go/bin:$PATH" env -u TMPDIR script/cibuild` step by `shell: ${{ steps.minimal-sdk.outputs.result }}\usr\bin\bash.exe --noprofile --norc -e -o pipefail {0}`. This would force it to use the minimal SDK always even if _something_ intercepts and modifies the `PATH` (although you then may need to really make sure that the `PATH` does not start with `/c/Program Files/Git/mingw64/bin:`...).

dscho:
By the way, a more robust solution might be to move off of Git for Windows' SDK altogether, as I cannot guarantee that `prove` will always be shipped with the minimal SDK. I know I helped with https://github.com/git-lfs/git-lfs/pull/5666 but I should have thought harder back then.

What I have in mind is to use regular MSYS2 instead, via `shell: C:\msys64\usr\bin\bash.exe --noprofile --norc -e -o pipefail {0}`. 
The `run:` value would then _need_ to start with `. /etc/profile`, to configure the `PATH` correctly, and then you'd have to install `perl` (which includes `prove`) and either `gettext` or `mingw-w64-x86_64-gettext-tools` (for `msgfmt.exe`). (`windows-2019` [_may_ have installed this via `base-devel`](https://github.com/actions/runner-images/blob/650fb260cb8ba21ccccaff47470d3f43c94fbb3e/images/windows/toolsets/toolset-2019.json#L157-L231), but starting [with `windows-2022`](https://github.com/actions/runner-images/blob/650fb260cb8ba21ccccaff47470d3f43c94fbb3e/images/windows/toolsets/toolset-2022.json#L161-L164) only a bare-bones MSYS2 is installed on the hosted runners by default), via `pacman -S --noconfirm perl gettext`.

BTW the runner image documentation [suggests](https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#shells) that `msys2bash.cmd` should do the same, but I have no idea whether that works, I only stumbled across this part of the documentation today.

chrisd8088:
Thanks again for all the input and insight, @dscho!  I had likewise come to suspect that we're toggling between different versions of Bash, based on some kind of semi-random factor, such as a side effect of when some command completes or other timing-related variable.

I think we can start just by pulling out the Ruby setup steps, and over time I'll experiment with some of your suggestions to try to narrow down either a cause or an alternative implementation which runs consistently.

PR #5986, opened 2025-02-10
Create devcontainer.json

shamyuchiha:


bk2204:
I'll definitely let the maintainers decide whether they want to accept this change, but I'll mention that Git LFS doesn't pass its tests in a GitHub Codespace (nor, for that matter, does Git), in case using that was your goal.

The reason is that GitHub Codespaces sets an ACL on at least parts of the `/workspaces` directory that makes files world-writable and overrides other permissions, which breaks various parts of the testsuite that rely on being able to set files read-only.  I've provided that feedback to them, but I don't believe there are current plans to change it at the moment, and I've been informed that removing the ACL is not a supported approach (and I know it can break some things).

Of course, this could also be used for regular development with the Devcontainers CLI on a Linux system, in which case that problem wouldn't occur.

So, the Codespace problem may affect the views of the maintainers on whether they'd like to accept this patch (or it might not, who knows).  Of course, I'll defer to their judgment here as to what's best.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,5 @@
+{
+  "image": "mcr.microsoft.com/devcontainers/universal:2",
+  "features": {
+  }
```
chrisd8088, line 4: Is this empty `features` object actually required, or could it be removed?

chrisd8088 (commented):
Hey, thanks for the PR!

Would you be able to explain what the intended use case is here, and how this change is beneficial?

I assume the goal is to improve the development experience when using a tool like VSCode or GitHub Codespaces, as you've referenced the Universal Development Container Image [published](https://hub.docker.com/r/microsoft/devcontainers-universal) by Microsoft.

Can you expand on why this container is appropriate for your intended uses, and whether there are any limitations or alternatives?

As @bk2204 pointed out, there may be some issues with GitHub Codespaces, and a development environment where it's not possible to run out test suites (both the Go and Bash shell ones) is not especially useful.

PR #5987, opened 2025-02-11
Add security patch release process documentation

chrisd8088:
The [documentation](https://github.com/git-lfs/git-lfs/blob/d0d1ed06177d78e2222ed3d9a7bb07db6effb033/docs/howto/release-git-lfs.md) of our standard and patch release processes does not, at present, include the steps required to build a security patch release, so we add two sections which describe how we build such releases.

The first new section documents our normal security patch release process, which applies when we do not need to coordinate our release with other projects, or at least do not need to build release assets in private while under a publication embargo.

The second new section details the process of building new Git LFS release binaries in private, so they can be shared with other downstream projects while we are under an embargo on the publication of the security vulnerability and associated security patch release.


docs/howto/release-git-lfs.md
```diff
@@ -337,3 +337,269 @@ and then use that branch as the base for the PATCH release.
...
+  5. Publish the security advisory.
+
+  6. Follow our standard release process, starting with the step in which
+     we create a GPG-signed tag named `vM.N.P` on the merge commit to the
+     `release-M.N` branch.
+
+  7. After publication of the new release, update the `git-lfs.com` home
+     page with a banner message regarding the security PATCH release
+     by pushing a change to the `_includes/home/secondary.html` file
+     in the `git-lfs/git-lfs.github.com` repository.
```
larsxschneider, line 396: nit: Maybe add [this link](https://github.com/git-lfs/git-lfs.github.com/commit/3aaeafd3f729e2f2d696b7dd19e718b850222d34) as example? BTW: Do we always keep the announcement or do we remove it after some time?

docs/howto/release-git-lfs.md
```diff
@@ -337,3 +337,269 @@ and then use that branch as the base for the PATCH release.
...
+When coordinating the release of a security patch with one or more
+other projects, we must not follow the processes described above as
+they depend on the GitHub Actions release workflow in our public
+`git-lfs/git-lfs` repository.  Instead, we use a private repository
+with a modified release workflow to build the binaries and packages
+for our new version so that we can share them with the other projects
+in advance of the coordinated release date.
+
+  1. Open a draft security advisory, apply for a CVE identifier, and
+     create a private fork in which to develop a resolution of the
```
larsxschneider, line 414: Is "which to develop" correct? It reads a bit odd as non-native speaker.

docs/howto/release-git-lfs.md
```diff
@@ -337,3 +337,269 @@ and then use that branch as the base for the PATCH release.
...
+  7. Create a GPG-signed tag named `vM.N.P` on the merge commit, using
+     the same command from the equivalent step of our standard release
+     process:
+
+     ```ShellSession
+     $ git tag -s vM.N.P -m vM.N.P
+     ```
+
+  8. Check out the `main` branch of the private repository and make
+     several revisions to the `.github/workflows/release.yml` file,
```
larsxschneider, line 471: `several` --> `the following` ?

larsxschneider (approved):
Thank you so much for writing these detailed instructions @chrisd8088 🙇🙇🙇!

That will be super helpful in the future!

docs/howto/release-git-lfs.md
chrisd8088: That's better, thanks!

docs/howto/release-git-lfs.md
chrisd8088: Good questions.  I've expanded this step's instructions a bit more fully and included an example of the banner message and formatting for it, which I think will help clarify things somewhat.

docs/howto/release-git-lfs.md
chrisd8088: I'm pretty sure it's correct English grammar, although perhaps a bit formal or abstruse.

I'm no expert, but from a bit of research, I believe the "in which" is a prepositional phrase since it begins with the preposition "in" and ends with the object "which", where the "which" introduces a relative clause, in this case, the full infinitive "to develop".  (Corrections to that description are welcome!)

I've rephrased the sentence, though, as three individual sentences, which I hope makes it easier to read for everyone.

Issue #5968, opened 2025-01-19
Go 1.24: non-constant format strings break test build


QuLogic:
**Describe the bug**
We are rebuilding things in Fedora for upcoming version 42 against Go 1.24~rc1. `go test` fails on git-lfs due to non-constant format strings:
```
./command_clean.go:85:9: non-constant format string in call to github.com/git-lfs/git-lfs/v3/commands.Debug
./command_dedup.go:132:31: non-constant format string in call to github.com/git-lfs/git-lfs/v3/errors.Errorf
./command_fsck.go:173:10: non-constant format string in call to github.com/git-lfs/git-lfs/v3/commands.Debug
./command_fsck.go:217:8: non-constant format string in call to github.com/git-lfs/git-lfs/v3/commands.Debug
./command_logs.go:40:8: non-constant format string in call to github.com/git-lfs/git-lfs/v3/commands.Debug
./command_logs.go:54:8: non-constant format string in call to github.com/git-lfs/git-lfs/v3/commands.Debug
./command_logs.go:55:77: non-constant format string in call to github.com/git-lfs/git-lfs/v3/errors.Wrapf
./command_migrate.go:322:29: non-constant format string in call to github.com/git-lfs/git-lfs/v3/errors.Errorf
./command_migrate_export.go:38:31: non-constant format string in call to github.com/git-lfs/git-lfs/v3/errors.Errorf
./command_migrate_export.go:119:31: non-constant format string in call to github.com/git-lfs/git-lfs/v3/errors.Errorf
./command_migrate_import.go:47:32: non-constant format string in call to github.com/git-lfs/git-lfs/v3/errors.Errorf
./command_migrate_import.go:51:32: non-constant format string in call to github.com/git-lfs/git-lfs/v3/errors.Errorf
./command_migrate_import.go:69:32: non-constant format string in call to github.com/git-lfs/git-lfs/v3/errors.Errorf
./command_migrate_import.go:76:33: non-constant format string in call to github.com/git-lfs/git-lfs/v3/errors.Errorf
./command_migrate_import.go:83:37: non-constant format string in call to github.com/git-lfs/git-lfs/v3/errors.Wrapf
./command_migrate_import.go:127:32: non-constant format string in call to github.com/git-lfs/git-lfs/v3/errors.Errorf
./command_migrate_import.go:145:32: non-constant format string in call to github.com/git-lfs/git-lfs/v3/errors.Errorf
[...]

PR #5989, opened 2025-02-12
Replace unused debug mode with trace logging

chrisd8088:
In PR #68 the `commands` package was updated with several logging functions such as `Panic()`, `Print()`, and `Debug()`.  The last of these [only outputs](https://github.com/git-lfs/git-lfs/blob/d0d1ed06177d78e2222ed3d9a7bb07db6effb033/commands/commands.go#L251-L254) a log message if the `Debugging` [variable](https://github.com/git-lfs/git-lfs/blob/d0d1ed06177d78e2222ed3d9a7bb07db6effb033/commands/commands.go#L32) in the same package is set to `true`, which at the time was controlled by the `setupDebugging()` [function](https://github.com/git-lfs/git-lfs/blob/f83d26fa39acad6618ca537286ac710c3e6d3e8d/commands/commands.go#L163-L169).  If the user supplied the `--debug` option to a given Git LFS command, `setupDebugging()` function would set the `Debugging` variable to `true`, and any calls to `Debug()` would then generate log messages.

However,  in PR #84 the `setupDebugging()` function was removed, along with a lot of the earlier command-line parsing and setup code, and replaced with the Cobra library from the `spf13/cobra` project. Since then it has not been possible to set the `Debugging` variable to a value other than `false`, so the remaining references to the `Debug()` function will never generate log messages.

We therefore remove the `Debugging` variable and the `Debug()` function, and replace any calls to that function with ones to our standard trace logging function.

This change should help reduce the number of non-constant format strings we need to revise before adopting Go 1.24, as has been reported in #5968.

### DIFF
--- a/commands/command_clean.go
+++ b/commands/command_clean.go
@@ -8,6 +8,7 @@ import (
 	"github.com/git-lfs/git-lfs/v3/lfs"
 	"github.com/git-lfs/git-lfs/v3/tools"
 	"github.com/git-lfs/git-lfs/v3/tr"
+	"github.com/rubyist/tracerx"
 	"github.com/spf13/cobra"
 )
 
@@ -76,13 +77,13 @@ func clean(gf *lfs.GitFilter, to io.Writer, from io.Reader, fileName string, fil
 		if stat.Size() != cleaned.Size && len(cleaned.Pointer.Extensions) == 0 {
 			Exit("%s\n%s\n%s", tr.Tr.Get("Files don't match:"), mediafile, tmpfile)
 		}
-		Debug("%s exists", mediafile)
+		tracerx.Printf("%s exists", mediafile)
 	} else {
 		if err := os.Rename(tmpfile, mediafile); err != nil {
 			Panic(err, tr.Tr.Get("Unable to move %s to %s", tmpfile, mediafile))
 		}
 
-		Debug(tr.Tr.Get("Writing %s", mediafile))
+		tracerx.Printf("Writing %s", mediafile)
 	}
 
 	_, err = lfs.EncodePointer(to, cleaned.Pointer)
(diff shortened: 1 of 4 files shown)


commands/command_clean.go
```diff
@@ -76,13 +77,13 @@ func clean(gf *lfs.GitFilter, to io.Writer, from io.Reader, fileName string, fil
...
 			Exit("%s\n%s\n%s", tr.Tr.Get("Files don't match:"), mediafile, tmpfile)
 		}
-		Debug("%s exists", mediafile)
+		tracerx.Printf("%s exists", mediafile)
 	} else {
 		if err := os.Rename(tmpfile, mediafile); err != nil {
 			Panic(err, tr.Tr.Get("Unable to move %s to %s", tmpfile, mediafile))
 		}
 
-		Debug(tr.Tr.Get("Writing %s", mediafile))
```
larsxschneider, line 85: We don't want to translate these messages because it is debug output?

larsxschneider (approved):

commands/command_clean.go
chrisd8088: Exactly.  We don't offer localization of trace log messages (as per [few](https://github.com/git-lfs/git-lfs/blob/d0d1ed06177d78e2222ed3d9a7bb07db6effb033/tq/transfer_queue.go#L384) [random](https://github.com/git-lfs/git-lfs/blob/d0d1ed06177d78e2222ed3d9a7bb07db6effb033/tq/transfer_queue.go#L544) [examples](https://github.com/git-lfs/git-lfs/blob/d0d1ed06177d78e2222ed3d9a7bb07db6effb033/tq/transfer_queue.go#L560) from the transfer queue, just for illustrative purposes.)

Issue #5941, opened 2024-12-20
crypto 0.21.0 in git-lfs 3.6.0


daipok:
**Describe the issue**
We noticed this is already done:
https://github.com/git-lfs/git-lfs/commit/4e2d7b092ac51e296f1cc789a2457fdbb02bff2a

Mainly concerning about CVE-2024-45337

What is the possible timeline of another release?

Currently we are still on 3.5.1 thinking of upgrading to 3.6.0 but getting another vulnerability issue so we are evaluating

**System environment**
Rocky8

**Output of `git lfs env`**
The output of running `git lfs env` as a code block.
`
git-lfs/3.5.1 (GitHub; linux amd64; go 1.21.7; git e237bb3a)
git version 2.43.5

LocalWorkingDir=
LocalGitDir=
LocalGitStorageDir=
LocalMediaDir=lfs/objects
LocalReferenceDirs=
TempDir=lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
FetchRecentRefsIncludeRemotes=true
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneVerifyUnreachableAlways=false
PruneRemoteName=origin
LfsStorageDir=lfs
AccessDownload=none
AccessUpload=none
DownloadTransfers=basic,lfs-standalone-file,ssh
UploadTransfers=basic,lfs-standalone-file,ssh
GIT_EXEC_PATH=/usr/libexec/git-core
git config filter.lfs.process = ""
git config filter.lfs.smudge = ""
git config filter.lfs.clean = ""
`

**Additional context**
Any other relevant context about the problem here.

chrisd8088:
As noted in https://github.com/git-lfs/git-lfs/pull/5935#pullrequestreview-2504964858, the vulnerability addressed by the update to the `x/crypto` Go module in PR #5935, reported as CVE-2024-45337, only pertains to the `x/crypto/ssh` package's `ServerConfig` type, which the Git LFS client does not use.

The Git LFS client never acts as an SSH server, so there is no immediate necessity for us to upgrade the module and release a new version of Git LFS.

[...]

PR #5997, opened 2025-02-27
Upgrading x/crypto to v0.35.0 to solve CVE-2025-22869

gergelyfabian:


go.mod
```diff
@@ -39,9 +39,11 @@ require (
...
 	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
-	golang.org/x/crypto v0.31.0 // indirect
-	golang.org/x/text v0.21.0 // indirect
+	golang.org/x/crypto v0.35.0 // indirect
+	golang.org/x/text v0.22.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-go 1.21.0
+go 1.23.0
```
gergelyfabian, line 47: I'm not sure about this, it was automatically added by `go mod tidy`.

chrisd8088:
Hey, thanks for the PR and for jumping on this issue.  Dependabot should offer PRs for these kinds of issues too, but I think you've pre-empted it!  Fortunately, as with #5941 this is another case where the vulnerability doesn't affect the Git LFS client.

This new vulnerability, addressed by the update to the `x/crypto` Go module in v0.35.0 and reported as CVE-2025-22869, only pertains to the `x/crypto/ssh` package, which the Git LFS client does not use.

The Git LFS client never acts as an SSH server, so there is no immediate necessity for us to upgrade the module and release a new version of Git LFS.

While some of the Go packages we use indirectly import packages from the `x/crypto` module, none of them are the `x/crypto/ssh` package.  We can confirm this is the case by running `go list -json all | grep x/crypto/ssh`, which returns no output.

Nevertheless, we might as well upgrade the module, so that when we do release v3.7.0 of the Git LFS client, security scanners will not generate false positive reports about the version of the `x/crypto` module we use.

For reference, the details of the vulnerability and its mitigation are described in [GO-2025-3487](https://pkg.go.dev/vuln/GO-2025-3487) and also in the [release](https://github.com/golang/crypto/releases/tag/v0.35.0) [announcements](https://groups.google.com/g/golang-announce/c/qN_GDasRQSA) for version 0.35.0 of the `x/crypto` module.

---

As for when we'll merge this PR, I think we'll probably wait a few weeks, since it's not urgent and because we'll need to upgrade our CI and Docker builds to Go 1.24 first, and that in turn is likely to expose some other issues which we should address first (see #5968).  I have a set of changes in progress to resolve those, but until the other member of our core team is available to review them, they'll have to wait.

go.mod
chrisd8088: Looks good now that we've bumped up to Go 1.24.

chrisd8088 (approved):

PR #5999, opened 2025-03-01
Replace min/max helpers with built-in min/max

Juneezee:
We can use the built-in `min` and `max` functions since Go 1.21.

Reference: https://go.dev/ref/spec#Min_and_max

### DIFF
--- a/tq/meter.go
+++ b/tq/meter.go
@@ -241,26 +241,16 @@ func (m *Meter) str() string {
 		percentage,
 		m.finishedFiles, m.estimatedFiles,
 		humanize.FormatBytes(clamp(m.currentBytes)),
-		humanize.FormatByteRate(clampf(m.avgBytes), time.Second))
+		humanize.FormatByteRate(clamp(m.avgBytes), time.Second))
 }
 
 // clamp clamps the given "x" within the acceptable domain of the uint64 integer
 // type, so as to prevent over- and underflow.
-func clamp(x int64) uint64 {
+func clamp[T int64 | float64](x T) uint64 {
 	if x < 0 {
 		return 0
 	}
 	if x > math.MaxInt64 {
-		return math.MaxUint64
-	}
-	return uint64(x)
-}
-
-func clampf(x float64) uint64 {
-	if x < 0 {
-		return 0
-	}
-	if x > math.MaxUint64 {
 		return math.MaxUint64
 	}
 	return uint64(x)
--- a/tq/meter.go
+++ b/tq/meter.go
@@ -241,16 +241,26 @@ func (m *Meter) str() string {
 		percentage,
 		m.finishedFiles, m.estimatedFiles,
 		humanize.FormatBytes(clamp(m.currentBytes)),
-		humanize.FormatByteRate(clamp(m.avgBytes), time.Second))
+		humanize.FormatByteRate(clampf(m.avgBytes), time.Second))
 }
 
 // clamp clamps the given "x" within the acceptable domain of the uint64 integer
 // type, so as to prevent over- and underflow.
-func clamp[T int64 | float64](x T) uint64 {
+func clamp(x int64) uint64 {
 	if x < 0 {
 		return 0
 	}
 	if x > math.MaxInt64 {
+		return math.MaxUint64
+	}
+	return uint64(x)
+}
+
+func clampf(x float64) uint64 {
+	if x < 0 {
+		return 0
+	}
+	if x > math.MaxUint64 {
 		return math.MaxUint64
 	}
 	return uint64(x)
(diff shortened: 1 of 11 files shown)


tq/meter.go
```diff
@@ -241,12 +241,12 @@ func (m *Meter) str() string {
...
+		humanize.FormatByteRate(clamp(m.avgBytes), time.Second))
 }
 
 // clamp clamps the given "x" within the acceptable domain of the uint64 integer
 // type, so as to prevent over- and underflow.
-func clamp(x int64) uint64 {
+func clamp[T int64 | float64](x T) uint64 {
 	if x < 0 {
 		return 0
 	}
```
chrisd8088, line 252: The difference between the two functions, aside from their input types, is in how they apply their upper limits.  The `clampf()` performs a [check](https://github.com/git-lfs/git-lfs/blob/aa0ca91e2287d4b7b9e060e7141a61531290113e/tq/meter.go#L263-L265) for an input value greater than what can be stored in a `uint64`, and limits the value to that `math.MaxUint64` value before casting to a `uint64`.  This makes sense, because a `float64` can represent a value much larger than `math.MaxUint64`.

However, the `clamp()` function's code, while it looks identical, is slightly different, and both incorrect and redundant.  It [checks](https://github.com/git-lfs/git-lfs/blob/aa0ca91e2287d4b7b9e060e7141a61531290113e/tq/meter.go#L253-L255) if the input value is larger than `math.MaxInt64`, which can never be true because the input value is an `int64`, so the conditional never executes.  That's actually good, because if the conditional did execute, it would return `math.MaxUint64`, a value 2× larger than `math.MaxInt64`!  (Or more precisely, at least on some hardware it would return `math.MaxUint64`.)

As it is written now, this PR folds these two functions into a single `clamp()` function which accepts a generic input value.  This replacement function uses the upper-bounds check from original `clamp()` function, so it checks for a value larger than `math.MaxInt64`, but then tries to return `max.MaxUint64`.

I say "tries to return" because when you cast `max.MaxUint64` to a `float64` you encounter hardware-dependent behaviour.  The Go language specification [says](https://go.dev/ref/spec#Conversions) in the "Conversions between numeric types" subsection:

> In all non-constant conversions involving floating-point or complex values, if the result type cannot represent the value the conversion succeeds but the result value is implementation-dependent. 

We can test how `math.MaxUint64` is converted to a `float64` and then back to a `uint64` with a small program:

```go
package main

import (
	"fmt"
	"math"
)

func main() {
	n := float64(math.MaxUint64)
	fmt.Printf("       float64(math.MaxUint64) : %f\n", n)
	fmt.Printf("uint64(float64(math.MaxUint64)): %20d\n", uint64(n))
}
```

On an ARM 64-bit CPU, like an Apple Silicon CPU, we get the following, with a final value that's the same as `math.MaxUint64`:
```
       float64(math.MaxUint64) : 18446744073709551616.000000
uint64(float64(math.MaxUint64)): 18446744073709551615
```

On an x64 CPU, we instead get a final value that's much smaller, specifically `1<<32` or `math.MaxInt64 + 1`:
```
       float64(math.MaxUint64) : 18446744073709551616.000000
uint64(float64(math.MaxUint64)):  9223372036854775808
```

If you use `n := float64(math.MaxUint64 - 1024)` on the x64 CPU, though, we then get:
```
       float64(math.MaxUint64) : 18446744073709549568.000000
uint64(float64(math.MaxUint64)): 18446744073709549568
```

Fixing these bugs and discrepancies isn't urgent, though, since it's improbable any user would be dealing with values in the range of these maximums, so I think it's best deferred to another PR.

I've got some code written to address them, but I'll just wait until we've adjusted this PR to deal with the `min()`/`max()` changes only and merged it, and then I'll put those fixes into a followup PR.

chrisd8088 (commented):
Hey, thanks for this PR!  I've been meaning to get around to replacing the old min/max functions, and you've done it for us, which is great.

I also appreciate that you worked on simplifying the `clamp()` and `clampf()` functions in the `tq/meter.go` source file.  Looking at those changes, though, I realized there are some latent problems with both functions, and it's not quite as simple to merge the two into a generic function as we might like.

Would you mind just restoring those two functions, for the time being, so we can approve this PR's changes to all the min/max use cases?  Once that's done I've got some improvements to the `clamp` family of functions which should cover all the edge cases (I think!)

Again, thanks so much for the contribution and for getting this change underway!

tq/meter.go
Juneezee: Thanks for reviewing my PR and providing such a detailed comment :blush:. I have reverted the changes for `tq/meter.go` in commit 310b29f8bae6c6f36854cc44e4d3707a03f666d6

chrisd8088 (approved):
Thank you for helping to clean up and simplify our codebase!  PRs like this are always appreciated.

PR #4289, opened 2020-10-22
pull: gracefully handle merge conflicts

bk2204:
When a merge conflict occurs, it obviously isn't possible to check out the file using git lfs checkout or git lfs pull, since it's unclear which revision the user wants.  git lfs checkout handles this gracefully by just ignoring the conflict, whereas git lfs pull causes a full logged error, which is excessive.

Let's make git lfs pull match the behavior of git lfs checkout here by having it ignore malformed pointer files just like it ignores files that are not pointers.

Fixes #1608

Issue #1608, opened 2016-10-26
'git lfs pull' fails for a specific file


Sakamoto-Shigeru:
Accidentally, I have committed and pushed a text file which is set to track as a binary file by a git-lfs system.
Contents of the file is as follows:

```
version https://git-lfs.github.com/spec/v1
<<<<<<< Updated upstream
oid sha256:12b1af2edcc26a0b9e9218995d0cb50c9fd7cc6981d08f8e25f81a3be415c132
=======
oid sha256:13e36f2710b5c01f9f0c799ddbc804248cd3b25c65e466836a0bc82287c2cdbe
>>>>>>> Stashed changes
size 291328
```

Contents is not a binary but this file was correctly tracked as a binary file by lfs and contents of git-lfs storage is completely same above.

I face to error when doing below:
1. I clone with GIT_LFS_SKIP_SMUDGE=1, and succeeded.
2. 'git lfs pull' succeeded, so the file exists both in my work and .git/lfs/objects/.
3. 'git lfs pull' once more. It was failed.

Error output is below:

```
Problem accessing <PathToTheFile>

Errors logged to /XXXX/.git/lfs/objects/logs/20161026T143946.529695656.log
Use `git lfs logs last` to view the log.
```

Log output is below:

```
git-lfs/1.4.1 (GitHub; linux 386; go 1.7; git 2070e4c)
git version 2.10.0

$ git-lfs pull
Problem accessing <PathToTheFile>

pointer parsing: Expected key oid, got <<<<<<<
github.com/github/git-lfs/errors.newWrappedError
        /tmp/docker_run/src/github.com/github/git-lfs/errors/types.go:166
github.com/github/git-lfs/errors.NewBadPointerKeyError
        /tmp/docker_run/src/github.com/github/git-lfs/errors/types.go:303
github.com/github/git-lfs/lfs.decodeKVData
        /tmp/docker_run/src/github.com/github/git-lfs/lfs/pointer.go:265
github.com/github/git-lfs/lfs.decodeKV
        /tmp/docker_run/src/github.com/github/git-lfs/lfs/pointer.go:139
github.com/github/git-lfs/lfs.DecodeFrom
        /tmp/docker_run/src/github.com/github/git-lfs/lfs/pointer.go:120
github.com/github/git-lfs/lfs.DecodePointer
        /tmp/docker_run/src/github.com/github/git-lfs/lfs/pointer.go:107
[...]

PR #6006, opened 2025-03-05
Detect and react to potential conflict markers in pointer files

DarkDefender:
Before, we would create new hashes for pointer files that contained conflict markers. Now we do not do any extra processing for these kinds of files.

This also makes it so `git diff --check` return the correct result when we have conflicts in pointer files.

Note that I'm very unsure if my changes are safe and doesn't introduce any unintended side effects. `make test` seems to pass, but I'm not sure how exhaustive the test suite is.

Before I started working on this, I created a ticket about the problem here: https://github.com/git-lfs/git-lfs/issues/5990

In that issue, @chrisd8088 pointed out that detecting conflict markers might be very hard as the length of them are configurable. However from my point of view, I think that if we have detected a valid LFS header, then if we detect a "start conflict" marker symbol of any length, we can be quite certain that there are conflict markers in the file.

False positives in this case would mean that the pointer file has been corrupted in some other way and the end result should be the same in that case I think. IE we don't create a new LFS object for the corrupted file and instead just return/save the actual content of the pointer file to git.

In this change I explicitly did not make git lfs error our and return an error to git in the filter process.

This is because I don't think we should stray from how conflicts are handled in regular git for the following reasons:

1. Git GUI programs like VScode and SourceGIT etc, will have to adopt special workarounds for conflict resolution in LFS files as a lot of git commands will in that case error out. For example, they can't use regular git commands like `git diff --check` to check for conflicts anymore. (It also means regular terminal users have to modify their workflow when dealing with LFS conflict resolution)
2. I've personally been (and continue to be) in projects where not everyone on the team knows how to properly resolve a conflict. With git I've multiple times had scenarios where the easiest solution was for someone (that could potentially be on the other side of the world) to create a branch and commit all they have including files with conflict markers and then someone else resolves the conflicts in the branch. LFS erroring out would make this impossible.

On the basic level, I feel that conflict markers is a feature of git itself. If we can detect them, then I think that we should simply treat them as warnings and not errors, just like regular git does.

While this of course means that you can commit broken pointer files, I don't feel that this is a big issue because there is no data loss (I think). If someone needs to prevent people from doing this, I feel like the git server and client hooks should be enough. (IE the hook looks for conflicts markers and denies pushes or commits with them).

Perhaps we could have some example hooks that people could deploy if they wanted to?

### DIFF
--- a/commands/pull.go
+++ b/commands/pull.go
@@ -103,7 +103,7 @@ func (c *singleCheckout) Run(p *lfs.WrappedPointer) {
 				return
 			}
 		} else {
-			if errors.IsNotAPointerError(err) || errors.IsBadPointerKeyError(err) {
+			if errors.IsNotAPointerError(err) || errors.IsBadPointerKeyError(err) || errors.IsPointerConflictMarkerError(err) {
 				// File has non-pointer content, leave it alone
 				return
 			}
--- a/commands/pull.go
+++ b/commands/pull.go
@@ -103,7 +103,7 @@ func (c *singleCheckout) Run(p *lfs.WrappedPointer) {
 				return
 			}
 		} else {
-			if errors.IsNotAPointerError(err) || errors.IsBadPointerKeyError(err) || errors.IsPointerConflictMarkerError(err) {
+			if errors.IsNotAPointerError(err) || errors.IsPointerConflictMarkerError(err) {
 				// File has non-pointer content, leave it alone
 				return
 			}
--- a/lfs/gitfilter_clean.go
+++ b/lfs/gitfilter_clean.go
@@ -80,7 +80,7 @@ func (f *GitFilter) copyToTemp(reader io.Reader, fileSize int64, cb tools.CopyCa
 	n, rerr := buf.Read(by)
 	by = by[:n]
 
-	if rerr != nil || (err == nil && len(by) < blobSizeCutoff) {
+	if rerr != nil || errors.IsPointerConflictMarkerError(err) || (err == nil && len(by) < blobSizeCutoff) {
 		err = errors.NewCleanPointerError(ptr, by)
 		return
 	}
--- a/lfs/gitfilter_clean.go
+++ b/lfs/gitfilter_clean.go
@@ -80,7 +80,7 @@ func (f *GitFilter) copyToTemp(reader io.Reader, fileSize int64, cb tools.CopyCa
 	n, rerr := buf.Read(by)
 	by = by[:n]
 
-	if rerr != nil || errors.IsPointerConflictMarkerError(err) || (err == nil && len(by) < blobSizeCutoff) {
+	if rerr != nil || (err == nil && len(by) < blobSizeCutoff) {
 		err = errors.NewCleanPointerError(ptr, by)
 		return
 	}
@@ -90,6 +90,11 @@ func (f *GitFilter) copyToTemp(reader io.Reader, fileSize int64, cb tools.CopyCa
 		// If there is still more data to be read from the file, tack on
 		// the original reader and continue the read from there.
 		from = io.MultiReader(from, reader)
+	}
+
+	if errors.IsPointerConflictMarkerError(err) {
+		errors.SetContext(err, "reader", from)
+		return
 	}
 
 	size, err = tools.CopyWithCallback(writer, from, fileSize, cb)
--- a/lfs/pointer.go
+++ b/lfs/pointer.go
@@ -276,6 +276,16 @@ func decodeKVData(data []byte) (kvps map[string]string, exts map[string]string, 
 			continue
 		}
 
+		if strings.HasPrefix(text, "<") {
+			err = verifyVersion(kvps["version"])
+			if err == nil {
+				err = errors.NewPointerConflictMarkerError(errors.New(tr.Tr.Get("found potential conflict marker on line %d: %s", line, text)))
+				return
+			}
+			err = errors.NewNotAPointerError(errors.New(tr.Tr.Get("error reading line %d: %s", line, text)))
+			return
+		}
+
 		parts := strings.SplitN(text, " ", 2)
 		if len(parts) < 2 {
 			err = errors.NewNotAPointerError(errors.New(tr.Tr.Get("error reading line %d: %s", line, text)))
--- a/lfs/pointer.go
+++ b/lfs/pointer.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"regexp"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -268,24 +269,35 @@ func decodeKVData(data []byte) (kvps map[string]string, exts map[string]string, 
 	}
 
 	scanner := bufio.NewScanner(bytes.NewBuffer(data))
-	line := 0
-	numKeys := len(pointerKeys)
+	foundConflictMarker := false
+
+	var text_lines []string
 	for scanner.Scan() {
 		text := scanner.Text()
 		if len(text) == 0 {
 			continue
 		}
 
-		if strings.HasPrefix(text, "<") {
-			err = verifyVersion(kvps["version"])
-			if err == nil {
-				err = errors.NewPointerConflictMarkerError(errors.New(tr.Tr.Get("found potential conflict marker on line %d: %s", line, text)))
-				return
-			}
-			err = errors.NewNotAPointerError(errors.New(tr.Tr.Get("error reading line %d: %s", line, text)))
-			return
+		// Check for any potential confilct markers. ('|' marker can appear if changing merge.conflictStyle to "diff3").
+		// We only check the first character as the length of the conflict markers can be changed.
+		// We do not return early to make sure that the file actually seems to be a pointer file first.
+		switch text[0] {
+		case '<', '>', '=', '|':
+			foundConflictMarker = true
+			continue
 		}
 
+		text_lines = append(text_lines, text)
+	}
+
+	err = scanner.Err()
+	if err != nil {
+		return
+	}
+
+	line := 0
+	numKeys := len(pointerKeys)
+	for _, text := range text_lines {
 		parts := strings.SplitN(text, " ", 2)
 		if len(parts) < 2 {
 			err = errors.NewNotAPointerError(errors.New(tr.Tr.Get("error reading line %d: %s", line, text)))
@@ -294,6 +306,15 @@ func decodeKVData(data []byte) (kvps map[string]string, exts map[string]string, 
 
 		key := parts[0]
 		value := parts[1]
+
+		if foundConflictMarker {
+			if slices.Contains(pointerKeys, key) {
+				kvps[key] = value
+				continue
+			}
+			err = errors.NewNotAPointerError(errors.New(tr.Tr.Get("not a git-lfs pointer key on line %d: %s", line, text)))
+			return
+		}
 
 		if numKeys <= line {
 			err = errors.NewNotAPointerError(errors.New(tr.Tr.Get("extra line: %s", text)))
@@ -316,6 +337,15 @@ func decodeKVData(data []byte) (kvps map[string]string, exts map[string]string, 
 		kvps[key] = value
 	}
 
-	err = scanner.Err()
+	if foundConflictMarker {
+		if len(kvps) != numKeys {
+			// Not all pointer keys were present, not a pointer file.
+			err = errors.NewNotAPointerError(errors.New(tr.Tr.Get("Not all required pointer keys were present.")))
+			return
+		}
+		err = errors.NewPointerConflictMarkerError(errors.New(tr.Tr.Get("found potential conflict markers in pointer file.")))
+		return
+	}
+
 	return
 }
--- a/lfs/pointer.go
+++ b/lfs/pointer.go
@@ -312,6 +312,12 @@ func decodeKVData(data []byte) (kvps map[string]string, exts map[string]string, 
 				kvps[key] = value
 				continue
 			}
+			if extRE.MatchString(key) {
+				// The pointer file can contain custom extention entries.
+				// However we do not record if these are present or not as
+				// we only care if all standard keys as these are optional.
+				continue
+			}
 			err = errors.NewNotAPointerError(errors.New(tr.Tr.Get("not a git-lfs pointer key on line %d: %s", line, text)))
 			return
 		}
--- a/lfs/pointer.go
+++ b/lfs/pointer.go
@@ -278,7 +278,7 @@ func decodeKVData(data []byte) (kvps map[string]string, exts map[string]string, 
 			continue
 		}
 
-		// Check for any potential confilct markers. ('|' marker can appear if changing merge.conflictStyle to "diff3").
+		// Check for any potential conflict markers. ('|' marker can appear if changing merge.conflictStyle to "diff3").
 		// We only check the first character as the length of the conflict markers can be changed.
 		// We do not return early to make sure that the file actually seems to be a pointer file first.
 		switch text[0] {
--- a/t/t-pull.sh
+++ b/t/t-pull.sh
@@ -1044,6 +1044,59 @@ begin_test "pull with merge conflict"
 
   set -e
   [ "$res" = "2" ]
+
+  # Ensure that git lfs does not treat the file as a valid pointer file
+  git lfs pointer --check --file abc.bin && exit 1
+  # Make the result of the subshell a success.
+  true
+)
+end_test
+
+begin_test "pull with partially fixed merge conflict"
+(
+  set -e
+  cd pull-merge-conflict
+
+  cp abc.bin abc.bin_bak
+
+  # Define conflict marker removal patterns for us to try
+  patterns=(
+  '^[<]'
+  '^[=]'
+  '^[>]'
+  '^[<=]'
+  '^[<>]'
+  '^[=>]'
+  '^[<=>]'
+  )
+
+  i=1
+  for pattern in "${patterns[@]}"; do
+      echo "Partially removing conflict markers with: $pattern"
+      sed -i "/$pattern/d" abc.bin
+      if [ $i -lt 7 ]; then
+        # Check that the remaining confict markers are still detected by git
+        set +e
+        git diff --check > conficts.log 2>&1
+        res=$?
+
+        set -e
+        [ "$res" = "2" ]
+      else
+        # All conflict markers should have been removed so check that none are detected
+        set +e
+        git diff --check > conficts.log 2>&1
+        res=$?
+
+        set -e
+        [ "$res" = "0" ]
+
+        # Ensure that git lfs does not treat the file as a valid pointer file
+        git lfs pointer --check --file abc.bin && exit 1
+      fi
+      cp abc.bin_bak abc.bin
+      ((i++))
+  done
 )
 end_test
 
--- a/t/t-pull.sh
+++ b/t/t-pull.sh
@@ -1073,7 +1073,7 @@ begin_test "pull with partially fixed merge conflict"
   i=1
   for pattern in "${patterns[@]}"; do
       echo "Partially removing conflict markers with: $pattern"
-      sed -i "/$pattern/d" abc.bin
+      sed -e "/$pattern/d" abc.bin_bak > abc.bin
       if [ $i -lt 7 ]; then
         # Check that the remaining confict markers are still detected by git
         set +e
@@ -1094,7 +1094,6 @@ begin_test "pull with partially fixed merge conflict"
         # Ensure that git lfs does not treat the file as a valid pointer file
         git lfs pointer --check --file abc.bin && exit 1
       fi
-      cp abc.bin_bak abc.bin
       ((i++))
   done
 )
--- a/t/t-pull.sh
+++ b/t/t-pull.sh
@@ -1035,7 +1035,16 @@ begin_test "pull with merge conflict"
   # This will exit nonzero because of the merge conflict.
   GIT_LFS_SKIP_SMUDGE=1 git merge def || true
   git lfs pull > pull.log 2>&1
-  [ ! -s pull.log ]
+  # LFS should have caught the confict markers in the file
+  [ -s pull.log ]
+
+  # Check that conficts are detected by git correctly
+  set +e
+  git diff --check > conficts.log 2>&1
+  res=$?
+
+  set -e
+  [ "$res" = "2" ]
 )
 end_test
 
--- a/t/t-pull.sh
+++ b/t/t-pull.sh
@@ -1035,8 +1035,7 @@ begin_test "pull with merge conflict"
   # This will exit nonzero because of the merge conflict.
   GIT_LFS_SKIP_SMUDGE=1 git merge def || true
   git lfs pull > pull.log 2>&1
-  # LFS should have caught the confict markers in the file
-  [ -s pull.log ]
+  [ ! -s pull.log ]
 
   # Check that conficts are detected by git correctly
   set +e
--- a/t/t-pull.sh
+++ b/t/t-pull.sh
@@ -1012,90 +1012,12 @@ end_test
 begin_test "pull with merge conflict"
 (
   set -e
-  git init pull-merge-conflict
-  cd pull-merge-conflict
 
-  git lfs track "*.bin"
-  git add .
-  git commit -m 'gitattributes'
-  printf abc > abc.bin
-  git add .
-  git commit -m 'abc'
+  setup_local_repo_with_merge_conflict "pull-merge-conflict" \
+  "abc.bin" "other" "def" "ghi"
 
-  git checkout -b def
-  printf def > abc.bin
-  git add .
-  git commit -m 'def'
-
-  git checkout main
-  printf ghi > abc.bin
-  git add .
-  git commit -m 'ghi'
-
-  # This will exit nonzero because of the merge conflict.
-  GIT_LFS_SKIP_SMUDGE=1 git merge def || true
   git lfs pull > pull.log 2>&1
   [ ! -s pull.log ]
-
-  # Check that conficts are detected by git correctly
-  set +e
-  git diff --check > conficts.log 2>&1
-  res=$?
-
-  set -e
-  [ "$res" = "2" ]
-
-  # Ensure that git lfs does not treat the file as a valid pointer file
-  git lfs pointer --check --file abc.bin && exit 1
-  # Make the result of the subshell a success.
-  true
-)
-end_test
-
-begin_test "pull with partially fixed merge conflict"
-(
-  set -e
-  cd pull-merge-conflict
-
-  cp abc.bin abc.bin_bak
-
-  # Define conflict marker removal patterns for us to try
-  patterns=(
-  '^[<]'
-  '^[=]'
-  '^[>]'
-  '^[<=]'
-  '^[<>]'
-  '^[=>]'
-  '^[<=>]'
-  )
-
-  i=1
-  for pattern in "${patterns[@]}"; do
-      echo "Partially removing conflict markers with: $pattern"
-      sed -e "/$pattern/d" abc.bin_bak > abc.bin
-      if [ $i -lt 7 ]; then
-        # Check that the remaining confict markers are still detected by git
-        set +e
-        git diff --check > conficts.log 2>&1
-        res=$?
-
-        set -e
-        [ "$res" = "2" ]
-      else
-        # All conflict markers should have been removed so check that none are detected
-        set +e
-        git diff --check > conficts.log 2>&1
-        res=$?
-
-        set -e
-        [ "$res" = "0" ]
-
-        # Ensure that git lfs does not treat the file as a valid pointer file
-        git lfs pointer --check --file abc.bin && exit 1
-      fi
-      ((i++))
-  done
 )
 end_test
 
--- a/commands/command_clean.go
+++ b/commands/command_clean.go
@@ -61,6 +61,14 @@ func clean(gf *lfs.GitFilter, to io.Writer, from io.Reader, fileName string, fil
 
 		_, err = to.Write(errors.GetContext(err, "bytes").([]byte))
 		return nil, err
+	} else if errors.IsPointerConflictMarkerError(err) {
+		// If the contents read from the working directory are the
+		// result of a merge conflict between two pointers, we'll get
+		// a `PointerConflictMarkerError` with the context containing
+		// the data that we should write back out to Git.
+
+		_, err = io.Copy(to, errors.GetContext(err, "reader").(io.Reader))
+		return nil, err
 	}
 
 	if err != nil {
--- a/errors/types.go
+++ b/errors/types.go
@@ -94,6 +94,19 @@ func IsNotAPointerError(err error) bool {
 	return false
 }
 
+// IsPointerConflictMarkerError indicates that there are conflict markers in the LFS pointer file.
+func IsPointerConflictMarkerError(err error) bool {
+	if e, ok := err.(interface {
+		PointerConflictMarkerError() bool
+	}); ok {
+		return e.PointerConflictMarkerError()
+	}
+	if parent := parentOf(err); parent != nil {
+		return IsPointerConflictMarkerError(parent)
+	}
+	return false
+}
+
 // IsNotAPointerError indicates the parsed data is not an LFS pointer.
 func IsPointerScanError(err error) bool {
 	if e, ok := err.(interface {
@@ -347,6 +360,20 @@ func (e notAPointerError) NotAPointerError() bool {
 
 func NewNotAPointerError(err error) error {
 	return notAPointerError{newWrappedError(err, tr.Tr.Get("Pointer file error"))}
+}
+
+// Definitions for IsPointerConflictMarkerError()
+
+type pointerConflictMarkerError struct {
+	*wrappedError
+}
+
+func (e pointerConflictMarkerError) PointerConflictMarkerError() bool {
+	return true
+}
+
+func NewPointerConflictMarkerError(err error) error {
+	return pointerConflictMarkerError{newWrappedError(err, tr.Tr.Get("Pointer file conflict marker error"))}
 }
 
 // Definitions for IsPointerScanError()
--- a/lfs/pointer_test.go
+++ b/lfs/pointer_test.go
@@ -344,6 +344,149 @@ size 177735`,
 	}
 }
 
+func TestDecodeConflictMarkers(t *testing.T) {
+	// NOTE: The conflict markers examples are mostly single character examples here because of the configurable length.
+	examples := []string{
+		// Real work example
+		`version https://git-lfs.github.com/spec/v1
+<<<<<<< Updated upstream
+oid sha256:7d865e959b2466918c9863afca942d0fb89d7c9ac0c99bafc3749504ded97730
+=======
+oid sha256:bf07a7fbb825fc0aae7bf4a1177b2b31fcf8a3feeaf7092761e18c859ee52a9c
+>>>>>>> Stashed changes
+size 4`,
+
+		// very short conflict markers
+		`version https://git-lfs.github.com/spec/v1
+< Updated upstream
+oid sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+=
+oid sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+> Stashed changes
+size 4`,
+
+		// No conflict comments/hint strings
+		`version https://git-lfs.github.com/spec/v1
+<
+oid sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+=
+oid sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+>
+size 4`,
+
+		// Partially removed conflict markers
+		`version https://git-lfs.github.com/spec/v1
+oid sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+=
+oid sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+>
+size 4`,
+
+		`version https://git-lfs.github.com/spec/v1
+oid sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+oid sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+>
+size 4`,
+
+		`version https://git-lfs.github.com/spec/v1
+<
+oid sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+oid sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+size 4`,
+
+		`version https://git-lfs.github.com/spec/v1
+oid sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+=
+oid sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+size 4`,
+
+		// Test that multiple conflict markers and git-lfs entries can exist
+		// (For compounding merge conflicts)
+		`<
+version https://git-lfs.github.com/spec/v1
+=
+version https://git-lfs.github.com/spec/v2
+<
+>
+oid sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+=
+oid sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+>
+<
+size 10
+=
+size 4
+<`,
+
+		// Test "diff3" style conflict markers
+		`version https://git-lfs.github.com/spec/v1
+<
+oid sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+|
+oid sha256:cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
+=
+oid sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+>
+size 4`,
+
+		`version https://git-lfs.github.com/spec/v1
+oid sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+|
+oid sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+size 4`,
+	}
+
+	for _, ex := range examples {
+		_, err := DecodePointer(bytes.NewBufferString(ex))
+		if err == nil || !errors.IsPointerConflictMarkerError(err) {
+			t.Errorf("No conflict marker detected. From:\n%s", strings.TrimSpace(ex))
+		}
+	}
+}
+
+func TestDecodeConflictMarkersInvalid(t *testing.T) {
+	// NOTE: The conflict markers examples are mostly single character examples here because of the configurable length.
+	examples := []string{
+		// No git-lfs version string
+		`<
+oid sha256:aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
+=
+oid sha256:bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
+>
+size 4`,
+
+		// No size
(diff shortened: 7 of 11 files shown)


DarkDefender:
I didn't implement the next part of my proposal to have an alternate message for pointer file with conflict markers in them. IE, instead of:
```
Encountered 1 file that should have been a pointer, but wasn't:
        <path to conflict file>
```
The message would be something like:
```
Encountered 1 LFS pointer file that has conflict markers in it:
        <path to conflict file>
Select the pointer hash you want and after committing, make sure to run `git lfs checkout`.
See the git-lfs-checkout man page for how to inspect the contents of the different hashes
```

Perhaps a bit verbose, but I think it would be nice to give some hints what to do when you have a conflict with the pointer file.

However I didn't do this yet as I shouldn't really figure out the best way to bubble up the conflict marker error to `filterCommand` in `commands/command_filter_process.go`.

DarkDefender:
Ah, seems like I didn't realize that `make test` in the git root didn't run the tests in the `t` directory.

chrisd8088:
Hey, thanks so much for working on this issue!

As your PR's title indicates that it's still work-in-progress, I don't know if you want a full review yet or not.  I hope to get a chance over the next few days to think a bit more about the potential issues involved with this change; I have some `$DAYJOB` tasks I need to complete first, unfortunately.

A few quick initial notes:
- The Windows CI job failure is just due to a flake, and the Go 1.22.x job failure is unrelated to your changes here.  (We'll have to bump ourselves up to Go 1.23 and 1.24 soon, clearly.)
- I like your suggestion that it could be sufficient to just look for a single character from a conflict marker; that's definitely simpler than anything else!  We may want to look for `>`, `=`, and `|` characters as well as just `<` characters, just to be thorough.  (The `|` character is used in some of the markers Git generates when its `diff3` conflict marker style is configured.)
- While our test suite is reasonably comprehensive, it generally doesn't contain tests for situations we haven't previously considered in depth.  As this PR changes the `clean` filter, which is used when adding Git LFS files to the Git index, I think some new tests in scripts such as `t/t-clean.sh` and `t/t-filter-process.sh` to validate the expected behaviour would be extremely valuable.  There are some existing [tests](https://github.com/git-lfs/git-lfs/blob/79dfe3bc2ca8856060b97893c200c7885e5bf05c/t/t-filter-process.sh#L120-L158) in the latter script which could good models for additional tests of this PR's changes.
- On a quick read-through, it seems as though the nicely crafted error message from the `decodeKVData()` and `NewPointerConflictMarkerError()` functions are never actually output anywhere.  If we go in that direction, we probably don't need such detailed messages (i.e., with the line number and text showing the potential conflict marker).
- While it's a bit of a pathological case, I was able to commit a broken Git LFS pointer with the following content using this PR's changes as they stand now.  Offhand, I feel as though we ought to be a bit more strict in trying to check for conflict markers, such that for example we expect to see at least one each of the `oid` and `size` lines, and no extraneous content other than conflict-marker lines.
    ```
    < This is the version string.
    
    Anything goes here.
    ```
- You make an interesting point about errors versus warnings when conflict markers are detected, and I think I could be persuaded in that direction.  I'd love to see a broader set of test cases working through the common conditions involving merge conflicts on Git LFS pointers, though, so we have some concrete examples to discuss.

Thanks again very much for this PR!  We sincerely appreciate new ideas and contributions like this from our community; it's what really helps the project to evolve and improve!

DarkDefender:
> As your PR's title indicates that it's still work-in-progress, I don't know if you want a full review yet or not.

I added `WIP` as I wasn't sure at all if this was the correct approach, so I didn't want to seem like I was certain :sweat_smile: .
I also wanted to add the extra `Encountered 1 LFS pointer file that has conflict markers in it:` warning. But at this point I think it might be better to do that later and just focus on getting the current change reviewed and merged.

I'll remove the `WIP` tag.

> I hope to get a chance over the next few days to think a bit more about the potential issues involved with this change; I have some $DAYJOB tasks I need to complete first, unfortunately.

No problem! Take your time. I've already rolled this out to my local computer lab so I can just continue to use my patched version until this makes it in proper! :)

> - I like your suggestion that it could be sufficient to just look for a single character from a conflict marker; that's definitely simpler than anything else! We may want to look for >, =, and | characters as well as just < characters, just to be thorough. (The | character is used in some of the markers Git generates when its diff3 conflict marker style is configured.)
> - While it's a bit of a pathological case, I was able to commit a broken Git LFS pointer with the following content using this PR's changes as they stand now. Offhand, I feel as though we ought to be a bit more strict in trying to check for conflict markers, such that for example we expect to see at least one each of the oid and size lines, and no extraneous content other than conflict-marker lines.

I think these two points touches on the area of how strict we want to be with classifying pointer files as corrupted or not. Note that I don't really have much against checking for complete conflict markers, but I feel like we might instead go into an other direction:

----

To me, I feel like if the `version https://git-lfs.github.com/spec/v1` (or one of other the other two headers) are present, then we should always threat the file as a pointer file. If there is any other unexpected entries in the file, we treat it as corrupted.

This then leads into the other topic you touched on:

>  - On a quick read-through, it seems as though the nicely crafted error message from the decodeKVData() and NewPointerConflictMarkerError() functions are never actually output anywhere. If we go in that direction, we probably don't need such detailed messages (i.e., with the line number and text showing the potential conflict marker).

You are correct! Perhaps then we shouldn't treat conflict markers differently and instead just have it be a generic "pointer file is corrupt" error. This would then make the check even more generic. `decodeKVData` would then work like this:

1. If we encounter that is not withing the pointer file spec after `version https://git-lfs.github.com/spec/v1` has been found, we error out with a generic `PointerCorruptionError`. (This is just a rename of `PointerConflictMarkerError`).
2. We don't give conflict markers any special treatment as the end result in this case will be the same. (No new pointer is created and the corrupted pointer file is simply used as is
3. The error doesn't really contain detailed information as it will not be printed anywhere

----

We could of course also go the otherway of keeping the explicit conflict marker detection and error. The only question is if it is worth it to accurately detect conflicts as I feel like we probably want to treat other corruption errors in the same way.

> - While our test suite is reasonably comprehensive, it generally doesn't contain tests for situations we haven't previously considered in depth. As this PR changes the clean filter, which is used when adding Git LFS files to the Git index, I think some new tests in scripts such as t/t-clean.sh and t/t-filter-process.sh to validate the expected behaviour would be extremely valuable. There are some existing tests in the latter script which could good models for additional tests of this PR's changes.
> - You make an interesting point about errors versus warnings when conflict markers are detected, and I think I could be persuaded in that direction. I'd love to see a broader set of test cases working through the common conditions involving merge conflicts on Git LFS pointers, though, so we have some concrete examples to discuss.

I completely agree that adding more tests is good!
I'm a bit unsure on exactly what the tests should test out as everything seems to work correctly to me. And because I can't seem to discover any broken things, I don't know what the tests should encompass. I'm a bit afraid I will just conjure a lot of trivial tests that doesn't really test anything useful.

We should probably discuss this more as I agree that having more useful tests in this area is something we should have!

chrisd8088:
Thank you for all the thoughts and detailed notes!  I'll try to find some time to do a more detailed review in the next week or so.

> I think these two points touches on the area of how strict we want to be with classifying pointer files as corrupted or not.

I'd like to start by being pretty strict, as I think that's the more cautious approach.  We can always make the detection more permissive in future PRs, if that proves valuable and unproblematic.

At present we don't consider a file to be a Git LFS pointer unless it [includes](https://github.com/git-lfs/git-lfs/blob/79dfe3bc2ca8856060b97893c200c7885e5bf05c/lfs/pointer.go#L291-L301) only the expected elements in the expected order, is under a fixed size, and meets various other conditions.

While detecting conflict markers is a great enhancement, I think we want to be careful not to relax these rules too much until we've shaken out all the edge cases.  I'm inclined to think that we want to see all the Git LFS elements (`version`, `oid`, `size`), in that order, with limited duplications, plus conflict marker lines which at least generally match the expected patterns for those, and nothing else.

> I don't know what the tests should encompass.

I'm not entirely sure either, but at a high level, I think we probably want to cover cases such as:

- Various types of files with and without conflict markers, such as:
    - Partially edited files which have only one or two kinds of markers in them (e.g., a `=======` line but nothing else), or files where multiple conflicting merges have introduced several sets of conflict markers.
    - Files with some conflict marker characters in them but which lack Git LFS fields like `oid` and `size`, which I think we should not recognize them as a pointers-with-markers (at least in an initial implementation).
    - Expanding the Go [tests](https://github.com/git-lfs/git-lfs/blob/79dfe3bc2ca8856060b97893c200c7885e5bf05c/lfs/pointer_test.go) of the `DecodePointer()` function should help us cover many of these examples of recognized and unrecognized pointers-with-markers.
- Commands which invoke the `DecodePointer()` function:
    - As one example, we should probably confirm that `git lfs pointer --check --stdin` doesn't return a zero (success) exit code if passed a pointer-with-markers.
- Commands which invoke the `GitFilter.Clean()` method:
    - "Plumbing" commands `git lfs clean` and `git lfs filter-process`.
    - "Porcelain" commands like `git add` and `git diff`.
    - The `GitFilter.Clean()` method is [invoked](https://github.com/git-lfs/git-lfs/blob/79dfe3bc2ca8856060b97893c200c7885e5bf05c/commands/uploader.go#L388) for missing Git LFS objects during `git push`/`git lfs push` operations, but that can _probably_ be ignored as it may be a vestige of some old logic and effectively unused now.
 
We may find that a relatively small set of tests can cover most or all of these conditions, of course, and some of these new tests are going to be redundant so long as everything works as intended.

However, there's value in having tests which simply check that certain behaviour remains the same both with and without the changes in this PR.  That way anyone who revises or updates this code in the future can have some confidence that the test suite will detect a range of possible regressions.

And thank you again very much for working on this PR!

DarkDefender:
Just to give a quick update.
I got a lot of unexpected high priority tasks that takes priority over this.
I haven't lost interest, I'm just busy and I don't really know when I will be free. :(

chrisd8088:
> Just to give a quick update.
> I got a lot of unexpected high priority tasks that takes priority over this

@DarkDefender — Not to worry, and thank you for the update!  I started a review yesterday but have since been pulled back over to some other responsibilities, so I fully appreciate that you also have other things to focus on than this PR!

I think the intent of this PR is really valuable, though, and I very much hope to return to a review soon.

May I ask, would you be willing to send me an invitation through the GitHub UI to collaborate on your repository?  If so, and assuming I find some time to do so, I could add a few commits to this PR to try to implement some of the Go tests and other adjustments I have in mind, and then we could request a review from another member of the core Git LFS team.  But I will understand if you'd prefer not to follow that route, in which case I'll simply aim to finish a review as soon as I'm able.

Thanks again very much for contributing to Git LFS!

DarkDefender:
> May I ask, would you be willing to send me an invitation through the GitHub UI to collaborate on your repository?
Absolutely! If you have time to work on the code I will not try to stop you. Thanks you so much!

I've added you as a collaborator to on my fork repo.

DarkDefender:
@chrisd8088 long time no see!
I finally got some time to work on this after more than a year! :D

I added all of the tests (I think) you requested besides the `GitFilter.Clean()` tests as I'm not too sure what to test for. I'm guessing you want me to test if adding and committing the files will commit the conflict files as regular git files (not tracked by lfs)?

Thank you again for the thorough list of test cases. They managed to expose quite a few issues and weaknesses with my old code.

For example not handling files with partial conflict markers that `git diff --check` detects.
Because of this I had to rework the logic a bit where it now does a full pass over the 1024 byte file and checks for potential conflict markers first.

I don't think this should make the parsing much slower because of the minimal file sizes. But if it becomes an issue, we could check if the lines starts with any of the keyword characters (v,o,s) or any of the conflict markers and if not exit early.

chrisd8088:
> long time no see!
> I finally got some time to work on this after more than a year! :D

@DarkDefender — Hey, thanks for picking this back up again!

We're hoping to make a release in the next couple of weeks and there are a few issues I have to deal with as part of that process, so I may be a little delayed in re-reviewing the latest changes to this PR, but I'll take a look as soon as I can.

Thanks again for working on this, and my apologies for not finishing a full review of the original version.

DarkDefender:
@chrisd8088 Just to double check, I guess this will not make it into the upcoming release?
I realize it is not a critical fix, so no worries if not. 

To me, it would be nice so to get it in so that some of the people that I work remotely with would be able to get properly notified by their git clients when there are conflicts in the lfs files. Currently I have to  provide custom builds of git lfs for them with this patch.

chrisd8088:
> I guess this will not make it into the upcoming release?

Unfortunately, I think that's probably the case, mainly because I'm trying to prioritize one last bug fix (from PR #6011, which is almost as old as this PR!) and because I just haven't had time to give this a proper review yet.  I apologize for the delay in doing so, and will try to make time to take a close look as soon as I'm able.

Thank you again for your considerable patience and your help in trying to improve our project!

lfs/pointer.go
```diff
@@ -268,14 +269,35 @@ func decodeKVData(data []byte) (kvps map[string]string, exts map[string]string,
...
 		text := scanner.Text()
 		if len(text) == 0 {
 			continue
 		}
 
+		// Check for any potential confilct markers. ('|' marker can appear if changing merge.conflictStyle to "diff3").
+		// We only check the first character as the length of the conflict markers can be changed.
+		// We do not return early to make sure that the file actually seems to be a pointer file first.
+		switch text[0] {
+		case '<', '>', '=', '|':
```
chrisd8088, line 285: I think this is a good way to aligning our marker detection with the approach taken in Git's `is_conflict_marker()` [function](https://github.com/git/git/blob/v2.54.0/diff.c#L3499-L3520).  Ideally, we might try to also match the current marker length specified by the Git configuration and `gitattributes` files, but this is a good start.

commands/pull.go
```diff
@@ -103,7 +103,7 @@ func (c *singleCheckout) Run(p *lfs.WrappedPointer) {
 				return
 			}
 		} else {
-			if errors.IsNotAPointerError(err) || errors.IsBadPointerKeyError(err) {
+			if errors.IsNotAPointerError(err) || errors.IsBadPointerKeyError(err) || errors.IsPointerConflictMarkerError(err) {
```
chrisd8088: I was intrigued as to why this one particular method makes a point of checking whether attempting to decode a pointer resulted in a `BadPointerKeyError` or not, when none of our commands do so.

It turns out this is a because it is a relatively simple way to detect conflict markers and prevent error messages appearing because of them during `git lfs checkout` and `git lfs pull` commands.

I'll describe the history and behaviour in more detail below, but in brief, given that the intent of this PR is to introduce a more accurate way of detecting conflict markers, we should just remove the `errors.IsBadPointerKeyError()` check entirely:

```suggestion
			if errors.IsNotAPointerError(err) || errors.IsPointerConflictMarkerError(err) {
```

---

For my own reference, here are a few notes on how we came to check for `BadPointerKeyError`s in this conditional block.

We added the `errors.IsBadPointerKeyError()` check in commit 9e3bc45b934758e422c3e800161dabd0711775bf of PR #4289 in order to address the issue reported in #1608, which occurred when a working tree file contained a Git LFS pointer with conflict markers and `git lfs pull` was run.  Prior to PR #4289, this would produce an error message, and the purpose of the PR was to silence that message.

(Note that the PR's description states that `git lfs checkout` did not produce an error message in this case, but I was unable to reproduce that, even using an old version of Git LFS without the change from PR #4289.  Since the `git lfs checkout` command shared this same code as the `git lfs pull` command then as now, I think perhaps the PR description was just mistaken.)

The reason the addition of the `errors.IsBadPointerKeyError()` check worked is was due to the specific formatting usually found after two Git LFS pointers conflict during a merge.  First, suppose the working tree file contains something like the following:

```
<<<<<<< HEAD
oid sha256:b439f67c6e8bd45073d963848bcf5b736b7483a11d98bbad332afb7283398127
=======
oid sha256:4859dc3b10cc3d56efa9940e6d29e29cb6609572e73026991e01f1e9062adc23
>>>>>>> other
size 5
```

Because the `<<<<<<< HEAD` line contains two whitespace-separated fields, the `decodeKVData()` function in our `lfs` package [considers](https://github.com/git-lfs/git-lfs/blob/c3b78de708bd36470ad5cf423b7361d7d3b4d0c6/lfs/pointer.go#L279-L283) the line a possible pointer key/value pair and doesn't return a `NotAPointerError`.  It then [finds](https://github.com/git-lfs/git-lfs/blob/c3b78de708bd36470ad5cf423b7361d7d3b4d0c6/lfs/pointer.go#L293-L297) that the line doesn't match the expected `oid` key/value pair and isn't a Git LFS pointer extension line either, and so returns a `BadPointerKeyError`.

The `decodeKV()` function then [calls](https://github.com/git-lfs/git-lfs/blob/c3b78de708bd36470ad5cf423b7361d7d3b4d0c6/lfs/pointer.go#L171-L173) the `StandardizeBadPointerError()` function from our `errors` package, which [wraps](https://github.com/git-lfs/git-lfs/blob/c3b78de708bd36470ad5cf423b7361d7d3b4d0c6/errors/types.go#L142-L144) the `BadPointerKeyError` in a `NotAPointerError`, but only if the "bad" pointer key was the `version` key.  When two pointers cause a merge conflict, though, the most common result is like the example above where the first unexpected line appears in place of the `oid` key, so the `StandardizeBadPointerError()` function doesn't return a `NotAPointerError`.

Hence the change from PR #4289 was necessary to try to detect conflict markers because it wasn't sufficient to just look for a `NotAPointerError`.  However, because it isn't actually a check for any conflict markers in the file, it's actually an overly broad check.

For instance, if we do the following, the `git lfs checkout` and `git lfs pull` commands don't report an error at all, even though the working tree file is not the result of a merge conflict between pointers at all:

```
$ git init test
$ cd test
$ git lfs track "*.bin"
$ echo foo >foo.bin
$ git commit -m foo

$ cat <<EOF >foo.bin
foo bar
EOF

$ git lfs pull
# no error is output
```

The reason there's no error is because we're using the `errors.IsBadPointerKeyError()` check in this conditional block.  If, on the other hand, we do the following, we do get an error:

```
$ cat <<EOF >foo.bin
oid foo
EOF

$ git lfs pull
Checkout error for "foo.bin": Invalid OID value: foo
```

So I think it would actually be an improvement to use the more-accurate `errors.IsPointerConflictMarkerError()` check introduced in this PR in place of the `errors.IsBadPointerKeyError()` check.

Note that the `pull with merge conflict` [test](https://github.com/git-lfs/git-lfs/blob/c3b78de708bd36470ad5cf423b7361d7d3b4d0c6/t/t-pull.sh#L1012-L1040) will succeed with my suggested change, and that the test was explicitly added in PR #4289 to verify that PR's behaviour.

lfs/gitfilter_clean.go
```diff
@@ -80,7 +80,7 @@ func (f *GitFilter) copyToTemp(reader io.Reader, fileSize int64, cb tools.CopyCa
 	n, rerr := buf.Read(by)
 	by = by[:n]
 
-	if rerr != nil || (err == nil && len(by) < blobSizeCutoff) {
+	if rerr != nil || errors.IsPointerConflictMarkerError(err) || (err == nil && len(by) < blobSizeCutoff) {
```
chrisd8088: There's a lot going on in this conditional expression already, before we even add in any detection of conflict markers.

I'm fairly confident that the `rerr != nil` condition is wrong, first of all.  We've had that in place since commit 6f54232890174a119f7dbfd562d5c1ab57174d0b of PR #1796, but I still suspect it's not correct (but almost harmless).  I've added some notes on that further below, for my own reference.

Putting that issue aside for now, I do understand why it helps the `git diff --check` command if we consider pointers with conflict markers as "clean".

But I think the more common case is for someone to run `git add` to try to resolve a merge conflict, and end up with a broken pointer in their repository.

So I think we have to do something here like this (note that I'm explicitly deferring dealing with the `rerr != nil` condition to a future PR):

```suggestion
	if rerr != nil || (err == nil && len(by) < blobSizeCutoff) {
		err = errors.NewCleanPointerError(ptr, by)
		return
	} else if errors.IsPointerConflictMarkerError(err) {
		return
	}
```

That will generate errors of the form `Error cleaning Git LFS object: Pointer file conflict marker error: found potential conflict markers in pointer file.`  It's not as pretty in the `git diff --check` case, but I think it's a distinct improvement in the `git add` case.

In other words, I think my remarks from this [comment](https://github.com/git-lfs/git-lfs/issues/5990#issuecomment-2664777910) on #5990 still apply here:

>  [W]e could then return an `error` status from the `git-lfs-filter-process` command ... when we detected what appeared to be a Git LFS pointer file except that it had conflict markers in it.  If we did this, a Git command like `git add` or `git diff` would simply exit with a message like the following:
>
> ```
> Error cleaning Git LFS object: possible conflict markers found in Git LFS pointer file: test.bin
> error: external filter 'git-lfs filter-process' failed
> fatal: test.bin: clean filter 'lfs' failed
> ```
> 
> For `git diff`, that's not as nice as if we returned the contents of the original file in such a case.  If we could do that ... `git diff --check` would work as you expect.
>
> However, `git diff` isn't the only Git command which calls Git LFS's filter process.  The `git add` command, for instance, also calls the `clean` filter, and there we very likely don't want to just return the contents of the file, because then you'd add the broken Git LFS pointer file as-is into your repository.  It would be more valuable in this case to cause the error message to appear, which would prevent people from accidentally adding broken Git LFS pointers with conflict markers in them.  And unfortunately I don't think our filter process has a generic and reliable way of determining which Git command has invoked it, so we can't really vary it's behaviour based on that information.

---

Just for future reference, here's why I think the `rerr != nil` condition should result in an error, not a "clean" pointer.

In the rare case that the underlying `io.Reader` returns an error other than `io.EOF` and zero bytes when the `DecodeFrom()` function tries to [read](https://github.com/git-lfs/git-lfs/blob/c3b78de708bd36470ad5cf423b7361d7d3b4d0c6/lfs/pointer.go#L131) from it, then the `io.MultiReader` that the function [constructs](https://github.com/git-lfs/git-lfs/blob/c3b78de708bd36470ad5cf423b7361d7d3b4d0c6/lfs/pointer.go#L134-L141) and returns will do so as well when `Read()` is called on it above, so both `err` and `rerr` would be non-`nil`.  We shouldn't really treat that unusual case as a clean, empty pointer, even though no bytes have been read, because the `io.Reader` is reporting some kind of problem.

Note that if the underlying `io.Reader` returns any data at all, plus a non-EOF error, then we happen to handle that case correctly.  The `bytes.Reader` constructed by the `DecodeFrom()` function will contain the data read prior to the error, and so the first `Read()` we make on the `io.MultiReader` will just read data from the `bytes.Reader` and its internal byte slice, and so will return `(n>0, nil)`.  That means `rerr` is `nil` and `err` is non-`nil` (since the `DecodeFrom()` function preserves the error from the original `Read()` call), so this conditional expression evaluates `false || false`, we don't return a `CleanPointerError`, and proceed to try to read more, at which point the non-EOF error will be reported because we're reading from the underlying `io.Reader` and not the `bytes.Reader`.

lfs/pointer.go
```diff
@@ -285,6 +307,15 @@ func decodeKVData(data []byte) (kvps map[string]string, exts map[string]string,
 		key := parts[0]
 		value := parts[1]
 
+		if foundConflictMarker {
+			if slices.Contains(pointerKeys, key) {
+				kvps[key] = value
+				continue
```
chrisd8088, line 313: This is a definite improvement in how we look for potential Git LFS pointers with conflict markers!

I think it's still somewhat overly broad in how it works, though, while also not accounting for Git LFS pointer [extensions](https://github.com/git-lfs/git-lfs/blob/c3b78de708bd36470ad5cf423b7361d7d3b4d0c6/docs/extensions.md), which are a rare but valid possibility.

For instance, this is caught by the current logic and reported with a `PointerConflictMarkerError`:

```
=
oid foo
version 1
size git-lfs
size bar
```

But the key/value lines are distinctly out of order, there's no version [string](https://github.com/git-lfs/git-lfs/blob/c3b78de708bd36470ad5cf423b7361d7d3b4d0c6/lfs/pointer.go#L22-L24), the `oid` and `size` values aren't valid, there's an extra `size` line, etc.

Conversely, this example is not reported with a `PointerConflictMarkerError`, but it's a completely well-formed conflict between two pointers in a repository where our Git LFS pointer extension test [utility](https://github.com/git-lfs/git-lfs/blob/c3b78de708bd36470ad5cf423b7361d7d3b4d0c6/t/cmd/lfstest-caseinverterextension.go) was configured:

```
<<<<<<< HEAD
ext-0-caseinverter sha256:9491f5f5982344efb8b84111232199488769252ad7e73494fca8a74a639c85be
oid sha256:93fe3f2872a5b64013387bd7c8c539dc0cacaaed73abd02d19e8df1f497ade85
=======
ext-0-caseinverter sha256:d03246324a51c99af4c912e1141f2fde59afdbc92f3d23a99220baf2cb8700fa
oid sha256:81ab12d3ecf28f40b12632cdd34e17172d90b247a00168379d2b28e4e269df3d
>>>>>>> other
size 4
```

(For reference, you can configure this utility in a shell integration test using the `setup_case_inverter_extension()` shell function.  For example, the `pull: pointer extension` test in our `t/t-pull.sh` script [calls](https://github.com/git-lfs/git-lfs/blob/c3b78de708bd36470ad5cf423b7361d7d3b4d0c6/t/t-pull.sh#L1421) this function before running some checks on pointer extension support in the `git lfs pull` command.)

At a minimum, rather than just recognizing that one of the `version`, `oid`, and `size` keys has been found and continuing the loop (when `foundConflictMarker` is `true`), I suspect we need do something more complex.  Ideally, we'd allow for pointer extension lines using the `extRE` regular expression, which is currently [used](https://github.com/git-lfs/git-lfs/blob/c3b78de708bd36470ad5cf423b7361d7d3b4d0c6/lfs/pointer.go#L294) to ignore those lines when looking at the order of the three required keys (`version`, `oid`, and `size`).  I would be inclined to think about how we might count those keys and the extension lines when `foundConflictMarker` is `true`, and expect 1 or 2 of each, but not more than that.  (Also, note that there can in theory be multiple extensions configured, so one could have lines like `ext-0-foo`, `ext-1-bar`, etc.  We don't have a second test utility to exercise this right now, unfortunately, at least not in a simple way.)

lfs/pointer.go
```diff
@@ -268,14 +269,35 @@ func decodeKVData(data []byte) (kvps map[string]string, exts map[string]string,
...
+	foundConflictMarker := false
+
+	var text_lines []string
 	for scanner.Scan() {
 		text := scanner.Text()
 		if len(text) == 0 {
 			continue
 		}
 
+		// Check for any potential confilct markers. ('|' marker can appear if changing merge.conflictStyle to "diff3").
```
chrisd8088: Minor typo: `confilct` should be `conflict`.

t/t-pull.sh
```diff
@@ -1036,6 +1036,66 @@ begin_test "pull with merge conflict"
...
   git lfs pull > pull.log 2>&1
   [ ! -s pull.log ]
+
+  # Check that conficts are detected by git correctly
+  set +e
+  git diff --check > conficts.log 2>&1
+  res=$?
+
+  set -e
+  [ "$res" = "2" ]
```
chrisd8088: Minor typo: `conficts` should be `conflicts`.

Unless we're checking the output of the command, there's no need for a log file.  However, I would recommend that we do check the command's output, because I'm assuming we're going to generate errors rather than treat these pointers with conflicts as "clean" pointers:

```suggestion
  # Check that conflicts are detected by git correctly
  git diff --check 2>&1 | tee diff.log
  [ 128 -eq "${PIPESTATUS[0]}" ]

  grep "Error cleaning Git LFS object" diff.log
  grep "found potential conflict markers in pointer file" diff.log
  grep "abc.bin: clean filter 'lfs' failed" diff.log
```

This relies on a technique we use throughout our shell tests, which is that by piping a command's output into `tee(1)`, we don't need to tinker with the `errexit` shell option (i.e., by using `set +e`/`set -e`) because even if the command returns a non-zero exit code, the `tee` command will not and it's exit code will be used as the result of the pipeline, so the test won't halt.  We can then check the first command's exit code explicitly using the `PIPESTATUS` array variable.

Moreover, because this is really a check of our `clean` filter operations, and not of our `git lfs pull` command, I'm going to suggest moving it into a separate test in our `git-filter-process.sh` script.

I admit this is a somewhat subtle distinction, but as a general principle we try to test our own commands in their respective test scripts.  Those tests may include lots of other Git and Git LFS commands, of course, but as a guideline, if we're actively verifying that one of our Git LFS commands behaves a certain way, we should try to perform that check in a test and a script dedicated to that command.  (Note that this differs from when we rely on one of our commands, like `git lfs track "*.bin"`, to just set up the conditions for a check of another command's behaviour.)

To simplify the setup for multiple tests which all need to create merge conflicts, I'd suggest implementing a helper function like this:

```shell
setup_local_repo_with_merge_conflict() {
  local reponame="$1"
  local filename="$2"
  local branch="$3"
  local branch_contents="$4"
  local main_contents="$5"

  git init "$reponame"
  cd "$reponame"

  git lfs track "$filename"
  printf "%s" "$filename" >"$filename"
  git add .gitattributes "$filename"
  git commit -m "add $filename"

  git checkout -b "$branch"

  printf "%s" "$branch_contents" >"$filename"
  git add "$filename"
  git commit -m "$branch_contents"

  git checkout main

  printf "%s" "$main_contents" >"$filename"
  git add "$filename"
  git commit -m "$main_contents"

  # This will exit nonzero because of the merge conflict.
  git merge "$branch" || true
}
```

Then this existing `pull with merge conflict` test can be simplified to the following:

```shell
begin_test "pull with merge conflict"
(
  set -e

  setup_local_repo_with_merge_conflict "pull-merge-conflict" \
    "abc.bin" "other" "def" "ghi"

  git lfs pull > pull.log 2>&1
  [ ! -s pull.log ]
)
end_test
```

We'd then add a new `filter process: merge conflict causes error` test to the `t/t-filter-process.sh` script like this, which checks that the filter behaves as we expect with both the `git add` and `git diff` commands:

```shell
begin_test "filter process: merge conflict causes error"
(
  set -e

  setup_local_repo_with_merge_conflict "filter-process-merge-conflict" \
    "test.bin" "other" "other" "main"

  # Ensure Git will choose "filter.lfs.process" over "filter.lfs.clean".
  git config --global --unset filter.lfs.clean

  # Check that conflicts are detected by git correctly
  git diff --check 2>&1 | tee diff.log
  [ 128 -eq "${PIPESTATUS[0]}" ]

  grep "Error cleaning Git LFS object" diff.log
  grep "found potential conflict markers in pointer file" diff.log
  grep "test.bin: clean filter 'lfs' failed" diff.log

  git add test.bin 2>&1 | tee add.log
  [ 128 -eq "${PIPESTATUS[0]}" ]

  grep "Error cleaning Git LFS object" add.log
  grep "found potential conflict markers in pointer file" add.log
  grep "test.bin: clean filter 'lfs' failed" add.log
)
```

If one felt really conscientious, a similar `clean: merge conflict causes error` test could be added to the `t/t-clean.sh` script, except it would force the use of the legacy `git lfs clean` command.

t/t-pull.sh
```diff
@@ -1036,6 +1036,66 @@ begin_test "pull with merge conflict"
...
+
+  # Ensure that git lfs does not treat the file as a valid pointer file
+  git lfs pointer --check --file abc.bin && exit 1
+  # Make the result of the subshell a success.
+  true
```
chrisd8088: Since this is really a check of the behaviour of one of our own Git LFS commands, I think it belongs in a separate test in our `t/t-pointer.sh` script.

I realize there are a few [instances](https://github.com/git-lfs/git-lfs/blob/c3b78de708bd36470ad5cf423b7361d7d3b4d0c6/t/t-filter-process.sh#L291-L300) of the `git lfs pointer --check --file` command in our `t/t-filter-process.sh` script, but those are used to verify that the `smudge` filter operation of the `git lfs filter-process` command has worked as we expect.

In this case, though, we're really checking how the `git lfs pointer` command itself behaves, so I think that could reasonably be moved to its own test in the `t/t-pointer.sh` script.

t/t-pull.sh
```diff
@@ -1036,6 +1036,66 @@ begin_test "pull with merge conflict"
...
+)
+end_test
+
+begin_test "pull with partially fixed merge conflict"
+(
+  set -e
+  cd pull-merge-conflict
```
chrisd8088: I appreciate that's it's straightforward to make use of the conditions established by the previous test in this test, and we do have some legacy tests which [depend](https://github.com/git-lfs/git-lfs/blob/c3b78de708bd36470ad5cf423b7361d7d3b4d0c6/t/t-checkout.sh#L728-L731) on a previous test in the same script to set up a repository which they can reuse.

However, this approach makes it difficult to understand the dependencies between tests, and if one test starts to fail due to a regression, it can cause others to fail only because they lack their expected pre-conditions.  That makes debugging harder, and it is also makes it harder to refactor the test suite and extract certain tests into new or different files, as we sometimes want to do.

So I would recommend using the `setup_local_repo_with_merge_conflict()` helper I drafted for this test to make it completely independent of any others.

We should also move this test into the `t/t-filter-process.sh` script, since it's (indirectly) checking the behaviour of the `git lfs filter-process` command which the `git diff --check` command runs, and isn't checking the behaviour of the `git lfs pull` command at all.

t/t-pull.sh
```diff
@@ -1036,6 +1036,66 @@ begin_test "pull with merge conflict"
...
+      echo "Partially removing conflict markers with: $pattern"
+      sed -e "/$pattern/d" abc.bin_bak > abc.bin
+      if [ $i -lt 7 ]; then
+        # Check that the remaining confict markers are still detected by git
+        set +e
+        git diff --check > conficts.log 2>&1
+        res=$?
+
+        set -e
+        [ "$res" = "2" ]
```
chrisd8088: Here and in the `else` block below, my notes and suggestions about the previous test also apply.

There's a minor typo in `confict` in a few places, and you can avoid the need for `set +e`/`set -e` and actually check the output of the `git diff --check` command using `grep(1)`.

```suggestion
        # Check that the remaining conflict markers are still detected by git
        git diff --check 2>&1 | tee diff.log
        [ 128 -eq "${PIPESTATUS[0]}" ]

        grep "Error cleaning Git LFS object" diff.log
        grep "found potential conflict markers in pointer file" diff.log
        grep "test.bin: clean filter 'lfs' failed" diff.log
```

And I'd skip the final `git lfs pointer` check, for the same reason I noted above, namely that it's better done in a separate test in the `t/t-pointer.sh` script.

t/t-pull.sh
```diff
@@ -1036,6 +1036,66 @@ begin_test "pull with merge conflict"
...
+  '^[<>]'
+  '^[=>]'
+  '^[<=>]'
+  )
+
+  i=1
+  for pattern in "${patterns[@]}"; do
+      echo "Partially removing conflict markers with: $pattern"
+      sed -e "/$pattern/d" abc.bin_bak > abc.bin
+      if [ $i -lt 7 ]; then
```
chrisd8088: We can avoid the use of a hard-coded limit here, so that if the size of the `patterns` array changes in the future (e.g., to add patterns with `|` in them) we don't have to remember to adjust the limit as well:

```suggestion
      if [ $i -lt ${#patterns[@]} ]; then
```

chrisd8088 (commented):
Hey, thanks again for pursuing the ideas we discussed in #5990, and thank you for your patience!

I really like the direction this work is going.  I had a few thoughts regarding some of the edge cases we'll need to watch out for when trying to detect conflict markers, such as handling Git LFS pointer extension lines.

I also realized that this PR would obviate the need for the `errors.IsBadPointerKeyError(err)` [check](https://github.com/git-lfs/git-lfs/blob/c3b78de708bd36470ad5cf423b7361d7d3b4d0c6/commands/pull.go#L106) in the `git lfs pull` and `git lfs checkout` commands, because that's actually just there to try to detect conflict markers using a simpler and less precise approach than this PR would implement.

Most importantly, though, I think we have to report an error when a `clean` filter operation (i.e., one performed by a `git lfs filter-process` or `git lfs clean` command) encounters what seems like a Git LFS pointer with conflict markers.  If we treat it as a "clean", valid pointer, then people can easily commit those broken pointers into their repository.  Conversely, if we don't try to detect them at all (which is what we do now), then we convert them into Git LFS objects.  Either approach is inappropriate, so I think we have to report an error.

Because we can't know during a `clean` filter operation which Git command might have invoked us, we can't vary our approach.  It would be nice if we could treat pointer files with conflict markers as "clean" when our commands are run by a `git diff --check` command, but report an error when we're run by a `git add` or `git hash-object` command.  But we don't have a way to figure this out (at least, not that I'm aware of), so I think reporting an error is the least-bad option.

Again, thanks so much for pursuing this line of work!  It's turned up some interesting history and has been thought-provoking to review, that's for sure.  :-)

DarkDefender:
@chrisd8088 thanks for the review!

About making the git commands error out when there is a conflict, I think we probably should discuss that before I move onto the other review comments.
I also started with the goal of having it error out, but I changed my opinion after I implemented it that way initially. I could have sworn that I had written about my finding and motivation why I backtracked on that. But I can't find it so I'll write it here:

There are two reasons I changed my opinion with the first one being the most important:

1. It was a while ago, but if I remember correctly, last time I tested erroring out, VSCode and other git GUI clients would stop working properly if we returned an error when running `git diff --check`. I'm guessing it is because git gets the error and stops (because why would it continue if any of the subcommands errored out?). I think you can also see this in the cli where it wouldn't open `less` with the conflict output.

To me, the side effect of erroring out makes this change quite useless as I want to be able to get conflict warnings as I do with regular git. (IE clients doesn't need to handle conflicts errors differently if they are in lfs pointer files)

2. While working on productions with a lot of less technical people, it has happened quite a lot that someone remote runs into conflicts that they don't really know how to resolve. In a few of those cases, it got resolved by them knowingly committing files with conflict markers in them and me going in and fixing up the files for them.

If we make `git add` error out, then this would be impossible.

While it might not be nice to allow foot guns like this, I think that we should stick to what is possible with regular git and not introduce any new restrictions.

chrisd8088:
Thanks for the thoughtful reply, @DarkDefender, and the additional context.  I've been musing on your comments for a week or so and I think you've convinced me that your approach is the right one.

My principal concern has been the user experience when performing a `git add` command on a pointer file with conflict markers, as might occur after a merge conflict.  We have three basic options, as far as I can see: convert the file into a Git LFS object (the current behaviour of the Git LFS client), make no changes to the file and allow it to be added as-is, or report an error.

I think just about everyone would agree that the current behaviour is the least desirable of these options.  I still also think that reporting an error is the more ergonomic and user-friendly option.

However, upon reflection I remembered that we try to let our design choices be guided by those of the Git project, and Git assumes that when you run the `git add` command on a file it means you've reviewed the file's contents and intend to commit them to the repository as they stand.  (It's hardly necessary, but for the sake of completeness I'll include a simple example below.)

<details><summary>Example</summary>

```
$ git init test
$ cd test
$ echo test >test.txt
$ git add test.txt
$ git commit -m init

$ git checkout -b other
$ echo other >test.txt
$ git add test.txt
$ git commit -m other

$ git checkout main
$ echo main >test.txt
$ git add test.txt
$ git commit -m main

$ git merge other
Auto-merging test.txt
CONFLICT (content): Merge conflict in test.txt
Automatic merge failed; fix conflicts and then commit the result.

$ git add test.txt
$ git commit -m merge
$ git show HEAD:test.txt
<<<<<<< HEAD
main
=======
other
>>>>>>> other
```
</details>

Given that model, we should probably just follow Git's approach and allow Git LFS pointers to be added to the index and committed even if they clearly contain conflict markers.  Which is, of course, what you proposed.

I'll revise a couple of my review comments with this in mind.  Thanks again for the additional discussion, as it prompted some further reflection on my part.

lfs/gitfilter_clean.go
chrisd8088: In https://github.com/git-lfs/git-lfs/pull/6006#issuecomment-5007709239 I outlined why I now think we should just allow Git LFS pointers with conflict markers to be added to the index, as you proposed in the first place.

Unfortunately, though, we can't just naively construct a `CleanPointerError` the same way as we do when we've found a valid Git LFS pointer.

By definition valid pointers must be 1024 bytes or less in size.  This means that the byte slice and buffer [constructed](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/lfs/pointer.go#L130-L134) by the `DecodeFrom()` function are guaranteed to be sufficient to hold the entire contents of the working tree file, which the `copyToTemp()` method then [duplicates](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/lfs/gitfilter_clean.go#L80) into its own `by` byte slice and passes to our `errors` package's `NewCleanPointerError()` function, which [stores](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/errors/types.go#L334) it in a "context" field named `bytes`.  The `clean()` function in our `commands` package can then just assert that context field is a byte slice and [write](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/commands/command_clean.go#L62) it out to Git.

However, when we're dealing with what appears to be a pointer file with conflict markers, we can't be certain the total size of the file is 1024 bytes or less.  Usually it will be, of course.  But users can make conflict markers arbitrarily long by setting the `conflict-marker-size` Git attribute, and they can also set the `merge.conflictStyle` configuration option to `diff3`/`zdiff3`, which inflates the number of lines in the conflict description.  If one or two Git LFS pointer extensions are also configured, exceeding 1024 bytes is definitely possible.  An edge case, to be sure, but not unreasonably so.

Given that, I think we'll need to structure this portion of the `copyToTemp()` method like this:

```go
	if rerr != nil || (err == nil && len(by) < blobSizeCutoff) {
		err = errors.NewCleanPointerError(ptr, by)
		return
	}

	var from io.Reader = bytes.NewReader(by)
	if fileSize < 0 || int64(len(by)) < fileSize {
		// If there is still more data to be read from the file, tack on
		// the original reader and continue the read from there.
		from = io.MultiReader(from, reader)
	}

	if errors.IsPointerConflictMarkerError(err) {
		errors.SetContext(err, "reader", from)
		return
	}
```

Then in `commands/command_clean.go` where we currently only [handle](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/commands/command_clean.go#L57-L64) the `CleanPointerError`, we will now need to handle the two types of errors differently:

```go
	if errors.IsCleanPointerError(err) {
		// If the contents read from the working directory was _already_
		// a pointer, we'll get a `CleanPointerError`, with the context
		// containing the bytes that we should write back out to Git.
		_, err = to.Write(errors.GetContext(err, "bytes").([]byte))
		return nil, err
	} else if errors.IsPointerConflictMarkerError(err) {
		// If the contents read from the working directory are the
		// result of a merge conflict between two pointers, we'll get
		// a `PointerConflictMarkerError` with the context containing
		// the data that we should write back out to Git.
		_, err = io.Copy(to, errors.GetContext(err, "reader").(io.Reader))
		return nil, err
	}
```

One nice benefit of this change is that because the `Clean()` method of our `GitFilter` structure will now return a distinct `PointerConflictMarkerError` rather than conflating it with a `CleanPointerError`, the `git lfs pointer` command will typically be able to report the conflict markers with a clear error message.  For example:

```
$ git lfs pointer --file conflict.bin
Pointer file conflict marker error: found potential conflict markers in pointer file.
```

t/t-pull.sh
chrisd8088: In https://github.com/git-lfs/git-lfs/pull/6006#issuecomment-5007709239 I outlined why I now think we should just allow Git LFS pointers with conflict markers to be added to the index, as you proposed in the first place.

Most of my suggestions in the comment above are still applicable as I wrote them.  We should create a `setup_local_repo_with_merge_conflict()` test helper function and refactor the existing `pull with merge conflict` [test](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/t/t-pull.sh#L1012-L1040) to make use of it.

However, instead of adding a `filter process: merge conflict causes error` test as I wrote it above, we should add a `filter process: pointer merge conflict detected and not converted` test instead:

```shell
begin_test "filter process: pointer merge conflict detected and not converted"
(
  set -e

  setup_local_repo_with_merge_conflict "filter-process-merge-conflict" \
    "test.bin" "other" "other" "main"

  # Git will choose "filter.lfs.process" over "filter.lfs.clean".
  git config --global --unset filter.lfs.clean

  # Check that conflicts are detected by Git correctly.
  git diff --check 2>&1 | tee diff.log
  [ 2 -eq "${PIPESTATUS[0]}" ]

  [ 3 -eq "$(grep -c "leftover conflict marker" diff.log)" ]

  contents_oid="$(calc_oid_file "test.bin")"

  # Check that an unresolved merge conflict is not converted
  # to a Git LFS object.
  git add test.bin
  git commit -m "commit with merge conflict"

  refute_local_object "$contents_oid"
)
end_test
```

I'd also recommend we include a simpler version of the test in our `t/t-clean.sh` script as well:

```shell
begin_test "clean: pointer merge conflict detected and not converted"
(
  set -e

  setup_local_repo_with_merge_conflict "clean-merge-conflict" \
    "test.bin" "other" "other" "main"

  # Check that an unresolved merge conflict is not converted
  # to a Git LFS object.
  diff -u "test.bin" <(git lfs clean <"test.bin")
)
end_test
```

t/t-pull.sh
chrisd8088: Here's how I might write a test of the `git lfs pointer` command, FWIW:

```shell
begin_test "pointer: merge conflict detected"
(
  set -e

  setup_local_repo_with_merge_conflict "pointer-merge-conflict" \
    "test.bin" "other" "other" "main"

  git lfs pointer --check --file "test.bin" && exit 1

  git lfs pointer --file "test.bin" 2>&1 | tee pointer.log
  [ 0 -ne "${PIPESTATUS[0]}" ]

  grep "Pointer file conflict marker error" pointer.log
  grep "found potential conflict markers in pointer file" pointer.log
)
end_test
```

lfs/pointer.go
DarkDefender: We could add some sanity checks to see if there is more than 3 entries of each key. (If you use the diff3 style then a single merge conflict could potentially have three entries of the same key.

The reason I didn't do this is that you could potentially have more than one merge conflict in the same file depending on what the user was doing. For example merging multiple branches together in one go.

I feel that checking if the file only contains keys that we expect in a pointer file should be a strong enough of a heuristic for us to quite confidentially guess if it is pointer file with conflict markers. Of course if you insist on adding limits to how many times a key could appear, then I will do it. I just worry about us missing conflict markers and creating new pointer files again.

I tried to implement extension support but ran into issues. I'll elaborate in an other comment below.

DarkDefender:
@chrisd8088 I think I've implemented the majority of your feedback.
I think that the only things left (correct me if I'm wrong), is working extension support and maybe limiting the number of times a key can appear in a conflict file. (For the key amount topic see my code review reply)

I added support for having extension markers in the conflict files. I also added a test, but that is where I noticed some issues what I don't really know how I would resolve in a good way.

The issue is that when there are extensions, the file seems to fed to the extension pipeline without first checking if the file is actually a broken pointer file. Therefore if the user has any extensions turned on, the file will be fed through the extension filters before any pointer checks are done and thus we won't get any conflict markers.

The (failing) test I added in the last commit show cases this issue.

I'm guessing we could solve this by adding a pointer file sanity check first before we run the file through any extensions?

chrisd8088:
Thanks so much for continuing to work on this PR, @DarkDefender!

I apologize for the delay in reviewing the latest changes.  I have not forgotten about this PR and will circle back to it as soon as I can.  (At least you're in good company: there are quite a few PRs which deserve our attention, but that's been in very short supply of late.)

Thanks again very much for both your time and effort, and for your patience.

PR #6011, opened 2025-03-12
filter-process: Ignore file size in working copy when cleaning

### LINKED ISSUE #4974: Help triaging git-lfs checkout failure
**Describe the issue**
In our CI environment, we run `git lfs fetch` followed by a `git lfs checkout` to properly pull down/populate large fixture objects in our repo from git LFS pointers, needed to run our tests. Randomly (anecdotally, seems to be about 5% of the time) the `git lfs checkout` step silently fails (see logs below), leaving the local repo checkout in a dirty state (unexpected/undesired behavior for CI environment for obvious reasons)

To break this apart, there's at least 2 issues/problems for us here:
1. `git lfs checkout` is failing (see error message and stack trace below) and we cannot figure out why
2. when the above failure occurs, the `git lfs checkout` process still exits cleanly (exit code 0), despite failing and leaving our repo in a dirty state.

Any help in debugging the first problem would be very much appreciated, and any thoughts on if the second is an expected behavior (and if so, why) or if it's worth opening a separate bug report issue would also be helpful. Thanks!

**System environment**
CircleCI job, running with their Docker Executor, using a variety of images (observed on `cimg/go:1.17`, `cimg/ruby:2.7.5-node`, `cimg/node:14.18.1-browsers`, `cimg/node:14.18.1`, etc)

git lfs is not installed by default on any of these base `cimg` images, so it is installed each run using the following bash
```bash
  # Skip installing git-lfs if already present
  if ! command -v git-lfs &> /dev/null; then
    curl -sL https://github.com/git-lfs/git-lfs/releases/download/v3.1.2/git-lfs-linux-amd64-v3.1.2.tar.gz -o git-lfs.tar.gz \
    && tar -xzf git-lfs.tar.gz \
    && sudo ./install.sh
  fi
  git-lfs --version  # Print out version for debugging
  cd ~/build && git lfs install  # Finally, install it in the repo
```

Error logged to `~/build/.git/lfs/logs/20220428T065800.570779134.log`
```
git-lfs/3.1.2 (GitHub; linux amd64; go 1.17.6)
git version 2.35.1

$ git-lfs checkout
Error updating the Git index:
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/git-lfs/git-lfs/v3/git.(*FilterProcessScanner).readRequest(0xc0000a3b48)
	work/git-lfs/git-lfs/git/filter_process_scanner.go:188 +0x245
github.com/git-lfs/git-lfs/v3/git.(*FilterProcessScanner).Scan(0xc0000a3b48)
	work/git-lfs/git-lfs/git/filter_process_scanner.go:152 +0x76
github.com/git-lfs/git-lfs/v3/commands.filterCommand(0xc000125900, {0x9be0d9, 0x0, 0x0})
	work/git-lfs/git-lfs/commands/command_filter_process.go:74 +0x345
github.com/spf13/cobra.(*Command).execute(0xc000125900, {0xe710b8, 0x0, 0x0})
	go/pkg/mod/github.com/spf13/cobra@v0.0.3/command.go:766 +0x5f8
github.com/spf13/cobra.(*Command).ExecuteC(0xc000124000)
	go/pkg/mod/github.com/spf13/cobra@v0.0.3/command.go:852 +0x2dc
github.com/spf13/cobra.(*Command).Execute(...)
	go/pkg/mod/github.com/spf13/cobra@v0.0.3/command.go:800
github.com/git-lfs/git-lfs/v3/commands.Run()
	work/git-lfs/git-lfs/commands/run.go:107 +0x2ed
main.main()
	work/git-lfs/git-lfs/git-lfs.go:34 +0xea
error: packet write with format failed
fatal: title-vi/fixtures/orig_map_trips.json: clean filter 'lfs' failed


exit status 128

Current time in UTC:
2022-04-28 06:58:00

Environment:
LocalWorkingDir=/home/circleci/remix
LocalGitDir=/home/circleci/remix/.git
LocalGitStorageDir=/home/circleci/remix/.git
LocalMediaDir=/home/circleci/remix/.git/lfs/objects
LocalReferenceDirs=
TempDir=/home/circleci/remix/.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/home/circleci/remix/.git/lfs
AccessDownload=none
AccessUpload=none
GIT_TRACE=1

Client IP addresses:
192.168.240.3
```

> If relevant, include the output of `git config -l` as a code block.

```
url.ssh://git@github.com.insteadof=https://github.com
gc.auto=0
filter.lfs.required=true
core.filemode=true
core.bare=false
core.logallrefupdates=true
remote.origin.url=git@github.com:[redacted]/[redacted].git
branch.master.remote=origin
lfs.repositoryformatversion=0
```

**Output of `git lfs env`**
```
git-lfs/3.1.2 (GitHub; linux amd64; go 1.17.6)
git version 2.33.0

Endpoint=https://github.com/[redacted]/[redacted].git/info/lfs (auth=none)
  SSH=git@github.com:[redacted]/[redacted].git
LocalWorkingDir=/home/circleci/build
LocalGitDir=/home/circleci/build/.git
LocalGitStorageDir=/home/circleci/build/.git
LocalMediaDir=/home/circleci/build/.git/lfs/objects
LocalReferenceDirs=
TempDir=/home/circleci/build/.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/home/circleci/build/.git/lfs
AccessDownload=none
AccessUpload=none
```

**Additional context**
I was able to capture a failure while running `git lfs fetch && GIT_TRACE=1 git lfs checkout`. You can view the trace output in [this gist](https://gist.github.com/oderby/f4f1a894048e81b42782914248538174) (omitting only the redundant listing of all files in our repo and other repeated lines)

cynix:
Sometimes when Git sends us the clean command, the target file in the
working copy does not have the correct size (e.g. it's empty). Since
Git will feed us the blob to clean via stdin, we have to read it all
anyway, especially if we're invoked via filter-process where leaving
any data unread will cause our Pktline's reader to not be positioned
correctly for parsing the next command. Don't trust the file size on
disk when determining how much to copy.

Resolves #4974 (specifically, the secondary problem reported in that issue).


t/cmd/lfstest-badlocalfile.go
```diff
@@ -0,0 +1,72 @@
...
+	// For any file Git asks us to clean, rename it and create a
+	// 0-byte file in its place. This allows Git to continue to
+	// read from the original file and send us its contents via
+	// packets on stdin, while our filter-process command will
+	// see the 0-byte file as soon as the first flush packet is
+	// received.
+	//
+	// If we simply truncate the file to 0 bytes in situ, Git
+	// may not be able to finish reading its full contents, and
+	// might send us fewer packets than actually needed.
```
cynix, line 25: I don't know how to actually reproduce what Git is doing when it invokes our filter-process with incorrect files in the working copy, so I'm using this util to simulate it. Let me know if you figure out a way to reproduce this natively.

chrisd8088:
Hey, thanks very much for this PR and for the investigation into the problems reported #4974, which have proven challenging to replicate!  We really appreciate this kind of contribution from our community of users.

I think this PR deserves a thoughtful review and so I hope to devote a solid chunk of time to it as soon as I can.  My apologies in advance if that takes a week or two; I can assure you it's not because we're not interested in this potential fix for a longstanding issue!

chrisd8088:
Thank you again for this PR and for your patience, @cynix!

I've started working on a review, and will get that done as soon as possible so you can see my suggested revisions, which are for the most part quite minor, at least as regards the work in this PR at the moment.

However, there are some ancillary changes I think we'll need to make as a consequence of this PR, and so I may try to implement those in another PR which we could merge ahead of this one, or ask if I can contribute some commits to this PR (for which I'd need an invitation to collaborate on your repository).

My apologies for the delay in replying, and thanks again very much for this bug fix!

cynix:
@chrisd8088 Thanks for the update. I've invited you as a collaborator, feel free to add commits as you see fit. I'm guessing you're thinking of finding a way to still report progress to the user even though we can't trust the file size? (e.g. use the file size on disk as a hint for progress display only, not for determining how much to copy, or something like that) :D

chrisd8088:
> I'm guessing you're thinking of finding a way to still report progress to the user even though we can't trust the file size?

Yes, that's one of the issues, indeed.  At the moment, the file size in the progress log (when enabled with `GIT_LFS_PROGRESS`) will have `-1` as the denominator in the "number of bytes written so far" field, so I worked up some changes to provide an alternative (along with a bunch of tests), but then noticed a more subtle issue, which is that when `total` is `-1`, we inadvertently defeat the 200ms [throttling](https://github.com/git-lfs/git-lfs/blob/97f47b9cd2102273287066e0225a6c10457c401a/lfs/util.go#L58) of writes to the progress log, so we write to the log on every callback.

Since noticing that I've been pulled off in another direction, but will return to this topic shortly, I very much hope!

chrisd8088:
Hey, I just wanted to note that although I haven't had time to circle back to this PR and the outstanding progress log issues, I will do so in time because I think this is a valuable change and I'd like to get it merged.

Thanks again for your contribution and your patience!

chrisd8088:
It's taken me much longer than I expected to circle back to this PR, but I have not forgotten about it!

For my own reference, I wanted to note that the progress log (when enabled with `GIT_LFS_PROGRESS`) will already have a `-1` in the denominator of the bytes-received field if a server replies to an object transfer request with a `Content-Encoding: gzip` header and a gzipped response body.

The Go HTTP client automatically inserts an `Accept-Encoding: gzip` request header, as described in #6193, so a server is within its rights to reply with a gzipped response.  When the `net/http` package from the Go standard library has inserted the `Accept-Encoding: gzip` request header on its own, and the server returns a gzipped response with a `Content-Encoding: gzip` header, the `ContentLength` value of the `Response` structure is [set](https://github.com/golang/go/blob/9777ceceec8fee294d038182739cab7c845ad2d1/src/net/http/transport.go#L2461) to `-1`, and we then just pass that on to our progress meter.

Under these conditions, we'll also bypass our throttling of progress log messages, as I noted above in https://github.com/git-lfs/git-lfs/pull/6011#issuecomment-2761801198.

Enabling a progress log with `GIT_LFS_PROGRESS="$TRASHDIR/progress.log"` in our `storage upload with compression` [test](https://github.com/git-lfs/git-lfs/blob/bb65882304a655ffa8abf2be6922e53ff18af5a5/t/t-push.sh#L958-L998) is sufficient to demonstrate this behaviour.

cynix:
> It's taken me much longer than I expected to circle back to this PR, but I have not forgotten about it!

Thanks for keeping this on your radar. Keen to see it merged eventually :)

chrisd8088:
Believe it or not, I've been able to spend some time focusing on the issues raised by this PR, with the intent that we might include this PR in our upcoming (and long-delayed) v3.8.0 release.

PR #6285 is the first of two separate PRs I'd like to merge ahead of this PR.  The next PR will refactor our progress log callback mechanisms and expand the test suite for those.

Then I'll have just a handful of revisions for this PR, mostly to deal with a few specific details which I'll note in a final review.  I can push up the changes myself, as I've already written the necessary commits.

Thanks again for the diagnosis and fix for this bug, and for your considerable patience!

cynix:
Thanks for the update, very happy to see progress being made, can’t wait for your new release :)

chrisd8088:
PR #6288 is the second of two PRs I'd like to merge ahead of this one, as it resolves several problems with the progress log that would be much more apparent once we no longer try to guess the size of objects during `clean` operations by "peeking" at the corresponding size of the files in the working tree.

After that I just have a small set of tweaks for this PR, most importantly removing the [block](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/lfs/gitfilter_clean.go#L73-L75) which would otherwise disable the progress log entirely during `clean` operations.  I'll write up those commits and push them into this PR's branch once we've got PR #6288 in place.

t/cmd/lfstest-badlocalfile.go
chrisd8088: I think this is a perfectly reasonable way to simulate the problem!  Most likely some other process is involved which removes or alters the working copy file after Git opens it but before Git LFS can check its size, so this is a clever and effective way to simulate the issue for our test suite.

t/cmd/lfstest-badlocalfile.go
```diff
@@ -0,0 +1,72 @@
...
+							panic(err)
+						}
+
+						f, err := os.OpenFile(pathname, os.O_WRONLY|os.O_CREATE, 0o644)
+						if err != nil {
+							panic(err)
+						}
+						if err = f.Truncate(0); err != nil {
+							panic(err)
+						}
```
chrisd8088: Here I think we can slightly simplify the action of this utility by just forcing a new file to be created, which by definition will be empty:

```suggestion
						f, err := os.OpenFile(pathname, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0o644)
						if err != nil {
							panic(err)
						}
```

t/t-filter-process.sh
```diff
@@ -317,3 +317,33 @@ begin_test "filter process: git archive does not invoke SSH"
...
+
+  reponame="repo-issue-4974"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" "$reponame"
+
+  git lfs track "*.dat"
+  git add .gitattributes
+  git commit -m "initial commit"
+
+  dd if=/dev/zero of=large.dat bs=65537 count=1
```
chrisd8088: Rather than create a file of a fixed size, we can now take advantage of the `lfstest-getlimit` [helper](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/t/cmd/lfstest-getlimit.go) we added in PR #6285 to ensure that we create a file large enough to exceed the [size](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/lfs/scanner.go#L10-L12) of the internal buffer we use when [checking](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/lfs/gitfilter_clean.go#L79-L86) whether a file already contains a valid Git LFS pointer.

In PR #6285 we updated some [other tests](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/t/t-filter-process.sh#L193-L255) in this script to use the `lfstest-getlimit` helper, so following their current design, we can do something like the following here:

```suggestion
  # Test with a file size larger than the size of the read buffer used when
  # decoding pointers.  See the secondary issue reported in
  # https://github.com/git-lfs/git-lfs/issues/4974#issuecomment-2717907543
  # and https://github.com/git-lfs/git-lfs/pull/6011.
  max_pointer_size="$(lfstest-getlimit --max-pointer-size)"
  contents_size=$((max_pointer_size * 2))
  head -c "$contents_size" /dev/zero >large.dat
```

t/t-filter-process.sh
```diff
@@ -317,3 +317,33 @@ begin_test "filter process: git archive does not invoke SSH"
...
+  oid=$(calc_oid_file large.dat)
+
+  GIT_TRACE_PACKET=1 git \
+    -c "filter.lfs.process=lfstest-badlocalfile | git-lfs filter-process" \
+    -c "filter.lfs.clean=false"\
+    -c "filter.lfs.smudge=false" \
+    -c "filter.lfs.required=true" \
+    add large.dat
```
chrisd8088: If we capture the trace log output from the `git add` command, which includes the trace log output from the `git lfs filter-process` command as well, we can check to make sure the `unknown command` error message no longer appears:

```suggestion
    add large.dat 2>&1 | tee filter.log
  [ 0 -eq "${PIPESTATUS[0]}" ]

  [ 0 -eq "$(grep -c "unknown command" filter.log)" ]
```

commands/command_clean.go
chrisd8088: We should rewrite this comment to reflect the revised action of the function:

```diff
-// If fileSize is given as a non-negative (>= 0) integer, that value is used
-// with preference to os.Stat(fileName).Size(). If it is given as negative, the
-// value from the `stat(1)` call will be used instead.
+// If the total size of the object is unknown until it is read, fileSize
+// should be a negative integer.

```

chrisd8088 (commented):
Thank you again for this PR, and my apologies for the amount of time it's taken for us to get it merged!

I've already written commits which address all of my notes and suggestions from this belated review, along with some other minor cleanup issues, and I'll push those up to this PR once we've merged PR #6288 into `main` and I can cleanly pull its changes into this PR's branch.
 
In other words, there should be no need for you to try to address any of my comments; they are just here for future reference.

---

The key hurdle, as you [noted](https://github.com/git-lfs/git-lfs/pull/6011/#issuecomment-2761229659), is that by ignoring the file in the working tree which (usually!) corresponds to the object we're processing through our `clean` filter, that means we'll always pass a `-1` down to the `copyToTemp()` method of our `GitFilter` structure, and so if we make no other changes, that method will always [unset](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/lfs/gitfilter_clean.go#L73-L75) the callback function pointer we explicitly [create](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/commands/command_clean.go#L38) for `clean` filter operations.  As a result, we'd effectively disable progress logging during these operations, even when it should be enabled because the `GIT_LFS_PROGRESS` environment variable specifies a log file path.

We added the conditional block in the `copyToTemp()` method back in PR #3902 with intent that it would cause our `clean` filter operations to skip creating an object file in local storage when there was no need for it, for instance, if the invoking command was `git hash-object` rather than `git add`.  However, I believe we were under the misapprehension that the callback function is responsible for creating an object file, when its only purpose is to write an entry in the progress log (if a log file has been specified with the `GIT_LFS_PROGRESS` environment variable).

So we can just remove this conditional block and allow the `-1` values to be passed down to the callback function.  However, doing so exposes another problem, which is that in PR #5504 we revised the anonymous callback function [created](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/lfs/util.go#L56-L67) by the `CopyCallbackFile()` method of the `GitFilter` structure so that it throttles its writes to the progress log, and we explicitly did so to reduce the progress log I/O overhead during `clean` filter operations.

To throttle its writes, the callback function skips writing an entry to the log if less than 200 ms have elapsed since the previous write for the same object and the `written` parameter is less than the `total` parameter.  That logic ensures a final entry is written regardless of the amount of time elapsed since the previous write, but only if the `total` parameter is valid.  If it's `-1`, though, the throttling logic will be defeated and every callback will result in a write, which would re-introduce the problem PR #5504 was meant to solve.

That problem is what led to the work in PR #6288, which in part ensures the throttling works in all cases, even when `total` is negative.  That PR's changes also mean we won't write a literal `-1` into the progress log anymore (something we already do, per my notes in https://github.com/git-lfs/git-lfs/pull/6011#issuecomment-3866531617), and will use a `????`-style field instead.

There's also a lot of other cleanup that went into that PR, to deal with the fact that for historical reasons we actually have two functions which format [identical](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/lfs/util.go#L59) [messages](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/tq/meter.go#L278) for the progress log, and (separately) [two](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/tools/copycallback.go#L70-L75) [structures](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/tools/copycallback.go#L11-L16) which wrap `io.Reader` and `io.ReadSeekCloser`, respectively, and these are used in three different combinations to generate entries in the log.

---

We have two other commands besides `git lfs clean` and `git lfs filter-process` which also call the `clean()` function, the `git lfs migrate import` command and the `git lfs merge-driver` command.  In the case of the `git lfs migrate import` command, we actually know the size of a given object for certain in the [two](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/commands/command_migrate_import.go#L412) [instances](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/commands/command_migrate_import.go#L176) where we call the `clean()` function, because we're reading a blob from repository's Git history.

However, in the case of the `git lfs merge-driver` command, we [pipe](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/commands/command_merge_driver.go#L39) the output of another command into a temporary file (the one identified by the `%D` specifier), and then [make](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/commands/command_merge_driver.go#L76-L79) an `os.Stat()` call just to find the size of that file before [passing](https://github.com/git-lfs/git-lfs/blob/ac3a0f402316f8b1e29e6348f4a630adb6512539/commands/command_merge_driver.go#L88) it to the `clean()` function.

Since it's at least theoretically possible for the temporary file to change between the time we read its size and the time we open it, we should just skip the `os.Stat()` call and pass `-1` to the `clean()` command in its `fileSize` parameter.

larsxschneider (approved):

PR #6014, opened 2025-03-13
Honour 429 object transfer response `Retry-After` headers

### LINKED ISSUE #6007: Retry-After header wait time when HTTP 429 is received is not considered. Does not wait.
**Describe the bug**
When using a git server with git-lfs support, and the rate limit when requesting files is it and [HTTP 429](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) is emitted would expect `git-lfs` to wait the correct number of time, specified in [Retry-After](https://developer.mozilla.org/de/docs/Web/HTTP/Headers/Retry-After) header.

What actually happens is that `git-lfs` retries **HTTP 429** as often as possible, which puts additional load on the git server.

**To Reproduce**
1. Set a low rate limit on your git server for git lfs (GitHub, Gitlab etc.)
2. Try to clone a git repo with lots of `git lfs` tracked files by running
    `GIT_CURL_VERBOSE=1 GIT_TRANSFER_TRACE=1 git clone https://your-git-server.com/your-repo.git > trace.txt 2>&1`
3. Observe the logs in `trace.txt` and notice that `git-lfs` is constantly retrying

In the `trace.txt` below you can see the `Retry-After` number counting down, clearly indicating that its not waiting for the rate limit to pass.

Retrying with `git clone -c lfs.concurrenttransfers=1` or `-c lfs.concurrenttransfers=1 -c lfs.transfer.batchSize=1000` did not improve the situation.

**Expected behavior**
When a **HTTP 429** is received by a single worker and it contains the `Retry-After` response header with a number, `git-lfs` should wait for the specified amount of time and therefore respect `Retry-After` sent by the server.

**System environment**
Observed on Linux x86_64 (Fedora) and macOS 15.3.1 with `git-lfs`

**Output of `git lfs env`**

```
git version 2.48.1

Endpoint=https://githost.com/example-project.git/info/lfs (auth=none)
  SSH=git@githost.com:example-project.git
LocalWorkingDir=/Users/max/Documents/example-project
LocalGitDir=/Users/max/Documents/example-project/.git
LocalGitStorageDir=/Users/max/Documents/example-project/.git
LocalMediaDir=/Users/max/Documents/example-project/.git/lfs/objects
LocalReferenceDirs=
TempDir=/Users/max/Documents/example-project/.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/Users/max/Documents/example-project/.git/lfs
AccessDownload=none
AccessUpload=none
```

**Additional context**

**LFS Files downloaded (git lfs ls-files | wc -l)**: **652**
**Number of HTTP 429 received during `git clone` operation**: **925**

Related to https://github.com/git-lfs/git-lfs/issues/3122

[trace.txt](https://github.com/user-attachments/files/19126218/trace.txt)

chrisd8088:
This PR resolves the problem reported in #6007 regarding the Git LFS client's failure to honour the delay interval or fixed retry time specified in `Retry-After` headers, if the headers are sent with 429 HTTP response status codes in reply to individual Git LFS object transfer requests.

In addition, we revise a number of related tests so they now check the number of retry attempts made after 429 status code responses to both Batch API requests and individual object transfer requests, and then add the same type of checks to the appropriate [tests](https://github.com/git-lfs/git-lfs/blob/41999a064abfc4c30ff6864da6bab8fad32d3536/t/t-batch-storage-retries-ratelimit.sh#L5-L103) to verify that the bug fix in this PR is effective.

As well, we fix [several](https://github.com/git-lfs/git-lfs/blob/41999a064abfc4c30ff6864da6bab8fad32d3536/t/t-batch-retries-ratelimit.sh#L62-L132) [tests](https://github.com/git-lfs/git-lfs/blob/41999a064abfc4c30ff6864da6bab8fad32d3536/t/t-batch-storage-retries-ratelimit.sh#L72-L103) of our 429 status code response handling during `git lfs clone` operations.  At present, these tests only receive responses from our test server with 429 status codes and `Retry-After` headers when pushing objects to establish the test conditions, and not when performing the subsequent clone operations the tests are intended to validate.

Resolves #6007.
/cc @max-wittig as reporter.

### DIFF
--- a/tq/transfer_queue.go
+++ b/tq/transfer_queue.go
@@ -251,6 +251,7 @@ type objectTuple struct {
 	Size            int64
 	Missing         bool
 	ReadyTime       time.Time
+	retryLaterTime  time.Time
 }
 
 func (o *objectTuple) ToTransfer() *Transfer {
@@ -546,7 +547,10 @@ func (q *TransferQueue) enqueueAndCollectRetriesFor(batch batch) (batch, error) 
 	enqueueRetry := func(t *objectTuple, err error, readyTime *time.Time) {
 		count := q.rc.Increment(t.Oid)
 
-		if readyTime == nil {
+		if !t.retryLaterTime.IsZero() {
+			t.ReadyTime = t.retryLaterTime
+			t.retryLaterTime = time.Time{}
+		} else if readyTime == nil {
 			t.ReadyTime = q.rc.ReadyTime(t.Oid)
 		} else {
 			t.ReadyTime = *readyTime
@@ -809,7 +813,7 @@ func (q *TransferQueue) handleTransferResult(
 
 			if ok {
 				t := objects.First()
-				t.ReadyTime = readyTime
+				t.retryLaterTime = readyTime
 				retries <- t
 			} else {
 				q.errorc <- res.Error
--- a/tq/transfer_queue.go
+++ b/tq/transfer_queue.go
@@ -806,7 +806,7 @@ func (q *TransferQueue) handleTransferResult(
 			// If the object can't be retried now, but can be
 			// after a certain period of time, send it to
 			// the retry channel with a time when it's ready.
-			tracerx.Printf("tq: retrying object %s after %s seconds.", oid, time.Until(readyTime).Seconds())
+			tracerx.Printf("tq: retrying object %s after %.2fs", oid, time.Until(readyTime).Seconds())
 			q.trMutex.Lock()
 			objects, ok := q.transfers[oid]
 			q.trMutex.Unlock()
--- a/t/cmd/lfstest-gitserver.go
+++ b/t/cmd/lfstest-gitserver.go
@@ -276,13 +276,18 @@ var (
 	laterRetriesMu  sync.Mutex
 )
 
+func getResourceKey(api, direction, repo, oid string) string {
+	return strings.Join([]string{api, direction, repo, oid}, ":")
+}
+
 // checkRateLimit tracks the various requests to the git-server. If it is the first
 // request of its kind, then a times is started, that when it is finished, a certain
 // number of requests become available.
 func checkRateLimit(api, direction, repo, oid string) (seconds int, isWait bool) {
 	laterRetriesMu.Lock()
 	defer laterRetriesMu.Unlock()
-	key := strings.Join([]string{direction, repo, oid}, ":")
+
+	key := getResourceKey(api, direction, repo, oid)
 	if requestsRemaining, ok := requestTokens[key]; !ok || requestsRemaining == 0 {
 		if retryStartTimes[key] == (time.Time{}) {
 			// If time is not initialized, set it to now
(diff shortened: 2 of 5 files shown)


max-wittig:
@chrisd8088 I don't know the codebase to well to be of help here, but if I should test something let me know!

chrisd8088:
> I don't know the codebase to well to be of help here, but if I should test something let me know!

Thanks, @max-wittig!  It may be a while before I'm able to move this out of draft mode and get a review, but in the meantime, if you want to download the appropriate CI job artifacts for your platform (maybe the macOS [ones](https://github.com/git-lfs/git-lfs/actions/runs/13829402745/artifacts/2743503540)?) and give the `git-lfs` binary a try in your environment, that would be much appreciated!  The CI job summary [page](https://github.com/git-lfs/git-lfs/actions/runs/13829402745) has an "Artifacts" section down at the bottom-right, and you can pick the appropriate Zip archive from there.

tq/transfer_queue.go
```diff
@@ -546,7 +547,10 @@ func (q *TransferQueue) enqueueAndCollectRetriesFor(batch batch) (batch, error)
 	enqueueRetry := func(t *objectTuple, err error, readyTime *time.Time) {
 		count := q.rc.Increment(t.Oid)
 
-		if readyTime == nil {
+		if !t.retryLaterTime.IsZero() {
+			t.ReadyTime = t.retryLaterTime
```
larsxschneider, line 551: What happens if `retryLaterTime` is far in the future? Would `git-lfs` seem to hang? Should we restrict this to something sane like a maximum wait time of 10min or something?

larsxschneider (approved):

tq/transfer_queue.go
chrisd8088: That's a good question.  We already respect the `Retry-After` header when handling Batch API responses, so the scenario you describe is possible now, even without this PR's changes.

If we want to guard against excessive `Retry-After` settings, I think we should probably make that enhancement in another PR, since this one is intended to address the bug in #6007 in particular.  We probably also want to allow for some kind of configuration of whatever cutoff we establish, so users can choose to allow longer waits than our default maximum of, say, 10 minutes, or even to just always honour any `Retry-After` setting if they want.  And of course we'd want to expand our test suite to validate this new functionality with both Batch API and individual object transfer requests.

As for an actual implementation of such a maximum on `Retry-After` values, we could insert it at the [end](https://github.com/git-lfs/git-lfs/blob/d0e9d610c771a468cab62d1ed67121e18736bbef/tq/transfer_queue.go#L1046) of the `TransferQueue.canRetryObjectLater()` method, so it checks and limits the time value it otherwise simply returns, or, alternatively, check and limit the time value [calculated](https://github.com/git-lfs/git-lfs/blob/d0e9d610c771a468cab62d1ed67121e18736bbef/errors/types.go#L418-L432) when we parse the `Retry-After` itself (noting that there are two valid `return` paths in the `NewRetriableLaterError()` function, so we'd need to refactor it so both passed through the new code that implemented the limit).

All of which is to say, I think I'll go ahead and merge this PR, but make a note that we might want to revisit this question in a subsequent PR.

chrisd8088: Fortunately, this issue has now been addressed by PR #6304.

PR #3244, opened 2018-09-11
Create new lfshttp package

PastelMobileSuit:
This PR extracts much of the lower level, http request-handling functionality out of the `lfsapi` package and into a new `lfshttp` package. 

The first commit (d101bdb) largely just extracts files into the new package with minimal changes, and updates `lfsapi` and other packages to rely on it. The second commit (e6c4c63) refactors the way redirects are handled so that `lfsapi` can re-authenticate on a redirect when the original request was authenticated.

PR #6018, opened 2025-03-19
transform DoWithAuth from recurrent to loop approach to prevent endless authentication attepmpts

manturovDan:
In case of wrong credentials, the method DoWithAuth is executed recursively without termination. This is not such a big problem when users interact with LFS via the command line, but it may cause the process to hang when the command is called automatically (e.g. by CI tool) with credential.helper and core.askpass arguments

### DIFF
--- a/lfsapi/auth.go
+++ b/lfsapi/auth.go
@@ -25,21 +25,20 @@ var (
 // supporting basic authentication.
 func (c *Client) DoWithAuth(remote string, access creds.Access, req *http.Request) (*http.Response, error) {
 	count := 0
-	res, err := c.doWithAuth(remote, &count, access, req, nil)
-
-	if errors.IsAuthError(err) {
-		if len(req.Header.Get("Authorization")) == 0 {
-			// This case represents a rejected request that
-			// should have been authenticated but wasn't. Do
-			// not count this against our redirection
-			// maximum.
-			newAccess := c.Endpoints.AccessFor(access.URL())
-			tracerx.Printf("api: http response indicates %q authentication. Resubmitting...", newAccess.Mode())
-			return c.DoWithAuth(remote, newAccess, req)
+	for {
+		res, err := c.doWithAuth(remote, &count, access, req, nil)
+		if !errors.IsAuthError(err) || len(req.Header.Get("Authorization")) != 0 || count == defaultMaxAuthAttempts {
+			return res, err
 		}
-	}
 
-	return res, err
+		// This case represents a rejected request that
+		// should have been authenticated but wasn't. Do
+		// not count this against our redirection
+		// maximum.
+		newAccess := c.Endpoints.AccessFor(access.URL())
+		tracerx.Printf("api: http response indicates %q authentication. Resubmitting...", newAccess.Mode())
+		count++
+	}
 }
 
 // DoWithAuthNoRetry sends an HTTP request to get an HTTP response. It works in
@@ -86,9 +85,7 @@ func (c *Client) doWithAuth(remote string, count *int, access creds.Access, req 
 
 			if credWrapper.Creds != nil {
 				req.Header.Del("Authorization")
-				if multistage && *count < defaultMaxAuthAttempts && res != nil && res.StatusCode == 401 {
-					*count++
-				} else {
+				if !multistage || *count >= defaultMaxAuthAttempts || res == nil || res.StatusCode != 401 {
 					credWrapper.CredentialHelper.Reject(credWrapper.Creds)
 				}
 			}


chrisd8088:
Hey, thanks very much for this bug report and PR!

This looks like a plausible fix for the problem, but I've only taken a brief look so far.  Unfortunately we have a small queue of PRs awaiting review, so it might take me a while before I can give this one its proper due.  In the meantime I've started our CI suite, and it looks like there are some Go tests that are failing and so those will need to be reconsidered, at a minimum.  (The Go 1.22 job is also failing right now for an unrelated reason, so don't worry about that one.)

In addition to the Go tests, it would be great if we can add some shell tests to whichever of our `t/t-askpass.sh`, `t/t-credentials.sh`, and `t/t-credentials-no-prompt.sh` test scripts are appropriate.

As a start in that direction, would you be able to provide a reproduction test case here in a comment, at least?  That would really help us understand how we can reproduce the problem in our test suite, so that we can in turn demonstrate that this PR's changes have the intended effect of resolving the issue.

Thank you again for taking the time to contribute to our project and helping to improve Git LFS for all our users!

chrisd8088:
Thank you very much for the additional trace logs and the reproduction case and script; those will help a lot!

We've got a few PRs to review at the moment, each of which happens to involve somewhat non-trivial changes (including this PR), so I apologize again that it may take us a little while longer to give this PR a careful look.

Thank you again for investigating this problem and for writing up this PR, and for your continued patience!

chrisd8088:
Thanks again for reporting this bug and providing the logs and a PR!  I spent some time looking at the issues here and we appear to have a number of similar kinds of potential problems, and a lack of tests that would expose them.

The first thing I notice is that your Nexus server is presumably replying with a 401 status code when it receives invalid credentials, rather than a 403, and that's not a behaviour we handle well because the Git LFS client just keeps sending the same credentials again.  Arguably that's an issue with the Nexus server, but for our purposes, it reveals a bug on our side that we need to fix.

Looking back over the history of this code, it seems possible to me that we've never had adequate protections against a server which continued to return 401 status codes.

#### Replication in Tests

We can simulate the problem you've encountered in a various ways in our test suite.  In the `TestDoWithAuthApprove()` Go test function, replacing the [conditional](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth_test.go#L72-L81) lines in the test server code with a single unconditional `w.WriteHeader(http.StatusUnauthorized)` causes the test to never complete.

Or if we change our `lfstest-gitserver` utility's `skipIfBadAuth()` function to [return](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/t/cmd/lfstest-gitserver.go#L1731) a 401 instead of a 403, a number of shell tests enter an infinite loop.

A more subtle kind of reproduction case occurs in our `credentials can authenticate with multistage auth` test if we change the local credential data we [create](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/t/t-credentials.sh#L331) for our `git-credential-lfstest` helper so it causes the client to keep sending the same first-stage credentials for the second stage of a multi-stage authorization process.  To do this we alter the helper's data like this:
```diff
@@ -328,7 +328,7 @@ begin_test "credentials can authenticate with multistage auth"
   reponame="auth-multistage-token"
   setup_remote_repo "$reponame"
 
-  printf 'Multistage::cred2:state1:state2:\nMultistage::cred1::state1:true' > "$CREDSDIR/127.0.0.1--$reponame"
+  printf 'Multistage::cred1:state1:state1:\nMultistage::cred1::state1:true' > "$CREDSDIR/127.0.0.1--$reponame"
 
   clone_repo "$reponame" "$reponame"
   git checkout -b new-branch
```

Yet another, different reproduction case can be triggered by making the following change to our `download authenticated object` [test](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/t/t-object-authenticated.sh#L8-L23), which results in the test never completing:

```diff
@@ -8,10 +8,12 @@ ensure_git_version_isnt $VERSION_LOWER "2.3.0"
 begin_test "download authenticated object"
 (
   set -e
-  reponame="$(basename "$0" ".sh")"
+  reponame="requirecreds-$(basename "$0" ".sh")"
   setup_remote_repo "$reponame"
   clone_repo "$reponame" without-creds
 
+  printf ":requirecreds:pass" > "$CREDSDIR/127.0.0.1--$reponame"
+
   git lfs track "*.dat"
   printf "object-authenticated" > hi.dat
   git add hi.dat
```

#### Recursion in HTTP Handling

The penultimate test mentioned above is interesting because it dates from PR #5803, and in commit 67830784dbdc54c7fddf926653c18c4097124a8f of that PR we added the `count` [variable](67830784dbdc54c7fddf926653c18c4097124a8f) to the `DoWithAuth()` method of the `lfsapi` package's `Client` structure, along with the [conditional](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L89-L90) which tries to increment that value on repeated multi-stage authorization requests.  The commit description notes that this check was intended "to avoid a credential helper getting stuck in an infinite loop if it keeps handing back the same credentials."  However, if a server returns 401, the `doWithCreds()` method will [return](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L118-L120) with a non-`nil` error, and although that will cause `count` to be [incremented](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L90), that new value will simply be discarded when `doWithAuth()` [returns](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L103) to `DoWithAuth()`.  That method will then [recurse](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L38) and call itself again, still with a zero value for `count`, as you noticed.  Meanwhile, if the server returns a recognized 3xx code, `doWithCreds()` will not return with an error but instead [call](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L127) `doWithAuth()` again, and while that will pass the `count` value through the recursive call chain, the value will still be zero.

(_UPDATE:_ The description above is largely correct, but omits the detail that the `count` variable is [passed](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L28) as a pointer to an integer, and so when the `doWithAuth()` method [increments](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L90) it, the value is actually updated in a variable in the scope of the calling `DoWithAuth()` method.  However, when that method [recurses](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L38), its next invocation [instantiates](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L27) another `count` variable local to that level of the calling stack, and so the incremented value is effectively ignored.)

On this point, we have two levels of potential recursion: `DoWithAuth()` calls itself if it wants to retry a request after receiving a 401, expecting that the server will send back a 403 if the credentials included in the next request are invalid.  Meanwhile, `doWithAuth()` calls `doWithCreds()` which may call `doWithAuth()` again if a recognized 3xx status code is received, in order to follow an HTTP redirection.

There's also a very similar [recursion](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfshttp/client.go#L375) in the `doWithRedirects()` method of the `lfshttp` package's `Client` structure.  That internal method is [used](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfshttp/client.go#L194) by the `do()` method, which is called by the `Do()` and `DoWithAccess()` methods, both of which are used in specialized conditions where we know the access mode (if any) in advance.

What's interesting about both of these implementations which handle 3xx redirections through recursion is that they suffer from effectively the same problem you reported, where the Git LFS client doesn't give up after a certain number of HTTP requests.  If a server or servers send a chain of 3xx responses which result in an endless loop of redirections, the client will follow them without ceasing, and authentication doesn't even need to be part of the picture in this case.

As for the final test mentioned above, that exercises yet a fourth recursive method, the `makeRequest()` method of the `tq` package's `basicUploadAdapter` structure.  This method handles the HTTP requests used to transfer individual Git LFS objects to a server, and through its [call](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/tq/basic_upload.go#L229) to the `doHTTP()` method, may [invoke](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/tq/adapterbase.go#L229) the `Do()` method from the `lfshttp` package if the server has informed the client that no local credentials need to be sent with the object transfer request.  The server does this by returning an `authenticated: true` [element](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/docs/api/batch.md?plain=1#L141-L143) in its Batch API response.  However, by modifying the test as described above, we cause our test server to both send this JSON element but also actually require credentials for the individual object transfers, so it sends a 401 status code to the `PUT` request for the object, which causes the `makeRequest()` method to [call](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/tq/basic_upload.go#L240) itself without making any changes, and so the cycle repeats endlessly.

#### Recursion Limits

Intriguingly, we appear to have once have had functioning protection against following an endless loop of HTTP redirection, with some logic originally added in PR #1791 and then simplified in PR #1839.  The `doWithRedirects()` method of the `Client` structure in the `lfsapi` package [accepted](https://github.com/git-lfs/git-lfs/blob/ac9ad7d8175a181fdd2a6ffab2bf07f9c3938080/lfsapi/client.go#L78) a slice of `http.Request` objects named `via`, and if the current HTTP request received a 307 status code, the method would [append](https://github.com/git-lfs/git-lfs/blob/ac9ad7d8175a181fdd2a6ffab2bf07f9c3938080/lfsapi/client.go#L104-L107) the current `http.Request` to the slice and then return an error if the `via` slice had grown to three consecutive requests.  Only if fewer requests than three had been made so far would the method [recurse](https://github.com/git-lfs/git-lfs/blob/ac9ad7d8175a181fdd2a6ffab2bf07f9c3938080/lfsapi/client.go#L114) by calling itself, passing in the `via` slice.

In PR #3028 this logic was updated so as to distinguish requests where the response was a 401 status code and not include those in count of redirections, and also avoid sending the `Authorization` header from a prior request to another server after a redirection, which is an important security measure.

In the latter situation the `via` slice was retained and [passed](https://github.com/git-lfs/git-lfs/blob/37bc00aa3e5b222f0df527291337d2b6f1e83645/lfsapi/client.go#L239) when `doWithRedirects()` [called](https://github.com/git-lfs/git-lfs/blob/37bc00aa3e5b222f0df527291337d2b6f1e83645/lfsapi/client.go#L239) `doWithAuth()` to follow the redirection but ignore any previously authenticated credentials.  This meant the length of the slice increased with each redirection.

However, when `doWithAuth()` processed a 401 status code and initiated another request to be made with credentials, it [called](https://github.com/git-lfs/git-lfs/blob/37bc00aa3e5b222f0df527291337d2b6f1e83645/lfsapi/auth.go#L58) the new `DoWithAuth()` method, which didn't accept a `via` argument and thus restarted the count of HTTP requests.

I believe the changes in PR #3244, when the `lfshttp` package was refactored out of the `lfsapi` package, introduced a subtle bug which caused the `via` slice to never grow during a recursive chain of calls.  In commit e6c4c6386ba8cce6fb3cc31f105d0eefa43b444e, the `doWithRedirects()` method was refactored into a `DoWithRedirect()` method that performed the HTTP request, and a smaller `doWithRedirects()` method which called `DoWithRedirect()` and then, if a redirection was received, recursively [called](https://github.com/git-lfs/git-lfs/blob/e6c4c6386ba8cce6fb3cc31f105d0eefa43b444e/lfshttp/client.go#L334) itself.  As well, the `doWithCreds()` method in the `lfsapi` package was updated to also recursively [call](https://github.com/git-lfs/git-lfs/blob/e6c4c6386ba8cce6fb3cc31f105d0eefa43b444e/lfsapi/auth.go#L87) `doWithAuth()` under the same conditions (i.e., when a redirection was received).  The design of these methods remains essentially the same today.

Unfortunately, in both cases, although the `DoWithRedirect()` method still [appends](https://github.com/git-lfs/git-lfs/blob/e6c4c6386ba8cce6fb3cc31f105d0eefa43b444e/lfshttp/client.go#L311-L314) to the `via` slice it is passed, it does nothing else with the slice, and doesn't pass it to any other methods, whereas it used to pass it back to itself in a recursive call.  Although the other methods which call `DoWithRedirect()` pass their `via` slice arguments along, Go passes slices by value, or at least, their internal header structures are passed by value.  If the `DoWithRedirect()` method changed the existing values in the array pointed to by the slice, those changes would be propagated through the recursion, but because it just appends to the slice, which only affects the slice header (and maybe causes a new backing array to be allocated and populated), the added value at the end of the `via` slice is lost at the end of the method.

Hence the intended [limit](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfshttp/client.go#L351) of at most three HTTP redirections is not respected.

Meanwhile, as noted above, a different attempt at limiting the number of multi-stage authentication requests was added in PR #5803, but that doesn't seem to be as effective as we might like.  It's definitely possible to set up multi-stage authorization which exceeds the [limit](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L20) of three authorization attempts.  That might be acceptable, of course, but presumably we do want to limit the number of times we send identical `Authorization` headers to the same server—which is exactly the situation I think this PR aims to resolve.

<a id="comprehensive-mitigation"></a>
#### Comprehensive Mitigation

One concern which all of foregoing raises, I think, is how we should handle these various different situations, as well as a few others which occurred to me.

For one thing, the code [comment](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L32-L35) in the `DoWithAuth()` method is a remnant of the logic added in commit 5257e587e4f8b0f03b5a86cb0da9baae8522883f of PR #3028, but no longer reflects the behaviour of that method.  Originally, the comment signalled that after receiving a 401, the `via` slice would effectively be [reset](https://github.com/git-lfs/git-lfs/blob/37bc00aa3e5b222f0df527291337d2b6f1e83645/lfsapi/auth.go#L58) to `nil`, thus resetting the count of the number of consecutive HTTP redirections the client had followed.  But at the moment, there's no longer any redirection handling or counting in the `DoWithAuth()` method, since it doesn't even receive a `via` slice which it should ignore under specific circumstances.

We've discussed a few potential infinite loop conditions above, including:
- Server(s) send a looping chain of HTTP redirections.
- Server sends a 401 response after receiving invalid credentials, causing the client to keep resending them.
- Server claims an object transfer `PUT` request requires no local credentials but then requests them.  (Right now this causes the client to loop, but it should either just report an error or try sending the credentials.)
- Git credential helper returns credentials which cause a loop during multi-stage authentication.

Another, more challenging condition to consider might be if a series of servers send a mixed series of 3xx and 401 responses which ultimately cause a loop.  For instance, server A first replies with a 401, causing the client to look up credentials and send them, at which point the server sends with a 307 sending the client to server B.  Server B then requests credentials with a 401, after which it replies with a 307 sending the client back to server A.  Ideally, the client should cache the credentials it retrieves for both servers and send them without being prompted by a 401, at which point the servers' responses should devolve into a chain of redirections, which is just the first condition in the list above.

I think we'll have to address these problems in a series of PRs, rather than just one or even two.  As for this PR, tackling the problem of repeated 401 responses and maybe also making the multi-stage authentication limits function as they were intended to is lots for one PR!  We do, however, need to add some fairly extensive tests in both our Go and shell test suites to prove the fixes are effective.  As well, since we may not want to limit valid multi-stage authentication to three stages, we probably either need to make the limit configurable, or find a way to detect when we're resending the same `Authorization` header to the same URL and only increment the count in this case.  That might actually be valuable in the more common non-multi-stage situation as well.

Finally, thank you again very much for the bug report and proposed code changes!  Analyzing this problem has certainly produced some interesting findings.

chrisd8088:
As an addendum to my comments above regarding the code from PR #5803, I should note that it is possible for the `count` variable to be [incremented](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L89-L90) within a chain of `doWithAuth()` and `doWithCreds()` method calls, but I believe it requires a relatively unusual set of conditions involving HTTP redirections.

Moreover, because the increment occurs after `doWithAuth()` calls `doWithCreds()`, which then recursively calls `doWithAuth()` to handle an HTTP redirection, the `count` variable is increased only as the chain of these recursive calls is unwound.  Therefore the [check](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L64-L66) at the start of `doWithAuth()` is never exercised in this case, and so we never see a `too many authentication attempts` error generated.  All that happens is that if the `defaultMaxAuthAttempts` limit is reached while unwinding the series of calls, some of them will [discard](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L92) the credentials they retrieved in order to make their requests.

I believe that check was intended to guard against a different form of recursive call, where `DoWithAuth()` [invokes](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L38) itself in order to generate another HTTP request after a server's 401 response indicates that credentials are required.  But as we have noted above, `DoWithAuth()` always [instantiates](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L27) a new `count` variable and set it to zero, so when it calls `doWithAuth()` that value is always zero and thus the `too many authentication attempts` error is never output.

---

To exercise the [check](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L89-L90) as it stands now, we can simulate the following sequence of requests and responses.  First, a server replies to an unauthenticated request with a 401, which causes a single recursion through the `DoWithAuth()` method.  The client retrieves a local credential, one which we have already defined as the first stage of a multi-stage authentication process, and makes a second request, to which the server then replies with a 307 and a redirection to a URL with a different hostname.

This causes the client to make a recursive [call](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L127) back to `doWithAuth()` from `doWithCreds()`, but with a new request from which the `Authorization` header has been [removed](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfshttp/client.go#L563-L567).  However, the client has not reset its awareness that an authentication sequence is in progress, and so when `getCreds()` [runs](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L70) it retrieves the appropriate credentials for the new hostname from the local Git credential helper, which again we have defined in advance as the second stage of a multi-stage authentication process.  The client makes a request to the new server, and the server then replies with a 307 response which redirects the client to a URL with the original hostname (or a third hostname).

The client therefore makes another recursive call to `doWithAuth()` from `doWithCreds()`, and will repeat this sequence as long as the server keeps replying with redirections.  Note that each `doWithAuth()` call in the stack has not advanced past the `doWithCreds()` [call](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L76) and so none of the subsequent [logic](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L77-L103) has yet run (except after the very first request, when a 401 was received, and `doWithAuth()` returned back to `DoWithAuth()`, which then called itself to start the authentication sequence).  Because none of this subsequent logic in `doWithAuth()` has run for any of the calls on the stack, the `count` variable has never been [incremented](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L90), so the [check](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L64) at the start of the method has only ever seen a `count` value of zero, no matter how many redirections occur.

Eventually, once the server returns a non-redirection status code, the final `doWithCreds()` and `doWithAuth()` pair return, and then all the others do as well, unwinding the recursion chain.  During this process, if the final status code sent by the server was another 401, then the `errors.IsAuthError(err)` [conditional](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L78) will be true in every instance as the `doWithAuth()` methods complete the logic which follows their calls to `doWithCreds()`.  As a result, the `count` variable will finally be incremented, and if the stack has enough levels, the check against the `defaultMaxAuthAttempts` maximum will fail, and the Git credential helper will be [executed](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/lfsapi/auth.go#L92) with a `reject` argument for the credentials of the hostname from that particular level in the chain of redirections.

While interesting, I don't think that behaviour is what was intended when the multi-stage authentication support was added, so we'll have to consider what the behaviour should be instead as part of this PR, and write some tests to verify that whatever revisions we make perform as we expect.

---

For my own future reference, in our shell test suite we can establish a sufficiently long multi-stage authentication sequence on the client side using test credential record files with content like the following:
```
Multistage::cred5:state4:state5:true
Multistage::cred4:state3:state4:true
Multistage::cred3:state2:state3:true
Multistage::cred2:state1:state2:true
Multistage::cred1::state1:true
```
We then need to create appropriate copies for each of the hostnames involved, so that our `git-credential-lfstest` helper program will find them.  In our test server, we then need to expand the `Multistage` authorization [case](https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/t/cmd/lfstest-gitserver.go#L1719-L1728) to return 307s with `Location` headers referring to at least one hostname other than the default, and add additional handlers to accept the redirected requests from the client.

chrisd8088 (commented):
Thanks very much again for identifying this problem and tackling a potential solution!

While doing my own investigation and code analysis I've turned up a few related issues, and used this PR to make a bunch of notes for the future—my apologies for that.

I've written up a first set of review comments and suggestions, which I think should further simplify our code and also resolve the immediate Go test failures seen in the CI jobs for this PR at the moment.  For clarity, I'll include my final suggested versions of the three key methods below.

We will, though, still need to work on adding some Go and shell tests which exercise the authorization request limit, as we don't have anything right now.  If we did, we'd have noticed the potential for endless recursion!  So we'll have to collaborate on a set of tests to demonstrate the effectiveness of the new code.

Again, thank you for reporting this issue and finding a solution!

---

Here is my suggested version of the `DoWithAuth()` method:
```go
func (c *Client) DoWithAuth(remote string, access creds.Access, req *http.Request) (*http.Response, error) {
	for range defaultMaxAuthAttempts {
		res, err := c.doWithAuth(remote, access, req, nil)
		if err == nil || !errors.IsAuthError(err) {
			return res, err
		}

		// We expect this condition should never occur because
		// doWithAuth() will delete any Authorization header
		// after a 401 status code is received.
		if len(req.Header.Get("Authorization")) != 0 {
			return res, err
		}

		access = c.Endpoints.AccessFor(access.URL())
		tracerx.Printf("api: http response indicates %q authentication. Resubmitting...", access.Mode())
	}

	return nil, fmt.Errorf("too many authentication attempts")
}
```

Next, we can remove `count` from the call to `doWithAuth()` in the `DoWithAuthNoRetry()` method:

```go
func (c *Client) DoWithAuthNoRetry(remote string, access creds.Access, req *http.Request) (*http.Response, error) {
	return c.doWithAuth(remote, access, req, nil)
}
```

We can remove the `count` variable from the `doWithAuth()` method, and also eliminate one level of indentation by combining two conditions that go well together:
```go
unc (c *Client) doWithAuth(remote string, access creds.Access, req *http.Request, via []*http.Request) (*http.Response, error) {
	req.Header = c.client.ExtraHeadersFor(req)

	credWrapper, err := c.getCreds(remote, access, req)
	if err != nil {
		return nil, err
	}
	c.credContext.SetStateFields(credWrapper.Creds["state[]"])

	res, err := c.doWithCreds(req, credWrapper, access, via)
	if err != nil && errors.IsAuthError(err) {
		multistage := credWrapper.Creds.IsMultistage()
		newMode, newModes, headers := getAuthAccess(res, access.Mode(), c.access, multistage)
		newAccess := access.Upgrade(newMode)
		if newAccess.Mode() != access.Mode() {
			c.Endpoints.SetAccess(newAccess)
			c.access = newModes
		}

		if credWrapper.Creds != nil {
			req.Header.Del("Authorization")
			if !multistage {
				credWrapper.CredentialHelper.Reject(credWrapper.Creds)
			}
		}
		c.credContext.SetWWWAuthHeaders(headers)
	}

	if res != nil && res.StatusCode < 300 && res.StatusCode > 199 {
		credWrapper.CredentialHelper.Approve(credWrapper.Creds)
	}

	return res, err
}
```

Lastly, the `count` variable can also be dropped from the `doWithCreds()` method:
```go
func (c *Client) doWithCreds(req *http.Request, credWrapper creds.CredentialHelperWrapper, access creds.Access, via []*http.Request) (*http.Response, error) {
	if access.Mode() == creds.NegotiateAccess {
		return c.doWithNegotiate(req, credWrapper)
	}

	req.Header.Set("User-Agent", lfshttp.UserAgent)

	client, err := c.client.HttpClient(req.URL, access.Mode())
	if err != nil {
		return nil, err
	}

	redirectedReq, res, err := c.client.DoWithRedirect(client, req, "", via)
	if err != nil || res != nil {
		return res, err
	}

	if redirectedReq == nil {
		return res, errors.New(tr.Tr.Get("failed to redirect request"))
	}

	return c.doWithAuth("", access, redirectedReq, via)
}
```

manturovDan:
Hi @chrisd8088, thank you for your help. Do you have any updates related to this issue?

chrisd8088:
Thank you again for the PR and for identifying this issue!

> Do you have any updates related to this issue?

The main first steps, I think, are the ones I identified in my code review.  Would you have time to look at those and adjust the PR accordingly?  I put the revised versions of the key methods into this [comment](https://github.com/git-lfs/git-lfs/pull/6018#pullrequestreview-2762083712) above, if that helps.

Once we've made those changes to this PR, we'll also need to expand the test suite, as I noted in that comment.  But as a first step, adopting the revisions from my PR review should at least allow the CI suite to pass.

If you don't have time to look at my suggestions, that's fine; I expect to circle back to this PR and several others in the not-too-distant future, and can draft my own version of this PR which would then replace yours.  If you're able to help in the meantime, though, that would of course be appreciated.

Either way, though, thank you again for pinpointing the problem and drafting a first potential solution!

tamaracvjetkovic:
Hi @chrisd8088!
I'm Tamara, here on behalf of @manturovDan to assist with this. :)

> Would you have time to look at those and adjust the PR accordingly?

Of course, we will take care of adjusting the PR.

Also, thank you so much for the detailed comments and explanations above, they are extremely helpful and very much appreciated!

chrisd8088:
Thank you very much @tamaracvjetkovic!  Just let me know if anything in my review comments and suggested revisions is unclear!

lfsapi/auth.go
```diff
@@ -25,21 +25,20 @@ var (
...
-	if errors.IsAuthError(err) {
-		if len(req.Header.Get("Authorization")) == 0 {
-			// This case represents a rejected request that
-			// should have been authenticated but wasn't. Do
-			// not count this against our redirection
-			// maximum.
-			newAccess := c.Endpoints.AccessFor(access.URL())
-			tracerx.Printf("api: http response indicates %q authentication. Resubmitting...", newAccess.Mode())
-			return c.DoWithAuth(remote, newAccess, req)
+	for {
```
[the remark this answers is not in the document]
tamaracvjetkovic: I agree, removing the count variable makes everything easier to read! 

In the latest commit, I removed the count variable entirely. I also wanted to use the `for...range` loop as I think it looks more readable, but because the module targets Go 1.21 and `for...range` loop feature requires Go 1.22 or later, I went with a classic counted for loop instead.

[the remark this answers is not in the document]
chrisd8088: We should [require](https://github.com/git-lfs/git-lfs/blob/469c8da0478f4bee02836f6dee12bce9fe737d2b/go.mod#L48) Go 1.23 at a minimum now, so feel free to use any feature that's available in 1.23 or higher, including the `for...range` feature.

lfsapi/auth.go
```diff
@@ -25,21 +25,20 @@ var (
...
+		res, err := c.doWithAuth(remote, &count, access, req, nil)
+		if !errors.IsAuthError(err) || len(req.Header.Get("Authorization")) != 0 || count == defaultMaxAuthAttempts {
+			return res, err
 		}
```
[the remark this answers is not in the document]
tamaracvjetkovic: I've removed the unnecessary conditions, and added the additional check for `err == nil`. 

I also kept the `len(req.Header.Get("Authorization")) != 0` check in `DoWithAuth` as a separate condition after `err == nil || !errors.IsAuthError(err)`. I think it's okay to keep it maybe for extra safety, even though I also think it shouldn't be non-empty in the scenario you mentioned.

lfsapi/auth.go
```diff
@@ -25,21 +25,20 @@ var (
...
 
-	return res, err
+		// This case represents a rejected request that
+		// should have been authenticated but wasn't. Do
+		// not count this against our redirection
+		// maximum.
+		newAccess := c.Endpoints.AccessFor(access.URL())
+		tracerx.Printf("api: http response indicates %q authentication. Resubmitting...", newAccess.Mode())
+		count++
+	}
```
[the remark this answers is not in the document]
tamaracvjetkovic: The check has been removed (along with the count variable), and I added the error return as you suggested!

lfsapi/auth.go
```diff
@@ -25,21 +25,20 @@ var (
...
+		if !errors.IsAuthError(err) || len(req.Header.Get("Authorization")) != 0 || count == defaultMaxAuthAttempts {
+			return res, err
 		}
-	}
 
-	return res, err
+		// This case represents a rejected request that
+		// should have been authenticated but wasn't. Do
+		// not count this against our redirection
+		// maximum.
```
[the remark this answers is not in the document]
tamaracvjetkovic: Done!

lfsapi/auth.go
```diff
@@ -25,21 +25,20 @@ var (
...
+		newAccess := c.Endpoints.AccessFor(access.URL())
+		tracerx.Printf("api: http response indicates %q authentication. Resubmitting...", newAccess.Mode())
```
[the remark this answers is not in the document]
tamaracvjetkovic: Done, thank you very much for pointing that out!

lfsapi/auth.go
```diff
@@ -86,9 +85,7 @@ func (c *Client) doWithAuth(remote string, count *int, access creds.Access, req
 
 			if credWrapper.Creds != nil {
 				req.Header.Del("Authorization")
-				if multistage && *count < defaultMaxAuthAttempts && res != nil && res.StatusCode == 401 {
-					*count++
-				} else {
+				if !multistage || *count >= defaultMaxAuthAttempts || res == nil || res.StatusCode != 401 {
```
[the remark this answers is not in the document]
tamaracvjetkovic: Done! Now there is only check for `!multistage`. Thank you very much for the detailed description, it helped a lot and made the code easier to follow. :)

lfsapi/auth.go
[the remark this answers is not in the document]
tamaracvjetkovic: Thank you, I overlooked that this is quite an old PR. I will use the `for...range` then!

tamaracvjetkovic:
Hi @chrisd8088!

In the previous commits, I included the changes you suggested, thank you very much for the helpful comments!
Could you, please, approve the workflow so we can verify that everything is correct?

I tried running the GitHub Actions workflow in a fork of git-lfs, and two jobs failed on both the PR branch and the latest main, so I don't believe the failures are related to these changes.

I also added some Go tests that should target the 401 retry loop you mentioned.

tamaracvjetkovic:
Also, regarding [this](https://github.com/git-lfs/git-lfs/pull/6018#:~:text=A%20more%20subtle,b%20new%2Dbranch) reproduction case - it makes sense that the issue would happen if the client keeps sending the same credentials for the second stage of the multi-stage auth process. However, when I ran the test with the changes you mentioned (on the latest main branch), I didn't see the behavior you described, and the test passed?

I also tried to reproduce the same behavior by modifying the shell test on my own, but I couldn't get the expected failure.

Related to that, I wrote a [Go test](https://github.com/git-lfs/git-lfs/pull/6018/files#:~:text=TestDoWithAuthMultistageRetryLimitExceeded) that should simulate a similar scenario by using a custom [helper](https://github.com/git-lfs/git-lfs/pull/6018/files#:~:text=func-,newNonAdvancingMultistageHelper,-()%20*nonAdvancingMultistageHelper) which always returns the same credentials, causing the "stuck" effect. When I run this test in the main branch, it does get stuck, on the PR branch, it passes, so I believe that the test is configured as described and exercises the correct behavior. Please let me know if I'm misunderstanding anything.

tamaracvjetkovic:
One more thing, while writing the tests and thinking about the 401 loop scenario, I was wondering what is the expected behavior if the server returns 401 without any authentication challenge headers (no Lfs-Authenticate and/or WWW-Authenticate)?

In that case, is the behavior to still retry `defaultMaxVerifyAttempts` times, or to return immediately since there is no challenge to respond to? I’m not fully sure how these scenarios are currently handled, it may already work this way and I just overlooked it, but I'm interested and wanted to double check.

chrisd8088:
Thank you very much for all the work on this PR, @tamaracvjetkovic!  I will take a close look as soon as I can.

As a first step, I've kicked off the CI jobs as you requested, so we can see if those pass.

Thank you again so much for picking up this PR and working on the complicated set of issues it exposed!

lfsapi/auth_test.go
```diff
@@ -236,6 +236,135 @@ func TestDoWithAuthNoRetry(t *testing.T) {
...
+			"lfs.url": srv.URL + "/repo/lfs",
+		},
+	))
+	require.Nil(t, err)
+	c.Credentials = cred
+
+	req, err := http.NewRequest("POST", srv.URL+"/repo/lfs/foo", nil)
+	require.Nil(t, err)
+
+	req.Header.Set("Authorization", basicAuth("user", "pass"))
```
chrisd8088, line 313: Thank you for all these great Go tests!

It occurs to me that this specific line might benefit from a comment which explains that by explicitly setting the `Authorization` header here, rather than letting the `getCreds()` method do it for us, we can exercise the "safety" check for a non-empty `Authorization` header in the `DoWithAuth()` method, because otherwise the `doWithAuth()` method will delete the header on each iteration of the loop in `DoWithAuth()`.

This all pertains to the discussion in https://github.com/git-lfs/git-lfs/pull/6018#discussion_r2041265774 regarding commit b3ac11b8d2ea6ef22f74f469ba5e95ab4191f14d; I had to go back to my own notes there to refresh my memory of how some of the pieces fit together in this puzzle.

Specifically, if we don't set `Authorization` header in the request here, then when `doWithAuth()` [calls](https://github.com/git-lfs/git-lfs/blob/071e19e8ea03b1e40b181706909fb8c18d928e29/lfsapi/auth.go#L70) `getCreds()` the first time, it [returns](https://github.com/git-lfs/git-lfs/blob/071e19e8ea03b1e40b181706909fb8c18d928e29/lfsapi/auth.go#L161-L163) without creating an `Authorization` header because the access mode is set to `NoneAccess`.  For the second request, though, the access mode has been set to `BasicAccess`, so `getCreds()` [gathers](https://github.com/git-lfs/git-lfs/blob/071e19e8ea03b1e40b181706909fb8c18d928e29/lfsapi/auth.go#L175-L179) the appropriate credentials (as [returned](https://github.com/git-lfs/git-lfs/blob/071e19e8ea03b1e40b181706909fb8c18d928e29/lfsapi/auth_test.go#L317-L320) by the `mockCredentialHelper`) and populates the request with an `Authorization` header.  However, because the `Creds` element of the `CredentialHelperWrapper` structure type is now non-`nil`, the `doWithAuth()` method will [remove](https://github.com/git-lfs/git-lfs/blob/071e19e8ea03b1e40b181706909fb8c18d928e29/lfsapi/auth.go#L87-L88) the `Authorization` header when a 401 is received from the server, so the "safety" check in our revised `DoWithAuth()` method won't be executed by this test.

Therefore, I'd suggest we add a comment here to help folks like me remember how this all works when we look at it again the future, maybe something like:
```suggestion
	// We set the Authorization header explicitly so as to exercise the
	// check for an extant Authorization header in DoWithAuth().
	//
	// Otherwise an Authorization header will be added automatically for
	// each request subsequent to the first request, since the server
	// always returns 401, but the header will then be removed by
	// doWithAuth() when it receives another 401, so DoWithAuth() will
	// never find an extant Authorization header in any requests.
	req.Header.Set("Authorization", basicAuth("user", "pass"))
```

chrisd8088:
@tamaracvjetkovic — My sincere apologies for my delayed response to your work on this PR, and your questions.  (Getting the v3.7.1 release written and tested consumed more or less all of my available time for the last few months.)

A few quick responses to your questions:

> Also, regarding [this](https://github.com/git-lfs/git-lfs/pull/6018#:~:text=A%20more%20subtle,b%20new%2Dbranch) reproduction case - it makes sense that the issue would happen if the client keeps sending the same credentials for the second stage of the multi-stage auth process. However, when I ran the test with the changes you mentioned (on the latest main branch), I didn't see the behavior you described, and the test passed?

Ah, that's because I made a slight typo, which I've now corrected.  The patch required to make the `credentials can authenticate with multistage auth` test in `t/t-checkout.sh` go into an infinite loop actually should be as follows; note that we need a `Multistage::cred1:state1:state1` line, whereas previously I had a `Multistage::cred2:state1:state1` line, and that's not sufficient:

```patch
--- a/t/t-credentials.sh
+++ b/t/t-credentials.sh
@@ -328,7 +328,7 @@ begin_test "credentials can authenticate with multistage auth"
   reponame="auth-multistage-token"
   setup_remote_repo "$reponame"
 
-  printf 'Multistage::cred2:state1:state2:\nMultistage::cred1::state1:true' > "$CREDSDIR/127.0.0.1--$reponame"
+  printf 'Multistage::cred1:state1:state1:\nMultistage::cred1::state1:true' > "$CREDSDIR/127.0.0.1--$reponame"
 
   git checkout -b new-branch
```

> Related to that, I wrote a [Go test](https://github.com/git-lfs/git-lfs/pull/6018/files#:~:text=TestDoWithAuthMultistageRetryLimitExceeded) that should simulate a similar scenario by using a custom [helper](https://github.com/git-lfs/git-lfs/pull/6018/files#:~:text=func-,newNonAdvancingMultistageHelper,-()%20*nonAdvancingMultistageHelper) which always returns the same credentials, causing the "stuck" effect. When I run this test in the main branch, it does get stuck, on the PR branch, it passes, so I believe that the test is configured as described and exercises the correct behavior. Please let me know if I'm misunderstanding anything.

That's great!  I don't see anything missing there at all.

> One more thing, while writing the tests and thinking about the 401 loop scenario, I was wondering what is the expected behavior if the server returns 401 without any authentication challenge headers (no Lfs-Authenticate and/or WWW-Authenticate)?
>
> In that case, is the behavior to still retry `defaultMaxVerifyAttempts` times, or to return immediately since there is no challenge to respond to? I’m not fully sure how these scenarios are currently handled, it may already work this way and I just overlooked it, but I'm interested and wanted to double check.

RFC 7235 [states](https://www.rfc-editor.org/rfc/rfc7235#section-3.1): "The server generating a 401 response MUST send a `WWW-Authenticate` header field containing at least one challenge applicable to the target resource."

So I think it would be entirely appropriate if we simply stopped retrying after a 401 response without a `WWW-Authenticate` header (and also without our custom `Lfs-Authenticate` header, which is not technically valid according to the HTTP specification, but was added to avoid triggering password prompts in browsers).

However, since our `defaultMaxAuthAttempts` [setting](https://github.com/git-lfs/git-lfs/blob/071e19e8ea03b1e40b181706909fb8c18d928e29/lfsapi/auth.go#L20) is quite low, it doesn't bother me too much if we just retry until that's exhausted.

Either way, it would be ideal if we have a test to check our handling of that case too ... thank you for the thoughtful question!

chrisd8088:
@tamaracvjetkovic — Thanks again for all your work on this PR!  I'd love to merge these changes, as I've left this PR to linger for quite some time (for which I sincerely apologize!)

If by chance you're not free right now to work on it further, I fully understand.  If you're willing to invite me to collaborate on your branch, I can add a few shell tests and the one code comment I suggested in my review.  Otherwise, I'll just leave the PR as it stands for now and circle back to it again once I've reviewed some of the other PRs we have pending.

Thank you again very much for helping to improve Git LFS!

manturovDan:
Hello @chrisd8088, I am sorry for the late response. I have added you to the list of collaborators. Thank you!

chrisd8088:
> I am sorry for the late response. I have added you to the list of collaborators. Thank you!

Thank you very much, @manturovDan!

And I'm the one who should apologize for a late response, not you!  With the v3.7.1 security patch release complete, I'm glad to have some time to circle back to important pending PRs like this one.

Thanks to both you and @tamaracvjetkovic for all the help with this PR!

chrisd8088:
I've added a few commits to the `fix_endless_authentication_attempts` branch of this PR, with the aim that we can get another review from the other member of our core team and merge this PR soon!

For my own future reference, in the "Comprehensive Mitigation" section of my [comment](https://github.com/git-lfs/git-lfs/pull/6018#issuecomment-2796175323) above I made a list of a few potential infinite loop conditions.  This PR now specifically addresses two of them:

- Server sends a 401 response after receiving invalid credentials, causing the client to keep resending them.
- Git credential helper returns credentials which cause a loop during multi-stage authentication.

That leaves two others to be tackled in separate, future PRs:

- Server(s) send a looping chain of HTTP redirections.
- Server claims an object transfer `PUT` request requires no local credentials but then requests them.

lfsapi/auth_test.go
chrisd8088: I included a slightly edited version of this comment in one of the commits I added to this PR's branch.

lfsapi/auth.go
```diff
@@ -24,30 +24,43 @@ var (
...
+		// should have been authenticated but wasn't.
+		access = c.Endpoints.AccessFor(access.URL())
+		tracerx.Printf("api: http response indicates %q authentication. Resubmitting...", access.Mode())
 	}
 
-	return res, err
+	c.credContext.SetStateFields(nil)
+
+	tracerx.Printf("api: too many authentication attempts")
+	return nil, fmt.Errorf("too many authentication attempts")
```
larsxschneider, line 65: maybe `too many failed authentication attempts`?

lfsapi/auth.go
```diff
@@ -24,30 +24,43 @@ var (
...
+		// status code is received, doWithAuth() will remove any
+		// Authorization header that it added to the request.
+		if len(req.Header.Get("Authorization")) != 0 {
+			return res, err
 		}
+
+		// This case represents a rejected request that
+		// should have been authenticated but wasn't.
+		access = c.Endpoints.AccessFor(access.URL())
+		tracerx.Printf("api: http response indicates %q authentication. Resubmitting...", access.Mode())
```
larsxschneider: Shouldn't we print this only if there is an auth attempt left?

t/cmd/lfstest-gitserver.go
```diff
@@ -1717,18 +1718,25 @@ func skipIfBadAuth(w http.ResponseWriter, r *http.Request, id string) bool {
...
-		} else if cred == "cred2" {
-			return false
+		if matches := multiStageCredRE.FindStringSubmatch(cred); len(matches) == 3 {
+			if matches[1] == matches[2] {
+				return false
+			} else {
+				wantedAuth = "Multistage type=bar"
+				w.Header().Add(authHeader, wantedAuth)
+				w.WriteHeader(401)
+				debug(id, "auth stage %d of %d succeeded: %q", matches[1], matches[2], auth)
```
larsxschneider: we match `\d` but it is still a string:
```suggestion
				debug(id, "auth stage %s of %s succeeded: %q", matches[1], matches[2], auth)
```

lfsapi/auth.go
```diff
@@ -24,30 +24,43 @@ var (
...
+	if access.Mode() == creds.NoneAccess {
+		maxAuthAttempts++
+	}
+
+	for range maxAuthAttempts {
+		res, err := c.doWithAuth(remote, access, req, nil)
+		if err == nil || !errors.IsAuthError(err) {
+			return res, err
+		}
+
```
larsxschneider, line 37: Should we close the unused response body?
```
     if res != nil && res.Body != nil {
         res.Body.Close()
     }
```

chrisd8088: Great catch!  Yes, we should definitely be doing this, because the Go documentation [states](https://github.com/golang/go/blob/go1.26.1/src/net/http/response.go#L59-L64) that it's the caller's responsibility to do so.

Apparently we don't need to check that `Body` is `nil`, because it's guaranteed not to be, but it's probably safest to do so regardless.

lfsapi/auth.go
chrisd8088: Good point.  Sadly, this means we have to track the request count in a local variable, which I'd hoped to avoid purely for the sake of neatness, but it's the right thing to do.

lfsapi/auth.go
chrisd8088: I started to implement this change, but on reflection, I think the message is probably sufficient as it stands, for a few reasons.

First, I would hope that the statement that "too many attempts" have been made should imply to most readers that these attempts have not all been successful.

Second, while each attempt has received a 401 response, they haven't been outright denied with a 403.  So in that sense they haven't failed—but they also haven't completely succeeded.

As a concrete example, in a long multi-stage authentication sequence, each stage could "succeed" in the sense that a 401 is returned and so prompt the client to proceed to the next stage, but we might still report this error if the number of stages exceeds our allowed maximum.  While this is probably not going to occur in practice for anyone, I think this was the spirit in which the error message was first written, back in commit 67830784dbdc54c7fddf926653c18c4097124a8f of PR #5803.

In other words, the client has made multiple authentication requests, but hasn't reached a point of either being accepted or denied.  Its attempts to authenticate haven't specifically failed or succeeded, just reached an internal limit on the number of attempts.

t/cmd/lfstest-gitserver.go
chrisd8088: Thanks for spotting this; I've applied your fix.

larsxschneider (approved):
LGTM! Thank you :)

PR #6035, opened 2025-04-21
Fix language in code snippets in API docs

slonopotamus:
These snippets are not `js` (JavaScript), they are JSON. Since many snippets also include comments, I marked them as JSON5 because raw JSON doesn't support comments.


chrisd8088 (approved):
Looks good, thank you for the formatting corrections!

PR #6037, opened 2025-04-22
lfshttp/certs_darwin: filter CA certs that are only fuzzy matches of CN

### LINKED ISSUE #6036: darwin: Git LFS improperly adds certificates that cause SSL validation to fail
**Describe the bug**

In our organization (`example.org`), we use the [Jamf endpoint management system](https://www.jamf.com/), which adds a self-signed SSL certificate with the CN:

```
CN=example.org JSS Built-in Certificate Authority
```

This causes SSL validation to fail when accessing Git LFS on `example.org`:

```
Post "https://example.org/group/project/info/lfs/locks/verify": x509: certificate signed by unknown authority
```

This is happening because:

1. When accessing `example.org` endpoints, Git LFS checks if there are any custom certificates to trust for that host.
1. It calls `appendRootCAsForHostFromKeychain` with the hostname `example.org`.
1. This function executes [`/usr/bin/security find-certificate -a -p -c example.org <keychain>`](https://github.com/git-lfs/git-lfs/blob/2b26694f481f21c86a725e9d406b48fa47084a31/lfshttp/certs_darwin.go#L61), which performs a substring match on the Common Name.
1. If the System keychain contains a certificate with CN `example.org JSS Built-in Certificate Authority`, this gets returned by the fuzzy match.
1. Git LFS then adds this CA certificate to the trust store for connections to `example.org`.
1. When connecting to the real `example.org`, Git LFS is now trying to validate the official TLS certificate against this self-signed JAMF CA certificate, which will always fail.

**To Reproduce**

1. Install a self-signed cert that has a matching Git LFS server hostname in the CN.
2. Attempt to pull or push LFS files.

**System environment**

macOS

**Output of `git lfs env`**

```
git version 2.49.0

LocalWorkingDir=/Users/stanhu/github/git-lfs
LocalGitDir=/Users/stanhu/github/git-lfs/.git
LocalGitStorageDir=/Users/stanhu/github/git-lfs/.git
LocalMediaDir=/Users/stanhu/github/git-lfs/.git/lfs/objects
LocalReferenceDirs=
TempDir=/Users/stanhu/github/git-lfs/.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/Users/stanhu/github/git-lfs/.git/lfs
AccessDownload=none
AccessUpload=none
GIT_APPEND_BUILD_OPTIONS=LIBPCREDIR=/opt/homebrew/opt/pcre2
```

**Additional context**

It isn't possible to make `security find-certificate` return a strict match, but we can make the Git LFS code perform filtering: https://github.com/git-lfs/git-lfs/issues/6037

stanhu:
In our organization (`example.org`), we use the Jamf endpoint, which adds a self-signed SSL certificate with the CN:

```
CN=example.org JSS Built-in Certificate Authority
```

This causes SSL validation to fail when accessing Git LFS on `example.org`:

```
Post "https://example.org/group/project/info/lfs/locks/verify": x509: certificate signed by unknown authority
```

This is happening because:

1. When accessing `example.org` endpoints, Git LFS checks if there are any custom certificates to trust for that host.
2. It calls `appendRootCAsForHostFromKeychain` with the hostname `example.org`.
3. This function executes `/usr/bin/security find-certificate -a -p -c example.org <keychain>`, which performs a substring match on the Common Name.
4. If the System keychain contains a certificate with CN `example.org JSS Built-in Certificate Authority`, this gets returned by the fuzzy match.
5. Git LFS then adds this CA certificate to the trust store for connections to `example.org`.
6. When connecting to the real `example.org`, Git LFS is now trying to validate the official TLS certificate against this self-signed JAMF CA certificate, which will always fail.

Because `/usr/bin/security find-certificate` does not support exact matches, this commit filters out non-matching CNs to avoid using the wrong cert for the given domain.

Resolves #6036.

### DIFF
--- a/lfshttp/certs_darwin.go
+++ b/lfshttp/certs_darwin.go
@@ -1,7 +1,9 @@
 package lfshttp
 
 import (
+	"bytes"
 	"crypto/x509"
+	"encoding/pem"
 	"regexp"
 	"strings"
 
@@ -57,8 +59,8 @@ func appendRootCAsForHostFromPlatform(pool *x509.CertPool, host string) *x509.Ce
 	return pool
 }
 
-func appendRootCAsFromKeychain(pool *x509.CertPool, name, keychain string) *x509.CertPool {
-	cmd, err := subprocess.ExecCommand("/usr/bin/security", "find-certificate", "-a", "-p", "-c", name, keychain)
+func appendRootCAsFromKeychain(pool *x509.CertPool, hostname, keychain string) *x509.CertPool {
+	cmd, err := subprocess.ExecCommand("/usr/bin/security", "find-certificate", "-a", "-p", "-c", hostname, keychain)
 	if err != nil {
 		tracerx.Printf("Error getting command to read keychain %q: %v", keychain, err)
 		return pool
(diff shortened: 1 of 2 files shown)


lfshttp/certs_darwin.go
```diff
@@ -68,5 +70,60 @@ func appendRootCAsFromKeychain(pool *x509.CertPool, name, keychain string) *x509
...
+		cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			continue
+		}
+
+		// Skip CA certificates where the hostname is a substring of the CN
+		// but not an exact match (like "example.com JSS Built-in Certificate Authority")
+		if cert.IsCA &&
+			strings.Contains(cert.Subject.CommonName, hostname) &&
+			!strings.EqualFold(cert.Subject.CommonName, hostname) {
```
stanhu, line 112: Hmm, this is problematic if you happen to have a TLS cert that requires this bundle:

1. `example.org`
2. `example.org Intermediate CA`
3. `example.org Root CA`

It's possible all three certs are needed here? I guess this is really a problem with `/usr/bin/security find-certificate` not doing exact matching, or these intermediate certificates having a misleading CN?

chrisd8088:
Hey, thanks for this pull request and the description of the issue it is designed to address!  We really appreciate contributions like these which include tests and a thorough explanation of how the changes are intended to operate.

I'll aim to take a closer look as soon as possible, with the caveat that it might take me a little while before I can give this the careful review it deserves.

Thank you again very much for this PR!

stanhu:
Closing in favor of https://github.com/git-lfs/git-lfs/pull/6049.

PR #6039, opened 2025-04-23
Fix leaking directories

adrianschroeter:
Please consider this more a bugreport, then a valid fix. However, it solves the issue for us that our servers run "out of disk space" errors because no available inodes on an ext4 file system. The reason behind is that the /tmp/sock-XXXXX directories of lfs.sock never get cleaned up.

The fix assuming that upper directory of the lfs.sock must be removed is a bit dangerous, but we don't have a handle for the directory itself atm.

It solves the issue, but most likely some refactoring should be done to have a proper handle for cleanup instead.

### DIFF
--- a/ssh/connection.go
+++ b/ssh/connection.go
@@ -2,6 +2,9 @@ package ssh
 
 import (
 	"bytes"
+	"fmt"
+	"os"
+	"path/filepath"
 	"sync"
 
 	"github.com/git-lfs/git-lfs/v3/config"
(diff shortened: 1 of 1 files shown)


chrisd8088:
Hey, thanks for the bug report and first pass at a potential resolution!

I'll mark this PR as a bug fix and put it on the project board so we make sure to revisit the problem and find a comprehensive solution.

PR #6042, opened 2025-04-28
Fix migrations where `.git/objects` is a symlink

### LINKED ISSUE #5426: Bad revision fatal error when .git is a symbolic link with symbolic links inside
**Describe the bug**

I'm migrating Android repo tree from Gerrit to GitHub and I need to convert some repositories to LFS due to the file size limit. Android uses the [`repo`](https://android.googlesource.com/tools/repo) tool for repository management, which uses symbolic links for `.git` and directories inside `.git` for a given repo.

For example `.git` points to `../../../../../.repo/projects/xyz/core.git/` and inside of it:
```
-rw-r--r-- 1 docker docker    414 May 12 11:49 .repo_config.json
-rw-r--r-- 1 docker docker    176 Jul 13 21:30 FETCH_HEAD
-rw-r--r-- 1 docker docker     25 Jul 13 21:30 HEAD
drwxr-xr-x 1 docker docker      0 May 12 11:45 branches/
-rw-r--r-- 1 docker docker    408 Jul 13 21:30 config
lrwxrwxrwx 1 docker docker     70 May 12 11:45 description -> ../../../../../../project-objects/platform/system/core.git/description
lrwxrwxrwx 1 docker docker     64 May 12 11:45 hooks -> ../../../../../../project-objects/platform/system/core.git/hooks/
-rw-r--r-- 1 docker docker 215946 Jul 13 21:30 index
lrwxrwxrwx 1 docker docker     63 May 12 11:45 info -> ../../../../../../project-objects/platform/system/core.git/info/
drwxr-xr-x 1 docker docker     16 Jun  7 19:35 logs/
lrwxrwxrwx 1 docker docker     66 May 12 11:45 objects -> ../../../../../../project-objects/platform/system/core.git/objects/
-rw-r--r-- 1 docker docker     39 May 12 11:49 packed-refs
drwxr-xr-x 1 docker docker     32 May 12 12:05 refs/
lrwxrwxrwx 1 docker docker     67 May 12 11:45 rr-cache -> ../../../../../../project-objects/platform/system/core.git/rr-cache/
lrwxrwxrwx 1 docker docker     62 May 12 11:45 svn -> ../../../../../../project-objects/platform/system/core.git/svn/
```

When trying to perform any `git lfs migrate` operation, it fails with:
```console
$ git lfs migrate info --skip-fetch
Error in git rev-list --reverse --topo-order --do-walk --stdin --: exit status 128 fatal: bad revision 'refs/heads/upstream'
```

Once I copy the `.git` with `cp -rL` git-lfs works as expected.

**To Reproduce**
1. Clone a repository, e.g., with the `repo` tool, so that `.git` is a relative symlink and all directories inside are relative symlinks.
2. Try to run `git lfs migrate info --skip-fetch`

**Expected behavior**
Migrate tool works.

**System environment**

Bare metal Ubuntu 22.04.2 LTS

**Output of `git lfs env`**

```
git-lfs/3.0.2 (GitHub; linux amd64; go 1.18.1)
git version 2.34.1

Endpoint (upstream)=xxx/platform/system/core.git/info/lfs (auth=none)
  SSH=xxx:/platform/system/core
Endpoint (github)=xxx/platform_system_core.git/info/lfs (auth=none)
  SSH=xxx/platform_system_core.git
LocalWorkingDir=/opt/src/xxx/LINUX/android/system/core
LocalGitDir=/opt/src/.repo/projects/xxx/LINUX/android/system/core.git
LocalGitStorageDir=/opt/src/.repo/projects/xxx/LINUX/android/system/core.git
LocalMediaDir=/opt/src/.repo/projects/xxx/LINUX/android/system/core.git/lfs/objects
LocalReferenceDirs=
TempDir=/opt/src/.repo/projects/xxx/LINUX/android/system/core.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/opt/src/.repo/projects/xxx/LINUX/android/system/core.git/lfs
AccessDownload=none
AccessUpload=none
```

**Additional context**

--

### LINKED ISSUE #6030: `git-lfs-migrate import` doesn't work when `.git/objects` is a symlink
**Describe the bug**
Running `git lfs migrate import` errors when `.git/objects` is a symlink.

**To Reproduce**
```sh
# create a sandbox repo
mkdir -p sandbox/repo
cd sandbox/repo
git init --initial-branch=main

# make a commit
echo abc >> foo
git add .
git commit -m 'initial'

# copy objects folder to sandbox/ and symlink to it inside .git/
cp -r .git/objects ..
rm -rf .git/objects
ln -s ../../objects .git/objects

# try to migrate, observe error
git lfs migrate import --everything
```

```
Error in `git rev-list --reverse --topo-order --do-walk --stdin --`: exit status 128 fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

**System environment**
Fedora 40

**Additional context**
The reason that this error occurs is clear: `git-rev-list` is executed from inside the `objects/` directory, but this doesn't work if `objects/` is not in `.git/`: https://github.com/git-lfs/git-lfs/blob/d5db3c873ed3cf90b667d796a2c39b0dc7674578/git/rev_list_scanner.go#L175-L177

However, even if you remove this conditional such that `rev-list` is executed inside the cwd and thus succeeds, `git lfs migrate import` still doesn't work. No history is rewritten; it appears to be a no-op. You can observe this by doing the following, after hacking out the conditional logic to fix the `rev-list` error:

```sh
rm -rf .git/objects
cp -r ../objects .git/
git lfs migrate import --everything

git reset --hard HEAD@{1}
rm -rf .git/objects
ln -s ../../objects .git/
git lfs migrate import --everything
```

The outputs will look something like
```
Sorting commits: ..., done.
Rewriting commits: 100% (1/1), done.
  main  da0d9adcb169cc66ebd8f2d82d3c39a80495295d -> da62d0d420c23399dfaadc9ff63687c9926fb59f
Updating refs: ..., done.
Checkout: ..., done.

Sorting commits: ..., done.
Rewriting commits: 100% (1/1), done.
Updating refs: ..., done.
Checkout: ..., done.
```

The first case (no symlink), the history is rewritten. The second case (with symlink), nothing happens.

If you're wondering what the reason could be for having `.git/objects/` symlinked, it's because that's how Google's `git-repo` tool sets up the working directories for an AOSP checkout. LFS users working on AOSP projects don't have much choice in this matter.

bk2204:
Android's `repo` tool creates `.git/objects` as a symlink to a place outside the `.git` directory for unknown reasons.  When that happens, Git operations where we set the working directory to that directory fail.  This is because when Git performs discovery on the working directory, it first canonicalizes the path, which takes it outside of the `.git` directory and it then fails to find anything that looks like a Git repository.

In our codebase, we set the path to the object directory root in a couple of places which cause this problem.  To avoid this problem, simply join the repository root with `..`, which is always applied before canonicalization and will result in the path being properly set to the `.git` directory, which results in `git lfs migrate import` working correctly in this configuration.

Fixes #5426
Fixes #6030

### DIFF
--- a/git/githistory/rewriter.go
+++ b/git/githistory/rewriter.go
@@ -4,6 +4,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"sync"
 
@@ -313,7 +314,7 @@ func (r *Rewriter) Rewrite(opt *RewriteOptions) ([]byte, error) {
 			return nil, errors.Wrap(err, tr.Tr.Get("could not find refs to update"))
 		}
 
-		root, _ := r.db.Root()
+		root, _ := r.gitDirectory()
 
 		updater := &refUpdater{
 			cacheFn: r.uncacheCommit,
@@ -436,6 +437,15 @@ func copyEntryMode(e *gitobj.TreeEntry, mode int32) *gitobj.TreeEntry {
 	return copied
 }
 
+// gitDirectory returns a path to the .git directory on success.
+func (r *Rewriter) gitDirectory() (string, bool) {
+	root, ok := r.db.Root()
+	if ok {
+		root = filepath.Join(root, "..")
+	}
+	return root, ok
+}
+
 func (r *Rewriter) allows(typ gitobj.ObjectType, abs string) bool {
 	switch typ {
 	case gitobj.BlobObjectType:
@@ -529,7 +539,7 @@ func (r *Rewriter) refsToMigrate() ([]*git.Ref, error) {
 	var refs []*git.Ref
 	var err error
 
-	if root, ok := r.db.Root(); ok {
+	if root, ok := r.gitDirectory(); ok {
 		refs, err = git.AllRefsIn(root)
 	} else {
 		refs, err = git.AllRefs()
@@ -568,7 +578,7 @@ func (r *Rewriter) scannerOpts() *git.ScanRefsOptions {
 		Names:       make(map[string]string),
 	}
 
-	if root, ok := r.db.Root(); ok {
+	if root, ok := r.gitDirectory(); ok {
 		opts.WorkingDir = root
 	}
 	return opts
--- a/t/t-migrate-import.sh
+++ b/t/t-migrate-import.sh
@@ -142,6 +142,31 @@ begin_test "migrate import (given branch with filter)"
 )
 end_test
 
+begin_test "migrate import (.git objects/symlink)"
+(
+  set -e
+  mkdir other
+
+  setup_multiple_local_branches
+
+  mv .git/objects ../other/
+  ln -s ../../other/objects .git/objects
+
+  md_oid="$(calc_oid "$(git cat-file -p :a.md)")"
+  txt_oid="$(calc_oid "$(git cat-file -p :a.txt)")"
+  md_feature_oid="$(calc_oid "$(git cat-file -p my-feature:a.md)")"
+
+  git lfs migrate import
+
+  assert_pointer "refs/heads/main" "a.md" "$md_oid" "140"
+  assert_pointer "refs/heads/main" "a.txt" "$txt_oid" "120"
+
+  assert_local_object "$md_oid" "140"
+  assert_local_object "$txt_oid" "120"
+  refute_local_object "$md_feature_oid" "30"
+)
+end_test
+
 begin_test "migrate import (default branch, exclude remote refs)"
 (
   set -e


chrisd8088 (approved):
Thanks very much for jumping in with a fix for these issues!

I was puzzled for a few minutes as to how the existing code could have functioned at all, if adding a `..` path segment at the end of a so-called "root" path also works without problems (and solves some others), until I realized what is meant by the term "root" in this context.

Thanks again very much for fixing these problems!

#### Reference Notes

For my own future reference, the "root" referred to here is not the top of the repository's working tree or the top of Git's internal directory structure (i.e., `.git`), but the top of Git's internal object database directory structure (i.e., `.git/objects`).  So appending a `..` path segment always takes us back up to the top of Git's internal directory structure.

We construct the path to the `.git/objects` directory by first [calling](https://github.com/git-lfs/git-lfs/blob/58b1b77cc9827727011ecfba20fce976dad731c8/commands/command_migrate.go#L87) the `GitCommonDir()` method of our `git` package, which usually [runs](https://github.com/git-lfs/git-lfs/blob/58b1b77cc9827727011ecfba20fce976dad731c8/git/git.go#L889) `git rev-parse --git-common-dir`, and then [passing](https://github.com/git-lfs/git-lfs/blob/58b1b77cc9827727011ecfba20fce976dad731c8/commands/command_migrate.go#L92) that path to the `ObjectDatabase()` method, which [appends](https://github.com/git-lfs/git-lfs/blob/58b1b77cc9827727011ecfba20fce976dad731c8/git/git.go#L1674) the `objects` path segment, unless `GIT_OBJECT_DIRECTORY` is specified directly.  The resultant path is what is then [used](https://github.com/git-lfs/git-lfs/blob/58b1b77cc9827727011ecfba20fce976dad731c8/git/git.go#L1684) in the call to create the `ObjectDatabase` using our `git-lfs/gitobj` package.

demunion (approved):

PR #6048, opened 2025-05-03
Use a git-update-ref script to update references

alexkad0:
Invoking `git-update-ref` to rewrite a single reference is a noticeable performance bottleneck during LFS migration in a repository with a lot of references. Particularly, it may take ~3 minutes to update ~20k references.

The pull request makes Git LFS update all references at once by generating a reference update script and piping it into `git-update-ref`.

Please note that logic that handles nested tags in `updateOneRef()` assumes that the tag referenced by the tag being updated has already been updated in the Git storage, this assumption no longer holds. The pull request tackles the issue by caching the new state of the reference in the map `seen`.


git/githistory/ref_updater.go
```diff
@@ -48,13 +50,35 @@ func (r *refUpdater) updateRefs() error {
...
+	}
+	cmd.Dir = r.root
+	go func() {
+		defer input.Close()
+		io.WriteString(input, gitUpRefScript)
+	}()
+	o, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("git update-ref failed: %s, output: %s", err.Error(), string(o))
+	}
```
chrisd8088, line 80: Using `git update-ref --stdin` here is a great idea!!

First, I think we might want to add the `-z` option and use the NUL-delimited input [format](https://github.com/git/git/blob/6f84262c44a89851c3ae5a6e4c1a9d06b2068d75/Documentation/git-update-ref.adoc#L75-L93) so we don't run into problems caused by unusual ref names.  Fortunately, that option has been available since Git v2.0.0, so we don't have to worry about backward-compatibility if we use that format.

We also usually encapsulate the business of initializing (and, often, running) Git commands in small functions our `git` package so we can reuse them later from multiple locations in our code base.  For instance, our `git.UpdateIndexFromStdin()` [function](https://github.com/git-lfs/git-lfs/blob/4504241be48ba6e1969e11665703bea78c9e4a71/git/git.go#L648-L650) initializes a `git update-index --stdin ...` command, which we then [utilize](https://github.com/git-lfs/git-lfs/blob/4504241be48ba6e1969e11665703bea78c9e4a71/commands/pull.go#L171) in our `commands` package.  So I think we want to do the same thing here, perhaps with a new function that follows the `UpdateRefIn()` [function](https://github.com/git-lfs/git-lfs/blob/4504241be48ba6e1969e11665703bea78c9e4a71/git/git.go#L551-L567), for instance:

```go
// UpdateRefsFromStdinInDir initializes a "git update-ref" command which will
// read a series of reference update instructions from standard input and
// execute them. The command will operate within the given working
// directory "wd". The instructions should follow the NUL-terminated format.
func UpdateRefsFromStdinInDir(wd string) (*subprocess.Cmd, error) {
	cmd, err := gitNoLFS("update-ref", "--stdin", "-z")
	cmd.Dir = wd
	return cmd, err
}
```

As a side benefit, we can then effectively undo the change from commit bcc8e85e6e7a77928d5a6e0458abb86e579f30f2 in PR #2339, which was only made to support the Go test suite of our `git/githistory` package, by merging the `git.UpdateRefIn()` function back into the `git.UpdateRef()` function and simplifying it:
```go
func UpdateRef(ref *Ref, to []byte, reason string) error {
	args := []string{"update-ref", ref.Refspec(), hex.EncodeToString(to)}
	if len(reason) > 0 {
		args = append(args, "-m", reason)
	}

	_, err := gitNoLFSSimple(args...)
	return err
}
```

As for this `updateRefs()` method of the `refUpdater` structure, I had two thoughts.  The first is that we won't be able to use the new transactional `start`/`prepare`/`commit` instructions unless the Git version in use is v2.27.0 or higher, because they were only [introduced]() in that release.  With an older version of Git, we get the error `fatal: unknown command: start`, and we do try to support quite old versions of Git; our `build-earliest` CI [job](https://github.com/git-lfs/git-lfs/blob/4504241be48ba6e1969e11665703bea78c9e4a71/.github/workflows/ci.yml#L142-L165) is definitely not going to pass if we try to use these instructions.

Fortunately, according to the notes in commit git/git@e48cf33b61b8fecf1558fb32fcf5ee2987e82890 (which added support for explicit `git update-ref` transactions to Git) it should be fine to simply not send these instructions:

> By design, `git-update-ref(1)` will commit as soon as standard input is being closed. While fine in a non-transactional world, it is definitely unexpected in a transactional world. Because of this, as soon as any of the new transactional commands is used, the default will change to aborting without an explicit `commit`.

I think we might as well send the new instructions if Git is sufficiently modern, but it seems to work fine in my tests to also not send them, since we're really just performing one large transaction.

This does raise the question of how many ref updates we should pack into a single transaction.  I assume you're dealing with repositories with very large numbers of refs, so presumably your own tests indicate that using a single transaction for them all is feasible.

(As a possible enhancement, I suppose we could break up large numbers of ref updates into smaller transactions, assuming the user has Git v2.27.0 or newer installed.  And with future versions of Git we might also make use of the new `--batch-updates` flag that was just recently added in commit git/git@221e8fcb7f543f056246a901bcb1269a13145fa9, to allow for partial transaction failures.  But I'd suggest we leave this type of complexity alone for now, and wait to see if it's needed at all.)

My other thought with respect to this `updateRefs()` function is that we can perhaps avoid the need to collect a full script as an in-memory string if we start the `git update-ref` command before we loop over the refs, and then simply write out each `update` instruction to the command's input pipe in that loop.

Here's what I came up with as a possible version which uses the NUL-terminated instruction format, avoids failing with older Git versions, and writes to the command's input pipe directly rather than once after completing the loop over the set of refs to be updated.

Please do let me know what you think of this approach, and if you're able to test it against your repositories with large numbers of refs!

<details><summary><code>updateRefs()</code> Method</summary>

```go
func (r *refUpdater) updateRefs() error {
	list := r.logger.List(tr.Tr.Get("Updating refs"))
	defer list.Complete()

	var maxNameLen int
	for _, ref := range r.refs {
		maxNameLen = max(maxNameLen, len(ref.Name))
	}

	cmd, err := git.UpdateRefsFromStdinInDir(r.root)
	if err != nil {
		return err
	}

	var output bytes.Buffer
	cmd.Stdout = &output
	cmd.Stderr = &output

	input, err := cmd.StdinPipe()
	if err != nil {
		return err
	}

	err = cmd.Start()
	if err != nil {
		return err
	}
	defer func() {
		if cmd != nil {
			cmd.Wait()
		}
	}()

	if git.IsGitVersionAtLeast("2.27.0") {
		if _, err = io.WriteString(input, "start\x00"); err != nil {
			return err
		}
	}

	seen := make(map[string][]byte)
	for _, ref := range r.refs {
		if err = r.updateOneRef(list, maxNameLen, seen, ref, input); err != nil {
			return err
		}
	}

	if git.IsGitVersionAtLeast("2.27.0") {
		if _, err = io.WriteString(input, "prepare\x00commit\x00"); err != nil {
			return err
		}
	}
	if err = input.Close(); err != nil {
		return err
	}

	err = cmd.Wait()
	cmd = nil
	if err != nil {
		return fmt.Errorf("git update-ref failed: %s, output: %s", err.Error(), output.String())
	}

	return nil
}
```
</details>

<details><summary>Patch Diff for <code>/git/githistory/ref_updater.go</code></summary>

```diff
diff --git a/git/githistory/ref_updater.go b/git/githistory/ref_updater.go
index cc6ced7c..af486c84 100644
--- a/git/githistory/ref_updater.go
+++ b/git/githistory/ref_updater.go
@@ -1,6 +1,7 @@
 package githistory
 
 import (
+	"bytes"
 	"encoding/hex"
 	"fmt"
 	"strings"
@@ -10,7 +11,6 @@ import (
 	"github.com/git-lfs/git-lfs/v3/git"
 	"github.com/git-lfs/git-lfs/v3/tasklog"
 	"github.com/git-lfs/git-lfs/v3/tr"
-	"github.com/git-lfs/git-lfs/v3/subprocess"
 	"github.com/git-lfs/gitobj/v2"
 )
 
@@ -50,33 +50,56 @@ func (r *refUpdater) updateRefs() error {
 		maxNameLen = max(maxNameLen, len(ref.Name))
 	}
 
-	gitUpRefScript := "start\n";
+	cmd, err := git.UpdateRefsFromStdinInDir(r.root)
+	if err != nil {
+		return err
+	}
+
+	var output bytes.Buffer
+	cmd.Stdout = &output
+	cmd.Stderr = &output
+
+	input, err := cmd.StdinPipe()
+	if err != nil {
+		return err
+	}
+
+	err = cmd.Start()
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if cmd != nil {
+			cmd.Wait()
+		}
+	}()
+
+	if git.IsGitVersionAtLeast("2.27.0") {
+		if _, err = io.WriteString(input, "start\x00"); err != nil {
+			return err
+		}
+	}
 
 	seen := make(map[string][]byte)
 	for _, ref := range r.refs {
-		if err := r.updateOneRef(list, maxNameLen, seen, ref, &gitUpRefScript); err != nil {
+		if err = r.updateOneRef(list, maxNameLen, seen, ref, input); err != nil {
 			return err
 		}
 	}
 
-	gitUpRefScript += "prepare\ncommit\n";
-
-	cmd, err := subprocess.ExecCommand("git", "update-ref", "--stdin")
-	if err != nil {
-		return err;
+	if git.IsGitVersionAtLeast("2.27.0") {
+		if _, err = io.WriteString(input, "prepare\x00commit\x00"); err != nil {
+			return err
+		}
 	}
-	input, err := cmd.StdinPipe()
-	if err != nil {
-		return err;
+	if err = input.Close(); err != nil {
+		return err
 	}
-	cmd.Dir = r.root
-	go func() {
-		defer input.Close()
-		io.WriteString(input, gitUpRefScript)
-	}()
-	o, err := cmd.CombinedOutput()
+
+	err = cmd.Wait()
+	cmd = nil
 	if err != nil {
-		return fmt.Errorf("git update-ref failed: %s, output: %s", err.Error(), string(o))
+		return fmt.Errorf("git update-ref failed: %s, output: %s", err.Error(), output.String())
 	}
 
 	return nil
@@ -98,7 +121,7 @@ func (r *refUpdater) updateOneTag(tag *gitobj.Tag, toObj []byte) ([]byte, error)
 	return newTag, nil
 }
 
-func (r *refUpdater) updateOneRef(list *tasklog.ListTask, maxNameLen int, seen map[string][]byte, ref *git.Ref, gitUpRefScript *string) error {
+func (r *refUpdater) updateOneRef(list *tasklog.ListTask, maxNameLen int, seen map[string][]byte, ref *git.Ref, cmdInput io.WriteCloser) error {
 	sha1, err := hex.DecodeString(ref.Sha)
 	if err != nil {
 		return errors.Wrap(err, tr.Tr.Get("could not decode: %q", ref.Sha))
@@ -122,7 +145,7 @@ func (r *refUpdater) updateOneRef(list *tasklog.ListTask, maxNameLen int, seen m
 					return err
 				}
 
-				err = r.updateOneRef(list, maxNameLen, seen, old, gitUpRefScript)
+				err = r.updateOneRef(list, maxNameLen, seen, old, cmdInput)
 				if err != nil {
 					return err
 				}
@@ -158,7 +181,10 @@ func (r *refUpdater) updateOneRef(list *tasklog.ListTask, maxNameLen int, seen m
 		return nil
 	}
 
-	*gitUpRefScript += fmt.Sprintf("update %s %s\n", ref.Refspec(), hex.EncodeToString(to))
+	update := fmt.Sprintf("update %s\x00%s\x00\x00", ref.Refspec(), hex.EncodeToString(to))
+	if _, err := io.WriteString(cmdInput, update); err != nil {
+		return err
+	}
 
 	namePadding := max(maxNameLen-len(ref.Name), 0)
 	list.Entry(fmt.Sprintf("  %s%s\t%s -> %x", ref.Name, strings.Repeat(" ", namePadding), ref.Sha, to))
```

git/githistory/ref_updater.go
```diff
@@ -74,7 +98,7 @@ func (r *refUpdater) updateOneTag(tag *gitobj.Tag, toObj []byte) ([]byte, error)
 	return newTag, nil
 }
 
-func (r *refUpdater) updateOneRef(list *tasklog.ListTask, maxNameLen int, seen map[string]struct{}, ref *git.Ref) error {
+func (r *refUpdater) updateOneRef(list *tasklog.ListTask, maxNameLen int, seen map[string][]byte, ref *git.Ref, gitUpRefScript *string) error {
```
chrisd8088, line 101: In my proposed version where we pass an input pipe to this function, the function's signature would be something like:
```suggestion
func (r *refUpdater) updateOneRef(list *tasklog.ListTask, maxNameLen int, seen map[string][]byte, ref *git.Ref, cmdInput io.WriteCloser) error {
```

git/githistory/ref_updater.go
```diff
@@ -139,11 +158,10 @@ func (r *refUpdater) updateOneRef(list *tasklog.ListTask, maxNameLen int, seen m
 		return nil
 	}
 
-	if err := git.UpdateRefIn(r.root, ref, to, ""); err != nil {
-		return err
-	}
+	*gitUpRefScript += fmt.Sprintf("update %s %s\n", ref.Refspec(), hex.EncodeToString(to))
```
chrisd8088, line 161: In my proposed version where we pass an input pipe to this function, and also use the NUL-delimited instruction format, this line might become:

```suggestion
	update := fmt.Sprintf("update %s\x00%s\x00\x00", ref.Refspec(), hex.EncodeToString(to))
	if _, err := io.WriteString(cmdInput, update); err != nil {
		return err
	}
```

chrisd8088 (commented):
Thank you so much for this proposal and PR!

Using the `--stdin` option to `git update-ref` seems like a very good change and one we should definitely try to adopt for use in the `git lfs migrate` command.

I've put a few notes and possible additional changes into this review, so please take a look when you get a chance and let me know what you think.  As a quick summary, my main suggestions are:
- create a small wrapper in the `git` package for the command itself
- merge the existing `UpdateRefIn()` wrapper back into `UpdateRef()` and simplify it
- use the `-z` option and NUL-delimited input instructions
- avoid using the transactional instructions with Git prior to v2.27.0
- start the command before looping over the refs to be updated and then write `update` instructions to the command's input pipe within the loop

I hope those things all make sense!  Please do let me know your thoughts, and if you get a chance to test my suggestions against your real-world use cases, that would be fantastic.

Thank you very much against for this great improvement!

git/githistory/ref_updater.go
alexkad0: > First, I think we might want to add the -z option and use the NUL-delimited input

Yes, I think it would be great.

> This does raise the question of how many ref updates we should pack into a single transaction.

I asked myself the same question and, unfortunately, I forgot to answer it.

> I assume you're dealing with repositories with very large numbers of refs, so presumably your own tests indicate that using a single transaction for them all is feasible.

Yes, in my case it's fine to update ~20k references at once. This seems to be a reasonable limit.

git/githistory/ref_updater.go
```diff
@@ -4,6 +4,8 @@ import (
 	"encoding/hex"
 	"fmt"
 	"strings"
+	"io"
+	"bytes"
```
chrisd8088, line 8: I suspect our Go linter will ask for these to be ordered alphabetically, so that's a small change we'll need before the CI suite would pass.

chrisd8088 (approved):
Thank you for the additional changes!  I think this is looking really good.  I'll the CI suite run, although I expect it may complain about the ordering of modules in the `import` statement.

I'm also going to ask for a second pair of eyes from @git-lfs/core because this PR now includes a chunk of my own code.  Other than that, though, I think we're set to go.

As for handling the ref updates in a series of transactions instead of one, my own sense is that we should defer that to a follow-up PR if it proves to be necessary, but I'm interested in other opinions.

Thanks again very much for this idea and implementation!

git/githistory/ref_updater.go
chrisd8088: Awesome, thank you!

git/git.go
```diff
@@ -566,6 +566,16 @@ func UpdateRefIn(wd string, ref *Ref, to []byte, reason string) error {
 	return cmd.Run()
 }
 
+// UpdateRefsFromStdinInDir initializes a "git update-ref" command which will
+// read a series of reference update instructions from standard input and
+// execute them. The command will operate within the given working
+// directory "wd". The instructions should follow the NUL-terminated format.
+func UpdateRefsFromStdinInDir(wd string) (*subprocess.Cmd, error) {
```
larsxschneider, line 561: minuscule nit: `func UpdateRefsFromStdin` might match similar function names better?

larsxschneider (approved):
Nice! LGTM

git/git.go
chrisd8088: That's a good point, thank you!  We're removing `UpdateRefIn()`, so that in and of itself doesn't provide much guidance, which is why I tried to pick a new name that would be a bit more self-explanatory than the previous one.

However, your comment prompted me to double-check and indeed we do have one other function in the `git` package which explicitly [sets](https://github.com/git-lfs/git-lfs/blob/a920fbc115c4dd1e0792ed1db576de8fa791e080/git/git.go#L1469-L1478) the `cmd.Dir` value, the `AllRefsIn()` function.  So there is that model, which would suggest `UpdateRefsFromStdinIn()` here.

However ... I confess I find the doubled "in in" part of `...StdinIn` a bit mystifying, which is why I went with `...InDir()` instead.

On the other hand, our `NewLsFiles()` and `NewRevListScanner()` functions also set `cmd.Dir`, and they don't have any special suffix to their names.  Which would suggest just `UpdateRefsFromStdin()`, as you did.

Since your suggestion is the simplest, I decided to just go with that.  Thank you for the suggestion!  Naming things is hard.  :-)

(As a side note and future reference for myself, the reason this function changes the working directory at all is documented in PR #2339; it's solely because the Go tests in `git/githistory/ref_updater_test.go` [call](https://github.com/git-lfs/git-lfs/blob/a920fbc115c4dd1e0792ed1db576de8fa791e080/git/githistory/ref_updater_test.go#L32) the `updateRefs()` method on one of our `fixtures` [directories](https://github.com/git-lfs/git-lfs/tree/a920fbc115c4dd1e0792ed1db576de8fa791e080/git/githistory/fixtures), and need to change the current working directory to be inside the fixture's directory before running a Git command.)

Issue #5182, opened 2022-11-13
`.netrc` files ignored by Git LFS on Windows but used by Git via cURL


chrisd8088:
**Describe the bug**

On Windows, the cURL library has traditionally supported files in the user's home directory named `_netrc` rather than the `.netrc` files supported on Unix systems.  Since Git uses the cURL library, it had the same behaviour, and Git LFS's implementation of support for these files followed suit.  On Windows, the filename we search for in the user's home directory when looking for credentials is [configured](https://github.com/git-lfs/git-lfs/blob/4e767fd9540426ecd31a297ac6ece539dbbff8a3/config/netrc_windows.go#L6) as `_netrc`, while on Unix systems the filename is [configured](https://github.com/git-lfs/git-lfs/blob/4e767fd9540426ecd31a297ac6ece539dbbff8a3/config/netrc_nix.go#L6) as `.netrc`.

However, as of commit curl/curl@f9c7ba9096ec29db2536481d8e9ebe314e007f0c from August 2019, the cURL library searches first for a `.netrc` file in the user's home directory on Windows, and only if that is not found does it search for a `_netrc` file there.

The consequence is that if a Git user has a valid `.netrc` file and attempts to clone a repository containing Git LFS objects, the Git clone succeeds except that the Git LFS client fails while attempting to download the Git LFS objects, particularly if a `_netrc` file exists with different, invalid credentials.  Git will ignore the latter but Git LFS will not.

**To Reproduce**

On a Windows system with a relatively recent Git and cURL, using the Bash shell in minGW-64 as provided by Git for Windows, and assuming there are no `~/.netrc` or `~/_netrc` files already extant:

```
$ cat <<EOF >~/_netrc
machine github.com
login unused
password <GitHub personal access token>
machine api.github.com
login unused
password <GitHub personal access token>
EOF

# Clone a repository which requires the PAT for access and which contains LFS objects
$ git clone https://github.com/<user>/<repo>.git test
Cloning into 'test'...
...
Filtering content: 100% ..., done.

$ cd test
$ git lfs fsck
Git LFS fsck OK

$ cd ..
$ rm -r test

$ mv ~/_netrc ~/.netrc
$ cat <<EOF >~/_netrc
machine github.com
login unused
password foo
machine api.github.com
login unused
password foo
EOF

$ git clone https://github.com/<user>/<repo>.git test
Cloning into 'test'...
...
[...]

PR #6055, opened 2025-05-19
Prefer .netrc on windows if present

johanvdw:
Closes https://github.com/git-lfs/git-lfs/issues/5182

Aligns the project with curl and git.


chrisd8088:
Hey, thanks very much for this PR and proposed enhancement!  It would excellent if we can include this in a future release.

It's always great to see some files being fully deleted!  As it happens, these changes highlight another thing we can simplify, because it turns out that we can drop the `config/netrc.go` file entirely as it's not used anywhere.  That will eliminate some duplicative code and another whole source code file.

---

We will want some updates to our shell tests as part of this PR, to make sure it has the intended effect.  Right now there are a few [tests](https://github.com/git-lfs/git-lfs/blob/4504241be48ba6e1969e11665703bea78c9e4a71/t/t-credentials.sh#L448-L610) in the `t/t-credentials.sh` test script which use either `.netrc` or `_netrc`, [depending](https://github.com/git-lfs/git-lfs/blob/4504241be48ba6e1969e11665703bea78c9e4a71/t/t-credentials.sh#L442) on the OS platform.

One option might be to run all of these tests nested inside a `for` loop, which iterates over just `.netrc` on Unix platforms but over `.netrc` and `_netrc` on Windows, and so runs each test twice on Windows.

I think that probably a bit excessive, though; it's probably sufficient just to repeat the initial `credentials from netrc` [test](https://github.com/git-lfs/git-lfs/blob/4504241be48ba6e1969e11665703bea78c9e4a71/t/t-credentials.sh#L448-L485) for this purpose with both possible file names.

(While we're at it, we can also make the [condtition](https://github.com/git-lfs/git-lfs/blob/4504241be48ba6e1969e11665703bea78c9e4a71/t/t-credentials.sh#L441) which checks the OS platform a little more consistent with our other tests by using `if [ "$IS_WINDOWS" -eq 1 ]`, which is a bit more readable than what we have now, and also [covers](https://github.com/git-lfs/git-lfs/blob/4504241be48ba6e1969e11665703bea78c9e4a71/t/testenv.sh#L13-L15) more POSIX environments on Windows.)

So this might look something like the following:

<details><summary><code>t/t-credentials.sh</code> edits</summary>

```bash
NETRCFILES=".netrc"
if [ "$IS_WINDOWS" -eq 1 ]; then
  NETRCFILES+=" _netrc"
fi

for netrcfile in $NETRCFILES; do
  begin_test "credentials from netrc ($netrcfile)"
  (
    set -e

    printf "machine localhost\nlogin netrcuser\npassword netrcpass\n" >"$HOME/$netrcfile"
    ...
  )
  end_test
done

begin_test "credentials from netrc with unknown keyword"
(
  set -e

  printf "machine localhost\nlogin netrcuser\nnot-a-key something\npassword netrcpass\n" >"$HOME/.netrc"
  ...
)
end_test

# ... and then also use "$HOME/.netrc" in the subsequent tests instead of "$NETRCFILE"
```
</details>

---

I've added two commits on top of yours in my `windows-netrc` [branch](https://github.com/chrisd8088/git-lfs/tree/windows-netrc) and see that our Windows CI job [ran](https://github.com/chrisd8088/git-lfs/actions/runs/15156685076/job/42613049808) the `t/t-credentials.sh` script successfully, so I think these commits could be added to this PR, if you're willing to pull them into your branch.  (Or if you prefer to issue me an invitation to push to your fork, I can do that instead.)

I tried to include some relevant details in the commit messages so we won't have to dig around in the Git history as much in the future if there are any concerns down the road.

Thanks again very much for tackling this issue and putting this PR together!

creds/netrc.go
```diff
@@ -24,7 +28,17 @@ func ParseNetrc(osEnv config.Environment) (NetrcFinder, string, error) {
...
 	if _, err := os.Stat(nrcfilename); err != nil {
-		return &noFinder{}, nrcfilename, nil
+		// If on Windows, try _netrc instead
+		if runtime.GOOS == "windows" {
+			altFilename := filepath.Join(home, netrcAltBasename)
+			if _, errAlt := os.Stat(altFilename); errAlt == nil {
+				nrcfilename = altFilename
+			} else {
+				return &noFinder{}, altFilename, nil
+			}
```
chrisd8088, line 38: As a very minor stylistic suggestion, because we're returning from one branch, we can make this a bit shorter and also follow the general Go convention of handling errors first, e.g.:
```suggestion
			if _, errAlt := os.Stat(altFilename); errAlt != nil {
				return &noFinder{}, altFilename, nil
			}
			nrcfilename = altFilename
```

(I'm just thinking of the old "Effective Go" page in the [section](https://go.dev/doc/effective_go#if) where they say, "The code reads well if the successful flow of control runs down the page, eliminating error cases as they arise. Since error cases tend to end in `return` statements, the resulting code needs no `else` statements."  But this is definitely not a hard-and-fast rule, and I'm sure there are lots of counterexamples littering our code.)

chrisd8088 (commented):
Thank you for this great improvement, and welcome to the Git LFS project!

I think if we add some test changes like the ones I drafted, that should cover the main thing we'll want to include in this PR.

We might as well also just drop the `config/netrc.go` file, too, I think; there's no point in keeping it around.  I also drafted a commit to do that, with some notes on the tangled history of these files.

Thanks again very much for working on this!

johanvdw:
> I've added two commits on top of yours in my `windows-netrc` [branch](https://github.com/chrisd8088/git-lfs/tree/windows-netrc) and I see that our Windows CI job [ran](https://github.com/chrisd8088/git-lfs/actions/runs/15156685076/job/42613049808) the `t/t-credentials.sh` script with my changes successfully, so I think these commits could be added to this PR, if you're willing to pull them into your branch. (Or if you prefer to issue me an invitation to push to your fork, I can do that instead.)

As you noted, I merged your suggestions (and also run an extra test on our system where the issue surfaced).

chrisd8088 (approved):
This looks great, thank you!

I usually try to solicit a second opinion from another @git-lfs/core member when I've contributed to a PR, but in this case I think my changes are sufficiently minimal (just deleting an unused file and tweaking the test cases) that I can approve this PR as it stands.

Thanks again very much!

PR #6058, opened 2025-05-24
Create devcontainer.json

Osama2035:
Itcv


chrisd8088:
This looks like an exact copy of PR #5986, which we've closed due to a lack of response and explanation as to its utility.  Because this PR appears to be a spurious copy, I'm going to close it as well.

Since I've asked previously that you refrain from this behaviour, and you did not, I'm also going to block your account from this project.

PR #6084, opened 2025-07-26
Create devcontainer.json

Vichanlee:


harveyhauling (commented):
[Google Passwords (1).csv](https://github.com/user-attachments/files/21441636/Google.Passwords.1.csv)

chrisd8088:
Hey, thanks for the PR!

As it happens, this is an duplicate of PR #5986, which was closed due to lack of response from the contributor.  Please take a look at the comments in that PR, as they all apply here too.

Specifically, I asked whether the empty `features` block is required, and [posed](https://github.com/git-lfs/git-lfs/pull/5986#pullrequestreview-2607711050) a series of other questions, since like that PR this one also lacks any description or explanation:

> Would you be able to explain what the intended use case is here, and how this change is beneficial?
>
> I assume the goal is to improve the development experience when using a tool like VSCode or GitHub Codespaces, as you've referenced the Universal Development Container Image [published](https://hub.docker.com/r/microsoft/devcontainers-universal) by Microsoft.
>
> Can you expand on why this container is appropriate for your intended uses, and whether there are any limitations or alternatives?
> 
> As @bk2204 [pointed out](https://github.com/git-lfs/git-lfs/pull/5986#issuecomment-2649125717), there may be some issues with GitHub Codespaces, and a development environment where it's not possible to run out test suites (both the Go and Bash shell ones) is not especially useful.

chrisd8088:
As we've had no response to this PR and it's not useful for our own development processes, I'm going to close it now.

Please re-open it if you feel it's useful, and in that case, please reply to our questions so we can judge the value of this change and whether there are any potential alternatives.

PR #6088, opened 2025-07-31
fetch: allow providing refs via stdin

larsxschneider:
Add an option to `git lfs fetch` which allows passing the refs via stdin. This helps users to avoid "Argument list too long" errors if they pass in many refs.


commands/command_fetch.go
```diff
@@ -99,7 +101,27 @@ func fetchCommand(cmd *cobra.Command, args []string) {
 		}
 	}
 
-	if len(args) > 1 {
+	if fetchUseStdin {
+		if len(args) > 1 {
+			Print(tr.Tr.Get("Further command line arguments are ignored with --stdin"))
+			os.Exit(1)
+		}
```
chrisd8088, line 108: Looking at the `git fetch` command's manual I noticed that it [permits](https://git-scm.com/docs/git-fetch#Documentation/git-fetch.txt---stdin) the `--stdin` option to be combined with refs specified on the command line, and I wondered if we should try to match that behaviour.

If we stick with this design for now, though, we could just use the `Exit()` function to output the warning message, like we do in several other instances below (e.g., for the `Cannot combine --all with --recent` message, etc.)

chrisd8088 (approved):
Great idea, thank you for implementing it!

I had one suggestion only, but this could be merged without implementing it, so I've marked this as approved.

Thanks again!

commands/command_fetch.go
larsxschneider: Thank you for looking up the `git fetch` behavior!

I wrote this code to be consistent with the [`git lfs push` command](https://github.com/git-lfs/git-lfs/blame/5c6daf0b1e62e3f54a49c3d6596fb4661e0eba08/commands/command_push.go#L54-L57). Therefore, and for simplicity, I would like to keep the decision to ignore command line arguments if `--stdin` is given. If you feel strongly about it, then this would change my mind, though.

I agree on the usage of `Exit()`. However, I will create a new PR to make this change here and, for consistency, in `git lfs push`.

larsxschneider: See https://github.com/git-lfs/git-lfs/pull/6095

PR #6111, opened 2025-09-11
docs: Provide clearer instructions for non destructive imports

vikhik:
ref: https://github.com/git-lfs/git-lfs/issues/3421

### DIFF
--- a/docs/man/git-lfs-migrate.adoc
+++ b/docs/man/git-lfs-migrate.adoc
@@ -469,10 +469,39 @@ $ git lfs migrate import --no-rewrite test.zip some_specific_file.mp3 another_sp
 Migrating to Git LFS in a single new commit, with a specified commit message:
 
 ....
-$ git lfs migrate import --no-rewrite \
-  -m "Import test.zip, .mp3, .psd files in root of repo" \
-  test.zip some_specific_file.mp3 another_specific_file.psd
+$ git lfs track "*.zip" "*.mp3" "*.psd"
+$ git add .gitattributes
+$ git commit -m "add Git LFS attributes"
+$ git lfs migrate import --no-rewrite --yes \
+  -m "Import test.zip, audios/*.mp3, and images/*.psd files" \
+  test.zip audios/*.mp3 images/*.psd
 ....
+
+Note that the `*.mp3` and `*.psd` filename patterns are not quoted and
+so will be expanded by the shell, not by the `git lfs migrate import`
+command. Using quotation marks around such patterns with the
+`--no-rewrite` option is not supported, as the command expects a list
+of files (possibly as expanded by the shell) and will not search for
+all files matching a quoted filename pattern.
+Additional commands may be used to find and list all files matching a set
+of patterns, of course. In a POSIX-compliant environment such as Linux,
+macOS, or Git Bash on Windows, the example above could be combined with
+the `find(1)` and `sed(1)` commands as follows, so long as the resulting
+list of files does not cause the length of the expanded command line to
+exceed the maximum permitted by the shell:
+
+....
+$ git lfs migrate import --no-rewrite --yes \
+  $(find . -name "*.zip" -o -name "*.mp3" -o -name "*.psd" | sed "s/^\.\///")
+....
+
+Note also that the `git lfs track` command will by default change the
+modification timestamps on any existing files which match the given
+filename patterns so the files will appear "dirty" to Git and thereby
+prompt the user to convert them to Git LFS files. The subsequent
+`git lfs migrate import` command will then caution that the working tree
+contains modifications and ask the user for confirmation to proceed
+unless the `--yes` option is specified, as in the examples above.
 
 As an alternative to using `git lfs migrate import` with the `--no-rewrite`
 option to migrate files to Git LFS without modifying the existing history
--- a/docs/man/git-lfs-migrate.adoc
+++ b/docs/man/git-lfs-migrate.adoc
@@ -461,8 +461,16 @@ your repository.
 Migrating to Git LFS in a single new commit, without specifying a commit message:
 
 ....
-$ git lfs migrate import --no-rewrite test.zip some_specific_file.mp3 another_specific_file.psd
+$ git lfs track "*.zip" "*.mp3" "*.psd"
+$ git add .gitattributes
+$ git commit -m "add Git LFS attributes"
+
+$ git lfs migrate import --no-rewrite --yes test.zip audios/*.mp3 images/*.psd
 ....
+
+Note that the `*.mp3` and `*.psd` filename patterns are not quoted and
+so will be expanded by the shell, not by the `git lfs migrate import`
+command.  See the remarks following the next example for further details.
 
 Migrating to Git LFS in a single new commit, with a specified commit message:
 
--- a/docs/man/git-lfs-migrate.adoc
+++ b/docs/man/git-lfs-migrate.adoc
@@ -474,12 +474,18 @@ $ git lfs migrate import --no-rewrite \
   test.zip some_specific_file.mp3 another_specific_file.psd
 ....
 
-Alternatively, you can use `git add` with the `--normalize` flag to stage *all* changes which (now) match your .gitattributes
+As an alternative to using `git lfs migrate import` with the `--no-rewrite`
+option to migrate files to Git LFS without modifying the existing history
+of your repository, if you have Git v2.16.0 or higher installed, you can
+use the `--renormalize` option of the `git add` command to achieve
+the same result.  For example:
 
 ....
+$ git lfs track "*.zip" "*.mp3" "*.psd"
+$ git add .gitattributes
 $ git add --renormalize .
+$ git commit -m "Import all *.zip, *.mp3, and *.psd files to Git LFS"
 ....
-
 
 === Migrate from Git LFS
 
--- a/docs/man/git-lfs-migrate.adoc
+++ b/docs/man/git-lfs-migrate.adoc
@@ -460,19 +460,26 @@ your repository. Note that in the examples below, files in
 subdirectories are not included because they are not explicitly
 specified.
 
-Without a specified commit message:
+Using lfs migrate import without a specified commit message:
 
 ....
-$ git lfs migrate import --no-rewrite test.zip *.mp3 *.psd
+$ git lfs migrate import --no-rewrite test.zip some_specific_file.mp3 another_specific_file.psd
 ....
 
-With a specified commit message:
+Using lfs migrate import with a specified commit message:
 
 ....
 $ git lfs migrate import --no-rewrite \
   -m "Import test.zip, .mp3, .psd files in root of repo" \
-  test.zip *.mp3 *.psd
+  test.zip some_specific_file.mp3 another_specific_file.psd
 ....
+
+Alternatively, you can use `git add` with the `--normalize` flag to stage *all* changes which (now) match your .gitattributes
+
+....
+$ git add --renormalize .
+....
+
 
 === Migrate from Git LFS
 
--- a/docs/man/git-lfs-migrate.adoc
+++ b/docs/man/git-lfs-migrate.adoc
@@ -453,7 +453,7 @@ the `--all` option when force-pushing may be convenient if many local refs
 were updated, e.g., after importing to Git LFS with the `--everything`
 option.
 
-=== Migrate without rewriting local history
+=== Migrate without rewriting local history (using migrate import --no-rewrite)
 
 You can also migrate files without modifying the existing history of
 your repository.
@@ -511,7 +511,7 @@ prompt the user to convert them to Git LFS files. The subsequent
 contains modifications and ask the user for confirmation to proceed
 unless the `--yes` option is specified, as in the examples above.
 
-Alternative to using --no-rewrite:
+=== Migrate without rewriting local history (using add --renormalize)
 
 As an alternative to using `git lfs migrate import` with the `--no-rewrite`
 option to migrate files to Git LFS without modifying the existing history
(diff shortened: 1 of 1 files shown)


docs/man/git-lfs-migrate.adoc
```diff
@@ -460,20 +460,27 @@ your repository. Note that in the examples below, files in
 subdirectories are not included because they are not explicitly
 specified.
 
-Without a specified commit message:
+Using lfs migrate import without a specified commit message:
```
chrisd8088: We typically use uppercase for "LFS" unless it's part of a command line example, in which case we try to place the command in backticks and spell out the full command, i.e., `git lfs migrate import`.  So I would suggest we either phrase this simply using plain English, or include the full command name.

Personally, I think I prefer the former (the plain English version), but please take your pick, or offer another alternative if you prefer:

```suggestion
Migrating to Git LFS in a single new commit, without specifying a commit message:
```

Or:

```suggestion
Using `git lfs migrate import --no-rewrite` without a specified commit message:
```

docs/man/git-lfs-migrate.adoc
```diff
@@ -460,20 +460,27 @@ your repository. Note that in the examples below, files in
...
 
 ....
-$ git lfs migrate import --no-rewrite test.zip *.mp3 *.psd
+$ git lfs migrate import --no-rewrite test.zip some_specific_file.mp3 another_specific_file.psd
 ....
 
-With a specified commit message:
+Using lfs migrate import with a specified commit message:
```
chrisd8088: As above, I would either phrase this in plain English or include the full command name, in backticks.  My own suggestion would be:

```suggestion
Migrating to Git LFS in a single new commit, with a specified commit message:
```

docs/man/git-lfs-migrate.adoc
```diff
@@ -460,20 +460,27 @@ your repository. Note that in the examples below, files in
...
-  test.zip *.mp3 *.psd
+  test.zip some_specific_file.mp3 another_specific_file.psd
 ....
 
+Alternatively, you can use `git add` with the `--normalize` flag to stage *all* changes which (now) match your .gitattributes
+
+....
+$ git add --renormalize .
+....
+
```
chrisd8088: I appreciate the idea of mentioning this alternative and much simpler approach here!

The `git add` [option](https://git-scm.com/docs/git-add#Documentation/git-add.txt---renormalize) is actually `--renormalize`, not just `--normalize`, so we should correct that in the text, first of all.  (It's already correct in the example, just not in the text preceding it.)

As well, I'd suggest that we clarify in the description that this obviates any need to use the `git lfs migrate import` command at all, and that we expand the example to illustrate how to update your `.gitattributes` file as well.  Perhaps something like the following?

```suggestion
As an alternative to using `git lfs migrate import` with the `--no-rewrite`
option to migrate files to Git LFS without modifying the existing history
of your repository, if you have Git v2.16.0 or higher installed, you can
use the `--renormalize` option of the `git add` command to achieve
the same result.  For example:

....
$ git lfs track "*.zip" "*.mp3" "*.psd"
$ git add .gitattributes
$ git add --renormalize .
$ git commit -m "Import all *.zip, *.mp3, and *.psd files to Git LFS"
....
```

docs/man/git-lfs-migrate.adoc
```diff
@@ -460,20 +460,27 @@ your repository. Note that in the examples below, files in
...
 ....
 
-With a specified commit message:
+Using lfs migrate import with a specified commit message:
 
 ....
 $ git lfs migrate import --no-rewrite \
   -m "Import test.zip, .mp3, .psd files in root of repo" \
-  test.zip *.mp3 *.psd
+  test.zip some_specific_file.mp3 another_specific_file.psd
```
chrisd8088: I appreciate the goal here is to disambiguate the way in which the current example shows the use of a `*` "glob" filename pattern which is expanded by the shell, as opposed to other examples where we use quotes so that the `*` character is interpreted by the given command, such as `git lfs track "*.bin"` or `git lfs migrate import --include="*.bin"`.

One challenge with only showing specific filenames in our example, though, is that typically a user might want to try to match a wildcard pattern they've added to `.gitattributes`.  The wording of the preceding statement alludes to this by saying, "Note that in the examples below, files in subdirectories are not included because they are not explicitly specified."  But that's not really very clear, and our examples also don't illustrate how one needs to add patterns to `.gitattributes` before running `git lfs migrate import --no-rewrite`.

I think we need to expand these examples more fully, add some explanatory notes, and remove the preceding statement about files in subdirectories not being included.  For this section, perhaps something like:

```suggestion
$ git lfs track "*.zip" "*.mp3" "*.psd"
$ git add .gitattributes
$ git commit -m "add Git LFS attributes"

$ git lfs migrate import --no-rewrite --yes \
  -m "Import test.zip, audios/*.mp3, and images/*.psd files" \
  test.zip audios/*.mp3 images/*.psd
....

Note that the `*.mp3` and `*.psd` filename patterns are not quoted and
so will be expanded by the shell, not by the `git lfs migrate import`
command. Using quotation marks around such patterns with the
`--no-rewrite` option is not supported, as the command expects a list
of files (possibly as expanded by the shell) and will not search for
all files matching a quoted filename pattern.

Additional commands may be used to find and list all files matching a set
of patterns, of course. In a POSIX-compliant environment such as Linux,
macOS, or Git Bash on Windows, the example above could be combined with
the `find(1)` and `sed(1)` commands as follows, so long as the resulting
list of files does not cause the length of the expanded command line to
exceed the maximum permitted by the shell:

....
$ git lfs migrate import --no-rewrite --yes \
  $(find . -name "*.zip" -o -name "*.mp3" -o -name "*.psd" | sed "s/^\.\///")
....

Note also that the `git lfs track` command will by default change the
modification timestamps on any existing files which match the given
filename patterns so the files will appear "dirty" to Git and thereby
prompt the user to convert them to Git LFS files. The subsequent
`git lfs migrate import` command will then caution that the working tree
contains modifications and ask the user for confirmation to proceed
unless the `--yes` option is specified, as in the examples above.
```

docs/man/git-lfs-migrate.adoc
```diff
@@ -460,20 +460,27 @@ your repository. Note that in the examples below, files in
 subdirectories are not included because they are not explicitly
 specified.
 
-Without a specified commit message:
+Using lfs migrate import without a specified commit message:
 
 ....
-$ git lfs migrate import --no-rewrite test.zip *.mp3 *.psd
+$ git lfs migrate import --no-rewrite test.zip some_specific_file.mp3 another_specific_file.psd
 ....
```
chrisd8088: As below, I'd suggest we include some wildcard examples, but follow them with a note to see a detailed explanation like the one I've suggested for the following example.  So here I might suggest:

```suggestion
$ git lfs track "*.zip" "*.mp3" "*.psd"
$ git add .gitattributes
$ git commit -m "add Git LFS attributes"

$ git lfs migrate import --no-rewrite --yes test.zip audios/*.mp3 images/*.psd
....

Note that the `*.mp3` and `*.psd` filename patterns are not quoted and
so will be expanded by the shell, not by the `git lfs migrate import`
command.  See the remarks following the next example for further details.
```

docs/man/git-lfs-migrate.adoc
```diff
@@ -460,20 +460,27 @@ your repository. Note that in the examples below, files in
 subdirectories are not included because they are not explicitly
 specified.
```
chrisd8088: Since we're going to expand our examples below and add more discussion and explanation, I think we should just remove this last sentence entirely, the one which reads, "Note that in the examples below, files in subdirectories are not included because they are not explicitly specified."

As you've pointed out, because this sentence appears before the examples and doesn't specifically mention quoting and shell filename expansion, it's easy for a reader to not understand its full import.

docs/man/git-lfs-migrate.adoc
vikhik: Manually applied the suggested change (GitHub UI couldn't do it)

docs/man/git-lfs-migrate.adoc
vikhik: Updated, and added a single sentence "heading" for the alternative method to make the document flow more uniform

vikhik:
Updated the PR with the requested changes, and a touch more to unify the document flow with the added method!

Thanks for the collab and speedy reply :)

docs/man/git-lfs-migrate.adoc
chrisd8088: Great, thank you very much!

docs/man/git-lfs-migrate.adoc
```diff
@@ -456,22 +456,72 @@ option.
...
+so will be expanded by the shell, not by the `git lfs migrate import`
+command.  See the remarks following the next example for further details.
+
+Migrating to Git LFS in a single new commit, with a specified commit message:
+
+....
+$ git lfs track "*.zip" "*.mp3" "*.psd"
+$ git add .gitattributes
+$ git commit -m "add Git LFS attributes"
+$ git lfs migrate import --no-rewrite --yes \
```
chrisd8088, line 482: Let's add back one line of whitespace here, just to align with the other example above and to highlight the `git lfs migrate import` command:

```suggestion

$ git lfs migrate import --no-rewrite --yes \
```

docs/man/git-lfs-migrate.adoc
```diff
@@ -456,22 +456,72 @@ option.
...
 ....
 
-With a specified commit message:
+Note that the `*.mp3` and `*.psd` filename patterns are not quoted and
+so will be expanded by the shell, not by the `git lfs migrate import`
+command. Using quotation marks around such patterns with the
+`--no-rewrite` option is not supported, as the command expects a list
+of files (possibly as expanded by the shell) and will not search for
+all files matching a quoted filename pattern.
+Additional commands may be used to find and list all files matching a set
```
chrisd8088, line 494: Likewise, let's add back a line of whitespace here as well:

```suggestion

Additional commands may be used to find and list all files matching a set
```

docs/man/git-lfs-migrate.adoc
```diff
@@ -456,22 +456,72 @@ option.
...
+
+Note also that the `git lfs track` command will by default change the
+modification timestamps on any existing files which match the given
+filename patterns so the files will appear "dirty" to Git and thereby
+prompt the user to convert them to Git LFS files. The subsequent
+`git lfs migrate import` command will then caution that the working tree
+contains modifications and ask the user for confirmation to proceed
+unless the `--yes` option is specified, as in the examples above.
+
+Alternative to using --no-rewrite:
```
chrisd8088: I think including a sub-heading here is very good idea!

If we use a `===` formatting, the Asciidoctor `manpage` [converter](https://docs.asciidoctor.org/asciidoctor/latest/manpage-backend/) will turn that into a nice section title for us in the HTML and roff versions of our manual pages.  What would you think about something like this, which further clarifies that you don't need the `git lfs migrate` command at all if you use this alternative approach?

```suggestion
=== Migrate without rewriting local history and without --no-rewrite
```

---

For reference, here's my thinking on this proposed change; you can skip this entirely, though, if you like!  :-)

There are actually a lot of constraints on these section titles, so the text we can use is quite limited.  Normally I would suggest using backticks or other formatting around the command name, such as `` `--no-rewrite` `` or `"--no-rewrite"`.

Unfortunately, those all run afoul of the very restrictive regular expression we [use](https://github.com/git-lfs/git-lfs/blob/463ed9727d21155585721489fdb8cc81030f979b/docs/man/mangen.go#L72) to recognize section titles when we convert the AsciiDoc source into plain text to be displayed by the `git lfs help` command, since it only permits ASCII alphanumerics, the space character, and the hyphen character.

Meanwhile, we also use the proper Asciidoctor conversion tool to create roff and HTML versions to be included in our release packages.  And finally there's GitHub's UI, which has its own converter for rendering AsciiDoc in HTML.  Anyway, it's all a bit of a mess right now, so our best option is to just use the simplest possible text formatting here, with hyphens and nothing else.

And while it would be handy to use a section title here that was indented one extra level, which could be easily formatted in AsciiDoc with four equals sign characters (e.g., `==== Alternative without --no-rewrite` or something like that), the GNU manual page formatting [standards](https://man7.org/linux/man-pages/man7/groff_man.7.html) only support two levels of section titles, and so the Asciidoctor converter [maps](https://github.com/asciidoctor/asciidoctor/blob/main/lib/asciidoctor/converter/manpage.rb#L141-L144) any section title with three or more equals signs to a simple second-level section title, as our custom converter does as well.

Since we're stuck with a section title at the same level as the preceding one, I thought it was best to just repeat the same initial "Migrate without rewriting local history" text but then try to clarify why this section is different from the previous one.

chrisd8088 (commented):
Hey, thank you so much for continuing to work on this PR, and for accepting my (rather verbose) suggestions!

I like the idea of adding a distinct sub-heading to call out the use of `git add --renormalize` as a completely separate approach from using `git lfs migrate import --no-rewrite`.

My only concrete suggestion there is to format that sub-heading as a section title, and so I proposed some text which might work with the limited formatting available to us for those kinds of titles.

Please do take a look and let me know what you think, and thanks again for your help improving our documentation!

docs/man/git-lfs-migrate.adoc
vikhik: I took it a step further and added context for both, what do you think?

chrisd8088: That looks great, thank you!

chrisd8088 (approved):
This looks great, thank you!

chrisd8088:
Looks like we need to bump our CI jobs up to Go 1.25/1.24, so I'll merge PR #6116 first, and then this one.  My apologies for the delay!

Issue #5181, opened 2022-11-12
Macro attributes handled inconsistently by `git lfs fsck`


chrisd8088:
**Describe the bug**

The `git lfs fsck` command processes macro attribute definitions and resolves macro attribute references as it encounters them, rather than the way Git does, which is to read all attribute definitions from all relevant files (letting definitions from higher-precedence files override definitions for the same file patterns and macro attributes in lower-precedence files), and then resolve any macro attribute references based on that final set of definitions.

The `git lfs fsck` command's macro attribute handling differs from Git's and from that of the `git lfs track` command because the `git lfs fsck` command uses the `git ls-tree -r` command to find all `.gitattributes` files, and then processes each file once in the order they are returned by that command, and each definition once in order of appearance within a single file.  A number of inconsistent behaviours are exhibited as a consequence.

**To Reproduce**

```
$ git init test
$ cd test
$ git commit --allow-empty -m init

$ echo foo >foo.top
$ cp foo.top foo.bot
$ git add foo.*
$ git commit -m foo
$ cat <<EOF >.gitattributes
*.top lfs
[attr]lfs filter=lfs diff=lfs merge=lfs -text
*.bot lfs
EOF
$ git add .gitattributes
$ git commit -m attrs
$ git check-attr filter foo.*
foo.bot: filter: lfs
foo.top: filter: lfs
$ git lfs fsck
pointer: unexpectedGitObject: "foo.bot" (treeish 6775e3dd3f9367101874dbf9d4a220b8f9576f93) should have been a pointer but was not

$ git reset --hard HEAD^
$ cat <<EOF >.gitattributes
[attr]lfs filter=lfs diff=lfs merge=lfs -text
*.top lfs
*.bot lfs
EOF
$ git add .gitattributes
$ git commit -m attrs
$ git lfs fsck
pointer: unexpectedGitObject: "foo.bot" (treeish 00f772c3dfbc0d15c11a20b4510318a7d4d23901) should have been a pointer but was not
pointer: unexpectedGitObject: "foo.top" (treeish 00f772c3dfbc0d15c11a20b4510318a7d4d23901) should have been a pointer but was not

$ git reset --hard HEAD^^
$ mkdir dir
$ echo foo >dir/foo.bin
$ echo "*.bin lfs" >dir/.gitattributes
$ cp -r dir .dir
$ git add dir .dir
$ git commit -m foo
[...]

Issue #5179, opened 2022-11-11
Macro attributes handled inconsistently by `git lfs track`


chrisd8088:
**Describe the bug**

The `git lfs track` command processes macro attribute definitions and resolves macro attribute references as it encounters them, rather than the way Git does, which is to read all attribute definitions from all relevant files (letting definitions from higher-precedence files override definitions for the same file patterns and macro attributes in lower-precedence files), and then resolve any macro attribute references based on that final set of definitions.

The `git lfs track` command's macro attribute handling differs from Git's and from that of the `git lfs fsck` command because the `git lfs track` command orders `.gitattributes` files, including the `.git/info/attributes` file, by full path length from the root of the repository, and then processes each file once in order, and each definition once in order of appearance within a single file.  A number of inconsistent behaviours are exhibited as a consequence.

**To Reproduce**

The examples below illustrate the behaviour of `git lfs track` without any arguments.  When a pattern is supplied (e.g., `git lfs track "*.bin"`), the behaviour is similar and varies only in terms of whether an attribute definition in the current directory's `.gitattributes` file is written or not, depending on whether the command recognizes one as having a defined `filter=lfs` attribute based on how it processes macro attribute references; the macro attribute reference resolution logic remains the same as shown in the example below, however.

```
$ git init test
$ cd test

$ cat <<EOF >.gitattributes
*.top lfs
[attr]lfs filter=lfs diff=lfs merge=lfs -text
*.bot lfs
EOF
$ touch foo.top foo.bot
$ git check-attr filter foo.*
foo.bot: filter: lfs
foo.top: filter: lfs
$ git lfs track
Listing tracked patterns
    *.bot (.gitattributes)
Listing excluded patterns

$ echo "[attr]lfs filter=lfs diff=lfs merge=lfs -text" >.gitattributes
$ mkdir dir
$ echo "*.bin lfs" >dir/.gitattributes
$ touch dir/foo.bin
$ git check-attr filter dir/foo.bin
dir/foo.bin: filter: lfs
$ git lfs track

$ rm .gitattributes
$ echo "[attr]lfs filter=lfs diff=lfs merge=lfs -text" >.git/info/attributes
$ cp -r dir thisisalongdir
$ git check-attr filter dir/foo.bin thisisalongdir/foo.bin
dir/foo.bin: filter: lfs
thisisalongdir/foo.bin: filter: lfs
$ git lfs track
Listing tracked patterns
    dir/*.bin (dir/.gitattributes)
[...]

Issue #5173, opened 2022-11-10
Special Git attributes files not read by `git lfs fsck` command


chrisd8088:
**Describe the issue**
In order to determine which file patterns should be checked for valid and invalid Git LFS pointer files, the `git lfs fsck` command only examines `.gitattributes` files which are committed to the Git repository.

This stands in contrast to how the `git lfs track` command operates, as it parses both the system-wide and global Git attributes [files](https://github.com/git-lfs/git-lfs/blob/be47d92db33851c4631cf9792e9e13ad6e0316b4/commands/command_track.go#L292-L293) and the `$GIT_DIR/info/attributes` [file](https://github.com/git-lfs/git-lfs/blob/be47d92db33851c4631cf9792e9e13ad6e0316b4/git/attribs.go#L175), although it only parses the system-wide and global ones to resolve macro attribute references.

Arguably this is correct behaviour for the `git lfs fsck` command as the special Git attributes files may vary with different user and system configurations, and looking only at `.gitattributes` files that are part of the repository contents (and by ignoring files that only in the working tree or Git index) should result in repeatable, consistent output between users.

If that's the case, we should update the `git lfs fsck` documentation to clarify how it resolves Git macro attribute references and how and why it does not match the behaviour of either Git or the `git lfs track` command.

**To Reproduce**
Steps to reproduce the behaviour (note that this example assumes `~/.git/config/attributes` and `${prefix}/etc/gitattributes` do not exist; `${prefix}` is the installation prefix for Git, if any, e.g., `/usr/local` on macOS with Homebrew):
```
$ git init test
$ cd test

$ git lfs uninstall
$ echo foo >foo.bin
$ git add foo.bin
$ git commit -m init

$ git lfs install
$ git check-attr filter foo.bin
foo.bin: filter: unspecified

$ echo "*.bin filter=lfs diff=lfs merge=lfs -text" >.git/info/attributes
$ git check-attr filter foo.bin
foo.bin: filter: lfs
$ git lfs fsck
Git LFS fsck OK

$ rm .git/info/attributes

$ echo "*.bin filter=lfs diff=lfs merge=lfs -text" >~/.config/git/attributes
$ git check-attr filter foo.bin
foo.bin: filter: lfs
$ git lfs fsck
Git LFS fsck OK

$ rm ~/.config/git/attributes

$ command -v git
/usr/local/bin/git

[...]

PR #6126, opened 2025-10-16
Support Macro in migrate `--fixup`

### LINKED ISSUE #5383: Macro attributes ignored by `--fixup` option to `git lfs migrate` commands
**Describe the bug**

The `git lfs migrate info` and `git lfs migrate import` commands, when used with their `--fixup` options, do not recognize as requiring migration blobs which are not Git LFS objects but to which the `filter=lfs` Git attributes is applied by existing `.gitattributes` files in the repository, if the attribute is applied by a `.gitattributes` reference to a macro attribute whose definition sets the `filter=lfs` attribute.

**To Reproduce**

```
$ git init test
$ cd test

$ git lfs uninstall

$ echo foo >foo.bin
$ cat <<EOF >.gitattributes
[attr]lfs filter=lfs diff=lfs merge=lfs -text
*.bin lfs
EOF
$ git commit -m attrs
$ git check-attr filter foo.bin
foo.bin: filter: lfs
$ git show HEAD:foo.bin
foo

$ git lfs install

$ git lfs migrate info --fixup
migrate: Fetching remote refs: ..., done.                                       
migrate: Examining commits: 100% (1/1), done.

$ git lfs migrate import --fixup
migrate: Fetching remote refs: ..., done.                                       
migrate: Rewriting commits: 100% (1/1), done.                                   
  master	2467c885c36776305dc42db9f40202656440c59c -> 2467c885c36776305dc42db9f40202656440c59c
migrate: Updating refs: ..., done.                                              
migrate: checkout: ..., done.   
...
```

**Expected behaviour**

The `git lfs migrate info --fixup` command should report blobs which are not Git LFS objects but to which the `filter=lfs` Git attribute applies, even when that attribute is applied by a `.gitattributes` reference to a macro attribute whose definition sets the `filter=lfs` attribute.

Equally, the `git lfs migrate import --fixup` command should convert blobs which are not Git LFS objects into Git LFS objects when the `filter=lfs` Git attribute applies to them, even when that attribute is applied by a `.gitattributes` reference to a macro attribute whose definition sets the `filter=lfs` attribute.

**System environment**

The example above is performed on macOS with Git installed under `/usr/local`, but applies to any system with appropriate choice of `$prefix` for the commands which touch `$prefix/etc/gitattributes`.

**Output of `git lfs env`**

Note that this is a custom build of the Git LFS client with PR #5382 merged.

```
git-lfs/3.3.0 (GitHub; darwin amd64; go 1.20.1; git 2bdd529d)
git version 2.39.0
```

**Additional context**

Macro attribute support was initially added in PR #3391, but the `git lfs migrate import --fixup` command option was not updated at that time to process macro attributes, and when the `git lfs migrate info --fixup` command option was introduced in commit 4800c5e9882f9cf1e271a855baeb32dbe1f767ec of PR #4501, it too lacked support for Git macro attributes.

Consequently, using `git lfs migrate info --fixup` with any macro attribute would lead to a panic, as reported in #5332, while `git lfs migrate import --fixup` with the same input data would stall indefinitely.  This behaviour has been resolved in PRs #5375 and #5382.  In commit 07b1adf4d059dc522c6413ff8af5133e2050f225 of the latter PR, a commented-out [test](https://github.com/git-lfs/git-lfs/pull/5375) was added to the `t/t-migrate-fixup.sh` file and [another](https://github.com/git-lfs/git-lfs/blob/2bdd529dbb2c0f38fbf327abd963fc673137a74f/t/t-migrate-info.sh#L800-L820) to the `t/t-migrate-info.sh` file which illustrate the problem.

/cc #5332 as the issue whose investigation led to this issue

Diamondemon:
This pull request fixes #5383  by making the `gitattr.Tree` structure able to process macro attributes.

### New Tree behaviour:
- the Tree class now processes macros, and respects the order of processing given by git documentation

### Detailled behaviour:
- The `gitattr.Tree` structure now has a reference to the 3 special attributes trees, namely system, user and repository. These references can be nil for now.
- The original `gitattr.New` recursive logic has been moved to an internal `gitattr.newFromTree` ([here](https://github.com/Diamondemon/git-lfs/blob/7e01ef2968e0125920658f720a77ff286318b4c8/git/gitattr/tree.go#L36C1-L71C2)). 
- `gitattr.New` now assumes that it is creating a `Tree` "located" at the root of the repository, with its own `gitattr.MacroProcessor`. This `MacroProcessor` is relayed to its children through the recursive creation in `gitattr.newFromTree`
- The original recursive logic of `Tree.Applied`  has been moved to an internal `Tree.applied` method ([here](https://github.com/Diamondemon/git-lfs/blob/7e01ef2968e0125920658f720a77ff286318b4c8/git/gitattr/tree.go#L148C1-L165C2)).
- `Tree.Applied` now assumes the Tree it is called on is the root tree (which may reference the special trees).
It starts by using the `MacroProcessor` referenced in the `Tree`, which is apparently similar to the way `$ git check-attr` works. The `MacroProcessor` is assumed as never nil.
The processing of macros is done respecting the order indicated in the [git documentation](https://git-scm.com/docs/gitattributes) (system --> user --> tree --> repo). Only macros are read, `PatternLine` conversions are ignored at that step.
Once macros are parsed, the whole tree structure is processed again in the same order, this time without reading macros.
- To avoid parsing macros every time `Tree.Applied` is called, a [new boolean](https://github.com/Diamondemon/git-lfs/blob/7e01ef2968e0125920658f720a77ff286318b4c8/git/gitattr/macro.go#L5) has been added to the `MacroProcessor` structure, which is set to `true` once macros were processed at least once (i.e. through calling `MacroProcessor.ProcessLines(lines, true)`. This boolean is checked before parsing macros in `Tree.Applied` and only used there for now.
It is thus assumed that a `Tree` instance `t` is never modified after the first call of `t.Applied`.

### Other features:
-  A new function `gitattr.NewReader` ([here](https://github.com/Diamondemon/git-lfs/blob/7e01ef2968e0125920658f720a77ff286318b4c8/git/gitattr/tree.go#L73C1-L82C2)) has been added, following the idea of unifying the parsing of attributes to avoid future inconsistencies. For now, it takes a `MacroProcessor` and an `io.Reader` as arguments. It may eventually get adapted depending when fixing #5173, #5181 and #5179
- Existing tests were adapted to the requirements of the new behaviour (19bb112).
- [New tests](https://github.com/Diamondemon/git-lfs/blob/7e01ef2968e0125920658f720a77ff286318b4c8/git/gitattr/tree_test.go#L256C1-L382C2) were added to test the new parsing behaviours but they may lack precision. Any insight would be appreciated.
- A [new test](https://github.com/Diamondemon/git-lfs/blob/7e01ef2968e0125920658f720a77ff286318b4c8/git/gitattr/macro_test.go#L187C1-L217C2) was added to verify macro override when reading re-definitions of a macro, to stay consistent with `$ git check-attr`.

### Work left to do in later PRs:
- The Tree class now joins #5173 because the current package structure prevents from getting the special attribute files without duplicating existing code.
- Full documentation of the `gitattr` will be done once main relevant issues are resolved.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
           FORCE_LOCALIZE: true
     - run: mkdir -p bin/assets
     - run: find bin/releases -name "*$(uname -s | tr A-Z a-z)*" | xargs -I{} cp {} bin/assets
-    - uses: actions/upload-artifact@v4
+    - uses: actions/upload-artifact@v5
       with:
         name: ${{ matrix.os }}
         path: bin/assets
@@ -111,7 +111,7 @@ jobs:
       shell: bash
     - run: mv *.exe bin/assets
       shell: bash
-    - uses: actions/upload-artifact@v4
+    - uses: actions/upload-artifact@v5
       with:
         name: windows-latest
         path: bin/assets
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -58,7 +58,7 @@ jobs:
       shell: bash
       env:
         FORCE_LOCALIZE: true
-    - uses: azure/trusted-signing-action@v0.5.9
+    - uses: azure/trusted-signing-action@v0.5.10
       with:
         azure-tenant-id: ${{ secrets.SPN_GIT_LFS_SIGNING_TENANT_ID }}
         azure-client-id: ${{ secrets.SPN_GIT_LFS_SIGNING_CLIENT_ID }}
@@ -73,7 +73,7 @@ jobs:
         timestamp-digest: SHA256
     - run: make release-windows-stage-2
       shell: bash
-    - uses: azure/trusted-signing-action@v0.5.9
+    - uses: azure/trusted-signing-action@v0.5.10
       with:
         azure-tenant-id: ${{ secrets.SPN_GIT_LFS_SIGNING_TENANT_ID }}
         azure-client-id: ${{ secrets.SPN_GIT_LFS_SIGNING_CLIENT_ID }}
@@ -90,7 +90,7 @@ jobs:
       shell: bash
     - run: make release-windows-rebuild
       shell: bash
-    - uses: actions/upload-artifact@v4
+    - uses: actions/upload-artifact@v5
       with:
         name: windows-assets
         path: bin/releases
@@ -125,7 +125,7 @@ jobs:
         DARWIN_DEV_USER: ${{secrets.MACOS_DEV_USER}}
         DARWIN_DEV_PASS: ${{secrets.MACOS_DEV_PASS}}
         DARWIN_DEV_TEAM: ${{secrets.MACOS_DEV_TEAM}}
-    - uses: actions/upload-artifact@v4
+    - uses: actions/upload-artifact@v5
       with:
         name: macos-assets
         path: bin/releases
@@ -151,11 +151,11 @@ jobs:
     - run: sudo apt-get update && sudo apt-get -y install gettext libarchive-tools
       env:
           DEBIAN_FRONTEND: noninteractive
-    - uses: actions/download-artifact@v5
+    - uses: actions/download-artifact@v6
       with:
         name: windows-assets
         path: windows-assets
-    - uses: actions/download-artifact@v5
+    - uses: actions/download-artifact@v6
       with:
         name: macos-assets
         path: macos-assets
@@ -166,7 +166,7 @@ jobs:
     - run: rm -f bin/releases/git-lfs-windows-assets*.tar.gz
     - run: mkdir -p release-assets/bin
     - run: mv bin/releases release-assets/bin
-    - uses: actions/upload-artifact@v4
+    - uses: actions/upload-artifact@v5
       with:
         name: release-assets
         path: release-assets
--- a/commands/command_pointer.go
+++ b/commands/command_pointer.go
@@ -16,17 +16,20 @@ import (
 )
 
 var (
-	pointerFile     string
-	pointerCompare  string
-	pointerStdin    bool
-	pointerCheck    bool
-	pointerStrict   bool
-	pointerNoStrict bool
+	pointerFile         string
+	pointerCompare      string
+	pointerStdin        bool
+	pointerCheck        bool
+	pointerStrict       bool
+	pointerNoStrict     bool
+	pointerNoExtensions bool
 )
 
 func pointerCommand(cmd *cobra.Command, args []string) {
+	var comparePointer *lfs.Pointer
 	comparing := false
 	something := false
+	hasExtensionsConfig := false
 	buildOid := ""
 	compareOid := ""
 
@@ -39,7 +42,7 @@ func pointerCommand(cmd *cobra.Command, args []string) {
 		}
 
 		if len(pointerCompare) > 0 {
-			ExitWithError(errors.New(tr.Tr.Get("Cannot combine --check with --compare")))
+			ExitWithError(errors.New(tr.Tr.Get("Cannot combine --check with --pointer")))
 		}
 
 		if len(pointerFile) > 0 {
@@ -53,7 +56,7 @@ func pointerCommand(cmd *cobra.Command, args []string) {
 		} else if pointerStdin {
 			r = io.NopCloser(os.Stdin)
 		} else {
-			ExitWithError(errors.New(tr.Tr.Get("Must specify either --file or --stdin with --compare")))
+			ExitWithError(errors.New(tr.Tr.Get("Must specify either --file or --stdin with --check")))
 		}
 
 		p, err := lfs.DecodePointer(r)
@@ -79,17 +82,37 @@ func pointerCommand(cmd *cobra.Command, args []string) {
 			os.Exit(1)
 		}
 
-		oidHash := sha256.New()
-		size, err := io.Copy(oidHash, buildFile)
+		var ptr *lfs.Pointer
+		if pointerNoExtensions || !cfg.InRepo() {
+			oidHash := sha256.New()
+			size, err := io.Copy(oidHash, buildFile)
+			if err != nil {
+				Error(err.Error())
+				buildFile.Close()
+				os.Exit(1)
+			}
+
+			ptr = lfs.NewPointer(hex.EncodeToString(oidHash.Sum(nil)), size, nil)
+		} else {
+			gitfilter := lfs.NewGitFilter(cfg)
+			hasExtensionsConfig = len(cfg.Extensions()) > 0
+			cleaned, err := gitfilter.Clean(buildFile, pointerFile, -1, nil)
+			if err != nil {
+				Error(err.Error())
+				buildFile.Close()
+				os.Exit(1)
+			}
+
+			ptr = cleaned.Pointer
+		}
 		buildFile.Close()
 
-		if err != nil {
-			Error(err.Error())
-			os.Exit(1)
+		fmt.Fprint(os.Stderr, tr.Tr.Get("Git LFS pointer for %s", pointerFile), "\n")
+		if !pointerNoExtensions && hasExtensionsConfig {
+			fmt.Fprint(os.Stderr, tr.Tr.Get("warning: Using LFS extensions, use --no-extensions for a plain pointer."), "\n")
 		}
+		fmt.Fprint(os.Stderr, "\n")
 
-		ptr := lfs.NewPointer(hex.EncodeToString(oidHash.Sum(nil)), size, nil)
-		fmt.Fprint(os.Stderr, tr.Tr.Get("Git LFS pointer for %s", pointerFile), "\n\n")
 		buf := &bytes.Buffer{}
 		lfs.EncodePointer(io.MultiWriter(os.Stdout, buf), ptr)
 
@@ -115,7 +138,7 @@ func pointerCommand(cmd *cobra.Command, args []string) {
 
 		buf := &bytes.Buffer{}
 		tee := io.TeeReader(compFile, buf)
-		_, err = lfs.DecodePointer(tee)
+		comparePointer, err = lfs.DecodePointer(tee)
 		compFile.Close()
 
 		pointerName := "STDIN"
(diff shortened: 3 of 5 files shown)


git/gitattr/tree.go
```diff
@@ -34,56 +48,65 @@ func New(db *gitobj.ObjectDatabase, t *gitobj.Tree) (*Tree, error) {
 
 		// For every entry in the current tree, parse its sub-trees to
 		// see if they might contain a .gitattributes.
-		t, err := db.Tree(entry.Oid)
+		subT, err := db.Tree(entry.Oid)
```
chrisd8088: This is a pretty minor thought, but I'd be inclined to suggest a name other than `subT` here, only because it's neither a short single-letter name nor a more fully descriptive one.

I might just stick with the existing reuse of the `t` variable name, honestly.  (It's obviously not ideal, though, since it means we're still masking the `t` function argument.)

If that doesn't sit well with you, perhaps a fully descriptive name like `subGitTree`, and renaming the function `newFromGitTree()`?

git/gitattr/tree.go
```diff
@@ -99,11 +122,10 @@ func (t *Tree) Applied(to string) []*Attr {
 		}
 	}
 
-	splits := strings.SplitN(to, "/", 2)
-	if len(splits) == 2 {
-		car, cdr := splits[0], splits[1]
-		if child, ok := t.Children[car]; ok {
-			attrs = append(attrs, child.Applied(cdr)...)
+	dirPath, remainingPath, found := strings.Cut(to, "/")
```
chrisd8088, line 157: This is a nice simplification!  (I'm a bit sad to lose the Lisp [terminology](https://en.wikipedia.org/wiki/CAR_and_CDR), but I think your names are going to be more helpful for most people.)

git/gitattr/tree.go
```diff
@@ -11,18 +12,31 @@ import (
...
+
+	SystemAttributes *Tree
+
+	UserAttributes *Tree
 	// Lines are the lines of the .gitattributes at this level of the tree.
 	Lines []Line
 	// Children are the named child directories in the repository.
 	Children map[string]*Tree
+
+	RepoAttributes *Tree
```
chrisd8088: I believe the `Lines` and `Children` elements of this structure are not actually accessed outside of our `gitattr` package, and so could instead have lowercase names like `lines` and `children`.

Since the new elements being added here will also not be used outside of the package, I'd suggest using lowercase names for them like `mp`, `systemAttributes`, etc., so we avoid unnecessarily exporting the fields.

git/gitattr/tree.go
```diff
@@ -114,18 +114,50 @@ func linesInTree(db *gitobj.ObjectDatabase, t *gitobj.Tree, mp *MacroProcessor)
...
+		}
+		if t.UserAttributes != nil {
+			t.MP.ProcessLines(t.UserAttributes.Lines, true)
+		}
+		t.MP.ProcessLines(t.Lines, true)
+
+		if t.RepoAttributes != nil {
+			t.MP.ProcessLines(t.RepoAttributes.Lines, true)
+		}
+	}
```
chrisd8088: IIUC, the purpose of these steps is to load all the macros found in the various `gitattributes(5)` files (in the appropriate order) using just the `MacroLine` [case](https://github.com/git-lfs/git-lfs/blob/071e19e8ea03b1e40b181706909fb8c18d928e29/git/gitattr/macro.go#L62-L65) in the `ProcessLines()` method.

Since we don't care about the other lines in the files at this stage, the `PatternLine` [case](https://github.com/git-lfs/git-lfs/blob/071e19e8ea03b1e40b181706909fb8c18d928e29/git/gitattr/macro.go#L30-L61) is overhead which might be skipped.  Perhaps we could substitute for the whole `ProcessLines()` method a simpler `ProcessMacros()` method which only looked for macro definitions and ignored all other lines, and then use that new method here?

git/gitattr/tree.go
Diamondemon: You're right, I was mainly motivated by the idea of not shadowing the `t` variable name from the function scope. So yeah, the name was not thought through at all.
I also agree with the `newFromGitTree` name for the recursive function. Should I also change `New` to `NewFromGitTree`? I was trying to keep the names as they were but this may add confusion now :/

git/gitattr/tree.go
Diamondemon: Actually, you're right, they don't have much of a reason for being public apart from the fact that I kept the original scheme. Even in my vision I don't see any reason why they would need to be used as public attributes.

git/gitattr/tree.go
Diamondemon: I was trying to minimize changes and thought "oh well since it happens only once it does not matter much". However that may be a big overlook in the case of many pattern lines with only few macros (which is generally the case I would assume).

Diamondemon:
> If those two tests now succeed, it would be great to then also expand the `t/t-migrate*.sh` test suite a bit further to check the behaviour of our `git lfs migrate` commands with attribute macros defined in various other ways, such as in the `$GIT_DIR/info/attributes` file or the `$HOME/.config/git/attributes` file, and with references to those attributes in `.gitattributes` files at various levels of the test repositories' directory hierarchies. (It should be safe to write to both of those files in our test scripts, because we reset `$GIT_DIR` and `$HOME` to temporary locations, and we unset `$XDG_CONFIG_HOME`.)

As I said, this I not something that can be done in this PR as finding the special attributes files and processing them the right way is done in the `git.go` file of the `git` package. Using the file-finding logic from a function in `gitattr/tree.go` creates a dependency loop `git <--> gitattr`. Plus the use of the `config.Environment` interface it would imply creates another dependency loop `git --> gitattr --> config --> git`.
I have already done such refactoring work on local/personal instances, but this is a gross modification and I will share those changes in a later draft PR. I will need your input as these may be seen as drastic changes.

> One particular change which I think would be really valuable to add to this PR would be to uncomment the `migrate info (--fixup, .gitattributes with LFS macro)` [test](https://github.com/git-lfs/git-lfs/blob/071e19e8ea03b1e40b181706909fb8c18d928e29/t/t-migrate-fixup.sh#L177-L196) in our `t/t-migrate-fixup.sh` shell test script and see if it now passes!
> 
> Likewise, does the `migrate info (potential fixup, --fixup, .gitattributes with LFS macro)` [test](https://github.com/git-lfs/git-lfs/blob/071e19e8ea03b1e40b181706909fb8c18d928e29/t/t-migrate-info.sh#L811-L831) in the `t/t-migrate-info.sh` script now also work?

I will study that case. I did not look into the shell testing files. Thanks for reminding me they do in fact exist.

> FWIW, there are also some other commented-out [lines](https://github.com/git-lfs/git-lfs/blob/071e19e8ea03b1e40b181706909fb8c18d928e29/t/t-attributes.sh#L130-L146) in the `t/t-attributes.sh` test script because the `git lfs track` command doesn't fully support macro attribute definitions in the same manner as Git. I assume this PR won't fix that problem, but it's a great step in the right direction, so thank you again for all your help!

Yes exactly. However, I already have some work on that on my personal instances. My vision so far has been to turn the `Tree` system into the lawmaker when it comes to processing attributes files (how to find them, in which order to process their macros and patterns, so on and so forth).
This implies adding some logic considering that when converting attributes, `git lfs track` ignores pattern lines (but not  macro lines) coming from the special attributes files.
For now, as this is still a WIP, I added conversion from `Tree` to `[]AttributePath` as I have not completely identified how to make a `Tree` that would be transparent with the current use of `AttributePath`. It is also built on the refactor I had to make, which is still susceptible to many changes after I'll draft that one PR.
So yeah, all of this will be coming in later PRs once this one has been completed, corrected and merged.

git/gitattr/tree.go
chrisd8088: > Should I also change New to `NewFromGitTree`?

I think it's fine the way it is, at least for now.  I suppose ideally we might rename it, but let's keep the changes localized to the `gitattr` package for this PR, and if we want to rename the `New()` function, make that a discrete change in a separate PR.

chrisd8088:
> I have already done such refactoring work on local/personal instances, but this is a gross modification and I will share those changes in a later draft PR. I will need your input as these may be seen as drastic changes.

> So yeah, all of this will be coming in later PRs once this one has been completed, corrected and merged.

Understood!  Thanks for experimenting with these revisions, and I look forward to seeing them in some future PR!

git/gitattr/tree.go
chrisd8088: Thanks for making a change along these lines in this PR!  I think the advantage is mainly going to be that this really clarifies the intent of each method; the performance gains will be minimal in most cases.

I personally tend to find it a bit confusing when I try to understand code which seems to be doing important work, but actually just discards some of the results.  We probably have a few examples of this lurking on our codebase already, so if we can avoid adding another, so much the better.  Thanks again!

chrisd8088:
I gave this PR's changes a try with the two commented-out tests in the `t/t-migrate-info.sh` and `t/t-migrate-fixup.sh` test scripts, and it works, which is great!

There is a bug in the final [line](https://github.com/git-lfs/git-lfs/blob/071e19e8ea03b1e40b181706909fb8c18d928e29/t/t-migrate-fixup.sh#L194) of the `migrate info (--fixup, .gitattributes with LFS macro)` test I added back in PR #5382; it should instead be replaced with these two lines:

```shell
  echo "$main_attrs" | grep -q "\[attr\]lfs filter=lfs diff=lfs merge=lfs"
  echo "$main_attrs" | grep -q "*\.txt lfs"
```

Once that's done, the changes in this PR are otherwise sufficient to allow the test to pass, and the `migrate info (potential fixup, --fixup, .gitattributes with LFS macro)` [test](https://github.com/git-lfs/git-lfs/blob/071e19e8ea03b1e40b181706909fb8c18d928e29/t/t-migrate-info.sh#L813-L831) passes without any adjustment.

If you wouldn't mind uncommenting those two tests, removing the `NOTE:` comment lines above them, and fixing my bug with the two lines noted above, I think that's all we need to get this PR over the finish line.

Thanks again for all the help!

chrisd8088:
I've added one commit which just enables the two shell tests I left commented-out back in PR #5382 and fixes a couple of minor issues with one of them, so I think we're good to merge this PR.

Thank you again for all the help!

Issue #5937, opened 2024-12-17
Markup remains in the help text


mslinn:
Help text for `git-lfs` can be generated by simply typing `git-lfs` at a command prompt. Some of the resulting help text appears to have formatting left in it. For example, notice the dots and + characters in this exerpt:

```
Examples
--------

To get started with Git LFS, the following commands can be used.

. Setup Git LFS on your system. You only have to do this once per user
account:
+

git lfs install

. Choose the type of files you want to track, for examples all ISO
images, with git lfs track:
+

git lfs track "*.iso"

. The above stores this information in gitattributes(5) files, so that
file needs to be added to the repository:
+

git add .gitattributes

. Commit, push and work with the files normally:
+
```

chrisd8088:
Hey, thanks for the report!  This looks like something we should be able to improve, and I suspect there are some other formatting anomalies in our manual pages as well.

The code which converts our AsciiDoctor manual page source files to text is in the `docs/man/mangen.go` [file](https://github.com/git-lfs/git-lfs/blob/fba9303dee3c94751e5ecdccb294c07c27e1fcb5/docs/man/mangen.go), if anyone wants to take a look and try to craft a PR to address this issue.

In the meantime, I'll put this into our backlog as an improvement we'd like to make.

slonopotamus:
Would you consider using a [proper AsciiDoc -> man converter](https://docs.asciidoctor.org/asciidoctor/latest/manpage-backend/) or that is a too big dependency?

chrisd8088:
> Would you consider using a [proper AsciiDoc -> man converter](https://docs.asciidoctor.org/asciidoctor/latest/manpage-backend/) or that is a too big dependency?

@slonopotamus — Since PR #5054, we do [use](https://github.com/git-lfs/git-lfs/blob/97f47b9cd2102273287066e0225a6c10457c401a/Makefile#L60-L63) `asciidoctor` for this purpose already, to [generate](https://github.com/git-lfs/git-lfs/blob/97f47b9cd2102273287066e0225a6c10457c401a/Makefile#L724-L732) the `roff` format and HTML format manual pages we include in release packages.

[...]

PR #6138, opened 2025-11-06
Fix manual page formatting errors from AsciiDoc conversion

chrisd8088:
In PR #5054 we migrated our manual page source files from a format supported by the `ronn` [utility](https://github.com/rtomayko/ronn) to the AsciiDoc format, and while this conversion was extremely successful, a handful of formatting errors were introduced by the scripts we used.

We detected many of those errors at the time and repaired them in the same PR, and have corrected a few others in PRs #5414 and #5944.

This PR corrects all of remaining errors of this type of which we are presently aware, and also removes some `RONN*` variables from our `Makefile` which are no longer relevant.

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -40,13 +40,11 @@ func checkoutCommand(cmd *cobra.Command, args []string) {
 		Exit(tr.Tr.Get("Error parsing args: %v", err))
 	}
 
-	rootedPaths := rootedPaths(args)
-
 	if checkoutTo != "" && stage != git.IndexStageDefault {
 		if len(args) != 1 {
 			Exit(tr.Tr.Get("--to requires exactly one Git LFS object file path"))
 		}
-		checkoutConflict(rootedPaths[0], stage)
+		checkoutConflict(args[0], stage)
 		return
 	} else if checkoutTo != "" || stage != git.IndexStageDefault {
 		Exit(tr.Tr.Get("--to and exactly one of --theirs, --ours, and --base must be used together"))
@@ -56,6 +54,8 @@ func checkoutCommand(cmd *cobra.Command, args []string) {
 	if err != nil {
 		Panic(err, tr.Tr.Get("Could not checkout"))
 	}
+
+	rootedPathPatterns := rootedPathPatterns(args)
 
 	// will chdir to root of working tree, if one exists
 	singleCheckout := newSingleCheckout(cfg.Git, "")
@@ -85,7 +85,7 @@ func checkoutCommand(cmd *cobra.Command, args []string) {
 		pointers = append(pointers, p)
 	})
 
-	chgitscanner.Filter = filepathfilter.New(rootedPaths, nil, filepathfilter.GitIgnore)
+	chgitscanner.Filter = filepathfilter.New(rootedPathPatterns, nil, filepathfilter.GitIgnore)
 
 	if err := chgitscanner.ScanLFSFiles(ref.Sha, nil); err != nil {
 		ExitWithError(err)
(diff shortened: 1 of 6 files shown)


docs/man/git-lfs-migrate.adoc
```diff
@@ -193,8 +192,9 @@ that set will be shown.
   change this behavior.
 `--unit=<unit>`::
   Format the number of bytes in each entry as a quantity of the storage unit
-  provided. Valid units include: * b, kib, mib, gib, tib, pib - for IEC storage
-  units * b, kb, mb, gb, tb, pb - for SI storage units
+  provided. Valid units include:
+** b, kib, mib, gib, tib, pib - for IEC storage units
+** b, kb, mb, gb, tb, pb - for SI storage units
```
larsxschneider, line 197: The `**` does not need to be indented here?

chrisd8088: Excellent question!  This PR is really just a first step in a series of corrections, and so I only included a basic change here to untangle the lines that were inadvertently concatenated by the conversion to AsciiDoc, and align them with similar formatting in the `git-lfs-config(5)` manual page.

This change alone improves the rendering somewhat, and no, the `**` lines don't need to be indented for AsciiDoctor to indent them for us (this screenshot is from a `man 1 git-lfs-migrate` command's output):

Screenshot 2025-11-10 at 1 19 24 PM

But there's at least one remaining problem, in that the trailing paragraph is now indented as if it pertained solely to the second bullet point.

And things are worse in our `git-lfs-config(5)` manual page, which [includes](https://github.com/git-lfs/git-lfs/blob/6f73c9bbb9c01c9228f43217b0c856dadce71e9d/docs/man/git-lfs-config.adoc?plain=1#L304-L307) [three](https://github.com/git-lfs/git-lfs/blob/6f73c9bbb9c01c9228f43217b0c856dadce71e9d/docs/man/git-lfs-config.adoc?plain=1#L304-L307) [lists](https://github.com/git-lfs/git-lfs/blob/6f73c9bbb9c01c9228f43217b0c856dadce71e9d/docs/man/git-lfs-config.adoc?plain=1#L381-L387) of the same kind, all of which render very poorly right now, for instance:

Screenshot 2025-11-10 at 1 36 02 PM

In brief, we should be using AsciiDoc [list continuations](https://docs.asciidoctor.org/asciidoc/latest/lists/continuation/#list-continuation) here, combined with [open blocks](https://docs.asciidoctor.org/asciidoc/latest/blocks/open-blocks/).  Ideally we might reformat these top-level entries to use a single list continuation, followed by an open block which contains the entire contents of the entry, including our nested lists along with all the separate paragraphs of text.  This would be relatively elegant, but would require more extensive changes in our `mangen` [program](https://github.com/git-lfs/git-lfs/blob/6f73c9bbb9c01c9228f43217b0c856dadce71e9d/docs/man/mangen.go) to recognize this alternative formatting technique for multi-paragraph list entries.  At the moment, when parsing lists, our program simply [discards](https://github.com/git-lfs/git-lfs/blob/6f73c9bbb9c01c9228f43217b0c856dadce71e9d/docs/man/mangen.go#L198-L199) lines with a single `+` character, and expects that an empty list signals the end of list.

The approach I plan to take instead in subsequent PRs is to enclose just our second-level lists in open blocks with the appropriate `--` delimiters on separate lines before and after the lists, attach these open blocks to the preceding paragraphs in their top-level list entries using a `+` list continuation line, and attach any trailing paragraphs which need to be out-dented using [ancestor list continuations](https://docs.asciidoctor.org/asciidoc/latest/lists/continuation/#ancestor-list-continuation), which require an extra empty line.  That means this section should eventually look something like:

```asciidoc
`--unit=<unit>`::
  Format the number of bytes in each entry as a quantity of the storage unit
  provided. Valid units include:
+
--
** `b`, `kib`, `mib`, `gib`, `tib`, `pib` - Used for IEC storage units.
** `b`, `kb`, `mb`, `gb`, `tb`, `pb` - Used for SI storage units.
--

+
If a `--unit` is not specified, the largest unit that can fit the number
of counted bytes as a whole number quantity is chosen.
```

The `mangen` program will have to be updated to support this formatting; I hope to include those updates as part of any work we do to address #5937.

RatStar811 (approved):

PR #6139, opened 2025-11-06
Feature schannel sslcert

fredriksimon:


chrisd8088:
Hey, thanks very much for this PR, it looks really interesting!

Would you be able to flesh out the PR description to explain what problem this PR solves, and provide some context on how it does so?

For instance, it seems like `NCryptSignHash()` is the key Windows API [function](https://learn.microsoft.com/en-us/windows/win32/api/ncrypt/nf-ncrypt-ncryptsignhash) used in these changes.  If you could include a link to any documentation on how that works and why it's needed here, that would be great.

As a quick first note, I noticed that the current version of this PR calls the `windows.MustLoadDLL()` Go [function](https://pkg.go.dev/golang.org/x/sys/windows#MustLoadDLL), which will panic if it is unable to load the DLL.  From the information I could find, it appears that all recent versions of Windows include `ncrypt.dll`, so this is probably fine, but is there any chance we'd want to gracefully handle the condition where that DLL was absent by using the `windows.LoadDLL()` function instead and checking the `error` value it returns?

Before we could merge this PR, we'll want to make sure it includes appropriate tests of its new functionality, either in our Go test suite or, maybe more likely, in our Bash shell test suite.

The `t/t-clone.sh` script, for instance, [includes](https://github.com/git-lfs/git-lfs/blob/6f73c9bbb9c01c9228f43217b0c856dadce71e9d/t/t-clone.sh#L113-L444) a number of our current tests of various kinds of SSL/TLS certificate support.  The certificates for these tests are [generated](https://github.com/git-lfs/git-lfs/blob/6f73c9bbb9c01c9228f43217b0c856dadce71e9d/t/cmd/lfstest-gitserver.go#L76-L91) by the `t/cmd/lfstest-gitserver.go` program we use to simulate a Git and Git LFS server in our shell test suite.  I imagine we'd want to expand this framework to generate whatever types of certificates are required to validate the new code added in this PR.

Thank you again very much for your interest in Git LFS and for contributing to the project!

fredriksimon:
I didn’t mean to push this to the git-lfs repo — still getting used to GitHub. I was just trying to get the CI tests running in our fork. 🙂
I’ll create a new pull request once I’ve addressed your suggestions.

chrisd8088:
> I didn’t mean to push this to the git-lfs repo — still getting used to GitHub. I was just trying to get the CI tests running in our fork. 🙂

Hey, no worries at all!  I'm glad you're working on Git LFS and hope to see whatever you come up if you think it's valuable to share with the community.  Good luck!

PR #6142, opened 2025-11-14
Detect invalid object IDs and repo names in test functions

chrisd8088:
This PR updates a number of our shell test helper and assertion functions so that they verify that the parameters they are passed are not empty or invalid.

In particular, the test helper and assertion functions which could succeed even when given an empty or invalid value in place of their `oid` parameter now check that this parameter contains a 64-character hexadecimal value.

As well, those functions which could succeed even if their `reponame` parameter did not correspond to a remote test repository will now fail if a directory does not exist with the given repository name.

These changes implement a suggestion from @larsxschneider during the development of the v3.7.1 release, when we corrected an error in the initial `pull` test of our `t/t-pull.sh` test script where we failed to pass a valid `oid` parameter to the `refute_server_object()` assertion.  (For details, see commit b7e4910ccb420252101203b2f296eb49303760de of our `main` development branch or commit 779d7a3e40db56f877e6159059f99d06bd516ba6 of our `release-3.7` branch.)

The additional checks in this PR revealed another, similar error, in this case in the `init fetch unclean paths` test in our `t/t-fetch-paths.sh` test script where we accidentally [call](https://github.com/git-lfs/git-lfs/blob/b4833940ef3cf2e4440ccf023b60f024e3f25718/t/t-fetch-paths.sh#L33) the `refute_server_object()` assertion without passing a `reponame` parameter.

We therefore correct this problem, and also update the tests in the `t/t-credentials-protect.sh` script so they create remote repositories with the same names as they configure in the `lfs.url` option, since they then pass those repository names to the `refute_server_object()` function and it now requires that its `reponame` parameter identifies an extant Git repository.

### DIFF
--- a/t/t-credentials-protect.sh
+++ b/t/t-credentials-protect.sh
@@ -34,6 +34,12 @@ begin_test "credentials rejected with line feed"
   testreponame="test%0a$reponame"
   git config lfs.url "$gitserver/$testreponame.git/info/lfs"
 
+  # The refute_server_object() assertion below will fail unless its
+  # repository name parameter matches an extant "remote" directory.
+  pushd .
+    setup_remote_repo "$testreponame"
+  popd
+
   GIT_TRACE=1 git lfs push origin main 2>&1 | tee push.log
   if [ "0" -eq "${PIPESTATUS[0]}" ]; then
     echo >&2 "fatal: expected 'git lfs push' to fail ..."
@@ -79,6 +85,12 @@ begin_test "credentials rejected with carriage return"
   testreponame="test%0d$reponame"
   git config lfs.url "$gitserver/$testreponame.git/info/lfs"
 
+  # The refute_server_object() assertion below will fail unless its
+  # repository name parameter matches an extant "remote" directory.
+  pushd .
+    setup_remote_repo "$testreponame"
+  popd
+
   GIT_TRACE=1 git lfs push origin main 2>&1 | tee push.log
   if [ "0" -eq "${PIPESTATUS[0]}" ]; then
     echo >&2 "fatal: expected 'git lfs push' to fail ..."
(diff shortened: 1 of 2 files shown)


t/t-credentials-protect.sh
```diff
@@ -34,6 +34,12 @@ begin_test "credentials rejected with line feed"
   testreponame="test%0a$reponame"
   git config lfs.url "$gitserver/$testreponame.git/info/lfs"
 
+  # The refute_server_object() assertion below will fail unless its
+  # repository name parameter matches an extant "remote" directory.
```
larsxschneider, line 38: I thought I spotted the first ever typo in your comments. But I was wrong and learned https://www.merriam-webster.com/dictionary/extant ☺️

t/t-credentials-protect.sh
```diff
@@ -34,6 +34,12 @@ begin_test "credentials rejected with line feed"
...
+  pushd .
+    setup_remote_repo "$testreponame"
```
larsxschneider, line 40: I am surprised that this test works on Windows. I would expect Windows file systems to fail on newline or CR (below) characters.

chrisd8088: Indeed, these tests would fail on Windows if we had bare LF, CR, or null characters in the repository names, but because those characters are [URL-encoded](https://github.com/git-lfs/git-lfs/blob/b4833940ef3cf2e4440ccf023b60f024e3f25718/t/t-credentials-protect.sh#L34), the actual directory names just have `%` in them, which Windows allows.

When those repository names are used in URLs, though, Git LFS decodes the characters before passing the request paths to any Git credential helper, which is where the security problem occurred that we fixed in our v3.6.1 release.

PR #6148, opened 2025-11-22
Hierarchy rework

Diamondemon:
### What: Reorganizing the hierarchy

This pull request aims to reorganize package hierarchy to dependency loops that appear when trying to avoid knowledge duplication.

### Why?

When working on the fix for `git lfs migrate import --fixup` #6126 and all issues mention in discussion #6105  I realized that trying to also integrate the support of special attributes resulted in a few dependency loops, mainly the two following:
- Using the "special file"-finding logic from a function in `gitattr/tree.go` creates a dependency loop `git <--> gitattr`
- Using the `config.Environment` interface for the previous point creates another dependency loop `git --> gitattr --> config --> git`

### Proposed solution

The creation of a `git/core` package that contains files  with core git logic that a lot of other packages depend on AND only depend on very atomic packages like the `tr` package or the `subprocess` package.

### Why a draft?

Although I tried to keep the movement minimal, the result is very coarse and rough. All movements of files were motivated mainly by dependencies to the `config.Environment` interface and the `git/git.go` file. However this resulted in the movement of other files that these two files relied on, without much afterthought. This small change then creates a modification of almost all files of this repository, underlying the importance of those few files.

Despite all packages being explicitly documented as non-api, such drastic changes deserve a deeper look with broader knowledge I do not have. For example, I single-handedly decided to erase any mention of an `Env(ironment)` interface  other than `config.Environment` for unification more because I was confused by the many numbers of re-definitions that were done, than functionality, and even though these re-definitions did not represent much of a "knowledge decentralization" so to speak.

I look forward to any insight on these changes (@chrisd8088)

### DIFF
--- a/tq/custom.go
+++ b/tq/custom.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/git-lfs/git-lfs/v3/errors"
 	"github.com/git-lfs/git-lfs/v3/fs"
+	"github.com/git-lfs/git-lfs/v3/git/core"
 	"github.com/git-lfs/git-lfs/v3/tools"
 	"github.com/git-lfs/git-lfs/v3/tr"
 
@@ -351,7 +352,7 @@ const (
 	standaloneFileName = "lfs-standalone-file"
 )
 
-func configureDefaultCustomAdapters(git Env, m *concreteManifest) {
+func configureDefaultCustomAdapters(git core.Environment, m *concreteManifest) {
 	newfunc := func(name string, dir Direction) Adapter {
 		standalone := m.standaloneTransferAgent != ""
 		return newCustomAdapter(m.fs, standaloneFileName, dir, "git-lfs", "standalone-file", false, standalone)
@@ -361,7 +362,7 @@ func configureDefaultCustomAdapters(git Env, m *concreteManifest) {
 }
 
 // Initialise custom adapters based on current config
-func configureCustomAdapters(git Env, m *concreteManifest) {
+func configureCustomAdapters(git core.Environment, m *concreteManifest) {
 	configureDefaultCustomAdapters(git, m)
 
 	pathRegex := regexp.MustCompile(`lfs.customtransfer.([^.]+).path`)
--- a/tq/manifest.go
+++ b/tq/manifest.go
@@ -376,12 +376,3 @@ func (m *concreteManifest) NewDownloadAdapter(name string) Adapter {
 func (m *concreteManifest) NewUploadAdapter(name string) Adapter {
 	return m.NewAdapterOrDefault(name, Upload)
 }
-
-// Env is any object with a config.Environment interface.
-type Env interface {
-	Get(key string) (val string, ok bool)
-	GetAll(key string) []string
-	Bool(key string, def bool) (val bool)
-	Int(key string, def int) (val int)
-	All() map[string][]string
-}


chrisd8088:
Hey, thanks so much for pursuing this initiative, and for asking for some feedback at this stage!

I suppose I have two somewhat separate sets of thoughts, neither of which is entirely fully-formed.  :-)

I tried to guess where your work was going a little bit, based on the first two commits in this draft, and I wasn't able to replicate the circular package dependencies, so my comments are a little bit theoretical.  If you're able to point to a branch (doesn't have to be a full PR) which illustrates the problem you ran into, that would be great.

In summary, I'm hopeful we might be able to avoid a substantial reorganization, and if one is necessary, I put my initial thoughts on that subject further below.

#### Moving Git Attributes Functions

Looking at the first two commits in this draft to try to get a sense of the goal of the restructuring, I guessed that the principal first step was to consolidate all the functions related to Git attribute handling into the `git/gitattr` package.

With that idea in mind, I tried just moving the `git/attribs.go` source file into the `git/gitattr` package and fixing up the references which had to be changed to make the project compile again.  I then cherry-picked the first two commits from this draft on top of those changes.

In case that's helpful, my experiments can be found in the `gitattr-reorg-wip` [branch](https://github.com/chrisd8088/git-lfs/tree/gitattr-reorg-wip) in my fork of the project.

Is there any chance this is sufficient to unblock your other efforts to improve our handling of Git attributes?

#### Package Reorganization

Regarding a potential reorganization of the codebase, I'm certainly not opposed to that in principle.  (The last major reorganization was probably when we refactored the `lfsapi` package to extract the `lfshttp` package in PR #3244, and that proved to be very successful.)

However, I would be inclined to try to avoid shifting around too much code in the `git` package, if we can find any way to avoid it.  As I understand the issue you encountered, it goes like this:

- `config` depends on `git`
- `git` depends on `gitattr`
- if `gitattr` is changed to depend on `config`, we get: `gitattr` ⇒ `config` ⇒ `git` ⇒ `gitattr`

If we have to extract the `Environment` structure type and methods from the `config` package, my first instinct would be to create a `config/env` package to contain it, and then reference that new, simple `env` package from the `config`, `fs`, `git`, and `tq` packages.

In particular, I'd like to avoid having the fairly low-level and minimal `fs` package depend on the much more substantial and complex `git` package.

As for a `git/core` package, if such a thing is necessary, I'd propose the name `git/gitcore`, as I think that would clarify what "core" means in this context and would also align with the names `githistory` and `gitattr`.  (But I kind of hope don't need a `git/gitcore` package at all.)

Diamondemon:
> I tried to guess where your work was going a little bit, based on the first two commits in this draft, and I wasn't able to replicate the circular package dependencies, so my comments are a little bit theoretical. If you're able to point to a branch (doesn't have to be a full PR) which illustrates the problem you ran into, that would be great.

Yeah I actually didn't think about that... I'll try to replicate the most minimal issue that I was having with both dependency loops covered.

> In case that's helpful, my experiments can be found in the `gitattr-reorg-wip` [branch](https://github.com/chrisd8088/git-lfs/tree/gitattr-reorg-wip) in my fork of the project.
> 
> Is there any chance this is sufficient to unblock your other efforts to improve our handling of Git attributes?

I'll look into that too.

> If we have to extract the `Environment` structure type and methods from the `config` package, my first instinct would be to create a `config/env` package to contain it, and then reference that new, simple `env` package from the `config`, `fs`, `git`, and `tq` packages.

Yeah I may have ended up mixing a bunch of things together unnecessarily.



> In particular, I'd like to avoid having the fairly low-level and minimal `fs` package depend on the much more substantial and complex `git` package.
> 
> As for a `git/core` package, if such a thing is necessary, I'd propose the name `git/gitcore`, as I think that would clarify what "core" means in this context and would also align with the names `githistory` and `gitattr`. (But I kind of hope don't need a `git/gitcore` package at all.)

I get it, let's keep the change to the minimum if it is not mandatory. I agree about the name, it would fit better (in case it would be needed of course).

Diamondemon:
So, uh, I realized the first commit of the branch has nothing to do with the actual focus of the branch... It was a start of work that I committed to not lose it when doing changes for my previous PR, but that's all. Thank god that's a draft.

Diamondemon:
Thanks @chrisd8088 for the second insight because I was so caught up in that dependency loop that I did not realize the only file that created it was actually `attribs.go` which anyway is relevant to move into the `gitattr` package. Since this means no more dependency from `git` to `gitattr`, none of the two loops I had exist anymore. I spent a lot of time on a quite useless endeavour, damn.

So, after retrying and confirming that I only need to move the `attribs.go` file, and checking [your branch](https://github.com/chrisd8088/git-lfs/tree/gitattr-reorg-wip) I came to this conclusion:

- I really like that `Var` function you added to the `git.go` [file](https://github.com/git-lfs/git-lfs/blob/846bb81440bba4318ecf4c7179fbda3bb6fcd52d/git/git.go#L240C1-L242C2).
- Maybe creating a simple `gitattr.Environment` with only the `Get` function is a simpler idea than depending on the big `config.Environment` on the long term (rejoining your comment on `fs` depending on a big package just because I had removed the definition of `fs.Environment`). I don't think the `gitattr` package will need to use methods other than `Get`.
- I am thinking of closing this draft PR and include the changes we mention here in the PR that also adds processing of special attribute files, so the benefits of such a move become obvious.

Are you good with that?

chrisd8088:
> I really like that `Var` function you added to the `git.go` [file](https://github.com/git-lfs/git-lfs/blob/846bb81440bba4318ecf4c7179fbda3bb6fcd52d/git/git.go#L240-L242).

Thanks!  I forgot to mention that detail in my terse work-in-progress commit message; my apologies.  That type of function is pretty typical of how we wrap various Git commands in the `git` package.

> I am thinking of closing this draft PR and include the changes we mention here in the PR that also adds processing of special attribute files, so the benefits of such a move become obvious.

Sure thing, that sounds like a good plan to me!

The way I try to approach revisions like this is to make each commit a discrete unit containing logically related changes, and where the whole test suite still passes after each commit.  So for instance, one commit might look very similar to what I tried in commit 846bb81440bba4318ecf4c7179fbda3bb6fcd52d, which was just to move `git/attribs.go` to `git/gitattr/attribs.go` and make the minimal alternations necessary to get everything to compile and pass the test suite again, like adding the `git.Var()` method and using it in the moved `attribs.go` file.  Then subsequent commits can start introducing new functionality, one logical piece at a time.

> I spent a lot of time on a quite useless endeavour

Hey, I hope you don't feel that too intensely!  We certainly appreciate people trying out new ideas, and you never know which ones will stick.  Over the years I have spent countless hours experimenting with changes that I ultimately discard, on all kinds of projects, but as a result I always end up with a better knowledge of the codebase I'm working with, so I find it pays dividends in the end.

PR #6151, opened 2025-11-26
Correct and improve `grep -v` and `grep -c` usage in tests

chrisd8088:
This PR revises and corrects a number of instances in our test suite where we use either the `-v` or `-c` options of the `grep(1)` command, especially those in which we mistakenly depend on the `-v` option to cause the command to fail if an input line matches the given pattern.

For example, in our `t/t-migrate-import.sh` shell test script, we invoke `grep` with its `-v` option to try to verify that no `.gitattributes` entry has been created for the `*.txt` path pattern:

```shell
  echo "$main_attrs" | grep -vq "*.txt filter=lfs diff=lfs merge=lfs"
```

However, this simply filters any lines which match that pattern from the input; the command succeeds because there happen to be other lines in the input which do not match the pattern.  The `grep` command only returns a non-zero exit status code in the case of an error, or when no input lines are selected based on the command's options and arguments.

In some cases, our incorrect use of the `-v` option masks other problems in our test suite, so we correct those issues in the process of revising or removing the use of the `-v` option.

In many cases, we can replace the use of the `-v` option with the `-c` option, and then compare the line count reported by the `grep` command with the expected line count, whether that is zero or a positive integer.

We also replace a number of instances where we check that the output of the `grep` command is empty with checks that the line count reported by the `-c` option is zero, because the latter technique is more robust against accidental regressions.

We then revise all the instances where we use the `-c` option so they follow a consistent idiom, and simplify instances where we check the line count with the `wc(1)` command so we use the `grep` command's `-c` option instead.

Note that in this PR we intentionally avoid rewriting the patterns used in our `grep` commands, even when improvements or corrections would be valuable, and defer that work to a future PR.  Likewise, we also avoid changing the behaviour of any Git LFS commands.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
         go: ['1.25.x']
     runs-on: ${{ matrix.os }}
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         fetch-depth: 0
         persist-credentials: false
@@ -48,7 +48,7 @@ jobs:
         go: ['1.24.x']
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         fetch-depth: 0
         persist-credentials: false
@@ -61,7 +61,7 @@ jobs:
     name: Build on Windows
     runs-on: windows-latest
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         fetch-depth: 0
         persist-credentials: false
@@ -122,7 +122,7 @@ jobs:
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         fetch-depth: 0
         persist-credentials: false
@@ -146,7 +146,7 @@ jobs:
         os: [ubuntu-latest, macos-latest]
     runs-on: ${{ matrix.os }}
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         fetch-depth: 0
         persist-credentials: false
@@ -167,7 +167,7 @@ jobs:
     name: Build Linux packages
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         fetch-depth: 0
         persist-credentials: false
@@ -179,7 +179,7 @@ jobs:
     name: Build Linux ARM packages
     runs-on: ubuntu-24.04-arm
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         fetch-depth: 0
         persist-credentials: false
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         go: ['1.25.x']
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         fetch-depth: 0
         persist-credentials: false
@@ -102,7 +102,7 @@ jobs:
       matrix:
         go: ['1.25.x']
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         fetch-depth: 0
         persist-credentials: false
@@ -139,7 +139,7 @@ jobs:
       matrix:
         go: ['1.25.x']
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         fetch-depth: 0
         persist-credentials: false
@@ -175,7 +175,7 @@ jobs:
     environment: production
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         fetch-depth: 0
         persist-credentials: false
@@ -192,7 +192,7 @@ jobs:
     name: Build Linux ARM packages
     runs-on: ubuntu-24.04-arm
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         fetch-depth: 0
         persist-credentials: false
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -413,6 +413,6 @@ func init() {
 		cmd.Flags().BoolVarP(&trackNoModifyAttrsFlag, "no-modify-attrs", "", false, "skip modifying .gitattributes file")
 		cmd.Flags().BoolVarP(&trackNoExcludedFlag, "no-excluded", "", false, "skip listing excluded paths")
 		cmd.Flags().BoolVarP(&trackFilenameFlag, "filename", "", false, "treat this pattern as a literal filename")
-		cmd.Flags().BoolVarP(&trackJSONFlag, "json", "", false, "print output in JSON")
+		cmd.Flags().BoolVarP(&trackJSONFlag, "json", "j", false, "print output in JSON")
 	})
 }
--- a/docs/man/git-lfs-dedup.adoc
+++ b/docs/man/git-lfs-dedup.adoc
@@ -23,6 +23,13 @@ contents before they are written to the Git LFS storage directory, and
 therefore the working tree files should not be copy-on-write clones of
 the LFS object files.
 
+== OPTIONS
+
+`-t`::
+`--test`::
+  Checks whether the current operating system and file system support
+  copy-on-write file creation, without actually cloning any Git LFS files.
+
 == SEE ALSO
 
 Part of the git-lfs(1) suite.
--- a/docs/man/git-lfs-fetch.adoc
+++ b/docs/man/git-lfs-fetch.adoc
@@ -27,9 +27,11 @@ This does not update the working copy.
 `--exclude=<paths>`::
   Specify lfs.fetchexclude just for this invocation; see
   <<_include_and_exclude>>.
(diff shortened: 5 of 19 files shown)


t/t-migrate-import.sh
```diff
@@ -106,7 +106,7 @@ begin_test "migrate import (default branch with filter)"
   [ -z "$(git cat-file -p "$feature:.gitattributes")" ]
 
   echo "$main_attrs" | grep -q "*.md filter=lfs diff=lfs merge=lfs"
-  echo "$main_attrs" | grep -vq "*.txt filter=lfs diff=lfs merge=lfs"
+  [ 0 -eq "$(echo "$main_attrs" | grep -c "*.txt filter=lfs diff=lfs merge=lfs")" ]
```
larsxschneider, line 109: Should we use single quotes with the grep here to avoid * expansion? Which according to your commit message does not happen because * is the first character ... but I believe single quotes make this more explicit?!

t/t-migrate-import.sh
```diff
@@ -195,7 +195,7 @@ begin_test "migrate import (default branch, exclude remote refs)"
   [ -z "$(git cat-file -p "$remote:.gitattributes")" ]
 
   echo "$main_attrs" | grep -q "*.md filter=lfs diff=lfs merge=lfs"
-  echo "$main_attrs" | grep -vq "*.txt filter=lfs diff=lfs merge=lfs"
+  echo "$main_attrs" | grep -q "*.txt filter=lfs diff=lfs merge=lfs"
 )
```
larsxschneider, line 199: Nit: Should we just make this to emphasize that the file types are not relevant here?
```
echo "$main_attrs" | grep -q "filter=lfs diff=lfs merge=lfs"
```

t/t-pre-push.sh
```diff
@@ -363,7 +363,7 @@ begin_test "pre-push allow missing object (found on server)"
     tee push.log
 
   # make sure there were no errors reported
-  [ -z "$(grep -i 'Error' push.log)" ]
+  [ 0 -eq "$(grep -c -i 'Error' push.log)" ]
```
larsxschneider, line 366: Nit: I think this would succeed if the `push.log` is not created due to a previous error. We likely would abort with a previous error and therefore that should be no issue.

t/t-credentials.sh
```diff
@@ -57,15 +57,15 @@ begin_test "credentials without useHttpPath, with bad path password"
...
-  [ "1" -eq "$(cat push.log | grep "creds: git credential fill" | wc -l)" ]
+  [ 1 -eq "$(grep -c "creds: git credential fill" push.log)" ]
 
   echo "credential calls have no path:"
   credcalls="$(grep "creds: git credential" push.log)"
-  [ "0" -eq "$(echo "$credcalls" | grep "no-httppath-bad-password" | wc -l)" ]
+  [ 0 -eq "$(echo "$credcalls" | grep -c "no-httppath-bad-password")" ]
   expected="$(echo "$credcalls" | wc -l)"
-  [ "$expected" -eq "$(printf "%s" "$credcalls" | grep '", "")' | wc -l)" ]
+  [ "$expected" -eq "$(printf "%s" "$credcalls" | grep -c '", "")')" ]
```
larsxschneider, line 68:
```sh
$ credcalls='foo", "")bar", "")\nbaz", "")"'

# %s never prints newlines
$ printf "%s" "$credcalls"
foo", "")bar", "")\nbaz", "")"%

# grep -c increments if the string is at least found once in a line
$ printf "%s" "$credcalls" | grep -c '", "")'
1
```

I believe that means this check works for the case that we expect 1 cred call. Is this good enough or does this need to be more robust?

t/t-migrate-import.sh
chrisd8088: Well, there's two kinds of expansion to consider, shell expansion and then the regular expression matching.

Whether we use single or double quotes, it should be enough to prevent the shell expanding the `*.txt` to match any files that happen to exist with that filename pattern.  The shell will pass the entire quoted string (`*.txt filter=lfs diff=lfs merge=lfs`) to the `grep` command as a single argument.

As for the regular expression parser, what I mentioned in my commit message was that because the `*` is the first character in the expression, it's not parsed as a metacharacter and instead as a literal.  However, the `.` is parsed as a metacharacter, so it technically matches any character and not just `.` characters.  As I noted, though, this is a fairly pervasive mistake in our test suite and deserves a separate PR to either use the `-F` flag to match strings as literals, or escape the characters as necessary in the regular expressions.  That kind of escape would be one that's processed not by the shell but by the `grep` command, of course.

All of which is just to say that I don't think we need single quotes here because it won't change the string as it's perceived by the `grep` command:

```
$ echo "foo" | grep "*"
$ echo "*foo" | grep "*"
*foo

$ echo "foo" | grep '*'
$ echo "*foo" | grep '*'
*foo
```

t/t-migrate-import.sh
chrisd8088: We could do that in a [few](https://github.com/git-lfs/git-lfs/blob/a18bd24111bbfb4d814cd26023d35136bfe80fc2/t/t-migrate-import.sh#L237-L238) [places](https://github.com/git-lfs/git-lfs/blob/a18bd24111bbfb4d814cd26023d35136bfe80fc2/t/t-migrate-import.sh#L278-L281) in this file and likely a [few](https://github.com/git-lfs/git-lfs/blob/a18bd24111bbfb4d814cd26023d35136bfe80fc2/t/t-migrate-export.sh#L254-L255) [others](https://github.com/git-lfs/git-lfs/blob/a18bd24111bbfb4d814cd26023d35136bfe80fc2/t/t-migrate-export.sh#L189-L192), for sure.

I wanted to try to avoid tinkering with the patterns as much as possible to keep the scope of this PR somewhat constrained to just fixing the immediate issues that arose after resolving the misuse of `grep -v`, and then making the use of `grep -c` consistent.  I realize that's a somewhat arbitrary distinction, but I thought I'd better draw the line somewhere.

When we fix the problems with metacharacters that aren't properly escaped or treated as literals with `-F`, though, most or all of these same patterns will turn up as candidates for those corrections, and I think that would be a good opportunity to combine some of these checks into simpler ones.  So for example, we can just do `grep -F "filter=lfs diff=lfs merge=lfs"` here as we don't need to match anything other than a fixed string.

t/t-pre-push.sh
chrisd8088: I agree it's not an issue.  One of the reasons I wanted to change over from using `-z` to see if a `grep` command's output was empty to using `grep -c` and checking that it reports `0` is that means we now catch problems like the one you describe, where an expected log file is absent.

If we forcibly remove the `push.log` file and run the test as it stands (with `-z`), the test incorrectly passes.  However, if we try the same thing with the revised version, we get a test failure:

```
#     ++ grep -c -i Error push.log
#     grep: push.log: No such file or directory
#     + '[' 0 -eq '' ']'
#     t-pre-push.sh: line 367: [: : integer expression expected
#     + test_status=2
```

In all cases, `grep` returns a non-zero status code, either a `1` if `push.log` exists and no matching lines are found, or `2` if `push.log` does not exist.  In all these cases, however, the non-zero status code won't cause the test to fail on its own, despite the fact that we've enabled the `errexit` shell attribute with `set -e`.  That's because the `grep` command is run inside a subshell in a `[` shell test builtin, and the subshell's status code isn't used as the result of the `[` builtin.

Instead, the result of the `[` test builtin depends on what it's testing, and when `push.log` is missing, the `grep` command will return nothing on `stdout`, so `-z ...` will succeed (even though there's an error message on `stderr`).  But the `0 -eq ...` will fail because it's expecting an integer and gets an empty string; hence the `integer expression expected` error message, which is what actually triggers the test failure with our revised approach.

So that's why I thought we should switch over to using this technique consistently.

t/t-credentials.sh
chrisd8088: In practice we see 3 lines which match the `creds: git credential` pattern in the log file, and they look like this:
```
12:27:59.054428 trace git-lfs: creds: git credential fill ("http", "127.0.0.1:50879", "")
12:27:59.087163 trace git-lfs: creds: git credential approve ("http", "127.0.0.1:50879", "")
12:27:59.161620 trace git-lfs: creds: git credential cache ("http", "127.0.0.1:50879", "")
```

So the value of the `$expected` variable is `3` (or more precisely, `       3`, because we've captured the output of `wc -l` and without removing the leading whitespace that command outputs, but that's OK because the whitespace will be ignored by the `-eq` test operator when parsing its integer operands).

We're then just checking that all of those lines match the pattern `", "")` by confirming that we get the same number of lines as we found with `wc -l`, which looks for newlines just like `grep`, I believe.

I think what you're noticing in your example is that if you use single quotes to define the `$credcalls` variable, then `\n` is treated as two literal characters, so `printf` will just pass those through:

```
$ credcalls='foo", "")\nbar", "")\nbaz", "")\n'

$ printf "%s" "$credcalls"
foo", "")\nbar", "")\nbaz", "")\n
```

However, if we define it with LF characters instead, `printf` will write out those newlines:

```
$ credcalls='foo","")
bar","")
baz","")
'

$ printf "%s" "$credcalls"
foo","")
bar","")
baz","")
```

So this latter situation is akin to what happens in our test, because the `$credcalls` variable contains multiple lines as output by the `grep` command which [finds](https://github.com/git-lfs/git-lfs/blob/a18bd24111bbfb4d814cd26023d35136bfe80fc2/t/t-credentials.sh#L65) all the lines matching the pattern `creds: git credential`, and so when we echo that verbatim using `printf "%s" "$credcalls"` we get the same number of lines.  (In short, that `printf` technique is a handy way to avoid all the historical [complexity](https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo/65819#65819) of `echo -n` and output an exact copy of the variable's contents, including LF characters.)

PR #6152, opened 2025-11-27
Add missing options to manual pages and make JSON options consistent

chrisd8088:
This PR updates our manual pages so all available command options are documented, including the short versions of any options for which those are defined.

We also ensure that when a short version of an option exists, we list it ahead the long version, to be consistent with general manual page formatting practices and also with those of the Git project.

Finally, we update the `git lfs track` command so that it now accepts the `-j` short option version of the `--json` option, as all of our other commands which support JSON-formatted output already do.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
         go: ['1.25.x']
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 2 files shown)


larsxschneider (approved):
🙇

josephjamesmorris700-hub (commented):
Thank you

PR #6170, opened 2025-12-20
Full support of special attributes in `migrate`

Diamondemon:
This pull request completes pull request #6126 by making the `gitattr` package able to access special attributes files. It paves the way for thorough correction of #5173, #5179 and #5181
New method for the Tree class:

- `FindSpecialAttributes` allows the Tree to have access to the system-level, user-level and repository-specific attributes files.

### Detailed behaviour:

- The gitattr.Tree structure uses its reference to the 3 special attributes trees, namely system, user and repository. These references can be filled by `Tree.FindSpecialAttributes` or stay `nil`.
- `FindSpecialAttributes` takes as arguments the config environments that contain information about where to find the system-level and user-level special attributes files, and the path of the git working directory to be able to get the repository-specific special attributes. It then builds their respective trees and keeps them into the respective references.
- These new trees are already taken into account by the `Applied` method.
- The new way to use the `Tree` class to its full potential is the following: `t = New --> t.FindSpecialAttributes --> t.Applied (as many times as necessary)` This order must be respected due to the mecanism in Applied that avoids re-processing macros.

### Other features:

- in order to be able to access special attributes files from the `gitattr` package without creating a dependency loop, the `attribs.go` file (which contains such logic) has been moved from the `git` package to the `git/gitattr` package.
- A new `git.Var` function has been added to be able to access git environment variables necessary for the retrieval of the aforementioned files (thanks @chrisd8088 ).
- In `attribs.go`, file-retrieval logic has been separated from the `AttributePath` parsing logic. `Tree.FindSpecialAttributes` uses the former only.
- Functions to get special attributes files that are used in `command_track.go` have been renamed to better fit their actual effect.

### Work left to do in later PRs:
- Make full use of this new logic in `command_track.go` and any other file that parses attributes, to have a unified process
- Full documentation of the `gitattr` package will be done once main related issues are resolved.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -58,7 +58,7 @@ jobs:
       shell: bash
       env:
         FORCE_LOCALIZE: true
-    - uses: azure/trusted-signing-action@v0.5.11
+    - uses: azure/trusted-signing-action@v1.0.0
       with:
         azure-tenant-id: ${{ secrets.SPN_GIT_LFS_SIGNING_TENANT_ID }}
         azure-client-id: ${{ secrets.SPN_GIT_LFS_SIGNING_CLIENT_ID }}
@@ -73,7 +73,7 @@ jobs:
         timestamp-digest: SHA256
     - run: make release-windows-stage-2
       shell: bash
-    - uses: azure/trusted-signing-action@v0.5.11
+    - uses: azure/trusted-signing-action@v1.0.0
       with:
         azure-tenant-id: ${{ secrets.SPN_GIT_LFS_SIGNING_TENANT_ID }}
         azure-client-id: ${{ secrets.SPN_GIT_LFS_SIGNING_CLIENT_ID }}


chrisd8088:
Hey, thanks so much for continuing to work on this topic, and I'll try to get a review done over the holiday, if at all possible.

As an initial quick response, I see that there are some CI job failures, so we'll obviously need to make sure those are all resolved.

As well, I would just ask that we rewrite the initial commit (the one I contributed to), as I didn't mean for it to be included as-is in a final PR.  (It really was just a quick experiment!)  For one thing, we should probably not just rename `git/attribs.go` to `git/gitattr/attribs.go` but call it something like `git/gitattr/files.go` instead.

As for the commit message, we don't need anything too elaborate, but instead of `[EXPERIMENT] consolidate Git attribute handling` I would suggest something like the following:

```
consolidate Git attribute handling in one package

At present, our "git/gitattr" package defines functions and methods
related to the parsing of Git attributes files, while our "git" package
defines higher-level functions to locate Git attributes files and parse
their contents using the "gitattr" package.

One consequence of this design is that we are unable to utilize certain
functions from the "git" package within the "gitattr" package,
specifically those which detect and parse Git attributes files other
than ".gitattributes" files.

To simplify future changes and support the detection of all Git
attributes files, especially during "git lfs migrate" operations,
we move the functions of the "git" package related to Git attribute
files into the "gitattr" package.

To make this change, we rename our "git/attribs.go" source file to
"git/gitattr/files.go", and revise the GetSystemAttributePaths() function
to call a new Var() function in the "git" package in order to execute the
"git var" command, rather than executing that command directly with the
gitNoLFS() helper function, since that function is not exported outside
the "git" package.
```

Diamondemon:
> As an initial quick response, I see that there are some CI job failures, so we'll obviously need to make sure those are all resolved.

Yes of course. I would have done so sooner if I had realized that doing `make test` from the root directory and from the `t` directory does not run the same tests ^^"
It looks like I inadvertently modified the behaviour of the `git lfs track` command which I thought I had done everything to keep the same way. So yeah, I'm on it.

> As well, I would just ask that we rewrite the initial commit (the one I contributed to), as I didn't mean for it to be included as-is in a final PR. (It really was just a quick experiment!) For one thing, we should probably not just rename git/attribs.go to git/gitattr/attribs.go but call it something like git/gitattr/files.go instead.
>
> As for the commit message, we don't need anything too elaborate, but instead of [EXPERIMENT] consolidate Git attribute handling I would suggest something like the following

Sure thing, I'll change the message. I also agree with the change in file name, the old naming feels like a duplicate in the package now.

Diamondemon:
I fixed it. I did not understand the `defer` keyword the right way so I was using it wrong.

I also changed the first commit's message, changed the `attribs.go` file to `files.go`, and rearranged the number of commits.

Only missing point should be some new shell tests.

Diamondemon:
I looked into the shell tests and I realized that the system `gitattributes` file cannot really be checked in tests. There does not seem to be a really safe way to change it since the path `/etc/gitattributes` cannot be overridden the same way `$HOME` can be changed.
Thus we ignore it, right?

I created a test that at least verifies that both the user-specific attribute file and the repo-wide file are parsed and priorities are applied in the appropriate order.

I did not check for the override of macros since this is done both in the `macro_test.go` and `tree_test.go` files.

The test failed on Windows because of the user-level file I create. I there an equivalent on Windows?

Diamondemon:
I removed the creation of the user-level attributes file (and related tested files) for windows tests and now the build on windows works.

chrisd8088:
> There does not seem to be a really safe way to change it since the path `/etc/gitattributes` cannot be overridden the same way `$HOME` can be changed.
> Thus we ignore it, right?

Yes and no.  It's not something we can change solely in a shell test, because the path to the system attributes file is [wired](https://github.com/git/git/blob/b5c409c40f1595e3e590760c6f14a16b6683e22c/attr.c#L878) into Git at compile time.  However, it's not impossible that we change that in our CI jobs, because we compile Git in several of them, so we could define the `ETC_GITATTRIBUTES` [variable](https://github.com/git/git/blob/b5c409c40f1595e3e590760c6f14a16b6683e22c/configure.ac#L371-L377) when doing so.  But we'd require some coordination between our CI workflows and our shell tests, so that the tests only tried to alter the system attributes file if they ran in a workflow context where the system attributes file path had been altered, and the tests would also need to ensure they reset the contents of the file even if the test failed.  The former condition is one the tests could check by running `git var GIT_ATTR_SYSTEM`, but as that only works with Git v2.42.0 and above, we'd likely only want to set the `ETC_GITATTRIBUTES` variable when [compiling](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/.github/workflows/ci.yml#L133-L140) Git in our `build-latest` CI jobs.

On the other hand, the global attributes file can be relocated at runtime by setting the `core.attributesFile`, so we should definitely have some shell tests of our migration commands with Git LFS patterns set in that file.  (At the moment, I think there's only one [example](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/t/t-track.sh#L621) of another shell test which sets this configuration option, but now we can add similar tests for our migration commands.)

chrisd8088:
> The test failed on Windows because of the user-level file I create. I there an equivalent on Windows?

The global Git attributes file is definitely supported on Windows.  (As you know, Git's documentation uses "global" to mean "affects all of a user's repositories", not "affects all users' repositories", since that's what the system attributes file does.)

I looked at some of the CI jobs runs for this PR and I suspect this particular [run](https://github.com/git-lfs/git-lfs/actions/runs/20682472794/attempts/1) may be the one which suggested to you that the tests you'd written in your commit 41ca81e0a2458f99635d626a8d27 were failing on Windows due to a problem with the global attributes file.

However, I don't actually see a failure of the `t-migrate*.sh` scripts in that job; rather, I think we had one of our (unfortunately not uncommon) test flakes on Windows, in this case in the unrelated `t-status.sh` test script.

I re-ran that particular CI job and while there were still [failures](https://github.com/git-lfs/git-lfs/actions/runs/20682472794/attempts/2), they are definitely unrelated to your PR.  The tests in the `t-migrate*.sh` scripts all passed, as did those in the `t-status.sh` script.  The "new" failures are due to a known regression in upstream Git which made it way into Git for Windows 2.53.0.rc0.windows.1, and appear in the `t-fetch-include.sh`, `t-unusual-filenames.sh`, and `t-zero-len-file.sh` scripts.  Once Git 2.53.0-rc1 is tagged and a corresponding Git for Windows tag is made, we should see these failures disappear from our CI suite.

In the meantime, though, I don't think you need to exempt your tests of the global attributes file from running on Windows, so you probably can drop your commit 554c569aef9d485e6cbc6bceeb8459121d5f622a.

_UPDATE:_ The issue with the upstream Git for Windows version has been resolved.  I also have to partially correct my
 statement above, because I realized that re-running your CI job wasn't a valid demonstration because the job picked up the latest version of your PR's branch, rather than using the exact same commit from the previous version.  So I copied your latest version into my own branch and removed the final commit 554c569aef9d485e6cbc6bceeb8459121d5f622a (`fix[tests]: avoid creating user-level attributes file on windows`), leaving commit fb04406fbec0be4a1e9fd6f4ae7aa8ef90db9c81 (`feat[tests]: test migrate fixup with special attribute files`) as tip of the branch, and then ran a fresh [set](https://github.com/chrisd8088/git-lfs/actions/runs/21321310986) of CI jobs on that commit.  These all passed, as I expected, including the Windows job [run](https://github.com/chrisd8088/git-lfs/actions/runs/21321310986/job/61371557084).  I therefore believe we can be confident that we don't need to skip any tests of the global user attributes file on Windows.

---

One detail which may be worth stating with respect to our Windows tests is that we run them in the Git Bash environment, which uses MSYS2 to emulate aspects of a Unix system.  As a consequence, if you construct a path like `$HOME/.git/config/attributes` using the `HOME` environment variable, it will look something like `/tmp/git-lfs_TEMP.lITl9S/t-migrate-fixup.sh-1440/home/.git/config/attributes`, with a leading `/tmp` directory.  But if you ask Git to report the path to the global attributes file with `git var GIT_ATTR_GLOBAL`, it will look like `C:/Users/RUNNER~1/AppData/Local/Temp/git-lfs_TEMP.lITl9S/t-migrate-fixup.sh-1440/home/.config/git/attributes`.  Both of these paths refer to the same filesystem location, but the first will be translated by MSYS2, while the latter is the "canonical" Windows path.

---

If you want to be explicit, you could set the path to the global attributes file by setting the `core.attributesFile` configuration variable, instead of creating a file in `$HOME/.git/config/attributes`.  For example, you could do:

```
global="$(canonical_path "$TRASHDIR/gitattributes-global")"
git config core.attributesFile "$global"

echo "*.bat filter=overridden diff=lfs merge=lfs -text" >"$global"
```

That should function on all systems too, while using an explicitly specified path.  (Note also that it's OK to set global Git configuration values because we reset the global configuration file at the start of each test.)

git/gitattr/files.go
```diff
@@ -47,44 +47,86 @@ func (s *AttributeSource) String() string {
 	return s.Path
 }
 
-// GetRootAttributePaths beahves as GetRootAttributePaths, and loads information
-// only from the global gitattributes file.
-func GetRootAttributePaths(mp *MacroProcessor, cfg Environment) []AttributePath {
-	af, _ := cfg.Get("core.attributesfile")
-	af, err := tools.ExpandConfigPath(af, "git/attributes")
+// GetRepoAttributePaths behaves as GetAttributePaths, and loads information
+// only from the repo attributes file.
```
chrisd8088: I think this comment should read:

```suggestion
// GetUserAttributePaths behaves as GetAttributePaths, and loads information
// only from the user's global attributes file.
```

git/gitattr/files.go
```diff
@@ -47,44 +47,86 @@ func (s *AttributeSource) String() string {
...
+		return nil, nil
 	}
+	defer reader.Close()
+	// The working directory for the user gitattributes file is blank.
+	return AttrPathsFromReader(mp, path, "", reader, true), nil
+}
 
-	if _, err := os.Stat(af); os.IsNotExist(err) {
-		return nil
+func GetUserAttributesFile(gitEnv Environment) (io.ReadCloser, string, error) {
```
chrisd8088: For consistency, I think this should be named `GetUserAttributeFile()` (i.e., `Attribute` not `Attributes`).

git/gitattr/files.go
```diff
@@ -47,44 +47,86 @@ func (s *AttributeSource) String() string {
...
+	}
+	if reader == nil {
+		return nil, nil
+	}
+	defer reader.Close()
+	// The working directory for the system gitattributes file is blank.
+	return AttrPathsFromReader(mp, path, "", reader, true), nil
+}
+
+func GetSystemAttributesFile(osEnv Environment) (io.ReadCloser, string, error) {
```
chrisd8088: For consistency, I think this should be named `GetSystemAttributeFile()` (i.e., `Attribute` not `Attributes`).

git/gitattr/files.go
```diff
@@ -47,44 +47,86 @@ func (s *AttributeSource) String() string {
...
+		return nil, "", nil
 	}
 
-	// The working directory for the root gitattributes file is blank.
-	return attrPathsFromFile(mp, af, "", true)
+	reader, err := os.Open(path)
+	if err != nil {
+		return nil, "", err
+	}
+	return reader, path, nil
```
chrisd8088: As a general rule, we try to avoid situations where one function opens a resource (like a file) but other functions are expected to close it.  Instead, we aim to follow the principles outlined in the Effective Go guide in its [section](https://go.dev/doc/effective_go#defer) on the `defer` statement.

I'd therefore ask that all of these `Get*AttributeFile()` functions only return the file path and an `error`, if any, and not attempt to open the file.

It then becomes the responsibility of the `Get*AttributePaths()` functions to try to open the file at the given path, handle any error which might occur, and also handle closing the file.  That way we can't accidentally introduce resource leaks in the future, say by calling `GetUserAttributeFile()` but forgetting to close the open file descriptor it returns.  As the Effective Go guide says:

> Deferring a call to a function such as `Close` has two advantages. First, it guarantees that you will never forget to close the file, a mistake that's easy to make if you later edit the function to add a new return path.  Second, it means that the close sits near the open, which is much clearer than placing it at the end of the function. 

The `GetUserAttributePaths()` function should then have a series of statements like the following, and the other `Get*AttributePaths()` functions should look similar:

```go
	path, err := GetUserAttributeFile(gitEnv)
	if err != nil {
		return nil, err
	}

	reader, err := os.Open(path)
	if err != nil {
		return nil, err
	}
	defer reader.Close()
```

git/gitattr/tree.go
```diff
@@ -88,6 +88,40 @@ func NewFromReader(mp *MacroProcessor, rdr io.Reader) (*Tree, error) {
...
 }
 
+func (t *Tree) FindSpecialAttributes(gitEnv, osEnv Environment, gitDir string) error {
+	systemReader, _, err := GetSystemAttributesFile(osEnv)
+	if err != nil {
+		return err
+	}
+	if systemReader != nil {
+		t.systemAttributes, err = NewFromReader(t.mp, systemReader)
+	}
```
chrisd8088: I think we may not ever close the `systemReader` (and `userReader` and `repoReader`) file resources, which would be an example of the type of resource leak I described in my other comments.

What I'd recommend here is to use the following idiom, and the same again for both of the other resources:

```suggestion
	systemFile, _, err := GetSystemAttributeFile(osEnv)
	if err != nil {
		return err
	}
	systemReader, err := os.Open(systemFile)
	if err != nil {
		return err
	}
	defer systemReader.Close()
	t.systemAttributes, err = NewFromReader(t.mp, systemReader)
```

commands/command_clone.go
```diff
@@ -53,10 +53,7 @@ func cloneCommand(cmd *cobra.Command, args []string) {
 	clonedir, err := filepath.Abs(args[len(args)-1])
 	if err != nil || !tools.DirExists(clonedir) {
 		// Derive from clone URL instead
-		base := path.Base(args[len(args)-1])
-		if strings.HasSuffix(base, ".git") {
-			base = base[:len(base)-4]
-		}
+		base := strings.TrimSuffix(path.Base(args[len(args)-1]), ".git")
```
chrisd8088: This is a nice simplification, but I think it should go into a separate PR, because it's not logically related to all the other work in this PR.

Out of interest, is there a particular tool or linter you're using which outputs a warning about the existing code?

commands/command_merge_driver.go
```diff
@@ -132,7 +132,7 @@ func mergeProcessInput(gf *lfs.GitFilter, filename string, fileSpecifiers map[st
 		Exit(tr.Tr.Get("could not create callback: %s", err))
 	}
 	defer fp.Close()
-	_, err = gf.Smudge(file, pointer, file.Name(), true, getTransferManifestOperationRemote("download", cfg.Remote()), cb)
+	_, _ = gf.Smudge(file, pointer, file.Name(), true, getTransferManifestOperationRemote("download", cfg.Remote()), cb)
```
chrisd8088: This is a nice simplification, but I think it should go into a separate PR, because it's not logically related to all the other work in this PR.

Out of interest, is there a particular tool or linter you're using which outputs a warning about the existing code?

t/fixtures/migrate.sh
```diff
@@ -240,6 +240,25 @@ setup_single_local_branch_tracked_corrupt() {
       mv .gitattributes gitattrs
 
       add_symlink gitattrs .gitattributes
+    elif [[ $1 == "special" ]]; then
+      echo "*.dat filter=lfs" > .git/info/attributes
+      echo "*.bin filter=lfs diff=lfs merge=lfs -text" >> .git/info/attributes
+      echo "*.dat filter=overriden diff=lfs merge=lfs -text" >> .gitattributes
```
chrisd8088: Minor typo here and below (`overridden` not `overriden`):

```suggestion
      echo "*.dat filter=overridden diff=lfs merge=lfs -text" >> .gitattributes
```

t/fixtures/migrate.sh
```diff
@@ -240,6 +240,25 @@ setup_single_local_branch_tracked_corrupt() {
...
+      mkdir -p $HOME/.config/git
+      echo "*.bat filter=overriden diff=lfs merge=lfs -text" > $HOME/.config/git/attributes
+      echo "*.out filter=lfs" >> $HOME/.config/git/attributes
```
chrisd8088: We typically add double quotation marks around strings which expand shell variables like `HOME`:

```suggestion
      mkdir -p "$HOME/.config/git"
      echo "*.bat filter=overridden diff=lfs merge=lfs -text" > "$HOME/.config/git/attributes"
      echo "*.out filter=lfs" >> "$HOME/.config/git/attributes"
```

The reason for this practice is to handle cases where a variable contains multiple words separated by whitespaces, which is possible for file paths since filenames may contains spaces.  The shell will expand the string and then interpret each unquoted word separately.  For example, if you run the following shell commands, they create a file named `foo bar`:

```
$ FOO="foo bar"
$ echo "test" >"$FOO"
$ ls -1 foo*
foo bar
$ cat foo\ bar
test
```

But in the Bash shell, if you run the following commands you'll get an error because Bash expects a single target of a redirection but instead gets two words from `$FOO`:
```
$ FOO="foo bar"
$ echo "test" >$FOO
$FOO: ambiguous redirect
```

t/t-migrate-fixup.sh
```diff
@@ -47,8 +50,10 @@ begin_test "migrate import (--fixup, special attributes)"
   assert_pointer "refs/heads/main" "a.dat" "$dat_oid" "120"
   assert_local_object "$dat_oid" "120"
 
-  assert_pointer "refs/heads/main" "a.out" "$out_oid" "120"
-  assert_local_object "$out_oid" "120"
+  if [ $IS_WINDOWS -eq 0 ]; then
+    assert_pointer "refs/heads/main" "a.out" "$out_oid" "120"
+    assert_local_object "$out_oid" "120"
+  fi
```
chrisd8088: As I mentioned in my longer [reply](https://github.com/git-lfs/git-lfs/pull/6170#issuecomment-3770628811) to your question, global user-level attributes files can and should be supported on Windows.

I think you just happened to encounter one of our several irritating test flakes on Windows; the particular CI job run saw a failure of the unrelated `t/t-status.sh` script, not this one.  I've subsequently run a successful and complete set of CI jobs on your PR with this final commit's changes removed, so I believe you can just drop this last commit, other than perhaps retaining portions of the initial comment.

t/fixtures/migrate.sh
```diff
@@ -217,6 +217,10 @@ setup_single_local_branch_complex_tracked() {
 #   added to the .gitattributes file.
 #   If "link" is passed as an argument, the .gitattributes file is created
 #   as a symlink to a gitattrs file.
+#   If "special" is passed as an argument, a user-level attribute file is
+#   created for non-windows platforms and a repository-wide attribute file is
+#   created. a.out, a.bin, a.bat, a.dat are added to the commit so they get
+#   modified by fixup.
```
chrisd8088: I believe your tests should work on Windows too, so I would just write:

```
#   If "special" is passed as an argument, user-level (i.e., global) and
#   repository-specific attribute files are created, and corresponding files
#   named "a.out", "a.bin", "a.bat", and "a.dat" are added to the commit.
```

Diamondemon: That's a much better phrasing indeed.

t/fixtures/migrate.sh
Diamondemon: Sure, I forgot about that.

commands/command_merge_driver.go
Diamondemon: I think that's just the official Go extension in Visual Studio Code. That's the only linter running in my workspace.

commands/command_clone.go
Diamondemon: I think that's just the official Go extension in Visual Studio Code. That's the only linter running in my workspace.

git/gitattr/files.go
Diamondemon: I went for `Attributes` because there are multiple attributes in one file, and I did not verify for consistency. I compare it to Get*AttributePaths where it's multiple `AttributePath`s instances that get returned.

git/gitattr/files.go
Diamondemon: Oops, typical copy-paste mistake.

git/gitattr/tree.go
Diamondemon: I realized I did understand the `defer` keyword in the wrong way then changed something in `files.go` and I forgot to edit the `tree.go` file.

git/gitattr/files.go
Diamondemon: > As a general rule, we try to avoid situations where one function opens a resource (like a file) but other functions are expected to close it. Instead, we aim to follow the principles outlined in the Effective Go guide in its [section](https://go.dev/doc/effective_go#defer) on the `defer` statement.

That is why I first tried to do by calling `defer reader.Close()` from my `Get*AttributesFile` function but it closes the reader at the end of the scope which is obviously not what we want. So I thought that anyway since all the method does is call the path finding method and open the file, that would not be that bad to use the `defer reader.Close()` statement one level above.

However I see why one may forget to close the reader, even more if they only want the file path for some reason.

Diamondemon:
I read all your comments so here are the differences in my new push:

- I removed the Windows condition. Somehow I had not read the test result properly and blamed my code.
- I changed the documentation addition of `setup_single_local_branch_tracked_corrupt` to the phrasing you suggested.
- I fixed the typo you mentioned on `overridden`
- I added the quotes around `$HOME` in my tests
- I added reworked the file opening logic as per your suggestion
- I renamed all the methods to `Get*AttributeFilePath` for consistency
- I put my unrelated warning fix out of the PR for now.

> One detail which may be worth stating with respect to our Windows tests is that we run them in the Git Bash environment, which uses MSYS2 to emulate aspects of a Unix system. As a consequence, if you construct a path like $HOME/.git/config/attributes using the HOME environment variable, it will look something like /tmp/git-lfs_TEMP.lITl9S/t-migrate-fixup.sh-1440/home/.git/config/attributes, with a leading /tmp directory. But if you ask Git to report the path to the global attributes file with git var GIT_ATTR_GLOBAL, it will look like C:/Users/RUNNER~1/AppData/Local/Temp/git-lfs_TEMP.lITl9S/t-migrate-fixup.sh-1440/home/.config/git/attributes. Both of these paths refer to the same filesystem location, but the first will be translated by MSYS2, while the latter is the "canonical" Windows path. 
> [...]

Does that mean that one way should be better than the other?

> As for system-level attribute files, let's not complicate this PR further by trying to test those, although as I described in reply to your question, it's not completely out of the question. But we'd have to revise some aspects of our GitHub Actions workflows to make this possible, and while I think that would be a good idea, I also think it expands the scope of this PR beyond what's reasonable.

Yeah, that goes a bit out of scope indeed.

> Thank you so much for this PR, it was a pleasure to read! Each commit has a nicely isolated and logical set of changes, which makes it so much easier to understand how they build on each other.

Thanks, I appreciate that. I'm glad this works well with you.

git/gitattr/files.go
chrisd8088: I agree it would read more idiomatically to say "user attributes file" than "user attribute file", especially since Git refers to them as `gitattributes` files.  I don't have a strong preference, though, so long as similar functions are named similarly.  The current `Get*AttributeFilePath` naming scheme is just fine, thank you!

git/gitattr/files.go
chrisd8088: I agree it would read more idiomatically to say "user attributes file" than "user attribute file", especially since Git refers to them as `gitattributes` files.  I don't have a strong preference, though, so long as similar functions are named similarly.  The current `Get*AttributeFilePath` naming scheme is just fine, thank you!

commands/command_merge_driver.go
chrisd8088: > I think that's just the official Go extension in Visual Studio Code. 

Interesting, thank you!  And thanks for moving this change out of the PR!

chrisd8088:
>> Both of these paths refer to the same filesystem location, but the first will be translated by MSYS2, while the latter is the "canonical" Windows path.

> Does that mean that one way should be better than the other?

No, not particularly.  We generally aim to use the Unix-style paths as much as we can in our shell tests, since these allow our tests to be agnostic about the operating system (Linux, macOS, or Windows).

However, there are times when the abstraction the MSYS2 environment provides doesn't quite suffice, for one reason or another.  For instance, when we test the output of the `git lfs env` command, we [expect](https://github.com/git-lfs/git-lfs/blob/9afaf60b9c47e96f031f86d04089ef2e0b79a1fa/t/t-env.sh#L78-L83) to see Windows-style paths in the output because the command [resolves](https://github.com/git-lfs/git-lfs/blob/9afaf60b9c47e96f031f86d04089ef2e0b79a1fa/config/config.go#L398-L400) symbolic links, which
[canonicalizes](https://github.com/git-lfs/git-lfs/blob/9afaf60b9c47e96f031f86d04089ef2e0b79a1fa/tools/filetools.go#L63-L65) file paths [using](https://github.com/git-lfs/git-lfs/blob/9afaf60b9c47e96f031f86d04089ef2e0b79a1fa/tools/filetools_windows.go#L34) a Windows API call that of course only returns Windows-style paths.

PR #6171, opened 2025-12-20
attempt to fix merge ff with multiple remotes and multiple lfs commits

fwag:
This code is meant to fix the following issue reproducer 

## Clone the same repo just from  two different git instances
```
git clone gitea@src.oss.org:joe/nodejs22 oss-nodejs
git clone gitea@src.iss.org:joe/nodejs22 iss-nodejs
```

## Add new LFS objects to OSS
```
cd oss-nodejs
git checkout main

dd if=/dev/random of=foo.xz count=100
git add foo.xz
git commit -a -m 'test1'

dd if=/dev/random of=foo.xz count=100
git add foo.xz
git commit -a -m 'test2'
```

## Now we have two new commits with 2 LFS objects to push
```
git push origin main
```

## Move to the repo hosted on another instance with a branch in common
```
cd ../iss-nodejs
git checkout main
git remote add oss gitea@src.oss.org
git fetch oss

git merge --ff-only oss/main
Updating f442893..67b6897
Downloading foo.xz (51 KB)
Error downloading object: foo.xz (bbdd1b0): Smudge error: Error downloading foo.xz (bbdd1b0b9131658ad7820176879a65d9a24fde1f2a1d8fffd0b79aa663bd4076): could not get connection for batch request: pure SSH connection unavailable (#0)

Errors logged to '/tmp/node/iss-nodejs22/.git/lfs/logs/20251219T123627.95839924.log'.
fatal: foo.xz: smudge filter lfs failed
```

lfs.remote.autodetect fixes the issue, why is the default setting not true ?        
```                   
git config lfs.remote.autodetect true                                      
Updating f442893..67b6897
Fast-forward
 foo.xz | 3 +++
 create mode 100644 foo.xz
```

## OK, so things look like they merged… Try to push
```
> git push origin main                                               
  $ git config lfs.https://src.iss.org/joe/nodejs.git/info/lfs.locksverify true
Git LFS upload failed:   0% (0/1), 0 B | 0 B/s      
Uploading LFS objects:   0% (0/2), 0 B | 0 B/s, done.
  (missing) foo.xz (02b3b4dd03aff81ca43fdd0941ace2ef36de6cd5786559e079cefcab7bcd9fe0)
hint: Your push was rejected due to missing or corrupt local objects.
hint: You can disable this check with: `git config lfs.allowincompletepush true`
```

With a fast forward merge, only the HEAD LFS object is fetched, leaving the LFS objects from intermediate commits unfetched.

### DIFF
--- a/commands/command_post_merge.go
+++ b/commands/command_post_merge.go
@@ -3,6 +3,8 @@ package commands
 import (
 	"os"
 
+	"github.com/git-lfs/git-lfs/v3/git"
+	"github.com/git-lfs/git-lfs/v3/lfs"
 	"github.com/git-lfs/git-lfs/v3/tr"
 	"github.com/rubyist/tracerx"
 	"github.com/spf13/cobra"
@@ -17,6 +19,8 @@ func postMergeCommand(cmd *cobra.Command, args []string) {
 		Print(tr.Tr.Get("This should be run through Git's post-merge hook.  Run `git lfs update` to install it."))
 		os.Exit(1)
 	}
+
+	fetchMissingLfsObjects("ORIG_HEAD", "HEAD")
 
 	// Skip entire hook if lockable read only feature is disabled
 	if !cfg.SetLockableFilesReadOnly() {
@@ -43,6 +47,41 @@ func postMergeCommand(cmd *cobra.Command, args []string) {
 	err := lockClient.FixAllLockableFileWriteFlags()
 	if err != nil {
 		LoggedError(err, tr.Tr.Get("Warning: post-merge locked file check failed: %v", err))
+	}
+}
+
+func fetchMissingLfsObjects(pre, post string) {
+	remote := cfg.Remote()
+	if r := git.FirstRemoteForTreeish(post); r != "" {
+		remote = r
+	}
+
+	q := newDownloadQueue(
+		getTransferManifestOperationRemote("download", remote),
+		remote,
+	)
+
+	gitscanner := lfs.NewGitScanner(cfg, func(p *lfs.WrappedPointer, err error) {
+		if err != nil {
+			LoggedError(err, tr.Tr.Get("Scanner error: %s", err))
+			return
+		}
+		lfs.LinkOrCopyFromReference(cfg, p.Oid, p.Size)
+		if cfg.LFSObjectExists(p.Oid, p.Size) {
+			return
+		}
+
+		tracerx.Printf("post-merge: found LFS object %s (%s)", p.Oid, p.Name)
+		q.Add(downloadTransfer(p))
+	})
+
+	if err := gitscanner.ScanRefRange(post, pre, nil); err != nil {
+		LoggedError(err, tr.Tr.Get("Scanner error: %s", err))
+	}
+
+	q.Wait()
+	for _, err := range q.Errors() {
+		LoggedError(err, tr.Tr.Get("Download error: %s", err))
 	}
 }
 
--- a/config/config.go
+++ b/config/config.go
@@ -225,7 +225,7 @@ func (c *Configuration) IsDefaultRemote() bool {
 }
 
 func (c *Configuration) AutoDetectRemoteEnabled() bool {
-	return c.Git.Bool("lfs.remote.autodetect", false)
+	return c.Git.Bool("lfs.remote.autodetect", true)
 }
 
 func (c *Configuration) SearchAllRemotesEnabled() bool {


chrisd8088:
Hey, thanks for this proposed change, and I'm sorry you're having trouble.  I'll try to take a closer look over the holidays, if at all possible.

> `lfs.remote.autodetect` fixes the issue, why is the default setting not true?

This option enables the automatic selection of a Git remote from which to derive a Git LFS remote URL based on the output of a command like `git branch -r --contains HEAD`, and the logic to do so was only introduced in v3.3.0 with PR #5066.  As [mentioned](https://github.com/git-lfs/git-lfs/pull/5066#pullrequestreview-1036397080) in the [comments](https://github.com/git-lfs/git-lfs/pull/5066/files#r919361861) on the PR, this logic deviates from the legacy behaviour of the Git LFS client and causes certain existing configuration options like `remote.lfsDefault` to be ignored if a remote can be selected automatically from those which contain the current commit.

As a consequence, to maintain backwards compatibility, a new configuration option had to be introduced and the additional logic made conditional on that option being explicitly set, rather than being the default.

When and if we release another major version such as v4.0.0, we would then have an opportunity to revisit the existing defaults for our configuration options and remote selection logic and break with the previously established behaviour of the Git LFS client, although we'd still want to try to minimize the consequences for our users.  For instance, we might only try to determine the remote automatically from the current commit if no `remote.lfsDefault` option was defined, rather than potentially ignoring that option as we do now when `lfs.remote.autoDetect` is set to enabled.

All of which is to say that we are unfortunately not able to make `lfs.remote.autoDetect` enabled by default, at least not until a future major release of the Git LFS client.

commands/command_post_merge.go
```diff
@@ -18,6 +20,8 @@ func postMergeCommand(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
+	fetchMissingLfsObjects("ORIG_HEAD", "HEAD")
```
chrisd8088, line 23: If we want to introduce this type of behaviour to our post-merge hook program, I think we should make it optional, and also not the default.  Perhaps we could introduce an `lfs.fetchPostMerge` configuration option for this purpose, whose default value is `false`.

My rationale for this suggestion is based on two concerns.  First, I suspect that users may be surprised if a `git merge` operation, which typically requires only local access, now requires connectivity to a remote in order to retrieve Git LFS objects, and possibly even new credentials.

Second, while some users may wish to retrieve all the Git LFS objects referenced in intermediate commits in the histories of the merged branches (for instance, if they wish to then push the merged result to another remote, as in your case), other users may want to keep as few Git LFS objects as possible in local storage.  These users typically deal with large repositories where the Git LFS client's duplicative local storage is a burden, and so they may not want to automatically fetch a full history's worth of objects when they perform a merge.

commands/command_post_merge.go
```diff
@@ -46,6 +50,41 @@ func postMergeCommand(cmd *cobra.Command, args []string) {
 	}
 }
 
+func fetchMissingLfsObjects(pre, post string) {
+	remote := cfg.Remote()
+	if r := git.FirstRemoteForTreeish(post); r != "" {
+		remote = r
+	}
```
chrisd8088, line 57: Our only other [use](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/commands/command_filter_process.go#L100) of the `FirstRemoteForTreeish()` function is [governed](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/commands/command_filter_process.go#L98) by the `lfs.remote.autoDetect` configuration option, and so I think this one should be as well.

The `FirstRemoteForTreeish()` function will [run](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/git/git.go#L1752) a `git branch -r --contains HEAD` [command](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/git/git.go#L1695-L1702) and then [select](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/git/git.go#L1742-L1747) the first reference from that command's output to select a remote.  While this works, it's not the established behaviour of the Git LFS client, and as I've noted already, we shouldn't change that behaviour (at least not for the present).

All of which is to say that I think if we are going to fetch Git LFS objects in situations where we did not previously do so, we should select the remote(s) from which to fetch following the same logic we do now when "smudging".

For one thing, as was [pointed out](https://github.com/git-lfs/git-lfs/pull/5066/files#r919366243) during the review of PR #5066 when the `lfs.remote.autoDetect` and `lfs.remote.searchAll` options were introduced, if a user has multiple remotes and we choose one other than what we've selected previously, the user may be prompted for credentials.  We should therefore make sure they've opted into this behaviour by enabling the relevant options and that we don't otherwise try to use remotes different than the one we would normally select.

config/config.go
```diff
@@ -225,7 +225,7 @@ func (c *Configuration) IsDefaultRemote() bool {
 }
 
 func (c *Configuration) AutoDetectRemoteEnabled() bool {
-	return c.Git.Bool("lfs.remote.autodetect", false)
+	return c.Git.Bool("lfs.remote.autodetect", true)
```
chrisd8088, line 228: Alas, as I noted in my initial comments on this PR, we can't change this option's default setting at this time.

In general, we try to avoid altering the defaults for our configuration options unless there's a critical bug which we need to correct.  Barring that type of problem, if we do change the default settings for our options, that change should be part of a new major version of the Git LFS client and should be accompanied by a suitably lengthy period of advance notice to our users.

chrisd8088 (commented):
Hey, thanks again for this PR, and for clearly documenting the issue you encountered!  It's taken me a while to think through some of the implications of the changes you've proposed; my apologies for the delay.

In the conditions you describe, I believe our existing tools at least provide a manual step users can take to fetch the necessary Git LFS objects, which is to run a `git lfs fetch --all <remote> <remote>/<ref>` command.  (I'll put a longer example in a separate comment with some other notes.)

However, I acknowledge that this is less than ideal as it does require a separate step on the part of users.

As I've noted already, we can't change the default setting of our `lfs.remote.autoDetect` or `lfs.remote.searchAll` configuration options, at least not at this time.

As for the changes to our post-merge hook program, though, those could be valuable for some users.  However, I think they (a) need to respect the existing remote selection logic, and (b) need to run only when a new `lfs.fetchPostMerge` configuration option is also enabled.

In other words, if a user has `lfs.fetchPostMerge` enabled, our post-merge hook program would try to fetch any Git LFS objects found between `ORIG_HEAD` and `HEAD`, using our existing logic to select a remote.  That means respecting `lfs.remote.autoDetect` if it's enabled, and otherwise using the remote selected by `remote.lfsDefault` and ` lfs.url`, etc., but also respecting the `lfs.remote.searchAll` fallback mechanism if it's enabled.

It's a bit unfortunate that Git doesn't (yet) provide a post-fetch hook, which would be great place for us to register a program that would automatically fetch any Git LFS objects corresponding to the Git objects that were fetched.  (If you feel like proposing such a hook on the Git mailing list, please do so!)

We'll also need this PR to include an appropriate set of new tests to verify the expanded behaviour of the post-merge hook, likely in our `t/t-post-merge.sh` test [script](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/t/t-post-merge.sh).

As well, we'll want this PR to document the new operations of the post-merge hook in our `git-lfs-post-merge(1)` manual [page](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/docs/man/git-lfs-post-merge.adoc), and we'll need an additional entry in our `git-lfs-config(5)` manual [page](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/docs/man/git-lfs-config.adoc#fetch-settings) to document the new `lfs.fetchPostMerge` option.

Thank you again for taking an interest in trying to improve the experience of our users, and for writing up this proposal!

chrisd8088:
For my own future reference, here's a reproduction case of the issue described in this PR, using only local repositories:

```
$ git init src1
$ cd src1
$ git lfs track "*.bin"
$ git add .gitattributes
$ git commit -m attrs
$ git config receive.denyCurrentBranch ignore

$ cd ..
$ git clone src1 src2
$ cd src2
$ echo foo >foo.bin
$ git add foo.bin
$ git commit -m foo
$ echo FOO >foo.bin
$ git add foo.bin
$ git commit -m FOO

$ cd ..
$ git clone src1 dst
$ cd dst
$ git remote add src2 ../src2
$ git fetch src2
$ git merge src2/main
Error downloading object: foo.bin (1706b5c): Smudge error: Error downloading foo.bin (1706b5c18e4358041b463995efc30f8f721766fab0e018d50d85978b46df013c): error transferring "1706b5c18e4358041b463995efc30f8f721766fab0e018d50d85978b46df013c": [0] remote missing object 1706b5c18e4358041b463995efc30f8f721766fab0e018d50d85978b46df013c

fatal: foo.bin: smudge filter lfs failed
```

We can resolve the immediate issue by using `git lfs fetch` to retrieve the Git LFS objects associated with the final commit in the `src2/main` branch, but this still results in a problem when we try to push back to the `origin` remote:

```
$ git lfs fetch src2 src2/main
$ git merge src2/main
Updating d99d959..807a418
Fast-forward

$ git push origin main
Git LFS upload failed:  50% (1/2), 0 B | 0 B/s                                  
  (missing) foo.bin (b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c)
hint: Your push was rejected due to missing or corrupt local objects.
Uploading LFS objects:  50% (1/2), 0 B | 0 B/s, done.
```

To fix that issue, we need to run `git lfs fetch` with the `--all` option.  For example, we could have done this instead of the commands above:

```
$ git lfs fetch --all src2 src2/main
$ git merge src2/main
Updating d99d959..807a418
Fast-forward

$ git push origin main
Uploading LFS objects: 100% (2/2), 0 B | 0 B/s, done.
```

commands/command_post_merge.go
fwag: What if we move it to pre push hook ?

chrisd8088: > What if we move it to pre push hook ?

Interesting question!  I think my comments would largely all still apply.  In part that's because it might be surprising to a user if they push to remote A and are asked for credentials for remote B in order to download Git LFS objects, especially if remote A is just another copy of the repository on their local disk.

I skimmed through the Git mailing list archives for discussions of a post-fetch hook, and the topic has come up a few times.  One relevant early [message](https://lore.kernel.org/git/7vbq7ibxhh.fsf@gitster.siamese.dyndns.org/) from the Git maintainer outlines conditions for a new hook, one of which would seem to broadly apply to the Git LFS use case:

> There are five valid reasons you might want a hook to a git operation:
> ...
> (2) A hook that operates on data generated after the command starts to run.  The ability to munge the commit log message by the commit-msg hook is an example.

In another [message](https://lore.kernel.org/git/7v4nwpbaxq.fsf@alter.siamese.dyndns.org/) from a thread about a potential "tweak-fetch" hook the Git maintainer points out that a typical `git pull` operation invokes `git fetch` and then `git merge`.  So if we have to work with just the existing hooks provided by Git, I think the post-merge hook is (somewhat) more appropriate than the pre-push hook.

In practice, of course, we primarily rely on the "smudge" operation to download the Git LFS objects the user requires in order to perform the final updates of the working tree, both when `git pull` is run but also when the user runs `git checkout` or `git merge`.

The challenge with this, as you've pointed out, is that we expect Git LFS objects to be available from the default Git LFS remote the user has configured, and we only "see" the objects in the current checkout, not any intermediate objects in the Git history (with the caveat that not all Git LFS users want those intermediate objects to be automatically retrieved).

A post-fetch hook, if one existed, would presumably inform the hook program about the remote from which Git objects were fetched, as well as the IDs of those objects.  Hypothetically, a Git LFS post-fetch hook might then be able to download any corresponding Git LFS objects *before* any subsequent checkout stage, so our "smudge" filter would find that all the Git LFS objects necessary for the working tree checkout already exist locally and thus wouldn't need to download anything at all.

We'd want such a hook program to respect all the `lfs.fetch*` configuration options, like `lfs.fetchExclude`.

I'm slightly uncertain, though, as to how our existing remote server discovery [process](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/docs/api/server-discovery.md) could or should apply if the remote provided to the post-fetch hook differs from, say, an explicitly configured `lfs.url` setting.  I think we'd need to work through the full range of potential conditions here and make sure we weren't doing anything too surprising for users.

But that's all hypothetical, of course, since Git doesn't have a post-fetch hook right now.

PR #6172, opened 2025-12-21
commands: show root help for -h

### LINKED ISSUE #6168: git lfs -h not working
**Describe the bug**

﻿git lfs -h
Sorry, no usage text found for "-h"

**To Reproduce**

```
> git lfs -h
Sorry, no usage text found for "-h"
```

**Expected behavior**

It should show the help as seen in `git lfs`.

**System environment**
The version of your operating system, plus any relevant information about platform or configuration (e.g., container or CI usage, Cygwin, WSL, or non-Basic authentication).  If relevant, include the output of `git config -l` as a code block.

git-lfs/3.7.1 (GitHub; darwin arm64; go 1.25.3)

osasisorae:
## Summary
- treat `-h` the same as `--help` so `git lfs -h` prints the root man page instead of “Sorry, no usage text…”
- add a regression test around `printHelp("-h")` to ensure the short flag keeps working

## Testing
- go test ./commands -run TestPrintHelp

Fixes #6168


chrisd8088:
Thanks again very much for this PR!

I've added the `t/t-usage.sh` test script I proposed above to my own PR #6188, and so once that's merged, I think we can drop the `commands/run_test.go` file from this PR and just add two lines to the new shell test, specifically these two lines:

```
git lfs -h | grep 'git lfs <command> \[<args>\]'
```

```
git-lfs -h | grep 'git lfs <command> \[<args>\]'
```

osasisorae:
Alright. Lemme know how it goes.

chrisd8088:
> Alright. Lemme know how it goes.

Thanks, @osasisorae!  I expect the issue in upstream Git which is causing parts of our CI suite to fail at the moment will be resolved next week, at which point I can merge my PR #6188 and then we can adjust this PR to make use of the new `t/t-usage.sh` test script.

If you'd prefer me to make those adjustments, I'd be happy to do so; you'd just need to grant me permission to push to the branch in your repository.

chrisd8088:
Thanks!  I was able to push one extra commit to this PR which just makes use of the new shell test we added in PR #6188 and drops the additional Go test file, since the shell test should be sufficient and effective.

I'll let the CI suite run even though I know the Windows job will fail due to a regression in upstream Git which made its way into the latest Git for Windows pre-release candidate.  We'll unfortunately have to wait for the next Git for Windows candidate to be tagged before that job will pass again, but once that's the case (and I've dealt with what are likely some test flakes as well), I can merge this PR.

Thanks again for all your help, and welcome to the Git LFS project!

PR #6188, opened 2026-01-13
Improve formatting of manual page synopses

chrisd8088:
This PR updates the synopsis sections of our manual pages to resolve several errors and to introduce additional formatting to make our documentation more clear and accessible.

Specifically, we now format our manual pages' synopsis sections using AsciiDoc source listing blocks with custom settings that permit us to demarcate all literal text in boldface and all replaceable arguments in italics.  This additional formatting will be visible whenever our manual pages are rendered in HTML or displayed by the `man(1)` program from roff sources.

The use of boldface for literal text and italics for placeholder arguments is common to the manual page conventions of many systems, including the [style guide](https://man7.org/linux/man-pages/man7/man-pages.7.html#STYLE_GUIDE) of the Linux `man-pages(7)` manual and the BSD manual page [guide](https://manpages.bsd.lv/mdoc.html).

#### Line Breaks

One major advantage to the use of source listing blocks is that in our command synopses with multiple lines, we no longer need to remember to include trailing `+` characters to force breaks between the lines.

For example, in PR #6088 we updated the `git-lfs-fetch(1)` manual page and added a second line to the synopsis, but accidentally did not add a trailing `+` character to the first line, so at present the GitHub UI renders both lines of the page's [synopsis](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/docs/man/git-lfs-fetch.adoc#synopsis) consecutively on a single line:

&nbsp;&nbsp;&nbsp;&nbsp;Screenshot 2026-01-12 at 9 53 22 PM

By using source listing blocks for our synopses, we ensure that problems of this type can not occur again.

#### Rendered Examples

The [synopsis](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/docs/man/git-lfs-checkout.adoc#synopsis) of our `git-lfs-checkout(1)` command's manual page renders in the GitHub UI as shown below, first without the changes from this PR and then with them.  Notice that at present two separate fonts are used, one monospace and the other a proportional-width font, and only the text in the monospace font has a grey background:

- Without this PR's changes:
    Screenshot 2026-01-12 at 8 46 55 PM
- With this PR's changes:
    Screenshot 2026-01-12 at 8 47 46 PM

The revisions made in this PR ensure that the entire section appears in monospace with a common grey background, as well as distinguishing the literal text and placeholders both from each other and from the syntactic characters such as `[`, `{`, and `|`.

The same manual page's synopsis is rendered by the `man git-lfs-checkout` command on a macOS system as shown below, first without the changes from this PR and then with them:

- Without this PR's changes:
    Screenshot 2026-01-12 at 7 43 39 PM
- With this PR's changes:
    Screenshot 2026-01-12 at 7 44 06 PM

Note that the indentation remains the same, while boldface and underlined text demarcates the literals and placeholders, respectively.

#### Asciidoctor Extension

One inconvenient consequence of the use of AsciiDoc source listing blocks for our manual page synopses is that the Asciidoctor `manpage` [converter](https://docs.asciidoctor.org/asciidoctor/latest/manpage-backend/#backend-and-converter) we use to generate copies of our manual pages in the roff format expected by the `man(1)` program always [indents](https://github.com/asciidoctor/asciidoctor/blob/fc0d033577d30adbffba73ce06709292fc2cf3ce/lib/asciidoctor/converter/manpage.rb#L253-L260) these blocks.  This results in an extra level of indentation, as shown below using the `man git-lfs-env` command as an example (the red arrow highlights the excess indentation):

&nbsp;&nbsp;&nbsp;&nbsp;Screenshot 2026-01-12 at 11 00 45 PM

Unfortunately, it is not possible at present to change this aspect of the `manpage` converter's behaviour via a configuration option.  Instead, we add a custom `role=synopsis` attribute to the source blocks in our synopsis sections, and introduce a small Asciidoctor extension program which removes the extra indentation from the roff formatting generated by the converter, but only when the block has the `role=synopsis` attribute defined.  The result is that our synopsis sections are aligned with all the other sections of our manual pages, as shown below:

&nbsp;&nbsp;&nbsp;&nbsp;Screenshot 2026-01-12 at 11 22 15 PM

#### Plain Text Output

When a user runs the `git lfs help` command, at present we do not pass through the output of a `man(1)` command in the same manner as the `git help` command, but instead output a plain-text version of the appropriate Git LFS manual page from an internal map compiled into the `git-lfs` binary itself.  This map is [generated](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/Makefile#L230-L239) during our build process by our standalone `mangen` [program](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/docs/man/mangen.go), which attempts to remove all AsciiDoc formatting from our manual page source files while still retaining their basic structure.

Because our AsciiDoc manual page source files now include inline formatting markup within their synopsis sections, we  revise the `mangen` program to remove this markup, specifically the `*` and `_` characters.

Fortunately there are no instances of the `*` character in any of our synopses, and the `_` character only appears in variable argument names in two of our manual pages, the `git-lfs-post-checkout(1)` [page](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/docs/man/git-lfs-post-checkout.adoc?plain=1#L9) and and the `git-lfs-post-merge(1)` [page](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/docs/man/git-lfs-post-merge.adoc?plain=1#L9).

Since there is no semantic difference between the use of underscore characters to separate words in argument names and the use of hyphen (`-`) characters, and the latter are both more consistent with our [other](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/docs/man/git-lfs-checkout.adoc?plain=1#L9-L10) [pages](https://github.com/git-lfs/git-lfs/blob/1f2f7c697b5568b4bf1ed76e7ed4115dd1f0b3ba/docs/man/git-lfs-migrate.adoc?plain=1#L311) and are also [recommended](https://github.com/git/git/blob/c4a0c8845e2426375ad257b6c221a3a7d92ecfda/Documentation/CodingGuidelines#L871-L878) by the Git documentation style guide, we simply replace the underscores with hyphens in the few instances where they appear in argument names.

We also introduce a new `t/t-usage.sh` shell test script with a single initial test to verify that our `mangen` program correctly removes the AsciiDoc formatting from the synopsis sections of our manual pages.  (We expect to later enhance this test when we merge the changes from PR #6172.)

#### Unrelated CI Failures

Note that our `Build with latest Git` CI jobs are currently failing due to an unrelated issue caused by a recent change to Git's `master` branch, specifically that from commit git/git@3e2836a742d8b2b2da25ca06e9d0ac3a539bd966.  I've advised the Git mailing list of this issue and will await feedback from the upstream Git developers on whether we should or should not revise our tests to accommodate this change.

### DIFF
--- a/docs/man/git-lfs-pointer.adoc
+++ b/docs/man/git-lfs-pointer.adoc
@@ -6,12 +6,12 @@ git-lfs-pointer - Build, compare, and check pointers
 
 == SYNOPSIS
 
-[source,role=synopsis]
+[source,role=synopsis,subs="verbatim,quotes"]
 ----
-git lfs pointer --file=path/to/file
-git lfs pointer --file=path/to/file --pointer=path/to/pointer
-git lfs pointer --file=path/to/file --stdin
-git lfs pointer --check --file=path/to/file
+*git lfs pointer* **--file=**__path/to/file__
+*git lfs pointer* **--file=**__path/to/file__ **--pointer=**__path/to/pointer__
+*git lfs pointer* **--file=**__path/to/file__ *--stdin*
+*git lfs pointer --check* **--file=**__path/to/file__
 ----
 
 == Description
--- a/docs/man/git-lfs-checkout.adoc
+++ b/docs/man/git-lfs-checkout.adoc
@@ -6,10 +6,10 @@ git-lfs-checkout - Update working copy with file content if available
 
 == SYNOPSIS
 
-[source,role=synopsis]
+[source,role=synopsis,subs="verbatim,quotes"]
 ----
-git lfs checkout [<glob-pattern>...]
-git lfs checkout --to <file> {--base|--ours|--theirs} <conflict-obj-path>
+*git lfs checkout* [_<glob-pattern>_...]
+*git lfs checkout --to* _<file>_ {*--base*|*--ours*|*--theirs*} _<conflict-obj-path>_
 ----
 
 == DESCRIPTION
--- a/docs/man/git-lfs-clean.adoc
+++ b/docs/man/git-lfs-clean.adoc
@@ -6,9 +6,9 @@ git-lfs-clean - Git clean filter that converts large files to pointers
 
 == SYNOPSIS
 
-[source,role=synopsis]
+[source,role=synopsis,subs="verbatim,quotes"]
 ----
-git lfs clean <path>
+*git lfs clean* _<path>_
 ----
 
 == DESCRIPTION
--- a/docs/man/git-lfs-clone.adoc
+++ b/docs/man/git-lfs-clone.adoc
@@ -6,9 +6,9 @@ git-lfs-clone - Efficiently clone a LFS-enabled repository
 
 == SYNOPSIS
 
-[source,role=synopsis]
+[source,role=synopsis,subs="verbatim,quotes"]
 ----
-git lfs clone [git clone options] <repository> [<directory>]
+*git lfs clone* [_git clone options_] _<repository>_ [_<directory>_]
 ----
 
 == DESCRIPTION
--- a/docs/man/git-lfs-completion.adoc
+++ b/docs/man/git-lfs-completion.adoc
@@ -6,11 +6,11 @@ git-lfs-completion - Shell tab-completion script generation for Git LFS
 
 == SYNOPSIS
 
-[source,role=synopsis]
+[source,role=synopsis,subs="verbatim,quotes"]
 ----
-git lfs completion bash
-git lfs completion fish
-git lfs completion zsh
+*git lfs completion bash*
+*git lfs completion fish*
+*git lfs completion zsh*
 ----
 
 == DESCRIPTION
--- a/docs/man/git-lfs-dedup.adoc
+++ b/docs/man/git-lfs-dedup.adoc
@@ -6,9 +6,9 @@ git-lfs-dedup - Deduplicate Git LFS files
 
 == SYNOPSIS
 
-[source,role=synopsis]
+[source,role=synopsis,subs="verbatim,quotes"]
 ----
-git lfs dedup
+*git lfs dedup*
 ----
 
 == DESCRIPTION
--- a/docs/man/git-lfs-env.adoc
+++ b/docs/man/git-lfs-env.adoc
@@ -6,9 +6,9 @@ git-lfs-env - Display the Git LFS environment
 
 == SYNOPSIS
 
-[source,role=synopsis]
+[source,role=synopsis,subs="verbatim,quotes"]
 ----
-git lfs env
+*git lfs env*
 ----
 
 == DESCRIPTION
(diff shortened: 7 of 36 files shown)


docs/man/git-lfs-pointer.adoc
```diff
@@ -6,12 +6,12 @@ git-lfs-pointer - Build, compare, and check pointers
...
+[source,role=synopsis,subs="verbatim,quotes"]
 ----
-git lfs pointer --file=path/to/file
-git lfs pointer --file=path/to/file --pointer=path/to/pointer
-git lfs pointer --file=path/to/file --stdin
-git lfs pointer --check --file=path/to/file
+*git lfs pointer* **--file=**__path/to/file__
+*git lfs pointer* **--file=**__path/to/file__ **--pointer=**__path/to/pointer__
+*git lfs pointer* **--file=**__path/to/file__ *--stdin*
+*git lfs pointer --check* **--file=**__path/to/file__
```
larsxschneider, line 14: `**foo**` prints text bold too, right? Why is `*` not sufficient here?

larsxschneider (approved):
Thanks for cleaning this up! I’m always amazed by how much thought you put into every detail 🙇

docs/man/git-lfs-pointer.adoc
chrisd8088: Yes, `**` can also be used to demarcate bold text, and `__` can be used for italics.  These are used in [unconstrained](https://docs.asciidoctor.org/asciidoc/latest/text/#unconstrained) pairs, while the single-character form is used in [constrained](https://docs.asciidoctor.org/asciidoc/latest/text/#constrained) pairs.

The key difference is that constrained pairs require a word boundary to be parsed correctly, so we could write `*--file* _<file>_` and that would be rendered as we expect, but if we write `*--file=*_<file>_`, that doesn't work because the `=` character (and lack of any space character) effectively turns the whole phrase into a single "word".

Now this is all actually just a temporary state of affairs, I very much hope!  I intend to write up another PR to make our various manual pages use consistent terms and syntax for things like placeholder arguments.  Right now, we have a mix of `<file>`, `<path>`, `<directory>`, and in this case, `path/to/file`, which is very non-standard; we also mix `<options>` and plain `options`, etc.

We should also remove the `=` characters from this page's synopsis as we don't use them in any other manual pages, at which point we can just use constrained markup pairs instead of the `**` and `__` unconstrained pairs.

I thought about trying to include those changes in this PR too, but decided it was better to tackle one type of change at a time, so this PR just converts our manual pages to a new synopsis format, and the next PR can adjust the actual text of those synopses and the other related parts of the pages, like options lists.

Issue #6187, opened 2026-01-08
git-lfs interacts with case sensitivity across platforms poorly


doctorpangloss:
**Describe the bug**

Git LFS pattern matching in `.gitattributes` is case-sensitive on Linux, but files may be pushed from case-insensitive systems (Windows, macOS) with varying case extensions. When checking out on Linux, files with uppercase extensions (e.g., `.TGA`, `.PNG`, `.BMP`) are not automatically fetched/smudged because the `.gitattributes` patterns use lowercase (e.g., `*.tga`).

**To Reproduce**

1. On Windows or macOS, create and push files with uppercase extensions (e.g., `texture.TGA`)
2. Ensure `.gitattributes` has lowercase patterns: `*.tga filter=lfs diff=lfs merge=lfs -text`
3. Clone the repository on Linux
4. Run `git lfs ls-files` - files with uppercase extensions show `*` (pointer, not downloaded)
5. Run `git lfs pull` - uppercase extension files remain as pointers
6. Unity (or other tools) fail to read these files: `Could not create asset from .../T_Facade_02_N.TGA: File could not be read`

**Expected behavior**

Either:
- LFS pattern matching should be case-insensitive by default (or have an option for it)
- `git lfs pull` should have a warning when patterns don't match due to case
- Documentation should clearly warn about cross-platform case sensitivity issues

**Workaround**

Files can be pulled individually using the full path:
```bash
git lfs pull --include="path/to/file.TGA"
```

Or force checkout specific files:
```bash
git checkout HEAD -- "path/to/file.TGA"
```

Or add duplicate patterns with uppercase extensions to `.gitattributes`:
```
*.tga filter=lfs diff=lfs merge=lfs -text**Files affected in this case:**
- 29 `.TGA` files
- 5 `.PNG` files
- 2 `.BMP` files
*.TGA filter=lfs diff=lfs merge=lfs -text
*.png filter=lfs diff=lfs merge=lfs -text
*.PNG filter=lfs diff=lfs merge=lfs -text
*.bmp filter=lfs diff=lfs merge=lfs -text
*.BMP filter=lfs diff=lfs merge=lfs -text
```

**System environment**

- OS: Linux 6.14.0-37-generic (Ubuntu)
- Files originally pushed from Windows/macOS (case-insensitive filesystems)

**Output of `git lfs env`**

```
git-lfs/3.4.1 (GitHub; linux amd64; go 1.22.2)
git version 2.43.0

[...]

PR #6192, opened 2026-01-16
Docs: recommend bracket-glob patterns for cross-platform case safety in FAQ (refs #6187)

star-ga:
Addresses part of #6187 by documenting the portable bracket-glob workaround in git-lfs-faq(7).

Adds a new FAQ entry under Tracking and Migration explaining that .gitattributes patterns are case-sensitive, which can cause assets committed on macOS/Windows with uppercase extensions to fail matching on Linux CI, leaving LFS pointer files.

Docs-only change; no behavior or performance impact.

Test/verification:
Verified that *.[pP][nN][gG] matches both file.png and file.PNG on a case-sensitive filesystem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -741,13 +741,13 @@ man : $(MAN_ROFF_TARGETS) $(MAN_HTML_TARGETS)
 # Generate ROFF-style man pages from the corresponding .adoc files.
 man/man1/%.1 : docs/man/%.adoc
 	@mkdir -p man/man1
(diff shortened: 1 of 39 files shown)


star-ga:
Updated the FAQ entry to keep wording filesystem/OS-neutral, use the bracket-glob examples directly in `.gitattributes`, and avoid shell-session formatting—aiming for the minimal docs-only guidance discussed in #6187.

PR #6196, opened 2026-01-21
add lfs.httpDownloadEncoding with zstd and gzip support

mzr:
This is to address the issue https://github.com/git-lfs/git-lfs/issues/6193.

Currently the default encoding is gzip, which golang's library adds automatically and transparently from the point of view of the git-lfs.

This commit adds configs `lfs.httpDownloadEncoding` / `lfs.<url>.httpDownloadEncoding`, which allow the git-lfs client to specify which encoding should the backend send the objects encoded with. The HTTP's Accept-Encoding header is used.

The config can be scoped to a url, and only affects the http urls that are returned as a download action from the batch endpoint response. It's because not every object provider can support `zstd`.


chrisd8088:
Hey, thanks very much for drafting this PR!  I'll take a detailed look as soon as I'm able.

tq/basic_download.go
```diff
@@ -131,6 +134,22 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
 		req.Header.Set("Range", fmt.Sprintf("bytes=%d-%d", fromByte, t.Size-1))
 	}
 
+	// Set Accept-Encoding header if configured to zstd
+	// (Go's http client handles gzip automatically when no Accept-Encoding is set)
+	uc := config.NewURLConfig(a.apiClient.GitEnv())
+	httpDownloadEncoding, _ := uc.Get("lfs", rel.Href, "httpdownloadencoding")
```
chrisd8088: Thanks for adopting the `httpDownloadEncoding` terminology!

If you don't mind, I'd like to also change the prefix from just `lfs` to `lfs.transfer`, as I think this new configuration option belongs with the set of other `lfs.transfer.*` [settings](https://github.com/git-lfs/git-lfs/blob/bb65882304a655ffa8abf2be6922e53ff18af5a5/docs/man/git-lfs-config.adoc?plain=1#L192-L231), such as `lfs.transfer.enableHrefRewrite` and `lfs.transfer.maxRetries`.

My reasoning on this point is that like those settings, this new option only affects Git LFS object transfers and not all HTTP requests.  For example, Locking API and Batch API requests will still use the default `gzip` encoding, even if this is set to `zstd`.

I've included this change in one of the commits in my `zstd-download-encoding` [branch](https://github.com/chrisd8088/git-lfs/tree/zstd-download-encoding).

tq/basic_download.go
```diff
@@ -223,8 +242,20 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
 		authOkFunc()
 	}
 
+	// Handle Content-Encoding decompression for zstd
+	// (gzip is handled automatically by Go's http client when we don't set Accept-Encoding)
+	var bodyReader io.Reader = res.Body
+	if res.Header.Get("Content-Encoding") == "zstd" {
```
chrisd8088: If my reading of the Go standard library's `net/http` package code is correct, I believe they [perform](https://github.com/golang/go/blob/d61c0c9283fb1cc3195f113428296976f1fc03ce/src/net/http/transport.go#L2457-L2463) a case-insensitive comparison of the `Content-Encoding` response header's value to the string `gzip`.

That makes sense given the RFC 9110 specification for Content Codings (section 8.4.1), which [states](https://httpwg.org/specs/rfc9110.html#rfc.section.8.4.1) that "All content codings are case-insensitive".

Hence I think we should something like the following here in our comparison:

```suggestion
	if strings.ToLower(res.Header.Get("Content-Encoding")) == "zstd" {
```

I've included this change in one of the commits in my `zstd-download-encoding` [branch](https://github.com/chrisd8088/git-lfs/tree/zstd-download-encoding).

docs/man/git-lfs-config.adoc
```diff
@@ -135,6 +135,22 @@ the version is less than 2.34.0, false otherwise.
...
 
+* `lfs.httpDownloadEncoding` / `lfs.<url>.httpDownloadEncoding`
++
+Specifies the compression encoding to use when downloading LFS objects.
+Valid values are `gzip` or `zstd`.
++
+When set to `gzip`, no explicit header is added and Go's HTTP client
+handles gzip compression automatically. When set to `zstd`, the
+`Accept-Encoding: zstd` header is sent and Git LFS will decompress
+the response.
```
chrisd8088: This is all beautifully written documentation, thank you!

I think it's probably simpler for most users if we restrict the details of Go's HTTP client and how we do or do not override its behaviour to comments in our Go code, as you've already done.  Here I would just say something brief like the following:

```suggestion
Specifies the compression encoding to use when downloading LFS objects
over HTTP. Valid values are `gzip` or `zstd`. Default is `gzip`.
```

docs/man/git-lfs-config.adoc
```diff
@@ -135,6 +135,22 @@ the version is less than 2.34.0, false otherwise.
...
++
+Supports URL config lookup as described in:
+https://git-scm.com/docs/git-config#Documentation/git-config.txt-httplturlgt.
++
+Example: `git config lfs.httpDownloadEncoding "zstd"`
++
+Example (URL-specific): `git config lfs.https://example.com/.httpDownloadEncoding "zstd"`
```
chrisd8088: I know we already have this same [paragraph](https://github.com/git-lfs/git-lfs/blob/bb65882304a655ffa8abf2be6922e53ff18af5a5/docs/man/git-lfs-config.adoc?plain=1#L338-L339) and URL in our entry for the `lfs.<url>.locksverify` configuration option, so this is fine as it stands.

I am slowly in the process of trying to improve the formatting of our manual pages, though, and one of my to-do items is to add link text to any bare URLs using Asciidoc macros.  So if you feel like it, I'd suggest we could format this paragraph similar to the following:

```suggestion
URL configuration lookup follows the logic described in the Git
documentation of the `http.<url>.*`
https://git-scm.com/docs/git-config#Documentation/git-config.txt-httplturlgt[configuration options],
so to set this value for a specific host such as `example.com`, use:
+
....
git config lfs.transfer.https://example.com/.httpDownloadEncoding zstd
....
```

t/t-httpdownloadencoding.sh
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env bash
```
chrisd8088: These shell tests are awesome; thank you so much for writing them!

If you don't mind, I'd like to move these tests into a file named `t/t-batch-transfer-encoding.sh`, as I think that will group them nicely with some of our other tests which pertain specifically to batch object transfers.

I've started this process in PR #6207, which I hope we can merge before this PR.  In particular, that PR moves the sole existing [test](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/t-push.sh#L958-L998) of gzip encoding into a new `t/t-batch-transfer-encoding.sh` file, and updates the test so it can serve as a template for this PR's tests.

tq/basic_download.go
```diff
@@ -223,8 +242,20 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
...
+		zstdReader, err := zstd.NewReader(res.Body)
+		if err != nil {
+			return errors.Wrap(err, tr.Tr.Get("failed to create zstd decompressor"))
+		}
+		defer zstdReader.Close()
+		bodyReader = zstdReader
```
chrisd8088: After the Go standard library's `net/http` package inserts an `Accept-Encoding: gzip` header into a request, if a gzip-compressed response with a `Content-Encoding: gzip` header is received, the package [sets](https://github.com/golang/go/blob/d61c0c9283fb1cc3195f113428296976f1fc03ce/src/net/http/transport.go#L2461) the `ContentLength` element in the `Response` structure it returns to the caller to `-1`.  This signals to the caller that until the response body is read and decompressed, its size is unknown.

As we are in a similar position when handling a response with a zstd-compressed body, we should likewise set the `ContentLength` element in the `Response` structure to `-1`.

One consequence of this will be that users who define the `GIT_LFS_PROGRESS` environment variable will see the `-1` value in their progress logs, rather than the expected size of the Git LFS objects being downloaded.  However, as I noted in https://github.com/git-lfs/git-lfs/pull/6011#issuecomment-3866531617, this is a pre-existing issue which already occurs with gzip-compressed responses.  We should therefore resolve it comprehensively in the progress log callback code, and not try to address it here.

(Technically, we could set the `ContentLength` element to the value from `t.Size` to reflect what we expect the decompressed Git LFS object size to be, but I think we should avoid doing so because we can't guarantee that will actually be the length of the server's response after decompression.)

I'd also suggest that we want to generate a trace log message here similar to the one we [output](https://github.com/git-lfs/git-lfs/blob/bb65882304a655ffa8abf2be6922e53ff18af5a5/lfshttp/client.go#L325) when the `net/http` package has automatically decompressed a response.  This will allow us to easily verify in our shell tests that we are actually reading a zstd-compressed response body.

I've included these changes in two of the commits in my `zstd-download-encoding` [branch](https://github.com/chrisd8088/git-lfs/tree/zstd-download-encoding), and the tests in that branch now search for the additional trace log message when appropriate.

```suggestion

		tracerx.Printf("http: decompressing zstd-encoded response")
		bodyReader = zstdReader

		// Set ContentLength to -1 to match Go's default behaviour
		// when decompressing gzipped responses; see:
		// https://github.com/golang/go/blob/go1.25.7/src/net/http/response.go#L90-L92
		// https://github.com/golang/go/blob/go1.25.7/src/net/http/transport.go#L2385
		res.ContentLength = -1
```

tq/basic_download.go
```diff
@@ -131,6 +134,22 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
 		req.Header.Set("Range", fmt.Sprintf("bytes=%d-%d", fromByte, t.Size-1))
 	}
 
+	// Set Accept-Encoding header if configured to zstd
+	// (Go's http client handles gzip automatically when no Accept-Encoding is set)
+	uc := config.NewURLConfig(a.apiClient.GitEnv())
+	httpDownloadEncoding, _ := uc.Get("lfs", rel.Href, "httpdownloadencoding")
+	if httpDownloadEncoding != "" {
+		httpDownloadEncoding = strings.TrimSpace(httpDownloadEncoding)
```
chrisd8088: We can likely skip the `strings.TrimSpace()` call here, since Git should provide configuration values without leading or trailing whitespace.  (If we needed to trim off whitespace, we should also do so before checking for a non-empty value, but I don't think we need to bother with that.)

I've included this change in one of the commits in my `zstd-download-encoding` [branch](https://github.com/chrisd8088/git-lfs/tree/zstd-download-encoding).

tq/basic_download.go
```diff
@@ -131,6 +134,22 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
...
+		switch httpDownloadEncoding {
+		case "gzip":
+			// Don't set header, let Go's http client handle gzip automatically
+		case "zstd":
+			req.Header.Set("Accept-Encoding", "zstd")
+		default:
```
chrisd8088: We should perhaps one day also allow for a user to set either `gzip,zstd` or `zstd,gzip` as valid options.  This does increase number of test conditions, though, and of course also would require additional documentation.

As an additional complication, though, once we set the `Accept-Encoding` header explicitly, Go won't automatically handle decompression even if the response body was compressed using the `gzip` encoding, as [noted](https://github.com/golang/go/blob/d61c0c9283fb1cc3195f113428296976f1fc03ce/src/net/http/transport.go#L2865-L2868) in the `net/http` source and documentation.  So we'd then have to make sure to handle `gzip` decompression ourselves in this case.

t/t-httpdownloadencoding.sh
```diff
@@ -0,0 +1,117 @@
...
+  # Now test download with invalid Accept-Encoding value
+  rm -rf .git/lfs/objects
+
+  git config lfs.httpDownloadEncoding "br"
+
+  # Should fail with an error
+  set +e
+  git lfs pull 2>&1 | tee pull.log
+  exit_code=$?
+  set -e
```
chrisd8088: This is good defensive programming!

However, we actually don't need to worry about changing the shell's `errexit` shell attribute with `set +e`/`set -e` here because the shell will only exit early if the final command in a pipeline returns a non-zero 
status.  Since our `git lfs pull` command is not the last command in the pipeline, even if it fails, it won't cause the test to fail.

The idiom we often use in these cases is to write instead:

```suggestion
  git lfs pull 2>&1 | tee pull.log
  if [ "0" -eq "${PIPESTATUS[0]}" ]; then
    echo >&2 "fatal: expected 'git lfs pull' to fail ..."
    exit 1
  fi
```

t/t-httpdownloadencoding.sh
```diff
@@ -0,0 +1,117 @@
...
+
+  git config lfs.httpDownloadEncoding "gzip"
+
+  GIT_CURL_VERBOSE=1 git lfs pull 2>&1 | tee curl.log
+
+  # Should not have explicit Accept-Encoding: zstd header
+  if grep "> Accept-Encoding: zstd" curl.log; then
+    echo "Accept-Encoding should not be zstd when gzip is configured"
+    exit 1
+  fi
```
chrisd8088: Our most common idiom for these kinds of checks that a given log message was not output is to use a single-line construction like:

```suggestion
  [ 0 -eq "$(grep -c "Accept-Encoding: zstd" pull.log)" ]
```

I've included examples of this style of check in PR #6207 and in the other comments on this PR where I've tried to expand on these tests and also update our test server utility.

That said, I don't think we need a separate test to check that `Accept-Encoding: zstd` does *not* appear in this case, so long we check that `Accept-Encoding: gzip` appears the correct number of times.  I've included checks of this type in one of the commits in my `zstd-download-encoding` [branch](https://github.com/chrisd8088/git-lfs/tree/zstd-download-encoding), so I think we can probably skip this separate test.

t/t-httpdownloadencoding.sh
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env bash
+
+. "$(dirname "$0")/testlib.sh"
+
+begin_test "lfs.httpDownloadEncoding zstd"
```
chrisd8088: It's great to have a set of tests like these, and I think we can expand on them in two key ways!

First, if we merge the latest version of our `main` development branch into this PR's branch, that will pick up the change from my PR #6195 and so the `Accept-Encoding: gzip` headers output by the `net/http` package will be visible to these tests.

Second, if we modify our `lfstest-gitserver` test utility so that it recognizes an `Accept-Encoding: zstd` header and responds with a zstd-encoded response body, we can verify that the Git LFS client is actually able to successfully decompress such a response.

We can use the technique developed for our existing `storage upload with compression` [test](https://github.com/git-lfs/git-lfs/blob/9afaf60b9c47e96f031f86d04089ef2e0b79a1fa/t/t-push.sh#L958-L998) in PR #4345 where it [creates](https://github.com/git-lfs/git-lfs/blob/9afaf60b9c47e96f031f86d04089ef2e0b79a1fa/t/t-push.sh#L966-L968) a Git LFS object containing a specific string, and the test utility then [detects](https://github.com/git-lfs/git-lfs/blob/9afaf60b9c47e96f031f86d04089ef2e0b79a1fa/t/cmd/lfstest-gitserver.go#L816-L822) that, checks that an appropriate `Accept-Encoding` request header has been sent, and if so [replies](https://github.com/git-lfs/git-lfs/blob/9afaf60b9c47e96f031f86d04089ef2e0b79a1fa/t/cmd/lfstest-gitserver.go#L882-L888) with a matching `Content-Encoding` header and compressed response body.

Back when we wrote PR #4345, I think we overlooked the issue I described in PR #6195 with respect to our historical use of an inappropriate function from the `net/http/httputil` package to capture HTTP request and response headers.  Nevertheless, the test we added PR #4345 is effective at detecting the use of gzip compression on Git LFS object transfer downloads over HTTP.

However, the existing test is located in our `t/t-push.sh` script and its name (`storage upload with compression`) implies that gzip compression is also used during object transfer uploads.  Since that is not the case, I've moved and renamed the test as part of my PR #6207, and reworked the test somewhat to work as a template for this PR's tests.

Here are the changes to the `t/cmd/lfstest-gitserver.go` source file as I implemented them (in a series of logically separate commits) in my `zstd-download-encoding` [branch](https://github.com/chrisd8088/git-lfs/tree/zstd-download-encoding).  That branch is based on top of my PR #6207, so I've only include the differences from that branch below.

<details><summary>Patch for Test Server</summary>

```patch
--- a/t/cmd/lfstest-gitserver.go
+++ b/t/cmd/lfstest-gitserver.go
@@ -36,6 +36,8 @@ import (
 	"sync"
 	"time"
 	"unicode"
+
+	"github.com/klauspost/compress/zstd"
 )
 
 var (
@@ -62,7 +64,8 @@ var (
 		"object-authenticated", "storage-upload-retry", "storage-upload-retry-later", "storage-upload-retry-later-no-header", "unknown-oid",
 		"storage-download-retry-later", "storage-download-retry-later-no-header", "storage-download-retry",
 		"storage-download-retry-range", "storage-download-retry-range-rejected", "storage-download-retry-no-invalid-range",
-		"storage-download-encoding-gzip",
+		"storage-download-encoding-gzip", "storage-download-encoding-zstd", "storage-download-encoding-zstd-retry-range",
+		"storage-download-encoding-zstd-1", "storage-download-encoding-zstd-2", "storage-download-encoding-zstd-3",
 		"send-verify-action", "send-deprecated-links", "redirect-storage-upload", "batch-hash-algo-empty", "batch-hash-algo-invalid",
 		"auth-bearer", "auth-multistage",
 	}
@@ -691,6 +694,8 @@ func verifyHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+var zstdHeaderRE = regexp.MustCompile(`(?:\A|, *)(?i:zstd)(?:, *|\z)`)
+
 // handles any /storage/{oid} requests
 func storageHandler(w http.ResponseWriter, r *http.Request) {
 	id, ok := reqId(w)
@@ -789,6 +794,7 @@ func storageHandler(w http.ResponseWriter, r *http.Request) {
 		statusCode := 200
 		byteLimit := 0
 		compress := false
+		zstdCompress := false
 
 		if by, ok := largeObjects.Get(repo, oid); ok {
 			switch oidHandlers[oid] {
@@ -810,7 +816,6 @@ func storageHandler(w http.ResponseWriter, r *http.Request) {
 					statusCode = 500
 					by = []byte("malformed content")
 				}
-			case "storage-download-retry-range":
 				// Resume if header includes range, otherwise deliberately interrupt
 				if handleRangeRequest(w, r, by) {
 					return
@@ -841,7 +846,38 @@ func storageHandler(w http.ResponseWriter, r *http.Request) {
 				} else {
 					compress = true
 				}
+			case "storage-download-encoding-zstd",
+				"storage-download-encoding-zstd-1",
+				"storage-download-encoding-zstd-2",
+				"storage-download-encoding-zstd-3":
+				if zstdHeaderRE.MatchString(r.Header.Get("Accept-Encoding")) {
+					zstdCompress = true
+				} else {
+					statusCode = http.StatusInternalServerError
+					by = []byte("not encoded")
+				}
+			case "storage-download-encoding-zstd-retry-range":
+				// Resume if header includes range, otherwise deliberately interrupt
+				if handleRangeRequest(w, r, by) {
+					return
+				}
+
+				if zstdHeaderRE.MatchString(r.Header.Get("Accept-Encoding")) {
+					zstdCompress = true
+
+					// Note that this value is an offset
+					// into the second encoded frame, so
+					// we interrupt the response after a
+					// complete first frame.  This ensures
+					// the client will write the decoded
+					// first frame to a temporary file.
+					byteLimit = 6
+				} else {
+					statusCode = http.StatusInternalServerError
+					by = []byte("not encoded")
+				}
 			}
+
 			var wrtr io.Writer = w
 			if compress {
 				w.Header().Set("Content-Encoding", "gzip")
@@ -849,7 +885,39 @@ func storageHandler(w http.ResponseWriter, r *http.Request) {
 				defer gz.Close()
 
 				wrtr = gz
-			}
+			} else if zstdCompress {
+				w.Header().Set("Content-Encoding", "zstd")
+				enc, err := zstd.NewWriter(w)
+				if err == nil {
+					defer enc.Close()
+
+					// We need the encoded data's length to
+					// set the Content-Length header when
+					// byteLimit > 0, so we encode the data
+					// here rather than set wrtr = enc,
+					// which would use the stream encoder.
+					// This results in a less-compact
+					// encoding, with a different length.
+					//
+					// We create two encoded frames so we
+					// can interrupt the second one when
+					// byteLimit > 0, allowing the client
+					// to fully decode the first frame and
+					// write it to a temporary file.
+					mid := len(by) / 2
+					frame1 := enc.EncodeAll(by[0:mid], make([]byte, 0, mid*2))
+					frame2 := enc.EncodeAll(by[mid:], make([]byte, 0, mid*2))
+					by = append(frame1, frame2...)
+
+					if byteLimit > 0 {
+						byteLimit += len(frame1)
+					}
+				} else {
+					statusCode = http.StatusInternalServerError
+					by = []byte("not encoded")
+					byteLimit = 0
+	                        }
+	                }
 
 			if byteLimit > 0 {
 				// Force Content-Length header to report the
```
</details>

As for these tests, I think we can combine the first three into two and expand them a bit to cover various use cases.  (The "invalid value" test would still be a separate test.)

My PR #6207 already moves the `storage upload with compression` test from the `t/t-push.sh` file into a new `t/t-batch-storage-encoding.sh` script, renames the test to `batch storage HTTP download with gzip encoding`, and refactors it a bit to better serve as a template for additional tests to be added with this PR.

In my `zstd-download-encoding` [branch](https://github.com/chrisd8088/git-lfs/tree/zstd-download-encoding) I then move the commits from this `t/t-httpdownloadencoding.sh` file into the new `t/t-batch-storage-encoding.sh` script, refactor them, and expand the gzip-encoding test.

My goal was to incorporate a few changes for the following purposes:

- Check multiple related configuration settings in a single test
- Use `GITSERVER` variable in URL-specific configurations for simplicity
- Check presence or absence of all `Accept-Encoding` headers with `grep -c`
- Use `GIT_TRACE=1` to capture trace log messages re content decompression
- Rename trace log files to `pull.log` per our usual style
- Check `git lfs pull` command success using `PIPESTATUS` array
- Check for local objects using `assert_local_object()`
- Rephrase test names to reflect batch storage HTTP download context

Here are the changes to the `t/t-batch-storage-encoding.sh` script as I implemented them (in a series of logically separate commits) in my `zstd-download-encoding` [branch](https://github.com/chrisd8088/git-lfs/tree/zstd-download-encoding).  That branch is based on top of my PR #6207, so I've only included the differences from that branch below.

<details><summary>Patch for Shell Tests</summary>

```patch
--- a/t/t-batch-storage-encoding.sh
+++ b/t/t-batch-storage-encoding.sh
@@ -40,5 +40,264 @@ begin_test "batch storage HTTP download with gzip encoding"
   [ 1 -eq "$(grep -c "decompressed gzipped response" pull.log)" ]
 
   assert_local_object "$contents_oid" "${#contents}"
+
+  # Test again with an explicit configuration.
+  rm -rf .git/lfs/objects
+
+  git config lfs.transfer.httpDownloadEncoding gzip
+
+  GIT_TRACE=1 GIT_CURL_VERBOSE=1 git lfs pull 2>&1 | tee pull.log
+  if [ "0" -ne "${PIPESTATUS[0]}" ]; then
+    echo >&2 "fatal: expected 'git lfs pull' to succeed ..."
+    exit 1
+  fi
+
+  # We expect one "Accept-Encoding: gzip" header from the Batch API request,
+  # prior to the object transfer download request.
+  [ 2 -eq "$(grep -c "Accept-Encoding: gzip" pull.log)" ]
+
+  [ 1 -eq "$(grep -c "decompressed gzipped response" pull.log)" ]
+
+  assert_local_object "$contents_oid" "${#contents}"
+
+  # Test again with a URL-specific configuration.
+  rm -rf .git/lfs/objects
+
+  git config --unset lfs.transfer.httpDownloadEncoding
+  git config "lfs.transfer.$GITSERVER.httpDownloadEncoding" gzip
+
+  GIT_TRACE=1 GIT_CURL_VERBOSE=1 git lfs pull 2>&1 | tee pull.log
+  if [ "0" -ne "${PIPESTATUS[0]}" ]; then
+    echo >&2 "fatal: expected 'git lfs pull' to succeed ..."
+    exit 1
+  fi
+
+  # We expect one "Accept-Encoding: gzip" header from the Batch API request,
+  # prior to the object transfer download request.
+  [ 2 -eq "$(grep -c "Accept-Encoding: gzip" pull.log)" ]
+
+  [ 1 -eq "$(grep -c "decompressed gzipped response" pull.log)" ]
+
+  assert_local_object "$contents_oid" "${#contents}"
+)
+end_test
+
+begin_test "batch storage HTTP download with zstd encoding"
+(
+  set -e
+
+  reponame="batch-storage-download-encoding-zstd"
+
+  git lfs track "*.dat"
+
+  # This content announces to the server that it should expect an
+  # "Accept-Encoding: zstd" header and send a zstd-compressed response.
+  contents="storage-download-encoding-zstd"
+  printf "%s" "$contents" >a.dat
+
+
+  git push origin main
+
+  # Test object transfer download with an "Accept-Encoding: zstd" header.
+  rm -rf .git/lfs/objects
+
+  git config lfs.transfer.httpDownloadEncoding zstd
+
+  GIT_TRACE=1 GIT_CURL_VERBOSE=1 git lfs pull 2>&1 | tee pull.log
+  if [ "0" -ne "${PIPESTATUS[0]}" ]; then
+    echo >&2 "fatal: expected 'git lfs pull' to succeed ..."
+    exit 1
+  fi
+
+  # We expect one "Accept-Encoding: gzip" header from the Batch API request,
+  # prior to the object transfer download request.
+  [ 1 -eq "$(grep -c "Accept-Encoding: gzip" pull.log)" ]
+  [ 1 -eq "$(grep -c "Accept-Encoding: zstd" pull.log)" ]
+
+  [ 1 -eq "$(grep -c "Content-Encoding: zstd" pull.log)" ]
+  [ 1 -eq "$(grep -c "decompressing zstd-encoded response" pull.log)" ]
+
+  assert_local_object "$contents_oid" "${#contents}"
+
+  # Test again with a URL-specific configuration.
+  rm -rf .git/lfs/objects
+
+  git config --unset lfs.transfer.httpDownloadEncoding
+  git config "lfs.transfer.$GITSERVER.httpDownloadEncoding" zstd
+
+  GIT_TRACE=1 GIT_CURL_VERBOSE=1 git lfs pull 2>&1 | tee pull.log
+  if [ "0" -ne "${PIPESTATUS[0]}" ]; then
+    echo >&2 "fatal: expected 'git lfs pull' to succeed ..."
+    exit 1
+  fi
+
+  # We expect one "Accept-Encoding: gzip" header from the Batch API request,
+  # prior to the object transfer download request.
+  [ 1 -eq "$(grep -c "Accept-Encoding: gzip" pull.log)" ]
+  [ 1 -eq "$(grep -c "Accept-Encoding: zstd" pull.log)" ]
+
+  [ 1 -eq "$(grep -c "Content-Encoding: zstd" pull.log)" ]
+  [ 1 -eq "$(grep -c "decompressing zstd-encoded response" pull.log)" ]
+
+  assert_local_object "$contents_oid" "${#contents}"
+)
+end_test
+
+begin_test "batch storage HTTP download with zstd encoding and multiple objects"
+(
+  set -e
+
+  reponame="batch-storage-download-encoding-zstd-multiple"
+
+  git lfs track "*.dat"
+
+  # This content announces to the server that it should expect an
+  # "Accept-Encoding: zstd" header and send a zstd-compressed response.
+  contents1="storage-download-encoding-zstd-1"
+  contents2="storage-download-encoding-zstd-2"
+  contents3="storage-download-encoding-zstd-3"
+  contents1_oid=$(calc_oid "$contents1")
+  contents2_oid=$(calc_oid "$contents2")
+  contents3_oid=$(calc_oid "$contents3")
+  printf "%s" "$contents1" >test1.dat
+  printf "%s" "$contents2" >test2.dat
+  printf "%s" "$contents3" >test3.dat
+
+  git add .gitattributes test*.dat
+
+  git push origin main
+
+  # Test object transfer download with an "Accept-Encoding: zstd" header.
+  rm -rf .git/lfs/objects
+
+  git config lfs.transfer.httpDownloadEncoding zstd
+
+  # Allow no more than two concurrent transfer workers, each with a
+  # dedicated zstd decoder.
+  git config lfs.concurrentTransfers 2
+
+  GIT_TRACE=1 GIT_CURL_VERBOSE=1 git lfs pull 2>&1 | tee pull.log
+  if [ "0" -ne "${PIPESTATUS[0]}" ]; then
+    echo >&2 "fatal: expected 'git lfs pull' to succeed ..."
+    exit 1
+  fi
+
+  # We expect one "Accept-Encoding: gzip" header from the Batch API request,
+  # prior to the object transfer download requests.
+  [ 1 -eq "$(grep -c "Accept-Encoding: gzip" pull.log)" ]
+  [ 3 -eq "$(grep -c "Accept-Encoding: zstd" pull.log)" ]
+
+  [ 3 -eq "$(grep -c "Content-Encoding: zstd" pull.log)" ]
+  [ 3 -eq "$(grep -c "decompressing zstd-encoded response" pull.log)" ]
+
+  # We expect one zstd decoder to be initialized for each transfer worker.
+  [ 2 -eq "$(grep -c "initialized zstd decoder" pull.log)" ]
+  [ 2 -eq "$(grep -c "closed zstd decoder" pull.log)" ]
+
+  assert_local_object "$contents1_oid" "${#contents1}"
+  assert_local_object "$contents2_oid" "${#contents2}"
+  assert_local_object "$contents3_oid" "${#contents3}"
+)
+end_test
+
+begin_test "batch storage HTTP download with zstd encoding retries with Range header and without encoding"
+(
+  set -e
+
+  reponame="batch-storage-download-encoding-zstd-retry-range"
+
+  git lfs track "*.dat"
+
+  # This content announces to the server that it should expect an
+  # "Accept-Encoding: zstd" header and send a zstd-encoded response
+  # unless a Range header with a positive offset was sent.  As well,
+  # the server should interrupt the initial response after at least one
+  # complete zstd-encoded frame has been delivered.
+  contents="storage-download-encoding-zstd-retry-range"
+  printf "%s" "$contents" >a.dat
+
+
+  git push origin main
+
+  # Test object transfer download with an "Accept-Encoding: zstd" header
+  # and an interrupted initial response, after which the client should
+  # fetch the remaining bytes using a request with a Range header and
+  # without an "Accept-Encoding: zstd" header.
+  rm -rf .git/lfs/objects
+
+  git config lfs.transfer.httpDownloadEncoding zstd
+
+  GIT_TRACE=1 GIT_CURL_VERBOSE=1 git lfs fetch 2>&1 | tee fetch.log
+  if [ "0" -ne "${PIPESTATUS[0]}" ]; then
+    echo >&2 "fatal: expected 'git lfs fetch' to succeed ..."
+    exit 1
+  fi
+
+  # We expect one "Accept-Encoding: gzip" header from the Batch API request
+  # prior to each object transfer download request.  We expect one object
+  # transfer download request with an "Accept-Encoding: zstd" header, then a
+  # retried request with a Range header but without a Accept-Encoding header.
+  [ 2 -eq "$(grep -c "Accept-Encoding: gzip" fetch.log)" ]
+  [ 1 -eq "$(grep -c "Accept-Encoding: zstd" fetch.log)" ]
+
+  [ 1 -eq "$(grep -c "Content-Encoding: zstd" fetch.log)" ]
+  [ 1 -eq "$(grep -c "decompressing zstd-encoded response" fetch.log)" ]
+
+  # The server should encode the object data in two frames of equal length and
+  # interrupt the first download response after the first frame, so we expect
+  # the next download attempt to request just the second half of the data.
+  grep "Attempting to resume download of \"$contents_oid\"" fetch.log
+  grep "tq: retrying object $contents_oid" fetch.log
+  grep "Range: bytes=$((${#contents} / 2))-$((${#contents} - 1))" fetch.log
+
+  grep "206 Partial Content" fetch.log
+  grep "Content-Range: bytes $((${#contents} / 2))-$((${#contents} - 1))/${#contents}" fetch.log
+  grep "xfer: server accepted resume .*$contents_oid" fetch.log
+
+  assert_local_object "$contents_oid" "${#contents}"
+)
+end_test
+
+begin_test "batch storage HTTP download with invalid encoding"
+(
+  set -e
+
+  reponame="batch-storage-download-encoding-invalid"
+
+  git lfs track "*.dat"
+  printf "contents" > a.dat
+
+  git push origin main
+
+  # Now test download with invalid Accept-Encoding value
+  rm -rf .git/lfs/objects
+
+  git config lfs.transfer.httpDownloadEncoding "br"
+
+  # Should fail with an error
+  git lfs pull 2>&1 | tee pull.log
+  if [ "0" -eq "${PIPESTATUS[0]}" ]; then
+    echo >&2 "fatal: expected 'git lfs pull' to fail ..."
+    exit 1
+  fi
+
+  grep "unsupported lfs\.transfer\.httpDownloadEncoding" pull.log
 )
 end_test
```
</details>

chrisd8088:
Hey, thanks so much for this PR!

I've been reviewing it over the last few days and am still writing up my notes, but in the process I've turned up a number of problems with our existing test suite which I'd like to fix first, so that I don't have to complicate this PR's review with pre-existing bug fixes.

Thanks again and I'll update this PR as soon as I can!

tq/basic_download.go
```diff
@@ -131,6 +134,22 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
 		req.Header.Set("Range", fmt.Sprintf("bytes=%d-%d", fromByte, t.Size-1))
 	}
```
chrisd8088: We should enclose the entirety of the new code section below in an `else` block of the `fromByte > 0` conditional above, because we should only send an `Accept-Encoding: zstd` request header if we are not sending a `Range` header.

The byte offsets we provide with a `Range` header are offsets into the uncompressed content and therefore won't make sense when used with a compressed response.  It's for this reason that the Go standard library's `net/http` package will [not insert](https://github.com/golang/go/blob/d61c0c9283fb1cc3195f113428296976f1fc03ce/src/net/http/transport.go#L2872) an `Accept-Encoding: gzip` request header if a `Range` header is specified, and so we need to follow the same logic.

To test this behaviour we'll need to further patch our `lfstest-gitserver` test utility as well.

I've included these changes in one of the commits in my `zstd-download-encoding` [branch](https://github.com/chrisd8088/git-lfs/tree/zstd-download-encoding), making use of the refactored `Range` header parsing in the `lfstest-gitserver` utility from my PR #6207.

tq/basic_download.go
```diff
@@ -223,8 +242,20 @@ func (a *basicDownloadAdapter) download(t *Transfer, cb ProgressCallback, authOk
...
 
+	// Handle Content-Encoding decompression for zstd
+	// (gzip is handled automatically by Go's http client when we don't set Accept-Encoding)
+	var bodyReader io.Reader = res.Body
+	if res.Header.Get("Content-Encoding") == "zstd" {
+		zstdReader, err := zstd.NewReader(res.Body)
+		if err != nil {
+			return errors.Wrap(err, tr.Tr.Get("failed to create zstd decompressor"))
+		}
+		defer zstdReader.Close()
```
chrisd8088: According to the documentation of the `github.com/klauspost/compress/zstd` package, by default four goroutines are [started](https://github.com/klauspost/compress/blob/v1.18.4/zstd/README.md?plain=1#L296) for each stream-based decompressor, which serve various [purposes](https://github.com/klauspost/compress/blob/v1.18.4/zstd/README.md?plain=1#L374-L379) with the goal of making data available for output as fast as possible.

By default, we [run](https://github.com/git-lfs/git-lfs/blob/9afaf60b9c47e96f031f86d04089ef2e0b79a1fa/tq/adapterbase.go#L91-L97) [eight](https://github.com/git-lfs/git-lfs/blob/9afaf60b9c47e96f031f86d04089ef2e0b79a1fa/tq/manifest.go#L17) separate goroutines to download Git LFS objects, which means we could be creating as many as 32 goroutines for zstd decompression.  I think that's likely more than will be effective in the typical use case, and so we many need to limit the number of goroutines each decompressor creates with a `WithDecoderConcurrency(n)` setting.

A related issue is that according to the package's documentation, every time we create a new decompressor with `zstd.NewReader()` it spends time [allocating](https://github.com/klauspost/compress/blob/v1.18.4/zstd/README.md?plain=1#L355-L359) internal resources, which can be reused to achieve better performance.  Since downloading multiple Git LFS objects is very common, we should make sure to reuse the decompressors as much as possible.

The obvious thing to do is to add a decompressor to our `basicDownloadAdapter` structure [type](https://github.com/git-lfs/git-lfs/blob/9afaf60b9c47e96f031f86d04089ef2e0b79a1fa/tq/basic_download.go#L20-L22), except that we share those structures between all the worker goroutines, so we'd have to add a mutex around the use of the decompressor, which is not ideal.  Instead, we should initialize a decompressor in the `WorkerStarting()` method and store it as an element in a new `basicDownloadAdapterWorkerContext` structure type, and then return a pointer to an instance of that type as the `ctx` return value, so that the `WorkerEnding()` method can close the decompressor.  There's already a [model](https://github.com/git-lfs/git-lfs/blob/9afaf60b9c47e96f031f86d04089ef2e0b79a1fa/tq/custom.go#L148-L168) for this in the custom transfer adapter code, where each worker starts an external command process.

Here's what I implemented in one of the commits in my `zstd-download-encoding` [branch](https://github.com/chrisd8088/git-lfs/tree/zstd-download-encoding), along with a test and some minor changes to our `lfstest-gitserver` utility so that we can verify this works as intended:

```go
type basicDownloadAdapterWorkerContext struct {
	zstdDecoder *zstd.Decoder
}

func (a *basicDownloadAdapter) WorkerStarting(workerNum int) (interface{}, error) {
	return &basicDownloadAdapterWorkerContext{}, nil
}

func (a *basicDownloadAdapter) WorkerEnding(workerNum int, ctx interface{}) {
	context, ok := ctx.(*basicDownloadAdapterWorkerContext)
	if !ok {
		tracerx.Printf("context object for basic download transfer adapter was of the wrong type")
		return
	}

	if context.zstdDecoder != nil {
		context.zstdDecoder.Close()
		tracerx.Printf("http: closed zstd decoder")
	}
}
```

The very start of the `DoTransfer()` method then performs a quick type assertion check so we don't have to repeat that in the `download()` method, since that method may recurse:

```go
	context, ok := ctx.(*basicDownloadAdapterWorkerContext)
	if !ok {
		return errors.New(tr.Tr.Get("context object for basic download transfer adapter was of the wrong type"))
	}
```

The signature of the `download()` method gains a `context *basicDownloadAdapterWorkerContext` argument so when `DoTransfer()` [called](https://github.com/git-lfs/git-lfs/blob/9afaf60b9c47e96f031f86d04089ef2e0b79a1fa/tq/basic_download.go#L97) the `download()` method can just pass `context` (and the same thing again when the `download()` method [recurses](https://github.com/git-lfs/git-lfs/blob/9afaf60b9c47e96f031f86d04089ef2e0b79a1fa/tq/basic_download.go#L152) and calls itself):

```go
	err = a.download(context, t, cb, authOkFunc, f, fromByte, hash)
```

Finally, to handle an individual object's download transfer, the `download()` method does the following to ensure we only create decompressors when and if they are needed:

```suggestion
		zstdDecoder := context.zstdDecoder
		if zstdDecoder == nil {
			zstdDecoder, err := zstd.NewReader(res.Body, zstd.WithDecoderConcurrency(1))
			if err != nil {
				return errors.Wrap(err, tr.Tr.Get("failed to create zstd decompressor"))
			}
			context.zstdDecoder = zstdDecoder
			tracerx.Printf("http: initialized zstd decoder")
		} else {
			zstdDecoder.Reset(res.Body)
		}
		bodyReader = context.zstdDecoder
```

go.mod
```diff
@@ -36,6 +36,7 @@ require (
 	github.com/jcmturner/goidentity/v6 v6.0.1 // indirect
 	github.com/jcmturner/gokrb5/v8 v8.4.2 // indirect
 	github.com/jcmturner/rpc/v2 v2.0.3 // indirect
+	github.com/klauspost/compress v1.18.3 // indirect
```
chrisd8088: I don't think this entry should be in the `require` section with the other indirectly-referenced modules, since we import the `klauspost/compress/zstd` package directly.

If I run the `go mod tidy` command, it rewrites the `go.mod` file and moves this entry up to the preceding `require` section, so we should do the same.

```patch
diff --git a/go.mod b/go.mod
index 0683709e..afee0683 100644
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/git-lfs/wildmatch/v2 v2.0.1
 	github.com/golang/groupcache v0.0.0-20241129210726-2c02b8208cf8
 	github.com/jmhodges/clock v1.2.0
+	github.com/klauspost/compress v1.18.3
 	github.com/leonelquinteros/gotext v1.5.0
 	github.com/mattn/go-isatty v0.0.4
 	github.com/olekukonko/ts v0.0.0-20171002115256-78ecb04241c0
@@ -36,7 +37,6 @@ require (
 	github.com/jcmturner/goidentity/v6 v6.0.1 // indirect
 	github.com/jcmturner/gokrb5/v8 v8.4.2 // indirect
 	github.com/jcmturner/rpc/v2 v2.0.3 // indirect
-	github.com/klauspost/compress v1.18.3 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/xeipuuv/gojsonpointer v0.0.0-20180127040702-4e3ac2762d5f // indirect
```

chrisd8088 (commented):
Thank you so much for this PR!

Reviewing it has certainly been an educational experience and one which led to multiple avenues of investigation!

My PRs #6195 and #6207 were prompted by this PR, and in addition, your PR led to an illuminating discussion in klauspost/compress#1129 regarding zstd frames and unexpected EOFs, and also highlighted a pre-existing (but minor) [bug](https://github.com/git-lfs/git-lfs/pull/6011#issuecomment-3866531617) with our progress logs.

I see there is a proposal in golang/go#62492 to add support in the `net/http` package for automatic zstd decompression, which looks to depend on the introduction of a public `compress/zstd` module, as tracked by golang/go#62513.  However, neither appears likely to be addressed in the short term, so I think this PR is valuable.

I developed PR #6207 in the process of writing up my review comments, and at a minimum I'd like to see that merged first so that this PR's tests can mirror the gzip-encoding test I refactored out of an existing [test](https://github.com/git-lfs/git-lfs/blob/8509f933d2bd1bb1dcc444611dfef6f46889f824/t/t-push.sh#L958-L998).  Note that I renamed the test and moved it into a new script named `t/t-batch-storage-encoding.sh`.  

I also experimented with a number of potential enhancements to this PR to try to address the issues which turned up during my review, and ultimately drafted a series of commits in my `zstd-download-encoding` [branch](https://github.com/chrisd8088/git-lfs/tree/zstd-download-encoding) that I hope could be pulled into this PR.

Note that my branch starts from this PR's current state, merges PR #6207's branch, and then adds another group of commits on top.

I didn't try to address everything in that branch, just some of the key issues regarding transfer queue workers, multi-threading, and HTTP retry handling, plus a couple of other details.

I also tried to make sure the server we use in our shell test suite will send zstd-encoded responses when requested, so our tests can actually verify that we decode those responses correctly.

However, I didn't make any changes to the documentation or the placement of the `github.com/klauspost/compress` module in our `go.mod` file; those points still need to addressed.

At any rate, let me know what you think of all of that, and if you're willing to pull my changes into this PR, then we can ask for a review from the other member of our core team to get a third pair of eyes on the code.

Thank you again for proposing this new feature and writing up this PR!  We really appreciate initiatives like this from our community!

tq/basic_download.go
mzr: Ok, that sounds good!

mzr:
Thanks @chrisd8088 for a very thorough review! I'm impressed by your attention to detail in this PR and in the PRs you handled beforehand to make this one fit better into git-lfs. 

I looked into some easy-to-address comments, but as I went through them I realised I need a proper sitting with this.

I agree it would be ideal to just reuse zstd from golang, but if that's doesn't get traction, it shouldn't stop us from doing it ourselves here.

First thing I'll do, I'll rebase on top [your branch](https://github.com/chrisd8088/git-lfs/tree/zstd-download-encoding), then I'll address your comments.

chrisd8088:
Just FYI, we've merged PR #6207 now, so if you rebase on `main`, I think that should simplify this PR a bit, since the `t/t-batch-storage-encoding.sh` script is now part of our main development branch.

mzr:
Squashed & rebased

chrisd8088:
Thank you for rebasing this PR, it's looking really good to me!

Because I contributed a fair amount to this PR, I'm going to ask for a review from another maintainer of the project before we merge this PR.

Before I do that, though, I'd like to "un-squash" the changes from my `zstd-download-encoding` [branch](https://github.com/chrisd8088/git-lfs/tree/zstd-download-encoding), specifically the commits from 52892610d8f3ada35865da142e7c772352963abe to ec9ba2e9bccad0bd348e61a1ececdbfd2bcb6115, inclusive (i.e., the range `16fc463b8a01..ec9ba2e9bcca`).  I'm happy to take care of this, though, if you don't mind my force-pushing to your branch.

I divided up my work into a series of commits so that each one has a discrete, logically-related set of changes and a detailed description of the purpose and context for those changes.  That should make it easier for my colleague to review the PR as a whole and understand how the pieces all fit together.

(The other reason I try to include notes with each of my commits is so that in the future they'll help me—or anyone else looking at the history of the project—figure out why we did something a particular way!)

Other than "un-squashing", though, I think we're ready for a second review.

FWIW, I ran a few tests just to make sure that the addition of the per-URL configuration option search to the `download()` method of the `tq` package's `basicDownloadAdapter` structure didn't have a material impact on the performance of the download transfer queue, and it seems fine to me.  On my older macOS system with a modestly-sized Git configuration file, the search for a matching `lfs.transfer.<url>.httpDownloadEncoding` option adds about 20 μs to each Git LFS object download, which is dwarfed by the HTTP transfer time and the filesystem I/O.

(It's maybe worth noting here that the current Git configuration is only read once.  Each [search](https://github.com/git-lfs/git-lfs/blob/dd3714055863a3eb158950e75f34c8e352eaf9b8/config/url_config.go#L28-L39) is then just performed [using](https://github.com/git-lfs/git-lfs/blob/dd3714055863a3eb158950e75f34c8e352eaf9b8/config/url_config.go#L72) a copy of the [cached](https://github.com/git-lfs/git-lfs/blob/dd3714055863a3eb158950e75f34c8e352eaf9b8/config/git_fetcher.go#L141-L154) configuration data.)

Thanks again so much for contributing this PR and initiating this new feature for Git LFS!

mzr:
Thank you for this great summary and driving a ton of changes to deliver this feature! I don't mind you force pushing to my branch at all.

chrisd8088:
> I don't mind you force pushing to my branch at all.

Wonderful, thank you very much.  I've done so, and now we'll just look for a second reviewer on this before we merge.

Thanks again!

chrisd8088 (approved):

tq/basic_download.go
```diff
@@ -31,13 +38,28 @@ func (a *basicDownloadAdapter) tempDir() string {
...
 func (a *basicDownloadAdapter) WorkerStarting(workerNum int) (interface{}, error) {
-	return nil, nil
+	return &basicDownloadAdapterWorkerContext{}, nil
 }
 
 func (a *basicDownloadAdapter) WorkerEnding(workerNum int, ctx interface{}) {
+	context, ok := ctx.(*basicDownloadAdapterWorkerContext)
+	if !ok {
+		tracerx.Printf("context object for basic download transfer adapter was of the wrong type")
+		return
```
larsxschneider, line 48: If this would happen, then this would be a bug in the code, no? Shouldn't we rather panic to make any test fail instead of a silent log?

Granted, the error in `DoTransfer` would trigger first anyways and return surface the error.

larsxschneider (approved):
Nice work. Thank you 🙇

tq/basic_download.go
chrisd8088: Indeed, it would be a bug!

This is the idiom we already [use](https://github.com/git-lfs/git-lfs/blob/653422d164356dd68d705cf33791eeab7e6dda7d/tq/custom.go#L247-L250) in `tq/custom.go` in the equivalent `WorkerEnding()` method for the `customAdapter` structure.  (It's been in place since PR #1367 in 2016, apparently.)

Meanwhile, in `tq/ssh.go`, we don't even bother to check the type [assertion](https://github.com/git-lfs/git-lfs/blob/653422d164356dd68d705cf33791eeab7e6dda7d/tq/ssh.go#L180).

I'm comfortable with the logic here whereby the `DoTransfer()` method will raise an error if there's a problem, and so `WorkerEnding()` just needs to log the same, since at this point the process must be shutting down anyway.

If we run a test with the existing custom transfer adapter after injecting an invalid type in the [return](https://github.com/git-lfs/git-lfs/blob/653422d164356dd68d705cf33791eeab7e6dda7d/tq/custom.go#L168) value from the `WorkerStarting()`  method, we get errors like the following, which seem sufficient to me:

```
context object for custom transfer "testcustom" was of the wrong type
context object for custom transfer "testcustom" was of the wrong type
...
error: failed to push some refs to 'http://127.0.0.1:56559/test-custom-transfer-1'
```

If you feel we ought to do more than this, though, I think that should be something we change in all the transfer adapters at the same time in a separate PR, so I'm still inclined to leave this as written in order to match the existing idiom for now.

slonopotamus:
Is there a reason why zstd is not accepted by default?

Also: is it possible to enable zstd via repo `.lfsconfig`?

chrisd8088:
> Is there a reason why zstd is not accepted by default?

@slonopotamus — There's no reason which prevents this being made the default in a future release.  I held back from doing so in this PR both because we try to be a bit cautious about changing long-standing defaults, and because it will complicate the implementation for the reason I described in https://github.com/git-lfs/git-lfs/pull/6196#discussion_r2778584876 above.

Go's `net/http` package automatically inserts an `Accept-Encoding: gzip` request header if no `Accept-Encoding` header is specified (and if the `DisableCompression` [field](https://github.com/golang/go/blob/go1.27.0/src/net/http/transport.go#L199-L207) is not set to `true` in the `http.Transport` structure), and then if the response includes a `Content-Encoding: gzip` header, Go automatically decompresses the response body

However, the `net/http` package only [performs](https://github.com/golang/go/blob/go1.27.0/src/net/http/transport.go#L2586-L2592) that automatic gzip-decompression if the package itself added the `Accept-Encoding: gzip` request header.

So, if we explicitly set an `Accept-Encoding: zstd, gzip` request header, then we have handle gzip-decompression ourselves.  While that's perfectly possible, it was more than I thought we should try to achieve in a first implementation.

If there's demand for zstd support by default, that's obviously something which can be addressed in a future PR.  Of course, there's the possibility that some Git LFS services may not correctly handle an `Accept-Encoding: zstd, gzip` request header, but since that's a well-formed HTTP header, it would be a problem those services would need to address on their own.

> Also: is it possible to enable zstd via repo `.lfsconfig`?

Not at the moment, but if someone wants to propose adding that feature, or simply making zstd support the default, we'd certainly be glad to take a look at any PRs.

PR #6200, opened 2026-02-03
lfshttp: fix auth response ExpiresIn data type

aymanbagabas:
The git config value "lfs.defaulttokenttl" can be larger than max int on i386 systems. We need to change this field to int64/time.Duration so that it can hold larger values.

This commit also updates the Environment interface and add Int64 method to support fetching int64 values from git config.

Fixes: https://github.com/charmbracelet/soft-serve/issues/781

### DIFF
--- a/commands/command_migrate_import.go
+++ b/commands/command_migrate_import.go
@@ -64,7 +64,7 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 
 		root := commit.TreeID
 
-		filter := git.GetAttributeFilter(cfg.LocalWorkingDir(), cfg.LocalGitDir())
+		filter := gitattr.GetAttributeFilter(cfg.LocalWorkingDir(), cfg.LocalGitDir())
 		if len(filter.Include()) == 0 {
 			ExitWithError(errors.New(tr.Tr.Get("No Git LFS filters found in '.gitattributes'")))
 		}
@@ -194,6 +194,9 @@ func migrateImportCommand(cmd *cobra.Command, args []string) {
 
 				fixups, err = gitattr.New(db, t)
 				if err != nil {
+					return err
+				}
+				if err = fixups.FindSpecialAttributes(cfg.GitEnv(), cfg.OSEnv(), cfg.LocalGitDir()); err != nil {
 					return err
 				}
 			}
--- a/commands/command_migrate_info.go
+++ b/commands/command_migrate_info.go
@@ -172,6 +172,9 @@ func migrateInfoCommand(cmd *cobra.Command, args []string) {
 					if err != nil {
 						return err
 					}
+					if err = fixups.FindSpecialAttributes(cfg.GitEnv(), cfg.OSEnv(), cfg.LocalGitDir()); err != nil {
+						return err
+					}
 				}
 				return nil
 			}
--- a/commands/command_track.go
+++ b/commands/command_track.go
@@ -59,9 +59,9 @@ func trackCommand(cmd *cobra.Command, args []string) {
 
 	// Intentionally do _not_ consider global- and system-level
 	// .gitattributes here.  Parse them still to expand any macros.
-	git.GetSystemAttributePaths(mp, cfg.Os)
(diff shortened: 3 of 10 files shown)


config/delayed_environment.go
```diff
@@ -43,6 +43,13 @@ func (e *delayedEnvironment) Int(key string, def int) int {
 	return e.env.Int(key, def)
 }
 
+// Get is shorthand for calling the e.Load(), and then returning
```
chrisd8088: Although there are similar typos in the comments above, we should at least avoid adding another one, so I'd suggest using the new method's name in the comment:

```suggestion
// Int64 is shorthand for calling the e.Load(), and then returning
```

The first of these typos was introduced back in commit 04c27548a2f40f9af42dc399ea4f31dbca0fd416 from PR #1450, and we've inadvertently retained and duplicated it through various refactoring PRs since then.

config/environment.go
```diff
@@ -41,6 +41,17 @@ type Environment interface {
 	// then it will be returned wholesale.
 	Int(key string, def int) (val int)
 
+	// Int64 returns the int64 value associated with a given key, or the value
+	// "def", if no value was associated.
+	//
+	// To convert from a the string value attached to a given key,
```
chrisd8088: Although there are similar typos in the comments above, we should at least avoid adding another one, so I'd suggest using the new method's name in the comment:

```suggestion
	// To convert from the string value attached to a given key,
```

The first of these typos was introduced back in commit 18cb9257f90f2d7b122c4e0bf087ce0ec72cccb0 from PR  #1419, and we've inadvertently retained and duplicated it through various PRs since then.

config/environment.go
```diff
@@ -101,6 +117,28 @@ func Int(value string, def int) int {
 	return i
 }
 
+// Int64 returns the int64 value associated with the given value, or the value
+// "def", if the value is blank.
+//
+// To convert from a the string value attached to a given key,
```
chrisd8088: Although there are similar typos in the comments above, we should at least avoid adding another one, so I'd suggest using the new method's name in the comment:

```suggestion
// To convert from the string value attached to a given key,
```

The first of these typos was introduced back in commit 18cb9257f90f2d7b122c4e0bf087ce0ec72cccb0 from PR  #1419, and we've inadvertently retained and duplicated it through various PRs since then.

lfshttp/client.go
```diff
@@ -26,8 +26,10 @@ import (
 	"golang.org/x/net/http2"
 )
 
-const MediaType = "application/vnd.git-lfs+json"
-const RequestContentType = MediaType + "; charset=utf-8"
+const (
+	MediaType          = "application/vnd.git-lfs+json"
+	RequestContentType = MediaType + "; charset=utf-8"
+)
```
chrisd8088: This is a nice bit of code cleanup, thank you!

As a general rule we typically make such changes in a separate commit (or even PR) from functional changes, which helps clarify exactly which alterations are required to implement the new functionality, as opposed to those which are optional improvements.

Would you mind moving this change into a distinct commit from the one with the new `int64` support?

lfshttp/client.go
```diff
@@ -626,14 +628,19 @@ func (e testEnv) Int(key string, def int) int {
...
+
 func (e testEnv) Bool(key string, def bool) bool {
 	s, _ := e.Get(key)
 	return config.Bool(s, def)
 }
 
 func (e testEnv) All() map[string][]string {
 	m := make(map[string][]string)
-	for k, _ := range e {
+	for k := range e {
```
chrisd8088: This is a nice bit of code cleanup, thank you!

As a general rule we typically make such changes in a separate commit (or even PR) from functional changes, which helps clarify exactly which alterations are required to implement the new functionality, as opposed to those which are optional improvements.

Would you mind moving this change into a distinct commit from the one with the new `int64` support?

aymanbagabas:
> I made a few minor suggestions, but the core functional changes all look great. Thanks again for this improvement!

I've updated the PR with the suggested changes 🙂

PR #6204, opened 2026-02-09
Fix extra header duplication on retry

### LINKED ISSUE #6202: ExtraHeadersFor duplicates http.extraHeader on auth retry, causing infinite 401 loop
**Describe the bug**

`ExtraHeadersFor()` duplicates `http.<url>.extraHeader` values on every authentication retry, producing malformed headers that cause an infinite 401 loop.

When a custom authentication header is configured via `http.extraHeader` and `lfs.access` is not set to `basic`, Git LFS retries authentication using the same `*http.Request` object. Each retry calls `ExtraHeadersFor()`, which copies all existing headers (including previously appended extra headers) and appends the extra headers again. After two iterations the header value is already malformed, causing every subsequent request to fail with 401 — which triggers another retry, ad infinitum.

**To Reproduce**

1. Configure a custom auth header:
   ```bash
   git config --global "http.https://git.example.com.extraHeader" "Authorization: bearer <token>"
   ```

2. Clone a repo with LFS content, **without** `lfs.access` set to `basic`:
   ```bash
   git clone https://git.example.com/org/repo.git
   ```

3. Git LFS prompts for credentials. Entering valid credentials leads to an infinite loop of password prompts.

**Expected behavior**

Extra headers should be applied exactly once per request, regardless of how many auth retries occur. Valid credentials should be accepted on the retry.

**Root cause**

Two issues interact:

**1. Header duplication in `ExtraHeadersFor`** ([`lfshttp/client.go:237–254`](https://github.com/git-lfs/git-lfs/blob/main/lfshttp/client.go#L237-L254))

```go
func (c *Client) ExtraHeadersFor(req *http.Request) http.Header {
    extraHeaders := c.extraHeaders(req.URL)
    // ...
    copy := make(http.Header, len(req.Header))
    for k, vs := range req.Header {
        copy[k] = vs                        // copies ALL headers, including previous extra headers
    }
    for k, vs := range extraHeaders {
        for _, v := range vs {
            copy[k] = append(copy[k], v)    // appends extra headers AGAIN → duplicates
        }
    }
    return copy
}
```

This is called on every auth retry via `doWithAuth` ([`lfsapi/auth.go:68`](https://github.com/git-lfs/git-lfs/blob/main/lfsapi/auth.go#L68)):

```go
req.Header = c.client.ExtraHeadersFor(req)  // same req on every iteration
```

Result per retry:
```
Attempt 1: Authorization: bearer <token>
Attempt 2: Authorization: bearer <token>, bearer <token>           ← malformed
Attempt 3: Authorization: bearer <token>, bearer <token>, bearer …  ← malformed
```

**2. No recursion depth limit in `DoWithAuth`** ([`lfsapi/auth.go:26–43`](https://github.com/git-lfs/git-lfs/blob/main/lfsapi/auth.go#L26-L43))

```go
func (c *Client) DoWithAuth(remote string, access creds.Access, req *http.Request) (*http.Response, error) {
    count := 0  // fresh counter on every recursive call
    res, err := c.doWithAuth(remote, &count, access, req, nil)

    if errors.IsAuthError(err) {
        if len(req.Header.Get("Authorization")) == 0 {
            // ...
            return c.DoWithAuth(remote, newAccess, req)  // recursive, bypasses defaultMaxAuthAttempts
        }
    }
    return res, err
}
```

Each recursive call to `DoWithAuth` resets `count` to 0, so `defaultMaxAuthAttempts` is never reached.

**Proposed fix**

Replace `append` with assignment in `ExtraHeadersFor` so extra headers are set idempotently:

```go
for k, vs := range extraHeaders {
    copy[k] = vs  // replace instead of append — idempotent across retries
}
```

This ensures extra headers are applied exactly once regardless of how many times the function is called with the same request.

**Workaround**

Set `lfs.access` to `basic` so credentials are included on the first request, avoiding the retry path:

```bash
git config --global lfs.access basic
```

**System environment**

Not environment-specific. The bug is in the auth retry logic and affects all platforms. Verified against current `main` (commit `bb65882`).

**Output of `git lfs env`**

N/A — the bug is reproducible with any environment that uses `http.extraHeader` for authentication without `lfs.access=basic`.

ypfaff:
Fixes the header duplication issue reported in #6202 by adding a deduplication check in `ExtraHeadersFor` before appending headers. This prevents values like `Authorization: Bearer x, Bearer x` when the same `*http.Request` is reused during auth retries.

Rebased on top of #6018.

Fixes #6202


chrisd8088:
Thanks very much for this PR!

If you don't mind, I'll finish PR #6018 first and get it merged, as it should resolve the HTTP recursion issue you also reported, and I'm going to introduce a suite of new shell and Go tests to demonstrate the problem is fixed.

Once that's merged, we can then look at what other fixes are still required to deal with any remaining problems related to the `http.extraHeader` configuration option.

chrisd8088:
Just FYI, I updated PR #6018 and am now just waiting for a second review before merging it.

If you feel like rebasing and refactoring this PR on top of PR #6018's branch, please go ahead, and as you should then be able to focus this PR on just the header duplication issue.

ypfaff:
Rebased on top of #6018 and dropped the auth retry loop fix, keeping only the header deduplication fix in `ExtraHeadersFor`. Thanks for the heads-up! 🙂

lfshttp/client.go
```diff
@@ -247,7 +247,16 @@ func (c *Client) ExtraHeadersFor(req *http.Request) http.Header {
...
+			exists := false
+			for _, existing := range copy[k] {
+				if existing == v {
+					exists = true
+					break
+				}
+			}
+			if !exists {
+				copy[k] = append(copy[k], v)
+			}
```
chrisd8088: This looks good, but I think we could write it more compactly with the help of the `slices` package that was added to the Go standard library in v1.21.

If we insert `slices` into our alphabetical [list](https://github.com/git-lfs/git-lfs/blob/0a45b99f9fa8d1942cc44f089029a031805461f1/lfshttp/client.go#L4-L18) of the standard library packages we import, we can then just do:

```suggestion
			if !slices.Contains(copy[k], v) {
				copy[k] = append(copy[k], v)
			}
```

lfshttp/client_test.go
```diff
@@ -409,3 +409,30 @@ func TestHttpVersion(t *testing.T) {
...
+	req.Header = c.ExtraHeadersFor(req)
+	t.Logf("Headers after 1st call: %v", req.Header)
+
+	assert.Equal(t, []string{extraHeaderValue}, req.Header[extraHeaderKey], "Header should be present once after first call")
+
+	// Simulate retry with same request object
+	req.Header = c.ExtraHeadersFor(req)
+
+	// Ensure headers are not duplicated on retry (regression test for #6202)
+	assert.Equal(t, []string{extraHeaderValue}, req.Header[extraHeaderKey], "Header should still be present only once after second call")
```
chrisd8088, line 453: This is a great test, thank you!

It occurs to me that we could further expand this test to also check that we handle multi-valued headers correctly too.

For instance, I think something like the following would help verify that we're checking all the values of a single header for any duplicates and only appending distinct ones:

```suggestion
	extraHeader2Key := "Cache-Control"
	extraHeader2Value1 := "no-cache"
	extraHeader2Value2 := "no-store"

	context := &testContext{
		gitConfig: git.NewConfig("", ""),
		osEnv:     make(testEnv),
		gitEnv: config.EnvironmentOf(config.MapFetcher(map[string][]string{
			"http.extraheader": []string{
				fmt.Sprintf("%s: %s", extraHeaderKey, extraHeaderValue),
				fmt.Sprintf("%s: %s", extraHeader2Key, extraHeader2Value1),
				fmt.Sprintf("%s: %s", extraHeader2Key, extraHeader2Value2),
			},
		})),
	}

	c, err := NewClient(context)
	require.Nil(t, err)

	t.Logf("Configured extraHeader: %v", c.GitEnv().GetAll("http.extraHeader"))

	req, err := http.NewRequest("GET", targetURL, nil)
	require.Nil(t, err)

	req.Header = c.ExtraHeadersFor(req)
	t.Logf("Headers after 1st call: %v", req.Header)

	assert.Equal(t, []string{extraHeaderValue}, req.Header[extraHeaderKey], "Header should be present once after first call")
	assert.Equal(t, []string{extraHeader2Value1, extraHeader2Value2}, req.Header[extraHeader2Key], "Header with multiple values should be present twice after first call")

	// Simulate retry with same request object
	req.Header = c.ExtraHeadersFor(req)

	// Ensure headers are not duplicated on retry (regression test for #6202)
	assert.Equal(t, []string{extraHeaderValue}, req.Header[extraHeaderKey], "Header should still be present only once after second call")
	assert.Equal(t, []string{extraHeader2Value1, extraHeader2Value2}, req.Header[extraHeader2Key], "Header with multiple values should still be present only twice after first call")
```

Note that we have to work around the limitation that our `testEnv` [type](https://github.com/git-lfs/git-lfs/blob/0a45b99f9fa8d1942cc44f089029a031805461f1/lfshttp/client.go#L610) in the `lfshttp` package doesn't support multi-valued headers, but we can use the `mapFetcher` [type](https://github.com/git-lfs/git-lfs/blob/0a45b99f9fa8d1942cc44f089029a031805461f1/config/map_fetcher.go#L5) from our `config` package for that purpose instead.  (I think trying to enhance the `testEnv` type to do the same thing is outside the scope of this PR.)

chrisd8088 (commented):
Thanks again for this PR, and for your patience while we get PR #6018 finalized!

I looked at the last commit in this PR and it looks very good!  Thanks especially for writing a test.  I made a couple of suggestions only, so please take a look whenever you get a chance.

Thanks again very much for your help!

lfshttp/client.go
ypfaff: Good call, replaced it with `slices.Contains`. 🙂

lfshttp/client_test.go
ypfaff: Great suggestion, expanded the test with `config.MapFetcher` as you described. Thanks for the hint on `testEnv`'s limitation 😊

PR #6207, opened 2026-02-26
Fix and refactor HTTP encoding and retry tests

chrisd8088:
This PR corrects, updates, and reorganizes our shell tests of Batch API and object transfer HTTP request retries and our shell tests of gzip-encoding during object transfer HTTP download requests.

These changes should make the relationships between these tests more apparent, and should provide a framework in which to add similar tests in the future, for example in PR #6196.

#### Test Reorganization

We move the `storage upload with compression` [test](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/t-push.sh#L958-L998) from the `t/t-push.sh` script into a new `t/t-batch-storage-encoding.sh` script, and rename the test to `batch storage HTTP download with gzip encoding` since it only tests the use of gzip compression during HTTP downloads, not uploads.

Next, we move the tests from the `t/t-batch-retries.sh` [script](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/t-batch-retries.sh) and the `t/t-resume-http-range.sh` [script](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/t-resume-http-range.sh) into a single new `t/t-batch-storage-retries.sh` script, and rename these tests and the tests in both the `t/t-batch-retries-ratelimit.sh` [script](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/t-batch-retries-ratelimit.sh) and the `t/t-batch-storage-retries-ratelimit.sh` [script](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/t-batch-storage-retries-ratelimit.sh) to all follow a common pattern.

Finally, we move the tests from the `t/t-resume-tus.sh` [script](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/t-resume-tus.sh) to a new `t/t-batch-storage-upload-tus.sh` script, and rename the tests to align with the test names we've introduced in our other changes.

#### Test Fixes and Updates

We correct the `Content-Range` [headers](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/cmd/lfstest-gitserver.go#L831) [returned](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/cmd/lfstest-gitserver.go#L877) to several tests by our `lfstest-gitserver` utility so they read, for instance, `bytes 10-27/28` instead of `bytes 10-28/-18`, as they do now.

We revise the `lfstest-gitserver` utility so that when it [returns](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/cmd/lfstest-gitserver.go#L890-L891) truncated object data to simulate an interrupted HTTP object download, it explicitly sets the `Content-Length` header to the full object's length.  Otherwise, as occurs now, the `net/http` package from the Go standard library automatically inserts the `Content-Length` header and uses the length of the truncated data, which fails to provide a true simulation of an interrupted response.  The result is that the Git LFS client understands the transfer to be complete, but assumes that the object must be corrupted on the server because the SHA-256 hash of the downloaded data does not match the OID.  For this reason, [several](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/t-resume-http-range.sh#L31-L35) of our [tests](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/t-resume-http-range.sh#L72-L76) expect the Git LFS client to report an error due to an interrupted download, when in fact the client should just retry the request immediately with a `Range` header.

At present, the `resume-http-range-retry` test (which we also move and rename, per the notes above) passes trivially in all cases.  The test [creates](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/t-resume-http-range.sh#L120) a file with a `.tmp` extension in the `.git/lfs/incomplete` directory to try to simulate an interrupted object transfer download.  However, since PR #3826 the files the client creates in that directory have the extension `.part`, not `.tmp`.  As a consequence, the client ignores the file with the `.tmp` extension and simply makes a fresh, complete download of the object.  We therefore update the file extension of the temporary file, and also expand the test so it will guard against similar regressions in the future.

We remove a misleading [comment](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/t-resume-tus.sh#L52-L53) from the `tus-upload-interrupted-resume` test (which we also move and rename, per the notes above) and simplify the string [used](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/t-resume-tus.sh#L54) to create an object since the test does not actually depend on the object's contents at all.

We refactor the [two](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/cmd/lfstest-gitserver.go#L825-L832) [blocks](https://github.com/git-lfs/git-lfs/blob/f49a7d70485d94b187d7b373cdc973d11874878c/t/cmd/lfstest-gitserver.go#L850-L879) in the `lfstest-gitserver` utility where we parse `Range` request headers into a common `handleRangeRequest()` function, which we be able to reuse in subsequent PRs when we need to add new tests that require the utility to validate `Range` headers.

In addition, we try to refine many of the tests we are otherwise moving, renaming, or correcting in this PR.  For instance, we revise the internal code comments in a number of the tests.  We also expand the sets of messages and HTTP headers the tests look for in Git LFS command trace logs, add checks of commands' exit codes where appropriate, eliminate unnecessary log files, and in general try to make the tests as consistent with each other as possible.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
-        go: ['1.25.x']
+        go: ['1.26.x']
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v6
@@ -45,7 +45,7 @@ jobs:
     name: Build with specific Go
     strategy:
       matrix:
-        go: ['1.24.x']
+        go: ['1.25.x']
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v6
@@ -69,7 +69,7 @@ jobs:
     - run: gem install asciidoctor
     - uses: actions/setup-go@v6
       with:
-        go-version: '1.25.x'
+        go-version: '1.26.x'
     - run: mkdir -p "$HOME/go/bin"
       shell: bash
     - run: set GOPATH=%HOME%\go
@@ -129,7 +129,7 @@ jobs:
         ref: ${{ github.ref }}
     - uses: actions/setup-go@v6
       with:
-        go-version: '1.25.x'
+        go-version: '1.26.x'
     - run: git clone -b master https://github.com/git/git.git "$HOME/git"
     - run: |
         echo "GIT_INSTALL_DIR=$HOME/git" >> "$GITHUB_ENV"
@@ -153,7 +153,7 @@ jobs:
         ref: ${{ github.ref }}
     - uses: actions/setup-go@v6
       with:
-        go-version: '1.25.x'
+        go-version: '1.26.x'
     - run: git clone -b v2.0.0 https://github.com/git/git.git "$HOME/git"
     - run: |
         echo "GIT_INSTALL_DIR=$HOME/git" >> "$GITHUB_ENV"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: windows-latest
     strategy:
       matrix:
-        go: ['1.25.x']
+        go: ['1.26.x']
     steps:
     - uses: actions/checkout@v6
       with:
@@ -58,13 +58,16 @@ jobs:
       shell: bash
       env:
         FORCE_LOCALIZE: true
-    - uses: azure/trusted-signing-action@v1.1.0
+    - uses: azure/artifact-signing-action@v1.1.0
       with:
         azure-tenant-id: ${{ secrets.SPN_GIT_LFS_SIGNING_TENANT_ID }}
         azure-client-id: ${{ secrets.SPN_GIT_LFS_SIGNING_CLIENT_ID }}
         azure-client-secret: ${{ secrets.SPN_GIT_LFS_SIGNING }}
-        endpoint: https://wus.codesigning.azure.net/
-        trusted-signing-account-name: GitHubInc
+        # Note that the endpoint URI must match the Azure region where our
+        # Signing Account and Certificate Profile were created:
+        # https://github.com/Azure/artifact-signing-action/blob/v1.1.0/README.md?plain=1#L138-L139
+        endpoint: https://wus3.codesigning.azure.net/
+        signing-account-name: GitHubInc
         certificate-profile-name: GitHubInc
         files-folder: ${{ github.workspace }}/tmp/stage1
         files-folder-filter: exe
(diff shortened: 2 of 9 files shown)


t/t-batch-storage-retries.sh
```diff
@@ -90,30 +90,38 @@ begin_test "batch storage HTTP download retries with Range header"
...
+  git commit -m "add a.dat"
   git push origin main
 
   assert_server_object "$reponame" "$contents_oid"
 
-  # delete local copy then fetch it back
-  # server will abort the transfer mid way (so will error) when not resuming
-  # then we can restart it
+  # Test object transfer download with an interrupted initial response,
+  # after which the client should fetch the remaining bytes using a request
```
larsxschneider, line 108: Maybe add this to the comment?
```
# Test object transfer download with an interrupted initial response (truncated after 10 bytes).
```

Maybe set and use a variable like this to make the "magic" `10` below a bit more clear?
```
truncated_content=10
```

t/t-batch-storage-retries.sh
```diff
@@ -0,0 +1,266 @@
...
+  reponame="batch-storage-download-retry-range"
+  setup_remote_repo "$reponame"
+
+  clone_repo "$reponame" $reponame
+
+  git lfs track "*.dat" 2>&1 | tee track.log
+  grep "Tracking \"\*.dat\"" track.log
+
+ # This content announces to the server that it should interrupt the
+ # object download unless a Range header with a positive offset was sent.
```
larsxschneider: nit nit: identation

t/cmd/lfstest-gitserver.go
larsxschneider: nit: should that be renamed to `storageDownloadRetryRangeRejectedAttempts`?

t/t-batch-storage-retries.sh
```diff
@@ -0,0 +1,266 @@
...
+end_test
+
+begin_test "batch storage HTTP download retries with Range header"
+(
+  set -e
+
+  reponame="batch-storage-download-retry-range"
+  setup_remote_repo "$reponame"
+
+  clone_repo "$reponame" $reponame
```
larsxschneider: nit: Sometimes we quote $reponame and sometimes we don't. I guess we don't need the quotes anywhere but maybe we want to be consistent?

chrisd8088: That's a good point.  This particular discrepancy appears to date back to the original versions of these tests added in commit c9c457c2706a15948e371fb02c82ef813613876a of PR #1297, but we might as well tidy it up along with all the other changes in this PR.

t/cmd/lfstest-gitserver.go
chrisd8088: Great idea, thank you!  Done.

t/t-batch-storage-retries.sh
chrisd8088: Good catch; I've fixed that.

t/t-batch-storage-retries.sh
chrisd8088: Your suggestions prompted me to reconsider these hard-coded offsets entirely.

The original tests from PR #1297 used fixed offsets in the test server utility, so I'd just carried those over into the tests themselves.

However, if we change the test server to simply return the first half of an object in these cases, then the tests can calculate the same midpoint programmatically without any need for magic values, so that's what I opted to do.

PR #6211, opened 2026-03-02
Various gopls lint cleanups for package `git`

bk2204:
I was in the `git` package and noticed that my editor was providing me with some linting suggestions, which apparently come from gopls, the Go LSP.  I've fixed the two major issues that I found there, which can be listed with `gopls check --severity=hint ./git/*.go`.

I did not fix every issue, however.  I skipped removing the unused functions which implement a complete interface and I only chose to fix items that were at least several Go releases old (I think the latest version required is Go 1.21) in order to reduce the burden on downstream maintainers.

### DIFF
--- a/git/git.go
+++ b/git/git.go
@@ -1364,15 +1364,14 @@ func CachedRemoteRefs(remoteName string) ([]*Ref, error) {
 
 func parseShowRefLine(refPrefix, line string) (sha, name string, ok bool) {
 	// line format: <sha> <space> <ref>
-	space := strings.IndexByte(line, ' ')
-	if space < 0 {
+	sha, ref, found := strings.Cut(line, " ")
+	if !found {
 		return "", "", false
 	}
-	ref := line[space+1:]
 	if !strings.HasPrefix(ref, refPrefix) {
 		return "", "", false
 	}
-	return line[:space], strings.TrimSpace(ref[len(refPrefix):]), true
+	return sha, strings.TrimSpace(ref[len(refPrefix):]), true
 }
 
 // Fetch performs a fetch with no arguments against the given remotes.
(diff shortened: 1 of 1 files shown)


git/git.go
```diff
@@ -1364,15 +1364,14 @@ func CachedRemoteRefs(remoteName string) ([]*Ref, error) {
 
 func parseShowRefLine(refPrefix, line string) (sha, name string, ok bool) {
 	// line format: <sha> <space> <ref>
-	space := strings.IndexByte(line, ' ')
-	if space < 0 {
+	sha, ref, ok := strings.Cut(line, " ")
+	if !ok {
```
chrisd8088: Minor request: would you mind renaming `ok` to `found`, which would match the terminology from the `Cut()` function's [documention](https://pkg.go.dev/strings#Cut)?

Since the final return value doesn't really indicate whether an error occurred or not, which is the context in which we often use `ok`, I think `found` makes a bit more sense here.

```suggestion
	sha, ref, found := strings.Cut(line, " ")
	if !found {
```

git/git.go
```diff
@@ -1438,11 +1437,10 @@ func parseLsRemoteLine(line string) (sha, ns, name string, ok bool) {
 	const tagPrefix = "refs/tags/"
 
 	// line format: <sha> <tab> <ref>
-	tab := strings.IndexByte(line, '\t')
-	if tab < 0 {
+	sha, ref, ok := strings.Cut(line, "\t")
+	if !ok {
```
chrisd8088: As above, just one minor request to substitute `ok` with `found` here too.

```suggestion
	sha, ref, found := strings.Cut(line, "\t")
	if !found {
```

git/git.go
bk2204: Sure, I can do that.  I'll try to re-roll a little later today.

bk2204: Done in both places.

PR #6212, opened 2026-03-03
Use `core.ignorecase` to determine case sensitivity

bk2204:
Right now, we always determine the case sensitivity of the repository based on the operating system default: case insensitive on Windows and macOS and case sensitive on other Unix systems.  This works in most cases, but it is possible to have case sensitive file systems on macOS, case sensitive directories on Windows, and case insensitive file systems (e.g., JFS) on Linux, in which case we do not function correctly.

Add a function which uses the `core.ignorecase` setting in Git to properly detect whether our repository is case sensitive or not.

### DIFF
--- a/filepathfilter/filepathfilter.go
+++ b/filepathfilter/filepathfilter.go
@@ -83,10 +83,10 @@ func NewFromPatterns(include, exclude []Pattern, setters ...Option) *Filter {
 	return f
 }
 
-func New(include, exclude []string, ptype PatternType, setters ...Option) *Filter {
+func New(include, exclude []string, ptype PatternType, gitEnv Environment, setters ...Option) *Filter {
 	return NewFromPatterns(
-		convertToWildmatch(include, ptype),
-		convertToWildmatch(exclude, ptype), setters...)
+		convertToWildmatch(include, ptype, gitEnv),
+		convertToWildmatch(exclude, ptype, gitEnv), setters...)
 }
 
 // Include returns the result of calling String() on each Pattern in the
@@ -184,7 +184,7 @@ const (
 	sep byte = '/'
 )
 
-func NewPattern(p string, ptype PatternType) Pattern {
+func NewPattern(p string, ptype PatternType, gitEnv Environment) Pattern {
 	tracerx.Printf("filepathfilter: creating pattern %q of type %v", p, ptype)
 
 	switch ptype {
@@ -193,7 +193,7 @@ func NewPattern(p string, ptype PatternType) Pattern {
 			p: p,
 			w: wildmatch.NewWildmatch(
 				p,
-				wildmatch.SystemCase,
+				caseFromConfig(gitEnv),
 				wildmatch.Contents,
 			),
 		}
@@ -202,7 +202,7 @@ func NewPattern(p string, ptype PatternType) Pattern {
 			p: p,
 			w: wildmatch.NewWildmatch(
 				p,
-				wildmatch.SystemCase,
+				caseFromConfig(gitEnv),
 				wildmatch.Basename,
 				wildmatch.GitAttributes,
 			),
@@ -228,10 +228,10 @@ func join(paths ...string) string {
 	return joined
 }
 
-func convertToWildmatch(rawpatterns []string, ptype PatternType) []Pattern {
+func convertToWildmatch(rawpatterns []string, ptype PatternType, gitEnv Environment) []Pattern {
 	patterns := make([]Pattern, len(rawpatterns))
 	for i, raw := range rawpatterns {
-		patterns[i] = NewPattern(raw, ptype)
+		patterns[i] = NewPattern(raw, ptype, gitEnv)
 	}
 	return patterns
 }
--- a/filepathfilter/filepathfilter_test.go
+++ b/filepathfilter/filepathfilter_test.go
@@ -107,28 +107,28 @@ func TestPatternMatch(t *testing.T) {
 }
 
 func assertPatternMatch(t *testing.T, pattern string, filenames ...string) {
-	p := NewPattern(pattern, GitAttributes)
+	p := NewPattern(pattern, GitAttributes, nil)
 	for _, filename := range filenames {
 		assert.True(t, p.Match(filename), "%q should match pattern %q", filename, pattern)
 	}
 }
 
 func assertPatternMatchIgnore(t *testing.T, pattern string, filenames ...string) {
-	p := NewPattern(pattern, GitIgnore)
+	p := NewPattern(pattern, GitIgnore, nil)
 	for _, filename := range filenames {
 		assert.True(t, p.Match(filename), "%q should match pattern %q", filename, pattern)
 	}
 }
 
 func refutePatternMatch(t *testing.T, pattern string, filenames ...string) {
-	p := NewPattern(pattern, GitAttributes)
+	p := NewPattern(pattern, GitAttributes, nil)
 	for _, filename := range filenames {
 		assert.False(t, p.Match(filename), "%q should not match pattern %q", filename, pattern)
 	}
 }
 
 func refutePatternMatchIgnore(t *testing.T, pattern string, filenames ...string) {
-	p := NewPattern(pattern, GitIgnore)
+	p := NewPattern(pattern, GitIgnore, nil)
 	for _, filename := range filenames {
 		assert.False(t, p.Match(filename), "%q should not match pattern %q", filename, pattern)
 	}
@@ -142,13 +142,13 @@ type filterTest struct {
 }
 
 func TestFilterReportsIncludePatterns(t *testing.T) {
-	filter := New([]string{"*.foo", "*.bar"}, nil, GitAttributes)
+	filter := New([]string{"*.foo", "*.bar"}, nil, GitAttributes, nil)
 
 	assert.Equal(t, []string{"*.foo", "*.bar"}, filter.Include())
 }
 
 func TestFilterReportsExcludePatterns(t *testing.T) {
-	filter := New(nil, []string{"*.baz", "*.quux"}, GitAttributes)
+	filter := New(nil, []string{"*.baz", "*.quux"}, GitAttributes, nil)
 
 	assert.Equal(t, []string{"*.baz", "*.quux"}, filter.Exclude())
 }
--- a/git/gitattr/files.go
+++ b/git/gitattr/files.go
@@ -219,14 +219,14 @@ func AttrPathsFromReader(mp *MacroProcessor, fpath, workingDir string, rdr io.Re
 // file paths can be matched against
 // workingDir is the root of the working copy
 // gitDir is the root of the git repo
-func GetAttributeFilter(workingDir, gitDir string) *filepathfilter.Filter {
+func GetAttributeFilter(gitEnv filepathfilter.Environment, workingDir, gitDir string) *filepathfilter.Filter {
 	paths := GetAttributePaths(NewMacroProcessor(), workingDir, gitDir)
 	patterns := make([]filepathfilter.Pattern, 0, len(paths))
 
 	for _, path := range paths {
 		// Convert all separators to `/` before creating a pattern to
 		// avoid characters being escaped in situations like `subtree\*.md`
-		patterns = append(patterns, filepathfilter.NewPattern(filepath.ToSlash(path.Path), filepathfilter.GitAttributes))
+		patterns = append(patterns, filepathfilter.NewPattern(filepath.ToSlash(path.Path), filepathfilter.GitAttributes, gitEnv))
 	}
 
 	return filepathfilter.NewFromPatterns(patterns, nil)
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -85,7 +85,7 @@ func checkoutCommand(cmd *cobra.Command, args []string) {
 		pointers = append(pointers, p)
 	})
 
-	chgitscanner.Filter = filepathfilter.New(rootedPathPatterns, nil, filepathfilter.GitIgnore)
+	chgitscanner.Filter = filepathfilter.New(rootedPathPatterns, nil, filepathfilter.GitIgnore, cfg.Git)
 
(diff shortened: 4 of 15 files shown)


tools/filetools.go
```diff
@@ -256,8 +256,8 @@ type FastWalkCallback func(parentDir string, info os.FileInfo, err error)
 //   - Automatically ignores any .git directories
 //
 // rootDir - Absolute path to the top of the repository working directory
-func FastWalkDir(rootDir string, cb FastWalkCallback) {
-	fastWalkCallback(fastWalkWithExcludeFiles(rootDir), cb)
+func FastWalkDir(rootDir string, cfg filepathfilter.Environment, cb FastWalkCallback) {
```
chrisd8088: As a quick first thought only, I suspect we can avoid the need to alter this function and the `tools` package, and also avoid altering anything our `fs` package, with a bit of advance cleanup.

The `FastWalkDir()` function is called in a [couple](https://github.com/git-lfs/git-lfs/blob/f4ebf5fecd848f7a8e6e8f2e4c5cc5d775870bc5/fs/fs.go#L53) of [instances](https://github.com/git-lfs/git-lfs/blob/f4ebf5fecd848f7a8e6e8f2e4c5cc5d775870bc5/fs/cleanup.go#L28), in the `EachObject()` and `cleanupTmp()` methods.  Within the `FastWalkDir()` function itself, we [call](https://github.com/git-lfs/git-lfs/blob/f4ebf5fecd848f7a8e6e8f2e4c5cc5d775870bc5/tools/filetools.go#L260) `fastWalkWithExcludeFiles()`, which [establishes](https://github.com/git-lfs/git-lfs/blob/f4ebf5fecd848f7a8e6e8f2e4c5cc5d775870bc5/tools/filetools.go#L293-L296) some file path filters to [exclude](https://github.com/git-lfs/git-lfs/blob/f4ebf5fecd848f7a8e6e8f2e4c5cc5d775870bc5/tools/filetools.go#L353-L355) `.git` directories.

The `cleanupTmp()` method uses the `FastWalkDir()` function to traverse the `.git/lfs/tmp` directory looking for stale files to delete.  This directory is always within the `.git` directory, there's no value to trying to avoid `.git` directories while traversing its contents.

The `EachObject()` method, meanwhile, uses the `FastWalkDir()` function to traverse the `.git/lfs/objects` directory looking for object files on which to [invoke](https://github.com/git-lfs/git-lfs/blob/f4ebf5fecd848f7a8e6e8f2e4c5cc5d775870bc5/fs/fs.go#L62) a callback function.  Again, since this directory is always within the `.git` directory, there's no need to try to exclude `.git` directories while performing the traversal.

The fact that the `FastWalkDir()` function bothers to try to exclude `.git` directories at all is an artifact of its original use case, which was to walk through the whole Git working tree.  We added the precursor of the current `FastWalkDir()` function in commit 13a8af6dfe711fe4e25e0d9e721f404221735610 of PR #1616, at which time the function was named `FastWalkGitRepo()`.  It was initially used in the `git lfs track` command and offered a performance improvement over the `Walk()` [function](https://pkg.go.dev/path/filepath#Walk) provided by the Go standard library's `filepath` package.

Subsequent PRs gradually added calls to the `FastWalkGitRepo()` implementation, until it was largely replaced in commit 83d7f76e763e819793afdb9132b6df4ec58f0946 of PR #3823 with the `NewLsFiles()` function in our `git` package, which executes a `git ls-files` command and processes its output.  The `FastWalkGitRepo()` function was renamed to `FastWalkDir()`, and left in place to handle the two use cases where we need to traverse a directory hierarchy within the `.git` directory itself.  However, some of the original design remains in place, although it's no longer useful.

When I was first investigating this development history a while ago, I realized it explained something I'd noticed but never fully understood about the Git LFS client's trace logs, which is that they typically end with lines like the following:
```
18:21:21.340497 trace git-lfs: filepathfilter: creating pattern ".git" of type gitignore
18:21:21.340532 trace git-lfs: filepathfilter: creating pattern "**/.git" of type gitignore
18:21:21.340562 trace git-lfs: filepathfilter: accepting "tmp"
```

This happens because every Git LFS command ends by [calling](https://github.com/git-lfs/git-lfs/blob/f4ebf5fecd848f7a8e6e8f2e4c5cc5d775870bc5/git-lfs.go#L35) the `Cleanup()` function in our `commands` package, which [indirectly](https://github.com/git-lfs/git-lfs/blob/f4ebf5fecd848f7a8e6e8f2e4c5cc5d775870bc5/commands/commands.go#L276) [invokes](https://github.com/git-lfs/git-lfs/blob/f4ebf5fecd848f7a8e6e8f2e4c5cc5d775870bc5/config/config.go#L460) the `Cleanup()` function in our `fs` package, which [runs](https://github.com/git-lfs/git-lfs/blob/f4ebf5fecd848f7a8e6e8f2e4c5cc5d775870bc5/config/config.go#L460) the `cleanupTmp()` function.

At any rate, I think we could maybe start to shrink the scope of this PR if we first eliminated some of these vestigial uses of filepath filtering.

chrisd8088: Whoops ... I forgot this review comment wouldn't appear until I hit the "Submit review" button!  My apologies, since I wrote this a while ago.

Your PR prompted me to tackle some long-pending cleanup in PR #6223.  If we merge that first, I think it will make this set of changes somewhat simpler.  (Or if you'd prefer to rebase this PR on that PR's branch, by all means do so.)

fs/fs.go
```diff
@@ -29,6 +29,7 @@ var (
 // For more information, see config/environment.go.
 type Environment interface {
 	Get(key string) (val string, ok bool)
+	Bool(key string, def bool) (val bool)
```
chrisd8088: If we rebase this PR's changes on my PR #6223 (or merge that PR's changes first) then I think we can skip this change, as we won't need to modify the `config`, `fs`, or `tools` packages at all.

filepathfilter/filepathfilter.go
```diff
@@ -235,3 +235,16 @@ func convertToWildmatch(rawpatterns []string, ptype PatternType) []Pattern {
...
 	return patterns
 }
+
+type Environment interface {
+	Get(key string) (val string, ok bool)
+	Bool(key string, def bool) (val bool)
+}
+
+func caseFromConfig(cfg Environment) func(w *wildmatch.Wildmatch) {
+	if cfg.Bool("core.ignorecase", false) {
```
chrisd8088: I think we can perhaps avoid the need for a `FakeEnvironment` structure and related methods if we allow our test functions to just pass a `nil` for the `Environment` interface parameter, and then do something here like:

```go
	if cfg != nil && cfg.Bool("core.ignorecase", false) {
		return wildmatch.CaseFold
	} else {
		return func(w *wildmatch.Wildmatch) {}
	}
```

I'd also be inclined to name the parameter `gitEnv` rather than `cfg`, partly for additional clarity and partly because that follows the pattern already established in other packages that also have a local, minimal `Environment` interface.  For instance, the `ObjectDatabase()` [function](https://github.com/git-lfs/git-lfs/blob/0a45b99f9fa8d1942cc44f089029a031805461f1/git/git.go#L1664) in our `git` package takes `gitEnv` and `osEnv` parameters which both conform to the `Environment` [interface](https://github.com/git-lfs/git-lfs/blob/0a45b99f9fa8d1942cc44f089029a031805461f1/git/config.go#L19-L24) from that package.

So my suggestion here would be:

```suggestion
func caseFromConfig(gitEnv Environment) func(w *wildmatch.Wildmatch) {
	if gitEnv != nil && gitEnv.Bool("core.ignorecase", false) {
```

filepathfilter/filepathfilter.go
```diff
@@ -83,10 +83,10 @@ func NewFromPatterns(include, exclude []Pattern, setters ...Option) *Filter {
 	return f
 }
 
-func New(include, exclude []string, ptype PatternType, setters ...Option) *Filter {
+func New(include, exclude []string, ptype PatternType, cfg Environment, setters ...Option) *Filter {
```
chrisd8088: As I noted in another comment, I'd suggest we use the parameter name `gitEnv` in place of `cfg` here and elsewhere, just to match up with some of our [existing](https://github.com/git-lfs/git-lfs/blob/0a45b99f9fa8d1942cc44f089029a031805461f1/git/gitattr/tree.go#L92) [usage](https://github.com/git-lfs/git-lfs/blob/0a45b99f9fa8d1942cc44f089029a031805461f1/git/git.go#L1664) in other places.

filepathfilter/filepathfilter.go
```diff
@@ -228,14 +228,28 @@ func join(paths ...string) string {
...
+func convertToWildmatch(rawpatterns []string, ptype PatternType, cfg Environment) []Pattern {
 	patterns := make([]Pattern, len(rawpatterns))
 	for i, raw := range rawpatterns {
-		patterns[i] = NewPattern(raw, ptype)
+		patterns[i] = NewPattern(raw, ptype, cfg)
 	}
 	return patterns
 }
 
+type FakeEnvironment struct{}
```
chrisd8088: As I noted in another comment, I think we can probably avoid the need for this test-only structure and its methods to be defined in a non-test source file if we check for a `nil` in the `caseFromConfig()` function.

filepathfilter/filepathfilter_test.go
```diff
@@ -107,28 +107,28 @@ func TestPatternMatch(t *testing.T) {
 }
 
 func assertPatternMatch(t *testing.T, pattern string, filenames ...string) {
-	p := NewPattern(pattern, GitAttributes)
+	p := NewPattern(pattern, GitAttributes, NewFakeEnvironment())
```
chrisd8088: Assuming we adjust the `caseFromConfig()` function to check for a `nil` value of its parameter, we can then just pass `nil` here and in the other test functions:

```suggestion
	p := NewPattern(pattern, GitAttributes, nil)
```

chrisd8088:
FWIW, we've merged PR ##6223 now, which produces some conflicts with this PR.

In case it's helpful, I updated my `ignorecase` [branch](https://github.com/chrisd8088/git-lfs/tree/ignorecase) so that commit 55baf1f48fd7160eb4e1235de2a7f51988400d65 resolves just the merge conflicts, the next two commits do some additional post-merge cleanup, and then the final two commits implement my other suggestions for this PR.

bk2204:
I'll try to get to rebasing this onto the main branch this week or early next week.

bk2204:
Okay, I think I've got all of your suggestions here, but if I've missed one or more, please let me know and I'll try to get it addressed in a more timely manner than last time.

config/config.go
```diff
@@ -441,6 +441,7 @@ func (c *Configuration) Filesystem() *fs.Filesystem {
 		lfsdir, _ := c.Git.Get("lfs.storage")
 		c.fs = fs.New(
 			c.Os,
+			c.Git,
```
chrisd8088: I think we can drop this change now, as per commit 294e2511bf73753196e06a0df4fd620c22fdc6f1 in my `ignorecase` [branch](https://github.com/chrisd8088/git-lfs/tree/ignorecase).  We should be able to avoid touching the `config` and `fs` packages entirely, I believe.

Please feel free to just grab that commit and add it to this PR's branch!

fs/fs.go
```diff
@@ -41,6 +41,7 @@ type Filesystem struct {
 	GitStorageDir string   // parent of objects/lfs (may be same as GitDir but may not)
 	LFSStorageDir string   // parent of lfs objects and tmp dirs. Default: ".git/lfs"
 	ReferenceDirs []string // alternative local media dirs (relative to clone reference repo)
+	GitConfig     Environment
```
chrisd8088: I think we can drop this change and the one below now, as per commit 294e2511bf73753196e06a0df4fd620c22fdc6f1 in my `ignorecase` [branch](https://github.com/chrisd8088/git-lfs/tree/ignorecase).  We should be able to avoid touching the `config` and `fs` packages entirely, I believe.

Please feel free to just grab that commit and add it to this PR's branch!

git/gitattr/files.go
```diff
@@ -219,14 +219,14 @@ func AttrPathsFromReader(mp *MacroProcessor, fpath, workingDir string, rdr io.Re
 // file paths can be matched against
 // workingDir is the root of the working copy
 // gitDir is the root of the git repo
-func GetAttributeFilter(workingDir, gitDir string) *filepathfilter.Filter {
+func GetAttributeFilter(cfg filepathfilter.Environment, workingDir, gitDir string) *filepathfilter.Filter {
```
chrisd8088: May I suggest that we also use `gitEnv` here in place of `cfg`?  I think that aside from the `config` and `fs` packages, that's the only remaining difference between my `ignorecase` [branch](https://github.com/chrisd8088/git-lfs/tree/ignorecase) and this PR's branch.

```suggestion
func GetAttributeFilter(gitEnv filepathfilter.Environment, workingDir, gitDir string) *filepathfilter.Filter {
```

git/gitattr/files.go
```diff
@@ -219,14 +219,14 @@ func AttrPathsFromReader(mp *MacroProcessor, fpath, workingDir string, rdr io.Re
...
 	paths := GetAttributePaths(NewMacroProcessor(), workingDir, gitDir)
 	patterns := make([]filepathfilter.Pattern, 0, len(paths))
 
 	for _, path := range paths {
 		// Convert all separators to `/` before creating a pattern to
 		// avoid characters being escaped in situations like `subtree\*.md`
-		patterns = append(patterns, filepathfilter.NewPattern(filepath.ToSlash(path.Path), filepathfilter.GitAttributes))
+		patterns = append(patterns, filepathfilter.NewPattern(filepath.ToSlash(path.Path), filepathfilter.GitAttributes, cfg))
```
chrisd8088:
```suggestion
		patterns = append(patterns, filepathfilter.NewPattern(filepath.ToSlash(path.Path), filepathfilter.GitAttributes, gitEnv))
```

bk2204:
I've picked up all those changes, I believe.  My apologies for the oversight on the issues you'd mentioned.

Issue #6213, opened 2026-03-09
git-lfs ignores URL scheme when matching scoped `http.SCHEME://<url>.*` configs


mzr:
**Describe the bug**

Git LFS ignores the URL scheme when looking up scoped `http.<url>.*` gitconfig settings for SSL certificates. Config scoped to `https://` URLs is incorrectly applied to `http://` requests to the same host. Native Git correctly distinguishes schemes per [git-config documentation](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httplturlgt):

> Scheme (e.g., https in https://example.com/). This field must match exactly between the config key and the URL.

This causes Git LFS to attempt loading client certificates for plain HTTP requests and fail when they're not present.

**To Reproduce**

```bash
#!/usr/bin/env bash

set -euo pipefail

TESTDIR=$(mktemp -d)
trap 'rm -rf "$TESTDIR"' EXIT

echo "Repo directory: $TESTDIR/repo"
echo

cd "$TESTDIR"
git init repo >/dev/null 2>&1
cd repo
git checkout -b main >/dev/null 2>&1

cat <<EOF > file.bin
version https://git-lfs.github.com/spec/v1
oid sha256:45942f670e64f1a015f3ca66ac758d2b2f1eaa42c086f15662e36e6485ea4977
size 49042944
EOF

echo "*.lfs filter=lfs diff=lfs merge=lfs -text" > .gitattributes

git add . >/dev/null  2>&1
git commit -m "initial commit" >/dev/null

git lfs install --local >/dev/null 2>&1

LFS_HOST="example.com"

# Point to a fake HTTP remote and set the LFS URL
git remote add origin "http://$LFS_HOST/repo.git"
git config lfs.url "http://$LFS_HOST/repo/info/lfs"
git config lfs.locksverify false

# Set HTTPS-scoped client cert config — this should ONLY apply
# to https://$LFS_HOST, NOT http://
git config "http.https://$LFS_HOST/.sslcert" "/nonexistent/cert.pem"
git config "http.https://$LFS_HOST/.sslkey" "/nonexistent/key.pem"

git lfs fetch
```

yields:
```
Repo directory: /tmp/tmp.YLeVATTd9m/repo

Fetching reference refs/heads/main
batch response: Error reading client cert file "/nonexistent/cert.pem": open /nonexistent/cert.pem: no such file or directory
error: failed to fetch some objects from 'http://example.com/repo/info/lfs'
```

**Expected behavior**

[...]

PR #6214, opened 2026-03-09
Make git-lfs consider url scheme when picking up TLS-related configs

mzr:
This is to address #6213.

### DIFF
--- a/lfshttp/certs.go
+++ b/lfshttp/certs.go
@@ -17,9 +17,9 @@ import (
 )
 
 // isCertVerificationDisabledForHost returns whether SSL certificate verification
-// has been disabled for the given host, or globally
-func isCertVerificationDisabledForHost(c *Client, host string) bool {
-	hostSslVerify, _ := c.uc.Get("http", fmt.Sprintf("https://%v", host), "sslverify")
+// has been disabled for the given URL, or globally
+func isCertVerificationDisabledForHost(c *Client, u *url.URL) bool {
+	hostSslVerify, _ := c.uc.Get("http", fmt.Sprintf("%s://%s", u.Scheme, u.Host), "sslverify")
 	if hostSslVerify == "false" {
 		return true
 	}
@@ -28,10 +28,10 @@ func isCertVerificationDisabledForHost(c *Client, host string) bool {
 }
 
 // isClientCertEnabledForHost returns whether client certificate
-// are configured for the given host
-func isClientCertEnabledForHost(c *Client, host string) bool {
-	_, hostSslKeyOk := c.uc.Get("http", fmt.Sprintf("https://%v/", host), "sslKey")
-	_, hostSslCertOk := c.uc.Get("http", fmt.Sprintf("https://%v/", host), "sslCert")
+// are configured for the given URL
+func isClientCertEnabledForHost(c *Client, u *url.URL) bool {
+	_, hostSslKeyOk := c.uc.Get("http", fmt.Sprintf("%s://%s/", u.Scheme, u.Host), "sslKey")
+	_, hostSslCertOk := c.uc.Get("http", fmt.Sprintf("%s://%s/", u.Scheme, u.Host), "sslCert")
 
 	return hostSslKeyOk && hostSslCertOk
 }
@@ -68,11 +68,11 @@ func decryptPEMBlock(c *Client, block *pem.Block, path string, key []byte) ([]by
 	return buf, nil
 }
 
-// getClientCertForHost returns a client certificate for a specific host (which may
-// be "host:port" loaded from the gitconfig
-func getClientCertForHost(c *Client, host string) (*tls.Certificate, error) {
-	hostSslKey, _ := c.uc.Get("http", fmt.Sprintf("https://%v/", host), "sslKey")
-	hostSslCert, _ := c.uc.Get("http", fmt.Sprintf("https://%v/", host), "sslCert")
+// getClientCertForHost returns a client certificate for a specific URL loaded
+// from the gitconfig
+func getClientCertForHost(c *Client, u *url.URL) (*tls.Certificate, error) {
+	hostSslKey, _ := c.uc.Get("http", fmt.Sprintf("%s://%s/", u.Scheme, u.Host), "sslKey")
+	hostSslCert, _ := c.uc.Get("http", fmt.Sprintf("%s://%s/", u.Scheme, u.Host), "sslCert")
 
 	hostSslKey, err := tools.ExpandPath(hostSslKey, false)
 	if err != nil {
@@ -120,15 +120,15 @@ func getClientCertForHost(c *Client, host string) (*tls.Certificate, error) {
 // Git configuration and environment settings, if any are defined.
 // May return nil if it doesn't have anything to add, in which case the default
 // RootCAs will be used if passed to TLSClientConfig.RootCAs
-func getRootCAsForHostFromGitconfig(c *Client, host string) *x509.CertPool {
+func getRootCAsForHostFromGitconfig(c *Client, u *url.URL) *x509.CertPool {
 	// don't init pool, want to return nil not empty if none found; init only on successful add cert
 	var pool *x509.CertPool
 
-	url := fmt.Sprintf("https://%v/", host)
+	rawurl := fmt.Sprintf("%s://%s/", u.Scheme, u.Host)
 	uc := config.NewURLConfig(c.gitEnv)
 
-	backend, _ := uc.Get("http", url, "sslbackend")
-	schannelUseSslCaInfoStrValue, _ := uc.Get("http", url, "schannelusesslcainfo")
+	backend, _ := uc.Get("http", rawurl, "sslbackend")
+	schannelUseSslCaInfoStrValue, _ := uc.Get("http", rawurl, "schannelusesslcainfo")
 	schannelUseSslCaInfo := config.Bool(schannelUseSslCaInfoStrValue, false)
 
 	if backend == "schannel" && !schannelUseSslCaInfo {
@@ -142,7 +142,7 @@ func getRootCAsForHostFromGitconfig(c *Client, host string) *x509.CertPool {
 		return appendCertsFromFile(pool, cafile)
 	}
 	// http.<url>/.sslcainfo or http.<url>.sslcainfo
-	if cafile, ok := uc.Get("http", url, "sslcainfo"); ok {
+	if cafile, ok := uc.Get("http", rawurl, "sslcainfo"); ok {
 		return appendCertsFromFile(pool, cafile)
 	}
 	// GIT_SSL_CAPATH
--- a/lfshttp/certs.go
+++ b/lfshttp/certs.go
@@ -16,10 +16,10 @@ import (
 	"github.com/rubyist/tracerx"
 )
 
-// isCertVerificationDisabledForHost returns whether SSL certificate verification
+// isCertVerificationDisabledForURL returns whether SSL certificate verification
 // has been disabled for the given URL, or globally
-func isCertVerificationDisabledForHost(c *Client, u *url.URL) bool {
-	hostSslVerify, _ := c.uc.Get("http", fmt.Sprintf("%s://%s", u.Scheme, u.Host), "sslverify")
+func isCertVerificationDisabledForURL(c *Client, u *url.URL) bool {
+	hostSslVerify, _ := c.uc.Get("http", u.String(), "sslverify")
 	if hostSslVerify == "false" {
 		return true
 	}
@@ -27,11 +27,11 @@ func isCertVerificationDisabledForHost(c *Client, u *url.URL) bool {
 	return c.SkipSSLVerify
 }
 
-// isClientCertEnabledForHost returns whether client certificate
+// isClientCertEnabledForURL returns whether client certificate
 // are configured for the given URL
-func isClientCertEnabledForHost(c *Client, u *url.URL) bool {
-	_, hostSslKeyOk := c.uc.Get("http", fmt.Sprintf("%s://%s/", u.Scheme, u.Host), "sslKey")
-	_, hostSslCertOk := c.uc.Get("http", fmt.Sprintf("%s://%s/", u.Scheme, u.Host), "sslCert")
+func isClientCertEnabledForURL(c *Client, u *url.URL) bool {
+	_, hostSslKeyOk := c.uc.Get("http", u.String(), "sslKey")
+	_, hostSslCertOk := c.uc.Get("http", u.String(), "sslCert")
 
 	return hostSslKeyOk && hostSslCertOk
 }
@@ -68,11 +68,11 @@ func decryptPEMBlock(c *Client, block *pem.Block, path string, key []byte) ([]by
 	return buf, nil
 }
 
-// getClientCertForHost returns a client certificate for a specific URL loaded
+// getClientCertForURL returns a client certificate for a specific URL loaded
 // from the gitconfig
-func getClientCertForHost(c *Client, u *url.URL) (*tls.Certificate, error) {
-	hostSslKey, _ := c.uc.Get("http", fmt.Sprintf("%s://%s/", u.Scheme, u.Host), "sslKey")
-	hostSslCert, _ := c.uc.Get("http", fmt.Sprintf("%s://%s/", u.Scheme, u.Host), "sslCert")
+func getClientCertForURL(c *Client, u *url.URL) (*tls.Certificate, error) {
+	hostSslKey, _ := c.uc.Get("http", u.String(), "sslKey")
+	hostSslCert, _ := c.uc.Get("http", u.String(), "sslCert")
 
 	hostSslKey, err := tools.ExpandPath(hostSslKey, false)
 	if err != nil {
@@ -115,16 +115,16 @@ func getClientCertForHost(c *Client, u *url.URL) (*tls.Certificate, error) {
 	return &certobj, nil
 }
 
-// getRootCAsForHostFromGitconfig returns a certificate pool for that
-// specific host (which may be "host:port" or just "host") loaded from the
-// Git configuration and environment settings, if any are defined.
+// getRootCAsForURLFromGitconfig returns a certificate pool for that
+// specific URL loaded from the Git configuration and environment settings,
+// if any are defined.
 // May return nil if it doesn't have anything to add, in which case the default
 // RootCAs will be used if passed to TLSClientConfig.RootCAs
-func getRootCAsForHostFromGitconfig(c *Client, u *url.URL) *x509.CertPool {
+func getRootCAsForURLFromGitconfig(c *Client, u *url.URL) *x509.CertPool {
 	// don't init pool, want to return nil not empty if none found; init only on successful add cert
 	var pool *x509.CertPool
 
-	rawurl := fmt.Sprintf("%s://%s/", u.Scheme, u.Host)
+	rawurl := u.String()
 	uc := config.NewURLConfig(c.gitEnv)
 
 	backend, _ := uc.Get("http", rawurl, "sslbackend")
--- a/lfshttp/certs_test.go
+++ b/lfshttp/certs_test.go
@@ -49,9 +49,13 @@ var sslCAInfoMatchedHostTests = []struct {
 	{"wronghost.com", false},
 }
 
-func clientForHost(c *Client, host string) *http.Client {
+func httpsURLForHost(host string) *url.URL {
 	u, _ := url.Parse(fmt.Sprintf("https://%v", host))
-	client, _ := c.HttpClient(u, creds.BasicAccess)
+	return u
+}
+
+func clientForHost(c *Client, host string) *http.Client {
+	client, _ := c.HttpClient(httpsURLForHost(host), creds.BasicAccess)
 	return client
 }
 
@@ -73,7 +77,7 @@ func TestCertFromSSLCAInfoConfig(t *testing.T) {
 		assert.Nil(t, err)
 
 		for _, matchedHostTest := range sslCAInfoMatchedHostTests {
-			pool := getRootCAsForHostFromGitconfig(c, matchedHostTest.hostName)
+			pool := getRootCAsForHostFromGitconfig(c, httpsURLForHost(matchedHostTest.hostName))
 
 			var shouldOrShouldnt string
 			if matchedHostTest.shouldMatch {
@@ -96,7 +100,7 @@ func TestCertFromSSLCAInfoConfig(t *testing.T) {
 
 	// Should match any host at all
 	for _, matchedHostTest := range sslCAInfoMatchedHostTests {
-		pool := getRootCAsForHostFromGitconfig(c, matchedHostTest.hostName)
+		pool := getRootCAsForHostFromGitconfig(c, httpsURLForHost(matchedHostTest.hostName))
 		assert.NotNil(t, pool)
 	}
 }
@@ -117,7 +121,7 @@ func TestCertFromSSLCAInfoEnv(t *testing.T) {
 
 	// Should match any host at all
 	for _, matchedHostTest := range sslCAInfoMatchedHostTests {
-		pool := getRootCAsForHostFromGitconfig(c, matchedHostTest.hostName)
+		pool := getRootCAsForHostFromGitconfig(c, httpsURLForHost(matchedHostTest.hostName))
 		assert.NotNil(t, pool)
 	}
 }
@@ -140,7 +144,7 @@ func TestCertFromSSLCAInfoEnvIsIgnoredForSchannelBackend(t *testing.T) {
 
 	// Should match any host at all
 	for _, matchedHostTest := range sslCAInfoMatchedHostTests {
-		pool := getRootCAsForHostFromGitconfig(c, matchedHostTest.hostName)
+		pool := getRootCAsForHostFromGitconfig(c, httpsURLForHost(matchedHostTest.hostName))
 		assert.Nil(t, pool)
 	}
 }
@@ -164,7 +168,7 @@ func TestCertFromSSLCAInfoEnvWithSchannelBackend(t *testing.T) {
 
 	// Should match any host at all
 	for _, matchedHostTest := range sslCAInfoMatchedHostTests {
-		pool := getRootCAsForHostFromGitconfig(c, matchedHostTest.hostName)
+		pool := getRootCAsForHostFromGitconfig(c, httpsURLForHost(matchedHostTest.hostName))
 		assert.NotNil(t, pool)
 	}
 }
@@ -183,7 +187,7 @@ func TestCertFromSSLCAPathConfig(t *testing.T) {
 
 	// Should match any host at all
 	for _, matchedHostTest := range sslCAInfoMatchedHostTests {
-		pool := getRootCAsForHostFromGitconfig(c, matchedHostTest.hostName)
+		pool := getRootCAsForHostFromGitconfig(c, httpsURLForHost(matchedHostTest.hostName))
 		assert.NotNil(t, pool)
 	}
 }
@@ -201,7 +205,7 @@ func TestCertFromSSLCAPathEnv(t *testing.T) {
 
 	// Should match any host at all
 	for _, matchedHostTest := range sslCAInfoMatchedHostTests {
-		pool := getRootCAsForHostFromGitconfig(c, matchedHostTest.hostName)
+		pool := getRootCAsForHostFromGitconfig(c, httpsURLForHost(matchedHostTest.hostName))
 		assert.NotNil(t, pool)
 	}
 }
@@ -277,4 +281,92 @@ func TestCertVerifyDisabledHostConfig(t *testing.T) {
 	if assert.True(t, ok) {
 		assert.False(t, tr.TLSClientConfig.InsecureSkipVerify)
 	}
+}
+
+func TestCertVerifyDisabledRespectsScheme(t *testing.T) {
+	c, err := NewClient(NewContext(nil, nil, map[string]string{
+		"http.https://example.com.sslverify": "false",
+	}))
+	assert.Nil(t, err)
+
+	// HTTPS URL should match HTTPS-scoped sslverify config
+	httpsURL, _ := url.Parse("https://example.com/repo")
+	httpsClient, _ := c.HttpClient(httpsURL, creds.BasicAccess)
+	tr, ok := httpsClient.Transport.(*http.Transport)
+	if assert.True(t, ok) {
+		assert.True(t, tr.TLSClientConfig.InsecureSkipVerify,
+			"HTTPS URL should have InsecureSkipVerify from HTTPS-scoped config")
+	}
+
+	// HTTP URL should NOT match HTTPS-scoped sslverify config
+	httpURL, _ := url.Parse("http://example.com/repo")
+	httpClient, _ := c.HttpClient(httpURL, creds.BasicAccess)
+	tr, ok = httpClient.Transport.(*http.Transport)
+	if assert.True(t, ok) {
+		assert.False(t, tr.TLSClientConfig.InsecureSkipVerify,
+			"HTTP URL should not have InsecureSkipVerify from HTTPS-scoped config")
+	}
+}
+
+func TestClientCertRespectsScheme(t *testing.T) {
+	c, err := NewClient(NewContext(nil, nil, map[string]string{
+		"http.https://example.com/.sslkey":  "/path/to/key.pem",
+		"http.https://example.com/.sslcert": "/path/to/cert.pem",
+	}))
+	assert.Nil(t, err)
+
+	// HTTPS URL should find cert config
+	httpsURL, _ := url.Parse("https://example.com/repo")
+	assert.True(t, isClientCertEnabledForHost(c, httpsURL),
+		"client cert should be enabled for HTTPS URL with HTTPS-scoped config")
+
+	// HTTP URL should NOT find cert config scoped to HTTPS
+	httpURL, _ := url.Parse("http://example.com/repo")
+	assert.False(t, isClientCertEnabledForHost(c, httpURL),
+		"client cert should not be enabled for HTTP URL with HTTPS-scoped config")
+}
+
+func TestTransportRespectsSchemeForCerts(t *testing.T) {
+	c, err := NewClient(NewContext(nil, nil, map[string]string{
+		"http.https://git.example.com/.sslkey":  "/nonexistent/key.pem",
+		"http.https://git.example.com/.sslcert": "/nonexistent/cert.pem",
+	}))
+	assert.Nil(t, err)
+
+	// HTTP URL should not try to load certs from HTTPS-scoped config
+	httpURL, _ := url.Parse("http://git.example.com/repo.git/info/lfs")
+	_, err = c.Transport(httpURL, creds.BasicAccess)
+	assert.Nil(t, err,
+		"HTTP URL should not try to load certs from HTTPS-scoped config")
+
+	// HTTPS URL should try to load certs and fail (files don't exist)
+	httpsURL, _ := url.Parse("https://git.example.com/repo.git/info/lfs")
+	_, err = c.Transport(httpsURL, creds.BasicAccess)
+	assert.NotNil(t, err,
+		"HTTPS URL should try to load certs from HTTPS-scoped config")
+}
+
+func TestRootCAsRespectsScheme(t *testing.T) {
+	tempfile, err := os.CreateTemp("", "testcert")
+	assert.Nil(t, err, "Error creating temp cert file")
+	defer os.Remove(tempfile.Name())
+
+	_, err = tempfile.WriteString(testCert)
+	assert.Nil(t, err, "Error writing temp cert file")
+	tempfile.Close()
+
+	c, err := NewClient(NewContext(nil, nil, map[string]string{
+		"http.https://example.com.sslcainfo": tempfile.Name(),
+	}))
+	assert.Nil(t, err)
+
+	// HTTPS URL should find the CA cert
+	httpsURL, _ := url.Parse("https://example.com/repo")
+	pool := getRootCAsForHostFromGitconfig(c, httpsURL)
+	assert.NotNil(t, pool, "HTTPS URL should find CA cert from HTTPS-scoped config")
+
+	// HTTP URL should NOT find the CA cert scoped to HTTPS
+	httpURL, _ := url.Parse("http://example.com/repo")
+	pool = getRootCAsForHostFromGitconfig(c, httpURL)
+	assert.Nil(t, pool, "HTTP URL should not find CA cert from HTTPS-scoped config")
 }
--- a/lfshttp/certs_test.go
+++ b/lfshttp/certs_test.go
@@ -346,6 +346,32 @@ func TestTransportRespectsSchemeForCerts(t *testing.T) {
 		"HTTPS URL should try to load certs from HTTPS-scoped config")
 }
 
+func TestRootCAsPathScoped(t *testing.T) {
+	tempfile, err := os.CreateTemp("", "testcert")
+	assert.Nil(t, err, "Error creating temp cert file")
+	defer os.Remove(tempfile.Name())
+
+	_, err = tempfile.WriteString(testCert)
+	assert.Nil(t, err, "Error writing temp cert file")
+	tempfile.Close()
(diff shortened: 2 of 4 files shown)


chrisd8088:
Hey, thanks for the PR!  I'll take a close look as soon as I'm able.

One detail I noticed on a first glance is that you don't actually need to do the `set +e`/`set -e` dance around the `git lfs fetch` commands in your shell tests.  Even though you expect the command to fail, it won't cause the test to fail so long as the command is part of a pipeline and the final `tee(2)` command in the pipeline succeeds, which it always does.

We now typically try to write something like the following, which allows us to check that the command did in fact fail as expected, and `set +e`/`set -e` are not needed at all:

```
  git lfs fetch 2>&1 | tee fetch.log
    echo >&2 "fatal: expected 'git lfs fetch' to fail ..."
    exit 1
  fi
```

mzr:
Alrighty, I just changed it. Thanks for pointing that out!

lfshttp/certs.go
```diff
@@ -17,9 +17,9 @@ import (
 )
 
 // isCertVerificationDisabledForHost returns whether SSL certificate verification
-// has been disabled for the given host, or globally
-func isCertVerificationDisabledForHost(c *Client, host string) bool {
-	hostSslVerify, _ := c.uc.Get("http", fmt.Sprintf("https://%v", host), "sslverify")
+// has been disabled for the given URL, or globally
+func isCertVerificationDisabledForHost(c *Client, u *url.URL) bool {
+	hostSslVerify, _ := c.uc.Get("http", fmt.Sprintf("%s://%s", u.Scheme, u.Host), "sslverify")
```
chrisd8088: I think you've identified an interesting historical oversight in our support for URL-scoped `http.<url>.*` Git configuration [options](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httplturlgt), and while these changes partially address that issue, I suspect they don't go far enough.

In brief, I believe we'll likely want to change the names of these three functions (including in the preceding comments) and just use the whole URL parameter when searching for `http.<url>.*` configuration options.

For example:

```go
// isCertVerificationDisabledForURL returns whether SSL certificate verification
// has been disabled for the given URL, or globally
func isCertVerificationDisabledForURL(c *Client, u *url.URL) bool {
	hostSslVerify, _ := c.uc.Get("http", u, "sslverify")

	if hostSslVerify == "false" {
		return true
	}

	return c.SkipSSLVerify
}
```
<p/>

There's a lot of history here which I'll put into a longer and more detailed set of review comments, but the gist of the problem is that the current state of affairs likely stems from an incomplete original implementation from PRs #1787 and #1839 which was then incompletely migrated in PR #2160 to our full emulation of Git's matching rules for `http.<url>.*` options when that was added in PRs #1912 and #3392.

I still haven't unpacked all the consequences of this situation, so I've held off on a full review, but I thought I should provide an update and a brief summary of the likely issues we'll want to address.

(As one example of the unintended complexity here, the `Get()` method of our `URLConfig` type falls back to [returning](https://github.com/git-lfs/git-lfs/blob/52e7457830d9f16e97ee2c8fd5051db9013e1efa/config/url_config.go#L38) the `http.sslVerify` option's value, if any, when no matching `http.<url>.sslVerify` option is found.  So the fact that we also look up the value of the `http.sslVerify` option when [setting](https://github.com/git-lfs/git-lfs/blob/52e7457830d9f16e97ee2c8fd5051db9013e1efa/lfshttp/client.go#L89) the `SkipSSLVerify` field is effectively redundant and a remnant of the original implementation.)

Thanks very much for bringing all this to our attention!

t/t-http-scheme-config.sh
```diff
@@ -0,0 +1,92 @@
...
+version https://git-lfs.github.com/spec/v1
+oid sha256:45942f670e64f1a015f3ca66ac758d2b2f1eaa42c086f15662e36e6485ea4977
+size 49042944
+EOF
+  git add file.bin
+  git commit -m "initial commit"
+
+  # Configure an http:// LFS endpoint
+  git remote add origin "http://192.0.2.1:9999/repo.git"
+  git config lfs.url "http://192.0.2.1:9999/repo/info/lfs"
```
chrisd8088: Having some shell tests for this PR is fantastic, thank you!  We'll definitely want these to help verify the changes in this PR.

As an initial thought only, I suspect we can make use of our test server [utility](https://github.com/git-lfs/git-lfs/blob/52e7457830d9f16e97ee2c8fd5051db9013e1efa/t/cmd/lfstest-gitserver.go) to actually perform both non-TLS and TLS connections in the tests, and so might not need to use a test IP like `192.0.2.1` at all.

Just as an example, the `clone ClientCert` [test](https://github.com/git-lfs/git-lfs/blob/52e7457830d9f16e97ee2c8fd5051db9013e1efa/t/t-clone.sh#L192-L310) in our `t/t-clone.sh` script makes use of the server's TLS variant and the client certificates it [creates](https://github.com/git-lfs/git-lfs/blob/52e7457830d9f16e97ee2c8fd5051db9013e1efa/t/cmd/lfstest-gitserver.go#L82-L94).

chrisd8088:
Thank you for all the updates to this PR!  I'm going to try to give it a close look again over the weekend.

lfshttp/certs.go
```diff
@@ -16,22 +16,22 @@ import (
...
 )
 
-// isCertVerificationDisabledForHost returns whether SSL certificate verification
-// has been disabled for the given host, or globally
-func isCertVerificationDisabledForHost(c *Client, host string) bool {
-	hostSslVerify, _ := c.uc.Get("http", fmt.Sprintf("https://%v", host), "sslverify")
+// isCertVerificationDisabledForURL returns whether SSL certificate verification
+// has been disabled for the given URL, or globally
+func isCertVerificationDisabledForURL(c *Client, u *url.URL) bool {
+	hostSslVerify, _ := c.uc.Get("http", u.String(), "sslverify")
```
chrisd8088: I'd suggest we rename the local variable here to just `sslVerify`, since it's not specific to the host name any more.

lfshttp/certs.go
```diff
@@ -16,22 +16,22 @@ import (
...
-// isClientCertEnabledForHost returns whether client certificate
-// are configured for the given host
-func isClientCertEnabledForHost(c *Client, host string) bool {
-	_, hostSslKeyOk := c.uc.Get("http", fmt.Sprintf("https://%v/", host), "sslKey")
-	_, hostSslCertOk := c.uc.Get("http", fmt.Sprintf("https://%v/", host), "sslCert")
+// isClientCertEnabledForURL returns whether client certificate
+// are configured for the given URL
+func isClientCertEnabledForURL(c *Client, u *url.URL) bool {
+	_, hostSslKeyOk := c.uc.Get("http", u.String(), "sslKey")
+	_, hostSslCertOk := c.uc.Get("http", u.String(), "sslCert")
```
chrisd8088, line 34: I'd suggest we rename the local variables here to just `sslKeyOK` and `sslCertOK`, since they're not specific to the host name any more.

The same is also true in the `getRootCAsForURLFromGitconfig()` function below (i.e., `sslKey` and `sslCert`).

One nice thing is that it's an opportunity to more closely adhere to Google's Go naming style [guidelines](https://google.github.io/styleguide/go/decisions.html#initialisms) for initialisms, so for example, we can replace `hostSslKeyOk` with `sslKeyOK` and keep the cases of `ssl` and `OK` internally consistent.

lfshttp/client.go
```diff
@@ -467,9 +468,9 @@ func (c *Client) Transport(u *url.URL, access creds.AccessMode) (http.RoundTripp
 		Renegotiation: tls.RenegotiateFreelyAsClient,
 	}
 
-	if isClientCertEnabledForHost(c, host) {
+	if isClientCertEnabledForURL(c, u) {
 		tracerx.Printf("http: client cert for %s", host)
```
chrisd8088, line 472: Let's change this trace log message to include the full URL used to look for a client TLS certificate, e.g.:

```suggestion
		tracerx.Printf("http: client cert for %q", u.String())
```

We might as well also use `%q` here, just to make the output a bit tidier.

Note that when we make this change, there will be no need to [initialize](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfshttp/client.go#L420) the local `host` variable at all in this method, so we can just drop that line.

lfshttp/certs.go
```diff
@@ -16,22 +16,22 @@ import (
...
-	hostSslVerify, _ := c.uc.Get("http", fmt.Sprintf("https://%v", host), "sslverify")
+// isCertVerificationDisabledForURL returns whether SSL certificate verification
+// has been disabled for the given URL, or globally
+func isCertVerificationDisabledForURL(c *Client, u *url.URL) bool {
+	hostSslVerify, _ := c.uc.Get("http", u.String(), "sslverify")
 	if hostSslVerify == "false" {
 		return true
 	}
 
 	return c.SkipSSLVerify
```
chrisd8088, line 27: In my previous [comment](https://github.com/git-lfs/git-lfs/pull/6214#discussion_r3020303864) I alluded to our development history and how it's left us with a slightly confusing and redundant code design:

> ... the current state of affairs likely stems from an incomplete original implementation from PRs #1787 and #1839 which was then incompletely migrated in PR #2160 to our full emulation of Git's matching rules for `http.<url>.*` options when that was added in PRs #1912 and #3392.

I've expanded on the details below, and I'll put notes on the whole tangled history in a comment, but here's what I think we should do:

```suggestion
	if _, ok := c.osEnv.Get("GIT_SSL_NO_VERIFY"); ok {
		return true
	}
	return !c.uc.Bool("http", u.String(), "sslVerify", true)
```
</p>

We can then remove the `SkipSSLVerify` [field](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfshttp/client.go#L66) from the `Client` structure entirely, as this function is the only place it's used, and we don't need it here at all if we rewrite the function as in my suggestion above.

That does mean we also have to move and refactor the two Go [tests](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfshttp/client_test.go#L194-L234) in the `lfshttp/client_test.go ` source file.

As well, there's another test in the `lfshttp/client_test.go` source file which [sets](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfshttp/client_test.go#L118-L121) multiple `http.<url>.sslVerify` options, but so far as I can tell, these are unnecessary and can be removed.  Other tests just [set](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfshttp/client_test.go#L310) the `http.sslVerify` option only, and given the limitation this PR addresses, the extra `http.<url>.sslVerify` options in the `TestClientRedirect()` test function would never have had any effect.  (The `TestClientRedirect()` function was added first, in PR #2235, while the `TestHttp2()` and `TestHttpVersion()` functions were added later.)

#### Git Implementation

It's worth reviewing how the precedence and parsing of the `http.sslVerify`/`http.<url>.sslVerify` option and the `GIT_SSL_NO_VERIFY` environment variable are implemented by Git.

First, URL matching [rules](https://github.com/git/git/blob/v2.54.0/urlmatch.c#L479-L495) are [applied](https://github.com/git/git/blob/v2.54.0/http.c#L1341) to look for the most-specific variant of the `http.sslVerify`/`http.<url>.sslVerify` options.  These options are [expected](https://github.com/git/git/blob/v2.54.0/http.c#L398-L399) to have Boolean [values](https://github.com/git/git/blob/v2.54.0/config.c#L1265), which are [parsed](https://github.com/git/git/blob/v2.54.0/parse.c#L157-L182) such that the terms `false`, `no`, and `off`, along with the integer `0`, all count as `false`, while their opposite terms and all non-zero integers count as `true`.

Next, if the `GIT_SSL_NO_VERIFY` environment variable is defined at all, with any value (even an empty string), this [overrides](https://github.com/git/git/blob/v2.54.0/http.c#L1403-L1404) the configuration option settings with a `false` value.  Finally, if no option and no environment variable were found, the default value [used](https://github.com/git/git/blob/v2.54.0/http.c#L1418-L1419) is `true`.

#### Current Git LFS Implementation

Our implementation is different from Git's in a few respects, some of which stem from the early development history of our project.

In commit 8009d17bbc97c57f1cf553a4218078b228780bf3 of PR #1787 (which was part of a larger set of PRs that were all merged into the main development branch in PR #1839), we implemented a simplified version of Git's URL-matching logic for `http.<url>.*` configuration options, which just searched for an `http.https://<host>/.sslVerify` option.

There was already a discrepancy in this initial version of the code between how the generic `http.sslVerify` option and the URL-specific variant of that option were treated.  The former was [parsed](https://github.com/git-lfs/git-lfs/blob/8009d17bbc97c57f1cf553a4218078b228780bf3/lfsapi/lfsapi.go#L70) using the `environment.Bool()` method from our `config` package, which recognizes a number of possible values like `0` and `false` as being equivalent to `false`, whereas the latter was retrieved with the `environment.Get()` method and then [checked](https://github.com/git-lfs/git-lfs/blob/8009d17bbc97c57f1cf553a4218078b228780bf3/lfsapi/certs.go#L16) only for the value `false`.  The following example illustrates how this discrepancy remains the current code:

<details>

```
$ git init test
$ cd test
$ git lfs track "*.bin"
$ echo foo >foo.bin
$ git commit -m foo

$ git remote add origin https://expired.badssl.com/test/test.git
$ git lfs push --object-id origin b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c                
batch response: Post "https://expired.badssl.com/test/test.git/info/lfs/objects/batch": tls: failed to verify certificate: x509: certificate has expired or is not yet valid: “*.badssl.com” certificate is expired

# "false" is accepted for both "http.sslVerify" and "http.https://<host>/.sslVerify"
$ git -c http.sslVerify=false lfs push --object-id origin b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c
batch response: Repository or object not found: https://expired.badssl.com/test/test.git/info/lfs/objects/batch
$ git -c http.https://expired.badssl.com/.sslVerify=false lfs push --object-id origin b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c 
batch response: Repository or object not found: https://expired.badssl.com/test/test.git/info/lfs/objects/batch

# "0" is accepted only for "http.sslVerify"
$ git -c http.sslVerify=0 lfs push --object-id origin b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c
batch response: Repository or object not found: https://expired.badssl.com/test/test.git/info/lfs/objects/batch
$ git -c http.https://expired.badssl.com/.sslVerify=0 lfs push --object-id origin b5bb9d8014a0f9b1d61e21e796d78dccdf1352f23cd32812f4850b878ae4944c                     
batch response: Post "https://expired.badssl.com/test/test.git/info/lfs/objects/batch": tls: failed to verify certificate: x509: certificate has expired or is not yet valid: “*.badssl.com” certificate is expired
```
</details>

So that's one problem: we use our `config.Bool()` [function](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/config/environment.go#L142-L165) to parse only the generic version of the `http.sslVerify` / `http.<url>.sslVerify` option, whereas for the URL-specific version we only accept `false` as a possible value, which doesn't match Git's implementation.

(As a sidebar, our `config.Bool()` function doesn't [parse](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/config/environment.go#L152-L165) Boolean values in exactly the same way as Git, so for instance Git treats any non-zero integer as `true`, but we only accept the integer value `1` as equivalent to `true`.  But this is a discrepancy we should postpone resolving to another, future PR.)

Another discrepancy pertains to how we handle the `GIT_SSL_NO_VERIFY` environment variable.  If the variable is defined with any value, even the empty string, then Git does not perform TLS/SSL certificate validation.  As the Git documentation [states](https://github.com/git/git/blob/v2.54.0/Documentation/git.adoc?plain=1#L743-L746):

> Setting and exporting this environment variable to any value tells Git not to verify the SSL certificate when fetching or pushing over HTTPS.

When we first introduced support for the `GIT_SSL_NO_VERIFY` variable, in commit f83b028792a68f8ac07b86f2e8999ed8aee9d6b2 of PR #145, we [accepted](https://github.com/git-lfs/git-lfs/blob/f83b028792a68f8ac07b86f2e8999ed8aee9d6b2/gitmediaclient/client.go#L158) any non-empty value of the variable as indicating that certificate validation should be disabled.  That was almost equivalent to Git's implementation, except for the case of an empty string.

However, in commit 410830ab73ef15477c62d2a7537f1c6f65b4ac51 of PR #401 we began [parsing](https://github.com/git-lfs/git-lfs/blob/410830ab73ef15477c62d2a7537f1c6f65b4ac51/lfs/http.go#L124) the value of the `GIT_SSL_NO_VERIFY` variable as a Boolean field, which is quite different from how Git interprets it, we continue to mis-parse this variable as a Boolean field today.

Finally, there's a lesser issue that we currently often read the value of the `http.sslVerify` option twice.

We parse it first when we [set](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfshttp/client.go#L110) the value of the `SkipSSLVerify` field of our `lfshttp.Client` structure.  That design dates back to our original introduction of support for this configuration option in PRs #1787 and #1839.

Later, in commit 9a17bd26d49c93ced23afff713aeb9cd49f1e14f of PR #2160, we started using the initial version of our `config.URLConfig` structure's `Get()` method to [retrieve](https://github.com/git-lfs/git-lfs/blob/9a17bd26d49c93ced23afff713aeb9cd49f1e14f/lfsapi/certs.go#L17) the value of an `http.https://<host>/.sslVerify` option if one was set.  Because we still constructed the `<url>` part from just a `<host>` value, this change was largely ineffective at properly supporting the full range of possible `http.<url>.sslVerify` options.  But one consequence was that we now often read the value of the generic version of the `http.sslVerify` / `http.<url>.sslVerify` option a second time, because when no host-specific version matching `http.https://<host>/.sslVerify` is found the `Get()` method [falls back](https://github.com/git-lfs/git-lfs/blob/9a17bd26d49c93ced23afff713aeb9cd49f1e14f/config/url_config.go#L37) to looking for the generic version of the option.

#### Revised Git LFS Implementation

With all that said, I believe my proposed revision to the `isCertVerificationDisabledForURL()` function makes the following corrections and refinements:

1. We check the environment variable [using](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/config/os_fetcher.go#L47) the `os.LookupEnv()` function, and if the variable has any defined value at all, we interpret that to mean we should not perform TLS/SSL certificate verification (i.e., the `isCertVerificationDisabledForURL()` function should return `true`).  This allows us to skip the more complicated URL-matching phase entirely if the variable is defined, and corrects our historical mis-parsing of the variable.

2. If the environment variable isn't defined, then we use the `Bool()` method of our `config.URLConfig` structure to look for the best-matching `http.sslVerify`/`http.<url>.sslVerify` option, passing a default value of `true` in case none of these options are defined.  This avoids any extra lookups of the generic version of the option and allows us to fully support Boolean values for the URL-specific version of the option.

lfshttp/certs_test.go
```diff
@@ -278,3 +282,117 @@ func TestCertVerifyDisabledHostConfig(t *testing.T) {
...
+		"HTTP URL should not try to load certs from HTTPS-scoped config")
+
+	// HTTPS URL should try to load certs and fail (files don't exist)
+	httpsURL, _ := url.Parse("https://git.example.com/repo.git/info/lfs")
+	_, err = c.Transport(httpsURL, creds.BasicAccess)
+	assert.NotNil(t, err,
+		"HTTPS URL should try to load certs from HTTPS-scoped config")
+}
+
+func TestRootCAsPathScoped(t *testing.T) {
```
chrisd8088, line 349: These two new `TestRootCAsPathScoped()` and `TestRootCAsRespectsScheme()` functions both exercise the `getRootCAsForURLFromGitconfig()` function, which is what the first six test [functions](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfshttp/certs_test.go#L58-L199) in this source file also validate, so at a minimum we should order them together with those existing functions.

But I think we can also consolidate all these test functions and also expand the set of checks we perform.  That's going to require quite a large refactoring of the tests, though.

lfshttp/certs.go
```diff
@@ -115,20 +115,20 @@ func getClientCertForHost(c *Client, host string) (*tls.Certificate, error) {
...
 
-	url := fmt.Sprintf("https://%v/", host)
+	rawurl := u.String()
 	uc := config.NewURLConfig(c.gitEnv)
 
-	backend, _ := uc.Get("http", url, "sslbackend")
-	schannelUseSslCaInfoStrValue, _ := uc.Get("http", url, "schannelusesslcainfo")
+	backend, _ := uc.Get("http", rawurl, "sslbackend")
+	schannelUseSslCaInfoStrValue, _ := uc.Get("http", rawurl, "schannelusesslcainfo")
 	schannelUseSslCaInfo := config.Bool(schannelUseSslCaInfoStrValue, false)
```
chrisd8088, line 132: We can simplify this section by using the `URLConfig.Bool()` method:

```suggestion
	backend, _ := uc.Get("http", rawurl, "sslbackend")
	schannelUseSslCaInfo := uc.Bool("http", rawurl, "schannelusesslcainfo", false)
```

t/t-http-scheme-config.sh
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env bash
```
chrisd8088, line 1: I think we should probably just put these tests in our `t/t-config.sh` file, as they pertain to the validation of how we read and apply configuration settings.

lfshttp/certs.go
```diff
@@ -142,7 +142,7 @@ func getRootCAsForHostFromGitconfig(c *Client, host string) *x509.CertPool {
 		return appendCertsFromFile(pool, cafile)
 	}
 	// http.<url>/.sslcainfo or http.<url>.sslcainfo
-	if cafile, ok := uc.Get("http", url, "sslcainfo"); ok {
+	if cafile, ok := uc.Get("http", rawurl, "sslcainfo"); ok {
 		return appendCertsFromFile(pool, cafile)
 	}
 	// GIT_SSL_CAPATH
```
chrisd8088, line 148: We apparently have never supported `http.<url>.sslCAPath` configuration options, even though Git does.

Like some of our issues with the `http.sslVerify`/`http.<url>.sslVerify` option, this problem goes back to the original implementation in commit 8009d17bbc97c57f1cf553a4218078b228780bf3 of PRs #1787 and #1839.

If that was the only issue, I would recommend just the following change:

```patch
 		return appendCertsFromFilesInDir(pool, cadir)
 	}
 	// http.sslcapath
-	if cadir, ok := c.gitEnv.Get("http.sslcapath"); ok {
+	if cadir, ok := uc.Get("http", rawurl, "sslcapath"); ok {
 		return appendCertsFromFilesInDir(pool, cadir)
 	}
```

However, I believe there's another couple of problems here as well.

First, we treat the `GIT_SSL_CAINFO` environment variable and the `http.sslCAInfo` / `http.<url>.sslCAInfo` option as if they should [override](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfshttp/certs.go#L140-L147) the `GIT_SSL_CAPATH` environment variable and the `http.sslCAPath` / `http.<url>.sslCAPath` option.

And second, we also [ignore](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfshttp/certs.go#L130-L136) the `GIT_SSL_CAPATH` environment variable and the `http.sslCAPath` / `http.<url>.sslCAPath` option if the `http.sslBackend` / `http.<url>.sslBackend` option is set to `schannel` and the `http.schannelUseSSLCAInfo` / `http.<url>.schannelUseSSLCAInfo` option is not set to `true`.

However, that's not how Git handles these options.  Regardless of any other configuration settings, Git [sets](https://github.com/git/git/blob/v2.54.0/http.c#L1121-L1122) the libcurl library's `CURLOPT_CAPATH` [option](https://curl.se/libcurl/c/CURLOPT_CAPATH.html) from the `GIT_SSL_CAPATH` environment variable or the `http.sslCAPath` / `http.<url>.sslCAPath` configuration option.

Then, if the `http.sslBackend` / `http.<url>.sslBackend` option is not set to `schannel` or the `http.schannelUseSSLCAInfo` / `http.<url>.schannelUseSSLCAInfo` option is set to `true`, Git [sets](https://github.com/git/git/blob/v2.54.0/http.c#L1130-L1131) the libcurl library's `CURLOPT_CAINFO` [option](https://curl.se/libcurl/c/CURLOPT_CAINFO.html) from the `GIT_SSL_CAINFO` environment variable or the `http.sslCAInfo` / `http.<url>.sslCAInfo` option.

In other words, the "certificate file" (i.e., "info") and "certificate path" options are not incompatible, but we treat them as such right now, and let the "info" setting override any "path" setting.  In particular, if either is set, the libcurl library [passes both](https://github.com/curl/curl/blob/curl-8_20_0/lib/vtls/openssl.c#L3020-L3072) through to the OpenSSL library (assuming that's the SSL/TLS backend in use) via the `X509_STORE_load_file()` and `X509_STORE_load_path()` [functions](https://github.com/openssl/openssl/blob/openssl-4.0.0/crypto/x509/x509_d2.c#L65-L80).  These either directly or indirectly invoke the `X509_STORE_add_lookup()` function, which [pushes](https://github.com/openssl/openssl/blob/openssl-4.0.0/crypto/x509/x509_lu.c#L311-L320) the appropriate lookup function into the list of locations to check for valid certificates.

Note that when Git [reads](https://github.com/git/git/blob/v2.54.0/http.c#L1298-L1302) these environment variables, even an empty string value is accepted, while we [ignore](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfshttp/certs.go#L141) [empty](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfshttp/certs.go#L149) strings, which is another discrepancy between our implementation and Git's.

Note, too, that the `CURLOPT_CAPATH` option is documented to have no effect on Windows due to an OpenSSL limitation (which is likely why there's no `http.schannelUseSSLCAPath` option in Git).  But we read certificates from the path specified by the `GIT_SSL_CAPATH` environment variable or the `http.sslCAPath` / `http.<url>.sslCAPath` option and parse them even on Windows, so we should correct that inconsistency.

All of which is to say that I think we need to rewrite this whole [section](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfshttp/certs.go#L130-L155) to look something like this:

```go
	if runtime.GOOS != "windows" {
		// GIT_SSL_CAPATH overrides http.sslCAPath/http.<url>.sslCAPath
		if cadir, ok := c.osEnv.Get("GIT_SSL_CAPATH"); ok {
			pool = appendCertsFromFilesInDir(pool, cadir)
		} else if cadir, ok := uc.Get("http", rawurl, "sslcapath"); ok {
			pool = appendCertsFromFilesInDir(pool, cadir)
		}
	}

	backend, _ := uc.Get("http", rawurl, "sslbackend")
	schannelUseSslCaInfo := uc.Bool("http", rawurl, "schannelusesslcainfo", false)

	if backend == "schannel" && !schannelUseSslCaInfo {
		return pool
	}

	// GIT_SSL_CAINFO overrides http.sslCAInfo/http.<url>.sslCAInfo
	if cafile, ok := c.osEnv.Get("GIT_SSL_CAINFO"); ok {
		pool = appendCertsFromFile(pool, cafile)
	} else if cafile, ok := uc.Get("http", rawurl, "sslcainfo"); ok {
		pool = appendCertsFromFile(pool, cafile)
	}
```

t/t-http-scheme-config.sh
```diff
@@ -0,0 +1,73 @@
...
+  git add .gitattributes a.dat
+  git commit -m "add lfs file"
+
+  # Extract host:port from $GITSERVER (e.g. "127.0.0.1:PORT") so we can
+  # build a config key that is scoped to the https:// scheme for that host.
+  gitserver_hostport="${GITSERVER#*://}"
+
+  # Set client cert config scoped to https:// — should NOT apply to http://
+  git config "http.https://$gitserver_hostport/.sslCert" "/nonexistent/cert.pem"
+  git config "http.https://$gitserver_hostport/.sslKey" "/nonexistent/key.pem"
```
chrisd8088, line 24: Since we want to to ensure that the certificate and key files don't exist, I think we want to use paths which we can guarantee are unique and local to the tests, rather than using paths which are just unlikely to exist on a system.

Admittedly, it's very unlikely that a system would have a root-level directory named `nonexistent` with those particular files in it, but partly for the sake of consistency with other tests which [check](https://github.com/git-lfs/git-lfs/blob/fcc19a9333a17344d08c4ee8edf597ab14c4e801/t/t-pull.sh#L1257) that files don't exist, I'd suggest we do:

```suggestion
  git config "http.https://$gitserver_hostport/.sslCert" "$TRASHDIR/nonexistent/cert.pem"
  git config "http.https://$gitserver_hostport/.sslKey" "$TRASHDIR/nonexistent/key.pem"
```

We should make the same change to the similar lines in the subsequent test as well, of course.

t/t-http-scheme-config.sh
```diff
@@ -0,0 +1,73 @@
...
+  if [ "${PIPESTATUS[0]}" -ne 0 ]; then
+    echo >&2 "FAIL: git-lfs tried to load client cert for http:// URL from https-scoped config"
+    cat push.log >&2
+    exit 1
+  fi
+
+  if grep -q "Error reading client cert file" push.log; then
+    echo >&2 "FAIL: git-lfs tried to load client cert for http:// URL from https-scoped config"
+    exit 1
+  fi
```
chrisd8088, line 39: This is an interesting case, because if the `git lfs push` command fails, we'd ideally like to check if it did so specifically because it tried to load the nonexistent certificate file, rather than for some other spurious reason (e.g., because the test server had crashed).

A [common](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/t/t-unlock.sh#L604-L607) [pattern](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/t/t-push.sh#L56-L62) in our shell tests is to check that a command failed and then check that a specific message appears in the logged output from that command.

We certainly also have [lots](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/t/t-checkout.sh#L244-L251) of [instances](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/t/t-batch-storage-retries.sh#L254-L263) in our shell tests where we check that a log file does not contain a given string, but we normally do this along with various "positive" assertions, such as that the log file does contain other strings or that a command succeeded.  Occasionally, though, these may even [follow](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/t/t-install.sh#L37-L41) a check that a command has failed.

What's unusual about this situation, though, is that we don't have any such positive assertions, and that if the command does unexpectedly fail, we'd like to check for a specific error message first before checking the command's exit status.  Otherwise we're making an assumption that the command failed for a specific reason which may not actually be the case.

So I'd suggest we adopt a modified version of a [pattern](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/t/t-custom-transfers.sh#L28-L36) that turns up in a few other shell tests, where we save the exit status from the command into a variable, then check the log for the a specific message, and then check if the exit status.  Something like this:

```suggestion
  res="${PIPESTATUS[0]}"

  [ 0 -eq "$(grep -c -F "Error reading client cert file" push.log)" ]

  [ 0 -eq "$res" ]
```

So first we confirm that there are zero instances of the string `Error reading client cert file` in the log file, and then we confirm that the command returned successfully with an exit status of zero.

One detail I should perhaps mention here is that we don't need to explicitly output the `push.log` file in the case of an error, because our test framework, combined with the use of the `tee(1)` command, will do that for us.  In fact, that's why we use the `tee` command: we get a log file local to the test which it can search for specific patterns, but the command's output is also written to `stdout`, which our test harness [captures](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/t/testlib.sh#L66-L71).  If a test fails, then the harness [outputs](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/t/testlib.sh#L110-L118) those captured logs (which are separate from any local log file created within the test).

So in the case of this test, with my suggested changes above, if it fails then we get diagnostic test output like the following, where the `stderr` section indicates that the `grep(1)` command failed and the `stdout` section includes the full output from the `git lfs push` command:

```
# -- stdout --
...
#     batch response: Error reading client cert file "/.../git-lfs_TEMP.9O8m3R7c6R/t-config.sh-72750/nonexistent/cert.pem": open /.../git-lfs_TEMP.9O8m3R7c6R/t-config.sh-72750/nonexistent/cert.pem: no such file or directory
#     Uploading LFS objects:   0% (0/1), 0 B | 0 B/s, done.
# -- stderr --
...
#     + git lfs push origin main
#     + tee push.log
#     + res=2
#     ++ grep -c -F 'Error reading client cert file' push.log
#     + '[' 0 -eq 1 ']'
#     + test_status=1
```

t/t-http-scheme-config.sh
```diff
@@ -0,0 +1,73 @@
...
+
+  # Push should fail trying to load the client cert (files don't exist), confirming
+  # that the https-scoped config is applied to the https:// connection.
+  git lfs push origin main 2>&1 | tee push.log
+  if [ "${PIPESTATUS[0]}" -eq 0 ]; then
+    echo >&2 "FAIL: expected 'git lfs push' to fail with a cert error"
+    exit 1
+  fi
+
+  grep -q "Error reading client cert file" push.log
```
chrisd8088, line 71: My only suggestion here is that we could add the `-F` option to the `grep(1)` command since we're just searching for a fixed string rather than a pattern.

Admittedly, this is an optimization we neglect to use in many of our shell tests, but it's something we might as well include whenever we can going forward.

chrisd8088 (commented):
Thank you so much for this PR, and for patiently waiting for comments on it from us!

As I worked through the review process I found the issues raised by these changes kept expanding the scope, because there are a lot of interrelated problems with our existing support for the `http.sslVerify`, `http.sslCAInfo`, and `http.sslCAPath` configuration options and their respective variations and associated environment variables.

I also started on a full rewrite of the `lfshttp/certs_test.go` file because I think it misses quite a few test cases, which in turn is part of the reason we haven't detected some of the problems with our legacy implementation.  Rather than writing complete Go test functions for each individual case, I think we'd be better off executing multiple test cases in a few functions.

Here's an example of how that might work to test a range of possible settings of the `http.sslVerify` / `http.<url>.sslVerify` option and the `GIT_SSL_NO_VERIFY` environment variable.  I've been developing another set for the `http.sslCAInfo` and `http.sslCAPath` options and all their variants and related environment variables, but I've been pulled off in other directions and haven't finished it yet.

<details><summary>Test Cases for SSL/TLS Verification</summary>

```go

type certSkipVerifyTestCase struct {
	url    string
	osEnv  map[string]string
	gitEnv map[string]string

	parsedURL     *url.URL
	httpClient    *http.Client
	httpTransport *http.Transport

	expected bool
}

func (c *certSkipVerifyTestCase) setup(t *testing.T) bool {
	var err error
	c.parsedURL, err = url.Parse(c.url)
	if err != nil {
		t.Error(fmt.Errorf("unable to parse URL: %w", err))
		return false
	}

	context := NewContext(nil, c.osEnv, c.gitEnv)

	client, err := NewClient(context)
	if err != nil {
		t.Error(fmt.Errorf("unable to create client: %w", err))
		return false
	}

	c.httpClient, err = client.HttpClient(c.parsedURL, creds.BasicAccess)
	if err != nil {
		t.Error(fmt.Errorf("unable to create HTTP client: %w", err))
		return false
	}

	var ok bool
	c.httpTransport, ok = c.httpClient.Transport.(*http.Transport)
	if !ok {
		t.Error(errors.New("unable to extract HTTP transport"))
		return false
	}

	return true
}

func (c *certSkipVerifyTestCase) AssertVerificationDisabledForURL(t *testing.T) {
	if ok := c.setup(t); !ok {
		return
	}

	assert.Equal(t, c.expected, c.httpTransport.TLSClientConfig.InsecureSkipVerify)
}

func TestCertVerificationDisabledForURL(t *testing.T) {
	for desc, c := range map[string]*certSkipVerifyTestCase{
		"no settings": {
			url:                "http://example.com/repo",
			expected: false,
		},
		"generic option set false matches host": {
			url: "http://example.com/repo",
			gitEnv: map[string]string{
				"http.sslverify": "false",
			},
			expected: true,
		},
		"generic option set false matches other host": {
			url: "http://otherhost.com/repo",
			gitEnv: map[string]string{
				"http.sslverify": "false",
			},
			expected: true,
		},
		"generic option set true matches host": {
			url: "http://example.com/repo",
			gitEnv: map[string]string{
				"http.sslverify": "true",
			},
			expected: false,
		},
		"specific option set false matches host": {
			url: "http://example.com/repo",
			gitEnv: map[string]string{
				"http.http://example.com/.sslverify": "false",
			},
			expected: true,
		},
		"specific option set false does not match other host": {
			url: "http://otherhost.com/repo",
			gitEnv: map[string]string{
				"http.http://example.com/.sslverify": "false",
			},
			expected: false,
		},
		"specific option set false overrides generic option": {
			url: "http://example.com/repo",
			gitEnv: map[string]string{
				"http.sslverify":                     "true",
				"http.http://example.com/.sslverify": "false",
			},
			expected: true,
		},
		"specific option set true overrides generic option": {
			url: "http://example.com/repo",
			gitEnv: map[string]string{
				"http.sslverify":                     "false",
				"http.http://example.com/.sslverify": "true",
			},
			expected: false,
		},
		"specific option set false matches scheme and host": {
			url: "https://example.com/repo",
			gitEnv: map[string]string{
				"http.https://example.com/.sslverify": "false",
			},
			expected: true,
		},
		"specific option set false matches host but not scheme": {
			url: "http://example.com/repo",
			gitEnv: map[string]string{
				"http.https://example.com/.sslverify": "false",
			},
			expected: false,
		},
		"specific option set false matches host and path": {
			url: "http://example.com/repo",
			gitEnv: map[string]string{
				"http.http://example.com/repo/.sslverify": "false",
			},
			expected: true,
		},
		"specific option set false matches host but not path": {
			url: "http://example.com/otherrepo",
			gitEnv: map[string]string{
				"http.http://example.com/repo/.sslverify": "false",
			},
			expected: false,
		},
		"empty variable overrides generic option set true": {
			url: "http://example.com/repo",
			gitEnv: map[string]string{
				"http.sslverify": "true",
			},
			osEnv: map[string]string{
				"GIT_SSL_NO_VERIFY": "",
			},
			expected: true,
		},
		"non-empty variable overrides generic option set true": {
			url: "http://example.com/repo",
			gitEnv: map[string]string{
				"http.sslverify": "true",
			},
			osEnv: map[string]string{
				"GIT_SSL_NO_VERIFY": "0",
			},
			expected: true,
		},
		"non-empty variable overrides specific option set true": {
			url: "http://example.com/repo",
			gitEnv: map[string]string{
				"http.http://example.com/.sslverify": "true",
			},
			osEnv: map[string]string{
				"GIT_SSL_NO_VERIFY": "1",
			},
			expected: true,
		},
		"empty variable overrides specific and generic options set true": {
			url: "http://example.com/repo",
			gitEnv: map[string]string{
				"http.sslverify":                     "true",
				"http.http://example.com/.sslverify": "true",
			},
			osEnv: map[string]string{
				"GIT_SSL_NO_VERIFY": "",
			},
			expected: true,
		},
	} {
		t.Run(desc, c.AssertVerificationDisabledForURL)
	}
}
```
</details>

In any case, please do take some time to read through my comments and see if anything stands out as incorrect or unhelpful, and many thanks again for all your help on our project!

PR #6215, opened 2026-03-09
lfs: fix Access denied on Windows Server 2019 in SmudgeToFile

### LINKED ISSUE #6157: Access denied during `git lfs pull` for some objects
**Describe the bug**

We updated our git lfs version to 3.7.1 and some of our repositories struggle with checking out LFS files on Windows. With git-lfs 3.7.0 there was no such problem.

**To Reproduce**

We have no influence on the following commands as this is done by the gitlab-runner custom executor we're relying on:

```powershell
$env:GIT_LFS_SKIP_SMUDGE="1"
git init C:/users/nice/desktop/mypkg --template C:\gitlab-runner\builds\mypkg.tmp\git-template
cd C:\users\nice\desktop\mypkg
git remote add origin https://internal-url.com/mypkg.git
git fetch origin --no-recurse-submodules +refs/heads/master --depth 20 --prune --quiet
git -c submodule.recurse=false checkout -f -q master
git clean -ffdx
git lfs pull
```

The `git lfs pull` command fails with:

```log
could not check out "tests/testdata/cs.foobar_clean-1.2.3-py3-none-any.whl": could not create working directory file "tests/testdata/cs.foobar_clean-1.2.3-py3-none-any.whl": open C:\Users\nice\Desktop\mypkg\tests\testdata\cs.foobar_clean-1.2.3-py3-none-any.whl: Access is denied.
```

The `.gitattributes` file looks like:

```
tests/testdata/*.egg filter=lfs diff=lfs merge=lfs -text
tests/testdata/*.whl filter=lfs diff=lfs merge=lfs -text
```
... and all files are properly tracked.

BTW: A second `git lfs pull` works just fine

**Expected behavior**

It should work just like it did with 3.7.0

**System environment**

```
Windows Server >= 2019
```

Windows defender exclusions are set to exclude the builds and mypkg directories

The same repositories do not fail when run on Windows 11 with the "same" setup.

**Output of `git lfs env`**
```
git-lfs/3.7.1 (GitHub; windows amd64; go 1.25.1; git b84b3384)
git version 2.52.0.windows.1

Endpoint=https://internal-url.com/mypkg.git/info/lfs (auth=none)
LocalWorkingDir=C:\Users\nice\Desktop\mypkg
LocalGitDir=C:\Users\nice\Desktop\mypkg\.git
LocalGitStorageDir=C:\Users\nice\Desktop\mypkg\.git
LocalMediaDir=C:\Users\nice\Desktop\mypkg\.git\lfs\objects
LocalReferenceDirs=
TempDir=C:\Users\nice\Desktop\mypkg\.git\lfs\tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=C:\Users\nice\Desktop\mypkg\.git\lfs
AccessDownload=none
AccessUpload=none
GIT_LFS_SKIP_SMUDGE=1
```

stanhu:
The security fix in v3.7.1 (CVE-2025-26625) revised SmudgeToFile to call os.Remove before os.OpenFile(O_EXCL) to prevent writing through symbolic and hard links. This delete-then-create sequence introduced a race on Windows Server 2019 where os.Remove marks the file as pending-deletion but the directory entry is not freed until all open handles are closed. If any external process (e.g. antivirus, search indexer) holds a transient handle to the file, the subsequent os.OpenFile(O_EXCL) call fails immediately with "Access is denied". Windows 11 and Server 2022 are unaffected because Go uses FILE_DISPOSITION_POSIX_SEMANTICS on those versions, which unlinks the name immediately regardless of open handles.

Fix this by writing LFS content to a short-named temporary file in the same directory and then renaming it over the destination. os.Rename uses MoveFileExW(MOVEFILE_REPLACE_EXISTING) on Windows, which atomically replaces the target path without a pending-deletion window. For regular files no prior Remove is needed. For symlinks and other irregular entries a Remove is still performed before the rename, since rename(2) and its Windows equivalent may follow a final symlink rather than replace it.

The temporary file pattern is kept intentionally short (.lfs-*) to avoid exceeding MAX_PATH on Windows in deep working trees.

Fixes #6157

### DIFF
--- a/lfs/gitfilter_smudge.go
+++ b/lfs/gitfilter_smudge.go
@@ -29,24 +29,65 @@ func (f *GitFilter) SmudgeToFile(path string, ptr *WrappedPointer, download bool
 		mode = stat.Mode().Perm()
 	}
 
-	if err := os.Remove(path); err != nil && !os.IsNotExist(err) {
-		return errors.Wrap(err, tr.Tr.Get("could not remove working directory file %q", path))
+	abs, err := filepath.Abs(path)
+	if err != nil {
+		return errors.New(tr.Tr.Get("could not produce absolute path for %q", path))
 	}
 
-	file, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_EXCL, mode)
+	// Write to a temp file in the same directory, then rename atomically.
+	// This avoids the Windows Server 2019 race where os.Remove() leaves
+	// a pending-deletion entry that blocks os.OpenFile(O_EXCL).
+	tmp, err := os.CreateTemp(filepath.Dir(abs), ".lfs-*")
 	if err != nil {
 		return errors.Wrap(err, tr.Tr.Get("could not create working directory file %q", path))
 	}
-	defer file.Close()
-	if _, err := f.Smudge(file, ptr.Pointer, ptr.Name, download, manifest, cb); err != nil {
+	tmpName := tmp.Name()
+	defer func() {
+		tmp.Close()
+		os.Remove(tmpName) // no-op if rename succeeded
+	}()
+
+	if err := tmp.Chmod(mode); err != nil {
+		return errors.Wrap(err, tr.Tr.Get("could not set permissions on temporary file"))
+	}
+
+	if _, err := f.Smudge(tmp, ptr.Pointer, ptr.Name, download, manifest, cb); err != nil {
 		if errors.IsDownloadDeclinedError(err) {
 			// write placeholder data instead
-			file.Seek(0, io.SeekStart)
-			ptr.Encode(file)
+			tmp.Seek(0, io.SeekStart)
+			ptr.Encode(tmp)
+			if closeErr := tmp.Close(); closeErr == nil {
+				os.Rename(tmpName, abs)
+			}
 			return err
-		} else {
-			return errors.New(tr.Tr.Get("could not write working directory file: %v", err))
 		}
+		return errors.New(tr.Tr.Get("could not write working directory file: %v", err))
+	}
+
+	if err := tmp.Close(); err != nil {
+		return errors.Wrap(err, tr.Tr.Get("could not close temporary file"))
+	}
+
+	// Remove symlinks/irregular entries before rename; regular files are
+	// replaced atomically by os.Rename without needing a prior Remove.
+	if lstat, _ := os.Lstat(abs); lstat != nil {
+		if lstat.Mode().IsRegular() {
+			// MoveFileExW(MOVEFILE_REPLACE_EXISTING) fails on Windows if the
+			// destination is read-only, as may be the case for locked LFS files.
+			if lstat.Mode().Perm()&0200 == 0 {
+				if err := os.Chmod(abs, lstat.Mode().Perm()|0200); err != nil {
+					return errors.Wrap(err, tr.Tr.Get("could not make working directory file writable %q", path))
+				}
+			}
+		} else {
+			if err := os.Remove(abs); err != nil && !os.IsNotExist(err) {
+				return errors.Wrap(err, tr.Tr.Get("could not remove working directory file %q", path))
+			}
+		}
+	}
+
+	if err := os.Rename(tmpName, abs); err != nil {
+		return errors.Wrap(err, tr.Tr.Get("could not replace working directory file %q", path))
 	}
 	return nil
 }


stanhu:
I'm going to close this because I think Go's `os.Remove` should be setting `FILE_DISPOSITION_POSIX_SEMANTICS`: https://github.com/golang/go/blob/e30e65f7a8bda0351d9def5a6bc91471bddafd3d/src/internal/syscall/windows/at_windows.go#L299-L303

lfs/gitfilter_smudge.go
```diff
@@ -29,25 +29,66 @@ func (f *GitFilter) SmudgeToFile(path string, ptr *WrappedPointer, download bool
...
+	abs, err := filepath.Abs(path)
+	if err != nil {
+		return errors.New(tr.Tr.Get("could not produce absolute path for %q", path))
 	}
 
-	file, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_EXCL, mode)
+	// Write to a temp file in the same directory, then rename atomically.
+	// This avoids the Windows Server 2019 race where os.Remove() leaves
+	// a pending-deletion entry that blocks os.OpenFile(O_EXCL).
+	tmp, err := os.CreateTemp(filepath.Dir(abs), ".lfs-*")
```
chrisd8088, line 40: If we ever do need to create temporary files, the ideal place for them is in our `.lfs/tmp` directory, because then there's no chance that we'll leave spurious files behind in the user's Git working tree if the Git LFS client is abruptly terminated (e.g., by a `SIGKILL` signal).

Then when another Git LFS command is run in the future, it will automatically [clean up](https://github.com/git-lfs/git-lfs/blob/f4ebf5fecd848f7a8e6e8f2e4c5cc5d775870bc5/git-lfs.go#L35) the `.lfs/tmp` directory by [removing](https://github.com/git-lfs/git-lfs/blob/f4ebf5fecd848f7a8e6e8f2e4c5cc5d775870bc5/fs/cleanup.go#L75-L79) any files more than an hour old.

chrisd8088 (commented):
Hey, thanks very much for this draft proposal!  I know you've closed the PR, but I thought I'd make a few notes here for my own future reference anyway.

The issue you've raised seems like a classic case of a [leaky abstraction](https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/), where Go tries to provide consistent file deletion behaviour on all systems, but some of the details of the underlying OS version affect its ability to do so.

I notice that our Windows CI jobs for this PR had some `Access is denied` [errors](https://github.com/git-lfs/git-lfs/actions/runs/22886982148/job/66401935336), so if we did want to pursue this approach we'd need to figure out what's causing those.

However, I think that if the Go standard libraries aren't updated in the manner you suggested in golang/go#78072, the way we should address this issue is to adopt the same technique used by Git itself.  In general, we try to follow Git's example wherever possible, even though we're using a different programming language.

As we noted in commit 5c11ffce9a4f095ff356bc781e2a031abb46c1a8, when the `git checkout` command is asked to update a specific pathspec (e.g., with a command such as `git checkout -- file.txt`), the `checkout_entry_ca()` function first [calls](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/entry.c#L577-L578) the `unlink(2)` system call, and then the `create_file()` function [invokes](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/entry.c#L577-L578) the `open(2)` system call with the `O_CREAT` and `O_EXCL` flags.

So we'd like to continue to follow that model, since it works very cleanly on most systems.

However, on Windows we may have to look a bit more deeply into Git's own abstraction of the underlying filesystem API.  Since the `unlink(2)` and `open(2)` system calls don't actually exist on Windows, Git [defines](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw-posix.h#L205-L207) [them](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw-posix.h#L214) as C macros which call the `mingw_unlink()` and `mingw_open()` functions, respectively.

Here we care about the `mingw_unlink()` function in particular, which first [tries](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L459-L460) a simple file deletion using the `DeleteFileW()` Windows API [function](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-deletefilew), and then if that doesn't succeed, enters a [loop](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L462-L480) which repeatedly attempts to delete the file (or other directory entry) using more complex steps.

The loop [sets](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L464) the file permissions to read-write, then [calls](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L465-L466) the `_wunlink()` C Runtime Library [function](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/unlink-wunlink) to try to delete the file again.  If it gets an error code which doesn't [match](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L165-L167) either `ERROR_ACCESS_DENIED` or `ERROR_SHARING_VIOLATION`, it [exits](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L467-L468) the loop and returns an error; otherwise, it [invokes](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L474-L475) the `_wrmdir()` CRL [function](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/rmdir-wrmdir) in case it's trying to delete a symbolic link to a directory.

We likely should not need to replicate all these bits of logic because the Go standard library's `os.Remove()` function [performs](https://github.com/golang/go/blob/0dc89195f9aece70476320be3fc9d6d657904056/src/os/file_windows.go#L230-L255) a similar series of steps, albeit in a slightly different order.  The `os.Remove()` function even [changes](https://github.com/golang/go/blob/0dc89195f9aece70476320be3fc9d6d657904056/src/os/file_windows.go#L248-L249) the file permissions and tries again to delete the file, much as Git's `mingw_unlink()` function does.

What's different about Git's abstraction from Go's abstraction is the loop.  Git's loop is [controlled](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L479-L480) by the `retry_ask_yes_no()` function, which allows for five additional attempts to delete the file with [delays](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L243) of 0, 1, 10, 20, and 40 milliseconds before each one.  If the file still can't be deleted, additional attempts may be made, [depending](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L260) on what the the `ask_yes_no_if_possible()` function returns.

The `ask_yes_no_if_possible()` function either [runs](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L218-L224) a program specified by the `GIT_ASK_YESNO` environment variable, or [prints](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L231-L234) a user-visible question and waits for an acceptable answer.  If the user chooses to not to retry again, then an error will be returned by the calling `mingw_unlink()` function.

If we were to replicate the looping behaviour in the Git LFS client, we'd want to do so in a new `RobustRemove()` function in our `tools` package, akin to the `RobustRename()` and `RobustOpen()` functions we already [define](https://github.com/git-lfs/git-lfs/blob/f4ebf5fecd848f7a8e6e8f2e4c5cc5d775870bc5/tools/robustio_windows.go#L19-L40) separately for Windows.  The Windows variants of these functions add loops around the Go standard library's `os.Rename()` and `os.Open()` functions, respectively, while the non-Windows Unix variants just [call](https://github.com/git-lfs/git-lfs/blob/f4ebf5fecd848f7a8e6e8f2e4c5cc5d775870bc5/tools/robustio.go#L8-L14) the standard library functions directly.

We could then simply change the `os.Remove()` [call](https://github.com/git-lfs/git-lfs/blob/f4ebf5fecd848f7a8e6e8f2e4c5cc5d775870bc5/lfs/gitfilter_smudge.go#L32) in our `lfs` package's `SmudgeToFile()` function to invoke `tools.RobustRemove()` instead, and that's the only alteration we'd need to make to the `SmudgeToFile()` function.  I think that's a significantly cleaner approach than what Claude proposed for this draft PR, and it would also be much more closely aligned with how Git works when performing a checkout on Windows.

PR #6221, opened 2026-03-16
tools: add RobustRemove() for Windows to handle file lock errors

### LINKED ISSUE #6157: Access denied during `git lfs pull` for some objects

stanhu:
Add RobustRemove() function that retries on ERROR_SHARING_VIOLATION and ERROR_ACCESS_DENIED, using the same delay pattern as Git on Windows (0, 1, 10, 20, 40 ms). This works around "Access is denied" errors during git lfs pull when antivirus or file indexers hold files open.

Closes #6157


tools/robustio_windows.go
```diff
@@ -16,6 +17,24 @@ func isEphemeralError(err error) bool {
...
 
+// isFileInUseError returns true if remove err may be resolved by waiting.
+// See https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L162-L171.
+func isFileInUseError(err error) bool {
+	return errors.Is(err, windows.ERROR_SHARING_VIOLATION) ||
+		errors.Is(err, windows.ERROR_ACCESS_DENIED)
+}
+
+// removeRetryOptions uses the same delay pattern as Git on Windows.
+// See https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L243.
```
chrisd8088, line 28: Thanks for updating the link in the comment to point to the corresponding part of Git's retry logic!

If this PR doesn't also implement the same `Unlink of file '%s' failed. Should I try again? (y/n)` [prompt](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L479-L480) that Git [asks](https://github.com/git/git/blob/7f19e4e1b6a3ad259e2ed66033e01e03b8b74c5e/compat/mingw.c#L231) the user when the automatic retries are exhausted, then we should expand our code comments to note that difference.

The same is true if we choose not to support the `GIT_ASK_YESNO` environment variable; we should document that decision in our comments.

tools/robustio_windows.go
```diff
@@ -16,6 +17,24 @@ func isEphemeralError(err error) bool {
...
+var removeRetryOptions = []retry.Option{
+	retry.Delay(0),
+	retry.Delay(1 * time.Millisecond),
+	retry.Delay(10 * time.Millisecond),
+	retry.Delay(20 * time.Millisecond),
+	retry.Delay(40 * time.Millisecond),
```
chrisd8088: This looks like such a clean implementation, I was really hoping it would work to mirror Git's behaviour!

Unfortunately, I don't think it will.

I ran a small test program which passes multiple `Delay()` option functions, and only the last one takes effect because the option functions are [run](https://github.com/avast/retry-go/blob/2.4.2/retry.go#L90-L92) in the order they are passed to the `Do()` function.  So the final `Delay()` option function [overwrites](https://github.com/avast/retry-go/blob/2.4.2/options.go#L48) the internal `delay` configuration setting that the previous option functions set to smaller values.  Here's my example program:

```go
package main

import (
	"errors"
	"fmt"
	"time"

	"github.com/avast/retry-go"
)

func main() {
	retry.Do(
		func() error {
			fmt.Println(time.Now().Format("15:04:05"))
			return errors.New("oops")
		},
		retry.Delay(0),
		retry.Delay(1 * time.Second),
		retry.Delay(5 * time.Second),
	)
}
```

In my example, this means the first retry delay is 5 seconds, not 0 or 1 seconds.  Then because the [default](https://github.com/avast/retry-go/blob/2.4.2/retry.go#L85) `delayType` configuration setting is `BackOffDelay`, each subsequent delay period is [twice](https://github.com/avast/retry-go/blob/2.4.2/options.go#L62) the previous one:

```
18:59:50
18:59:55
19:00:05
19:00:25
19:01:05
19:02:25
19:05:05
19:10:25
```

We use an older 2.4.2 version of the `github.com/avast/retry-go` package, so we could certainly upgrade to the latest 5.0.0 version if that was helpful.  I'm not sure it's necessary, though, as I think something more like this could work.  Here's another example program, and its output:

```go
package main

import (
	"errors"
	"fmt"
	"time"

	"github.com/avast/retry-go"
)

var delaySecs = [5]int{0, 1, 10, 20, 40}

func main() {
	retry.Do(
		func() error {
			fmt.Println(time.Now().Format("15:04:05"))
			return errors.New("oops")
		},
		retry.Attempts(6),
		retry.DelayType(func(n uint, config *retry.Config) time.Duration {
			return time.Duration(delaySecs[n]) * time.Second
		}),
	)
}
```

```
21:48:46
21:48:46
21:48:47
21:48:57
21:49:17
21:49:57
```

So here we set a maximum of 6 attempts, and provide our own anonymous `DelayTypeFunc` function to calculate the next delay interval.  That seems to work as we'd like (except that I'm using seconds in my examples, not milliseconds).

I'll suggest a change to the `RobustRemove()` function below based on this experiment.

tools/robustio_windows.go
```diff
@@ -38,3 +57,12 @@ func RobustOpen(name string) (*os.File, error) {
 		retry.LastErrorOnly(true),
 	)
 }
+
+func RobustRemove(path string) error {
+	return retry.Do(
+		func() error {
+			return os.Remove(path)
+		},
+		append(removeRetryOptions, retry.RetryIf(isFileInUseError))...,
```
chrisd8088: Here's what I would suggest instead, which I think will get us closer to what we would like:

```suggestion
		retry.RetryIf(isFileInUseError),
		retry.DelayType(func(n uint, config *retry.Config) time.Duration {
			return time.Duration(robustRemoveRetryDelaysMs[n]) * time.Millisecond
		}),
		retry.Attempts(uint(len(robustRemoveRetryDelaysMs)+1)),
		retry.LastErrorOnly(true),
```

with an array defined like this:

```go
var robustRemoveRetryDelaysMs = [5]int{0, 1, 10, 20, 40}
```

chrisd8088 (commented):
Hey, thanks for taking the time to tackle this enhancement!

I made some suggestions which I believe will be necessary, based on running a few tests.

Aside from that, I think it's OK for us to start with a simpler implementation which doesn't try to prompt the user the way Git does if all the automatic retry attempts fail.  We can always add that logic, and possibly also support for the `GIT_ASK_YESNO` environment variable, in a future PR.

Thanks again for trying to help improve the experience on older Windows systems!

tools/robustio_windows.go
stanhu: Thanks for looking at this in detail!

chrisd8088 (approved):
This looks great, thank you!

stanhu:
Thanks @chrisd8088! 

@git-lfs/releases Is there a timeline for the next git-lfs release?

chrisd8088:
> Is there a timeline for the next git-lfs release?

Yes, we're expecting to make a release shortly.  There are more details in my reply to #6253.  It won't be this week due to outside factors, so we are now targeting the first part of May.

PR #6222, opened 2026-03-17
Update Windows release signing action

chrisd8088:
The name of the `azure/trusted-signing-action` GitHub repository was recently changed to `azure/artifact-signing-action` in PR azure/artifact-signing-action#108, so we update our GitHub Actions release workflow to use the new name in our `build-windows` job, as the job would otherwise fail.

We also rename one of the input parameters we pass to the Action, as the same PR azure/artifact-signing-action#108 deprecated the parameter name `trusted-signing-account-name` in favour of the simpler term `signing-account-name`.

Finally, we update the value of the `endpoint` input parameter we pass to the Action, as our Windows Artifact Signing account and certificate profile have apparently moved to a new Azure region.

Note that this PR's changes were tested successfully in a private repository.

Many thanks to @divz44 and @padraignix for their help diagnosing the region change!

### DIFF
--- a/tools/filetools.go
+++ b/tools/filetools.go
@@ -16,7 +16,6 @@ import (
 	"sync/atomic"
 
 	"github.com/git-lfs/git-lfs/v3/errors"
-	"github.com/git-lfs/git-lfs/v3/filepathfilter"
 	"github.com/git-lfs/git-lfs/v3/tr"
 )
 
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -58,13 +58,13 @@ jobs:
...
         FORCE_LOCALIZE: true
-    - uses: azure/trusted-signing-action@v1.1.0
+    - uses: azure/artifact-signing-action@v1.1.0
       with:
         azure-tenant-id: ${{ secrets.SPN_GIT_LFS_SIGNING_TENANT_ID }}
         azure-client-id: ${{ secrets.SPN_GIT_LFS_SIGNING_CLIENT_ID }}
         azure-client-secret: ${{ secrets.SPN_GIT_LFS_SIGNING }}
-        endpoint: https://wus.codesigning.azure.net/
-        trusted-signing-account-name: GitHubInc
+        endpoint: https://wus3.codesigning.azure.net/
```
larsxschneider, line 69: Maybe add a note here that this aligns to the region where the Artifact Signing Account was created.

.github/workflows/release.yml
```diff
@@ -58,13 +58,13 @@ jobs:
       shell: bash
       env:
         FORCE_LOCALIZE: true
-    - uses: azure/trusted-signing-action@v1.1.0
+    - uses: azure/artifact-signing-action@v1.1.0
```
larsxschneider, line 61: FYI 1.2.0 was just released. No urgent reasons to upgrade, though.

.github/workflows/release.yml
chrisd8088: Sure thing, that's a good idea.

.github/workflows/release.yml
chrisd8088: Thanks!  I think I'll leave it for now and let Dependabot figure it out for us, if it's important.

PR #6228, opened 2026-03-24
Fix `--remote` option when a git push remote is set

markshep-wbg:
When running `git lfs` commands in a local branch that has a push remote set that's different from the fetch remote the `--remote` option gets ignored and the push remote is always queried.  For example if you have a repository which is configured like this:

```
$ git config branch.$(git branch --show-current).remote
origin
$ git config branch.$(git branch --show-current).pushRemote
preview
```
Then the command `git lfs locks --remote origin` will actually query the `preview` remote for the locks rather than the requested `origin` remote.

This PR fixes that bug so that the `--remote` option is honoured for the `lock`, `locks` and `unlock` commands.


chrisd8088:
Hey, thanks very much for identifying this issue and providing a potential fix!

Looking back through our history, it appears that our locking commands were not updated back in PR #2715 when support for the `branch.<name>.pushRemote` Git configuration options was first added to Git LFS.  (The locking commands themselves predate that PR, having been introduced in PR #1256.)

My only suggestion for this PR is that it would be ideal if we could add some tests which exercise the `--remote` option of the three locking commands, with and without various Git configuration option settings.

For example, here's a test we could add to our `t/t-locks.sh` shell script which should only pass with the change to the `git lfs locks` command from your PR:

```shell
begin_test "list a single lock (--remote overrides push default)"
(
  set -e

  reponame="locks-list-remote"
  setup_remote_repo_with_file "$reponame" "f.dat"

  git lfs lock --json "f.dat" | tee lock.log

  id=$(assert_lock lock.log f.dat)
  assert_server_lock "$reponame" "$id" "refs/heads/main"

  git remote add bad-remote "invalid-url"
  git config remote.pushDefault bad-remote

  git lfs locks 2>&1 | tee locks.log
    echo >&2 "fatal: expected 'git lfs locks' to fail ..."
    exit 1
  fi

  git lfs locks --remote origin --path "f.dat" | tee locks.log

  [ 1 -eq "$(wc -l <locks.log)" ]
  grep "f.dat" locks.log
  grep "Git LFS Tests" locks.log
)
end_test
```

Would you be willing to add a test similar to this to each of the `t/t-lock.sh`, `t/t-locks.sh`, and `t/t-unlock.sh` scripts?  If not, that's OK; I can try to write some tests and then if you're able to grant me permission, I could push them into this PR's branch.

Thank you again very much for taking the time to help improve our project!

chrisd8088 (approved):
Thanks for this PR!  I'm going to approve it, as the changes to the Git LFS client itself look good to me.

If we can add some tests to it to demonstrate that the changes are effective, that would be even better, and then we can kick off the CI suite and make sure it passes.

chrisd8088:
I hope you don't mind; I went ahead and added four new tests I wrote to this PR so that our test suite now exercises the `--remote` option of the locking commands and checks that it is preferred to any remote specified with `branch.<name>.pushRemote` or `remote.pushDefault`.

larsxschneider (approved):

PR #6231, opened 2026-03-31
Fix infinite loop in LFS requests rejected with 401

dmach:
The infinite loop was caused by a recursive call in lfsapi/auth.go:DoWithAuth. When an LFS request was rejected with a 401 but didn't have an Authorization header, the function would call itself recursively. Because the count variable was local to the function, it was reset to 0 on every recursion, effectively bypassing the defaultMaxAuthAttempts limit. In cases where the user lacked permissions (401 response despite valid credentials), git-lfs would continuously "resubmit" the request.

Fix:
 1. Converted the recursion in DoWithAuth into a for loop.
 2. Persisted the count variable across retry attempts.
 3. Added logic to increment the counter if a retry is attempted with the same access mode, ensuring the process terminates after defaultMaxAuthAttempts (3 attempts) even if the server continues to challenge.

### DIFF
--- a/lfsapi/auth.go
+++ b/lfsapi/auth.go
@@ -25,21 +25,31 @@ var (
 // supporting basic authentication.
 func (c *Client) DoWithAuth(remote string, access creds.Access, req *http.Request) (*http.Response, error) {
 	count := 0
-	res, err := c.doWithAuth(remote, &count, access, req, nil)
+	for {
+		oldCount := count
+		res, err := c.doWithAuth(remote, &count, access, req, nil)
 
-	if errors.IsAuthError(err) {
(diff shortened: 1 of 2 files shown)


chrisd8088:
Hey, thanks very much for diagnosing this issue and proposing this PR!

As it happens, we have PR #6018 in progress to deal with this issue, and it just went through another round of reviews yesterday, as it happens.  We should be closing in on a comprehensive fix for this particular concern soon.

So if you don't mind, I'll close this PR as a duplicate, but we really appreciate your help trying to improve Git LFS!

dmach:
:+1: Closing as duplicate was definitely fine, I should have looked for duplicates first.
I'm glad this is getting fixed.

PR #6234, opened 2026-04-07
Improve throughput by pipelining batch API calls

joshfriend:
Previously, each batch of LFS objects had to fully download before the next batch API call was made (`lfs.transfer.batchSize` let users increase the number of objects per batch. but GitHub has a hard cap of 100). I measured ~400ms of time per batch API call, and download workers spent significant time idle waiting for the next batch response. This PR pipelines batch API calls using a pool of parallel workers so that downloads from previous batches run concurrently with new API requests.

Additionally, the default concurrent transfers is now scaled to `3 * NCPU` (min 8) instead of the hardcoded default of 8 from 2017. A bug is also fixed where an empty `"transfer"` field in the batch API response (which per spec means "use default") was treated as a name mismatch, causing all download workers to be torn down and recreated every batch cycle.

Tested on a repo with ~19,400 LFS objects (~20KB avg, ~900MB total, GitHub hosted) on a 16-core M3 macbook pro: network throughput increased from ~1.8 MB/s to ~18 MB/s (~10x) and wall time dropped from 6m 4s to 48s (~7.5x) for `git lfs pull` after nuking `.git/lfs/objects/*`.

I added `lfs.concurrentbatchrequests` (default 4) to control the number of parallel batch API workers. In testing, 3-4 was the sweet spot for my laptop and kept the download workers busy, though I needed 8 in a server environment.

#### Cloud worker (192 CPUs)

  I did a benchmark sweep of git-lfs v3.7.1 and this patch with various settings on a large cloud worker.

  ##### Current git-lfs v3.7.1 (serial batch API)

  | concurrenttransfers |   Time | vs default |
  |--------------------:|-------:|-----------:|
  |          8 (default)| 511.6s |   baseline |
  |                  32 | 245.5s |      2.1x  |
  |                  96 | 191.0s |      2.7x  |
  |                 192 | 179.0s |      2.9x  |

  Bumping transfer concurrency alone hits diminishing returns fast, the serial batch API calls are the bottleneck, not the download workers.

  ##### This PR (pipelined batch API)

  | transfers | batch workers |   Time | vs v3.7.1 default | vs best v3.7.1 |
  |----------:|--------------:|-------:|-------------------:|---------------:|
  |         8 |             1 | 346.3s |              1.5x  |          0.5x  |
  |         8 |             4 | 323.8s |              1.6x  |          0.6x  |
  |        32 |             4 |  79.1s |              6.5x  |          2.3x  |
  |        96 |             4 |  27.3s |             18.7x  |          6.6x  |
  |        96 |             8 |  26.8s |             19.1x  |          6.7x  |
  |       192 |             4 |  25.3s |             20.2x  |          7.1x  |
  |       192 |             8 |  14.6s |             35.0x  |         12.3x  |
  |       288 |             8 |  14.3s |             35.8x  |         12.5x  |
  |       384 |             8 |  14.0s |             36.5x  |         12.8x  |
  |       384 |            12 |  11.7s |             43.7x  |         15.3x  |
  |       384 |            16 |   9.0s |             56.8x  |         19.9x  |
  |       576 |      4 (def.) |  26.8s |             19.1x  |          6.7x  |
  |       576 |             8 |  15.0s |             34.1x  |         11.9x  |
  |       576 |            16 |   9.5s |             53.9x  |         18.8x  |

  Best: **9.0s** at transfers=384, batch=16, **56.8x** faster than stock.
  Transfer concurrency plateaus above ~192; batch workers scale nearly linearly.

  #### Local machine (M3 MacBook Pro, 16 CPUs, baseline 364s with v3.7.1)

  | transfers | batch workers |  Time | vs v3.7.1 |
  |----------:|--------------:|------:|-------------------:|
  |         8 |             1 |  331s |              1.1x  |
  |         8 |             4 |  306s |              1.2x  |
  |        16 |             4 |  147s |              2.5x  |
  |        32 |             4 |   70s |              5.2x  |
  |        48 |             4 |   48s |              7.6x  |
  |        48 |             8 |   50s |              7.3x  |
  |        64 |             4 |   36s |             10.1x  |
  |        64 |             8 |   37s |              9.8x  |
  |        96 |             4 |   29s |             12.6x  |
  |        96 |             8 |   29s |             12.6x  |
  |        96 |            16 |   33s |             11.0x  |

  Best: **29s** at transfers=96, batch=4 or 8, **12.6x** faster than v3.7.1.
  Batch workers beyond 4 show no benefit.

  ---

  ### ios repo — 65,387 objects, 7.3 GB

  #### Cloud worker (192 CPUs)

  I also tested this PR against our larger iOS repo, using batch worker concurrency of 1 to somewhat mimic the current state.

  | transfers | batch workers |    Time | vs baseline |
  |----------:|--------------:|--------:|------------:|
  |         8 |             1 | 1139.8s |    baseline |
  |         8 |             4 |  996.7s |       1.1x  |
  |        32 |             4 |  214.4s |       5.3x  |
  |        96 |             4 |   71.5s |      15.9x  |
  |       192 |             4 |   57.3s |      19.9x  |
  |       576 |      4 (def.) |  157.9s |       7.2x  |
  |       576 |             1 |  206.7s |       5.5x  |
  |       576 |             8 |  113.9s |      10.0x  |

  Best: **57.3s** at transfers=192, batch=4, **19.9x** faster than baseline.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ lfstest-*
 !lfstest-*.go
 
 vendor/
+.idea/
--- a/docs/man/git-lfs-config.adoc
+++ b/docs/man/git-lfs-config.adoc
@@ -139,6 +139,12 @@ occurs.
 +
 The number of concurrent uploads/downloads. Default is 3x the number of
 logical CPUs, with a minimum of 8.
+* `lfs.concurrentbatchrequests`
++
+The number of concurrent batch API requests to make when fetching download
+URLs from the server. Higher values keep download workers saturated when
+there are many objects to transfer. Default scales with `lfs.concurrenttransfers`
+(1 per 12 concurrent transfers, minimum 4).
 * `lfs.basictransfersonly`
 +
 If set to true, only basic HTTP upload/download transfers will be used,
--- a/tq/adapterbase.go
+++ b/tq/adapterbase.go
@@ -104,24 +104,29 @@ type job struct {
 
 	results chan<- TransferResult
 	wg      *sync.WaitGroup
+	done    *sync.WaitGroup
 }
 
 func (j *job) Done(err error) {
 	j.results <- TransferResult{j.T, err}
 	j.wg.Done()
+	j.done.Done()
 }
 
 func (a *adapterBase) Add(transfers ...*Transfer) <-chan TransferResult {
 	results := make(chan TransferResult, len(transfers))
 
+	// done tracks when this batch's jobs complete, independent of
+	// other batches. jobWait tracks all jobs globally for End().
+	var done sync.WaitGroup
+	done.Add(len(transfers))
 	a.jobWait.Add(len(transfers))
 
 	go func() {
 		for _, t := range transfers {
-			a.jobChan <- &job{t, results, a.jobWait}
+			a.jobChan <- &job{T: t, results: results, wg: a.jobWait, done: &done}
 		}
-		a.jobWait.Wait()
-
+		done.Wait()
 		close(results)
 	}()
 
--- a/tq/manifest.go
+++ b/tq/manifest.go
@@ -22,6 +22,7 @@ type Manifest interface {
 	MaxRetries() int
 	MaxRetryDelay() int
 	ConcurrentTransfers() int
+	ConcurrentBatchRequests() int
 	IsStandaloneTransfer() bool
 	batchClient() BatchClient
 	GetAdapterNames(dir Direction) []string
@@ -70,6 +71,10 @@ func (m *lazyManifest) MaxRetryDelay() int {
 
 func (m *lazyManifest) ConcurrentTransfers() int {
 	return m.Upgrade().ConcurrentTransfers()
+}
+
+func (m *lazyManifest) ConcurrentBatchRequests() int {
+	return m.Upgrade().ConcurrentBatchRequests()
 }
 
 func (m *lazyManifest) IsStandaloneTransfer() bool {
@@ -138,6 +143,7 @@ type concreteManifest struct {
 	maxRetries              int
 	maxRetryDelay           int
 	concurrentTransfers     int
+	concurrentBatchRequests int
 	basicTransfersOnly      bool
 	standaloneTransferAgent string
 	tusTransfersAllowed     bool
@@ -164,6 +170,10 @@ func (m *concreteManifest) MaxRetryDelay() int {
 
 func (m *concreteManifest) ConcurrentTransfers() int {
 	return m.concurrentTransfers
+}
+
+func (m *concreteManifest) ConcurrentBatchRequests() int {
+	return m.concurrentBatchRequests
 }
 
 func (m *concreteManifest) IsStandaloneTransfer() bool {
@@ -225,6 +235,9 @@ func newConcreteManifest(f *fs.Filesystem, apiClient *lfsapi.Client, operation, 
 		if v := git.Int("lfs.concurrenttransfers", 0); v > 0 {
 			m.concurrentTransfers = v
 		}
+		if v := git.Int("lfs.concurrentbatchrequests", 0); v > 0 {
+			m.concurrentBatchRequests = v
+		}
 		m.basicTransfersOnly = git.Bool("lfs.basictransfersonly", false)
 		m.standaloneTransferAgent = findStandaloneTransfer(
 			apiClient, operation, remote,
@@ -243,7 +256,6 @@ func newConcreteManifest(f *fs.Filesystem, apiClient *lfsapi.Client, operation, 
 	if m.concurrentTransfers < 1 {
 		m.concurrentTransfers = lfshttp.DefaultConcurrentTransfers()
 	}
-
 	if sshTransfer != nil {
 		if !useSSHMultiplexing {
 			m.concurrentTransfers = 1
@@ -253,6 +265,22 @@ func newConcreteManifest(f *fs.Filesystem, apiClient *lfsapi.Client, operation, 
 		m.batchClientAdapter = &SSHBatchClient{
 			maxRetries: m.maxRetries,
 			transfer:   sshTransfer,
+		}
+		// SSH batch requests are serialized through a single
+		// connection, so parallel batch API workers don't help.
+		m.concurrentBatchRequests = 1
+	}
+
+	if m.concurrentBatchRequests < 1 {
+		// Scale batch API workers with transfer concurrency so the
+		// download pipeline stays saturated. This ratio was benchmarked
+		// across large cloud workers and local laptops.
+		m.concurrentBatchRequests = (m.concurrentTransfers + 11) / 12
+		if m.concurrentBatchRequests < 4 && m.concurrentTransfers >= 4 {
+			m.concurrentBatchRequests = 4
+		}
+		if m.concurrentBatchRequests < 1 {
+			m.concurrentBatchRequests = 1
 		}
 	}
 
--- a/tq/transfer.go
+++ b/tq/transfer.go
@@ -244,7 +244,7 @@ type Adapter interface {
 	// or more Add calls. The passed in callback will receive updates on progress.
 	Begin(cfg AdapterConfig, cb ProgressCallback) error
 	// Add queues a download/upload, which will complete asynchronously and
-	// notify the callbacks given to Begin()
+	// whose results will be sent on the returned channel.
 	Add(transfers ...*Transfer) (results <-chan TransferResult)
 	// Indicate that all transfers have been scheduled and resources can be released
 	// once the queued items have completed.
--- a/tq/transfer_queue.go
+++ b/tq/transfer_queue.go
@@ -252,6 +252,7 @@ type objectTuple struct {
 	Missing         bool
 	ReadyTime       time.Time
 	retryLaterTime  time.Time
+	lastErr         error
 }
 
 func (o *objectTuple) ToTransfer() *Transfer {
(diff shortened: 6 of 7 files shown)


chrisd8088:
Thank you very much for this PR!  It's very exciting to see performance improvements of this magnitude!

May I ask one favour, before we dig into a full review?  I think it would be quite beneficial if we could separate out both the bug fix for the `transfer` Batch API element and the change to the default value of the `lfs.concurrentTransfers` configuration option into separate commits, or even better, into separate PRs.  That will really help us analyze the changes specific to each of them (including documentation and test updates) from the implementation of the new "pipelining" functionality.

The other thing we'll likely try to include are some tests, if we can work out the best way to implement those.  It's not always straightforward or even possible, of course, but anything we can come up with would be appreciated.  The `t/t-batch-transfer-size.sh` [script](https://github.com/git-lfs/git-lfs/blob/dd3714055863a3eb158950e75f34c8e352eaf9b8/t/t-batch-transfer-size.sh) might provide a partial template, for instance.

As well, I believe the tests in our `t/t-env.sh` script are going to need modification in the same commit or PR where we change the default `lfs.concurrentTransfers` value to be variable based on the number of CPUs, because these tests [expect](https://github.com/git-lfs/git-lfs/blob/dd3714055863a3eb158950e75f34c8e352eaf9b8/t/t-env.sh#L42) a value of `8` for the `ConcurrentTransfers` field in the output of `git lfs env` commands.

joshfriend:
> I think it would be quite beneficial if we could separate out both the bug fix for the transfer Batch API element and the change to the default value of the lfs.concurrentTransfers configuration option into separate commits, or even better, into separate PRs

Sure, that's reasonable to request, I'll make a new PR with only the bugfix I found. I tested performance with and without that fix and found it to be inconsequential so I did not bother initially, but I'm happy to do so.

Updating the behavior of `concurrentTransfers` can be a separate PR too. That will likely have some performance benefits on it's own, as you see above I did already benchmark the current impl with higher count of concurrent transfers.

> The other thing we'll likely try to include are some tests

Of course 😄, I'll see what I can come up with.

joshfriend:
Bugfix PR broken out: https://github.com/git-lfs/git-lfs/pull/6240

joshfriend:
concurrentTransfers default separated out: https://github.com/git-lfs/git-lfs/pull/6241

chrisd8088:
Thanks very much for crafting the separate PRs!

joshfriend:
I've rebased this on top of https://github.com/git-lfs/git-lfs/pull/6241

New changes:

* batch api workers now scale relative to concurrenttransfers automatically (1:12). I found this to be the optimal ratio for both the test cloud environment and my laptop in the two repos tested.

I also tried dynamically scaling up batch requests by detecting download transfer starvation, but it was not any faster than just using a simple ratio so I kept the simple solution 😄 

Moving to draft state for a bit while I make sure tests pass and do code cleanup

JeremieA:
Hello, thanks for this work on making LFS faster ! I'm also trying various ways to improve performances (#6242), and as part of that work I ran [our LFS benchmarks](https://github.com/InSimo/lfs-experiments) with this PR (the version before your recent rebases, i.e. with b5aeacd439be1a01cee5cb371ff12f78b03898bd). I encountered a crash:
```
fatal error: concurrent map writes

goroutine 23 gp=0xc00028c700 m=12 mp=0xc000101008 [running]:
runtime.fatal({0xaf8785, 0x15})
        /go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.13.linux-amd64/src/runtime/panic.go:1123 +0x5c fp=0xc00053d0f0 sp=0xc00053d0c0 pc=0x43dddc
internal/runtime/maps.fatal({0xaf8785?, 0xa6c684c?})
        /go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.13.linux-amd64/src/runtime/panic.go:1058 +0x18 fp=0xc00053d110 sp=0xc00053d0f0 pc=0x473e18
runtime.mapassign_faststr(0xa4f6c0, 0xc000220750, {0xc000251260, 0x2c})
        /go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.13.linux-amd64/src/internal/runtime/maps/runtime_faststr_swiss.go:273 +0x65 fp=0xc00053d1c8 sp=0xc00053d110 pc=0x40c785
github.com/git-lfs/git-lfs/v3/lfshttp.(*sshCache).Resolve(0xc000254348, {{0xc000251230, 0x2f}, {{0xc0003ec190, 0xe}, {0x0, 0x0}, {0x7ffcd2958d2c, 0x14}}, {0xaeecf4, ...}, ...}, ...)
        lfs-experiments/git-lfs/lfshttp/ssh.go:51 +0x507 fp=0xc00053d4b0 sp=0xc00053d1c8 pc=0x8b3567
github.com/git-lfs/git-lfs/v3/lfshttp.(*Client).sshResolveWithRetries(0xc000230140, {{0xc000251230, 0x2f}, {{0xc0003ec190, 0xe}, {0x0, 0x0}, {0x7ffcd2958d2c, 0x14}}, {0xaeecf4, ...}, ...}, ...)
        lfs-experiments/git-lfs/lfshttp/client.go:252 +0x40e fp=0xc00053d748 sp=0xc00053d4b0 pc=0x8ae3ae
github.com/git-lfs/git-lfs/v3/lfshttp.(*Client).NewRequest(0xc000230140, {0xaee567, 0x4}, {{0xc000251230, 0x2f}, {{0xc0003ec190, 0xe}, {0x0, 0x0}, {0x7ffcd2958d2c, ...}}, ...}, ...)
        lfs-experiments/git-lfs/lfshttp/client.go:122 +0x1b8 fp=0xc00053d8f8 sp=0xc00053d748 pc=0x8ad098
github.com/git-lfs/git-lfs/v3/lfsapi.(*Client).NewRequest(...)
        lfs-experiments/git-lfs/lfsapi/client.go:13
github.com/git-lfs/git-lfs/v3/tq.(*tqClient).Batch(0xc00002cff0, {0x7ffcd2958d18, 0x28}, 0xc000257680)
        lfs-experiments/git-lfs/tq/api.go:82 +0x1ff fp=0xc00053db20 sp=0xc00053d8f8 pc=0x8c3fbf
github.com/git-lfs/git-lfs/v3/tq.Batch({0xc0c310, 0xc00022c700}, 0x0, {0x7ffcd2958d18, 0x28}, 0xc0001b0d50, {0xc0001ae0a0, 0x13, 0x13})
        lfs-experiments/git-lfs/tq/api.go:46 +0x272 fp=0xc00053dba8 sp=0xc00053db20 pc=0x8c3cf2
github.com/git-lfs/git-lfs/v3/tq.(*TransferQueue).enqueueAndCollectRetriesFor(0xc0000fce00, {0xc0002ac008, 0x13, 0x0?})
        lfs-experiments/git-lfs/tq/transfer_queue.go:665 +0x2db fp=0xc00053df10 sp=0xc00053dba8 pc=0x8da81b
github.com/git-lfs/git-lfs/v3/tq.(*TransferQueue).collectBatches.func2()
        lfs-experiments/git-lfs/tq/transfer_queue.go:472 +0xe7 fp=0xc00053dfe0 sp=0xc00053df10 pc=0x8da147
runtime.goexit({})
        /go/pkg/mod/golang.org/toolchain@v0.0.1-go1.24.13.linux-amd64/src/runtime/asm_amd64.s:1700 +0x1 fp=0xc00053dfe8 sp=0xc00053dfe0 pc=0x47b9c1
created by github.com/git-lfs/git-lfs/v3/tq.(*TransferQueue).collectBatches in goroutine 15
        lfs-experiments/git-lfs/tq/transfer_queue.go:469 +0x18b
```
This is on scenarios using HTTP transport for LFS (with Forgejo or Gitea on the server-side), but ssh for git (including the LFS authentication with git-lfs-authenticate). A possible fix is to use either a `RWLock` or a `sync.Map` to wrap `endpoints` accesses in `sshCache`:
```
diff --git a/lfshttp/ssh.go b/lfshttp/ssh.go
index bc973aa5..16d24d16 100644
--- a/lfshttp/ssh.go
+++ b/lfshttp/ssh.go
@@ -4,6 +4,7 @@ import (
        "bytes"
        "encoding/json"
        "strings"
+       "sync"
        "time"
 
        "github.com/git-lfs/git-lfs/v3/config"
@@ -18,14 +19,11 @@ type SSHResolver interface {
 }
 
 func withSSHCache(ssh SSHResolver) SSHResolver {
-       return &sshCache{
-               endpoints: make(map[string]*sshAuthResponse),
-               ssh:       ssh,
-       }
+       return &sshCache{ssh: ssh}
 }
 
 type sshCache struct {
-       endpoints map[string]*sshAuthResponse
+       endpoints sync.Map // map[string]*sshAuthResponse
        ssh       SSHResolver
 }
 
@@ -35,7 +33,9 @@ func (c *sshCache) Resolve(e Endpoint, method string) (sshAuthResponse, error) {
        }
 
        key := strings.Join([]string{e.SSHMetadata.UserAndHost, e.SSHMetadata.Port, e.SSHMetadata.Path, method}, "//")
-       if res, ok := c.endpoints[key]; ok {
+
+       if val, ok := c.endpoints.Load(key); ok {
+               res := val.(*sshAuthResponse)
                if _, expired := res.IsExpiredWithin(5 * time.Second); !expired {
                        tracerx.Printf("ssh cache: %s git-lfs-authenticate %s %s",
                                e.SSHMetadata.UserAndHost, e.SSHMetadata.Path, endpointOperation(e, method))
@@ -48,7 +48,7 @@ func (c *sshCache) Resolve(e Endpoint, method string) (sshAuthResponse, error) {
 
        res, err := c.ssh.Resolve(e, method)
        if err == nil {
-               c.endpoints[key] = &res
+               c.endpoints.Store(key, &res)
        }
        return res, err
 }
```

joshfriend:
@JeremieA thanks for the additional testing ❤️

I had found one other race bug related to ssh just from the existing CI tests (`Upgrade()` in manifest.go), but yours was new. I added a new commit with a test that duplicated the race when using `go test -race`. 

Once you had patched the race, how did this pipeline optimization perform in your benchmarks?

@chrisd8088 I moved the concurrency bug fixes to a new pr (https://github.com/git-lfs/git-lfs/pull/6243) and cherry-picked the commit back here as well.

chrisd8088:
> I moved the concurrency bug fixes to a new pr (https://github.com/git-lfs/git-lfs/pull/6243) and cherry-picked the commit back here as well.

Great, thank you, that's much appreciated.

JeremieA:
> Once you had patched the race, how did this pipeline optimization perform in your benchmarks?

I had initial runs with around 10% speedup, which is significant (as in my benchmarks everything is local, so there is no network latency to hide, I'm essentially measuring the "light-speed" efficiency of the full stack). However I then tried to run the full experiment and it was either deadlocking somewhere or making the tests very slow, hitting the timeouts. I will investigate further later this week to see if I can see where it is coming from (it could be on the server-side that there are contentions when more than 100 objects are checked in parallel)

joshfriend:
@JeremieA I found another deadlock, workers that finished processing a batch couldn't signal completion because the collector only checked for completion signals during shutdown, so they froze before picking up the next batch, which blocked the collector from dispatching more work. I think your local "light-speed" test setup hit the problem pretty easily, I've now reproduced what I think is a similar setup in a test.

PR #6235, opened 2026-04-08
Run git lfs install as original user when using sudo

### LINKED ISSUE #6230: Using `sudo install.sh` (on MacOS) corrupts permissions on local `.gitconfig`
**Describe the bug**

There's a permission conflict between the permissions required to install `git-lfs` into `/usr/local` and to update the local account with git-lfs support.

Assuming a MacOS Unix environment:

The obvious way to install git-lfs to /usr/local is to run `sudo ./install.sh`.  All well and good.  However, at the end this "helpfully" runs `git lfs install`.  But because we used sudo to install the files, the "install" command is run as root.  This updates the ownership of `~/.gitconfig` to `root:staff` and permissions `rx-r-----`.

In other words, running `install.sh` as root (good for writing the files to `/usr/local` _also_ turns my local `.gitconfig` into a read-only file from my user.

**To Reproduce**
1. Unpack git release 3.7.1 Mac (Apple silicon)
2. From terminal, run `sudo ./install.sh`

**Expected behavior**

* Installs binaries in `/usr/local'.
* Doesn't try to configure "git lfs" for root.

**Actual behavior**

* Installs binaries in `/usr/local'.
* Configures user `.gitconfig` with root permission settings.

I share some blame for this situation in that section "From binary" of "README.md" says that it includes a script that will "Install Git LFS binaries onto the system `$PATH`." and "Run `git lfs install` to perform required global configuration changes.".  I mis-read this as saying that the install script would install the binaries and then I would need to run `git lfs install` as per "Step 1" of "getting started" on the "git-lfs.com" homepage.

However, no other tool is provided in the binary package to perform the binary install.

**System environment**

MacOS Tahoe 26.3, running Terminal.

**Proposed Fix**

If `./install.sh` is run _without_ the `-local` flag, do not attempt to perfom a `git lfs install` within the install script as the user and install environments are likely not compatible.

utafrali:
When install.sh runs with sudo on macOS, git lfs install was executing as root and messing up .gitconfig ownership. Fixed by checking for SUDO_USER and running the command as the original user. Fixes #6230.

### DIFF
--- a/script/install.sh
+++ b/script/install.sh
@@ -38,4 +38,8 @@ pushd "$( dirname "${BASH_SOURCE[0]}" )" > /dev/null
 popd > /dev/null
 
 PATH+=:"$prefix/bin"
-git lfs install
+if [ -n "${SUDO_USER:-}" ]; then
+  sudo -u "$SUDO_USER" git lfs install
+else
+  git lfs install
+fi


script/install.sh
```diff
@@ -38,4 +38,8 @@ pushd "$( dirname "${BASH_SOURCE[0]}" )" > /dev/null
 popd > /dev/null
 
 PATH+=:"$prefix/bin"
-git lfs install
+if [ -n "${SUDO_USER:-}" ]; then
+  sudo -u "$SUDO_USER" git lfs install
```
chrisd8088, line 42: I'm afraid this approach isn't going to function as we need it to in all conditions.

On a system where the user's `HOME` is retained by `sudo(8)`, this would likely be OK, because the user-specific configuration file that `git config --global` will modify likely to be the one belonging to the user who ran `sudo` and not the `root` user.

However, on systems where `HOME` is cleared by `sudo` and reset to the `root` user's home directory, then this is likely to try to modify `root`'s global Git configuration file but as the non-`root` user.

The difference between systems depends on their `sudoers(5)` configuration, and also on how the user runs `sudo`.  For instance, even on a macOS system where `HOME` normally retains the user's home directory path after running `sudo`, I can do the following:

```
$ sudo -s
# sudo -u chrisd git var GIT_CONFIG_GLOBAL
/Users/chrisd/.config/git/config
/Users/chrisd/.gitconfig

# exit

$ sudo -sH
# sudo -u chrisd git var GIT_CONFIG_GLOBAL
/var/root/.config/git/config
/var/root/.gitconfig

# exit
```

So with the changes from this PR, if I ran `sudo -H ./install.sh`, then the `git lfs install` command would try to modify `root`'s Git configuration file but do so as my regular user account.  Note that `git lfs install` just executes `git config --global` to make these changes, so it really depends on how Git determines the file it is going to alter.

For an alternative approach, see my notes in https://github.com/git-lfs/git-lfs/issues/6230#issuecomment-4202298798.  I suspect we'll probably need to record the location of the global Git configuration file that the `git config --global` command will modify, read its ownership, run `git lfs install` as we do now, and then reset its ownership afterwards (but only if we're running as `root`).

chrisd8088 (commented):
Hey, thanks for this PR!

Unfortunately, I don't think we can just use `SUDO_USER` to try to resolve this problem, as we really have to take into account the file Git is going to modify for us as well.

script/install.sh
chrisd8088: > However, on systems where `HOME` is cleared by `sudo` and reset to the `root` user's home directory, then this is likely to try to modify `root`'s global Git configuration file but as the non-`root` user.

Actually, I have to correct myself here; my apologies.  On systems like this, which include many Linux distributions, running `sudo -u "$SUDO_USER" ...` as `root` should re-reset the `HOME` environment variable to the invoking user's home directory.

The consequence should be that on such systems, when run using `sudo ./install.sh`, where our script formerly updated `root`'s global (i.e., user-specific) Git configuration file it would now update the invoking user's global Git configuration file.  This would be more akin to what occurs on a macOS system with a default `sudoers` configuration.

I can see an argument that this is what our script should have been doing all along; that is, when run via `sudo`, install the Git LFS binary into a standard system location as `root` and then update the invoking user's Git configuration, not `root`'s Git configuration.

However, that's not what our script has done for the past 8 years or so (at least, not consistently on all systems), which makes me hesitant to change it, given the heterogeneity of contexts in which people are likely using our script.

utafrali:
Done, pushed the fix.

utafrali:
@chrisd8088 Thanks for the detailed review and the correction. Just wanted to check -- does the latest push address your concerns about the `SUDO_USER` approach and the edge cases you mentioned? Happy to iterate further if needed.

chrisd8088:
@utafrali — Hey, thanks for checking in!  Perhaps I'm missing something, but I'm not sure I see a commit newer than 6ed03beef2c586d5271b12b484955898c9a8ec87.  Is there a more recent branch I should look at?

utafrali:
You're right, I don't think the SUDO_USER approach handles all those edge cases you outlined. Let me close this and take another look at the approach you mentioned in #6230 instead.

PR #6237, opened 2026-04-09
commands: fix format string type mismatch in lockverifier

opohorel:
The addLocks() method passes a *git.Ref value to a %q format verb in a tr.Tr.Get() call, but %q expects a string. Newer versions of Go recognize tr.Tr.Get() as a printf-like function and flag this as an error during go vet.

Pass ref.Refspec() instead, which returns the reference name as a string.

### DIFF
--- a/commands/lockverifier.go
+++ b/commands/lockverifier.go
@@ -90,7 +90,7 @@ func (lv *lockVerifier) addLocks(ref *git.Ref, locks []locking.Lock, set map[str
 	for _, l := range locks {
 		if rl, ok := set[l.Path]; ok {
 			if err := rl.Add(ref, l); err != nil {
-				Error(tr.Tr.Get("warning: error adding %q lock for ref %q: %+v", l.Path, ref, err))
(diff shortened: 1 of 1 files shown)


chrisd8088 (approved):
Thanks for this bit of cleanup!

Intriguingly, I'm unable to get `go vet` to complain about this particular issue, even with the latest Go v1.26.2, but it's definitely a mistake regardless, so I'm happy to approve this correction.

Thanks again!

opohorel:
I hit it when rebuilding CentOS Stream rpm package, where we run `go vet` in the %check section. It most likely didn't fail for you because we are building the package with `GO111MODULE=off`.

The error message looked like this:
```
./lockverifier.go:93:60: (*github.com/git-lfs/git-lfs/v3/vendor/github.com/leonelquinteros/gotext.Locale).Get format %q has arg ref of wrong type *github.com/git-lfs/git-lfs/v3/git.Ref
FAIL	github.com/git-lfs/git-lfs/v3/commands [build failed]
```

It mentions vendor directory in the path. I guess `go vet` in the GOPATH mode detects the tr.Tr.Get() as a printf-like function, but in MODULE mode it stays silent.

PR #6240, opened 2026-04-13
Fix adapter reuse when server returns empty transfer adapter name

joshfriend:
When the batch API response omits the transfer adapter name (empty string), useAdapter would fail the equality check against the current adapter and unnecessarily tear it down. Check for empty name first to preserve the existing adapter.

### DIFF
--- a/tq/transfer_queue.go
+++ b/tq/transfer_queue.go
@@ -889,6 +889,13 @@ func (q *TransferQueue) useAdapter(name string) {
 	q.adapterInitMutex.Lock()
 	defer q.adapterInitMutex.Unlock()
 
+	// The spec says clients MUST use the basic transfer adapter when
+	// the response omits the transfer property (docs/api/batch.md).
+	// Normalize here to match NewAdapterOrDefault.
+	if name == "" {
+		name = BasicAdapterName
+	}
+
 	if q.adapter != nil {
 		if q.adapter.Name() == name {
 			// re-use, this is the normal path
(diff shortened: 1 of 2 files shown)


tq/transfer_queue.go
```diff
@@ -889,9 +889,15 @@ func (q *TransferQueue) useAdapter(name string) {
...
+	// The spec says clients MUST use the basic transfer adapter when
+	// the response omits the transfer property (docs/api/batch.md).
+	// Normalize here to match NewAdapterOrDefault.
+	if name == "" {
+		name = BasicAdapterName
+	}
+
 	if q.adapter != nil {
 		if q.adapter.Name() == name {
-			// re-use, this is the normal path
```
chrisd8088, line 894: Is there any particular reason we should remove this comment?  I'd be inclined to leave it in place unless it's technically incorrect, which I don't think is the case.

chrisd8088 (commented):
This looks great, thank you!  And thanks very much for adding the tests.

There is one existing comment which I'd prefer to leave intact, unless I'm missing something and it's not correct for some reason.  Please let me know if that's the case!

tq/transfer_queue.go
joshfriend: In #6234 I replaced this comment with the one that is now above this by the `if name == ""` check. I forgot to put it back when I split the PR out and updated this section to be more like `NewAdapterOrDefault`.

I put it back and amended

chrisd8088: Great, thank you!  That's what I guessed.

chrisd8088 (approved):

PR #6241, opened 2026-04-13
Scale default concurrent transfers by CPU count

joshfriend:
Default to 6× logical CPUs (min 8) instead of a hard-coded 8, better utilizing available I/O parallelism. (See previous benchmarks in #6234)

### DIFF
--- a/t/t-env.sh
+++ b/t/t-env.sh
@@ -14,6 +14,8 @@ fi
 # despite the "GIT_" strings in its name and value.
 export TEST_GIT_EXAMPLE="GIT_EXAMPLE"
 
+setup_expected_concurrent_transfers
+
 begin_test "env with no remote"
 (
   set -e
@@ -39,7 +41,7 @@ LocalGitStorageDir=%s
 LocalMediaDir=%s
 LocalReferenceDirs=
 TempDir=%s
-ConcurrentTransfers=8
+ConcurrentTransfers=%d
 TusTransfers=false
 BasicTransfersOnly=false
 SkipDownloadErrors=false
@@ -58,7 +60,7 @@ DownloadTransfers=basic,lfs-standalone-file,ssh
 UploadTransfers=basic,lfs-standalone-file,ssh
 %s
 %s
-' "$(git lfs version)" "$(git version)" "$localwd" "$localgit" "$localgitstore" "$localmedia" "$tempdir" "$lfsstorage" "$envVars" "$envInitConfig")
+' "$(git lfs version)" "$(git version)" "$localwd" "$localgit" "$localgitstore" "$localmedia" "$tempdir" "$expectedConcurrentTransfers" "$lfsstorage" "$envVars" "$envInitConfig")
   actual=$(git lfs env | grep -v "^GIT_EXEC_PATH=")
 
   contains_same_elements "$expected" "$actual"
@@ -92,7 +94,7 @@ LocalGitStorageDir=%s
 LocalMediaDir=%s
 LocalReferenceDirs=
 TempDir=%s
-ConcurrentTransfers=8
+ConcurrentTransfers=%d
 TusTransfers=false
 BasicTransfersOnly=false
 SkipDownloadErrors=false
@@ -111,7 +113,7 @@ DownloadTransfers=basic,lfs-standalone-file,ssh
 UploadTransfers=basic,lfs-standalone-file,ssh
 %s
 %s
-' "$(git lfs version)" "$(git version)" "$endpoint" "$localwd" "$localgit" "$localgitstore" "$localmedia" "$tempdir" "$lfsstorage" "$envVars" "$envInitConfig")
+' "$(git lfs version)" "$(git version)" "$endpoint" "$localwd" "$localgit" "$localgitstore" "$localmedia" "$tempdir" "$expectedConcurrentTransfers" "$lfsstorage" "$envVars" "$envInitConfig")
   actual=$(git lfs env | grep -v "^GIT_EXEC_PATH=")
   contains_same_elements "$expected" "$actual"
 
@@ -152,7 +154,7 @@ LocalGitStorageDir=%s
 LocalMediaDir=%s
 LocalReferenceDirs=
 TempDir=%s
-ConcurrentTransfers=8
+ConcurrentTransfers=%d
 TusTransfers=false
 BasicTransfersOnly=false
 SkipDownloadErrors=false
@@ -171,7 +173,7 @@ DownloadTransfers=basic,lfs-standalone-file,ssh
 UploadTransfers=basic,lfs-standalone-file,ssh
 %s
 %s
-' "$(git lfs version)" "$(git version)" "$endpoint" "$endpoint2" "$localwd" "$localgit" "$localgitstore" "$localmedia" "$tempdir" "$lfsstorage" "$envVars" "$envInitConfig")
+' "$(git lfs version)" "$(git version)" "$endpoint" "$endpoint2" "$localwd" "$localgit" "$localgitstore" "$localmedia" "$tempdir" "$expectedConcurrentTransfers" "$lfsstorage" "$envVars" "$envInitConfig")
   actual=$(git lfs env | grep -v "^GIT_EXEC_PATH=")
   contains_same_elements "$expected" "$actual"
 
@@ -210,7 +212,7 @@ LocalGitStorageDir=%s
 LocalMediaDir=%s
 LocalReferenceDirs=
 TempDir=%s
-ConcurrentTransfers=8
+ConcurrentTransfers=%d
 TusTransfers=false
 BasicTransfersOnly=false
 SkipDownloadErrors=false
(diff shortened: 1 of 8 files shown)


t/t-env.sh
```diff
@@ -687,7 +695,7 @@ LocalGitStorageDir=%s
 LocalMediaDir=%s
 LocalReferenceDirs=
 TempDir=%s
-ConcurrentTransfers=8
+ConcurrentTransfers='"$expectedConcurrentTransfers"'
```
chrisd8088: There's a sneaky difference in this one `printf()` statement, because we use double quotes rather than single quotes for its format string argument, i.e., `printf "%s..." "$(git lfs version)" ...` rather than `printf '%s...' "$(git lfs version)" ...`).

So this particular line produces a `diff` like the following and consequent test failure:

```diff
-ConcurrentTransfers='8'
+ConcurrentTransfers=8
```

Rather than change the quoting, though, the way I'd suggest that we take another approach to interpolating the expected value we've calculated for `ConcurrentTransfers`, which is just to replace the original hard-coded value of `8` with a `%d` format specifier, and then add `"$expectedConcurrentTransfers"` into the `printf()` argument list further down.  We can then use this approach for all the tests in the `t/t-env.sh` and `t/t-worktree.sh` scripts.

```suggestion
ConcurrentTransfers=%d
```

Here's how we'd insert the additional value into all the `printf()` statements' argument lists:

```diff
-" "$(git lfs version)" "$(git version)" "$localgit" "$localgitstore" "$localmedia" "$tempdir" "$lfsstorage" "$envVars" "$envInitConfig")
+" "$(git lfs version)" "$(git version)" "$localgit" "$localgitstore" "$localmedia" "$tempdir" "$expectedConcurrentTransfers" "$lfsstorage" "$envVars" "$envInitConfig")
```

t/t-env.sh
```diff
@@ -14,6 +14,14 @@ fi
 # despite the "GIT_" strings in its name and value.
 export TEST_GIT_EXAMPLE="GIT_EXAMPLE"
 
+# Compute the expected default ConcurrentTransfers value (3 × NCPU,
+# clamped to [8, 384]) so tests stay correct on any machine.
+_ncpu=$(getconf _NPROCESSORS_ONLN 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || echo 1)
```
chrisd8088: It looks like the `getconf(1)` command should work in all our CI environments, including Git Bash on Windows, so that's great!

Since this test setup logic is a bit involved and we have to repeat it in two test scripts, I might suggest moving it into a `setup_expected_concurrent_transfers()` function in our `t/testhelpers.sh` library, and then just calling that function here and in `t/t-workflow.sh`.

---

If by chance we run into issues in the future where the value `getconf _NPROCESSORS_ONLN` returns doesn't match up with what Go's `runtime.NumCPU()` function returns, we can work around that by adding a small `t/cmd/lfstest-getnumcpu.go` test helper program which just outputs the return value from `runtime.NumCPU()`.  We have some existing helpers like the `t/cmd/lfstest-nanomtime.go` [program](https://github.com/git-lfs/git-lfs/blob/ef7a182c3829cf5742239f7bbb748a4883bb1834/t/cmd/lfstest-nanomtime.go) we use in our shell scripts to gather the same data as the `git-lfs` binary gets from Go standard library functions.

tq/manifest.go
```diff
@@ -180,6 +183,21 @@ func (m *concreteManifest) Upgraded() bool {
 	return true
 }
 
+// defaultConcurrentTransfers scales with CPU count, clamped to a
+// reasonable range. Downloads are I/O-bound (network + disk) so we use
+// 3× NCPU to keep many connections saturated while a few are stalled on
+// TLS handshakes or server latency.
+func defaultConcurrentTransfers() int {
```
chrisd8088: I noticed that on macOS, we had test failures showing an expected CPU count of 24, but only 8 in the `git lfs env` output:

```diff
-ConcurrentTransfers=24
+ConcurrentTransfers=8
```

This puzzled me until I realized that we actually read the `lfs.concurrentTransfers` option in another [place](https://github.com/git-lfs/git-lfs/blob/ef7a182c3829cf5742239f7bbb748a4883bb1834/lfshttp/client.go#L89), where we also set the default to `8`.  Plus we have a third [place](https://github.com/git-lfs/git-lfs/blob/ef7a182c3829cf5742239f7bbb748a4883bb1834/lfshttp/client.go#L414) where that default value is set hard-coded again.

To be clear, repeated hard-coded default values like these are not good program design or coding practice, but just the accidental result of our development history.

Since we're now going to determine the default value in a more dynamic manner, we need to consolidate this code into one place where we choose the final value only once.  That should probably be in the `lfshttp` package, I think, rather than here in the `tq` (i.e., transfer queue) package.

It's worth noting that we use the value of the option (whether specified explicitly or based on the default) not just in the transfer queue logic but also to [set](https://github.com/git-lfs/git-lfs/blob/ef7a182c3829cf5742239f7bbb748a4883bb1834/lfshttp/client.go#L434) the `MaxIdleConnsPerHost` [field](https://github.com/golang/go/blob/5bb4fb9ca8e687ed33d04486a690051f209d5569/src/net/http/transport.go#L210-L213) of the `http.Transport` structure we create when we perform an HTTP request.

tq/manifest.go
```diff
@@ -12,9 +13,11 @@ import (
...
 
 const (
-	defaultMaxRetries          = 8
-	defaultMaxRetryDelay       = 10
-	defaultConcurrentTransfers = 8
+	defaultMaxRetries    = 8
+	defaultMaxRetryDelay = 10
+
+	minConcurrentTransfers = 8
+	maxConcurrentTransfers = 384
```
chrisd8088: This seems like a reasonable contemporary maximum (3× for 128-core CPU, say).

That said, I wonder if we would be better served over the long run replacing hard-coded specific maximums like this with a default maximum scaling factor based on the number of CPUs.  But this seems fine for now, I think, given your benchmark results in PR #6234 on 192-core systems.

chrisd8088 (commented):
Thanks very much for splitting out this change as a separate PR!

I've made some suggestions and notes regarding the need to refactor these changes slightly to also update the `lfshttp` package, which should help the tests pass too.

Thanks again for working on these improvements for Git LFS!

t/t-env.sh
joshfriend: > I might suggest moving it into a setup_expected_concurrent_transfers() function in our t/testhelpers.sh library, and then just calling that function here and in t/t-workflow.sh.

done!

> If by chance we run into issues in the future where the value `getconf _NPROCESSORS_ONLN` returns doesn't match up with what Go's `runtime.NumCPU()` function returns

I was already on the fence about this having just dealt with this in a different context and cgroup limits, so I just added the helper proactively.

tq/manifest.go
joshfriend: > That should probably be in the lfshttp package, I think

makes sense to me, done!

tq/manifest.go
joshfriend: I'm experimenting with this multiplier some more in a local and CI worker type environment to see if 3x really is the best heuristic. So far for local (nCPU=16) it is pretty good but it gets a bit faster at higher multipliers. These tests are with my pipelining PR by the way, since that's the end goal of this PR series.

image

And for the ci environment with 192 cpus it doesn't make a whole lot of difference after 1x nCPU

image

I don't think 3x is a bad default, but 6x looks better for developer macbooks and doesn't hurt huge CI machines much.

joshfriend:
In addition to addressing the review comments (thank you!), I've updated the PR with these changes:

* No cap on concurrent transfers by default
* Use 6x multiplier instead of 3x

Based on the combination of this and the changes in https://github.com/git-lfs/git-lfs/pull/6234 that appears to be the best default for the two large repos we have that contain thousands of PNG snapshot test baselines. I did not see a practical downside in my testing to increasing that multiplier, 3-4 seconds in a huge CI worker isn't going to make a difference (and in workers like this we typically circumvent git and git lfs entirely with a snapshotting mechanism to avoid github ratelimits)

tq/manifest.go
chrisd8088: Very interesting, thank you!

I confess that I'm inclined to be a bit cautious about the scaling factor, only because there are a lot of systems out there and we don't have a full picture, especially for resource-constrained systems.

Would you be OK with starting out with just the 3× factor until we get some feedback from users?

chrisd8088 (approved):
This looks really good, thank you!

I'm going to approve this PR, but before I merge it, would you consider lowering the scaling factor just a bit, just to be a bit conservative?  We have quite a diverse user base and I'm inclined to play it safe for a while, at least.

joshfriend:
> before I merge it, would you consider lowering the scaling factor just a bit, just to be a bit conservative?

Done! (Set back to 3x) Happy to start wherever you are comfortable, I have the capability of configuring this using the override setting for my users and cloud environments to match my benchmarks.

chrisd8088:
> Done! (Set back to 3x)

Thank you very much!

PR #6243, opened 2026-04-17
Fix concurrent map/struct races in sshCache and lazyManifest

joshfriend:
In #6234 `sshCache.endpoints` is accessed from multiple goroutines when parallel batch API workers resolve SSH credentials simultaneously. Replace the plain map with `sync.Map` to prevent the "concurrent map writes" panic.

`lazyManifest.Upgrade()` can be called concurrently, causing multiple concreteManifest instances to be created. Add a mutex to ensure only one is created.

### DIFF
--- a/lfshttp/ssh.go
+++ b/lfshttp/ssh.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/git-lfs/git-lfs/v3/config"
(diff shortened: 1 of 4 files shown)


joshfriend:
Side note: one of these race condition tests only reproduces when run with `go test -race` which is not used by CI.

chrisd8088 (approved):
Hey, thanks for this PR, it looks great!

IIUC, the changes in the two packages (`lfshttp` and `tq`) are unrelated to each other, so they could be separate commits.  But that's merely a procedural detail, and the added thread safety is a clear improvement.

chrisd8088:
> Side note: one of these race condition tests only reproduces when run with `go test -race` which is not used by CI.

Thanks for the note!  I actually found both tests would report an issue when `make test-race` was run, if the corresponding code fixes were not in place, so that's good, and we could certainly add `make test-race` to our `script/cibuild` script in another PR.

PR #6244, opened 2026-04-17
locking: ignore missing files after unlock

### LINKED ISSUE #6167: Unlock --json reporting of non-existant file
**Describe the bug**
Unlock --json reporting on files which don't exist in local work tree is incorrect.

**To Reproduce**
Set-up: 
- two clones with the same user (so that user can edit the locks for testing purposes)
- cloneA creates an lfs tracked file, file1

1. On cloneA lock unstaged file1
2. On cloneB try to unlock file1
  - `git lfs unlock --json file1` 
**Expected behavior**
json report of unlock attempt on cloneB does not match behaviour.
```json
[
    {
        "path": "file1",
        "unlocked": false,
        "reason": "stat file1: no such file or directory"
    },
....
]
```
Reported as not unlocked.  
However lock on file1 is removed from the server, this can be checked with both:
- `git lfs locks`
- `git lfs locks --verify`

**System environment**
OS: Ubuntu 22.04

**Output of `git lfs env`**
The following output has been redacted.
The git forge server is an org hosted gitlab server.
The lfs server is unknown, but the endpoint appears to be on the same server.
```bash
git-lfs/3.7.1 (GitHub; linux amd64; go 1.24.4)
git version 2.51.2

Endpoint=<org hosted gitlab server lfs endpoint> (auth=none)
  SSH=git@<org hosted gitlab server>:<repo>.git
LocalWorkingDir=<repo_path>
LocalGitDir=<repo_path>/.git
LocalGitStorageDir=<repo_path>/.git
LocalMediaDir=<repo_path>/.git/lfs/objects
LocalReferenceDirs=
TempDir=<repo_path>/.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=<repo_path>/.git/lfs
AccessDownload=none
AccessUpload=none
```

**Additional context**
I understand that the server side implementation of the lfs protocol is the responsibility of the server (i.e. gitlab in this case). However, the report must be constructed locally (as it is including information about the local filesystem), and is also not populating with what the server executed (responds with a 200 response to the unlock POST).

lawrence3699:
Fixes #6167

When one clone has the lockable pattern but not the newly locked file, `git lfs unlock --json <path>` can remove the server lock and then turn that successful unlock into a local failure while trying to make the missing path read-only.

Before this change, the unlock POST succeeded but the command exited with `unlocked:false` and a local `stat ... no such file or directory` reason.

After this change, a successful unlock stays successful, and the post-unlock chmod step is skipped when the file is not present in the current worktree.

I also added a regression in `t/t-unlock.sh` that reproduces the two-clone lockable-pattern case from the issue.

Validation:
- `make -C t t-unlock.sh`
- `go test ./locking`

### DIFF
--- a/t/t-unlock.sh
+++ b/t/t-unlock.sh
@@ -508,27 +508,24 @@ begin_test "unlocking a missing lockable file (--json)"
 
   reponame="unlock_missing_lockable_path_json"
   setup_remote_repo "$reponame"
-  clone_repo "$reponame" seed
+  clone_repo "$reponame" "$reponame"
 
-  cd "$TRASHDIR/seed"
   git lfs track --lockable "*.dat"
   git add .gitattributes
   git commit -m "Add lockable pattern"
   git push origin main
 
-  cd "$TRASHDIR"
-  clone_repo "$reponame" cloneA
-  clone_repo "$reponame" cloneB
-
-  cd "$TRASHDIR/cloneA"
   echo "file1.dat" > file1.dat
   git lfs lock --json "file1.dat" | tee lock.log
 
   id=$(assert_lock lock.log file1.dat)
   assert_server_lock "$reponame" "$id"
 
-  cd "$TRASHDIR/cloneB"
-  git lfs unlock --json "file1.dat" 2>&1 | tee unlock.log
+  rm file1.dat
+
+  git lfs unlock --json "file1.dat" | tee unlock.log
+  [ 0 -eq "${PIPESTATUS[0]}" ]
+
   grep -F '[{"path":"file1.dat","unlocked":true}]' unlock.log
 
   refute_server_lock "$reponame" "$id"
--- a/t/t-unlock.sh
+++ b/t/t-unlock.sh
@@ -378,6 +378,39 @@ begin_test "unlocking a lock (--json)"
 )
 end_test
 
+begin_test "unlocking a missing lockable file (--json)"
+(
+  set -e
+
+  reponame="unlock_missing_lockable_path_json"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" seed
+
+  cd "$TRASHDIR/seed"
+  git lfs track --lockable "*.dat"
+  git add .gitattributes
+  git commit -m "Add lockable pattern"
+  git push origin main
+
+  cd "$TRASHDIR"
+  clone_repo "$reponame" cloneA
+  clone_repo "$reponame" cloneB
+
+  cd "$TRASHDIR/cloneA"
+  echo "file1.dat" > file1.dat
+  git lfs lock --json "file1.dat" | tee lock.log
+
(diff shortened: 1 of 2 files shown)


t/t-unlock.sh
```diff
@@ -378,6 +378,39 @@ begin_test "unlocking a lock (--json)"
...
+
+  cd "$TRASHDIR/cloneA"
+  echo "file1.dat" > file1.dat
+  git lfs lock --json "file1.dat" | tee lock.log
+
+  id=$(assert_lock lock.log file1.dat)
+  assert_server_lock "$reponame" "$id"
+
+  cd "$TRASHDIR/cloneB"
+  git lfs unlock --json "file1.dat" 2>&1 | tee unlock.log
```
chrisd8088: This is a great addition to our test suite, thank you!

I think we might be able to both simplify and expand it slightly by adopting some of the idioms used elsewhere in our shell tests.  How would you feel about applying the (relatively minor, I hope!) revisions below?

```suggestion
  clone_repo "$reponame" "$reponame"

  git lfs track --lockable "*.dat"
  git add .gitattributes
  git commit -m "Add lockable pattern"
  git push origin main

  echo "file1.dat" > file1.dat
  git lfs lock --json "file1.dat" | tee lock.log

  id=$(assert_lock lock.log file1.dat)
  assert_server_lock "$reponame" "$id"

  rm file1.dat

  git lfs unlock --json "file1.dat" | tee unlock.log
  if [ "0" -ne "${PIPESTATUS[0]}" ]; then
    echo >&2 "fatal: expected 'git lfs unlock' to succeed ..."
    exit 1
  fi

```

In brief, we can just use the initial repository for the whole test, and we likely don't need to capture `stderr` output from the `git lfs unlock` command because we really just care that it reports success and that the lock is removed.

However, we might want to confirm that the `git lfs unlock` command now returns a `0` exit code.  Right now it returns `2`, but that non-zero exit code won't cause the test to fail because its masked by the pipeline's final `tee(1)` command.

chrisd8088 (commented):
Thanks so much for this helpful bug fix and for including a comprehensive test!

I made some suggestions regarding the test, but only to try to simplify its setup steps and add a check of the exit code from the `git lfs unlock` command.  Please let me know what you think of those, and thanks again for this PR!

t/t-unlock.sh
chrisd8088: I went ahead and added these minor test simplifications to the PR; I hope that's OK!

chrisd8088 (approved):
Thanks again for fixing this bug!

PR #6250, opened 2026-04-22
Fix remote ref resolution for download operations

larsxschneider:
## Summary

Fix download operations to resolve the remote ref using the upstream tracking branch instead of `push.default` logic.

## Problem

`currentRemoteRef()` resolved the remote ref sent in LFS batch API requests using `push.default` rules via `cfg.PushRemote()`. This is semantically correct for pushes but wrong for downloads: a download needs to tell the LFS server which remote branch the client is *fetching from* (for authorization), not which branch a *push* would target.

When `remote.pushDefault` pointed to a different remote than the fetch remote, the download path could resolve to the wrong ref.

## Changes

Split `currentRemoteRef()` into two purpose-specific functions:

- **`pushRemoteRef()`** -- resolves the remote ref using `push.default` rules and the configured push remote (used by push operations)
- **`fetchRemoteRef()`** -- resolves the upstream tracking ref for the current local branch (used by download operations)

Export `TrackingRef` in `git/refs.go` so `fetchRemoteRef` can look up the tracking branch directly without going through push-oriented `defaultRemoteRef` logic.

## Testing

Added integration test "fetch with pushRemote configured" that:
1. Pushes to a `tracked` branch on origin
2. Configures `remote.pushDefault` to a separate push-only remote
3. Sets `branch.main.merge` to `refs/heads/tracked`
4. Verifies `git lfs fetch` sends the correct tracking ref

The test fails without the fix (server rejects the wrong ref) and passes with it.

### DIFF
--- a/commands/command_filter_process.go
+++ b/commands/command_filter_process.go
@@ -107,7 +107,7 @@ func filterCommand(cmd *cobra.Command, args []string) {
 					tq.Download,
 					getTransferManifestOperationRemote("download", cfg.Remote()),
 					cfg.Remote(),
-					tq.RemoteRef(currentRemoteRef()),
+					tq.RemoteRef(fetchRemoteRef()),
 					tq.WithBatchSize(cfg.TransferBatchSize()),
 				)
 				go infiniteTransferBuffer(q, available)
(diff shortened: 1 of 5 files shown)


chrisd8088 (approved):
This looks great, thank you!

Issue #6118, opened 2025-09-19
could not get connection for batch request: pure SSH connection unavailable


galets:
**Describe the bug**

I think I see the variant of #6060, which I believe was closed prematurely, as it was fixed in gitea, and not git-lfs. This is what I observe:

```bash
$ git lfs push --all origin
... 30 or more minutes passes here ...
could not get connection for batch request: pure SSH connection unavailable (#0)                                                                                                                                                           
could not get connection for batch request: pure SSH connection unavailable (#0)
could not get connection for batch request: pure SSH connection unavailable (#0)
could not get connection for batch request: pure SSH connection unavailable (#0)
could not get connection for batch request: pure SSH connection unavailable (#0)
...
```

**To Reproduce**

this is git push to a newly created bare repository using git-lfs-transfer via pure ssh connection. Repository is ~35GB in size, which I believe is the cause for a timeout.

**Expected behavior**

`git lfs push` succeeds

**System environment**

Unubtu 24.04
Server implementation - Ubuntu 24.04
Direct SSH connection, same local network, no firewall

**Output of `git lfs env`**

```
$ git lfs env
git-lfs/3.7.0 (GitHub; linux amd64; go 1.24.4)
git version 2.43.0

Endpoint=https://nas30.home/documents-lfs.git/info/lfs (auth=none)
  SSH=gitfam@nas30.home:documents-lfs
LocalWorkingDir=/home/galets/tmp/docs-lfs/Documents
LocalGitDir=/home/galets/tmp/docs-lfs/Documents/.git
LocalGitStorageDir=/home/galets/tmp/docs-lfs/Documents/.git
LocalMediaDir=/home/galets/tmp/docs-lfs/Documents/.git/lfs/objects
LocalReferenceDirs=
TempDir=/home/galets/tmp/docs-lfs/Documents/.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
FetchRecentRefsIncludeRemotes=true
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneVerifyUnreachableAlways=false
PruneRemoteName=origin
LfsStorageDir=/home/galets/tmp/docs-lfs/Documents/.git/lfs
AccessDownload=none
AccessUpload=none
[...]

PR #6254, opened 2026-04-26
Simplify client structure initializations

chrisd8088:
This PR simplifies all of the factory functions we use to construct new API, HTTP, and locking clients by removing their unused `error` return values.

For historical reasons, the `NewClient()` functions in our `lfsapi`, `lfshttp`, and `locking` packages all return a value with the `error` type, but at present none of the functions could ever actually encounter an error condition and return a non-`nil` `error` value.  We can therefore just eliminate these unused return values.

#### Avoiding Future Deadlocks

One specific advantage of this change is that the `getAPIClient()` function in our `commands` package will no longer need to include a conditional [block](https://github.com/git-lfs/git-lfs/blob/e09c0f6216dadd0624eb425e8965f28868f9cd47/commands/commands.go#L81-L83) which checks the `error` return value from the `lfsapi.NewClient()` constructor function.  This block handles the (hypothetical) case of a non-`nil` `error` return value by calling the `ExitWithError()` function, which causes the Git LFS client to exit immediately.

In a future PR, we may revise the `ExitWithError()` function so that it performs various cleanup operations prior to calling `os.Exit()` function, but only after first acquiring the same `global` [mutex](https://github.com/git-lfs/git-lfs/blob/e09c0f6216dadd0624eb425e8965f28868f9cd47/commands/commands.go#L44) that the `getAPIClient()` function also [acquires](https://github.com/git-lfs/git-lfs/blob/e09c0f6216dadd0624eb425e8965f28868f9cd47/commands/commands.go#L76-L77).  With this goal in mind, we need to ensure that the `getAPIClient()` function does not invoke the `ExitWithError()` function, since that would result in a deadlock condition.

### DIFF
--- a/lfshttp/client.go
+++ b/lfshttp/client.go
@@ -83,7 +83,7 @@ type Client struct {
 	sshTries int
 }
 
-func NewClient(ctx Context) (*Client, error) {
+func NewClient(ctx Context) *Client {
 	if ctx == nil {
 		ctx = NewContext(nil, nil, nil)
 	}
@@ -97,7 +97,7 @@ func NewClient(ctx Context) (*Client, error) {
 		sshResolver = withSSHCache(sshResolver)
 	}
 
-	c := &Client{
+	return &Client{
 		SSH:                 sshResolver,
 		DialTimeout:         gitEnv.Int("lfs.dialtimeout", 0),
 		KeepaliveTimeout:    gitEnv.Int("lfs.keepalive", 0),
@@ -112,8 +112,6 @@ func NewClient(ctx Context) (*Client, error) {
 		sshTries:            gitEnv.Int("lfs.ssh.retries", 5),
 		credHelperContext:   creds.NewCredentialHelperContext(gitEnv, osEnv),
 	}
-
-	return c, nil
 }
 
 func (c *Client) GitEnv() config.Environment {
--- a/lfshttp/client.go
+++ b/lfshttp/client.go
@@ -92,6 +92,10 @@ func NewClient(ctx Context) *Client {
 	osEnv := ctx.OSEnv()
 
 	cacheCreds := gitEnv.Bool("lfs.cachecredentials", true)
+
+	// SSHResolver resolves LFS endpoint authentication by running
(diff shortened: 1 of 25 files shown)


knightitcomp-sketch:
I don't know what this means, my ph is getting hacked, I put alerts on all
my credit companies, but there still hacking me, I don't think i even know
how to upload the code that would help? What can/should I do?

tq/manifest.go
```diff
@@ -185,18 +185,20 @@ func (m *concreteManifest) Upgraded() bool {
 	return true
 }
 
+func (m *concreteManifest) Shutdown() error {
+	if m.sshTransfer != nil {
+		return m.sshTransfer.Shutdown()
+	}
+	return nil
+}
```
larsxschneider: Where is this new method used?

lfshttp/client.go
```diff
@@ -97,7 +97,7 @@ func NewClient(ctx Context) (*Client, error) {
 		sshResolver = withSSHCache(sshResolver)
 	}
 
-	c := &Client{
+	return &Client{
 		SSH:                 sshResolver,
```
larsxschneider, line 105: SSH in the lfshttp package gave me a short pause :-).

Extra nit: Maybe we could add a comment or make it more clear that this is for auth only?!

```
 // SSHResolver resolves LFS endpoint authentication by running
 // git-lfs-authenticate over SSH. The returned credentials (URL
 // and headers) are then used for subsequent HTTPS API requests.
```

larsxschneider (approved):
A nice cleanup! Thank you 🙇 

(Unless we use it shortly, I would remove the unused Shutdown() method before merging)

tq/manifest.go
chrisd8088: Ah, good catch; that's part of the next round of work to address #6118 which builds on this PR.  Let me pull this out of this PR.

lfshttp/client.go
chrisd8088: Sure thing, that's a good idea.

PR #6255, opened 2026-04-27
Use consistent exit functions for all commands

chrisd8088:
This PR updates a number of Git LFS commands to ensure that all our commands use a consistent set of utility functions when they need to stop and exit immediately.

These changes will help minimize the revisions that will be necessary in a future PR to guarantee that our commands always perform a common set of cleanup steps before exiting.

#### Specific Changes

In this PR, we first update several commands' implementation functions to either use the existing `ExitWithError()` utility [function](https://github.com/git-lfs/git-lfs/blob/e09c0f6216dadd0624eb425e8965f28868f9cd47/commands/commands.go#L237-L241) or, if they intend to exit with a status code of zero, to simply return and allow our normal call stack to completely unwind.

In the particular case of our `git lfs standalone-file` command, we also revise our `lfshttp/standalone` package to eliminate the `standaloneFailure()` [function](https://github.com/git-lfs/git-lfs/blob/e09c0f6216dadd0624eb425e8965f28868f9cd47/lfshttp/standalone/standalone.go#L277-L281) which at present calls the `os.Exit()` function directly.  To make this possible, we adjust our handling of unknown Git LFS custom transfer protocol event messages so that we now report an error back up to the `standaloneFileCommand()` function in the `commands` package.

We then introduce a new `ExitWithCode()` utility function which, in its initial form, just calls the `os.Exit()` function with a given status code.  In a subsequent PR, however, we may expand this function and the small set of remaining functions which call the `os.Exit()` function to ensure they all execute a standard cleanup and shutdown sequence before exiting.

### DIFF
--- a/commands/command_checkout.go
+++ b/commands/command_checkout.go
@@ -32,7 +32,7 @@ func checkoutCommand(cmd *cobra.Command, args []string) {
 	// repository but will exit non-zero, as other commands already do.
 	if cfg.LocalWorkingDir() == "" {
 		Print(tr.Tr.Get("This operation must be run in a work tree."))
-		os.Exit(0)
+		return
 	}
 
 	stage, err := whichCheckout()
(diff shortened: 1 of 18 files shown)


larsxschneider (approved):
Again, nice cleanup 🙇

PR #6258, opened 2026-04-30
Temporarily restore concurrency limit to historical fixed default

chrisd8088:
This PR reverses the key change from PR #6241 and restores the fixed, historical default of 8 for the `lfs.concurrentTransfers` configuration option, so that we do not spawn large numbers of child processes when transferring objects using our "pure" SSH protocol or when using a custom transfer agent.

After the upcoming v3.8.0 Git LFS release we will revisit the issue of scaling default concurrency settings with CPU counts while limiting the number of child processes we might create.  See #6259 for that discussion.

#### Additional Context

As discussed in PRs #6241 and #6234, we would like to refine the concurrency controls for the Git LFS client so that they can scale with the number of available CPUs, at least when performing HTTP-based object transfers.

We adopted an initial implementation of in PR #6241, which was effective for HTTP object transfers and improved performance significantly on multi-core CPUs.

However, this change also has the unintended effect of scaling the number of child processes we execute when handling object transfers using a custom agent or when using multiplexed SSH connections.  While some systems may not be affected, creating large numbers of separate processes is at best not ideal and at worst may run afoul of system or container limits.

We therefore temporarily reverse the key change from PR #6241 and restore the fixed, historical default of 8 for the `lfs.concurrentTransfers` configuration option.

After our upcoming release of version 3.8.0 of the Git LFS client, we intend to revisit this issue and find a more comprehensive solution so that we can safely scale thread counts while keeping process counts within reasonable bounds.

#### Other Changes

We also take the opportunity to correct the placement and wording of a [comment](https://github.com/git-lfs/git-lfs/blob/4267a957a4ec729d3390dda642c24575b85fc5a8/tq/manifest.go#L252) regarding concurrency support for SSH-based object transfers.

/cc @joshfriend as the author of PRs #6241 and #6234.

### DIFF
--- a/docs/man/git-lfs-config.adoc
+++ b/docs/man/git-lfs-config.adoc
@@ -137,8 +137,7 @@ occurs.
 
 * `lfs.concurrenttransfers`
 +
-The number of concurrent uploads/downloads. Default is 3x the number of
-logical CPUs, with a minimum of 8.
+The number of concurrent uploads/downloads. Default is 8.
 * `lfs.basictransfersonly`
 +
 If set to true, only basic HTTP upload/download transfers will be used,
--- a/lfshttp/client.go
+++ b/lfshttp/client.go
@@ -12,7 +12,6 @@ import (
 	"net/url"
 	"os"
 	"regexp"
-	"runtime"
 	"slices"
 	"strconv"
 	"strings"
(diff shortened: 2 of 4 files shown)


chrisd8088:
Hey @joshfriend, I'm sorry about this PR!

It certainly wasn't my intention to reverse course when I approved and merged PR #6241, which seemed like a relatively straightforward enhancement with clear performance benefits.  I should have thought a bit more carefully about all the dusty edge cases we have to support.

While I don't think it's a disaster if we spin up, say, 10 or 12 custom transfer agents instead of 8, I wouldn't want to surprise users who discover we're now forking 384 processes on their 128-core machine, just to transfer one or two objects via their custom transfer process.

joshfriend:
Not a problem, Chris! In my setup with pretty vanilla github hosted LFS that was a pretty great default tuning, but I had no test harness available for some of the other transport methods and clearly missed a few things.

#6234 is the main unlock for performance though. We could leave the default parallelism pretty low as it sat before, with affordances for increasing it, and I think that would still be a big win!

larsxschneider (approved):

PR #6264, opened 2026-05-12
fix: chunked content length

telometto:
When a server's LFS batch response includes `Transfer-Encoding: chunked` in an upload action's `header` map (as Gitea 1.26.0 now does via go-gitea/gitea#36380), git-lfs correctly sets `req.TransferEncoding` but unconditionally also sets `req.ContentLength = t.Size`, which causes two problems:

Number one is around HTTP/1.1: (Go's) `net/http` client sends both `Transfer-Encoding: chunked` and `Content-Length` (which [RFC 7230 forbids](https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2)). Proxies that key their buffering decision on the presence of `Content-Length` (most notably Cloudflare, which has a 100 MB buffer cap on their free plan) ignore the chunked signal, buffer the body against the declared length, and return HTTP 413 for files exceeding that cap.

The second one is around HTTP/2: the protocol strips `Transfer-Encoding` entirely; Go's HTTP/2 client ignores `req.TransferEncoding` and falls back to `req.ContentLength`. So, basically, the chunked signal is silently discarded, which means that the Gitea server-side fix has no effect at all on TLS connections (where HTTP/2 is the default).

The solution is thus to set `req.ContentLength = -1`, which makes Go omit the `Content-Length` header entirely. In HTTP/1.1 this produces a clean chunked request with no conflicting header; in HTTP/2 it causes the client to use unknown-length streaming semantics, which achieves the same result. In both cases the proxy has no declared size to buffer against and must stream the body through.

### DIFF
--- a/tq/basic_upload.go
+++ b/tq/basic_upload.go
@@ -55,13 +55,13 @@ func (a *basicUploadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb Progres
 		return err
 	}
 
-	if req.Header.Get("Transfer-Encoding") == "chunked" {
+	if isChunkedTransferEncoding(req.Header.Get("Transfer-Encoding")) {
 		req.TransferEncoding = []string{"chunked"}
+		req.ContentLength = -1
 	} else {
 		req.Header.Set("Content-Length", strconv.FormatInt(t.Size, 10))
+		req.ContentLength = t.Size
 	}
-
-	req.ContentLength = t.Size
 
 	f, err := os.OpenFile(t.Path, os.O_RDONLY, 0644)
 	if err != nil {
@@ -208,6 +208,18 @@ func newStartCallbackReader(r lfsapi.ReadSeekCloser, cb func() error) *startCall
 		ReadSeekCloser: r,
 		cb:             cb,
 	}
+}
+
+// isChunkedTransferEncoding reports whether the Transfer-Encoding header value
+// contains "chunked" as a transfer-coding token. Tokens are case-insensitive
+// and the header may be a comma-separated list per RFC 7230 §3.3.1.
+func isChunkedTransferEncoding(v string) bool {
+	for _, tok := range strings.Split(v, ",") {
+		if strings.EqualFold(strings.TrimSpace(tok), "chunked") {
+			return true
+		}
+	}
+	return false
 }
 
 func configureBasicUploadAdapter(m *concreteManifest) {
--- a/tq/basic_upload_test.go
+++ b/tq/basic_upload_test.go
@@ -1,0 +1,89 @@
+package tq
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"strconv"
+	"testing"
+
+	"github.com/git-lfs/git-lfs/v3/lfsapi"
+	"github.com/git-lfs/git-lfs/v3/lfshttp"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBasicUploadAdapterContentLengthWithChunked(t *testing.T) {
+	const content = "hello chunked upload"
+
+	f, err := os.CreateTemp("", "git-lfs-upload-test-*")
+	require.Nil(t, err)
+	defer os.Remove(f.Name())
+	_, err = f.WriteString(content)
+	require.Nil(t, err)
+	f.Close()
+
+	tests := []struct {
+		name              string
+		actionHeaders     map[string]string
+		wantContentLength string
+	}{
+		{
+			name:              "chunked: Content-Length header must be absent",
+			actionHeaders:     map[string]string{"Transfer-Encoding": "chunked"},
+			wantContentLength: "",
+		},
+		{
+			name:              "chunked uppercase: Content-Length header must be absent",
+			actionHeaders:     map[string]string{"Transfer-Encoding": "Chunked"},
+			wantContentLength: "",
+		},
+		{
+			name:              "chunked in list: Content-Length header must be absent",
+			actionHeaders:     map[string]string{"Transfer-Encoding": "gzip, chunked"},
+			wantContentLength: "",
+		},
+		{
+			name:              "normal: Content-Length header must equal file size",
+			actionHeaders:     map[string]string{},
+			wantContentLength: strconv.Itoa(len(content)),
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			var capturedContentLength string
+			var requestReceived bool
+
+			srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				requestReceived = true
+				capturedContentLength = r.Header.Get("Content-Length")
+				w.WriteHeader(http.StatusOK)
+			}))
+			defer srv.Close()
+
+			c := lfsapi.NewClient(lfshttp.NewContext(nil, nil, nil))
+			bu := &basicUploadAdapter{newAdapterBase(nil, BasicAdapterName, Upload, nil)}
+			bu.transferImpl = bu
+			bu.apiClient = c
+
+			transfer := &Transfer{
+				Oid:           "abc123def456",
+				Size:          int64(len(content)),
+				Path:          f.Name(),
+				Authenticated: true,
+				Actions: map[string]*Action{
+					"upload": {
+						Href:   srv.URL + "/upload",
+						Header: tc.actionHeaders,
+					},
+				},
+			}
+
+			err := bu.DoTransfer(nil, transfer, nil, nil)
+			require.Nil(t, err)
+			assert.True(t, requestReceived, "expected upload request to reach the server")
+			assert.Equal(t, tc.wantContentLength, capturedContentLength)
+		})
+	}
+}


tq/basic_upload.go
```diff
@@ -57,12 +57,12 @@ func (a *basicUploadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb Progres
 
 	if req.Header.Get("Transfer-Encoding") == "chunked" {
 		req.TransferEncoding = []string{"chunked"}
+		req.ContentLength = -1
 	} else {
 		req.Header.Set("Content-Length", strconv.FormatInt(t.Size, 10))
+		req.ContentLength = t.Size
 	}
```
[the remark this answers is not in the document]
telometto: Fixed by https://github.com/git-lfs/git-lfs/pull/6264/commits/3d392f7414c7183614f4ac27f15adfad735553df and https://github.com/git-lfs/git-lfs/pull/6264/commits/7868936195cb398288ab5e24426720386e514d1a

chrisd8088:
Hey, thanks very much for this PR!  I will aim to take a close look at it early next week.

It looks like the original support for chunked transfer encoding during object uploads was added back in PR #386.

One quick initial thought I had is that it might be valuable to add one or more tests to our shell test suite which exercise this functionality.  The relatively new `t/t-batch-storage-encoding.sh` script might be a good place for those; we'd likely also want to update the `t/cmd/lfstest-gitserver.go` program to support the new tests.

That said, the Go tests look relatively comprehensive already; thank you very much for including those in the PR!

tq/basic_upload.go
```diff
@@ -55,14 +55,14 @@ func (a *basicUploadAdapter) DoTransfer(ctx interface{}, t *Transfer, cb Progres
 		return err
 	}
 
-	if req.Header.Get("Transfer-Encoding") == "chunked" {
+	if isChunkedTransferEncoding(req.Header.Get("Transfer-Encoding")) {
 		req.TransferEncoding = []string{"chunked"}
+		req.ContentLength = -1
```
chrisd8088, line 60: This is the key change in this PR, which according to the PR's description is required because "(Go's) `net/http` client sends both `Transfer-Encoding: chunked` and `Content-Length`" when making HTTP/1.1 requests".

However, that appears not to be the case.  The `net/http` package's internal `transferWriter.shouldSendContentLength()` method [returns](https://github.com/golang/go/blob/go1.26.3/src/net/http/transfer.go#L255-L256) `false` if the encoding stack from the `TransferEncoding` field of an `http.Request` structure [contains](https://github.com/golang/go/blob/go1.26.3/src/net/http/transfer.go#L609) the `chunked` string, and in this case skips writing out a `Content-Length` header and only [writes](https://github.com/golang/go/blob/go1.26.3/src/net/http/transfer.go#L302-L304) a `Transfer-Encoding` header.

In other words, Go ignores the `http.Request` structure's `ContentLength` field entirely when we've set the `TransferEncoding` field with the `chunked` encoding.

This can be easily demonstrated by dropping all of this PR's proposed changes from our `tq/basic_upload.go` source file, but leaving the new `tq/basic_upload_test.go` file in place, and running our `make test` command.  The result is that the `chunked: Content-Length header must be absent` test [case](https://github.com/git-lfs/git-lfs/blob/7868936195cb398288ab5e24426720386e514d1a/tq/basic_upload_test.go#L31-L35) succeeds regardless of the fact that we leave the the `http.Request` structure's `ContentLength` field set to a positive value.

We can also see this in a Git LFS trace log created using the `GIT_TRACE=1 GIT_CURL_VERBOSE=1` environment variable settings; there's a `Transfer-Encoding: chunked` header but no `Content-Length` header:

```
> PUT /storage/baf4f8a7a095efcbcba242539c3f9298a3ac5fb00fff444cbd6b88ca7abfde7b?r=test-chunked-transfer-encoding HTTP/1.1
> Host: 127.0.0.1:64716
> User-Agent: git-lfs/3.7.0 (GitHub; darwin arm64; go 1.26.3; git d7917d8f)
> Transfer-Encoding: chunked
> Content-Type: text/plain; charset=utf-8
> Accept-Encoding: gzip
>
```

As for HTTP/2, when the server supports that, Go ignores the `http.Request` structure's `TransferEncoding` field entirely, as it should.  So if we add the `GODEBUG=http2debug=2` environment variable setting, we see the simulated HTTP/1.1 headers returned by the `net/http/httputil` package's `DumpRequestOut()` function, which always outputs them in in HTTP/1.x form (per the [documentation](https://pkg.go.dev/net/http/httputil#DumpRequest)), but also the actual HTTP/2 headers.  While `Transfer-Encoding: chunked` appears in the HTTP/1.1 version, what's actually sent by the client includes an HTTP/2 `content-length` header field:

```
> PUT /storage/baf4f8a7a095efcbcba242539c3f9298a3ac5fb00fff444cbd6b88ca7abfde7b?r=test-chunked-transfer-encoding HTTP/1.1
> Host: 127.0.0.1:64716
> User-Agent: git-lfs/3.7.0 (GitHub; darwin arm64; go 1.26.3; git d7917d8f)
> Transfer-Encoding: chunked
> Content-Type: text/plain; charset=utf-8
> Accept-Encoding: gzip
>
http2: Transport encoding header ":authority" = "127.0.0.1:64716"
http2: Transport encoding header ":method" = "PUT"
http2: Transport encoding header ":path" = "/storage/baf4f8a7a095efcbcba242539c3f9298a3ac5fb00fff444cbd6b88ca7abfde7b?r=test-chunked-transfer-encoding"
http2: Transport encoding header ":scheme" = "https"
http2: Transport encoding header "content-type" = "text/plain; charset=utf-8"
http2: Transport encoding header "user-agent" = "git-lfs/3.7.0 (GitHub; darwin arm64; go 1.26.3; git d7917d8f)"
http2: Transport encoding header "content-length" = "31"
http2: Transport encoding header "accept-encoding" = "gzip"
```

While it's true that HTTP/2 servers do not require a `content-length` header field, they are [permitted](https://datatracker.ietf.org/doc/html/rfc9113#section-8.1.1) so long as they accurately match the length of the data, and the [advice](https://stackoverflow.com/a/73371305) I've found suggests it's a reasonable practice to include a `content-length` header field with HTTP/2 requests.

tq/basic_upload.go
```diff
@@ -210,6 +210,18 @@ func newStartCallbackReader(r lfsapi.ReadSeekCloser, cb func() error) *startCall
...
+// isChunkedTransferEncoding reports whether the Transfer-Encoding header value
+// contains "chunked" as a transfer-coding token. Tokens are case-insensitive
+// and the header may be a comma-separated list per RFC 7230 §3.3.1.
+func isChunkedTransferEncoding(v string) bool {
+	for _, tok := range strings.Split(v, ",") {
+		if strings.EqualFold(strings.TrimSpace(tok), "chunked") {
+			return true
+		}
+	}
+	return false
```
chrisd8088, line 222: This function raises some interesting points.  We're not actually parsing a "real" HTTP header, but rather the header the Git LFS Batch API server has asked us to send back to it.

We should definitely use case-insensitive matching, unlike what we do now.

However, I'm not 100% sure we want to accept the header if it has any values other than `chunked`.  Go's `net/http` package only [allows](https://github.com/golang/go/blob/go1.26.3/src/net/http/transfer.go#L643-L652) a single encoding name, specifically `chunked`, for security reasons.  If a server asks us to send back `gzip, chunked`, it's asking us to both gzip-encode and chunk our uploads, which we don't support.

Should we then try to at least support the `chunked` encoding, or just ignore the header, as we effectively do now?  I can see arguments for both possibilities.

---

Assuming we want to parse the header more fully, I would suggest writing it as follows:

```suggestion
// and the header may be a comma-separated list per RFC 9110 section 6.1.
func isChunkedTransferEncoding(h http.Header) bool {
	v := h.Get("Transfer-Encoding")
	if v == "" {
		return false
	}

	return slices.ContainsFunc(strings.Split(v, ","), func(c string) bool {
		return strings.EqualFold(strings.TrimSpace(c), "chunked")
	})
```

This references the latest HTTP specification here, so that would be the "Transfer-Encoding" [section](https://datatracker.ietf.org/doc/html/rfc9112#name-transfer-encoding) of RFC 9112 (specifically, section 6.1).  It's substantially similar to the corresponding section in the earlier RFC, of course.

It also makes use of the `slices` Go package, new with Go v1.21, to avoid needing to write our own loop in this function.

As well, we expect the caller to pass the map of the HTTP headers and let the function extract the appropriate header's value.  This is the pattern we use in the `isTraceableContent()` [function](https://github.com/git-lfs/git-lfs/blob/d7917d8f2166e5d2d368fe5ade1332567cb99658/lfshttp/verbose.go#L149-L157) in our `lfshttp` package, for instance (which could also now be rewritten to make use of the `slices` package, as it happens!)

The caller would then pass `req.Header` instead of `req.Header.Get("Transfer-Encoding")`, so the function encapsulates all the necessary logic for reading and parsing the header.

tq/basic_upload_test.go
```diff
@@ -0,0 +1,89 @@
...
+	"strconv"
+	"testing"
+
+	"github.com/git-lfs/git-lfs/v3/lfsapi"
+	"github.com/git-lfs/git-lfs/v3/lfshttp"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBasicUploadAdapterContentLengthWithChunked(t *testing.T) {
```
chrisd8088, line 16: This Go test function is effective, but as noted above, two of its test cases already pass without any other changes to the Git LFS client, specifically the `chunked: Content-Length header must be absent` and `normal: Content-Length header must equal file size` cases.  The fact that the former passes indicates that we don't actually need to set `Request.ContentLength` to `-1` when performing chunked transfers.

Regardless, though, I think it might be better to fix, move, revise, and expand our legacy integration [test](https://github.com/git-lfs/git-lfs/blob/d7917d8f2166e5d2d368fe5ade1332567cb99658/t/t-chunked-transfer-encoding.sh#L5-L63) from PR #386, which is currently in the `t/t-chunked-transfer-encoding.sh` source file.

While reviewing this PR I discovered two problems with this existing test.  First, it is virtually identical to our sample [test](https://github.com/git-lfs/git-lfs/blob/d7917d8f2166e5d2d368fe5ade1332567cb99658/t/t-happy-path.sh#L7-L65) and therefore relies entirely for its efficacy on our test server [detecting](https://github.com/git-lfs/git-lfs/blob/d7917d8f2166e5d2d368fe5ade1332567cb99658/t/cmd/lfstest-gitserver.go#L1603-L1605) the test repository by name and [changing](https://github.com/git-lfs/git-lfs/blob/d7917d8f2166e5d2d368fe5ade1332567cb99658/t/cmd/lfstest-gitserver.go#L573-L579) its behaviour to force the client to perform chunked transfers.  This means the test itself doesn't actually guarantee that chunked transfers are performed.

That problem has masked another problem for eight (!) years now, which is that ever since we moved our shell integration tests from the `test` directory to the `t` directory and renamed the scripts to begin with a `t-` prefix, back in PR #3125, our `lfstest-gitserver` program has failed to detect the test repository's name and so doesn't force the client to perform chunked transfers at all.  The test server is [looking](https://github.com/git-lfs/git-lfs/blob/d7917d8f2166e5d2d368fe5ade1332567cb99658/t/cmd/lfstest-gitserver.go#L1604) for a repository name which begins with `test-chunked-transfer-encoding`.  However, the test [names](https://github.com/git-lfs/git-lfs/blob/d7917d8f2166e5d2d368fe5ade1332567cb99658/t/t-chunked-transfer-encoding.sh#L12) the repository it creates based on the name of the containing script, so when the script was renamed from `test-chunked-transfer-encoding.sh` to its current name, the server stopped treating the test any differently than usual.  Because the test doesn't do any of its own verification of the specific behaviour of the client (other than that it successfully uploads and downloads objects), the name change just caused the test to silently become ineffective.

Anyway, I'll try to fix those problems in a separate PR which also revises the test and moves it into the relatively new `t/t-batch-storage-encoding.sh` script, which I think is a more appropriate place for it now.

chrisd8088 (commented):
Hey, thanks again for this PR!  It led to some interesting research and turned up some issues we should resolve.

That said, I'm not certain there's any need to adopt the key change from this PR as it stands now.

As far as I can determine, when making HTTP/1.1 `PUT` requests, the Git LFS client already sends either a `Content-Length` header or a `Transfer-Encoding: chunked` header, but not both.  This is because Go's `net/http` package itself only [sends](https://github.com/golang/go/blob/go1.26.3/src/net/http/transfer.go#L291-L308) one or the other.  The new Go test function provided in this PR actually demonstrates this when it's run without any other changes from the PR.

Is there any chance your own environment has a proxy of some kind which is altering the headers of the Git LFS client's `PUT` requests, and adding a `Content-Length` header in addition to the `Transfer-Encoding: chunked` header sent by the client?

With respect to HTTP/2 requests, my reading of the specification indicates that providing a `content-length` header field is fine, so long as its value matches the length of the request data.

Moreover, `Transfer-Encoding` headers are not permitted with HTTP/2 at all, so if a Git LFS Batch API server is returning those in the `header` elements of its JSON responses to HTTP/2 requests, that's probably not something the client should consider valid, even as a "signal" that it should omit a `content-length` HTTP/2 header field in its requests.

If the problem here is that certain proxies are limiting `PUT` upload bodies based on their size, that seems to me to be an issue which should be raised with those projects, since it's entirely valid for an HTTP/2 request to specify its size with a `content-length` header field.

That all said, I will open a separate PR to fix the issues that turned up with our legacy chunked transfer encoding shell test [script](https://github.com/git-lfs/git-lfs/blob/d7917d8f2166e5d2d368fe5ade1332567cb99658/t/t-chunked-transfer-encoding.sh).

PR #6266, opened 2026-05-19
Reuse SSH sessions when possible and only close at client exit

### LINKED ISSUE #6118: could not get connection for batch request: pure SSH connection unavailable

chrisd8088:
This PR updates the Git LFS client so that where possible, SSH sessions are reused for multiple related operations.

This change will permit the "pure" SSH-based Git LFS transfer protocol to be used when pushing or fetching multiple branches, resolving the issue reported in #6118.

As well, this PR revises the Git LFS client so that when exiting, it always tries to cleanly terminate any SSH sessions that were started for the "pure" SSH-based transfer protocol.

This PR will be most easily reviewed on a commit-by-commit basis, with whitespace differences ignored.

#### SSH Session Reuse

When we introduced support for the "pure" SSH-based Git LFS object transfer protocol in PR #4446, we designed the `SSHTransfer` [structure](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/ssh/connection.go#L15-L24) in our `ssh` package to be the basic abstraction with which we represent and manage SSH processes and connections for the protocol.

In commit 31d3fb7ceeef219bc7ee466d5938306000de052d of the same PR we added an `SSHTransfer()` [method](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfsapi/lfsapi.go#L43-L67) to the `Client` structure of our `lfsapi` package in order to provide a common interface through which we could initiate and share SSH sessions.  This method always returns either a new `SSHTransfer` structure, representing a new SSH session and underlying set of connections, or else returns a `nil` value to indicate that the "pure" SSH-based Git LFS protocols are not supported by the remote service, or that an error condition has occurred.
    
Because the `SSHTransfer()` method always returns a new SSH session (when it does not return `nil`), at present we [create](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/tq/manifest.go#L197) [separate](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/locking/api.go#L344) sessions for both object transfer requests and locking requests.

Further, as reported in #6118, when a transfer queue completes we shut down any SSH session it was using, but then we attempt to reuse that session in subsequent transfer queues, which causes an error.

Specifically, when uploading or downloading objects for multiple Git references, we [establish](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/commands.go#L69) a single transfer manifest for the entire operation.  The first time one of the methods of the `lazyManifest` structure in our `tq` package is invoked, it creates a `concreteManifest` structure by [calling](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/tq/manifest.go#L123) the `newConcreteManifest()` function, which [calls](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/tq/manifest.go#L197) the `SSHTransfer()` function of our `lfsapi` package.  Thus if that function returns a non-`nil` value (i.e., an `SSHTransfer` structure), then the manifest retains a pointer to that structure for the lifetime of the Git LFS client process.
    
We then pass that manifest to the `NewTransferQueue()` function [each](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/uploader.go#L122-L126) [time](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/command_filter_process.go#L106-L112) we start a new transfer queue.  After all the objects to be transferred have been added to the queue, we [then](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/uploader.go#L247) [call](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/command_filter_process.go#L154) the `Wait()` method of the `TransferQueue` structure.
    
When all the object transfers are complete, if the queue's manifest contains non-`nil` `sshTransfer` field, then the `Wait()` method [calls](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/tq/transfer_queue.go#L996-L999) the `Shutdown()` method of the `SSHTransfer` structure, which closes the underlying SSH connections.
    
The consequence is that if we then attempt to start another queue for the same operation and remote endpoint, we do attempt to reuse the same `SSHTransfer` structure and associated SSH session, but all of its SSH processes and connections have been terminated.  This leads to the error reported in #6118.

To avoid this problem, we first refactor the `SSHTransfer()` method in the `lfsapi` package into two methods.  The new `initSSHTransfer()` method performs the same actions as the `SSHTransfer()` method did before, and as such may return
a `nil` value if the remote service does not support the `git-lfs-transfer` command.

The `SSHTransfer()` method now maintains a map of the SSH sessions for which it has called the `initSSHTransfer()` method, and only calls that method if it is required.  Otherwise, the value returned by the `initSSHTransfer()` method for the given `operation` and `remote` parameters is returned again, even if it is a `nil` value.

To maintain this map safely, the` SSHTransfer()` method first acquires a mutex, which we add to the `Client` structure along with the map.  For keys, we use a concatenation of the unique pairs of `operation` and `remote` strings passed to the method as its parameters.  This technique follows the existing approach used in our `commands` package, where we [construct](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/commands.go#L67) keys for the global `tqManifest` map by concatenating the `operation` and `remote` string parameters of the `getTransferManifestOperationRemote()` function.

One advantage of this change is that it also resolves the problem whereby we would previously create an extra SSH process and connection for each locking request, even if they were to be made to the same remote service as subsequent object transfer requests, and for the same type of operation (i.e., an upload or download operation).

Another advantage is that when the Git LFS client is exiting, we can identify the set of SSH sessions that were opened for the SSH-based transfer protocol and attempt to terminate them all properly, instead of closing the individual session used for a transfer queue after that queue has completed.

To do this, we add a `closeSSHTransfers()` method to the `Client` structure in the `lfsapi` package, and then invoke this new method from the `Client` structure's `Close()` [method](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfsapi/client.go#L53-L55).  The `Close()` method is [called](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/commands.go#L91) by the `commands` package's `closeAPIClient()` function when a Git LFS command has [completed](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/run.go#L155) normally.

Our new `closeSSHTransfers()` method iterates over all the non-`nil` values in the map maintained by the `SSHTransfer()` method and calls the `Shutdown()` method of each `SSHTransfer` structure, thus attempting to terminate them properly.  We can then revise the `Wait()` method of the `TransferQueue structure` so that it no longer attempts to shut down the SSH session associated with the queue's manifest.
    
Collectively, these revisions and enhancements should resolve the problem reported in #6118.  To verify this, we add a new test to our `t/t-batch-transfer.sh` shell script which runs each of the `git push` and `git lfs fetch` commands for multiple branches, causing both commands to start and stop more than a single transfer queue.

#### Consistent Exit Handling

The changes described above ensure we reuse SSH sessions for the "pure" SSH-based Git LFS transfer protocols, and that we only attempt to terminate them when the client is exiting, at least under normal circumstances.

However, if [certain](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/command_filter_process.go#L184) [errors](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/lockverifier.go#L75) occur or if one of the signals we [trap](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/git-lfs.go#L16) is received, then the Git LFS client may halt immediately, and in these cases we would like it to at least attempt to terminate any SSH sessions started for the SSH-based transfer protocol.

In PR #6255 we revised all of the individual command implementation functions in our `commands` package so that they always invoke one of a small set of exit wrapper functions when they need to force the client to halt.  We can now take advantage of this fact to ensure that regardless of circumstances, the Git LFS client will always attempt to terminate any active SSH sessions that were opened for the "pure" SSH-based Git LFS transfer protocols.

Most obviously, we add a call to the `Cleanup()` [function](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/commands.go#L282-L286) of our `commands` package to each of the exit wrapper functions, just before they call the `Exit()` function of the Go standard library's `os` package.

Next, we refactor the `Cleanup()` function into two functions, with the actual cleanup activities performed by a new `doCleanup()` function, which is then invoked by the `Cleanup()` function exactly one time only, via the `Do()` method of a `Once` structure from the `sync` package of the Go standard library.  We assign this `Once` structure to a new `cleanupOnce` global variable in our `commands` package, which replaces the `once` [variable](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/git-lfs.go#L18) of the same type in our `main` package.  In our `main()` function, we then just call the `commands` package's `Cleanup()` directly in the [two](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/git-lfs.go#L23) [instances](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/git-lfs.go#L35) where we previously did so via the `once` variable's `Do()` method.

Finally, we add to the `doCleanup()` function a call to the `closeAPIClient()` [function](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/commands.go#L85-L92), which acquires the necessary `global` mutex before invoking the global `apiClient` variable's `Close()` method.  This supplants the legacy [call](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/run.go#L155) to the `closeAPIClient()` function from our `Run()` function, so we just remove that call.

#### Other Changes

In addition to the principal changes described above, this PR also makes a number of small refinements, including:

- We fix the SSH connection IDs in the trace log messages we [output](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/ssh/connection.go#L170) when uninitialized connections are ignored while reducing the number of active connections.
- We output new messages to help clarify when lock [verification](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/lockverifier.go#L49-L87) begins and ends in our trace logs.
- We update our existing [tests](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/t/t-batch-transfer.sh#L235-L356) of the "pure" SSH-based Git LFS object transfer protocol to perform additional checks and to make use of the `setup_expected_concurrent_transfers()` test helper [function](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/t/testhelpers.sh#L1088-L1097) we added in PR [#6241](https://github.com/git-lfs/git-lfs/pull/6241).
- We rename [one](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/t/t-batch-transfer.sh#L313-L356) of these tests and revise our `assert_ssh_transfer_sessions()` test helper [function](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/t/t-batch-transfer.sh#L178-L233) to correctly reflect the conditions checked by the test, in which the number of SSH connections is limited to fewer than the number of objects to be transferred in a single batch.  This causes the transfer of some objects to be delayed, but still performed within the same batch as the others (and to be performed by the same transfer queue).
- We revise a number of our Go test functions and secondary programs so that we consistently call the `Close()` [method](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfsapi/client.go#L53-L55) of the `lfsapi` package's `Client` structure type, whenever such a structure has reached the end of its lifetime.  For the same reason, we update one instance in the Git LFS client where we [create](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/lfs/lfs.go#L22) a temporary `Client` structure, and eliminate another similar [instance](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/tq/manifest.go#L193-L195) which exists solely to support some of our Go test functions.

Resolves #6118.
/cc @galets as reporter.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -39,9 +39,10 @@ var (
 	ManPages     = make(map[string]string, 20)
 	tqManifest   = make(map[string]tq.Manifest)
 
-	cfg       *config.Configuration
-	apiClient *lfsapi.Client
-	global    sync.Mutex
+	cfg         *config.Configuration
+	apiClient   *lfsapi.Client
+	global      sync.Mutex
+	cleanupOnce sync.Once
 
 	oldEnv = make(map[string]string)
 
@@ -280,6 +281,10 @@ func Panic(err error, format string, args ...interface{}) {
 }
 
 func Cleanup() {
+	cleanupOnce.Do(doCleanup)
+}
+
+func doCleanup() {
 	if err := cfg.Cleanup(); err != nil {
 		fmt.Fprintln(os.Stderr, tr.Tr.Get("Error clearing old temporary files: %s", err))
 	}
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -288,6 +288,10 @@ func Cleanup() {
 }
 
 func doCleanup() {
+	if err := closeAPIClient(); err != nil {
+		fmt.Fprintln(os.Stderr, tr.Tr.Get("Error closing API client: %s", err))
+	}
+
 	if err := cfg.Cleanup(); err != nil {
 		fmt.Fprintln(os.Stderr, tr.Tr.Get("Error clearing old temporary files: %s", err))
 	}
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -206,6 +206,7 @@ func uninstallHooks() error {
 
 // ExitWithCode exits immediately with the given code.
 func ExitWithCode(code int) {
+	Cleanup()
 	os.Exit(code)
 }
 
@@ -232,6 +233,7 @@ func Print(format string, args ...interface{}) {
 // Exit prints a formatted message and exits.
 func Exit(format string, args ...interface{}) {
 	Error(format, args...)
+	Cleanup()
 	os.Exit(2)
 }
 
@@ -277,6 +279,7 @@ func LoggedError(err error, format string, args ...interface{}) {
 // a log file before exiting.
 func Panic(err error, format string, args ...interface{}) {
 	LoggedError(err, format, args...)
+	Cleanup()
 	os.Exit(2)
 }
 
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -89,7 +89,11 @@ func closeAPIClient() error {
 	if apiClient == nil {
 		return nil
 	}
-	return apiClient.Close()
+
+	err := apiClient.Close()
+	apiClient = nil
+
+	return err
 }
 
 func newLockClient() *locking.Client {
--- a/lfsapi/lfsapi.go
+++ b/lfsapi/lfsapi.go
@@ -65,8 +65,11 @@ func (c *Client) SSHTransfer(operation, remote string) *ssh.SSHTransfer {
 
 	sshTransfer, ok := c.sshTransfers[k]
 	if !ok {
-		sshTransfer = c.initSSHTransfer(operation, remote)
-		c.sshTransfers[k] = sshTransfer
+		var err error
+		sshTransfer, err = c.initSSHTransfer(operation, remote)
+		if err == nil {
+			c.sshTransfers[k] = sshTransfer
+		}
 	}
 
 	return sshTransfer
@@ -75,16 +78,16 @@ func (c *Client) SSHTransfer(operation, remote string) *ssh.SSHTransfer {
 // initSSHTransfer returns either an suitable transfer object or nil if the
 // server is not using an SSH remote or the git-lfs-transfer style of SSH
 // remote.
-func (c *Client) initSSHTransfer(operation, remote string) *ssh.SSHTransfer {
+func (c *Client) initSSHTransfer(operation, remote string) (*ssh.SSHTransfer, error) {
 	endpoint := c.Endpoints.Endpoint(operation, remote)
 	if len(endpoint.SSHMetadata.UserAndHost) == 0 {
-		return nil
+		return nil, nil
 	}
 
 	uc := config.NewURLConfig(c.context.GitEnv())
 	if val, ok := uc.Get("lfs", endpoint.OriginalUrl, "sshtransfer"); ok && val != "negotiate" && val != "always" {
 		tracerx.Printf("skipping pure SSH protocol connection by request (%s, %s)", operation, remote)
-		return nil
+		return nil, nil
 	}
 
 	ctx := c.Context()
@@ -92,10 +95,10 @@ func (c *Client) initSSHTransfer(operation, remote string) *ssh.SSHTransfer {
 	sshTransfer, err := ssh.NewSSHTransfer(ctx.OSEnv(), ctx.GitEnv(), &endpoint.SSHMetadata, operation)
 	if err != nil {
 		tracerx.Printf("pure SSH protocol connection failed (%s, %s): %s", operation, remote, err)
-		return nil
+		return nil, err
 	}
 
-	return sshTransfer
+	return sshTransfer, nil
 }
 
 func (c *Client) closeSSHTransfers() error {
--- a/lfsapi/lfsapi.go
+++ b/lfsapi/lfsapi.go
@@ -1,6 +1,9 @@
 package lfsapi
 
 import (
+	"fmt"
+	"sync"
+
 	"github.com/git-lfs/git-lfs/v3/config"
 	"github.com/git-lfs/git-lfs/v3/creds"
 	"github.com/git-lfs/git-lfs/v3/lfshttp"
@@ -17,6 +20,9 @@ type Client struct {
 	client  *lfshttp.Client
 	context lfshttp.Context
 	access  []creds.AccessMode
+
+	sshTransfers      map[string]*ssh.SSHTransfer
+	sshTransfersMutex *sync.Mutex
 }
 
 func NewClient(ctx lfshttp.Context) *Client {
@@ -33,6 +39,9 @@ func NewClient(ctx lfshttp.Context) *Client {
 		context:     ctx,
 		credContext: creds.NewCredentialHelperContext(gitEnv, osEnv),
 		access:      creds.AllAccessModes(),
+
+		sshTransfers:      make(map[string]*ssh.SSHTransfer),
+		sshTransfersMutex: &sync.Mutex{},
 	}
 }
 
@@ -40,22 +49,43 @@ func (c *Client) Context() lfshttp.Context {
 	return c.context
 }
 
-// SSHTransfer returns either an suitable transfer object or nil if the
-// server is not using an SSH remote or the git-lfs-transfer style of SSH
-// remote.
+// SSHTransfer returns either an suitable transfer object or nil.  For a
+// given operation and remote, the same transfer object (or nil) will
+// always be returned.
 func (c *Client) SSHTransfer(operation, remote string) *ssh.SSHTransfer {
 	if len(operation) == 0 {
 		return nil
 	}
+
+	k := fmt.Sprintf("%s.%s", operation, remote)
+
+	c.sshTransfersMutex.Lock()
+	defer c.sshTransfersMutex.Unlock()
+
+	sshTransfer, ok := c.sshTransfers[k]
+	if !ok {
+		sshTransfer = c.initSSHTransfer(operation, remote)
+		c.sshTransfers[k] = sshTransfer
+	}
+
+	return sshTransfer
+}
+
+// initSSHTransfer returns either an suitable transfer object or nil if the
+// server is not using an SSH remote or the git-lfs-transfer style of SSH
+// remote.
+func (c *Client) initSSHTransfer(operation, remote string) *ssh.SSHTransfer {
 	endpoint := c.Endpoints.Endpoint(operation, remote)
 	if len(endpoint.SSHMetadata.UserAndHost) == 0 {
 		return nil
 	}
+
 	uc := config.NewURLConfig(c.context.GitEnv())
 	if val, ok := uc.Get("lfs", endpoint.OriginalUrl, "sshtransfer"); ok && val != "negotiate" && val != "always" {
 		tracerx.Printf("skipping pure SSH protocol connection by request (%s, %s)", operation, remote)
 		return nil
 	}
+
 	ctx := c.Context()
 	tracerx.Printf("attempting pure SSH protocol connection (%s, %s)", operation, remote)
 	sshTransfer, err := ssh.NewSSHTransfer(ctx.OSEnv(), ctx.GitEnv(), &endpoint.SSHMetadata, operation)
@@ -63,5 +93,6 @@ func (c *Client) SSHTransfer(operation, remote string) *ssh.SSHTransfer {
 		tracerx.Printf("pure SSH protocol connection failed (%s, %s): %s", operation, remote, err)
 		return nil
 	}
+
 	return sshTransfer
 }
--- a/lfsapi/lfsapi.go
+++ b/lfsapi/lfsapi.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/git-lfs/git-lfs/v3/config"
 	"github.com/git-lfs/git-lfs/v3/creds"
+	"github.com/git-lfs/git-lfs/v3/errors"
 	"github.com/git-lfs/git-lfs/v3/lfshttp"
 	"github.com/git-lfs/git-lfs/v3/ssh"
 	"github.com/rubyist/tracerx"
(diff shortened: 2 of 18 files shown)


t/t-batch-transfer.sh
```diff
@@ -255,12 +255,16 @@ begin_test "batch transfers with ssh endpoint (git-lfs-transfer)"
   # enforce their use in order to match other platforms' connection counts.
   git config --global lfs.ssh.autoMultiplex true
 
-  GIT_TRACE=1 git push origin main >push.log 2>&1
+  GIT_TRACE=1 git push origin main 2>&1 | tee push.log
+  [ 0 -eq "${PIPESTATUS[0]}" ]
```
larsxschneider, line 253: This captures the exit status for all SSH tests as the commit message describes it. Should we do the same for other tests? (e.g. https://github.com/git-lfs/git-lfs/blame/b84a8a68c81e1710ec456922ee9eac615458104a/t/t-batch-transfer.sh#L91 )

commands/commands.go
```diff
@@ -276,10 +279,19 @@ func LoggedError(err error, format string, args ...interface{}) {
...
+	Cleanup()
 	os.Exit(2)
 }
 
 func Cleanup() {
+	cleanupOnce.Do(doCleanup)
+}
+
+func doCleanup() {
+	if err := closeAPIClient(); err != nil {
```
larsxschneider, line 295: Should we set `apiClient = nil` within ` closeAPIClient()` for extra safety?

lfsapi/lfsapi.go
```diff
@@ -33,35 +40,76 @@ func NewClient(ctx lfshttp.Context) *Client {
...
+
+	k := fmt.Sprintf("%s.%s", operation, remote)
+
+	c.sshTransfersMutex.Lock()
+	defer c.sshTransfersMutex.Unlock()
+
+	sshTransfer, ok := c.sshTransfers[k]
+	if !ok {
+		sshTransfer = c.initSSHTransfer(operation, remote)
+		c.sshTransfers[k] = sshTransfer
```
larsxschneider: When this fails, then `c.sshTransfers[k]` is set to `nil`. Does this mean a transient error would be permanent for a given command invocation? Previously lock verification and object transfers each called `initSSHTransfer()` independently, so a transient failure in one wouldn't affect the other. Now they share the cached result.

larsxschneider (approved):
LGTM! Thank you 🙇

t/t-batch-transfer.sh
chrisd8088: I thought about that, most especially the [test](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/t/t-batch-transfer.sh#L143-L164) of the `git-lfs-authenticate` flow (i.e., using SSH only to acquire HTTP credentials), but decided to keep the scope of the PR somewhat more focused.

There are certainly lots of instances in our test suite where we could standardize how we perform command checks, but I figure that's best left for a more comprehensive PR that tackles all our tests at once.

commands/commands.go
chrisd8088: That's a good idea!  Done.

lfsapi/lfsapi.go
chrisd8088: Good question.  Given the current state of affairs, it wouldn't make much difference, because of a quirk in the sequence by which we try to establish an SSH session for the "pure" Git LFS protocols.

In brief, we [perform](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/uploader.go#L24) lock verification requests when uploading, and those [currently](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/commands/lockverifier.go#L164) [cause](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/tq/manifest.go#L76) the transfer queue manifest for the object transfers to be [upgraded](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/tq/manifest.go#L123) first, which results in a [call](https://github.com/git-lfs/git-lfs/blob/14ae1707938bf0a7b77e5bfe333b196574bbc843/tq/manifest.go#L197) to the `SSHTransfer()` method to create an SSH session for the manifest.

So if that initial call to the `SSHTransfer()` method succeeds in establishing an SSH session, the session will now be cached and used for the both the lock and object transfer requests.

On the other hand, if the call doesn't succeed, then the object transfer requests would eventually fail regardless of whether the lock verification ones do or don't.  (Note that unless `lfs.locksVerify` is explicitly set to `true`, we allow lock verification to "fail open" and the upload operation to proceed.)

Nevertheless, it's not a bad idea to try to be a bit more tolerant of errors when trying to establish an SSH session.  I've revised the `initSSHTransfer()` method to return an `error` type, and if it's non-`nil`, the `SSHTransfer()` method now skips caching the result.

This should mean that if, in the future, we change the sequence of events during upload operations, we might benefit from the lack of caching after error conditions.  For instance, if the `SSHTransfer()` method were to be called specifically for the lock verification requests first, those could all fail to establish a session and so long as `lfs.locksVerify` was not `true`, that wouldn't prevent the object transfer requests from potentially being able to establish a session and proceed.

PR #6269, opened 2026-05-27
chore: improve git-lfs maintenance path

otjdiepluong:
Summary:
- Add or tighten focused edge-case tests or type assertions in fs/fs_test.go, tq/api_test.go related to CI, Docker, Kubernetes, build tooling, release automation; avoid docs-only changes and broad refactors.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/fs/fs_test.go
+++ b/fs/fs_test.go
@@ -42,3 +42,9 @@ func TestRepositoryPermissions(t *testing.T) {
 		assert.Equal(t, v, fs.RepositoryPermissions(false))
 	}
 }
+
+func TestRepositoryPermissionsZero(t *testing.T) {
+	fs := Filesystem{repoPerms: 0}
+	assert.Equal(t, os.FileMode(0), fs.RepositoryPermissions(false))
+	assert.Equal(t, os.FileMode(0), fs.RepositoryPermissions(true))
+}
--- a/tq/api_test.go
+++ b/tq/api_test.go
@@ -18,8 +18,8 @@ import (
 )
 
 func TestAPIBatch(t *testing.T) {
-	require.NotNil(t, batchReqSchema, batchReqSchema.Source)
-	require.NotNil(t, batchResSchema, batchResSchema.Source)
+	require.NotNil(t, batchReqSchema)
+	require.NotNil(t, batchResSchema)
 
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.URL.Path != "/api/objects/batch" {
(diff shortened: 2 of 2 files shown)


tq/api_test.go
```diff
@@ -18,8 +18,8 @@ import (
 )
 
 func TestAPIBatch(t *testing.T) {
-	require.NotNil(t, batchReqSchema, batchReqSchema.Source)
-	require.NotNil(t, batchResSchema, batchResSchema.Source)
+	require.NotNil(t, batchReqSchema)
+	require.NotNil(t, batchResSchema)
```
chrisd8088, line 22: I assume the reasoning here is that `batchReqSchema` were to be `nil`, then trying to dereference its `Source` element would lead to a panic?  If so, this seems like a reasonable fix to me.

fs/fs_test.go
```diff
@@ -42,3 +42,9 @@ func TestRepositoryPermissions(t *testing.T) {
 		assert.Equal(t, v, fs.RepositoryPermissions(false))
 	}
 }
+
+func TestRepositoryPermissionsZero(t *testing.T) {
+	fs := Filesystem{repoPerms: 0}
+	assert.Equal(t, os.FileMode(0), fs.RepositoryPermissions(false))
+	assert.Equal(t, os.FileMode(0), fs.RepositoryPermissions(true))
+}
```
chrisd8088, line 50: I'm not quite sure I understand the value of this separate test.  What would be the circumstances where we explicitly need to confirm that when passed zero, we [map](https://github.com/git-lfs/git-lfs/blob/d7917d8f2166e5d2d368fe5ade1332567cb99658/tools/filetools.go#L462) that to zero?

The preceding `TestRepositoryPermissions()` function already demonstrates that `0700` is mapped to `0600`, so we already know zero bits are left unchanged in most positions.

If this check is truly required, though, or just desired for completeness, we only need to add a single line to the preceding test's map of test cases, e.g.:

```patch
--- a/fs/fs_test.go
+++ b/fs/fs_test.go
@@ -35,6 +35,7 @@ func TestRepositoryPermissions(t *testing.T) {
 		0777: 0666,
 		0755: 0644,
 		0700: 0600,
+		0000: 0000,
 	}
 	for k, v := range m {
 		fs := Filesystem{repoPerms: v}
```

chrisd8088 (commented):
Hey, thanks for this PR.

Would you be able to explain what its purpose is in a bit more detail?  I'm not sure how these minor test enhancements pertain to "CI, Docker, Kubernetes, build tooling, release automation", for instance.

Regardless, the changes in the `tq/api_test.go` source file seem reasonable, but the ones in the `fs/fs_test.go` source file seem like something we might not need, or could implement with a single extra line in an existing test.

PR #1375, opened 2016-07-15
teach script/integration.go about GIT_LFS_TEST_MAXPROCS

technoweenie:
Found an intermittent test issue running `script/cibuild` on mac. While investigating, I noticed that `GIT_LFS_TEST_MAXPROCS` isn't even used now. The intermittent issue goes away when `GIT_LFS_TEST_MAXPROCS=1` too.

PR #6273, opened 2026-06-02
Fix and revise chunked transfer encoding test

chrisd8088:
This PR replaces the legacy, non-functional test of the Git LFS client's support for chunked transfer encoding during object uploads with a test which verifies that the client correctly sends a `Transfer-Encoding: chunked` header when instructed to do so by a Git LFS Batch API service, and does not send a `Content-Length` header at the same time.

As well, this PR adds a second test which confirms that the client does not send a `Transfer-Encoding` header when the Batch API service has not requested one.

Finally, this test adjusts several functions in our test server utility and also updates the related shell tests to make their repository names more consistent and to help avoid problems similar to the ones which caused our legacy chunked transfer encoding test to be ineffective.

#### First Legacy Test Problem

When support for chunked transfer encoding during object uploads was first introduced in PR #386, we added the `chunked transfer encoding` [test](https://github.com/git-lfs/git-lfs/blob/cc3a8c45d1aca62aedab3cf6fd937c2639784708/t/t-chunked-transfer-encoding.sh#L5-L63) in what is now our `t/t-chunked-transfer-encoding.sh` shell script.  This test is essentially identical to the generic, sample `happy path` [test](https://github.com/git-lfs/git-lfs/blob/cc3a8c45d1aca62aedab3cf6fd937c2639784708/t/t-happy-path.sh#L7-L94) in our `t/t-happy-path.sh` script, and as such relies on our `lfstest-gitserver` utility to detect the name of the test's repository and alter its behaviour to validate that the client sends a `Transfer-Encoding: chunked` header when requested to do so.

However, while the Batch API handler in our server utility used to detect the test's repository name correctly, it appears that the object upload handler has never done so, and so the test has been ineffective since it was first written. 

Both handlers [make](https://github.com/git-lfs/git-lfs/blob/cc3a8c45d1aca62aedab3cf6fd937c2639784708/t/cmd/lfstest-gitserver.go#L469) [calls](https://github.com/git-lfs/git-lfs/blob/cc3a8c45d1aca62aedab3cf6fd937c2639784708/t/cmd/lfstest-gitserver.go#L767) to the `testingChunkedTransferEncoding()` function to check for a special repository name and this function [compares](https://github.com/git-lfs/git-lfs/blob/cc3a8c45d1aca62aedab3cf6fd937c2639784708/t/cmd/lfstest-gitserver.go#L1604) the URL path of the current request to the prefix `/test-chunked-transfer-encoding`.  However, only Batch API requests to our test server begin with the repository name as an initial URL path component; the URLs for individual objects [begin](https://github.com/git-lfs/git-lfs/blob/cc3a8c45d1aca62aedab3cf6fd937c2639784708/t/cmd/lfstest-gitserver.go#L273) with `/storage/` instead.

So as the test was originally written, the Batch API handler would have [received](https://github.com/git-lfs/git-lfs/blob/962be37a79f7c2575c832cd9af0faf16c01c25e9/test/cmd/lfstest-gitserver.go#L212) a `true` value from the `testingChunkedTransferEncoding()` [function](https://github.com/git-lfs/git-lfs/blob/962be37a79f7c2575c832cd9af0faf16c01c25e9/test/cmd/lfstest-gitserver.go#L353-L354) during the test, and would then have [instructed](https://github.com/git-lfs/git-lfs/blob/962be37a79f7c2575c832cd9af0faf16c01c25e9/test/cmd/lfstest-gitserver.go#L225-L227) the client to send a `Transfer-Encoding: chunked` header when uploading.  The subsequent object upload request, however, would have a URL path which [began](https://github.com/git-lfs/git-lfs/blob/962be37a79f7c2575c832cd9af0faf16c01c25e9/test/cmd/lfstest-gitserver.go#L219) with `/storage/` and so the object storage handler would never have [checked](https://github.com/git-lfs/git-lfs/blob/962be37a79f7c2575c832cd9af0faf16c01c25e9/test/cmd/lfstest-gitserver.go#L251-L262) whether a `Transfer-Encoding` header was sent with the object upload `PUT` request.

This problem still persists today, because the URL paths generated by our test server utility for individual object requests still begin with `/storage/`, and the `testingChunkedTransferEncoding()` function still checks for a prefix of `/test-chunked-transfer-encoding`.  We resolve this issue now by passing just a repository name to the `testingChunkedTransferEncoding()` function rather than an URL path, since both the Batch API and object storage handlers have a local `repo` variable which they populate with the repository name based on information in the request. 

#### Second Legacy Test Problem

The original design of our legacy test expected the storage handler in our test server utility to [halt](https://github.com/git-lfs/git-lfs/blob/962be37a79f7c2575c832cd9af0faf16c01c25e9/test/cmd/lfstest-gitserver.go#L260) with a fatal error if it found that no `Transfer-Encoding: chunked` header had been sent when one was expected (although as noted above, the server would never actually perform this check at all).

In commit e2046fd8b7d6f08fd6b6b3fcfbec4f4cb66761d3 of PR #1375 we introduced the `debug()` helper function to our server utility and revised many of our calls to the Go standard library's `log` package to call our new helper function instead.  In most cases this was an innocuous change, but in the particular instance of the check in the storage handler for a `Transfer-Encoding` header, the server was intended to exit if no header was found when one was expected, with no alternative indication that an error had occurred.

The object storage handler in our test server utility still only [outputs](https://github.com/git-lfs/git-lfs/blob/cc3a8c45d1aca62aedab3cf6fd937c2639784708/t/cmd/lfstest-gitserver.go#L776) a message to its debug log if no `Transfer-Encoding` header is found when one is expected, so our new test would never receive any indication of an error condition if the client failed to send a `Transfer-Encoding` header unless we resolve this problem.  We therefore update the object storage handler so that in such a case, it will now return a 400 Bad Request status code to the client, which would cause the client to retry the request and then ultimately report an error, resulting in a test failure.

We also design our new shell test such that it independently verifies that a `Transfer-Encoding: chunked` header was sent by the Git LFS client in the object upload `PUT` request, and that no `Content-Length` header was sent in this request.  This guarantees that even we introduce a regression in the test server in the future, the test itself should be able to verify whether the client sends `Transfer-Encoding: chunked` headers when instructed to do so.

#### Third Legacy Test Problem

As described in https://github.com/git-lfs/git-lfs/pull/6264#discussion_r3327568836, in PR #3125 we renamed our shell test scripts so all their names now begin with `t-` rather than `test-`, and in doing so we inadvertently added a third problem which made our legacy `chunked transfer encoding` test fully ineffective.

The `chunked transfer encoding` test [derives](https://github.com/git-lfs/git-lfs/blob/cc3a8c45d1aca62aedab3cf6fd937c2639784708/t/t-chunked-transfer-encoding.sh#L12) the name of its primary repository from the name of the enclosing shell script, so the repository name is now `t-chunked-transfer-encoding`.  Because the `testingChunkedTransferEncoding()` function compares the leading portion of the URL path to the prefix `test-chunked-storage-encoding`, it currently never returns `true`, even when called from the server utility's Batch API handler.

We resolve this issue in our new test by using a fixed string for the name of the repository, and then changing the `testingChunkedTransferEncoding()` function to compare the given repository name to exactly that string.

We also revise the three other [functions](https://github.com/git-lfs/git-lfs/blob/cc3a8c45d1aca62aedab3cf6fd937c2639784708/t/cmd/lfstest-gitserver.go#L1607-L1615) in our test server utility that have been modelled on the design of the `testingChunkedTransferEncoding()` function.  We change each function's input parameter to a simple repository name, and where possible we refine the comparisons performed by the functions.  We also take the opportunity to update the names of the repositories created by the related shell tests so they are more consistent with those of the other tests in the same scripts.

/cc PR #6264


t/t-batch-storage-encoding.sh
```diff
@@ -301,3 +301,75 @@ begin_test "batch storage HTTP download with invalid encoding"
...
+
+  # To confirm that when the object was uploaded, the PUT request included a
+  # Transfer-Encoding header and did not include a Content-Length header,
+  # we filter out the initial portion of the trace log with the headers
+  # from the Batch and Locking API requests.
+  prog="$(printf 'index($0, " PUT %s/storage/") > 0 { f=1 } f' "$GITSERVER")"
+  awk "$prog" push.log >put.log
+
+  grep "> Transfer-Encoding: chunked" put.log
+  [ 0 -eq "$(grep -c -i "> Content-Length:" put.log)" ]
```
larsxschneider, line 337: The awk program feels fragile but I am unable to come up with something better.

larsxschneider (approved):

t/t-batch-storage-encoding.sh
chrisd8088: I thought about this too, but I think it's OK.

For one thing, we have a positive check in that `Transfer-Encoding` must appear in the resulting file, so if the `awk` call started failing to match anything and `put.log` ended up empty, we'd get a failure and could go fix the test.  And if we start matching too early and `put.log` ends up with multiple requests' worth of trace logs, the negative check for `Content-Length` would likely start failing because it would find those headers from the Batch and Locking API requests.

PR #6274, opened 2026-06-02
Add corrupt object download tests

chrisd8088:
This PR adds tests which confirm that the Git LFS client verifies the contents of the objects it downloads from remote services, and that the client does not save an object to the local filesystem if the object's contents do not match its expected SHA-256 hash value.

Each of the download transfer adapters we support already verify the contents of the objects they download.

At present, however, we do not have tests which confirm these checks are effective, so we add those tests now to a new `t/t-batch-storage-corrupt.sh` shell script.

#### HTTP Downloads

To test our default `basic` download adapter, which makes object transfer requests using HTTP, we add a `batch storage HTTP download of corrupt object` test which first creates an object with the special content key `storage-download-corrupt` and uploads it to our `lfstest-gitserver` test server.  The test then removes the object from the local storage directories and the working tree, and attempts to download it using a `git lfs pull` command.

We also adjust the `lfstest-gitserver` program so that when handling a `GET` request for an object, if the object's content matches the special key `storage-download-corrupt`, the content is replaced with an uppercase version of the same key before being provided to the client.  Our new test then checks that the `basic` download adapter detects the invalid object content, reports an error, and does not write the object content into the local storage directories.

#### SSH Downloads

For our `ssh` download adapter, which makes object transfer requests using SSH, we add a `batch storage SSH download of corrupt object` test that works in a similar fashion to the test of the `basic` adapter, except that in this case the test itself has to corrupt the object's contents in the remote repository before attempting to download the object.

Because our tests of the SSH-based Git LFS transfer protocol do not use a custom test utility like the `lfstest-gitserver` program to provide the remote side of the protocol, our test has to corrupt the objects itself.  (Note that for our tests of the SSH-based transfer protocol we use the `git-lfs-transfer` [program](https://github.com/bk2204/scutiger/tree/614a4f01cfbe6aafa355ca92ed6d1581b5cfc088/scutiger-lfs#git-lfs-transfer) from the `scutiger-lfs` Rust [crate](https://crates.io/crates/scutiger-lfs), and we configure this program to read and write from a local repository.)

#### Custom Transfer Agent Downloads

Finally, for our `custom` download adapter, which makes object transfer requests via an external custom transfer agent, we add a `batch storage custom adapter download of corrupt object` test, which is very similar to our test of the `basic` adapter, except that we set the `lfs.customTransfer.testcustom.path` configuration option so that our `lfstest-customadapter` transfer agent is used to upload and download objects.  This test utility simply acts as a proxy, responding to custom transfer protocol JSON messages from the Git LFS client, and then making HTTP-based object transfers to our `lfstest-gitserver` program and relaying the data back over the custom transfer protocol.

As a consequence, this test can also create an object with the special content key `storage-download-corrupt`, and expect that the `lfstest-gitserver` utility will recognize this key during a download request and replace the object's content with an uppercase version, so the test does not need to do so itself the way our test of the SSH-based object transfer protocol does.

### DIFF
--- a/t/t-batch-storage-corrupt.sh
+++ b/t/t-batch-storage-corrupt.sh
@@ -1,0 +1,144 @@
+#!/usr/bin/env bash
+
+. "$(dirname "$0")/testlib.sh"
+
+begin_test "batch storage HTTP download of corrupt object"
+(
+  set -e
+
+  reponame="batch-storage-download-corrupt-http"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" "$reponame"
+
+  git lfs track "*.dat"
+
+  # This content announces to the server that it should corrupt the
+  # contents of the object before returning it.
+  contents="storage-download-corrupt"
+  contents_oid="$(calc_oid "$contents")"
+  printf "%s" "$contents" > a.dat
+
+  git add .gitattributes a.dat
+  git commit -m "initial commit"
+
+  git push origin main
+
+  assert_server_object "$reponame" "$contents_oid"
+
+  rm -f a.dat
+  rm -rf .git/lfs/objects
+
+  # We expect the server to corrupt the object's contents by inverting
+  # the case of each character.
+  inverted_contents_oid="$(calc_oid "$(invert_case "$contents")")"
+
+  GIT_TRACE=1 git lfs pull 2>&1 | tee pull.log
+  [ 0 -ne "${PIPESTATUS[0]}" ]
+
+  grep "expected OID $contents_oid, got $inverted_contents_oid after ${#contents} bytes written" pull.log
+  grep "Failed to fetch some objects" pull.log
+
+  refute_local_object "$contents_oid"
+)
+end_test
+
+begin_test "batch storage SSH download of corrupt object"
+(
+  set -e
+
+  setup_pure_ssh
+
+  reponame="batch-storage-download-corrupt-ssh"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" "$reponame"
+
+  sshurl=$(ssh_remote "$reponame")
+  git config lfs.url "$sshurl"
+
+  git lfs track "*.dat"
+
+  contents="storage-download-corrupt"
+  contents_oid="$(calc_oid "$contents")"
+  printf "%s" "$contents" > a.dat
+
+  git add .gitattributes a.dat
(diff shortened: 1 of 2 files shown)


t/t-batch-storage-corrupt.sh
```diff
@@ -0,0 +1,144 @@
...
+  # the case of each character.
+  inverted_contents_oid="$(calc_oid "$(invert_case "$contents")")"
+
+  GIT_TRACE=1 git lfs pull 2>&1 | tee pull.log
+  [ 0 -ne "${PIPESTATUS[0]}" ]
+
+  grep "expected OID $contents_oid, got $inverted_contents_oid after ${#contents} bytes written" pull.log
+  grep "Failed to fetch some objects" pull.log
+
+  refute_local_object "$contents_oid"
```
larsxschneider, line 45: nit: should we also refute that `$refute_local_object` exists for good measure?

t/t-batch-storage-corrupt.sh
```diff
@@ -0,0 +1,144 @@
...
+  assert_server_object "$reponame" "$contents_oid"
+
+  rm -f a.dat
+  rm -rf .git/lfs/objects
+
+  # We expect the server to corrupt the object's contents by inverting
+  # the case of each character.
+  inverted_contents_oid="$(calc_oid "$(invert_case "$contents")")"
+
+  GIT_TRACE=1 git lfs pull 2>&1 | tee pull.log
```
larsxschneider, line 39: should we do the same test for the more commonly used `git pull`, too?

we expect the client to be in an undefined state after this, right? meaning at least the problem object is not present and others may or may not be present. should we check that `git status`/`git lfs ls-files` communicates that there is at least one broken object?

larsxschneider (approved):
That is a great test to have 👍

t/t-batch-storage-corrupt.sh
chrisd8088: I'm guessing you meant that we should refute that `$inverted_contents_oid` exists?  That would make sense to me as an addition, so I'll add that to the tests.

t/t-batch-storage-corrupt.sh
chrisd8088: Sure, that makes sense.  The same download adapters are used in each case, so it's not really necessary, but we might as well demonstrate that fact in our tests.

PR #6277, opened 2026-06-08
deps: bump golang.org/x/net v0.47.0  to  v0.54.0 (CVE-2026-39821)

cristian-gherghina:
## Summary

`go.mod` pins `golang.org/x/net v0.47.0`, which is affected by **CVE-2026-39821**: `idna.ToASCII`/`ToUnicode` incorrectly accept Punycode-encoded labels that decode to an ASCII-only label (e.g. `ToUnicode("xn--example-.com")` returns `example.com` instead of erroring). This can defeat hostname-based authorization checks. Fixed upstream in `golang.org/x/net v0.54.0`.

This is a public, already-disclosed dependency CVE, so filing as a normal PR rather than a private report.

## Change

No source changes.

```
go get golang.org/x/net@v0.54.0 && go mod tidy
```
- `golang.org/x/net` v0.47.0  to **v0.54.0**
-  other deps impacted: 
  `x/sync` 0.18 to 0.20
  `x/sys` 0.38 to 0.44
  `x/crypto` 0.45 to 0.51 (indirect)
  `x/text` 0.31 to 0.37 (indirect)
- **`go` directive `1.24.0` t0  `1.25.0`**  `x/net v0.54.0` requires Go ≥ 1.25, so `go mod tidy` raises the `go` directive from `1.24.0` to `1.25.0`.  This raises git-lfs's minimum supported Go version. Flagging explicitly in case that affects your CI matrix / support policy,  happy to adjust if you'd prefer a different approach.

## Verification

- `go build ./...`   passes
- `go test ./...`  passes
- `go vet ./...` only the two findings already present on `main` (`lfshttp/client.go:515` lock copy, `git-lfs.go:15` unbuffered signal channel); no new findings introduced.

Refs: CVE-2026-39821, go.dev/issue/78760


go.mod
```diff
@@ -21,9 +21,9 @@ require (
 	github.com/ssgelm/cookiejarparser v1.0.1
 	github.com/stretchr/testify v1.6.1
 	github.com/xeipuuv/gojsonschema v0.0.0-20170210233622-6b67b3fab74d
-	golang.org/x/net v0.47.0
-	golang.org/x/sync v0.18.0
-	golang.org/x/sys v0.38.0
+	golang.org/x/net v0.54.0
```
chrisd8088, line 24: If I'm not mistaken, the Punycode-related issue [reported](https://pkg.go.dev/vuln/GO-2026-5026) as GO-2026-5026 and [CVE-2026-39821](https://www.cve.org/CVERecord?id=CVE-2026-39821) is actually only resolved in v0.55.0 of the `x/net` module.  That's based on the notes in the module's version [listing](https://pkg.go.dev/golang.org/x/net?tab=versions) and the GO-2026-5026 [report](https://pkg.go.dev/vuln/GO-2026-5026).

So I think we want to update to v0.55.0, as I've tried to do in PR #6278, where I've credited you as co-author on the change.

chrisd8088 (commented):
Hey, thanks for this PR!

As it happens, we're trying to release a new version of Git LFS soon, so this is a timely reminder to upgrade our Go module dependencies.

Because I think we actually want an even newer version of the `x/net` module than the one from this PR, and because we also want to update our Dependabot configuration, I opened PR #6278 and credited you for its Go module changes.

If you don't mind, once that's approved and merged I'll close this PR.  Thanks again for help!

go.mod
cristian-gherghina: thank you @chrisd8088 that is even better :).

PR #6284, opened 2026-06-17
Autominsize support

gnkow:
This PR adds a --min-size flag to git lfs track and fixes a build failure when no git tags are present.

The new --min-size flag automatically tracks files exceeding a given size threshold:
- git lfs track --min-size 10MB sets lfs.autotracksize in .lfsconfig and adds * filter=lfs to .gitattributes
- Existing files above the threshold are touched so they appear as modified on the next git add
- Uses human-readable size units (KB, MB, GB, KiB, MiB, GiB)
- Supports --dry-run to preview without making changes
- The clean filter passes through files below the threshold unchanged


chrisd8088 (commented):
Hey, thanks very much for this draft PR, and for trying to tackle a long-standing issue!

I typically avoid reviewing on PRs in a draft state, as I assume they are still a work in progress, but from your [comment](https://github.com/git-lfs/git-lfs/issues/282#issuecomment-4733130661) I think you are hoping for some discussion, so I thought I would at least make some initial comments.

As I understand it, the key design principle of this PR is to pass all files through Git LFS; that's the effect of a `.gitattributes` pattern which matches `*`.

This idea has been considered before; I think our primary concern is going to be the performance cost of running every file through the Git LFS filters.  We'd definitely want to see some comparative timing measurements done on each of the platforms we support (Linux, macOS, and Windows) using a reasonably large Git repository as test data.

Git LFS was originally developed several large Git hosting services (Atlassian, GitHub, GitLab) as a kind of workaround which allowed them to avoid storing large binary files in their own platforms.   By using `.gitattributes` patterns to invoke the Git LFS filters, though, they could limit the impact of this additional program and system to just a specific set of files.

As a high-level concern, if we were to now start recommending that users process every file through Git LFS, that seems like we are trending away from direction I think we likely would prefer, where Git itself can do more of the work necessary to store and transfer large files efficiently.

A few other notes I made while thinking about this approach:

- If the minimum size is relatively small, files such as `.gitattributes` and `.gitignore` may be converted to Git LFS objects, which is not appropriate as Git expects these to have a specific format that it can read.  All such Git special files would need to be excluded automatically.
- Since files managed with Git LFS can be quite large, using `os.ReadAll()` may consume a considerable amount of memory.  It would likely be necessary to spool the incoming data to a temporary file; our `tools.Spool()` function does this already and is used in our `smudge` filter code to avoid deadlocking the pipe with Git (see PR [#1932](https://github.com/git-lfs/git-lfs/pull/1932)).
- If a file below the threshold is removed from the working tree and then `git checkout` is run to recreate it, Git will invoke Git LFS as a `smudge` filter, which assumes the file should contain a pointer and will report it as malformed when it does not.  So the `smudge` filter would also need to be aware of the file size threshold and ignore non-pointers below that limit.
- I happen to have spent the last week working to deal with the ramifications of the bug fix proposed in PR [#6011](https://github.com/git-lfs/git-lfs/pull/6011), and so I know that once it's merged, our `clean()` function will rarely know the size of a file in advance.  Specifically, the `fileSize` parameter is [passed](https://github.com/git-lfs/git-lfs/blob/023ef527434d5bd636b3f8e22222069c85fed15e/commands/command_filter_process.go#L88) as `-1` during `clean` filter operations because Git doesn't provide the size of the data it's sending, and although we currently try to work around this by [checking](https://github.com/git-lfs/git-lfs/blob/023ef527434d5bd636b3f8e22222069c85fed15e/commands/command_clean.go#L32-L45) the size of the corresponding file in the working tree, this introduces a race condition because the file might change after Git starts streaming it to our filter.  Ultimately, we have to remove this check and so only by reading the data arriving from Git would be know if its size exceeds a given threshold.
- Specifying the threshold value in our `.lfsconfig` file is, I think, going to be a bit fragile, because it means the overall mechanism depends on both a global-match pattern in `.gitattributes` and a Git configuration option.  Instead, if we were going to try to implement a threshold like this at all, I would want to see if we could specify it directly in the `.gitattributes` entry as a custom key/value attribute along with `filter=lfs`.  Of course, this does mean Git LFS would have to read and parse the `.gitattributes` file to learn what the current threshold is, if any.

Those are just some quick initial thoughts.  Whether or not we pursue the approach taken in this PR, it's definitely worth considering the implications of such a design, so thank you very much for opening the PR and starting the discussion!

gnkow:
I read your review very carefully and proceed to address each topic one by one which resulted in several commits. I pushed to my fork branch which I assumed that would trigger this PR to be updated. Here is the summary of what was done:

1. Configuration in .gitattributes (265bb629e)
- Changed approach: Moved from standalone flag to .gitattributes-based configuration
- Implementation: Added autotracksize=<bytes> attribute support in .gitattributes
- Files modified: 7 files (188 insertions, 44 deletions)
- Key changes:
- Extended gitattr parser to read autotracksize values
- Updated git lfs track to write attributes to .gitattributes
- Added comprehensive test coverage in t/t-track.sh
2. Fixed Defaults for Exclusion (7cc59b77c)
- Added safety mechanisms: Implemented lfs.autotrackexclude config for exclusion patterns
- Default exclusions: Protected Git-critical paths (.git/, .gitignore, etc.)
- Files modified: 6 files (453 insertions, 20 deletions)
- Key changes:
- Added exclusion logic in clean/smudge filters
- Comprehensive unit tests in commands_test.go and attr_test.go
- Documentation updates
3. Memory and Clean/Smudge Optimizations (3f7fd50e1)
- Pass-through optimization: Improved clean filter to efficiently handle small files
- Memory safety: Added checks to avoid unnecessary buffering
- Files modified: 2 files (180 insertions, 10 deletions)
- Key changes:
- Refactored clean filter pass-through logic
- Added 137 lines of unit tests
4. Changed Minsize Setup Logic (50c28b183)
- User interface improvements: Made --min-size require explicit pattern argument
- No more * catch-all by default: Users must specify git lfs track "*" --min-size <size> explicitly
- Files modified: 3 files (324 insertions, 55 deletions)
- Key changes:
- Updated command_track.go to require pattern with --min-size
- Allows re-tracking existing patterns to add/update autotracksize
- Comprehensive documentation in man pages
- 310 lines of integration tests
5. Benchmark Toolset (62a43c66e)
- Performance measurement infrastructure: Comprehensive benchmark suite for performance analysis
- Native Go tooling: Replaced Python dependency with Go-based plotting tool
- Files added: 6 files (1,604 insertions)
- Components:
- commands/bench_sweep_test.go: 17 benchmark functions covering all critical paths (722 lines)
- tools/benchplot/: Native Go plotting tool (730 lines) with full documentation
- Makefile target: make benchplot for easy building
- Generates 22 plots: Performance visualization for all aspects of the feature


Performance optimisation is deferred to future work such as pattern caching, memory allocation, I/O.

PR #6285, opened 2026-06-23
Update tests to dynamically match internal program limits

chrisd8088:
This PR revises a number of our shell tests so they are able to dynamically adjust the conditions they create to match or exceed certain internal limits in the Git LFS client.

Previously, these tests depended on hard-coded values which could drift out of sync with the actual limits of the client.

In the case of the `clean pseudo pointer with extra data` [test](https://github.com/git-lfs/git-lfs/blob/023ef527434d5bd636b3f8e22222069c85fed15e/t/t-clean.sh#L45-L58) in particular, this has already occurred.  When the test was first developed, in PRs #271, #336, and #447, the size of the internal buffer used to read and parse Git LFS pointers was [only](https://github.com/git-lfs/git-lfs/blob/1f2453349d70360be5dfc57b20b81bd4ed82550c/lfs/pointer.go#L65) 512 bytes, and so the test intentionally creates an invalid Git LFS pointer with sufficient padding to exceed that limit.

However, the size of the internal buffer was later raised to its current [value](https://github.com/git-lfs/git-lfs/blob/023ef527434d5bd636b3f8e22222069c85fed15e/lfs/scanner.go#L12) of 1024 bytes, but the `clean pseudo pointer with extra data`  test was never updated to reflect this change.  (Note that the test happens to [specify](https://github.com/git-lfs/git-lfs/blob/023ef527434d5bd636b3f8e22222069c85fed15e/t/t-clean.sh#L53) the value of `1024` as the `size` field of the invalid pointer it creates, but this is unrelated to the actual size of the string containing the invalid pointer.  To help clarify the arbitrary nature of the value in this `size` field, we now change it to `9999`.)

To allow our shell tests to retrieve the various current internal limits of the Git LFS client, we add a test helper utility program named `lfstest-getlimit` which returns one of several values depending on the option it is passed.  These values include:

- The [size](https://github.com/git-lfs/git-lfs/blob/023ef527434d5bd636b3f8e22222069c85fed15e/lfs/scanner.go#L12) of the internal buffer used when decoding Git LFS pointers.
- The maximum [length](https://github.com/git-lfs/pktline/blob/ca444d533ef1e474d0aab99cdbeed9b048d65241/pkt_line.go#L16) of a Git packet line.
- The [size](https://github.com/git-lfs/git-lfs/blob/023ef527434d5bd636b3f8e22222069c85fed15e/tools/iotools.go#L19) of the internal buffer used before [spooling](https://github.com/git-lfs/git-lfs/blob/023ef527434d5bd636b3f8e22222069c85fed15e/commands/command_smudge.go#L104) non-pointer `smudge` filter data to a temporary file.
- The maximum [size](https://pkg.go.dev/bufio#pkg-constants) of the buffer used when parsing tokens with a `Scanner` structure from the Go standard library's `bufio` package.

We also adjust the behaviour of our `calc_oid()` test helper function so that it no longer [treats](https://github.com/git-lfs/git-lfs/blob/023ef527434d5bd636b3f8e22222069c85fed15e/t/testhelpers.sh#L844) its parameter as a format specification for the `printf` shell built-in command.  This guarantees that even if we accidentally pass special formatting characters like `\` or `%` in the input to the `calc_oid()` function, it will return the same SHA-256 hash value as the Git LFS client would calculate for the same data.

---

Note that the new test added in PR #6011 requires input data for the `git lfs filter-process` command which exceeds the pointer read buffer size.  We therefore expect to revise that PR's test to make use of the `lfstest-getlimit` helper utility introduced in this PR, thereby avoiding the use of a hard-coded value.

### DIFF
--- a/lfs/gitfilter_clean.go
+++ b/lfs/gitfilter_clean.go
@@ -76,11 +76,11 @@ func (f *GitFilter) copyToTemp(reader io.Reader, fileSize int64, cb tools.CopyCa
 
 	ptr, buf, err := DecodeFrom(reader)
 
-	by := make([]byte, blobSizeCutoff)
+	by := make([]byte, BlobSizeCutoff)
 	n, rerr := buf.Read(by)
 	by = by[:n]
 
-	if rerr != nil || (err == nil && len(by) < blobSizeCutoff) {
+	if rerr != nil || (err == nil && len(by) < BlobSizeCutoff) {
 		err = errors.NewCleanPointerError(ptr, by)
 		return
 	}
--- a/lfs/gitscanner_catfilebatch.go
+++ b/lfs/gitscanner_catfilebatch.go
@@ -138,7 +138,7 @@ func (s *PointerScanner) next(blob string) (string, string, *WrappedPointer, err
 
 	var buf *bytes.Buffer
 	var to io.Writer = sha
-	if size < blobSizeCutoff {
+	if size < BlobSizeCutoff {
 		buf = bytes.NewBuffer(make([]byte, 0, size))
 		to = io.MultiWriter(to, buf)
 	}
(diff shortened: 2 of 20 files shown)


larsxschneider (approved):
Thank you 😊

PR #6286, opened 2026-06-24
fall back to guessed endpoint when git-lfs-authenticate is unavailable

zbleness:
When an SSH remote's server does not provide the git-lfs-authenticate command, a usable LFS endpoint has already been guessed from the remote URL. Previously any non-zero exit from git-lfs-authenticate was treated as fatal, so such servers could not be used over SSH without explicitly configuring lfs.url.

Detect the unavailable command in sshAuthClient.Resolve and signal it with a distinct sshAuthenticateUnavailableError. sshResolveWithRetries then falls back to the guessed endpoint (authenticating via the normal Git credential helper) instead of failing, and stops retrying since a missing command will not become available.

Detection covers both the conventional command-not-found exit code (127) and servers that return a generic exit code with a 'not found'-style message. Gerrit, for example, runs the command through a restricted shell and responds with exit code 1 and the message:

  fatal: Gerrit Code Review: git-lfs-authenticate: not found

Message matching is anchored to the command name so that unrelated errors such as a missing repository are not misclassified. Any other non-zero exit code remains fatal.


lfshttp/ssh_test.go
```diff
@@ -207,6 +211,157 @@ func TestSSHCacheResolveWithError(t *testing.T) {
...
+
+func (r *countingResolver) Resolve(e Endpoint, method string) (sshAuthResponse, error) {
+	r.calls++
+	return sshAuthResponse{}, r.err
+}
+
+func newTestClient(t *testing.T, resolver SSHResolver) *Client {
+	t.Helper()
+	c, err := NewClient(NewContext(nil, nil, nil))
+	assert.Nil(t, err)
```
chrisd8088: The `NewClient()` [function](https://github.com/git-lfs/git-lfs/blob/b8850997c5a67a711627bbe866f6c32d40e64767/lfshttp/client.go#L85-L118) doesn't return an `error` value, so this will fail to compile and no Go tests in this package will actually run.  It needs to look like this instead:

```suggestion
	c := NewClient(NewContext(nil, nil, nil))
```

lfshttp/ssh_test.go
```diff
@@ -207,6 +211,157 @@ func TestSSHCacheResolveWithError(t *testing.T) {
...
+		},
+	}
+	c := newTestClient(t, resolver)
+
+	req, err := c.NewRequest("GET", testSSHEndpoint(), "objects/batch", nil)
+	assert.Nil(t, err)
+	assert.Equal(t,
+		"https://git-server.com/foo/bar.git/info/lfs/objects/batch",
+		req.URL.String())
+}
```
chrisd8088: Although this test function is obviously very closely related to the other `TestSSHResolve*()` functions added in this PR, it's actually a test of the `NewRequest()` method rather than the `sshResolveWithRetries()` method.

Hence I would be inclined to call it `TestNewRequestUsesGuessedEndpointWhenSSHAuthenticateUnavailable()` or something along those lines.

I'd also be inclined to move it over with the other `TestNewRequest*()` test [functions](https://github.com/git-lfs/git-lfs/blob/b8850997c5a67a711627bbe866f6c32d40e64767/lfshttp/client_test.go#L236-L266).

chrisd8088 (commented):
Hey, thanks very much for this PR and for the analysis of the issue it addresses!

Both the general approach and the specific implementation look good to me.

There's a test failure which will need to be resolved, and I'd recommend renaming and moving one Go test function.

The other recommendation I'd make is that we'll also want an integration test in our shell test suite.

We can easily modify our `t/cmd/lfs-ssh-echo.go` utility to return a 127 exit code when a certain repository name is provided in its arguments.  (We use this utility in our integration tests to simulate an SSH client.)

The challenge we'll run into is that our `t/cmd/lfstest-gitserver.go` test utility, which simulates remote Git and Git LFS services, starts its HTTP services on non-standard ports, but the fallback URL [constructed](https://github.com/git-lfs/git-lfs/blob/b8850997c5a67a711627bbe866f6c32d40e64767/lfshttp/endpoint.go#L62-L64) by the Git LFS client explicitly drops any port number that specified as part of an SSH URL.

For our existing tests, when our `lfs-ssh-echo` utility returns a JSON response to a `git-lfs-authenticate` command invocation, it [provides](https://github.com/git-lfs/git-lfs/blob/b8850997c5a67a711627bbe866f6c32d40e64767/t/cmd/lfs-ssh-echo.go#L141) an `href` element whose HTTP URL contains the same port that was used for the SSH command.  The Git LFS client then [prefers](https://github.com/git-lfs/git-lfs/blob/b8850997c5a67a711627bbe866f6c32d40e64767/lfshttp/client.go#L146-L148) that URL to the fallback one it has constructed.

However, if our `lfs-ssh-echo` utility returns a 127 exit code and no JSON response, the client will try an HTTP URL without a port, and since our `lfstest-gitserver` utility doesn't listen on port 443, the test will likely fail (but it depends on whether a "real" Git LFS service happens to be listening on port 443 and accepting requests).

Here's what I'd suggest.  First, apply this patch to our `t/cmd/lfs-ssh-echo.go` utility:

```patch
diff --git a/t/cmd/lfs-ssh-echo.go b/t/cmd/lfs-ssh-echo.go
index ecb862e5..4bb9af87 100644
--- a/t/cmd/lfs-ssh-echo.go
+++ b/t/cmd/lfs-ssh-echo.go
@@ -53,8 +53,8 @@ func checkSufficientArgs(offset int) {
 
 func main() {
 	// expect args:
-	//   lfs-ssh-echo [-p PORT [--]] git@127.0.0.1 "git-lfs-authenticate REPO OPERATION"
-	//   lfs-ssh-echo [-p PORT [--]] git@127.0.0.1 "git-lfs-transfer REPO OPERATION"
+	//   lfs-ssh-echo [-p PORT [--]] git@127.0.0.1[.invalid] "git-lfs-authenticate REPO OPERATION"
+	//   lfs-ssh-echo [-p PORT [--]] git@127.0.0.1[.invalid] "git-lfs-transfer REPO OPERATION"
 	//   lfs-ssh-echo git@127.0.0.1 "git-upload-pack REPO"
 	//   lfs-ssh-echo git@127.0.0.1 "git-receive-pack REPO"
 	offset := 1
@@ -114,8 +114,9 @@ func main() {
 	}
 
 	checkSufficientArgs(offset)
-	if os.Args[offset] != "git@127.0.0.1" {
-		fmt.Fprintf(os.Stderr, "expected \"git@127.0.0.1\", got %q", os.Args[offset])
+
+	if os.Args[offset] != "git@127.0.0.1" && os.Args[offset] != "git@127.0.0.1.invalid" {
+		fmt.Fprintf(os.Stderr, "expected \"git@127.0.0.1\" or \"git@127.0.0.1.invalid\", got %q", os.Args[offset])
 		os.Exit(1)
 	}
 
@@ -137,6 +138,10 @@ func main() {
 
 	repo := remoteCmd[1]
 
+	if repo == "/ssh-unavailable" {
+		os.Exit(127)
+	}
+
 	r := &sshResponse{
 		Href: fmt.Sprintf("http://127.0.0.1:%s/%s.git/info/lfs", os.Args[2], repo),
 	}
```

Next, add this test to our `t/t-ssh.sh` script:

```shell
begin_test "SSH git-lfs-authenticate unavailable (fallback to HTTPS)"
(
  set -e

  # This repository name announces to the SSH test utility that it should
  # exit as if the "git-lfs-authenticate" command was not found.
  reponame="ssh-unavailable"
  setup_remote_repo "$reponame"

  # We construct an invalid host name (127.0.0.1.invalid) to guarantee that
  # the HTTPS fallback URL, which will default to port 443, cannot contact
  # any local or remote services not controlled by our test suite.
  gitserver_hostport="${GITSERVER#http://}"
  gitserver_host="${gitserver_hostport/%:*/}"
  gitserver_port="${gitserver_hostport/#*:/}"
  invalid_host="$gitserver_host.invalid"
  git config lfs.url "ssh://git@$invalid_host:$gitserver_port/$reponame"

  git lfs track "*.dat"

  contents="test"
  printf "%s" "$contents" >test.dat

  git add .gitattributes test.dat
  git commit -m "initial commit"

  GIT_TRACE=1 git push origin main 2>&1 | tee push.log
  [ 0 -ne "${PIPESTATUS[0]}" ]

  # Requests to both the Locking API and the Batch API should fall back to
  # an HTTPS URL, which should in turn fail because the host name is invalid.
  [ 2 -eq "$(grep -c "exec: lfs-ssh-echo.*git-lfs-authenticate /$reponame upload" push.log)" ]
  [ 2 -eq "$(grep -c -F "ssh: git@$invalid_host does not provide git-lfs-authenticate, falling back to guessed LFS endpoint" push.log)" ]
  [ 2 -eq "$(grep -c -F "HTTP: POST https://$invalid_host/$reponame/" push.log)" ]
)
end_test
)
end_test
```

Other tests could be added as well, for example to verify that an HTTPS fallback URL is attempted when the `lfs-ssh-echo` utility returns an exit code of `1` and a message like `git-lfs-authenticate: command not found` in reply to an SSH URL request for the `git-lfs-authenticate` command.  You would just need to use a different repository name such as `ssh-unavailable-not-found` and further tweak the `lfs-ssh-echo` utility to recognize it and exit with the appropriate code and message.

Thanks again for working on this PR and for your help improving the Git LFS client!

lfshttp/ssh_test.go
zbleness: Done

zbleness:
Thanks for the detailed and substantial comments @chrisd8088. I think I have now implemented them all as requested. 

I also added a `begin_test "SSH Gerrit use-case (fallback to HTTPS)"` to protect that particular use-case as we are using it at Volvo.

t/cmd/lfs-ssh-echo.go
```diff
@@ -137,6 +137,15 @@ func main() {
...
 	repo := remoteCmd[1]
 
+	if repo == "/ssh-unavailable" {
+		os.Exit(127)
+	}
+
+	if repo == "/ssh-gerrit-without-lfs-plugin" {
+		fmt.Fprintf(os.Stderr, "fatal: Gerrit Code Review: git-lfs-authenticate: not found")
+		os.Exit(1)
+	}
```
chrisd8088: Since we have two conditions here now, I'd probably combine them with a `switch` statement.  I think I'd also just use a somewhat more generic name for the test repository, for example, `ssh-unavailable-message`:

```suggestion
	switch repo {
	case "/ssh-unavailable":
		os.Exit(127)
	case "/ssh-unavailable-message":
		// Simulate the error message returned by a third-party utility
		// which does not support the "git-lfs-authenticate" command.
		fmt.Fprintf(os.Stderr, "fatal: Gerrit Code Review: git-lfs-authenticate: not found")
		os.Exit(1)
	}
```

t/t-ssh.sh
```diff
@@ -61,3 +61,79 @@ begin_test "ssh with proxy command in lfs.url (custom variant)"
...
 )
 end_test
+
+begin_test "SSH Gerrit use-case (fallback to HTTPS)"
+(
+  set -e
+
+  # This repository name announces to the SSH test utility that it should
+  # exit with code 1 and the
+  reponame="ssh-gerrit-without-lfs-plugin"
```
chrisd8088: Thanks for adding another test!  It's definitely a good idea to check this use case as well.

Since this test is effectively identical to the other new shell test, we could simplify them using the approach we take in our `t/t-push-failures-remote.sh` script where we [define](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/t/t-push-failures-remote.sh#L5-L42) a common test function and then just [invoke](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/t/t-push-failures-remote.sh#L44-L122) it multiple times with different repository names.

How would you feel about defining this test and the one above it as follows?

```shell
# ssh_authenticate_unavailable_test performs a test which expects "git push"
# to fail, but only after the Git LFS client correctly defaults to a
# fallback HTTPS URL when the "git-lfs-authenticate" command is reported as
# not found on the SSH remote.
#
# The SSH client used by the tests is triggered by the repository name to
# simulate a specific exit code and optional error message indicating that
# the "git-lfs-authenticate" command is not available.
ssh_authenticate_unavailable_test() {
  local reponame="$1"

  set -e

  setup_remote_repo "$reponame"

  # We construct an invalid host name (127.0.0.1.invalid) to guarantee that
  # the HTTPS fallback URL, which will default to port 443, cannot contact
  # any local or remote services not controlled by our test suite.
  gitserver_hostport="${GITSERVER#http://}"
  gitserver_host="${gitserver_hostport/%:*/}"
  gitserver_port="${gitserver_hostport/#*:/}"
  invalid_host="$gitserver_host.invalid"
  git config lfs.url "ssh://git@$invalid_host:$gitserver_port/$reponame"

  git lfs track "*.dat"

  contents="test"
  printf "%s" "$contents" >test.dat

  git add .gitattributes test.dat
  git commit -m "initial commit"

  GIT_TRACE=1 git push origin main 2>&1 | tee push.log
  [ 0 -ne "${PIPESTATUS[0]}" ]

  # Requests to both the Locking API and the Batch API should fall back to
  # an HTTPS URL, which should in turn fail because the host name is invalid.
  [ 2 -eq "$(grep -c "exec: lfs-ssh-echo.*git-lfs-authenticate /$reponame upload" push.log)" ]
  [ 2 -eq "$(grep -c -F "ssh: git@$invalid_host does not provide git-lfs-authenticate, falling back to guessed LFS endpoint" push.log)" ]
  [ 2 -eq "$(grep -c -F "HTTP: POST https://$invalid_host/$reponame/" push.log)" ]
}

begin_test "SSH git-lfs-authenticate unavailable (fallback to HTTPS)"
(
  set -e

  # This repository name announces to the SSH test utility that it should
  # exit as if the "git-lfs-authenticate" command was not found.
  reponame="ssh-unavailable"
  ssh_authenticate_unavailable_test "$reponame"
)
end_test

begin_test "SSH git-lfs-authenticate unavailable message (fallback to HTTPS)"
(
  set -e

  # This repository name announces to the SSH test utility that it should
  # exit with a generic error status code and a message indicating that the
  # "git-lfs-authenticate" command was not found.
  reponame="ssh-unavailable-message"
  ssh_authenticate_unavailable_test "$reponame"
)
end_test
```

chrisd8088 (commented):
Hey, thanks again for this PR and for adopting my suggestions and adding another test!

This is looking great.  I had only a couple of minor suggestions as to how we might simplify the implementation of the two new shell integration tests, but otherwise I think this is ready to go.

I'll let the CI suite run to see how it does (sometimes we have unexpected regressions, so one of those may pop up and have to be fixed separately).

Thanks again for working on this enhancement to Git LFS!

t/t-ssh.sh
zbleness: Sure, make sense.

Added both your suggestions now and `make ./t-ssh.sh` passes nicely on my machine.

chrisd8088: Thank you very much!

chrisd8088 (approved):
Thanks again for all the work on this PR!

I tweaked the names and indentation of the two shell tests with one final commit, and expanded one test's comment to help clarify that it simulates the Gerrit Code Review condition you described.

Once our CI jobs pass, we can merge the PR.  Thanks again for all the help!

Issue #5475, opened 2023-08-17
Abort LFS pull after rate limit is hit


nick-wellinghoff:
**Describe the bug**
I have a LFS server where I limit the amount of requests per day a ip can hit the server to avoid abusive clients that are constantly clean cloning and pulling . However the lfs client will just hang and wait forever when my LFS server returns headers

X-Ratelimit-Remaining 0
Retry-After 86400 

I purpose a feature is added where the LFS server can add a header or if the LFS client detects the retry-after is very long the process aborts and shows the http response provided by the 429 error. 

**To Reproduce**
1. Setup a LFS server with a reverse proxy with a rate limiter 
2. When the rate limit is hit set a 429 return code and set the above headers.
3. Observe the LFS pull operation waits indefinitely and never times out. The lfs user has no idea they have hit their rate limit quota. 

**Expected behavior**
I purpose a feature is added where the LFS server can add a header or if the LFS client detects the retry-after is very long the process aborts and shows the http response provided by the 429 error. 

**System environment**
N/A

**Output of `git lfs env`**
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
FetchRecentRefsIncludeRemotes=true
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/home/ec2-user/test/gecko_sdk_lfs/.git/lfs
AccessDownload=none
AccessUpload=none
DownloadTransfers=basic,lfs-standalone-file,ssh
UploadTransfers=basic,lfs-standalone-file,ssh
GIT_EXEC_PATH=/usr/libexec/git-core
git config filter.lfs.process = "git-lfs filter-process"
git config filter.lfs.smudge = "git-lfs smudge -- %f"
git config filter.lfs.clean = "git-lfs clean -- %f"

**Additional context**
Any other relevant context about the problem here.

bk2204:
Hey,

I think we can look into providing the error string if one's provided by the server here.

PR #6304, opened 2026-07-15
Enhance LFS transfer retries with maximum retry time limit

TheRootDaemon:
Adds `lfs.transfer.maxretryafter` config option with a default value of `300` seconds.
Also treats `509` as `429` so that `Retry-After` headers on 509 responses are recognized.

Addresses #5475


lfsapi/response_test.go
```diff
@@ -180,5 +178,164 @@ func TestWithNonFatal500WithoutBody(t *testing.T) {
...
+	c := NewClient(nil)
+
+	var called uint32
+
+	tests := []struct {
+		name          string
+		statusCode    int
+		retryAfter    string
+		wantRetriable bool
+		wantLater     bool
```
chrisd8088: While `want` is a perfectly reasonable choice of term, a lot of our existing Go tests use the term `expected`, so I'd suggest using that instead.

lfsapi/response_test.go
```diff
@@ -180,5 +178,164 @@ func TestWithNonFatal500WithoutBody(t *testing.T) {
...
+		wantPrefix    string
```
chrisd8088: I see that values for `wantPrefix` are supplied in the individual test cases below, but I don't see anything in the test itself which sets or checks these strings.  Are they necessary at all?

tq/transfer_queue.go
```diff
@@ -1050,13 +1051,34 @@ func (q *TransferQueue) canRetryObject(oid string, err error) bool {
...
 func (q *TransferQueue) canRetryObjectLater(oid string, err error) (time.Time, bool) {
 	if count, ok := q.rc.CanRetry(oid); !ok {
 		tracerx.Printf("tq: refusing to retry %q, too many retries (%d)", oid, count)
 		return time.Time{}, false
 	}
 
-	return q.canRetryLater(err)
+	retryAt, canRetry := q.canRetryLater(err)
+	if !canRetry {
+		return time.Time{}, false
```
chrisd8088: I'd suggest that we just return the `retryAt` value from the `canRetryLater()` method here, rather than initializing a new `Time` structure.

Technically it doesn't matter at the moment, because our callers [both](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/tq/transfer_queue.go#L594-L595) [discard](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/tq/transfer_queue.go#L815-L829) the `retryAt` value when the `canRetry` value is `false`.  However, by returning the `retryAt` value we guard against any situations in the future where for some reason we want to examine that value even when `canRetry` is `false`.

```suggestion
		return retryAt, false
```

lfshttp/errors.go
```diff
@@ -62,7 +62,7 @@ func (c *Client) handleResponse(res *http.Response) error {
 		return errors.NewUnprocessableEntityError(err)
 	}
 
-	if res.StatusCode == 429 {
+	if res.StatusCode == 429 || res.StatusCode == 509 {
```
chrisd8088: In another comment l'll expand on the question of whether or not to treat 509 `Bandwidth Limit Exceeded` status codes in the same manner as a 429 `Too Many Requests` status code; in brief, I don't think we want to do this.

Here, I'll just note that if we did want to handle the two status codes in the same manner, then we'd also want to remove the `res.StatusCode != 509` condition from the [line](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/lfshttp/errors.go#L75) below.  But as I said, I don't think we should treat the two types of responses identically, so I don't think that change is actually necessary.

chrisd8088 (commented):
Hey, thanks very much for picking up an issue from our project backlog and proposing a potential implementation!

I think we should name the new configuration option `lfs.transfer.maxRetryTime` in order to align with Git's `http.maxRetryTime` [option](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httpmaxRetryTime), which looks to me to have the same effect.

(Arguably, we should also not have implemented our own `lfs.transfer.*` versions of Git's `http.maxRetries` and `http.retryAfter` options, but since we have our own settings, going forward we can at least try to keep the nomenclature we use as similar to Git's as possible.)

We'll also need to add a section to our `git-lfs-config(5)` manual [page]() to describe the new `lfs.transfer.maxRetryTime` option.  That section should follow the [sections](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/docs/man/git-lfs-config.adoc?plain=1#L220-L237) about the `lfs.transfer.maxRetries` and `lfs.transfer.maxRetryDelay` options.

The Go test functions in this PR generally look good, thank you for including those!  We should also expand the integration tests in our shell test suite so they exercise the new configuration option.

We already have a range of integration tests which exercise the Git LFS client when 429 status code responses are received while making requests to the Batch API endpoint and to upload or download individual objects.  The `t/t-batch-retries-ratelimit.sh` script checks requests made to the Batch API endpoint, while the `t/t-batch-storage-retries-ratelimit.sh` script checks requests made to upload and download individual objects.  These tests generally operate by either [creating](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/t/t-batch-retries-ratelimit.sh#L9) repositories whose name includes a special string like `batch-retry-later` or by [creating](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/t/t-batch-storage-retries-ratelimit.sh#L13) objects whose content includes a special string like `storage-upload-retry-later`.  Our `t/cmd/lfstest-gitserver.go` test server [then](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/t/cmd/lfstest-gitserver.go#L449-L458) [detects](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/t/cmd/lfstest-gitserver.go#L750-L758) these repository names or object contents and responds by generating a certain number of 429 status code responses.  The server's `checkRateLimit()` function in should [cause](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/t/cmd/lfstest-gitserver.go#L306-L313) a 429 status code to be returned for each request until 10 seconds have passed since the first request of the same type, as [set](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/t/cmd/lfstest-gitserver.go#L279) by the `secondsToRefillTokens` constant.

Ideally, we'd include several new integration tests, one for the Batch API request and one each for object upload and download requests, which would set the `lfs.transfer.maxRetryTime` option to something like `5` and then confirm that the Git LFS client abandons the request when it receives a `Retry-After` header with a larger value.

Finally, with respect to our handling of 509 `Bandwidth Limit Exceeded` HTTP status code responses, I'm inclined to say that we should defer any changes to another PR.  This [comment](https://github.com/git-lfs/git-lfs/issues/5475#issuecomment-1682622340) on issue #5475 suggests to me that some users are explicitly converting 429 status code responses into 509 status code responses in order to force the Git LFS client to abandon its requests, so I wouldn't want to alter how we handle those without deeper consideration.

For one thing, our current behaviour differs depending on whether a 509 response is returned for a Batch API request or for an individual object transfer request.  If a `POST` request to the Batch API endpoint receives a 509 status code response, the Git LFS client [treats](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/tq/api.go#L90-L93) it as an error which should not be retried, so the [call](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/tq/transfer_queue.go#L592) to the `TransferQueue` structure's `canRetryObject()` method results in a `false` return value and the objects listed in the request are not re-enqueued.

However, when making an individual `GET` or `PUT` object transfer request, our upload and download HTTP transfer adapters [both](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/tq/basic_download.go#L200) [treat](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/tq/basic_upload.go#L134) a 509 status code response as an error which should be retried, so the [call](https://github.com/git-lfs/git-lfs/blob/d72db1e533a1d6ee5543e02e9f8ccac97e0fcd34/tq/transfer_queue.go#L815) to the `TransferQueue` structure's `canRetryObject()` method results in a `true` return value and the request is repeated, up to the maximum number of times specified by the `lfs.transfer.maxRetries` option.

If we're going to address how we handle 509 status code responses, we should ideally make these behaviours consistent with each other, while also respecting the possibility that servers already return the 509 status code for Batch API requests with the expectation that the client will not retry them.  So for various reasons, I'd suggest we don't try to change anything related to 509 status codes in this PR, and instead just focus on implementing the `lfs.transfer.maxRetryTime` option for 429 status code responses with `Retry-After` headers.

We really appreciate contributions like this, and I hope you'll enjoy working on our project!  Thanks so much for taking the time to draft this PR!

TheRootDaemon:
HI @chrisd8088,
Thanks so much for the thorough review !!!

I've addressed the requested changes:

- Renamed the configuration option to `lfs.transfer.maxRetryTime` to align with Git's `http.maxRetryTime`.
- Added documentation for `lfs.transfer.maxRetryTime` in the `adoc` page.
- Added integration tests covering the Batch API as well as upload and download object transfers, verifying that the client abandons retries when `Retry-After` exceeds `lfs.transfer.maxRetryTime`.

Per your suggestion, I left the handling of HTTP 509 responses unchanged so this PR remains focused on the new `lfs.transfer.maxRetryTime` functionality for 429 responses.

Thanks.

lfsapi/response_test.go
```diff
@@ -120,7 +120,6 @@ func TestAuthErrWithoutBody(t *testing.T) {
 	_, err = c.Do(req)
 	assert.NotNil(t, err)
 	assert.True(t, errors.IsAuthError(err))
-	assert.True(t, strings.HasPrefix(err.Error(), "Authentication required: Authorization error:"), err.Error())
```
chrisd8088, line 123: I'd like to restore this assertion and the one below it in the `TestFatalWithoutBody()` test function.

I appreciate that the preceding assertion of the return value from the `errors.IsAuthError()` function implicitly serves to verify that the error message contains the string `Authentication required`, since the `errors.NewAuthError()` function [wraps](https://github.com/git-lfs/git-lfs/blob/e8128e707419e8f08bf2ebd8cba3021ed27a2c9e/errors/types.go#L300) the original error message with that string.

However, that's not an explicit check, as the `NewAuthError()` function could change that string in the future.  More significantly, the existing assertion also checks that the original error message contains the string `Authorization error`, which is [generated](https://github.com/git-lfs/git-lfs/blob/e8128e707419e8f08bf2ebd8cba3021ed27a2c9e/lfshttp/errors.go#L108-L109) by our `lfshttp` package's `defaultError()` function.

The same is true of the assertion in the TestFatalWithoutBody()` test function which checks for the `Fatal error` and `Server error` strings in the error message, so I'd like to restore that assertion also.

lfsapi/response_test.go
```diff
@@ -182,3 +180,125 @@ func TestWithNonFatal500WithoutBody(t *testing.T) {
...
+		statusCode         int
+		retryAfter         string
+		expectedRetriable  bool
+		expectedRetryLater bool
+		serverMessage      string
+	}{
+		{
+			name:              "429 without Retry-After",
+			statusCode:        429,
+			retryAfter:        "",
```
chrisd8088: Since the default value of a string in Go is the empty string, we can just drop this line and the `retryAfter` field will still be assigned the empty string.

lfsapi/response_test.go
```diff
@@ -182,3 +180,125 @@ func TestWithNonFatal500WithoutBody(t *testing.T) {
...
+		name               string
+		statusCode         int
+		retryAfter         string
+		expectedRetriable  bool
+		expectedRetryLater bool
+	}{
+		{
+			name:              "429 without Retry-After",
+			statusCode:        429,
+			retryAfter:        "",
```
chrisd8088: Again, we can just drop this line, as the `retryAfter` field will automatically be assigned the empty string.

tq/transfer_queue_test.go
```diff
@@ -162,3 +169,72 @@ func TestUseAdapterSwitchesWhenNameDiffers(t *testing.T) {
...
+		wantReadyTimeZero bool
+	}{
+		{
+			name: "retry count exceeded",
+			err:  errors.NewRetriableLaterError(errors.New("rate limit"), "10"),
+			setup: func(q *TransferQueue) {
+				q.rc.MaxRetries = 1
+				q.rc.Increment("oid")
+			},
+			wantCanRetry:      false,
```
chrisd8088: Here and in each of the test cases below we can drop any lines which assign the `false` value, as that's the default value of a Boolean variable in Go.

tq/transfer_queue_test.go
```diff
@@ -162,3 +169,72 @@ func TestUseAdapterSwitchesWhenNameDiffers(t *testing.T) {
...
+		),
+	)
+	defer cli.Close()
+
+	tests := []struct {
+		name              string
+		err               error
+		setup             func(*TransferQueue)
+		wantCanRetry      bool
+		wantReadyTimeZero bool
```
chrisd8088: To be consistent with the similar test case structures for our other Go test functions, let's also rename these fields to use the prefix `expected` rather than `want`, as you've already done for the structure in the `lfsapi/response_test.go` source file.

t/t-batch-retries-ratelimit.sh
```diff
@@ -192,3 +192,32 @@ begin_test "batch API HTTP upload causes retries (missing header)"
...
+
+begin_test "batch API HTTP upload abandons when Retry-After exceeds maxRetryTime"
+(
+  set -e
+
+  reponame="upload-batch-retry-later-maxretrytime"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" batch-repo-upload-maxretrytime
+
+  git config --local lfs.transfer.maxretrytime 5
```
chrisd8088: This test doesn't succeed as it is currently written because our `lfstest-gitserver` test server utility won't respond to Batch API requests with a 429 status code and a `Retry-After` unless it [detects](https://github.com/git-lfs/git-lfs/blob/e8128e707419e8f08bf2ebd8cba3021ed27a2c9e/t/cmd/lfstest-gitserver.go#L449-L468) that the repository name in the URL has the suffix `batch-retry-later` or the `suffix `batch-retry-later-no-header`.  So it just permits the client's Batch API request to proceed, which means the Git LFS object is uploaded and the `git push` command returns a zero exit status.

We can resolve this by adjusting the test server utility to detect this test's repository name.  I'm going to suggest that we use the suffix `batch-retry-later-exceeds-max` to help clarify the condition the test is expected to validate.  Thus we need to patch the server as follows:

```patch
--- a/t/cmd/lfstest-gitserver.go
+++ b/t/cmd/lfstest-gitserver.go
@@ -446,7 +446,7 @@ func lfsBatchHandler(w http.ResponseWriter, r *http.Request, id, repo string) {
                }
        }
 
-       if strings.HasSuffix(repo, "batch-retry-later") {
+       if strings.HasSuffix(repo, "batch-retry-later") || strings.HasSuffix(repo, "batch-retry-later-exceeds-max") {
                if timeLeft, isWaiting := checkRateLimit("batch", "", repo, ""); isWaiting {
                        w.Header().Set("Retry-After", strconv.Itoa(timeLeft))
                        w.WriteHeader(http.StatusTooManyRequests)
```

We can then revise this test to use the repository name `upload-batch-retry-later-exceeds-max`.

I'm also going to suggest a few other adjustments here, none of which are critical but which help bring the test into alignment with our current idioms for our shell tests (even though the existing tests in this file don't exactly follow those same patterns).

For instance, we can just use the `reponame` variable twice in the `clone_repo()` function call, use camel-case for our configuration option's name, drop the `--local` flag in the `git config` command because our test suite reconstitutes a fresh "global" Git configuration for every test, and add some comments to help explain how the repository name signals the test server to behave a certain way.

```suggestion
  # This repository name announces to the server that it should
  # respond to all Batch API requests with a 429 Too Many Requests
  # status code and a Retry-After header until at least 10 seconds
  # have passed after the first request to the Batch API.
  reponame="upload-batch-retry-later-exceeds-max"
  setup_remote_repo "$reponame"
  clone_repo "$reponame" "$reponame"

  git config lfs.transfer.maxRetryTime 5
```

t/t-batch-retries-ratelimit.sh
```diff
@@ -192,3 +192,32 @@ begin_test "batch API HTTP upload causes retries (missing header)"
...
+  git commit -m "initial commit"
+
+  GIT_TRACE=1 git push origin main 2>&1 | tee push.log
+  if [ "0" -eq "${PIPESTATUS[0]}" ]; then
+    echo >&2 "fatal: expected \`git push origin main\` to fail ..."
+    exit 1
+  fi
+
+  grep "tq: refusing to retry" push.log
+  grep "exceeds maximum" push.log
```
chrisd8088: I think it would be good to add a call to our `refute_server_object()` test helper function at the end of this test.  And as above, we can also adopt some of our current idioms simplifying our checks of the `git push` command's exit status and the resultant log file:

```suggestion
  [ 0 -ne "${PIPESTATUS[0]}" ]

  grep "tq: refusing to retry \"$oid\", retry after \d\d*s exceeds maximum 5s" push.log

  refute_server_object "$reponame" "$oid"
```

t/t-batch-storage-retries-ratelimit.sh
```diff
@@ -187,3 +187,75 @@ begin_test "batch storage HTTP download causes retries (missing header)"
...
+begin_test "batch storage HTTP upload abandons when Retry-After exceeds maxRetryTime"
+(
+  set -e
+
+  reponame="batch-storage-upload-retry-later-maxretrytime"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" batch-storage-repo-upload-maxretrytime
+
+  git config --local lfs.transfer.maxretrytime 5
+
```
chrisd8088: This test and the one below it already succeed as written, which is great!

I'm going to suggest refine them slightly just to make them as similar as possible to the new test in the `t/t-batch-retries-ratelimit.sh` shell script and to some of most-recent [shell](https://github.com/git-lfs/git-lfs/blob/e8128e707419e8f08bf2ebd8cba3021ed27a2c9e/t/t-batch-storage-corrupt.sh#L5-L78) [tests](https://github.com/git-lfs/git-lfs/blob/e8128e707419e8f08bf2ebd8cba3021ed27a2c9e/t/t-batch-storage-corrupt.sh#L5-L78).  (Admittedly, this is a moving target, because we keep refining our "house style" for these types of tests.)

```suggestion
  reponame="batch-storage-upload-retry-later-exceeds-max"
  setup_remote_repo "$reponame"
  clone_repo "$reponame" "$reponame"

  git config lfs.transfer.maxRetryTime 5

  # This content announces to the server that it should respond to all
  # object storage PUT requests with a 429 Too Many Requests
  # status code and a Retry-After header until at least 10 seconds
  # have passed after the first object storage PUT request.
```

t/t-batch-storage-retries-ratelimit.sh
```diff
@@ -187,3 +187,75 @@ begin_test "batch storage HTTP download causes retries (missing header)"
...
+  git commit -m "initial commit"
+
+  GIT_TRACE=1 git push origin main 2>&1 | tee push.log
+  if [ "0" -eq "${PIPESTATUS[0]}" ]; then
+    echo >&2 "fatal: expected \`git push origin main\` to fail ..."
+    exit 1
+  fi
+
+  grep "tq: refusing to retry" push.log
+  grep "exceeds maximum" push.log
```
chrisd8088: As above, I think we can compress these lines a bit, while also adding a call to our `refute_server_object()` test assertion:

```suggestion
  [ 0 -ne "${PIPESTATUS[0]}" ]

  grep "tq: refusing to retry \"$oid\", retry after \d\d*s exceeds maximum 5s" push.log

  refute_server_object "$reponame" "$oid"
```

t/t-batch-storage-retries-ratelimit.sh
```diff
@@ -187,3 +187,75 @@ begin_test "batch storage HTTP download causes retries (missing header)"
...
+end_test
+
+begin_test "batch storage HTTP download abandons when Retry-After exceeds maxRetryTime"
+(
+  set -e
+
+  reponame="batch-storage-download-retry-later-maxretrytime"
+  setup_remote_repo "$reponame"
+  clone_repo "$reponame" batch-storage-repo-download-maxretrytime
+
```
chrisd8088: As above, let's adjust this slightly by adding a comment and using the `-exceeds-max` suffix in the repository name:

```suggestion
  reponame="batch-storage-download-retry-later-exceeds-max"
  setup_remote_repo "$reponame"
  clone_repo "$reponame" "$reponame"

  # This content announces to the server that it should respond to all
  # object storage GET requests with a 429 Too Many Requests
  # status code and a Retry-After header until at least 10 seconds
  # have passed after the first object storage GET request.
```

t/t-batch-storage-retries-ratelimit.sh
```diff
@@ -187,3 +187,75 @@ begin_test "batch storage HTTP download causes retries (missing header)"
...
+  git config --local lfs.transfer.maxretrytime 5
+
+  GIT_TRACE=1 git lfs pull origin main 2>&1 | tee pull.log
+  if [ "0" -eq "${PIPESTATUS[0]}" ]; then
+    echo >&2 "fatal: expected \`git lfs pull origin main\` to fail ..."
+    exit 1
+  fi
+
+  grep "tq: refusing to retry" pull.log
+  grep "exceeds maximum" pull.log
```
chrisd8088: We should indent these lines two spaces, and again, I think we can simplify our checks while also adding a call to an assertion function, in this case our `refute_local_object()` function:

```suggestion
    git \
      -c "filter.lfs.process=" \
      -c "filter.lfs.smudge=cat" \
      -c "filter.lfs.required=false" \
      clone "$GITSERVER/$reponame" "$reponame-assert"

    cd "$reponame-assert"

    git config credential.helper lfstest
    git config lfs.transfer.maxRetryTime 5

    GIT_TRACE=1 git lfs pull origin main 2>&1 | tee pull.log
    [ 0 -ne "${PIPESTATUS[0]}" ]

    grep "tq: refusing to retry \"$oid\", retry after \d\d*s exceeds maximum 5s" pull.log

    refute_local_object "$oid"
```

docs/man/git-lfs-config.adoc
```diff
@@ -235,6 +235,17 @@ Must be an integer which is not negative. Use zero to disable delays
...
++
+Specifies the maximum time in seconds LFS will wait before aborting a
+transfer when a server requests a delay using the `Retry-After` header.
+If the server indicates a wait time longer than this value, LFS will
+log a warning and mark the transfer as failed rather than sleeping for
+the full requested duration.
++
+Must be an integer which is at least than one. If the value is not an
+integer, is less than one, or is not given, a value of three hundred
+will be used instead.
```
chrisd8088: Thank you for this description of the new option!  We really appreciate the help keeping our documentation up to date.

There's one small typo here ("is at least one" not "is at least than one"), and I think we can also simplify the formatting slightly by using the numeral 300 rather than the words "three hundred".  I admit that the preceding options' descriptions use words ("eight", "ten") for numbers, but in general we numerals, especially for larger values.  (For instance, see the [entry](https://github.com/git-lfs/git-lfs/blob/e8128e707419e8f08bf2ebd8cba3021ed27a2c9e/docs/man/git-lfs-config.adoc#L251-L254) for `lfs.transfer.batchSize`.)

I'm also going to suggest we use the camel-case version of the option's name.  We don't do that consistently in our documentation (at least, not yet), but Git does in its documentation and I'm gradually trying to adjust ours to match that style.

```suggestion
* `lfs.transfer.maxRetryTime`
+
Specifies the maximum time in seconds LFS will wait before aborting a
transfer when a server requests a delay using the `Retry-After` header.
If the server indicates a wait time longer than this value, LFS will
log a warning and mark the transfer as failed rather than sleeping for
the full requested duration.
+
Must be an integer which is at least one. If the value is not an integer,
is less than one, or is not given, a value of 300 will be used instead.
```

chrisd8088 (commented):
Thank you very much for addressing our review comments and adding some new shell tests and documentation!  This PR is looking quite good now.

I've made another round of suggestions and did some debugging of one shell test, and in the interest of trying to move this PR forward I just wrote up a small set of commits which implement my proposed changes.

Since they are all minor test and documentation changes, I thought I would try pushing my commits up to your branch and seeing if our CI jobs pass.  I hope you don't mind my taking that approach!

Thank you again for helping to improve the Git LFS client; we really approach the assistance.

chrisd8088 (approved):

TheRootDaemon:
Hi @chrisd8088,
No issues, I am very glad that I found this project :)

Issue #6303, opened 2026-07-10
Git LFS filter-process continuously fills .git/lfs/tmp with hundreds of gigabytes of temporary files


xavdev1:
**Describe the bug**
Git LFS continuously creates large temporary files inside `.git/lfs/tmp` without an intentional Git LFS download, fetch, checkout, add, commit, push, or pull operation.

The temporary directory initially grew to approximately 265 GB with around 850 files and completely filled my C: drive. After deleting the files and stopping `git.exe` and `git-lfs.exe`, the files began regenerating.

During a later inspection, the directory had regenerated 118 files totalling 53,428,017,152 bytes. New files appeared every few seconds.

A minimal standalone reproduction is not yet available, but the issue repeatedly occurs in the affected repository:

**To Reproduce**
1. The repository contains large untracked `.safetensors` files under `model/Qwen3-4B`.
2. A nested `.gitattributes` file assigns `filter=lfs` to `*.safetensors`.
3. Leave the repository open in the development environment without intentionally running a Git LFS operation.
4. Observe `git-lfs.exe` running and `.git/lfs/tmp` continuously growing.
5. Stop Git LFS and delete the temporary files.
6. Observe the files being recreated and the directory growing again.

**Expected behavior**
Git LFS temporary files should be removed after an operation completes or is cancelled. The temporary directory should not continuously grow until the drive runs out of space.

If another application repeatedly invokes `git-lfs filter-process`, cancelled or incomplete operations should not leave hundreds of gigabytes of temporary files behind.

**System environment**
Operating system: Windows 11 Home 25H2
Build: 26200.8655
Architecture: x64
Git: git version 2.51.0.windows.1
Git LFS: git-lfs/3.7.0 (GitHub; windows amd64; go 1.24.4; git 92dddf56)
Local repository: C:\XavAi
Environment: Local Windows workspace; no container, CI, or Cygwin
WSL: Enabled with default version 2, but no Linux distributions are installed or running. The repository and Git LFS processes are running natively on Windows.
Proxy/VPN: None in use.
Antivirus/firewall: Windows default configuration with no known custom rules affecting Git LFS.

**Output of `git lfs env`**
```
git-lfs/3.7.0 (GitHub; windows amd64; go 1.24.4; git 92dddf56)
git version 2.51.0.windows.1
Endpoint=https://github.com/<redacted>/<redacted>.git/info/lfs (auth=none)
LocalWorkingDir=C:\XavAi
[...]

Issue #6242, opened 2026-04-16
Proposals for improving Git LFS custom transfer agents and local storage


JeremieA:
_AI disclaimer: The code and documentation mentioned below were created with the assistance of AI agents. The design, implementation, and tests were reviewed and directed by human developers. All communications in issues and PRs will be handled manually._

**Describe the issue**

I will be submitting 3 proposals for improving Git LFS custom transfer agents and local storage. This issue is meant as a first discussion prior to submitting separate issues or PRs for each individual proposal, followed by their implementations. Please indicate which one we should start with, if you have a preference.

Our use case: we maintain CI infrastructure where build artefacts (~2 GB per version, 500+ versions) are stored in git repositories. Plain git handles this surprisingly well thanks to delta compression (1.6 GB total), but breaks down at scale: repacking times explode as the repository grows. Git LFS is the natural solution, but storing each file version as an independent blob without deduplication increases storage from 1.6 GB to 23 GB for this dataset.

To address this, we developed a custom transfer agent backed by [desync](https://github.com/folbricht/desync), which adds content-defined chunking, deduplication, and zstd compression on top of Git LFS. This brings storage back down to 4 GB (5.6x reduction vs standard LFS) while preserving the standard `git push` / `git pull` workflow.

While building and benchmarking this integration, we identified 3 areas where git-lfs itself could be improved to better support custom transfer agents and to handle local storage efficiently to avoid manual maintenance operations on client/workers computers.

### Benchmarks

We ran a comprehensive benchmark comparing 15 scenarios across direct SSH, Forgejo, and Gitea with various storage backends. The full experiment results, dataset descriptions, and methodology are available at: https://github.com/InSimo/lfs-experiments/

Key observations:
- Desync deduplication achieves **5.6x** storage reduction (4 GB vs 22.6 GB)
- The **pipelined client-side agent** is the fastest push at 4m48s, **26% faster** than the scutiger baseline (6m30s) and **10% faster** than the non-pipelined agent (5m19s), using less CPU and RAM
- Client-side desync clones transfer only ~3.8 GB of compressed chunks over the wire (vs 22.5 GB for standard LFS)

**System environment**

The experimental implementations have been tested on Linux (Ubuntu 24.04) using a local network of Docker containers. We intend to test on Windows as well before submitting our implementations. Help from others to test on macOS or other platforms would be welcome.

**Output of `git lfs env`**

N/A

**Additional context**

### Proposals

[...]

PR #6306, opened 2026-07-19
Proposal: bundled custom transfers RFC

bogomolets-owl:
Hey git lfs team,  hope you're doing great!

This is the proposal-only follow-up to the discussion in #6119, as we agreed
there to develop an RFC first and move to the implementation once the design
is settled.

This PR contains two documents and no code changes:

- `docs/proposals/custom-transfers-bundling.md` - the proposal itself:
  problem statement, concepts, configuration and compatibility considerations.
- `docs/proposals/custom-transfers-bundling-protocol.md` - the companion
  specification of the proposed custom transfer protocol version 2.

The proposal allows custom transfer agents to process a bundle of objects as
a single unit of work instead of a single object per transfer, which enables
transfer strategies like compressing a bundle into a single archive per
download/upload session. This addresses the performance limitations for
repositories with large numbers of small objects and for large well-compressible
objects.

Compared to the state discussed in #6119, the main changes are:

1. The terminology moved from "bulk"/"batch" to "bundles" to stop clashing
   with the Batch API terms. 
2. The Batch API extension is dropped entirely.
3. The protocol version negotiation on the init stage is included.
4. The bundle composition is defined as a Git LFS responsibility.
5. The optional `retry` flag for completion messages is included.

#6119 is left open as the draft implementation playground for now.

### DIFF
--- a/docs/proposals/custom-transfers-bundling-protocol.md
+++ b/docs/proposals/custom-transfers-bundling-protocol.md
@@ -1,0 +1,557 @@
+# Custom Transfer Protocol Version 2: Bundles (Proposed Specification)
+
+> **Status: proposal.** This document is a companion to the
+> [bundled custom transfers proposal](./custom-transfers-bundling.md) and
+> describes a proposed version 2 of the custom transfer protocol. Nothing in
+> this document is implemented yet.
+> The custom transfer protocol version 1 is specified in the
+> [custom transfers documentation](../custom-transfers.md) and remains
+> unchanged by this proposal. This document only specifies the changes
+> introduced by the version 2.
+
+## Overview
+
+Protocol version 2 extends the custom transfer protocol with:
+1. Protocol version negotiation on the initiation stage.
+2. The `bundles` transfer mode, where a bundle of objects is the unit of work
+for a custom transfer agent process, instead of a single object.
+3. An optional `retry` flag on the completion messages, allowing agents to
+report retriable errors without forcing Git LFS to fall back to the default
+transfer agent.
+
+The general communication mechanics are unchanged: the git-lfs client
+communicates with the custom transfer process via the stdin and stdout
+streams, no file content is communicated on these streams, and every message
+is a single line of JSON as per the version 1 specification.
+
+## Configuration
+
+The following new configuration values are proposed:
+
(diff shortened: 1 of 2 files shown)


chrisd8088:
Hey, thanks so much for this new proposal!

I confess I have not yet had time to do more than skim it quickly, but it certainly looks promising.

We should perhaps also consider the ideas outlined in #6242 regarding another way of enhancing the custom transfer protocol.  I hope that after the long-delayed v3.8.0 release is done, we'll have some time to review these together and see how they might be compatible, and where they might conflict.

Thanks again for separating out the proposal into its own PR; that will make the next steps much easier!

bogomolets-owl:
Hey @chrisd8088! 
Thank you for reply!

Regarding #6242 - yep, i agree, we definetly should consider this proposal.
#6306 and #6242 pretty much share common goals and values. 
#6242 proposes a simpler solution approach for #6306 goals and simpler seems better here.

One thing i want to outline specifically is error handling/retries which is explicitly described in #6306 and i have yet to examine it in #6242 closely. Based on my experience of using the current 3.7.1 version with custom transfer - the error handling is a pain point on day-to-day basis. 

On #6119 discussion you proposed to split different proposals (bundles delivery separate from error handling/retries behaviour) and now seems a good point to pivot the #6303 proposal to error handling approach only having #6242  in mind. It seems possible since the #6242 proposed pipelined mode can cover a lot the bundles has to offer not introducing a separate complicated procotol version.

One idea on bundles vs pipelined mode though: bundles are explicitly defined for agent, the pipelined mode defines only the concurrent items via concurrenttransfers. It effectively means there is no well defined way to say "okay, transfer queue has no objects left to offer and agent should proceed with whats left, we will not hit concurrenttransfers number of objects again". The agent has to guess that via time interval waiting and if nothing comes in - flush the actual transfers. 
This is what bundles define explicitly to avoid any guessing on agent side. 
In pipelined mode it may be worked around via addtional explicit "flush" protocol message which effectively mean that we have reached that moment so agent know that without guessing and assume the queue has something to offer untill explicitly said otherwise.

Everything above is absolutely discussable, just sharing my perspective.

PR #6325, opened 2026-08-12
config: normalize URL-scoped matches

### LINKED ISSUE #6112: Inconsistent git config parsing
**Describe the bug**

Due to maps being unordered in golang, git sometimes can choose wrong config as the so called "best candidate". ([here](https://github.com/git-lfs/git-lfs/blob/main/config/url_config.go#L83))

[Git docs explicitly lists in which order configs should be loaded](https://git-scm.com/docs/git-config#FILES):

> The files are read in the order given above, with last value found taking precedence over values read earlier. When multiple values are taken then all values of a key from all files will be used.

**To Reproduce**

Setup the repo:

```
$ git init $HOME/test_repo
$ cd $HOME/test_repo
$ git remote add origin https://example.com/repo
$ openssl req -x509 -newkey rsa:2048 -keyout key.pem -out cert.pem -days 365 -nodes -subj "/CN=localhost"
$ openssl req -x509 -newkey rsa:2048 -keyout key2.pem -out cert2.pem -days 365 -nodes -subj "/CN=localhost"
$ echo '*.lfs filter=lfs diff=lfs merge=lfs -text' > .gitattributes
$ cat <<EOF > lfsfile.lfs
oid sha256:96407fa5898f91a2770dfef30241c16c49e385c09cd4565e159cb37f66ac841b
size 12
EOF
$ git add .
$ git commit -m "initial"
```

Setup the configs (make sure to replace `${HOME}` with the actual value).

Append this to your /etc/gitconfig:
```
[http "https://example.com"]
    proxy =
    sslcert = ${HOME}/test_repo/cert2.pem
```

Append this to your ~/.gitconfig:
```
[http "https://example.com/"]
  sslCert = ${HOME}/test_repo/cert.pem
  sslKey = ${HOME}/test_repo/key.pem
```

Now inside the repo run this 

`GIT_CURL_VERBOSE=1 git lfs fetch`

couple of times and observe how it can behave in 2 different ways at random due to sometimes cert2.pem being chosen and sometimes cert.pem being chosen. It depends on which entry appears first in [this iteration of the loop over an unordered map](https://github.com/git-lfs/git-lfs/blob/main/config/url_config.go#L83).

It will either start to download the file after loading key and cert correctly, or fail with:
`api error: Error reading client cert/key: tls: private key does not match public key` if it loads cert2.pem and key.pem.

**Expected behavior**

The sslCert option from ~/.gitconfig should always take precedence and be chose in a stable way.

**System environment**

```
$ cat /etc/os-release
NAME="CentOS Stream"
VERSION="9"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="9"
PLATFORM_ID="platform:el9"
PRETTY_NAME="CentOS Stream 9"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:centos:centos:9"
HOME_URL="https://centos.org/"
BUG_REPORT_URL="https://issues.redhat.com/"
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux 9"
REDHAT_SUPPORT_PRODUCT_VERSION="CentOS Stream"
```

**Output of `git lfs env`**
```
git-lfs/3.6.1 (GitHub; linux amd64; go 1.24.4 (Red Hat 1.24.4-3.el9) X:strictfipsruntime)
git version 2.47.3

Endpoint=https://example.com/repo.git/info/lfs (auth=none)
LocalWorkingDir=/home/mzr/test_repo
LocalGitDir=/home/mzr/test_repo/.git
LocalGitStorageDir=/home/mzr/test_repo/.git
LocalMediaDir=/home/mzr/test_repo/.git/lfs/objects
LocalReferenceDirs=
TempDir=/home/mzr/test_repo/.git/lfs/tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneRemoteName=origin
LfsStorageDir=/home/mzr/test_repo/.git/lfs
AccessDownload=none
AccessUpload=none
GIT_CEILING_DIRECTORIES=/home
GIT_EXEC_PATH=/usr/libexec/git-core
GIT_TRACE2=/tmp/.tmpGVGBtl
GIT_TRACE2_BRIEF=1
GIT_TRACE2_PARENT_SID=20250912T155313.198307Z-Hb2d21a63-P003eed46
```

fzlzjerry:
## Summary

- normalize URL-scoped configuration before matching, including scheme and host casing, percent escapes, default and zero-padded ports, and dot path segments
- expose a source-ordered `Environment.SortedAll()` snapshot so equally specific canonical matches follow Git configuration order without secondary index lookups
- preserve Git LFS's `.git/info/lfs` path compatibility and cover both the reported trailing-slash case and canonical-path matching

Fixes #6112

## Testing

- `make test`
- `make PKGS=config test-race`
- `go vet ./config`
- `make -C t PROVE_EXTRA_ARGS=-v t-config-url-match.sh t-extra-header.sh t-clone.sh t-lock.sh t-fetch.sh`

### DIFF
--- a/config/delayed_environment.go
+++ b/config/delayed_environment.go
@@ -56,14 +56,10 @@ func (e *delayedEnvironment) All() map[string][]string {
 	return e.env.All()
 }
 
-// SourceIndex returns a configuration key's last position in source order when
-// available.
-func (e *delayedEnvironment) SourceIndex(key string) (int, bool) {
+// SortedAll returns all configuration entries in source order when available.
+func (e *delayedEnvironment) SortedAll() []EnvironmentEntry {
 	e.Load()
-	if env, ok := e.env.(orderedEnvironment); ok {
-		return env.SourceIndex(key)
-	}
-	return 0, false
+	return e.env.SortedAll()
 }
 
 // Load reads and parses the .gitconfig by calling ReadGitConfig. It
--- a/config/delayed_environment.go
+++ b/config/delayed_environment.go
@@ -56,6 +56,16 @@ func (e *delayedEnvironment) All() map[string][]string {
 	return e.env.All()
 }
 
+// SourceIndex returns a configuration key's last position in source order when
+// available.
+func (e *delayedEnvironment) SourceIndex(key string) (int, bool) {
+	e.Load()
+	if env, ok := e.env.(orderedEnvironment); ok {
+		return env.SourceIndex(key)
+	}
+	return 0, false
+}
+
 // Load reads and parses the .gitconfig by calling ReadGitConfig. It
 // also sets values on the configuration instance `g.config`.
 //
--- a/config/environment.go
+++ b/config/environment.go
@@ -55,11 +55,10 @@ type Environment interface {
 	// All returns a copy of all the key/value pairs for the current
 	// environment.
 	All() map[string][]string
-}
 
-type orderedEnvironment interface {
-	Environment
-	SourceIndex(key string) (int, bool)
+	// SortedAll returns all key/value pairs ordered by each key's final source
+	// occurrence, or nil when the underlying source does not retain ordering.
+	SortedAll() []EnvironmentEntry
 }
 
 type environment struct {
@@ -100,11 +99,11 @@ func (e *environment) All() map[string][]string {
 	return e.Fetcher.All()
 }
 
-func (e *environment) SourceIndex(key string) (int, bool) {
-	if fetcher, ok := e.Fetcher.(orderedFetcher); ok {
-		return fetcher.SourceIndex(key)
+func (e *environment) SortedAll() []EnvironmentEntry {
+	if fetcher, ok := e.Fetcher.(sortedFetcher); ok {
+		return fetcher.SortedAll()
 	}
-	return 0, false
+	return nil
 }
 
 // Int returns the int value associated with the given value, or the value
--- a/config/environment.go
+++ b/config/environment.go
@@ -57,6 +57,11 @@ type Environment interface {
 	All() map[string][]string
 }
 
+type orderedEnvironment interface {
+	Environment
+	SourceIndex(key string) (int, bool)
+}
+
 type environment struct {
 	// Fetcher is the `environment`'s source of data.
 	Fetcher Fetcher
@@ -93,6 +98,13 @@ func (e *environment) Int64(key string, def int64) int64 {
 
 func (e *environment) All() map[string][]string {
 	return e.Fetcher.All()
+}
+
+func (e *environment) SourceIndex(key string) (int, bool) {
+	if fetcher, ok := e.Fetcher.(orderedFetcher); ok {
+		return fetcher.SourceIndex(key)
+	}
+	return 0, false
 }
 
 // Int returns the int value associated with the given value, or the value
--- a/config/fetcher.go
+++ b/config/fetcher.go
@@ -20,3 +20,9 @@ type Fetcher interface {
 	// environment.
 	All() map[string][]string
 }
+
+// orderedFetcher provides access to configuration key positions in source order.
+// Fetchers which do not preserve source order need not implement it.
+type orderedFetcher interface {
+	SourceIndex(key string) (int, bool)
+}
--- a/config/fetcher.go
+++ b/config/fetcher.go
@@ -21,8 +21,15 @@ type Fetcher interface {
 	All() map[string][]string
 }
 
-// orderedFetcher provides access to configuration key positions in source order.
-// Fetchers which do not preserve source order need not implement it.
-type orderedFetcher interface {
-	SourceIndex(key string) (int, bool)
+// EnvironmentEntry contains all values associated with a configuration key.
+type EnvironmentEntry struct {
+	Key    string
+	Values []string
+}
+
+// sortedFetcher provides all configuration entries ordered by each key's final
+// source occurrence. Fetchers which do not preserve source order need not
+// implement it.
+type sortedFetcher interface {
+	SortedAll() []EnvironmentEntry
 }
--- a/config/git_fetcher.go
+++ b/config/git_fetcher.go
@@ -12,12 +12,15 @@ import (
 )
 
 type GitFetcher struct {
-	vmu  sync.RWMutex
-	vals map[string][]string
+	vmu   sync.RWMutex
+	vals  map[string][]string
+	order map[string]int
 }
 
 func readGitConfig(configs ...*git.ConfigurationSource) (gf *GitFetcher, extensions map[string]Extension, uniqRemotes map[string]bool) {
 	vals := make(map[string][]string)
+	order := make(map[string]int)
+	nextIndex := 0
 	ignored := make([]string, 0)
 
 	extensions = make(map[string]Extension)
@@ -97,6 +100,8 @@ func readGitConfig(configs ...*git.ConfigurationSource) (gf *GitFetcher, extensi
 				continue
 			}
 
+			order[key] = nextIndex
+			nextIndex++
 			vals[key] = append(vals[key], val)
 		}
 	}
@@ -108,7 +113,7 @@ func readGitConfig(configs ...*git.ConfigurationSource) (gf *GitFetcher, extensi
 		}
 	}
 
-	gf = &GitFetcher{vals: vals}
+	gf = &GitFetcher{vals: vals, order: order}
 
 	return
 }
@@ -151,6 +156,15 @@ func (g *GitFetcher) All() map[string][]string {
 	}
 
 	return newmap
+}
+
+// SourceIndex returns a configuration key's last position in source order.
+func (g *GitFetcher) SourceIndex(key string) (int, bool) {
+	g.vmu.RLock()
+	defer g.vmu.RUnlock()
+
+	index, ok := g.order[g.caseFoldKey(key)]
+	return index, ok
 }
 
 func (g *GitFetcher) caseFoldKey(key string) string {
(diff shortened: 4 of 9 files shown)


chrisd8088 (commented):
Hey, thanks for taking time to try to help fix the issue described in #6112!

While the changes in this PR might resolve the specific problem as it was reported, I think we will need to take a different, more comprehensive approach, which I alluded to in one of my [comments](https://github.com/git-lfs/git-lfs/issues/6112#issuecomment-3305689243) on that issue (emphasis added):

> We either have to sort keys in the exactly the same order as they appeared in the `git config -l` output, _or (maybe better) ensure that we convert any URLs in keys into canonical versions of themselves, following the same rules as Git does for this conversion_.

Let me outline my thoughts here, in order of increasing importance.

1. In instances like this one, where we are trying to resolve an issue that was reported to us, we usually try to include a shell integration test which replicates the original issue as it was reported and confirms that it is resolved.

    Here is a test that I drafted for the purpose of reproducing the specific issue from [#6112](https://github.com/git-lfs/git-lfs/issues/6112).  (It's not actually guaranteed to reproduce the issue every time, but it is very likely to, given the odds involved.)  I would recommend we place this test in a new `t/t-config-url-match.sh` script:

    <details>

    ```
    #!/usr/bin/env bash

    . "$(dirname "$0")/testlib.sh"

    begin_test "SSL certificate matches root path"
    (
      set -e

      if [ "$IS_WINDOWS" -eq 1 ]; then
        git config --global "http.sslBackend" "openssl"
      fi

      git config --global "http.$LFS_CLIENT_CERT_URL.sslCert" "$LFS_CLIENT_CERT_FILE"
      git config --global "http.$LFS_CLIENT_CERT_URL.sslKey" "$LFS_CLIENT_KEY_FILE"

      reponame="config-ssl-cert-match-root-path"
      setup_remote_repo "$reponame"
      clone_repo_clientcert "$reponame" "$reponame"

      git lfs track "*.bin"

      contents="test"
      printf "%s" "$contents" >a.bin

      git add .gitattributes a.bin
      git commit -m "initial commit"

      git push origin main

      # Test with the global "http.<url>.sslCert" option reset to an invalid
      # certificate file and a local "http.<url>/.sslCert" option set to the
      # valid file.  The global option should be ignored in favour of the local
      # one, even though the local option's URL has a trailing slash character.
      # See https://github.com/git-lfs/git-lfs/issues/6112.
      git config --global "http.$LFS_CLIENT_CERT_URL.sslCert" "$TRASHDIR/nonexistent/cert.pem"
      git config "http.$LFS_CLIENT_CERT_URL/.sslCert" "$LFS_CLIENT_CERT_FILE"

      # The reported issue occurred only half the time, on average, due
      # to Go's arbitrary map key ordering.  To increase the likelihood of
      # reproducing the problem, repeat the check up to eight times.
      for ((i=0; i<8; i++)); do
        rm -rf .git/lfs/objects

        git lfs pull

        assert_local_object "$(calc_oid "$contents")" "${#contents}"
      done
    )
    end_test
    ```
    </details>

2. In the current version of this PR, the `getAll()` method of our `URLConfig` structure would [ignore](https://github.com/git-lfs/git-lfs/blob/abee7887343245d6bf4a29b03ad347a6f9b59a60/config/url_config.go#L99-L101) the `bool` value returned by the `SourceIndex()` method of the proposed `orderedFetcher` interface:

    ```go
    		if ordered != nil {
    			match.sourceIndex, _ = ordered.SourceIndex(k)
    		}
    ```

    In the current implementation this always succeeds because the `readGitConfig()` function has always populated an index value for any value the `getAll()` method will pass to the `SourceIndex()` method.

    However, ignoring a sentinel return value like this is usually not good programming practice, as it opens up the possibility for a future change to silently introduce problems.  If for some reason we altered how the data was populated so that the `SourceIndex()` method legitimately began to return `false` values to some of these calls, the `getAll()` method would just ignore them.

3. Given that concern, if we wanted to pursue the approach I initially suggested in my [comment](https://github.com/git-lfs/git-lfs/issues/6112#issuecomment-3305689243) on [#6112](https://github.com/git-lfs/git-lfs/issues/6112), I think we would be best served by introducing an API for the `Environment` interface which would return all key/value pairs and the keys' sort order in a single data structure.  This would avoid the need to call the `All()` [method](https://github.com/git-lfs/git-lfs/blob/af13aa8f259f958049d3703252068e125b412d5e/config/environment.go#L55-L57), get an unsorted set of key/value pairs, and then make repeated secondary calls to retrieve the keys' indexes.

    For instance, a hypothetical `SortedAll()` method might return a slice where each element was a structure with a key and a set of values, or else just return `nil` if the underlying data was unsorted or not available, as the `OsFetcher` structure's `All()` method already [does](https://github.com/git-lfs/git-lfs/blob/af13aa8f259f958049d3703252068e125b412d5e/config/os_fetcher.go#L66-L68) now in the latter case.

4. However, upon investigation I now think there are more issues here than can be solved by just respecting the sort order of the configuration options returned from a `git config -l` command.  That was my hunch when I replied to [#6112](https://github.com/git-lfs/git-lfs/issues/6112), but having written some shell integration tests to experiment, I can confirm it's the case.

    Here's a test which illustrates that Git matches URL paths correctly when one of them contains a spurious `/./` path segment and the other does not, but Git LFS fails to do so (even with the changes from this PR):

    <details>

    ```
    begin_test "SSL certificate matches canonical path"
    (
      set -e

      if [ "$IS_WINDOWS" -eq 1 ]; then
        git config --global "http.sslBackend" "openssl"
      fi

      git config --global "http.$LFS_CLIENT_CERT_URL.sslCert" "$LFS_CLIENT_CERT_FILE"
      git config --global "http.$LFS_CLIENT_CERT_URL.sslKey" "$LFS_CLIENT_KEY_FILE"

      reponame="config-ssl-cert-match-canonical-path"
      setup_remote_repo "$reponame"
      clone_repo_clientcert "$reponame" "$reponame"

      git lfs track "*.bin"

      contents="test"
      printf "%s" "$contents" >a.bin

      git add .gitattributes a.bin
      git commit -m "initial commit"

      git push origin main

      # Test with the global "http.<url>.sslCert" option reset to an invalid
      # certificate file and a local "http.<url>/./.sslCert" option set to the
      # valid file.  The global option should be ignored in favour of the local
      # one, even though the local option's URL has a spurious path segment.
      git config --global "http.$LFS_CLIENT_CERT_URL.sslCert" "$TRASHDIR/nonexistent/cert.pem"
      git config "http.$LFS_CLIENT_CERT_URL/./.sslCert" "$LFS_CLIENT_CERT_FILE"

      # First push a new commit that does not include any Git LFS objects to
      # confirm Git matches the local option's URL to the remote, regardless
      # of the spurious path segment.
      echo "test" >a.txt
      git add a.txt
      git commit -m "second commit"

      git push origin main

      # Next confirm Git LFS also matches the local option's URL to the remote,
      # regardless of the spurious path segment.
      rm -rf .git/lfs/objects

      git lfs pull

      assert_local_object "$(calc_oid "$contents")" "${#contents}"
    )
    end_test
    ```
    </details>

    Git [normalizes](https://github.com/git/git/blob/v2.55.0/urlmatch.c#L707) [both](https://github.com/git/git/blob/v2.55.0/http.c#L1401) of the URLs it passes to its `url_match()` [function](https://github.com/git/git/blob/v2.55.0/http.c#L1401) using a fairly complex `url_normalize_1()` [function](https://github.com/git/git/blob/v2.55.0/urlmatch.c#L115-L437) (and `url_normalize()` wrapper function).

    Git LFS relies on Go's URL handling to do some of this for us, but it's not sufficient.  Because we fail to normalize the URLs, when our `comparePaths()` function compares `/` and `/./`, it [finds](https://github.com/git-lfs/git-lfs/blob/af13aa8f259f958049d3703252068e125b412d5e/config/url_config.go#L222-L224) they have different numbers of path segments, and deems them different.

    There are of course many other ways to produce URLs which are logically equivalent but appear different unless they are normalized.  We may be able to make use of the `Clean()` function in the Go standard library's `path` package, but however we implement normalization for URL matching, we should endeavour to match Git's behaviour exactly, including any case-folding and handling of URL-encoded characters.

    Conveniently, the Git project has a set of [unit tests](https://github.com/git/git/blob/v2.55.0/t/unit-tests/u-urlmatch-normalization.c) and [integration tests](https://github.com/git/git/blob/v2.55.0/t/t1300-config.sh#L1874-L2058) which we can replicate for our Go and shell test suites to help verify any changes we make align with Git's implementation.

So, to summarize, we unfortunately need to address the larger issue of URL normalization in our handling of `http.<url>.*` configuration options, rather than simply retaining the order of the options as they are output by the `git config -l` command and trying to respect that ordering.  This is why we haven't tackled the issue reported in #6112 yet, as it's a bit more complicated than it appears on the surface.

Thank you again for taking the time to help improve Git LFS, we really appreciate the effort!

fzlzjerry:
Thanks for the detailed review. I reworked the implementation in 7178e057:

- added both certificate regression cases to a new `t/t-config-url-match.sh` integration test;
- replaced the per-key `SourceIndex()` lookup with a single `Environment.SortedAll()` snapshot;
- normalize both requested and configured URLs before scoring, covering scheme/host casing, percent escapes, zero-padded/default ports, and dot-segment resolution; and
- added unit coverage based on the Git normalization cases, while retaining the existing Git LFS `.git/info/lfs` matching behavior.

The final head passes `make test`, `make PKGS=config test-race`, `go vet ./config`, and the URL-related integration group (`t-config-url-match.sh`, `t-extra-header.sh`, `t-clone.sh`, `t-lock.sh`, and `t-fetch.sh`).

PR #6328, opened 2026-08-14
subprocess: use Go standard executable lookup

### LINKED ISSUE #5612: Revisit use of Go's standard library to locate Git executable
In our remediation of the security issues CVE-2020-27955, CVE-2021-21237, and CVE-2022-24826, we addressed a series of problems stemming from the fact that on Windows the Go language's `os/exec` package would search for executable programs in the current working directory before searching other directories specified by the `PATH` environment variable, and so Git LFS could be convinced to run a malicious executable when it attempted to run Git.

As of Go 1.19, the `os/exec` package in the Go standard library will no longer execute programs found in the current working directory, and in fact will report a new `ErrDot` [error](https://github.com/golang/go/blob/dcbe77246922fe7ef41f07df228f47a37803f360/src/os/exec/exec.go#L1296-L1303) if it is passed a file name to execute and Go discovers an executable in the current working directory which matches that file name, as [described](https://github.com/golang/go/blob/dcbe77246922fe7ef41f07df228f47a37803f360/src/os/exec/exec.go#L24-L90) in the code comments.  The full explanation for this change is discussed in an accompanying blog article:

https://go.dev/blog/path-security

Our current approach uses a version of the internal `lookPath()` functions from the `os/exec` package which we  have merged and [imported](https://github.com/git-lfs/git-lfs/blob/2db011fb5e5685e8bf150c5dafa50ecae067462c/subprocess/path.go#L39-L70) and then [call](https://github.com/git-lfs/git-lfs/blob/2db011fb5e5685e8bf150c5dafa50ecae067462c/subprocess/subprocess_windows.go#L13-L15) to set the `Path` member of an `exec.Cmd` structure after we have created it.  This ensures Go will execute the program we have located as opposed to whatever it first locates when the `Cmd` structure is initialized.

However, as noted in PR #5611, as of Go 1.19 if the executable file name matches a file in the current working directory then an `ErrDot` error is thrown regardless of whether we later change the `Path` structure member or not.

Also, the [current](https://github.com/golang/go/blob/1d45a7ef560a76318ed59dfdb178cecd58caf948/src/os/exec/lp_unix.go#L44-L82) [versions](https://github.com/golang/go/blob/1d45a7ef560a76318ed59dfdb178cecd58caf948/src/os/exec/lp_windows.go#L134-L212) of the `lookPath()` function for Unix and Windows have diverged from [older](https://github.com/golang/go/blob/7bb721b9384bdd196befeaed593b185f7f2a5589/src/os/exec/lp_unix.go#L31-L59) [ones](https://github.com/golang/go/blob/7bb721b9384bdd196befeaed593b185f7f2a5589/src/os/exec/lp_windows.go#L54-L94) we originally imported into this project and modified.  This means we are not benefitting from any of the subsequent enhancements or bug fixes in this code since we "forked" it, not just the introduction of the new `ErrDot` checks but any others as well.

We should determine if we are comfortable reverting to use of the `os/exec` package to search for executables, and if so, make that change.  We do, however, want to retain all of our existing tests which validate that Git LFS will not run a Git binary or any other executable if it appears in the current working directory.

/cc #5611

m0g3r:
## Summary

- resolve executable names with `exec.LookPath` before constructing `exec.Cmd`
- remove the stale copied lookup implementation and update the Go-derived license notice
- retain the current-directory executable-hijacking coverage without setting `GODEBUG=execerrdot=0`

This adopts the standard library behavior added in Go 1.19 while preserving the synchronous lookup-error contract and original argument zero.

Fixes #5612.

## Testing

- full configured Go test suite on Go 1.26.6
- `go test ./subprocess` on Go 1.25.13 and Go 1.26.6
- `make -C t t-path.sh` with all three path-security cases passing
- Windows amd64 test-binary cross-compilation on Go 1.25.13 and Go 1.26.6
- SSH, credential, custom-transfer, and happy-path integration suites
- Go vet, goimports, `bash -n`, and `git diff --check`

Windows execution was not available locally, so upstream CI remains the final native-platform verification.

## Assistance

AI assistance was used during implementation and test analysis.


t/t-path.sh
```diff
@@ -33,56 +33,21 @@ begin_test "does not look in current directory for git with credential helper"
...
-  # before trying to execute the program.
-  #
-  # Since we explicitly test Git LFS's avoidance of programs in the current
-  # working directory using the "git-lfs pull" command further below, here
-  # we just want "git add" to succeed, and so for the time being we disable
-  # Go's new security checks for this command only.  We will revisit this
-  # when we address the larger issue of re-adopting Go's own logic for
-  # locating executable programs.
+  # We add our malicious Git to the index before creating any attributes which
+  # select LFS filters, then remove it so it is not found before our key test.
```
chrisd8088: While this comment is correct, it's a bit terse, given the overall complexity of the test.  I'd rather retain the relevant portion of the original comment:

```suggestion
  # We add our malicious Git to the index and then remove it from the
  # work tree so it is not found early, before we perform our key test.
  # Specifically, our "git push" below will run git-lfs, which then runs
  # "git credential", so if we are looking for Git in the current directory
  # first when running a credential helper, we will fail at that point
  # because our malicious Git will be found first.
  #
  # We prefer to check for this behavior during our "git-lfs pull" further
  # below when we are populating LFS objects into a clone of this repo
  # (which contains the malicious Git), so for now we remove the malicious
  # Git as soon as possible.
```

Issue #6131, opened 2025-10-28
Disallow `git lfs checkout` command in bare repositories


chrisd8088:
The `git lfs checkout` command has no purpose when run in a bare repository, and performs no actions in this case, but nevertheless returns a zero exit code, which is intended to indicate that a command has succeeded in its operation.

As of our v3.7.1 release, the `git lfs checkout` command now outputs the warning message "This operation must be run in a work tree" when the command is run in a bare repository.

However, as described in commit 5327ac2dca9b50f044b43d73cc2d3a98240df78f of the v3.7.1 release and the corresponding commit d02bd13f02ef76f6807581cd6b34709069cb3615 in our `main` development branch, the command still returns a zero (i.e., successful) exit code when run in a bare repository.

We made this decision so as to avoid introducing unexpected failures for any users with CI jobs or other scripts which (perhaps inadvertently) invoke `git lfs checkout` in bare repositories and except the command to succeed.

In the future, though, we would prefer for the command to return a non-zero exit code when run in a bare repository, in the same manner as commands such as `git lfs track`.

We could accomplish this change simply by revising the command's implementation to call `setupWorkingCopy()` instead of [calling](https://github.com/git-lfs/git-lfs/blob/main/commands/command_checkout.go#L27) `setupRepository()`, and we could then also remove the subsequent conditional [block](https://github.com/git-lfs/git-lfs/blob/main/commands/command_checkout.go#L23-L36) where we explicitly check if the current repository is bare and return a zero exit code if that is the case.

This issue is therefore our public [notice](https://github.com/git-lfs/git-lfs/blob/main/commands/command_checkout.go#L29-L32) that we expect to make this change with our next major release, where we permit ourselves to break from our usual goal of full backwards compatibility.

Issue #6004, opened 2025-03-04
git lfs pull not working correctly with partial checkout in 3.6.1


tetreaultmarc:
**Describe the bug**
When using partial checkout with `git lfs pull --include`, LFS files are not being downloaded properly in version 3.6.1. Instead, only pointer files are being pulled.

**To Reproduce**
I created [a repo](https://github.com/tetreaultmarc/git-lfs-bug-demo) you can use to reproduce the issue.
Simply run these commands in a new directory with version `3.6.1` (or `3.6.0`).
```
git init .
git remote add origin git@github.com:tetreaultmarc/git-lfs-bug-demo.git
git fetch origin master
git update-ref HEAD $(git rev-parse origin/master)
git checkout $(git rev-parse origin/master) -- test-path
git lfs pull --include test-path
```

**Expected behavior**
The PNG file should be downloaded and replaced with its actual contents (as it works in git-lfs `3.5.1`)

**Actual behavior**
Only the LFS pointer file is pulled, not the actual file contents

**System environment**
git-lfs version `3.6.1`
Previously working in git-lfs `3.5.1`
OS: Windows 11 Pro 10.0.26100 Build 26100
Git version: 2.48.1.windows.1

**Output of `git lfs env`**
```
git-lfs/3.6.1 (GitHub; windows amd64; go 1.23.3; git ea47a34b)
git version 2.48.1.windows.1

Endpoint=https://github.com/tetreaultmarc/git-lfs-bug-demo.git/info/lfs (auth=none)
  SSH=git@github.com:tetreaultmarc/git-lfs-bug-demo.git
LocalWorkingDir=C:\git\git-lfs-testing
LocalGitDir=C:\git\git-lfs-testing\.git
LocalGitStorageDir=C:\git\git-lfs-testing\.git
LocalMediaDir=C:\git\git-lfs-testing\.git\lfs\objects
LocalReferenceDirs=
TempDir=C:\git\git-lfs-testing\.git\lfs\tmp
ConcurrentTransfers=8
TusTransfers=false
BasicTransfersOnly=false
SkipDownloadErrors=false
FetchRecentAlways=false
FetchRecentRefsDays=7
FetchRecentCommitsDays=0
FetchRecentRefsIncludeRemotes=true
PruneOffsetDays=3
PruneVerifyRemoteAlways=false
PruneVerifyUnreachableAlways=false
PruneRemoteName=origin
[...]

PR #6329, opened 2026-08-15
Return a non-zero exit code when checkout fails

g-chauvel:
Commit 3d8f497ea50ef550d15907bcc91868e608e56c14 of PR #5629 added the "read-only directory" tests, which check that `git lfs checkout` and `git lfs pull` report errors when a Git LFS object's contents cannot be written to a file in the working tree, and noted in its commit message:

> Note that for historical reasons we expect that under these conditions
> the commands will return a zero exit code, indicating success, rather
> than a non-zero exit code, which would indicate failure.  We may adjust
> this behaviour in the future, however, so that the commands also return
> a non-zero exit code in these cases.

This PR makes that adjustment: both commands now count the files they could not check out and exit with a status of 2 when the count is non-zero, matching the download stage of `git lfs pull`. The zero exit status makes checkout failures invisible to callers such as CI jobs. This was raised several times in #4974:

* [oderby, 2022](https://github.com/git-lfs/git-lfs/issues/4974#issuecomment-1122743446): "caused us to miss this in CI for quite some time"
* [jwatte, 2024](https://github.com/git-lfs/git-lfs/issues/4974#issuecomment-2305872017): "The command also exits with status 0, so we don't detect that it failed easily."
* [ravenAtSafe, 2025](https://github.com/git-lfs/git-lfs/issues/4974#issuecomment-3181785718): "We run `git lfs pull` in a Linux CI environment, it returns without indicating any error"

A green GitLab CI pipeline with a missing file was recently traced back to it (https://gitlab.com/gitlab-org/gitlab-runner/-/work_items/39678).

Working tree conflicts, such as a file found where a directory is expected, are also counted as checkout failures. Commit 0cffe93176b870055c9dadbb3cc9a4a440e98396, which added these conflict checks, notes:

> That function returns an error if any of the
> directories in the given path do not already exist and cannot be created,
> and the "git lfs checkout" and "git lfs pull" commands would just report
> that error rather than attempt to resolve it by removing anything.

### DIFF
--- a/commands/pull.go
+++ b/commands/pull.go
@@ -48,9 +48,9 @@ func newSingleCheckout(gitEnv config.Environment, remote string) abstractCheckou
 type abstractCheckout interface {
 	Manifest() tq.Manifest
 	Skip() bool
-	Run(*lfs.WrappedPointer)
+	Run(*lfs.WrappedPointer) error
 	RunToPath(*lfs.WrappedPointer, string) error
-	Close()
+	Close() error
 }
 
 type singleCheckout struct {
@@ -71,9 +71,9 @@ func (c *singleCheckout) Skip() bool {
 	return false
 }
 
-func (c *singleCheckout) Run(p *lfs.WrappedPointer) {
+func (c *singleCheckout) Run(p *lfs.WrappedPointer) error {
 	if !c.hasWorkTree {
-		return
+		return nil
 	}
 
 	dirWalker := tools.NewDirWalkerForFile("", p.Name, cfg)
@@ -83,7 +83,7 @@ func (c *singleCheckout) Run(p *lfs.WrappedPointer) {
 	if err != nil {
 		if !os.IsNotExist(err) {
 			LoggedError(err, tr.Tr.Get("Checkout error trying to check path for %q: %s", p.Name, err))
-			return
+			return err
 		}
 	} else {
 		// Check the content - either missing or still this pointer (not exist is ok)
@@ -95,34 +95,34 @@ func (c *singleCheckout) Run(p *lfs.WrappedPointer) {
 			output, err := git.DiffIndexWithPaths("HEAD", true, []string{p.Name})
 			if err != nil {
 				LoggedError(err, tr.Tr.Get("Checkout error trying to run diff-index: %s", err))
-				return
+				return err
 			}
 			if strings.HasPrefix(output, ":100644 000000 ") || strings.HasPrefix(output, ":100755 000000 ") {
 				// This file is deleted in the index.  Don't try
 				// to check it out.
-				return
+				return nil
 			}
 		} else {
 			if errors.IsNotAPointerError(err) || errors.IsBadPointerKeyError(err) {
 				// File has non-pointer content, leave it alone
-				return
+				return nil
 			}
 
 			LoggedError(err, tr.Tr.Get("Checkout error for %q: %s", p.Name, err))
-			return
+			return err
 		}
 	}
 
 	if filepointer != nil && filepointer.Oid != p.Oid {
 		// User has probably manually reset a file to another commit
 		// while leaving it a pointer; don't mess with this
-		return
+		return nil
 	}
 
 	if err != nil && os.IsNotExist(err) {
 		if err := dirWalker.WalkAndCreate(); err != nil {
 			LoggedError(err, tr.Tr.Get("Checkout error trying to create path for %q: %s", p.Name, err))
-			return
+			return err
 		}
 	}
 
@@ -130,16 +130,20 @@ func (c *singleCheckout) Run(p *lfs.WrappedPointer) {
 		if errors.IsDownloadDeclinedError(err) {
 			// acceptable error, data not local (fetch not run or include/exclude)
 			Error(tr.Tr.Get("Skipped checkout for %q, content not local. Use fetch to download.", p.Name))
-		} else {
-			FullError(errors.Wrap(err, tr.Tr.Get("could not check out %q", p.Name)))
+			return nil
 		}
-		return
+
+		err = errors.Wrap(err, tr.Tr.Get("could not check out %q", p.Name))
+		FullError(err)
+		return err
 	}
 
 	// errors are only returned when the gitIndexer is starting a new cmd
 	if err := c.gitIndexer.Add(p.Name); err != nil {
 		Panic(err, tr.Tr.Get("Could not update the index"))
 	}
+
+	return nil
 }
 
 // RunToPath checks out the pointer specified by p to the given path.  It does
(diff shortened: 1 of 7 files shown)


commands/pull.go
```diff
@@ -50,16 +51,27 @@ type abstractCheckout interface {
...
 
+func (c *singleCheckout) fail() {
+	atomic.AddInt64(&c.failures, 1)
+}
+
+// Failures reports the number of files which could not be checked out.
+func (c *singleCheckout) Failures() int64 {
+	return atomic.LoadInt64(&c.failures)
+}
+
```
chrisd8088: I appreciate that this design minimizes the changes needed to the existing callers of the `Run()` method.

However, I think it introduces some fragility into the `Run()` method itself, because we now have to remember to invoke the `fail()` method after logging an error and before returning, e.g.:

```go
			LoggedError(err, tr.Tr.Get("Checkout error for %q: %s", p.Name, err))
			c.fail()
			return
```

If we update the `Run()` method in the future to handle some new, additional error condition, we might overlook the need to call the `fail()` method before returning.  (For that matter, we might also forget to call a function like `LoggedError()` to log the error, which I think points to an existing design problem in how we handle errors in the `Run()` method and its callers.)

Conversely, it's also possible to accidentally call the `fail()` method in instances where we shouldn't.  (Per my other comment below, I think that's already the case in the current version of this PR, where the `fail()` method is called from the `Close()` method.)

---

Rather than adding `fail()` and `Failures()` methods, we might instead consider refactoring the `Run()` method to return an `error` value.

To start with, where we now call the `LoggedError()` or `FullError()` functions we would now also return the `error` value.  For example, in place of the code above or the existing [code](https://github.com/git-lfs/git-lfs/blob/af13aa8f259f958049d3703252068e125b412d5e/commands/pull.go#L111-L112), we would have:

```go
			LoggedError(err, tr.Tr.Get("Checkout error for %q: %s", p.Name, err))
			return err
```

Callers would then follow the standard Go error-handling idiom of checking the `err` value and handling it appropriately.

A naive approach might be to skip counting errors and have each of the two commands just keep track of whether at least one invocation of the `Run()` method returned an error.  After processing all the Git LFS pointers, if any errors were encountered, the commands would use the `ExitWithCode()` function to return a non-zero exit status.  That wouldn't produce a nice summary of the number of errors, but users will still see all the individual per-object error messages.

If a summary message is really important, the two commands could count the non-nil `error` values returned by the `Run()` method.  In the case of the `git lfs pull` command, which calls the `Run()` method [inside](https://github.com/git-lfs/git-lfs/blob/af13aa8f259f958049d3703252068e125b412d5e/commands/command_pull.go#L85) a goroutine, that might involve an error channel or some other mechanism to ensure a correct count.

Personally, though, I think this PR could just focus on ensuring the exit status is non-zero when any error has occurred, and leave further enhancements to another PR.

I'm inclined toward changing the `Run()` method to return an `error` value for a couple of overlapping reasons.  First, it's just a more conventional way for Go methods and functions to operate.

Second, as a longer-term project, it would be ideal to refactor the `Run()` and `Close()` methods so that they didn't make a variety of separate calls to the `LoggedError()` and `FullError()` functions, but just returned `error` values and deferred to the caller to collect those and report them appropriately.

Unfortunately, the `LoggedError()` and `FullError()` functions have some non-obvious legacy behaviour in terms of how and when they generate error messages and log files, so making sure that equivalent error handling is re-implemented outside of the `Run()` method is going to be a fiddly business and one that I think is best left to a separate PR.

commands/pull.go
```diff
@@ -152,6 +169,7 @@ func (c *singleCheckout) RunToPath(p *lfs.WrappedPointer, path string) error {
 func (c *singleCheckout) Close() {
 	if err := c.gitIndexer.Close(); err != nil {
 		LoggedError(err, "%s\n%s", tr.Tr.Get("Error updating the Git index:"), c.gitIndexer.Output())
+		c.fail()
```
chrisd8088: In my comment above, I outlined why I'm a bit uncomfortable with the introduction of a `fail()` method at all, but even if we were to adopt this approach, I don't think we want to call the `fail()` method here.

The `Close()` method is [only](https://github.com/git-lfs/git-lfs/blob/af13aa8f259f958049d3703252068e125b412d5e/commands/command_checkout.go#L105) [invoked](https://github.com/git-lfs/git-lfs/blob/af13aa8f259f958049d3703252068e125b412d5e/commands/command_pull.go#L93) [once](https://github.com/git-lfs/git-lfs/blob/af13aa8f259f958049d3703252068e125b412d5e/commands/command_pull.go#L93) for the whole `singleCheckout` structure, after all the individual Git LFS pointers have been passed to the `Run()` method.

If an error occurred here, and we called the `fail()` method, that would increment the count of the number of failures.  But in the current version of this PR, the count returned by the `Failures()` method is reported in a new error message.  So if exactly one `Run()` method called the `fail()` method, but then the `Close()` method also did so, the error message would be `Failed to check out 2 objects`.  That seems wrong, as any problem terminating the `git update-index` command might affect an unknown number of objects, not just one extra one.

chrisd8088 (commented):
Hey, thanks very much for looking at this issue and drafting this PR to address it!

First of all, I think we need to provide ample advance notice before we change the exit status returned by the `git lfs checkout` and `git lfs pull` commands in response to error conditions, even if it is a strict improvement to the commands' behaviour.

As Hyrum's Law [asserts](https://www.hyrumslaw.com/), we can be fairly confident that a number of existing workflows and CI systems rely implicitly on the existing behaviour.

When we revised our internal implementation of certain commands like `git lfs pull` in PR #5699, we didn't foresee that a number of users were effectively relying on the existing behaviour, as was subsequently reported in #6004 and several other issues.  This experience is why we've held off changing the `git lfs checkout` command to return a non-zero exit code when it is run in a bare repository; we'll make that change in v4.0.0 instead, per #6131.

For the same reason, for now at least I'm going to tag this PR as one for us to consider in a future v4.0.0 release.

To be clear, though, I agree that the `git lfs checkout` and `git lfs pull` commands should return a non-zero exit code after encountering errors, and I acknowledge that the existing behaviour causes trouble for some users.

With that goal in mind, I did leave a couple of comments on the specific implementation proposed in this PR.  The gist of my suggestions is that I think we ought to consider changing the `Run()` method to return an `error` value, then revising the callers set a local flag variable to `true` if any instances of the `Run()` method encountered an error.  If the flag is `true` at the end of a command's activity, it can then return a non-zero exit status.

Let us know your thoughts on that approach, and thanks again very much for taking the time to try to address this issue!

commands/pull.go
g-chauvel: The counter was modelled on the existing atomic counters in tq/meter.go.

Agreed.

commands/pull.go
g-chauvel: Yes, the count was misleading here.

PR #6337, opened 2026-08-27
Fix typo in git-lfs-prune man page

ssgelm:
Super simple typo fix on the `git-lfs-prune` man page that was flagged by Debian tooling. `exceution` -> `execution`


chrisd8088 (approved):
Thanks very much for the correction!

I'll merge this PR soon as soon as I get the fixes for the issues I bumped into during the recent release merged first.  (To be clear, though, that's only for the sake of my own future reference and not an actual dependency of any kind between the two PRs.)

